PR #98, opened 2016-08-31
Added ownCloud chart

juan131:


migmartri:
This one seems to work for the most part if you set a `owncloudLoadBalancerIP` 

I would not merged it though until we decide what to do with this option. Either make it mandatory or improve the documentation.

PR #99, opened 2016-08-31
Added Magento chart

juan131:


magento/values.yaml
```diff
@@ -0,0 +1,64 @@
+## Bitnami Magento image version
+## ref: https://hub.docker.com/r/bitnami/magento/tags/
+##
+imageTag: 2.0.5-r1
```
migmartri: This tag does not seem to exist in https://hub.docker.com/r/bitnami/magento/tags/

magento/templates/deployment.yaml
```diff
@@ -0,0 +1,83 @@
...
+          value: {{ template "mariadb.fullname" . }}
+        - name: MARIADB_PORT
+          value: "3306"
+        - name: MARIADB_PASSWORD
+          valueFrom:
+            secretKeyRef:
+              name: {{ template "mariadb.fullname" . }}
+              key: mariadb-root-password
+        - name: MAGENTO_HOST
+          value: {{ default .Values.magentoLoadBalancerIP .Values.magentoHost | quote }}
```
migmartri: The linter returns the following linter warnings.

We need to think what to do in these cases.

```
helm lint .
==> Linting .
[WARNING] templates/deployment.yaml: these substitution functions are returning no value: [{{ default .Values.magentoLoadBalancerIP .Values.magentoHost | quote }}]
[WARNING] templates/svc.yaml: these substitution functions are returning no value: [{{ default "" .Values.magentoLoadBalancerIP }}]

1 chart(s) linted, no failures
```

migmartri:
I added some inline comments. 

It crashed to me with the following error:

I am using minikube cluster and default values in the charts.

```
kubectl logs -f fair-penguin-magento-3227626210-kq0ks
harpoon INFO  Initializing apache
harpoon INFO  apache successfully initialized
harpoon INFO  Initializing magento
magento INFO  Preparing webserver environment...
magento INFO  Preparing PHP environment...
Error executing 'postInstallation': File '/opt/bitnami/php/conf/php.ini' does not exists
```

magento/values.yaml
migmartri: For some reason the image 2.0.5-r1 was not built. I have triggered the build and built it in any case.

I am still getting the same error:

```
kubectl logs -f hardy-guppy-magento-3000609057-17wap
harpoon INFO  Initializing apache
harpoon INFO  apache successfully initialized
harpoon INFO  Initializing magento
magento INFO  Preparing webserver environment...
magento INFO  Preparing PHP environment...
Error executing 'postInstallation': File '/opt/bitnami/php/conf/php.ini' does not exists
```

sameersbn: > For some reason the image 2.0.5-r1 was not built. I have triggered the build and built it in any case.

I've seen this happen before. For some reason we need to manually register the `Docker` as a hook in the Github repo (or delete the DH repo and recreate it).

magento/templates/deployment.yaml
sameersbn: > Does Magento work if we do not add magentoLoadBalancerIP ? I mean, is it usable?

My tests reveal that specifying `magentoLoadBalancerIP` does indeed result in a working deployment (needed to make some unrelated changes though).

> If it is not the case we should think of making it mandatory.

For certain apps (magento, owncloud, ghost, etc.) it's mandatory to specify the FQDN or external IP address of the service (a.k.a `loadBalancerIP` in case of GKE).

The `magentoLoadBalancerIP` is a way to provide an IP address to associate to the service. This feature is cloud/platform specific and hence cannot be called mandatory IMO. For other clouds users would need to configure and provide the correct hostname via the `magentoHost` parameter at the time of deployment.

sameersbn:
>```
> kubectl logs -f fair-penguin-magento-3227626210-kq0ks
> harpoon INFO  Initializing apache
> harpoon INFO  apache successfully initialized
> harpoon INFO  Initializing magento
> magento INFO  Preparing webserver environment...
> magento INFO  Preparing PHP environment...
> Error executing 'postInstallation': File '/opt/bitnami/php/conf/php.ini' does not exists
>```

Fixed in https://github.com/bitnami/bitnami-docker-magento/pull/6 but I haven't released a new tag yet. If your interested in testing, you can try the following command after removing the liveness and readiness probes from  the `deployment.yaml` file.

```bash
helm install magento/ --set imageTag=development,magentoLoadBalancerIP=x.x.x.x
```

magento/charts/mariadb/templates/deployment.yaml
```diff
@@ -0,0 +1,60 @@
...
+          valueFrom:
+            secretKeyRef:
+              name: {{ template "fullname" . }}
+              key: mariadb-password
+        - name: MARIADB_DATABASE
+          value: {{ default "" .Values.mariadbDatabase | quote }}
+        ports:
+        - name: mysql
+          containerPort: 3306
+        livenessProbe:
```
sameersbn: @juan131 The liveness and readiness probes fail because the application should only be access using the particular ip or hostname. We had a similar issue with the phabricator chart and resolved it by checking the status of the `nami` module instead, please see https://github.com/bitnami/charts/blob/master/phabricator/templates/deployment.yaml#L67-L82

PR #100, opened 2016-08-31
Added PrestaShop chart

juan131:


prestashop/values.yaml
```diff
@@ -0,0 +1,63 @@
+## Bitnami PrestaShop image version
+## ref: https://hub.docker.com/r/bitnami/prestashop/tags/
+##
+imageTag: 1.6.1.6-r1
```
migmartri: This image does not exist.

```
harping-hog-prestashop-456225883-8ru2o   0/1       ErrImagePull   0          6s
```

migmartri:
Same issues than Magento in terms of the linter errors.

Is the existence of prestashopHost or prestashopLoadBalancerIP mandatory? In that case, it should not be commented out and probably now even let you deploy it. @sameersbn what's your take on this?

```
helm lint .
==> Linting .
[WARNING] templates/deployment.yaml: these substitution functions are returning no value: [{{ default .Values.prestashopLoadBalancerIP .Values.prestashopHost | quote }}]
[WARNING] templates/svc.yaml: these substitution functions are returning no value: [{{ default "" .Values.prestashopLoadBalancerIP }}]
```

prestashop/values.yaml
prydonius: I triggered the build and now it exists.

prydonius:
I'm getting the same issue as @migmartri, though it works in Docker Compose.

prestashop/values.yaml
sameersbn: @prydonius can you please check if the Docker service hook is setup in Github.

sameersbn:
> Same issues than Magento in terms of the linter errors.
> [WARNING] templates/deployment.yaml: these substitution functions are returning no value: [{{ default .Values.prestashopLoadBalancerIP .Values.prestashopHost | quote }}]

Not sure how to silence this error.

> [WARNING] templates/svc.yaml: these substitution functions are returning no value: [{{ default "" .Values.prestashopLoadBalancerIP }}]

This one can be fixed with `{{ default "" .Values.prestashopLoadBalancerIP | quote }}`, but I can't remember why i didn't do it for the other charts. I'll run some tests to make sure if its safe to make this change.

> Is the existence of prestashopHost or prestashopLoadBalancerIP mandatory? 

see my comment https://github.com/bitnami/charts/pull/99#discussion_r77354621

To reiterate, the `prestashopHost` param is mandatory.

> In that case, it should not be commented out and probably now even let you deploy it

I thought I was highlighting that part of the text :laughing: 

> Same error than Magento. @prydonius could you try and let me know?

Similar [fix](https://github.com/bitnami/bitnami-docker-magento/pull/6) needs to be implemented in the prestashop image, or we need to stop mounting `/bitnami/php` in the chart  and docker-compose files.

PR #128, opened 2016-09-17
README, NOTES.txt, values.yaml updates and fixes

prydonius:
For Redmine, Drupal and WordPress

General:
- Wordpress -> WordPress

READMEs:
- Add persistence configuration values
- Update persistence sections to describe Persistent Volume Claim usage

NOTES.txt:
- Fixes `kubectl` command for retrieving URL for ClusterIP service type
- Use root (homepage) URLs instead of login URLs
- Fix forwarding commands for ClusterIP service type, and use
  unprivileged port
- Improve LoadBalancer note printout
- Improve credentials printout

values.yaml:
- redmineUser -> redmineUsername
- drupalUser -> drupalUsername

/cc @migmartri @sameersbn

### DIFF
--- a/drupal/templates/deployment.yaml
+++ b/drupal/templates/deployment.yaml
@@ -19,8 +19,8 @@ spec:
     spec:
       containers:
       - name: {{ template "fullname" . }}
-        image: "bitnami/drupal:{{ .Values.imageTag }}"
-        imagePullPolicy: {{ default "" .Values.imagePullPolicy | quote }}
+        image: "{{ .Values.image.repo }}:{{ .Values.image.tag }}"
+        imagePullPolicy: {{ default "" .Values.image.pullPolicy | quote }}
         env:
         - name: MARIADB_HOST
           value: {{ template "mariadb.fullname" . }}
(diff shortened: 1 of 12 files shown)


drupal/templates/deployment.yaml
```diff
@@ -19,8 +19,8 @@ spec:
     spec:
       containers:
       - name: {{ template "fullname" . }}
-        image: "bitnami/drupal:{{ .Values.imageTag }}"
-        imagePullPolicy: {{ default "" .Values.imagePullPolicy | quote }}
+        image: "{{ .Values.image.repo }}:{{ .Values.image.tag }}"
```
migmartri: What do you think if instead of splitting it make values.image to include the tag as well if wanted?

That seems to me easier to understand by user since it is the same way people configure docker-compose files, etc.

redmine/values.yaml
```diff
@@ -1,18 +1,20 @@
 ## Bitnami Redmine image version
 ## ref: https://hub.docker.com/r/bitnami/redmine/tags/
 ##
-imageTag: 3.3.0-r2
+image:
```
migmartri: Same thing here about grouping both keys.

drupal/templates/deployment.yaml
prydonius: Yeah you're right, that's way easier :+1:

redmine/values.yaml
prydonius: so I'll have two top-level keys: `image` and `imagePullPolicy`. Sound good?

migmartri: :+1:

PR #134, opened 2016-09-20
bump jenkins imageTag to '2.23-r0'

jotadrilo:


jotadrilo:
I will merge once the tag is available in DH

PR #139, opened 2016-09-27
Added SuiteCRM Chart

juan131:


sameersbn:
I'm seeing random session terminations with this chart, which I also found to be more frequent with firefox. A sure fire way to encounter this issue is to visit the link **Create > Create Account**

PR #141, opened 2016-10-03
bump Drupal imageTag to 8.1.10-r0

jotadrilo:


jotadrilo:
This image is not built yet, I will update the PR when it is done

jotadrilo:
The new tagged image is built

prydonius:
Thanks!

PR #142, opened 2016-10-04
Update Magento and ownCloud image version

tompizmor:


prydonius:
Can you bump the minor version in Chart.yaml for these?

PR #159, opened 2016-10-26
Updated Nginx tag to 1.10.2-r0

Fortiz2305:


sameersbn:
@Fortiz2305 could you bump the chart patch version as well?

PR #161, opened 2016-10-27
Update PrestaShop revision to 1.6.1.7-r1

juan131:


sameersbn:
chart was updated with image: 1.6.1.8-r0 in #166

PR #167, opened 2016-10-31
Bump Tomcat version

Fortiz2305:


incubator/tomcat/values.yaml
```diff
@@ -1,7 +1,7 @@
 ## Bitnami Tomcat image version
 ## ref: https://hub.docker.com/r/bitnami/tomcat/tags/
 ##
-imageTag: 8.0.35-r4
+imageTag: 8.0.38-r0
```
sameersbn: image does not exist at https://hub.docker.com/r/bitnami/tomcat/tags/

PR #170, opened 2016-10-31
Add Osclass Chart

andresbono:


incubator/osclass/templates/deployment.yaml
```diff
@@ -0,0 +1,90 @@
...
+              name: {{ template "fullname" . }}
+              key: smtp-password
+        - name: SMTP_PROTOCOL
+          value: {{ default "" .Values.smtpProtocol | quote }}
+        ports:
+        - name: http
+          containerPort: 80
+        - name: https
+          containerPort: 443
+        livenessProbe:
```
prydonius: `nc` should be used instead, since the httpGet probe fails

sameersbn: @andresbono FWIW, `nc` has also been used in the probes of the phabricator chart. see: https://github.com/bitnami/charts/blob/master/incubator/phabricator/templates/deployment.yaml#L67-L82

PR #186, opened 2016-11-09
docs: added a template for creating README's for charts

sameersbn:


prydonius:
Cool! We should see if we can get something like this generated with the `helm create` command.

PR #191, opened 2016-11-16
Update phabricator version to 2016.46

carrodher:


incubator/phabricator/Chart.yaml
```diff
@@ -15,7 +15,7 @@ keywords:
 - git
 - mercurial
 - subversion
-home: https://www.phacility.com
+home: https://www.phacility.com/phabricator
```
sameersbn: funny thing is the url is broken if its not terminated with `/`, i.e. `https://www.phacility.com/phabricator/`

PR #193, opened 2016-11-17
Update owncloud version to 9.1.2

carrodher:
Correct password is _bitnami_, I don't know the reason why it was established to _bitnami1_


prydonius:
Weird - thanks for finding that fix @carrodher!

PR #259, opened 2017-01-05
bulk updates

sameersbn:
- update mariadb dependency chart version
 - explicitly set `IfNotPresent` as the default `imagePullPolicy`


prydonius:
@sameersbn don't forget to update the chart versions in the future

sameersbn:
@prydonius i did not update the chart versions because the same chart are submitted to official and should be accepted soon. bumping the version here would make the chart in this repo bear a higher version.

PR #300, opened 2017-02-25
Reduce labels for pod spec

prydonius:
Reduces the labels for the pod spec to ensure the chart version changes don't affect the match selector.


prydonius:
cc @jbianquetti-nami @sameersbn

sameersbn:
why drop the release and heritage lables?
Also I wonder if we should incorporate the revision in the labels.

sameersbn:
> Also I wonder if we should incorporate the revision in the labels.

Maybe not since that will be associated with the chart version.

PR #378, opened 2017-06-19
Add tensorflow-inception chart

tompizmor:
This chart deploys a tensorflow-serving server serving the inception model for image recognition.


incubator/tensorflow-inception/templates/inception-job.yaml
```diff
@@ -0,0 +1,26 @@
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: {{ template "fullname" . }}-seed-inception
```
prydonius: we should add the `app` label and probably others

incubator/tensorflow-inception/templates/inception-pvc.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: {{ template "fullname" . }}-seed-inception
```
prydonius: labels

incubator/tensorflow-inception/templates/inception-job.yaml
```diff
@@ -0,0 +1,26 @@
+apiVersion: batch/v1
+kind: Job
```
prydonius: It seems like this job should be a pre/post-install hook. Actually, it might even be better as an init-container, but if we intend to scale the server out that would mean we would have multiple seed volumes.

incubator/tensorflow-inception/templates/pvc.yaml
```diff
@@ -0,0 +1,18 @@
+{{- if .Values.persistence.enabled }}
+kind: PersistentVolumeClaim
+apiVersion: v1
+metadata:
+  name: {{ template "fullname" . }}
```
prydonius: labels

incubator/tensorflow-inception/templates/inception-job.yaml
tompizmor: yes, right

incubator/tensorflow-inception/templates/inception-job.yaml
tompizmor: the problem with pre/post-install hooks is that the resources created in those hooks are not deleted with `helm delete release` and also the helm notes are not shown until the Job has finished.

As you said the init-container will make scale the server odd. 
I think it is fine to launch the Job at the same time as the other resources as the Deployment has readiness and liveness probes and it will be restarting pods a couple of times until the Job finishes.

incubator/tensorflow-inception/templates/inception-job.yaml
tompizmor: done

incubator/tensorflow-inception/templates/inception-pvc.yaml
tompizmor: done

incubator/tensorflow-inception/templates/pvc.yaml
tompizmor: done

incubator/tensorflow-inception/templates/inception-job.yaml
prydonius: thanks for the explanation

incubator/tensorflow-inception/templates/inception-job.yaml
prydonius: They're in the best practices but optional: https://github.com/kubernetes/helm/blob/master/docs/chart_best_practices/labels.md. It's always good to have at least one label that can be used to query resources from a chart

incubator/tensorflow-inception/Chart.yaml
```diff
@@ -0,0 +1,19 @@
+name: tensorflow-serving
+version: 0.0.1
+app_version: 0.5.1
```
prydonius: this should be `appVersion`

incubator/tensorflow-inception/.helmignore
```diff
@@ -0,0 +1 @@
+.git
```
prydonius: just fyi - this isn't the right `.helmignore`, you can copy the one from here https://github.com/kubernetes/charts/blob/master/stable/acs-engine-autoscaler/.helmignore but it's not that important

incubator/tensorflow-inception/README.md
```diff
@@ -0,0 +1,84 @@
...
+- PV provisioner support in the underlying infrastructure
+
+## Get this chart
+
+Download the latest release of the chart from the [releases](../../../releases) page.
+
+Alternatively, clone the repo if you wish to use the development snapshot:
+
+```console
+$ git clone https://github.com/bitnami/tensorflow.git
```
sameersbn: Should be `git clone https://github.com/bitnami/bitnami-docker-tensorflow-serving.git`

incubator/tensorflow-inception/values.yaml
```diff
@@ -0,0 +1,35 @@
+## TensorFlow Serving server image version
+## ref: https://hub.docker.com/r/bitnami/tensorflow-serving/tags/
+##
+server:
+  image: tompizmor/tensorflow-serving:latest
```
sameersbn: Did you mean to use `bitnami/tensorflow-serving` instead? Also it's not recommended to use images tagged `latest` in charts. It does not follow best practices

incubator/tensorflow-inception/values.yaml
```diff
@@ -0,0 +1,35 @@
...
+  port: 9000
+
+## TensorFlow Serving Client image version
+## ref: https://hub.docker.com/r/bitnami/tensorflow-serving-client/tags/
+##
+client:
+  image: tompizmor/tensorflow-serving-client:latest
```
sameersbn: I think you meant to use `bitnami/tensorflow-serving-client`, but I don't see it exist https://hub.docker.com/r/bitnami/tensorflow-serving-client

incubator/tensorflow-inception/values.yaml
```diff
@@ -0,0 +1,35 @@
...
+
+## Specify a imagePullPolicy
+## Defaults to 'Always' if image tag is 'latest', else set to 'IfNotPresent'
+## ref: http://kubernetes.io/docs/user-guide/images/#pre-pulling-images
+##
+# imagePullPolicy:
```
sameersbn: If we're going to use images tagged `latest` , then we should defaulting this to `Always`

incubator/tensorflow-inception/values.yaml
tompizmor: yes, I forget to rename the image. I will use a specific tag instead of latest

incubator/tensorflow-inception/values.yaml
tompizmor: Finally it was renamed to bitnami/tensorflow-inception

incubator/tensorflow-inception/values.yaml
tompizmor: we will use tags so `imagePullPolicy: IfNotPresent`

incubator/tensorflow-inception/Chart.yaml
tompizmor: fixed

incubator/tensorflow-inception/.helmignore
tompizmor: updated

incubator/tensorflow-inception/README.md
tompizmor: well it should be `git clone https://github.com/bitnami/charts.git` as it is where the chart will be

incubator/tensorflow-inception/templates/inception-job.yaml
tompizmor: ok, thanks!

incubator/tensorflow-inception/templates/inception-job.yaml
```diff
@@ -0,0 +1,31 @@
...
+spec:
+  template:
+    metadata:
+      name: seed-inception
+    spec:
+      containers:
+      - name: seed
+        image: {{ .Values.client.image }}
+        command: ["/bin/sh"]
+        args: ["-c", "curl -o /seed/inception-v3-2016-03-01.tar.gz http://download.tensorflow.org/models/image/imagenet/inception-v3-2016-03-01.tar.gz && \
```
prydonius: I think this will be easier to read if you do it like https://github.com/kubernetes/charts/blob/master/incubator/etcd/templates/etcd-statefulset.yaml#L92

tompizmor: thanks, I didn't know that was possible

PR #402, opened 2017-07-08
[incubator/tomcat] Add sidecar war container

dhcdata:
Implementation of sidecar war files as described in[ Kubernetes example docs](https://github.com/kubernetes/kubernetes/tree/master/examples/javaweb-tomcat-sidecar)

The sidecar Dockerfile should look like this

```
# Dockerfile
FROM busybox:latest
ADD sample.war sample.war
CMD "tail" "-f" "/dev/null"
```

### DIFF
--- a/incubator/tomcat/templates/deployment.yaml
+++ b/incubator/tomcat/templates/deployment.yaml
@@ -15,7 +15,7 @@ spec:
     spec:
       containers:
       {{- if .Values.sidecarContainers }}
-{{ toYaml .Values.sidecarContainers }}
+{{ toYaml .Values.sidecarContainers | indent 6}}
       {{- end }}
       - name: {{ template "fullname" . }}
         image: "{{ .Values.image }}"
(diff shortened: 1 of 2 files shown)


sameersbn:
Hi @dhcdata, this is an excellent idea. Thanks for the contribution.

I really think we should add this functionality to the chart. However, I am of the opinion that instead of a sidecar container we could start the container as an init container and copy the war file. After the war file is copied the container will terminate and the tomcat container will be launched. As a we result we'll not have a sidecar container running for the lifetime of the pod.

For reference see https://github.com/bitnami/charts/blob/master/incubator/mean/templates/deployment.yaml#L36-L47

What do you think?

dhcdata:
I just added another commit to include ad hoc volume mounts. This will allow the individual tomcat config files to be overridden.

I have not used Init Containers yet. After briefly looking at the docs, Inits look like an improved approach.  The same can be accomplished with this PR by providing a `sidecar.command` that terminates.

I'll read more about Inits and see if I can get it to work.

dhcdata:
Added another commit that changes the primary use to an init container.  I'm about to test it.  The sidecar approach should remain for those using a development/testing deployment that want to continue to pull from github.

I agree that a production use should be using the init container approach (or single image if start time becomes problematic).

dhcdata:
This now works as expected.

sameersbn:
@dhcdata thanks for the contribution. I would need a little time to fully test the changes in the PR.

carrodher:
Hi, I just found this pull request, sorry, I do not know how we could ignore it for so long. 

I've been looking at the changes and they are strong changes, it may make sense in the development containers, but I think that we should have the same approach to all development charts, therefore it is something to discuss.

Any opinion about it? cc @sameersbn @prydonius @tompizmor @juan131 @javsalgar

PR #428, opened 2017-09-04
Add allowRemoteManagement as a parameter for the Tomcat Helm Chart

dbarranco:
This PR is intended to add the parameter to remotely get access to the Tomcat management services and make us able to test all the services in our CI/CD systems.

### DIFF
--- a/incubator/tomcat/README.md
+++ b/incubator/tomcat/README.md
@@ -55,6 +55,8 @@ The following tables lists the configurable parameters of the Tomcat chart and t
 | `persistence.accessMode`   | PVC Access Mode for Tomcat volume     | `ReadWriteOnce`                             |
 | `persistence.size`         | PVC Storage Request for Tomcat volume | `8Gi`                                       |
 | `resources`                | CPU/Memory resource requests/limits   | Memory: `512Mi`, CPU: `300m`                |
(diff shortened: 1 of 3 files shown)


dbarranco:
@sameersbn I've been trying to enable the management services in my minikube cluster but it seems adding the value it is not enough, cause I still don't have access to the management services.

According to the parameters of [the Tomcat container](https://github.com/bitnami/bitnami-docker-tomcat/blob/master/8.5/rootfs/tomcat-inputs.json)

```
  "allowRemoteManagement": "{{$global.env.TOMCAT_ALLOW_REMOTE_MANAGEMENT}}",
```
That is the parameter we should add here, but I'm unsure if I'm missing something.

sameersbn:
@dbarranco can you check now. I amended a commit that adds `TOMCAT_ALLOW_REMOTE_MANAGEMENT` to the container environment.

dbarranco:
@sameersbn Tested and working as it should. 
Looks good to merge

sameersbn:
we're you able to login as the manager? I can't seem to login

dbarranco:
@sameersbn proceeding to land along with the changes in the Tomcat nami module

PR #506, opened 2018-02-16
Refactor node chart to be production ready

tompizmor:
This refactor includes the following features:

- Supports ingress (disabled by default) 
- Supports ingress TLS termination (the secret with the TLS certificate must exist before deploying the chart or use [kube-lego](https://github.com/kubernetes/charts/tree/master/stable/kube-lego) )
- The number of replicas of the application is configurable via values.
- It uses a minimal node image to run the application (`bitnami/node:TAG-prod`)
- If your application requires a database you can provision a production-ready one running in the Azure cloud using k8s service catalog and the Open Service Broker for Azure.

### DIFF
--- a/incubator/node/README.md
+++ b/incubator/node/README.md
@@ -5,14 +5,14 @@
 ## TL;DR;
 
 ```console
-$ helm install incubator/node
+$ helm repo add bitnami-incubator https://charts.bitnami.com/incubator
+$ helm install bitnami-incubator/node
 ```
-
 ## Introduction
 
 This chart bootstraps a [Node](https://github.com/bitnami/bitnami-docker-node) deployment on a [Kubernetes](http://kubernetes.io) cluster using the [Helm](https://helm.sh) package manager.
 
-It clones and deploys a Node.js application from a git repository.
+It clones and deploys a Node.js application from a git repository. Optionally you can set un an Ingress resource to access your application and provision an external database using the k8s service catalog and the Open Service Broker for Azure.
 
 ## Prerequisites
 
@@ -24,7 +24,7 @@ It clones and deploys a Node.js application from a git repository.
 To install the chart with the release name `my-release`:
 
 ```console
-$ helm install --name my-release incubator/node
+$ helm install --name my-release bitnami-incubator/node
 ```
 
 The command deploys Node.js on the Kubernetes cluster in the default configuration. The [configuration](#configuration) section lists the parameters that can be configured during installation. Also includes support for MariaDB chart out of the box.
@@ -47,15 +47,27 @@ The command removes all the Kubernetes components associated with the chart and 
 
 The following tables lists the configurable parameters of the Node chart and their default values.
 
-|           Parameter           |             Description             |                        Default                        |
-|-------------------------------|-------------------------------------|-------------------------------------------------------|
-| `image`                       | Node image                          | `bitnami/node:{VERSION}`                              |
-| `imagePullPolicy`             | Image pull policy                   | `IfNotPresent`                                        |
-| `repository`                  | Repo of the application             | `git@github.com:jbianquetti-nami/simple-node-app.git` |
-| `revision`                    | Revision  to checkout               | `master`                                              |
-| `mariadb.mariadbRootPassword` | MariaDB admin password              | `nil`                                                 |
-| `serviceType`                 | Kubernetes Service type             | `LoadBalancer`                                        |
-| `resources`                   | CPU/Memory resource requests/limits | Memory: `512Mi`, CPU: `300m`                          |
+|              Parameter               |            Description                                    |                        Default                            |
+|--------------------------------------|-----------------------------------------------------------|-----------------------------------------------------------|
+| `image`                              | Node image                                                | `bitnami/node:{VERSION}`                                  |
+| `devImage`                           | Image used for initContainers                             | `bitnami/node:{VERSION}`                                  |
+| `imagePullPolicy`                    | Image pull policy                                         | `IfNotPresent`                                            |
+| `repository`                         | Repo of the application                                   | `https://github.com/jbianquetti-nami/simple-node-app.git` |
+| `revision`                           | Revision  to checkout                                     | `master`                                                  |
+| `replicas`                           | Number of replicas for the application                    | `1`                                                       |
+| `applicationPort`                    | Port where the application will be running                | `3000`                                                    |
+| `serviceType`                        | Kubernetes Service type                                   | `LoadBalancer`                                            |
+| `persistence.enabled`                | Enable persistence using PVC                              | `false`                                                   |
+| `persistence.path`                   | Path to persisted directory                               | `/app/data`                                               |
+| `persistence.accessMode`             | PVC Access Mode                                           | `ReadWriteOnce`                                           |
+| `persistence.size`                   | PVC Storage Request                                       | `1Gi`                                                     |
+| `externalDatabase.azure.enabled`     | Create a database on Azure cloud with k8s service catalog | `false`                                                   |
+| `externalDatabase.azure.location`    | The Azure region in which to deploy the database          | `eastus`                                                  |
+| `externalDatabase.azure.servicePlan` | The plan to request for the Azure database                | `mongo-db`                                                |
+| `ingress.enabled`                    | Enable ingress creation                                   | `false`                                                   |
+| `ingress.path`                       | Ingress path                                              | `/`                                                       |
+| `ingress.host`                       | Ingress host                                              | `example.local`                                           |
+| `ingress.tls`                        | TLS configuration for the ingress                         | `{}`                                                      |
 
 The above parameters map to the env variables defined in [bitnami/node](http://github.com/bitnami/bitnami-docker-node). For more information please refer to the [bitnami/node](http://github.com/bitnami/bitnami-docker-node) image documentation.
 
@@ -63,11 +75,11 @@ Specify each parameter using the `--set key=value[,key=value]` argument to `helm
 
 ```console
 $ helm install --name my-release \
-  --set repository=https://github.com/jbianquetti-nami/simple-node-app.git,mariadb.mariadbRootPassword=secretpassword \
-    incubator/node
+  --set repository=https://github.com/jbianquetti-nami/simple-node-app.git,replicas=2 \
+    bitnami-incubator/node
 ```
 
-The above command clones the remote git  repository to the `/app/` directory  of the container. Additionally it sets the MariaDB `root` user password to `secretpassword`.
+The above command clones the remote git  repository to the `/app/` directory  of the container. Additionally it sets the number or `replicas` to `2`.
 
 Alternatively, a YAML file that specifies the values for the above parameters can be provided while installing the chart. For example,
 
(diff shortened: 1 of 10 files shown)


incubator/node/README.md
```diff
@@ -83,3 +95,58 @@ The [Bitnami Node](https://github.com/bitnami/bitnami-docker-node) image stores
...
+
+1. Install Service Catalog in your Kubernetes cluster following [this instructions](https://kubernetes.io/docs/tasks/service-catalog/install-service-catalog-using-helm/)
+2. Install the Open Service Broker for Azure in your Kubernetes cluster following [this instructions](https://github.com/Azure/helm-charts/tree/master/open-service-broker-azure)
+3. Deploy the helm chart:
+
+    ```
+    $ helm install --name node-app --set externalDatabase.azure.enabled=true bitnami-incubator/node
+    ```
+
+Setting `externalDatabase.azure.enabled` to `true` makes the chart to create a `ServiceInstance` and a `ServiceBinding` kubernetes resources.
```
arapulido: Mmm, is this what we wanted to do? I thought the idea was to make it more generic (not only for Azure), but have instructions on how to use the sevice catalog to create an instance in Azure, and connect it using the generic externalDatabase values.

tompizmor: We thought to do it in this way because the ServiceInstance and ServiceBinding resources are automatically created for the use. Make it more generic means that the user will need to write and create these resources on its own before deploys the chart.  
Also, the idea was to consume the database credentials from the Secret instead of passing them as standard helm values.

prydonius: So I think what we should do here is consume the secret, and not create the ServiceInstance and ServiceBinding. We can assume that those have been created, out of band, and just take in the result of that binding (i.e. the secret).

The main reason for this is that ServiceInstances usually require a range of parameters that are very specific to the cloud platform. Whilst we could easily make those parameters configurable via chart values, I think it's better to not create a direct dependency on the ServiceInstance.

This means that another scenario we can support is when you already have a ServiceInstance (and maybe also a ServiceBinding) and want to use that instead of have the chart provision one for you. With the current approach, it's not possible to use an existing ServiceInstance.

We should document the steps needed to create the ServiceInstance and ServiceBinding, either by having example YAML files or using the svcat CLI. The chart should then take in a secret (`--set externalDatabase.secretName=<name>`) and we should document what keys we expect that secret to have (for now, it should be whatever the OSBA ServiceBinding creates for us).

WDYT?

incubator/node/templates/deployment.yaml
```diff
@@ -1,22 +1,23 @@
...
-        app: {{ template "fullname" . }}
+        app: {{ template "node.name" . }}
+        release: {{ .Release.Name | quote }}
     spec:
       initContainers:
       - name: git-clone-app
-        image: {{ .Values.image }}
+        image: {{ .Values.devImage }}
         imagePullPolicy: {{ .Values.imagePullPolicy }}
         command: [ '/bin/sh', '-c' , 'git clone {{ .Values.repository }} /app && git checkout {{ .Values.revision }}']
```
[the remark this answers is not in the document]
tompizmor: I tried using the gitRepo volume but it clones the repo in ${mountPoint}/${repoName} while I want to always clone in `/app` to not depends on the name. I am using no the alpine/git image

incubator/node/templates/deployment.yaml
```diff
@@ -1,22 +1,23 @@
 apiVersion: extensions/v1beta1
 kind: Deployment
 metadata:
-  name: {{ template "fullname" . }}
+  name: {{ template "node.fullname" . }}
   labels:
-    app: {{ template "fullname" . }}
+    app: {{ template "node.name" . }}
     chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
```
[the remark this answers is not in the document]
tompizmor: good catch!

incubator/externaldb-helper/Chart.yaml
```diff
@@ -0,0 +1,13 @@
+name: externaldb-helper
```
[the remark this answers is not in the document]
tompizmor: @prydonius @sameersbn thoughts?

incubator/node/README.md
```diff
@@ -97,7 +97,7 @@ The [Bitnami Node](https://github.com/bitnami/bitnami-docker-node) image stores
 Persistent Volume Claims are used to keep the data across deployments. This is known to work in GCE, AWS, and minikube.
 See the [Configuration](#configuration) section to configure the PVC or to disable persistence.
 
-## Set an Ingress
+## Set an Ingress controller
```
sameersbn: `Set` -> `Set up`

incubator/node/templates/deployment.yaml
```diff
@@ -17,9 +17,9 @@ spec:
     spec:
       initContainers:
       - name: git-clone-app
-        image: {{ .Values.devImage }}
+        image: {{ .Values.gitImage }}
         imagePullPolicy: {{ .Values.imagePullPolicy }}
-        command: [ '/bin/sh', '-c' , 'git clone {{ .Values.repository }} /app && git checkout {{ .Values.revision }}']
+        command: ['git', 'clone', '{{ .Values.repository }}', '-b', '{{ .Values.revision }}', '/app']
```
sameersbn: `-b` argument for `git clone` will not work for revisions. It should either be a tag or branch.

incubator/node/README.md
```diff
@@ -47,27 +47,40 @@ The command removes all the Kubernetes components associated with the chart and
...
-| `resources`                   | CPU/Memory resource requests/limits | Memory: `512Mi`, CPU: `300m`                          |
+|              Parameter                  |            Description                                    |                        Default                            |
+|-----------------------------------------|-----------------------------------------------------------|-----------------------------------------------------------|
+| `image`                                 | Node image                                                | `bitnami/node:{VERSION}`                                  |
+| `gitImage`                              | Image used for initContainers                             | `alpine/git`                                              |
+| `imagePullPolicy`                       | Image pull policy                                         | `IfNotPresent`                                            |
+| `repository`                            | Repo of the application                                   | `https://github.com/jbianquetti-nami/sample-mean.git`     |
```
sameersbn: should be https://github.com/bitnami/sample-mean.git

incubator/node/README.md
```diff
@@ -83,3 +96,98 @@ The [Bitnami Node](https://github.com/bitnami/bitnami-docker-node) image stores
...
+      app: mongodb
+  spec:
+    clusterServiceClassExternalName: azure-cosmos-mongo-db
+    clusterServicePlanExternalName: mongo-db
+    parameters:
+      location: YOUR_AZURE_LOCATION
+      resourceGroup: mongodb-k8s-service-catalog
+      sslEnforcement: disabled
+      firewallStartIPAddress: 0.0.0.0
+      firewallEndIPAddress: 255.255.255.255
```
sameersbn: the firewall specification has changed for osba, see https://github.com/Azure/helm-charts/commit/75ab2232e80c1469ad262faa8039ac64262c5c8c#diff-d04a73ad2755189100f39bf28c84a22e for the necessary changes.

tompizmor: thanks!

tompizmor:
@prydonius @sameersbn I moved the bitnami-common chart to this other PR https://github.com/bitnami/charts/pull/639 . Until that chart is not merged the ci will fail for this PRL

incubator/node/README.md
```diff
@@ -5,14 +5,14 @@
 ## TL;DR;
```
sameersbn: `TL;DR;` -> `TL;DR`

incubator/node/README.md
```diff
@@ -47,30 +47,47 @@ The command removes all the Kubernetes components associated with the chart and
 
 The following tables lists the configurable parameters of the Node chart and their default values.
```
sameersbn: `The following tables lists` -> `The following table lists`

incubator/node/README.md
```diff
@@ -47,30 +47,47 @@ The command removes all the Kubernetes components associated with the chart and
...
-| `resources`                   | CPU/Memory resource requests/limits | Memory: `512Mi`, CPU: `300m`                              |
+|              Parameter                  |            Description                                    |                        Default                            |
+|-----------------------------------------|-----------------------------------------------------------|-----------------------------------------------------------|
+| `image.registry`                        | NodeJS image registry                                     | `docker.io`                                               |
+| `image.repository`                      | NodeJS Image name                                         | `bitnami/node`                                            |
+| `image.tag`                             | NodeJS Image tag                                          | `{VERSION}`                                               |
+| `image.pullPolicy`                      | NodeJS image pull policy                                  | `Always` if `imageTag` is `latest`, else `IfNotPresent`   |
```
sameersbn: default value is `IfNotPresent`.

incubator/node/README.md
```diff
@@ -47,30 +47,47 @@ The command removes all the Kubernetes components associated with the chart and
...
+| `gitImage.registry`                     | Git image registry                                        | `docker.io`                                               |
+| `gitImage.repository`                   | Git Image name                                            | `alpine/git`                                              |
+| `gitImage.tag`                          | Git Image tag                                             | `latest`                                                  |
+| `gitImage.pullPolicy`                   | Git image pull policy                                     | `Always` if `imageTag` is `latest`, else `IfNotPresent`   |
+| `gitImage`                              | Image used for initContainers                             | `alpine/git`                                              |
+| `repository`                            | Repo of the application                                   | `https://github.com/bitnami/sample-mean.git`              |
+| `revision`                              | Revision  to checkout                                     | `master`                                                  |
```
sameersbn: remove whitespaces in `Revision to checkout`

incubator/node/README.md
```diff
@@ -47,30 +47,47 @@ The command removes all the Kubernetes components associated with the chart and
...
+| `replicas`                              | Number of replicas for the application                    | `3`                                                       |
```
sameersbn: defaults to `1`

incubator/node/README.md
```diff
@@ -47,30 +47,47 @@ The command removes all the Kubernetes components associated with the chart and
...
 ```console
 $ helm install --name my-release \
-  --set repository=https://github.com/jbianquetti-nami/simple-node-app.git,mariadb.mariadbRootPassword=secretpassword \
-    incubator/node
+  --set repository=https://github.com/jbianquetti-nami/simple-node-app.git,replicas=2 \
+    bitnami-incubator/node
 ```
 
-The above command clones the remote git  repository to the `/app/` directory  of the container. Additionally it sets the MariaDB `root` user password to `secretpassword`.
+The above command clones the remote git repository to the `/app/` directory  of the container. Additionally it sets the number or `replicas` to `2`.
```
sameersbn: number ~or~of replicas

incubator/node/README.md
```diff
@@ -86,3 +103,100 @@ The [Bitnami Node](https://github.com/bitnami/bitnami-docker-node) image stores
 
 Persistent Volume Claims are used to keep the data across deployments. This is known to work in GCE, AWS, and minikube.
 See the [Configuration](#configuration) section to configure the PVC or to disable persistence.
+
+## Set an Ingress controller
```
sameersbn: `Set an` -> `Set up an`

PR #639, opened 2018-04-03
Add bitnami-common chart with custom templates

tompizmor:
This chart will be a place to define common templates like the exterdal db helpers for service catalog mapping.

### DIFF
--- a/incubator/bitnami-common/Chart.yaml
+++ b/incubator/bitnami-common/Chart.yaml
@@ -1,0 +1,15 @@
+name: bitnami-common
+version: 0.0.1
+appVersion: 0.0.1
+description: Chart with custom tempaltes used in Bitnami charts.
+keywords:
+- helper
+- template
+- bitnami
+- broker
+- service
+- catalog
+maintainers:
+- name: Bitnami
+  email: containers@bitnami.com
+engine: gotpl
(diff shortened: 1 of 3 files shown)


sameersbn:
A table in the readme with a list of available helpers would really be helpful

prydonius:
Can we create a new repo/folder for this? I would prefer it not be in the main repo because it's not `helm install`-able. I suggest creating a folder at the root here called `library`, which would give us `https://charts.bitnami.com/library` automatically I believe.

tompizmor:
It is not installable directly but it is a dependency for other charts.
Creating a new repo/folder for this means that the user will need to also add this `https://charts.bitnami.com/library` repo in order to update the dependencies.

prydonius:
> Creating a new repo/folder for this means that the user will need to also add this https://charts.bitnami.com/library repo in order to update the dependencies.

Not if they're getting the packaged version from the chart repo, if they're checking out the source and building the chart locally, then yes, but that can be documented.

tompizmor:
oh, you are right! Then my only concern is if we end up having only one chart repo per branch

tompizmor:
well yes! By the moment I will move it to the `library/` folder.

PR #647, opened 2018-04-12
use statefulset for master nodes

sameersbn:
When the chart release is deleted all resources, with the exception of the pvc (storage) that were associated with the slave statefulset, are deleted. This is a policy of the statefulset resource. This means that if we deploy the chart again with the same release name, the existing pvc will be associated with the new release and hence the slave pods will not go through the initialization and in essence reuse the state from the previous release. As you can imaging this will end up in a broken deploy since the master will be re-initialized while the slaves will be re-using the past state. This is demostrated below in use case 1.

To fix we should use a statefulset for the master as well; so the entire state of the cluster is preserved as a whole.

Use Case: Release is deleted and redeployed with same release name
---

```bash
# install the chart
$ helm install --name mariadb incubator/mariadb-cluster --set root.password=password,replication.password=anotherpassword

# verify that the cluster is installed properly
$ kubectl get pods -l "release=mariadb"
NAME                                     READY     STATUS    RESTARTS   AGE
mariadb-mariadb-master-b64d8c6dc-s2xn6   1/1       Running   0          1m
mariadb-mariadb-slave-0                  1/1       Running   0          1m

# delete the release
$ helm delete --purge mariadb

# notice that the slave pvc is still around
$ kubectl get pvc -l "release=mariadb"
NAME                           STATUS    VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS   AGE
data-mariadb-mariadb-slave-0   Bound     pvc-cd93f5bb-3e20-11e8-9c7e-42010aa00030   8Gi        RWO            standard       6m

# install the chart with the same release name
$ helm install --name mariadb incubator/mariadb-cluster --set root.password=password,replication.password=anotherpassword

# slave pods will fail repeatedly (broken deployment)
$ kubectl get pods -l "release=mariadb"
NAME                                     READY     STATUS    RESTARTS   AGE
mariadb-mariadb-master-b64d8c6dc-d4qt9   1/1       Running   0          9m
mariadb-mariadb-slave-0                  0/1       Running   3          9m
```


beltran-rubo:
Thanks. This is a valid scenario so moving the master to statefulset would simplify reusing the PVCs. I also tested it in my local environment.

PR #652, opened 2018-04-13
Add Consul to bitnami charts

tompizmor:
List of things tested and pending ones:

- [x] Upscaling/Downscaling the cluster from 3 to 5 nodes and 5 to 3 (those numbers are the recommended by the Consul team)
- [x] Secret with the private key it's taken by the nami module and set in the right place
- [x] Values parametrized in values.yaml file
- [x] Load configuration via env var
- [x] Load configuration via configMap
- [x] Ingress to access the user interface
- [x] Node affinity settings
- [x] README
- [x]  Review notes.txt
- [x]  Prometheus Consul exporter
- [x]  Check structure of values.yaml. (image part)
- [x]  Secret of TLS for ingress (Documentation)
- [x]  Secret for TLS between servers (secret structure and documentation)
- [x] Check the logic of the Nami module for the persistence of the configuration files outside k8s.

### DIFF
--- a/bitnami/consul/values.yaml
+++ b/bitnami/consul/values.yaml
@@ -1,0 +1,206 @@
+## Bitnami Osclass image version
+## ref: https://hub.docker.com/r/bitnami/osclass/tags/
+##
+image:
+  registry: docker.io
+  repository: bitnami/consul
+  tag: 1.0.6
+  ## Specify a imagePullPolicy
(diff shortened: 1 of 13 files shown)


bitnami/consul/values.yaml
```diff
@@ -0,0 +1,206 @@
+## Bitnami Osclass image version
+## ref: https://hub.docker.com/r/bitnami/osclass/tags/
```
sameersbn: OSclass is referred in these comments

PR #672, opened 2018-05-28
Using Bitnami Image for ES Exporter

juan131:
Bitnami recently published a Docker Image for the ES exporter that will be maintained, up-to-date and free or CVEs. 

This PR adapt the chart so this new image is used for the Metrics pods.


tompizmor:
update also the image name in the README

PR #674, opened 2018-05-29
[bitnami/postgresql] - Allow configure replication for postgresql chart 

tompizmor:
This PRs add support to the postgresql chart to configures a replication master-slave. 
It also add other improvements like:
- configurable readiness and liveness probes
- ability to enable/disable a metrics exporter
- ability to mount custom config file via config map


bitnami/postgresql/templates/NOTES.txt
```diff
@@ -1,8 +1,54 @@
+{{- if contains .Values.service.type "LoadBalancer" }}
+{{- if not .Values.postgresPassword }}
```
juan131: Typo. Use `.Values.postgresqlPassword`

bitnami/postgresql/templates/secrets.yaml
```diff
@@ -1,16 +1,23 @@
...
-    chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+    app: {{ template "postgresql.name" . }}
+    chart: {{ template "postgresql.chart" . }}
     release: "{{ .Release.Name }}"
     heritage: "{{ .Release.Service }}"
 type: Opaque
 data:
   {{ if .Values.postgresqlPassword }}
-  postgresql-password: {{ default "" .Values.postgresqlPassword | b64enc | quote }}
+  postgresql-password: {{ .Values.postgresqlPassword | b64enc | quote }}
```
juan131: What happens if `.Values.postgresqlPassword` is not defined?

bitnami/postgresql/templates/statefulset.yaml
```diff
@@ -0,0 +1,148 @@
+apiVersion: apps/v1beta1
```
juan131: I see different solutions for the StatefulSet API version.. Some charts use `apps/v1beta1`, others `apps/v1beta2` and others `apps/v1`. We should standardise it.

bitnami/postgresql/templates/svc.yaml
```diff
@@ -1,17 +1,19 @@
 apiVersion: v1
```
juan131: Should we call it `svc-master.yaml`?

bitnami/postgresql/files/README.md
```diff
@@ -0,0 +1 @@
+Copy here your postgresql.conf file to use it as a config map.
```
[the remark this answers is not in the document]
tompizmor: ok

bitnami/postgresql/templates/NOTES.txt
tompizmor: thanks

bitnami/postgresql/templates/secrets.yaml
tompizmor: Then a 10 character alphanumeric password is generated:

```
  {{ if .Values.postgresqlPassword }}
  postgresql-password: {{ .Values.postgresqlPassword | b64enc | quote }}
  {{ else }}
  postgresql-password: {{ randAlphaNum 10 | b64enc | quote }}
  {{ end }}
```

bitnami/postgresql/templates/statefulset.yaml
tompizmor: You are right! I think we should use `apps/v1beta2`. 
The `apps/v1` only works in very new kubernetes clusters and we have already had issues in the past with this API version. https://github.com/kubernetes/charts/issues/4908

bitnami/postgresql/templates/svc.yaml
tompizmor: I prefer to keep it just `svc.yaml` because it also applies when you deploy a single postgresql node.

PR #681, opened 2018-06-13
[bitnami/kafka] - Add kafka chart

tompizmor:
This is the first version of the Bitnami Kafka chart with the following features:

- A scalable number of kafka nodes.
- It uses the bitnami/zookeeper chart as dependency but it can use any other external zookeeper server.
- Configures and expose liveness and readiness probes
- Allow overriding default configuration via config map
- The kafka SASL_SSL listener can be enabled using the `auth.enabled` switch. The user must use its own certificates.
- Run as non-root by default. Kubernetes security context is configured so volumes have the correct ownership.
- Metrics using kafka-exporter can be enabled with the `metrics.kafka.enabled` switch.
- Metrics using jmx-exporter can be enabled with the `metrics.kafka.enabled` switch.


bitnami/kafka/README.md
```diff
@@ -0,0 +1,215 @@
...
+| `auth.zookeeperUser`             | Kafka Zookeeper user.                                                                                      | `nil`                                                              |
+| `auth.zookeeperPassword`         | Kafka Zookeeper password.                                                                                  | `nil`                                                              |
+| `securityContext.enabled`        | Enable security context (redis master pod)                                                                 | `true`                                                             |
+| `securityContext.fsGroup`        | Group ID for the container (redis master pod)                                                              | `1001`                                                             |
+| `securityContext.runAsUser`      | User ID for the container (redis master pod)                                                               | `1001`                                                             |
```
javsalgar: Remove redis master pod in the three lines above

bitnami/kafka/README.md
```diff
@@ -0,0 +1,215 @@
...
+| `brokerId`                       | ID of the Kafka node                                                                                       | `-1`                                                               |
+| `deleteTopicEnable`              | Switch to enable topic deletion or not.                                                                    | `false`                                                            |
+| `heapOpts`                       | Kafka's Java Heap size.                                                                                    | `-Xmx1024m -Xms1024m`                                              |
+| `logFlushIntervalMessages`       | The number of messages to accept before forcing a flush of data to disk.                                   | `10000`                                                            |
+| `logFlushIntervalMs`             | The maximum amount of time a message can sit in a log before we force a flush.                             | `1000`                                                             |
```
javsalgar: Don't we want to group these parameters in a kafka group?

bitnami/kafka/values-production.yaml
```diff
@@ -0,0 +1,329 @@
...
+## ref: https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#taints-and-tolerations-beta-feature
+nodeSelector: {}
+tolerations: []
+
+## Configure resource requests and limits
+## ref: http://kubernetes.io/docs/user-guide/compute-resources/
+##
+resources:
+  requests:
+    memory: 256Mi
```
javsalgar: Given the JAVA memory requests you are setting above, it may not make sense to set 256Mi as the minimum memory request

bitnami/kafka/values.yaml
```diff
@@ -0,0 +1,329 @@
+## Bitnami Kafka image version
+## ref: https://hub.docker.com/r/bitnami/kafka/tags/
+##
+image:
+  registry: docker.io
+  repository: tompizmor/kafka
```
javsalgar: I assume this will be changed, right?

tompizmor: yes!

bitnami/kafka/README.md
tompizmor: done

bitnami/kafka/README.md
```diff
@@ -0,0 +1,215 @@
...
+| `readinessProbe.periodSeconds`        | How often to perform the probe                                                                        |  10                                                                |
+| `readinessProbe.timeoutSeconds`       | When the probe times out                                                                              |  5                                                                 |
+| `readinessProbe.failureThreshold`     | Minimum consecutive failures for the probe to be considered failed after having succeeded.            |  6                                                                 |
+| `readinessProbe.successThreshold`     | Minimum consecutive successes for the probe to be considered successful after having failed           |  1                                                                 |
+| `jmx.configMap.enabled`               | Enable the default ConfigMap for JMX                                                                  | `true`                                                             |
```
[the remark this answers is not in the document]
tompizmor: Yes

bitnami/kafka/README.md
tompizmor: The only reason why I didn't group these parameters is when someone uses the kafka chart as a dependency. 
In that case, in order to override a value from the parent chart would be sth like:

`--set kafka.kafka.brokerID`

bitnami/kafka/values-production.yaml
tompizmor: actually, the resource requests and limits should be commented by default.

bitnami/kafka/templates/jmx-configmap.yaml
```diff
@@ -13,7 +13,7 @@ data:
 {{- if .Values.jmx.configMap.overrideConfig }}
```
javsalgar: Use new approach here

PR #697, opened 2018-06-27
[bitnami/etcd] - First version of etcd chart

tompizmor:
This PR adds the first version of the etcd chart with the following features:

- **Using values.yaml**
  - Deploy 1 node
  - Persistence is enabled by default
  - etcd rbac auth enabled by default

- **Using values-production.yaml**
  - Deploy a cluster of 3 nodes
  - Persistence is enabled by default
  - etcd rbac auth enabled by default
  - peer-to-peer communication encrupted with auto-generated certificates
  - client-to-server communication requires cert files created by the user in a secret


bitnami/etcd/README.md
```diff
@@ -0,0 +1,178 @@
...
+| `image.tag`                        | etcd Image tag                                                                                             | `{VERSION}`                                                        |
+| `image.pullPolicy`                 | etcd image pull policy                                                                                     | `Always`                                                           |
+| `image.pullSecrets`                | Specify image pull secrets                                                                                 | `nil` (does not add image pull secrets to deployed pods)           |
+| `image.debug`                      | Specify if debug values should be set                                                                      | `false`                                                            |
+| `updateStrategy`                   | Update strategy for the stateful set                                                                       | `RollingUpdate`                                                    |
```
juan131: Add **rollingUpdatePartition** parameter to configure the partition update strategy as we did on other charts such as Redis.

bitnami/etcd/README.md
```diff
@@ -0,0 +1,178 @@
...
+| `securityContext.enabled`          | Enable security context                                                                                    | `true`                                                             |
+| `securityContext.fsGroup`          | Group ID for the container                                                                                 | `1001`                                                             |
+| `securityContext.runAsUser`        | User ID for the container                                                                                  | `1001`                                                             |
+| `service.type`                     | Kubernetes Service type                                                                                    | `ClusterIP`                                                        |
+| `service.port`                     | etcd client port                                                                                           | `2379`                                                             |
```
juan131: Shouldn't we include a `service.nodePorts` group of parameters with `service.nodePorts.client` and `service.nodePorts.peer`. This way, the user can configure what node ports to use when setting `service.type` to **NodePort**

bitnami/etcd/README.md
```diff
@@ -0,0 +1,178 @@
...
+| `persistence.annotations`          | Annotations for the PVC                                                                                    | `{}`                                                               |
+| `nodeSelector`                     | Node labels for pod assignment                                                                             | `{}`                                                               |
+| `tolerations`                      | Toleration labels for pod assignment                                                                       | `[]`                                                               |
+| `resources`                        | CPU/Memory resource requests/limits                                                                        | Memory: `256Mi`, CPU: `250m`                                       |
+| `livenessProbe.enabled`               | would you like a livessProbed to be enabled                                                             |  `true`                                                            |
```
juan131: `would` -> `Would`

bitnami/etcd/README.md
```diff
@@ -0,0 +1,178 @@
...
+| `livenessProbe.periodSeconds`         | How often to perform the probe                                                                          |  10                                                                |
+| `livenessProbe.timeoutSeconds`        | When the probe times out                                                                                |  5                                                                 |
+| `livenessProbe.failureThreshold`      | Minimum consecutive failures for the probe to be considered failed after having succeeded.              |  2                                                                 |
+| `livenessProbe.successThreshold`      | Minimum consecutive successes for the probe to be considered successful after having failed             |  1                                                                 |
+| `readinessProbe.enabled`              | would you like a readinessProbe to be enabled                                                           |  `true`                                                            |
```
juan131: `would` -> `Would`

javsalgar: I would change it to: Turn on and off readiness probe (like we have in the Redis chart)

bitnami/etcd/templates/NOTES.txt
```diff
@@ -0,0 +1,75 @@
...
+    "NodePort". As alternative, you can also switch to "usePassword=true"
+    providing a valid password on "password" parameter.
+
+-------------------------------------------------------------------------------
+{{- end }}
+{{- end }}
+
+** Please be patient while the chart is being deployed **
+
+etcd can be accessed via port 2379 on the following DNS name from within your cluster:
```
javsalgar: Here you should use the etcd port parameter that you specified in Values.yaml

bitnami/etcd/templates/NOTES.txt
```diff
@@ -0,0 +1,75 @@
...
+ * As rbac is enabled you should add the flag `-u root:$ETCD_ROOT_PASSWORD` to the etcdctl commands. Export the password with this command:
+
+    export ETCD_ROOT_PASSWORD=${kubectl get secret --namespace {{ .Release.Namespace }} {{ template "etcd.fullname" . }} -o jsonpath="{.data.etcd-root-password}" | base64 --decode)}
+
+{{ end }}
+{{- if .Values.auth.client.secureTransport }}
+{{- if .Values.auth.client.useAutoTLS }}
+ * As TLS secured transport is enabled you should add the flag `--cert-file {{ template "etcd.dataDir" . }}/fixtures/client/cert.pem --key-file {{ template "etcd.dataDir" .}}/fixtures/client/key.pem` to the etcdctl commands.
+{{- else }}
+ * As TLS secured transport is enabled you should add the flag `--cert-file /opt/bitnami/etcd/certs/client/cert.pem --key-file /opt/bitnami/etcd/certs/client/key.pem` to the etcdctl commands.
```
javsalgar: I would remove "secured transport"

bitnami/etcd/templates/statefulset.yaml
```diff
@@ -0,0 +1,357 @@
...
+        - name: peer
+          containerPort: {{ .Values.service.port }}
+        lifecycle:
+          preStop:
+            exec:
+              command:
+                - "/bin/sh"
+                - "-ec"
+                - |
+                  AUTH_OPTIONS=""
```
javsalgar: After seeing this amount of bash code in the yaml, would it make sense to move it as a template?

bitnami/etcd/templates/svc.yaml
```diff
@@ -0,0 +1,21 @@
...
+  labels:
+    app: {{ template "etcd.name" . }}
+    chart: {{ template "etcd.chart" . }}
+    release: "{{ .Release.Name }}"
+    heritage: "{{ .Release.Service }}"
+spec:
+  type: {{ .Values.service.type }}
+  ports:
+  - name: client
+    port: 2379
```
javsalgar: Shouldn't we use the values in Values.yaml here?

bitnami/etcd/templates/NOTES.txt
tompizmor: ok

bitnami/etcd/README.md
tompizmor: done

bitnami/etcd/README.md
tompizmor: done

bitnami/etcd/README.md
tompizmor: done

bitnami/etcd/templates/NOTES.txt
tompizmor: right

bitnami/etcd/templates/statefulset.yaml
tompizmor: it makes sense yes. I also thought about that as an option but I choose the bash path as it was more straightforward for me. I will try to create a template in any case

bitnami/etcd/templates/svc.yaml
tompizmor: Usually, we have this port hardcoded as it is the port the user will use to access the app and not the port where the app is running (because of the port translation).

We can re-check if this makes sense but we are doing this in all charts

PR #711, opened 2018-07-11
Metrics Server chart

carrodher:


bitnami/metrics-server/README.md
```diff
@@ -0,0 +1,89 @@
...
+
+### Configure RBAC
+
+In order to enable Role-based access control for Metrics Servier you can run the following command:
+
+```console
+$ helm install --name my-release --set rbac.create bitnami/metrics-server
+```
+
+## Persistence
```
juan131: Is this true? Does this image storage data at **/bitnami/** directory?

bitnami/metrics-server/templates/metric-server-service.yaml
```diff
@@ -0,0 +1,17 @@
...
+metadata:
+  name: {{ template "metrics-server.fullname" . }}
+  labels:
+    app: {{ template "metrics-server.name" . }}
+    chart: {{ template "metrics-server.chart" . }}
+    release: {{ .Release.Name | quote }}
+    heritage: {{ .Release.Service | quote }}
+spec:
+  ports:
+    - port: 8443
```
tompizmor: this port should 443 to the outside. Otherwise, other resources will not fetch the metrics properly. I tested with this example: https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale-walkthrough/

PR #715, opened 2018-07-19
Add new chart: ExternalDNS

juan131:
**What this PR does / why we need it:**

ExternalDNS is a Kubernetes addon that configures public DNS servers with information about exposed Kubernetes services to make them discoverable.. More information at: https://github.com/kubernetes-incubator/external-dns

Adds a new Chart that allows to deploy ExternalDNS on your k8s cluster.

### DIFF
--- a/bitnami/external-dns/README.md
+++ b/bitnami/external-dns/README.md
@@ -1,0 +1,124 @@
+# external-dns
+
+[ExternalDNS](https://github.com/kubernetes-incubator/external-dns) is a Kubernetes addon that configures public DNS servers with information about exposed Kubernetes services to make them discoverable.
+
+## TL;DR;
+
+```console
+$ helm install bitnami/external-dns
+```
+
+## Introduction
+
+This chart bootstraps a [ExternalDNS](https://github.com/bitnami/bitnami-docker-external-dns) deployment on a [Kubernetes](http://kubernetes.io) cluster using the [Helm](https://helm.sh) package manager.
+
+## Prerequisites
+
+- Kubernetes 1.4+ with Beta APIs enabled
+- PV provisioner support in the underlying infrastructure
+
+## Installing the Chart
+
+To install the chart with the release name `my-release`:
+
+```console
+$ helm install --name my-release bitnami/external-dns
+```
+
+The command deploys ExternalDNS on the Kubernetes cluster in the default configuration. The [configuration](#configuration) section lists the parameters that can be configured during installation.
+
+> **Tip**: List all releases using `helm list`
+
+## Uninstalling the Chart
+
+To uninstall/delete the `my-release` deployment:
+
+```console
+$ helm delete my-release
+```
+
+The command removes all the Kubernetes components associated with the chart and deletes the release.
+
+## Configuration
+
+The following table lists the configurable parameters of the external-dns chart and their default values.
+
+
+| Parameter                             | Description                                                                                              | Default                                                  |
+| ------------------------------------- | -------------------------------------------------------------------------------------------------------- | -------------------------------------------------------- |
+| `image.registry`                      | ExternalDNS image registry                                                                               | `docker.io`                                              |
+| `image.repository`                    | ExternalDNS Image name                                                                                   | `bitnami/external-dns`                                   |
+| `image.tag`                           | ExternalDNS Image tag                                                                                    | `{VERSION}`                                              |
+| `image.pullPolicy`                    | ExternalDNS image pull policy                                                                            | `Always`                                                 |
+| `image.pullSecrets`                   | Specify image pull secrets                                                                               | `nil` (does not add image pull secrets to deployed pods) |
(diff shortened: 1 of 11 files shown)


bitnami/external-dns/README.md
```diff
@@ -0,0 +1,124 @@
...
+
+```console
+$ helm delete my-release
+```
+
+The command removes all the Kubernetes components associated with the chart and deletes the release.
+
+## Configuration
+
+The following table lists the configurable parameters of the external-dns chart and their default values.
```
javsalgar: I do not know if it makes sense to point to a documentation page where the user can understand all the available values. For example, in the case of `policy` or `logLevel` I do not know which values I can set

bitnami/external-dns/templates/NOTES.txt
```diff
@@ -0,0 +1,45 @@
+{{- if eq .Values.provider "aws" }}
+{{- if or (empty .Values.aws.accessKey) (empty .Values.aws.secretKey) }}
+  ###############################################################################
+  ### ERROR: You chose 'aws' as your DNS provider and you did not set the     ###
+  ### AWS access&secret keys                                                  ###
+  ###############################################################################
+
+Indicate your AWS access&secret keys by installing this chart as shown below:
+
+    $ helm install --name my-release bitnami/external-dns --set provider='aws',aws.accessKey='YOUR_ACCESS_KEY',aws.secretKey='YOUR_SECRET_KEY'
```
javsalgar: I think that telling the user to upgrade the chart would be better than deploying a completely new one. That's what we do in other charts AFAIK

bitnami/external-dns/templates/deployment.yaml
```diff
@@ -0,0 +1,203 @@
...
+          timeoutSeconds: {{ .Values.readinessProbe.timeoutSeconds }}
+          successThreshold: {{ .Values.readinessProbe.successThreshold }}
+          failureThreshold: {{ .Values.readinessProbe.failureThreshold }}
+          {{- end }}
+        resources:
+{{ toYaml .Values.resources | indent 10 }}
+        volumeMounts:
+        {{- if eq .Values.provider "aws" }}
+        - name: aws-credentials
+          mountPath: /root/.aws
```
javsalgar: I imagine this is a root container, right?

bitnami/external-dns/values.yaml
```diff
@@ -0,0 +1,166 @@
+## Bitnami etcd image version
```
javsalgar: Aren't we using a values-production.yaml as well?

juan131: I'm not sure if it makes sense in this chart. I mean, I don't know if it makes sense to replicate it.

bitnami/external-dns/templates/NOTES.txt
juan131: 👍

bitnami/external-dns/README.md
juan131: We can also provide the valid values. For instance: 

```
| Verbosity of the ExternalDN logs (accepted values: X, Y, Z,...)     |
```

 What do you think?

bitnami/external-dns/templates/deployment.yaml
juan131: It's not a root container... We could use env. variable to point to the path where we mount the AWS credentials (see https://docs.aws.amazon.com/cli/latest/userguide/cli-environment.html). What do you think?

bitnami/external-dns/README.md
javsalgar: Works for me

bitnami/external-dns/templates/deployment.yaml
javsalgar: In other non root solutions we set `HOME=/`

javsalgar: But it is set in the container

bitnami/external-dns/templates/clusterrole.yaml
```diff
@@ -0,0 +1,23 @@
+{{- if .Values.rbac.create }}
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
```
tompizmor: I see that the official helm chart is parametrizing the apiVersion (https://github.com/helm/charts/blob/master/stable/external-dns/templates/clusterrole.yaml#L2)

Do you think is a good feature to add to this chart? I would say yes

bitnami/external-dns/templates/deployment.yaml
tompizmor: yes, probably the image will try to look the aws credentials at `/.aws`. I think I would make explicit in the chart where to look for the credentials using env vars as Juan suggests

javsalgar:
Closing as it is continued here: https://github.com/bitnami/charts/pull/717

PR #718, opened 2018-07-23
Adapt chart to new image solving non-root issues

carrodher:
New metrics-server images were released solving some issues with the non-root user. This PR removed some workarounds used previously.


tompizmor:
bump the chart version before merge

PR #721, opened 2018-07-24
Fix replication issue with MySQL 8. It is also compatible with 5.7

beltran-rubo:


javsalgar:
LGTM after applying @tompizmor suggestion

PR #724, opened 2018-07-25
Update dependency in node chart

javsalgar:


javsalgar:
I've also updated the node version (@jotamartos request). Hence it is a major version

PR #735, opened 2018-07-30
Add missing release and charts labels in pods

javsalgar:


bitnami/zookeeper/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 name: zookeeper
-version: 0.0.3
+version: 0.0.4
```
tompizmor: I don't see any other change in zookeeper, did you forget to commit it?

carrodher:
Can you take advantage of this PR and remove `debian-9` from `appVersion` in the applications that contains this wrong parameter?

bitnami/zookeeper/Chart.yaml
javsalgar: That didn't require changes, reverting it.

bitnami/zookeeper/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 name: zookeeper
 version: 0.0.3
```
tompizmor: now you should increase chart version because you are changing the app version

Issue #738, opened 2018-08-01
[bitnami/postgresql] feature: override configmaps


soujiro32167:
# Use case

There is a chart that requires bitnami/postgres as a dependency. I would like to pass in initialization scripts and override `postgresql.conf`. The directory structure is:
```
mychart
|- charts
|-- postgresql-2.1.0.tgz 
```

# Problem

The only way I found to do this was clone the chart into the `charts` folder and insert the files directly:

```
mychart
|- charts
|-- postgresql
|--- files
|---- postgresql.conf
|---- docker-entrypoint-initdb.d
|----- setup.sql
```

This, however, breaks the helm workflow for chart management

# Solution

A potential solution is to override the init configmaps with my own, which would include files/templates taken from the top level. Thoughts?

tompizmor:
Hi @soujiro32167,

Agree! Your solution seems valid to me. We could add a parameter in the values.yaml so that if you provide a configmap name the chart will use that instead of creating a new one.

We will work on adding this feature to the chart but I cannot provide an ETA about when this will be done.

soujiro32167:
Czech out the PR #739. Its working for us with the top chart as follows:

values.yaml
```yaml
postgresql:
  config:
    customInitConfig: true
```

templates/postgres-init-config.yaml
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: {{ .Release.Name }}-postgresql-init-scripts # the name must match the original configmap
  labels:
    app: "postgresql" # I can't get the child chart properties, which makes me a sad panda
    chart: "postgresql-2.1.1"
    release: {{ .Release.Name | quote }}
    heritage: {{ .Release.Service | quote }}
data:
  registry-init.sql: |-
    CREATE DATABASE registry;
# Or other things like globbing
```

tompizmor:
@soujiro32167 I just checked the PR and added one comment about the implementation. Let me know if it makes sense to you!

blakebyrnes:
Would love to see this merged in

tompizmor:
@blakebyrnes check my comment in the related PR: https://github.com/bitnami/charts/pull/739#issuecomment-431864502

PR #739, opened 2018-08-01
issue #738: added ability to override configmaps

soujiro32167:
This is a simple, if inelegant fix for #738

### DIFF
--- a/bitnami/postgresql/values.yaml
+++ b/bitnami/postgresql/values.yaml
@@ -156,6 +156,7 @@ metrics:
 
 ## For use by charts who depend on postrges
 ## Allows the depending chart to inject its own init scripts and postgres.conf by overriding the configmaps
-# config:
-#   customInitConfig: false
-#   customConfig: false
+config:
+  enabled: false
+#   customInitConfigName: my-init-configmap
+#   customConfigName: my-configmap
--- a/bitnami/postgresql/values.yaml
+++ b/bitnami/postgresql/values.yaml
@@ -153,3 +153,9 @@ metrics:
     timeoutSeconds: 5
     failureThreshold: 6
     successThreshold: 1
+
+## For use by charts who depend on postrges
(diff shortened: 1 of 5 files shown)


bitnami/postgresql/values.yaml
```diff
@@ -153,3 +153,9 @@ metrics:
     timeoutSeconds: 5
     failureThreshold: 6
     successThreshold: 1
+
+## For use by charts who depend on postrges
+## Allows the depending chart to inject its own init scripts and postgres.conf by overriding the configmaps
```
tompizmor: Instead, I would ask for the config map name to be used. That way you don't need to create the config map with the same name (if you don't give a name while installing you don't really know the name in advance).

```
config:
    customInitConfigMapName:
    customConfigMapName:
```

So if you provide a configMapName it will use that, otherwise, it will create one.

tompizmor: Of course, this change requires to update the logic in the other files too

soujiro32167: That makes sense

soujiro32167:
@tompizmor I've addressed your comment

tompizmor:
Thanks for updating the PR. This is definitely a feature we must add.

However, I am not sure if we should merge this feature in the Bitnami chart or wait until the Bitnami chart became the official helm chart (there is a PR about to be merged for this https://github.com/helm/charts/pull/8004)

cc/ @carrodher

carrodher:
Hi, thanks for the PR!!
As @tompizmor rightly said, the PostgreSQL chart was moved to the official Helm repository (https://github.com/helm/charts) some weeks ago, now you can take a look at the chart in the following link: https://github.com/helm/charts/blob/master/stable/postgresql

The chart suffered numerous changes and a great refactor, would it make sense to apply these changes there? If yes, could you create the PR in the new location? Feel free to point to this one if needed

PR #750, opened 2018-08-14
[Kafka] - Fix issue with maxMessageBytes default value

tompizmor:
There is an issue in helm that converts large numbers in values.yaml to scientific notation when passing the value to the docker image. So Kafka is configured with a numeric value that it cannot understand and so fails.

https://github.com/helm/helm/issues/3001
https://github.com/helm/helm/issues/1707


javsalgar:
LGTM

PR #770, opened 2018-09-03
Nginx Ingress Controller chart

carrodher:


bitnami/nginx-ingress-controller/templates/controller-daemonset.yaml
```diff
@@ -0,0 +1,205 @@
...
+            {{- end }}
+          {{- end }}
+          {{- if (semverCompare ">=0.16.0" .Values.image.tag) }}
+          securityContext:
+            capabilities:
+                drop:
+                - ALL
+                add:
+                - NET_BIND_SERVICE
+            # www-data -> 33
```
juan131: Are we sure we want to use this user? We're using 1001 on the image:

https://github.com/bitnami/bitnami-docker-nginx-ingress-controller/blob/master/0/debian-9/Dockerfile

bitnami/nginx-ingress-controller/templates/controller-deployment.yaml
```diff
@@ -0,0 +1,199 @@
...
+          {{- end }}
+          {{- if (semverCompare ">=0.16.0" .Values.image.tag) }}
+          securityContext:
+            capabilities:
+                drop:
+                - ALL
+                add:
+                - NET_BIND_SERVICE
+            # www-data -> 33
+            runAsUser: 33
```
juan131: Same concern about using the user 33 instead of `1001`

bitnami/nginx-ingress-controller/templates/controller-daemonset.yaml
juan131: We normally allow the user to choose the user, check https://github.com/bitnami/charts/blob/master/bitnami/etcd/values.yaml#L90

bitnami/nginx-ingress-controller/values.yaml
```diff
@@ -0,0 +1,382 @@
+## Bitnami NGINX image version
```
juan131: `Bitnami NGINX image` -> `Bitnami NGINX Ingress controller Image`

bitnami/nginx-ingress-controller/values.yaml
```diff
@@ -0,0 +1,382 @@
...
+## ref: https://hub.docker.com/r/bitnami/nginx/tags/
```
juan131: `ref: https://hub.docker.com/r/bitnami/nginx/tags/` -> `ref: https://hub.docker.com/r/bitnami/nginx-ingress-controller/tags/`

bitnami/nginx-ingress-controller/templates/controller-daemonset.yaml
```diff
@@ -0,0 +1,205 @@
...
+      {{- if .Values.imagePullSecrets }}
+      imagePullSecrets:
+{{ toYaml .Values.imagePullSecrets | indent 8 }}
+      {{- end }}
+{{- if .Values.priorityClassName }}
+      priorityClassName: "{{ .Values.priorityClassName }}"
+{{- end }}
+      containers:
+        - name: {{ template "nginx-ingress.name" . }}-{{ .Values.name }}
+          image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
```
tompizmor: you are not using the `image.registry` property. Add a helper to define the image name as we already have in other charts.

bitnami/nginx-ingress-controller/templates/controller-deployment.yaml
```diff
@@ -0,0 +1,199 @@
...
+      {{- if .Values.imagePullSecrets }}
+      imagePullSecrets:
+{{ toYaml .Values.imagePullSecrets | indent 8 }}
+      {{- end }}
+{{- if .Values.priorityClassName }}
+      priorityClassName: "{{ .Values.priorityClassName }}"
+{{- end }}
+      containers:
+        - name: {{ template "nginx-ingress.name" . }}-{{ .Values.name }}
+          image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
```
tompizmor: same comment about the docker image and the helper

PR #772, opened 2018-09-04
[bitnami/postgresql] fix non-string env values

akhy:
Env values must be string


carrodher:
Hi, thanks for the PR!!
The PostgreSQL chart was moved to the official Helm repository (https://github.com/helm/charts) some weeks ago, now you can take a look to the chart in the following link
https://github.com/helm/charts/blob/master/stable/postgresql

Regarding your changes, as you can see [here](https://github.com/helm/charts/blob/master/stable/postgresql/templates/statefulset-slaves.yaml#L85), your suggestions was applied. Feel free to contribute to this chart in the new location

PR #774, opened 2018-09-05
[bitnami/mysql] Improve Helpers functions to handle image names

juan131:
**What this PR does / why we need it:**

This PR improves the image naming functions used on _helpers. I found some weird behaviour with the current ones and I decided to update them with the same workflow used on latest charts.

It also fix the **values-production.yaml** images pull policy.

### DIFF
--- a/bitnami/mysql/templates/_helpers.tpl
+++ b/bitnami/mysql/templates/_helpers.tpl
@@ -29,7 +29,7 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this 
 
 
 {{/*
-Return the proper image name
+Return the proper MySQL image name
 */}}
 {{- define "mysql.image" -}}
 {{- $registryName :=  .Values.image.registry -}}
@@ -39,7 +39,7 @@ Return the proper image name
 {{- end -}}
 
 {{/*
-Return the proper image name
+Return the proper MySQL metrics exporter image name
 */}}
 {{- define "metrics.image" -}}
 {{- $registryName :=  .Values.metrics.image.registry -}}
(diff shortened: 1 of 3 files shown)


bitnami/mysql/templates/_helpers.tpl
```diff
@@ -27,20 +27,24 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this
 {{- printf "%s-%s" .Chart.Name .Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-" -}}
 {{- end }}
 
+
 {{/*
-Return the proper MySQL image name
```
tompizmor: I would keep that is the MySQL image

bitnami/mysql/templates/_helpers.tpl
```diff
@@ -27,20 +27,24 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this
...
-{{- printf "%s/%s:%s" $registryName .Values.image.repository $tag -}}
-{{- end }}
+{{- $registryName :=  .Values.image.registry -}}
+{{- $repositoryName := .Values.image.repository -}}
+{{- $tag := .Values.image.tag | toString -}}
+{{- printf "%s/%s:%s" $registryName $repositoryName $tag -}}
+{{- end -}}
 
 {{/*
-Return the proper MySQL metrics exporter image name
```
tompizmor: I would keep is the metrics exporter image

bitnami/mysql/values-production.yaml
```diff
@@ -9,7 +9,7 @@ image:
   ## Defaults to 'Always' if image tag is 'latest', else set to 'IfNotPresent'
   ## ref: http://kubernetes.io/docs/user-guide/images/#pre-pulling-images
   ##
-  pullPolicy: Always
+  pullPolicy: IfNotPresent
```
tompizmor: why? The tag used is still a rolling tag no?

juan131: These **values** are intended for production. You can check other **values-production.yaml* files (Redis, MariaDB, etc)

Issue #784, opened 2018-09-12
MySQL chart existing volume claim flag missing


pentago:
Just noticed that even though at the end of the README it's said that existing volume claim can be used, I do not see **master.persistence.existingClaim** config existing there in a same way as in MariaDB chart. Is this the issue with chart description or this functionality hasn't been implemented yet?

beltran-rubo:
Hi @pentago 

It seems the container functionality is there but we should add that parameter to the Chart files. We will look into it.

pentago:
Thanks a ton @beltran-rubo !

carrodher:
@pentago Thanks for letting us know!
I just implemented your suggestion (you can see the changes in the above PR).
If you find something weird do not hesitate to let us know.

pentago:
Thanks a ton @carrodher. I'll upgrade later on and get back to you if anything weird occurs.

pentago:
Just tried installing fresh chart and first installation works great, no problems whatsoever.
To test successfulness of the patch I performed ```helm del --purge mysql```  to delete deployment and tried installing it again to check if PVC is picked up. This time, chart installs but causes _CrashLoopBackOff_ with _exit code  1_.

Following is error log:

> mysqld_safe error: log-error set to '/opt/bitnami/mysql/logs/mysqld.log', however file don't exists. Create writable for user 'mysql'.

Same issue occurs even if I do not use existing PVC.

With stable/mariadb chart there's no such error and deployment works and mounts existing data with existing PVC I specify in custom values.yaml

values.yaml:
```
image:
  tag: 8.0.12-debian-9-r25

root:
  password: password

db:
  user: ''
  password: ''
  name: ''

replication:
  enabled: true
  user: replicator
  password: password

master:
  persistence:
    enabled: true
#    existingClaim: mysql-master-storage
#    mountPath: /bitnami
    storageClass: ssd
    accessModes:
    - ReadWriteOnce
    size: 100Gi
  resources:
    requests:
      cpu: 2
      memory: 3Gi
    limits:
      cpu: 2.5
      memory: 4Gi

slave:
  replicas: 2
  persistence:
    enabled: true
    storageClass: ssd
    accessModes:
    - ReadWriteOnce
    size: 100Gi
  resources:
    requests:
      cpu: 2
      memory: 3Gi
    limits:
      cpu: 2.5
      memory: 4Gi

metrics:
  enabled: true
  resources:
    requests:
      cpu: 0.3
      memory: 100Mi
    limits:
[...]

PR #787, opened 2018-09-14
Add existing volume claim logic to MySQL

carrodher:
As requested on #784 I added existing volume claim logic to MySQL chart as we are doing currently in the upstream MariaDB chart


tompizmor:
LGTM!

PR #790, opened 2018-09-15
[bitnami/elasticsearch] Fix custom config file is not applied

zerda:
Per [libelasticsearch.sh](https://github.com/bitnami/bitnami-docker-elasticsearch/blob/6.4.0-debian-9-r28/6/debian-9/rootfs/libelasticsearch.sh), the configuration directory is specified as `/opt/bitnami/elasticsearch/config` in docker.

The custom config file should be there too.

### DIFF
--- a/bitnami/elasticsearch/Chart.yaml
+++ b/bitnami/elasticsearch/Chart.yaml
@@ -1,6 +1,6 @@
 name: elasticsearch
 version: 2.0.4
-appVersion: 6.4.1
+appVersion: 6.4.2
 description: A highly scalable open-source full-text search and analytics engine
 keywords:
 - elasticsearch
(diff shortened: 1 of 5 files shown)


XBeg9:
Agree, helm is broken and zen discovery doesn't work without this changes.

XBeg9:
@tompizmor latest version of ElasticSearch is 6.4.1, you did merge `appVersion: 6.4.2`. Not sure if this can break anything, but it looks like it's wrong.

tompizmor:
You're right! I bumped the wrong version in the chart. Fixed here: https://github.com/bitnami/charts/commit/d8a44261f60fffbebf98b732a30ad74c7e2be409

XBeg9:
> You're right! I bumped the wrong version in the chart. Fixed here: [d8a4426](https://github.com/bitnami/charts/commit/d8a44261f60fffbebf98b732a30ad74c7e2be409)

No problem, thanks for fix!

Issue #791, opened 2018-09-16
ETCD statefulset failing to scale - error #0: dial tcp: lookup lumpy-tarsier-etcd-0 on 10.96.0.10:53: server misbehaving


fmrtl73:
When trying to scale the statefulset to 2 from 1 the container logs reports:
```
==> The ID of the host is 1
==> Creating data dir...
==> Adding member to existing cluster.
==> Adding new member
client: etcd cluster is unavailable or misconfigured; error #0: dial tcp: lookup lumpy-tarsier-etcd-0 on 10.96.0.10:53: server misbehaving
```

10.96.0.10:53 is my Kube DNS server

When I do an nslookup with lumpy-tarsier-etcd-0 it doesn't resolve but when I do an nslookup with lumpy-tarsier-etcd-0.lumpy-tarsier-etcd-headless.default.svc.cluster.local it works fine.

tompizmor:
hi @fmrtl73,

I was able to reproduce the issue. We will take a look to the chart to see what is happening. 
In the meantime, you can install the chart directly with 3 replicas using this command:

```
helm install bitnami/etcd --statefulset.replicaCount=3
```

tompizmor:
Hi @fmrtl73,

I just merged a PR that fixes the issue. Please let us know if it works for you or not.
Remember to use the latest version of the chart by running `helm repo update`

PR #820, opened 2018-09-24
[bitnami/etcd] Fix scaling etcd nodes

tompizmor:
Fixes https://github.com/bitnami/charts/issues/791.

The main issue was loading the env-vars for the new cluster state after add a new node.

### DIFF
--- a/bitnami/etcd/templates/statefulset.yaml
+++ b/bitnami/etcd/templates/statefulset.yaml
@@ -61,6 +61,7 @@ spec:
       {{- $clientPort := int .Values.service.port }}
       {{- $peerPort := int .Values.service.peerPort }}
       {{- $etcdFullname := include "etcd.fullname" . }}
+      {{- $releaseNamespace := .Release.Namespace }}
       {{- $etcdHeadlessServiceName := printf "%s-%s" $etcdFullname "headless" }}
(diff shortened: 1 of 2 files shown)


bitnami/etcd/templates/statefulset.yaml
```diff
@@ -164,11 +166,11 @@ spec:
         - name: ETCD_DATA_DIR
           value: {{ template "etcd.dataDir" . }}
         - name: ETCD_ADVERTISE_CLIENT_URLS
-          value: "{{ $etcdClientProtocol }}://$(MY_POD_NAME):{{ .Values.service.port }}"
+          value: "{{ $etcdClientProtocol }}://$(MY_POD_NAME).{{ $etcdHeadlessServiceName }}.{{ .Release.Namespace }}.svc.cluster.local:{{ .Values.service.port }}"
```
juan131: Why do you still need `$(MY_POD_NAME)`?

javsalgar: I suppose because it is using headless services

tompizmor: because it is different for each replica

etcd-0, etcd-1, etcd-2,...

juan131: 👍

PR #821, opened 2018-09-25
Add new Jenkins env variables

carrodher:


bitnami/jenkins/templates/deployment.yaml
```diff
@@ -31,12 +31,18 @@ spec:
...
         - name: JENKINS_PASSWORD
           valueFrom:
             secretKeyRef:
               name: {{ template "fullname" . }}
               key: jenkins-password
+        - name: JENKINS_HOME
+          value: {{ .Values.jenkinsHome | quote }}
+        - name: DISABLE_JENKINS_INITIALIZATION
+          value: {{ .Values.disableInitialization | quote }}
+        - name: JAVA_OPTS
```
juan131: Could you check whether `.Values.javaOpts ` is defined and only set *JAVA_OPTS* when it is?

PR #823, opened 2018-09-25
[bitnami/mysql] Use Existing PVCs to store MySQL data

juan131:

Currently we're creating a PVC even if the user indicates its own existing PVC. Furthermore, the one which is dynamically created is being mounted at **/bitnami/data**.

This is a very weird scenario since the user who already has a PVC (with his corresponding PV) can't use it to store the MySQL data.

#### Which issue this PR fixes

fixes https://github.com/bitnami/charts/issues/784

### DIFF
--- a/bitnami/mysql/values-production.yaml
+++ b/bitnami/mysql/values-production.yaml
@@ -85,10 +85,11 @@ master:
     ## If true, use a Persistent Volume Claim, If false, use emptyDir
     ##
     enabled: true
+    mountPath: /bitnami/mysql
     ## Enable persistence using an existing PVC
     ##
     # existingClaim:
-    # mountPath:
+
     ## Persistent Volume Storage Class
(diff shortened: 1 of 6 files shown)


pentago:
From what I saw, this should work.. Thanks @juan131

bitnami/mysql/values-production.yaml
```diff
@@ -85,10 +85,11 @@ master:
     ## If true, use a Persistent Volume Claim, If false, use emptyDir
     ##
     enabled: true
+    mountPath: /bitnami/mysql
```
javsalgar: The reasoning for setting the mountPath here is for flexibility with other images, right?

juan131: Yes... I mean, the user can also use its own configuration file using a different path for data. Therefore, the mountPath for the data volume should be flexible.

PR #844, opened 2018-10-03
Update values.yaml

yanivoliver:
As per the chart documentation, RBAC should not be enabled by default (especially if there is no password set by default)


tompizmor:
Hi @yanivoliver,

I think it is good to have the RBAC authentication enabled by default. If no password is set, a random one will be generated and instructions about how to obtain will be displayed after installing the chart.

So instead of disabling the RBAC authentication, I would just fix the README.

Also, take into account that any change in the chart requires to update the chart version in the Chart.yaml file.

carrodher:
Hi @yanivoliver, I just created https://github.com/bitnami/charts/pull/944 applying @tompizmor suggestions. Thanks for pointing us in the right direction

PR #848, opened 2018-10-04
[postgresql] Change name truncation to 63 instead of 24 fixes #780

### LINKED ISSUE #780: [bitnami/postgres] svc name truncates to 24
can someone confirm if this is still necessary? 

deploying the chart as a subchart means that svc name gets completely foobarred as it will pick up the release name, tag on `postgres` then truncate to 24 characters.

e.g.
`hhs-feature-devops-64-en` when it was supposed to be `hhs-feature-devops-64-env-postgres`.

The only workaround I have (which is a horrible one) is all my charts have to implement the same truncation.

original issue: https://github.com/kubernetes/kubernetes/pull/25041

gtseres:


gtseres:
@carrodher thanks a lot! If I'm not mistaken, this is considered more like a fix/minor enhancement, that's why I bumped the patch version. Please let me know if this is not the case. Thanks! 👍

carrodher:
Perfect! Thanks for the collaboration 👏

PR #858, opened 2018-10-08
[bitnami/memcached] - Memcached security context

tompizmor:
Add securityContext feature to memcached helm chart.


tompizmor:
you're welcome! 😄

PR #864, opened 2018-10-11
Add global registry option

carrodher:
- Add `global.registry` at the top of `values.yaml` (commented out at the beginning).
- Add the required logic to use the global value if it is defined.
- Add a link to the `values.yaml` of Zookeeper in the Kafka `values.yaml` chart (in the section of the Zookeeper options).


carrodher:
Closing this POC to send a PR with the production changes in all charts https://github.com/bitnami/charts/pull/865

PR #880, opened 2018-10-16
[bitnami/tensorflow-inception] Simplify chart deleting unnecessary pvc

tompizmor:
We were using a PVC to store the model data another one just to persist a TensorFlow configuration file.

It does not make sense to persists the configuration file as it is regenerated in each initialization so we  can save resources deleting this configuration.


bitnami/tensorflow-inception/templates/deployment.yaml
```diff
@@ -43,16 +43,7 @@ spec:
         volumeMounts:
         - name: seed
           mountPath: "/bitnami/model-data"
-        - name: data
```
juan131: Why are you removing the data persistence?

bitnami/tensorflow-inception/templates/deployment.yaml
```diff
@@ -43,16 +43,7 @@ spec:
...
         persistentVolumeClaim:
-          claimName: {{ template "fullname" . }}-seed-inception
-      - name: data
-      {{- if .Values.persistence.enabled }}
-        persistentVolumeClaim:
-          claimName: {{ template "fullname" . }}
-      {{- else }}
-        emptyDir: {}
-      {{- end -}}
+          claimName: {{ template "fullname" . }}-seed-inception
```
juan131: Are you requesting a claim on every scenario? That's not what we did on other apps

bitnami/tensorflow-inception/templates/deployment.yaml
juan131: 👍

bitnami/tensorflow-inception/templates/deployment.yaml
juan131: 👍

PR #895, opened 2018-10-25
 Replace busybox image by minideb

carrodher:
- Replace busybox image by minideb
- Adapt commands to install the package before running it
- Comment and Document ES initContainer modifying kernel settings


bitnami/elasticsearch/templates/master-deploy.yaml
```diff
@@ -56,13 +56,14 @@ spec:
...
       {{- end}}
       {{- end }}
-      initContainers:
-      - name: sysctl
-        image: {{ template "sysctl.image" . }}
-        imagePullPolicy: {{ .Values.image.pullPolicy | quote }}
-        command: ["sysctl", "-w", "vm.max_map_count=262144"]
-        securityContext:
-          privileged: true
+      ## Image that performs the sysctl operation to modify Kernel settings (needed sometimes to avoid boot errors)
```
javsalgar: I wouldn't comment it. I would put it as an option in the values

beltran-rubo: +1

PR #922, opened 2018-11-12
[bitnami/cassandra] Add more customization settings

javsalgar:
- Set endpoint snitch
- Set rack name
- Set datacenter name
- Set JAVA jvm settings
- Enable rpc


bitnami/cassandra/templates/statefulset.yaml
```diff
@@ -116,8 +142,10 @@ spec:
           containerPort: 7199
         - name: cql
           containerPort: 9042
+{{- if .Values.cluster.enableRPC }}
```
juan131: Why this indentation?

bitnami/cassandra/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 name: cassandra
-version: 0.0.1
+version: 0.0.2
```
juan131: Increase minor version

PR #933, opened 2018-11-20
[metrics-server] Adding ability to pull docker image from private registry

pmontanari:
Signed-off-by: Patrick Montanari <patrick.montanari@gmail.com>


bitnami/metrics-server/README.md
```diff
@@ -52,6 +52,7 @@ The following tables lists the configurable parameters of the Metrics Server cha
 | `image.repository`       | Metrics Server image name                                                   | `bitnami/metrics-server`               |
 | `image.tag`              | Metrics Server image tag                                                    | `{VERSION}`                            |
 | `image.pullPolicy`       | Metrics Server image pull policy                                            | `Always`                               |
+| `image.imagePullSecrets` | Reference to secrets to be used when pulling images                         | []                                 |
```
carrodher: As we are under the `image` level, `imagePullSecrets` sounds redundant, can you replace it with `pullSecrets`?

```suggestion
| `image.pullSecrets`                  | Specify image pull secrets                                 | `nil`                                         |
```

bitnami/metrics-server/values.yaml
```diff
@@ -17,6 +17,10 @@ image:
   ##
   pullPolicy: Always
 
+  # For supporting pulling from private registries
+  imagePullSecrets: []
+  # - name: "image-pull-secret"
```
carrodher: For other solutions we are using something like:
```yaml
  ##
  # pullSecrets:
  #   - myRegistrKeySecretName
```

We've the option commented and we added some documentation

bitnami/metrics-server/templates/serviceaccount.yaml
```diff
@@ -8,4 +8,6 @@ metadata:
     chart: {{ template "metrics-server.chart" . }}
     release: {{ .Release.Name | quote }}
     heritage: {{ .Release.Service | quote }}
+imagePullSecrets:
```
carrodher: First we check if the value exists and after that we read it
```yaml
      {{- if .Values.image.pullSecrets }}
      imagePullSecrets:
      {{- range .Values.image.pullSecrets }}
        - name: {{ . }}
      {{- end}}
      {{- end }}
```

bitnami/metrics-server/README.md
pmontanari: well... I never saw the abreviation in other charts.. .Not sure it is very clear

carrodher:
Before merge the changes, can you take a look at the above suggestions?

pmontanari:
> Before merge the changes, can you take a look at the above suggestions?

Agree

PR #963, opened 2018-12-06
Fix indentation in external-dns deployment

ajmadsen:
**Description of the change**

This change fixes an indentation issue in one of the templates of the `external-dns` chart. It caused any `extraArgs` to end up joined to the previous command line arg, causing interesting results.


juan131:
Hi @ajmadsen 

ajmadsen:
Thank you for the direction. I've bumped the patch version for the chart per the guidelines.

PR #965, opened 2018-12-06
[bitnami/external-dns] only set google specific config if they are not empty

Starefossen:
**Description of the change**

This makes all google specific provider configurations optional if they are empty `""`. This is in order to make it work on Google Kubernetes Engine (GKE) which, if configured correctly, does not require a separate service account. 

Source: https://github.com/kubernetes-incubator/external-dns/blob/master/docs/tutorials/gke.md

**Benefits**

This will make the helm install work out of the box on GKE clusters without having to create a dedicated service account.

**Possible drawbacks**

This should be fully backward compatible.

**Applicable issues**

**Additional information**

### DIFF
--- a/bitnami/external-dns/templates/deployment.yaml
+++ b/bitnami/external-dns/templates/deployment.yaml
@@ -106,7 +106,9 @@ spec:
         {{- end }}
 
         {{- if eq .Values.provider "google" }}
+        {{- if .Values.google.project }}
         - --google-project={{ .Values.google.project }}
+        {{- end }}
         {{- end }}
 
         env:
@@ -147,8 +149,10 @@ spec:
         {{- end }}
 
         {{- if eq .Values.provider "google" }}
+        {{- if .Values.google.serviceAccountSecret }}
         - name: GOOGLE_APPLICATION_CREDENTIALS
           value: /etc/secrets/service-account/credentials.json
+        {{- end }}
         {{- end }}
 
         ports:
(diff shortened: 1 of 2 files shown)


Starefossen:
Ping @juan131 or @carrodher 🙂

carrodher:

bitnami/external-dns/templates/deployment.yaml
```diff
@@ -106,8 +106,10 @@ spec:
         {{- end }}
 
         {{- if eq .Values.provider "google" }}
+        {{- if ne .Values.google.project "" }}
```
carrodher: Not tested, just a suggestion, would the solution work by grouping both `if` into one?
```suggestion
        {{- if and (eq .Values.provider "google") (ne .Values.google.project "") -}}
```

If yes, can you apply it for everything and remove the unnecessaries final `{{- end }}`?

Starefossen: Good suggestion @carrodher, however, if we want to do it like this I suggest we do it consistent across the entire chart since the double if switch seems to be the "preferred" way so far.

Starefossen:
Updated the Chart.yaml version 🙂

bitnami/external-dns/templates/deployment.yaml
juan131: It depends. If I'm not wrong **Go Templates** evaluates both conditions even if the 1st one is false when using a single `if`.
Therefore, there are cases where it's better to use a "double if"

PR #982, opened 2018-12-17
Fix <nil> jvmFlags

mkmik:
**Description of the change**

Fixes a bug in the chart what causes `"<nil>"` to be passed as the default value of the JVMFLAGS env

**Benefits**

Now the chart works without having to override the `jvmFlags` variable.

**Possible drawbacks**

No idea


mkmik:
@tompizmor beat me with #981

Issue #984, opened 2018-12-17
GKE data nodes failing to join cluster with masters - failed send ping


hawksight:
**Which chart**:

bitnami/elasticsearch --version 4.2.5

**Description**

After initialising a clean cluster on GKE (v1.10.9-gke.7) installing helm with tiller in the `tiller` namespace. The data nodes seem to error constantly on startup with the following error:

```
[2018-12-17T16:22:47,587][WARN ][o.e.d.z.UnicastZenPing   ] [staging-elasticsearch-data-2] [35] failed send ping to {10.44.3.15:9300}{NBRfGnyIS--x1dOB6asT-g}{staging-elasticsearch-discovery}{10.44.3.15:9300}
java.lang.IllegalStateException: handshake failed, mismatched cluster name [Cluster [<nil>]] - {10.44.3.15:9300}{NBRfGnyIS--x1dOB6asT-g}{staging-elasticsearch-discovery}{10.44.3.15:9300}
	at org.elasticsearch.transport.TransportService.handshake(TransportService.java:448) ~[elasticsearch-6.5.2.jar:6.5.2]
	at org.elasticsearch.transport.TransportService.handshake(TransportService.java:412) ~[elasticsearch-6.5.2.jar:6.5.2]
	at org.elasticsearch.discovery.zen.UnicastZenPing$PingingRound.getOrConnect(UnicastZenPing.java:366) ~[elasticsearch-6.5.2.jar:6.5.2]
	at org.elasticsearch.discovery.zen.UnicastZenPing$3.doRun(UnicastZenPing.java:471) [elasticsearch-6.5.2.jar:6.5.2]
	at org.elasticsearch.common.util.concurrent.ThreadContext$ContextPreservingAbstractRunnable.doRun(ThreadContext.java:723) [elasticsearch-6.5.2.jar:6.5.2]
	at org.elasticsearch.common.util.concurrent.AbstractRunnable.run(AbstractRunnable.java:37) [elasticsearch-6.5.2.jar:6.5.2]
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149) [?:1.8.0_191]
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624) [?:1.8.0_191]
	at java.lang.Thread.run(Thread.java:748) [?:1.8.0_191]
```

I deployed a quick alpine debugging pod to check the ports:
```
[...]

PR #985, opened 2018-12-17
[elasticseach] - Correct empty plugins case

hawksight:
**Description of the change**

Move end of if function so empty `.Values.plugins` does not overwrite the cluster name with `nil` and cause cluster connectivity issues between master deployment & the rest of the elasticsearch components.

**Benefits**

Elasticsearch deployment will work when no plugins are provided

**Possible drawbacks**

None that I'm aware.

**Applicable issues**

Possibly resolving #984 

**Additional information**

Testing an upgrade `--dry-run` it appears the value of `ELASTICSEARCH_CLUSTER_NAME` is no longer `"<nil>"`.


juan131:
LGTM!! Thanks so much for the fix!!

PR #1006, opened 2018-12-28
[bitnami/etcd] start temporary etcd single node cluster for root password setting when start cluster with node number >1 and auth.rbac.rootPassword

Kuleshov-Arsentii:
When start with replicas >1, etcd wait for all nodes to elect leader, and don't accept clients.
So, first node can't set root password:

```
==> The ID of the host is 0
==> Creating data dir...
==> There is no data at all. Creating new cluster
==> Configuring RBAC authentication!
stty: 'standard input': Inappropriate ioctl for device
stty: 'standard input': Inappropriate ioctl for device
New password: 
client: etcd cluster is unavailable or misconfigured; error #0: client: endpoint http://etcd-0.etcd-headless.default.svc.cluster.local:2379 exceeded header timeout
```

### DIFF
--- a/bitnami/etcd/Chart.yaml
+++ b/bitnami/etcd/Chart.yaml
@@ -1,5 +1,5 @@
 name: etcd
-version: 1.4.0
+version: 1.4.1
 appVersion: 3.3.10
 description: etcd is a distributed key value store that provides a reliable way to store data across a cluster of machines
 keywords:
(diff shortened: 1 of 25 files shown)


tompizmor:
@Kuleshov-Arsentii Thanks for the fix!

Can you please increase the version field in the Chart.yaml file to `1.4.1`?

Kuleshov-Arsentii:
> @Kuleshov-Arsentii Thanks for the fix!
> 
> Can you please increase the version field in the Chart.yaml file to `1.4.1`?

It is done.

tompizmor:
Thanks! Merged!

PR #1014, opened 2019-01-06
etcd: Allow non-default service domains

bryanalves:
**Description of the change**
Provide a new configuration value -- `service.dnsBase` -- with a
sensible default that allows users to run this chart on clusters whose
internal service DNS domain for services is something other than
`svc.cluster.local`

**Benefits**

Allow clusters with non-standard domains to deploy the etcd chart

**Possible drawbacks**

The default setting should make this PR be effectively a no-op for clusters who don't have a need for this.

**Applicable issues**

**Additional information**


tompizmor:
@bryanalves  Thanks for updating the PR!

PR #1023, opened 2019-01-15
Document the correct format for pullSecrets in READMEs

carrodher:
`XXX.pullSecrets` should be an array of secrets, this PR add some comments in the README to clarify it.
Basically, it replaces different ways of
```
| `image.pullSecrets` | Specify image pull secrets | `nil` |
```
by
```
| `image.pullSecrets` | Specify docker-registry secret names as an array | `[]` (does not add image pull secrets to deployed pods) |
```
Fixes: https://community.bitnami.com/t/helm-install-error-with-pullsecrets/63753


jotadrilo:
this changes the `appVersion` field of some _Chart.yaml_ files, which is wrong

PR #1042, opened 2019-02-04
 [bitnami/node, bitnami/mean] Enable securityContexts

javsalgar:
**Description of the change**
Enable security contexts so the charts are compatible with OpenShift

**Benefits**

**Possible drawbacks**

**Applicable issues**

**Additional information**

### DIFF
--- a/bitnami/mean/requirements.yaml
+++ b/bitnami/mean/requirements.yaml
@@ -3,6 +3,3 @@ dependencies:
   repository: https://kubernetes-charts.storage.googleapis.com/
   version: 4.x.x
   condition: mongodb.install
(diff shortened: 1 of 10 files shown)


bitnami/mean/requirements.yaml
```diff
@@ -3,6 +3,3 @@ dependencies:
   repository: https://kubernetes-charts.storage.googleapis.com/
   version: 4.x.x
   condition: mongodb.install
-- name: bitnami-common
```
tompizmor: why are you deleting this? This chart is required if we want to deploy cloud resources using k8s service-catalog

Issue #1034, opened 2019-01-31
Can not deploy the charts in kuberente v1.11


huikang:
**Steps to reproduce the issue:**

1.  `helm install bitnami/etcd --set persistence.enabled=false`

**Describe the results you received:**
```
 kubectl get pod
NAME                        READY     STATUS             RESTARTS   AGE
vigilant-bumblebee-etcd-0   0/1       CrashLoopBackOff   5          5m
```
```
kubectl logs -f vigilant-bumblebee-etcd-0
==> The ID of the host is 0
==> Creating data dir...
==> There is no data at all. Creating new cluster
==> Configuring RBAC authentication!
stty: 'standard input': Inappropriate ioctl for device
stty: 'standard input': Inappropriate ioctl for device
New password: 


client: etcd cluster is unavailable or misconfigured; error #0: client: endpoint http://vigilant-bumblebee-etcd-0.vigilant-bumblebee-etcd-headless.default.svc.cluster.local:2379 exceeded header timeout
```

**Describe the results you expected:**

**Additional information you deem important (e.g. issue happens only occasionally):**

**Version of Helm and Kubernetes**:

- Output of `helm version`:
```
Client: &version.Version{SemVer:"v2.11.0", GitCommit:"2e55dbe1fdb5fdb96b75ff144a339489417b146b", GitTreeState:"clean"}
Server: &version.Version{SemVer:"v2.11.0", GitCommit:"2e55dbe1fdb5fdb96b75ff144a339489417b146b", GitTreeState:"clean"}
```

- Output of `kubectl version`:
```
Client Version: version.Info{Major:"1", Minor:"11", GitVersion:"v1.11.3", GitCommit:"a4529464e4629c21224b3d52edfe0ea91b072862", GitTreeState:"clean", BuildDate:"2018-09-09T18:02:47Z", GoVersion:"go1.10.3", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"11", GitVersion:"v1.11.6+IKS", GitCommit:"002d263ed027db260968616b951fb46f2bab9bb1", GitTreeState:"clean", BuildDate:"2019-01-09T08:07:22Z", GoVersion:"go1.10.3", Compiler:"gc", Platform:"linux/amd64"}
```

carrodher:
What is the output of the `helm install` command?
```bash
$ helm install bitnami/etcd --set persistence.enabled=false
NAME:   existing-pig
[...]

PR #1043, opened 2019-02-05
[bitnami/etcd] Wait for cluster response before setting authentication

krwrang:
**Description of the change**
Instead of a fixed 5 second sleep, wait until the cluster is ready to answer to etcdctl
**Benefits**
When adding to root user via etcdctl, if a multi-node cluster does not have atleast two live nodes, the step will fail with `exceeded header timeout`.
This can happen in multiple cases:

1. Readiness probe enabled
2. Using persistence with a storage class that takes some time to create volumes

**Possible drawbacks**
No known limitations, creation tested with:

- Single and multi-node clusters
- With and without persistence

**Applicable issues**
#1034 
**Additional information**


tompizmor:
Thanks for the PR @krwrang!

Issue #1032, opened 2019-01-28
Add affinity rules setting ability to bitnami/elasticsearch


Troyhy:
**bitnami/elasticsearch**:

**Description**

I'm not able to select nodes witch elasticsearch pods are created.

rafariossaa:
Hi @Troyhy,
  Could you provide more information of what are you trying to do and the errors you get?

thanks forehand

Troyhy:
Hi
I have cluster provisioned in Hezner cloud. Elastic needs  sysctl -w vm.max_map_count=262144 in order to start. This is not defined in all of the nodes in cluster.

So I need to do something like this:
``` yaml
spec:
  affinity:
    nodeAffinity:
      requiredDuringSchedulingIgnoredDuringExecution:
        nodeSelectorTerms:
        - matchExpressions:
          - key: elasticsearch-shedulable
            operator: Exists

 ```

Troyhy:
@rafariossaa is this correct way to approach this issue?

javsalgar:
I created this PR that I think covers your needs https://github.com/bitnami/charts/pull/1046

carrodher:
I just removed the `stale` tag because we are reviewing the PR (#1046) that fixes this issue

PR #1046, opened 2019-02-07
[bitnami/elasticsearch] Allow custom node affinity policy

javsalgar:
**Description of the change**
Elasticsearch requires to be run in nodes with certain kernel settings. This PR enables to set custom node affinity policies so you can select a node with the proper kernel settings

Fixes https://github.com/bitnami/charts/issues/1032

**Benefits**

**Possible drawbacks**

**Applicable issues**

**Additional information**


bitnami/elasticsearch/values-production.yaml
```diff
@@ -44,6 +44,8 @@ master:
   replicas: 3
   heapSize: 128m
   antiAffinity: "soft"
+  ## node affinity
```
carrodher: Can you use uppercase in those sentences? In this way we can differentiate the comment from the variable, if not they are pretty similar

PR #1052, opened 2019-02-12
Updated READMEs

vikram-bitnami:
**Description of the change**

Updated READMEs with info on BKPR

**Benefits**

Informs users that the chart can also be used with BKPR

**Possible drawbacks**

None

**Applicable issues**

None

**Additional information**

None


tompizmor:
@vikram-bitnami the changes look good to me but even a change in the documentation requires to update the chart version field in the Chart.yaml file.
You should increase the version for each of the charts you have modified the README

PR #1054, opened 2019-02-12
[bitnami/cassandra] Add Prometheus Operator support

raehik:
Adds a ServiceMonitor for PrometheusOperator, and a separate Service to expose the
metrics for pods.

**Description of the change**

Adds easy Prometheus Operator integration for the chart. Metrics come built-in (if enabled), so a relatively trivial change.

**Benefits**

Easy setup for Prometheus Operator, which generates scrape configs for Services automatically. Should work just the same when scaling up and down, without any intereference required, I believe.

**Possible drawbacks**

  * Metrics port and path are hardcoded (8080 and `/metrics`), as in the StatefulSet. Not sure this is best practice.
  * Prometheus resource requires editing to select the ServiceMonitor (if "use all" isn't configured). I haven't found a good workaround for this. Might be intended ^^
  * Created a separate Service for the metrics port. Should it instead be another open port in `service.yaml` ?
  * Dependent on `metrics.create == true`. It'll run if not, but of course no metrics will be scraped.

**Applicable issues**

None.

**Additional information**

Thanks for this chart, it's saving us a lot of time and I hope we can give back a bit :)

### DIFF
--- a/bitnami/cassandra/templates/service-metrics.yaml
+++ b/bitnami/cassandra/templates/service-metrics.yaml
@@ -1,0 +1,21 @@
+{{- if .Values.serviceMonitor.create }}
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ template "cassandra.fullname" . }}-metrics
+  labels:
+    app: {{ template "cassandra.name" . }}
+    chart: {{ template "cassandra.chart" . }}
+    release: {{ .Release.Name }}
+    heritage: {{ .Release.Service }}
+spec:
+  type: ClusterIP
+  ports:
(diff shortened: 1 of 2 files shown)


raehik:
As of yet untested.

raehik:
Seems to work with a local `helm install .` I can share my values if it's useful.

javsalgar:
Hi,

Thanks for sharing the PR, we will take a look and let you know any changes. I think that sharing your values could help

javsalgar:
Ping @raehik , could you share your values.yaml?

Issue #1057, opened 2019-02-19
[Feature Request] MongoDB allow specify some parameters for arbiters separately


md2k:
Hi guys, it is possible to allow setup for arbiters separately resources `limits`, `podafinity`, `nodeselector`, `tolerations` separately. 
Currently if we setup Limits and other parameters from above `arbiter` pods gets parameters same as primary and masters, what is useless because this type of pods not include any data and just helps maintain election's flow. it will be a westing of resources to grant them same cpu/memory limits, also probably `arbiters` can be placed on different nodes which is for example much smaller versus nodes who gets mongodb primary and slaves

Thank you in advance.

javsalgar:
Hi,

Thanks for the feedback. Regarding the PR, you need to do it in the helm/charts repo (where Bitnami mongodb is) in order to work. The upstreamed folder just syncs with the helm/charts repo

PR #1058, opened 2019-02-20
[upstreamed/mongodb] add limits, affinity, node selector and tolerations for arbiter

md2k:
**Description of the change**
Since arbiter is just a mongodb cluster instance to manage vote election procedure, it is not require same amount of resources as other type of mongodb nodes. by default with current chart if we set limits  arbiter pod also get same limits as primary or secondaries, what is westing of resources.
Also pod affinity, node selection and tolerations may vary for arbiter because for example we have pool of big nodes to satisfy primary and secondary nodes and lets say pool which is hold various tools like some controllers and etc. since arbiter not required same resources as real cluster nodes we can deploy it to `utility pool`, while rest cluster will be rolled into specially prepared node pool.

**Benefits**

Save resource when resources limits in use to achieve better pods distribution 

**Possible drawbacks**

People who use arbiters and affected parameters probably should update their value files to keep arbiters deployed to same places where it was before this change (replicate resources limits, nodeselector, affinity, tolerations if was set)

**Applicable issues**

This is about my issue #1057 

**Additional information**

This is correct place to make PR if i want became it available in `stable/mongodb`repository?

### DIFF
--- a/upstreamed/mongodb/templates/statefulset-arbiter-rs.yaml
+++ b/upstreamed/mongodb/templates/statefulset-arbiter-rs.yaml
@@ -38,17 +38,17 @@ spec:
       securityContext:
         fsGroup: {{ .Values.securityContext.fsGroup }}
       {{- end }}
-      {{- if .Values.affinity }}
+      {{- if .Values.arbiter.affinity }}
       affinity:
-{{ toYaml .Values.affinity | indent 8 }}
+{{ toYaml .Values.arbiter.affinity | indent 8 }}
       {{- end -}}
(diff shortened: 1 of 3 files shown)


javsalgar:
Hi,

Thanks for the PR. Could you send the PR  in the helm/charts repo (where Bitnami mongodb is)? The upstreamed folder just syncs with the helm/charts repo

md2k:
Hi @javsalgar , sure , i'll do this bit later.

carrodher:
Hi, feel free to point to this PR in the new one if needed, in the meantime, we will close this one.
Thanks!

Issue #902, opened 2018-10-30
Single node etcd cluster cannot recovery from restart


canhnt:
Reproduction step:
- Deploy an etcd cluster with single node
```
helm install --name foo bitnami/etcd
```
- Go to pod shell, write some key via etcdctl (v3)
- Delete the `foo-etcd-0` pod.
- New `foo-etcd-0` pod is created but cannot start successfully. After several restarting time, the pod goes to `CrashLoopBackOff` state.
- pod crashing log:
```
==> The ID of the host is 0
==> Creating data dir...
==> There is no data at all. Creating new cluster
==> Configuring RBAC authentication!
stty: 'standard input': Inappropriate ioctl for device
stty: 'standard input': Inappropriate ioctl for device
New password:
User root created
auth: already enabled
```

tompizmor:
Thanks for reporting the issue. I could reproduce this issue as well. We will work on a fix and update this issue once it is fixed.

qiuyuzhou:
I have same issue. Is there a workaround for fix it?

juan131:
Hi everyone,

I just merged a PR from @tompizmor which addresses this issue. Could you please confirm the issue disappeared using the latest version of the chart?

canhnt:
I cannot install etcd chart v1.4.1. After installation, the `etcd-0` pod failed to start:
```
==> The ID of the host is 0
==> Data exists. Re-joining etcd member
cat: /bitnami/etcd/data/member_id: No such file or directory
```

If I scale to 3 replicas, `etcd-1` and `etcd-2` pods start successfully. However they do not join the single cluster but cause the split-brain (i.e. 2 single-node clusters).

tompizmor:
mmm, I think this was fixed at https://github.com/bitnami/charts/pull/1010 and broke again with https://github.com/bitnami/charts/pull/1006. I will work on a fix!

canhnt:
Tested the chart v1.4.2, `etcd-0` node starts successfully and can recover from failure (i.e. restart/deletion).

However split-brain issue https://github.com/bitnami/charts/issues/901 still exists

tompizmor:
Correct! The fix I did does not fix that issue.

canhnt:
I suspect the v1.4.2 cannot create a multi-node etcd cluster anymore. I try to deploy from the beginning with 3 nodes but it create 3 clusters, each has 1 node. If it is the case, then it breaks a feature.

canhnt:
[...]

PR #1064, opened 2019-02-27
[bitnami/etcd] Fix etcd node 0 recovery from failure

tompizmor:
**Description of the change**

This change allows node-0 to recovery from failures. Previously, we were only relying on node-0 as an etcd endpoint so if this node was not running some etcd commands failed. Now, we configure all the nodes as possible endpoints so it does not matter if the node-0 is down as long as there are other nodes up and running in the cluster.

**Benefits**

Increase the robustness of the chart.

**Possible drawbacks**

Unknown

**Applicable issues**

Fixes https://github.com/bitnami/charts/issues/902

### DIFF
--- a/bitnami/etcd/Chart.yaml
+++ b/bitnami/etcd/Chart.yaml
@@ -1,5 +1,5 @@
 name: etcd
-version: 1.5.5
+version: 1.5.6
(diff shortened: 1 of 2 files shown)


juan131:
LGTM!

Issue #1061, opened 2019-02-26
Unable to specify KAFKA_LOG_MESSAGE_FORMAT_VERSION


rjanovski:
**Which chart**:

bitnami/kafka

**Description**

the KAFKA_LOG_MESSAGE_FORMAT_VERSION is expose in the docker, but helm does not expose it to the user to config.

This is needed when configuring brokers to work with older clients (while upgrading prod env client by client)

**Steps to reproduce the issue:**

no way to configure e.g. KAFKA_LOG_MESSAGE_FORMAT_VERSION=0.9.0.1

javsalgar:
Hi

Thanks for reporting. I created this PR that solves the issue https://github.com/bitnami/charts/pull/1065

rjanovski:
Thanks @javsalgar !

juan131:
Hi @rjanovski 

We just merged the PR which addresses this issue. Could you please confirm it solved your problem?

rjanovski:
just tried it, works great!
thanks for the quick fix :)

PR #1065, opened 2019-02-27
[bitnami/kafka] Expose KAFKA_LOG_MESSAGE_FORMAT_VERSION

javsalgar:
**Description of the change**
Expose the environment variable LOG_MESSAGE_FORMAT_VERSION as a value

**Benefits**

**Possible drawbacks**

**Applicable issues**
Fixes https://github.com/bitnami/charts/issues/1061

**Additional information**


bitnami/kafka/templates/statefulset.yaml
```diff
@@ -155,6 +155,8 @@ spec:
           value: {{ .Values.logRetentionCheckIntervalMs | quote }}
         - name: KAFKA_LOG_RETENTION_HOURS
           value: {{ .Values.logRetentionHours | quote }}
+        - name: KAFKA_LOG_MESSAGE_FORMAT_VERSION
+          value: {{ .Values.logMessageFormatVersion | replace "_" "" | quote }}
```
tompizmor: You don't need the replace thing here. That was just a workaround to avoid helm parsing big numbers in values.yaml as scientific notation.

bitnami/kafka/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 name: kafka
-version: 1.4.0
+version: 1.4.1
```
tompizmor: Increase the minor version instead

PR #1070, opened 2019-03-02
[bitnami/zookeeper] Add a metrics svc to be able to scrape with prometheus-operator

lcostea:
The additional label app.kubernetes.io/component can separate the exporter
pod from the statefulset ones when used in the svc. There was no change for
the existing labels in order to make it smooth for upgrades. That's why in
the exporter deploy/spec/selector the new label is not added.

**Description of the change**

A new metrics svc that allows to scrape using prometheus-operator.
The current labels set on the metrics pod are the same with the statefulset ones, so
it doesn't allow you to create a metrics service (it will select all pods).
I tried to follow the [recommended kubernetes labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/) for the new label. 

We are using bitnami zookeeper in our kubernetes clusters and we can't scrape it with prometheus-operator without making changes to the bitnami chart.
**Benefits**

Possibility to create a ServiceMonitor to scrape the metrics pods

**Possible drawbacks**

Change is done to be easy for upgrading an existing release that's why it adds one more label to the metrics pod.

**Additional information**

I was able to create a ServiceMonitor and to display the zookeeper metrics in prometheus.
Here is how the ServiceMonitor should look (assuming prometheus-operator is installed on the monitoring ns and it has the label ```release: prometheus-operator```):
```
apiVersion: monitoring.coreos.com/v1
kind: ServiceMonitor
metadata:
  name: {{ template "zookeeper.fullname" . }}-exporter
  namespace: monitoring
  labels:
    release: prometheus-operator
spec:
  selector:
    matchLabels:
      app: {{ template "zookeeper.name" . }}
      app.kubernetes.io/component: {{ template "zookeeper.name" . }}-metrics
  endpoints:
    - port: metrics
      path: /metrics
      interval: 10s
  namespaceSelector:
    any: true
```


lcostea:
Zookeeper was deployed with original values.yaml file, just set metrics.enabled to true.

javsalgar:
LGTM, however, please increase the chart minor version

lcostea:
Done, version increased to 1.2.7

Not the first time I forget to bump the version, need to get used to it.

lcostea:
One colleague pointed me out that the same issue is for kafka also. And it seems he is right, the kafka-exporter deployment has the same labels like the statefulset, though here is a little more complicated with the jmx exporter running as a side container in the statefulset and the kafka-exporter running separately. I might have some time in the next days to address also this.

javsalgar:
Thank you! If you create a PR, ping me and I'll take a look :)

PR #1074, opened 2019-03-07
[bitnami/apache] Add global 'imagePullSecrets' to overwrite any other existing one

juan131:

**Description of the change**

This PR allows using a global parameter to indicate the array of secret names to use as Image Registry Secret.

Use:

```bash
$ kubectl create secret XXXX ...
$ helm install bitnami/apache --name apache --set global.imagePullSecrets=XXXX ...
````

### DIFF
--- a/bitnami/apache/values.yaml
+++ b/bitnami/apache/values.yaml
@@ -1,8 +1,11 @@
-## Global Docker image registry
-## Please, note that this will override the image registry for all the images, including dependencies, configured to use the global value
+## Global Docker image parameters
+## Please, note that this will override the image parameters, including dependencies, configured to use the global value
+## Current available global Docker image parameters: imageRegistry and imagepullSecrets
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: bitnami/apache/values.yaml [hunk 1]
```diff
@@ -1,8 +1,11 @@
-## Global Docker image registry
-## Please, note that this will override the image registry for all the images, including dependencies, configured to use the global value
+## Global Docker image parameters
+## Please, note that this will override the image parameters, including dependencies, configured to use the global value
+## Current available global Docker image parameters: imageRegistry and imagepullSecrets
 ##
 # global:
 #   imageRegistry:
```
beltran-rubo: should we not add myRegistryName here as well?

juan131: Sure. We can add it.

PR #1079, opened 2019-03-11
[bitnami/tensorflow-inception] Add global 'imagePullSecrets' to overwrite any other existing one

carrodher:


bitnami/tensorflow-inception/templates/_helpers.tpl
```diff
@@ -70,3 +70,39 @@ Return the proper image name (for the metrics image)
...
+Also, we can not use a single if because lazy evaluation is not an option
+*/}}
+{{- if .Values.global }}
+{{- if .Values.global.imagePullSecrets }}
+imagePullSecrets:
+{{- range
+.Values.global.imagePullSecrets }}
+  - name: {{ . }}
+{{- end }}
+{{- else if or .Values.image.pullSecrets .Values.metrics.image.pullSecrets }}
```
tompizmor: `.Values.image.pullSecrets` does not exists. There are different images for server and client

- `.Values.server.image.pullSecrets`
- `.Values.client.image.pullSecrets`

PR #1083, opened 2019-03-11
[bitnami/wildfly] Add global 'imagePullSecrets' to overwrite any other existing one

carrodher:


bitnami/wildfly/values.yaml
```diff
@@ -1,8 +1,11 @@
-## Global Docker image registry
-## Please, note that this will override the image registry for all the images, including dependencies, configured to use the global value
+## Global Docker image parameters
+## Please, note that this will override the image parameters, including dependencies, configured to use the global value
+## Current available global Docker image parameters: imageRegistry and imagepullSecrets
```
tompizmor: `imagepullSecrets` should be `imagePullSecrets`

PR #1089, opened 2019-03-11
[bitnami/kafka] Add global 'imagePullSecrets' to overwrite any other existing one

carrodher:


bitnami/kafka/templates/_helpers.tpl
```diff
@@ -80,3 +80,39 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this
...
+Also, we can not use a single if because lazy evaluation is not an option
+*/}}
+{{- if .Values.global }}
+{{- if .Values.global.imagePullSecrets }}
+imagePullSecrets:
+{{- range
+.Values.global.imagePullSecrets }}
+  - name: {{ . }}
+{{- end }}
+{{- else if or .Values.image.pullSecrets .Values.metrics.image.pullSecrets }}
```
tompizmor: I see that `.Values.metrics.image.pullSecrets` does not exists for kafka because it has two ways ox exporting metrics.

Instead, the keys that exists are `.Values.metrics.kafka.image.pullSecrets` and `.Values.metrics.jmx.image.pullSecrets`

PR #1093, opened 2019-03-11
[bitnami/cassandra] Add global 'imagePullSecrets' to overwrite any other existing one

carrodher:


bitnami/cassandra/templates/statefulset.yaml
```diff
@@ -184,18 +184,12 @@ spec:
         - name: configurations
           mountPath: /bitnami/cassandra/conf
       {{- end }}
-      {{- if .Values.image.pullSecrets }}
-      imagePullSecrets:
-        - name: {{ .Values.image.pullSecrets }}
-      {{- end }}
+{{- include "cassandra.imagePullSecrets" . | indent 6 }}
```
tompizmor: here you are specifying imagePullSecrets at pod level and below, for the metrics image, at container level. I don't think we should mix it

PR #1096, opened 2019-03-12
Global options for production and pullSecrets in metrics/other containers

carrodher:
Add global options to `values-production.yaml` and add `pullSecrets` to metrics or other images without it previously set


bitnami/metrics-server/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: metrics-server
-version: 2.3.0
+version: 2.4.0
```
tompizmor: this change should be just a patch version update

PR #1102, opened 2019-03-14
Use global registry in secondary, metrics and other non-default images

carrodher:
Signed-off-by: Carlos Rodriguez Hernandez <crhernandez@bitnami.com>

### DIFF
--- a/bitnami/cassandra/templates/_helpers.tpl
+++ b/bitnami/cassandra/templates/_helpers.tpl
@@ -68,7 +68,7 @@ Also, we can't use a single if because lazy evaluation is not an option
 {{/*
 Return the proper metrics image name
 */}}
-{{- define "metrics.image" -}}
+{{- define "cassandra.metrics.image" -}}
 {{- $registryName := .Values.metrics.image.registry -}}
 {{- $repositoryName := .Values.metrics.image.repository -}}
 {{- $tag := .Values.metrics.image.tag | toString -}}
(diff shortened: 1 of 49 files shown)


bitnami/cassandra/templates/_helpers.tpl
```diff
@@ -65,27 +65,39 @@ Also, we can't use a single if because lazy evaluation is not an option
 {{- end -}}
 {{- end -}}
 
-
 {{/*
 Return the proper metrics image name
 */}}
-{{- define "cassandra.metrics.image" -}}
-{{- $registryName :=  .Values.metrics.image.registry -}}
+{{- define "metrics.image" -}}
```
tompizmor: it is a good practice to use namespace for defined templates

https://helm.sh/docs/chart_best_practices/#names-of-defined-templates

bitnami/node/templates/_helpers.tpl
```diff
@@ -90,7 +90,7 @@ Also, we can't use a single if because lazy evaluation is not an option
 {{/*
 Return the proper Docker Image Registry Secret Names
 */}}
-{{- define "node.imagePullSecrets" -}}
+{{- define "mean.imagePullSecrets" -}}
```
tompizmor: this should be `node`

bitnami/elasticsearch/templates/_helpers.tpl
```diff
@@ -102,19 +102,47 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this
...
+    {{- end -}}
+{{- else -}}
+    {{- printf "%s/%s:%s" $registryName $repositoryName $tag -}}
+{{- end -}}
 {{- end -}}
 
 {{/*
 Return the proper sysctl image name
 */}}
 {{- define "sysctl.image" -}}
```
tompizmor: can you also use a namespace here?

PR #1104, opened 2019-03-16
[bitnami/zookeeper] Manage disruptions when using more than 1 replicas

lcostea:
**Description of the change**

Created a pod distruption budget when replicas is bigger than 1
By default only 1 pod can be unavailable at any point in time
Added additional label to the statefulset so the pdb doesn't select the metrics pod

**Benefits**

A pdb is very useful for replicated application and it will allow us to safely drain a node

**Possible drawbacks**

Can't see any right now

**Additional information**

Tests made: 
- pdb is only applied when replicaCount is 2 or more. 
 - if the additional label wasn't added and the metrics was enabled, then the metrics pod was counted in the pdb (can be easily checked with ```kubectl describe pdb```)

### DIFF
--- a/bitnami/zookeeper/templates/poddisruptionbudget.yaml
+++ b/bitnami/zookeeper/templates/poddisruptionbudget.yaml
@@ -1,0 +1,20 @@
+{{- if gt .Values.replicaCount 1.0 -}}
+apiVersion: policy/v1beta1
+kind: PodDisruptionBudget
+metadata:
+  name: {{ template "zookeeper.fullname" . }}
+  labels:
+    app: {{ template "zookeeper.name" . }}
+    chart: {{ template "zookeeper.chart" . }}
+    release: {{ .Release.Name | quote }}
+    heritage: {{ .Release.Service | quote }}
+    app.kubernetes.io/component: {{ template "zookeeper.name" . }}
+    app.kubernetes.io/instance: {{ .Release.Name | quote }}
+spec:
+  selector:
+    matchLabels:
+      app: {{ template "zookeeper.name" . }}
+      release: {{ .Release.Name | quote }}
+      app.kubernetes.io/component: {{ template "zookeeper.name" . }}
+{{ toYaml .Values.podDisruptionBudget | indent 2 }}
+{{- end }}
--- a/bitnami/zookeeper/values-production.yaml
+++ b/bitnami/zookeeper/values-production.yaml
@@ -36,6 +36,14 @@ image:
 ##
 updateStrategy: RollingUpdate
 
+
+## Limits the number of pods of the replicated application that are down simultaneously from voluntary disruptions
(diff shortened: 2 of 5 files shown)


lcostea:
@javsalgar Can I get any feedback on this? (you approved my last PR on the zookeeper chart)
We are thinking of running zookeeper in production, so a pod disruption budget would be helpful.

bitnami/zookeeper/templates/poddisruptionbudget.yaml
```diff
@@ -0,0 +1,19 @@
+{{- if gt .Values.replicaCount 1.0 -}}
```
tompizmor: just curiosity, why `1.0` and not `1` ?

bitnami/zookeeper/values-production.yaml
```diff
@@ -36,6 +36,14 @@ image:
 ##
 updateStrategy: RollingUpdate
 
+
+## Limits the number of pods of the replicated application that are down simultaneously from voluntary disruptions
```
tompizmor: In values.yaml you also added the following sentence:

```
## The PDB will only be created if replicaCount is greater than 1
```

Can you add it also here?

lcostea: done

bitnami/zookeeper/templates/poddisruptionbudget.yaml
lcostea: It seems that either yaml or Helm is setting Values.replicaCount to float. And when using `1` the `gt` function gives the error that it can't compare two different types. 
I actually fixed it by following this: https://stackoverflow.com/questions/46009933/kubernetes-helm-chart-if-condition-check

lcostea:
@tompizmor I think I done the changes you asked, can you please take a look?
Thanks

srueg:
It might make sense to wait for #1127 and use the `component` label introduced there?

lcostea:
Ok, lets wait for #1127

tompizmor:
I requested a small change in #1127 that you should also include in this PR. For each change in the chart, we should update the chart version. 

In this case, as you are adding a new feature, you should update the minor version, so please update from `1.5.0` to `1.6.0` [here](https://github.com/bitnami/charts/blob/master/bitnami/zookeeper/Chart.yaml#L3)

lcostea:
@tompizmor if I am waiting for #1127 I will postpone the minor version bump, because others might increase it before.
Once #1127 is cleared I will update the chart minor version

tompizmor:
@lcostea OK! If #1127 is not updated in the following couple of days I will make the changes and let you know!

tompizmor:
hi @lcostea , @srueg sent https://github.com/bitnami/charts/pull/1144 and I already merge it so you can now update this PR

lcostea:
@tompizmor PR updated with the changes from #1127

PR #1113, opened 2019-03-26
Fix syntax issue in consul chart

tompizmor:

**Description of the change**

Fix following syntax error:

```
Error: parse error in "consul/templates/statefulset.yaml": template: consul/templates/statefulset.yaml:259: unexpected EOF
```


juan131:
LGTM!

PR #1127, opened 2019-04-08
bitnami/zookeeper: remove zookeeper-exporter pod from zookeeper service

wuzhihui1123:
service of x-zookeeper and x-zookeeper-headless must not contains the pod of zookeeper-exporter

PR #1136, opened 2019-04-12
[bitnami/zookeeper] Set metadata labels

srueg:
**Description of the change**

Add further metadata labels like component and instance.

**Benefits**

Implements some of the [Chart best practices](https://helm.sh/docs/chart_best_practices/#labels-and-annotations).

**Additional information**

To set up proper Pod AntiAffinity rules, at least the component label is necessary to target only the ZooKeeper pods and not the metrics exporter pod.

### DIFF
--- a/bitnami/zookeeper/templates/configmap.yaml
+++ b/bitnami/zookeeper/templates/configmap.yaml
@@ -8,6 +8,7 @@ metadata:
     chart: {{ template "zookeeper.chart" . }}
     release: {{ .Release.Name | quote }}
     heritage: {{ .Release.Service | quote }}
+    app.kubernetes.io/instance: {{ .Release.Name | quote }}
 data:
   zoo.cfg: |-
 {{ .Values.config | indent 4 }}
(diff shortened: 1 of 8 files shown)


srueg:
Hmm, parts of this actually duplicate #1127

srueg:
And #1104 also sets the `component` label :smile:

bitnami/zookeeper/templates/configmap.yaml
```diff
@@ -8,6 +8,7 @@ metadata:
     chart: {{ template "zookeeper.chart" . }}
     release: {{ .Release.Name | quote }}
     heritage: {{ .Release.Service | quote }}
+    app.kubernetes.io/instance: {{ .Release.Name }}
```
juan131: I'd use `{{ .Release.Name | quote }}` for consistency with other labels. This comment applies to every manifest file.

srueg: Done. I also removed the `component` label again, as it will be implemented by #1127

PR #1145, opened 2019-04-25
[bitnami/zookeeper] Mount emptyDir for ZooKeeper

srueg:
**Description of the change**

If persistence is disabled, mount the emptyDir volume instead.

**Benefits**

Even if persistence is disabled, an emptyDir is used.

**Possible drawbacks**

**Applicable issues**

Previously the emptyDir was defined but not mounted. 

**Additional information**

### DIFF
--- a/bitnami/zookeeper/Chart.yaml
+++ b/bitnami/zookeeper/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: zookeeper
-version: 1.7.0
+version: 1.7.1
 appVersion: 3.4.14
 description: A centralized service for maintaining configuration information, naming, providing distributed synchronization, and providing group services for distributed applications.
 keywords:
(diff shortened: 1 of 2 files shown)


srueg:
@juan131 can you please have a look at this?

bitnami/zookeeper/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: zookeeper
-version: 1.6.1
+version: 1.7.0
```
juan131: This is a fix, you're not introducing any new parameter or functionality. Therefore, I'd increase the "patch version":

```diff
- version: 1.6.1
+ version: 1.6.2
```

srueg: Totally agree

PR #1146, opened 2019-04-26
[bitnami/zookeeper] metrics exporter timeout

srueg:
**Description of the change**

Change the exporter image and set a connection timeout.

**Benefits**

Usually a metrics endpoint is scraped with an interval of between 10s - 1m, so a lower connection timeout makes sense. Also the default timeout for a scrape is 10s, which will be exceeded if even one ZK instance is down.

**Possible drawbacks**

**Applicable issues**

**Additional information**

### DIFF
--- a/bitnami/zookeeper/Chart.yaml
+++ b/bitnami/zookeeper/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: zookeeper
-version: 2.1.0
+version: 2.2.0
 appVersion: 3.4.14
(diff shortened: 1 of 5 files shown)


srueg:
@tompizmor any chance you can take a look a this?

javsalgar:
Hi,

Thanks for the PR. Could you document the variable in the README.md, as well as adding the timeout in values-production.yaml?

srueg:
@javsalgar anything else I need to do here?

PR #1158, opened 2019-04-30
Add new chart: MinIO

juan131:

**Description of the change**

This PR adds a new chart: MinIO

[MinIO](https://min.io) is an object storage server, compatible with Amazon S3 cloud storage service, mainly used for storing unstructured data (such as photos, videos, log files, etc.)

This chart supports two modes: **standalone** and **distributed** and it's based in the [Bitnami Minio Image](https://github.com/bitnami/bitnami-docker-minio)


bitnami/minio/Chart.yaml
```diff
@@ -0,0 +1,18 @@
+apiVersion: v1
+version: 0.0.1
+appVersion: 2019.4.23
+description: MinIO is an object storage server, compatible with Amazon S3 cloud storage service, mainly used for storing unstructured data (such as photos, videos, log files, etc.)
+keywords:
+- minio
+- storage
+- object-storage
+- s3
```
javsalgar: If it can be used for a cluster, then we could add the cluster keyword just like we do with postgresql

https://github.com/helm/charts/blob/master/stable/postgresql/Chart.yaml

bitnami/minio/README.md
```diff
@@ -0,0 +1,179 @@
...
+| `statefulset.updateStrategy`         | Statefulset update strategy policy                                                           | `RollingUpdate`                                         |
+| `statefulset.podManagementpolicy`    | Statefulset pods management policy                                                           | `Parallel`                                              |
+| `existingSecret`                     | Existing secret with MinIO credentials                                                       | `nil`                                                   |
+| `accessKey.password`                 | MinIO Access Key. Ignored if existing secret is provided.                                    | _random 10 character alphanumeric string_               |
+| `accessKey.forcePassword`            | Force users to specify an Access Key                                                         | `false`                                                 |
+| `secretKey.password`                 | MinIO Secret Key. Ignored if existing secret is provided.                                    | _random 40 character alphanumeric string_               |
```
javsalgar: It would be nice if we can have `usePasswordFile` just like we have with PostgreSQL and Redis https://github.com/helm/charts/tree/master/stable/postgresql

bitnami/minio/templates/statefulset.yaml
```diff
@@ -0,0 +1,175 @@
...
+          tcpSocket:
+            port: minio
+        {{- end }}
+        {{- if .Values.resources }}
+        resources: {{ toYaml .Values.resources | nindent 10 }}
+        {{- end }}
+        volumeMounts:
+        - name: "data"
+          mountPath: {{ .Values.persistence.mountPath }}
+{{- if and .Values.persistence.enabled .Values.persistence.existingClaim }}
```
javsalgar: This will not to work unless it is a ReadWriteMany PVC. I would leave the existingClaim only for the standalone mode.

bitnami/minio/templates/deployment-standalone.yaml
```diff
@@ -0,0 +1,131 @@
...
+        {{- end }}
+        env:
+        {{- if .Values.image.debug }}
+        - name: BITNAMI_DEBUG
+          value: "true"
+        {{- end }}
+        - name: MINIO_ACCESS_KEY
+          valueFrom:
+            secretKeyRef:
+              {{- if .Values.existingSecret }}
```
javsalgar: Move this logic to a helper function. This is useful when an umbrella chart uses this one as a subchart. Example, postgresql

bitnami/minio/templates/statefulset.yaml
juan131: I agree. It makes no sense to use the same claim for more that one node in distributed mode.

bitnami/minio/values.yaml
```diff
@@ -7,6 +7,7 @@
 #   imagePullSecrets:
 #     - myRegistryKeySecretName
 
+
```
javsalgar: Remove extra line

bitnami/minio/templates/_helpers.tpl
```diff
@@ -97,6 +97,52 @@ imagePullSecrets:
...
+{{- end -}}
+
+{{/*
+Return MinIO secretKey
+*/}}
+{{- define "minio.secretKey" -}}
+{{- if .Values.global.minio.secretKey }}
+    {{- if .Values.global.minio.secretKey.password }}
+        {{- .Values.global.minio.secretKey.password -}}
+    {{- else if (not .Values.global.minio.secretKey.forcePassword) }}
```
javsalgar: It is not clear to me why you would want the forcePassword parameter to be accessible in other subcharts

juan131: So you suggest sth like this:
```
{{/*
Return MinIO secretKey
*/}}
{{- define "minio.secretKey" -}}
{{- if .Values.global.minio.secretKey }}
    {{- if .Values.global.minio.secretKey.password }}
        {{- .Values.global.minio.secretKey.password -}}
    {{- else if (not .Values.secretKey.forcePassword) }}
        {{- randAlphaNum 40 -}}
    {{- else -}}
        {{ required "A Secret Key is required!" .Values.global.minio.secretKey.password }}
    {{- end -}}
{{- else if .Values.secretKey -}}
    {{- if .Values.secretKey.password }}
        {{- .Values.secretKey.password -}}
    {{- else if (not .Values.secretKey.forcePassword) }}
        {{- randAlphaNum 40 -}}
    {{- else -}}
        {{ required "A Secret Key is required!" .Values.secretKey.password }}
    {{- end -}}
{{- end -}}
{{- end -}}
```

juan131: ???

javsalgar: Yes

javsalgar:
LGTM

javsalgar:
@WarheadsSE FYI

WarheadsSE:
@javsalgar Thanks for bring this to my attention. :bow: Thanks for taking our needs and practices into consideration.

I'll make a note to review it in-depth, there may be a few things from `gitlab/gitlab` that would be useful here. 

1. `$` as a global property anchor removes the need to deal with `if else` logic for stored properties
1. [checkConfig](https://gitlab.com/charts/gitlab/blob/master/templates/_checkConfig.tpl) logic is being expanded, to handle the "whoops! bad / conflicting values!" scenarios in your `NOTES.txt`
1. The use of `.Capabilities` is problematic for "tiller free" usage. It's super handy as long as all features are reachable from _just_ `helm template`, but breaks if these are only available within Tiller pod.

juan131:
Hi @WarheadsSE 

Thanks so much for your feedback!! 

> $ as a global property anchor removes the need to deal with if else logic for stored properties

I'm not sure I fully understand this point. Could you please provide more information?

> 2. checkConfig logic is being expanded, to handle the "whoops! bad / conflicting values!" scenarios in your NOTES.txt

It's something very interesting. I think we can implement sth like on this chart.

> 3. The use of .Capabilities is problematic for "tiller free" usage. It's super handy as long as all features are reachable from just helm template, but breaks if these are only available within Tiller pod.

I guess you're not using Tiller for security reasons and you cannot use charts that made user of some of the features Tiller pod offers. What alternative do you suggest to ensure we're using the right **ApiVersion**? Do you use anything different?

WarheadsSE:
> > $ as a global property anchor removes the need to deal with if else logic for stored properties
>
> I'm not sure I fully understand this point. Could you please provide more information?

Sure. See [chart_template_guide/#varibles](https://helm.sh/docs/chart_template_guide/#variables) for the `$` defintion:

> However, there is one variable that is always global - $ - this variable will always point to the root context.

As for the use of assignments, you _should_ be able to get some simplification.
One example, [this template](https://github.com/bitnami/charts/pull/1158/files#diff-fc0aed23cd2af10f2bd5637d80f9fb8bR51) could be greatly simplified. As the `.Values.global` _should_ always be present, pre-checking for it via lazy should be safe. If, for some reason, you're not seeing this, then simply providing an empty set in `values.yaml` should prevent the need for the complication of the template logic itself.

```
{{- $registryName := coalesce .Values.global.imageRegistry .Values.clientImage.registry -}}
{{- $repositoryName := .Values.clientImage.repository -}}
{{- $tag := .Values.clientImage.tag | toString -}}
{{- printf "%s/%s:%s" $registryName $repositoryName $tag -}}
```
> What alternative do you suggest to ensure we're using the right ApiVersion? Do you use anything different?

In our case, we're actually not supporting Kubernetes < `1.8.x`. We've documented that clearly, and its due to the needs of the application not being met by versions prior to that. I do realize that there are _some_ users still using <= `1.7.x`, but that is also more than 2 years old at this point, and not longer getting updates. The `1.7.x` series hasn't seen a non-beta tag since 2018-04-04.

This [issue discussion on LTS policy](https://github.com/kubernetes/community/issues/2720) is quite informative.


> It's something very interesting. I think we can implement sth like on this chart.

If you liked that, see where it originated, in our [Deprecations templates](https://gitlab.com/charts/gitlab/blob/master/doc/development/deprecations.md). You'd probably find our [Design Decisions](https://gitlab.com/charts/gitlab/blob/master/doc/architecture/decisions.md ) enlightening as well.

juan131:
Hi @WarheadsSE 

Thanks so much for your suggestions and the detailed explanations! I apply most on them on the PR below, please take a look when you have a chance.

https://github.com/bitnami/charts/pull/1168

Regarding not supporting `K8s < 1.8.x`, I think it's just fine. Indeed, as long as it's reflected in the README.md (as it is) we can consider it obsolete.

PR #1059, opened 2019-02-21
Name truncation fix

heironeous:
**Description of the change**

Changes the truncation of certain helm charts from 24 to 63. Most charts already support 63 char long naming, standardized them

**Benefits**

Can use longer strings for the name of the deployments (63 instead of 24)

**Possible drawbacks**

If the name of the service/pods are used within the containers, they might cause issues

**Applicable issues**

**Additional information**

PR #1163, opened 2019-05-02
[bitnami/zookeeper] Name truncation fix

heironeous:
**Description of the change**

Change the truncated name functions from 24 characters to 63 characters. This is already implemented for many charts, simply copied the code from others.

**Benefits**

The helm deployment names could be longer than 24 chars, but the DNS names currently cannot. This fixes that. Also, fixes an issue where a truncation to 24 characters creates an invalid DNS name (such as a name 'aaaaa-bbbbb-ccccc-dddd-')

**Possible drawbacks**

-NA-

**Applicable issues**

-NA-

**Additional information**

Recreated the PR, missed the notifications, and the previous PR went stale - [link to stale PR](https://github.com/bitnami/charts/pull/1059#event-2257858084)


javsalgar:
Hi,

Thanks for the PR. Note that, if you update several charts, you would need to bump the versions in all of them. If you are only interested in Zookeeper, then it would be better to not include the changes in the rest of the charts

PR #1173, opened 2019-05-07
[bitnami/kafka] Update dependencies

juan131:
**Description of the change**

This PR updates the Kafka chart dependency: Zookeeper

### DIFF
--- a/bitnami/kafka/Chart.yaml
+++ b/bitnami/kafka/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: kafka
-version: 1.10.1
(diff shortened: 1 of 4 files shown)


bitnami/kafka/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: kafka
-version: 1.10.1
+version: 1.11.0
```
tompizmor: I think we should bump the major version

juan131: Bumping major version

PR #1174, opened 2019-05-07
[bitnami/kafka] Adapt kafka to new labels format

juan131:

Update labels to the official format.

This is a follow-up PR of #1173
That's the reason why a major version hasn't been bumped


juan131:
This is a follow-up PR of https://github.com/bitnami/charts/pull/1173
That's the reason why a major version hasn't been bumped

PR #1176, opened 2019-05-07
[bitnami/nginx] Add ability to configure Liveness and Readiness Probes

RothAndrew:
**Description of the change**

Adds the ability to configure the Liveness and Readiness Probes

**Benefits**

The user is able to configure the Liveness and Readiness Probes to their liking

**Possible drawbacks**

None. The defaults identically match the previous hard-coded values in the Deployment template.

**Applicable issues**

None that are open.

**Additional information**

Documentation updates still need to be done, but I wanted to get this out there to get some eyeballs on it.

### DIFF
--- a/bitnami/nginx/templates/deployment.yaml
+++ b/bitnami/nginx/templates/deployment.yaml
@@ -40,18 +40,18 @@ spec:
           containerPort: 8080
         livenessProbe:
           httpGet:
-            path: /
-            port: http
-          initialDelaySeconds: 30
-          timeoutSeconds: 5
-          failureThreshold: 6
+            path: {{ .Values.livenessProbe.httpGet.path }}
+            port: {{ .Values.livenessProbe.httpGet.port }}
+          initialDelaySeconds: {{ .Values.livenessProbe.initialDelaySeconds }}
+          timeoutSeconds: {{ .Values.livenessProbe.timeoutSeconds }}
+          failureThreshold: {{ .Values.livenessProbe.failureThreshold }}
         readinessProbe:
           httpGet:
-            path: /
-            port: http
-          initialDelaySeconds: 5
-          timeoutSeconds: 3
-          periodSeconds: 5
+            path: {{ .Values.readinessProbe.httpGet.path }}
+            port: {{ .Values.readinessProbe.httpGet.port }}
+          initialDelaySeconds: {{ .Values.readinessProbe.initialDelaySeconds }}
+          timeoutSeconds: {{ .Values.readinessProbe.timeoutSeconds }}
+          periodSeconds: {{ .Values.readinessProbe.periodSeconds }}
         volumeMounts:
         {{- if .Values.serverBlock }}
         - name: nginx-server-block
--- a/bitnami/nginx/values.yaml
+++ b/bitnami/nginx/values.yaml
@@ -110,6 +110,27 @@ ingress:
 #     }
 #   }
 
+## Liveness Probe
+## ref: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/
+##
+livenessProbe:
+  httpGet:
+    path: /
+    port: http
+  initialDelaySeconds: 30
+  timeoutSeconds: 5
+  failureThreshold: 6
+
+## Readiness Probe
+## ref: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/
+readinessProbe:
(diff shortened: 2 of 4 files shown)


RothAndrew:
ping @juan131, mind taking a look at this one for me?

bitnami/nginx/values.yaml
```diff
@@ -110,6 +110,27 @@ ingress:
...
+
+## Readiness Probe
+## ref: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/
+readinessProbe:
+  httpGet:
+    path: /
+    port: http
+  initialDelaySeconds: 5
+  timeoutSeconds: 3
+  periodSeconds: 5
```
juan131: What about the rest of

bitnami/nginx/templates/deployment.yaml
```diff
@@ -40,18 +40,18 @@ spec:
...
-            path: /
-            port: http
-          initialDelaySeconds: 5
-          timeoutSeconds: 3
-          periodSeconds: 5
+            path: {{ .Values.readinessProbe.httpGet.path }}
+            port: {{ .Values.readinessProbe.httpGet.port }}
+          initialDelaySeconds: {{ .Values.readinessProbe.initialDelaySeconds }}
+          timeoutSeconds: {{ .Values.readinessProbe.timeoutSeconds }}
+          periodSeconds: {{ .Values.readinessProbe.periodSeconds }}
```
juan131: I'd create some extra values `livenessProbe.enabled` and `readinessProbe.enabled` (set by default to `true`) so you can apply some logic like the one below:

```yaml
        {{- if . Values.livenessProbe.enabled }}
        livenessProbe:
            path: {{ .Values.livenessProbe.httpGet.path }}
            port: {{ .Values.livenessProbe.httpGet.port }}
          initialDelaySeconds: {{ .Values.livenessProbe.initialDelaySeconds }}
          timeoutSeconds: {{ .Values.livenessProbe.timeoutSeconds }}
          failureThreshold: {{ .Values.livenessProbe.failureThreshold }}
        {{- end }}
        {{- if . Values.readinessProbe.enabled }}
        readinessProbe:
          path: {{ .Values.readinessProbe.httpGet.path }}
          port: {{ .Values.readinessProbe.httpGet.port }}
        initialDelaySeconds: {{ .Values.readinessProbe.initialDelaySeconds }}
        timeoutSeconds: {{ .Values.readinessProbe.timeoutSeconds }}
        failureThreshold: {{ .Values.readinessProbe.failureThreshold }}
        {{- end }}
```

bitnami/nginx/values.yaml
juan131: What about other values such as the one below?

```yaml
    successThreshold:
    failureThreshold:
```

RothAndrew: My goal was to make the probes configurable while changing as little as possible. I have no problem with adding some new functionality if that is the desire.

What would you like the defaults to be for the new fields?

Alternatively, something I've done before is to use `{{ toYaml .Values.livenessProbe }}`, which lets the user configure whatever parameters they want. That's definitely a more radical change but it gives the user more options.

bitnami/nginx/templates/deployment.yaml
RothAndrew: Sure I can definitely do that.

bitnami/nginx/values.yaml
juan131: I agree it's more flexible for the user. I'd follow that approach.

bitnami/nginx/templates/deployment.yaml
juan131: If you're moving to the approach to use `{{ toYaml .Values.livenessProbe }}`, do not implement this change. You should check whether the `Values.livenessProbe` object is set or not instead.

RothAndrew:
@juan131 I like both of your comments, but I think making the requested change is going to take longer than I have right now.

The PR as-is identically duplicates the current functionality, except it lets you modify the values for the parameters like `path`.

Would you be opposed to accepting the PR as-is and letting me submit another one that has the new feature?

juan131:
Hi @RothAndrew 

I'd rather wait for the final changes. It does not make much sense to release to different chart versions to include the same feature.

Feel free to take as much time as you need, no hurries.

RothAndrew:
reopen please. I'm working on making the changes @juan131 requested.

PR #1177, opened 2019-05-09
[bitnami/airflow] Update Airflow dependencies

juan131:
Signed-off-by: juan131 juan@bitnami.com

This PR updates Airflow so it uses the latest major version of PostgreSQL as dependency.

#### Checklist

- [x] [DCO](https://github.com/helm/charts/blob/master/CONTRIBUTING.md#sign-your-work) signed
- [x] Chart Version bumped
- [x] title of the PR contains starts with chart name e.g. `[stable/chart]`

### DIFF
--- a/bitnami/airflow/Chart.yaml
+++ b/bitnami/airflow/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: airflow
-version: 0.0.2
+version: 1.0.0
 appVersion: 1.10.3
 description: Apache Airflow is a platform to programmatically author, schedule and monitor workflows.
 keywords:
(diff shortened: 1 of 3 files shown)


bitnami/airflow/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: airflow
-version: 0.0.2
+version: 1.0.0
```
tompizmor: I would not update the major version in this case as the airflow chart has not reached the version 1.0.0 and is not considered stable yet.

Use 0.0.3 instead

juan131: Done

PR #1178, opened 2019-05-09
[bitnami/kafka] Create a pod disruption budget when replicaCount is bigger than 1

lcostea:
**Description of the change**

Created a pod distruption budget when replicas is bigger than 1
By default only 1 pod can be unavailable at any point in time

**Benefits**

A pdb is very useful for replicated application and it will allow us to safely drain a node

**Possible drawbacks**

Can't see any right now

**Additional information**

Tested that pdb is not created when replicaCount is 1


lcostea:
@javsalgar could you please take a look when you have some time?

PR #1180, opened 2019-05-13
[bitnami/airflow] Change gitImage to git in README

alemorcuq:
**Description of the change**

The configuration options for the `git` image are called `git` instead of `gitImage`. This fixes the docs to reflect the correct option.

**Benefits**

Better documentation.


alemorcuq:
> Be careful with the trailing spaces

The trailing spaces are there to make the table pretty.

Issue #1157, opened 2019-04-29
kafka volume mounts could use a persistence.existingClaim similar to the postgres chart


janitha09:
**Which chart**:
| chart | application |
| kafka-1.10.1 | 2.2.0 | 
                  

**Description**

The postgres char located here https://github.com/helm/charts/tree/master/stable/postgresql has a `persistence.existingClaim` configuration. The kafka chart does not. 

You can work around this by ensuring that the PVCs have a name as expected by the chart
```
NAME                    REVISION        UPDATED                         STATUS          CHART                           APP VERSION      NAMESPACE
kafka-janitha           2               Fri Apr 26 18:39:44 2019        DEPLOYED        kafka-1.10.1                    2.2.0            kafka
```
```
NAME                             STATUS   VOLUME           CAPACITY   ACCESS MODES   STORAGECLASS    AGE
data-kafka-janitha-0             Bound    kafka-data       8Gi        RWO            kafka-data      11d
data-kafka-janitha-zookeeper-0   Bound    zookeeper-data   8Gi        RWO            zookeeper-efs   11d
```

**Steps to reproduce the issue:**

1. `helm install kafka-janitha --set livenessProbe.initialDelaySeconds=120,livenessProbe.periodSeconds=120,livenessProbe.timeoutSeconds=120 bitnami/kafka`

**Describe the results you received:**

PV is not created
PVC created for kafka not for zookeeper
```persistence.storageClass``` creates a PVC for kafka only per the documentation. There isn't a way to create it for zookeeper. with the provided switches

**Describe the results you expected:**

Specify or create PVC for zookeeper and kafka

**Version of Helm and Kubernetes**:

- Output of `helm version`:

```
Client: &version.Version{SemVer:"v2.13.1", GitCommit:"618447cbf203d147601b4b9bd7f8c37a5d39fbb4", GitTreeState:"clean"}
Server: &version.Version{SemVer:"v2.13.1", GitCommit:"618447cbf203d147601b4b9bd7f8c37a5d39fbb4", GitTreeState:"clean"}
```

- Output of `kubectl version`:

```
Client Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.1", GitCommit:"b7394102d6ef778017f2ca4046abbaa23b88c290", GitTreeState:"clean", BuildDate:"2019-04-08T17:11:31Z", GoVersion:"go1.12.1", Compiler:"gc", Platform:"linux/amd64"}
[...]

PR #1186, opened 2019-05-15
[bitnami/kakfa] Add option to use existingClaim

tompizmor:

**Description of the change**

Add option to specify existing volume claim.

**Benefits**

It will be possible to reuse a managed persistent volume claim

**Possible drawbacks**

None

**Applicable issues**

#1157


bitnami/kafka/templates/statefulset.yaml
```diff
@@ -272,14 +272,25 @@ spec:
...
+{{- if and .Values.persistence.enabled .Values.persistence.existingClaim }}
+      - name: data
+        persistentVolumeClaim:
+{{- with .Values.persistence.existingClaim }}
+          claimName: {{ tpl . $ }}
+{{- end }}
+{{- else if not .Values.persistence.enabled }}
+      - name: data
+        emptyDir: {}
+{{- else if and .Values.persistence.enabled (not .Values.persistence.existingClaim) }}
```
juan131: I think it's simpler to use sth like:
```yaml
{{- if not .Values.persistence.enabled }}
...
{{- else if .Values.persistence.existingClaim }}
...
{{- else }}
...
{{- end }}
```

tompizmor: done

carrodher:
There are some conflicts in the README, @tompizmor can you take a look? Apart from that, LGTM

Issue #1130, opened 2019-04-09
Zookeeper log messages show malformed connections from the readiness probe


rjemanuele:
**Which chart**:

bitnami/zookeeper

**Description**

The log messages contain many sequences such as this:
```
2019-04-09 04:58:12,368 [myid:6] - INFO  [NIOServerCxn.Factory:0.0.0.0/0.0.0.0:2181:NIOServerCnxnFactory@215] - Accepted socket connection from /10.46.128.0:53334
2019-04-09 04:58:12,368 [myid:6] - WARN  [NIOServerCxn.Factory:0.0.0.0/0.0.0.0:2181:NIOServerCnxn@376] - Unable to read additional data from client sessionid 0x0, likely client has closed socket
2019-04-09 04:58:12,369 [myid:6] - INFO  [NIOServerCxn.Factory:0.0.0.0/0.0.0.0:2181:NIOServerCnxn@1056] - Closed socket connection for client /10.46.128.0:53334 (no session established for client)
```
These connections without sessions are the Kubernetes readiness probes checking the process in order to include it in a Kubernetes Service.

**Steps to reproduce the issue:**

1. Install Zookeeper using the helm chart
2. Check the logs using `kubectl logs`

**Describe the results you received:**

See the Description above.

**Describe the results you expected:**

These extraneous message should not be in the log as they are noise and distract from actual , important log message.

**Potential Solution**

I would suggest doing something like this with an ncat or such.  https://stackoverflow.com/questions/29106546/how-to-check-if-zookeeper-is-running-or-up-from-command-prompt

Over ssl you could use ncat and stunnel to do the same test.

rjemanuele:
Some of this was discussed in #1097 off-topic

javsalgar:
Hi,

I agree that we should do something about it. I do not know if changing the readiness probe or maybe changing the log level. That's something we will need to investigate

miguelaeh:
Hi @rjemanuele ,
I have done a new update on the image to support the log level, and the pull request to Chart. I put the log level on `ERROR` by default, so the readiness probes's logs will not appear, but the drawback is that only ERROR messages will be shown.
Link to the pull request: https://github.com/bitnami/charts/pull/1187

carrodher:
I am closing this issue as it should be fixed in the mentioned PR, please reopen it if needed.

PR #1187, opened 2019-05-15
Add ZOO_LOG_LEVEL env var

miguelaeh:
Set the Zookeeper log level to `ERROR` to avoid the verbosity of the readiness probe.

### DIFF
--- a/bitnami/zookeeper/README.md
+++ b/bitnami/zookeeper/README.md
@@ -43,7 +43,7 @@ The command removes all the Kubernetes components associated with the chart and 
 
 ## Log level
 
-You can configure the Zookeeper log level using the `ZOO_LOG_LEVEL` environment variable. By default, it is set to `FALSE` because of each readiness prove produce an `INFO` message on connection and a `WARN` message on disconnection.
+You can configure the Zookeeper log level using the `ZOO_LOG_LEVEL` environment variable. By default, it is set to `ERROR` because of each readiness prove produce an `INFO` message on connection and a `WARN` message on disconnection.
 
 ## Configuration
 
--- a/bitnami/zookeeper/README.md
+++ b/bitnami/zookeeper/README.md
@@ -43,7 +43,7 @@ The command removes all the Kubernetes components associated with the chart and 
 
 ## Log level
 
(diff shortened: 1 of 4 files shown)


bitnami/zookeeper/README.md
```diff
@@ -41,6 +41,10 @@ $ helm delete my-release
 
 The command removes all the Kubernetes components associated with the chart and deletes the release.
 
+## Log level
+
+You can configure the Zookeeper log level using the `ZOO_LOG_LEVEL` environment variable. By default, it is set to `FALSE` because of each readiness prove produce an `INFO` message on connection and a `WARN` message on disconnection.
```
srueg: This should probably say "set to `ERROR`"

bitnami/zookeeper/values.yaml
```diff
@@ -13,7 +13,7 @@
 image:
   registry: docker.io
   repository: bitnami/zookeeper
-  tag: 3.4.14
+  tag: 3.4.15
```
carrodher: Why are you bumping the image tag?

bitnami/zookeeper/README.md
carrodher: According to the README table some lines below:
```
| `logLevel`                            | Log level of Zookeeper server                                       | `ERROR`       |
```
it is set to `ERROR` by default

miguelaeh: Yes, you are right, it was a mistake.

carrodher:
Does it work with the current image version (`3.4.14`)?

bitnami/zookeeper/values.yaml
miguelaeh: I thought my changes on the image would create a new version, but as it will create a new revision, so I change this back.

miguelaeh:
> Does it work with the current image version (`3.4.14`)?

Yes, it does. 
And the output is the following:
```zookeep INFO  ==> Validating inputs...
nami    INFO  zookeeper successfully initialized
ZooKeeper JMX enabled by default
Using config: /opt/bitnami/zookeeper/bin/../conf/zoo.cfg
Starting zookeeper ... STARTED```

As you can see the changes avoid the readiness prove logs.

carrodher:
This PR fixed #1130

PR #1191, opened 2019-05-19
Cassandra: Add support for overriding environment variables

sukrit007:
**Description of the change**
Adds support for passing extra environment variables to Cassandra image using helm variable: extraVars.

**Benefits**
At times we need to pass in additional environment variables to override Cassandra setting that is not available in the helm chart. Instead of adding all the options to the chart, we can add the environment variable that we need to override. (e.g: NAMI_DEBUG)

**Possible drawbacks**
N/A

**Applicable issues**
N/A

**Additional information**
N/A


carrodher:

PR #1196, opened 2019-05-23
kafka: add protocol map and broker listener options

philwinder:
**Description of the change**

Adds kafka protocol map and broker listener options to the chart, based upon the environmental variables listed in the [Docker container](https://hub.docker.com/r/bitnami/kafka/)

**Benefits**

Allows the use of internal and external listeners (for allowing remote connections) straight from the helm values file.


carrodher:

philwinder:
Thanks @carrodher. I have rebased and incremented the chart patch version. Over to you.

PR #1203, opened 2019-05-28
Use immutable tags

carrodher:
- Bump patch version
- Use immutable tags
- Check rolling tags in the images (except minideb and exporter) and show a warning in _NOTES.txt_
- Add a section about rolling/immutable tags in _README.md_

### DIFF
--- a/bitnami/minio/values-production.yaml
+++ b/bitnami/minio/values-production.yaml
@@ -18,7 +18,7 @@ global:
 image:
   registry: docker.io
   repository: bitnami/minio
-  tag: 2019.5.14
+  tag: 2019.5.14-debian-9-r5
   ## Specify a imagePullPolicy
   ## Defaults to 'Always' if image tag is 'latest', else set to 'IfNotPresent'
   ## ref: http://kubernetes.io/docs/user-guide/images/#pre-pulling-images
--- a/bitnami/minio/values-production.yaml
+++ b/bitnami/minio/values-production.yaml
@@ -42,7 +42,7 @@ image:
 clientImage:
   registry: docker.io
   repository: bitnami/minio-client
-  tag: latest
+  tag: 2019.5.23-debian-9-r5
 
 ## MinIO server mode. Allowed values: standalone or distributed.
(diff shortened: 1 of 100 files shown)


bitnami/tensorflow-resnet/templates/NOTES.txt
```diff
@@ -27,4 +27,10 @@
 
 2. Test the server with a sample image.
 
-  docker run --rm -it bitnami/tensorflow-resnet bash -c "curl -Lo /tmp/cat.jpg https://tensorflow.org/images/blogs/serving/cat.jpg && resnet_client_cc --server_port=$APP_HOST:$APP_PORT --image_file=/tmp/cat.jpg"
\ No newline at end of file
+  docker run --rm -it bitnami/tensorflow-resnet bash -c "curl -Lo /tmp/cat.jpg https://tensorflow.org/images/blogs/serving/cat.jpg && resnet_client_cc --server_port=$APP_HOST:$APP_PORT --image_file=/tmp/cat.jpg"
+{{- if and (contains bitnami/ .Values.image.repository) (not (regexMatch -r\d+$ .Values.image.tag)) }}
```
tompizmor: this specific chart does not have `.Values.image.` but `.Values.server.image` and `.Values.client.image.`

carrodher: Done

bitnami/minio/values-production.yaml
```diff
@@ -42,7 +42,7 @@ image:
 clientImage:
   registry: docker.io
   repository: bitnami/minio-client
-  tag: latest
+  tag: 2019.5.23-debian-9-r5
```
beltran-rubo: the version is different for the server & client 5.14 vs 5.23

carrodher: Yes, it is because MinIO releases are failing in the pipeline but MinioClient is working fine, that is the reason.
At this moment we are working on a fix for MinIO, once fixed in the pipeline a new release will be triggered and both versions should be the same.
Although this change doesn't affect, I mean, `latest` is pointing to  `2019.5.23-debian-9-r5` in the client; so everything is in the same state than previously.

PR #1209, opened 2019-05-29
[bitnami/etcd] Support auto disaster recovery

### LINKED ISSUE #1125: [bitnami/etcd] Multi replica cluster can't recover when all replicas fail
**Which chart**:
bitnami/etcd:2.2.2

**Description**
When the chart is deployed in a multi replica (with persistent storage) configuration it isn't able to recover when all of the replicas fail.
This scenario is critical when you encounter an infrastructure failure that causes all of the replica pods to fail simultaneously (e.g. all of the pods are scheduled on the same failing Kubernetes node).

(This test was performed on AKS, but with replacing the storageClass below should reproduce in any Kubernetes environment)
1. helm install bitnami/etcd --set-string persistence.storageClass=managed-premium,statefulset.replicaCount=2
2. Wait for all of the replicas to come online
3. kubectl delete pod -l chart=etcd-2.2.2

All of the pods enter a CrashLoopBackOff state.
Both replicas expect to be able to connect to the other. For example, the logs of the first replica:
```
>> kubectl get pods
NAME                    READY   STATUS             RESTARTS   AGE
dunking-rodent-etcd-0   0/1     CrashLoopBackOff   5          6m

>> kubectl logs dunking-rodent-etcd-0
==> The ID of the host is 0
==> Adding member to existing cluster.
client: etcd cluster is unavailable or misconfigured; error #0: dial tcp 10.244.7.150:2379: connect: connection refused
; error #1: dial tcp: lookup dunking-rodent-etcd-1.dunking-rodent-etcd-headless.etcd-test.svc.cluster.local on 10.0.0.10:53: no such host
; error #2: http: no Host in request URL

>> kubectl get statefulsets
NAME                  DESIRED   CURRENT   AGE
dunking-rodent-etcd   2         1         17m
```

The etcd cluster to come back online successfully.

- Output of `helm version`:

```
Client: &version.Version{SemVer:"v2.11.0", GitCommit:"2e55dbe1fdb5fdb96b75ff144a339489417b146b", GitTreeState:"clean"}
Server: &version.Version{SemVer:"v2.11.0", GitCommit:"2e55dbe1fdb5fdb96b75ff144a339489417b146b", GitTreeState:"clean"}
```

```
Client Version: version.Info{Major:"1", Minor:"13", GitVersion:"v1.13.3", GitCommit:"721bfa751924da8d1680787490c54b9179b1fed0", GitTreeState:"clean", BuildDate:"2019-02-04T04:48:03Z", GoVersion:"go1.11.5", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"11", GitVersion:"v1.11.4", GitCommit:"bf9a868e8ea3d3a8fa53cbb22f566771b3f8068b", GitTreeState:"clean", BuildDate:"2018-10-25T19:06:30Z", GoVersion:"go1.10.3", Compiler:"gc", Platform:"linux/amd64"}
```

juan131:
## Description of the change

This PR implements a way to **automatically recover the cluster from a disaster**. On the other hand, it includes the changes below:

- **etcdctl** uses v3 API.
- Labels are adapted to follow the Helm charts best practices.

## Benefits

The etcd cluster will survive to a situation where (N-1)/2 member die (e.g. a K8s node dies)

## Limitations

Only v3 API data is supported.

## Applicable issues

- fixes #1125 

### Current situation

Etcd pods define a "preStop" hook which runs the script below:

```
#!/bin/bash

set -o errexit
set -o pipefail
set -o nounset

# Constants
HOSTNAME="$(hostname -s)"
AUTH_OPTIONS="{{ $etcdAuthOptions }}"
ETCDCTL_ENDPOINTS="{{range $i, $e := until $replicaCount }}{{ $etcdClientProtocol }}://{{ $etcdFullname }}-{{ $e }}.{{ $etcdHeadlessServiceName }}.{{ $releaseNamespace }}.{{ $dnsBase }}:{{ $clientPort }},{{ end }}"
# Remove the last comma "," introduced in the string
export ETCDCTL_ENDPOINTS="${ETCDCTL_ENDPOINTS::-1}"

etcdctl $AUTH_OPTIONS member remove "$(cat "$ETCD_DATA_DIR/member_id")" && rm -rf $ETCD_DATA_DIR/*
```

Therefore, whenever a pod is killed, it tries to remove itself from the etcd cluster.

In a cluster scenario like the one below one (or several) etcd pods could die:

```
|---------------|---------------|---------------|
| Node1         | Node2         | Node3         |
|---------------|---------------|---------------|
|-------|-------|-------|-------|---------------|
| Etcd1 | Etcd2 | Etcd3 |       |               |
|-------|-------|-------|-------|---------------|
|       |       |       |       |               |
|       |       |       |       |               |
```

Depending on the number of members dying we can have the following scenarios:

- Scenario A: one of the members is killed, the "pre-stop" hook is executed and it's removed from the cluster successfully.
- Scenario B: one of the nodes crashes and one or more members are deleted without executing the "pre-stop" hook properly (unable to remove member from cluster).
- Scenario C: **(N-1)/2** etcd members die (e.g. Node1 and Node2 crashes) causing the etcd cluster to collapse!!!!

We can ensure the etcd cluster can survive to every scenario with the following solutions:

- Scenario A: When the stafulset creates a new pod to substitute the member that was killed, **the pod should join the etcd cluster as a new member.**
- Scenario B: When the stafulset creates new pods to substitute the old members, **the pod should reuse the data, claim they're already existing members and update the etcd cluster.**
- Scenario C: When the stafulset creates new pods to substitute the old ones, **the pod should start a "disaster recovery process" restoring the cluster from a snapshot.**

### Auto Disaster Recovery

The most complex solution is the Scenario C where **(N-1)/2** etcd members have died. Here you have the suggested solution to **Automate Disaster Recovery**:

- To restore a cluster from a snapshot, we need to previously automate the creation of snapshots. To do so, a **cron job** is created so it **periodically creates a snapshot of the keyspace**. This snapshot will be (over)written in a **ReadWriteMany** PV.
- Once we have a snapshot, the container logic is modified so it's able to detect the etcd cluster collapsed and it restores it using the snapshot.


tompizmor:
@juan131 I tried deleting just one pod and it was not able to recover it:

Install the chart with: 

```
helm install --name etcd --set statefulset.replicaCount=3 .
```

Then kill the pod-2

```
kubectl delete pod etcd-etcd-2
```

This use case was working before.

juan131:
@tompizmor thanks for reporting it, it works when `disasterRecovery.enabled=true`:

```
$ helm install --name etcd --set statefulset.replicaCount=3 --set disasterRecovery.enabled=true bitnami/etcd
$ kubectl get pod
NAME                       READY   STATUS    RESTARTS   AGE
etcd-etcd-0                2/2     Running   0          58s
etcd-etcd-1                2/2     Running   0          36s
etcd-etcd-2                2/2     Running   0          22s
$ kubectl delete pod etcd-etcd-2
pod "etcd-etcd-2" deleted
$ kubectl get pod
NAME                       READY   STATUS    RESTARTS   AGE
etcd-etcd-0                2/2     Running   0          2m58s
etcd-etcd-1                2/2     Running   0          2m36s
etcd-etcd-2                2/2     Running   0          40s
```

But I need to fix the scenario you mentioned.

juan131:
@tompizmor could you please retry? It should be fixed now

bitnami/etcd/templates/scripts-configmap.yaml
```diff
@@ -0,0 +1,184 @@
...
+        local -r min_endpoints=$((({{ $replicaCount }} + 1)/2))
+
+        for e in "${endpoints_array[@]}"; do
+            if [[ "$e" != "$ETCD_ADVERTISE_CLIENT_URLS" ]] && etcdctl $AUTH_OPTIONS  endpoint health --endpoints="$e" >/dev/null 2>&1; then
+                active_endpoints=$((active_endpoints + 1))
+            fi
+        done
+
+        if [[ -f "/snapshots/.disaster_recovery" ]]; then
+            if [[ $active_endpoints -eq $(({{ $replicaCount }} - 1)) ]]; then
```
javsalgar: It may be worth specifying in the docs that scaling using `kubectl scale` will not work (in order to allow that we would need to query `kube-apiserver` for the number of replicas), only `helm upgrade`. It would be interesting exploring `kube-apiserver` access for future iterations of the chart.

bitnami/etcd/templates/statefulset.yaml
```diff
@@ -27,10 +27,10 @@ spec:
     metadata:
       name: "{{ template "etcd.fullname" . }}"
       labels:
-        app: {{ template "etcd.name" . }}
-        chart: {{ template "etcd.chart" . }}
-        release: {{ .Release.Name | quote }}
-        heritage: {{ .Release.Service | quote }}
+        app.kubernetes.io/name: {{ template "etcd.name" . }}
```
javsalgar: Changing the templates in a StatefulSet would cause helm upgrade from a previous version fail, right?

bitnami/etcd/templates/snapshot-pvc.yaml
```diff
@@ -0,0 +1,18 @@
+{{- if .Values.disasterRecovery.enabled -}}
```
javsalgar: It would also be interesting to allow a user to specify an already existing PVC for snapshotting

javsalgar: It would also be interesting to allow a user to specify an already existing PVC for snapshotting

bitnami/etcd/templates/cronjob.yaml
```diff
@@ -0,0 +1,68 @@
...
+          {{- end }}
+          containers:
+          - name: etcd-snapshotter
+            image: "{{ template "etcd.image" . }}"
+            imagePullPolicy: "{{ .Values.image.pullPolicy }}"
+            command:
+            - /scripts/save-snapshot.sh
+            env:
+            - name: ETCDCTL_API
+              value: "3"
```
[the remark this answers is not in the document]
juan131: I don't think so. The commands used on the scripts used the API v3 format

bitnami/etcd/templates/statefulset.yaml
juan131: Yes, it's documented in the **README.md**
This PR breaks backwards compatibility

bitnami/etcd/templates/snapshot-pvc.yaml
juan131: Sure! I'll try to update the PR adding this feature

bitnami/etcd/templates/scripts-configmap.yaml
juan131: 👍 I'll update the **README.md** warning the users to use `helm upgrade` exclusively

tompizmor:
Thanks, @juan131! That case is already solved, but I have seen another issue that we can easily address:

If you install 3 replicas cluster without disaster recovery and then kill 2 replicas. They show the following error:

```
==> Detected data from previous deployments...
touch: cannot touch '/snapshots/.disaster_recovery': No such file or directory
==> Cluster not responding!!
```

I see that you are printing this error for the same situation but in another part of the code.

```
==> There was no snapshot to perform data recovery!!
```

Should you print a similar user-friendly error for this specific case?

bitnami/etcd/templates/scripts-configmap.yaml
```diff
@@ -0,0 +1,184 @@
...
+                  --initial-cluster $ETCD_INITIAL_CLUSTER \
+                  --initial-cluster-token $ETCD_INITIAL_CLUSTER_TOKEN \
+                  --initial-advertise-peer-urls $ETCD_INITIAL_ADVERTISE_PEER_URLS
+                store_member_id &
+            else
+                echo "==> There was no snapshot to perform data recovery!!"
+                exit 1
+            fi
+{{- else }}
+            exit 1
```
tompizmor: Should we print here that disasterRecovery was disabled and so we cannot recover the cluster?

juan131:
@tompizmor @javsalgar I updated the PR addressing your suggestions

bitnami/etcd/templates/scripts-configmap.yaml
juan131: done

tompizmor:
@juan I see there has been a regression with the latest changes. I have deployed a cluster with 3 replicas and it was not able to recover 1 replica failure: This was the issue:

```
==> Detected data from previous deployments...
==> Updating member in existing cluster...
Error: etcdserver: member not found
```

juan131:
@tompizmor could you give it a second try?

tompizmor:
@juan131 I tried again and it is worked fine. Thanks!

bitnami/etcd/README.md
```diff
@@ -217,22 +223,55 @@ Once the secret is created, you can deploy the helm chart with these options:
...
 
+### To 3.0.0
+
+Backwards compatibility is not guaranteed. The following notables changes were included:
+
+- **etcdctl** uses v3 API.
+- Adds support for auto disaster recovery.
+- Labels are adapted to follow the Helm charts best practices.
+
+To upgrade from previous charts versions, create a snapshot of the keyspace and restore it in a new etcd cluster. Only v3 API data can be restored.
```
javsalgar: I would extend this with the steps to perform

juan131:
@javsalgar as you can check, in the last commit, I provided a way to install a new etcd cluster based on an exiting snapshot

PR #1212, opened 2019-05-30
[bitnami/apache] Remove apache volume

carrodher:
Signed-off-by: Carlos Rodriguez Hernandez <crhernandez@bitnami.com>

The Apache configuration volume (_/bitnami/apache_) has been deprecated, and support for this feature will be dropped in the near future. Until then, the container will enable the Apache configuration from that volume if it exists. By default, and if the configuration volume does not exist, the configuration files will be regenerated each time the container is created. Users wanting to apply custom Apache configuration files are advised to mount a volume for the configuration at _/opt/bitnami/apache/conf_, or mount specific configuration files individually.

You can find more info in the container README: https://github.com/bitnami/bitnami-docker-apache/blob/master/README.md#notable-changes

### DIFF
--- a/bitnami/apache/templates/deployment.yaml
+++ b/bitnami/apache/templates/deployment.yaml
@@ -58,9 +58,6 @@ spec:
           initialDelaySeconds: 5
           timeoutSeconds: 3
           periodSeconds: 5
-        volumeMounts:
-        - name: apache-data
-          mountPath: /bitnami/apache
 {{- if .Values.metrics.enabled }}
       - name: metrics
         image: {{ template "apache.metrics.image" . }}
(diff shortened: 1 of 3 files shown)


bitnami/apache/templates/deployment.yaml
```diff
@@ -58,9 +58,6 @@ spec:
           initialDelaySeconds: 5
           timeoutSeconds: 3
           periodSeconds: 5
```
javsalgar: Now that I think of it. How is the user supposed to mount their own web page? Having just an Apache server deployed does not add much value to the chart, does it?

carrodher: I will create a task to investigate it

PR #1230, opened 2019-06-07
[bitnami/mysql] Unify and document production values

carrodher:


bitnami/mysql/values.yaml
```diff
@@ -28,7 +28,6 @@ image:
 
 service:
   ## Kubernetes service type
-  # nodePort:
```
carrodher: Not sure about this one, it doesn't appear in the production one and it is not documented cc @tompizmor

tompizmor: Is it used in the service? Maybe we forgot to document it..

carrodher: No, it is not used:
```
$ ag 'nodePort' bitnami/mysql
bitnami/mysql/values.yaml
31:  # nodePort:
```

tompizmor: then we should delete it from the values.yaml

PR #1235, opened 2019-06-07
[bitnami/elasticsearch] Unify and document production values

carrodher:


bitnami/elasticsearch/README.md
```diff
@@ -197,6 +197,174 @@ $ helm install --name my-release -f values.yaml bitnami/elasticsearch
 
 > **Tip**: You can use the default [values.yaml](values.yaml). [values-production.yaml](values-production.yaml) has defaults optimized for use in production environments.
 
+### Production configuration
```
juan131: I'd update the "Tip" note with a link to this section. I mean this one:

```
> **Tip**: You can use the default [values.yaml](values.yaml). [values-production.yaml](values-production.yaml) has defaults optimized for use in production environments.
```

bitnami/elasticsearch/values-production.yaml
```diff
@@ -117,16 +125,11 @@ securityContext:
...
-##
-discovery:
-  name: discovery
-
 ## Elasticsearch coordinating-only node parameters
 ##
 coordinating:
   name: coordinating-only
-  replicas: 2
+  replicas: 3
```
juan131: Why this change?

bitnami/elasticsearch/values-production.yaml
```diff
@@ -243,7 +251,7 @@ data:
 ingest:
   enabled: true
   name: ingest
-  replicas: 2
+  replicas: 3
```
juan131: Why this change?

bitnami/elasticsearch/values.yaml
```diff
@@ -18,7 +18,7 @@ image:
   ## Defaults to 'Always' if image tag is 'latest', else set to 'IfNotPresent'
   ## ref: http://kubernetes.io/docs/user-guide/images/#pre-pulling-images
   ##
-  pullPolicy: Always
+  pullPolicy: IfNotPresent
```
juan131: Why did you change the `pullPolicy` in the **values.yaml**?

bitnami/elasticsearch/values.yaml
```diff
@@ -310,7 +307,7 @@ metrics:
     registry: docker.io
     repository: bitnami/elasticsearch-exporter
     tag: 1.0.2
-    pullPolicy: Always
+    pullPolicy: IfNotPresent
```
juan131: Why did you change the `pullPolicy` in the **values.yaml**?

carrodher: Defaults to 'Always' if image tag is 'latest', else set to 'IfNotPresent'
ref: http://kubernetes.io/docs/user-guide/images/#pre-pulling-images

bitnami/elasticsearch/values.yaml
carrodher: Defaults to 'Always' if image tag is 'latest', else set to 'IfNotPresent'
ref: http://kubernetes.io/docs/user-guide/images/#pre-pulling-images

bitnami/elasticsearch/values-production.yaml
carrodher: We are increasing the number of nodes from 2 to 3 in the case of `master` and `data`, why not for `coordinating` and `ingest`?

bitnami/elasticsearch/values-production.yaml
carrodher: We are increasing the number of nodes from 2 to 3 in the case of `master` and `data`, why not for `coordinating` and `ingest`?

PR #1258, opened 2019-06-27
[bitnami/airflow] Update Airflow dependencies

sameersbn:
- postgresql: 5.x.x
 - redis: 8.x.x

**Applicable issues**
- https://github.com/helm/charts/issues/14766

### DIFF
--- a/bitnami/airflow/Chart.yaml
+++ b/bitnami/airflow/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: airflow
-version: 0.1.6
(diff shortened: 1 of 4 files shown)


carrodher:
As PostgreSQL is bumping from a major version in the server itself (not only in the chart version) maybe it produces some breaking changes. It's ok in the sense that we need to update the dependency and we are bumping the Airflow chart version in a major, but if it's the case we should add a note in the proper section of the README

sameersbn:
done!

PR #1283, opened 2019-07-10
Standardize component.name & component.fullname macros

juan131:
Signed-off-by: juan131 <juan@bitnami.com>

**Description of the change**

Standardize `component.name` & `component.fullname` macros with format:

```
{{/*
Expand the name of the chart.
*/}}
{{- define "xxx.name" -}}
{{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" -}}
{{- end -}}

{{/*
Create a default fully qualified app name.
We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
*/}}
{{- define "xxx.fullname" -}}
{{- if .Values.fullnameOverride -}}
{{- .Values.fullnameOverride | trunc 63 | trimSuffix "-" -}}
{{- else -}}
{{- $name := default .Chart.Name .Values.nameOverride -}}
{{- if contains $name .Release.Name -}}
{{- .Release.Name | trunc 63 | trimSuffix "-" -}}
{{- else -}}
{{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" -}}
{{- end -}}
{{- end -}}
{{- end -}}
```
**Checklist** [Place an '[X]' (no spaces) in all applicable fields. Please remove unrelated fields.]

- [x] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [x] Variables are documented in the README.md
- [x] If the chart contains a `values-production.yaml` apart from `values.yaml`, ensure that you implement the changes in both files

PR #1288, opened 2019-07-17
Revert pull request #1283

carrodher:
Changes in #1283 are breaking changes when used in some configurations, for example `helm upgrade` doesn't work using a static IP in the `loadBalancer`. 
According to [semver](http://semver.org/):

Those changes are not backwards-compatible, so we should have increased the _MAJOR_ version instead of the _PATCH_ one. In this PR, we are reverting the breaking changes (https://github.com/bitnami/charts/commit/a95471ae70c6d727be5e8ba70772fd8c96e541e5) so the current _MAJOR_ version will continue working without issues. 
In the same way, we will create a new PR adding again those changes but bumping the chart version in a _MAJOR_

**Process followed:**
```bash
$ git pull
$ git checkout -b revertChanges
$ git revert --mainline 1 a95471ae
# Resolve conflicts (harbor/Chart.yaml, kafka/Chart.yaml and minio/Chart.yaml)
# Bump the patch version in every Chart.yaml
$ git push origin revertChanges
```


javsalgar:
LGTM

PR #1297, opened 2019-07-22
[WIP] Add volumePermissions initContainer

tompizmor:

**Description of the change**

Sometimes the built-in Kubernetes functionality to adjust file permissions in the volumes is not available (`securityContext`). If that is the case and we want to run non-root containers, we have to adjust the permissions of the volume previously with a privileged init container.

**Benefits**

The charts will be more flexible and they could be used in more Kubernetes scenarios.

**Possible drawbacks**

**Applicable issues**

**Additional information**

Changes planned. Review but nor merge


bitnami/cassandra/templates/statefulset.yaml
```diff
@@ -58,6 +58,19 @@ spec:
 {{ toYaml . | indent 8 }}
       {{- end }}
 {{- include "cassandra.imagePullSecrets" . | indent 6 }}
+      {{- if and .Values.volumePermissions.enabled .Values.persistence.enabled }}
```
javsalgar: I presume we checked that we do not need an init container when persistence is disabled, right?

javsalgar: That is, the container in the Dockerfile sets the right permissions already

tompizmor: Yes, that should be the case. If it is not, we should fix the image

bitnami/cassandra/README.md
```diff
@@ -46,81 +46,87 @@ The command removes all the Kubernetes components associated with the chart and
...
-| `metrics.image.pullSecrets`          | Specify docker-registry secret names as an array                                                                                                    | `[]` (does not add image pull secrets to deployed pods)      |
-| `metrics.podAnnotations`             | Additional annotations for Metrics exporter                                                                                                         | `{prometheus.io/scrape: "true", prometheus.io/port: "8080"}` |
-| `metrics.resources`                  | Exporter resource requests/limit                                                                                                                    | `{}`                                                         |
+| Parameter                            | Description                                                                                                                                               | Default                                                      |
```
javsalgar: I would add a section explaining the Volume Permissions init container. 

In addition, in the NOTES I will add a section where, in case the volume container is not enabled, let the user know that, in some platforms, it may be necessary enabling it.

tompizmor: I agree with the README section but not sure about the section in the NOTES. By default, the volumePermission initContainer will be disabled and it is supposed to be used only with a minority of kubernetes versions. 
My concern is that we will add "noise" for user that are not affected by the issue and the could be confused.

javsalgar: Ok, then just the README

PR #1299, opened 2019-07-22
[bitnami/etcd] Fix livenessProbe and readinessProbe

### LINKED ISSUE #1286: [bitnami/etcd] liveness/readiness checks fail for all nodes if one node is down
-------------------------------
    BUG REPORT INFORMATION
-------------------------------

**Which chart**:

etcd

**Description**

The liveness and readiness checks of this chart seem completely broken by design to me when running this as a statefulset on Kubernetes. The script executes the `etcdctl endpoint health` command with all configured nodes listed in the `ETCDCTL_ENDPOINTS` environment variable. The consequence of this is that when one pod is temporarily down (a normal occurrence on Kubernetes), then the liveness and readiness checks of every single pod in the etcd cluster will fail! So any pod going down will tear down the whole cluster.

The idea of liveness and readiness checks on Kubernetes is to have a customized method to determine when a pod needs to be removed/restarted (liveness check) or when a pod should not receive traffic because it's not ready (readiness check). This must be pod-specific, otherwise it defeats the purpose of its existence.

1. Spin up a statefulset with 3 replicas and liveness checks enabled
2. Wait for the cluster to initialize
3. Scale the statefulset down to 2 replicas

All cluster pods go into a crash loop as soon as the next liveness check executes.

The liveness of the remaining 2 pods should be OK and the cluster should continue to serve clients.

**Additional information you deem important:**

For full transparency I'll say that I'm not using Tiller at all because I don't have enough permissions on our corporate Kubernetes cluster for that. I'm only using the templating engine of Helm to create the manifests locally and then submit those via kubectl, but that shouldn't matter for this bug that I'm reporting.

I have tried with and without persistence (I also found separate bugs in that, but if I decide to report those it will be in a separate issue. 

I have `allowNoneAuthentication: true` set during my experiments, but RBAC is enabled.

Metrics and disasterRecovery are off.

I'm using this the bitnami etcd docker image with the tag `3.3.13-debian-9-r44` for this.

- Output of `helm version`:

```
Client: &version.Version{SemVer:"v2.14.2", GitCommit:"a8b13cc5ab6a7dbef0a58f5061bcc7c0c61598e7", GitTreeState:"clean"}
Error: Get https://redacted/api/v1/namespaces/kube-system/pods?labelSelector=app%3Dhelm%2Cname%3Dtiller: Not Found
```

```
Client Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.1", GitCommit:"b7394102d6ef778017f2ca4046abbaa23b88c290", GitTreeState:"clean", BuildDate:"2019-04-08T17:11:31Z", GoVersion:"go1.12.1", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"12", GitVersion:"v1.12.3", GitCommit:"435f92c719f279a3a67808c80521ea17d5715c66", GitTreeState:"clean", BuildDate:"2018-11-26T12:46:57Z", GoVersion:"go1.10.4", Compiler:"gc", Platform:"linux/amd64"}
```

### LINKED ISSUE #1287: [bitnami/etcd] Cluster cannot start with readiness checks configured
-------------------------------
    BUG REPORT INFORMATION
-------------------------------

**Which chart**:

etcd

**Description**

I'm trying to spin up a brand new cluster of etcd (3 replicas), with readiness checks set to enabled. I cannot seem to do it though, because the failing readiness check of the first pod prevents the statefulset from starting the second pod, and as long as the second pod is not started the readiness check cannot succeed. It's a chicken and egg problem.

The readiness checks of this chart seem completely broken by design to me when running this as a statefulset on Kubernetes. The script executes the `etcdctl endpoint health` command with all configured nodes listed in the `ETCDCTL_ENDPOINTS` environment variable. The consequence of this is that when only the first node is up during a normal startup sequence, then the readiness checks can never succeed since it only succeeds when the whole cluster is up and healthy.

The idea of readiness checks on Kubernetes is to have a customized method to determine when a single pod should start to receive traffic because it's ready. This must be pod-specific, otherwise it defeats the purpose of its existence.

1. Install this chart with 3 replicas and readiness checks enabled

The first pod/replica will never be viewed as ready, so the 2nd and 3rd pods will never be started.

All 3 pods should come up in sequence as intended by the rules of the statefulset.

**Additional information you deem important:**

For full transparency I'll say that I'm not using Tiller at all because I don't have enough permissions on our corporate Kubernetes cluster for that. I'm only using the templating engine of Helm to create the manifests locally and then submit those via kubectl, but that shouldn't matter for this bug that I'm reporting.

I have tried with and without persistence (I also found separate bugs in that, but if I decide to report those it will be in a separate issue. 

I have `allowNoneAuthentication: true` set during my experiments, but RBAC is enabled.

Metrics and disasterRecovery are off.

I'm using this the bitnami etcd docker image with the tag `3.3.13-debian-9-r44` for this.

- Output of `helm version`:

```
Client: &version.Version{SemVer:"v2.14.2", GitCommit:"a8b13cc5ab6a7dbef0a58f5061bcc7c0c61598e7", GitTreeState:"clean"}
Error: Get https://redacted/api/v1/namespaces/kube-system/pods?labelSelector=app%3Dhelm%2Cname%3Dtiller: Not Found
```

```
Client Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.1", GitCommit:"b7394102d6ef778017f2ca4046abbaa23b88c290", GitTreeState:"clean", BuildDate:"2019-04-08T17:11:31Z", GoVersion:"go1.12.1", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"12", GitVersion:"v1.12.3", GitCommit:"435f92c719f279a3a67808c80521ea17d5715c66", GitTreeState:"clean", BuildDate:"2018-11-26T12:46:57Z", GoVersion:"go1.10.4", Compiler:"gc", Platform:"linux/amd64"}
```

pantaoran:
**Description of the change**

This solves fundamental problems which resulted in the livenessProbe and readinessProbe features of the chart being broken.

The livenessProbe is fixed by removing the ETCDCTL_ENDPOINTS in the probes.sh script. This makes the check only execute against the one current/local node and not the whole cluster.

The readinessProbe is fixed by configuring the podManagementPolicy to Parallel instead of OrderedReady. 
ETCD is a system that needs to have all nodes up and running during bootstrap, otherwise it will fail. This is in contradiction with the behavior of a statefulSet when readinessProbe is configured, as the next pod will only start once the previous pod's readinessProbe succeeds. I first discarded this idea of changing the podManagementPolicy because I thought it would throw away the benefits of using a statefulSet in the first place. But I have found in the [documentation](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#parallel-pod-management) that:

> This option only affects the behavior for scaling operations. Updates are not affected.

Since this chart doesn't benefit from sequential scaling I think it's an adequate solution. For scaling this chart (beyond the initial config) everything would need to be redeployed anyway since the scripts and environment variables will change, so I conclude that regular scaling shouldn't be done here.

**Benefits**

The probes become usable, the cluster is able to bootstrap and doesn't fail completely if one node goes down.
I also improved a regex because I saw cases where additional whitespace was included due to shorter identifiers. This previously made the regex not match and the node couldn't start.

**Possible drawbacks**

I don't know of any.

**Applicable issues**

  - fixes #1286
  - fixes #1287 

**Additional information**

I wasn't sure what the new chart version should be; I decided to increase the minor rather than the patch version because this PR effectively adds functionality (features that were broken before work now).

**Checklist**

### DIFF
--- a/bitnami/etcd/Chart.yaml
+++ b/bitnami/etcd/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: etcd
-version: 4.0.0
+version: 4.1.0
 appVersion: 3.3.13
 description: etcd is a distributed key value store that provides a reliable way to store data across a cluster of machines
 keywords:
--- a/bitnami/etcd/templates/scripts-configmap.yaml
+++ b/bitnami/etcd/templates/scripts-configmap.yaml
@@ -94,7 +94,7 @@ data:
     ## Check wether the member was succesfully removed from the cluster
     should_add_new_member() {
         return_value=0
-        if (grep -E "^Member [a-z0-9]+ removed from cluster [a-z0-9]+$" "$(dirname "$ETCD_DATA_DIR")/member_removal.log" > /dev/null) || \
+        if (grep -E "^Member[[:space:]]+[a-z0-9]+ removed from cluster [a-z0-9]+$" "$(dirname "$ETCD_DATA_DIR")/member_removal.log" > /dev/null) || \
            ! ([[ -d "$ETCD_DATA_DIR/member/snap" ]] && [[ -f "$ETCD_DATA_DIR/member_id" ]]); then
             rm -rf $ETCD_DATA_DIR/*
         else
@@ -193,9 +193,6 @@ data:
 
     # Constants
     AUTH_OPTIONS="{{ $etcdAuthOptions }}"
-    ETCDCTL_ENDPOINTS="{{range $i, $e := until $replicaCount }}{{ $etcdClientProtocol }}://{{ $etcdFullname }}-{{ $e }}.{{ $etcdHeadlessServiceName }}.{{ $releaseNamespace }}.{{ $dnsBase }}:{{ $clientPort }},{{ end }}"
-    # Remove the last comma "," introduced in the string
-    export ETCDCTL_ENDPOINTS="$(sed 's/,/ /g' <<< $ETCDCTL_ENDPOINTS | awk '{$1=$1};1' | sed 's/ /,/g')"
 
     etcdctl $AUTH_OPTIONS endpoint health >/dev/null 2>&1
 {{- if .Values.disasterRecovery.enabled }}
--- a/bitnami/etcd/values-production.yaml
+++ b/bitnami/etcd/values-production.yaml
@@ -57,7 +57,7 @@ statefulset:
   ## Pod management policy
   ## https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#pod-management-policies
   ##
-  podManagementPolicy: OrderedReady
+  podManagementPolicy: Parallel
   ## Number od replicas
   ##
   replicaCount: 3
--- a/bitnami/etcd/values.yaml
+++ b/bitnami/etcd/values.yaml
@@ -51,7 +51,7 @@ statefulset:
   ## Pod management policy
   ## https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#pod-management-policies
   ##
-  podManagementPolicy: OrderedReady
+  podManagementPolicy: Parallel
   ## Number od replicas
   ##
   replicaCount: 1


javsalgar:

LGTM

bhiravabhatla:
Might have to relook at rbac configuration - I guess - because we dont guarantee the order of deployment any more - 

```go
configure_rbac() {
        # When there's more than one replica, we can assume the 1st member
        # to be created is "{{ $etcdFullname }}-0" since a statefulset is used
        if [[ -n "${ROOT_PASSWORD:-}" ]] && [[ "$HOSTNAME" == "{{ $etcdFullname }}-0" ]]; then
            echo "==> Configuring RBAC authentication!" 1>&3 2>&4
            etcd &
            ETCD_PID=$!
            while ! etcdctl $AUTH_OPTIONS member list; do sleep 1; done
            echo "$ROOT_PASSWORD" | etcdctl $AUTH_OPTIONS user add root --interactive=false
            etcdctl $AUTH_OPTIONS auth enable
            kill "$ETCD_PID"
            sleep 5
```

FraPazGal:
Hello @bhiravabhatla,

Thanks for bringing this up! To follow this properly, could you please create a separate issue for this? If you want, you can also try creating a PR to fix the issue, we will be happy to review!

PR #1319, opened 2019-07-27
[bitnami/harbor] Add 'behindReverseProxy' config

niiku:
This configuration value enables to exclude the 'proxy_set_header X-Forwarded-Proto $scheme;' setting from the nginx configuration so Harbor can be used behind other reverse proxy such as HAProxy.

The exclusion of the 'X-Forwarded-Proto' field from the header is already suggested in the current nginx config, so I just added a simple `if` around every occurrence of this field.

**Checklist**

### DIFF
--- a/bitnami/harbor/Chart.yaml
+++ b/bitnami/harbor/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: harbor
-version: 2.1.1
+version: 2.2.0
 appVersion: 1.8.1
 description: Harbor is an an open source trusted cloud native registry project that stores, signs, and scans content
 keywords:
--- a/bitnami/harbor/templates/nginx/configmap-http.yaml
+++ b/bitnami/harbor/templates/nginx/configmap-http.yaml
@@ -53,8 +53,9 @@ data:
           proxy_set_header X-Real-IP $remote_addr;
           proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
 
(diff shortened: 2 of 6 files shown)


bitnami/harbor/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: harbor
-version: 2.1.1
+version: 2.1.2
```
alemorcuq: This should be a minor according to [SemVer](https://semver.org/):
>
>1. MAJOR version when you make incompatible API changes,
>2. MINOR version when you add functionality in a backwards-compatible manner, and
>3. PATCH version when you make backwards-compatible bug fixes.
>

bitnami/harbor/templates/nginx/configmap-http.yaml
```diff
@@ -53,8 +53,10 @@ data:
           proxy_set_header X-Real-IP $remote_addr;
           proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
 
+          {{- if not .Values.nginx.behindReverseProxy }}
           # When setting up Harbor behind other proxy, such as an Nginx instance, remove the below line if the proxy already has similar settings.
```
alemorcuq: No reason for this comments anymore. Could you please remove them?

alemorcuq:
Hi, @niiku!

Thank you very much for this PR! I've left some comments for you.

Regards,
Alejandro

niiku:
Hi @alemorcuq thank you for your feedback. I applied the requested changes and squashed them in the existing commit.

PR #1332, opened 2019-08-05
[bitnami/etcd] Various small improvements

pantaoran:
**Description of the change**

- _helpers.tpl: I don't see why the path of the dataDir should be different depending on whether persistence is enabled or not. This simplifies things.
- scripts-configmap.yaml: It was my experience that the volumeMount one level above the dataDir erased many files during the mounting of the empty persistence directory. My changes make it so that only the actual dataDir directly is mounted to prevent erasing files at the parent level. I also fixed some problems with unescaped double quotes.
- statefulset.yaml: Fixed some hardcoded paths to use the existing variable. Added an empty /tmp directory volume for environments like mine where containers are fully read-only by policy and only mounted volumes are exempted from that. Adding this volume will not change anything for writeable environments, but it's needed for read-only environments.
- values-*.yaml: I added the possibility to have additional securityContext options taken into consideration by using the entire dictionary in the template. In my environment for example the `runAsNonRoot: true` key is mandatory. For this I had to move the `enable` key out of the dictionary into its own key.

**Benefits**

See description above.

**Possible drawbacks**

None that I know of.

**Applicable issues**

I didn't open issues for these first.

**Checklist**


bitnami/etcd/values.yaml
```diff
@@ -129,10 +129,13 @@ auth:
 ## Kubernetes Security Context
 ## https://kubernetes.io/docs/tasks/configure-pod-container/security-context/
 ##
+securityContextEnabled: true
 securityContext:
-  enabled: true
```
juan131: Please revert this change. It includes `backwards` incompatibilities.

bitnami/etcd/templates/statefulset.yaml
```diff
@@ -280,6 +281,8 @@ spec:
         configMap:
           name: {{ template "etcd.fullname" . }}-scripts
           defaultMode: 0755
+      - name: tmp
+        emptyDir: {}
```
juan131: Why do we need this volume?

bitnami/etcd/templates/_helpers.tpl
```diff
@@ -79,12 +79,8 @@ Return the proper etcd client protocol
 Return the proper etcd data dir
 */}}
 {{- define "etcd.dataDir" -}}
-{{- if .Values.persistence.enabled -}}
-{{- print "/bitnami/etcd/data" -}}
-{{- else -}}
 {{- print "/opt/bitnami/etcd/data" -}}
 {{- end -}}
-{{- end -}}
```
juan131: Did you test this change? What happens when you disable the persistence?

bitnami/etcd/templates/statefulset.yaml
juan131: I just saw the explanation in your summary. Ignore this comment please

bitnami/etcd/values.yaml
pantaoran: Sorry could you elaborate? How is it incompatible?

bitnami/etcd/templates/_helpers.tpl
pantaoran: Worked fine when I quickly deployed without persistence for a smoke test.

bitnami/etcd/values.yaml
juan131: It modifies the structure of the **values.yaml**. 

It's true it doesn't imply a major change but a user who's using a specific **values.yaml** needs to adapt to continue enabling the securityContext. Therefore, if we want to go ahead with the change (at least) we need to mention the change in the README.md (Upgrading section) to alert users about the change.

andresbono:
Hi @pantaoran, what do you think about @juan131's comment here https://github.com/bitnami/charts/pull/1332#discussion_r312022136?

andresbono:
Hi again @pantaoran, just pinging you again.

If we don't get an answer, we may apply some minor changes to the PR and merge it so that your work is not lost!

raoofm:
@pantaoran @juan131 @andresbono just checking if this was supposed to be closed or just the bot accidentally did it.

carrodher:
It was not accidentally, there are some incomplete requests, so we can't merge this PR until the suggestions are applied. To avoid the accumulation of old Issues/PRs, the bot closes it after some days without answers.

PR #1352, opened 2019-08-12
[bitnami/mariadb-galera] Update LDAP configuration

andresbono:

**Description of the change**

Update LDAP configuration

### DIFF
--- a/bitnami/mariadb-galera/README.md
+++ b/bitnami/mariadb-galera/README.md
@@ -101,6 +101,9 @@ The following table lists the configurable parameters of the MariaDB Galera char
 | `ldap.binddn`                        | LDAP bind DN                                                                                                                                                | `nil`                                                             |
 | `ldap.bindpw`                        | LDAP bind password                                                                                                                                          | `nil`                                                             |
 | `ldap.bslookup`                      | LDAP base lookup                                                                                                                                            | `nil`                                                             |
+| `ldap.nss_initgroups_ignoreusers`    | LDAP ignored users                                                                                                                                          | `root,nslcd`                                                      |
+| `ldap.scope`                         | LDAP search scope                                                                                                                                           | `nil`                                                             |
+| `ldap.tls_reqcert`                   | LDAP TLS check on server certificates                                                                                                                       | `nil`                                                             |
 | `mariadbConfiguration`               | Configuration for the MariaDB server                                                                                                                        | `_default values in the values.yaml file_`                        |
 | `configurationConfigMap`             | ConfigMap with the MariaDB configuration files (Note: Overrides `mariadbConfiguration`). The value is evaluated as a template.                              | `nil`                                                             |
 | `initdbScripts`                      | Dictionary of initdb scripts                                                                                                                                | `nil`                                                             |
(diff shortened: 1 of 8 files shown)


bitnami/mariadb-galera/templates/statefulset.yaml
```diff
@@ -154,9 +154,21 @@ spec:
               name: {{ template "mariadb-galera.fullname" . }}
             {{- end }}
               key: ldap-bindpw
+        - name: MARIADB_LDAP_NSS_INITGROUPS_IGNOREUSERS
+          value: {{ .Values.ldap.nss_initgroups_ignoreusers | default "root,nslcd" | quote }}
```
juan131: Move these defaults to **values.yaml** (uncomment them) and remove the default command. Please note that unless `.Values.ldap.enabled` is `true`, they won't be used.

bitnami/mariadb-galera/values.yaml
```diff
@@ -158,14 +158,14 @@ ldap:
...
-  # scope:
-  # tls_reqcert:
+  uri:
+  base:
+  binddn:
+  bindpw:
+  bslookup:
+  nss_initgroups_ignoreusers: root,nslcd
+  scope:
+  tls_reqcert:
```
juan131: What values are mandatory? I'd add a `mariadb-galera.validateValues` function (see other charts such as MinIO) where you can validate the values. 

For instance, if someone sets `ldap.enabled=true` and doesn't provide the `ldap.uri` we could show an error explaining the expected values when enabling LDAP.

andresbono:
Please @juan, commandeer the PR.

bitnami/mariadb-galera/README.md
```diff
@@ -183,17 +186,23 @@ LDAP support can be enabled in the chart by specifying the `ldap.` parameters wh
...
 For example:
 
 ```bash
 $ helm install --name my-release bitnami/mariadb-galera \
     --set ldap.enabled="true" \
-    --set ldap.url="ldap://my_ldap_server" \
-    --set ldap.base="dc=example,dc=org" \
-    --set ldap.binddn="cn=admin,dc=example,dc=org" \
+    --set ldap.uri="ldap://my_ldap_server" \
+    --set ldap.base="dc=example\,dc=org" \
```
marcosbc: Why is the backslash necessary? AFAIK a comma is a valid character and does not need to be escaped.

bitnami/mariadb-galera/templates/_helpers.tpl
```diff
@@ -147,5 +147,65 @@ imagePullSecrets:
...
+{{- if $message -}}
+{{-   printf "\nVALUES VALIDATION:\n%s" $message | fail -}}
+{{- end -}}
+{{- end -}}
+
+{{/* Validate values of MariaDB Galera - must provide passwords when forced */}}
+{{- define "mariadb-galera.validateValues.rootPassword" -}}
+{{- if and .Values.rootUser.forcePassword (empty .Values.rootUser.password) -}}
+mariadb-galera: rootUser.password
+    A MariaDB Database Root Password is required (`rootUser.forcePassword=true` is set)
```
marcosbc: If these errors are meant to be shown in a CLI environment, we should not need to escape code with "`".

bitnami/mariadb-galera/values-production.yaml
```diff
@@ -158,11 +158,14 @@ ldap:
   ## Enable LDAP support
   ##
   enabled: false
-  # uri:
-  # base:
-  # binddn:
-  # bindpw:
-  # bslookup:
+  uri:
```
marcosbc: Why are we un-commenting this if the values are empty?

bitnami/mariadb-galera/values.yaml
```diff
@@ -158,11 +158,14 @@ ldap:
   ## Enable LDAP support
   ##
   enabled: false
-  # uri:
-  # base:
-  # binddn:
-  # bindpw:
-  # bslookup:
+  uri:
```
marcosbc: IMHO we should not leave these values un-commented in the non-production `values.yaml` file.

marcosbc:
The chart version should also be updated to a minor version, if not major in case it could break any existing installations.

It seems like it was already broken before so maybe a minor version change is OK.

bitnami/mariadb-galera/README.md
juan131: Removing backslashes

bitnami/mariadb-galera/templates/_helpers.tpl
juan131: Applying this suggestion

bitnami/mariadb-galera/values-production.yaml
juan131: Because they're ignored as far as `ldap.enabled=false` and we want to avoid to use the `default` function in the templates.

PR #1357, opened 2019-08-13
[bitnami/tensorflow-resnet] Enable Rest API and configure metrics

andresbono:

**Description of the change**

- Enable Rest API for Tensorflow Serving
- Configure Tensorflow Serving to expose Prometheus metrics without needing the exporter.

**Checklist**
- [x] ~If the chart contains a `values-production.yaml` apart from `values.yaml`, ensure that you implement the changes in both files~


bitnami/tensorflow-resnet/values.yaml
```diff
@@ -86,44 +87,11 @@ readinessProbe:
...
-  ## Metrics exporter pod Annotation and Labels
   podAnnotations:
     prometheus.io/scrape: "true"
-    prometheus.io/port: "9118"
-    ## Metrics exporter resource requests and limits
-    ## ref: http://kubernetes.io/docs/user-guide/compute-resources/
-    ##
-  # resources: {}
+    prometheus.io/path: "/monitoring/prometheus/metrics"
+    prometheus.io/port: "8501"
```
juan131: Correct me if I'm wrong but we're setting the Prometheus annotations in the Tensorflow pods, aren't we? 

I wonder if it makes sense to create a specific service `metrics-svc.yaml` with the annotations and redirecting traffic to the restApiPort. If we do so, and we add the annotations to this service, we don't nee to add the annotations to the pods.

andresbono: I not sure about it. The port `8501` exposes the Rest API of Tensorflow serving. Among other things, it includes metrics for Prometheus in the `/monitoring/prometheus/metrics`, but it also has other endpoints not related to metrics, i.e. `/models/resnet`.

juan131: 🤔 in this case, I'd leave it as it's. Thanks for the explanation.

PR #1373, opened 2019-08-22
[bitnami/elasticsearch] Bump exporter version and fix issue with initContainer

juan131:

**Description of the change**

This PR bumps the version of `elasticsearch-exporter` and adapts the port to be used. On the other hand, it fixes the `initContainer` since `sysctl` needs the package `procps`.

**Checklist**


juan131:
> Why do you need procps? Otherwise lgtm.

You need both `systemd` and `procps` system to use `systctl`

juan131:
Port fixed in the annotations. Good catch

PR #1374, opened 2019-08-22
[bitnami/harbor] Add Harbor Redis password when using redis chart dep…

miguelaeh:
…endency

Signed-off-by: Miguel A. Cabrera Minagorri <macabrera@bitnami.com>

**Description of the change**
There was a bug with the Redis password as it was only set to Harbor for the external Redis while if using the Redis Chart dependency and setting a Redis password ,the Harbor chart was not using it so it couldn't connect to Redis.

**Benefits**
This PR allow to deploy Harbor Chart using the Redis Chart dependency with a custom password.

**Possible drawbacks**

**Applicable issues**

  - fixes #

**Additional information**

### DIFF
--- a/bitnami/harbor/templates/_helpers.tpl
+++ b/bitnami/harbor/templates/_helpers.tpl
@@ -227,8 +227,11 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this 
 {{- end -}}
 
 {{- define "harbor.redis.rawPassword" -}}
-  {{- if and (eq .Values.redis.enabled false) .Values.externalRedis.password -}}
+  {{- if and (not .Values.redis.enabled) .Values.externalRedis.password -}}
     {{- .Values.externalRedis.password -}}
+  {{- end -}}
+  {{- if and .Values.redis.enabled .Values.redis.password -}}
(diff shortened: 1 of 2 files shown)


bitnami/harbor/templates/_helpers.tpl
```diff
@@ -230,6 +230,9 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this
   {{- if and (eq .Values.redis.enabled false) .Values.externalRedis.password -}}
     {{- .Values.externalRedis.password -}}
   {{- end -}}
+  {{- if and (eq .Values.redis.enabled true) .Values.redis.password -}}
```
juan131: I don't like using `eq` function to compare with booleans:
```
-   {{- if and (eq .Values.redis.enabled true) .Values.redis.password -}}
+  {{- if and .Values.redis.enabled .Values.redis.password -}}
```
It also applies to line 230:
```diff
-  {{- if and (eq .Values.redis.enabled false) .Values.externalRedis.password -}}
+  {{- if and (not .Values.redis.enabled) .Values.externalRedis.password -}}
```

PR #1390, opened 2019-09-04
[bitnami/zookeeper] Update container entrypoint to make it compatible with bash logic

andresbono:
**Description of the change**

- Updates container entrypoint to make it compatible with bash logic. Related: https://github.com/bitnami/bitnami-docker-zookeeper/pull/15
- Adds a new variable that the ZooKeeper container supports.

### DIFF
--- a/bitnami/zookeeper/Chart.yaml
+++ b/bitnami/zookeeper/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: zookeeper
-version: 4.3.4
+version: 4.4.4
 appVersion: 3.5.5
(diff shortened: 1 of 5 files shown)


bitnami/zookeeper/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: zookeeper
-version: 4.3.4
+version: 4.4.4
```
marcosbc: We should be setting this as a major version (`5.0.0`) due to there being a breaking change (the entrypoint is not compatible anymore).

PR #1398, opened 2019-09-05
[bitnami/kafka] Fix name of environment variables set in chart

marcosbc:
**Description of the change**

The environment variables for the Kafka container were fixed some time ago. However, they were left outdated in the chart. This PR fixes this issue.

**Benefits**

Environment variables set in `values.yaml` are not properly detected by Kafka.

**Possible drawbacks**

**Applicable issues**

  - Fixes https://github.com/bitnami/bitnami-docker-kafka/issues/62

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/kafka/Chart.yaml
+++ b/bitnami/kafka/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: kafka
-version: 6.0.0
+version: 5.4.0
 appVersion: 2.3.0
 description: Apache Kafka is a distributed streaming platform.
 keywords:
--- a/bitnami/kafka/Chart.yaml
+++ b/bitnami/kafka/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: kafka
-version: 5.4.0
+version: 5.3.0
 appVersion: 2.3.0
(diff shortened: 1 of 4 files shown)


marcosbc:
The reasoning is this is a breaking change in the sense that it could break misconfigured installations that were working before, but with the new changes it may fail.

This is because our container sets any environment variable starting with `KAFKA_CFG_` to the configuration file. I.e. `KAFKA_CFG_LOGS_DIR` sets `logs.dir` to `server.key`.

However, some the variables set in the chart were wrong before meaning they were not being properly taken into account from `values.yaml`. That means it did not work before, and adding it now means a user with a wrong chart configuration could end up with a broken deployment.

For example, in the previous chart we were setting `log.dirs` with what was added to `values.yaml`. Due to the correct value being `logs.dir` it was not being applied, and if a user had put a wrong path it may have worked, but with the new deployment it could fail.

bitnami/kafka/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: kafka
-version: 5.2.4
+version: 6.0.0
```
juan131: I'd bump the minor version.

PR #1405, opened 2019-09-09
Add git hooks 

carrodher:
**Description of the change**

Add git hooks to reduce issues.

Error message after changes in kubeapps and/or upstreamed charts (commit scope):
Screen Shot 2019-09-09 at 18 01 08

Error message after `helm lint` and `helm install --dry-run` validation (push scope):
Screen Shot 2019-09-09 at 17 59 44

### DIFF
--- a/githooks/pre-push/helm-lint
+++ b/githooks/pre-push/helm-lint
@@ -1,0 +1,34 @@
+#!/bin/bash
+
+currentBranch="$(git rev-parse --abbrev-ref HEAD)"
+originCommit="$(git rev-parse --short "$(git merge-base master "$currentBranch")")"
+filesToBePushed="$(git diff --name-only "$originCommit")"
+failed=false
+
+for chartName in $( cut -d'/' -f1,2 <<< "$filesToBePushed" | uniq ); do
+    printf '\033[01;33mValidating %s:\n\033[0m' "$chartName"
+    chartPath="$(git rev-parse --show-toplevel)"/"$chartName"
+
+    printf '\033[0;34m- Running helm lint %s\n\033[0m' "$chartPath"
+    if helm lint "$chartPath"; then
+        printf '\033[0;32m\U2705 helm lint %s\n\n\033[0m' "$chartPath"
+    else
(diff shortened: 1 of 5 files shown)


githooks/pre-push/helm-lint
```diff
@@ -0,0 +1,37 @@
...
+        if helm install --dry-run "$chartPath"; then
+            printf '\033[0;32m\U2705 helm install --dry-run %s\n\n\033[0m' "$chartPath"
+        else
+            printf '\033[0;31m\U1F6AB helm install --dry-run %s failed. Pull Request cancelled.\n\n\033[0m' "$chartPath"
+            failed=true
+        fi
+    fi
+done
+
+if $failed ; then
```
alemorcuq: You could make `failed` a numeric argument `0|1` and just do `exit $failed` at the end.

carrodher:
> Does it make sense to merge the `kubeapps` and the `upstreamed` hooks into the same file? LGTM either way.

I was thinking about it, but in the future, both logic can diverge, I think it's not a problem to maintain both files at this time.

githooks/pre-push/helm-lint
```diff
@@ -0,0 +1,34 @@
...
+    # Avoid running 'helm lint|install' when modified dirs are not charts
+    if [[ $chartName = bitnami/* ]]; then
+        printf '\033[01;33mValidating %s:\n\033[0m' "$chartName"
+        chartPath="$(git rev-parse --show-toplevel)"/"$chartName"
+
+        printf '\033[0;34m- Running helm lint %s\n\033[0m' "$chartPath"
+        if helm lint "$chartPath"; then
+            printf '\033[0;32m\U2705 helm lint %s\n\n\033[0m' "$chartPath"
+        else
+            printf '\033[0;31m\U1F6AB helm lint %s failed. Pull Request cancelled.\n\n\033[0m' "$chartPath"
```
andresbono: Should this say `Push cancelled`? The same in the below line

carrodher: mmm... well, being purists it can be, although the objective of the push is to create a PR, there are some cases that not

PR #1439, opened 2019-09-26
[bitnami/etcd] Fix conflict between ETCDCTL_ENDPOINTS env var and --endpoints

matt-everett:
**Description of the change**

etcdctl does not seem to support the use of both the `ETCDCTL_ENDPOINTS` environment variable and the `--endpoints` command line parameter in the `etcdctl endpoint health` call, it gives an error about how the setting is being shadowed.

Have unset the `ETCDCTL_ENDPOINTS` environment variable in a subshell so the unset only affects the execution of the `etcdctl endpoint health` call. This appears to work as intended now with minimal code changes.

Also added debug output in the setup.sh script to indicate the member id has successfully been stored.

**Benefits**

I tested that a pod in a 3 node cluster recovers from failure by running `kill -15 1` from within the pod to simulate a crash. The pod recovers correctly. Before this change, the `etcdctl endpoint health` call always fails and the pod goes into a crash loop.

The debug output gives greater confidence that the member id has been stored correctly.

**Possible drawbacks**

None known

**Applicable issues**

None

**Additional information**

None

**Checklist** 

### DIFF
--- a/bitnami/etcd/templates/scripts-configmap.yaml
+++ b/bitnami/etcd/templates/scripts-configmap.yaml
@@ -76,7 +76,7 @@ data:
         local -r min_endpoints=$((({{ $replicaCount }} + 1)/2))
 
         for e in "${endpoints_array[@]}"; do
-            if [[ "$e" != "$ETCD_ADVERTISE_CLIENT_URLS" ]] && (unset -v ETCDCTL_ENDPOINTS; exec etcdctl $AUTH_OPTIONS  endpoint health --endpoints="$e"); then
+            if [[ "$e" != "$ETCD_ADVERTISE_CLIENT_URLS" ]] && (unset -v ETCDCTL_ENDPOINTS; etcdctl $AUTH_OPTIONS  endpoint health --endpoints="$e"); then
                 active_endpoints=$((active_endpoints + 1))
             fi
         done
@@ -132,14 +132,12 @@ data:
               --initial-cluster-token $ETCD_INITIAL_CLUSTER_TOKEN \
               --initial-advertise-peer-urls $ETCD_INITIAL_ADVERTISE_PEER_URLS 1>&3 2>&4
             store_member_id & 1>&3 2>&4
-
         else
             echo "==> There was no snapshot to perform data recovery!!" 1>&3 2>&4
             exit 1
         fi
 {{- else }}
         store_member_id & 1>&3 2>&4
-
         configure_rbac
 {{- end }}
     else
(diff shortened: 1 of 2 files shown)


bitnami/etcd/templates/scripts-configmap.yaml
```diff
@@ -130,13 +131,15 @@ data:
               --initial-cluster $ETCD_INITIAL_CLUSTER \
               --initial-cluster-token $ETCD_INITIAL_CLUSTER_TOKEN \
               --initial-advertise-peer-urls $ETCD_INITIAL_ADVERTISE_PEER_URLS 1>&3 2>&4
-            store_member_id &
+            store_member_id & 1>&3 2>&4
+
```
miguelaeh: Please remove extra lines. This apply to the rest of changes too.

bitnami/etcd/templates/scripts-configmap.yaml
```diff
@@ -75,7 +76,7 @@ data:
         local -r min_endpoints=$((({{ $replicaCount }} + 1)/2))
 
         for e in "${endpoints_array[@]}"; do
-            if [[ "$e" != "$ETCD_ADVERTISE_CLIENT_URLS" ]] && etcdctl $AUTH_OPTIONS  endpoint health --endpoints="$e"; then
+            if [[ "$e" != "$ETCD_ADVERTISE_CLIENT_URLS" ]] && (unset -v ETCDCTL_ENDPOINTS; exec etcdctl $AUTH_OPTIONS  endpoint health --endpoints="$e"); then
```
miguelaeh: It is not clear to me why is the `exec` necessary. Can you clarify it?

bitnami/etcd/templates/scripts-configmap.yaml
matt-everett: Good idea, done!

bitnami/etcd/templates/scripts-configmap.yaml
matt-everett: I copied a snippet to do this. I think you're right that the 'exec' is unnecessary as the extra child process isn't a burden and your suggestion simplifies the code. The chart seems to work fine without so I've removed the 'exec'.

matt-everett:
Thanks for the feedback @miguelaeh! I hope those updates are ok. Let me know if you need anything else.

ncsibra:
Hi, it looks like a critical issue, because the disaster recovery script does not work correctly when this will be merged?

miguelaeh:
Sorry for the delay, we just merge it into master.

ncsibra:
Thanks!

matt-everett:
Thanks @miguelaeh!

PR #1466, opened 2019-10-08
[bitnami/kibana] Reduce the number of default Elasticsearch nodes

alemorcuq:
**Description of the change**
Add some parameters to the `values.yaml` in order to be able to customize the number of Elasticsearch nodes that will be deployed:

- `elasticsearch.master.replicas`
- `elasticsearch.coordinating.replicas`
- `elasticsearch.data.replicas`

It also sets their default values to `1` instead of `2`.

**Benefits**
Much less resources needed to deploy the chart with the default settings.

**Possible drawbacks**
Existing deployments may see an unexpected reduction in the number of Elasticsearch nodes when upgrading.

**Additional information**
As the `values-production.yaml` disables the bundled Elasticsearch by default, I've set the same values as in the `values.yaml`.

**Checklist** 

### DIFF
--- a/bitnami/kibana/values-production.yaml
+++ b/bitnami/kibana/values-production.yaml
@@ -314,6 +314,15 @@ elasticsearch:
   ## Enable sysctl image for the bundled Elasticsearch
   sysctlImage:
     enabled: false
+  ## Elasticsearch master-eligible node parameters
+  master:
+    replicas: 2
+  ## Elasticsearch coordinating-only node parameters
+  coordinating:
+    replicas: 2
+  ## Elasticsearch data node parameters
+  data:
+    replicas: 2
   ## Properties when enabled is false
   external:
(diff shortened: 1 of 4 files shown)


bitnami/kibana/values-production.yaml
```diff
@@ -314,6 +314,15 @@ elasticsearch:
   ## Enable sysctl image for the bundled Elasticsearch
   sysctlImage:
     enabled: false
+  ## Elasticsearch master-eligible node parameters
+  master:
```
javsalgar: I wouldn't change it for the production values.

alemorcuq: Note that Elasticsearch is disabled by default in the production configuration. I've changed it nonetheless.

javsalgar:
LGTM

PR #1505, opened 2019-10-18
Add pre-push test to avoid breaking existing schemas

andresmgot:
**Description of the change**

Add a check that, in case the chart to push has a schema, validates that the schema is valid with the default values.

**Applicable issues**

  - ref https://github.com/kubeapps/kubeapps/issues/1182

### DIFF
--- a/githooks/pre-push/helm-lint
+++ b/githooks/pre-push/helm-lint
@@ -28,6 +28,11 @@ for chartName in $( cut -d'/' -f1,2 <<< "$filesToBePushed" | uniq ); do
             printf '\033[0;31m\U1F6AB helm template %s failed. Push cancelled.\n\n\033[0m' "$chartPath"
             failed=1
         fi
+
+        # Verify that the schema is still valid
+        if [[ -f "${chartPath}/values.schema.json" ]]; then
+            helm3 install --generate-name --dry-run $chartPath >> /dev/null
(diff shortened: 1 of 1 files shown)


carrodher:
This kind of check is executed locally, so we need to ensure that the team has `helm v3` installed and included in the PATH as `helm3` if not, the hook will fail and the PRs will be blocked.
Before merging it, can you update the handbook and sent some messages to the team in order to ensure that they have the correct tools?

andresmgot:
TBH this is a test that can be either bypassed (if the githooks are not enabled) or that will fail and will need to be removed when `helm3` becomes the default. If there is already a test somewhere that check that the charts are compatible with Helm 3 we can skip this. WDYT?

carrodher:
In the content team, everyone should have githooks enabled and it should be executed before each PR (unless someone is bypassing it 🔪). If we want to add this test at this time we need to ensure that the command works for the team. I think the easy way is notifying the team via Slack with a link to the handbook with the instructions.

andresmgot:
> If we want to add this test at this time we need to ensure that the command works for the team

Yes, that's my point. I don't think it's worth it to add the test here because it only affect to half of the charts (not the ones upstreamed) so I'd rather not give a false feeling of testing this when it's only done partially.

PR #1512, opened 2019-10-22
README.md PoC

carrodher:
- There are two common sections across charts ("Rolling VS Immutable tags" and "Production configuration and horizontal scaling"). The first one is present in all the charts, while the second one is present when the chart contains a `values-production.yaml`. As both sections are very usual, I placed just after the table.
- After those common sections, there are other sections specific per chart.
- Avoid CLI commands when possible
```
.
├── Parameters
│  └── (Table)
└── Configuration and installation details
   ├── Rolling VS Immutable tags
   ├── Production configuration and horizontal scaling
   └── Forwarding the logs to another service
```

### DIFF
--- a/bitnami/fluentd/README.md
+++ b/bitnami/fluentd/README.md
@@ -44,7 +44,7 @@ $ helm delete my-release
 
 The command removes all the Kubernetes components associated with the chart and deletes the release. Use the option `--purge` to delete all history too.
 
-## Configuration
+## Parameters
 
 The following tables lists the configurable parameters of the kibana chart and their default values.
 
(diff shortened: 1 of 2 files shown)


bitnami/fluentd/README.md
```diff
@@ -234,39 +260,3 @@ $ helm install bitnami/fluentd \
     --set aggregator.extraEnv[1].name=ELASTICSEARCH_PORT \
     --set aggregator.extraEnv[1].value=your-port-here \
 ```
```
carrodher: IMHO it's interesting to maintain those CLI commands to show users how to use this configmap

bitnami/fluentd/README.md
```diff
@@ -234,39 +260,3 @@ $ helm install bitnami/fluentd \
...
-```
-
-To horizontally scale this chart once it has been deployed:
-
-```console
-$ helm upgrade my-release bitnami/fluentd \
-  -f ./values-production.yaml \
-  --set statefulset.replicaCount=3
-```
-
```
carrodher: Replaced commands by text

bitnami/fluentd/README.md
beltran-rubo: I would remove the command and mention the parameters the user needs to configure.

bitnami/fluentd/README.md
```diff
@@ -140,6 +140,32 @@ $ helm install --name my-release -f values.yaml bitnami/fluentd
...
+
+- Enable prometheus to access fluentd metrics endpoint:
+```diff
+- metrics.enabled: false
++ metrics.enabled: true
+```
+
+To horizontally scale this chart once it has been deployed you can upgrade the deployment using a new value for the `aggregator.replicaCount` parameter.
+
+### [Rolling VS Immutable tags](https://docs.bitnami.com/containers/how-to/understand-rolling-tags-containers/)
```
beltran-rubo: I would move this one to the first in the "Configuration and installation details" as it is included to all of them

PR #1526, opened 2019-10-24
[bitnami/grafana] adds servicemonitor support

sameersbn:
**Description of the change**

Adds Prometheus Operator ServiceMonitor support

### DIFF
--- a/bitnami/grafana/values-production.yaml
+++ b/bitnami/grafana/values-production.yaml
@@ -279,3 +279,38 @@ affinity: {}
 ## ref: https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/
 ##
 podAnnotations: {}
+
+## Prometheus metrics
+##
+metrics:
+  enabled: false
+
+  ## Prometheus Operator ServiceMonitor configuration
+  ##
+  service:
+    annotations:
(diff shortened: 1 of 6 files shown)


bitnami/grafana/values-production.yaml
```diff
@@ -277,3 +277,38 @@ affinity: {}
...
 ##
 podAnnotations: {}
+
+## Prometheus metrics
+##
+metrics:
+  enabled: false
+
+  ## Prometheus Operator ServiceMonitor configuration
+  ##
```
juan131: I think this comment doesn't correspond this block but the "serviceMonitor" one.

bitnami/grafana/values-production.yaml
```diff
@@ -277,3 +277,38 @@ affinity: {}
...
+  service:
+    annotations:
+      prometheus.io/scrape: "true"
+      prometheus.io/port: "{{ .Values.service.port }}"
+      prometheus.io/path: "/metrics"
```
juan131: Where are you using these `metrics.service.annotations`? It seems there's no metrics service using them

sameersbn: fixed!

PR #1542, opened 2019-10-28
[bitnami/fluentd] Add an option to disable aggregator

carrodher:
**Description of the change**

Add an option to disable the aggregator

**Checklist** 


bitnami/fluentd/templates/forwarder-configmap.yaml
```diff
@@ -77,6 +77,7 @@ data:
       @type kubernetes_metadata
     </filter>
 
+    {{- if .Values.aggregator.enabled -}}
```
alemorcuq: If the aggregator is disabled we should do something else with the logs. I think we can throw them to the stdout of each forwarder (see `aggregator-configmap.yaml`) and then add a note in the NOTES.txt about this new default configuration when aggregators are disabled.

bitnami/fluentd/templates/NOTES.txt
```diff
@@ -3,13 +3,20 @@
...
     kubectl get all -l "app.kubernetes.io/name={{ include "fluentd.name" . }},app.kubernetes.io/instance={{ .Release.Name }}"
-{{- if and .Values.aggregator.enabled (not .Values.aggregator.configMap) -}}
+{{ if and .Values.aggregator.enabled (not .Values.aggregator.configMap) }}
   Logs are captured on each node by the forwarder pods and then sent to the aggregator pods. By default, the aggregator pods send the logs to the standard output.
   You can see all the logs by running this command:
 
     kubectl logs -l "app.kubernetes.io/component=aggregator"
   
   You can mount your own configuration files to the aggregators and the forwarders. For example, this is useful if you want to forward the aggregated logs to Elasticsearch or another service.
+{{- else }}
```
alemorcuq: I would change this for:

`{{- else if and (not .Values.aggregator.enabled) (not .Values.forwarder.configMap) }}`

As if the user is loading their own configuration, the forwarders may be doing something different.

PR #1548, opened 2019-10-30
[bitnami/postgresql] Fix error when running init-container

mgrachev:
**Description of the change**

When I use another image like `postgres:11.5` I always get the following error:
```
Error: failed to create containerd task: OCI runtime create failed: container_linux.go:337: starting container process caused "exec: \"sh\": executable file not found in $PATH": unknown
```

Using the full path to `sh` (`/bin/sh`) fixes that problem.

**Benefits**

People can use other images without get the above error.

**Possible drawbacks**

No.

**Applicable issues**

**Additional information**

**Checklist** 

### DIFF
--- a/upstreamed/postgresql/Chart.yaml
+++ b/upstreamed/postgresql/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: postgresql
-version: 6.5.4
+version: 6.5.5
 appVersion: 11.5.0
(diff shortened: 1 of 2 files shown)


alemorcuq:
Thank you very much for your PR, @mgrachev.

All the charts under the `upstreamed/` directory are mirrored from the [Helm Stable repository](https://github.com/helm/charts/tree/master/stable). You can open a PR there and we will gladly review and merge it.

Issue #1557, opened 2019-11-05
[fluentd]: Only run aggregator part


davidkarlsen:
**Which chart**:
fluentd

**Description**

Be able to run aggregator only.
Often folks use fluent-bit on the nodes, and hence only want to run fluentd as aggregator.

**Describe the results you received:**

Forwarder daemonset will be created - we don't want that.

**Describe the results you expected:**

Only aggregator sts.

**Additional information you deem important (e.g. issue happens only occasionally):**

alemorcuq:
Hi, @davidkarlsen! I can see why an user would want this and I agree with you. I will make a PR supporting it.

alemorcuq:
We just merged #1558, implementing this feature. Let us know if this works for you.

davidkarlsen:
Thanks - I will!

davidkarlsen:
Works like a charm!

PR #1558, opened 2019-11-06
[bitnami/fluentd] Implement new features

alemorcuq:
**Description of the change**
Adds some features to the Fluentd chart:

  - Allow defining custom containerPorts in both the forwarders and the
aggregators
  - Add services for the forwarders and the aggregator
  - Allow exposing custom ports in the services
  - Add an option to deploy only the aggregator

**Benefits**
- Allowing the definition of custom ports and exposing them makes the user able to get logs from other sources, like syslog, for example.
- Deploying only the aggregator is useful for people that are already getting the logs from the nodes (e.g. with fluent-bit) and only want to aggregate them.

**Applicable issues**
  - #1557

**Checklist** 

Signed-off-by: Alejandro Moreno <amoreno@bitnami.com>


bitnami/fluentd/templates/aggregator-svc.yaml
```diff
@@ -0,0 +1,32 @@
+{{- if .Values.aggregator.service.ports }}
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ include "fluentd.fullname" . }}-aggregator
+  labels: {{- include "fluentd.labels" . | nindent 4 }}
+    app.kubernetes.io/component: aggregator
+  {{- if .Values.aggregator.service.annotations }}
+  annotations: {{- tpl (toYaml .Values.aggregator.service.annotations) $ | nindent 4 }}
```
juan131: Create the macro below at `_helpers.tpl`:
```
{{/*
Usage:
{{ include "fluentd.tplValue" ( dict "value" .Values.path.to.the.Value "context" $) }}
*/}}
{{- define "fluentd.tplValue" -}}
        {{- tpl .value .context }}
    {{- else }}
    {{- end }}
{{- end -}}
```
And use:
```
  annotations: {{- include "fluentd.tplValue" (dict "value" .Values.aggregator.service.annotations "context" $) | nindent 4 }}
```

bitnami/fluentd/values-production.yaml
```diff
@@ -59,6 +60,68 @@ forwarder:
...
+      #   protocol: TCP
+      #   nodePort: 31514
+      # tcp:
+      #   port: 24224
+      #   targetPort: tcp
+      #   protocol: TCP
+
+    ## loadBalancerIP for the forwarders service (optional, cloud specific)
+    ## ref: http://kubernetes.io/docs/user-guide/services/#type-loadbalancer
+    # loadBalancerIP:
```
juan131: Follow conventions for comments:

```
## loadBalancerIP for the forwarders service (optional, cloud specific)
## ref: http://kubernetes.io/docs/user-guide/services/#type-loadbalancer
##
# loadBalancerIP:
```

PR #1561, opened 2019-11-06
[bitnami/nginx-ingress-controller] Add fullnameOverride for default backend

T1loc:
Enable fullnameOverride for the default-backend


carrodher:

T1loc:
Thanks for your answer, 
So if I understand properly the new chart version should be 5.1.0, right ?
I will edit the name of the MR also, it's about the fullnameOverride and not the nameoverride

carrodher:
Yes, a minor version bump works fine in this case. Thanks

PR #1563, opened 2019-11-07
[bitnami/prometheus-operator] helm v3 compatibility

sameersbn:
It appears helm 3 does not create the CRD resources before any other checks are run, leading to the following error while installing the chart `Error: unable to build kubernetes objects from release manifest: [unable to recognize "": no matches for kind "Alertmanager" in version "monitoring.coreos.com/v1", unable to recognize "": no matches for kind "Prometheus" in version "monitoring.coreos.com/v1", unable to recognize "": no matches for kind "ServiceMonitor" in version "monitoring.coreos.com/v1"]`

A new feature of helm 3 is the addition of a `crds/` directory where the CRDS manifests can be added. Helm 3 will install these manifests before the chart manifests are installed, thereby mitigating the scenario we are currently facing with crds. But if we move all the crd definitions to the `crds/` directory, we will no longer be able to deploy the chart to helm 2.

To get around this ~we are making a copy of all the manifests in the `crds/` directory which will be ignored by helm 2 making the chart compatible with both helm versions at the expense of duplication~, we have created stub templates in the `templates/prometheus-operator` directory that source the crd definitions from the `crds/` directory.

**Checklist** 

**edit: updated description**


juan131:
Hi @sameersbn 

> To get around this we are making a copy of all the manifests in the crds/ directory which will be ignored by helm 2 making the chart compatible with both helm versions at the expense of duplication, we have created stub templates in the templates/prometheus-operator directory that source the crd definitions from the crds/ directory.

Could you please detailed what's a "stub templates"? I don't see anything in the `templates/prometheus-operator` directory sourcing the CRD definitions.

sameersbn:
> I don't see anything in the `templates/prometheus-operator` directory sourcing the CRD definitions.

Take a look at the changes in this file for example. Towards the bottom you should see the line `{{ .Files.Get "crds/crd-podmonitor.yaml" }}`, which is essentially sourcing the file 

 https://github.com/bitnami/charts/pull/1563/files#diff-696c2bbe2f4803892d3e5a135ba8117dR2

juan131:
Wouldn't it try to create the CRD(s) twice when using Helm 3? First, when it discovers the `crd` directory. Secondly, when it sees the "stub templates".

sameersbn:
it will evaluate the manifests and the resulting manifests will be the same as currently deployed. so k8s will not apply the changes

PR #1567, opened 2019-11-07
[bitnami/kibana] Remove ES as Kibana dependency

carrodher:
**Description of the change**

Remove Elasticsearch as dependency bundled in Kibana. From now on, Kibana will be included as a dependency in the ES chart.

**Additional information**

⚠️ Please, review it but not merge the PR because we need to review what happens with marketplaces where `helm install` should work by default. In this case, as it is shown in the README, you need to use the already running ES host and port.

**Checklist** 


carrodher:
I just created a new PR after some Git mess...
liada

PR #1573, opened 2019-11-08
[bitnami/prometheus-operator] Remove `helm.sh/hook-delete-policy` on crds

carrodher:
The `"helm.sh/hook-delete-policy": "before-hook-creation"` annotation results in existing crds being deleted while the chart is being deployed leading to the error:

```
the server could not find the requested resource (post alertmanagers.monitoring.coreos.com)
```


carrodher:
Closed because those changes are part of other PR in progress

PR #1574, opened 2019-11-08
[bitnami/elasticsearch] Remove `helm.sh/hook-delete-policy` on hooks

carrodher:
The `"helm.sh/hook-delete-policy": "before-hook-creation"` annotation results in existing hooks being deleted while the chart is being deployed leading to the error:

```
the server could not find the requested resource (post alertmanagers.monitoring.coreos.com)
```


carrodher:
Smae as https://github.com/bitnami/charts/pull/1573

PR #1581, opened 2019-11-11
[bitnami/fluentd] Use different securityContext for aggregator and forwarder pods

juan131:

**Description of the change**

This PR allows using different securityContexts for aggregator and forwarder pods. It also sets the **UID** and **fsGroup** for forwarder pods to **0** by default, so it can be configured to retrieve system-wide logs.

**Benefits**

Configuring Fluentd to retrieve system-wide logs.

**Possible drawbacks**

None

**Checklist**


bitnami/fluentd/values.yaml
```diff
@@ -39,6 +39,15 @@ clusterDomain: cluster.local
...
   enabled: true
+
+  ## K8s Security Context for forwarder pods
+  ## https://kubernetes.io/docs/tasks/configure-pod-container/security-context/
+  ##
+  securityContext:
+    enabled: true
+    runAsUser: 0
+    fsGroup: 0
+
```
alemorcuq: On the `README` you are saying that the default values are 1001. I assume the `README` is wrong.

bitnami/fluentd/values-production.yaml
```diff
@@ -187,6 +196,14 @@ aggregator:
   ##
   replicaCount: 2
 
+  ## K8s Security Context for forwarder pods
+  ## https://kubernetes.io/docs/tasks/configure-pod-container/security-context/
+  ##
+  securityContext:
+    enabled: true
+    runAsUser: 1001
+    fsGroup: 1001
```
sameersbn: in the `values.yaml` this is configured to have uid `0`, while in the `values-production,yaml` it's `1001`. Was it intended to be this way?

bitnami/fluentd/values-production.yaml
```diff
@@ -39,6 +39,15 @@ clusterDomain: cluster.local
...
   enabled: true
+
+  ## K8s Security Context for forwarder pods
+  ## https://kubernetes.io/docs/tasks/configure-pod-container/security-context/
+  ##
+  securityContext:
+    enabled: true
+    runAsUser: 1001
+    fsGroup: 1001
+
```
alemorcuq: This is different than in the `values.yaml` file, if that is intended please document the differences on the README.

bitnami/fluentd/values-production.yaml
juan131: It was not intended. Both should be using `0`.

PR #1583, opened 2019-11-11
[bitnami/kafka] Use port name instead of the value

carrodher:
As a continuation of https://github.com/bitnami/charts/pull/1582

It is not possible to use the port value; instead of that, we should use the port name:
```
spec.endpoints.port in body must be of type string instead of "integer"
```


bitnami/kafka/templates/kafka-exporter.yaml
```diff
@@ -34,7 +34,8 @@ spec:
           - --kafka.server={{ template "kafka.fullname" . }}:{{ .Values.service.port }}
           - --web.listen-address=:{{ .Values.metrics.kafka.port }}
         ports:
-          - containerPort: {{ .Values.metrics.kafka.port }}
+        - name: kafka-port
+          containerPort: {{ .Values.metrics.kafka.port }}
```
juan131: Indentation

PR #1584, opened 2019-11-11
[bitnami/elasticsearch] configure `minimum_master_nodes` to avoid split brain

sameersbn:
**Description of the change**
ref: https://www.elastic.co/guide/en/elasticsearch/reference/6.4/modules-node.html#split-brain

**Checklist** 

### DIFF
--- a/bitnami/elasticsearch/templates/master-statefulset.yaml
+++ b/bitnami/elasticsearch/templates/master-statefulset.yaml
@@ -85,6 +85,8 @@ spec:
           {{- $elasticsearchMasterFullname := include "elasticsearch.master.fullname" . }}
           {{- $replicas := int .Values.master.replicas }}
           value: {{range $i, $e := until $replicas }}{{ $elasticsearchMasterFullname }}-{{ $e }} {{ end }}
+        - name: ELASTICSEARCH_MINIMUM_MASTER_NODES
+          value: {{ add (div .Values.master.replicas 2) 1 | quote }}
         {{- if .Values.plugins }}
         - name: ELASTICSEARCH_PLUGINS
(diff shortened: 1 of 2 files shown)


bitnami/elasticsearch/templates/master-statefulset.yaml
```diff
@@ -82,9 +82,9 @@ spec:
         - name: ELASTICSEARCH_CLUSTER_HOSTS
           value: {{ template "elasticsearch.discovery.fullname" . }}
         - name: ELASTICSEARCH_CLUSTER_MASTER_HOSTS
-          {{- $elasticsearchMasterFullname := include "elasticsearch.master.fullname" . }}
-          {{- $replicas := int .Values.master.replicas }}
-          value: {{range $i, $e := until $replicas }}{{ $elasticsearchMasterFullname }}-{{ $e }} {{ end }}
+          value: {{ template "elasticsearch.master.fullname" . }}-0
```
juan131: I don't understand this change. There can be several master node replicas, why do you hardcode it to the `-0` replica?

sameersbn: I inspected the `elasticsearch.yml` file and found that it only list one at any given time (i guess there is a bug).

juan131: Do you mean a bug in ES or our setup bash logic? Anyway, I don't think we should change this but fix the issue in that case.

sameersbn: okay.. i'll remove this bit

sameersbn: reverted this change. please take a look

PR #1635, opened 2019-11-20
[bitnami/nginx] Ability to add labels for ingress

nirsv:
I use traefik as ingress controller. Traefik uses service labels for labelSelector. I want to have the ability to have additional labels for nginx's ingress

### DIFF
--- a/bitnami/nginx/templates/ingress.yaml
+++ b/bitnami/nginx/templates/ingress.yaml
@@ -8,7 +8,7 @@ metadata:
     chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
     release: "{{ .Release.Name }}"
     heritage: "{{ .Release.Service }}"
-  {{- range $key, $value := .Values.ingress.labels }}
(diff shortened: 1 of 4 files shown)


bitnami/nginx/templates/ingress.yaml
```diff
@@ -8,6 +8,9 @@ metadata:
     chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
     release: "{{ .Release.Name }}"
     heritage: "{{ .Release.Service }}"
+  {{- range $key, $value := .Values.ingress.labels }}
```
juan131: Could you switch the parameter from `.Values.ingress.labels` to `.Values.ingress.selectors`?
Please also update the **values.yaml** including the new parameter and update the list of parameters in the README.md file.
Finally, you'll have to bump the minor version in the **Chart.yaml**.

nirsv: HI Juan.

Issue #1513, opened 2019-10-22
Single-node etcd stuck in CrashLoopBackOff after restart


jwalton:
**Which chart**:

etcd v4.3.10

**Description**

If you bring up a single-node etcd cluster, then terminate the pod, when it comes back up it will be stuck in CrashLoopBackOff.  Logs are:

```
==> Bash debug is off
==> Detected data from previous deployments...
==> Cluster not responding!!
==> Disaster recovery is disabled, the cluster cannot be recovered!
```

**Steps to reproduce the issue:**

With minikube and helm 3:

1. `helm install testetcd --set persistence.storageClass="standard" --set persistence.size="256Mi" --set auth.rbac.enabled=false ./charts/etcd-4.3.10.tgz`
2. Wait for etcd to come up
3. `kubectl delete pod testetcd-0`

**Describe the results you received:**

etcd will be stuck in crash loop.

**Describe the results you expected:**

etcd should come back up.

**Additional information you deem important (e.g. issue happens only occasionally):**

**Version of Helm and Kubernetes**:

- Output of `helm version`:

```
version.BuildInfo{Version:"v3.0.0-beta.4", GitCommit:"7ffc879f137bd3a69eea53349b01f05e3d1d2385", GitTreeState:"dirty", GoVersion:"go1.13.1"}
```

- Output of `kubectl version`:

```
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.1", GitCommit:"d647ddbd755faf07169599a625faf302ffc34458", GitTreeState:"clean", BuildDate:"2019-10-02T23:49:20Z", GoVersion:"go1.12.9", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.0", GitCommit:"2bd9643cee5b3b3a5ecbd3af49d09018f0773c77", GitTreeState:"clean", BuildDate:"2019-09-18T14:27:17Z", GoVersion:"go1.12.9", Compiler:"gc", Platform:"linux/amd64"}
```

rafariossaa:
Hi @jwalton, 
  I am trying to reproduce it.

rafariossaa:
The problem is that etcd it thought to be run with more than one node, to be able to rsync with the others nodes.
If you "kill" one node you can see that it is not able to recover from the disaster.

```$ kubectl logs testetcd-0 
==> Bash debug is off
==> Detected data from previous deployments...
[...]

PR #1642, opened 2019-11-21
[bitnami/etcd] redirect output to new_member_envs

### LINKED ISSUE #1640: [bitnami/etcd]
**Which chart**:
bitnami/etcd version 4.4.3

**Description**
Deployed a etcd cluster with 3 replicas. If a pod restarts (e.g. triggered by a rolling update or a manual pod delete) it is unable to rejoin the cluster. 

The pod keeps restarting with the following output:
```
==> Detected data from previous deployments...
rm: cannot remove '/bitnami/etcd/member_removal.log': No such file or directory
==> Updating member in existing cluster...
2019-11-21 10:13:52.649453 I | pkg/flags: recognized and used environment variable ETCD_INITIAL_CLUSTER_STATE=new
...
2019-11-21 10:13:52.675047 E | etcdserver: the member has been permanently removed from the cluster
2019-11-21 10:13:52.675059 I | etcdserver: the data-dir used by this member must be removed.
2019-11-21 10:13:52.675092 E | etcdserver: publish error: etcdserver: request cancelled
2019-11-21 10:13:52.675106 I | etcdserver: aborting publish because server is stopped
```

Unexpected `ETCD_INITIAL_CLUSTER_STATE=new`, as it should be joining a existing cluster.
As the pod gets restarted I might have missed the log of the very first start. So this might just be the log of restart > 1

The cluster still has 2 healthy members, so it should be able to recover on its own as the quorum is 2. So, I expect a rolling update to work, and to recover from a single member failure.

sboschman:
Saving the env vars to file combined with stdout/stderr redirection does not seem to work. The `new_member_envs` stays empty and `ETCD_INITIAL_CLUSTER_STATE` is always `NEW` triggering the wrong path in the setup script.

fixes #1640

### DIFF
--- a/bitnami/etcd/Chart.yaml
+++ b/bitnami/etcd/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: etcd
-version: 4.4.3
+version: 4.4.4
 appVersion: 3.4.3
 description: etcd is a distributed key value store that provides a reliable way to store data across a cluster of machines
 keywords:
--- a/bitnami/etcd/templates/scripts-configmap.yaml
+++ b/bitnami/etcd/templates/scripts-configmap.yaml
@@ -163,7 +163,7 @@ data:
 {{- end }}
         elif should_add_new_member; then
             echo "==> Adding new member to existing cluster..." 1>&3 2>&4
-            etcdctl $AUTH_OPTIONS member add "$HOSTNAME" --peer-urls="{{ $etcdPeerProtocol }}://${HOSTNAME}.{{ $etcdHeadlessServiceName }}.{{ .Release.Namespace }}.svc.{{ $clusterDomain }}:{{ $peerPort }}" | grep "^ETCD_" > "$ETCD_DATA_DIR/new_member_envs" 1>&3 2>&4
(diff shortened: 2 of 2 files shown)


sameersbn:
@sboschman I'm curious to know if this change actually made a difference to resolve the issue? I wonder of the pipe operation followed by the redirect somehow causes the issue. Could you be able to test it out with this instead:

```bash
(etcdctl $AUTH_OPTIONS member add "$HOSTNAME" --peer-urls="{{ $etcdPeerProtocol }}://${HOSTNAME}.{{ $etcdHeadlessServiceName }}.{{ .Release.Namespace }}.svc.{{ $clusterDomain }}:{{ $peerPort }}" | grep "^ETCD_" > "$ETCD_DATA_DIR/new_member_envs") 1>&3 2>&4
```

sboschman:
Since #1650 it has become harder to fix my etcd clusters. It depends on which member you are trying to get up and running again what goes wrong. But setup.sh gets itself into a state it can't recover from. Steps to get it running again involve removing the data dirs from etcd-1 and/or etcd-2, running etcdctl to remove an unstarted member and multiple hacks in setup.sh. One thing is getting the new_member_envs file actually filled (before #1650 this seemed to be the only change needed). E.g. at https://github.com/bitnami/charts/blob/master/bitnami/etcd/templates/scripts-configmap.yaml#L136 it tries to save the member id of etcd-2, but etcd-2 is not a member yet, so member_id is empty and later on something fails on this missing member_id. So I added https://github.com/bitnami/charts/blob/master/bitnami/etcd/templates/scripts-configmap.yaml#L165 to it, so I can rejoin etcd-2 to an existing cluster.

juan131:
@sameersbn there are several known issues with the current logic to "automagically" recover the etcd cluster for each of the different scenarios.

I already created a task to deeply investigate/fix this issue but we didn't have the chance to work on it yet.

More info at: https://github.com/bitnami/charts/issues/1513

PR #1668, opened 2019-11-26
[bitnami/cassandra] Fix indentation for cassandra statefulset pod annotations

mohansitaram:
When the cassandra chart is built with the metrics.enabled=true flag, the template is rendered like so:

        heritage: Tiller
      annotations:        prometheus.io/port: "8080"
        prometheus.io/scrape: "true"

which is not valid.yaml. This PR fixes the indentation for this.

### DIFF
--- a/bitnami/cassandra/Chart.yaml
+++ b/bitnami/cassandra/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: cassandra
-version: 4.1.8
+version: 4.1.9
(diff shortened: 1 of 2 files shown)


alemorcuq:
Thank you very much for the fix, @mohansitaram. Looks good, there's only one thing pending.

mohansitaram:
Thanks @alemorcuq . Increased the chart version.

alemorcuq:
Thank you, @mohansitaram !

PR #1677, opened 2019-11-28
[bitnami/nginx] Add JSON Schema to NGINX chart

juan131:

**Description of the change**

This PR includes a basic JSON schema for the NGINX chart. It doesn't cover all the possible values of the chart but just the ones that, in my opinion, are more likely to be modified.

This is an example of the file rendered by Kubeapps:

Screenshot 2019-11-28 at 15 42 55

**Checklist**

### DIFF
--- a/bitnami/nginx/values.schema.json
+++ b/bitnami/nginx/values.schema.json
@@ -1,0 +1,78 @@
+{
+  "$schema": "http://json-schema.org/schema#",
+  "type": "object",
+  "properties": {
+    "ingress": {
+      "type": "object",
+      "form": true,
+      "title": "Ingress details",
+      "properties": {
+        "enabled": {
+          "type": "boolean",
+          "form": true,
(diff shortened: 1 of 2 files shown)


bitnami/nginx/values.schema.json
```diff
@@ -0,0 +1,78 @@
...
+        "certManager": {
+          "type": "boolean",
+          "form": true,
+          "title": "Enable TLS annotations via cert-manager",
+          "description": "Set this to true in order to add the corresponding annotations for cert-manager",
+          "hidden": {
+            "condition": false,
+            "value": "ingress.enabled"
+          }
+        },
```
andresmgot: I removed this field from the WordPress because it was confusing. It seems that just enabling this enables TLS but it's actually needed to go to the advanced form to change the `tls` section (to modify the host/secret there).

PR #1679, opened 2019-11-28
[bitnami/jenkins] Add JSON Schema

juan131:

**Description of the change**

This PR includes a basic JSON schema for the Jenkins chart. It doesn't cover all the possible values of the chart but just the ones that, in my opinion, are more likely to be modified.

This is an example of the file rendered by Kubeapps:

Screenshot 2019-11-28 at 16 01 43

**Checklist**

### DIFF
--- a/bitnami/jenkins/values.schema.json
+++ b/bitnami/jenkins/values.schema.json
@@ -80,16 +80,6 @@
           "title": "Use a custom hostname",
           "description": "Enable the ingress resource that allows you to access the Jenkins installation."
         },
-        "certManager": {
-          "type": "boolean",
-          "form": true,
-          "title": "Enable TLS annotations via cert-manager",
-          "description": "Set this to true in order to add the corresponding annotations for cert-manager",
-          "hidden": {
-            "condition": false,
-            "value": "ingress.enabled"
-          }
-        },
         "hostname": {
           "type": "string",
           "form": true,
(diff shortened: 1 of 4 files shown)


bitnami/jenkins/values.schema.json
```diff
@@ -0,0 +1,118 @@
...
+              "render": "slider",
+              "title": "CPU Request",
+              "sliderMin": 10,
+              "sliderMax": 2000,
+              "sliderUnit": "m"
+            }
+          }
+        }
+      }
+    },
```
andresmgot: The problem specifying this is that there is not a good default value. If there is not a default value, the minimum will be used by default (which is probably not a good value to leave). My recommendation is to set proper default values for `resources.requests` (or remove this section)

bitnami/jenkins/values.schema.json
```diff
@@ -0,0 +1,118 @@
...
+        "certManager": {
+          "type": "boolean",
+          "form": true,
+          "title": "Enable TLS annotations via cert-manager",
+          "description": "Set this to true in order to add the corresponding annotations for cert-manager",
+          "hidden": {
+            "condition": false,
+            "value": "ingress.enabled"
+          }
+        },
```
andresmgot: The same issue I told you about `certManager`, just setting this to true doesn't give you TLS support

bitnami/jenkins/values.schema.json
juan131: Adding default values for `resources.requests`

PR #1684, opened 2019-11-30
[bitnami/prometheus-operator] Fix exporters.enabled

carrodher:
**Description of the change**

Add `exporters.enabled` parameter to the _values*.yaml_ that is a necessary condition to install node-exporter:
https://github.com/bitnami/charts/blob/master/bitnami/prometheus-operator/requirements.yaml#L5

Parameter disabled in the _values.yaml_ to not modify the default behavior and enabled in the _values-production.yaml_.

**Checklist** 

### DIFF
--- a/bitnami/prometheus-operator/values-production.yaml
+++ b/bitnami/prometheus-operator/values-production.yaml
@@ -754,6 +754,7 @@ alertmanager:
 
 ## Exporters
 exporters:
+  enabled: true
   node-exporter:
     ## Enable node-exporter
     enabled: true
--- a/bitnami/prometheus-operator/Chart.yaml
+++ b/bitnami/prometheus-operator/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: 0.34.0
 description: The Prometheus Operator for Kubernetes provides easy monitoring definitions for Kubernetes services and deployment and management of Prometheus instances.
 name: prometheus-operator
-version: 0.6.0
+version: 0.6.1
 keywords:
 - prometheus
 - alertmanager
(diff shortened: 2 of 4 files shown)


bitnami/prometheus-operator/values-production.yaml
```diff
@@ -754,6 +754,7 @@ alertmanager:
 
 ## Exporters
 exporters:
+  enabled: true
```
sameersbn: Because of the way Helm resolves the `conditions` in the `requirements.yaml` file I deliberately left out the `exporters.enabled` flag from the `values.yaml` field.

> Condition - The condition field holds one or more YAML paths (delimited by commas). If this path exists in the top parent’s values and resolves to a boolean value, the chart will be enabled or disabled based on that boolean value. Only the first valid path found in the list is evaluated and if no paths exist then the condition has no effect.
> ref: https://v3.helm.sh/docs/topics/charts/

Consider the chart can deploy multiple exporters and that we only wanted to disable the `node-exporter`. The logical thing would be to do the following:

```bash
helm install . --set exporters.enabled=true --set exporters.node-exporter.enabled=false
```

But this will not work because helm would evaluate the condition to `true` because `exporters.enabled` is `true`, it would not consider what `exporters.node-exporter.enabled` is set to.

--- 

Given the above behavior, my design intention for `exporters.enabled` flag was to acts as a global flag to disable **ALL** exporters. Since the `exporters.enabled` path does not exist in the `values.yaml` it will skip the condition. Therefore if user wants to disable specific exporters they would need to only specify `exporters.node-exporter.enabled` flag.

If you think `node-exporter` should be disabled by default, then we should just flip the condition in the `values.yaml`.

PR #1703, opened 2019-12-04
[bitnami/airflow] Custom Airflow Workers

cccs-cat001:
**Description of the change**
Firstly, I added the ability to specify a secret in kubernetes which holds SSH keys for the dag repo. That way the repo can be private and you won't have to worry about logging in. 
Secondly, I made it possible to clone more than just the dag repo into the workers. 
Thirdly, it's now possible to spawn a custom worker, using whatever image you want, cloning any repos you want, setting any environment variables you want, and mounting any volumes you want.

**Benefits**
This will be useful so that now your workers can be more broad. One issue I was facing with the original worker was that it was very locked down. So this will allow you to make your own worker, give it whatever libraries you want, whatever configurations you want, etc. This change is in addition to having the plain airflow worker too, so now you have multiple workers, each specialized for their own purpose. 

**Possible drawbacks**
None that I'm aware of. 

**Applicable issues**
We use these changes because we host our dags in a private repo, so we needed the ssh keys to be mounted into all the airflow pods it creates. We also use papermill on airflow to schedule parameterized jupyter notebooks,and the simplest solution to that we thought up was to have our notebooks in a git repo (private also) and clone them into the workers similar to how the dags are done. 
Then we came up with this idea to create a custom worker, pulling from a separate queue. These separate workers can do their specific tasks on their queues. But.. you're probably wondering how these workers are pulling from a queue. Well, I'm creating a pull request next in the bitnami-docker-airflow-worker repo with that fix too. [will put the link here](https://github.com/bitnami/bitnami-docker-airflow-worker/pull/2)

**Additional information**

**Checklist** 


bitnami/airflow/templates/deployment-scheduler.yaml
```diff
@@ -58,6 +58,8 @@ spec:
           volumeMounts:
             - name: git-cloned-dag-files
               mountPath: /dags
+            - name: dag-git-keys
+              mountPath: /root/.ssh
```
alemorcuq: This volume should only be mounted if `airflow.cloneDagFilesFromGit.sshKeySecret` is set.

bitnami/airflow/templates/deployment-scheduler.yaml
```diff
@@ -201,6 +205,10 @@ spec:
       {{- if .Values.airflow.cloneDagFilesFromGit.enabled }}
         - name: git-cloned-dag-files
           emptyDir: {}
+        - name: dag-git-keys
+          secret:
+            secretName: {{ .Values.airflow.cloneDagFilesFromGit.sshKeySecret }}
+            defaultMode: 0600
```
alemorcuq: This volume should only exist if `airflow.cloneDagFilesFromGit.sshKeySecret` is set.

bitnami/airflow/templates/deployment-scheduler.yaml
```diff
@@ -73,6 +75,8 @@ spec:
           volumeMounts:
             - name: git-cloned-dag-files
               mountPath: /dags
+            - name: dag-git-keys
+              mountPath: /root/.ssh
```
alemorcuq: Same here.

bitnami/airflow/templates/deployment-web.yaml
```diff
@@ -58,6 +58,8 @@ spec:
           volumeMounts:
             - name: git-cloned-dag-files
               mountPath: /dags
+            - name: dag-git-keys
+              mountPath: /root/.ssh
```
alemorcuq: Same here.

bitnami/airflow/templates/deployment-web.yaml
```diff
@@ -73,6 +75,8 @@ spec:
...
```
alemorcuq: Same here.

bitnami/airflow/templates/deployment-web.yaml
```diff
@@ -231,6 +245,10 @@ spec:
         {{- if .Values.airflow.cloneDagFilesFromGit.enabled }}
         - name: git-cloned-dag-files
           emptyDir: {}
+        - name: dag-git-keys
+          secret:
+            secretName: {{ .Values.airflow.cloneDagFilesFromGit.sshKeySecret }}
+            defaultMode: 0600
```
alemorcuq: Same here.

bitnami/airflow/templates/deployment-web.yaml
```diff
@@ -170,6 +174,16 @@ spec:
               {{- else }}
               value: "no"
               {{- end }}
+              {{- if .Values.airflow.pool.enabled }}
+            - name: AIRFLOW_POOL
+              value: "true"
+            - name: AIRFLOW_POOL_NAME
+              value: {{ .Values.airflow.pool.name }}
+            - name: AIRFLOW_POOL_SIZE
+              value: "32"
```
alemorcuq: This shouldn't be hardcoded.

bitnami/airflow/templates/deployment-web.yaml
```diff
@@ -170,6 +174,16 @@ spec:
...
+              {{- if .Values.airflow.pool.enabled }}
+            - name: AIRFLOW_POOL
+              value: "true"
+            - name: AIRFLOW_POOL_NAME
+              value: {{ .Values.airflow.pool.name }}
+            - name: AIRFLOW_POOL_SIZE
+              value: "32" 
+            - name: AIRFLOW_POOL_DESCRIPTION
+              value: {{ .Values.airflow.pool.description }}
+              {{- end }}
```
alemorcuq: The Airflow container should know about this variables. Are you doing a PR there too? I didn't see anything in your message.

bitnami/airflow/templates/statefulset-worker.yaml
```diff
@@ -59,8 +60,29 @@ spec:
           volumeMounts:
             - name: git-cloned-dag-files
               mountPath: /dags
+            - name: dag-git-keys
+              mountPath: /root/.ssh
```
alemorcuq: Same comment.

bitnami/airflow/templates/statefulset-worker.yaml
```diff
@@ -200,23 +247,40 @@ spec:
...
+          configMap:
+            name: {{ .Values.airflow.dagsConfigMap }}
+        {{- end }}
+        {{- if .Values.airflow.cloneDagFilesFromGit.enabled }}
+        - name: git-cloned-dag-files
+          emptyDir: {}
+        - name: dag-git-keys
+          secret:
+            secretName: {{ .Values.airflow.cloneDagFilesFromGit.sshKeySecret }}
+            defaultMode: 0600
```
alemorcuq: Same comment

bitnami/airflow/templates/statefulset-worker.yaml
```diff
@@ -200,23 +247,40 @@ spec:
...
+          configMap:
+            name: {{ .Values.airflow.configurationConfigMap }}
+        {{- end }}
+        {{- range .Values.additionalRepos.repos }}
+        - name: git-cloned-{{ .name }}
+          emptyDir: {}
+        - name: {{ .name }}-git-keys  
+          secret:
+            secretName: {{ .sshKeySecret }}
+            defaultMode: 0600
```
alemorcuq: Same comment.

bitnami/airflow/templates/statefulset-worker.yaml
```diff
@@ -156,6 +201,8 @@ spec:
                   key: airflow-fernetKey
             - name: AIRFLOW_WEBSERVER_HOST
               value: {{ template "airflow.fullname" . }}
+            - name: AIRFLOW_QUEUE
+              value: "default"
```
alemorcuq: This value shouldn't be hardcoded.

bitnami/airflow/templates/statefulset-worker.yaml
```diff
@@ -74,6 +96,29 @@ spec:
...
+            - |
+              while true; do
+                  cd /{{ .name }} && git pull origin {{ .branch }}
+                  sleep {{ default "60" .interval }}
+              done
+          volumeMounts:
+            - name: git-cloned-{{ .name }}
+              mountPath: /{{ .name }}
+            - name: {{ .name }}-git-keys
+              mountPath: /root/.ssh
```
alemorcuq: Same comment.

bitnami/airflow/templates/statefulset-worker.yaml
```diff
@@ -74,6 +96,29 @@ spec:
           volumeMounts:
             - name: git-cloned-dag-files
               mountPath: /dags
+            - name: dag-git-keys
+              mountPath: /root/.ssh
```
alemorcuq: Same comment.

bitnami/airflow/templates/statefulset-worker.yaml
```diff
@@ -59,8 +60,29 @@ spec:
...
+          command:
+            - /bin/bash
+            - -ec
+            - |
+              git clone {{ .repository }} --branch {{ .branch }} /{{ .name }}
+          volumeMounts:
+            - name: git-cloned-{{ .name }}
+              mountPath: /{{ .name }}
+            - name: {{ .name }}-git-keys
+              mountPath: /root/.ssh
```
alemorcuq: Same comment.

bitnami/airflow/templates/statefulset-worker.yaml
```diff
@@ -59,8 +60,29 @@ spec:
           volumeMounts:
             - name: git-cloned-dag-files
               mountPath: /dags
+            - name: dag-git-keys
+              mountPath: /root/.ssh
+        {{- end }}
+        {{- if .Values.additionalRepos.enabled }}
+        {{- range .Values.additionalRepos.repos }}
```
alemorcuq: This is very inefficient. This is creating a new container for each repo we want to clone. We could clone all the repos using one single container instead.

bitnami/airflow/templates/statefulset-worker.yaml
```diff
@@ -74,6 +96,29 @@ spec:
...
```
alemorcuq: Same comment about multiple containers.

alemorcuq:
Hi, @cccs-cat001.

Thank you very much for this, this is an incredible contribution. Since it involves several repositories and makes a big addition, I'd like to proceed slowly with it.

@juan131, @tompizmor, @javsalgar could you take a look at this too?

Thank you again!

bitnami/airflow/templates/deployment-web.yaml
cccs-cat001: Sorry yes, I've got a pull request for the airflow image too - https://github.com/bitnami/bitnami-docker-airflow/pull/19

cccs-cat001:
> I don't think I understand very well the reasoning behind the additional StatefulSets for different configurations, could you explain it, maybe with some examples?

I made them stateful sets because I don't really know kubernetes that well. All I did was copy the `statefulset-worker.yaml` file, and modified it to work for a list of workers. If there is a more reasonable way to make custom workers without it being a stateful set I'm open to suggestions. 

> I also think that, instead of adding an array for extra repos and another one for extra workers, maybe we should just use an array in the first place.

So you mean define the default worker also in this `workers` list? 

> Changes in the `values.yaml` should be replicated in the `values-production.yaml`. If there are any different default settings, they should be documented in the README.

Sorry thought I did this before, guess I didn't save

> And last, you should increase the chart version.

I'm not sure what to increase to. 4.1.16 or 4.0.17?

bitnami/airflow/templates/statefulset-worker.yaml
Rkbhasan1: I am not sure.

bitnami/airflow/templates/statefulset-worker.yaml
Rkbhasan1: I don't know.

bitnami/airflow/templates/statefulset-worker.yaml
cccs-cat001: I changed it so it spawns only one pod, and then clones all the repos into that pod, and mounts a volume for each ssh key. 

In saying that, I realize that if all the repos require an ssh key, it will not work correctly. You can only use an ssh key for one repo. So we do need all the containers to be created - one for each repo.

bitnami/airflow/README.md
```diff
@@ -149,6 +155,26 @@ The following tables lists the configurable parameters of the Kafka chart and th
 | `metrics.image.pullPolicy`                | Image pull policy                                                                                    | `IfNotPresent`                                               |
 | `metrics.image.pullSecrets`               | Specify docker-registry secret names as an array                                                     | `[]` (does not add image pull secrets to deployed pods)      |
 | `metrics.podAnnotations`                  | Additional annotations for Metrics exporter                                                          | `{prometheus.io/scrape: "true", prometheus.io/port: "9112"}` |
+| `additionalRepos.enabled`                 | Enables the use of additional repos in the default Airflow Workers                                   | `false` 							  |
+| `additionalRepos.repos[o].name`	    | Descriptive name of a repo to sync			 					   | `nil`							  |
```
juan131: Typo: `additionalRepos.repos[o].name` should be `additionalRepos.repos[0].name`
You used `o` instead of `0`

bitnami/airflow/README.md
```diff
@@ -149,6 +155,26 @@ The following tables lists the configurable parameters of the Kafka chart and th
...
+| `workers[0].image.registry`    	    | Registry for this workers image									   | `nil`							  |
+| `workers[0].image.repository`		    | Name of this worker image										   | `nil`							  | 
+| `workers[0].image.tag`		    | Tag for this worker image										   | `nil`							  | 
+| `workers[0].image.pullPolicy` 	    | Pull policy for this worker image									   | `nil`							  | 
+| `workers[0].additionalRepos.enabled`	    | Enables the use of additional repos for this worker						   | `false`							  |
+| `workers[0].additionalRepos.repos[0].name` | Descriptive name of a repo to sync with this worker 	 					   | `nil`							  |
+| `workers[0].additionalRepos.repos[0].repository` | Link to the additional repository to sync with this worker					   | `nil` 							  |
+| `workers[0].additionalRepos.repos[0].branch` | Branch from repository to sync with this worker							   | `nil`							  | 
+| `workers[0].additionalRepos.repos[0].interval` | Interval to pull the repository for this worker							   | 60 							  |
+| `workers[0].additionalRepos.repos[0].sshKeySecret` | Kubernetes secret containing SSH keys for this repo						   | `nil`							  |
```
juan131: I would name this parameter `workers[0].additionalRepos.repos[0].sshKeysSecretName`

bitnami/airflow/values.yaml
```diff
@@ -218,7 +227,7 @@ ingress:
   ## If certManager is set to true, annotation kubernetes.io/tls-acme: "true" will automatically be set
   annotations:
   #  kubernetes.io/ingress.class: nginx
-
+
```
juan131: Remove this extra space

bitnami/airflow/values.yaml
```diff
@@ -365,9 +374,38 @@ metrics:
...
+    #  tag:
+    #  pullPolicy:
+    # additionalRepos:
+      # enabled: false
+      # repos:
+      # - name:
+      #   repository:
+      #   branch:
+      #   interval:
+      #   sshKeySecret:
```
juan131: What about the following approach:
```yaml
      ## SSH keys to clone the repo
      ## 
      # sshKeys:
      ## Name of existing secret containing the SSH keys to clone the repo
      ## When provided, `workers[0].additionalRepos.repos[0].sshKeys` will be ignored
      ##
      # sshKeysSecretName:
```
Then you can create a **ssh-keys-secrets.yaml** manifest like the one below:
```yaml
{{- range .Values.workers }}
{{- if .additionalRepos.enabled }}
{{- range .additionalRepos.repos }}
{{- if empty .sshKeysSecretName }}
apiVersion: v1
kind: Secret
metadata:
  name: {{ .name }}
  labels:
    app.kubernetes.io/name: {{ include "airflow.name" $ }}
    helm.sh/chart: {{ include "airflow.chart" $ }}
    app.kubernetes.io/instance: {{ $.Release.Name }}
    app.kubernetes.io/managed-by: {{ $.Release.Service }}
data: 
  ssh-keys: {{ .sshKeys | b64enc }}
---
{{- end }}
{{- end }}
{{- end }}
{{- end }}
```

bitnami/airflow/templates/statefulset-worker-list.yaml
```diff
@@ -0,0 +1,306 @@
+{{- range $.Values.workers }}
```
juan131: I don't think the `$` is necessary here

bitnami/airflow/values.yaml
cccs-cat001: I'm not sure I understand here. Are you saying we could include the sshkeys directly here, and then have it generate the secrets with those keys, as opposed to how I have it set up now where it's assumed that the ssh keys are already a secret, and you tell it which secret they're in?

bitnami/airflow/templates/statefulset-worker-list.yaml
cccs-cat001: I would agree

bitnami/airflow/values.yaml
juan131: Exactly. You support both taken the ssh keys from an already existing secret, and passing them in the values so you create the secret

bitnami/airflow/README.md
```diff
@@ -149,6 +155,21 @@ The following tables lists the configurable parameters of the Kafka chart and th
 | `metrics.image.pullPolicy`                | Image pull policy                                                                                    | `IfNotPresent`                                               |
 | `metrics.image.pullSecrets`               | Specify docker-registry secret names as an array                                                     | `[]` (does not add image pull secrets to deployed pods)      |
 | `metrics.podAnnotations`                  | Additional annotations for Metrics exporter                                                          | `{prometheus.io/scrape: "true", prometheus.io/port: "9112"}` |
+| `additionalRepos.enabled`                 | Enables the use of additional repos in the default Airflow Workers                                   | `false` 							  |
```
juan131: I think this parameter is not necessary anymore

bitnami/airflow/README.md
```diff
@@ -233,8 +254,42 @@ airflow.cloneDagFilesFromGit.enabled=true
...
+You can specify additional repositories to sync with your worker for additional files for your DAGs to use, or whatever pleases you. You can specify multiple. To do so: 
+
+```console
+additionalRepos.enabled=true
+additionalRepos.repos[0].name=additional-repo
+additionalRepos.repos[0].repository=https://github.com/USERNAME/REPOSITORY
+additionalRepos.repos[0].branch=master
+additioanlRepos.repos[0].interval=60
+additioanlRepos.repos[0].sshKeySecretName=KUBERNETES_SECRET
+```
```
juan131: This isn't true anymore now we're using a single list for workers, is it?

bitnami/airflow/README.md
juan131: The `workerImage` parameter is also deprecated now

bitnami/airflow/values.yaml
```diff
@@ -56,30 +56,6 @@ schedulerImage:
...
-  ## ref: https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/
-  ##
-  # pullSecrets:
-  #   - myRegistryKeySecretName
-
-  ## Set to true if you would like to see extra information on logs
-  ## It turns BASH and NAMI debugging in minideb
-  ## ref:  https://github.com/bitnami/minideb-extras/#turn-on-bash-debugging
-  debug: false
-
```
juan131: Update the **values-production.yaml**

bitnami/airflow/values.yaml
cccs-cat001: I think this assumes that you're definitely supplying ssh keys to every repo eh? I made an enabled option to turn it on in case you don't need ssh keys.

bitnami/airflow/templates/deployment-scheduler.yaml
```diff
@@ -58,6 +58,10 @@ spec:
           volumeMounts:
             - name: git-cloned-dag-files
               mountPath: /dags
+            {{- if .Values.airflow.cloneDagFilesFromGit.sshKeySecret }}
```
juan131: `sshKeySecret` -> `sshKeySecretName`

juan131: You need to apply these change several times on this manifest

bitnami/airflow/templates/deployment-web.yaml
```diff
@@ -58,6 +58,10 @@ spec:
           volumeMounts:
             - name: git-cloned-dag-files
               mountPath: /dags
+            {{- if .Values.airflow.cloneDagFilesFromGit.sshKeySecret }}
```
juan131: `sshKeySecret` -> `sshKeySecretName`

juan131: You need to apply these change several times on this manifest

bitnami/airflow/values-production.yaml
```diff
@@ -155,7 +129,7 @@ airflow:
     ## Option to force users to specify a password. That is required for 'helm upgrade' to work properly.
     ## If it is not forced, a random password will be generated.
     ##
-    forcePassword: true
+    forcePassword: false
```
juan131: Revert this changes in the production values

bitnami/airflow/values-production.yaml
```diff
@@ -205,7 +186,7 @@ service:
 ##
 ingress:
   ## Set to true to enable ingress record generation
-  enabled: true
+  enabled: false
```
juan131: Revert this changes in the production values

juan131:
Hi @cccs-cat001 

It seems that sth definitely went wrong when you updated your branch (as you can see there are changes related to other charts)

If I were you, I'd close this PR and create a new one including your latest changes, since I think it's going be much easier than fixing all the conflicts you have at the moment.

PR #1710, opened 2019-12-06
[bitnami/memcached] Durable memcache. Fixes #1685

### LINKED ISSUE #1685: Support memcached warm-restart
**Which chart**:

stable/memcached, latest version

**Description**

WDYT Adding support for https://github.com/memcached/memcached/wiki/WarmRestart ?
It could be an opt-in feature using a PVC.
This will make memcached upgrades less impacting, as the cache will survive restart.

1. implement feature
2. do upgrades / restart
3. verify that cached elements are still present in mc

State will be lost on upgrade

Keep state

N/A

N/A

- Output of `helm version`:
N/A

N/A

I've also forked the helm-project stable/mcrouter chart and I'm cleaning this up - let me know if you want it upstream.

davidkarlsen:
Signed-off-by: David J. M. Karlsen <david@davidkarlsen.com>

WorkInProgress - please provide feedback.

**Description of the change**
Provides HA by using PVC to store memory state on.

**Benefits**

Cache survives pod-recreation/upgrades.

**Possible drawbacks**

Added complexity and state - however the default is to run a single deployment as before depending on the `architecture` parameter.

**Applicable issues**

  - fixes #1685

**Additional information**

I'd like to add readonlyRootFilesystem too

**Checklist** 

---
This change is [Reviewable](https://reviewable.io/reviews/bitnami/charts/1710)

### DIFF
--- a/bitnami/memcached/templates/_helpers.tpl
+++ b/bitnami/memcached/templates/_helpers.tpl
@@ -129,6 +129,42 @@ imagePullSecrets:
 {{- end -}}
 {{- end -}}
 
+
+{{/*
+Return the proper Storage Class
+*/}}
+{{- define "memcached.storageClass" -}}
+{{/*
+Helm 2.11 supports the assignment of a value to a variable defined in a different scope,
+but Helm 2.9 and 2.10 does not support it, so we need to implement this if-else logic.
+*/}}
+{{- if .Values.global -}}
+    {{- if .Values.global.storageClass -}}
+        {{- if (eq "-" .Values.global.storageClass) -}}
+            {{- printf "storageClassName: \"\"" -}}
+        {{- else }}
+            {{- printf "storageClassName: %s" .Values.global.storageClass -}}
+        {{- end -}}
+    {{- else -}}
+        {{- if .Values.persistence.storageClass -}}
+              {{- if (eq "-" .Values.persistence.storageClass) -}}
+                  {{- printf "storageClassName: \"\"" -}}
+              {{- else }}
+                  {{- printf "storageClassName: %s" .Values.persistence.storageClass -}}
+              {{- end -}}
+        {{- end -}}
+    {{- end -}}
+{{- else -}}
+    {{- if .Values.persistence.storageClass -}}
+        {{- if (eq "-" .Values.persistence.storageClass) -}}
+            {{- printf "storageClassName: \"\"" -}}
+        {{- else }}
+            {{- printf "storageClassName: %s" .Values.persistence.storageClass -}}
+        {{- end -}}
+    {{- end -}}
+{{- end -}}
+{{- end -}}
+
 {{/*
 Renders a value that contains template.
 Usage:
--- a/bitnami/memcached/templates/_helpers.tpl
+++ b/bitnami/memcached/templates/_helpers.tpl
@@ -166,6 +166,42 @@ but Helm 2.9 and 2.10 does not support it, so we need to implement this if-else 
 {{- end -}}
 
 {{/*
+Compile all warnings into a single message, and call fail.
+*/}}
+{{- define "memcached.validateValues" -}}
+{{- $messages := list -}}
+{{- $messages := append $messages (include "memcached.validateValues.architecture" .) -}}
+{{- $messages := append $messages (include "memcached.validateValues.replicaCount" .) -}}
+{{- $messages := without $messages "" -}}
+{{- $message := join "\n" $messages -}}
+
+{{- if $message -}}
+{{-   printf "\nVALUES VALIDATION:\n%s" $message | fail -}}
+{{- end -}}
+{{- end -}}
+
+{{/* Validate values of Memcached - must provide a valid architecture */}}
+{{- define "memcached.validateValues.architecture" -}}
+{{- if and (ne .Values.architecture "standalone") (ne .Values.architecture "high-availability") -}}
+memcached: architecture
+    Invalid architecture selected. Valid values are "standalone" and
+    "high-availability". Please set a valid architecture (--set architecture="xxxx")
+{{- end -}}
+{{- end -}}
+
+{{/* Validate values of InfluxDB - number of replicas */}}
+{{- define "memcached.validateValues.replicaCount" -}}
+{{- $replicaCount := int .Values.replicaCount }}
+{{- if and (eq .Values.architecture "standalone") (gt $replicaCount 1) -}}
+memcached: replicaCount
+    The standalone architecture doesn't allow to run more than 1 replica.
+    Please set a valid number of replicas (--set memcached.replicaCount=1) or
+    use the "high-availability" architecture (--set architecture="high-availability")
+{{- end -}}
+{{- end -}}
+
+
+{{/*
 Renders a value that contains template.
 Usage:
 {{ include "memcached.tplValue" ( dict "value" .Values.path.to.the.Value "context" $) }}
--- a/bitnami/memcached/templates/_helpers.tpl
+++ b/bitnami/memcached/templates/_helpers.tpl
@@ -189,7 +189,7 @@ memcached: architecture
 {{- end -}}
 {{- end -}}
 
-{{/* Validate values of InfluxDB - number of replicas */}}
+{{/* Validate values of Memcached - number of replicas */}}
 {{- define "memcached.validateValues.replicaCount" -}}
 {{- $replicaCount := int .Values.replicaCount }}
 {{- if and (eq .Values.architecture "standalone") (gt $replicaCount 1) -}}
--- a/bitnami/memcached/templates/deployment.yaml
+++ b/bitnami/memcached/templates/deployment.yaml
@@ -40,7 +40,8 @@ spec:
         - name: memcached
           image: {{ template "memcached.image" . }}
           imagePullPolicy: {{ .Values.image.pullPolicy | quote }}
-          command: {{- toYaml .Values.command | nindent 12 }}
+          args:
+          {{ toYaml .Values.arguments | nindent 12 }}
           env:
             - name: MEMCACHED_USERNAME
               value: {{ default "" .Values.memcachedUsername | quote }}
--- a/bitnami/memcached/templates/deployment.yaml
+++ b/bitnami/memcached/templates/deployment.yaml
@@ -68,6 +68,9 @@ spec:
           {{- if .Values.resources }}
           resources: {{- toYaml .Values.resources | nindent 12 }}
           {{- end }}
+          volumeMounts:
+            - name: tmp
+              mountPath: /tmp
           {{- if .Values.securityContext.enabled }}
           securityContext:
             readOnlyRootFilesystem: {{ .Values.securityContext.readOnlyRootFilesystem }}
@@ -95,4 +98,7 @@ spec:
           resources: {{- toYaml .Values.metrics.resources | nindent 12 }}
           {{- end }}
         {{- end }}
+      volumes:
+        - name: tmp
+          emptyDir: {}
 {{- end }}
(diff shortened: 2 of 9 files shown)


bitnami/memcached/templates/deployment.yaml
```diff
@@ -38,6 +40,8 @@ spec:
         - name: memcached
           image: {{ template "memcached.image" . }}
           imagePullPolicy: {{ .Values.image.pullPolicy | quote }}
+          command:
+{{ toYaml .Values.command | indent 12 }}
```
juan131: Suggestion:
```yaml
          command: {{- toYaml .Values.command | nindent 12 }}
```

bitnami/memcached/templates/statefulset.yaml
```diff
@@ -0,0 +1,119 @@
...
+apiVersion: apps/v1
+kind: StatefulSet
+metadata:
+  name: {{ template "memcached.fullname" . }}
+  labels: {{- include "memcached.labels" . | nindent 4 }}
+spec:
+  selector:
+    matchLabels: {{- include "memcached.matchLabels" . | nindent 6 }}
+  replicas: {{ .Values.replicaCount }}
+  serviceName: {{ template "memcached.fullname" . }}-headless
```
juan131: Am I missing the manifest for this "headless" service?

bitnami/memcached/templates/statefulset.yaml
```diff
@@ -0,0 +1,119 @@
...
+      securityContext:
+        fsGroup: {{ .Values.securityContext.fsGroup }}
+        runAsUser: {{ .Values.securityContext.runAsUser }}
+      {{- end }}
+      containers:
+        - name: memcached
+          image: {{ template "memcached.image" . }}
+          imagePullPolicy: {{ .Values.image.pullPolicy | quote }}
+          command:
+{{ toYaml .Values.command | indent 12 }}
```
juan131: Same suggestion about using a single line:
```yaml
          command: {{- toYaml .Values.command | nindent 12 }}
```

bitnami/memcached/values-production.yaml
```diff
@@ -112,6 +120,22 @@ nodeSelector: {}
 ##
 tolerations: {}
 
+## Persistence - used for dumping and restoring states between recreations
+## Ref: https://github.com/memcached/memcached/wiki/WarmRestart
+persistence:
+  enabled: false
+  storageClassName: ""
+  # requests:
+  #   storage: 5Gi
```
juan131: Suggestion (so it's consistent with other charts):
```yaml
persistence:
  enabled: true
  ## Persistent Volume Storage Class
  ## If defined, storageClassName: <storageClass>
  ## If set to "-", storageClassName: "", which disables dynamic provisioning
  ## If undefined (the default) or set to null, no storageClassName spec is
  ##  set, choosing the default provisioner.  (gp2 on AWS, standard on
  ##  GKE, AWS & OpenStack)
  ##
  # storageClass: "-"
  ## Persistent Volume Claim annotations
  ##
  annotations: {}
  ## Persistent Volume Access Mode
  ##
  accessModes:
    - ReadWriteOnce
  ## Persistent Volume size
  ##
  size: 8Gi
```
Then, you'll need to adapt the `volumeClaimTemplates` to use these parameters.

bitnami/memcached/values-production.yaml
```diff
@@ -112,6 +120,22 @@ nodeSelector: {}
...
+
+## Args for running memcached
+## Ref: https://github.com/memcached/memcached/wiki/ConfiguringServer
+command:
+  - memcached
+  # - -m <maxMemoryLimit>
+  # - -I <maxItemSize>
+  # - -vv
```
juan131: We're using some scripts (`entypoint.sh` and `run.sh`) to start Memcached on this image. See:

- https://github.com/bitnami/bitnami-docker-memcached/blob/master/1/debian-9/Dockerfile#L25
- https://github.com/bitnami/bitnami-docker-memcached/blob/master/1/debian-9/rootfs/entrypoint.sh
- https://github.com/bitnami/bitnami-docker-memcached/blob/master/1/debian-9/rootfs/run.sh

Basically, the `entypoint.sh` is designed to do some setup (see https://github.com/bitnami/bitnami-docker-memcached/blob/master/1/debian-9/rootfs/setup.sh) when the command is "run.sh". This setup is based on some functions defined at  https://github.com/bitnami/bitnami-docker-memcached/blob/master/1/debian-9/rootfs/libmemcached.sh

We should maintain using these scripts since they provide us a lot of flexibility to configure Memcached.

Finally, if you check the line below, you'll notice that we're already appending every extra parameter to the list of parameters used to start memcached:

https://github.com/bitnami/bitnami-docker-memcached/blob/master/1/debian-9/rootfs/run.sh#L26

bitnami/memcached/templates/statefulset.yaml
davidkarlsen: I missed the whole command vs args thing. I think we're back to where it should be and properly using entrypoint (e.g. default command) now. Might be more to tweak.

bitnami/memcached/values-production.yaml
davidkarlsen: Had a look at kafka/jenkins charts and did the same here.

bitnami/memcached/templates/statefulset.yaml
davidkarlsen: helm 3 failed without it. Not sure what it is about

bitnami/memcached/values-production.yaml
davidkarlsen: The approach is a bit convoluted - but let's stick to it for consistency. The current chart has no way of adding env-vars, for instance to configure MEMCACHED_CACHE_SIZE. Also the env-vars are not documented - or am I missing something here?

bitnami/memcached/templates/statefulset.yaml
davidkarlsen: Added headless svc when doing ha / statefulset

bitnami/memcached/values.yaml
```diff
@@ -26,17 +26,28 @@ image:
...
 ## String to fully override memcached.fullname template
 ##
 # fullnameOverride:
 
 # Cluster domain
 clusterDomain: cluster.local
 
+## Memcahed architecture. Allowed values: standalone or high-availability
+##
+architecture: standalone
```
juan131: If you check the helpers of InfluxDB (lines 280 to 302): https://github.com/bitnami/charts/blob/master/bitnami/influxdb/templates/_helpers.tpl
We created a "validateValues" macro that can include some validations. For instance, we create a validation (`validateValues.architecture`) to ensure that only valid values are set on that parameter.

Could we do sth similar here?  

This macro is used in the NOTES.txt: https://github.com/bitnami/charts/blob/master/bitnami/influxdb/templates/NOTES.txt#L140

bitnami/memcached/templates/statefulset.yaml
```diff
@@ -0,0 +1,138 @@
...
+apiVersion: apps/v1
+kind: StatefulSet
+metadata:
+  name: {{ template "memcached.fullname" . }}
+  labels: {{- include "memcached.labels" . | nindent 4 }}
+spec:
+  selector:
+    matchLabels: {{- include "memcached.matchLabels" . | nindent 6 }}
+  replicas: {{ .Values.replicaCount }}
+  serviceName: {{ template "memcached.fullname" . }}
```
juan131: The current service is not a headless service, right?

davidkarlsen: It is if statefulset

bitnami/memcached/templates/_helpers.tpl
```diff
@@ -129,6 +129,78 @@ imagePullSecrets:
...
+{{/* Validate values of Memcached - must provide a valid architecture */}}
+{{- define "memcached.validateValues.architecture" -}}
+{{- if and (ne .Values.architecture "standalone") (ne .Values.architecture "high-availability") -}}
+memcached: architecture
+    Invalid architecture selected. Valid values are "standalone" and
+    "high-availability". Please set a valid architecture (--set architecture="xxxx")
+{{- end -}}
+{{- end -}}
+
+{{/* Validate values of InfluxDB - number of replicas */}}
```
juan131: Typo: `InfluxDB` -> `Memcached`

bitnami/memcached/templates/service.yaml
```diff
@@ -8,6 +8,9 @@ metadata:
   {{- end }}
 spec:
   type: {{ .Values.service.type }}
+  {{- if eq .Values.architecture "high-availability" }}
+  clusterIP: None
+  {{- end }}
```
juan131: How can I access Memcached when using "high-availability" architecture? 
We're using the service as a headless svc when using a statefulset, but there's no other alternative service. So I cannot configure my app to use it and I my requests are not balanced between the replicas.

davidkarlsen: When running several memcache-servers you'd make them all known to the client, which uses consistent-hashing to route directly to the right instance: https://www.mikeperham.com/2009/01/14/consistent-hashing-in-memcache-client/

OR you stick mcrouter in front: https://github.com/facebook/mcrouter/wiki/Replicated-pools-setup in front, to let it deal with the hashing and/or replication/routing.

juan131: It makes sense. Should we document this in the README.md? 
We should also adapt the NOTES.txt since it says:

```
Memcached can be accessed on port 11211 on the following DNS name from within your cluster: {{ template "memcached.fullname" . }}.{{ .Release.Namespace }}.svc.{{ .Values.clusterDomain }}
```

That won't be true when using "high-availability" architecture

davidkarlsen: PTAL

bitnami/memcached/templates/statefulset.yaml
```diff
@@ -0,0 +1,138 @@
...
+          {{- end }}
+          env:
+            - name: MEMCACHED_USERNAME
+              value: {{ default "" .Values.memcachedUsername | quote }}
+            - name: MEMCACHED_PASSWORD
+              valueFrom:
+                secretKeyRef:
+                  name: {{ template "memcached.fullname" . }}
+                  key: memcached-password
+            {{- toYaml .Values.extraEnv | nindent 12 }}
```
juan131: Could you apply this please?
```diff
+            {{- if .Values.extraEnv }}
             {{- toYaml .Values.extraEnv | nindent 12 }}
+            {{- end }}
```
Same reason as above

bitnami/memcached/templates/deployment.yaml
```diff
@@ -46,6 +49,7 @@ spec:
                 secretKeyRef:
                   name: {{ template "memcached.fullname" . }}
                   key: memcached-password
+            {{- toYaml .Values.extraEnv | nindent 12 }}
```
juan131: Could you apply this please?
```diff
+            {{- if .Values.extraEnv }}
             {{- toYaml .Values.extraEnv | nindent 12 }}
+            {{- end }}
```
The reason is that we are linting our YAMLs. If you run the command below, you'll find an error:

```console
$ helm template --values bitnami/memcached/values.yaml bitnami/memcached -x templates/deployment.yaml | yamllint -s -d "{extends: default, rules: {line-length: disable, trailing-spaces: disable, truthy: enable, document-start: disable, empty-lines: {max-end: 2} }}" -

  47:11     error    syntax error: could not find expected ':'
```

Why? Because the resulting YAML has the following aspect:

```yaml
...       
          env:
            - name: MEMCACHED_USERNAME
              value: ""
            - name: MEMCACHED_PASSWORD
              valueFrom:
                secretKeyRef:
                  name: release-name-memcached
                  key: memcached-password
            []
...
```

Despite it's a valid YAML to be run on K8s, we're trying to avoid this "ugly" YAMLs

juan131:
Again, thanks so much for this. It's awesome

PR #1720, opened 2019-12-10
[bitnami/harbor] Fix wrong values in helpers and configmap related to the external database

carrodher:
**Description of the change**

Fix wrong values in helpers and configmap related to the external database

**Checklist** 


bitnami/harbor/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: harbor
-version: 3.0.7
+version: 3.0.8
```
sameersbn: should we bump the minor instead?

carrodher: I usually follow this "approach":
> - MINOR version when you add functionality in a backwards-compatible manner

Why do you think it should be a minor?

sameersbn: oh sorry. I felt like the values.yaml was moved around. That is not the case. You can go ahead with the merge

PR #1721, opened 2019-12-10
[bitnami/kibana] Remove ES as a dependency

carrodher:
Reviewed and accepted in https://github.com/bitnami/charts/pull/1567

**Description of the change**

Remove Elasticsearch as dependency bundled in Kibana. From now on, Kibana will be included as a dependency in the ES chart.

**Additional information**

⚠️ Please, review it but not merge the PR because we need to review what happens with marketplaces where `helm install` should work by default. In this case, as it is shown in the README, you need to use the already running ES host and port.

**Checklist** 

### DIFF
--- a/bitnami/kibana/Chart.yaml
+++ b/bitnami/kibana/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: kibana
-version: 4.0.0
+version: 5.0.0
 appVersion: 7.5.0
 description: Kibana is an open source, browser based analytics and search dashboard for Elasticsearch.
 keywords:
--- a/bitnami/kibana/README.md
+++ b/bitnami/kibana/README.md
@@ -6,12 +6,7 @@
 
 ```console
 $ helm repo add bitnami https://charts.bitnami.com/bitnami
-
-# Option 1: With an existing Elasticsearch instance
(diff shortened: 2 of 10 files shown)


carrodher:
Answering the note I added in the description, just for the records:
> ⚠️ Please, review it but not merge the PR because we need to review what happens with marketplaces where helm install should work by default. In this case, as it is shown in the README, you need to use the already running ES host and port.

Now, Kibana doesn’t contain a bundled ES so users need to provide it via parameters.
The `helm install` command doesn’t fail but there are no pods running. That is because the deployment is not created if there is not a valid ES instance.
A warning in the _notes.txt_ was added:
```
######################################################################################################
### ERROR: You did not provide the Elasticsearch external host or port in your 'helm install' call ###
######################################################################################################

Complete your Kibana deployment by running:

  helm upgrade {{ .Release.Name }} bitnami/kibana \
    --set elasticsearch.hosts[0]=YOUR_ES_HOST,elasticsearch.port=YOUR_ES_PORT

Replacing "YOUR_ES_HOST" and "YOUR_ES_PORT" placeholders by the proper values of your Elasticsearch deployment.
```
Following those instructions, users can update the existing installation providing host and port, after that, the deployment and pods are created.

PR #1728, opened 2019-12-11
[bitnami/etcd] Use debug env parameter in bash scripts

sboschman:
According to https://github.com/bitnami/charts/blob/master/bitnami/etcd/values.yaml#L33 this chart value should also enable bash debugging.

But this chart value sets BITNAMI_DEBUG, https://github.com/bitnami/charts/blob/master/bitnami/etcd/templates/statefulset.yaml#L97

The bash scripts on the other check BASH_DEBUG, https://github.com/bitnami/charts/blob/master/bitnami/etcd/templates/scripts-configmap.yaml#L30

This PR adds another env parameter to toggle bash script debugging based on the chart debug value. An other possibility would be to change the scripts and have them check the BITNAMI_DEBUG env parameter as well, so we only have to set 1 env parameter. Any thoughts?

### DIFF
--- a/bitnami/etcd/Chart.yaml
+++ b/bitnami/etcd/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: etcd
-version: 4.4.5
+version: 4.4.6
 appVersion: 3.4.3
 description: etcd is a distributed key value store that provides a reliable way to store data across a cluster of machines
 keywords:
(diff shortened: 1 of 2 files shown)


alemorcuq:
Thanks for the PR, @sboschman. I would change the scripts to check `BITNAMI_DEBUG` instead, since that's the variable that we are using in other infrastructure containers.

alemorcuq:
Thanks!

PR #1736, opened 2019-12-13
[bitnami/elasticsearch] Fix nameOverride issue

carrodher:
**Description of the change**
The source of the issue is that we need to use the same template for the main chart and the subchart. This template is using:
```
{{- define "elasticsearch.coordinating.fullname" -}}
{{- $name := default .Chart.Name .Values.nameOverride -}}
{{- printf "%s-%s-%s" .Release.Name $name .Values.global.coordinating.name | trunc 63 | trimSuffix "-" -}}
{{- end -}}
```
the problem is that when the template is evaluated in the main chart `.Chart.Name` is `elasticsearch` while when it is evaluated in the subchart, `.Chart.Name` is `kibana`.

Because of that, the coordinating service name is not the same in both charts. To avoid this issue I used the `nameOverride` option in the kibana subchart:
```yaml
## Enable bundled Kibana
##
kibana:
  ## String to partially override elasticsearch.fullname template in the kibana subchart
  ##
  nameOverride: elasticsearch
```

Using it, the service is the same in both charts, because in the ES chart is using `.Chart.Name` (`elasticsearch`) and in the Kibana subchart is using `nameOverride` set to the same value.

The problem now is that `nameOverride` set to `elasticsearch` in the Kibana subchart implies that the Kibana resources are renamed to `elasticsearch`, for example:
```
▶ kubectl get pods --namespace assets-privileged
NAME                                                             READY   STATUS    RESTARTS   AGE
elasticsearch-5cd6db77c6-2rgbx                                   1/1     Running   0          2m36s             <-- This is the Kibana pod
elasticsearch-elasticsearch-coordinating-only-55c49b6f8f-tk7hr   1/1     Running   0          2m36s
elasticsearch-elasticsearch-data-0                               1/1     Running   0          2m36s
elasticsearch-elasticsearch-master-0                             1/1     Running   0          2m36s

▶ kubectl get svc --namespace assets-privileged
NAME                                            TYPE        CLUSTER-IP      EXTERNAL-IP   PORT(S)     AGE
elasticsearch                                   ClusterIP   10.30.250.191   <none>        80/TCP      5m49s     <-- This is the Kibana service
elasticsearch-elasticsearch-coordinating-only   ClusterIP   10.30.241.253   <none>        9200/TCP    5m49s
elasticsearch-elasticsearch-discovery           ClusterIP   None            <none>        9300/TCP    5m49s
elasticsearch-elasticsearch-master              ClusterIP   10.30.252.252   <none>        9300/TCP    5m49s
```

This PR avoid using `nameOverride` in Kibana, so Kibana resources maintain the same naming as previously. The change is in the coordinating service name if we know that Kibana is enabled the coordinating service name doesn't use 
```
{{- $name := default .Chart.Name .Values.nameOverride -}}
```
so the service name is not following the rest of the services approach (only when Kibana is enabled):
```
kubectl get pods --namespace assets-privileged
NAME                                       READY   STATUS    RESTARTS   AGE
petete-coordinating-only-ff9b57694-9mx2p   1/1     Running   0          29m
petete-coordinating-only-ff9b57694-v2w4w   1/1     Running   0          29m
petete-elasticsearch-data-0                1/1     Running   0          29m
petete-elasticsearch-data-1                1/1     Running   0          28m
petete-elasticsearch-master-0              1/1     Running   0          29m
petete-elasticsearch-master-1              1/1     Running   0          28m
petete-kibana-5d4d9b7789-k6c46             1/1     Running   0          29m
tiller-deploy-75ccc659f8-pvhmt             1/1     Running   0          2d21h
```

**Checklist** 

### DIFF
--- a/bitnami/elasticsearch/Chart.yaml
+++ b/bitnami/elasticsearch/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: elasticsearch
-version: 10.0.0
+version: 10.0.1
 appVersion: 7.5.0
 description: A highly scalable open-source full-text search and analytics engine
 keywords:
--- a/bitnami/elasticsearch/Chart.yaml
+++ b/bitnami/elasticsearch/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: elasticsearch
-version: 10.0.1
+version: 10.1.0
 appVersion: 7.5.0
 description: A highly scalable open-source full-text search and analytics engine
 keywords:
--- a/bitnami/elasticsearch/templates/_helpers.tpl
+++ b/bitnami/elasticsearch/templates/_helpers.tpl
@@ -126,13 +126,8 @@ Create a default fully qualified coordinating name.
 We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
 */}}
 {{- define "elasticsearch.coordinating.fullname" -}}
-{{/* If Kibana is enabled. Evaluated in the Kibana subchart */}}
-{{- if .Values.enabled -}}
+{{- if .Values.global.kibanaEnabled -}}
 {{- printf "%s-%s" .Release.Name .Values.global.coordinating.name | trunc 63 | trimSuffix "-" -}}
-{{/* If Kibana is enabled. Evaluated in the ES chart */}}
-{{- else if .Values.kibana.enabled -}}
-{{- printf "%s-%s" .Release.Name .Values.global.coordinating.name | trunc 63 | trimSuffix "-" -}}
-{{/* If Kibana is not enabled */}}
 {{- else -}}
 {{- $name := default .Chart.Name .Values.nameOverride -}}
 {{- printf "%s-%s-%s" .Release.Name $name .Values.global.coordinating.name | trunc 63 | trimSuffix "-" -}}
--- a/bitnami/elasticsearch/templates/_helpers.tpl
+++ b/bitnami/elasticsearch/templates/_helpers.tpl
@@ -126,8 +126,17 @@ Create a default fully qualified coordinating name.
 We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
 */}}
 {{- define "elasticsearch.coordinating.fullname" -}}
+{{/* If Kibana is enabled. Evaluated in the Kibana subchart */}}
+{{- if .Values.enabled -}}
+{{- printf "%s-%s" .Release.Name .Values.global.coordinating.name | trunc 63 | trimSuffix "-" -}}
+{{/* If Kibana is enabled. Evaluated in the ES chart */}}
+{{- else if .Values.kibana.enabled -}}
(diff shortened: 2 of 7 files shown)


bitnami/elasticsearch/templates/_helpers.tpl
```diff
@@ -126,9 +126,18 @@ Create a default fully qualified coordinating name.
 We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
 */}}
 {{- define "elasticsearch.coordinating.fullname" -}}
+{{/* If Kibana is enabled. Evaluated in the Kibana subchart */}}
+{{- if .Values.enabled -}}
```
sameersbn: `.Values.enabled`?

juan131: What is this parameter?

bitnami/elasticsearch/values.yaml
```diff
@@ -742,6 +742,3 @@ kibana:
     hosts:
       - '{{ include "elasticsearch.coordinating.fullname" . }}'
     port: 9200
-  ## String to partially override elasticsearch.fullname template in the kibana subchart
-  ##
-  nameOverride: elasticsearch
```
juan131: I guess you're just removing the default value but you can use this parameter, right?

carrodher: Yes, users can use it, but we don't need to set a default value

bitnami/elasticsearch/templates/_helpers.tpl
carrodher: In the ES `values.yaml` we have a section with parameters affecting to the subchart:
```yaml
## Enable bundled Kibana
##
kibana:
  enabled: false
  elasticsearch:
    hosts:
      - '{{ include "elasticsearch.coordinating.fullname" . }}'
    port: 9200
```
Those parameters are available in the ES chart (main chart) as `.Values.kibana.XXX`, in the same way, those paremeters are used in the Kibana chart (subchart) as `.Values.XXX`.

In this section of the `helpers.tpl` I need to use both ways to determine if Kibana chart is used as part of an ES deployment. So the first `if` is evaluated in the subchart while the second `else if` is evaluated in the main chart; in both cases, the service name that returns the function is the same, that is the final goal

carrodher: The same I replied in Juan's comment, it is because this function is evaluated in two charts, while in ES is `.Values.kibana.enabled` in the Kibana chart is `.Values.enabled`

juan131: I think this is very hacky. Why not using `global` to ensure both the chart and the subchart can access to these "global parameters"?

bitnami/elasticsearch/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: elasticsearch
-version: 10.0.0
+version: 10.0.1
```
juan131: Maybe minor version? A parameter is changing

Issue #1509, opened 2019-10-22
Elasticsearch never recovers after cluster restart. Single restart fails cluster forever


virtuman:
**Which chart**: elasticsearch
all versions

**Description**
Cluster never successfully recovers after restart (data persistence enabled) with message 
`Caused by: org.elasticsearch.cluster.coordination.CoordinationStateRejectedException: join validation on cluster state with a different cluster uuid E-nlXOHsSHa_Z8d1S3jDMA than local cluster uuid 4dUYLGC0RvysrkX0-nfqIQ, rejecting`

**Steps to reproduce the issue:**

1. Install helm chart bitnami/elasticsearch with values-production.yml 
2. delete all pods: kubectl delete pods --all
3. notice that some pods recover successfully, but data nodes fail to register with the above mentioned error message

**Describe the results you received:**

`kubectl logs -f elastic-elasticsearch-data-0`

```
[elastic-elasticsearch-data-0 elasticsearch] [2019-10-22T10:46:30,766][INFO ][o.e.c.c.JoinHelper       ] [elastic-elasticsearch-data-0] failed to join {elastic-elasticsearch-master-0}{cCKVF8XTTDCYX0m9FZ14pw}{XYhsq7QeSOyBDA688K8B4g}{10.42.5.237}{10.42.5.237:9300}{m} with JoinRequest{sourceNode={elastic-elasticsearch-data-0}{O1dbIj9JTBKQolnFhAIXAg}{EMV2G2adT7W1EyGVc8H94A}{10.42.5.139}{10.42.5.139:9300}{d}, optionalJoin=Optional.empty}
[elastic-elasticsearch-data-0 elasticsearch] org.elasticsearch.transport.RemoteTransportException: [elastic-elasticsearch-master-0][10.42.5.237:9300][internal:cluster/coordination/join]
[elastic-elasticsearch-data-0 elasticsearch] Caused by: java.lang.IllegalStateException: failure when sending a validation request to node
[elastic-elasticsearch-data-0 elasticsearch] 	at org.elasticsearch.cluster.coordination.Coordinator$2.onFailure(Coordinator.java:513) ~[elasticsearch-7.4.0.jar:7.4.0]
[elastic-elasticsearch-data-0 elasticsearch] 	at org.elasticsearch.action.ActionListenerResponseHandler.handleException(ActionListenerResponseHandler.java:59) ~[elasticsearch-7.4.0.jar:7.4.0]
[...]

PR #1737, opened 2019-12-13
Elasticsearch master nodes persistency

tzachalon:
This is a small fix regarding the inability to attach a persistence storage for elasticsearch master nodes, as described on: https://github.com/bitnami/charts/issues/1509

Adding double-quotes to the volumeMount's name fixed the issue on my kubernetes cluster and now I'm able to take the entire elasticsearch cluster down and up without any issues with differences in cluster uuid.

This is my first contribution on Github, let me know if anything else is needed.

### DIFF
--- a/bitnami/elasticsearch/templates/master-statefulset.yaml
+++ b/bitnami/elasticsearch/templates/master-statefulset.yaml
@@ -63,7 +63,7 @@ spec:
           resources: {{- toYaml .Values.volumePermissions.resources | nindent 12 }}
           {{- end }}
           volumeMounts:
-            - name: data
+            - name: "data"
               mountPath: "/bitnami/elasticsearch/data"
         {{- end }}
       {{- end }}
--- a/bitnami/elasticsearch/templates/master-statefulset.yaml
+++ b/bitnami/elasticsearch/templates/master-statefulset.yaml
@@ -63,7 +63,7 @@ spec:
           resources: {{- toYaml .Values.volumePermissions.resources | nindent 12 }}
           {{- end }}
           volumeMounts:
-            - name: "data"
+            - name: data
               mountPath: "/bitnami/elasticsearch/data"
         {{- end }}
       {{- end }}


carrodher:
mmm I am not sure if this fixed the issue because other parallel reason, but I was taking a look at the rest of Bitnami charts and in all of them we're using
```yaml
- name: whatever
```
without double quotes, I can't understand why it makes a difference in this case.

Do you have any idea about that @juan131 @sameersbn?

sameersbn:
@carrodher I don't think is should make a difference, but if it does then it's probably a kubernetes issue. From #1509 I see that nobody has been able to reproduce this issue.

@tzachalon can you please share some information about you k8s cluster? are you using gke, version, etc?

tzachalon:
> @carrodher I don't think is should make a difference, but if it does then it's probably a kubernetes issue. From #1509 I see that nobody has been able to reproduce this issue.
> 
> @tzachalon can you please share some information about you k8s cluster? are you using gke, version, etc?

Sure.
I'm using GKE. tried the installation on master version 1.12.10-gke.17 and also 1.13.11-gke.14. All I'm trying to say is that if on some k8s types we do see some issues when the quotes are not aligned, why not adding it anyway?

sameersbn:
@tzachalon of course there isn't any issue adding the quotes. I we are trying to ascertain is where the quotes actually made a difference. Cause if it did then we may have to do the same for every single chart published by us.  

I will follow you report and see if I am able to reproduce this issue.

tzachalon:
@sameersbn any update? Can I assist in any way?

sameersbn:
@tzachalon Please allow me some time to report back

tzachalon:
@sameersbn Sure thing. Thanks a lot!

javsalgar:
Hi,

Just like my colleagues, I think this is a very weird issue. Could you try checking the kubectl describe pod to see the issue mounting the volume?

carrodher:
Closing this PR due to lack of activity

PR #1749, opened 2019-12-18
[bitnami/postgresql-ha] Update NOTES.txt to correct postgresql client version

rene-demonsters:
**Description of the change**

Update NOTES.txt so that the "Connect to your database command" uses the postgresql client version that matches the version of the database (`11.x`) being used instead of `10.x`

**Benefits**

Postgresql client matches the server version

**Possible drawbacks**

None

**Checklist** 


alemorcuq:
Thanks for the PR!

PR #1754, opened 2019-12-19
[bitnami/airflow] Updated airflow to support multiple workers

cccs-cat001:
All information for this PR can be found [here](https://github.com/bitnami/charts/pull/1703). Merging master messed up so I created a fresh PR that works.


cccs-cat001:
@juan131 @alemorcuq

rafariossaa:
@cccs-cat001 ,
It seems that there is some leftovers from a conflicting rebase (<<<< HEAD). 
Could you check that ?

bitnami/airflow/templates/worker-statefulsets.yaml
```diff
@@ -0,0 +1,306 @@
...
+              mountPath: /{{ .name }}
+            {{- if .sshKeySecretName }}
+            - name: {{ .name }}-git-keys
+              mountPath: /root/.ssh
+            {{- end }}
+        {{- end }}
+        {{- end }}
+      containers:
+        {{- if $.Values.airflow.cloneDagFilesFromGit.enabled }}
+        - name: dag-repo-syncer
```
javsalgar: Now that I see this, and taking into account that it is a major change. I would add an option for not allowing the syncing of the repository. Something like `syncFilesFromGit`

bitnami/airflow/templates/metrics-deployment.yaml
```diff
@@ -22,8 +22,7 @@ spec:
         app.kubernetes.io/name: {{ include "airflow.name" . }}
         app.kubernetes.io/component: metrics
         app.kubernetes.io/instance: {{ .Release.Name }}
-    spec:
-    {{- include "airflow.imagePullSecrets" . | nindent 6 }}
+    spec: {{- include "airflow.imagePullSecrets" . | nindent 6 }}
```
javsalgar: Was the macro changed to cover the multiple worker use case?

javsalgar:
Thanks! I left some comments

bitnami/airflow/templates/metrics-deployment.yaml
cccs-cat001: I'm not sure what this macro does or if it needs to be changed?

bitnami/airflow/templates/worker-statefulsets.yaml
javsalgar: Not done yet

bitnami/airflow/templates/worker-statefulsets.yaml
```diff
@@ -0,0 +1,309 @@
...
+        {{- end }}
+        {{- else }}
+      containers:
+      {{- end }}
+      - name: {{ .name | camelcase | kebabcase }}
+        {{- if .image }}
+        image: "{{ .image.registry }}/{{ .image.repository }}:{{ .image.tag }}"
+        imagePullPolicy: {{ .image.pullPolicy default "IfNotPresent" }} 
+        {{- else }}
+        image: "docker.io/bitnami/airflow-worker:1.10.6-debian-9-r1"
```
javsalgar: We have some chart update automations in place that may not work correctly with this approach. How important is it that each worker has a different image? Maybe we can have that the different workers have the same image and only the configuration is the different thing. That would make the chart more maintainable and potentially more usable. Does it make sense?

bitnami/airflow/templates/worker-statefulsets.yaml
```diff
@@ -0,0 +1,309 @@
...
+            valueFrom:
+              secretKeyRef:
+              {{- if $.Values.redis.enabled }}
+                name: {{ template "airflow.redis.secretName" $ }}
+                key: redis-password
+              {{- else }}
+                name: {{ printf "%s-%s" .Release.Name "externalredis" }}
+                key: redis-password
+              {{- end }}
+          - name: AIRFLOW_EXECUTOR
```
javsalgar: I presume that this value should be configurable, right? Different type of workers having different executors

bitnami/airflow/templates/worker-statefulsets.yaml
```diff
@@ -0,0 +1,309 @@
...
+          - name: AIRFLOW_FERNET_KEY
+            valueFrom:
+              secretKeyRef:
+                name: {{ if $.Values.airflow.auth.existingSecret }}{{ $.Values.airflow.auth.existingSecret }}{{ else }}{{ template "airflow.fullname" $ }}{{ end }}
+                key: airflow-fernetKey
+          - name: AIRFLOW_WEBSERVER_HOST
+            value: {{ template "airflow.fullname" $ }}
+          - name: AIRFLOW_QUEUE
+            value: "{{ .queue | default "default" }}"
+          {{- if $.Values.airflow.extraEnvVars }}
```
javsalgar: This should be configurable as well, right?

bitnami/airflow/templates/worker-statefulsets.yaml
```diff
@@ -0,0 +1,309 @@
...
+            {{ toYaml $.Values.airflow.extraEnvVars | nindent 8 }}
+          {{- end }}
+          {{- if .environmentVars }}
```
javsalgar: I would call it extraEnvVars so it has the same name as the "common" one

bitnami/airflow/templates/worker-statefulsets.yaml
```diff
@@ -0,0 +1,309 @@
...
+          {{- end }}
+          {{- if $.Values.airflow.dagsConfigMap }}
+          - name: external-dag-files
+            mountPath: /opt/bitnami/airflow/dags/external
+          {{- end }}
+          {{- if $.Values.airflow.cloneDagFilesFromGit.enabled }}
+          - name: git-cloned-dag-files
+            mountPath: /opt/bitnami/airflow/dags/git
+          {{- end }}
+          {{- if $.Values.airflow.configurationConfigMap }}
```
javsalgar: Does it make sense for all type of workers to have the same cfg file?

javsalgar:
Also please fix the merge issues

bitnami/airflow/templates/worker-statefulsets.yaml
cccs-cat001: The entire idea of my pull request was to allow the user to provide fully customized workers. So you could specify your own image, own configurations, everything. This is because the default bitnami images are very basic and provide no additional libraries. For instance, I want to be able to use [papermill](https://papermill.readthedocs.io/en/latest/) with airflow, so I need to add it manually to the image. The ability to have multiple workers, each with different images is to allow for each worker to be dedicated to its individual task.

iamsudip:
@javsalgar Can we re look at this PR and get this merged? We also have requirements to periodically sync dags from multiple git repos.

marcosbc:
@iamsudip Sure, feel free to create a new PR and make the relevant changes. Just make sure to link to this PR to avoid confusion.

PR #1778, opened 2019-12-26
[bitnami/rabbitmq] sort helm parameters

jcpunk:
**Description of the change**

Sort the helm parameters by name so that they can be much more easily located.

**Benefits**

It is easier to find a parameter and its relatives due to sorting.

**Possible drawbacks**

If the existing order was important for users of the readme, they will be disappointed.

**Applicable issues**

None

**Additional information**

**Checklist** 


carrodher:
Hi, thanks for your contribution but we are following a specific order in all the Bitnami charts and this change doesn't follow this standard.
Usually, parameters are documented in the README in the same order that appears in the _values.yaml_.

PR #1805, opened 2020-01-06
[bitnami/elasticsearch] Change empty tolerations values to [] from {}

RobAtticus:
Using the empty value of {} instead of [] causes warnings when helm
attempts to override this with user-specified tolerations with:
```
Warning: ... Cannot overwrite table item 'tolerations', with non table value: map[]
```

Changing this to [] resolves the warning. Solution was found in
the following Github issue:
https://github.com/helm/charts/issues/12918

**Checklist**


RobAtticus:
Awesome thank you!

PR #1811, opened 2020-01-08
[bitnami/elasticsearch] Remove obsolete parameters from README

RobAtticus:
nodeAffinity and antiAffinity appear to have been replaced with
just passing in `affinity` values for coordinating/data/master/ingest.
Therefore to avoid confusion, these parameters should be removed
from the README. There may be more obsolete parameters, but these
were the few I discovered in my course of using the chart.


carrodher:
Thanks for your contribution! Before continuing you should increment the chart version in the `Chart.yml` according to [semver](http://semver.org/):

In this case bumping the patch version is fine.

RobAtticus:
I didn't bump the chart since nothing in the code part of the chart actually changed. Happy to bump it up if you think its worth it though.

carrodher:
Yes, the chart version should be bumped even in the case of changes in the README; in the end, everything is packed in the same tarball so it should be considered a new version.
Thanks!

RobAtticus:
Sure thing, done.

PR #1823, opened 2020-01-14
[bitnami/postgresql-ha] Allow specifying pgpool configuration via configmap

juan131:

**Description of the change**

This PR allows specifying pgpool configuration (`pgpool.conf) via configmap as it's possible for other configuration files (e.g. `postgresql.conf` or `pg_hba.conf`)

**Benefits**

More flexibility to configure Pgpool 

**Possible drawbacks**

None

**Checklist**

### DIFF
--- a/bitnami/postgresql-ha/templates/_helpers.tpl
+++ b/bitnami/postgresql-ha/templates/_helpers.tpl
@@ -525,6 +525,17 @@ Return the PostgreSQL extended configuration configmap.
 {{- end -}}
 
 {{/*
+Return the Pgpool configuration configmap.
+*/}}
+{{- define "postgresql-ha.pgpoolConfigurationCM" -}}
+{{- if .Values.pgpool.configurationCM -}}
+{{- printf "%s" (tpl .Values.pgpool.configurationCM $) -}}
+{{- else -}}
+{{- printf "%s-configuration" (include "postgresql-ha.pgpool" .) -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
 Return the PostgreSQL initdb scripts configmap.
 */}}
 {{- define "postgresql-ha.postgresqlInitdbScriptsCM" -}}
--- a/bitnami/postgresql-ha/templates/pgpool/configmap.yaml
+++ b/bitnami/postgresql-ha/templates/pgpool/configmap.yaml
@@ -10,6 +10,6 @@ data:
   {{- (.Files.Glob "files/pgpool.conf").AsConfig | nindent 2 }}
   {{- else if .Values.pgpool.configuration }}
   pgpool.conf: |
-  {{- .Values.pgpool.configuration | nindent 4 }}
+  {{- include "postgresql-ha.tplValue" (dict "value" .Values.pgpool.configuration "context" $) | nindent 4 }}
   {{- end }}
 {{- end }}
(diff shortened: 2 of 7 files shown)


bitnami/postgresql-ha/templates/pgpool/deployment.yaml
```diff
@@ -210,6 +210,10 @@ spec:
           resources: {{- toYaml .Values.pgpool.resources | nindent 12 }}
           {{- end }}
           volumeMounts:
+            {{- if or (.Files.Glob "files/pgool.conf") .Values.pgpool.configuration }}
+            - name: pgpool-config
+              mountPath: /opt/bitnami/pgpool/conf/
```
javsalgar: Is this the only file that will be written in this folder? Can you confirm that pgpool will never write in this file (I imagine examples like sentinel.conf in redis, that gets written)?

bitnami/postgresql-ha/templates/_helpers.tpl
```diff
@@ -524,6 +524,17 @@ Return the PostgreSQL extended configuration configmap.
 {{- end -}}
 {{- end -}}
 
+{{/*
+Return the Pgpool configuration configmap.
+*/}}
+{{- define "postgresql-ha.pgpoolConfigurationCM" -}}
+{{- if .Values.pgpool.configurationCM -}}
+{{- printf "%s" (tpl .Values.pgpool.configurationCM $) -}}
```
javsalgar: With this you will only allow a fixed string or something like `{{ .Values. ... }}` (but no if statements or similar). If this is expected then fine, if not then use tplValue.

bitnami/postgresql-ha/templates/pgpool/configmap.yaml
```diff
@@ -0,0 +1,15 @@
...
+metadata:
+  name: {{ printf "%s-configuration" (include "postgresql-ha.pgpool" .) }}
+  labels: {{- include "postgresql-ha.labels" . | nindent 4 }}
+    app.kubernetes.io/component: pgpool
+data:
+  {{- if (.Files.Glob "files/pgpool.conf") }}
+  {{- (.Files.Glob "files/pgpool.conf").AsConfig | nindent 2 }}
+  {{- else if .Values.pgpool.configuration }}
+  pgpool.conf: |
+  {{- .Values.pgpool.configuration | nindent 4 }}
```
javsalgar: Would it make sense to allow templates here? I do not know if there are settings like addresses that can be set here.

bitnami/postgresql-ha/templates/_helpers.tpl
juan131: I think it's enough with values like `{{ .Values. ... }`. We're using the same approach for every custom configmap/secret.

bitnami/postgresql-ha/templates/pgpool/deployment.yaml
juan131: I can confirm. See https://github.com/bitnami/bitnami-docker-pgpool#configuration-file

bitnami/postgresql-ha/templates/pgpool/configmap.yaml
juan131: It'd make sense. E.g. to set properties like the ones below:
```
backend_hostname0 = 'postgresql-ha-postgresql-0.postgresql-ha-postgresql-headless.default.svc.cluster.local'
```
Let me address this

PR #1843, opened 2020-01-21
[bitnami/prometheus-operator] adds service monitors for kube-apiserver and kubelet services

sameersbn:
**Description of the change**

adds service monitors for kube-apiserver and kubelet services

**Checklist** 

### DIFF
--- a/bitnami/prometheus-operator/Chart.yaml
+++ b/bitnami/prometheus-operator/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: 0.35.0
 description: The Prometheus Operator for Kubernetes provides easy monitoring definitions for Kubernetes services and deployment and management of Prometheus instances.
 name: prometheus-operator
-version: 0.8.3
+version: 0.9.0
 keywords:
 - prometheus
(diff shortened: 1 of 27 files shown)


sameersbn:
> What do you think about adding some documentation in the README about this new feature?

I am not sure what we can add. We have just adding servicemonitors manifests to scrape the kubelet and apiserver endpoints which users could have created themselves

sameersbn:
@carrodher please note I've added another commit that re-indents all the manifests as per bitnami's standard formatting practices

carrodher:
> > What do you think about adding some documentation in the README about this new feature?
> 
> I am not sure what we can add. We have just adding servicemonitors manifests to scrape the kubelet and apiserver endpoints which users could have created themselves

I was thinking about something if the user need to change anything, I mean, installing the new chart version the metrics from cadvisor will appear in Prometheus without any extra configuration?

Just this case, if users need to change something to make it work

sameersbn:
> Just this case, if users need to change something to make it work
enabled by default!

carrodher:
Then LGTM!

PR #1844, opened 2020-01-21
Fix typo: README.md

rlukin:
Fix typo: README.md


carrodher:
Thanks for the contribution, you should increment the chart version in the `Chart.yml` according to [semver](http://semver.org/):

In this case, bumping the patch version is enough.

PR #1858, opened 2020-01-24
[bitnami/harbor] Fix documentation about service exposure

migmartri:
Changed documentation to match the value set in the values.yaml file.


carrodher:
We should bump the chart version even for those small changes

sameersbn:
@carrodher oh sorry about it. I overlooked the requirement

migmartri:
Thanks!

PR #1920, opened 2020-02-13
[bitnami/etcd] keep snapshot history

dacodas:
**Description of the change**

Keep a timestamped etcd snapshot history. 

**Benefits**

For organizations that would like some history of changes, this might be helpful. One could argue for rotating these via another Kubernetes `cronjob` acting on the same persistent volume. Having this as part of the same chart with just a few lines is more helpful for the end user.

**Possible drawbacks**

Depending on your setting for history length, more disk usage. With the default history length and with the use of hardlinks, there is teensy impact by default.

Tried to limit any blast from this `find | rm -f` by minimizing the depth of `find`. Perhaps another filter for file names would be helpful---this carries the assumption that the PV is meant primarily for backups. 

I believe hard links [depend on the filesystem](https://en.wikipedia.org/wiki/Hard_link). FAT or ReFS users, for example, will not be able to use this [according to Wikipedia](https://en.wikipedia.org/wiki/Hard_link). I assert the vast majority of users would not experience this issue.

The use of the `$(( 1 + $VARIABLE ))` may not work with certain `sh`s. I see that these scripts have `bash` in the she-bang, so I don't foresee this to be an issue. 

**Applicable issues**

None

**Additional information**

**Checklist** 


bitnami/etcd/templates/scripts-configmap.yaml
```diff
@@ -260,6 +261,11 @@ data:
     if etcdctl $AUTH_OPTIONS endpoint health; then
         echo "Snapshotting the keyspace..." 1>&3 2>&4
         etcdctl $AUTH_OPTIONS snapshot save "/snapshots/db" 1>&3 2>&4
+        ln /snapshots/db /snapshots/db-$(date "+%Y-%m-%d_%H:%M") 1>&3 2>&4
```
nvtkaszpir: Please note that some systems may not allow filenames with colon.
Also ensure to use UTC time - it makes a lot of things easier.

dacodas: Alright, I added the `-u` flag for UTC time and replaced the colon with a dash :) Thanks!

bitnami/etcd/templates/scripts-configmap.yaml
```diff
@@ -10,6 +10,7 @@
 {{- $etcdPeerProtocol := include "etcd.peerProtocol" . }}
 {{- $etcdClientProtocol := include "etcd.clientProtocol" . }}
 {{- $initSnapshotFilename := .Values.startFromSnapshot.snapshotFilename }}
+{{- $snapshotHistoryLimit := .Values.cronjob.snapshotHistoryLimit }}
```
nvtkaszpir: Variable in template is incorrect, because it is nested under `disasterRecovery`, below is code change suggestion (not tested, though):

```suggestion
{{- $snapshotHistoryLimit := .Values.disasterRecovery.cronjob.snapshotHistoryLimit }}
```

dacodas:
My apologies for this; I was a bit cheeky and made this pull request without testing. I have now tested, and will be submitting a different pull requests that both implements this features and addresses #1921

dacodas:
This PR includes the commits from #1958

bitnami/etcd/templates/scripts-configmap.yaml
```diff
@@ -245,8 +246,9 @@ data:
     exec 3>&1
     exec 4>&2
 
-    if [ "${BITNAMI_DEBUG}" = true ]; then
+    if [ "${BITNAMI_SNAPSHOT_DEBUG}" = true ]; then
```
rafariossaa: Hi, I would suggest you to leave this environment variable as `BITNAMI_DEBUG`, it is our standard env var to enable debug.

bitnami/etcd/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: etcd
-version: 4.4.14
+version: 4.6.0
```
rafariossaa: Just bump it to 4.5.0

bitnami/etcd/templates/scripts-configmap.yaml
dacodas: I am receiving conflicting request from @rafariossaa and @miguelaeh [here](https://github.com/bitnami/charts/pull/1958#pullrequestreview-363271624). I am partial towards going with @miguelaeh, but if you have a discussion and come to a consensus, please let me know.

dacodas: PR #1958 that implemented this and enabled debug for either `BITNAMI_DEBUG` or `BITNAMI_SNAPSHOT_DEBUG` was merged this morning, so this is now solved.

bitnami/etcd/Chart.yaml
dacodas: My apologies about this, this PR here was actually dependent on PR #1958. Now that #1958 was merged, this change makes more sense. It's a change in the public API that is backwards compatible so I think the bump from 4.5.0 to 4.6.0 makes sense now. Let me know!

dacodas:
Alright, the related PR #1958 was merged this morning, so the details of this PR here are a bit different. I am requesting review from @miguelaeh, and re-requesting from @rafariossaa if you would be so kind.

bitnami/etcd/Chart.yaml
rafariossaa: ok, thanks.
Now that the other PR is merged is not a jump of 2.

bitnami/etcd/templates/scripts-configmap.yaml
```diff
@@ -268,6 +269,11 @@ data:
         if etcdctl $AUTH_OPTIONS endpoint health; then
             echo "Snapshotting the keyspace..." 1>&3 2>&4
             etcdctl $AUTH_OPTIONS snapshot save "/snapshots/db" 1>&3 2>&4
+            ln /snapshots/db /snapshots/db-$(date -u "+%Y-%m-%d_%H-%M") 1>&3 2>&4
```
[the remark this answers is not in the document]
dacodas: Yes, I wanted to change the restore logic as little as possible. I thought making the hard-link was the best way to keep anyone relying on there being a `/snapshots/db` file happy while also not creating two completely separate files and taking twice the disk space.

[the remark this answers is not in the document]
dacodas: Since making the hard-link only adds new files but doesn't change the restore functionality I was more confident making this change. I can see why saving one line might be helpful. If bitnami prefers to do it without the hard link, I think I will need a bit more help to be confident that I'm not messing anyone's restore up.

bitnami/etcd/templates/scripts-configmap.yaml
```diff
@@ -268,6 +269,11 @@ data:
...
+            find /snapshots/ -maxdepth 1 -type f -name 'db-*' \
+                    | sort -r \
+                    | tail -n+$((1 + {{ $snapshotHistoryLimit }})) \
```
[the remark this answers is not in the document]
dacodas: I think the two main parts here are that the `find` is filtering on `-name 'db-*'` and `tail -n+2` returns not two lines, but the entire tail after the first two lines. 

Here is an example of what I mean: 

```bash
strackd@home /tmp/etcd-tail-test
$ ls
test.bash

strackd@home /tmp/etcd-tail-test
$ cat test.bash
#!/bin/bash

NUMBER_TO_KEEP=1

touch db
touch db-$(date -u "+%Y-%m-%d_%H-%M")
touch db-$(date -d "tomorrow" -u "+%Y-%m-%d_%H-%M")
touch db-$(date -d "saturday" -u "+%Y-%m-%d_%H-%M")
touch db-$(date -d "sunday" -u "+%Y-%m-%d_%H-%M")

FIND_ALL=$(find . -maxdepth 1 -type f | sort -r)
FIND_FILTERED=$(find . -maxdepth 1 -type f -name 'db-*' | sort -r)
FIND_TO_DELETE=$(<<< "$FIND_FILTERED" tail -n+$((1 + $NUMBER_TO_KEEP)))
FIND_TO_KEEP=$( diff <( echo "$FIND_ALL" ) <(echo "$FIND_TO_DELETE" ) )

cat <<EOF
All files: 
$FIND_ALL

All filtered files:
$FIND_FILTERED

All files to delete:
$FIND_TO_DELETE

All files to keep:
$FIND_TO_KEEP
EOF


strackd@home /tmp/etcd-tail-test
$ ./test.bash 
All files: 
./test.bash
./db-2020-03-01_00-00
./db-2020-02-29_00-00
./db-2020-02-28_20-12
./db-2020-02-27_20-12
./db

All filtered files:
./db-2020-03-01_00-00
./db-2020-02-29_00-00
./db-2020-02-28_20-12
./db-2020-02-27_20-12

All files to delete:
./db-2020-02-29_00-00
./db-2020-02-28_20-12
./db-2020-02-27_20-12

All files to keep:
1,2d0
< ./test.bash
< ./db-2020-03-01_00-00
6d3
< ./db
```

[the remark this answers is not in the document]
miguelaeh: Thank you very much for the clarification!! It is fine then.

PR #1922, opened 2020-02-13
[bitnami/etcd] Ensure disaster recovery snapshot script works.

### LINKED ISSUE #1921: [bitnami/etcd] disaster recovery option is acually not working if replicas >1
**Which chart**:

bitnami/etcd version 4.4.14

**Description**

If you enable `disasterRecovery: true` and set `replicaCount: 3` then cronjob script to make snapshots will actually throw few errors:
- BITNAMI_DEBUG: unbound variable
- (and if you fix above) etcdctl snapshot actually works with only one member at a time - I believe this was caused by incorrect example in etcd docs itself (https://github.com/etcd-io/etcd/pull/11074) but unfortunately also assumes that snapshot feature was tested only with replicaCount 1.

First, install nfs-server and then etcd:

```bash
helm install stable/nfs-server-provisioner

helm repo add bitnami https://charts.bitnami.com/bitnami
helm install bitnami/etcd -f etcd-vals.yaml
```
etcd-vals.yaml
```yaml
statefulset:
  replicaCount: 3
auth:
  rbac:
    enabled: false
persistence:
  enabled: true
  size: 5Gi

disasterRecovery:
  enabled: true
  cronjob:
    schedule: "*/5 * * * *"
  pvc:
    size: 20Gi
    storageClassName: nfs

```

See what is going on with cronjobs, they are failing and it is visible if we decrease schedule even more.

When cronjob starts it fails with message:

```text
BITNAMI_DEBUG: unbound variable
```
After fixing that by editing configmap there will be another message:

```text
Error: snapshot must be requested to one selected node, not multiple [http://be-etcd-0.be-etcd-headless.my-namespace.svc.cluster.local:2379 http://be-etcd-1.be-etcd-headless.my-namespace.svc.cluster.local:2379 http://be-etcd-2.be-etcd-headless.my-namespace.svc.cluster.local:2379]```
```

1. no unbound vars in script
2. properly used etcd snapshot command according to the documentation - https://github.com/etcd-io/etcd/pull/11074

I cna provide working fix via PR.

- Output of `helm version`:

```
Client: &version.Version{SemVer:"v2.14.3", GitCommit:"0e7f3b6637f7af8fcfddb3d2941fcc7cbebb0085", GitTreeState:"clean"}
Server: &version.Version{SemVer:"v2.14.3", GitCommit:"0e7f3b6637f7af8fcfddb3d2941fcc7cbebb0085", GitTreeState:"clean"}

```

```
Client Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.0", GitCommit:"641856db18352033a0d96dbc99153fa3b27298e5", GitTreeState:"clean", BuildDate:"2019-03-25T15:53:57Z", GoVersion:"go1.12.1", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"14+", GitVersion:"v1.14.9-gke.23", GitCommit:"a392833a56a82ab79d7e14695365b6df08f9ae38", GitTreeState:"clean", BuildDate:"2020-01-17T23:04:32Z", GoVersion:"go1.12.12b4", Compiler:"gc", Platform:"linux/amd64"}

```
This was tested on GKE.

nvtkaszpir:
**Description of the change**

Ensure etcd snapshot works as expected.

**Benefits**

Snapshots works.

**Possible drawbacks**

None.

**Applicable issues**

  - fixes #1921 

**Additional information**

Example output with failed first node in the cluster
```text
kubectl get pods -n bitnami-r3
NAME                                      READY   STATUS      RESTARTS   AGE
bn-etcd-r3-0                              0/1     Error       11         34m
bn-etcd-r3-1                              1/1     Running     0          44m
bn-etcd-r3-2                              1/1     Running     0          44m
bn-etcd-r3-snapshotter-1581632100-7pptq   0/1     Completed   0          5m10s
bn-etcd-r3-snapshotter-1581632400-fdmps   1/1     Running     0          7s

kubectl get pods -n bitnami-r3
NAME                                      READY   STATUS             RESTARTS   AGE
bn-etcd-r3-0                              0/1     CrashLoopBackOff   11         35m
bn-etcd-r3-1                              1/1     Running            0          45m
bn-etcd-r3-2                              1/1     Running            0          45m
bn-etcd-r3-snapshotter-1581632400-fdmps   0/1     Completed          0          49s
```

Output from snapshot executing pod
```text
Using etcd endpoint http://bn-etcd-r3-0.bn-etcd-r3-headless.bitnami-r3.svc.cluster.local:2379
Warning - etcd endpoint not healthy http://bn-etcd-r3-0.bn-etcd-r3-headless.bitnami-r3.svc.cluster.local:2379
Trying another endpoint.
Using etcd endpoint http://bn-etcd-r3-1.bn-etcd-r3-headless.bitnami-r3.svc.cluster.local:2379
Snapshotting the keyspace...
Snapshot saved at /snapshots/db

```

And in case of failed cluster:
```text
NAME                                      READY   STATUS             RESTARTS   AGE
bn-etcd-r3-0                              0/1     CrashLoopBackOff   17         70m
bn-etcd-r3-1                              0/1     Running            0          27s
bn-etcd-r3-2                              1/1     Running            0          80m
bn-etcd-r3-snapshotter-1581634200-xv9ww   0/1     Completed          0          5m44s
bn-etcd-r3-snapshotter-1581634500-mm2sh   0/1     Error              1          41s

```

```text
Error - all etcd endpoits are unhealthy!
```

**Checklist** 

### DIFF
--- a/bitnami/etcd/templates/cronjob.yaml
+++ b/bitnami/etcd/templates/cronjob.yaml
@@ -31,6 +31,8 @@ spec:
               command:
                 - /scripts/save-snapshot.sh
               env:
+              - name: BITNAMI_DEBUG
+                value: {{ ternary "true" "false" .Values.image.debug | quote }}
               - name: ETCDCTL_API
                 value: "3"
               {{- if .Values.auth.client.secureTransport }}
--- a/bitnami/etcd/templates/scripts-configmap.yaml
+++ b/bitnami/etcd/templates/scripts-configmap.yaml
@@ -254,14 +254,27 @@ data:
     AUTH_OPTIONS="{{ $etcdAuthOptions }}"
     ETCDCTL_ENDPOINTS="{{range $i, $e := until $replicaCount }}{{ $etcdClientProtocol }}://{{ $etcdFullname }}-{{ $e }}.{{ $etcdHeadlessServiceName }}.{{ $releaseNamespace }}.svc.{{ $clusterDomain }}:{{ $clientPort }},{{ end }}"
     # Remove the last comma "," introduced in the string
-    export ETCDCTL_ENDPOINTS="$(sed 's/,/ /g' <<< $ETCDCTL_ENDPOINTS | awk '{$1=$1};1' | sed 's/ /,/g')"
+    export ETCDCTL_ENDPOINTS="${ETCDCTL_ENDPOINTS%%,}"
+    # prepare space separated list of endpoints
+    export etcdctl_endpoints_list="$(sed 's/,/ /g' <<< $ETCDCTL_ENDPOINTS)"
 
     mkdir -p "/snapshots" 1>&3 2>&4
-    if etcdctl $AUTH_OPTIONS endpoint health; then
-        echo "Snapshotting the keyspace..." 1>&3 2>&4
-        etcdctl $AUTH_OPTIONS snapshot save "/snapshots/db" 1>&3 2>&4
-    else
-        echo "Cluster isn't responding!!" 1>&3 2>&4
-        exit 1
-    fi
+
+    # walk over each node in the list, check if node healthy make snapshot and exit, otherwise try another node
+    for etcd_endpoint in $etcdctl_endpoints_list; do
+        export ETCDCTL_ENDPOINTS="${etcd_endpoint}"
+        echo "Using etcd endpoint ${etcd_endpoint}" 1>&3 2>&4
+        if etcdctl $AUTH_OPTIONS endpoint health; then
+            echo "Snapshotting the keyspace..." 1>&3 2>&4
+            etcdctl $AUTH_OPTIONS snapshot save "/snapshots/db" 1>&3 2>&4
+            exit 0
+        else
+            echo "Warning - etcd endpoint not healthy ${etcd_endpoint}" 1>&3 2>&4
+            echo "Trying another endpoint." 1>&3 2>&4
+        fi
+    done
+
+    # exit with error if all endpoints are bad
+    echo "Error - all etcd endpoits are unhealthy!" 1>&3 2>&4
+    exit 1
 {{- end }}
--- a/bitnami/etcd/Chart.yaml
+++ b/bitnami/etcd/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: etcd
-version: 4.4.14
+version: 4.4.15
 appVersion: 3.4.3
 description: etcd is a distributed key value store that provides a reliable way to store data across a cluster of machines
 keywords:


bitnami/etcd/templates/cronjob.yaml
```diff
@@ -31,6 +31,8 @@ spec:
               command:
                 - /scripts/save-snapshot.sh
               env:
+              - name: BITNAMI_DEBUG
```
dani8art: I would add a different value to reflect specifically this functionality. I mean, something like `disasterRecovery.debug`.

bitnami/etcd/templates/scripts-configmap.yaml
```diff
@@ -254,14 +254,27 @@ data:
...
-    if etcdctl $AUTH_OPTIONS endpoint health; then
-        echo "Snapshotting the keyspace..." 1>&3 2>&4
-        etcdctl $AUTH_OPTIONS snapshot save "/snapshots/db" 1>&3 2>&4
-    else
-        echo "Cluster isn't responding!!" 1>&3 2>&4
-        exit 1
-    fi
+
+    # walk over each node in the list, check if node healthy make snapshot and exit, otherwise try another node
+    for etcd_endpoint in $etcdctl_endpoints_list; do
```
dani8art: That's cool and will solve the problem, however, I'm wondering if we can avoid the loop. I mean, when we execute a snapshot whatever is the node, I presume, it is going to do a snapshot of the entire data. If this is the case and we do it three times then we are going to do two snapshots more, aren't we?

bitnami/etcd/templates/cronjob.yaml
nvtkaszpir: This is not for helm chart functionality, this is for debugging scripts and it is already there in the values: `images.debug: true`.
This conforms with the other uses such as in https://github.com/bitnami/charts/blob/master/bitnami/etcd/templates/statefulset.yaml#L97 and this is used for debugging bash scripts like in https://github.com/bitnami/charts/blob/master/bitnami/etcd/templates/scripts-configmap.yaml#L30 for etcd setup or https://github.com/bitnami/charts/blob/master/bitnami/etcd/templates/scripts-configmap.yaml#L193 pre-stop for etcd or probes https://github.com/bitnami/charts/blob/master/bitnami/etcd/templates/scripts-configmap.yaml#L245

So the cron job pod was missing this env var.
Another option would be to set:
```bash
: {BITNAMI_DEBUG:-"false"}
```
in every shell script but then still it would make it harder to use in the case of the cronjob (because it would always be set to false).

bitnami/etcd/templates/scripts-configmap.yaml
nvtkaszpir: I already thought about that, and that's why in line 270 there is `exit 0` statement, which ends the processing the whole script. This way we get only one snapshot.

There is no retry on failed snapshot execution but then the error processing defined in lines 237 will also end the script execution.

I've provided the example outputs in the description of the pull request to show how this works if for example first node is unavailable.

bitnami/etcd/templates/cronjob.yaml
nvtkaszpir: Edited the comment to make it more easier to understand.

bitnami/etcd/templates/scripts-configmap.yaml
```diff
@@ -254,14 +254,27 @@ data:
     AUTH_OPTIONS="{{ $etcdAuthOptions }}"
     ETCDCTL_ENDPOINTS="{{range $i, $e := until $replicaCount }}{{ $etcdClientProtocol }}://{{ $etcdFullname }}-{{ $e }}.{{ $etcdHeadlessServiceName }}.{{ $releaseNamespace }}.svc.{{ $clusterDomain }}:{{ $clientPort }},{{ end }}"
     # Remove the last comma "," introduced in the string
-    export ETCDCTL_ENDPOINTS="$(sed 's/,/ /g' <<< $ETCDCTL_ENDPOINTS | awk '{$1=$1};1' | sed 's/ /,/g')"
+    export ETCDCTL_ENDPOINTS="${ETCDCTL_ENDPOINTS%%,}"
+    # prepare space separated list of endpoints
+    export etcdctl_endpoints_list="$(sed 's/,/ /g' <<< $ETCDCTL_ENDPOINTS)"
```
dani8art: I would say we can avoid "remove the last comma..." when using this substitution.

bitnami/etcd/templates/cronjob.yaml
dani8art: Sorry, I didn't explain well, what I wanted to say here was that it would be good to have both features in different values. That would mean, `image.debug=true` if we want to debug the main container stuff, I mean,  statefulset containers, and `disasterRecovery.debug=true` if we want to debug this specific cronjob for the snapshots. @carrodher or @juan131 could you add your thought here, please?

nvtkaszpir: Ah. Well, in that case it's a feature and not a bug. :)

PR #1939, opened 2020-02-17
Fix stable repo to add bitnami charts

raquel-campuzano:
**Description of the change**

Replace the command you need to use to add Bitnami charts to use the helm stable repo instead. 

**Benefits**

Users of Helm 3 could install all the Helm charts catalog in their clusters.


raquel-campuzano:
The background of this change is when I tried to install Ghost in a cluster, the `helm repo add` command with the bitnami repo didn't work. And I didn't find any reference or link to the repo I have to add to install that or any other upstreamed chart in my cluster.  I think that those charts maintained by Bitnami should include that reference.

carrodher:
The problem is that both repositories are completely independent. You can install the Ghost chart using the Bitnami repo or the stable one.

If you want to install it **from the Bitnami repo** you should be able to follow [the instructions in this README](https://github.com/bitnami/charts#tldr) without any issue:
```
$ helm repo add bitnami https://charts.bitnami.com/bitnami
$ helm search repo bitnami
$ helm install my-release bitnami/<chart>
```

Those instructions translated to a real example should be something like:
- Starting without any repo in the cluster
```
▶ helm repo list
Error: no repositories to show
```
- Add Bitnami repo:
```
▶ helm repo add bitnami https://charts.bitnami.com/bitnami
"bitnami" has been added to your repositories
```
- Take a look at the charts present in this repo (I just added a filter to only show Ghost):
```
▶ helm search repo bitnami | grep ghost
NAME                            	CHART VERSION	APP VERSION            	DESCRIPTION
bitnami/ghost                   	9.1.7        	3.5.0                  	A simple, powerful publishing platform that all...
```
- Install it from this repo:
```
▶ helm install my-ghost bitnami/ghost
NAME: my-ghost
LAST DEPLOYED: Tue Feb 18 12:03:10 2020
NAMESPACE: default
STATUS: deployed
REVISION: 1
NOTES:
###############################################################################
### ERROR: You did not provide an external host in your 'helm install' call ###
###############################################################################

This deployment will be incomplete until you configure Ghost with a resolvable
host. To configure Ghost with the URL of your service:

1. Get the Ghost URL by running:

  NOTE: It may take a few minutes for the LoadBalancer IP to be available.
        Watch the status with: 'kubectl get svc --namespace default -w my-ghost'

  export APP_HOST=$(kubectl get svc --namespace default my-ghost --template "{{ range (index .status.loadBalancer.ingress 0) }}{{ . }}{{ end }}")
  export APP_PASSWORD=$(kubectl get secret --namespace default my-ghost -o jsonpath="{.data.ghost-password}" | base64 --decode)
  export APP_DATABASE_PASSWORD=$(kubectl get secret --namespace default my-ghost-mariadb -o jsonpath="{.data.mariadb-password}" | base64 --decode)

2. Complete your Ghost deployment by running:

  helm upgrade my-ghost stable/ghost \
    --set service.type=LoadBalancer,ghostHost=$APP_HOST,ghostPassword=$APP_PASSWORD,mariadb.db.password=$APP_DATABASE_PASSWORD
```

And that's all, everything is done following the instructions in the README and using Helm 3.

On the other hand, the ghost chart is also present in the [stable repo](https://github.com/helm/chart). In this repo, instructions and more or less the same but with a small change between Helm 2 and Helm 3.
In Helm 2, the stable repository was added by default while in Helm 3, users need to add it on demand.
They documented how to add the stable repo in Helm 3 in [this section](https://github.com/helm/charts#how-do-i-enable-the-stable-repository-for-helm-3) of their README.

To install the chart from stable instead of Bitnami, the process is the same but adding the stable repo instead of the bitnami one in the `helm repo add` command and then installing it using `helm install whatever stable/whatever` instead of `bitnami/whatever`.

Summarizing, there are different sources to install charts (at this moment, because stable will be deprecated in some months). So depend on what repo you want to use you should follow one or other instructions (all of them are very very similar).
But in this case, we are documenting in this README how to install charts from Bitnami repo, the instructions to install it from stable repo should be placed in the stable README.

raquel-campuzano:
Leave the `helm repo add` command as it was.

PR #1941, opened 2020-02-17
[bitnami/harbor] Store passwords of redis, clair and notary in secrets

marcostvz:
Taking some changes from:
https://github.com/goharbor/harbor-helm/commit/cb8c71d73aadf3c6c7fc796dbb87c2802d400eb9

Signed-off-by: Marcos Estevez <marcos.stvz@gmail.com>

**Description of the change**

Some passwords were shown in configmaps, moving those to secrets. Fixes some missing docs and misplaced configs in values.yaml.

**Benefits**

Passwords are in their place

**Possible drawbacks**

None

**Applicable issues**

None

**Checklist**


bitnami/harbor/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: harbor
-version: 4.0.0
+version: 4.0.1
```
miguelaeh: You should bump the minor version

bitnami/harbor/README.md
```diff
@@ -255,6 +255,27 @@ The following table lists the configurable parameters of the Harbor chart and th
 | `clair.podAnnotations`                                                      | Annotations to add to the clair pod                                      | `{}`                                                    |
 | `clair.livenessProbe`                                                       | Liveness probe configuration                                             | `Check values.yaml file`                                |
 | `clair.readinessProbe`                                                      | Readiness probe configuration                                            | `Check values.yaml file`                                |
+| **Notary**                                                                  |
+| `notary.enabled`                                                            | Enable Notary?                                                           | `true`                                                  |
```
miguelaeh:
```diff
-Enable Notary?
+Enable Notary
```

bitnami/harbor/conf/clair.yaml
```diff
@@ -0,0 +1,21 @@
+clair:
```
miguelaeh: Why are you duplicating the values on this file? They are already into the `values.yaml`, Please do no create a new file add the necessary to the `values.yaml`

bitnami/harbor/templates/clair/clair-dpl.yaml
```diff
@@ -15,7 +15,7 @@ spec:
       labels: {{- include "harbor.labels" . | nindent 8 }}
         app.kubernetes.io/component: clair
       annotations:
-        checksum/configmap: {{ include (print $.Template.BasePath "/clair/clair-cm.yaml") . | sha256sum }}
```
miguelaeh: Why are you doing this? This file has no password inside it is correct to be a configmap.

bitnami/harbor/templates/clair/clair-secret.yaml
```diff
@@ -0,0 +1,11 @@
+{{- if .Values.clair.enabled }}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ template "harbor.clair" . }}
+  labels:
+{{ include "harbor.labels" . | indent 4 }}
+type: Opaque
+data:
+  config.yaml: {{ tpl (.Files.Get "conf/clair.yaml") . | b64enc }}
```
miguelaeh: You are creating a secret with a configuration file that contains no passwords, could you explain the need to do this?.

bitnami/harbor/templates/core/core-secret-envvars.yaml
```diff
@@ -5,6 +5,8 @@ metadata:
   labels: {{- include "harbor.labels" . | nindent 4 }}
 type: Opaque
 data:
+  _REDIS_URL: "{{ template "harbor.redisForCore" . }}"
```
miguelaeh: The `_REDIS_URL` is not sensitive data it is not needed to move it to a secret.

bitnami/harbor/templates/notary/notary-secret.yaml
```diff
@@ -0,0 +1,18 @@
+{{- if .Values.notary.enabled }}
```
miguelaeh: The certificate are mounted as volumes so I don't see the need to put the path into a secret.

bitnami/harbor/templates/clair/clair-secret.yaml
marcostvz: if you use an external database, yes, the password is in the source section:
`clair:
      database:
        type: pgsql
        options:
          source: "postgres://harbor:MYPASSWORD@my-postgresql:5432/clair?sslmode=disable"`

bitnami/harbor/templates/core/core-secret-envvars.yaml
marcostvz: If you use an external redis with password, there you are:
```
$ kubectl get configmap -o yaml test-harbor-core-envvars
apiVersion: v1
data:
  _REDIS_URL: my-redis:6379,100,MYREDISPASSWORD
  _REDIS_URL_REG: redis://redis:MYREDISPASSWORD@test-redis:6379/2
...
```

bitnami/harbor/conf/clair.yaml
marcostvz: This file comes from moving this out from [clair-cm.yaml](https://github.com/bitnami/charts/pull/1941/files#diff-c1d79c0e047c00a17a428ed4347eb8f8L9-L34)

bitnami/harbor/templates/clair/clair-dpl.yaml
marcostvz: Answered in https://github.com/bitnami/charts/pull/1941/files#r380523144

bitnami/harbor/templates/notary/notary-secret.yaml
marcostvz: We moved server and signer jsons here because they stored our external DB password in plain within storage.db_url, same as happened with clair.

About the certificates, having tls.key in a Secret fits much more than in a Configmap.

marcostvz:
Hi @miguelaeh, 

thanks a lot for your comments and your quick review! I dropped a some replies to your comments, let me know if you need more clarification.

Also, take into account that with this PR I am attempting to bring the security improvements that goharbor/harbor chart applied several months ago in https://github.com/goharbor/harbor-helm/commit/cb8c71d73aadf3c6c7fc796dbb87c2802d400eb9

Kind regards :)

bitnami/harbor/conf/clair.yaml
miguelaeh: Well, in that case, move those values to the `values.yaml`

bitnami/harbor/templates/clair/clair-dpl.yaml
miguelaeh: For that case I think it should be enough with the `redis.password` parameter

bitnami/harbor/templates/clair/clair-secret.yaml
miguelaeh: There is also a `posgresql.postgresqlPassword` that should be use for that purpose.

bitnami/harbor/templates/core/core-secret-envvars.yaml
miguelaeh: (Already answered in the previous comment)

bitnami/harbor/templates/notary/notary-secret.yaml
miguelaeh: Not sure about this point, shouldn't it be the same case as the password above?

miguelaeh:
Hi @marcostvz ,
Thanks for addressing the changes, I answered again, basically, I think the passwords parameter should be used for that purpose but it seems they end into the URL anyway, now I understand your changes. Anyway, I don't think this is the best way to do this, maybe it could be possible to pass the passwords to the image and create the URL in the image so we don't have them in the config map nor the secret.
Anyway, thank you very much for clarifying the changes!!!

bitnami/harbor/templates/clair/clair-secret.yaml
marcostvz: Please, take into account that I am using the externalDatabase feature, so I have to set `postgresql.enabled: false` to avoid installing stable/postgresql that I won't use. This `postgresql.postgresqlPassword` has no effect in my use case.

bitnami/harbor/templates/clair/clair-dpl.yaml
marcostvz: Please, take into account that I am using the externalRedis feature, so I have to set `redis.enabled: false` to avoid installing stable/redis that I won't use.

bitnami/harbor/conf/clair.yaml
marcostvz: This is the config file that clair will use [here](https://github.com/bitnami/charts/pull/1941/files#diff-6b752c0b3398da15053a9352f3413995R79), it is not related to values.yaml

bitnami/harbor/templates/notary/notary-server.yaml
```diff
@@ -51,33 +51,34 @@ spec:
             - name: DB_URL
               value: {{ include "harbor.database.notaryServer" . }}
           volumeMounts:
-            - name: notary-config
-              mountPath: /etc/notary
+            - name: config
+              mountPath: /etc/notary/server-config.postgres.json
+              subPath: server.json
             - name: root-certificate
```
miguelaeh: This will through an error because you changed the name to the volume.

bitnami/harbor/templates/notary/notary-signer.yaml
```diff
@@ -52,26 +52,24 @@ spec:
             - name: NOTARY_SIGNER_DEFAULTALIAS
               value: defaultalias
           volumeMounts:
-            - name: notary-config
-              mountPath: /etc/notary
-            {{- if .Values.notary.secretName }}
-            - name: notary-cert
-              mountPath: /etc/ssl/notary/cert/notary-signer-ca.crt
```
miguelaeh: Deleting this certificate is throwing an error: 
```
{"level":"fatal","msg":"Unable to configure TLS to the trust service: could not read CA certificate \"/etc/notary/notary-signer-ca.crt\": open /etc/notary/notary-signer-ca.crt: no such file or directory","time":"2020-02-21T12:29:23Z"}
```

bitnami/harbor/templates/notary/notary-signer.yaml
```diff
@@ -52,26 +52,24 @@ spec:
...
-              subPath: ca
-            - name: notary-cert
-              mountPath: /etc/ssl/notary/cert/notary-signer.crt
```
miguelaeh: Also this one:
```
{"level":"fatal","msg":"unable to set up TLS: Could not load X509 key pair (cert: \"/etc/notary/notary-signer.crt\", key: \"/etc/notary/notary-signer.key\"): open /etc/notary/notary-signer.crt: no such file or directory","time":"2020-02-21T12:29:24Z"}
```

bitnami/harbor/templates/notary/notary-server.yaml
marcostvz: Sorry about this, I had notary disabled and overlooked this. It is fixed now

marcostvz:
Hey Miguel,

Thanks a lot for your comments and your accurate review.

I 100% agree with you that probably there could be a better way to solve this bug, but I'd ask you to consider this as not-a-perfect-but-good-enough solution which brings some improvements to what we have now.

Bumping major version now.

marcostvz:
[WIP] I am going to rollback the changes over notary certificates. It is okay to move them from ConfigMaps into Secrets, but paths should not change

marcostvz:
Hello @miguelaeh,

finally I got some time to restore the notary as it was, The only real change is that we're using secrets now instead of configmaps.

I'll appreciate any comment about this.

Marcos

bitnami/harbor/README.md
```diff
@@ -255,6 +255,27 @@ The following table lists the configurable parameters of the Harbor chart and th
 | `clair.podAnnotations`                                                      | Annotations to add to the clair pod          | `{}`        |
 | `clair.livenessProbe`                                                       | Liveness probe configuration                 | `Check values.yaml file` |
 | `clair.readinessProbe`                                                      | Readiness probe configuration                | `Check values.yaml file` |
+| **Notary**                                                                  |
```
miguelaeh: There are variables here that are not used

bitnami/harbor/README.md
```diff
@@ -255,6 +255,27 @@ The following table lists the configurable parameters of the Harbor chart and th
...
+| `notarySignerImage.tag`                                                     | Tag for notary signer image                                              | `1.10.1-debian-10-r0`                                   |
+| `notarySignerImage.pullPolicy`                                              | Harbor notary signer image pull policy                                   | `IfNotPresent`                                          |
+| `notary.server.replicas`                                                    | The replica count                                                        | `1`                                                     |
+| `notary.server.resources`                                                   | The [resources] to allocate for container                                | `{}`                                                    |
+| `notary.signer.image.repository`                                            | Repository for notary signer image                                       | `goharbor/notary-signer-photon`                         |
```
miguelaeh: This is duplicated with:
```
notarySignerImage.repository
```

bitnami/harbor/README.md
marcostvz: true, I just reviewed them one by one :)

bitnami/harbor/conf/clair.yaml
javsalgar: I'm not clear on this because I think this forces the user to clone the bitnami/charts repository in order to work. Would this file become part of the final tarball?

javsalgar:
Hi,

Added a comment

bitnami/harbor/conf/clair.yaml
marcostvz: Yes, `helm package` will package the whole directory, except files specified in a `.helmignore`.

This approach is the same as in here, where we import the `cert` folder: https://github.com/bitnami/charts/blob/master/bitnami/harbor/templates/core/core-secret.yaml#L24-L27

miguelaeh:
Hi @marcostvz,
Thank you very much for address the changes, let us test the PR in our pipeline before accept it.

marcostvz:
Yay! Thanks a lot!

Issue #1905, opened 2020-02-11
[bitnami/fluentd] Unreadable log mounts


austince:
**Which chart**:

Fluentd 0.4.10

**Description**

Mounted logs are unreadable. It seems similar to this issue: https://stackoverflow.com/questions/51671212/fluentd-log-unreadable-it-is-excluded-and-would-be-examined-next-time

I have already set the FLUENT_UID to '0'. What other information do you need to find out what else I could be doing wrong?

**Steps to reproduce the issue:**

1. Deploy the chart on EKS
2. Tail the logs of the daemonset
3. See error

**Describe the results you received:**

```
2020-02-11 00:28:26 +0000 [warn]: #0 /var/log/containers/prometheus-prometheus-operator-prometheus-1_monitoring_prometheus-config-reloader-5de238447ab71c5f5e77c3890822c9fe5e0caefb212003488747ba73d4a0e603.log unreadable. It is excluded and would be examined next time.
```

**Describe the results you expected:**

**Additional information you deem important (e.g. issue happens only occasionally):**

**Version of Helm and Kubernetes**:

- Output of `helm version`:

```
Client: &version.Version{SemVer:"v2.14.1", GitCommit:"5270352a09c7e8b6e8c9593002a73535276507c0", GitTreeState:"clean"}
Server: &version.Version{SemVer:"v2.14.1", GitCommit:"5270352a09c7e8b6e8c9593002a73535276507c0", GitTreeState:"clean"}
```

- Output of `kubectl version`:

```
Client Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.2", GitCommit:"59603c6e503c87169aea6106f57b9f242f64df89", GitTreeState:"clean", BuildDate:"2020-01-21T22:17:28Z", GoVersion:"go1.13.5", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"14+", GitVersion:"v1.14.9-eks-c0eccc", GitCommit:"c0eccca51d7500bb03b2f163dd8d534ffeb2f7a2", GitTreeState:"clean", BuildDate:"2019-12-22T23:14:11Z", GoVersion:"go1.12.12", Compiler:"gc", Platform:"linux/amd64"}
```

austince:
I have another Fluentd daemonset running as well - could they be interfering with one another?

javsalgar:
Hi,

Did you deploy it with the security group as root? 

```
[...]

PR #1952, opened 2020-02-19
[bitnami/fluentd] mount /var/lib/docker/containers as readonly

austince:
**Description of the change**

Mounts the `/var/lib/docker/containers` host path as readonly.

**Benefits**

The proper mount will be sent to the CSI driver: https://kubernetes.io/docs/concepts/storage/volumes/

**Possible drawbacks**

None.

**Applicable issues**

  - possibly related to #1905

**Additional information**

Other fluentd charts that do the same:
- https://github.com/kiwigrid/helm-charts/blob/92ae60e00f7753c28e62ca37e8e2f1d6005ab92d/charts/fluentd-elasticsearch/templates/daemonset.yaml#L137-L139
- https://github.com/helm/charts/blob/7ccc9f99d7ab6b9554624985d9c9b9723b7253c0/incubator/fluentd-cloudwatch/templates/daemonset.yaml#L69-L71

**Checklist** 

### DIFF
--- a/bitnami/fluentd/Chart.yaml
+++ b/bitnami/fluentd/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: fluentd
-version: 0.4.15
+version: 0.4.16
 appVersion: 1.9.2
 description: Fluentd is an open source data collector for unified logging layer
 keywords:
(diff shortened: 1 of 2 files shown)


rptaylor:
@juan131 @austince what about /var/log ?  It is also writeable by fluentd but I think/hope fluentd should not need to write to /var/log on the host filesystem.

PR #1959, opened 2020-02-20
[bitnami/mysql] Fix for initContainer volume-permissions

mayanksingh0895:
The initcontainer volume permission fails when the existing volume claim is specified and its value is not "data". As in case existing volume claim name is different from "data", then this specified name should be referenced in the initcontainer volume permission. But as per current implementation the volumeMount name of initcontainer is hardcoded to "data". Due to this reason the initcontainer fails due to volumeMount name not found when existing volume claim name is different from "data".
I have purposed to use default template function with default value as "data" in case existing volume claim name is not provided.

### DIFF
--- a/bitnami/mysql/Chart.yaml
+++ b/bitnami/mysql/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: mysql
-version: 6.9.0
+version: 6.9.1
 appVersion: 8.0.19
 description: Chart to create a Highly available MySQL cluster
 keywords:
--- a/bitnami/mysql/templates/master-statefulset.yaml
+++ b/bitnami/mysql/templates/master-statefulset.yaml
@@ -54,7 +54,7 @@ spec:
           resources: {{- toYaml .Values.volumePermissions.resources | nindent 12 }}
           {{- end }}
           volumeMounts:
-            - name: data
+            - name: {{ default "data" .Values.master.persistence.existingClaim | quote }}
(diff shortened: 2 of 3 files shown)


mayanksingh0895:
> Hi @mayanksingh0895 ,
> It is necessary to bump the version.
> Regarding the change, the `existingClaim` parameter is to use an existing claim it is not related to the volume name, I guess it should be better to change the following:
> 
> ```diff
> 251 {{- else if and .Values.master.persistence.enabled .Values.master.persistence.existingClaim }}
> -252         - name: {{ .Values.master.persistence.existingClaim }}
> +252         - name: "data"
> 253           persistentVolumeClaim:
> ```
> 
> I don't see why the volume name should not be hardcoded.

I agree with the changes requested and made some additional changes, so that these changes can work. There was reference of volumeName is volumeMount as well, so first commit was done to incorporate volumeMount name reference as well.

With the last commit I have reverted the changes for first commit and added changes suggested by you and added a change in volumeMount name as well so that the volume name is in sync.

mayanksingh0895:
Added similar fix for slave-statefulset.yaml

mayanksingh0895:
> I thank you very much for address the change you need also to bump the Chart version at `Chart.yaml`

Sure @miguelaeh , I have bumped the chart version to 6.9.1

mayanksingh0895:
> Thank you very much!!!!

My pleasure.
Thanks for the suggestions and for merging the PR.

PR #1973, opened 2020-02-25
[bitnami/harbor] - make core ca cert optional

### LINKED ISSUE #1863: [bitnami/harbor] Required certificate authority
Hi,

I deployed an instance of Harbor reading from a prepopulated k8s tls certificate generated through a trusted certificate authority.

To do so, after getting the certificate I 1) created the k8s secret by running 

```
kubectl create secret tls harbor-ingress-certs \
  --key certs//privkey1.pem \
  --cert certs/fullchain1.pem 
```

And 2) deployed harbor setting the external tls name `--set service.tls.secretName=harbor-ingress-certs`

The issue is that the `core` pod will not start because can not mount the CA file since it is not part of the secret https://github.com/bitnami/charts/blob/449f80cf4dda6fa105b3801292ad6148ae0d69b1/bitnami/harbor/templates/core/core-dpl.yaml#L134

To workaround it I downloaded the public, trusted CA from the internet and manually added it to the secret but, is a CA required even if it is from upstream? Would it make sense to make the CA optional?

Thanks!

dani8art:
**Description of the change**

Make the "ca.crt" optional in secret for certificate. When users are generating their own certificate and the secret contains no certificate for CA.

**Benefits**

Users can use certs signed by a trusted third-party CA.

**Possible drawbacks**

N/A

**Applicable issues**

  - fixes #1863

**Additional information**

N/A

### DIFF
--- a/bitnami/harbor/Chart.yaml
+++ b/bitnami/harbor/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: harbor
-version: 4.0.3
+version: 4.1.1
 appVersion: 1.10.1
 description: Harbor is an an open source trusted cloud native registry project that stores, signs, and scans content
 keywords:
--- a/bitnami/harbor/Chart.yaml
+++ b/bitnami/harbor/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: harbor
-version: 4.1.0
+version: 4.0.3
 appVersion: 1.10.1
 description: Harbor is an an open source trusted cloud native registry project that stores, signs, and scans content
 keywords:
--- a/bitnami/harbor/templates/core/core-dpl.yaml
+++ b/bitnami/harbor/templates/core/core-dpl.yaml
@@ -97,8 +97,7 @@ spec:
               subPath: tls.key
             {{- if .Values.service.tls.enabled }}
             - name: ca-download
-              mountPath: /etc/core/ca/ca.crt
-              subPath: ca.crt
+              mountPath: /etc/core/ca
             {{- end }}
             {{- if .Values.uaaSecretName }}
             - name: auth-ca-cert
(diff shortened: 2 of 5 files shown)


bitnami/harbor/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: harbor
-version: 4.0.2
+version: 4.1.0
```
juan131: This is just a fix, right? I would increase patch version in that case

bitnami/harbor/templates/core/core-dpl.yaml
```diff
@@ -140,9 +139,6 @@ spec:
           {{- else }}
             secretName: {{ .Values.service.tls.secretName }}
           {{- end }}
-            items:
-              - key: ca.crt
-                path: ca.crt
```
juan131: Well.. This is removing the `ca.crt` instead of making it optional. Or am I missing anything?

bitnami/harbor/Chart.yaml
dani8art: I'm not sure, in fact, we are changing the required content in the secret, it is that a fix or adding/modifying a feature. I don't really know, yes I would go for a patch. Changing it

bitnami/harbor/templates/core/core-dpl.yaml
dani8art: I would say that removing `items` we are going to mount just if the ca.crt exist but I will double-check it.

dani8art: **deployment with secret** ca.crt should only appear if the user add it to the secret
```
$ kubectl exec -ti hb-harbor-core-66ccf8c769-97rv4 ls /etc/core/ca
tls.crt  tls.key
```
**deployment without secret** ca.crt should appear always because we are generating it and it is self-signed
```
$k exec -ti hb-auto-harbor-core-7cccbc6c57-wlc6m ls /etc/core/ca
ca.crt  tls.crt  tls.key
```

It seems to be working as expected.

PR #1975, opened 2020-02-25
[bitnami/prometheus-operator] add extra args for thanos sidecar

ni-slamberti:
**Description of the change**
Adds the possibility to define some additional arguments for the thanos sidecar container in values.yaml.

**Benefits**
Allows us to use extra arguments for debug, tsdb path etc.

**Possible drawbacks**

**Applicable issues**

  - fixes #

**Additional information**

**Checklist** 


bitnami/prometheus-operator/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: 0.36.0
 description: The Prometheus Operator for Kubernetes provides easy monitoring definitions for Kubernetes services and deployment and management of Prometheus instances.
 name: prometheus-operator
-version: 0.12.6
+version: 0.12.7
```
miguelaeh: As you are adding a new feature this should be a minor change:
```diff
-version: 0.12.6
+version: 0.13.0
```

bitnami/prometheus-operator/values.yaml
```diff
@@ -548,6 +548,8 @@ prometheus:
       ##
       # pullSecrets:
       #   - myRegistryKeySecretName
+    
+    extraArgs: []
```
miguelaeh: It would be nice if you can add a comment here and an example if possible, it is really helpful for the rest of the users.

bitnami/prometheus-operator/Chart.yaml
ni-slamberti: ok, bumped to 0.13.0

bitnami/prometheus-operator/values.yaml
ni-slamberti: I added some comments and some example cases.

ni-slamberti:
@miguelaeh 
should be ok now, let me know if there's anything else I can improve.

PR #1993, opened 2020-03-02
enable_user_defined_functions

zhaxelll:
after update image https://github.com/bitnami/bitnami-docker-cassandra/pull/63
we want use this.
######
we need to create functions, but we do not want to use alternative cassandra.yaml


javsalgar:
Hi,

Thanks for the PR. Could you please update the chart minor version? It is adding a non-breaking feature.

zhaxelll:
Hi. 
I was updated Chart.yaml

bitnami/cassandra/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: cassandra
-version: 5.0.3
+version: 5.0.4
```
carrodher:
```suggestion
version: 5.1.0
```

zhaxelll: done

carrodher:
Thanks!

PR #2005, opened 2020-03-03
[bitnami/zookeeper] Added network policy

aksenk:
**Description of the change**

Added networkpolicy.yaml template, which is activated only when the parameter .Values.networkpolicy exists.

**Benefits**

It make possible to activate and configure network policy (ingress and egress) for zookeeper statefulset.
Configurable with bare network policy blocks for maximum flexibility.

**Possible drawbacks**

**Applicable issues**

**Additional information**

Example of configuration:
```
networkpolicy:
  ingress:
  - from:
    - podSelector:
        matchLabels:
          app: your-app
    ports:
    - protocol: TCP
      port: 2181
  egress: 
    - {}
```
Tested in our infrastructure.

**Checklist** 

### DIFF
--- a/bitnami/zookeeper/values-production.yaml
+++ b/bitnami/zookeeper/values-production.yaml
@@ -233,6 +233,21 @@ readinessProbe:
   failureThreshold: 6
   successThreshold: 1
 
+## Network policies
+## Ref: https://kubernetes.io/docs/concepts/services-networking/network-policies/
+##
+networkPolicy:
+  ## Specifies whether a NetworkPolicy should be created
+  ##
+  enabled: true
+
+  ## The Policy model to apply. When set to false, only pods with the correct
+  ## client label will have network access to the port Redis is listening
+  ## on. When true, zookeeper accept connections from any source
+  ## (with the correct destination port).
+  ##
+  allowExternal: true
+
 ## Prometheus Exporter / Metrics
 ##
 metrics:
--- a/bitnami/zookeeper/Chart.yaml
+++ b/bitnami/zookeeper/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: zookeeper
-version: 5.5.1
+version: 5.6.0
 appVersion: 3.5.7
 description: A centralized service for maintaining configuration information, naming, providing distributed synchronization, and providing group services for distributed applications.
 keywords:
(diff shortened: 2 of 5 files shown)


aksenk:
Got it, thanks. I'll change it to your format.

But why we **allow** clients with desired labels access to our app only if allowExternal is **false**?
In your template it looks like:
```
      {{- if not .Values.networkPolicy.allowExternal }}
        - podSelector:
            matchLabels:
              {{ include "cassandra.fullname" . }}-client: "true"
      {{- end }}
```
https://github.com/bitnami/charts/blob/master/bitnami/cassandra/templates/networkpolicy.yaml#L17

It should not be inverted? Like ` {{- if .Values.networkPolicy.allowExternal }}`

I hope I was able to present my thoughts correctly

aksenk:
https://github.com/bitnami/charts/blob/master/bitnami/cassandra/values.yaml#L271
  ## The Policy model to apply. When set to false, only pods with the correct
  ## client label will have network access to the port Redis (shoud be cassandra) is listening
  ## on. **When true, Redis (shoud be cassandra) will accept connections from any source**
  ## (**with the correct destination port**).

For correct behavior, I think it should be:
```
      {{- if .Values.networkPolicy.allowExternal }}
        - podSelector:
            matchLabels: {}
      {{- else }}
        - podSelector:
            matchLabels:
              {{ include "zookeeper.fullname" . }}-client: "true"
        - podSelector:
            matchLabels: {{- include "zookeeper.matchLabels" . | nindent 14 }}
      {{- end }}

```
Otherwise it just blocks **all** clients.

aksenk:
Hello. Please check my last edits.
It was done in your format, but with some changes:
1) if allowExternal is true it accept connections from all pods to port 2181 (in the cassandra template, this option blocks **all** clients)
2) if allowExternal is false it accept connections only from pods with desired label.

bitnami/zookeeper/values-production.yaml
```diff
@@ -233,6 +233,21 @@ readinessProbe:
...
+  ## Specifies whether a NetworkPolicy should be created
+  ##
+  enabled: true
+
+  ## The Policy model to apply. When set to false, only pods with the correct
+  ## client label will have network access to the port Redis is listening
+  ## on. When true, zookeeper accept connections from any source
+  ## (with the correct destination port).
+  ##
+  # allowExternal: true
```
juan131: Could you uncomment it and set it to "true" by default? To be consistent with what you're saying in the README.md

| Parameter  | Description    | Default  |
|------------|---------------|----------|
| `networkPolicy.allowExternal`  | Don't require client label for connections  | `true` |

aksenk: OK, no problem.
Done.

aksenk:
Done.

PR #2022, opened 2020-03-10
[stable/redis] Allow custom probes for other images

kremers:
Signed-off-by: Martin Kremers <info@martinkremers.de>


javsalgar:
Hi,

Thanks for the PR! Would moving the readiness probe to the values.yaml make more sense in this case?

cc @dani8art for possible standardization

kremers:
I had the same idea and created a previous version. But this will break compatibility since:
- Working with toYaml to allow the definition won't work since livenessProbe.enabled: true will disturb k8s 😢 
- Removing livenessProbe.enabled will be a breaking change for all previous users of the chart that disabled probes
- If we provide a default for the livenessProbe in values.yaml it might be merged with the user configuration, leading to duplicate probe definitions, so the pods cannot be scheduled

So i've gone along this way for users that use another image than the bitnami one. If you like I can show the other (breaking change) changeset as well. But I strongly recommend going along with this PR.

javsalgar:
Hi,

How about using functions like this?

```
(toYaml (unset (deepCopy .Values.master.readinessProbe) "enabled"))
```

kremers:
That does not resolve the problem, that the defaults getting merged with the user settings, resulting in double probes. In example if no cmd Probe is used but e.g. tcpsocket

https://docs.openshift.com/enterprise/3.0/dev_guide/application_health.html

Result: two probes - container won’t start 

Leaving the probe out completely from values.yaml will lead to weak defaults which is bad as well

javsalgar:
I think I didn't understand correctly. What I was proposing was not using the customLivenessProbe section but move the whole default liveness probles to the values.yaml, so the user can modify it. And, because you were saying that the enabled value would disturb k8s, I was proposing using that function that would remove the enabled value while leaving the rest.

kremers:
Not only the enabled bool will disturb k8s. You can not define two probe types "cmd" and "tcpSocket". 
If the chart is used as intended, there is no need to download it. Just 

`helm install relasenamehere bitnami/redis -f my_custom_values.yaml`

If chart values.yaml is already providing "cmd", defining "tcpSocket" in my_custom_values.yaml will lead to **cmd AND tcpSocket defined** causing an invalid manifest.

This is why I made the probe completely swappable by custom*Probe

javsalgar:
Hi,

What I'm proposing is something like this

values.yaml

```
        livenessProbe:
          enabled: true
          initialDelaySeconds: 5
          periodSeconds: 5
          timeoutSeconds: 5
          successThreshold: 5
          failureThreshold: 5
          exec:
            command:
            - "sh"
            - "-c"
            - "/health/ping_liveness_local.sh {{ .Values.master.livenessProbe.timeoutSeconds }}"

```

template

```
        {{- if .Values.master.livenessProbe.enabled }}
        livenessProbe:
{{- include "redis.tplValue" ( dict "value" (unset (deepCopy .Values.master.readinessProbe) "enabled") "context" $) }}
 ) | nindent 10 }}
        {{- end }}
```

helpers (this is something that we have in other charts)

```
{{/*
Usage:
{{ include "redis.tplValue" ( dict "value" .Values.path.to.the.Value "context" $) }}
*/}}
{{- define "redis.tplValue" -}}
        {{- tpl .value .context }}
    {{- else }}
    {{- end }}
{{- end -}}

```

This would make it compatible with the previous chart and it would not require an extra value, right?

kremers:
This does not work:


chart values.yaml
```
livenessProbe:
  enabled: true
  initialDelaySeconds: 5
  periodSeconds: 5
  timeoutSeconds: 5
  successThreshold: 5
  failureThreshold: 5
  exec:
    command:
    - "sh"
    - "-c"
    - "/health/ping_liveness_local.sh {{ .Values.master.livenessProbe.timeoutSeconds }}"
```

and 
custom-values.yml

```
livenessProbe:
  tcpSocket:
    port: 8080
  initialDelaySeconds: 15
  timeoutSeconds: 1
```

will produce:

```
livenessProbe:
  initialDelaySeconds: 15
  periodSeconds: 5
  timeoutSeconds: 1
  successThreshold: 5
  failureThreshold: 5
  tcpSocket:
    port: 8080
  exec:
    command:
    - "sh"
    - "-c"
    - "/health/ping_liveness_local.sh {{ .Values.master.livenessProbe.timeoutSeconds }}"
```

Which results in error.

It won't make it compatible, since 
tcpSocket:
      port: 8080
is yaml as well - dicts getting merged.

javsalgar:
Hi, 

Ok, now I understand it. Yes, you are right, we will need to think on a way to properly define this. Maybe the approach you suggest makes more sense then.

javsalgar:
cc @juan131 @dani8art any concerns on creating a customLivenessProbe section together with the livenessProbe section?

juan131:
No concerns @javsalgar 
I think both readiness/liveness probes for the "main container", as well as of other probes for "extra/sidecar" containers should be treated as a template. We can provide a "default" value in the **values.yaml** but they should be as customizable as possible.

kremers:
@juan131 same opinion. Holding on to the principle of the least surprise when moving templates, that will move "enabled: true" from liveness probe out of the templated key. (Just a Breaking change for people that disabled probes.) That will enable a maximum of customisation and not having to wonder about attributes that don't make it into the rendered config.

Having the custom*Probes is IMHO a workaround found in many charts for the problem to unset a key/structure in a nested struct. See: https://github.com/helm/helm/issues/5184

If this issue is resolved in the latest stable helm 3 release, we can suggest the user to unset the "exec" by "exec: null".

javsalgar:
Thank you, I will proceed to launch our automated job to update the tags

javsalgar:
Sorry, could you also bump the chart version?

PR #2028, opened 2020-03-11
[bitnami/kafka] Achieve consistent broker ids during cluster restart/rebuild

ghShu:
It is observed that current Kafka Chart leads to increasing broker ids after each restart/rebuild of
cluster when Persistent Volumes (PVs) are not deleted during restart/rebuild. This is
often the case since PVs are usually used to persist the log messages.

This commit provides a way to have deterministic broker ids after each cluster
restart/rebuild. Please refer to [this note](https://github.com/ghShu/kafka-notes/blob/master/notes/20200310-achieve-consistent-broker-ids-in-kafka-cluster.md) for detailed about the issue and explanation on the solution.

This solution assumes a deployment that uses this [Docker image](https://hub.docker.com/layers/bitnami/kafka/2.4.0-debian-10-r44/images/sha256-ce5ed55be807573b5902c5995982b346a01fe9f2cb77d2d2efe26d0043d83f46?context=explore) published in Dockerhub. Alternative solutions may exist with full access to Docker images. But this consistent broker ids behavior is believed to be helpful especially when PVs are used. This can help reduce the surprise in normal operation.


ghShu:
Thank you @miguelaeh, Kafka Chart version is bumped to 7.3.5.

yanjunli76:
Hi there,
There are two `command` if external access is enabled.

https://github.com/bitnami/charts/blob/5622f3120a2f5f8ed7bef73ced0c47e7133befc1/bitnami/kafka/templates/statefulset.yaml#L62-L65

https://github.com/bitnami/charts/blob/5622f3120a2f5f8ed7bef73ced0c47e7133befc1/bitnami/kafka/templates/statefulset.yaml#L207-L210

miguelaeh:
Hi @yanjunli76 ,
Thank you very much for reporting this, you are right. I missed that part.
Fortunately, there is already a PR fixing it: https://github.com/bitnami/charts/pull/2098

PR #2044, opened 2020-03-12
[bitnami/ghost] chart standardization

dani8art:
Signed-off-by: darteaga <darteaga@bitnami.com>

**Description of the change**

- Adding common things to `bitnami/ghost`

**Benefits**

Having clear charts and standardized for common things like affinity, sidecars, images, etc. 

**Possible drawbacks**

N/A

**Applicable issues**

N/A

**Additional information**

N/A

**Checklist** 

### DIFF
--- a/bitnami/ghost/README.md
+++ b/bitnami/ghost/README.md
@@ -210,6 +210,19 @@ See the [Parameters](#parameters) section to configure the PVC or to disable per
 
 ## Upgrading
 
+### To 10.0.0
+
+This version introduces `bitnami/common`, a [library chart](https://helm.sh/docs/topics/library_charts/#helm) as a dependency. More documentation about this new utility could be found [here](https://github.com/bitnami/charts/tree/master/bitnami/common#bitnami-common-library-chart). Please, make sure that you have updated the chart dependencies before executing any upgrade.
+
+Also, `allowEmptyPassword` has changed its value type from string to boolean, if you were using it please make sure that you are pasing the proper value.
+
+```console
+# before
+--set allowEmptyPassword="no"
+# after
+--set allowEmptyPassword=false
+```
+
 ### To 9.0.0
 
 Helm performs a lookup for the object based on its group (apps), version (v1), and kind (Deployment). Also known as its GroupVersionKind, or GVK. Changing the GVK is considered a compatibility breaker from Kubernetes' point of view, so you cannot "upgrade" those objects to the new GVK in-place. Earlier versions of Helm 3 did not perform the lookup correctly which has since been fixed to match the spec.
--- a/bitnami/ghost/README.md
+++ b/bitnami/ghost/README.md
@@ -50,88 +50,102 @@ The command removes all the Kubernetes components associated with the chart and 
 
 The following table lists the configurable parameters of the Ghost chart and their default values.
 
-| Parameter                           | Description                                                   | Default                                                  |
-|-------------------------------------|---------------------------------------------------------------|----------------------------------------------------------|
-| `global.imageRegistry`              | Global Docker image registry                                  | `nil`                                                    |
-| `global.imagePullSecrets`           | Global Docker registry secret names as an array               | `[]` (does not add image pull secrets to deployed pods)  |
-| `global.storageClass`               | Global storage class for dynamic provisioning                 | `nil`                                                    |
-| `image.registry`                    | Ghost image registry                                          | `docker.io`                                              |
-| `image.repository`                  | Ghost Image name                                              | `bitnami/ghost`                                          |
-| `image.tag`                         | Ghost Image tag                                               | `{TAG_NAME}`                                             |
-| `image.pullPolicy`                  | Image pull policy                                             | `IfNotPresent`                                           |
-| `image.pullSecrets`                 | Specify docker-registry secret names as an array              | `[]` (does not add image pull secrets to deployed pods)  |
-| `nameOverride`                      | String to partially override ghost.fullname template with a string (will prepend the release name) | `nil`               |
-| `fullnameOverride`                  | String to fully override ghost.fullname template with a string                                     | `nil`               |
-| `volumePermissions.image.registry`  | Init container volume-permissions image registry              | `docker.io`                                              |
-| `volumePermissions.image.repository`| Init container volume-permissions image name                  | `bitnami/minideb`                                        |
-| `volumePermissions.image.tag`       | Init container volume-permissions image tag                   | `buster`                                                 |
-| `volumePermissions.image.pullPolicy`| Init container volume-permissions image pull policy           | `Always`                                                 |
-| `ghostHost`                         | Ghost host to create application URLs                         | `nil`                                                    |
-| `ghostPort`                         | Ghost port to use in application URLs (defaults to `service.port` if `nil`) | `nil`                                      |
-| `ghostProtocol`                     | Protocol (http or https) to use in the application URLs       | `http`                                                   |
-| `ghostPath`                         | Ghost path to create application URLs                         | `nil`                                                    |
-| `ghostUsername`                     | User of the application                                       | `user@example.com`                                       |
-| `ghostPassword`                     | Application password                                          | Randomly generated                                       |
-| `ghostEmail`                        | Admin email                                                   | `user@example.com`                                       |
-| `ghostBlogTitle`                    | Ghost Blog name                                               | `User's Blog`                                            |
-| `smtpHost`                          | SMTP host                                                     | `nil`                                                    |
-| `smtpPort`                          | SMTP port                                                     | `nil`                                                    |
-| `smtpUser`                          | SMTP user                                                     | `nil`                                                    |
-| `smtpPassword`                      | SMTP password                                                 | `nil`                                                    |
-| `smtpFromAddress`                   | SMTP from address                                             | `nil`                                                    |
-| `smtpService`                       | SMTP service                                                  | `nil`                                                    |
-| `allowEmptyPassword`                | Allow DB blank passwords                                      | `yes`                                                    |
-| `livenessProbe.enabled`             | Would you like a livenessProbe to be enabled                  | `true`                                                   |
-| `livenessProbe.initialDelaySeconds` | Delay before liveness probe is initiated                      | 120                                                      |
-| `livenessProbe.periodSeconds`       | How often to perform the probe                                | 3                                                        |
-| `livenessProbe.timeoutSeconds`      | When the probe times out                                      | 5                                                        |
-| `livenessProbe.failureThreshold`    | Minimum consecutive failures to be considered failed          | 6                                                        |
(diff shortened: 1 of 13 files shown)


bitnami/ghost/templates/_helpers.tpl
```diff
@@ -190,3 +190,16 @@ Return the appropriate apiVersion for deployment.
...
+Renders a value that contains template.
+Usage:
+{{ include "ghost.tplValue" ( dict "value" .Values.path.to.the.Value "context" $) }}
+*/}}
+{{- define "ghost.tplValue" -}}
+    {{- if typeIs "string" .value }}
+        {{- tpl .value .context }}
+    {{- else }}
+        {{- tpl (.value | toYaml) .context }}
+    {{- end }}
```
javsalgar: This shouldn't be necessary anymore right?

dani8art: right! I will remove it

bitnami/ghost/templates/deployment.yaml
```diff
@@ -1,197 +1,189 @@
...
-        - mountPath: {{ .Values.persistence.path }}
-          name: ghost-data
+        - name: volume-permissions
+          image: {{ include "ghost.volumePermissions.image" . }}
+          imagePullPolicy: "{{ .Values.volumePermissions.image.pullPolicy }}"
+          command: ['sh', '-c', 'chmod -R g+rwX {{ .Values.persistence.path }}']
+          volumeMounts:
+            - mountPath: {{ .Values.persistence.path }}
+              name: ghost-data
+      {{- if .Values.initContainers }}
```
javsalgar, line 35: I've thinking a lot about this recently. How problematic would it be to put initContainers outside the if so you don't have to repeat the .Values.initContainers section twice? The worst that can happen is to get an empty initContainers yaml section, which I don't think is an issue (not in our linters) and the template code would be simpler. Thoughts?

bitnami/ghost/templates/deployment.yaml
```diff
@@ -1,197 +1,189 @@
...
-          httpGet:
-            path: {{ .Values.ghostPath }}
-            port: http
-            httpHeaders:
-            - name: Host
-              value: {{ include "ghost.host" . | quote }}
-            {{- if eq .Values.ghostProtocol "https" }}
-            - name: X-Forwarded-Proto
-              value: https
+        - name: {{ include "ghost.fullname" . }}
```
javsalgar, line 40: I'm not in favour of having a container with the name of the release. It doesn't make much sense to me and it is more problematic in terms of debugging. In my opinion, the container name should be `ghost`

javsalgar:
```suggestion
        - name: ghost
```

bitnami/ghost/templates/deployment.yaml
```diff
@@ -1,197 +1,189 @@
...
+          image: {{ include "ghost.image" . }}
+          imagePullPolicy: {{ .Values.image.pullPolicy | quote }}
+          env:
+            - name: ALLOW_EMPTY_PASSWORD
+              value: {{ ternary "yes" "no" .Values.allowEmptyPassword | quote }}
+            - name: MARIADB_HOST
+              value: {{ ternary (include "ghost.mariadb.fullname" .) .Values.externalDatabase.host .Values.mariadb.enabled | quote }}
+            - name: MARIADB_PORT_NUMBER
+              value: {{ ternary "3306" .Values.externalDatabase.port .Values.mariadb.enabled | quote }}
```
javsalgar, line 49: Is the mariadb sub-chart service port configurable? If so, you cannot assume that the sub-chart would use 3306.

javsalgar: (very weird case, to be honest)

bitnami/ghost/templates/deployment.yaml
```diff
@@ -1,197 +1,189 @@
...
+            - name: GHOST_DATABASE_USER
+              value: {{ ternary .Values.mariadb.db.user .Values.externalDatabase.user .Values.mariadb.enabled | quote }}
+            - name: GHOST_DATABASE_PASSWORD
+            {{- if .Values.mariadb.enabled }}
+              valueFrom:
+                secretKeyRef:
+                  name: {{ include "ghost.mariadb.fullname" . }}
+                  key: mariadb-password
+            {{- else }}
+              value: {{ .Values.externalDatabase.password | quote }}
```
javsalgar, line 61: Didn't see it was set like that. We wouldn't want to have plain text passwords in the yaml. In other charts what we do is to create a secret with the external db data. With that, you can use a helper that would return the correct secret name. I think we do it in Kong

bitnami/ghost/templates/ingress.yaml
```diff
@@ -2,42 +2,38 @@
...
-  labels:
-    app: "{{ template "ghost.fullname" . }}"
-    chart: "{{ template "ghost.chart" . }}"
-    release: {{ .Release.Name | quote }}
-    heritage: {{ .Release.Service | quote }}
-  annotations:
-    {{- if .Values.ingress.certManager }}
+  name: {{ include "ghost.fullname" . }}
+  labels: {{- include "common.labels.standard" . | nindent 4 }}
+  annotations:
```
javsalgar, line 7: Trailing space. There's an extension in vscode to automatically remove all trailing spaces when saving.

bitnami/ghost/templates/svc.yaml
```diff
@@ -27,5 +21,7 @@ spec:
       {{- if (and (eq .Values.service.type "NodePort") (not (empty .Values.service.nodePorts.http)))}}
       nodePort: {{ .Values.service.nodePorts.http }}
```
javsalgar, line 22: It is important to set the nodePort to null in case the service is of type ClusterIP. There were support cases that complained that helm upgrade would fail if you changed a deployment from LoadBalancer to ClusterIP. Check the WordPress chart svc.yaml for an example.

bitnami/ghost/templates/secrets.yaml
```diff
@@ -1,19 +1,15 @@
 apiVersion: v1
```
javsalgar, line 1: If .Values.existingSecret is set this should not be created, right? 

Well, I see an issue here because that would force the user to have both the ghostPassword and smtpPassword in the same secret.

bitnami/ghost/README.md
```diff
@@ -50,88 +50,98 @@ The command removes all the Kubernetes components associated with the chart and
...
+| `externalDatabase.password`          | Password for the above username                                                                    | `""`                                                                                          |
+| `externalDatabase.database`          | Name of the existing database                                                                      | `bitnami_ghost`                                                                               |
+| `mariadb.enabled`                    | Whether or not to install MariaDB (disable if using external)                                      | `true`                                                                                        |
+| `mariadb.rootUser.password`          | MariaDB admin password                                                                             | `nil`                                                                                         |
+| `mariadb.db.name`                    | MariaDB Database name to create                                                                    | `bitnami_ghost`                                                                               |
```
javsalgar, line 125: The following values are set in the values.yaml file but not explained here:

- mariadb.replication.enabled

- mariadb.master.persistence.enabled

- mariadb.master.persistence.accessMode

- mariadb.master.persistence.size

bitnami/ghost/templates/deployment.yaml
dani8art: yes, I think it could work fine and it would be the cleanest way, I did a try and an empty `initContainer` seems to not be a problem.

```yaml
    spec:
      {{- include "ghost.imagePullSecrets" . | indent 6 }}
      {{- if .Values.securityContext.enabled }}
      securityContext:
        fsGroup: {{ .Values.securityContext.fsGroup }}
      {{- end }}
      initContainers:
      {{- if not .Values.securityContext.enabled }}
        - name: volume-permissions
          image: {{ include "ghost.volumePermissions.image" . }}
          imagePullPolicy: "{{ .Values.volumePermissions.image.pullPolicy }}"
          command: ['sh', '-c', 'chmod -R g+rwX {{ .Values.persistence.path }}']
          volumeMounts:
            - mountPath: {{ .Values.persistence.path }}
              name: ghost-data
      {{- end }}
        {{- include "common.tplvalues.render" ( dict "value" .Values.initContainers "context" $ ) | indent 8 }}
      {{- end }}
```

```bash
$ echo "$(helm template g   --set service.type=ClusterIP   --set securityContext.enabled=true   -s templates/deployment.yaml   bitnami/ghost)" | kubeval -
# PASS - ghost/templates/deployment.yaml contains a valid Deployment
```

Thanks!

bitnami/ghost/templates/deployment.yaml
dani8art: You are right, it was a mistake. ;)

bitnami/ghost/templates/deployment.yaml
dani8art: Yes you are right, also I will fix a bug that I found with `ternary`.

```yaml
            {{- $host := .Values.externalDatabase.host -}}
            {{- $port := .Values.externalDatabase.port -}}
            {{- $database := .Values.externalDatabase.database -}}
            {{- $user := .Values.externalDatabase.user -}}

            {{- if .Values.mariadb.enabled -}}
              {{- $host = (include "ghost.mariadb.fullname" .) -}}
              {{- $port = .Values.mariadb.service.port -}}
              {{- $database = .Values.mariadb.db.name -}}
              {{- $user = .Values.mariadb.db.user -}}
            {{- end }}
            - name: MARIADB_HOST
              value: {{ $host | quote }}
            - name: MARIADB_PORT_NUMBER
              value: {{ $port | quote }}
            - name: GHOST_DATABASE_NAME
              value: {{ $database | quote }}
            - name: GHOST_DATABASE_USER
              value: {{ $user | quote }}
```

bitnami/ghost/templates/deployment.yaml
dani8art: yes, you are completely right!

bitnami/ghost/templates/ingress.yaml
dani8art: done

bitnami/ghost/templates/svc.yaml
dani8art: done

bitnami/ghost/templates/secrets.yaml
dani8art: > If .Values.existingSecret is set this should not be created, right?
I agree, and it is implementing.

Regarding, Let the user have different secrets per each secret implies to have the same logic for all secret data also in the deployment and to have them separately in values too, right? I don't know if it is a common use case and how we are handling it in other charts.

bitnami/ghost/README.md
dani8art: done

PR #2055, opened 2020-03-15
fix: don't request resources for dev environments

polRk:
I have about 30 releases of wordress and the default value not for production, forces kubernetes, with auto-scaling enabled, to create 10 additional nodes

### DIFF
--- a/bitnami/wordpress/Chart.yaml
+++ b/bitnami/wordpress/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: wordpress
-version: 9.0.3
+version: 9.0.4
 appVersion: 5.3.2
 description: Web publishing platform for building blogs and websites.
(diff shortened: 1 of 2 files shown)


polRk:
> Also, I didn't catch why is the resources section forcing to create 10 nodes, could you clarify that?

memory: 512Mi - It is a half of available memory in one pod. I use short-lived feature branches. 
cpu: 300m - One my deployment cost 9m - 30m. 

I fixed it by setting  ```resources: null```

miguelaeh:
Hi @polRk ,
I don't think we should merge these changes. As you can see here: https://kubernetes.io/docs/tasks/configure-pod-container/quality-service-pod/
The resources are necessary for the QoS. It could be fine for your use case to let them empty but it could not be the best option for teh rest of users.

polRk:
If someone would to remove resources value, you should use 
```
resources: null
```

PR #2059, opened 2020-03-16
[bitnami/postgresql] Evaluate extraEnvVarsCM as a template

evili:
Signed-off-by: Evili del Rio <evili.del.rio@gmail.com>

**Description of the change**

Use `tpl` to expand the value of `extraEnvVarsCM`.

**Benefits**

The config map name can vary from release to release when used as a subchart.

**Possible drawbacks**

None known.

**Applicable issues**

**Checklist** 


andresbono:
Hi @evili, thanks for submitting this PR! In principle, it looks good to me.

> **Benefits**
> The config map name can vary from release to release when used as a subchart.

Can you put an example to illustrate the benefits?

Thanks again!

evili:
Hi @andresbono 

I use this chart as a subchart and I need to share configuration between the parent (web app) and child. In my `values.yaml` I want to be able to put this:

```
postgresql:
  enabled: true
  extraEnvVarsCM: "{{ .Release.Name }}-postgres-config"
```

And then in one of the parent chart `templates/`

```
          envFrom:
            - configMapRef:
                name: {{ .Release.Name }}-postgres-config
```

So that, _by default_, the postgresql config map name can vary between releases of the parent chart _without_ user intervention (sensible default).

This is the same technique used for `persistence.existingClaim`.

andresbono:
Thanks for the explanation, @evili!

PR #2088, opened 2020-03-19
[bitnami/external-dns] reuse existing service account 

cw-sakamoto:
Signed-off-by: cw-sakamoto <sakamoto@chatwork.com>

**Description of the change**

-  This PR is able to reuse existing service account in external-dns chart
   - FYI: https://github.com/helm/charts/pull/21204

**Benefits**

- Easier to use the serviceaccount created by `eksctl` 

**Possible drawbacks**

**Applicable issues**

  - fixes #

**Additional information**

**Checklist** 


bitnami/external-dns/README.md
```diff
@@ -170,7 +170,8 @@ The following table lists the configurable parameters of the external-dns chart
...
 | `service.annotations`               | Annotations to add to service                                                                            | `{}`                                                        |
 | `rbac.create`                       | Weather to create & use RBAC resources or not                                                            | `true`                                                      |
-| `rbac.serviceAccountName`           | ServiceAccount (ignored if rbac.create == true)                                                          | `default`                                                   |
+| `rbac.serviceAccount.create`        | If true and rbac.create is also true, a service account will be created                                  | `true`                                                      |
+| `rbac.serviceAccount.name`          | existing ServiceAccount to use (ignored if rbac.create=true and rbac.serviceAccount.create=true)         | `default`                                                   |
```
juan131: Hi @cw-sakamoto 

Please note that you're modifying the values structure. Therefore, that would break backwards compatibility unless people adapt their values.yaml. We usually request to bump the major version for every change of these characteristics.

On the other hand, I don't think it's necessary to break this, you can use a different approach:

| Parameter | Description  | Default   |
| -----------| ------------| --------- |
| `rbac. serviceAccountName ` | ServiceAccount to use. A name is generated using the `external-dns.fullname` template if it is not set and `rbac.create` is `true` | `nil` |

Then, adapt the PR to ensure the approach previously defined is true.

cw-sakamoto: @juan131 Thanks for your comments!
As you say, maintaining backward compatibility is important, so I fixed it.
https://github.com/bitnami/charts/pull/2088/commits/0ec36b5b5c7f826f5cc5ee1be9df9e60d32e0805

I think that the charts ​​will work maintaining backward compatible, how about?

bitnami/external-dns/README.md
```diff
@@ -170,7 +170,8 @@ The following table lists the configurable parameters of the external-dns chart
...
 | `service.annotations`               | Annotations to add to service                                                                            | `{}`                                                        |
 | `rbac.create`                       | Weather to create & use RBAC resources or not                                                            | `true`                                                      |
-| `rbac.serviceAccountName`           | ServiceAccount (ignored if rbac.create == true)                                                          | `default`                                                   |
+| `rbac.serviceAccount.create`        | If true and rbac.create is also true, a service account will be created                                  | `true`                                                      |
+| `rbac.serviceAccountNamne`          | existing ServiceAccount to use (ignored if rbac.create=true and rbac.serviceAccount.create=true)         | `default`                                                   |
```
juan131: Typo: `serviceAccountNamne` -> `serviceAccountName`

bitnami/external-dns/README.md
```diff
@@ -170,7 +170,8 @@ The following table lists the configurable parameters of the external-dns chart
 | `service.loadBalancerSourceRanges`  | List of IP CIDRs allowed access to load balancer (if supported)                                          | `[]`                                                        |
 | `service.annotations`               | Annotations to add to service                                                                            | `{}`                                                        |
 | `rbac.create`                       | Weather to create & use RBAC resources or not                                                            | `true`                                                      |
-| `rbac.serviceAccountName`           | ServiceAccount (ignored if rbac.create == true)                                                          | `default`                                                   |
+| `rbac.serviceAccount.create`        | If true and rbac.create is also true, a service account will be created                                  | `true`                                                      |
```
juan131: I still think this parameter is useless, I mean we can obtain the same behaviour without it. See my other comments

bitnami/external-dns/templates/deployment.yaml
```diff
@@ -34,11 +34,7 @@ spec:
       {{- with .Values.tolerations }}
       tolerations: {{ tpl (toYaml .) $ | nindent 8 }}
       {{- end }}
-      {{- if .Values.rbac.create }}
-      serviceAccountName: {{ template "external-dns.fullname" . }}
-      {{- else }}
-      serviceAccountName: {{ .Values.rbac.serviceAccountName | quote }}
-      {{- end }}
+      serviceAccountName: {{ template "serviceaccount.name" . }}
```
juan131: `serviceAccountName` shouldn't be defined unless RBAC is enabled:
```diff
+     {{- if .Values.rbac.create }}
      serviceAccountName: {{ template "serviceaccount.name" . }}
+     {{- end }}
```

bitnami/external-dns/templates/_helpers.tpl
```diff
@@ -428,3 +428,14 @@ external-dns: transip.apiKey
...
+{{/*
+Create the name of the service account to use
+*/}}
+{{- define "serviceaccount.name" -}}
+{{- if and .Values.rbac.create .Values.rbac.serviceAccount.create -}}
+{{ include "external-dns.fullname" . }}
+{{- else -}}
+{{ .Values.rbac.serviceAccountName }}
+{{- end -}}
+{{- end -}}
```
juan131: Suggestion: you just need to check if `.Values.rbac.serviceAccountName` is defined. `.Values.rbac.serviceAccount.create` doesn't add any value.
```
{/*
Return the ExternalDNS service account name
*/}}
{{- define "external-dns.serviceAccountName" -}}
{{- if .Values.rbac.serviceAccountName -}}
    {{- printf "%s" (tpl .Values.rbac.serviceAccountName $) -}}
{{- else -}}
    {{- printf "%s" (include "external-dns.fullname" .) -}}
{{- end -}}
{{- end -}}
```

bitnami/external-dns/templates/serviceaccount.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if .Values.rbac.create }}
+{{- if and .Values.rbac.create .Values.rbac.serviceAccount.create -}}
```
juan131: Revert this change

bitnami/external-dns/values-production.yaml
```diff
@@ -411,6 +411,8 @@ rbac:
   ## Service Account for pods
   ## https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/
   ##
+  serviceAccount:
+    crete: true
   serviceAccountName: default
```
juan131: Suggestion: `serviceAccountName` should be undefined by default. This way, you can simply check in the rest of the templates whether it's defined or not:
```diff
-
```

bitnami/external-dns/values.yaml
```diff
@@ -422,6 +422,8 @@ rbac:
   ## Service Account for pods
   ## https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/
   ##
+  serviceAccount:
+    crete: true
```
juan131: Same comment i did at `values.yaml`

bitnami/external-dns/README.md
cw-sakamoto: Thank you for your polite review !

Sorry for not being able to communicate well in English...

`rbac.serviceAccount.create` parameter cannot be removed. 
The intent of this PR is to create status in which `clusterrole (role)` and `clusterrolebiding (rolebinding)` are created but `serviceAccount` is not created.

rolebinding is setting for the serviceAccount determined in https://github.com/bitnami/charts/pull/2088/files#diff-3d3105b5ff96a11099b64bae938ea60dR431-R441.

Because, when using iam role for sa (IRSA) in EKS, the simplest method is to create a serviceAccount that supports IRSA with `eksctl`. And by using serviceAccount created by eksctl, application can use iam role for sa.

FYI: https://aws.amazon.com/jp/blogs/opensource/introducing-fine-grained-iam-roles-service-accounts/

bitnami/external-dns/templates/serviceaccount.yaml
cw-sakamoto: Cannot revert for the same reason as https://github.com/bitnami/charts/pull/2088#discussion_r395482973

aaronj314:
@cw-sakamoto I submitted the same PR and please let me know if you need help with some of the questions there as we need this change ASAP

bitnami/external-dns/README.md
juan131: I see your point, but you can get the same without that parameter (check my comment at `_helpers.tpl`)
This is your current logic:
- `rbac.create=true` && `rbac.serviceAccount.create` -> Create SA using `{{ include "external-dns.fullname" . }}`
- `rbac.create=true` && `rbac.serviceAccount.create=false` -> Reuse existing SA (name taken from `rbac.serviceAccountName`)
This is what I suggest:
- `rbac.create=true` && `rbac.serviceAccountName is NOT defined` -> Create SA using `{{ include "external-dns.fullname" . }}`
- `rbac.create=true` && `rbac.serviceAccountName is defined` Reuse existing SA (name taken from `rbac.serviceAccountName`)
As you can see, the result is the same but you're just using one parameter: `rbac.serviceAccountName`

cw-sakamoto: what does it mean for https://github.com/bitnami/charts/pull/2088#discussion_r395470884 to be a revert?

If I revert this, the `rbac.serviceAccount.create` parameter will not work in `serviceaccount.yaml`, and the `rbac.create` parameter alone will determine whether the sa is created or not.

I'm sorry if I'm mistaken..

bitnami/external-dns/templates/serviceaccount.yaml
cw-sakamoto: Your suggest ?

```
{{- if and .Values.rbac.create .Values.rbac.serviceAccountName -}}
```

carrodher:
Hi, just to summarize, there are 3 PRs fixing the same issue:

- https://github.com/bitnami/charts/pull/2088 created on March 19 at 7:33 
- https://github.com/bitnami/charts/pull/2095 created by @aaronj314 on March 19 at 22:39
- https://github.com/bitnami/charts/pull/2104 created by @derrickburns on March 21

In order to avoid duplicating the effort in your side and our side as reviewers, we are going to close #2095 and #2104. it would be nice if everyone collaborates adding your thoughts to take into account the different use cases.

cw-sakamoto:
@juan131 
I've fixed the point you commented on, so please review it !
https://github.com/bitnami/charts/pull/2088/commits/5773c6027914250fba4844c1bbca43cfcd72a75b

bitnami/external-dns/values-production.yaml
```diff
@@ -411,6 +411,8 @@ rbac:
   ## Service Account for pods
   ## https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/
   ##
+  serviceAccount:
+    crete: true
```
juan131: Please remove this section

bitnami/external-dns/values.yaml
```diff
@@ -422,7 +422,8 @@ rbac:
   ## Service Account for pods
   ## https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/
   ##
-  serviceAccountName: default
+  serviceAccountName:
```
juan131: Add this change at `values-production.yaml` too

bitnami/external-dns/templates/serviceaccount.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if .Values.rbac.create }}
+{{- if and .Values.rbac.create (not .Values.rbac.serviceAccountName) -}}
```
juan131: Nice!

bitnami/external-dns/templates/deployment.yaml
```diff
@@ -34,11 +34,7 @@ spec:
       {{- with .Values.tolerations }}
       tolerations: {{ tpl (toYaml .) $ | nindent 8 }}
       {{- end }}
-      {{- if .Values.rbac.create }}
-      serviceAccountName: {{ template "external-dns.fullname" . }}
-      {{- else }}
-      serviceAccountName: {{ .Values.rbac.serviceAccountName | quote }}
-      {{- end }}
+      serviceAccountName: {{ template "external-dns.serviceAccountName" . }}
```
juan131: Do not remove this condition: `{{- if .Values.rbac.create }}`
```suggestion
      {{- if .Values.rbac.create }}
      serviceAccountName: {{ template "external-dns.serviceAccountName" . }}
      {{- end }}
```

cw-sakamoto:
@juan131 
Thank you for taking such a good look at all the details !
I fixed at https://github.com/bitnami/charts/pull/2088/commits/e9f11fad4604a238192781fee655e859d3da163b

PR #2095, opened 2020-03-19
[bitnami/external-dns] Resuse service account

aaronj314:
**Description of the change**

This PR allows to reuse an existing service accounts (for instance created with eksctl with its related IAM role) and still manage the role and cluster role bindings with the chart.

**Benefits**

An IAM role created by eksctl has a condition on the service account name it also creates, so we cannot just use the serviceAccountAnnotations value to do the mapping, unless we modify the role.
We are aware of the --override-existing-serviceaccounts eksctl argument will put the annotation on the service account if it already exists, and create the missing ones.
I'm interested in a workflow where the infra is created first (applying a YAML config file with eksctl specifying the cluster, including the IAM Roles for Service Accounts) and then the add-ons are deployed with Helm, including external-dns. Being able to specify all of that in config files in git and apply it to get a cluster fully configured is a challenge which today requires an extra step that could be eliminated with the help of this PR.

**Possible drawbacks**

None

**Applicable issues**

None

**Additional information**

I've tested the PR, with the default value it behaves as before. The new feature is enabled only when both rbac.create is true, and rbac.serviceAccount is either true or false.

Tested with Helm 3.1.1.

**Checklist** 

### DIFF
--- a/bitnami/external-dns/Chart.yaml
+++ b/bitnami/external-dns/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: external-dns
-version: 2.20.5
+version: 3.0.0
 appVersion: 0.7.0
 description: ExternalDNS is a Kubernetes addon that configures public DNS servers with information about exposed Kubernetes services to make them discoverable.
 keywords:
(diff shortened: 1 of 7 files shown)


juan131:
Hi @aaronj314 

Thanks so much for your PR! Closing it as it's a duplicate of https://github.com/bitnami/charts/pull/2088

PR #2099, opened 2020-03-20
[bitnami/mongodb] Issue 2096. Encode @ : in username / password when connect to mongo

lff0305:
If `@` or `:` in username / password, the metric exporter will not connect to mongodb, because
`mongo://username:password@localhost` will not be splitted correctly.

So here the `@` and `:` is replaced to `%40` and `%3A` (url-encode) by `sed` to fix the issue.


bitnami/mongodb/templates/statefulset-primary-rs.yaml
```diff
@@ -229,7 +229,7 @@ spec:
               secretKeyRef:
                 name: {{ if .Values.existingSecret }}{{ .Values.existingSecret }}{{- else }}{{ template "mongodb.fullname" . }}{{- end }}
                 key: mongodb-root-password
-          command: [ 'sh', '-c', '/bin/mongodb_exporter --mongodb.uri mongodb://root:${MONGODB_ROOT_PASSWORD}@localhost:{{ .Values.service.port }}/admin {{ .Values.metrics.extraArgs }}' ]
+          command: [ 'sh', '-c', '/bin/mongodb_exporter --mongodb.uri mongodb://root:`echo $MONGODB_ROOT_PASSWORD | sed -r "s/@/%40/g"`@localhost:{{ .Values.service.port }}/admin {{ .Values.metrics.extraArgs }}' ]
```
andresbono: This one is not making the substitutions for `:`

andresbono:
Please, bump the version of the chart in the Charts.yaml file.

bitnami/mongodb/templates/statefulset-primary-rs.yaml
lff0305: Missed this one. Will update.

andresbono:
Thanks for your contribution. We are now going to test the PR and merge it.

PR #2104, opened 2020-03-21
Make creation of service account optional.

### LINKED ISSUE #2056: Add ability to NOT generate service account for external DNS
When using Amazon IAM for Service Accounts, we create service accounts that need IAM identities outside of the helm chart.  We need the ability to NOT create the service account in the helm chart.

derrickburns:
Creating a service account for external DNS is not always necessary or desired if the service account already exists.

This is the case when using IAM for Service Accounts with external-dns on AWS EKS.  In that situation, the service account is created outside the helm chart with a token that provides an IAM role to the service account..  We need a way not to create a conflicting service account.

Fixes #2056

### DIFF
--- a/bitnami/external-dns/templates/deployment.yaml
+++ b/bitnami/external-dns/templates/deployment.yaml
@@ -37,7 +37,7 @@ spec:
       {{- if .Values.rbac.create }}
       serviceAccountName: {{ template "external-dns.fullname" . }}
       {{- else }}
-      serviceAccountName: {{ .Values.rbac.serviceAccountName | quote }}
+      serviceAccountName: {{ .Values.rbac.serviceAccount.name | quote }}
       {{- end }}
       {{- if .Values.priorityClassName }}
       priorityClassName: {{ .Values.priorityClassName | quote }}
--- a/bitnami/external-dns/templates/serviceaccount.yaml
+++ b/bitnami/external-dns/templates/serviceaccount.yaml
@@ -1,4 +1,4 @@
-{{- if .Values.rbac.create }}
+{{- if .Values.rbac.serviceAccount.create }}
 apiVersion: v1
 kind: ServiceAccount
 metadata:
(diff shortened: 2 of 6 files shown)


bitnami/external-dns/templates/serviceaccount.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if .Values.rbac.create }}
+{{- if .Values.rbac.serviceAccount.create }}
```
carrodher: You need to check both conditions, something like:
```suggestion
{{- if and .Values.rbac.create .Values.rbac.serviceAccount.create }}
```

bitnami/external-dns/values-production.yaml
```diff
@@ -411,6 +411,8 @@ rbac:
   ## Service Account for pods
   ## https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/
   ##
+  serviceAccount:
+    create: true
   serviceAccountName: default
```
carrodher: Now, maybe it has more sense to create the `name` parameter as part of the `serviceAccount` object:
```suggestion
  serviceAccount:
    create: true
    name: default
```

And then use the new `serviceAccount.name` in those places where `serviceAccountName` was used.

PS: Same is applied to the _values.yaml_ and _values-production.yaml_

bitnami/external-dns/templates/deployment.yaml
```diff
@@ -37,7 +37,7 @@ spec:
       {{- if .Values.rbac.create }}
```
carrodher: Thanks for your changes. I think we also need to check both parameters in this condition, what do you think?
```suggestion
      {{- if and .Values.rbac.create .Values.rbac.serviceAccount.create }}
```

carrodher:
And also, please, can you update the table that appears in the README? Adding the new parameter and changing the name to the `serviceAccountName` one

derrickburns:
Done

carrodher:
Hi, I am sorry but I just found that there are 3 PRs fixing the same issue:

- https://github.com/bitnami/charts/pull/2088 created by @cw-sakamotoon March 19 at 7:33 
- https://github.com/bitnami/charts/pull/2095 created by @aaronj314 on March 19 at 22:39
- https://github.com/bitnami/charts/pull/2104 created on March 21 (this one)

In order to avoid duplicating the effort in your side and our side as reviewers, we are going to close #2095 and #2104. it would be nice if everyone collaborates adding your thoughts to the open one (#2088) to take into account the different use cases.

PR #2127, opened 2020-03-24
[bitnami/airflow] Fix the wrong env var name for Airflow fernet key

boyanbonev:
**Description of the change**

There's an issue with the fernet key deployed with the current config. The vars which
are created throught the Airflow UI cannot be properly read by all workers because
they are using different fernet keys. The reason for this is the fact that the fernet
key env var name is not the right one. It is named AIRFLOW_FERNET_KEY and it must be
AIRFLOW__CORE__FERNET_KEY according to the Airflow documentation -
https://airflow.readthedocs.io/en/stable/howto/secure-connections.html

**Benefits**

It is possible to use Airflow variables. Previously it was not possible.

**Possible drawbacks**

No

**Applicable issues**

There's no open bug that I am aware of. This is a new issue that we discovered and fixed.

  - fixes #

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/airflow/templates/deployment-scheduler.yaml
+++ b/bitnami/airflow/templates/deployment-scheduler.yaml
@@ -161,7 +161,7 @@ spec:
                   key: redis-password
             - name: AIRFLOW_EXECUTOR
               value: "CeleryExecutor"
-            - name: AIRFLOW_FERNET_KEY
+            - name: AIRFLOW__CORE__FERNET_KEY
               valueFrom:
                 secretKeyRef:
                   name: {{ if .Values.airflow.auth.existingSecret }}{{ .Values.airflow.auth.existingSecret }}{{ else }}{{ template "airflow.fullname" . }}{{ end }}
--- a/bitnami/airflow/templates/deployment-web.yaml
+++ b/bitnami/airflow/templates/deployment-web.yaml
@@ -168,7 +168,7 @@ spec:
                 secretKeyRef:
                   name: {{ if .Values.airflow.auth.existingSecret }}{{ .Values.airflow.auth.existingSecret }}{{ else }}{{ template "airflow.fullname" . }}{{ end }}
                   key: airflow-password
-            - name: AIRFLOW_FERNET_KEY
+            - name: AIRFLOW__CORE__FERNET_KEY
               valueFrom:
                 secretKeyRef:
                   name: {{ if .Values.airflow.auth.existingSecret }}{{ .Values.airflow.auth.existingSecret }}{{ else }}{{ template "airflow.fullname" . }}{{ end }}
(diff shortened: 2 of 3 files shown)


dani8art:
Hi @boyanbonev thank you for your contribution, I would like to understand better your proposal or even the problems you are facing, `AIRFLOW_FERNET_KEY` should be working properly as it is a valid value for this configuration property, you can see it at https://github.com/bitnami/bitnami-docker-airflow#environment-variables

It is possible you have found other bug related to this `AIRFLOW_FERNET_KEY` and it is true that the best possible workaround is `AIRFLOW__CORE__FERNET_KEY`, however, we would like to understand what is happening when using `AIRFLOW_FERNET_KEY`. Did you do any other further investigation?

boyanbonev:
Hi Daniel,
You have defined a fernet key using a wrong environmental variable. If you
look at the link that I provided there's no such property named
*AIRFLOW_FERNET_KEY* supported by Airflow. The property to define the
shared fernet key should be named *AIRFLOW__CORE__FERNET_KEY - *
https://airflow.readthedocs.io/en/stable/howto/secure-connections.html

> Replace airflow.cfg fernet_key value with the one from Step 2.
> *Alternatively,* you can store your fernet_key in OS environment variable
> - You do not need to change airflow.cfg in this case as Airflow will use
> environment variable over the value in airflow.cfg:
>
> # Note the double underscoresexport AIRFLOW__CORE__FERNET_KEY=your_fernet_key
>
> Also, we've been using your deployment (I am from VMware) and noticed this
property *AIRFLOW_FERNET_KEY* is not taken into consideration. Once we
changed it to the one Airflow says it should be used everything was fixed
and the properties set by some of the Airflow workers were displayed in the
UI correctly.
Currently we don't have an automatic tests for this

Regards,
Boyan

На вт, 24.03.2020 г. в 12:42 Daniel Arteaga <notifications@github.com>
написа:

dani8art:
Yes, I agree with you @boyanbonev `AIRFLOW__CORE__FERNET_KEY` is the property which airflow requires to set `fernet_key`, however, `bitnami-docker-airflow` adds a new env var called `AIRFLOW_FERNET_KEY` which should do the same as `AIRFLOW__CORE__FERNET_KEY`. I mean, if `AIRFLOW_FERNET_KEY` is not doing the same as `AIRFLOW__CORE__FERNET_KEY` there is something wrong in `bitnami-docker-airflow` that we should fix instead of change it to `AIRFLOW__CORE__FERNET_KEY`.

boyanbonev:
Well I didn't see where AIRFLOW_FERNET_KEY is used in the repo, and since the AIRFLOW__CORE__FERNET_KEY fixed the issue at our end I thought maybe you have made a mistake with the env var name. I cannot see the code which uses it, and if you say that it is working as expected at your end them leave it as it is.

PR #2130, opened 2020-03-24
[bitnami/tomcat] Allow setting existing PVCs

javsalgar:
Signed-off-by: Javier J. Salmeron Garcia <jsalmeron@vmware.com>

**Description of the change**

Allow using an existing PVC so they can mount their custom .war files without having to create a forked image.

**Checklist** 


bitnami/tomcat/templates/pvc.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if .Values.persistence.enabled -}}
+{{- if and .Values.persistence.enabled (not .Values.persistence.existingClaim) -}}
 kind: PersistentVolumeClaim
 apiVersion: v1
 metadata:
```
carrodher: I can't comment in the next line because GitHub considers it out of the scope of this PR, but now the name of the PVC should be something like:
```
  name: {{ template "tomcat.pvc" . }}
```

javsalgar: No it shouldn't, `tomcat.pvc` is for deciding whether it should use the one created by the chart or an already existing one.

PR #2132, opened 2020-03-24
[external-dns] Fix for Cloudflare Api Key

emas80:
Fixed bug with cloudflare provider support for api key. 
The condition check inside of the deployment is wrong and it will always provide the env var "CF_API_TOKEN" to external-dns.

You still have to build a secret with both cloudflare_api_key and cloudflare_api_token, and you should specify always the email if you mean to use the secret even if you want to use the token, but in the current implementation I could not think at any better solution.

**Checklist** 
- [X ] Variables are documented in the README.md


emas80:
Hi @miguelaeh,

sorry I should have been more specific.

I am providing an external secret, and in my case I only have the api key (+ email). 

When providing an external secret, the check 
```
         {{- if or .Values.cloudflare.apiToken .Values.cloudflare.secretName }}
```
causes the deployment to ALWAYS be generated with the env variable `CF_API_TOKEN`, so for me it's impossible to have a deployment with `CF_API_KEY` and `CF_API_EMAIL` while providing an external secret.

Sure I can control what is inside of my secret, but how can we do that while generating the deployment? This is why the (bad) solution of the empty strings.

The only other solution I can think of would be to add more variables to specifically control how the deployment can be generated - something like 
```
 {{- if .Values.cloudflare.secretName AND .Values.I-HAVE-THE-KEY }}
 {{- else if .Values.cloudflare.secretName AND .Values.I-HAVE-THE-TOKEN }}
 {{- else if .Values.cloudflare.apiToken }}
 {{- else if .Values.cloudflare.apiKey}}
```

My knowledge of helm template language is not that great and maybe there are some other possibilities I am not aware

miguelaeh:
Hi @emas80 ,
Thanks for the clarification, I think I understand the problem. I have one question more, Has the `CF_API_TOKEN` inside your deployment a value? I mean, if your secret doesn't have the `cloudflare_api_token` key, which is the value of the `CF_API_TOKEN`? Is it an empty string?
Regarding the logic, I agree with you, it should be improved as it assumes if you provide a secret it will contain the `CF_API_TOKEN` and the `CF_API_KEY`.

emas80:
yes @miguelaeh it's an empty string. I've checked the code of external dns, and it should be all right, they check for the variable not being empty/empty string before going on and look for the pair key/email

miguelaeh:
Hi @emas80 ,
So if the code checks for the token to be empty or an empty string, and in that case fallback to the email and key, I don't understand why it is not working in the current approach, as it will try to take the token and the key form the secret, and if they are not in the secret they will be empty. The email will be taken always from the `.Values.cloudflare.email`.
So, 
> it will always provide the env var "CF_API_TOKEN" to external-dns.
That is okay, it will provide that env var with an empty string, but as you said
> they check for the variable not being empty/empty string before going on and look for the pair key/email
So it should be possible to provide the key and email without the token.
Maybe I am missing something, please correct me if I am wrong.

emas80:
This is the current code in the master

```
            {{- if or .Values.cloudflare.apiToken .Values.cloudflare.secretName }}
            - name: CF_API_TOKEN
              valueFrom:
                secretKeyRef:
                  name: {{ template "external-dns.secretName" . }}
                  key: cloudflare_api_token
            {{- else if or .Values.cloudflare.apiKey .Values.cloudflare.secretName }}
            - name: CF_API_KEY
              valueFrom:
                secretKeyRef:
                  name: {{ template "external-dns.secretName" . }}
                  key: cloudflare_api_key
            - name: CF_API_EMAIL
              value: {{ required "cloudflare.email is required if cloudflare.apiToken is not provided" .Values.cloudflare.email | quote }}
            {{- end }}
```

You see the line where it says

```
            {{- if or .Values.cloudflare.apiToken .Values.cloudflare.secretName }}
```

it means that if I specify a secret, then it enters ALWAYS in the first condition, so it always adds only the CF_API_TOKEN, there is no possibility to add the CF_API_KEY and CF_API_EMAIL to the deployment in that specific case.
Like I said, it's not possible to use the key/email if I have a secret.
Believe me, it's exactly what I am trying to do, and it doesn't work.

miguelaeh:
Thak you @emas80 for the clarification now I understand your point.
Then I thing we should change that code by something like this:
```
{{ if .Values.cloudflare.secretName }}
             - name: CF_API_KEY
              valueFrom:
                secretKeyRef:
                  name: {{ template "external-dns.secretName" . }}
                  key: cloudflare_api_key
            - name: CF_API_TOKEN
              valueFrom:
                secretKeyRef:
                  name: {{ template "external-dns.secretName" . }}
                  key: cloudflare_api_token
{{- else }}
           {{- if .Values.cloudflare.apiToken }}
            - name: CF_API_TOKEN
              value: {{ .Values.cloudflare.apiToken  }}
            {{- end }}
            {{- if .Values.cloudflare.apiKey }}
            - name: CF_API_KEY
              value: {{ .Values.cloudflare.apiKey }}
            {{- end }}
{{- end }}
           - name: CF_API_EMAIL
              value: {{ required "cloudflare.email is required if cloudflare.apiToken is not provided" .Values.cloudflare.email | quote }}
```
In the logic that is in the PR the if-else statement is not correct as it will use duplicated env vars providing a secret name.

lukasmrtvy:
@carrodher ping 
Fix: https://github.com/helm/charts/issues/21147#issuecomment-593499404

carrodher:
I am not taking a look at this case, I guess it was autoclosed because we are waiting for an answer or apply the pending changes.

@miguelaeh what is pending here? Is something we can implement on our side or take ownership of the changes?

miguelaeh:
Yes, I think the changes we talked about were not implemented.
The PR was closed because we didn't obtain an answer.
@emas80 would you plan to implement the changes? Or should we take the ownership of this PR?

lukasmrtvy:
@miguelaeh I think that everyone will be happy if You are decide to implement fix for it. Currently its not working, so waiting for something/someone will not help to anyone.

miguelaeh:
Closing this PR in favor of https://github.com/bitnami/charts/pull/2572

PR #2133, opened 2020-03-24
[bitnami/postgresql-ha] Move Pgpool to non-root

juan131:

#### Is this a new chart

No

This PR transforms the PostgreSQL-HA chart into a "non-root" chart by default. This practice is very extended on K8s and, indeed, it is mandatory for some K8s distributions such as **OpenShift** or **VMware Tanzu Kubernetes Grid.**

#### Checklist

### DIFF
--- a/bitnami/postgresql-ha/Chart.yaml
+++ b/bitnami/postgresql-ha/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: postgresql-ha
-version: 1.4.12
+version: 2.0.0
 appVersion: 11.7.0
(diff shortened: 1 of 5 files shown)


juan131:
Please **DO NOT MERGE** this PR until the CI/CD test everything and the image tags are updated

juan131:
> I would add a note on how to migrate the data from the previous major version to this one.

No backwards compatibility issues are expected since all the data is at PostgreSQL statefulset, and Pgpool uses a deployment without persistence. Therefore, upgrades should work smoothly

PR #2143, opened 2020-03-25
[bitnami/postgresql] fix replication with nodePort service type

### LINKED ISSUE #2063: [postgresql] NodePort + Relication fails
**Which chart**:
postgresql 8.6.4

**Description**
Running install with the following values fails:
```yaml
service:
  type: NodePort
  nodePort: 30042
replication:
  enabled: true
  slaveReplicas: 2
```
due to a collision in the nodePort for the master and slave services.

I'd propose a master.nodePort and slave.nodePort _but_ it occurs to me one might want master and slaves running different service types. We've also got that use case as master gets fed outside the cluster and the slaves serve internal workloads so given that I'd propose service entries under master and slave that override the top level service entry if present.

If that sounds reasonable I'll get to work on a PR.

Example proposed:
```yaml
service:
  type: NodePort
  nodePort: 30042
replication:
  enabled: true
  slaveReplicas: 2
slave:
  service:
    nodePort: 30043
```
alternatively expressible as:
```yaml
service:
  type: NodePort
replication:
  enabled: true
  slaveReplicas: 2
master:
  service:
    nodePort: 30042
slave:
  service:
    nodePort: 30043
```
More complex example:
```yaml
replication:
  enabled: true
  slaveReplicas: 2
master:
  service:
    type: NodePort
    nodePort: 30042
slave:
  service:
    type: ClusterIP
```

Git-Lucky:
**Description of the change**

When replication is enabled, if the service type is nodePort, there is a collision with the master and slave nodePorts.

**Benefits**

This allows for replication to be enabled and the services to avoid collision. This also allows for the master and slave nodes to be of different service types.

**Applicable issues**

Fixes #2063 for postgresql

**Checklist** 


bitnami/postgresql/README.md
```diff
@@ -144,6 +144,9 @@ The following tables lists the configurable parameters of the PostgreSQL chart a
 | `master.extraVolumeMounts`                    | Additional volume mounts to add to the pods (postgresql master)                                                                                                           | `[]`                                                          |
 | `master.extraVolumes`                         | Additional volumes to add to the pods (postgresql master)                                                                                                                 | `[]`                                                          |
 | `master.sidecars`                                         | Add additional containers to the pod                                                                                                                          | `[]`                                                     |
+| `master.service.type`                         | Kubernetes Master Service type (ovverides the service.type for differentiating from slave if desired)                                                                     | `nil`                                                         |
```
miguelaeh: I would change this by:
```
Allow to use a diferent service type for the master.             
```
And in the same way for the rest of comments

bitnami/postgresql/templates/_helpers.tpl
```diff
@@ -150,6 +150,96 @@ Return PostgreSQL port
...
 {{- end -}}
 
+{{/*
+Return PostgreSQL master service type
+*/}}
+{{- define "postgresql.masterServiceType" -}}
+{{- if .Values.master.service -}}
+    {{- if .Values.master.service.type -}}
+      {{- .Values.master.service.type -}}
+    {{- else if .Values.service.type -}}
```
miguelaeh: What do you think about change this by:
```
{{- define "postgresql.masterServiceType" -}}
{{- serviceType := .Values.service.type }}
{{- if .Values.master.service -}}
    {{- if .Values.master.service.type -}}
      {{- serviceType = .Values.master.service.type -}}
    {{- end -}}
{{- end -}}
{{- serviceType -}}
{{- end -}}
```
From my point this is more readable.
The rest of functions can be rafactored in the same way.

bitnami/postgresql/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: postgresql
-version: 8.6.9
+version: 8.6.10
```
miguelaeh: You should bump minor because you are adding new features.

bitnami/postgresql/README.md
Git-Lucky: Will do

bitnami/postgresql/templates/_helpers.tpl
Git-Lucky: Yes great suggestion

bitnami/postgresql/Chart.yaml
Git-Lucky: Sounds good...minor bump here

Git-Lucky:
@miguelaeh Is there a way I can add some tests for these new cases?

bitnami/postgresql/templates/_helpers.tpl
Git-Lucky: So after experimenting with our refactor suggestion I remembered why I had to settle on this solution in the first place. Because this logic also decides whether or not to include the value at all (i.e. Sometimes we include nodePort: and sometimes we don't, as in the case of clusterIP service type). When refactored in this way, it always assumes that we are returning a value. That is ok in the case of service type, but not for node port and clusterIP.

Git-Lucky: Oh and even in the case of service type there is an issue. For example, if master.service.clusterIP is defined, then the check for .Values.master.service returns true but when it checks for .Values.master.service.type it will be false and then no service type is returned at all

Git-Lucky: The control flow doesn't behave as might be expected and simply fall through to the original value

Git-Lucky: Please let me know if I am missing something about how this templating helper function works...I am definitely new to using them

miguelaeh: Well, I am assuming we are setting `ClusterIP` as the default value to the `.Values.service.type` in the `values.yaml` file.

miguelaeh:
Hi @Git-Lucky ,
We test the Helm Charts in our internal CI/CD, in the helm chart itself we don't have tests.
I answer one of your comments, please take a look: https://github.com/bitnami/charts/pull/2143#discussion_r400018384

bitnami/postgresql/templates/_helpers.tpl
Git-Lucky: ClusterIP doesn't have a default value, it is commented out by default

Git-Lucky:
I opened a new pull request with a better approach to this issue

PR #2187, opened 2020-04-01
[bitnami/mariadb-galera] Galera ssl replication traffic configuration parameters

rafariossaa:
**Description of the change**
Add support for the SSL replication traffic parameters.

**Checklist** 


bitnami/mariadb-galera/README.md
```diff
@@ -261,6 +266,34 @@ CREATE USER 'bitnami'@'localhost' IDENTIFIED VIA pam USING 'mariadb';
 
 With the above example, when the `bitnami` user attempts to login to the MariaDB server, he/she will be authenticated against the LDAP server.
 
+### SSL for Replication Traffic
+
+SSL support for replication traffic can be enabled in the chart by specifying the `ssl.rt.` parameters while creating a release. The followin parameters should be configured to properly enable the SSL support for replication traffic in the chart.
```
juan131, line 271: `followin` -> `following`
`the chart.` -> `the chart:`

bitnami/mariadb-galera/templates/_helpers.tpl
```diff
@@ -231,6 +231,28 @@ mariadb-galera: LDAP
...
+{{- end -}}
+
+{{/*
+Return the path to the CA cert file.
+*/}}
+{{- define "mariadb-galera.sslRTCACert" -}}
+{{- printf "/bitnami/mariadb/cert/%s" .Values.ssl.rt.certCAFilename -}}
+{{- end -}}
+
+
```
juan131, line 255: Avoid multiple empty lines

bitnami/mariadb-galera/templates/statefulset.yaml
```diff
@@ -128,6 +128,16 @@ spec:
             - name: MARIADB_EXTRA_FLAGS
               value: {{ .Values.extraFlags | quote }}
             {{- end }}
+            {{- if .Values.ssl.rt.enabled }}
+            - name:  MARIADB_ENABLE_SSL_RT
+              value: yes
```
juan131, line 133: You might want to move this environment variable before the `{{- if .Values.ssl.rt.enabled }}`, and use:
```
            - name:  MARIADB_ENABLE_SSL_RT
              value: {{ ternary "yes" "no" .Values.ssl.rt.enabled | quote }}
```
This way you're always setting the env. variable and you don't need to rely on the default value that comes from the container.

bitnami/mariadb-galera/values.yaml
```diff
@@ -178,6 +178,22 @@ ldap:
...
+    ## Enable SSL for replication traffic
+    enabled: false
+    ## Name of the secret that contains the certificates
+    certificatesSecret:
+    ## Certificate filename
+    certFilename:
+    ## Certificate Key filename
+    certKeyFilename:
+    ## CA Certificate filename
+    certCAFilename:
```
juan131, line 195: General conventions:

Use:

```
## Some comment
##
key: value
```

Instead of:

```
## Some comment
key: value
```

When you're not setting any default value, comment the key, instead of setting it to an empty value:

```diff
- key:
+ # key
```

PR #2188, opened 2020-04-01
[bitnami/common]: add initial functions

dani8art:
**Description of the change**

Adding bitnami/common with useful common template helpers to be reused as part of the standardization.

**Benefits**

Having clear charts and standardized for common things like affinity, sidecars, images, etc.

**Possible drawbacks**

N/A

**Applicable issues**

N/A

**Additional information**

N/A

**Checklist**

### DIFF
--- a/bitnami/common/Chart.yaml
+++ b/bitnami/common/Chart.yaml
@@ -1,11 +1,18 @@
 apiVersion: v1
 name: common
-description: A Library Helm Chart for grouping common logic between bitnami charts. This chart is not deployable by itself.
-
-# This is the chart version. This version number should be incremented each time you make changes
-# to the chart and its templates, including the app version.
 version: 0.1.0
-
-# This is the version number of the application being deployed. This version number should be
-# incremented each time you make changes to the application.
 appVersion: 0.1.0
+description: A Library Helm Chart for grouping common logic between bitnami charts. This chart is not deployable by itself.
+keywords:
+  - common
+  - helper
+  - template
+  - function
+  - bitnami
+icon: https://bitnami.com/downloads/logos/bitnami-mark.png
+sources:
+  - https://github.com/bitnami/charts
+maintainers:
+  - name: Bitnami
+    email: containers@bitnami.com
+engine: gotpl
--- a/bitnami/common/Chart.yaml
+++ b/bitnami/common/Chart.yaml
@@ -1,0 +1,11 @@
+apiVersion: v1
+name: common
+description: A Library Helm Chart for grouping common logic between bitnami charts. This chart is not deployable by itself.
+
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+version: 0.1.0
+
+# This is the version number of the application being deployed. This version number should be
+# incremented each time you make changes to the application.
+appVersion: 0.1.0
--- a/bitnami/common/README.md
+++ b/bitnami/common/README.md
@@ -1,0 +1,159 @@
+# Bitnami Common Library Chart
+
+A [Helm Library Chart](https://helm.sh/docs/topics/library_charts/#helm) for grouping common logic between bitnami charts.
+
+## TL;DR;
+
+```yaml
+dependencies:
+- name: common
+  version: 0.1.0
+  repository: https://charts.bitnami.com/bitnami
+```
+
+```bash
+$ helm dependency update
+```
+
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "common.names.fullname" . }}
+data:
+  myvalue: "Hello World"
+```
+## Introduction
+
+This chart provide a common template helpers which can be used to develop new charts using [Helm](https://helm.sh) package manager.
+
+Bitnami charts can be used with [Kubeapps](https://kubeapps.com/) for deployment and management of Helm Charts in clusters. This Helm chart has been tested on top of [Bitnami Kubernetes Production Runtime](https://kubeprod.io/) (BKPR). Deploy BKPR to get automated TLS certificates, logging and monitoring for your applications.
+
+## Prerequisites
+
+- Kubernetes 1.12+
+- Helm 2.16+ or Helm 3.0-beta3+
+- PV provisioner support in the underlying infrastructure
+- ReadWriteMany volumes for deployment scaling
+
+## Helpers
+
+The following table lists the helpers available in the library which are scoped in different sections.
+
+**Names**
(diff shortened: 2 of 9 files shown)


bitnami/common/README.md
```diff
@@ -0,0 +1,159 @@
...
+A [Helm Library Chart](https://helm.sh/docs/topics/library_charts/#helm) for grouping common logic between bitnami charts.
+
+## TL;DR;
+
+```yaml
+dependencies:
+- name: common
+  version: 0.1.0
+  repository: https://charts.bitnami.com/bitnami
+```
```
juan131, line 12: Indent the array two spaces

bitnami/common/README.md
```diff
@@ -0,0 +1,159 @@
...
+
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "common.names.fullname" . }}
+data:
+  myvalue: "Hello World"
+```
+## Introduction
```
juan131, line 26: Markdown good practices: always add an empty line between code blocks and text

bitnami/common/README.md
```diff
@@ -0,0 +1,159 @@
...
+
+This chart provide a common template helpers which can be used to develop new charts using [Helm](https://helm.sh) package manager.
```
juan131, line 28: `provide` -> `provides`

bitnami/common/README.md
```diff
@@ -0,0 +1,159 @@
...
+
+Bitnami charts can be used with [Kubeapps](https://kubeapps.com/) for deployment and management of Helm Charts in clusters. This Helm chart has been tested on top of [Bitnami Kubernetes Production Runtime](https://kubeprod.io/) (BKPR). Deploy BKPR to get automated TLS certificates, logging and monitoring for your applications.
+
+## Prerequisites
+
+- Kubernetes 1.12+
+- Helm 2.16+ or Helm 3.0-beta3+
+- PV provisioner support in the underlying infrastructure
+- ReadWriteMany volumes for deployment scaling
```
juan131, line 37: I don't think PV provisioner support is need, nor ReadWriteMany volumes (I guess you copy/pasted from other chart)

bitnami/common/README.md
```diff
@@ -0,0 +1,159 @@
...
+
+## Helpers
```
juan131, line 39: @carrodher correct me if I'm wrong but I think that some automation we're doing expects this section to be called `## Parameters`. Am i right?

bitnami/common/README.md
```diff
@@ -0,0 +1,159 @@
...
+|---------------------------------------------|------------------------------------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `common.names.name`                         | Expand the name of the chart or use `.Values.nameOverride` | `.` Chart context                                                                                                                                        |
+| `common.names.chart`                        | Chart name plus version                                    | `.` Chart context                                                                                                                                        |
+| `common.names.fullname`                     | Create a default fully qualified app name.                 | `.` Chart context                                                                                                                                        |
```
juan131, line 48: I'd switch the order between these two parameters in the table

bitnami/common/templates/_capabilities.tpl
```diff
@@ -0,0 +1,11 @@
...
+{{/*
+Return the appropriate apiVersion for deployment.
+*/}}
+{{- define "common.capabilities.deployment.apiVersion" -}}
+{{- if semverCompare "<1.14-0" .Capabilities.KubeVersion.GitVersion -}}
+{{- print "extensions/v1beta1" -}}
+{{- else -}}
+{{- print "apps/v1" -}}
+{{- end -}}
+{{- end -}}
```
juan131, line 11: This is great but we should extend this to other K8s objects: `PodSecurityPolicy`, `NetworkPolicy`

bitnami/common/README.md
```diff
@@ -0,0 +1,159 @@
...
+| `common.tplvalues.render`                   | Renders a value that contains template                     | `dict "value" .Values.path.to.the.Value "context" $`, value is the value should rendered as template, context frecuently is the chart context `$` or `.` |
+
+**Capabilities**
+| Helper identifier                           | Description                                                | Expected Input                                                                                                                                           |
+|---------------------------------------------|------------------------------------------------------------|----------------------------------------------------------------------------------------------------------------------------------------------------------|
+| `common.capabilities.deployment.apiVersion` | Return the appropriate apiVersion for deployment.          | `.` Chart context                                                                                                                                        |
+
+
```
juan131, line 77: Too many empty lines

bitnami/common/Chart.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
+name: common
+description: A Library Helm Chart for grouping common logic between bitnami charts. This chart is not deployable by itself.
+
```
juan131, line 4: We should include an icon. Maybe a "bitnami" icon or sth like that.

juan131: Other fields I'm missing: `engine`, `maintainers` and `keywords`

bitnami/common/templates/_capabilities.tpl
dani8art: Right totally agree, I think we could add it in an "on demand" approach. I mean, when we find a use case for `PodSecurityPolicy` or `NetworkPolicy`

bitnami/common/README.md
carrodher: There is a tool doing some replaces in the README before send it to marketplaces like IBM or Azure. If we are going to use this chart as a dependency for other charts we will need to send it to the Index of those companies, but I am not sure if the chart itself is going to be listed in the marketplaces as it is not installable by itself.

So, if you can use `## Parameters` is better, although we need to study this user case regarding marketplaces publishing

bitnami/common/templates/_images.tpl
```diff
@@ -0,0 +1,38 @@
+{{/* vim: set filetype=mustache: */}}
```
carrodher, line 1: For now it's ok, let's see what happens with the new Pivotal convention

dani8art: yes!

bitnami/common/templates/_capabilities.tpl
juan131: We already have cases for this (also jobs):
- https://github.com/bitnami/charts/blob/master/bitnami/elasticsearch/templates/cronjob.yaml
- https://github.com/bitnami/charts/blob/master/bitnami/redis/templates/psp.yaml
- https://github.com/bitnami/charts/blob/master/bitnami/postgresql/templates/networkpolicy.yaml

PR #2202, opened 2020-04-02
Update the phpMyAdmin requirements with the latest MariaDB

jotamartos:
Changes to use the latest version of MariaDB in the phpMyAdmin chart


carrodher:
Let's wait for the pipeline updating the tags and _requirements.lock_

PR #2219, opened 2020-04-03
[bitnami/rabbitmq] livenessProbe and readinessProbe command override

moskitone:
**Benefits**
Added ability to override livenessProbe and readinessProbe command.

**Possible drawbacks**
no

**Applicable issues**

**Additional information**

**Checklist** 


moskitone:
> Thanks for your PR.
> 
> Could you make the changes needed in values-production.yml.
> Also, in this case, it would be more appropriate to increase the minor digit form version of the chart instead of the release digit.

Thanks for your reply.
I've updated version and values-production.

PR #2229, opened 2020-04-05
[bitnami/mongodb-sharded ]- Add full support of sidecar containers.

### LINKED ISSUE #2228: [bitnami/mongodb-sharded] sidecar container definitions are ignored
**Which chart**:
bitnami/mongodb-sharded

**Describe the bug**
When defining sidecar containers in the helm yaml for any mongodb pod (configsrv, mongos, shardsvr.arbiter or shardsvr.dataNode) that definitions are ignored. No additional containers appear in the output yaml nor are any containers created in kubernetes.

**To Reproduce**
```
git clone https://github.com/bitnami/charts.git
cp charts/bitnami/mongodb-sharded/values.yaml myValues.yaml
.. edit ..
configsvr:
  ..
  sidecars:
    - name: mytest
      image: bitnami/node
      imagePullPolicy: IfNotPresent
.. and save.
helm install --dry-run --debug my-test -n my-test-space -f myValues.yaml charts/bitnami/mongodb-sharded | grep bitnami/node -C10
```
You see bitnami/node right after sidecars: two times - each time it is just the definition of the myValues.yaml.
But there is no image 'bitnami/node' added in the containers section of any output *-statefulset.yaml. 
If I do not do a dry run, but a real deploy, when I do
kubectl get pods -n my-test-space
```
NAME                                              READY   STATUS    RESTARTS   AGE
my-test-mongodb-sharded-configsvr-0               1/1     Running   1          3m10s
my-test-mongodb-sharded-mongos-547dcdccf5-k46xk   1/1     Running   0          3m10s
my-test-mongodb-sharded-shard0-data-0             1/1     Running   3          3m10s
my-test-mongodb-sharded-shard1-data-0             1/1     Running   3          3m10s
```

**Expected behavior**
When defining a sidecar like above, there should be an additional entry in the containers section of 
```
# Source: mongodb-sharded/templates/config-server/config-server-statefulset.yaml
..
spec:
  selector:
    matchLabels:
      app.kubernetes.io/name: mongodb-sharded
      app.kubernetes.io/instance: my-test
      app.kubernetes.io/component: configsvr
  ..
  template:
    ..
    spec:
      ..
      containers:
        - name: mongodb
          ..
        - name: mytest
          image: bitnami/node
          imagePullPolicy: IfNotPresent
Additionally I would expect certain mongo settings like the root password also be available in those sidecars:
          env:
            - name: MONGODB_ROOT_PASSWORD
              valueFrom:
                secretKeyRef:
                  name: my-test-mongodb-sharded
                  key: mongodb-root-password
```
If I do not do a dry run, but a real deploy, when I do
kubectl get pods -n my-test-space
I exprect to see:
```
NAME                                              READY   STATUS    RESTARTS   AGE
my-test-mongodb-sharded-configsvr-0               2/2     Running   1          3m10s
my-test-mongodb-sharded-mongos-547dcdccf5-k46xk   1/1     Running   0          3m10s
my-test-mongodb-sharded-shard0-data-0             1/1     Running   3          3m10s
my-test-mongodb-sharded-shard1-data-0             1/1     Running   3          3m10s
```
There is an additional container running. - configsvr-0 2/2

- Output of `helm version`:

```

```

```
Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.0", GitCommit:"9e991415386e4cf155a24b1da15becaa390438d8", GitTreeState:"clean", BuildDate:"2020-03-25T14:58:59Z", GoVersion:"go1.13.8", Compiler:"gc", Platform:"linux/amd64"}
```

**Additional context**
N/A

thomasgafner:
**Description of the change**

When defining sidecar containers in the helm yaml for any mongodb sharded pod (configsrv, mongos, shardsvr.arbiter or shardsvr.dataNode) that definitions are no longer ignored. When defining them in one of the five possible sections of the helm yaml (common, configsrv, mongos, shardsvr.arbiter or shardsvr.dataNode) additional containers appear in the output yaml and are created within the kubernetes cluster.

**Benefits**

Full sidecar support is added. Mongodb authentication is also available for sidecar containers.

**Possible drawbacks**

N/A

**Applicable issues**

  - fixes #2228

**Additional information**

N/A


bitnami/mongodb-sharded/templates/config-server/config-server-statefulset.yaml
```diff
@@ -271,6 +271,170 @@ spec:
           {{- end }}
           resources: {{ toYaml .Values.metrics.resources | nindent 12 }}
         {{- end }}
+        {{- range $car := $.Values.configsvr.sidecars }}
```
javsalgar, line 274: Why not using `tplValue` for the whole array? That would bring much more flexibility than having to set all the cases in the yaml. Users could use the templating language without issues (because it is using tplValue) and the yaml would become more readable

thomasgafner:
Good question. - This was also my first thought.
I decided against the easy one liner with tplValue because I presumed that
there will be some settings in the env section that should be done also to
each sidecar 'automatically' like the BITNAMI_DEBUG, MONGODB_ROOT_PASSWORD,
etc., which certainly are needed by some sidecars. - For example a percona
backup sidecar. - May be my solution even lacks copying some more MONGODB_*
values in the env section.
Another example is the livenessProbe, that has a global setting for
disabling or enabling: I thought for sidecars this global switch should
also be respected. So I did:
{{- if and $car.livenessProbe $.Values.livenessProbe.enabled }}
With just one single line of tplValue this would not be possible. - Or I do
not understand how.






Am Mo., 6. Apr. 2020 um 12:50 Uhr schrieb Javier J. Salmerón-García <
notifications@github.com>:

dani8art:
Yes, I would go for the `tplValue` helper too, it gives you more flexibility and also you can use all the common things since you can write it as a template in the `values.yaml`:

```
  sidecars:
    - name: mytest
      image: bitnami/node
      imagePullPolicy: IfNotPresent
      env:
      {{- if $.Values.usePasswordFile }}
        - name: MONGODB_ROOT_PASSWORD_FILE
           value: "/bitnami/mongodb/secrets/mongodb-root-password"
      {{- else }}
        - name: MONGODB_ROOT_PASSWORD
           valueFrom:
             secretKeyRef:
               name: {{ include "mongodb-sharded.secret" $ }}
               key: mongodb-root-password
```

And then it is going to be evaluated in the templates.

thomasgafner:
As an 'end user', if I go here
https://hub.helm.sh/charts/bitnami/mongodb-sharded
and read how to use the chart
```
$ helm repo add bitnami https://charts.bitnami.com/bitnami
$ helm install my-release bitnami/mongodb-sharded
```
I do so. But soon I want to provide my own configuration and I read that I can do 
either
```
$ helm install my-release --set ..
```
or
```
$ helm install my-release -f values.yaml bitnami/mongodb-sharded
```
(I also add --dry-run and --debug . And I can do all of this without downloadin the chart code - helm caches the repo for me.)
For a sidecar I certainly chose the -f values.yaml option, providing my own file.

In both ways no go templating is working. - Even if I download the chart code: If I use a yaml that is outside the bitnami chart folder (e.g. in my project folder, where I just want to use the chart - not develop it) I cannot even do a simple
  value: {{ .Release.Name }}
As soon as I add moustaches the compile fails.
So it was my expectation that the bitnami/mongodb-sharded chart will do all for me, which includes env variables like MONGODB_ROOT_PASSWORD just being ready to use without templating like:
```
    sidecars:
      - name: pbm-agent
        image: perconalab/percona-backup-mongodb
        imagePullPolicy: IfNotPresent
        command:
          - /bin/sh
          - -c
          - ls "$MONGODB_ROOT_PASSWORD_FILE" &> /dev/null && export MONGODB_ROOT_PASSWORD="$(< "${MONGODB_ROOT_PASSWORD_FILE}")" ; pbm-agent --mongodb-uri "mongodb://root:$(MONGODB_ROOT_PASSWORD)@localhost:27017"
```

So I did the implementation of sidecars accordingly.

But may be I am misunderstanding the scope or something.

thomasgafner:
Now I understand, that you ask me to change the impl in the according containers section to:
```
      containers:
        .. mongos and metrics ..
        {{- with $.Values.shardsvr.dataNode.sidecars }}
        {{- include "mongodb-sharded.tplValue" ( dict "value" . "context" $ ) | nindent 8}}
        {{- end }}
```
This means that in a sidecar env vars like MONGODB_ROOT_PASSWORD are only available using the go templating if someone is creating his own chart. If someone only want's to use the mongodb_sharded chart with a config yaml file (alternatively to use --set arguments - I am referencing the end of https://hub.helm.sh/charts/bitnami/mongodb-sharded - Parameters section), none of the env vars, no go templating nor tplValue are available.
Furthermore:
- If the sidecar is a bitnami image e.g. bitnami/node then I would also expect that some general settings are propagated 'automatically' to that sidecar too: like BITNAMI_DEBUG.
- My impl does not use tplValue for the array of sidecars, but within each sidecar it uses tplValue wherever possible.

So I think my impl serves both usecases the simple 'chart user' and a possible 'chart developer'. I presume the latter usecase to happen much less often than the first one.
Shall I really change the impl to use tplValue for the whole sidecars array?

javsalgar:
Hi,

> If someone only want's to use the mongodb_sharded chart with a config yaml file (alternatively to use --set arguments - I am referencing the end of https://hub.helm.sh/charts/bitnami/mongodb-sharded - Parameters section), none of the env vars, no go templating nor tplValue are available.

I think I didn't understand this. Could you explain this part in more detail? Indeed, the user would need to set as many env vars as necessary. Could you also explain the use case that you have in mind that would require propagating vars? I think that the more flexible it is, the better.

thomasgafner:
Ok, I try to explain it in more detail. - May be this could also reveal a misunderstanding on my part.

If someone just wants to USE the bitnami/mongodb-sharded chart in his project (may be a nodejs project) having also a kubernetes deployment task, then he will do as described in https://hub.helm.sh/charts/bitnami/mongodb-sharded (or https://github.com/bitnami/charts)
```
$ helm repo add bitnami https://charts.bitnami.com/bitnami
$ helm install my-release bitnami/mongodb-sharded

```
Helm will then download and cache the bitnami/mongodb-sharded chart code.
The helm install / upgrade command might also be part of a deployment script like:
```
npm run deployK8s
package.json
  "scripts": {
    "deployK8s": "helm upgrade --install my-release bitnami/mongodb-sharded",
```
But most probably the default settings (defined in the charts values.yaml) do not fit that projects needs and the owner configures the deployment either
- by adding --set arguments to the helm upgrade --install command (in the script)
or 
- by creating a values.yaml somewhere in his project (and reference it from the script helm upgrade --install -f <file> ..)

That values.yaml might be a copy of the charts values.yaml as proposed in https://hub.helm.sh/charts/bitnami/mongodb-sharded
"
Alternatively, a YAML file that specifies the values for the parameters can be provided while installing the chart. For example,
$ helm install my-release -f values.yaml bitnami/mongodb-sharded
    Tip: You can use the default values.yaml
"
If the project owner needs to configure a sidecar the --set option is too cumbersome and a values.yaml is quite a must.
Now if the project owner wants to add a sidecar that is very mongodb related (which is probable when using bitnami/mongodb-sharded) he could expect the MONGODB related vars of the mongodb container to be propagated also to his sidecar (in the same way it is also done for the metrics container automatically - by the way). One of these vars could be the root auth for accessing the mongodb - MONGODB_ROOT_PASSWORD and MONGODB_ROOT_PASSWORD_FILE.
Why? - Becuase if they are not propagated, what can the owner do to get them?
He CANNOT just insert 
```
      {{- if $.Values.usePasswordFile }}
        - name: MONGODB_ROOT_PASSWORD_FILE
           value: "/bitnami/mongodb/secrets/mongodb-root-password"
      {{- else }}
        - name: MONGODB_ROOT_PASSWORD
           valueFrom:
             secretKeyRef:
               name: {{ include "mongodb-sharded.secret" $ }}
               key: mongodb-root-password
```
as proposed by dani8art, because the project owner is not developing a chart - just using it - and no go templating (no {{ .. }} ) is possible in his copy of values.yaml.
Unfortunately also the following relies on templating and the charts _helpers.tpl - so is not available in the project - outside the chart:
```
            - name: MONGODB_ROOT_PASSWORD
              valueFrom:
                secretKeyRef:
                  name: {{ include "mongodb-sharded.secret" $ }}
                  key: mongodb-root-password
```
The best the project owner can do in his values.yaml is:
```
mongodbRootPassword: <HIS_OWN_PW>
..
    sidecars:
      - name: pbm-agent
        ..
        env:
          - name: MONGODB_ROOT_PASSWORD
            value: <HIS_OWN_PW>
        command:
          - /bin/sh
          - -c
          - pbm-agent --mongodb-uri --mongodb-uri "mongodb://root:$(MONGODB_ROOT_PASSWORD)@localhost:27017"
```
But this solution is nor very nice, because the password is written in the source code of the project - and goes to its git - and even if this was acceptable or solved the password appears redundant in the values.yaml, which is not error prone when updating.

The third option is that the project owner decides to become chart developer too. Then he might copy a full version of bitnami/mongodb-sharded into his project and do all changes to the original charts values.yaml where go templating works and all _helpers.tpl are available - code like dani8art's proposal is then possible. But that way - in my opinion - the helm repo and upgrade idea is somehow 'corrupted' and chart internas get mixed with each project using it, which widens the charts 'API' unnecessarily and can end up in a maintainance nightmare.

So I thought bitnami/mongodb-sharded should provide a solution, that fits the chart users needs as well as the chart developer ones. - And since both is possible (ok with some more code than just one line per sidecars section), why not provide it?

javsalgar:
Hi,

> as proposed by dani8art, because the project owner is not developing a chart - just using it - and no go templating (no {{ .. }} ) is possible in his copy of values.yaml.

The user can use `{{ ... }}` in the values.yaml, that's what the templating is there for. When you want to use more complex templating, the user should do something like this

```
sidecars: |-
      {{- if $.Values.usePasswordFile }}
        - name: MONGODB_ROOT_PASSWORD_FILE
           value: "/bitnami/mongodb/secrets/mongodb-root-password"
      {{- else }}
        - name: MONGODB_ROOT_PASSWORD
           valueFrom:
             secretKeyRef:
               name: {{ include "mongodb-sharded.secret" $ }}
               key: mongodb-root-password
```

Note that I turned `sidecars` into a string instead of an array (this is done when you want to use complex templating). That should be enough in that case. 

We are following this approach in most of our charts so I would prefer that you use the approach we mentioned as well. If users start asking for a different way (more aligned with what you propose), we can reconsider changing it for future versions.

thomasgafner:
Thank you javsalgar for clarification. Templating is possible.
It must have been an error in my setup, that made me think differently. Somehow for my complex setup it is only working when adding |- after sidecars: - as you indicated.
When I change my impl to that simple one liner with tplValue, in the projects yaml I can use {{ }} and everything from the _helpers.tpl. 

Only if I also try to add to my env this
```
          - name: MONGODB_PRIMARY_HOST
            value: {{ printf "%s-shard%d-data-0.%s-headless.%s.svc.%s" (include "mongodb-sharded.fullname" $ ) $i (include "mongodb-sharded.fullname" $ ) $.Release.Namespace $.Values.clusterDomain }}
            {{- if $.Values.common.useHostnames }}
```
it fails, because of the $i (everything else is fine). The $i is defined at the top of the charts yaml files to make the loop over the number of shards like that:
```
{{- $replicas := .Values.shards | int }}
{{- range $i, $e := until $replicas }}
```
It seems not to be available in my yaml, but I will figure out, how to get MONGODB_PRIMARY_HOST in the sidecar too. - Actually I just need to know whether it is the sidecar of a PRIMARY or SECONDARY data shard.

So I changed the impl of my fix as proposed by javsalgar and dani8art. It is flexible enough for all usecases.

PR #2242, opened 2020-04-06
[bitnami/node-exporter] Let the prometheus scrape annotation be optional

RichieSams:
The user may not want to add the prometheus scrape annotation. We have to keep the default behavior for backwards compatibility, so we can't remove it altogether.

Instead we put it behind a default true bool.

The variable name is a tad verbose. I'd be ok changing it to something else, if you would like.


juan131:
Waiting for CI/CD to update image tags

PR #2252, opened 2020-04-07
[bitnami/rabbitmq] Tune parameters to increase performance

### LINKED ISSUE #2189: RabbitMQ: nodes are pegged to a single CPU core in production configuration
**Which chart**:

RabbitMQ

**Describe the bug**

These [default production values](https://github.com/bitnami/charts/blob/master/bitnami/rabbitmq/values-production.yaml#L100-L104) for [runtime schedulers](https://www.rabbitmq.com/runtime.html#scheduling) are very questionable and peg the nodes
deployed using this chart to a single core, which has obvious implications.

It would make a lot more sense to use as many schedulers as a pod is allowed to use.
If that's not possible to compute, there should be very clear and visible documentation that says
that these values should be modified in actual production deployments.

**Additional context**

 * [RabbitMQ Runtime Tuning guide](https://www.rabbitmq.com/runtime.html#scheduling)

rafariossaa:
**Description of the change**

**Benefits**
This changes will fix the issue #2189 by allowing the Erlang VM to decide the number of threads 

**Possible drawbacks**

None

**Applicable issues**
Production scenarios.
  - fixes #2189

**Checklist** 


bitnami/rabbitmq/values.yaml
```diff
@@ -97,11 +97,12 @@ rabbitmq:
...
 
-  ## RabbitMQ maximum available scheduler threads and online scheduler threads
+  ## RabbitMQ maximum available scheduler threads and online scheduler threads. By default it will create a thread per CPU detected, with the following parameters you can tune it manually.
   ## ref: https://hamidreza-s.github.io/erlang/scheduling/real-time/preemptive/migration/2016/02/09/erlang-scheduler-details.html#scheduler-threads
+  ## ref: https://github.com/bitnami/charts/issues/2189
   ##
-  maxAvailableSchedulers: 2
-  onlineSchedulers: 1
+  # maxAvailableSchedulers: 2
+  # onlineSchedulers: 1
```
miguelaeh, line 105: Do you know which could be the implications of remove this configuration if you had the chart installed from the previous version setting this? Could this change break it?
For example, if a user installed the version `6.24.0` those parameters will have a value, then, if they upgrade the chart to this version the value of those paramters will be removed.

rafariossaa: The implications will be that as the Erlang VM is setting this it will choose an appropiate number of threads for the CPUs detected. Maybe you get an increase in performance.

miguelaeh: I will trust you

PR #2258, opened 2020-04-07
[bitnami/external-dns] azure secrets set in env by secretKeyRef

### LINKED ISSUE #2221: [bitnami/external-dns] Azure Private DNS : use secrets for sensitive env variable
**Which chart**:
bitnami/external-dns

Provider Azure private dns : sensitive data are being passed directly to deployment definition instead of via secrets

It would be way better to use env from secrets 

NONE...

**Additional context**
Use secrets for following values in https://github.com/bitnami/charts/blob/master/bitnami/external-dns/templates/deployment.yaml :
```
            {{- if eq .Values.provider "azure-private-dns" }}
            # Azure Private DNS variables
            {{- if .Values.azure.tenantId }}
            - name: AZURE_TENANT_ID
              value: {{ .Values.azure.tenantId }}
            {{- end }}
            {{- if .Values.azure.aadClientId }}
            - name: AZURE_CLIENT_ID
              value: {{ .Values.azure.aadClientId }}
            {{- end }}
            {{- if .Values.azure.aadClientSecret }}
            - name: AZURE_CLIENT_SECRET
              value: {{ .Values.azure.aadClientSecret }}
            {{- end }}
            {{- end }}
```

coopernetes:
**Description of the change**
* bugfix of Azure credential variables being set in plaintext environment variables in the deployment template. Store these values in a secret and reference by name & key
* combine `azure` and `azure-private-dns` provider validation & conditionals where they use the same values throughout
* add separate validation for `azure-private-dns` since it only works through env vars and not an `azure.json` configuration file

**Benefits**
* credentials that were in the clear in the deployment environment variables now reference the generated secret by name & key instead of plaintext
* expanding Helm validation for `azure` provider is now applied for the `azure-private-dns` provider where the two providers reuse the same values

**Possible drawbacks**
I'll be testing these changes against an Azure private DNS only so may need a hand validating this against public Azure DNS zones. Have not validated if there's broken compatibility with the `azure.secretName` value and its handling in the deployment template. Need to test this on a live cluster.

**Applicable issues**
  - fixes #2221 

**Additional information**
In the event that there are true semantic differences between these providers, a split is warranted in the current azure Values. Probably best to do that sooner rather than later since they are different APIs on the cloud provider and they may diverge in terms of feature set or functionality. 

**Checklist** 

### DIFF
--- a/bitnami/external-dns/Chart.yaml
+++ b/bitnami/external-dns/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: external-dns
-version: 2.20.11
+version: 2.20.12
 appVersion: 0.7.1
 description: ExternalDNS is a Kubernetes addon that configures public DNS servers with information about exposed Kubernetes services to make them discoverable.
 keywords:
--- a/bitnami/external-dns/templates/_helpers.tpl
+++ b/bitnami/external-dns/templates/_helpers.tpl
@@ -118,6 +118,8 @@ Return true if a secret object should be created
     {{- true -}}
 {{- else if and (eq .Values.provider "azure") (or (and .Values.azure.resourceGroup .Values.azure.tenantId .Values.azure.subscriptionId .Values.azure.aadClientId .Values.azure.aadClientSecret (not .Values.azure.useManagedIdentityExtension)) (and .Values.azure.resourceGroup .Values.azure.tenantId .Values.azure.subscriptionId .Values.azure.useManagedIdentityExtension)) (not .Values.azure.secretName) -}}
     {{- true -}}
+{{- else if and (eq .Values.provider "azure-private-dns") (or (and .Values.azure.resourceGroup .Values.azure.tenantId .Values.azure.subscriptionId .Values.azure.aadClientId .Values.azure.aadClientSecret (not .Values.azure.useManagedIdentityExtension)) (and .Values.azure.resourceGroup .Values.azure.tenantId .Values.azure.subscriptionId .Values.azure.useManagedIdentityExtension)) (not .Values.azure.secretName) -}}
+    {{- true -}}
 {{- else if and (eq .Values.provider "cloudflare") (or .Values.cloudflare.apiToken .Values.cloudflare.apiKey) (not .Values.cloudflare.secretName) -}}
     {{- true -}}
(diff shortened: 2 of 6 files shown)


coopernetes:
Thanks for catching that @juan131 :) 

I'd like to do some manual testing of this chart in an Azure environment before merge.

coopernetes:
Hey @juan131, when going through my manual testing, I encountered an odd behavior at the current commit. With my changes, a user isn't able to specify a pre-existing `secretName` for the azure-private-dns provider. I can see this as a valid use case for environments that may be using other secret management tools to populate their Kubernetes Secrets (a la [sealed secrets](https://github.com/bitnami-labs/sealed-secrets), etc). We should support this uniformly in both providers, even if they handle the Azure specific details differently.

Given these differences, the chart has to handle the following cases:

1. provider="azure" and no secretName specified. All values are required (rg, sub, tenant, client id & secret).
2. provider="azure" and secretName exists. All parameters are derived from this secret (**unique to azure provider**).
3. provider="azure-private-dns" and no secretName specified. All parameters are required (rg, sub, tenant, client id & secret).
4. Provider="azure-private-dns" and secretName is specified. Non-secret parameters are **still required** as they are specified as args & an environment variable. Client id and secret values are not used.
5. One additional variation for both providers when a managed identity is used.

Sorry if this is expanding the scope of this PR but I want to avoid any future bugs that will arise from this inconsistency. I might open an issue in upstream [external-dns](https://github.com/kubernetes-sigs/external-dns) to unify this across both so this complexity doesn't need to be maintained in this chart.

coopernetes:
Decided to push up those changes I mentioned anyways. Manual testing on both providers looks good :+1: 
https://github.com/coopernetes/kube-testing-repo/tree/bitnami/charts-pr2258#results

juan131:
Great @coopernetes !! Launching the CI/CD to test your changes

juan131:
The CI/CD tests passed successfully! Merging your PR @coopernetes 

Thanks so much for it!

ukphillips:
I am currently encountering issues with this logic in the public DNS zone. Resource Group is not in the secret - can you link any documentation indicating this?

andresbono:
Great, @ukphillips!

PR #2263, opened 2020-04-08
[bitnami/thanos] Add support to create Headless services

Reuuke:
**Description of the change**

- Removes separated Storegateway and Ruler Headless services
- Adds support to create headless service in the main service.yaml

**Benefits**

**Possible drawbacks**

**Applicable issues**

**Additional information**

**Checklist** 


bitnami/thanos/Chart.yaml
```diff
@@ -15,4 +15,4 @@ maintainers:
 name: thanos
 sources:
   - https://github.com/bitnami/bitnami-docker-thanos
-version: 0.3.1
\ No newline at end of file
+version: 0.3.2
```
juan131: Please rebase your branch from master and bump again the version, but this bump the "minor" version since this PR is introducing new parameters. 

Please also add a "newline" at the end of the file as requested in the error above.

bitnami/thanos/values-production.yaml
```diff
@@ -175,6 +175,9 @@ querier:
     ## Service type
     ##
     type: ClusterIP
+    ## Thanos Querier service clusterIP IP
+    ##
+    ClusterIP: ""
```
juan131, line 180: Instead of setting it to `""` by default, please use the approach below:
```yaml
  ## Set the Cluster IP to use
  ## ref: https://kubernetes.io/docs/concepts/services-networking/service/#choosing-your-own-ip-address
  ##
  # clusterIP: None
```

> Note: use camelCase style: `ClusterIP` -> `clusterIP`

Please apply this suggestion to the rest of parameter you added.

bitnami/thanos/templates/storegateway/service-headless.yaml
```diff
@@ -1,22 +0,0 @@
...
-      targetPort: http
-      protocol: TCP
-      name: http
-    - port: {{ .Values.storegateway.service.grpc.port }}
-      targetPort: grpc
-      protocol: TCP
-      name: rcp
-  selector: {{- include "thanos.matchLabels" . | nindent 4 }}
-    app.kubernetes.io/component: storegateway
-{{- end }}
```
juan131, line 22: Why did you remove this headless service? It's useful so I can have bot a headless service and a non-headless service for StoreGateway.

bitnami/thanos/Chart.yaml
Reuuke: Done

bitnami/thanos/values-production.yaml
Reuuke: Done

bitnami/thanos/templates/storegateway/service-headless.yaml
Reuuke: It seems to me that a dedicated headless service is redundant.
For both described cases, we can use one headless service.
Anyway, Thanos Querier gets a list of endpoints from the service.

juan131: > It seems to me that a dedicated headless service is redundant.

It's useful when you want to provide both the features below simultaneously:

- A) Provide a "single" access point for users, so they don't need to worry about the specific replica they're accessing. They just access some service hostname/IP and forget about the rest.
- B) Provide a stable network identifier to every replicas so the different micro-services can target a specific replica when needed.

I don't know under what circumstances the behaviour above is desired on Thanos, but keeping these dedicated headless services won't make any harm, and we'll cover that possibility for those users that find it useful.

Reuuke: Yes, you're right - I returned headless service.
I just added new parameter "storegateway.service.additionalHeadless" and "ruler.service.additionalHeadless" to enable/disable this service.
+ adjusted servicemonitor to avoid duplicated metrics.

bitnami/thanos/templates/ruler/service.yaml
```diff
@@ -5,11 +5,15 @@ metadata:
   name: {{ include "thanos.fullname" . }}-ruler
   labels: {{- include "thanos.labels" . | nindent 4 }}
     app.kubernetes.io/component: ruler
+    monitor: "true"
```
juan131, line 8: When using Prometheus (instead of Prometheus Operator), you can manage what endpoints to scrape using the annotation: `prometheus.io/scrape: 'true'`. However, annotations are ignored when using the Prometheus Operator.

We can use a similar naming here for this label: `prometheus-operator/monitor: 'true'`

```suggestion
    {{- if .Values.metrics.enabled .Values.metrics.serviceMonitor.enabled }}
    prometheus-operator/monitor: 'true'
    {{- end }}
```

bitnami/thanos/templates/ruler/servicemonitor.yaml
```diff
@@ -12,6 +12,7 @@ spec:
   selector:
     matchLabels: {{- include "thanos.matchLabels" . | nindent 6 }}
       app.kubernetes.io/component: ruler
+      monitor: "true"
```
juan131, line 15:
```suggestion
      prometheus-operator/monitor: 'true'
```

Reuuke:
> Thanks for changes @Reuuke ! Take a look to this suggestion

Done. StoreGateway servicemonitor also updated to use "prometheus-operator/monitor: 'true'"

juan131:
Wait for CI/CD to test it and update image tags

PR #2270, opened 2020-04-08
[bitnami/influxdb] .Values.initdbScripts does not exist and doesn't create the Config Map properly

rennokki:
I tried to get my head around why it doesn't run my commands from the scriptInitDb variable. It seems like in the `values.yaml` comes with this form:
```yaml
...

influxdb:
  initdbScripts:
    six_mo_retention.txt: |
      create retention policy "6mo" on "app" duration 26w replication 2 default
...
```

The old, changed in this PR code, was looking for this form, although everywhere throughout the code the `.Values.influxdb.*` form used:

```yaml
...

initdbScripts:
  six_mo_retention.txt: |
    create retention policy "6mo" on "app" duration 26w replication 2 default

influxdb:
  ...
```


lelithium:
Stumbled upon the same thing yesterday. Quick fix (while this PR gets merged) is to pass both, like this
```
initdbScripts:
   init.txt: ...
influxdb:
   initdbScripts: true
```

rennokki:
@lelithium 
Yea, same workaround used. 😁 

@miguelaeh 
My bad. I updated the PR.

miguelaeh:
Waiting the CI/CD to update the tags

PR #2291, opened 2020-04-10
[bitnami/kafka] Broaden jsonpath query

CharlieC3:
**Description of the change**

This change broadens the jsonpath query in the `auto-discovery.sh` to return any value it finds, instead of just the LB IP address.

**Benefits**

Prior to this change, the jsonpath query would always try returning the IP address of the load balancer, but not all load balancers have an IP; some have a hostname and no IP (like those created in AWS). This change makes the script compatible with LBs that have an IP or a hostname.

**Possible drawbacks**

Ideally the jsonpath query would return specifically the IP or the hostname if either keys were found, but it does not appear this is configurable using jsonpath, so I opted for the wildcard instead.

**Checklist** 

### DIFF
--- a/bitnami/kafka/Chart.yaml
+++ b/bitnami/kafka/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: kafka
-version: 10.2.0
+version: 10.2.1
 appVersion: 2.4.1
 description: Apache Kafka is a distributed streaming platform.
 keywords:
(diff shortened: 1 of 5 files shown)


CharlieC3:
Thanks for the response @miguelaeh, I've made the requested change to return the IP if it's not empty, but if it's empty to return the hostname. If both are empty, then it will return an empty string just like the original design prior to this PR.

PR #2327, opened 2020-04-15
[bitnami/memcached] Don't apply metrics service if disabled

saada:
**Description of the change**

This breaks pulumi deployments of the chart as it checks for the service to be up and ready

**Benefits**

Allows compatibility with Pulumi as well as prevent needless service creation.

**Possible drawbacks**

None I can think of

**Checklist**


saada:
Yeah I didn't realize I was on the wrong file. Thank you for reviewing!
Please check my latest changes.

PR #2330, opened 2020-04-15
[bitnami/wordpress] give the posibility of persist .htaccess

### LINKED ISSUE #2294: [bitnami/wordpress] Persist .htaccess
**Which chart**: 
bitnami/wordpress

I am managing an infrastructure for a lot of wordpress. Your charts are almost perfect ! Since I would like to let users manage their own wordpress, they need a `.htaccess` file. 

The problem is when the container reboot, `.htaccess` file disappear. Is there a way to persist it ?

I know that all persisted data are in `/bitnami` and I tried to put in it my `.htaccess` file. 
My suggestion is to restore the `.htaccess` to `/opt/bitnami/wordpress/` if it exists in `/bitnami/wordpress/` folder.

I tried to store my `.htaccess` file into `/bitnami/wordpress/` and made a symlink to make it available into `/opt/bitnami/wordpress/` but the problem is the same, my symlink is deleted on container reboot.

**Additional context**

dani8art:
**Description of the change**

 Adding a field for enabling/disabling .htaccess persistence. By default, its value will be `false`, so this functionality will be disabled. When its value is `true` then `/opt/bitnami/wordpress/.htaccess` will be created and persisted.

**Benefits**

Users will be able to add some rules in this file (via some wordpress plugins or bash).

**Possible drawbacks**

N/A

**Applicable issues**

  - fixes #2294
  - fixes https://github.com/bitnami/bitnami-docker-wordpress/issues/235

**Additional information**

N/A

**Checklist** 


marcosbc:
By the way, the Bitnami WordPress container image still does not have the `WORDPRESS_HTACCESS_PERSISTENCE_ENABLED` option. See: https://github.com/bitnami/bitnami-docker-wordpress/blob/master/5/debian-10/rootfs/wordpress-inputs.json

Also, don't forget to document it to the container image's README file.

dani8art:
Waiting for CI/CD

PR #2351, opened 2020-04-17
[bitnami/mongodb] adding local and global namespace overrides for mongodb

### LINKED ISSUE #2243: [bitnami/mongodb] override namespace through values.yaml
**Which chart**:
mongodb

We are trying to put together a chart for shared infrastructure running in each of our environment clusters, it would be great to be able to deploy dependency charts into their own namespaces, so that a single chart could deploy multiple components.

Ideally, I could add configuration such as:

```yaml
mongodb:
  namespaceOverride: mongo
```

to the `values.yaml` in my umbrella chart, to direct the `mongodb` dependency to install into its own namespace. This is a similar approach to the one taken in the `jenkins` and `prometheus` charts in stable:

https://github.com/helm/charts/pull/18986
https://github.com/helm/charts/pull/15202

Without the proposed change, the only alternative I can think of is to install the individual components separately, through some kind of custom script or additional tooling. This would work, of course, but you either lose the description of your environment in a single declarative format, or pay the cost of the additional complexity in purchasing, acquiring or developing and maintaining the extra tooling. It's not much, but if helm can do it already then why bother?

benjamin-wright:
**Description of the change**

As discussed in the issue: #2243

Additionally, as recommended by @mforutan in an earlier PR to stable, added a global namespaceOverride parameter as well, with priority: local -> global -> release

**Benefits**

As discussed in the issue: #2243

TLDR: Allows you to add

```yaml
  mongodb:
    namespaceOverride: mongodb
```

To override the release namespace and deploy the mongodb chart somewhere else.

Adding the global override also caters for the following scenario, where deploying a hierarchy of charts more than one level deep:

```yaml
subchart:
  somethingElse:
    namespaceOverride: ns1
  mongodb:
    namespaceOverride: ns1

otherChart:
  somethingElse:
    namespaceOverride: ns2
  mongodb:
    namespaceOverride: ns2
```

with the global override, this becomes:

```yaml
subchart:
  global:
    namespaceOverride: ns1

otherChart:
  global:
    namespaceOverride: ns2
```

i.e. using the `global` property of a dependency to encapsulate the `namespaceOverride` value, while still passing it down to all its children and not exposing the parent to any implementation details.

**Possible drawbacks**

The global override isn't a universal pattern by a long shot, meaning usage will be limited to those charts that have also embraced the pattern. So, primarily useful within nested charts of your own devising for now, and add any further sub-dependencies with care / update and PR if open-source.

**Applicable issues**

  - fixes #2243

**Additional information**

**Checklist** 


bitnami/mongodb/README.md
```diff
@@ -59,6 +59,8 @@ The following table lists the configurable parameters of the MongoDB chart and t
 | `image.pullPolicy`                                 | Image pull policy                                                                                                                                         | `IfNotPresent`                                           |
 | `image.pullSecrets`                                | Specify docker-registry secret names as an array                                                                                                          | `[]` (does not add image pull secrets to deployed pods)  |
 | `image.debug`                                      | Specify if debug logs should be enabled                                                                                                                   | `false`                                                  |
+| `global.namespaceOverride`                              | String to override the release namespace.                                                       | `nil`                                                    |
```
miguelaeh, line 62: please move this to the top with the rest of `global` values.

bitnami/mongodb/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: mongodb
-version: 7.10.11
```
miguelaeh, line 3: you should bump the minor version as you are adding a new feature

bitnami/mongodb/templates/_helpers.tpl
```diff
@@ -6,6 +6,31 @@ Expand the name of the chart.
 {{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" -}}
 {{- end -}}
 
+{{/*
+Allow the release namespace to be overridden for multi-namespace deployments in combined charts.
+*/}}
+{{- define "mongodb.namespace" -}}
```
miguelaeh, line 12: Following the approach of the rest of globals, the global should have preference over the rest of parameters

bitnami/mongodb/values-production.yaml
```diff
@@ -8,6 +8,12 @@
 #     - myRegistryKeySecretName
 #   storageClass: myStorageClass
 
+## Global namespace override
+## Override the namespace for resource deployed by the chart, but can itself be overridden by the local namespaceOverride
+##
+# global:
```
miguelaeh, line 14: Do not duplicate the `global` section, add the parameter to the already existing one

bitnami/mongodb/templates/_helpers.tpl
benjamin-wright: I can make this change if you want, but is that the most useful behaviour? If the local takes preference then a chart can still choose a specific namespace for its child when its own global namespace is overridden. If you switch the priority round then you force all dependencies of dependencies into a single namespace

mforutan: to fix this issue, I ended up not having local variable for other charts, for example: [mongodb-replicaset](https://github.com/helm/charts/blob/66f56ec2903cfdd45479e4835e58fcfaa5238291/stable/mongodb-replicaset/templates/_helpers.tpl#L80-L93) as global variables still can be set in local values file,

benjamin-wright: Ah, okay, that makes more sense now, thanks 😄

benjamin-wright:
@miguelaeh I think that's all your comment addressed?

benjamin-wright:
@miguelaeh bump!

bitnami/mongodb/templates/_helpers.tpl
```diff
@@ -6,6 +6,29 @@ Expand the name of the chart.
...
+    {{- if .values.global -}}
```
miguelaeh, line 13: This throws an error in our CI/CD as it should be `.Values` instead of`.values`. Note the `V` vs `v`

PR #2372, opened 2020-04-20
bitnami/rabbitMQ

### LINKED ISSUE #2371: better handling of rabbitmq secret
**Which chart**:
rabbitmq latest

Rather than use `existingBlahSecret`, simply don't derp around with the secret at all if we are in an upgrade situation. 

Don't mess with the secret during an upgrade, even if the secret isn't specified with `--set` or in `values`.

hardcode the secret. Yuk!

**Additional context**

tibers:
**Description of the change**
Add one liner to detect that we are upgrading and not touch the secret.

**Benefits**
We don't need to ever know or read the secret to upgrade rabbit, the chart is smart enough not to render the resource if it's an upgrade.

**Possible drawbacks**
May prevent users from rotating secrets via `upgrade` with helm.

**Applicable issues**
  - fixes #2371

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/rabbitmq/Chart.yaml
+++ b/bitnami/rabbitmq/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: rabbitmq
-version: 6.25.4
+version: 6.25.5
 appVersion: 3.8.3
 description: Open source message broker software that implements the Advanced Message Queuing Protocol (AMQP)
 keywords:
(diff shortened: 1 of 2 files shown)


miguelaeh:
Hi @tibers ,
Thank you very much for the PR,
I have been testing it manually, and after installing the chart I perform an upgrade changing the password and the secret is still changed.
Am I missing something here?

tibers:
Are you running the latest helm? 

https://helm.sh/docs/chart_template_guide/builtin_objects/

miguelaeh:
I am using the latest version, actually, I have tested it again to check if I did it in the wrong way but with the same result.
Can you share how did you test it?

miguelaeh:
Hi @tibers ,
There is a way that actually works using `pre-install` hook on the secret, but it has the drawback that the secret will not be part of the relelase.

PR #2379, opened 2020-04-21
[bitnami/jasperreports] Add nodeSelector and tolerations to deployment.

huang-tianwen:
**Description of the change**

Add nodeSelector and tolerations in deployment.

**Benefits**

Jasperreports need lot of CPU and Memory when running compare to other microservices. It might be a good options to put a dedicated server or nodes pool for reporting use.

**Checklist** 

### DIFF
--- a/bitnami/jasperreports/Chart.yaml
+++ b/bitnami/jasperreports/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: jasperreports
-version: 7.0.14
+version: 7.0.15
 appVersion: 7.2.0
 description: The JasperReports server can be used as a stand-alone or embedded reporting
   and BI server that offers web-based reporting, analytic tools and visualization,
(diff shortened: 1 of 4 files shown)


bitnami/jasperreports/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: jasperreports
-version: 7.0.14
+version: 7.0.15
```
miguelaeh, line 3: As you are adding a new feature you should bump the minor version

bitnami/jasperreports/templates/deployment.yaml
```diff
@@ -19,6 +19,12 @@ spec:
         chart: "{{ template "jasperreports.chart" . }}"
         release: {{ .Release.Name | quote }}
     spec:
+      {{- if .Values.nodeSelector }}
+      nodeSelector: {{ toYaml .Values.nodeSelector | nindent 8 }}
+      {{- end -}}
+      {{- with .Values.tolerations }}
```
miguelaeh, line 25: I think this can be done in the same way as the `nodeSelector`

PR #2380, opened 2020-04-21
[bitnami/kafka] Kafka externalAccess fixes.

### LINKED ISSUE #2373: [bitnami/kafka] externalAccess with port other than 19092 fails
**Which chart**:
Kafka: 10.2.3

**Describe the bug**
When the externalAccess.service.port is set to something other than 19092 the kafka-X-external service points to port 19092 even though the script in scripts-configmap.yaml would reconfigure the port that is actually exposed to be the configured value.

Some changes could be made to take this number into consideration but there is a secondary problem, if a port like 443 is used the container fails to start because port 443 is not allowed unless the container is run as root.

**To Reproduce**
helm install kafka bitnami/kafka \
  --set externalAccess.enabled=true \
  --set externalAccess.service.loadBalancerIPs[0]=10.10.10.10 \
  --set externalAccess.service.type=LoadBalancer \
  --set externalAccess.service.port=443

**Expected behavior**
Allow any port, including 443, to appear in externalAccess.service.port.

- Output of `helm version`:

```
3.1
```

```
1.15
```

**Additional context**

I'd be willing to make a PR, but I feel like it would be a pretty large change as I would probably be removing a lot of the EXTERNAL/INTERNAL listeners in favor of a single listener that is used for both and alter the scripts-configmap.yaml to not change the `KAFKA_CFG_LISTENERS`, only the `KAFKA_CFG_ADVERTISED_LISTENERS` after it has gotten the appropriate IP from the loadBalancerIPs list or SHARED_FILE.

kyle-winkelman:
**Description of the change**

externalAccess currently sets up an INTERNAL and an EXTERNAL listener. This is unnecessary because you can use the same listener for both use cases and forward externalAccess.service.port to the service.port.

**Benefits**

Simplifies the chart and is able to handle externalAccess.service.ports in restricted values (i.e. 443).

**Possible drawbacks**

**Applicable issues**

  - fixes #2373

**Additional information**

**Checklist** 


bitnami/kafka/templates/svc-external-access.yaml
```diff
@@ -19,28 +19,23 @@ metadata:
...
+  {{- if $root.Values.externalAccess.service.loadBalancerIPs }}
   loadBalancerIP: {{ index $root.Values.externalAccess.service.loadBalancerIPs $i }}
   {{- end }}
   {{- if $root.Values.externalAccess.service.loadBalancerSourceRanges }}
   loadBalancerSourceRanges: {{- toYaml $root.Values.service.loadBalancerSourceRanges | nindent 4 }}
   {{- end }}
   {{- end }}
   ports:
-    - name: tcp-kafka
+    - name: kafka
```
juan131, line 30: The new name doesn't follow [Istio standards](https://istio.io/docs/ops/deployment/requirements/), please revert the change to coninue using `tcp-kafka`.

bitnami/kafka/values-production.yaml
```diff
@@ -385,19 +385,14 @@ service:
...
   ##
-  nodePorts:
-    kafka: ""
-    ssl: ""
+  nodePort:
   ## Set the LoadBalancer service type to internal only.
   ## ref: https://kubernetes.io/docs/concepts/services-networking/service/#internal-load-balancer
   ##
-  # loadBalancerIP:
+  loadBalancerIP:
```
juan131, line 395: Please revert this change too. We faced some issues in the past with parameters with no value associated.
To avoid problems, it's better to have them commented.

bitnami/kafka/templates/svc.yaml
```diff
@@ -18,25 +18,11 @@ spec:
...
+    - name: kafka
       port: {{ .Values.service.port }}
-      protocol: TCP
       targetPort: kafka
-      {{- if and (or (eq .Values.service.type "NodePort") (eq .Values.service.type "LoadBalancer")) (not (empty .Values.service.nodePorts.kafka)) }}
-      nodePort: {{ .Values.service.nodePorts.kafka }}
-      {{- else if eq .Values.service.type "ClusterIP" }}
-      nodePort: null
+      {{- if eq .Values.service.type "NodePort" }}
+      nodePort: {{ .Values.service.nodePort }}
```
juan131, line 25: When you use a "LoadBalancer" service, a "nodePort" is also created (and associated to the external IP obtained via load balancer). Therefore, this should be:
```
    {{- if or (eq .Values.service.type "NodePort") (eq .Values.service.type "LoadBalancer") }}
```
Then, you're assuming that users will set the `.Values.service.nodePort` and that won't be true on many cases, so you have two alternatives:
- Generate a random `nodePort` when it's not specified (which implies reverting your changes)
- Make it mandatory, which I don't think it's necessary:

```
    nodePort: {{ required "A nodePort is mandatory!" .Values.service.nodePort }}
```

juan131: Please do not remove this block:
```
      {{- else if eq .Values.service.type "ClusterIP" }}
      nodePort: null
```

bitnami/kafka/templates/_helpers.tpl
```diff
@@ -276,6 +276,21 @@ but Helm 2.9 and 2.10 does not support it, so we need to implement this if-else
...
+{{- if and .Values.auth.ssl .Values.auth.enabled -}}
+SASL_SSL
+{{- else if and .Values.auth.ssl (not .Values.auth.enabled) -}}
+SSL
+{{- else if and .Values.auth.enabled (not .Values.auth.ssl) -}}
+SASL_PLAINTEXT
+{{- else -}}
+PLAINTEXT
+{{- end -}}
+{{- end -}}
```
juan131, line 292: Nice! I like this

bitnami/kafka/templates/NOTES.txt
```diff
@@ -116,9 +116,9 @@ To connect to your Kafka server from outside the cluster, follow the instruction
         export KAFKA_OPTS="-Djava.security.auth.login.config=/path/to/kafka_jaas.conf"
 {{- else }}
     PRODUCER:
-        kafka-console-producer.sh --broker-list 127.0.0.1:9092 --topic test
+        kafka-console-producer.sh --broker-list 127.0.0.1:{{ .Values.service.port }} --topic test
     CONSUMER:
-        kafka-console-consumer.sh --bootstrap-server 127.0.0.1:9092 --topic test --from-beginning
+        kafka-console-consumer.sh --bootstrap-server 127.0.0.1:{{ .Values.service.port }} --topic test --from-beginning
```
juan131, line 121: Good catch!

bitnami/kafka/templates/svc.yaml
kyle-winkelman: I didn't know about LoadBalancers getting nodePorts. Two questions:

- Why aren't we following the same approach for svc-external-access.yaml by allowing them to pass nodePorts as well as loadBalancerIPs? Or the opposite question why don't we set nodePort to null in svc.yaml (like we do in svc-external-access.yaml) when it is of type LoadBalancer?
- Why is the `nodePort: null` required for type `ClusterIP` isn't that the default?

bitnami/kafka/templates/statefulset.yaml
```diff
@@ -139,45 +139,18 @@ spec:
...
-              value: "INTERNAL:SASL_SSL,EXTERNAL:SASL_SSL"
-              {{- else if and .Values.auth.ssl (not .Values.auth.enabled) }}
-              value: "INTERNAL:SSL,EXTERNAL:SSL"
-              {{- else if and .Values.auth.enabled (not .Values.auth.ssl) }}
-              value: "INTERNAL:SASL_PLAINTEXT,EXTERNAL:SASL_PLAINTEXT"
-              {{- else }}
-              value: "INTERNAL:PLAINTEXT,EXTERNAL:PLAINTEXT"
-              {{- end }}
-            {{- else }}
             - name: KAFKA_CFG_LISTENERS
```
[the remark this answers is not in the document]
juan131: I agree, then let's configure INTERNAL listeners to PLAINTEXT so we have different ways to access the Kafka brokers, and we can use the internal one for probes.

bitnami/kafka/templates/statefulset.yaml
```diff
@@ -139,45 +139,18 @@ spec:
...
-              value: "SASL_SSL://$(MY_POD_NAME).{{ template "kafka.fullname" . }}-headless.{{.Release.Namespace}}.svc.{{ .Values.clusterDomain }}:$(KAFKA_PORT_NUMBER),SSL://$(MY_POD_NAME).{{ template "kafka.fullname" . }}-headless.{{.Release.Namespace}}.svc.{{ .Values.clusterDomain }}:9093"
-              {{- else if and .Values.auth.ssl (not .Values.auth.enabled) }}
-              value: "SSL://$(MY_POD_NAME).{{ template "kafka.fullname" . }}-headless.{{.Release.Namespace}}.svc.{{ .Values.clusterDomain }}:9093"
-              {{- else if and .Values.auth.enabled (not .Values.auth.ssl) }}
-              value: "SASL_PLAINTEXT://$(MY_POD_NAME).{{ template "kafka.fullname" . }}-headless.{{.Release.Namespace}}.svc.{{ .Values.clusterDomain }}:$(KAFKA_PORT_NUMBER)"
               {{- else }}
-              value: "PLAINTEXT://$(MY_POD_NAME).{{ template "kafka.fullname" . }}-headless.{{.Release.Namespace}}.svc.{{ .Values.clusterDomain }}:$(KAFKA_PORT_NUMBER)"
+              value: "{{ template "kafka.listenerType" . }}://$(MY_POD_NAME).{{ template "kafka.fullname" . }}-headless.{{ .Release.Namespace }}.svc.{{ .Values.clusterDomain }}:$(KAFKA_PORT_NUMBER)"
               {{- end }}
```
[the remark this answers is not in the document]
juan131: And how do you plant to manage both ports when using `SASL_SSL`?

bitnami/kafka/templates/svc.yaml
juan131: The `nodePort: null` is related to some weird behaviour reported by some users.
Regarding the "external-service", you need to chose whether configuring Kafka advertised listeners based on the nodeports or based on the external IPs (associated to LoadBalancer svc), that's why we only allow setting one of them.

kyle-winkelman:
@juan131, I have concerns with configuring INTERNAL only for the case of externalAccess.enabled. Why is it that for all the other cases we would not expose a PLAINTEXT port as well for the same reason?  I feel that if someone is interested in exposing two different listeners to support more advanced probes that is something they should do themselves (via listeners and advertisedListeners).

bitnami/kafka/templates/statefulset.yaml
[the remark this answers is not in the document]
kyle-winkelman: I don't believe we need the second port for `SASL_SSL`. Can you actually connect to the `SSL` port and do anything meaningful if you don't supply the necessary authentication for SASL?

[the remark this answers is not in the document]
juan131: You can have complex scenarios (see https://docs.confluent.io/current/security/security_tutorial.html#configure-brokers) where you have: 

- All broker/client communication use **SASL_SSL** security protocol, which ensures that the communication is encrypted and authenticated using **SASL/PLAIN**.
- All inter-broker communication use **SSL** security protocol, which ensures that the communication is encrypted and authenticated using SSL.

Therefore you need both ports to provide different authentication methods: one for client, and another for inter-broker communications.

juan131:
Sorry I totally lose the track of this PR!!

Theres's a comment I added:

> You can have complex scenarios (see https://docs.confluent.io/current/security/security_tutorial.html#configure-brokers) where you have:
>  
> - All broker/client communication use SASL_SSL security protocol, which ensures that the communication is encrypted and authenticated using SASL/PLAIN.
> - All inter-broker communication use SSL security protocol, which ensures that the communication is encrypted and authenticated using SSL.
> 
> Therefore you need both ports to provide different authentication methods: one for client, and another for inter-broker communications.

Did you see it? I have my concerns about exposing a single port in the Kafka container.

kyle-winkelman:
I understand that you can have many listeners for many different reasons, but this chart really only uses 1 port the whole time. Even in the case of SASL_SSL when used with external you end up with `INTERNAL:SASL_SSL,EXTERNAL:SASL_SSL`. If users of the chart want to use multiple listeners, they can use listeners and advertisedListeners Values to configure it themselves.

bitnami/kafka/templates/svc.yaml
```diff
@@ -22,21 +22,10 @@ spec:
...
-    - name: tls-kafka
-      port: {{ .Values.service.sslPort }}
-      protocol: TCP
-      targetPort: kafka-ssl
-      {{- if and (or (eq .Values.service.type "NodePort") (eq .Values.service.type "LoadBalancer")) (not (empty .Values.service.nodePorts.ssl)) }}
-      nodePort: {{ .Values.service.nodePorts.ssl }}
-      {{- else if eq .Values.service.type "ClusterIP" }}
-      nodePort: null
-      {{- end }}
-    {{- end }}
```
juan131, line 40: This doesn't convince me yet. I understand your point and I agree with you that "complex configurations" should be addressed by customising the `listeners` and `advertisedlisteners` parameters.
That said, with these changes I can only expose one port. This is fine for the "default behaviour" but it's not valid for the "complex configurations" where I can have use `SASL_SSL` for external access and `SSL` for inter-broker communication. In that case I need to expose 2 ports.
So.. if you want to deprecate the `service.sslPort` parameter (I"m fine with it) we should add an alternative (e.g. sth like `service.extraPorts`) to add more ports to expose when needed.

juan131:
I think this PR doesn't apply anymore now that we released a new major version refactoring the way to configure listeners, advertised listeners and authentication in Kafka. Now we support having different authentication methods on different listeners.

Find more information in the link below:

- https://github.com/bitnami/charts/tree/master/bitnami/kafka#to-1100

PR #2388, opened 2020-04-22
 [bitnami/mariadb-galera]Rename the enviroment vars used for TLS accordingly with the container

rafariossaa:
Description of the change
Renames the parameter for TLS usage, they are general and not only replication traffic related.
Also documents in the REAMDE how to use TLS in the client

Benefits
Use TLS for replication and clients

Possible drawbacks
Parameter renaming of the previous version

Checklist

Chart version bumped in Chart.yaml according to semver.
Variables are documented in the README.md
Title of the PR starts with chart name (e.g. [bitnami/chart])
If the chart contains a values-production.yaml apart from values.yaml, ensure that you implement the changes in both files

### DIFF
--- a/bitnami/mariadb-galera/Chart.yaml
+++ b/bitnami/mariadb-galera/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: mariadb-galera
-version: 1.1.3
+version: 2.0.0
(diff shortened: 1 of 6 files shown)


carrodher:
Reviewed in this one #2388

PR #2398, opened 2020-04-23
[bitnami/mongodb] add nodeSelector option for secondary statefulset

golgoth31:
**Description of the change**

This change allows the chart to push secondary and primary/arbiter on different nodeSelector by adding a new "nodeSelectorSecondary" value entry.

**Benefits**

On cloud platform, this can improve cluster resiliency by using different regions/availability zones.

**Possible drawbacks**

n/a

**Applicable issues**

n/a

**Additional information**

n/a

**Checklist** 


caiohasouza:
Thank you!

golgoth31:
Pleasure :)
My 2cts over your huge job on those charts

caiohasouza:
Guys, there is a problem. The secondary use a PVC (arbiter not), if i chose use a different aws zone of primary i need create a PVC in another zone, then we can create a persistence options similary of primary but to secondary, do you understand?

golgoth31:
This problem could happen but in some limited conditions:
- you update an existing release, your primary/secondary are in the same zone (az for AWS) and the nodeselector you want to use are located in different zones
- your storage class is not configured as "volumeBindingMode: WaitForFirstConsumer" and the scheduler creates the pv in an other zone before scheduling the pod in an other region
- you have manually created the pv before deploying this chart and you fall in the case 1 described

In any other cases, the scheduler will create a new pvc for the secondary (as it does for the primary) without any problem and the secondary will use it.
I think its also better to have one uniq storage size definition for primary and secondary.
Is it ok for you?

caiohasouza:
Hm, i have 3 storageclass with regions in your name, example:

- gp2-us-east-1a
- gp2-us-east-1c
- gp2-us-east-1d

For this reasons that i need a possibility that configure differente persistence volumes for primary/secondary, do you understand? Like this:

> mongodb
>   persistence:
>     enabled: true
>     storageClass: gp2-us-east-1a
>     accessMode: ReadWriteOnce
>     size: 10Gi
>   persistenceSecondary:
>     enabled: true
>     storageClass: gp2-us-east-1c
>     accessMode: ReadWriteOnce
>     size: 10Gi

And nodeSelectors:

>   nodeSelector:
>     environment: "stage-backend-us-east-1a"
>   nodeSelectorSecondary:
>     environment: "stage-backend-us-east-1c"
>   nodeSelectorArbiter:
>     environment: "stage-backend-us-east-1d"

With this approach i will have a primary on us-east-1a region, secondary on us-east-1c and arbiter on us-east-1d, a great MongoDB HA solution.

What do you think?

golgoth31:
Ok, I get your point.
I think it would be safer to only allow to change storageclass as all other option should remain the same (particularly the size); having multiple persistence entries will let users able to deploy primary and secondary with different storage size.
Perhaps, you can create an issue and I could work on it in the next few days

caiohasouza:
Ok, perfect! Issue created, thank you so much!

miguelaeh:
Thank you for notice about that, we will take care of the new issue, let's move the conversation there.

Issue #2378, opened 2020-04-21
[Postgres-HA] : Issue in restarting Postgres cluster


abhi7788:
**Which chart**:
Postgres-HA

**Describe the bug**
Unable to run the postgres database nodes after scaling down the stateful set to zero and then rescaling it back to 2

**To Reproduce**
1. Setup the cluster via helm chart
2. scale down the postgres statefulset to zero and pgpool to zero
3. scale back postgres statefulset to 2 and pgpool to 1
4. Statefulset goes into crashbackloop with following error:
Readiness probe failed: psql: could not connect to server: Connection refused Is the server running on host "127.0.0.1" and accepting TCP/IP connections on port 5432?
Back-off restarting failed container

**Additional context**
Detailed logs from the pod :

repmgr 17:26:07.14 INFO  ==> There are no nodes with primary role. Assuming the primary role...
repmgr 17:26:07.33 INFO  ==> Preparing PostgreSQL configuration...
postgresql 17:26:07.44 INFO  ==> postgresql.conf file not detected. Generating it...
repmgr 17:26:08.83 INFO  ==> Preparing repmgr configuration...
repmgr 17:26:08.93 INFO  ==> Initializing Repmgr...
postgresql 17:26:08.94 INFO  ==> Initializing PostgreSQL database...
postgresql 17:26:09.03 INFO  ==> Cleaning stale /bitnami/postgresql/data/postmaster.pid file
postgresql 17:26:09.23 INFO  ==> Custom configuration /opt/bitnami/postgresql/conf/postgresql.conf detected
postgresql 17:26:09.24 INFO  ==> Custom configuration /opt/bitnami/postgresql/conf/pg_hba.conf detected
postgresql 17:26:09.73 INFO  ==> Deploying PostgreSQL with persisted data...
postgresql 17:26:10.04 INFO  ==> Stopping PostgreSQL...
postgresql-repmgr 17:26:10.23 INFO  ==> ** PostgreSQL with Replication Manager setup finished! **
postgresql 17:26:11.94 INFO  ==> Starting PostgreSQL in background...
postgresql-repmgr 17:26:15.43 INFO  ==> ** Starting repmgrd **
[2020-04-21 17:26:15] [NOTICE] repmgrd (repmgrd 5.1.0) starting up
INFO:  set_repmgrd_pid(): provided pidfile is /opt/bitnami/repmgr/tmp/repmgr.pid
[...]

Issue #2193, opened 2020-04-02
[bitnami/postgresql-ha] HA does not survive worker node restart


ashokspeelyaal:
**Which chart**:
postgresql-ha-1.4.4

**Describe the bug**
I have an ubuntu worker node where both the pods( 0 and 1) were running. 
The node got restarted and see both the pods status remain CrashLoopBackOff. 
I do not see any issue in pod logs. 

From the logs pod-0: 
repmgr 04:56:39.99 INFO  ==> Preparing repmgr configuration...
repmgr 04:56:40.02 INFO  ==> Initializing Repmgr...
postgresql 04:56:40.03 INFO  ==> Initializing PostgreSQL database...
postgresql 04:56:40.12 INFO  ==> Custom configuration /opt/bitnami/postgresql/conf/postgresql.conf detected
postgresql 04:56:40.13 INFO  ==> Custom configuration /opt/bitnami/postgresql/conf/pg_hba.conf detected
postgresql 04:56:40.16 INFO  ==> Deploying PostgreSQL with persisted data...
postgresql 04:56:40.19 INFO  ==> Stopping PostgreSQL...
postgresql-repmgr 04:56:40.20 INFO  ==> ** PostgreSQL with Replication Manager setup finished! **

postgresql 04:56:40.35 INFO  ==> Starting PostgreSQL in background...
From the logs pod-1: 
postgresql-repmgr 04:54:00.19 INFO  ==> ** Starting PostgreSQL with Replication Manager setup **
repmgr 04:54:00.25 INFO  ==> Validating settings in REPMGR_* env vars...
postgresql 04:54:00.25 INFO  ==> Validating settings in POSTGRESQL_* env vars..
repmgr 04:54:00.26 INFO  ==> Querying all partner nodes for common upstream node...
repmgr 04:54:00.29 INFO  ==> Preparing PostgreSQL configuration...
postgresql 04:54:00.30 INFO  ==> postgresql.conf file not detected. Generating it...
repmgr 04:54:00.36 INFO  ==> Preparing repmgr configuration...
repmgr 04:54:00.38 INFO  ==> Initializing Repmgr...
repmgr 04:54:00.38 INFO  ==> Waiting for primary node...
postgresql 04:55:10.65 INFO  ==> Stopping PostgreSQL...

**To Reproduce**
1. create with default configurations
2. restart one of the worker nodes(with primary running)
3. all other pods are coming back to running stage
4. Except, postgresql pods go to status CrashLoopBackOff

[...]

Issue #1934, opened 2020-02-17
postgresql-ha is down size error


truyet:
**Which chart**: bitnami/postgresql-ha  version 1.4.3, app version 11.6.0

**Description**
Pods crash when it's restarted with error:
020-02-17 02:30:23.745 GMT [221] FATAL:  could not connect to the primary server: could not translate host name "postgresql-postgresql-ha-postgresql-1.postgresql-postgresql-ha-postgresql-headless.postgresql.svc.cluster.local" to address: Name or service not known
postgresql-repmgr 02:30:23.75 INFO  ==> ** Starting repmgrd **
2020-02-17 02:30:23.774 GMT [225] FATAL:  could not connect to the primary server: could not translate host name "postgresql-postgresql-ha-postgresql-1.postgresql-postgresql-ha-postgresql-headless.postgresql.svc.cluster.local" to address: Name or service not known
[2020-02-17 02:30:23] [NOTICE] repmgrd (repmgrd 5.0.0) starting up
INFO:  set_repmgrd_pid(): provided pidfile is /opt/bitnami/repmgr/tmp/repmgr.pid
[2020-02-17 02:30:23] [NOTICE] starting monitoring of node "postgresql-postgresql-ha-postgresql-0" (ID: 1000)
[2020-02-17 02:30:23] [ERROR] connection to database failed
[2020-02-17 02:30:23] [DETAIL]
could not translate host name "postgresql-postgresql-ha-postgresql-1.postgresql-postgresql-ha-postgresql-headless.postgresql.svc.cluster.local" to address: Name or service not known

[2020-02-17 02:30:23] [DETAIL] attempted to connect using:
  user=repmgr password=xxx connect_timeout=5 dbname=repmgr host=postgresql-postgresql-ha-postgresql-1.postgresql-postgresql-ha-postgresql-headless.postgresql.svc.cluster.local port=5432 fallback_application_name=repmgr
[2020-02-17 02:30:23] [ERROR] unable connect to upstream node (ID: 1001), terminating
[2020-02-17 02:30:23] [HINT] upstream node must be running before repmgrd can start

**Steps to reproduce the issue:**

1. Setup postgresql ha with replica = 2
2. Resize replica to 1
3. When postgresql is running only 1 pod, try to delete this pods, it will recreate new pod and crash forever

**Describe the results you received:**

NAME                                               READY   STATUS             RESTARTS   AGE
[...]

PR #2406, opened 2020-04-23
[bitnami/postgresql-ha] add graceful exit for primary nodes

### LINKED ISSUE #2033: [postgres-ha] unable to recover when primary pod is other than 0
**Which chart**:

name: postgresql-ha
version: 1.4.8

**Description**

I found an interesting scenario where `postgres-ha` is unable to start/recover after update of helm chart with modified `replicaCount`.

It happens to me during failover tests of latest chart version. First I delete manually primary pod so that `repmgr` promote another replica to be the primary pod like `postgresql-ha-postgresql-1`.

:thinking: :thought_balloon: Thinking that we may need another replica that serves as a witness in such conditions,

OR being able to promote itself as primary if no other postgres pods are running.

This issue https://github.com/bitnami/charts/issues/1934 seems same behavior to me but it was fixed by bumping pg-pool image version.

Thanks for all the effort here. This chart could be splendid!

1. Start `postgres-ha` chart with `2 replicas` using following `values.yaml`

```
---
  postgresql:
    username: postgres
    password: superuserpass
    database: app
    repmgrUsername: repmgr
    repmgrPassword: repmgrpass
    replicaCount: 2
  postgresqlImage:
    debug: true

  pgpool:
    adminPassword: pgpoolpass
    replicaCount: 2
```

2. Wait for pods to started

```
kubectl rollout status sts --namespace=postgresql-ha postgresql-ha-postgresql
```

3. Manually delete `postgresql-ha-postgresql-0` which has the primary role

```
kubectl delete po --namespace=postgresql-ha postgresql-ha-postgresql-0
```

4. Wait for deleted pod to be recreated

```
kubectl rollout status sts --namespace=postgresql-ha postgresql-ha-postgresql
```

5. Check which pod has the primary role using `repmgr`. I login to `postgresql-ha-postgresql-0` as user `postgres` which I manually created in order to run `repmgr` commands inside this non-root container.

```
postgres@postgresql-ha-postgresql-0:/$ repmgr cluster show -f /opt/bitnami/repmgr/conf/repmgr.conf --compact
 ID   | Name                       | Role    | Status    | Upstream                   | Location | Prio. | TLI
------+----------------------------+---------+-----------+----------------------------+----------+-------+-----
 1000 | postgresql-ha-postgresql-0 | standby |   running | postgresql-ha-postgresql-1 | default  | 100   | 2
 1001 | postgresql-ha-postgresql-1 | primary | * running |                            | default  | 100   | 2
```

6. Update helm chart to use `1 replica only`. Kubernetes first delete pod with index 1, then delete pod with index 0, and when it starts the new pod with index 0 it fails to recover forever with following message

```
[2020-03-11 12:06:18] [DETAIL] attempted to connect using:
user=repmgr password=repmgrpass connect_timeout=5 dbname=repmgr host=postgresql-ha-postgresql-1.postgresql-ha-postgresql-headless.postgresql-ha.svc.cluster.local port=5432 fallback_application_name=repmgr
[2020-03-11 12:06:18] [ERROR] unable connect to upstream node (ID: 1001), terminating
[2020-03-11 12:06:18] [HINT] upstream node must be running before repmgrd can start
```

At this moment, I think that repmgr on pod-0 is looking for pod-1 which was the last promoted primary, but since pod-1 does not exist, there is a missing logic for pod-0 to promote itself as primary.

After scale out, the postgresql pod is continuously restarting without opportunity to promote itself as primary. Server is unreachable and I was unable to recover even using `existingClaim` or `pdb: true`

The postgresql server to remain in working condition with pod-0 as primary role.

- Output of `helm version`:

```
Helm 2 using Rancher v2.3.5
```

```
Client Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.3", GitCommit:"06ad960bfd03b39c8310aaf92d1e7c12ce618213", GitTreeState:"clean", BuildDate:"2020-02-11T18:14:22Z", GoVersion:"go1.13.6", Compiler:"gc", Platform:"linux/amd64"}

```

dani8art:
**Description of the change**

This PR adds a graceful exit mechanism for primary nodes in the postgres cluster. Now when a primary node tries to leave the cluster it will wait until any other node promotes to the primary node. 

**Benefits**

Before these changes, when the primary was not the pod 0 (e.g: in a cluster with 3 pods the primary node was pod 1) scale down action bring the cluster to an unrecovered state. Now, We can scale down when the primary node is not the pod 0.

**Possible drawbacks**

N/A

**Applicable issues**

  - fixes #2033 #1934 #2193 #2378

**Additional information**

Steps to reproduce and test.

1. Install the chart

```bash
$ helm install pg --set postgresql.replicaCount=2 --set pgpool.replicaCount=1 \
  bitnami/postgresql-ha
```
2. For a change in the primary role

```bash
$ kubectl delete pod pg-postgresql-ha-postgresql-0
```
3. Wait until new primary is elected

```bash
$ kubectl rollout status sts postgresql-ha-postgresql
```
4. Get data to update the cluster size

```bash
$ export PGPOOL_PASSWORD=$(kubectl get secret pg-postgresql-ha-pgpool -o jsonpath="{.data.admin-password}" | base64 --decode)
$ export POSTGRES_PASSWORD=$(kubectl get secret pg-postgresql-ha-postgresql -o jsonpath="{.data.postgresql-password}" | base64 --decode)
$ export REPMGR_PASSWORD=$(kubectl get secret pg-postgresql-ha-postgresql -o jsonpath="{.data.repmgr-password}" | base64 --decode)
```

5. Scale down the cluster

```bash
$ helm upgrade pg \
  --set postgresql.replicaCount=1 \
  --set postgresql.password="$POSTGRES_PASSWORD" \
  --set postgresql.repmgrPassword="$REPMGR_PASSWORD" \
  --set pgpool.replicaCount=1 \
  --set pgpool.adminPassword="$PGPOOL_PASSWORD" \
  bitnami/postgresql-ha
```

**Checklist** 

### DIFF
--- a/bitnami/postgresql-ha/Chart.yaml
+++ b/bitnami/postgresql-ha/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: postgresql-ha
-version: 3.1.1
+version: 3.2.0
 appVersion: 11.7.0
 description: Chart for PostgreSQL with HA architecture (using Replication Manager (repmgr) and Pgpool).
 keywords:
--- a/bitnami/postgresql-ha/templates/postgresql/hooks-scripts-configmap.yaml
+++ b/bitnami/postgresql-ha/templates/postgresql/hooks-scripts-configmap.yaml
@@ -1,0 +1,67 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ printf "%s-hooks-scripts" (include "postgresql-ha.postgresql" .) }}
+  labels: {{- include "postgresql-ha.labels" . | nindent 4 }}
+    app.kubernetes.io/component: postgresql
+data:
+  pre-stop.sh: |-
+    #!/bin/bash
+    set -o errexit
+    set -o pipefail
+    set -o nounset
+
+    # Debug section
+    exec 3>&1
+    exec 4>&2
+
+    # Load Libraries
+    . /opt/bitnami/scripts/liblog.sh
+    . /opt/bitnami/scripts/libpostgresql.sh
+    . /opt/bitnami/scripts/librepmgr.sh
+
+    # Auxiliar functions
+    is_new_primary_ready() {
+        return_value=1
+        currenty_primary_node="$(repmgr_get_primary_node)"
+        currenty_primary_host="$(echo $currenty_primary_node | awk '{print $1}')"
+
+        info "$currenty_primary_host != $REPMGR_NODE_NETWORK_NAME"
+        if [[ $(echo $currenty_primary_node | wc -w) -eq 2 ]] && [[ "$currenty_primary_host" != "$REPMGR_NODE_NETWORK_NAME" ]]; then
+            info "New primary detected, leaving the cluster..."
+            return_value=0
+        else
+            info "Waiting for a new primary to be available..."
+        fi
+        return $return_value
+    }
+
+    export MODULE="pre-stop-hook"
+
+    if [[ "${BITNAMI_DEBUG}" == "true" ]]; then
+        info "Bash debug is on"
+    else
+        info "Bash debug is off"
+        exec 1>/dev/null
+        exec 2>/dev/null
+    fi
+
+    # Load PostgreSQL & repmgr environment variables
+    eval "$(repmgr_env)"
+    eval "$(postgresql_env)"
+
+    postgresql_enable_nss_wrapper
+
+    # Prepare env vars for managing roles
+    primary_node="$(repmgr_get_primary_node)"
+    primary_host="$(echo $primary_node | awk '{print $1}')"
+
+    # Stop postgresql for graceful exit.
+    postgresql_stop
+
+    if [[ "$primary_host" == "$REPMGR_NODE_NETWORK_NAME" ]]; then
+        info "Primary node need to wait for a new primary node before leaving the cluster"
+        retry_while is_new_primary_ready 10 5
+    else
+        info "Standby node doesn't need to wait, leaving the cluster."
+    fi
--- a/bitnami/postgresql-ha/templates/postgresql/statefulset.yaml
+++ b/bitnami/postgresql-ha/templates/postgresql/statefulset.yaml
@@ -68,6 +68,11 @@ spec:
       {{- end }}
       containers:
         - name: postgresql
+          lifecycle:
+            preStop:
+              exec:
+                command:
+                  - /pre-stop.sh
           image: {{ include "postgresql-ha.postgresqlImage" . }}
           imagePullPolicy: {{ .Values.postgresqlImage.pullPolicy | quote }}
           {{- if .Values.postgresql.securityContext.enabled }}
@@ -205,6 +210,9 @@ spec:
             - name: data
               mountPath: {{ .Values.persistence.mountPath }}
             {{- end }}
+            - name: hooks-scripts
+              mountPath: /pre-stop.sh
+              subPath: pre-stop.sh
         {{- if .Values.metrics.enabled }}
         - name: metrics
           image: {{ include "postgresql-ha.metricsImage" . }}
@@ -264,6 +272,10 @@ spec:
             {{- end }}
         {{- end }}
       volumes:
+        - name: hooks-scripts
+          configMap:
+            name: {{ printf "%s-hooks-scripts" (include "postgresql-ha.postgresql" .) }}
+            defaultMode: 0755
         {{- if or (.Files.Glob "files/repmgr.conf") (.Files.Glob "files/postgresql.conf") (.Files.Glob "files/pg_hba.conf") .Values.postgresql.repmgrConfiguration .Values.postgresql.configuration .Values.postgresql.pgHbaConfiguration .Values.postgresql.configurationCM }}
         - name: postgresql-config
           configMap:
--- a/bitnami/postgresql-ha/values-production.yaml
+++ b/bitnami/postgresql-ha/values-production.yaml
@@ -244,7 +244,6 @@ postgresql:
   ## ref: https://github.com/bitnami/bitnami-docker-postgresql-repmgr#configuration-file
   ##
   # repmgrConfiguration: |-
-
   ## PostgreSQL configuration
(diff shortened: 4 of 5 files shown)


stv0g:
Thanks for working on this :) I am ready to test it 👍 

Just one comment: wouldnt it make more sense to include the preStop script into the image like all the other bitnami shell scripts?

bitnami/postgresql-ha/templates/postgresql/scripts-configmap.yaml
```diff
@@ -0,0 +1,63 @@
...
+
+    set -o errexit
+    set -o pipefail
+    set -o nounset
+
+    # Debug section
+    exec 3>&1
+    exec 4>&2
+
+    if [ "${BITNAMI_DEBUG}" = true ]; then
```
juan131, line 19: Use: `[[ ... ]]` syntax (to be consistent)

bitnami/postgresql-ha/templates/postgresql/scripts-configmap.yaml
```diff
@@ -0,0 +1,63 @@
...
+
+    # Stop postgresql for graceful exit.
+    postgresql_stop
+
+    if [[ "$REPMGR_CURRENT_PRIMARY_HOST" == "$REPMGR_NODE_NETWORK_NAME" ]]; then
+      echo "Primary node need to wait for a new primary node before leaving the cluster"
+
+      wait="1"
+      while [ "$wait" -eq "1" ]
+      do
```
juan131, line 48: Use `while [[ ]]; do` syntax

bitnami/postgresql-ha/templates/postgresql/scripts-configmap.yaml
```diff
@@ -0,0 +1,63 @@
...
+    # Debug section
+    exec 3>&1
+    exec 4>&2
+
+    if [ "${BITNAMI_DEBUG}" = true ]; then
+      echo "==> Bash debug is on"
+    else
+      echo "==> Bash debug is off"
+      exec 1>/dev/null
+      exec 2>/dev/null
```
juan131, line 24: Please use 4 spaces indentation (apply this to the whole script)

bitnami/postgresql-ha/templates/postgresql/scripts-configmap.yaml
juan131: I would borrow the `retry_while` function from https://github.com/bitnami/bitnami-docker-postgresql-repmgr/blob/master/11/debian-10/prebuildfs/opt/bitnami/scripts/libos.sh
It's more elegant

bitnami/postgresql-ha/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: postgresql-ha
-version: 3.0.1
+version: 3.0.2
```
juan131, line 3: I'd increase the minor version. This is quite a feature to be considered just a pach

bitnami/postgresql-ha/templates/postgresql/scripts-configmap.yaml
dani8art: I did a try using `retry_while` but it doesn't fit at all, we must wait for a condition, not for command execution. Also, I had to do sth like:
```
am_i_primary() {
  # use some libraries
  # check condition
}

retry_while "am_i_primary" 10 2
```
When using it inside `retry_while` the scope change and library functions don't work properly.

dani8art:
> Thanks for working on this :) I am ready to test it 👍
> 
> Just one comment: wouldnt it make more sense to include the preStop script into the image like all the other bitnami shell scripts?

Controlling when a container is going to leave the cluster just into the image is sth that would take a huge effort, even I don't know it was possible at all. We are going to take the advantages that working in a k8s environment gives us to solve this scaling problem.

bitnami/postgresql-ha/templates/postgresql/scripts-configmap.yaml
juan131: You can wait for a condition using `retry_while`, you just create an auxiliary function, sth like `is_new_primary_ready` that returns "0/1" based on that condition. Then you can use:
```
    retry_while "is_new_primary_ready " 10 2
```
The scope shouldn't be an issue either. You should be able to access the same env. variables in the auxiliary function.

dani8art: I've tried again but it doesn't work

```
    is_new_primary_ready() {
        return_value=1
        eval "$(repmgr_set_role)"

        info "$REPMGR_CURRENT_PRIMARY_HOST == $REPMGR_NODE_NETWORK_NAME"
        if [[ "$REPMGR_CURRENT_PRIMARY_HOST" != "$REPMGR_NODE_NETWORK_NAME" ]]; then
            info "New primary detected, Leaving the cluster"
            return_value=0
        else
            info "I'm still primary node, waiting for a new primary"
        fi

        return $return_value
    }

    if [[ "$REPMGR_CURRENT_PRIMARY_HOST" == "$REPMGR_NODE_NETWORK_NAME" ]]; then
        info "Primary node need to wait for a new primary node before leaving the cluster"

        retry_while "is_new_primary_ready " 10 2
    else
        info "Standby node doesn't need to wait, leaving the cluster."
    fi
```
I would say this `repmgr_set_role` is missing sth inside `libos.sh`, I don't know but it works without `retry_while`

dani8art: `$REPMGR_CURRENT_PRIMARY_HOST` is not well resolved when using `retry_while`

bitnami/postgresql-ha/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: postgresql-ha
-version: 3.1.0
+version: 3.1.1
```
juan131, line 3: This should be `3.2.0`, right?

dani8art: yes!

PR #2407, opened 2020-04-23
[bitnami/mongodb] add storageClass field for secondary in persistence

### LINKED ISSUE #2405: [bitnami/mongodb] Allow secondary use different storageclass
**Which chart**:
mongodb 7.13.x

No

I want to be possible configure a differente storageclass to secondary, this approach allow i run primary/secondary/arbiter in differente AWS zones.

Add a persistent option to a secondary.

**Additional context**
this has already been discussed in PR https://github.com/bitnami/charts/pull/2398

golgoth31:
**Description of the change**

Add ability to declare different storageClass for primary and secondary pods

**Benefits**

flexibility

**Possible drawbacks**

n/a

**Applicable issues**

  - fixes #2405 

**Additional information**

n/a

**Checklist**


dani8art:
Hi @golgoth31, thank you for your PR. Could you share with us your use case? you have different `storageClass` per node? why do you need different `storageClass` for distributing mongo nodes along with your cluster node? I would like to understand the use case a bit more.

Also, I would go for a different approach regarding values yaml.
```
  storageClass: myCommonStorageClass
  storageClassOverride:
    primary: myPrimaryStorageClass
    secondary: mySecondaryStorageClass
```

golgoth31:
Hi,
this is not my use case as we use antiaffynity with zone as topology key to encure pods are deployed in different zones. This use case comes from [https://github.com/bitnami/charts/pull/2398#issuecomment-618409546](https://github.com/bitnami/charts/pull/2398#issuecomment-618409546) :
> Hm, i have 3 storageclass with regions in your name, example:
> 
> * gp2-us-east-1a
> * gp2-us-east-1c
> * gp2-us-east-1d
> 
> For this reasons that i need a possibility that configure differente persistence volumes for primary/secondary, do you understand? Like this:
> 
> > mongodb
> > persistence:
> > enabled: true
> > storageClass: gp2-us-east-1a
> > accessMode: ReadWriteOnce
> > size: 10Gi
> > persistenceSecondary:
> > enabled: true
> > storageClass: gp2-us-east-1c
> > accessMode: ReadWriteOnce
> > size: 10Gi
> 
> And nodeSelectors:
> 
> > nodeSelector:
> > environment: "stage-backend-us-east-1a"
> > nodeSelectorSecondary:
> > environment: "stage-backend-us-east-1c"
> > nodeSelectorArbiter:
> > environment: "stage-backend-us-east-1d"
> 
> With this approach i will have a primary on us-east-1a region, secondary on us-east-1c and arbiter on us-east-1d, a great MongoDB HA solution.
> 
> What do you think?

I wrote it in a way to not introduce breaking change in the values file to ease update for users

dani8art:
Ok, I understand so you would need to have a storage class for each Mongo node type

```
storageClass: "-"  # which is common  
primaryStorageClass: "-" # which will override storageClass for primary nodes
secondaryStorageClass: "-" # which will override storageClass for secondary nodes
arbiterStorageClass: "-" # which will override storageClass for arbiter nodes
```

Also, you would need a different PVC for each type of node? 

I'm wondering, what about keeping then as "-" GKE doesn't apply as default storage class the one inside de region of the node? Or you could establish as default in each node?

caiohasouza:
@dani8art yes, but to arbiter is not necessary,thay don't use StorageClass. I think that this PR is good!

PR #2429, opened 2020-04-26
[bitnami/phabricator] fix ingress in phabricator chart with tls

jabdoa2:
**Description of the change**

Fix ingress in chart with tls

**Benefits**

It works with tls.

**Possible drawbacks**

Crashing "feature" when setting "tls: true" no longer works.

**Applicable issues**

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/phabricator/Chart.yaml
+++ b/bitnami/phabricator/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: phabricator
-version: 9.1.0
+version: 9.1.1
 appVersion: 2020.7.0
 description: Collection of open source web applications that help software companies build better software.
 keywords:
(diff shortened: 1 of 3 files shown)


jabdoa2:
Is there a way to add a CI test for this case? Like some test values? It has been obviously untested because even "helm template" shows an error when setting tls to true.

marcosbc:
@jabdoa2 Unfortunately we don't have validations for Helm charts, and it's not something we have planned to include for now. In any case, I'll open up an internal discussion for this to see if there's anything we can do.

jabdoa2:
@marcosbc should be relatively easy to put a few test-values.yaml into a directory and run "helm template" on them. Would also be very welcome examples if you want to use a chart. Tested examples are the gold standard in my opinion

marcosbc:
Thanks for the suggestions. We're a bit concerned with test-values.yaml examples as they can easily get outdated, and to avoid this we would need to enforce each PR to run some tests to ensure they are correct.

Anyhow, we will discuss internally to see the best case for this. It sure is a problem and if your proposal is a possible solution for it.

PR #2456, opened 2020-04-28
[bitnami/nginx] Change in the nginx deployment template for containerPort as variable

shankarmohannamakkal:
**Description of the change**

Change in the nginx deployment template for containerPort as variable

**Benefits**

containerPort as variable in deployment of nginx since it can be configured to port 80 also

**Possible drawbacks**

No drawbacks

**Applicable issues**

  - When we use internally we need to configure the containerPort as 80

**Additional information**

Please provide feedback of this Pull Request ASAP. Thanks.

**Checklist** 


marcosbc:
This change looks does not look like it would break anything, in that aspect it looks good.

However I am a bit concerned with the change. There are a few issues changing this to a different value than 8080, and it looks like we are assuming both things were done:

* Changing containerPort is not enough, you would also need to change the port used by NGINX via the NGINX_HTTP_PORT_NUMBER environment variable.
* If you want to use port 80 (or any value less than 1024), you would also need to run the container as 'root' user.

Could you provide more context on this change and why you require it? Also, were you able to successfully try this change

shankarmohannamakkal:
When i try to host my application using this `nginx` chart, the containerPort 8080 is not working (exact reason i don't know why) and so when i use the port as 80, its successfully hosted and working.

So i hope the port 8080 may not be work in all the cases and need to change it as 80. So i changed this as variable.

marcosbc:
Are you basing on the Bitnami NGINX container? Or are you using a custom container image with non-Bitnami NGINX binaries?

shankarmohannamakkal:
> Are you basing on the Bitnami NGINX container? Or are you using a custom container image with non-Bitnami NGINX binaries?

I am using the custom container image for my application and its non-Bitnami NGINX binaries.

marcosbc:
@shankarmohannamakkal Would it be an option to change your NGINX container to run on port 8080 instead? It should not be so complex, and it would allow you to use the NGINX chart without any changes to it (apart from the Docker image ID).

shankarmohannamakkal:
> @shankarmohannamakkal Would it be an option to change your NGINX container to run on port 8080 instead? It should not be so complex, and it would allow you to use the NGINX chart without any changes to it (apart from the Docker image ID).

I my department, there are multiple applications which uses the same NGINX server which is configured under port 80. So I don't have an option to change the NGINX port to 8080 since its at organization level. I am planning to use the bitnami NGINX chart for one of my application here. Please suggest.

marcosbc:
Well, since it is a container you could make the port configurable via env. vars so it is 8080 by default.

In any case, I'll accept this PR. Please fix the conflicts (by merging the latest changes from master) so I can approve and start the merging process.

shankarmohannamakkal:
> Well, since it is a container you could make the port configurable via env. vars so it is 8080 by default.
> 
> In any case, I'll accept this PR. Please fix the conflicts (by merging the latest changes from master) so I can approve and start the merging process.

Thanks. Conflict resolved. Please check and approve.

marcosbc:
LGTM, thanks! I'll launch our CI to test and merge these changes.

PR #2457, opened 2020-04-28
[bitnami/cassandra] Allow to specify maxUnavailable pods

sheerun:
**Description of the change**

- fixes name of setting that configured minimum disruption budget (minimumAvailable -> minAvailable)
- adds setting that configures maximum disruption budget (maxUnavailable)

**Benefits**

As above

**Possible drawbacks**

No breaking changes

**Checklist** 

### DIFF
--- a/bitnami/cassandra/README.md
+++ b/bitnami/cassandra/README.md
@@ -92,8 +92,9 @@ The following tables lists the configurable parameters of the cassandra chart an
 | `cluster.rack`                       | Rack name                                                                                                                                                 | `rack1`                                                      |
 | `cluster.enableRPC`                  | Enable Thrift RPC endpoint                                                                                                                                | `true`                                                       |
 | `cluster.enableUDF'                  | Enable CASSANDRA_ENABLE_USER_DEFINED_FUNCTIONS                                                                                                            | `false`                                                      |
-| `cluster.pdbEnabled`                 | Enable the creation of a Pod Disruption Budget to ensure a minimun number of pods are running.                                                            |               `true`                                                      |
-| `cluster.minimumAvailable`           | Minimum number of instances that must be available in the cluster (used of PodDisruptionBudget) Needs `cluster.pdbEnabled=true`                           | `1`                                                          |
+| `cluster.pdbEnabled`                 | Enable the creation of a Pod Disruption Budget to ensure a minimun number of pods are running.                                                            | `true`                                                       |
+| `cluster.minAvailable`               | Minimum number of instances that must be available in the cluster (used of PodDisruptionBudget) Needs `cluster.pdbEnabled=true`                           | `nil`                                                        |
+| `cluster.maxUnavailable`             | Maximum number of instances that can be unavailable in the cluster (used of PodDisruptionBudget) Needs `cluster.pdbEnabled=true`                          | `nil`                                                        |
 | `cluster.internodeEncryption`        | Set internode encryption. NOTE: A value different from 'none' requires setting `tlsEncryptionSecretName`                                                  | `none`                                                       |
 | `cluster.clientEncryption`           | Set client-server encryption. NOTE: A value different from 'false' requires setting `tlsEncryptionSecretName`                                             | `false`                                                      |
 | `cluster.jvm.extraOpts`              | Set the value for Java Virtual Machine extra optinos (JVM_EXTRA_OPTS)                                                                                     | `nil`                                                        |
@@ -257,6 +258,10 @@ As an alternative, this chart supports using an initContainer to change the owne
 You can enable this initContainer by setting `volumePermissions.enabled` to `true`.
 
 ## Upgrade
+
+### 5.4.0
+
+The `minimumAvailable` option has been renamed to `minAvailable` for consistency with other charts. This is not a breaking change as `minimumAvailable` never worked before because of an error in chart templates.
 
 ### 5.0.0
 
--- a/bitnami/cassandra/Chart.yaml
+++ b/bitnami/cassandra/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: cassandra
-version: 5.3.3
+version: 5.4.0
 appVersion: 3.11.6
 description: Apache Cassandra is a free and open-source distributed database management system designed to handle large amounts of data across many commodity servers, providing high availability with no single point of failure. Cassandra offers robust support for clusters spanning multiple datacenters, with asynchronous masterless replication allowing low latency operations for all clients.
 keywords:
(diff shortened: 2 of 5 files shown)


bitnami/cassandra/README.md
```diff
@@ -92,8 +92,9 @@ The following tables lists the configurable parameters of the cassandra chart an
...
-| `cluster.pdbEnabled`                 | Enable the creation of a Pod Disruption Budget to ensure a minimun number of pods are running.                                                            |               `true`                                                      |
-| `cluster.minimumAvailable`           | Minimum number of instances that must be available in the cluster (used of PodDisruptionBudget) Needs `cluster.pdbEnabled=true`                           | `1`                                                          |
+| `cluster.pdbEnabled`                 | Enable the creation of a Pod Disruption Budget to ensure a minimun number of pods are running.                                                            | `true`                                                       |
+| `cluster.minAvailable`               | Minimum number of instances that must be available in the cluster (used of PodDisruptionBudget) Needs `cluster.pdbEnabled=true`                           | `nil`                                                        |
```
rafariossaa, line 96: Hi,
  Could you rename this value to keep the same name ?.
  This way there won't be changes for the users.

bitnami/cassandra/README.md
```diff
@@ -92,8 +92,9 @@ The following tables lists the configurable parameters of the cassandra chart an
...
+| `cluster.maxUnavailable`             | Maximum number of instances that can be unavailable in the cluster (used of PodDisruptionBudget) Needs `cluster.pdbEnabled=true`                          | `nil`                                                        |
```
rafariossaa, line 97: For coherence , this should be `cluster.maximumUnavailable`

bitnami/cassandra/README.md
sheerun: Please notice that in go code this variable was named minAvailable. Less breaking change is fixing documentation rather than fixing go-lang code, no? :)

bitnami/cassandra/README.md
sheerun: minAvailable and maxUnavailable are used in other charts + minAvailable is currently used in go code. This is documentation change for minAvailable and for coherence new variable should be maxUnavailable, correct me if I'm wrong

bitnami/cassandra/README.md
rafariossaa: Sorry, I think I not expressed my thought very well.
I wanted to mean to keep it the same with respect to previous versions. This way backwards compatibility is kept.
And yes, you'd have to change not only in this file.

bitnami/cassandra/README.md
rafariossaa: OK, I have checked what you mentioned about other charts.
I agree with keeping minAvailable and maxUnavailable.
Please add a note to readme mentioning the renaming of minimunAvailable, and bump the major version of the chart instead of the minor.
Thanks

sheerun:
@rafariossaa I've updated the README mentioning renaming, but I don't agree this is breaking change as minimumAvailable never worked before (it had no effect) because of an error in chart template. Effectively this is a new feature so just minor bump is necessary (the same goes for maxUnavailable)

carrodher:
That is right, because of the wrong `if` statement the previous changes didn't work. I am going to test this PR in our pipeline, and in case everything works as expected we will merge and release the changes.

Thanks for your contribution.

carrodher:
The CI/CD is failing but I am not able to reproduce the issue locally. The issue appears during the upgrade tests, we are installing the first version of this major (`5.0.0`) and then upgrading to the current version in this PR:
```
$ helm install t100113-cassand --set dbUser.password=random --set statefulset.updateStrategy=RollingUpdate bitnami/cassandra --version 5.0.0
$ helm upgrade t100113-cassand .
```
⬆️ where `.` is the path to this code.

Then the following issue appears:
```
Error: UPGRADE FAILED: cannot patch "t100113-cassand-cassandra-headless" with kind PodDisruptionBudget: PodDisruptionBudget.policy "t100113-cassand-cassandra-headless" is invalid: spec: Forbidden: updates to poddisruptionbudget spec are forbidden.

...

18m         Normal    NoPods                   poddisruptionbudget/t09
```

I checked the differences between the `5.0.0` and this code when rendering  _pdb.yaml_ (using `helm template`) and there is nothing significant:
```diff
# Source: cassandra/templates/pdb.yaml
apiVersion: policy/v1beta1
kind: PodDisruptionBudget
metadata:
  name: t100113-cassand-cassandra-headless
  labels:
    app: cassandra
-   chart: cassandra-5.0.0
+   chart: cassandra-5.4.0
    release: t100113-cassand
    heritage: Helm
spec:
  selector:
    matchLabels:
      app: cassandra
      release: t100113-cassand
- maxUnavailable: 1
+ minAvailable: 1
```

We will continue investigating the issue in the upgrade tests.

sheerun:
The `if` in chart was wrong before so previous version set maxUnavailable even if user set minimumAvailable... I guess the fix would be to make `maxUnavailable: 1` the default in `values.yaml` right?

carrodher:
> The `if` in chart was wrong before so previous version set maxUnavailable even if user set minimumAvailable... I guess the fix would be to make `maxUnavailable: 1` the default in `values.yaml` right?

Maybe setting `maxUnavailable: 1` by default fix the issue upgrading (I will test in a while), but I don't think the previous behavior was wrong as they shouldn't conflict with each other

sheerun:
They conflict with each other. minAvailable and maxUnavailable cannot be set at the same time

carrodher:
Yes, you're right, `minAvailable` and `maxUnavailable` cannot be both set at the same time.
I will revert the tests here and fix our internal pipeline

carrodher:
Issue solved and chart released, you can find the new version updating the repository:
```
▶ helm repo update
Hang tight while we grab the latest from your chart repositories...
...Successfully got an update from the "bitnami" chart repository
Update Complete. ⎈ Happy Helming!⎈

▶ helm search repo bitnami/cassandra
NAME             	CHART VERSION	APP VERSION	DESCRIPTION
bitnami/cassandra	5.4.0        	3.11.6     	Apache Cassandra is a free and open-source dist...
```

Thanks for your contribution

sheerun:
Tests were ok, just correct setting in values.yaml probably should be:

```
# minAvailable: 1
maxUnavailable: 1
```

This way there would be still `maxUnavailable: 1` set, just like (by mistake) before

carrodher:
Let's see, we can choose one or the other option as the default parameter in the `values.yaml`, I don't have any preference on this. We can see what is the opinion of other users.

PR #2463, opened 2020-04-29
[bitnami/fluentd] Add Istio compatibility

### LINKED ISSUE #2462: [bitnami/fluentd]
**Which chart**:
[bitnami/fluentd] 1.0.4

**Describe the bug**
The chart as is it's not compatible with Istio [requirements](https://istio.io/docs/ops/deployment/requirements/).
* `app` label is missing
* Service [port names](https://github.com/bitnami/charts/blob/32e9c184bc4c3cfdd1f50054bfc1d77fe1ac6ed6/bitnami/fluentd/templates/metrics-svc.yaml#L14) are incorrect 
* Service's are overlapping, e.g https://github.com/bitnami/charts/blob/master/bitnami/fluentd/templates/aggregator-svc.yaml and https://github.com/bitnami/charts/blob/master/bitnami/fluentd/templates/svc-headless.yaml both map to the same Pod (the aggregators).

**Expected behavior**
The chart should be compatible with Istio

- Output of `helm version`:

```

```

```
Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.0", GitCommit:"9e991415386e4cf155a24b1da15becaa390438d8", GitTreeState:"clean", BuildDate:"2020-03-25T14:58:59Z", GoVersion:"go1.13.8", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.0", GitCommit:"9e991415386e4cf155a24b1da15becaa390438d8", GitTreeState:"clean", BuildDate:"2020-03-25T14:50:46Z", GoVersion:"go1.13.8", Compiler:"gc", Platform:"linux/amd64"}

```

irizzant:
**Description of the change**

This PR adds Istio compatibility
**Benefits**

This PR adds Istio compatibility

**Possible drawbacks**

None AFAIK

**Applicable issues**

  - fixes #2462

**Additional information**

**Checklist** 


bitnami/fluentd/templates/aggregator-svc.yaml
```diff
@@ -1,32 +0,0 @@
-{{- if and .Values.aggregator.enabled .Values.aggregator.service.ports }}
```
dani8art, line 1: Why are we removing it?

bitnami/fluentd/values-production.yaml
```diff
@@ -274,10 +274,10 @@ aggregator:
     ## Service ports
     ##
     ports:
-      # http:
-      #   port: 9880
-      #   targetPort: http
-      #   protocol: TCP
+      http:
```
dani8art, line 277: Why did you put it as default value?

bitnami/fluentd/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: fluentd
-version: 1.0.4
+version: 1.0.5
```
dani8art, line 3: I think it could be a minor change so we are adding a new feature

bitnami/fluentd/templates/aggregator-svc.yaml
irizzant: Because this Service is useless, the aggregators Pod are already selected and covered by the [headless service](https://github.com/irizzant/charts-1/blob/fluentd-istio/bitnami/fluentd/templates/svc-headless.yaml)

bitnami/fluentd/values-production.yaml
irizzant: This port was already specified in the [containerPorts](https://github.com/irizzant/charts-1/blob/271dfd8b8cd014ede5beb293487ffcf20d8e5797/bitnami/fluentd/values-production.yaml#L95), I added it to the Service ports to reflect this so that Istio doesn't detect it as an unknown endpoint and it doesn't trigger Push warnings.

bitnami/fluentd/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: fluentd
-version: 1.0.4
+version: 1.1.4
```
carrodher, line 3: That should be 
```suggestion
version: 1.1.0
```

dani8art:
Hi @irizzant, it is failing in our CI/CD pipeline due to the service removal. We need to find another approach without removing it because there are some tests that depend on this service to be working. Could you try removing the headless service instead?

irizzant:
Hi @dani8art , the right approach should be rewriting/removing those tests that are failing, not keeping the unnecessary Service.
Also, removing the headless Service is by definition wrong, because aggregators are deployed as StatefulSet.

bitnami/fluentd/templates/_helpers.tpl
```diff
@@ -39,6 +39,8 @@ app.kubernetes.io/name: {{ include "fluentd.name" . }}
 helm.sh/chart: {{ include "fluentd.chart" . }}
 app.kubernetes.io/instance: {{ .Release.Name }}
 app.kubernetes.io/managed-by: {{ .Release.Service }}
+app.kubernetes.io/version: {{ .Chart.AppVersion }}
+version: {{ .Chart.AppVersion }}
```
juan131, line 43: Please undo these changes. Labels including versions are not compatible with upgrades (we found several issues in the past)

More context: 

This label will be different depending on the chart version. So, when you upgrade from version `X.X.X` to `Y.Y.Y`, the k8s don't use the same labels. This is highly problematic when dealing with PVCs, replicasets, etc.

bitnami/fluentd/templates/aggregator-statefulset.yaml
```diff
@@ -59,7 +62,7 @@ spec:
             {{- toYaml .Values.aggregator.containerPorts | nindent 12 }}
             {{- end }}
             {{- if .Values.metrics.enabled }}
-            - name: metrics
+            - name: tcp-metrics
```
juan131, line 65: Not necessary. Istio just requires the port name to follow its standards in the service.

irizzant: As reported [here](https://istio.io/docs/ops/configuration/traffic-management/protocol-selection/) in order for Istio to detect the traffic type either you rely on automatic protocol detection or you can be explicit and specify the traffic type manually.
Wherever the traffic type is known it's considered best practice to avoid relying on automatic protocol detection, so this change is correct.

bitnami/fluentd/templates/_helpers.tpl
irizzant: Istio doesn't need the `version` label to be exactly `{{ .Chart.AppVersion }}` at all but it does need to have a `version` label to avoid emitting validation warnings in Kiali user interface because it is missing.
I added a `version: v1` label as default which is a reasonable compromise to me.

dani8art: Please, you still need to remove `app.kubernetes.io/version: {{ .Chart.AppVersion }}`.

if you need to take the risk of having version labels we could do implementing another approach like:

https://github.com/bitnami/charts/blob/a0f9f97ed0a05ad6e26ee6ad492326830afb9f67/bitnami/mongodb-sharded/values.yaml#L597

https://github.com/bitnami/charts/blob/a0f9f97ed0a05ad6e26ee6ad492326830afb9f67/bitnami/mongodb-sharded/templates/shard/shard-data-statefulset.yaml#L18-L22

you could use a `extraLabels` or even `aggregator.extraLabels`

irizzant: I don't see any risk in having a `version: v1` default label applied on Pods which would be constant and never updated unless canary deployments occurs.
Sorry I missed the `app.kubernetes.io/version: {{ .Chart.AppVersion }}` and I'm going to remove it

dani8art: I wouldn't like to add `version: v1` to the chart if it means nothing, why v1? a version for the chart? the deployment? the app? it is a bit confusing and I agree with @juan131 it could create conflict. 

Having the second option is a good point so that you could add this label if you need and then you know why and your reasons and we keep the chart decoupled from this Kiali user interface.

irizzant: I need to understand exactly what kind of conflicts we're talking about.
I totally agree that second option would allow having extra needed labels regardless from the `version`, but if there is another option. 
Say for example I add `version: {{ .Chart.AppVersion }}` label directly to the Pod template as already done for the [app label](https://github.com/irizzant/charts-1/blob/fdd5b293001af3f8fd355346960a2702db3d9d57/bitnami/fluentd/templates/aggregator-statefulset.yaml#L9).
I deploy the chart with version X.X.X and:
* PVC are created and storage is claimed using only `Values.aggregator.persistence.accessMode`, `Values.aggregator.persistence.storageClass` and `.Values.aggregator.persistence.size` values
* StatefulSet, ReplicaSet, DaemonSet all use `spec.selector.matchLabels` to select the Pod they act on

After a while I switch the chart to Y.Y.Y version and:
* PVC keep on being tied to the right PV as per StatefulSet contract
* StatefulSet, ReplicaSet, DaemonSet will recreate their Pod according to their  `spec.selector.matchLabels`.

Given the above, I don't see any conflict rising.

irizzant: In addition to the above, I've just pushed a commit for the necessary changes.
At this point I'd just add the `version: {{ .Chart.AppVersion }}` label to the `template.metadata.labels`.

dani8art: Cool! thanks!

bitnami/fluentd/templates/aggregator-svc.yaml
alemorcuq: Can you share more information on this? The closest I can find is this paragraph from [Istio's documentation](https://istio.io/docs/ops/deployment/requirements/):

> Service association: A pod must belong to at least one Kubernetes service even if the pod does NOT expose any port. If a pod belongs to multiple Kubernetes services, the services cannot use the same port number for different protocols, for instance HTTP and TCP.

But this shouldn't be an issue as both services are using the same ports and the same protocols.

The service was actually needed if you wanted to aggregate logs from the outside via TCP. This isn't possible anymore.

irizzant: > But this shouldn't be an issue as both services are using the same ports and the same protocols.

Yes but Envoy listeners are created on IP - port combinations basis and if there are multiple Service hitting the same IP - port combination Istio generates conflicts.

>The service was actually needed if you wanted to aggregate logs from the outside via TCP. This isn't possible anymore.

What do you mean by this? Having the aggregator Service would expose the aggregator port 24224 outside the cluster, which would allow to send logs to the aggregators Pod and that's fine.
Nevertheless, in a typical Kubernetes cluster having NodePort open is a problematic choice since the cluster nodes could not be directly reachable, which would require extra configuration at networking and loadbalancer levels.
If this use case is still needed for any reason, nothing prevents you from using the forwarder Service instead.
You just need to open the very same port 24224 and configure forwarders to send traffic to their aggregators.

ngyl88: Hi @irizzant @dani8art , removal of this file disable the service creation of type "LoadBalancer" and "NodePort". Since this file does support service type "ClusterIP" (before the deletion), could you consider removing the headless service instead? It looks like the changes you that you done to the headless service is portable to this file. FYI, we are not using the headless service nor the port 24224 in my current project, but using NodePort service with another layer of load balancer setup. Thanks!

irizzant: Hi @ngyl88 
the headless Service is needed because aggregators are StatefulSet, they need persistence and they cannot be changed.
The forwarder Service does support ClusterIP, LoadBalancer and NodePort, so it supports also being deployed behind cloud providers, plus the forwarder configuration can be customized to forward logs to other services listening inside/outside the Kubernetes cluster.

PR #2464, opened 2020-04-29
[bitnami/mongodb] Change tag for a svc of mongo with replicaset enable 

### LINKED ISSUE #2304: MongoDB replicaset service
Hi, 

This is a doubt about the bitnami/mongodb chart. When I exec the chart with replicaset=true, the chart create two services: 

my-release-mongodb
my-release-mongodb-headless

On the first service (my-release-mongodb) there is only one endpoint to mongodb-primary pod. The second service (my-release-mongodb-headless) there are 3 endpoints (1 pod arbiter, 1 pod slave and 1 primary pod).

I want to expose this service from outside the cluster, if I expose the first service, the only pod that will receive connection is the "mongodb-primary".  If I expose the second service, all the pod will receive connections, including arbiter pod. It's correct?

There is a way to create a service only with the primary's and secondary's pods, without arbiters?

Mauraza:
**Description of the change**
Change in the tag component

**Benefits**
If the primary is dropped and becomes primary, the secondary can still access the Mongodb cluster

**Possible drawbacks**

**Applicable issues**

  - fixes #2304

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/mongodb/Chart.yaml
+++ b/bitnami/mongodb/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: mongodb
-version: 7.13.0
+version: 7.13.1
 appVersion: 4.2.6
 description: NoSQL document-oriented database that stores JSON-like documents with dynamic schemas, simplifying the integration of data in content-driven applications.
 keywords:
--- a/bitnami/mongodb/templates/statefulset-primary-rs.yaml
+++ b/bitnami/mongodb/templates/statefulset-primary-rs.yaml
@@ -28,14 +28,14 @@ spec:
     matchLabels:
       app: {{ template "mongodb.name" . }}
       release: {{ .Release.Name }}
-      component: primary
+      component: replicaset
   template:
     metadata:
       labels:
         app: {{ template "mongodb.name" . }}
         chart: {{ template "mongodb.chart" . }}
         release: {{ .Release.Name }}
-        component: primary
+        component: replicaset
       {{- if .Values.podLabels }}
 {{ toYaml .Values.podLabels | indent 8 }}
       {{- end }}
--- a/bitnami/mongodb/templates/statefulset-secondary-rs.yaml
+++ b/bitnami/mongodb/templates/statefulset-secondary-rs.yaml
@@ -21,7 +21,7 @@ spec:
     matchLabels:
       app: {{ template "mongodb.name" . }}
       release: {{ .Release.Name }}
-      component: secondary
+      component: replicaset
   podManagementPolicy: "Parallel"
   serviceName: {{ template "mongodb.fullname" . }}-headless
   replicas: {{ .Values.replicaSet.replicas.secondary }}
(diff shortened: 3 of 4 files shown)


marcosbc:
I think this can be problematic; the Bitnami MongoDB chart will not work properly when the master is down, and I think that is to be expected as failover is not implemented into the chart (even though MongoDB supports that) due to lack.

This change seems to be associating a secondary node to the 'primary' service, but the 'primary' service is only meant for the primary node, in the same way the 'secondary' service is meant for secondary nodes.

However, I might be wrong so I'm going to request a review from someone else.

derom:
Pls consider this kind of connection string as a replacement for this pr:
mongodb://root:password@my-service-mongodb-primary-0.my-service-mongodb-headless.default.svc.cluster.local:27017,my-service-mongodb-secondary-0.my-service-mongodb-headless.default.svc.cluster.local:27017/

marcosbc:
I asked more folks to check this and they did not like the approach. What @derom suggests should be enough to fix this issue, requiring no changes to the chart.

Mauraza:
If for whatever reason you lose the `mongodb-primar-0` node, the PRIMARY role is adopted by other nodes. If you now add a new node to the cluster, it is going to register to `mongodb-primary-0`, but this node is not longer alive or it is not the PRIMARY, so the new node won't be added to the cluster

emrahcetiner:
> This change seems to be associating a secondary node to the 'primary' service, but the 'primary' service is only meant for the primary node, in the same way the 'secondary' service is meant for secondary nodes.

I agree. This change may confuse primary reachability.

alankwon:
I initially was using the 2 headless services like how @derom suggested. Then, I saw the comment from @juan131 about following the direction from "helm status my-release" command. I'm accessing it from the cluster and according to the notes, I should be using my-release-mongodb.default.svc.cluster.local, rather than the headless service. 

So, the notes, the comments, all seem to suggest: use the primary-0. When it goes down, wait 'till it comes back up as it will take over the primary role. Otherwise, I don't see the point of my-release-mongodb service definition. So, if that's the case, what exactly is the secondary node created for? I get that the minimum recommendation in a replica set is 3 nodes, but if you are not taking advantage of secondary nodes taking over the primary role, why deploy this as replica set? Just trying to understand the reasoning of this deployment logic.

marcosbc:
Thanks for your input @alankwon. I agree that the current logic does not make much sense, and instead of having "mongodb-primary" and "mongodb-secondary" services, we should try to merge them into one (i.e. "mongodb-node" excluding arbiters). That way the failover would work properly.

vpapadopoulos:
Primary & Secondary are  2 of many ephemeral  mongodb nodes states , not properties.
All mongodb replica nodes are just nodes, elligible to become Primary of Secondar during their lifecycle. There is no preference  to one over the others to become Primary. (priority field is by default 1 to all)
The first node that creates the replica set becomes the first Primary node. But you can not predict or make assumptions over which one will be the Primary at any time in the future.

This is why clients need to include ALL replica nodes in their connection URI, in order to automaticaly connect to the Primary node after each replica set election.

So having separate Primary and Secondary stateful sets makes no sense. And there is no use to put a load balancer in front of the replica set.

marcosbc:
You're right @vpapadopoulos. It looks like the MongoDB Helm chart may require some refactor. @Mauraza is that something you will be able to do?

Mauraza:
This PR is closed since this has been done https://github.com/bitnami/charts/pull/2918

PR #2469, opened 2020-04-29
[bitnami/nginx] Adding health ingress.

shankarmohannamakkal:
**Description of the change**

Adding a separate health ingress for checking the health status of the application and it will be use for traffic manager.

**Benefits**

Create separate health ingress for the application.

**Possible drawbacks**

No drawbacks

**Additional information**

We can't use the same ingress which is available now for the health status also. Since i want to use the ingress annotation `nginx.ingress.kubernetes.io/rewrite-target: /health/status` which is going to use only for ingress health host and not for the application host. So created a separate health ingress.
**Note**: Exactly used the same existing ingress copy as a new health ingress

**Checklist** 

### DIFF
--- a/bitnami/nginx/Chart.yaml
+++ b/bitnami/nginx/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: nginx
-version: 5.2.4
+version: 5.2.5
 appVersion: 1.17.10
 description: Chart for the nginx server
 keywords:
--- a/bitnami/nginx/Chart.yaml
+++ b/bitnami/nginx/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: nginx
-version: 5.3.4
+version: 5.3.0
 appVersion: 1.17.10
 description: Chart for the nginx server
 keywords:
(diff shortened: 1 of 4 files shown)


bitnami/nginx/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: nginx
-version: 5.2.4
+version: 5.2.5
```
andresbono, line 3: What about bumping the minor version instead? This is adding a new feature, right?

andresbono:
Thanks so much for your contribution, @shankarmohannamakkal!! :tada:

bitnami/nginx/Chart.yaml
shankarmohannamakkal: Yes, you are correct. I incremented a minor version now, instead of patch increment.

bitnami/nginx/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: nginx
-version: 5.2.4
+version: 5.3.4
```
carrodher, line 3: The minor bump should start with a new patch
```suggestion
version: 5.3.0
```

andresbono:
Thanks for your contribution, @shankarmohannamakkal!! 🎉

PR #2472, opened 2020-04-30
[bitnami/nginx] Created new Default Config map volume mount

shankarmohannamakkal:
**Description of the change**

- The scope for this change is to deploy with n number of setting files in the application
- Included a new default config map for application settings to deploy with the existing custom web application deployment model
- This config map contains any number of key value pairs and those will deploy in the volume mount path as those number of files which is used for application settings

**Benefits**

Those settings files (like js, css etc.,) can be used in the application.

**Possible drawbacks**

No limitation or drawbacks.

**Additional information**

Don't want to change the existing volume mount deployment model. So created a new model for the deployment of any number of keys as a files in the application.

**Checklist** 


javsalgar:
Hi,

Thank you for the PR. As it is adding a feature, I would bump the minor version instead. Also, could you provide an example of usage?

shankarmohannamakkal:
- I have incremented the minor version
- I don't want to disturb the existing `serverBlock` or `staticSiteVolume` volumes. So i created a default one as `defaultSiteSetting` volume which is going to mount a new file in the path of `defaultSiteSettingVolumeMountPath`. Here the file name will be the default config map key name and the file content will be the value of the key.

Example in values yaml -
```
defaultSiteSetting:
  settings.js : "// Javascript function contains env specific values"
defaultSiteSettingVolumeMountPath: /usr/share/nginx/html/apps/admin-site/assets
```
This is going to create a new `settings.js` file with the content of `// Javascript function contains env specific values` and deploy in the path `/usr/share/nginx/html/apps/admin-site/assets`. The application going to use this settings.js file with the java script function which contains the environment specific values.

andresbono:
We have already have a set of parameters that covers this feature, maybe from different sources (volume, configmap, git repository,...):

| Parameter                                  | Description                                                                                 | Default                                                      |
| ------------------------------------------ | ------------------------------------------------------------------------------------------- | ------------------------------------------------------------ |
| `staticSiteConfigmap`                      | Name of existing ConfigMap with the server static content                                   | `nil`                                                        |
| `staticSitePVC`                            | Name of existing PVC with the server static content                                         | `nil`                                                        |
| `cloneStaticSiteFromGit.enabled`           | Get the server static content from a git repository                                         | `false`                                                      |

If I'm not wrong, you are suggesting adding a new one that will use literal values from a map in the `values.yaml`.

Maybe we should keep the same naming for the new way of specifying static site files. Something like:

```
staticSiteFiles:
  enabled: false
  map:
    file1: '// Contents 1'
    file2: '// Contents 2'
  mountPath: '/app'
```

Let me know your thoughts, @shankarmohannamakkal, @javsalgar.

shankarmohannamakkal:
Thanks. This approach also will work. 

Below 2 changes are only required.

1. The config map will be
```
{{- if .Values.staticSiteFiles.enabled }}
apiVersion: v1
kind: ConfigMap
metadata:
  name: {{ .Values.staticSiteConfigmap }}
  labels: {{- include "nginx.labels" . | nindent 4 }}
data:
  {{- range $key, $value := .Values.staticSiteFiles.map }}
  {{- end }}
{{- end }}
```

2. Deployment file may change
```
{{- if .Values.staticSiteFiles.enabled }}
- name: staticsite
  mountPath: {{ .Values.staticSiteFiles.mountPath}}
{{- end }}
```

Let me know can i proceed with these changes @javsalgar, @andresbono

javsalgar:
In some charts, what we have is the `extraVolumes` and `extraVolumeMounts` values that allow you to include whatever you want in the container. I think this case could be covered by adding these values instead of adding a new configmap (it could be confusing for users). Wouldn't that work for this case?

shankarmohannamakkal:
> In some charts, what we have is the `extraVolumes` and `extraVolumeMounts` values that allow you to include whatever you want in the container. I think this case could be covered by adding these values instead of adding a new configmap (it could be confusing for users). Wouldn't that work for this case?

I am not sure that will work or not. Can you please give me the link of which chart contains `extraVolumes` and `extraVolumeMounts` values. 

We need a separate config map here to create static files. So that we can create any number files and mount the same. As said previously, some thing like below will work. 

```
staticSiteFiles:
  enabled: false
  map:
    file1: '// Contents 1'
    file2: '// Contents 2'
  mountPath: '/app'
```

javsalgar:
Hi,

Here you can see the example: https://github.com/bitnami/charts/blob/master/bitnami/kibana/templates/deployment.yaml

If possible, the more that we have the same nomenclature and features, the better. So if using these two values does the job, I would use that instead.

shankarmohannamakkal:
> Hi,
> 
> Here you can see the example: https://github.com/bitnami/charts/blob/master/bitnami/kibana/templates/deployment.yaml
> 
> If possible, the more that we have the same nomenclature and features, the better. So if using these two values does the job, I would use that instead.

Thanks for the example. 

I see there are following 3 options we can use. Please let me know your suggestion on which option will be suitable here.

**Option 1**

I can use the `serverBlock` itself, but the following changes are required.

1. Deployment template - need to change the below mount path as variable
`mountPath: /opt/bitnami/nginx/conf/server_blocks`

2. server-block-config template - the entire data should be dynamic some thing like below,

```
data:
  {{- range $key, $value := .Values.serverBlock }}
  {{- end }}
```

**Option 2**

I can use the static site itself, but the below changes are required

1. Deployment template - need to change the below mount path as variable
`mountPath: /app`

2. Create a new config file may be like below

```
{{- if .Values.staticSite }}
apiVersion: v1
kind: ConfigMap
metadata:
  name: {{ template "nginx.fullname" . }}-static-site
  labels: {{- include "nginx.labels" . | nindent 4 }}
data:
  {{- range $key, $value := .Values.staticSite }}
  {{- end }}
{{- end }}
```

**Option 3**

Either as you mentioned i can create `extraVolumes` and `extraVolumeMounts` values in deployment (similar to the kibana deployment) or as i did in this PR. But any case i need a separate config map file is required like above case.

I hope you understand my need here. Please let me know what you think. @javsalgar, @andresbono

javsalgar:
I would go for Option 3, and have the configmap created before doing helm install. If you want to create the configmap as part of the values.yaml, we are exploring an extraDeploy feature in charts like kong where users have the freedom to add whatever fits their needs.

https://github.com/bitnami/charts/blob/master/bitnami/kong/templates/extra-list.yaml

shankarmohannamakkal:
Thanks.

Done the changes in deployment using `extraVolumes` and `extraVolumeMounts`. And as you suggested i will have the configmap created before doing helm install.

Please review the changes.

javsalgar:
Hi,

Great, thanks. Instead of `tpl`, could you use the helper `nginx.tplValue`? It is more versatile when using templated values.

shankarmohannamakkal:
> Hi,
> 
> Great, thanks. Instead of `tpl`, could you use the helper `nginx.tplValue`? It is more versatile when using templated values.

Hi @javsalgar 

Addressed your comment. Used the helper `nginx.tplValue`

shankarmohannamakkal:
> LGTM. Thanks! Could you check the merge issue?

Thanks! Merge issue resolved now. Please help to merge into master.

PR #2483, opened 2020-05-01
[bitnami/postgresql-ha] Default pgpool loadbalancing to false

RichieSams:
pgpool LoadBalancing is an advanced feature, and should be disabled by default. In the current default state of the chart, pgpool load balancing can lead to invalid read after write synchronization errors. 

Example: 
 1. application 1 INSERTs a new row in a table. pgpool sends this to the master node
 2. application 1 advertises this new row in a message passing system, say rabbitMQ
 3. application 2 receives the message, and tries to SELECT the row. pgpool with load balancing pushes the select to one of the slaves. BUT, the slave hasn't gotten the replicated data yet. The SELECT fails with no rows.

While there are solutions to the above problem, I think it illustrates that pgpool load balancing has enough footguns that it should be disabled by default and opt in.

**Possible drawbacks**

This does change default behavior. So if users don't specify it explicitly in their own values.yaml, their behavior will change. That said, this will be a minor version change, so that should show users that there is potentially changing behavior.


carrodher:
Hi, thanks for your contribution, we are going to check and test the implications of this change, although the code itself is easy to review, I want to check with other members.

carrodher:
Hi, I understand your concern but I would like to have more information about this advanced option. Can you point me to some docs that sustain this advanced feature?

Another thing, if it is an advanced option, it would make sense to you to enable it by default in the `values-production.yaml`?

RichieSams:
Here's the documentation for pgpool: https://www.pgpool.net/docs/41/en/html/read-query-load-balancing.html

I believe the issue is that pgpool load balancing assumes it's the one that's in control of replication. But, I *believe* this chart is using repmgr for replication. As a result, pgpool can't know the state of the replication, and can't safely know what statements can be offloaded to replicas.

carrodher:
Hi, we're trying to have the full context about the specific issue.

For replication, we are using [the master-slave mode](https://www.pgpool.net/docs/41/en/html/runtime-config-running-mode.html#RUNTIME-CONFIG-MASTER-SLAVE-MODE), you can check in the `pool.conf` that we have `master_slave_mode = on` and this should be compatible with the LOAD BALANCER according to https://www.pgpool.net/docs/pgpool-II-4.1.1/en/html/runtime-config-load-balancing.html.

On the other hand, in PostgreSQL, the replication is synchronous (parameter `synchronous_commit` in `postgresql.conf` which by default is `on`).

This means that "success" is not sent to the client until the WAL has propagated to the nodes: https://www.postgresql.org/docs/11/runtime-config-wal.html#GUC-SYNCHRONOUS-COMMIT

Therefore, what the described issue should not happens.

RichieSams:
@carrodher Thank you for following up. I apolgize for the delay in response.
I'm not super familiar with repmgr, pgpool, and postgres configuration, and how it's configured in this chart. We were just seeing read after write errors with pgpool load-balancing turned on, in multiple projects. Our one project does this (pseudo-code):

```go
package main

import (
	"context"
	"time"

	"github.com/jmoiron/sqlx"
	uuid "github.com/satori/go.uuid"
	"gitlab.ct.activision.com/indy/ark/v2/pkg/ark"
)

type Task struct {
	ID          uuid.UUID `gorm:"primary_key;type:uuid;not null" db:"id"`
	CreatedAt   time.Time `gorm:"default:'1970-01-01 00:00:00';not null" db:"created_at"`
	StartedAt   time.Time `gorm:"default:'1970-01-01 00:00:00';not null" db:"started_at"`
	CompletedAt time.Time `gorm:"default:'1970-01-01 00:00:00';not null" db:"completed_at"`
	Type        string    `gorm:"default:'';not null" db:"type"`
	Message     string    `gorm:"default:'';not null" db:"message"`
	Succeeded   bool      `gorm:"default:FALSE;not null" db:"succeeded"`
}

func Insert(ctx context.Context, db *sqlx.DB, body []byte) error {
	taskID := uuid.NewV4()
	_, err := db.NamedExecContext(
		ctx,
		`INSERT INTO tasks (id, type, created_at)
		 VALUES (:id, :type, :created_at)`,
		&ark.Task{
			ID:        taskID,
			Type:      "bom",
			CreatedAt: time.Now().UTC(),
		},
	)
	if err != nil {
		return err
	}

	// Publish via RabbitMQ
	// mq.Publish(taskID, body)

	return nil
}

func Consume(ctx context.Context, db *sqlx.DB) error {
	// Wait until message from RabbitMQ
	// taskID, body := mq.Get()
	taskID := ""
	body := []byte{}

	task := &ark.Task{}
	err := db.GetContext(ctx,
		task,
		`SELECT * from tasks
		WHERE id=$1`,
		taskID)

	return err
}
```

We have one server running a http REST api, and when it gets a POST, it calls `Insert()`, which INSERTs a row into Postgres, and then publishes an event in RabbitMQ about that row.

We have another server listening on RabbitMQ, and when it gets an entry, calls `Consume()`. The issue we ran into was the query in `Consume()` would return No rows returned. When I manually checked, the row did exist. So, I believe pgpool load balancing was sending the `Consume()` query to a slave that was lagging in replication. Once we turned off pgpool load balancing, the issue disappeared.

carrodher:
Yeah, it may be possible but I am not sure if that is something that should be implemented by default or just for some specific use cases.

I mean, this chart allows different customizations to cover different use cases, you can change this option when deploying the chart by using `--set pgpool.useLoadBalancing=false` as part of the `helm install` (or `upgrade`) command. At this moment we didn't see other users facing similar issues, which is weird as it's the default option.

My vote is to maintaining this PR open just in case there are other use cases that need the same change so in that case, it can be an option to change the default value if your issue can be solved by adding a parameter in the installation and maintain the default value as it is I will follow that approach. Are you able to have it working just by using `--set pgpool.useLoadBalancing=false`?

RichieSams:
I'm good with that vote. Leaving it open to see if others run into the same issue. And if it comes up again, it can be re-addressed.

Yes, our issues are solved by `--set pgpool.useLoadBalancing=false`

carrodher:
Nice, I just added the on-hold label to avoid the stale-bot closing this PR after 15 days without activity. Apart from that, I added a reminder to revisit it in a month, if there is no news about that in 1 month, maybe we can close it. Anyway, it can be reopened at any time if needed in the future.

Thank you very much for your time!

liyuntao:
Hi @carrodher @RichieSams 
We've met a [similar problem](https://github.com/bitnami/bitnami-docker-postgresql-repmgr/issues/19) and it's solved by  `--set pgpool.useLoadBalancing=false`.

svilensabev:
I also have an issue with pgpool's load balancing and had to turn it off. 

Our app is using [Postgraphile](https://www.graphile.org/) and pointing DB_HOST to the PGpool URL caused **some** GraphQL mutations to hit permission denied on schema api. 

I tried multiple configuration params from https://www.pgpool.net/docs/latest/en/html/runtime-config-load-balancing.html with no luck so far.

I do understand that Postgraphile has deep integration of pg roles than a regular app and that may be the problem here. The connection is made by authenticator user which parse a JWT and switch to one of the other pg roles that act as application roles e.g. app_user

My workaround so far would be to keep load balancing because I desperately want to route SELECT queries and relax primary db instance but also to provide URL to primary db instance for INSERT/UPDATE queries. This would require a logic on application level, that has to loop which DB_HOST is the primary db instance for that moment cause they may change any time.

carrodher:
Hi, I see. There are different use cases where it's necessary to set `pgpool.useLoadBalancing=false`, the thing is if we make this value the default option then other users will need to do the opposite `pgpool.useLoadBalancing=true`. Apart from that, changing this default value means a major bump in the chart version, because this change could modify the configuration for users that are using the `true` value after an upgrade.

I would like to investigate property if there is any configuration issue or misunderstanding between the `synchronous_commit=on` and `pg_stat_replication=sync` that we can solve without bump the major version. In the same way, if it is needed to bump the major version I think the best option is to do it when other changes that require the same bump are ready to be implemented.

What do you think about that? In the same way, I just created an internal task to properly investigate this issue and keep in mind when there is a new major version.

mboutet:
We are also experiencing this when selecting very shortly after inserting.

carrodher:
Hi, we're investigating the issue to determine what is the best default values. We will add a comment with more information when we advance in the decision, in the meantime, there is no problem modifying the default value when installing the chart, right? Thanks for reporting it!

mboutet:
Personally, there is no problem modifying the default values when installing the chart. I don't really mind the small lag given the added performance for read-only queries, but it would be great to find out the reason this is happening.

carrodher:
Yes, that is another option, as I mention we are going to investigate the different use cases deeply probably in our next iteration and update this PR with the conclusion. Thanks for your patience.

rafariossaa:
Hi,
  We are currently investigating this issue and which should be the default configuration.
  As @carrodher mentioned before, the `synchronous_commit` parameter is set to `on`, but there are other values possible, and I think that setting it to `remote_apply` would work in your scenario.
  The documentation [is here](https://www.postgresql.org/docs/11/runtime-config-wal.html#GUC-SYNCHRONOUS-COMMIT). 
  @RichieSams, @mboutet, do you think that would work in your scenario without having to disable load balancing ?. If you have many and quick transactions I think it would be a good idea to have the load balancing feature.

mboutet:
@rafariossaa, but if I understand correctly, `synchronous_commit=on` is the highest consistency level we can have (i.e. all online standbys must confirm the commit). So, how would `synchronous_commit=remote_apply` help?

rafariossaa:
Hi @mboutet ,

  Yes you are right, `on` should give the highest consistency level.
  Let me check the `delay_threshold`.   Thanks for sharing your findings.

amarbs:
For `synchronous_commit` to be truly distributed in a cluster, the `synchronous_standby_names` setting should be non-empty. However, setting `synchronous_standby_names` would make postgres master hang all queries until the specified standby is up. But in this chart, _repmgr_ is taking care of bringing up the standby. And _repmgr_ in turn needs to make DB changes to set itself up, which is blocked till standby is up. This is the catch-22 situation. There doesn't seem to be a mechanism to setup _repmgr_ with synchronous replication (at least as a global setting. Per transaction setting can still be possible).

rafariossaa:
Hi,
  I am creating an internal task to revisit the synchronous configuration.
  Thanks for sharing.

ceyhunn:
Hi @RichieSams, @carrodher, @rafariossaa, @mboutet ,
As you notice, same problem also occurred in my side, and it drives me crazy.
I think, we need `syncReplication: true`, this config have to affect synchronous_commit, synchronous_standby_names configs. However, it just set  `POSTGRESQL_NUM_SYNCHRONOUS_REPLICAS = 1 (replicaCount)`environment variable.


Screenshot from 2021-06-06 20-01-10

bitnami/postgresql-ha helm chart use  bitnami/postgresql-repmgr  as postgres image. In above screenshot you see, if `POSTGRESQL_NUM_SYNCHRONOUS_REPLICAS`  > 0, ` synchronous_commit, synchronous_standby_names` configs will be set. I don't know how this configs are overriden by bitnami/postgresql-ha helm chart? I think, the problem is that.

kdam0:
yes awaiting progress for this as well

RichieSams:
We gave up. helm-based solutions all suffered from issues. Like if all the nodes shut down, and came back up in a random order, the statefulset could we waiting for pod 0 to fully start up before creating pod 1. But pod 0 can't be fully up until the master on pod 1 is up. 

Deadlock.

We switched to an operator. Like https://github.com/zalando/postgres-operator
That way the operator can be an external controller, outside of the pods themselves. And the operator understands kubernetes and how to handle pod movements.

We had split brain twice with this helm chart. We've never had anything like that with an operator

rafariossaa:
Hi,
  There was a new release of the container images that should deal better with sync replication. 
  Could you give it a try to the latest version of the chart ?

kdam0:
> Hi, There was a new release of the container images that should deal better with sync replication. Could you give it a try to the latest version of the chart ?

```
2022-01-25 09:30:57.430308245 -0500 EST deployed        postgresql-ha-8.2.6     11.14.0 
```
Trying with the latest i think? Still no effect on `syncReplication: true` with 3 `replicaCount`. Master shows all replicas `sync_state` as `async`.
Doing a "two-step" deployment with `extendedConf` set on the second step, leads to nodes not joined to the cluster.
If I explicitly set `POSTGRESQL_NUM_SYNCHRONOUS_REPLICAS` to `"2"` as an `extraEnvVar`, with the 'two-step" deployment, I can get master to show 1 replica as `sync`, and the other as `potential`.

rafariossaa:
hi @RichieSams ,
  Did you have  a similar experience with the latest version ?

javsalgar:
Hi,

Just a note to let you know that we updated the chart  with a new revision of the PostgreSQL HA chart that fixes the synchronous commits issue. 

https://github.com/bitnami/charts/commit/bd69fac8068e7778b88d351b4bd3e335c680dc51

javsalgar:
Currently, we don't have plans to have the pgpool load balancing mechanism now that the synchronous commits are possible. We will close the PR for the time being. If more users want to change the default pgpool behaviour, we will gladly reopen it. Thank you everyone for the input and contributions!

PR #2518, opened 2020-05-05
[bitnami/postgresql-ha] Add option to set postgres user password

### LINKED ISSUE #2470: [bitnami/postgresql-ha] Method to set postgres password with other user
**Which chart**:
postgresql-ha 3.1.1

I'm attempting to initialize postgres with a specific user/database, such as with
```
postgres:
  username: my-db
  database: my-db
  password: ****
```

Doing that means the `postgres` user does not have a password set, as noted in https://github.com/bitnami/bitnami-docker-postgresql-repmgr#creating-a-database-user-on-first-run.

My motivation for having access as the postgres user is to do database backups, and to be able to restore a backup to a new cluster. The documented methods for that use ConfigMaps, which have a limit of 1MB (because they're stored in etcd), limiting their use to re-initialize the database from a backup.

I'd like to be able to set the `POSTGRESQL_POSTGRES_PASSWORD` setting in the container to provide a password for the `postgres` user when creating a separate database user.

I'd also be open to other suggestions on how to do backup/restore.

MikaelSmith:
When `postgresql.username` is not `postgres`, a new user is created that
only has permissions for the database created with `postgresql.database`
and the `postgres` user does not have a password and cannot be accessed
remotely.

This commit adds the `postgresql.postgresPassword` property that can be
used to specify the password for the `postgres` user when
`postgresql.username` is set to a different user.

Resolves #2470.

**Checklist** 


bitnami/postgresql-ha/templates/postgresql/secrets.yaml
```diff
@@ -7,6 +7,9 @@ metadata:
     app.kubernetes.io/component: postgresql
 type: Opaque
 data:
+  {{- if and .Values.postgresql.postgresPassword (not (eq .Values.postgresql.username "postgres")) }}
```
MikaelSmith, line 10: I've carried these conditionals over from https://github.com/bitnami/charts/blob/1765929fa8cf37819309ff975ebf0a8dfcf7824d/bitnami/postgresql/templates/secrets.yaml#L16, but they seem weird. If you set `username` or `postgresPassword` via the `global` section this won't trigger.

This seems better implemented as
```suggestion
  {{- if and (include "postgresql-ha.postgresqlPostgresPassword" .) (not (eq (include "postgresql-ha.postgresqlUsername" .) "postgres")) }}
```

However the template provides a random default if you don't specify one. Switching to checking the template result means that the `postgres` user would always have a password. If someone's intention was to disable that user they'd no longer have a way to do so.

juan131: Honestly, I think your solutions is much better.

> If someone's intention was to disable that user they'd no longer have a way to do so

That's not a problem, since it's required at this very moment due to the validations below:

- https://github.com/bitnami/bitnami-docker-postgresql-repmgr/blob/master/11/debian-10/rootfs/opt/bitnami/scripts/libpostgresql.sh#L247

PR #2536, opened 2020-05-06
[bitnami/harbor] Adds different credential fields for each database

schnatterer:
**Description of the change**

This PR adds the option of setting different credentials for each database. If no value is provided the field `externalDatabase.user` and/or `externalDatabase.password` is used. So this does not break compatibility.

**Benefits**

In case of a breach only one database is compromised.

**Possible drawbacks**

**Applicable issues**

**Additional information**

Some time ago, we've create a PR for the official harbor helm chart (without success), See goharbor/harbor-helm#501 and have since changed to bitnami/harbor.

**Checklist** 


andresbono:
Hi @schnatterer, thanks for your contribution! :tada:

We will review this PR shortly.

bitnami/harbor/templates/_helpers.tpl
```diff
@@ -101,6 +101,42 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this
   {{- end -}}
 {{- end -}}
 
+{{- define "harbor.database.clairUsername" -}}
+  {{- if eq .Values.postgresql.enabled true -}}
+    {{- printf "%s" "postgres" -}}
```
andresbono, line 106: Shouldn't this be: 
```suggestion
    {{- .Values.postgresql.postgresqlUsername -}}
```

The same for the rest if you think this change is appropriate.

schnatterer: You're right. I missed that one, when porting this PR from the official helm chart.
The same applies to `notaryServerUsername` and `notarySignerUsername`.

schnatterer:
@andresbono thanks for your review.
Rebased and implemented your suggestion.

PR #2545, opened 2020-05-07
[bitnami/etcd] Add PodDisruptionBudget

### LINKED ISSUE #2542: [bitnami/etcd] chart should include PodDisruptionBudget
**Which chart**:
The etcd chart

Pod evictions (for example because of a node-drain) can make etcd unavailable.

Add a PodDisruptionBudget for the etcd-pods that ensures that a majority of pods is available.

Every user of the chart could add the PodDisruptionBudget by himself.

carrodher:
**Description of the change**

Now it is possible to configure PodDisruptionBudget by using the parameters documented in the README.

```console
▶ helm template myrelease . --set podDisruptionBudget.enabled=true,podDisruptionBudget.minAvailable=4piVersion: policy/v1beta1
```
```yaml
kind: PodDisruptionBudget
metadata:
  name: myrelease-etcd
  labels:
    app.kubernetes.io/name: etcd
    helm.sh/chart: etcd-4.8.0
    app.kubernetes.io/instance: myrelease
spec:
  minAvailable: 4
  selector:
    matchLabels:
      app.kubernetes.io/name: etcd
      app.kubernetes.io/instance: myrelease
```

**Benefits**

Configure PodDisruptionBudget

**Possible drawbacks**

It is a new feature disabled by default, no expected drawbacks.

**Applicable issues**

  - fixes #2542

**Checklist** 


bitnami/etcd/README.md
```diff
@@ -102,6 +102,9 @@ The following tables lists the configurable parameters of the etcd chart and the
...
 | `persistence.annotations`                         | Annotations for the PVC                                                                                                                                   | `{}`                                                        |
+| `podDisruptionBudget.enabled`                     | Pod Disruption Budget toggle                                                                                                                              | `false`                                                     |
+| `podDisruptionBudget.minAvailable`                | Minimum available pods                                                                                                                                    | `nil`                                                       |
+| `podDisruptionBudget.maxUnavailable`              | Maximum unavailable pods                                                                                                                                  | `nil`                                                       |
```
juan131, line 107: Can I suggest using `pdb` instead of `podDisruptionBudget` for these parameters? We're using this on other charts and it's more shorter and easier to manage

dbaumgarten:
I am not sure if this is possible, but is it possible to calculate the number for minAvailable automatically based on the number of replicas (>50%)?
I think using a value by default that ensures a functional cluster is in the best interest of all users

carrodher:
> I am not sure if this is possible, but is it possible to calculate the number for minAvailable automatically based on the number of replicas (>50%)?
> I think using a value by default that ensures a functional cluster is in the best interest of all users

I am not 100% sure about that, not because it was not technically possible, but for two reasons:
- To guarantee the consistency with other bitnami charts
- Default values are always subjective, for each use case there are different needs. In this case, we are not setting any default value as the parameter is commented, so users can set whatever they want based on their requirements.

PR #2547, opened 2020-05-07
[bitnami/elasticsearch] Add support for custom initialization scripts and extra environment variables

marcosbc:
**Description of the change**

This allows for running a custom initialization script during the initialization process for all Elasticsearch containers.

To make it even more useful, we added support for specifying extra environment variables at deploy time, so the init scripts can make use of the data contained in them (such as credentials).

This would allow doing something such as:

* Changing Elasticsearch's configuration file.
* Performing additional plugin configuration before running. I.e. for configuring the Elasticsearch S3 repository plugins credentials during the initialization, with the credentials in a secret (or configmap, or via --set flags).
* Etc.

**Benefits**

Users will be able to customize the Elasticsearch chart even further!

**Possible drawbacks**

None.

**Applicable issues**

None

**Additional information**

None

**Checklist** 


bitnami/elasticsearch/README.md
```diff
@@ -70,6 +70,11 @@ The following table lists the configurable parameters of the Elasticsearch chart
...
 | `extraVolumes`                                    | Extra volumes                                                                                                                                             |                                                              |
 | `extraVolumeMounts`                               | Mount extra volume(s),                                                                                                                                    |                                                              |
+| `initdbScripts`                                   | Dictionary of initdb scripts                                                                                                                              | `nil`                                                        |
+| `initdbScriptsConfigMap`                          | ConfigMap with the initdb scripts (Note: Overrides `initdbScripts`)                                                                                       | `nil`                                                        |
```
javsalgar, line 74: In other charts we also allow init scripts as secrets in case they want to put sensitive data there. Using subfolders both can co-exist. Example: kong

https://github.com/bitnami/charts/blob/master/bitnami/kong/templates/deployment.yaml#L166

bitnami/elasticsearch/templates/configmap-initscripts.yaml
```diff
@@ -0,0 +1,15 @@
+{{- if and (or (.Files.Glob "files/docker-entrypoint-initdb.d/*.sh") .Values.initdbScripts) (not .Values.initdbScriptsConfigMap) }}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ template "elasticsearch.fullname" . }}-init-scripts
+  labels: {{- include "elasticsearch.labels" . | nindent 4 }}
+    component: master
+data:
+{{- if and (.Files.Glob "files/docker-entrypoint-initdb.d/*.sh") (not .Values.initdbScriptsConfigMap) }}
```
javsalgar, line 9: After seeing several issues wit the use of the `files/` section, we preferred to not using that subfolder as an option.

bitnami/elasticsearch/README.md
```diff
@@ -70,6 +70,11 @@ The following table lists the configurable parameters of the Elasticsearch chart
...
+| `extraEnvVars`                                    | Array containing extra env vars to be added to all pods (evaluated as a template)                                                                         | `[]`                                                         |
+| `extraEnvVarsConfigMap`                           | ConfigMap containing extra env vars to be added to all pods (evaluated as a template)                                                                     | `nil`                                                        |
```
javsalgar, line 76: We need to come with a naming standard. In other chart we have CM instead of ConfigMap (probably because it is too long 😅 )

bitnami/elasticsearch/templates/master-statefulset.yaml
```diff
@@ -109,6 +109,20 @@ spec:
               value: "yes"
             - name: ELASTICSEARCH_NODE_TYPE
               value: "master"
+            {{- if .Values.extraEnvVars }}
+            {{- include "elasticsearch.tplValue" ( dict "value" .Values.extraEnvVars "context" $ ) | nindent 12 }}
+            {{- end }}
+          {{- if or .Values.extraEnvVarsConfigMap .Values.extraEnvVarsSecret }}
```
javsalgar, line 115: Does it make sense for all the nodes to share the same env vars? Sometimes we've seen issues with that. In other charts we divide it by component type. However, if it makes sense maybe we can have a `commonExtraEnvVars` or, if for example all nodes require a plugin, maybe set the Plugins env var in the template.

bitnami/elasticsearch/values.yaml
```diff
@@ -130,6 +130,35 @@ discovery:
...
+## Alternatively, you can put your scripts under the files/docker-entrypoint-initdb.d directory
+##
+# initdbScripts:
+#   my_init_script.sh: |
+#      #!/bin/sh
+#      echo "Do something."
+#
+## ConfigMap with scripts to be run at first boot
+## Note: This will override initdbScripts
+# initdbScriptsConfigMap:
```
javsalgar, line 144: I would leave it uncommented but empty. It is very useful if you want to use automation tools for detecting if values are defined in the template or in the README.

javsalgar: For example, you are doing it with the values below

bitnami/elasticsearch/values.yaml
```diff
@@ -130,6 +130,35 @@ discovery:
 #     mountPath: /certs
 #     readOnly: true
 
+## initdb scripts
+## Specify dictionary of scripts to be run at first boot
+## Alternatively, you can put your scripts under the files/docker-entrypoint-initdb.d directory
+##
+# initdbScripts:
+#   my_init_script.sh: |
```
javsalgar, line 138: Be consistent with the amount of # in the examples. Below you use two but here only one.

bitnami/elasticsearch/templates/configmap-initscripts.yaml
marcosbc: OK! Removed

bitnami/elasticsearch/README.md
marcosbc: OK, changed, although I still am more a fan of initdbScriptsConfigMap.

bitnami/elasticsearch/values.yaml
marcosbc: OK.

bitnami/elasticsearch/values.yaml
marcosbc: Changed.

bitnami/elasticsearch/templates/master-statefulset.yaml
marcosbc: I agree that it would be the ideal situation, but for most cases this will be enough. I want to avoid making this a major version change of the chart, so I think this should be left out-of-scope.

The problem would come when someone wanted to change an Elasticsearch clustering flag, which is already performed with the chart. In that cases what you say would be required, but that means we should do it for everything: init scripts, environment variables **and volumes**. Currently volumes are mounted equal for all Elasticsearch containers, and changing that would require us to increase the major version.

bitnami/elasticsearch/README.md
marcosbc: OK, added.

PR #2551, opened 2020-05-07
[bitnami/elasticsearch] Fix gateway configuration based on node number

carrodher:
Add `ELASTICSEARCH_TOTAL_NODES` env. variable to configure the gateway section of the config file. Now, there is an issue when the number of nodes (data + master) is smaller than 3.

This PR needs changes in the ES image as well.

**Checklist** 

### DIFF
--- a/bitnami/elasticsearch/templates/coordinating-deploy.yaml
+++ b/bitnami/elasticsearch/templates/coordinating-deploy.yaml
@@ -71,6 +71,8 @@ spec:
               value: {{ .Values.name | quote }}
             - name: ELASTICSEARCH_CLUSTER_HOSTS
               value: {{ include "elasticsearch.hosts" . | quote }}
+            - name: ELASTICSEARCH_TOTAL_NODES
+              value: {{ add .Values.master.replicas .Values.data.replicas | quote }}
             {{- if .Values.plugins }}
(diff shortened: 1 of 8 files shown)


bitnami/elasticsearch/templates/coordinating-deploy.yaml
```diff
@@ -71,6 +71,8 @@ spec:
               value: {{ .Values.name | quote }}
             - name: ELASTICSEARCH_CLUSTER_HOSTS
               value: {{ include "elasticsearch.hosts" . | quote }}
+            - name: ELASTICSEARCH_TOTAL_NODES
+              value: {{ add .Values.master.replicas .Values.data.replicas | quote }}
```
juan131, line 75: Is it necessary to quote it? I mean, it's an integer, right?

carrodher: Maybe it is not necessary, I just adapted it from
```yaml
            - name: ELASTICSEARCH_MINIMUM_MASTER_NODES
              value: {{ add (div .Values.master.replicas 2) 1 | quote }}
```
I can try without the `quote`

juan131: Fine then

PR #2552, opened 2020-05-07
[bitnami/postgresql] Fix customMetrics example

mcritchlow:
The defined metric name `size_bytes` did not correspond to the `AS`
assignment in the example `customMetrics` SQL query, which was `size`. I've set the `AS` to `size_bytes` to match the metric.


javsalgar:
Hi,

Thanks for the PR!

PR #2560, opened 2020-05-08
[bitnami/prometheus-operator] Add the ability for users to create AlertManager config externally

RichieSams:
Alertmanager configuration can contain confidential tokens, etc. This change allows users to create the configuration secret themselves so their tokens aren't in plaintext in the helm values.


joancafom:
Hello @RichieSams ! Thanks for your PR. 

It is a nice idea to provide with an external alternative when it comes to confidential information. However, your implementation requires the existence of a Secret named after the specific release that will be deployed. While this is not wrong, changing the boolean property to a string referring to the name of the Secret that will be used provides much more flexibility.

We already implement this approach in other Charts:
https://github.com/bitnami/charts/blob/master/bitnami/airflow/templates/_helpers.tpl#L312
https://github.com/bitnami/charts/blob/master/bitnami/airflow/templates/secrets.yaml

Could you update your PR to match this? Thanks!

bitnami/prometheus-operator/README.md
```diff
@@ -237,65 +237,66 @@ The following table lists the configurable parameters of the Prometheus Operator
...
+| `alertmanager.config`                           | Alertmanager configuration directive                                                                                           | `{"global":{"resolve_timeout":"5m"},"route":{"group_by":["job"],"group_wait":"30s","group_interval":"5m","repeat_interval":"12h","receiver":"null","routes":[{"match":{"alertname":"Watchdog"},"receiver":"null"}]},"receivers":[{"name":"null"}]}` |
+| `alertmanager.externalConfig`                   | Alertmanager configuration is created externally. If true, `alertmanager.config` is ignored, and a secret will not be created. | `false` See [docs](https://github.com/coreos/prometheus-operator/blob/master/Documentation/user-guides/alerting.md#alerting) for details                                                                                                            |
```
joancafom, line 280: With a property named 'existingSecret', you could directly indicate the name of Secret the Chart may use to read the confidential information.

RichieSams: I do like that for a user. Unfortunately, alertmanager forces the secret to use the name: `alertmanager-<alertmanager_name>`.
That's why I went with a bool. Since the name is fixed, and is templatized.

RichieSams: I do like that for a user. Unfortunately, prometheus-operator forces the secret to use the name: alertmanager-<alertmanager_name>. https://github.com/coreos/prometheus-operator/blob/master/Documentation/user-guides/alerting.md#alerting
That's why I went with a bool. Since the name is fixed, and is templatized.

joancafom: Yes, you are right! Prometheus-operator enforces some naming rules over the Secrets (_alertmanager-<alertmanager_name>_ as you mentioned). Let's analyse that pattern:

**alertmanager-**
This part is always fixed, as it is imposed by Prometheus' conventions.

**<alertmanager_name>**
If we take a look to our `alertmanager/alertmanager.yaml` we find that the `name` uses the following template function:

         `name: {{ template "prometheus-operator.alertmanager.fullname" . }}`

This template uses the release name. Thus, although the Secret must always start with the _alermanager_ prefix, the name is not fixed and depends on the specific release.

That's why I think a string (as in the examples I have shown you) may be more convenient. Thanks!

RichieSams: The secret name isn't used anywhere in the chart. The boolean `externalConfig` just configures whether the chart will create secret itself or not.

externalConfig: true -> chart creates the secret, and names it `alertmanager-{{ template "prometheus-operator.alertmanager.fullname" . }}`

externalConfig: false -> chart does nothing. The user creates the secret themselves, and *must* name it `alertmanager-{{ template "prometheus-operator.alertmanager.fullname" . }}`

joancafom: Hi again @RichieSams ! You are absolutely right! Thanks for your contribution 😄

PR #2566, opened 2020-05-09
[bitnami/external-dns] add alibaba cloud configuration parameters

kahkeng:
**Description of the change**

This exposes some of the command-line arguments of external-dns that are specific to Alibaba Cloud into values.yaml, specifically, `--alibaba-cloud-zone-type` and `--alibaba-cloud-config-file`. Additionally, the latter is generated automatically via secret from `accessKeyId` and `accessKeySecret`, or taken from an existing secret specified by `secretName`.

**Benefits**

This lets people use external-dns via Alibaba Cloud keys. Due to external-dns binary requiring `--alibaba-cloud-config-file` to point to a file, it wasn't possible to use those keys simply by specifying `extraArgs` or `extraEnv` in the chart.

**Possible drawbacks**

- We don't let people specify the mount path currently (it's fixed at `/etc/kubernetes/alibaba-cloud.json` which is the default on the external-dns binary).

**Additional information**

**Checklist** 


bitnami/external-dns/templates/_helpers.tpl
```diff
@@ -163,6 +167,20 @@ Return the name of the Secret used to store the passwords
 {{- end -}}
 {{- end -}}
 
+{{- define "external-dns.alibabacloud-credentials" -}}
```
miguelaeh, line 170: From my point, it would be better to add this to a configmap instead of to a helper.

bitnami/external-dns/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: external-dns
-version: 2.22.1
+version: 2.22.2
```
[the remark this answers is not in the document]
kahkeng: Got it, I bumped it to 2.23.0.

bitnami/external-dns/templates/_helpers.tpl
kahkeng: Although it's called Alibaba Cloud configuration file, it actually stores access key id and secret, so I think it might be better to put it into a Secret instead of ConfigMap? 

Using the helper template is also the method used by the AWS and Azure provider for their respective credential files, so I was following the existing pattern here. What do you think?

miguelaeh: I agree a secret would be better.
Regarding the rest of the providers, maybe we should leave it in this way to maintain consistency and create a different PR to improve that, so you don't need to change it now.

PR #2599, opened 2020-05-15
[bitnami/mongodb-sharded] improve network policy to include internal traffic

jsemander:
**Description of the change**

Add Network Policy for RabbitMQ services to allow the cluster connectivity to support a denied network policy between services, only allowed is using the `mongodb-sharded-client` labels on the service.

**Benefits**

Ability to support a deny all default network policy so that services can be provided access based on need.

**Possible drawbacks**

If configured incorrectly, the cluster sets may not work with denying the network traffic

**Applicable issues**

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/mongodb-sharded/templates/networkpolicy.yaml
+++ b/bitnami/mongodb-sharded/templates/networkpolicy.yaml
@@ -1,0 +1,44 @@
+{{- if .Values.networkPolicy.enabled }}
+kind: NetworkPolicy
+{{- if .Values.global.networkPolicy }}
+apiVersion: {{ .Values.global.networkPolicy.apiVersion . | default "networking.k8s.io/v1" }}
+{{- else }}
+apiVersion: networking.k8s.io/v1
+{{- end }}
+metadata:
+  name: {{ template "mongodb-sharded.fullname" . }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app.kubernetes.io/name: {{ template "mongodb-sharded.name" . }}
+    helm.sh/chart: {{ template "mongodb-sharded.chart" . }}
+    app.kubernetes.io/instance: {{ .Release.Name }}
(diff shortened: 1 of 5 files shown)


bitnami/mongodb-sharded/templates/networkpolicy.yaml
```diff
@@ -0,0 +1,44 @@
+{{- if .Values.networkPolicy.enabled }}
+kind: NetworkPolicy
+{{- if .Values.global.networkPolicy }}
+apiVersion: {{ .Values.global.networkPolicy.apiVersion . | default "networking.k8s.io/v1" }}
+{{- else }}
+apiVersion: networking.k8s.io/v1
+{{- end }}
```
andresbono, line 7: I'm not sure if this block is something we may want to move to a helpers function. Can you add your thoughts, @juan131?

juan131: This should be part of `bitnami-common` chart. See this PR for instance: https://github.com/bitnami/charts/pull/2732

Then you can do sth like:

```suggestion
apiVersion: {{ include "common.capabilities.networkPolicy.apiVersion" . }}
```

andresbono: Thanks for your input. It makes more sense doing it like that.

andresbono: @jsemander, after double-checking, [bitnami/common](https://github.com/bitnami/charts/tree/master/bitnami/common) is not compatible with `bitnami/mongodb-sharded` yet. For this reason, we think it would be better to move the block to a helpers macro. Could you please update the PR?

andresbono:
@jsemander we added new comments to the PR. Could you check them? Thanks!

PR #2620, opened 2020-05-19
[bitnami/postgresql] Support all metrics env vars

### LINKED ISSUE #2611: [postgres] Add option to disable default metrics
**Which chart**:
Postgres (and presumably HA), latest.

My use case is to use the Postgres helm chart in our application and expose metrics for Prometheus/Grafana. Our DBA has no interest in the default metrics, or at least wants enough disabled that I'd prefer to build up what we want via `customMetrics`. 

The exporter image supports disabling the default metrics [as a flag](https://github.com/wrouesnel/postgres_exporter#flags) via `disable-settings-metrics` or [an environment variable](https://github.com/wrouesnel/postgres_exporter#environment-variables) via `PG_EXPORTER_DISABLE_DEFAULT_METRICS`, however this option does not appear to be made available via the Helm chart. 

What would be great, is a chart option along the lines of `metrics.defaultMetrics.enabled`, presumably with a default value of `true`.

Not having this isn't the end of the world :) I won't include these defaults in our Grafana dashboard, and we can definitely make due.

Also, if I'm just missing a way to leverage this with the chart as-is, I apologize!

**Additional context**
N/A

mcritchlow:
**Description of the change**

- Add `metrics.extraEnvVars` parameter to `README.md`
- Add sample `metrics.extraEnvVars` to `values.yaml` and `values-production.yaml`
- Add implementation to `statefulset.yaml`, using existing template.

**Benefits**

Provides support for any environment variable that the postgres-exporter image supports without extra chart configuration for maintainers.

**Possible drawbacks**

There are already some explicitly defined parameters which map to existing environment variables supported by the charts. This could potentially cause confusion, but the pattern seems to be used elsewhere in the bitnami charts.

**Applicable issues**

  - fixes #2611

**Additional information**

This seems to be working locally. Setting some local values via a values file `/tmp/pg-values.yaml`:

```
postgresqlDatabase: testing
metrics:
  enabled: true
  extraEnvVars:
    - name: PG_EXPORTER_DISABLE_DEFAULT_METRICS
      value: "true"
```

Installing an instance of the chart:
```
helm upgrade --install --namespace=postgres-testing --values=/tmp/pg-values.yaml pg-test ./
```

Run a shell on the metrics container and inspect `env` for the defined variable:
```
I have no name!@pg-test-postgresql-0:/opt/bitnami/postgres-exporter$ env | grep PG_EXPORTER
PG_EXPORTER_DISABLE_DEFAULT_METRICS=true
```

**Checklist** 


bitnami/postgresql/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: postgresql
-version: 8.9.8
+version: 8.9.9
```
juan131, line 3: Please bump the minor version instead, since we're adding a new feature to the chart:
```suggestion
version: 8.10.0
```

bitnami/postgresql/values-production.yaml
```diff
@@ -531,6 +531,13 @@ metrics:
   #       - size_bytes:
   #           usage: "GAUGE"
   #           description: "Size of the database in bytes"
+  ## An array to add extra env vars to configure postgres-exporter
+  ## see: https://github.com/wrouesnel/postgres_exporter#environment-variables
+  ## For example:
+  #  extraEnvVars:
+  #  - name: PG_EXPORTER_DISABLE_DEFAULT_METRICS
+  #    value: "true"
```
juan131, line 539: Set this parameter to an empty object by default:
```suggestion
  ## An array to add extra env vars to configure postgres-exporter
  ## see: https://github.com/wrouesnel/postgres_exporter#environment-variables
  ## For example:
  ##  extraEnvVars:
  ##  - name: PG_EXPORTER_DISABLE_DEFAULT_METRICS
  ##    value: "true"
  ##
  extraEnvVars: {}
```

bitnami/postgresql/values.yaml
```diff
@@ -537,6 +537,14 @@ metrics:
   #       - size_bytes:
   #           usage: "GAUGE"
   #           description: "Size of the database in bytes"
+  #
+  ## An array to add extra env vars to configure postgres-exporter
+  ## see: https://github.com/wrouesnel/postgres_exporter#environment-variables
+  ## For example:
+  #  extraEnvVars:
+  #  - name: PG_EXPORTER_DISABLE_DEFAULT_METRICS
+  #    value: "true"
```
juan131, line 546: Same here:
```suggestion
  ## An array to add extra env vars to configure postgres-exporter
  ## see: https://github.com/wrouesnel/postgres_exporter#environment-variables
  ## For example:
  ##  extraEnvVars:
  ##  - name: PG_EXPORTER_DISABLE_DEFAULT_METRICS
  ##    value: "true"
  ##
  extraEnvVars: {}
```

bitnami/postgresql/README.md
```diff
@@ -205,6 +205,7 @@ The following tables lists the configurable parameters of the PostgreSQL chart a
 | `metrics.image.pullPolicy`                    | PostgreSQL Exporter Image pull policy                                                                                                                                    | `IfNotPresent`                                                |
 | `metrics.image.pullSecrets`                   | Specify Image pull secrets                                                                                                                                                | `nil` (does not add image pull secrets to deployed pods)      |
 | `metrics.customMetrics`                       | Additional custom metrics                                                                                                                                                 | `nil`                                                         |
+| `metrics.extraEnvVars`                        | Extra environment variables to add to exporter  | `nil`    |
```
juan131, line 208: We should mention that it's being evaluated as a template (you're using  "postgresql.tplValue" function):
```suggestion
| `metrics.extraEnvVars`                        | Extra environment variables to add to exporter  | `{}` (evaluated as a template)   |
```

mcritchlow:
@juan131 - Thanks for the feedback!

Definitely agree with your suggestions. I just made them and pushed an updated commit.

PR #2621, opened 2020-05-19
[bitnami/mysql] fix bug chart never use existing secret

vinhngle:
**Description of the change**

If we use values.yaml/values-production.yaml as a config template and enable `existingSecret`, it will not work because of the wrong condition in secrets.yaml.
Update the condition in secrets.yaml to locate exactly existingSecret variable.

**Benefits**

Now the chart can use existing secret if we uncomment and fill value `existingSecret` in values.yaml/values-production.yaml.

**Possible drawbacks**

**Applicable issues**

**Additional information**

**Checklist** 
- [X ] Title of the PR starts with chart name (e.g. `[bitnami/chart]`)


vinhngle:
Sure, I updated the chart version.

lexoren:
hello, 

i'm not sure it'll work better because, the stateful templates still use the pre-define name for the secret

```
            - name: MYSQL_ROOT_PASSWORD
              valueFrom:
                secretKeyRef:
                  name: {{ template "mysql.fullname" . }}
```

https://github.com/bitnami/charts/blob/master/bitnami/mysql/templates/master-statefulset.yaml#L78

@marcosbc isn't `existingSecret` supposed to be the name of the secret to use rather than a boolean indicating whether or not to create the secret?

marcosbc:
You're right @alex-oren, thanks for the PR!

PR #2639, opened 2020-05-21
[bitnami/mariadb-galera] Add JSON Schema

juan131:

**Description of the change**

This PR includes a basic JSON schema for the MariaDB Galera chart. It doesn't cover all the possible values of the chart but just the ones that, in my opinion, are more likely to be modified.

**Checklist**


bitnami/mariadb-galera/values-production.yaml
```diff
@@ -186,10 +186,10 @@ ldap:
   ## Enable LDAP support
   ##
   enabled: false
-  uri:
-  base:
-  binddn:
-  bindpw:
+  uri: ""
```
andresmgot, line 189: I'm curious, is there a linter or something for things like this?

bitnami/mariadb-galera/values.schema.json
```diff
@@ -0,0 +1,193 @@
...
+      "form": true,
+      "properties": {
+        "create": {
+          "type": "boolean",
+          "form": true,
+          "title": "Create RBAC rules",
+          "description": "Specify whether RBAC resources should be created and used"
+        }
+      }
+    },
```
andresmgot, line 163: is this worth a parameter? I don't think disabling RBAC is a common use case

juan131: it's disabled by default. This parameter allows you enabling it

bitnami/mariadb-galera/values-production.yaml
juan131: This specific behaviour escapes linter's control

bitnami/mariadb-galera/values.schema.json
andresmgot: mmh, by default this chart doesn't work with a cluster with RBAC enabled?

PR #2643, opened 2020-05-21
[bitnami/thanos]: Add HPA for Thanos storegateway

hwoarang:
Add HPA resource for Thanos Store Gateway so we can dynamically scale in
and out based on resource utilization.

**Benefits**

- Automatic scaling instead of hardcoded number of replicas

**Checklist**


hwoarang:
Hello @juan131 do you have any thoughts on this (or someone else who have reviewed this chart in the past :)). I want to send further PRs for Thanos but I would like to avoid to resolve file conflicts :)

bitnami/thanos/Chart.yaml
```diff
@@ -15,4 +15,4 @@ maintainers:
 name: thanos
 sources:
   - https://github.com/bitnami/bitnami-docker-thanos
-version: 0.5.3
+version: 0.5.4
```
dani8art, line 18: I would say it should be a minor, it is a new feature for the chart.

hwoarang: Thank you @dani8art I pushed an updated tree

PR #2651, opened 2020-05-25
[bitnami/external-dns] Support RFC2136 minTTL

joancafom:
Add support for new argument [introduced](https://github.com/kubernetes-sigs/external-dns/blob/master/CHANGELOG.md#v070---2020-03-10) in v0.7.0 that let the user define a minimal TTL time for records.

**Description of the change**
A new argument specifying was added to the rest of the existing RFC2136 group of args.

**Benefits**
The end user can define the minimal TTL time for records.

**Possible drawbacks**
None

**Applicable issues**

  - fixes bitnami/bitnami-docker-external-dns/issues/5

**Additional information**

**Checklist** 


bitnami/external-dns/values-production.yaml
```diff
@@ -300,6 +300,7 @@ rfc2136:
   tsigSecretAlg: hmac-sha256
   tsigKeyname: externaldns-key
   tsigAxfr: true
+  minTTL: "0s"
```
dani8art, line 303: I would say the unit is optional [here](https://github.com/kubernetes-sigs/external-dns/blob/master/docs/ttl.md#configure-dns-record-ttl-time-to-live) and it is seconds by default, so you can avoid it. Also, I would add some docs about `time.Duration` and `time.ParseDuration` such as possible values.

```yaml
  ## Default unit `s`. Possible values [ns, us, ms, s, m, h], see more https://golang.org/pkg/time/#ParseDuration
  minTTL: "0"
```

bitnami/external-dns/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: external-dns
-version: 3.0.2
+version: 3.0.3
```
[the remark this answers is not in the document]
joancafom: That is true! Thanks!

bitnami/external-dns/values-production.yaml
joancafom: Initially, I implemented it without the unit, but errors show up. I will give it a try and post here the results. About the docs, that's a nice idea!

joancafom: This is the outcome you get when no units is provided (using a magnitude of 5):
```console
time="2020-05-27T08:11:10Z" level=fatal msg="flag parsing error: time: missing unit in duration 5"
```

dani8art: Cool, it seems they have implemented a validation in their side wrapping the original library, so let's keep the unit!

PR #2656, opened 2020-05-25
[bitnami/thanos]: Add bucketweb Ingress resource

hwoarang:
**Description of the change**

Add the ability to enable an Ingress resource for the bucketweb service.
This allows accessing the web UI for the storage bucket from outside the
cluster.

**Additional information**
This builds on top of #2643 which bumps the chart to 0.5.4

**Checklist** 

### DIFF
--- a/bitnami/thanos/Chart.yaml
+++ b/bitnami/thanos/Chart.yaml
@@ -15,4 +15,4 @@ maintainers:
 name: thanos
 sources:
   - https://github.com/bitnami/bitnami-docker-thanos
-version: 0.6.0
+version: 0.7.0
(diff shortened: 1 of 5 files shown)


carrodher:
Thank you very much for your contribution, changes look good to me, just a couple of questions:
- As it is a feature addition, can you bump the chart version in a minor?
- Do you think that this feature should be added to the rest of the components (querier, etc)?

hwoarang:
@carrodher yes this PR depends on #2643 so I will fix the version when this is merged. Also I would like to add ingress to more objects. Right now we use an ingress just for the querier but I would like to transform it to follow the path in this PR so we have a common way to add ingresses.

carrodher:
Nice thanks, changes LGTM, I will wait until the other PR and the version.

hwoarang:
Hello @carlossm I have rebased the PR against latest master and used the correct version

PR #2661, opened 2020-05-26
[bitnami/postgresql] update readme regarding postgres username and password

severi:
**Update documentation regarding postgres username and password and how they affect the admin user**

**Benefits**

Makes the documentation more clear regarding how setting of a different username than `postgres` works (i.e. `postgres`stays as the admin user)

**Possible drawbacks**

Haven't tested how the functionality goes if `postgresqlDatabase`is not set.

**Checklist** 


joancafom:
Hi @severi !

Thanks for your PR ! I would like to say that this information is already documented in `values.yml` and as well in the [bitnami/bitnami-docker-postgresql](https://github.com/bitnami/bitnami-docker-postgresql#creating-a-database-on-first-run) repo:

https://github.com/bitnami/charts/blob/fe359b2b2a830e78c32e4440d08ec8cdfd499dde/bitnami/postgresql/values.yaml#L122-L128

Do you think that this information is sufficient?

Thanks!

severi:
Hey! Good point! Probably should have realized to check the values.yaml also.

In my opinion all the documentation should be either 
A) kept up-to-date
B) or if it is duplicated documentation that kind of documents how something works but not as well  as some other source of documentation (as here `README.md` vs `values.yaml`) I would just remove the duplicated documentation from one place and refer to the proper file that serves as the primary source of documentation. For example here it might make sense to remove the documentation regarding the variables from the `README.md`and just state that the variables and their docs can be found in `values.yaml`. This would save you time of updating the docs in two places and make it easier for people the find the "right" documentation and not get confused with the "not-as-thorough" docs.

Even if you decide to stick with documenting these things in both the `README.md` and `values.yaml` I would add a line that clearly states that more thorough docs can be found from `values.yaml`. The problem with this approach is that people don't always read the docs word-by-word so they might just be happy when seeing the nice table with the variable definitions and directly focus on it, missing the hint that values.yaml has more thorough examples and docs.

To summarize: As long as its easy and clear for people to find the right documentation, I'm happy with the solution 🙂

joancafom:
Thanks for your feedback @severi !

Yes, the table in `README.md` is a nice and user-friendly way of presenting all the possible parameters. Usually, more details about them can be found in `values.yaml` as it is expected that the user opens this file if it wants to perform a change (and hence, will see the note 😃). 

As you said you experience difficulty in this case I am thinking that maybe the best approach it to make an explicit reference in the `README.md` to the `values.yaml` so that anyone seeing the table is able to get more info. Something like this:

"PostgreSQL admin password (used when `postgresqlUsername` is not `postgres`, [ref](https://github.com/bitnami/charts/blob/fe359b2b2a830e78c32e4440d08ec8cdfd499dde/bitnami/postgresql/values.yaml#L122-L124))"

What do you think of it? If you are happy with this approach you can change it if you like and I'll review it 😁

severi:
In this case I did not open the `values.yaml` file of this helm chart as I had added postgresql as a dependency to my own helm chart and was overriding the values in the `values.yaml` of my helm chart, hence I managed to miss the extra documentation that was provided here. I only tried to find the info via the Github UI and the README.md was therefore the first place I saw :) 

I think it would make sense to have a separate sentence before the table stating that more detailed docs may be found in the `values.yaml` and add link to it. I assume `postgresqlUsername` is not the only variable that is more thoroughly documented in the values.yaml so therefore would make sense to have a single statement that refers to all the variables in the table

joancafom:
HI @severi !

Thanks so much for your feedback, it is extremely valuable for us and we will consider adding that sentence in the `README.md`😁  On the other hand, I think the changes in this PR will come handy for any other user reading the table. Could you bump the Chart `patch` version so as I am able to merge your changes? 

Thanks!!

severi:
> Thanks for your PR! Please, bump Chart version according to semver.

done!

PR #2665, opened 2020-05-26
[bitnami/kibana] Add Istio compatibility

irizzant:
**Description of the change**

Add Istio compatibility to the chart and allows to specify additional labels to add to pods.

**Benefits**

Add Istio compatibility to the chart and allows to specify additional labels to add to pods.

**Possible drawbacks**

**Applicable issues**

**Additional information**

**Checklist** 
- [ X] Title of the PR starts with chart name (e.g. `[bitnami/chart]`)
- [ X] If the chart contains a `values-production.yaml` apart from `values.yaml`, ensure that you implement the changes in both files


bitnami/kibana/values-production.yaml
```diff
@@ -178,7 +178,7 @@ forceInitScripts: false
 ## Service configuration
 ##
 service:
-  port: 80
+  port: 5601
```
carrodher, line 181: Do you think this change is needed? This is the service port (not the container one), for example this is the generated note (you can see it in the first deployment or using `helm status kibana)`:
```
kubectl port-forward svc/kibana 8080:80
echo "Visit http://127.0.0.1:8080 to use your application"
```

irizzant: It's needed because this is the actual port used by the container as shown [here](https://github.com/bitnami/charts/blob/d798bcaa1c965a047ffcefe8dd5c6ac00b1f5a5f/bitnami/kibana/templates/deployment.yaml#L75):
```yaml
ports:
            - name: http
              containerPort: 5601
              protocol: TCP
```
The Service works because it uses `targetPort: http` to map port 80 to 5601, but this breaks Istio routing which shows the traffic as not belonging to the mesh.

The `targetPort: http` trick is actually not needed, by changing `.Values.service.port=5601` the generated note becomes:
```
kubectl port-forward svc/kibana 8080:5601
echo "Visit http://127.0.0.1:8080 to use your application"
```

carrodher:
Hi, there is an issue when installing the chart in our pipeline with ES as a sidecar to be fully functional:
```
$ helm install kibana . --namespace assets-privileged \
--set elasticsearch.hosts[0]=localhost \
--set elasticsearch.port=9200 \
--set sidecars[0].name=elasticsearch \
--set sidecars[0].image=bitnami/elasticsearch:latest \
--set sidecars[0].imagePullPolicy=IfNotPresent \
--set sidecars[0].ports[0].name=http \
--set sidecars[0].ports[0].containerPort=9200
Error: unable to build kubernetes objects from release manifest: error validating "": error validating data: ValidationError(Deployment.spec.template.metadata): unknown field "app" in io.k8s.apimachinery.pkg.apis.meta.v1.ObjectMeta
```

irizzant:
@carrodher 
Thanks for spotting it, I've updated the PR

PR #2675, opened 2020-05-26
[bitnami/mysql] Use existingSecret value as the secret's name

alex-oren:
**Description of the change**

With the help of a new defined named template `mysql.secretName`, it's possible to use the value `existingSecret` as the name of the secret rather than having to name a secret with the fullname

**Benefits**

It allows users to use any secret rather than forcing the name `mysql.fullname` so it's easier to share the same secrets between an app and a mysql dependency.

```
mysql:
  existingSecret: my-app-secret
```

Otherwise it either forces someone to set mysql `fullnameOverrides` *or* to force the name of the common secret to be `mysql.fullname`

**Possible drawbacks**

It's not clear in the documentation if existingSecret is supposed to be a boolean or a non nil value, but if someone uses `existingSecret` as a boolean, this is a breaking change since the name of their secret won't be the same

**Applicable issues**

  - fixes #

**Additional information**

It follows the change of PR #2621 

**Checklist** 


bitnami/mysql/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: mysql
-version: 6.13.2
+version: 6.14.0
```
andresbono, line 3: Do you think this change deserves bumping the version as minor instead of using patch?

alex-oren:
> * What if you change the README.md as well to make it clear that `existingSecret` should be the name of the existing secret?

You're right, i forgot this, done ✔️ 

> * Did you test this already? For instance with `helm template`.

Yes

PR #2684, opened 2020-05-27
[bitnami/thanos]: Move querier Ingress config to querier section

hwoarang:
**Description of the change**

Thanos querier is not the only component which can benefit from an ingress
object. bucketweb, storegateway, querier etc can all use an ingress
resource to be exposed outside of cluster. As such, it's cleaner to have
the ingress configuration for querier in the relevant section instead of
having it as a global configuration option. This enables us to form a
pattern on how to introduce ingresses to the rest of the resources.

**Benefits**
- Uniformity on how the Ingress resource for each component is defined

**Possible drawbacks**

- This is a breaking change so people need to perform certain steps during the update otherwise they risk ending up without a Querier ingress.

**Checklist**

### DIFF
--- a/bitnami/thanos/README.md
+++ b/bitnami/thanos/README.md
@@ -149,6 +149,17 @@ The following tables lists the configurable parameters of the Thanos chart and t
 | `querier.pdb.create`                            | Enable/disable a Pod Disruption Budget creation                                                        | `false`                                                 |
 | `querier.pdb.minAvailable`                      | Minimum number/percentage of pods that should remain scheduled                                         | `1`                                                     |
 | `querier.pdb.maxUnavailable`                    | Maximum number/percentage of pods that may be made unavailable                                         | `nil`                                                   |
+| `querier.ingress.enabled`                       | Enable ingress controller resource                                                                     | `false`                                                 |
+| `querier.ingress.certManager`                   | Add annotations for cert-manager                                                                       | `false`                                                 |
+| `querier.ingress.hostname`                      | Default host for the ingress resource                                                                  | `thanos.local`                                          |
+| `querier.ingress.annotations`                   | Ingress annotations                                                                                    | `[]`                                                    |
+| `querier.ingress.extraHosts[0].name`            | Additional hostnames to be covered                                                                     | `nil`                                                   |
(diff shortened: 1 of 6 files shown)


bitnami/thanos/README.md
```diff
@@ -42,6 +42,24 @@ To uninstall/delete the `my-release` chart:
 helm uninstall my-release
 ```
 
+## Upgrading existing release to new major version
```
carrodher, line 45: In order to be consistent with other charts, we are used to placing this section at the end of the README:
```
 ## Upgrading

 ### To 1.0.0
...
```

bitnami/thanos/README.md
```diff
@@ -42,6 +42,24 @@ To uninstall/delete the `my-release` chart:
...
+
+### To 1.0.0
+If you are upgrading from a `<0.0.1` release you need to move your Querier Ingress information to the new values settings:
```
carrodher, line 48: Not only `0.0.1`, right? You need to implement those changes from any `0.X.X` release

bitnami/thanos/values-production.yaml
```diff
@@ -236,6 +236,57 @@ querier:
     ##
     # maxUnavailable: 1
 
+  ## Configure the ingress resource that allows you to access Thanos Querier
```
carrodher, line 239: Can you remove this section as you did in the regular _values.yaml_?

bitnami/thanos/README.md
hwoarang: Yes by bad

bitnami/thanos/README.md
hwoarang: done

bitnami/thanos/README.md
hwoarang: done

bitnami/thanos/values-production.yaml
hwoarang: yes sorry about that. fixed now

hwoarang:
@carrodher thank you for the review. I have addressed all the problems now and pushed a new tree.

PR #2694, opened 2020-05-28
[bitnami/phpmyadmin] Migrate ingress to networking.k8s.io/v1beta1

mkilchhofer:
Signed-off-by: Marco Kilchhofer <marco.kilchhofer@bedag.ch>

**Description of the change**

The phpmyadmin ingress still uses the deprecated the API group `extensions/v1beta1`.
We should use `networking.k8s.io/v1beta1` for Ingress definitions now, which is available since K8s 1.14.
https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/

**Benefits**

Compatibilty with K8s 1.22+

**Possible drawbacks**

--

**Applicable issues**

--

**Additional information**

--

**Checklist** 


mkilchhofer:
I think the check has failed because quay.io is down:
https://status.quay.io/

bitnami/phpmyadmin/templates/ingress.yaml
```diff
@@ -1,5 +1,9 @@
 {{- if .Values.ingress.enabled -}}
+{{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1beta1" }}
+apiVersion: networking.k8s.io/v1beta1
+{{- else }}
 apiVersion: extensions/v1beta1
+{{- end }}
```
juan131, line 6: Can we use a macro as we did for WordPress? Check the links below:

- https://github.com/bitnami/charts/blob/master/bitnami/wordpress/templates/ingress.yaml#L2
- https://github.com/bitnami/charts/blob/master/bitnami/wordpress/templates/_helpers.tpl#L183

mkilchhofer:
Sure, I will adapt it.

Another way would be to move this template function from wordpress to the bitnami common chart and use it from there as a subchart.

juan131:
Correct, that's another alternative.

mkilchhofer:
Okay. What's your favorite?
- implement it in `_helpers.tpl` of the phpmyadmin chart
- implement it in the common chart

mkilchhofer:
Alright, I prepared everything. I need to wait for the merge of the [common-chart](https://github.com/bitnami/charts/pull/2732) then update the `requirements.lock` via the `helm dependency update` command

bitnami/phpmyadmin/requirements.yaml
```diff
@@ -1,4 +1,9 @@
 dependencies:
+  - name: common
+    version: "^0.3.0"
```
juan131, line 3: You can use:
```suggestion
    version: 0.x.x
```
I'll take the latest `0.x.x` version by default.

bitnami/phpmyadmin/templates/ingress.yaml
```diff
@@ -1,5 +1,5 @@
 {{- if .Values.ingress.enabled -}}
-apiVersion: extensions/v1beta1
+apiVersion: {{ include "common.capabilities.ingress.apiVersion" . }}
```
juan131, line 2: Nice!

bitnami/phpmyadmin/requirements.yaml
mkilchhofer: I can do this, but as the common chart only implement this in version 0.3.0 and higher (assumption), I used this notation to allow everything <2.0.0.

Ref: https://helm.sh/docs/topics/charts/

> Apart from version constrains employing operators `= != > < >= <=` the following shorthand notations are supported
> * hyphen ranges for closed intervals, where 1.1 - 2.3.4 is equivalent to >= 1.1 <= 2.3.4.
> * wildcards x, X and *, where 1.2.x is equivalent to >= 1.2.0 < 1.3.0.
> * tilde ranges (patch version changes allowed), where ~1.2.3 is equivalent to >= 1.2.3 < 1.3.0.
> * **caret ranges (major version changes allowed), where ^1.2.3 is equivalent to >= 1.2.3 < 2.0.0.**

juan131: It's fine. Let's leave as it's.

PR #2695, opened 2020-05-28
[bitnami/mariadb-galera] clean default MariaDB configuration

k3a:
**Description of the change**

- use underscores in the default config to make it consistent with [official documentation](https://mariadb.com/kb/en/server-system-variables/). MySQL unfortunately accepts both and it makes it messy and easy to define something twice (key-buffer-size and then key_buffer_size)
- remove duplicates of key_buffer_size and innodb_file_per_table options (set them only once)
- add new lines between some subsections

**Benefits**

Increased readability of default my.cnf options, underscores consistent with official documentation, easily searchable there

**Possible drawbacks**

none

**Applicable issues**

none

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/mariadb-galera/values-production.yaml
+++ b/bitnami/mariadb-galera/values-production.yaml
@@ -240,77 +240,75 @@ mariadbConfiguration: |-
   plugin_dir=/opt/bitnami/mariadb/plugin
 
   [mysqld]
-  default-storage-engine=InnoDB
+  default_storage_engine=InnoDB
   basedir=/opt/bitnami/mariadb
   datadir=/bitnami/mariadb/data
   plugin_dir=/opt/bitnami/mariadb/plugin
   tmpdir=/opt/bitnami/mariadb/tmp
   socket=/opt/bitnami/mariadb/tmp/mysql.sock
-  pid-file=/opt/bitnami/mariadb/tmp/mysqld.pid
-  bind-address=0.0.0.0
+  pid_file=/opt/bitnami/mariadb/tmp/mysqld.pid
+  bind_address=0.0.0.0
 
   ## Character set
-  collation-server=utf8_unicode_ci
-  init-connect='SET NAMES utf8'
-  character-set-server=utf8
+  collation_server=utf8_unicode_ci
+  init_connect='SET NAMES utf8'
+  character_set_server=utf8
 
   ## MyISAM
-  key-buffer-size=32M
-  myisam-recover-options=FORCE,BACKUP
+  key_buffer_size=32M
+  myisam_recover_options=FORCE,BACKUP
 
-  ## safety
-  skip-host-cache
-  skip-name-resolve
-  max-allowed-packet=16M
-  max-connect-errors=1000000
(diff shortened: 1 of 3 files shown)


bitnami/mariadb-galera/values-production.yaml
```diff
@@ -311,7 +308,9 @@ mariadbConfiguration: |-
   wsrep_cluster_address=gcomm://
   wsrep_cluster_name=galera
   wsrep_sst_auth="root:"
-  innodb-flush-log-at-trx-commit=2
+  # Not fully ACID compliant, up to 1sec transaction loss in the event of total cluster failure (across both regions)
+  # Enabled for performance per https://mariadb.com/kb/en/configuring-mariadb-galera-cluster/
+  innodb_flush_log_at_trx_commit=0
```
marcosbc, line 313: Before, we were setting `innodb_flush_log_at_trx_commit=0` to the `[mysqld]` section,  `innodb_flush_log_at_trx_commit=1` to the `[mariadb]` section and `innodb_flush_log_at_trx_commit=2` to the `[galera]` section.

But now we are setting `innodb_flush_log_at_trx_commit=1` to `[mariadb]`, `innodb_flush_log_at_trx_commit=0` to `[galera]` and none to MySQL (should be =0).

It looks like a typo was made here as one is missing. Could you fix that? I've also seen that in the file below.

marcosbc:
Per the docs:

> innodb_flush_log_at_trx_commit=0 — This is not usually recommended in the case of standard MariaDB. However, it is a bit safer with Galera Cluster, since inconsistencies can always be fixed by recovering from another node. 

My understanding with this is they say it is a "bit safer", but that does not mean they say it is "safe". I think it is best we avoid that change for now unless it causes huge improvements that can be extrapolated to all other MariaDB Galera chart installs.

Another thing is that the PR changes the version for the chart from `2.1.4` to `2.1.5`, consistent with a patch release as per the semver rules. If we were to modify not only the variable names, but also the `innodb_flush_log_at_trx_commit` property, we should also increase the "minor" version of the chart. But for now I would like to avoid that change unless it adds significant value.

k3a:
Ok, I kept `innodb_flush_log_at_trx_commit` without change. So this commit is just fixing underscores, updating cases in some comments, adding new line and moving some InnoDB options under `## InnoDB`.

Regarding your comment on `innodb_flush_log_at_trx_commit`. Previously (in the current master commit)  https://github.com/bitnami/charts/blob/de261fe/bitnami/mariadb-galera/values.yaml :

1. It was first set as `innodb_flush_log_at_trx_commit=0` under `[mysqld]` group
2. Then set as `innodb-flush-log-at-trx-commit=1` later (still under `[mysqld]`; you see - underscores are a must!)
3. Then `innodb-flush-log-at-trx-commit=2` under `[galera]`

So there were effectively only `innodb_flush_log_at_trx_commit=1` under `[mysqld]` which is default in both MySQL and MariaDB so it could eventually be removed but I think better keep for readability, especially as it is then overridden under `[galera]` to `=2`.

Do you really want `innodb_flush_log_at_trx_commit=0` in `[mysqld]` section? That is actually unsafe option. It is eventually only "safe" under galera, not bare mysqld.

PR #2702, opened 2020-05-29
[bitnami/thanos] Add Ingress for Querier GRPC service

hwoarang:
**Description of the change**

The querier uses a GRPC service which is useful when you have stacked
Thanos deployments. For example, it's possible to have a single observer
cluster which monitors many different clusters. Each one of these
clusters, including the observer, may have its own Thanos stack running.
However, from the observer cluster, the querier needs to connect to the
each cluster's querier in order to have a global view from all the
clusters. As such, in order to do that, we need to expose the Querier
GRPC service through the ingress in order to allow cluster-to-cluster
communication.

**Benefits**
- Allows hierarchical Thanos deployments

**Checklist** 

### DIFF
--- a/bitnami/thanos/values.yaml
+++ b/bitnami/thanos/values.yaml
@@ -14,7 +14,7 @@
 image:
   registry: docker.io
   repository: bitnami/thanos
-  tag: 0.12.2-scratch-r14
+  tag: 0.12.2-scratch-r15
   ## Specify a imagePullPolicy. Defaults to 'Always' if image tag is 'latest', else set to 'IfNotPresent'
   ## ref: http://kubernetes.io/docs/user-guide/images/#pre-pulling-images
   ##
--- a/bitnami/thanos/values.yaml
+++ b/bitnami/thanos/values.yaml
@@ -288,6 +288,39 @@ querier:
     ##   key:
     ##   certificate:
 
+    ## Create an ingress object for the GRPC service. This requires an HTTP/2
+    ## capable Ingress controller (eg. traefik using AWS NLB). Example annotations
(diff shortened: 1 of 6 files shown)


bitnami/thanos/values.yaml
```diff
@@ -288,6 +288,16 @@ querier:
...
 
+    ## Create an ingress object for the GRPC service. The options that are
+    ## accepted are identical to the HTTP one listed above
+    ## to the HTTP one above
+    grpc:
+      enabled: false
+      certManager: false
+      hostname: thanos-grpc.local
+      annotations: {}
+      secrets: []
```
juan131, line 299: The `tls`, `extraHosts` and `extraTls` parameters should be (at least) listed. You can leave the commented, but at least they should be listed so other users can identify them.

hwoarang: Yes I will do that. Thank you

hwoarang:
> Hi @hwoarang
> 
> Thanks so much for this contribution!
> 
> My main concern here is that most Ingress Controllers **do not support routing non-http traffic**. Therefore, this can lead to confusion. So I guess we need to include some warning when users enable this ingress rule.
> 
> On the other hand, using a "LoadBalancer" svc type for the querier svc would be a valid alternative to connect a Querier with other Thanos stacks.

Hi @juan131 

Yes sure. It assumes that your ingress can do HTTP/2 traffic (for example when it is backed by an AWS NLB). How would you like me to address this? Should I add a warning in the values files on in the README one?

hwoarang:
Hello @juan131 I have added some extra bits of information regarding the GRPC ingress. I hope it's more clear now.

PR #2708, opened 2020-05-29
[bitnami/kafka] Refactor listeners and authentication configuration

juan131:

**Description of the change**

This PR refactors how listeners and authentication is configured on Kafka, allowing users to have different listeners for different purposes, and configure different authentication protocols on them.

The supported listeners will be:

* INTERNAL: used for inter-broker communications.
* CLIENT: used for communications with clients that are within the K8s cluster.
* EXTERNAL: used for communications with clients that are outside the K8s cluster.

To use a more complex configuration, users will have to mount their own configuration files, or set the `listeners`, `advertisedListeners` and `listenerSecurityProtocolMap` parameters as needed.

This PR also fixes an issue with Kafka exporter being unable to export metrics when authentication was enabled. 

**Benefits**

We can achieve more complex configurations without assuming so many configuration details.

**Possible drawbacks**

Users will have to adapt their values.yaml and scripts files to the new format.

**Applicable issues**

None

**Additional information**

The current Kafka image is not compatible with these changes, you will have to use the changes at this PR to test it: https://github.com/bitnami/bitnami-docker-kafka/pull/104

**Checklist** 


bitnami/kafka/templates/NOTES.txt
```diff
@@ -50,13 +53,67 @@ IPs for Kafka brokers. To complete your deployment follow the steps below:
...
+    cat > /tmp/client.properties << EOF
+    {{- if eq .Values.auth.clientProtocol "sasl_tls" }}
+    ssl.truststore.location=/tmp/kafka.truststore.jks
+    {{- if .Values.auth.jksPassword }}
+    ssl.truststore.password={{ .Values.auth.jksPassword }}
+    {{- end }}
+    {{- end }}
+    security.protocol={{ $clientProtocol }}
+    sasl.mechanism=PLAIN
+    EOF
```
marcosbc, line 86: My concern with this "cat" command is that it cannot be copy-pasted, because the EOF should be right at the beginning of a line. Can we find a better way to do this? Even if we ask the users to create a file with the specific contents.

bitnami/kafka/templates/svc-headless.yaml
```diff
@@ -8,15 +8,13 @@ spec:
   type: ClusterIP
   clusterIP: None
   ports:
-    - name: tcp-kafka
+    - name: tcp-client
```
marcosbc, line 11: Why client? Wouldn't it be "service" in any case? Same for `svc.yaml`.

bitnami/kafka/values-production.yaml
```diff
@@ -233,46 +218,100 @@ extraVolumeMounts: []
...
+  ## Authentication protocol for client and inter-broker communications
+  ## Supported values: 'plaintext', 'tls', 'mtls', 'sasl' and 'sasl_tls'
+  ## This table shows the security provided on each protocol:
+  ## | Method    | Authentication                | Encryption via TLS |
+  ## | plaintext | None                          | No                 |
+  ## | tls       | None                          | Yes                |
+  ## | mtls      | Yes (two-way authentication)  | Yes                |
+  ## | sasl      | Yes (via SASL)                | No                 |
+  ## | sasl_tls  | Yes (via SASL)                | Yes                |
+  ##
```
marcosbc, line 230: This is great! I'm sure it will be very helpful for users.

bitnami/kafka/templates/svc-headless.yaml
juan131: Because it's the port for communications with clients, while the other port is the one for inter-broker communications.

bitnami/kafka/templates/NOTES.txt
juan131: Sure, modifying it

bitnami/kafka/templates/svc-headless.yaml
marcosbc: OK, it was just a bit confusing for me since in the end it is using the listener port. Wouldn't `tcp-listener` or something similar be easier to understand? In any case, feel free to keep it as-is.

PR #2710, opened 2020-05-29
[bitnami/mongodb] Rename serviceaccount.yml to serviceaccount.yaml

sergiu-svinarciuc:
**Description of the change**

Changing template extension so it is consistent with the other templates part of this chart.

**Benefits**
Consistency + when installing the chart through Jenkins X jx cli serviceaaccount.yml is skipped and not applied as it is having a non standard YAML file extension.

**Checklist**


marcosbc:
Thanks! We've found more occurrences like the ones you mention and will fix them shortly.

sergiu-svinarciuc:
Awesome, thank you @marcosbc for a quick turnaround!

Issue #2154, opened 2020-03-27
[bitnami/rabbitmq] Add community plugin download via init-container


bjoernw:
**Which chart**:
rabbitmq

I would like to be able to specify the location of community plugins I want to use, have an init container download them, and drop them in the right directory. Happy to help add this if this is something you're interested in.

carrodher:
Hi,

currently, you can use the `extraPlugins` parameter to enable already bundled plugins. To download 3rd-party plugins there is needed to implement a solution like the one you're proposing.

In fact, we already have this feature implemented in other charts like Kibana. We're happy to accept contributions and review your PR if you think it is something you can contribute.

I will point you to some Kibana references that I think can be useful:
- https://github.com/bitnami/charts/tree/master/bitnami/kibana#installing-plugins
- https://github.com/bitnami/charts/blob/e07d9d96e8c6677051841ea169c72c044c3c2d67/bitnami/kibana/values.yaml#L58
- https://github.com/bitnami/charts/blob/master/bitnami/kibana/templates/plugins-configmap.yml
- https://github.com/bitnami/charts/blob/e07d9d96e8c6677051841ea169c72c044c3c2d67/bitnami/kibana/templates/deployment.yaml#L107
- https://github.com/bitnami/charts/blob/e07d9d96e8c6677051841ea169c72c044c3c2d67/bitnami/kibana/templates/deployment.yaml#L150

PR #2729, opened 2020-06-01
[bitnami/rabbitmq] Add community plugins

ppoliani:
**Description of the change**

Add a script that will optionally install a list of community plugins. The list of comunity plugins can be defined as an addition parameter.

```
helm instal my-release bitnami/rabbitmq  \
  --set 'rabbitmq.communityPlugins[0]=https://github.com/rabbitmq/rabbitmq-delayed-message-exchange/releases/download/v3.8.0/rabbitmq_delayed_message_exchange-3.8.0.ez'\
```

**Benefits**

We need a way to allow the chart install community plugins. At the moment we can only install and enable those plugins that come with the official distribution of Rabbitmq.

**Applicable issues**

  - fixes https://github.com/bitnami/charts/issues/2154

**Checklist** 


bitnami/rabbitmq/templates/_helpers.tpl
```diff
@@ -24,6 +24,17 @@ If release name contains chart name it will be used as a full name.
 {{- end -}}
 {{- end -}}
 
+{{- define "rabbitmq.labels" -}}
```
miguelaeh, line 27: We know that the tags we are using at https://github.com/bitnami/charts/blob/master/bitnami/rabbitmq/templates/statefulset.yaml are deprecated, but changing them will need a major version because it breaks the backwards compatibility. To be consistent with the rest of the chart, please delete this section and update the configmap with the labels we use in that file.
This is something we plan to change in the next major version.

bitnami/rabbitmq/templates/plugins-configmap.yaml
```diff
@@ -0,0 +1,27 @@
...
+    pluginPath=/opt/bitnami/rabbitmq/plugins
+    downloadPath=/tmp/rabbitmq-plugins
+    mkdir $downloadPath
+    {{- $totalPlugins := len .Values.rabbitmq.communityPlugins }}
+    echo "Total plugins defined in chart installation: {{ $totalPlugins }}"
+    {{- range $i, $plugin := .Values.rabbitmq.communityPlugins }}
+    echo "Installing plugin {{ add $i 1 }} out of {{ $totalPlugins }}: {{ $plugin.name }}"
+    curl -X GET "{{ $plugin.url }}" -L -o $downloadPath/{{ $plugin.name }}.zip
+    unzip $downloadPath/{{ $plugin.name }}.zip -d $downloadPath
+    ls $downloadPath
```
miguelaeh, line 21: I think this is not necessary

bitnami/rabbitmq/templates/plugins-configmap.yaml
```diff
@@ -0,0 +1,27 @@
...
+    mv $downloadPath/*.ez $pluginPath/
+    rm $downloadPath/*
```
miguelaeh, line 23: I would use `rm -rf $downloadPath`

bitnami/rabbitmq/templates/plugins-configmap.yaml
```diff
@@ -0,0 +1,27 @@
...
+    echo $RABBITMQ_PLUGINS_DIR
+    pluginPath=/opt/bitnami/rabbitmq/plugins
+    downloadPath=/tmp/rabbitmq-plugins
+    mkdir $downloadPath
+    {{- $totalPlugins := len .Values.rabbitmq.communityPlugins }}
+    echo "Total plugins defined in chart installation: {{ $totalPlugins }}"
+    {{- range $i, $plugin := .Values.rabbitmq.communityPlugins }}
+    echo "Installing plugin {{ add $i 1 }} out of {{ $totalPlugins }}: {{ $plugin.name }}"
+    curl -X GET "{{ $plugin.url }}" -L -o $downloadPath/{{ $plugin.name }}.zip
+    unzip $downloadPath/{{ $plugin.name }}.zip -d $downloadPath
```
miguelaeh, line 20: Is not possible to unzip the plugins directly in the plugins directory? Doing that we wil save to move them later

bitnami/rabbitmq/templates/statefulset.yaml
```diff
@@ -100,6 +100,9 @@ spec:
                {{- if and .Values.persistence.enabled .Values.forceBoot.enabled }}
                if [ -d "{{ .Values.persistence.path }}/mnesia/${RABBITMQ_NODENAME}" ]; then rabbitmqctl force_boot; fi
                {{- end }}
+               {{- if .Values.rabbitmq.communityPlugins }}
+               ./opt/scripts/install-plugins.sh
```
miguelaeh, line 104: did you wanted to use
```
. /opt/scripts/install-plugins.sh
```
?

bitnami/rabbitmq/templates/statefulset.yaml
```diff
@@ -280,6 +287,12 @@ spec:
               - key: tls.key
                 path: server_key.pem
         {{- end }}
+        {{- if .Values.rabbitmq.communityPlugins }}
+        - name: {{ template "rabbitmq.fullname" . }}-plugins
+          configMap:
+            name: {{ template "rabbitmq.fullname" . }}-plugins
+            defaultMode: 0777
```
miguelaeh, line 294: Why is this needded? I think the configmap only needs execution and read permissions

bitnami/rabbitmq/values.yaml
```diff
@@ -107,6 +107,10 @@ rabbitmq:
   ## Plugins to enable
   plugins: "rabbitmq_management rabbitmq_peer_discovery_k8s"
 
+
+  ## Community plugins to install and enable
+  communityPlugins: {}
```
miguelaeh, line 112: Should not this be an array? the same applies to the values-production.yaml

bitnami/rabbitmq/README.md
```diff
@@ -68,7 +68,8 @@ The following table lists the configurable parameters of the RabbitMQ chart and
...
 | `rabbitmq.plugins`                        | List of plugins to enable                                                                                                                                  | `rabbitmq_management rabbitmq_peer_discovery_k8s`                                           |
-| `rabbitmq.extraPlugins`                   | Extra plugings to enable                                                                                                                                   | `nil`                                                                                       |
+| `rabbitmq.extraPlugins`                   | Extra plugings to enable                                                                                                                                   | `nil`          
+| `rabbitmq.communityPlugins`               | Community plugins to install and enable         | {}                                                |
```
miguelaeh, line 72: I think this should be an array (`[]`) instead of an object

bitnami/rabbitmq/templates/_helpers.tpl
ppoliani: @miguelaeh sure, I'll do that

bitnami/rabbitmq/templates/plugins-configmap.yaml
ppoliani: sounds like a better approach, indeed

bitnami/rabbitmq/templates/statefulset.yaml
ppoliani: I agree; that's a leftover from things I tried

bitnami/rabbitmq/values.yaml
ppoliani: We need two values for each plugin; 

1. The url to download the zip file
2. The name of the plugin file when we unzip it.

For example the urls might be 

https://bintray.com/rabbitmq/community-plugins/download_file?file_path=3.7.x%2Frabbitmq_delayed_message_exchange%2Frabbitmq_delayed_message_exchange-20171201-3.7.x.zip

However if I don't use a specific name I can't tell what the content of the rabbitmq_delayed_message_exchange-20171201-3.7.x.zip are. There is nothing that guarantees the name would be `abbitmq_delayed_message_exchange-20171201-3.7.`. So basically we need a safe way of renaming the plugin to a predictable name so we can later on call  `rabbitmq-plugins enable <plugin_name>`

bitnami/rabbitmq/templates/statefulset.yaml
ppoliani: I just wan to execute that sh file. `./opt/scripts/install-plugins.sh` does the job

PR #2734, opened 2020-06-02
[bitnami/kong] Fix for prometheus permissions

mostafahussein:
**Description of the change**

This PR fixes an issue that happens when metrics are enabled and service monitor namespace is specified. Prometheus user won't be able to get any information from kong's namespace so these two resources need to be created.

**Checklist** 


bitnami/kong/README.md
```diff
@@ -215,6 +215,7 @@ The following tables list the configurable parameters of the kong chart and thei
 | `metrics.service.annotations`                   | Annotations for Prometheus metrics service                                                                                                            | `Check values.yaml file`                                                                                    |
 | `metrics.serviceMonitor.enabled`                | if `true`, creates a Prometheus Operator ServiceMonitor (also requires `metrics.enabled` to be `true`)                                                | `false`                                                                                                     |
 | `metrics.serviceMonitor.namespace`              | Namespace in which Prometheus is running                                                                                                              | `nil`                                                                                                       |
+| `metrics.serviceMonitor.serviceAccount`         | Service account used by Prometheus                                                                                                                    | `nil`
```
juan131, line 218: Could we follow used on other charts (e.g. MariaDB Galera or Kafka) where we have 3 different parameters: `serviceAccount.create`, `serviceAccount.name`, and `rbac.create`. This approach allow us to use a custom ServiceAccount without enabling RBAC.

Check the link below:

- https://github.com/bitnami/charts/tree/master/bitnami/kafka#rbac-parameters

bitnami/kong/templates/kong-prometheus-rbac.yaml
```diff
@@ -0,0 +1,33 @@
...
+  {{- else }}
+  - namespace: {{ .Release.Namespace }}
+  {{- end }}        
+  kind: ServiceAccount
+  {{- if .Values.metrics.serviceMonitor.serviceAccount }}
+  name: {{ .Values.metrics.serviceMonitor.serviceAccount }}
+  {{- else }}
+  name: {{ template "kong.serviceAccount" . }}
+  {{- end }}
+{{- end }}
```
juan131, line 33: Please split this YAML into two files: `role.yaml` and `rolebinding.yaml`
Just for consistency with other charts.

bitnami/kong/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: kong
-version: 1.1.7
+version: 1.1.8
```
juan131, line 3: Please bump the minor version instead, you're adding a new feature to the chart.

bitnami/kong/README.md
mostafahussein: I am not sure if we should make it like the other charts, the idea is to use an existing `serviceAccount` which will be preconfigured for prometheus.

We you decide to integrate kong with prometheus, you have to write the namespace where prometheus is running. and write the serviceAccount which is used by prometheus.

I guess the creation would be only in case you need to create a serviceAccount for prometheus itself but this would be out of this chart's scope.

Am I getting what you meant correctly?

bitnami/kong/templates/kong-prometheus-rbac.yaml
mostafahussein: done

bitnami/kong/Chart.yaml
mostafahussein: updated to 1.2.7

bitnami/kong/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: kong
-version: 1.1.7
+version: 1.2.7
```
juan131, line 3:
```suggestion
version: 1.2.0
```

bitnami/kong/README.md
juan131: Yes, you want to provide Prometheus the permissions to access the Kong services and gather the metrics. As you said, it's a different use case.

That said, you are assuming that Role,Rolebinding must be created when `metrics.enabled` and `metrics.serviceMonitor.enabled` are `true`. And that's not true on clusters where RBAC is not enabled.

That's why I still think it's necessary a `rbac.create` parameter to decide whether create these resources or not.

bitnami/kong/templates/kong-prometheus-rolebinding.yaml
```diff
@@ -0,0 +1,23 @@
...
+  {{- if .Values.metrics.serviceMonitor.namespace }}
+  - namespace: {{ .Values.metrics.serviceMonitor.namespace }}
+  {{- else }}
+  - namespace: {{ .Release.Namespace }}
+  {{- end }}        
+  kind: ServiceAccount
+  {{- if .Values.metrics.serviceMonitor.serviceAccount }}
+  name: {{ .Values.metrics.serviceMonitor.serviceAccount }}
+  {{- else }}
+  name: {{ template "kong.serviceAccount" . }}
```
juan131, line 21: Why do we want to associate the Role for Prometheus with Kong's serviceAccount?

bitnami/kong/README.md
mostafahussein: I got it, I added the `rbac.create` condition which will create the resources if needed (enabled by default because we have another ingress rbac condition in this chart which is enabled by default too)

bitnami/kong/templates/kong-prometheus-rolebinding.yaml
mostafahussein: I just added to be like this condition in `servicemonitor.yaml`, Do you have a better suggestion? I am thinking about removing it

```
  {{- if .Values.metrics.serviceMonitor.namespace }}
  namespace: {{ .Values.metrics.serviceMonitor.namespace }}
  {{- else }}
  namespace: {{ .Release.Namespace }}
  {{- end }}
```

juan131: I would make the `.Values.metrics.serviceMonitor.serviceAccount` parameter mandatory when `metrics.serviceMonitor.rbac.create` is set to `true`, instead of binding the role to the Kong's ServiceAccount.

mostafahussein: I have updated the file

bitnami/kong/templates/kong-prometheus-rolebinding.yaml
```diff
@@ -0,0 +1,21 @@
...
+subjects:
+  {{- if .Values.metrics.serviceMonitor.namespace }}
+  - namespace: {{ .Values.metrics.serviceMonitor.namespace }}
+  {{- else }}
+  - namespace: {{ .Release.Namespace }}
+  {{- end }}        
+  kind: ServiceAccount
+  {{- if .Values.metrics.serviceMonitor.rbac.create }}
+  name: {{ required "A valid .Values.metrics.serviceMonitor.serviceAccount entry required!" .Values.metrics.serviceMonitor.serviceAccount }}
+  {{- end }}
```
juan131, line 20: You're already checking that RBAC is enabled at the top of the file. We can remove this "if" statement.
```suggestion
  name: {{ required "A valid .Values.metrics.serviceMonitor.serviceAccount entry required!" .Values.metrics.serviceMonitor.serviceAccount }}
```

PR #2644, opened 2020-05-21
[bitnami/phpmyadmin] Change to non-root container approach

marcosbc:
**Description of the change**

We are working on releasing the phpMyAdmin Docker image with a non-root approach. As part of that, we need to update the chart to use the new port numbers used by the container image.

**Benefits**

Support for environments such as OpenShift, where a non-root UID is required to run a container.

**Possible drawbacks**

Breaking change with previous chart versions, as the default HTTP/HTTPS ports are changed to 8080/8443 (previously 80/443).

**Applicable issues**

  - fixes #

**Additional information**

**Checklist** 
- [x] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [x] Variables are documented in the README.md
- [x] Title of the PR starts with chart name (e.g. `[bitnami/chart]`)
- [x] If the chart contains a `values-production.yaml` apart from `values.yaml`, ensure that you implement the changes in both files

:warning: Keep in mind that if you want to make changes to the kubeapps chart, please implement them in the [kubeapps repository](https://github.com/kubeapps/kubeapps/tree/master/chart/kubeapps). This is only a synchronized mirror.

PR #2740, opened 2020-06-03
[bitnami/phpmyadmin] Add non-root notice in README and add securityContext to change UID

marcosbc:
**Description of the change**

In #2644 we updated the chart so it uses a non-root Docker image. But we were missing some stuff we usually add to other images, an upgrading notice and securityContext support.

**Benefits**

Consistency with other Bitnami charts running non-root containers

**Possible drawbacks**

None

**Applicable issues**

None

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/phpmyadmin/README.md
+++ b/bitnami/phpmyadmin/README.md
@@ -78,9 +78,8 @@ The following table lists the configurable parameters of the phpMyAdmin chart an
 | `ingress.hosts[0].tls`       | Utilize TLS backend in ingress                                                                          | `false`                                                      |
 | `ingress.hosts[0].tlsHosts`  | Array of TLS hosts for ingress record (defaults to `ingress.hosts[0].name` if `nil`)                    | `nil`                                                        |
 | `ingress.hosts[0].tlsSecret` | TLS Secret (certificates)                                                                               | `phpmyadmin.local-tls-secret`                                |
-| `securityContext.enabled`    | Enable security context for phpMyAdmin pods                                                             | `true`                                                       |
-| `securityContext.fsGroup`    | Group ID for the phpMyAdmin filesystem                                                                  | `1001`                                                       |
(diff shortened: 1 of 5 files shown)


bitnami/phpmyadmin/README.md
```diff
@@ -78,6 +78,9 @@ The following table lists the configurable parameters of the phpMyAdmin chart an
...
 | `ingress.hosts[0].tlsHosts`  | Array of TLS hosts for ingress record (defaults to `ingress.hosts[0].name` if `nil`)                    | `nil`                                                        |
 | `ingress.hosts[0].tlsSecret` | TLS Secret (certificates)                                                                               | `phpmyadmin.local-tls-secret`                                |
+| `securityContext.enabled`    | Enable security context for phpMyAdmin pods                                                             | `true`                                                       |
+| `securityContext.fsGroup`    | Group ID for the phpMyAdmin filesystem                                                                  | `1001`                                                       |
+| `securityContext.runAsUser`  | User ID for the phpMyAdmin container                                                                    | `1001`                                                       |
```
juan131, line 83: We're deprecating using `securityContext` in favor of `podSecurityContext` and `containerSecurityContext` which provides more flexibility. Check Jenkins for instance:

- https://github.com/bitnami/charts/blob/master/bitnami/jenkins/values.yaml#L116
- https://github.com/bitnami/charts/blob/master/bitnami/jenkins/templates/deployment.yaml#L33
- https://github.com/bitnami/charts/blob/master/bitnami/jenkins/values.yaml#L123
- https://github.com/bitnami/charts/blob/master/bitnami/jenkins/templates/deployment.yaml#L61

As you can see the `.enabled` is also removed so we can treat the parameter as a YAML that's mapped to the specific value.

marcosbc: 👍

PR #2747, opened 2020-06-03
Add new chart: Spring Cloud Data Flow

juan131:

## Description of the change

This PR adds a new chart to the catalog: Spring Cloud Data Flow

Spring Cloud Data Flow is a microservices-based Streaming and Batch data processing pipeline in Cloud Foundry and Kubernetes.

Please read carefully the README.md and provide feedback about anything you think that doesn't make sense.


bitnami/spring-cloud-dataflow/values.yaml
```diff
@@ -0,0 +1,735 @@
...
+    # targetCPU: 50
+    # targetMemory: 50
+
+## Spring Cloud Skipper parameters.
+##
+skipper:
+  ## Set to true to enable Spring Cloud Skipper component.
+  ## Note: it'll be also enabled if streams are enabled in Dataflow server configuration.
+  ##
+  enabled: true
```
dani8art, line 311: This could cause conflict with `server.configuration.streamingEnabled`. Since `skipper` is required if streaming is enabled I would keep them as only one parameter (`server.configuration.streamingEnabled`) and make the rest of the logic, which now depends on `skipper.enabled`, depends on `server.configuration.streamingEnabled`.

So if we want to enable streaming then skipper will be enabled and if we want to disable streaming skipper will be disabled.

bitnami/spring-cloud-dataflow/README.md
```diff
@@ -0,0 +1,436 @@
...
+
+For each host indicated at `server.ingress.extraHosts`, please indicate a `name`, `path`, and any `annotations` that you may want the ingress controller to know about.
+
+For annotations, please see [this document](https://github.com/kubernetes/ingress-nginx/blob/master/docs/user-guide/nginx-configuration/annotations.md). Not all annotations are supported by all ingress controllers, but this document does a good job of indicating which annotation is supported by many popular ingress controllers.
+
+## Upgrading
+
+It's necessary to set the `mariadb.rootUser.password` parameter when upgrading for readiness/liveness probes to work properly. When you install this chart for the first time, unless you indicate set this parameter, a random value will be generated. Inspect the MariaDB secret to obtain the root password, then you can upgrade your chart using the command below:
```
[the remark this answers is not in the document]
juan131: done

bitnami/spring-cloud-dataflow/values.yaml
juan131: I agree and disagree. I mean, it makes no sense to disable Skipper if `server.configuration.streamingEnabled=true`. However, you may want to enable Skipper even if `server.configuration.streamingEnabled=false` (I cannot think of any use case for this at this very moment, but let's imagine someone wants to do so).
So my proposal is to create a validation in the `_helpers.tpl` to detect the 1st scenario and show an error.

juan131: Thinking about this again.. Let's imagine that you already have Skipper installed in your cluster, you might want to enable streaming without enabling Skipper (since you already have it).

dani8art: Yes, I agree with the second scenario and the proposal sounds good to me.

dani8art: Then we would need a value to provide the URL, right? something like `externalSkipper.url`

juan131: Yep, let me fix that

PR #2764, opened 2020-06-05
[bitnami/kafka] kafka-exporter-certificates chmod

guenaelgonnord:
**Description of the change**

the folder containing kafka-exporter-certificates is owned by root:root, so we need do adjust chmod permissions (0400 -> 0444). Because kafka prometheus exporter process is run by 1001:1001

**Benefits**

no more access denied from kafka prometheus exporter when accessing client certificates

**Possible drawbacks**

None

**Checklist** 


bitnami/kafka/templates/kafka-metrics-deployment.yaml
```diff
@@ -71,6 +71,6 @@ spec:
         - name: kafka-exporter-certificates
           secret:
             secretName: {{ .Values.metrics.kafka.certificatesSecret }}
-            defaultMode: 256
+            defaultMode: 0444
```
joancafom, line 74: What about changing this to `0440`? In that case, `1001` still has access to the files as it belongs to the `root` group. That way, we are being more restrictive, what do you think?

guenaelgonnord: yes it's better

PR #2766, opened 2020-06-05
[bitnami/external-dns] Support --exclude-domains flags

haines:
**Description of the change**

Add a new value `excludeDomains` to bitnami/external-dns, which accepts a list of subdomains to be excluded and adds [`--exclude-domains`](https://github.com/kubernetes-sigs/external-dns/blob/v0.7.2/pkg/apis/externaldns/types.go#L319) flags to the external-dns command arguments.

**Benefits**

This allows exclusion of more than one domain; currently the only way to add this argument is via `extraArgs`, which is a map so the flag can't be repeated.

**Possible drawbacks**

None.

**Applicable issues**

None.

**Checklist** 


haines:
@juan131 done! Thanks for reviewing 🙂

PR #2638, opened 2020-05-21
[bitnami/grafana] Add Grafana Image Renderer as optional deployment to render images

juan131:
Signed-off-by: juan131 <juan@bitnami.com>

**Description of the change**

This PR adds the possibility to create a deployment with Grafana Image Renderer, so it can be used as a rendering service for PNG images. It also adds the possibility to expose/manage the Prometheus metrics emitted by this container.

**Benefits**

You don't need to install the `grafana-image-renderer` plugin anymore.

**Possible drawbacks**

None

**Applicable issues**

Related to https://github.com/bitnami/bitnami-docker-grafana/issues/8 and https://github.com/bitnami/bitnami-docker-grafana/issues/7

**Checklist**

- [x] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [x] Variables are documented in the README.md
- [x] Title of the PR starts with chart name (e.g. `[bitnami/chart]`)
- [x] If the chart contains a `values-production.yaml` apart from `values.yaml`, ensure that you implement the changes in both files

PR #2773, opened 2020-06-06
[bitnami/grafana] add label to Grafana

alex1989hu:
**Description of the change**

Label based Kubernetes network policies can not properly distinguish Grafana and the recently added Image Renderer due to common label.

Image Renderer already have a separate label, thus Grafana also shall have a separate label to fulfill need for consistency.

Signed-off-by: Alex Szakaly <alex.szakaly@gmail.com>

**Benefits**

Label based Kubernetes network policies is going to properly distinguish Grafana and Image Renderer.

**Possible drawbacks**

Selector labels are changed, Chart version needed to be bumped to inform end users. Please verify `MAJOR` or `MINOR` Chart version change.

**Applicable issues**

N/A

**Additional information**
Related PR: https://github.com/bitnami/charts/pull/2638

**Checklist** 

### DIFF
--- a/bitnami/grafana/Chart.yaml
+++ b/bitnami/grafana/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: grafana
-version: 2.1.3
+version: 3.0.0
 appVersion: 7.0.5
 description: Grafana is an open source, feature rich metrics dashboard and graph editor for Graphite, Elasticsearch, OpenTSDB, Prometheus and InfluxDB.
 keywords:
--- a/bitnami/grafana/README.md
+++ b/bitnami/grafana/README.md
@@ -373,3 +373,15 @@ The [Bitnami Grafana](https://github.com/bitnami/bitnami-docker-grafana) image s
 
 Persistent Volume Claims are used to keep the data across deployments. This is known to work in GCE, AWS, and minikube.
 See the [Parameters](#parameters) section to configure the PVC or to disable persistence.
+
+## Upgrading
+
+### To 3.0.0
+
+Deployment label selector is immutable after it gets created, so you cannot "upgrade".
+
+In https://github.com/bitnami/charts/pull/2773 the deployment label selectors of the resources were updated to add the component name. Resulting in compatibility breakage.
+
+In order to "upgrade" from a previous version, you will need to [uninstall](#uninstalling-the-chart) the existing chart manually.
+
+This major version signifies this change.
--- a/bitnami/grafana/templates/configmap.yaml
+++ b/bitnami/grafana/templates/configmap.yaml
@@ -3,6 +3,7 @@ kind: ConfigMap
 metadata:
   name: {{ include "grafana.fullname" . }}-envvars
   labels: {{- include "grafana.labels" . | nindent 4 }}
+    app.kubernetes.io/component: grafana
 data:
   GF_SECURITY_ADMIN_USER: {{ .Values.admin.user | quote }}
   {{- if .Values.imageRenderer.enabled }}
--- a/bitnami/grafana/templates/dashboard-provider.yaml
+++ b/bitnami/grafana/templates/dashboard-provider.yaml
@@ -4,6 +4,7 @@ kind: ConfigMap
 metadata:
   name: {{ include "grafana.fullname" . }}-provider
   labels: {{- include "grafana.labels" . | nindent 4 }}
+    app.kubernetes.io/component: grafana
 data:
   default-provider.yaml: |-
     apiVersion: 1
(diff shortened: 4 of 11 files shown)


bitnami/grafana/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: grafana
-version: 2.0.0
+version: 3.0.0
```
miguelaeh, line 3: Do you think this change will break the backward compatibility? If not, as this is not an addition of a new feature (that will be a minor) we should bump the patch version.

bitnami/grafana/templates/configmap.yaml
```diff
@@ -3,6 +3,7 @@ kind: ConfigMap
 metadata:
   name: {{ include "grafana.fullname" . }}-envvars
   labels: {{- include "grafana.labels" . | nindent 4 }}
+    app.kubernetes.io/component: {{ include "grafana.name" . }}
```
miguelaeh, line 6: I would change all the lines to add a static value. The component will always be `grafana`, like in the case of the image renderer where the component is `image-renderer`: https://github.com/bitnami/charts/blob/5a09f0f84ac9077502857ba8b82fa0df03d602ad/bitnami/grafana/templates/image-renderer-deployment.yaml#L7

alex1989hu: Sure, will do

alex1989hu: I have just addressed 👍

bitnami/grafana/Chart.yaml
alex1989hu: I think yes due to `spec.selector.matchLabels` is immutable field. I have just "googled" some cases:
- https://github.com/jetstack/cert-manager/issues/1927
- https://github.com/istio/istio/issues/20142

miguelaeh: Oh, I see.
Then it would be great if you can add a section like this one: https://github.com/bitnami/charts/tree/master/bitnami/dokuwiki#upgrading to the README.md indicating the changes.

rafariossaa: I think a small description of what/why is the new version and some steps the user should follow if they want to upgrade.

miguelaeh:
Hi @alex1989hu ,
Sorry for the delay, we missed the notification, I fixed the conflicts but it is still missing the Upgrade section on the README. Do you think you will have the time to add it?

alex1989hu:
@miguelaeh: sorry. My only idea is to write upgrade section to notify users to uninstall the Chart before upgrading. What do you think?

miguelaeh:
I would write something like:
```
In https://github.com/bitnami/charts/pull/2773 the labels of the resources were updated to add the component name. Resulting in compatibility breakage.

In order to upgrade from a previous version, you will need to update manually the labels of the resources modified in the above PR adding: `app.kubernetes.io/component: grafana`.
That can be done by using the `kubectl edit <resource> <name>` command.
```

alex1989hu:
@miguelaeh: thanks for the example. Unfortunately the labels are [immutable](https://github.com/bitnami/charts/pull/2773#discussion_r438362987) so you can not either patch or edit it, only way is to uninstall the chart. I will update the PR with the same style but with caution that the only way is to uninstall.

miguelaeh:
Hi @alex1989hu ,
I think the labels are immutable from a Helm upgrade perspective, but not from a `kubectl` perspective. I tested it, and you can do the following:
1. Deploy the old version of Grafana: `helm install grafana bitnami/grafana`
2. Edit de deployment manually with `kubectl` to add the new label: `kubectl edit deployment grafana`
3. Take the new version of the chart (the one with the new label) and perform the Helm upgrade: `helm upgrade grafana .`
4. Get the deployment code and check that the labels are there: `kubectl get deployment grafana -o yaml`

bitnami/grafana/README.md
```diff
@@ -373,3 +373,15 @@ The [Bitnami Grafana](https://github.com/bitnami/bitnami-docker-grafana) image s
...
 Persistent Volume Claims are used to keep the data across deployments. This is known to work in GCE, AWS, and minikube.
 See the [Parameters](#parameters) section to configure the PVC or to disable persistence.
+
+## Upgrading
+
+### To 3.0.0
+
+Deployment label selector is immutable after it gets created, so you cannot "upgrade".
+
+In https://github.com/bitnami/charts/pull/2773 the deployment label selectors of the resources were updated to add the component name. Resulting in compatibility breakage.
```
alex1989hu, line 383: Is it okay? - I need to update the link

miguelaeh: 👍

PR #2774, opened 2020-06-06
[bitnami/redmine] Inject Root CA into image

ballj:
**Description of the change**

Add a certificate authority into the redmine image. Pulls CA from kubernetes secret using a sidecar init container.

**Benefits**

- Allows LDAP authentication using LDAPS to servers with an internal certificate.
- Adds a snakeoil certificate to the container. Used can be used for TLS sessions to the ingress.

**Possible drawbacks**

None

**Applicable issues**

None

**Additional information**

None

**Checklist** 


bitnami/redmine/values.yaml
```diff
@@ -470,3 +470,17 @@ mailReceiver:
...
+# Add custom certificate authorities to redmine container
+certificates:
+  customCA: []
+  # - secret: custom-CA
+  # - secret: more-custom-CAs
+  image:
+    repository: alpine
+    # pullPolicy:
+    # pullSecrets
+    #   - myRegistryKeySecretName
```
dani8art, line 483: Could you modify this following the standard binami image structure in `values.yaml`? Please take a look at https://github.com/bitnami/charts/tree/master/bitnami/common#imageroot

Also, please use `minideb` as https://github.com/bitnami/charts/blob/master/bitnami/ghost/values.yaml#L42

dani8art:
Hi @ballj , thank you so much for the contribution, it is a good feature! 

I have one question, would this cover the case where the user wants to provide the cert.pem and the key.pem? I think we should cover this as well.

ballj:
Hi @dani8art 
Will modify the structure as suggested

Can you clarify what you mean with the cert.pem and key.pem? Is this for replacing the snakeoil cert?

javsalgar:
Hi,

I don't have much experience with Redmine, but I suppose that this CA is used to create certificates (with public key and private key), am I correct? What if the user wants to provide these as well?

bitnami/redmine/values.yaml
ballj: These have been added along with the ability to use custom certificates

dani8art:
Waiting for CI/CD

PR #2794, opened 2020-06-09
[bitnami/kube-state-metrics] Add missing validatingwebhookconfigurations collector

hwoarang:
**Description of the change**
Commit 93a4afed8cd("[bitnami/kube-state-metrics] Add missing collectors")
added more collectors but it missed the `validatingwebhookconfigurations`
one. So lets add it to the list to achieve parity with the upstream
capabilities.

**Benefits**
- Adds a missing collector

**Checklist** 


bitnami/kube-state-metrics/README.md
```diff
@@ -50,95 +50,96 @@ The command removes all the Kubernetes components associated with the chart and
 
 The following table lists the configurable parameters of the kube-state-metrics chart and their default values.
 
-| Parameter                                  | Description                                                                                                   | Default                                                    |
```
miguelaeh, line 53: Could you try to preserve the old format? It is easier to review the changes.

bitnami/kube-state-metrics/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: 1.9.7
 description: kube-state-metrics is a simple service that listens to the Kubernetes API server and generates metrics about the state of the objects.
 name: kube-state-metrics
-version: 0.3.2
+version: 0.3.3
```
miguelaeh, line 5: Adding a new parameter requires bumping the minor version

bitnami/kube-state-metrics/values-production.yaml
```diff
@@ -99,6 +99,7 @@ collectors:
   statefulsets: true
   storageclasses: true
   verticalpodautoscalers: false
+  validatingwebhookconfigurations: true
```
miguelaeh, line 102: I would set this to `false` by default, so we avoid modifications when upgrading to the new version.
The same applies to the `values.yaml` file.

bitnami/kube-state-metrics/README.md
hwoarang: Unfortunately the columns will not be aligned with this new addition

bitnami/kube-state-metrics/Chart.yaml
hwoarang: Yes you are right.

bitnami/kube-state-metrics/values-production.yaml
hwoarang: I set it to `true` because this is enabled by default upstream. Would you like me to change that? It follows the same convention like the rest of the options where those which are enabled by default upstream are also set to `true` in the chart.

rafariossaa: Yes, we would like to keep changes (when upgrading) at minimum, and let the user choose when to change the new options.

hwoarang:
@miguelaeh @rafariossaa Thank you for the comments. I have now updated the PR

Issue #2749, opened 2020-06-03
[bitnami/redis] Call failover while gracefull shutdown of master


dm3ch:
**Which chart**:
bitnami/redis 10.6.18

**Is your feature request related to a problem? Please describe.**
When you are gracefull killing redis pod that is current master, election of new master starts only after configured timeout.

**Describe the solution you'd like**
I think it would be a good idea to add preStop hook for all sentinel containers. It should check if pod is actual master and trigger failover via `sentinel failover` command and wait until new master would be elected. Script should exit only after new master election (and maybe syncing needed amount of replicas) to provide flowless experience when k8s nodes is drained for maintanance.

**Describe alternatives you've considered**

**Additional context**

WarheadsSE:
This was bumped into while exploring turning on native `bitnami/redis` dependency with Sentinels @ https://gitlab.com/gitlab-org/charts/gitlab/-/issues/2132#note_353985755

miguelaeh:
Hi @dm3ch ,
Thank you very much for your suggestion,
would you like to open a PR for that feature? We will be pleased to review it.

dm3ch:
I'll try to create PR with adding such logic

miguelaeh:
Thank you very much!

dm3ch:
reopen

PR #2820, opened 2020-06-11
[bitnami/redis] Add failover during graceful shutdown of current sentinel master

dm3ch:
/resolves https://github.com/bitnami/charts/issues/2749

Current potential problems:
- [x] I'm not sure that preStop hook for sentinel container is a good idea, cause redis container could be stopped before sentinel container. Currently I implement preStop hook in sentinel pod cause we only need this logic in configurations with sentinel enabled.

- [ ] Currently graceful shutdown script not waits until needed number of slaves would be synced.
I'm not sure is that possible. Also I don't understand if current behaviour could cause dataloss.

- [ ] I still see 127.0.0.1 logs in sentinel logs
Example of logs:
```
1:X 11 Jun 2020 20:24:21.536 * +sentinel-address-switch master mymaster 10.233.94.202 6379 ip 10.233.69.68 port 26379 for fa465dba9473f478c0e5d3a242fa0a7f7ad29b85
1:X 11 Jun 2020 20:24:22.386 * +sentinel-address-switch master mymaster 10.233.94.202 6379 ip 127.0.0.1 port 26379 for fa465dba9473f478c0e5d3a242fa0a7f7ad29b85
1:X 11 Jun 2020 20:24:23.553 * +sentinel-address-switch master mymaster 10.233.94.202 6379 ip 10.233.69.68 port 26379 for fa465dba9473f478c0e5d3a242fa0a7f7ad29b85
1:X 11 Jun 2020 20:24:24.404 * +sentinel-address-switch master mymaster 10.233.94.202 6379 ip 127.0.0.1 port 26379 for fa465dba9473f478c0e5d3a242fa0a7f7ad29b85
```

- [ ] I still haven't properly tested this change

- [ ] Didn't found way to show shutdown script logs

- [ ] I'm not sure but maybe we should check that all sentinel nodes knows (returns) new master address

- [ ] Current master and slave are separate statefullsets so on helm upgrade at same time 2 pods are shutting down (one master and one slave). I believe it's a good idea to use PodDistributionBudget to prevent that - https://github.com/bitnami/charts/pull/2822

- [ ]  I caught one time situation when one master pod and one slave pod was scheduled on same node. I believe it could cause problems if that node will disappear, so I think it's a good idea to prevent such behaviour by affinity rule (for production config) or maybe by PodDistributionBudget

### DIFF
--- a/bitnami/redis/Chart.yaml
+++ b/bitnami/redis/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: redis
-version: 10.7.1
+version: 10.7.2
 appVersion: 6.0.5
 description: Open source, advanced key-value store. It is often referred to as a data structure server since keys can contain strings, hashes, lists, sets and sorted sets.
 keywords:
--- a/bitnami/redis/templates/configmap.yaml
+++ b/bitnami/redis/templates/configmap.yaml
@@ -77,15 +77,15 @@ data:
       MY_IP=$1
       MASTER_SET=$2
 
-      if [ $(get_master_ip $MASTER_SET) == $MY_IP]; then
+      if [ $(get_master_ip $MASTER_SET) == $MY_IP ]; then
         echo "This instance is current master"
       fi
 
-      if [ $(get_master_ip $MASTER_SET) == "127.0.0.1"]; then
+      if [ $(get_master_ip $MASTER_SET) == "127.0.0.1" ]; then
         echo "Current master is set to 127.0.0.1"
       fi
 
-      if [ $(redis-cli -h 10.233.81.152 -p 6379 -a $(cat /opt/bitnami/redis/secrets/redis-password) PING) != "PONG" ]; then
+      if [ "$(timeout 1 redis-cli -h $(get_master_ip $MASTER_SET) -p 6379 -a $(cat /opt/bitnami/redis/secrets/redis-password) PING)" != "PONG" ]; then
         echo "Current master is not available"
       fi
     }
@@ -105,13 +105,13 @@ data:
 
       echo "Waiting until master would be switched..."
 
-      while [ -z $(check_master_election $MY_IP $MASTER_SET) ]; do
+      while [ -n "$(check_master_election $MY_IP $MASTER_SET)" ]; do
         echo "$(check_master_election $MY_IP $MASTER_SET)"
         sleep 5
       done
 
-      echo "Master switched to $(get_master_ip $MASTER_SET)"
       echo ""
+      echo "Master switched to $(get_master_ip $MASTER_SET)"
     fi
 
     echo "Killing redis process"
--- a/bitnami/redis/templates/configmap.yaml
+++ b/bitnami/redis/templates/configmap.yaml
@@ -73,6 +73,23 @@ data:
       echo $($REDIS_COMMAND sentinel get-master-addr-by-name mymaster | head -n 1)
     }
 
+    function check_master_election {
+      MY_IP=$1
+      MASTER_SET=$2
+
+      if [ $(get_master_ip $MASTER_SET) == $MY_IP]; then
+        echo "This instance is current master"
+      fi
+
+      if [ $(get_master_ip $MASTER_SET) == "127.0.0.1"]; then
+        echo "Current master is set to 127.0.0.1"
+      fi
+
+      if [ $(redis-cli -h 10.233.81.152 -p 6379 -a $(cat /opt/bitnami/redis/secrets/redis-password) PING) != "PONG" ]; then
+        echo "Current master is not available"
+      fi
+    }
+
     MASTER_IP=$(get_master_ip $MASTER_SET)
     MY_IP=$(hostname -i)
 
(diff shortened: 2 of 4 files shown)


dm3ch:
cc @miguelaeh

dm3ch:
cc @WarheadSE FYI

bitnami/redis/templates/configmap.yaml
```diff
@@ -51,3 +51,69 @@ data:
...
+      fi
+    }
+
+    MASTER_IP=$(get_master_ip $MASTER_SET)
+    MY_IP=$(hostname -i)
+
+    echo "My ip: $MY_IP"
+    echo "Current master ip: $MASTER_IP"
+
+    if [ $MY_IP == $MASTER_IP ]; then
```
marcosbc, line 99:
```suggestion
    if [[ "$MY_IP" = "$MASTER_IP" ]]; then
```

bitnami/redis/templates/configmap.yaml
```diff
@@ -51,3 +51,69 @@ data:
...
+        echo "$(check_master_election $MY_IP $MASTER_SET)"
+        sleep 5
+      done
+
+      echo ""
+      echo "Master switched to $(get_master_ip $MASTER_SET)"
+    fi
+
+    echo "Killing redis process"
+    kill 1
```
marcosbc, line 118: Could we use "redis-cli SHUTDOWN" here?

bitnami/redis/templates/configmap.yaml
```diff
@@ -51,3 +51,69 @@ data:
...
+    heritage: {{ .Release.Service }}
+    release: {{ .Release.Name }}
+data:
+  shutdown.sh: |-
+    #!/bin/bash
+    MASTER_SET="{{ .Values.sentinel.masterSet }}"
+    REDIS_COMMAND="redis-cli -p 26379"
+
+    function get_master_ip {
+      echo $($REDIS_COMMAND sentinel get-master-addr-by-name mymaster | head -n 1)
```
marcosbc, line 73: Instead of using shell expansion for this, you could do the following:

```suggestion
    REDIS_CMD="redis-cli"
    REDIS_ARGS=("-p" "$REDIS_SENTINEL_PORT")

    get_master_ip() {
      echo $("$REDIS_CMD" "${REDIS_ARGS[@]}" sentinel get-master-addr-by-name mymaster | head -n 1) 
```

Also, like I included in the code snippet, we should be using the environment variable for the port number.

bitnami/redis/templates/configmap.yaml
```diff
@@ -51,3 +51,69 @@ data:
...
+    }
+
+    function check_master_election {
```
marcosbc, line 76:
```suggestion
    check_master_election() {
```

bitnami/redis/templates/configmap.yaml
```diff
@@ -51,3 +51,69 @@ data:
...
+
+      if [ $(get_master_ip $MASTER_SET) == $MY_IP ]; then
+        echo "This instance is current master"
+      fi
+
+      if [ $(get_master_ip $MASTER_SET) == "127.0.0.1" ]; then
+        echo "Current master is set to 127.0.0.1"
+      fi
+
+      if [ "$(timeout 1 redis-cli -h $(get_master_ip $MASTER_SET) -p 6379 -a $(cat /opt/bitnami/redis/secrets/redis-password) PING)" != "PONG" ]; then
```
marcosbc, line 88: Keeping in mind the style improvements mentioned throughout the review of the .sh file, is it guaranteed that `/opt/bitnami/redis/secrets/redis-password` will exist?

bitnami/redis/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: redis
-version: 10.7.1
+version: 10.7.2
```
marcosbc, line 3: I would bump the minor version for this (-> 10.8.0) since it is making important changes to the functionality of the chart.

marcosbc:
We just noticed some additional stuff after a more extensive review, related to the use of TLS. It looks like this change currently would not work then `tls.enabled` is set, since in those cases the following parameters would need to be configured to redis-cli:

- TLS port instead of default port
- `--tls`
- `--cert` and the path to the certificate
- `--key` and the path to the certificate key
- `--cacert` and the path to the cacert file

rabberbock:
@marcosbc Just curious if the tls blocker is being worked on and if there is any ETA? I am testing various redis setups on k8s and ran into the issue this PR solves (I deleted the master pod and when the master came back on right away the other pods could not connect to it anymore).

dm3ch:
@rabberbock Suddenly right now I'm not completely sure that this MR solves original problem. Also i think I wouldn't have time to continue working on this to the second week of July

marcosbc:
Thanks for the update @dm3ch. For now I'll try to keep this issue open.

@rabberbock If you prefer, you could create a new PR based on the changes from here, and applying the suggestions from code review. Thanks!

rabberbock:
@marcosbc Thanks for the update! If I have some free time I will try to put something together.

PR #2822, opened 2020-06-12
[bitnami/redis] Added pdb for redis chart

dm3ch:
Currently on sentinel configuration we have 2 stateful sets. When we upgrading helm release 2 pods are shutdown at same time (1 slave and 1 master).

**Description of the change**
Adds PodDisruptionBudget

**Benefits**
Allows prevent of shutting down slave and master pods at same time.

**Possible drawbacks**

**Applicable issues**

  - fixes #

**Additional information**

**Checklist** 


bitnami/redis/values-production.yaml
```diff
@@ -712,3 +712,11 @@ podSecurityPolicy:
   ## Specifies whether a PodSecurityPolicy should be created
   ##
   create: false
+
+## Define a disruption budget
+## ref: https://kubernetes.io/docs/concepts/workloads/pods/disruptions/
+##
+podDisruptionBudget:
+  enabled: false
+  # minAvailable: 1
```
andresbono, line 721: Would it make sense to uncomment `minAvailable` so it defaults to 1 when the `podDisruptionBudget.enabled` flag is set?

dm3ch: Yep, i think it's reasonable. 

But maybe setting `maxUnavailable: 1` would be a better option? What do you think?

bitnami/redis/README.md
```diff
@@ -136,6 +136,9 @@ The following table lists the configurable parameters of the Redis chart and the
...
+| `podDisruptionBudget.enabled`        | Pod Disruption Budget toggle                                                                                                                                                                                                                                                                                                                                                                         | `false`                                                 |
+| `podDisruptionBudget.minAvailable`   | Minimum available pods                                                                                                                                                                                                                                                                                                                                                                               | `nil`                                                   |
```
andresbono, line 140:
```suggestion
| `podDisruptionBudget.minAvailable`   | Minimum available pods                                                                                                                                                                                                                                                                                                                                                                               | `1`                                                   |
```

bitnami/redis/values.yaml
```diff
@@ -712,3 +712,11 @@ podSecurityPolicy:
   ## Specifies whether a PodSecurityPolicy should be created
   ##
   create: false
+
+## Define a disruption budget
+## ref: https://kubernetes.io/docs/concepts/workloads/pods/disruptions/
+##
+podDisruptionBudget:
+  enabled: false
+  # minAvailable: 1
```
andresbono, line 721:
```suggestion
  minAvailable: 1
```

dm3ch:
I just have tested this PR and it seems that it doesn't works as expected.

It seems that PDB controls pods for each matching sts separate.
If that's so it doesn't solves described problem

andresbono:
> I just have tested this PR and it seems that it doesn't works as expected.
> 
> It seems that PDB controls pods for each matching sts separate.
> If that's so it doesn't solves described problem

I see @dm3ch. Maybe for the specific case of two sts, each with only 1 replica, it doesn't make much sense. But for cases with more replicas (which is the default) I think your PR is still useful, right?

dm3ch:
You are right. But suddenly this doesn't fit for default cluster configuration (when there's 1 master and 2 slaves)

bitnami/redis/values-production.yaml
```diff
@@ -712,3 +712,11 @@ podSecurityPolicy:
   ## Specifies whether a PodSecurityPolicy should be created
   ##
   create: false
+
+## Define a disruption budget
+## ref: https://kubernetes.io/docs/concepts/workloads/pods/disruptions/
+##
+podDisruptionBudget:
+  enabled: true
```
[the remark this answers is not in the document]
dm3ch: Thi

> If the feature is disabled by default, it shouldn't be a problem with the default configuration (1 master, 2 slaves), right?

This change doesn't create problems with default configuration. It just seems that behaviour is not changed at all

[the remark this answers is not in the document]
andresbono: Well, depending on the specific use case, users will be able to set desired values for PDB configuration. This PR eases that.

It is true that the original issue is not going to be solved, and still slaves and master will be shutting down at the same time. But it can be addressed in a different Issue/PR. To be honest, I can't think of a straightforward solution for that. Suggestions are welcome!

dm3ch:
So what we needs to get this merged?

andresbono:
Our CI/CD automation will add a new commit and then it will be merged

PR #2825, opened 2020-06-12
[bitnami/rabbitmq] Custom nodePort expose per port

FernFerret:
This adds the ability to specify the node port to expose for all of the
currently exposed ports. This should **not** be used in production but
is very useful for running a local copy of the exact production chart
one would use but in a much smaller context.

I like to keep a "miniX" version of a lot of my charts for personal development that makes it easy
to spin up a persistent, yet insecure copy of some of the middleware so I have access to it in my development environment. All that was missing was the abiltiy to expose the management/metrics ports (I added the discovery/dist ports just because they might be useful to experiment with cross cluster testing or joining a local rmq to the one in k8s.

I use the `bitnami` RabbitMQ chart in production, but I also wanted a single-node dev instance that I could spin up without any fancy ingress controllers or other systems. I was able to easily expose the AMQP ports as nodeports but not the others.

Maybe I should add a note to the README about how this should **not** be used for production, but I think that the general guidance to avoid node-ports for these types of services (and the fact that it requires 2 seperate enables) would be enough, but am open to change.

To this end, I went with adding commented versions of the new nodePorts to the `values.yaml` but they are omitted entirely from the `values-production.yaml`, again, open to suggestions/changes here.

**Checklist** 


bitnami/rabbitmq/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: rabbitmq
-version: 6.27.0
+version: 6.27.1
```
miguelaeh, line 3: Please bump the minor version as you are adding a new feature.

FernFerret: Yea sorry about this, working on the merge but it's taking me a bit longer than normal because a new major version of the chart dropped while I was making these changes (bad luck).

I'm making a few changes to the PR based on the new major chart rev (they renamed `nodeTlsPort` -> `tlsNodePort` which I like a lot, so I've changed all my ports to `xxxNodePort`. I figure now's the time to make that change since it only affects me and isn't in the baseline yet.

FernFerret: Resolved In Patch, bumped to `7.1.0`

PR #2843, opened 2020-06-15
[bitnami/rabbitmq] Rely on container initialization + refactoring to follow Helm chart best practices

juan131:

**Description of the change**

This PR refactors the whole chart to follow Helm chart best practices. This includes several breaking changes, so please take a look to the "Upgrading" section describing all of them.

The most important change is that the chart doesn't use its own initialization scripts anymore but relies on the container's own scripts instead. Due to this, compatibility with non Bitnami images is not guaranteed anymore.

**Benefits**

- Standardization.
- We don't have duplicated initialisation logic and we can rely on container's initialization scripts.

**Possible drawbacks**

It breaks backwards compatibility.

**Applicable issues**

None

**Additional information**

The current RabbitMQ image is not compatible with these changes, you will have to use the changes at this PR to test it: https://github.com/bitnami/bitnami-docker-rabbitmq/pull/118

**Checklist** 


bitnami/rabbitmq/templates/networkpolicy.yaml
```diff
@@ -31,8 +25,8 @@ spec:
             matchLabels:
               app: {{ template "rabbitmq.name" . }}
               release: {{ .Release.Name }}
-        {{- with .Values.networkPolicy.additionalRules }}
-{{ toYaml . | indent 8 }}
+        {{- if .Values.networkPolicy.additionalRules }}
+        {{- include "common.tplvalues.render" (dict "value" .Values.networkPolicy.additionalRules "context" $) | nindent 8 }}
         {{- end }}
       {{- end }}
     # Allow prometheus scrapes
```
javsalgar, line 32: (I cannot select the exact line)

Still using `metrics.port` here, change it to metricsPort

bitnami/rabbitmq/templates/statefulset.yaml
```diff
@@ -265,23 +143,79 @@ spec:
...
+          livenessProbe: {{- include "common.tplvalues.render" (dict "value" .Values.livenessProbe "context" $) | nindent 12 }}
+          {{- end }}
+          {{- if .Values.readinessProbe }}
+          readinessProbe: {{- include "common.tplvalues.render" (dict "value" .Values.readinessProbe "context" $) | nindent 12 }}
+          {{- end }}
+          # yamllint disable rule:indentation
+          {{- if .Values.resources }}
+          resources: {{- toYaml .Values.resources | nindent 12 }}
+          {{- end }}
+          {{- if and .Values.clustering.rebalance (gt (.Values.replicas | int) 1) }}
```
javsalgar, line 183: replicas -> replicaCount

bitnami/rabbitmq/values.yaml
```diff
@@ -334,126 +472,178 @@ networkPolicy:
...
-##    maxUnavailable: 1
-##    minAvailable: 1
-##
-podDisruptionBudget: {}
-## Annotations for Rabbitmq pods
-##
-podAnnotations: {}
+  ## Node port Tls
+  ##
+  # nodeTlsPort: 30671
```
javsalgar, line 494: I'd rather set it as tlsNodePort

javsalgar:
Feel free to add these other standard values if you have time (taken from the Harbor chart):

  - commonLabels
  - commonAnnotations 
  These two are set in all of the resources (some users requested it for some charts)

  - lifecycleHooks: Allow to override the lifecycle hook
 

Other remarks

  - service.extraPorts may make sense to be added to headless?

bitnami/rabbitmq/templates/statefulset.yaml
```diff
@@ -290,69 +224,46 @@ spec:
...
+            {{- end }}
   {{- else }}
   volumeClaimTemplates:
     - metadata:
         name: data
-        labels:
-          app: {{ template "rabbitmq.name" . }}
-          release: "{{ .Release.Name }}"
-          heritage: "{{ .Release.Service }}"
+        labels: {{- include "common.labels.standard" . | nindent 10 }}
```
javsalgar, line 258: ### IMPORTANT

This helper contains the chart version, so it will change when upgrading and therefore mutating the volumeClaimTemplate field. This will cause all upgrades to fail

bitnami/rabbitmq/templates/_helpers.tpl
```diff
@@ -83,104 +81,72 @@ Get the password secret.
...
-            {{- printf "storageClassName: %s" .Values.persistence.storageClass -}}
-        {{- end -}}
-    {{- end -}}
-{{- end -}}
+Return the number of bytes given a value
+following a base 2 o base 10 number system.
+Usage:
+{{ include "rabbitmq.toBytes" .Values.path.to.the.Value }}
+*/}}
+{{- define "rabbitmq.toBytes" -}}
```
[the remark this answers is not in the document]
juan131: Correct. So far we only need this on this chart but that could change.

bitnami/rabbitmq/templates/statefulset.yaml
```diff
@@ -265,23 +143,79 @@ spec:
...
-        runAsUser: {{ .Values.securityContext.runAsUser }}
-        {{- if .Values.securityContext.extra }}
-        {{- toYaml .Values.securityContext.extra | nindent 8 }}
-        {{- end }}
-      {{- end }}
+          ports:
+            - name: epmd
+              containerPort: 4369
+            - name: amqp
+              containerPort: {{ .Values.service.port }}
```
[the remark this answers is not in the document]
juan131: I agree, this is technical debt that should removed.

bitnami/rabbitmq/templates/statefulset.yaml
juan131: Good catch

bitnami/rabbitmq/templates/networkpolicy.yaml
juan131: Good catch

juan131:
> service.extraPorts may make sense to be added to headless?

I wouldn't do so. What if a user define a `nodePort` that's not valid for a headless service?

bitnami/rabbitmq/templates/_helpers.tpl
```diff
@@ -12,68 +12,66 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this
...
-{{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" -}}
-{{- end -}}
-{{- end -}}
+{{- include "common.names.fullname" . -}}
 {{- end -}}
 
 {{/*
 Create chart name and version as used by the chart label.
 */}}
 {{- define "rabbitmq.chart" -}}
```
javsalgar, line 21: This is not used

bitnami/rabbitmq/templates/statefulset.yaml
```diff
@@ -265,23 +154,97 @@ spec:
...
+              containerPort: 4369
+            {{- if .Values.metrics.enabled }}
+            - name: metrics
+              containerPort: 9419
+            {{- end }}
+            {{- if .Values.extraContainerPorts }}
+            {{- toYaml .Values.extraContainerPorts | nindent 12 }}
+            {{- end }}
+          # yamllint disable rule:indentation
+          {{- if .Values.customLivenessProbe }}
```
javsalgar, line 195: How do you disable the livenessProbe? At this moment it is impossible. For the livenessProbe I would use the old approach

bitnami/rabbitmq/templates/statefulset.yaml
```diff
@@ -191,18 +191,38 @@ spec:
...
+                - -ec
+                - |
+                  /opt/bitnami/scripts/rabbitmq/healthcheck.sh "http://{{ .Values.auth.username }}:$RABBITMQ_PASSWORD@127.0.0.1:{{ .Values.service.managerPort }}/api/healthchecks/node" '{"status":"ok"}'
+            initialDelaySeconds: {{ .Values.readinessProbe.initialDelaySeconds }}
+            periodSeconds: {{ .Values.readinessProbe.periodSeconds }}
+            timeoutSeconds: {{ .Values.readinessProbe.timeoutSeconds }}
+            successThreshold: {{ .Values.readinessProbe.successThreshold }}
+            failureThreshold: {{ .Values.readinessProbe.failureThreshold }}
+          {{- else if .Values.customReadinessProbe }}
           livenessProbe: {{- include "common.tplvalues.render" (dict "value" .Values.customReadinessProbe "context" $) | nindent 12 }}
```
javsalgar, line 224: readinessProbe

connorearl:
Is there any reason the communityPlugins value was removed?

juan131:
@connorearl you have both `plugins` and `extraPlugins` values to add your custom plugins in the chart. Why do you need a third parameter?

connorearl:
I need the ability to install community plugins that aren't bundled in the docker image such as https://github.com/rabbitmq/rabbitmq-delayed-message-exchange

Before 7.0.0 I could reference this and it would download and install the plugin on startup. 

I didn't expect the helm chart to provide this functionality initially, and thought I'd have to implement this myself when I first needed it while using the 6.x.x chart, but when I noticed it, it proved super helpful. I'm confused why would we remove this useful functionality when it already had been implemented. I don't see an apparent reason it needed to be removed.

If this stays out of the chart, I'm stuck either forking the chart and re-implementing it, or forking the bitnami docker image. That plugin is required for my application.

Either option makes management and updates a bit of a headache for users of the chart

juan131:
@connorearl I just created these two PRs to recover having a way to download custom plugins during initialization:

- https://github.com/bitnami/bitnami-docker-rabbitmq/pull/121
- https://github.com/bitnami/charts/pull/2843

juan131:
@connorearl the PR including the changes to support custom plugins was merged. It can take a couple of hours until it's ready at https://charts.bitnami.com/bitnami
Thanks so much for warning us about this missing parameter

PR #2850, opened 2020-06-16
MongoDB tolerations default value

mendrugory:
Tolerations field accepts a map, so `{}` should be the default value in `values.yaml` and `values-production.yaml`


rafariossaa:
Hi,
  Thanks for your PR. I am running our CI/CD to check it.

Issue #2622, opened 2020-05-19
TLS for Redis


asgoel:
**Which chart**:
redis 10.6.14

**Is your feature request related to a problem? Please describe.**
Now that Redis 6.0 supports TLS, would it be possible to have native support in the chart for configuring TLS?

Generally I would use the `master.configmap` directive, but I also need to be able to pass in additional secrets for the TLS certs. Is there a way I can do that?

**Describe the solution you'd like**
An easy way to set up TLS for redis using helm values, with the option to provide a secret for the TLS certs.

carrodher:
Hi, thanks for letting us know, it seems an interesting feature that we already support in other charts.
As the change is not trivial because it requires changes in the container image and in the chart, I just created an internal task in our backlog. We will incorporate it into our work process according to other tasks and priorities. In order to avoid stale bot closing the issue, I will set the `on-hold` label and we will update this thread once we implemented the required changes.

Thanks for your suggestions and for using the Bitnami charts.

asgoel:
@carrodher thanks! is there any projected timeline on when this will get done?

Meanwhile, is there any way you know of that I could get this working with the current chart? The only thing missing is the ability to mount my own secrets into the container (which I would need to do for the certs). Is this possible?

carrodher:
Hi, we are already working on it, so probably it will be released during the next week. Now it is possible to mount secrets but only for a password, we need to implement the logic to support other secrets. Anyway, PRs are welcome, if you found a way on your side to make it works don't hesitate to create a PR, our team will be happy to review it.

rafariossaa:
Hi @asgoel ,
  We really appreciate any feedback provided.

Thanks forehand.

tbnguyen1407:
Hello, there is an error with if/else in this line causing the certs not mounted on slave-statefulset:

https://github.com/bitnami/charts/blob/365514ff6a116dacd30b5d5c3c144f788b799930/bitnami/redis/templates/redis-slave-statefulset.yaml#L481

joancafom:
Hi @tbnguyen1407 

Thanks so much for reporting! PR #2852 is going to be merged shortly and should be able to fix this problem 😄.

thomas-riccardi:
Hello, using redis chart 10.7.5 with `tls.authClients=false` the probes (readiness/health) still send client certificates, using server key/cert.
Somehow in my case it doesn't work:
[...]

PR #2852, opened 2020-06-16
[bitnami/redis] Fix conditional to mount TLS certificates on slave statefulset with persistence disabled

ahovgaard:
**Description of the change**

Currently, when persistence is disabled, the redis slave statefulset does not mount the `redis-certificates` volume from the secrets when TLS is enabled, due to an error with the conditional in the YAML.

**Benefits**

This will allow the use of TLS even when persistence is disabled.

**Possible drawbacks**

None.

**Applicable issues**

  - Issue identified by @tbnguyen1407 in #2622.

**Additional information**

None.

**Checklist** 

### DIFF
--- a/bitnami/redis/templates/redis-slave-statefulset.yaml
+++ b/bitnami/redis/templates/redis-slave-statefulset.yaml
@@ -478,16 +478,16 @@ spec:
           emptyDir: {}
         - name: redis-tmp-conf
           emptyDir: {}
-    {{- if not .Values.slave.persistence.enabled }}
-        - name: redis-data
-          emptyDir: {}
-    {{- else }}
     {{- if .Values.tls.enabled }}
         - name: redis-certificates
           secret:
             secretName: {{ required "A secret containing the certificates for the TLS traffic is required when TLS in enabled" .Values.tls.certificatesSecret }}
             defaultMode: 256
     {{- end }}
+    {{- if not .Values.slave.persistence.enabled }}
+        - name: redis-data
+          emptyDir: {}
+    {{- else }}
   volumeClaimTemplates:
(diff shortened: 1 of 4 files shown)


bitnami/redis/templates/redis-slave-statefulset.yaml
```diff
@@ -481,7 +481,7 @@ spec:
     {{- if not .Values.slave.persistence.enabled }}
         - name: redis-data
           emptyDir: {}
-    {{- else }}
+    {{- end }}
```
joancafom, line 484: I think the `else` here was doing fine. Instead, by moving the volume above the `if-else` statement, the problem would be solved 😁. 
```diff
+   {{- if .Values.tls.enabled }}
+       - name: redis-certificates
+         secret:
+           secretName: {{ required "A secret containing the certificates for the TLS traffic is required when TLS in enabled" .Values.tls.certificatesSecret }}
+           defaultMode: 256
+   {{- end }}
    {{- if not .Values.slave.persistence.enabled }}
        - name: redis-data
          emptyDir: {}
    {{- else }}
-    {{- if .Values.tls.enabled }}
-        - name: redis-certificates
-          secret:
-            secretName: {{ required "A secret containing the certificates for the TLS traffic is required when TLS in enabled" .Values.tls.certificatesSecret }}
-            defaultMode: 256
-    {{- end }}
  volumeClaimTemplates:
```

Could you try to do that?

ahovgaard: Thanks for the suggestion. :) I'm not used to YAML conditional and I got a bit confused by the indentation. Let me try that.

PR #2856, opened 2020-06-16
[bitnami/owncloud] Add extraEnv, extraVolumes/extraVolumeMounts, and containerPort to values

almahmoud:
**Description of the change**

This endeavor started when we tried to deploy ownCloud with this chart, with the ingress running at a specific path (i.e. `example.org/myproject/owncloud`). We couldn't get it working with the bitnami container, but using `owncloud/server` image it was possible. However, we needed to add the extra environment variables for that configuration. Rather than doing it just for our purposes, figured it'd be a good general change that would benefit the community. The `extraEnv` variable follows the standard used by most charts I've encountered.
As an artefact of changing the container image, the newly used one exposes owncloud at port 8080 rather than 80, so also added the `containerPort` variable so that users have the ability to swap out images but still expose the new container. Given that it's only in the initial container port definition that matters and everything else (i.e. service, ingress) go unchanged, this seemed like a good variable to have and is very straightforward.
Finally, adding extra Volumes and the corresponding mounts for them is a useful feature that many charts have. In this case, we're specifically running owncloud alongside other apps, and have a shared filesystem that we want to include in the admin owncloud data. These variables are not meant to deploy PV's/PVC's, but rather attach existing resources to the container, as is the case for other charts including this option.

**Benefits**

Admins deploying the chart can inject extra env variables and extra storage to the container, and readily change the exposed container port if they choose to use other images

**Additional Information**
All of these changes are packaged at https://github.com/almahmoud/helm-charts/tree/master/charts (i.e. `
helm repo add alex https://github.com/almahmoud/helm-charts/raw/master`)
All of these changes were tested with versions:

```
$ kubectl version

Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.2", GitCommit:"c97fe5036ef3df2967d086711e6c0c405941e14b", GitTreeState:"clean", BuildDate:"2019-10-15T19:18:23Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.5", GitCommit:"20c265fef0741dd71a66480e35bd69f18351daea", GitTreeState:"clean", BuildDate:"2019-10-15T19:07:57Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"linux/amd64"}


$ helm version

version.BuildInfo{Version:"v3.0.2", GitCommit:"19e47ee3283ae98139d98460de796c1be1e3975f", GitTreeState:"clean", GoVersion:"go1.13.5"}
```

**Checklist**


bitnami/owncloud/values.yaml
```diff
@@ -217,6 +221,23 @@ persistence:
...
+    #     mountPath: /mnt/data/files/admin/files/shared-data
+
+extraEnv: {}
+  # - name: OWNCLOUD_ADMIN_PASSWORD
+  #   valueFrom:
+  #     secretKeyRef:
+  #       name: my-admin-password
+  #       key: password
+  # - name: OWNCLOUD_SUB_URL
+  #   value: "/mysubpath/owncloud"
```
juan131, line 241: If you take a look to other charts (e.g. RabbitMQ), you'll find out that we provide different ways to add extra environment variables:

- Listing the extra environment variables to add
- Retrieving them from an existing ConfigMap
- Retrieving them from an existing Secret (when they contain credentials)

See:

- https://github.com/bitnami/charts/blob/master/bitnami/rabbitmq/values.yaml#L153
- https://github.com/bitnami/charts/blob/master/bitnami/rabbitmq/templates/statefulset.yaml#L159
- https://github.com/bitnami/charts/blob/master/bitnami/rabbitmq/README.md#adding-extra-environment-variables

Could we do exactly the same here?

bitnami/owncloud/values.yaml
```diff
@@ -217,6 +221,23 @@ persistence:
...
     accessMode: ReadWriteOnce
     size: 8Gi
+    # extraVolumes:
+    #   - name: shared-data
+    #     persistentVolumeClaim:
+    #       claimName: shared-data-pvc
+
+    # extraVolumeMounts:
+    #   - name: shared-data
+    #     mountPath: /mnt/data/files/admin/files/shared-data
```
juan131, line 231: I wouldn't include these parameters under the "persistence" section since volumes can be used for other stuff (e.g. mounting a ConfigMap as a volume). You can use RabbitMQ as a reference to see what we do in other charts, see:

- https://github.com/bitnami/charts/blob/master/bitnami/rabbitmq/values.yaml#L262

bitnami/owncloud/values.yaml
almahmoud: Thank you for the quick review, I'm happy to see you're receptive to the PR :)

The `extraEnvVars` as defined there are wrong, unless i'm missing something

For the templating in [statefulset.yaml](https://github.com/bitnami/charts/blob/master/bitnami/rabbitmq/templates/statefulset.yaml#L159), the value would need to be defined as
```
extraEnvVars:
  FOO: BAR
```
Doing it as an array of names/values as suggested in the README and values with the current templating results in:
```
            - name: 0
              value: "map[FOO value:BAR]"
```
I definitely agree with adding the possibility to retrieve from configmaps & secrets, and I don't mind the name change (although I've generally seen `extraEnv` not `extraEnvVars` in other charts personally).
I do however think that the adding should be just templated and added to the envs directly, rather than only accepting key-value pairs like is the templating code now. i.e. I agree with the example in the values and the README, but the templating in `statefulset.yaml` should be changed to just this imo:
```
        {{- if .Values.extraEnvVars }}
        {{- tpl (toYaml .Values.extraEnvVars) . | nindent 12 }}
        {{- end }}
```
I made a quick PR in Github that should fix it: https://github.com/bitnami/charts/pull/2867
I believe this is better because it allows users to define the envs any way they want (eg: they could do 
```
   - name: OWNCLOUD_ADMIN_PASSWORD
     valueFrom:
       secretKeyRef:
         name: my-admin-password
         key: password
```
which imo is a very nice functionality, and I don't see any point in removing it in favor of only key-value pairs.

I can then work on adding the other env variables in the next few days, hopefully by the end of next week, but I have a few other things that I need to deal with first for our projects before getting back to doing this. In the meantime, i'd be happy to continue discussing so we establish exactly what it should look like at the end, so that I can make the changes relatively quickly when i get back to it rather than too much back and forth at that time, if that's okay with you.

bitnami/owncloud/values.yaml
almahmoud: I personally think it'd be nice to have a high-level `owncloud` value that indicates the changes are made to the main owncloud container. This would both be in line with how [nextcloud chart](https://github.com/helm/charts/tree/master/stable/nextcloud#installing-the-chart) evolved, keeping the two more in sync, and personally think it makes it easier to understand.
This would mean having `owncloud.extraEnv` and `owncloud.extraVolumes` which seems better to me, but also it's ultimately just a naming issue so I am not at all set in stone in that opinion if you have other reasons like internal consistency for preferring to not nest things under `owncloud`

almahmoud: P.S. I definitely think highest scope is better than inside persistence though if `owncloud` value is not to be added. The choice to put it in persistence was mostly because some other chart developers are more inclined to accept nested than high-level values and wanted this to not look too intrusive :)

bitnami/owncloud/values.yaml
juan131: We have supported both ways in the past, and I agree that using a function such as the one below is more flexible. Thanks for pointing it out!!
```
        {{- if .Values.extraEnvVars }}
        {{- tpl (toYaml .Values.extraEnvVars) . | nindent 12 }}
        {{- end }}
```

juan131: PS: Use `extraEnvVars`, we're trying to make that our standard

bitnami/owncloud/values.yaml
juan131: We usually have a high-level value (such as `owncloud`, `mariadb`, etc.) when the chart deploys several components, so we can differentiate between what parameters affect componentA or componentB (e.g see [Thanos](https://github.com/bitnami/charts/tree/master/bitnami/thanos)). It's not vey common to do so on charts such as OwnCloud that only one component (plus MariaDB which is managed by a different chart).
Regarding including these parameters in a higher scope than `persistence`, as I said before volumes can be used for different purposes and that's why I think it should be treated as different parameters.

juan131:
@almahmoud just a friendly reminder to update this PR when you have a chance

juan131:
Ping @almahmoud

marcosbc:
Since now may not be the best time to check this due to summer vacations, I'll just remove the stale tag once more.

PR #2863, opened 2020-06-17
Add note about OpenEBS for persistent storage option

briankmatheson:
Hi all,

Some of us at MayaData were discussing the docs on this chart (primarily how we often find the most useful docs and info in the Bitnami charts as opposed to other sources :^).  We were wondering if you would update it and mention OpenEBS as an option for persistence.

**Description of the change**

Docs update to point out that there are other non-provider based ways to create persistent volumes.

**Benefits**

OnPrem users can't take advantage of orchestrated storage in AWS, GCP, etc.  OpenEBS orchestrates local storage on prem or in instances with local block devs on board.

**Possible drawbacks**

None.  Minor change.  Happy to provide more info and details about how to use openebs with Cassandra.  Would love to collaborate on some chart mods to automate setup for storage in a few common scenarios.  

**Additional information**

www.openebs.io for more info :^)
Thanks much,
Brian

### DIFF
--- a/bitnami/cassandra/README.md
+++ b/bitnami/cassandra/README.md
@@ -246,7 +246,7 @@ initDBConfigMap=init-db
 
 The [Bitnami cassandra](https://github.com/bitnami/bitnami-docker-cassandra) image stores the cassandra data at the `/bitnami/cassandra` path of the container.
 
-Persistent Volume Claims are used to keep the data across deployments. This is known to work in GCE, AWS, and minikube.  Persistent volumes can also be created dynamically with other CNCF software like OpenEBS.  See the [Parameters](#parameters) section to configure the PVC or to disable persistence.
+Persistent Volume Claims are used to keep the data across deployments. This is known to work in GCE, AWS, and minikube. Persistent volumes can also be created dynamically with other CNCF software like OpenEBS. See the [Parameters](#parameters) section to configure the PVC or to disable persistence.
 
 ### Adjust permissions of persistent volume mountpoint
 
--- a/bitnami/cassandra/README.md
+++ b/bitnami/cassandra/README.md
@@ -246,8 +246,7 @@ initDBConfigMap=init-db
 
 The [Bitnami cassandra](https://github.com/bitnami/bitnami-docker-cassandra) image stores the cassandra data at the `/bitnami/cassandra` path of the container.
 
-Persistent Volume Claims are used to keep the data across deployments. This is known to work in GCE, AWS, and minikube.
-See the [Parameters](#parameters) section to configure the PVC or to disable persistence.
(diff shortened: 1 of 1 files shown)


bitnami/cassandra/README.md
```diff
@@ -246,8 +246,7 @@ initDBConfigMap=init-db
 
 The [Bitnami cassandra](https://github.com/bitnami/bitnami-docker-cassandra) image stores the cassandra data at the `/bitnami/cassandra` path of the container.
 
-Persistent Volume Claims are used to keep the data across deployments. This is known to work in GCE, AWS, and minikube.
-See the [Parameters](#parameters) section to configure the PVC or to disable persistence.
+Persistent Volume Claims are used to keep the data across deployments. This is known to work in GCE, AWS, and minikube.  Persistent volumes can also be created dynamically with other CNCF software like OpenEBS.  See the [Parameters](#parameters) section to configure the PVC or to disable persistence.
```
andresbono, line 249:
```suggestion
Persistent Volume Claims are used to keep the data across deployments. This is known to work in GCE, AWS, and minikube. Persistent volumes can also be created dynamically with other CNCF software like OpenEBS. See the [Parameters](#parameters) section to configure the PVC or to disable persistence.
```

andresbono:
Hi @briankmatheson, thanks for showing interests on this chart! 😄 A couple of additional notes:

- Do you think it would be interesting to add the mention in more charts?

- Don't worry about the automated check failing, for this kind of change, it is not needed to bump the chart version.

briankmatheson:
Re: other charts: Absolutely!  We'd love to get the word out as much as possible, and for as many stateful apps as possible :^)  

Cassandra is front-of-mind for a lot of folks right now, which is why I started here.  Also of particular interest to us at MayaData are the charts for elasticsearch, magento, solr, kafka, postgres and mariadb, just off the top of my head.  

Also, I'd love to work with you on what the best way is to put in a parameterized option in some or all of the above charts to set up a new openebs storage class.  We could maybe control this with a boolean option, and when that option is turned on helm would install openebs if necessary, configure a storage pool and appropriate storage classes for the app, and then install the app using that set of storage classes.  

For cassandra in particular, it would be very easy to include support for openebs' dynamic LocalPV Device.  For other apps like Postgres, it makes more sense to create a pool and replicated storage class, so those apps would be a little more involved to support.  At any rate, this gives the charts / project the advantage of supporting bare metal deployment scenarios, as well as a simple mechanism for linking in local storage in scale-out situations in the various clouds without having to manually intervene.

briankmatheson:
One additional question came up: would you like similar updates to the other charts mentioned collected in this PR, or in separate PRs?

andresbono:
@briankmatheson, we are open to receive PRs and we will be more than happy to review them, at least if would be nice to receive a first draft for Cassandra as starting point. The implementation that you are describing sounds good!

> One additional question came up: would you like similar updates to the other charts mentioned collected in this PR, or in separate PRs?

It would be better to continue using this PR for the README updates. All the applications that you mentioned should have the same note about persistence.

PR #2867, opened 2020-06-17
[bitnami/rabbitmq] Fixing extraEnvVars

almahmoud:
Quick PR related to discussion in https://github.com/bitnami/charts/pull/2856


bitnami/rabbitmq/templates/statefulset.yaml
```diff
@@ -156,9 +156,8 @@ spec:
                   key: rabbitmq-password
             - name: RABBITMQ_PLUGINS
               value: {{ include "rabbitmq.plugins" . | quote }}
-            {{- range $key, $value := .Values.extraEnvVars }}
-            - name: {{ $key }}
-              value: {{ $value | quote }}
+            {{- if .Values.extraEnvVars }}
+            {{- tpl (toYaml .Values.extraEnvVars) . | nindent 12 }}
```
juan131, line 160: Thanks so much! Please use the `common.tplvalues.render` macro instead of `tpl (toYaml ...)`:
```suggestion
            {{- if .Values.extraEnvVars }}
            {{- include "common.tplvalues.render" (dict "value" .Values.extraEnvVars "context" $) | nindent 12 }}
```

PR #2881, opened 2020-06-18
[bitnami/elasticsearch] allow existing claims in master and data

zenbones:
For bitnami/elasticsearch...

This change allows master and data to use an existing volume claim for the data volume, as opposed to a claim template. Many other bitnami charts allow existing volume claims in this way, which are a convenience for the user.

This change, especially in light of stateful sets, means that the user does have to be responsible for the pvc, and all that entails. But, by using the existingClaim value, that is exactly what they're asking for. They can always get the previous behavior merely by not including the value, so I don't think there's any downside.


bitnami/elasticsearch/templates/data-statefulset.yaml
```diff
@@ -216,18 +216,22 @@ spec:
...
+        {{- end }}        
       spec:
         accessModes: {{- toYaml .Values.data.persistence.accessModes | nindent 10 }}
         {{ $storage := dict "global" .Values.global "local" .Values.data }}
         {{ include "elasticsearch.storageClass" $storage }}
         resources:
           requests:
             storage: {{ .Values.data.persistence.size | quote }}
-{{- end }}
+{{- end }}
```
marcosbc, line 237: Could you fix the "No newline at end of file" error?

bitnami/elasticsearch/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: elasticsearch
-version: 12.3.6
+version: 12.3.7
```
marcosbc, line 3: This would be in any case a minor version change (i.e. 12.4.0)

marcosbc:
Thanks for the contribution! I added a few comments on a few improvements, let me know if you have any question.

bitnami/elasticsearch/templates/data-statefulset.yaml
zenbones: Passed the linter, so I'm not sure where this error is. I'll add an extraneous new line to both files I changed. Maybe that will solve it?

zenbones: fixed

bitnami/elasticsearch/Chart.yaml
zenbones: changed to 12.4.0

zenbones:
Did you want me to take a pass at the values.yaml, values-production.yaml and README?

zenbones:
Closed, the conversation... not sure that was right, but in case yo can't see them, I made the suggested changes.

bitnami/elasticsearch/templates/data-statefulset.yaml
```diff
@@ -216,13 +216,17 @@ spec:
...
+          persistentVolumeClaim:
+            claimName: {{ .Values.data.persistence.existingClaim }}
 {{- else }}
   volumeClaimTemplates:
     - metadata:
         name: "data"
         {{- if .Values.data.persistence.annotations }}
         annotations: {{- toYaml .Values.data.persistence.annotations | nindent 10 }}
-        {{- end }}
+        {{- end }}
```
marcosbc, line 229: It looks like some trailing spaces were added, could you remove them?

zenbones: removed trailing spaces, added documentation to values.yaml, values-production.yaml and README.md

zenbones:
Let me know if you want any other changes.

Issue #2891, opened 2020-06-20
[bitnami/mongodb] MongoDB primary node and secondary node are always deployed in the same node in AWS EKS


chowwaihei1228:
**Which chart**:
mongodb (Version 7.14.8)

**Describe the bug**
I find the primary node and secondary node are deployed in the same node by using values-production.yaml. 

Then I try to use Pod Anti Affinity with **requiredDuringSchedulingIgnoredDuringExecution** to prevent such issue but failed. I observe the persistent volumes are created and mounted by the same node before the pods are created, and therefore both primary and secondary pods must be scheduled in that node. 

I see that there is no such issue in redis master and slave nodes, where all pods are install in distinct nodes in different AZs. 

**To Reproduce**
1. Edit values-production.yaml
```
affinity: 
  podAntiAffinity:
    requiredDuringSchedulingIgnoredDuringExecution:
    - labelSelector:
        matchExpressions:
        - key: app
          operator: In
          values:
          - mongodb
      topologyKey: failure-domain.beta.kubernetes.io/zone
```
2. helm install mongodb bitnami/mongodb -f values-production.yaml
3. kubectl get pod 
```
NAME                  READY   STATUS    RESTARTS   AGE
mongodb-arbiter-0     1/1     Running   0          19m
mongodb-primary-0     2/2     Running   0          19m
mongodb-secondary-0   0/2     Pending   0          19m  
```
4. kubectl describe pod mongodb-secondary-0
```
Events:
  Type     Reason             Age                From                Message
  ----     ------             ----               ----                -------
  Normal   NotTriggerScaleUp  27s                cluster-autoscaler  pod didn't trigger scale-up (it wouldn't fit if a new node is added): 1 max node group size reached
  Warning  FailedScheduling   25s (x4 over 31s)  default-scheduler   pod has unbound immediate PersistentVolumeClaims (repeated 3 times)
  Warning  FailedScheduling   21s                default-scheduler   0/3 nodes are available: 1 node(s) didn't match pod affinity/anti-affinity, 1 node(s) didn't satisfy existing pods anti-affinity rules, 2 node(s) had volume node affinity conflict.
```

**Expected behavior**
Either primary node or secondary node cannot be created due to volume node affinity conflict.

**Version of Helm and Kubernetes**:

- Output of `helm version`:

```
[...]

Issue #2848, opened 2020-06-16
[bitnami/mongodb] MongoDB primary pod restart fails to connect to mongodb-secondary


ganchandrasekaran:
**Which chart**:
The name (and version) of the affected chart
chart: bitnami/mongodb 
version: 7.14.6

**Describe the bug**
Once MongoDB cluster is deployed with replicaSet enabled and persistence disabled. 
persistence. enabled: false
replicaSet. enabled: true

**To Reproduce**
1. Deploy MongoDB chart
```
# kubectl get po
mongodb-arbiter-0                                                 1/1     Running   0          40s
mongodb-primary-0                                                 1/1     Running   0          40s
mongodb-secondary-0                                               1/1     Running   0          30s
```
2. Restart Primary node:
```
# kubectl rollout restart statefulset mongodb-primary
```
3. Check logs of primary to see error, note there is no error show on the status of the pod. 
```
# kubectl logs mongodb-primary-0
2020-06-16T13:24:31.670+0000 I  CONNPOOL [Replication] Connecting to mongodb-arbiter-0.mongodb-headless.default.svc.cluster.local:27017
2020-06-16T13:24:31.732+0000 I  REPL_HB  [replexec-1] Heartbeat to mongodb-arbiter-0.mongodb-headless.default.svc.cluster.local:27017 failed after 2 retries, response status: InvalidReplicaSetConfig: replica set IDs do not match, ours: 5ee8c801d5cdd538b8153c1c; remote node's: 5ee8c7b17f3ff3a4aec1d0d7
```

**Expected behavior**
Primary should bind with the existing the replication cluster

**Additional context**
Based on similar bug reports before such as: https://github.com/bitnami/bitnami-docker-mongodb/issues/110
1, This was fixed for Pod using PV.  Is it possible to make changes to make it work with emptyDir{} as well. Perhaps some way to sync the replica set IDs with existing secondary. 
2, What would be the exact role of Secondary and arbiter?can Secondary take the role of Primary as I understand primary is always rescheduled by K8S. Does this mean, Write operation has some downtime when primary pod is down?

carrodher:
Hi, thanks for the detailed explanation.

[...]

Issue #2460, opened 2020-04-29
[bitnami/mongodb] MongoDB is not rejoining cluster


estahn:
**Which chart**:
`bitnami/mongodb` and `7.13.0`

**Describe the bug**
Given a cluster with 1 primary and 2 secondary pods. We're terminating the primary which results in the secondary to take over. After the primary comes back it doesn't rejoin and the secondary stays "primary". To make the primary rejoin we execute the following on the secondary: 

```
cfg = rs.conf()
rs.reconfig(cfg)
```

**To Reproduce**
1. Create a cluster with 1 primary and 2 secondary
2. Delete the primary
3. Check logs of secondary

**Expected behavior**
Primary rejoins the cluster.

**Version of Helm and Kubernetes**:

- Output of `helm version`: We use ArgoCD, which uses `helm template` internally.
- Output of `kubectl version`:

```
Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.2", GitCommit:"52c56ce7a8272c798dbc29846288d7cd9fbae032", GitTreeState:"clean", BuildDate:"2020-04-16T23:35:15Z", GoVersion:"go1.14.2", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"12", GitVersion:"v1.12.10", GitCommit:"e3c134023df5dea457638b614ee17ef234dc34a6", GitTreeState:"clean", BuildDate:"2019-07-08T03:40:54Z", GoVersion:"go1.10.8", Compiler:"gc", Platform:"linux/amd64"}
```

**Additional context**
n/a

estahn:
@horihel makes a good observation in #2461. A cluster with the following settings is working fine:

```yaml
clusterDomain: cluster.local
replicaSet:
  clusterDomain: true
```

javsalgar:
Hi,

This is weird. So the thing is that the `search` settings in the cluster don't add the `cluster.local` section and so the resolution fails, right?

moh-abed:
The clusterDomain change didnt resolve the issue actually, primary did not rejoin unless manually calling rs.reconfig()

javsalgar:
Hi mod-abed, did setting the values here fix the issue?

https://github.com/bitnami/charts/issues/2460#issuecomment-621169403

SwannMR:
Hi,

I experienced the same issue while running a cluster with 1 primary and 2 secondary mongos.

[...]

PR #2918, opened 2020-06-25
[bitnami/mongodb] Refactor MongoDB exposure

juan131:

**Description of the change**

This PR refactor the way to expose MongoDB nodes outside the K8s cluster, so there's one service per node. It also groups every secondary and primary node under the same statefulset.

The PR also ensure the chart follow the latest Helm Best Practices.

**Benefits**

- Users can access the primary node when the role is taken by one of the secondary nodes (e.g. the primary node is down).
- Users can expose both secondary and primary nodes outside the cluster.

**Possible drawbacks**

- Breaking backwards compatibility.
- Ingress support has been removed until we can automate the process to configure the controller to expose TCP/UDP ports.

**Applicable issues**

- fixes https://github.com/bitnami/charts/issues/2848
- fixes https://github.com/bitnami/charts/issues/2891
- fixes https://github.com/bitnami/charts/issues/2460

**Checklist**


bitnami/mongodb/templates/initialization-configmap.yaml
```diff
@@ -1,14 +1,11 @@
...
-    release: {{ .Release.Name | quote }}
-    heritage: {{ .Release.Service | quote }}
+  name: {{ include "mongodb.fullname" . }}-init-scripts
+  namespace: {{ include "mongodb.namespace" . }}
+  labels: {{- include "common.labels.standard" . | nindent 4 }}
+    app.kubernetes.io/component: mongodb
 data:
-{{ tpl (.Files.Glob "files/docker-entrypoint-initdb.d/*[sh|js|json]").AsConfig . | indent 2 }}
-{{ end }}
+{{- include "common.tplvalues.render" (dict "value" .Values.initdbScripts "context" .) | nindent 2 }}
```
marcosbc, line 10: Wouldn't it be "indent 2" instead of "nindent 2" here? Since there is already a newline.

bitnami/mongodb/values.yaml
```diff
@@ -45,255 +45,319 @@ image:
...
-    ## ref: https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/
-    ##
-    # pullSecrets:
-    #   - myRegistryKeySecretName
-  resources: {}
+# schedulerName:
+
+## MongoDB architecture. Allowed values: standalone or replicaset
+##
+architecture: standalone
```
marcosbc, line 59: I have a small concern, and it is that we're encapsulating the concept of replicaset inside an "architecture" value. I'm thinking of a user who would just run "helm install mongodb bitnami/mongodb".

Why is it we can't keep the previous "replicaset.enable" variable? For me it is a lot more intuitive. If we need to decide between an architecture or another, we could always create a helper for deciding the architecture depending on "replicaset.enable".

bitnami/mongodb/templates/initialization-configmap.yaml
juan131: Not really since we're using `{{- ...`
These are equivalent:
```
{{- ... | nindent ... }}
{{ ... | indent ... }}
```
I prefer the 1st one to be consistent of the manifests where I use mainly `nindent`

bitnami/mongodb/values.yaml
juan131: There are two reasons why I prefer an "architecture" parameter, similar to the one we already use in MinIO (name 'mode') or InfluxDB:

- If, in the future, you want to add a 3rd supported architecture setup, you can still use the same parameter.
- It helps the user to understand that enabling replicaset has several implications. It's not like adding a feature but switching the whole architecture.

bitnami/mongodb/templates/standalone/svc.yaml
```diff
@@ -27,18 +24,14 @@ spec:
   loadBalancerSourceRanges: {{- toYaml .Values.service.loadBalancerSourceRanges | nindent 4 }}
   {{- end }}
   ports:
-    - name: mongodb
+    - name: tcp-mongodb
```
chris922, line 27: Hey @juan131 , we just installed mongo using version 8.0.7 of the mongodb chart and noticed that we aren't able to connect to the deployed mongo using this connection string: `mongodb+srv://mongodb-headless.persistence.svc.cluster.local`

I found out that the reason is, that the port within the `mongodb-headless` service is named `tcp-mongodb` instead of `mongodb` and then I found this change here.

Could you please have a look again at your changes and ensure that connecting to mongo via `srv` still works?
This also affects the other service definitions that you created during your refactorings.

More information about this: 
* https://www.mongodb.com/blog/post/mongodb-3-6-here-to-SRV-you-with-easier-replica-set-connections -> mongo prepends `_mongodb._tcp` to the hostname 
* https://kubernetes.io/docs/concepts/services-networking/service/#dns: k8s is using the port name + protocol in the SRV DNS entry as a prefix

javsalgar: Hi,

Thank you so much for the feedback. Could create an issue instead?  This will likely get lost if we use PR comments. I can copy-paste the message in a new issue if necessary.

chris922: Good idea, I just created #3067 out of this discussion

PR #2935, opened 2020-06-26
[bitnami/postgresql-ha]Conditional rendering of .Values.pgpool.priorityClassName

ttinkr:
**Description of the change**

Wrapped `priorityClassName: {{ .Values.pgpool.priorityClassName }}` in a conditional statement, because if this value is not set in values.yaml it will result in `priorityClassName:` inside the deployment.yaml which is not correct yaml and leads to the error:
`Error: <nil> is expected to be either a string or a []interface{}`

**Benefits**

If the value is not set, the fix will result in valid yaml, which also can be further customized using kustomize. Without this fix the yaml can only be used by helm itself, but not by e.g. kubectl apply.

**Possible drawbacks**

None I can think of.

**Applicable issues**

None.

**Additional information**

None.

**Checklist** 


joancafom:
Reopening this PR and implementing the version bump. It should be a patch, since we are applying a patch for a bug here.

joancafom:
LGTM! Implemented requested changes myself in order to close this bug fix.

PR #2954, opened 2020-06-29
[bitnami/harbor] Encode base64 CSRF_KEY

### LINKED ISSUE #2939: Harbor Chart - CSRF Key invalid
Harbor, v6.x

After installing the chart, I cannot log into harbor due a CSRF_Key invalid error. Looking in the core-envvars, I see that the string for the CSRF_KEY has been distorted. Replacing it with the string I had defined in the install resolves the issue. In the template for this secret, I see that it is not base64 encoding the value of the CSRF_KEY value, which I believe to be the issue.

1. Install the chart
2. Try to log into harbor

**Expected behavior**
The CSRF_KEY to be stored correctly.

joancafom:

**Description of the change**
Correctly encode CSRF tokens

**Benefits**
The value is rendered properly and the user can log in again

**Possible drawbacks**
N/A

**Applicable issues**

  - fixes #2939

**Additional information**

**Checklist** 


damoon:
thanks for the fix.

just got the error
`OCI runtime create failed: container_linux.go:349: starting container process caused "process_linux.go:449: container init caused \"setenv: invalid argument\"": unknown`
that is caused by this (see https://github.com/kubernetes/kubernetes/issues/89906).

joancafom:
Hi @damoon ,

Thanks for your comment! As far as I was able to test, I could not face your issue. I'll be merging this to solve #2939 . Should you further face the problem, you can create a new issue detailing the case and our team will be happy to provide with an answer.

Thanks!

damoon:
The issue i saw happens when by random chance a NUL byte is part of the secret value. Then the env var can not be set and the container creatation is stuck in a loop. Addding the missing base64 encoding will wrap the NUL and avoid the error. Thanks for the fix again.

PR #2286, opened 2020-04-10
[bitnami/elasticsearch] POC: ES to be compatible with Istio

juan131:
Signed-off-by: juan131 <juan@bitnami.com>

**Description of the change**

This is a POC to make ES compatible with Istio. In the associated issue you can find extended information about the concerns for this chart to be compatible with Istio.

**Benefits**

You can use ES with Istio

**Possible drawbacks**

Breaks backwards compatibility

**Applicable issues**

  - fixes https://github.com/bitnami/charts/issues/2199

**Additional information**

Please do not merge this PR. We should discuss about the changes before.

**Checklist**

- [x] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [x] Variables are documented in the README.md
- [x] Title of the PR starts with chart name (e.g. `[bitnami/chart]`)
- [x] If the chart contains a `values-production.yaml` apart from `values.yaml`, ensure that you implement the changes in both files

PR #2964, opened 2020-06-29
[bitnami/minio] make minio service istio-compatible

funkypenguin:
**Description of the change**

Hey guys! This change simply prefixes the Minio service port name with `http-`, in order to make Minio work with Istio mTLS (which requires [strict port naming](https://istio.io/latest/docs/ops/configuration/traffic-management/protocol-selection/))

**Benefits**

Istio users will be able to use Minio without having to deploy an _additional_ service (which is my current workaround) with a port matching the Istio protocol/port-naming requirements.

**Possible drawbacks**

None - port name is immaterial to non-Istio users

**Applicable issues**

None

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/minio/Chart.yaml
+++ b/bitnami/minio/Chart.yaml
@@ -1,5 +1,5 @@
 apiVersion: v1
-version: 3.4.12
+version: 3.4.13
 appVersion: 2020.6.22
 description: MinIO is an object storage server, compatible with Amazon S3 cloud storage service, mainly used for storing unstructured data (such as photos, videos, log files, etc.)
 keywords:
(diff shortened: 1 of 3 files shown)


andresbono:
Hi @funkypenguin, thank you very much for contributing with this PR!

Please check this comment: https://github.com/bitnami/charts/pull/2286#issuecomment-617098010. In there, you can see a list of the checks for all the requirements to be compatible with Istio. It seems that if you deploy the `bitnami/minio` chart with default `values.yaml` all the requirements are fulfilled. But it is not the case for `values-production.yaml`. There is a headless service that may conflict with the main service.

In any case, your PR is just changing the name of the ports in the end. So I don't think it hurts to merge it.

Please, let me know your thoughts on this!

PR #2971, opened 2020-06-30
[bitnami/discourse] do not pass quoted string to printf template functions

stv0g:
Passing a quoted string to `printf %d` does not work..


javsalgar:
Hi!

Thanks for the PR! LGTM

PR #2990, opened 2020-07-01
[bitnami/drupal] Move container to bash and non-root

andresbono:
**Description of the change**

We are working on releasing the Drupal Docker image with a non-root approach. As part of that, we need to update the chart to use the new port numbers used by the container image.

**Benefits**

**Possible drawbacks**

Breaking change with previous chart versions, as the default HTTP/HTTPS ports are changed to 8080/8443 (previously 80/443).

**Checklist** 


marcosbc:
> That said, since we're breaking backwards compatibility, we should include some "standardisation changes" in this PR that breaks backwards compatibility (e.g. using K8s recommended labels or using the "common" library)

I disagree. If that needs to be done, then it should be part of a bigger task where we do that on all charts. What we cannot do is "take advantage" of there being a major version bump, to enforce standarization of that Bitnami chart, as we have limited resources for that.

bitnami/drupal/templates/tls-secrets.yaml
```diff
@@ -4,11 +4,14 @@ apiVersion: v1
 kind: Secret
 metadata:
   name: {{ .name }}
-  labels:
-    app: {{ template "drupal.fullname" $ }}
-    chart: {{ template "drupal.chart" $ }}
-    release: {{ $.Release.Name }}
-    heritage: {{ $.Release.Service }}
+  name: {{ template "drupal.fullname" $ }}-drupal
```
marcosbc, line 7: duplicated name?

bitnami/drupal/templates/extra-list.yaml
```diff
@@ -0,0 +1,5 @@
+{{- if .Values.extraDeploy }}
+apiVersion: v1
+kind: List
+items: {{- include "kong.tplValue" (dict "value" .Values.extraDeploy "context" $) | nindent 2 }}
```
marcosbc, line 4: kong -> drupal

bitnami/drupal/templates/deployment.yaml
```diff
@@ -1,145 +1,220 @@
...
+          securityContext:
+            runAsUser: {{ .Values.containerSecurityContext.runAsUser }}
+          {{- end }}
+          env:
+            - name: ALLOW_EMPTY_PASSWORD
+            {{- if .Values.allowEmptyPassword }}
+              value: "yes"
+            {{- else }}
+              value: "no"
+            {{- end }}
```
juan131, line 77:
```suggestion
            - name: ALLOW_EMPTY_PASSWORD
              value: {{ ternary "yes" "no" .Values. allowEmptyPassword | quote }}
```

bitnami/drupal/templates/deployment.yaml
```diff
@@ -1,145 +1,220 @@
...
+            {{- include "common.tplvalues.render" (dict "value" .Values.extraVolumeMounts "context" $) | nindent 12 }}
+            {{- end }}
+        {{- if .Values.metrics.enabled }}
+        - name: metrics
+          image: {{ template "drupal.metrics.image" . }}
+          imagePullPolicy: {{ .Values.metrics.image.pullPolicy | quote }}
+          command: [ '/bin/apache_exporter', '--scrape_uri', 'http://status.localhost:80/server-status/?auto']
+          ports:
+          - name: metrics
+            containerPort: 9117
```
juan131, line 190: Indentation

bitnami/drupal/templates/deployment.yaml
```diff
@@ -1,145 +1,220 @@
...
+          command: {{- include "common.tplvalues.render" (dict "value" .Values.command "context" $) | nindent 12 }}
+          {{- end }}
+          {{- if .Values.args }}
+          args: {{- include "common.tplvalues.render" (dict "value" .Values.args "context" $) | nindent 12 }}
+          {{- end }}
+          {{- if .Values.containerSecurityContext.enabled }}
+          securityContext:
+            runAsUser: {{ .Values.containerSecurityContext.runAsUser }}
+          {{- end }}
+          env:
```
marcosbc, line 71: We should also be adding support for `BITNAMI_DEBUG`:

```

```

Don't forget to document in the README and values.yaml.

bitnami/drupal/README.md
```diff
@@ -48,69 +48,131 @@ The command removes all the Kubernetes components associated with the chart and
...
+|---------------------------|-------------------------------------------------|---------------------------------------------------------|
+| `global.imageRegistry`    | Global Docker image registry                    | `nil`                                                   |
+| `global.imagePullSecrets` | Global Docker registry secret names as an array | `[]` (does not add image pull secrets to deployed pods) |
+| `global.storageClass`     | Global storage class for dynamic provisioning   | `nil`                                                   |
+
+### Common parameters
+
+| Parameter           | Description                                                                  | Default                                                 |
+|---------------------|------------------------------------------------------------------------------|---------------------------------------------------------|
+| `image.registry`    | Drupal image registry                                                        | `docker.io`                                             |
```
marcosbc, line 65: Don't forget to document image.debug here and in values.yaml

andresbono: It's documented, line 70. The same in values.yaml

marcosbc: Oh, nice, thanks!

cpjolly:
I have some questions about how to control these changes. 

- Is it enough just to set containerSecurityContext.enabled and podSecurityContext.enabled to false or must you set runAsUser & fsGroup ?
- Is the README correct that just setting "containerSecurityContext.runAsUser to root" is sufficient to re-enable root user or do you also need to change podSecurityContext.fsGroup ?
- Should these parameters be set to "root" or "0" ? For example see https://github.com/bitnami/charts/blob/master/bitnami/mariadb-galera/README.md where it says
"You can revert this behavior by setting the parameters securityContext.runAsUser, and securityContext.fsGroup to 0"

marcosbc:
Hi @cpjolly, it would be enough to set `containerSecurityContext.runAsUser=0` (in other words, to the `root` user's UID). If you find it confusing where we mention to set it to `root`, feel free to contribute a PR, we'd be glad to review it!

PR #2996, opened 2020-07-01
[bitnami/elascticsearch] Allow existing volume in master and data

zenbones:
Allow an addition value in data and master persistence called 'existingVolume'. If that value is defined than add a volume match label selector for it to the volumClaimTemplate, which will allow use of a predefined persistent volume, as opposed to create a slew of dynamic ones. This is a quality of life improvement option for the chart consumer. I don't see any real drawbacks as the current charts use no match labels at all for the volumeClaimTemplate. In the future, maybe greater flexibility will be required, but at that point the volumeClaimTemplate should be complete defined in the values.yaml, which is a larger change, and probably unnecessary.

- [ x] Variables are documented in the README.md
- [ x] Title of the PR starts with chart name (e.g. `[bitnami/chart]`)
- [ x] If the chart contains a `values-production.yaml` apart from `values.yaml`, ensure that you implement the changes in both files


bitnami/elasticsearch/templates/master-statefulset.yaml
```diff
@@ -237,4 +237,9 @@ spec:
         resources:
           requests:
             storage: {{ .Values.master.persistence.size | quote }}
+          {{- if .Values.master.persistence.existingVolume }}
+          selector:
+            matchLabels:
+              volume: {{ .Values.master.persistence.existingVoume }}
```
marcosbc, line 243: Typo here

```suggestion
              volume: {{ .Values.master.persistence.existingVolume }}
```

zenbones: I don't think extraVolumes would cause the current volumeClaimTemplate to be ignored, and I think mounting an extra volume at exactly the same name and mountPath would be confusing if it did work, and I wouldn't want to depend upon such behavior. But, more importantly, there are 2 volumeClaimTemplate(s), one defined for .data. and one for .master., but extraVolumes is defined only at the root level (values.exrtaVolumes), so even if it could swap into place seamlessly, I'd have to define 'name: data' twice at two different mountPath(s)? I think the small addition in this request is cleaner and more comprehensible. Not that I'm a kubernetes expert at this point, so I'm open if you think that analysis is incorrect.

zenbones: Also, corrected that typo, thank you for catching that.

PR #3005, opened 2020-07-02
[bitnami/nats] adding pod disruption budget support

alexsn:
**Description of the change**
Adding pod disruption budget support to nats chart

**Benefits**
Provide ability to configure pdb for nats pods.

**Checklist**


bitnami/nats/templates/poddisruptionbudget.yaml
```diff
@@ -0,0 +1,17 @@
...
+    app: "{{ template "nats.name" . }}"
+    chart: "{{ template "nats.chart" . }}"
+    release: {{ .Release.Name | quote }}
+    heritage: {{ .Release.Service | quote }}
+spec:
+  selector:
+    matchLabels:
+      app: "{{ template "nats.name" . }}"
+      release: {{ .Release.Name | quote }}
+{{ toYaml .Values.podDisruptionBudget | indent 2 }}
```
joancafom, line 16: This works well for sure, but I think the most common approach followed by the majority of Charts is to split this into `minAvailable` and `maxUnavailable`. You can see an example in the [consul](https://github.com/bitnami/charts/blob/1f6bb260ebe2f43808939a60a307a24a124c1b24/bitnami/consul/templates/pod-dist-budget.yaml#L9) chart.

https://github.com/bitnami/charts/blob/1f6bb260ebe2f43808939a60a307a24a124c1b24/bitnami/consul/templates/pod-dist-budget.yaml#L9-L14

Could you please change this to follow an standardized approach?

bitnami/nats/templates/poddisruptionbudget.yaml
```diff
@@ -0,0 +1,17 @@
+{{- if .Values.podDisruptionBudget }}
```
joancafom, line 1: Here as well, you may want to use the standard approach `.Values.pdb.create` (see my comment below)

alexsn:
Thanks for the review, addressed all comments.

bitnami/nats/values-production.yaml
```diff
@@ -72,6 +72,12 @@ podAnnotations: {}
 ##
 podLabels: {}
 
+## Pod disruption budget
+## ref: https://kubernetes.io/docs/tasks/run-application/configure-pdb/
+##
+pdb:
```
joancafom, line 78: You should also reflect the new changes here 😄

bitnami/nats/templates/poddisruptionbudget.yaml
```diff
@@ -0,0 +1,22 @@
...
+    app: "{{ template "nats.name" . }}"
+    chart: "{{ template "nats.chart" . }}"
+    release: {{ .Release.Name | quote }}
+    heritage: {{ .Release.Service | quote }}
+spec:
+  selector:
+    matchLabels:
+      app: "{{ template "nats.name" . }}"
+      release: {{ .Release.Name | quote }}
+  {{- with .Values.pdb.minAvailable }}
```
joancafom, line 16: While this is completely valid, could you please the following pattern (used in the rest of the charts with `pdb`):

```yml
  {{- if .Values.pdb.minAvailable }}
  minAvailable: {{ .Values.pdb.minAvailable }}
  {{- end }}
  {{- if .Values.pdb.maxUnavailable }}
  maxUnavailable: {{ .Values.pdb.maxUnavailable }}
  {{- end }}
```

These makes all Charts have an standardized approach and are easier to maintain 😁

alexsn: Done even though I prefer to do less typing :)

PR #3024, opened 2020-07-03
[bitnami/kubwatch] Implement SMTP config for kubewatch 0.1.0

mkmik:
Signed-off-by: Marko Mikulicic <mkmik@vmware.com>

kubewatch 0.1.0 added SMTP handlers; this PR adds the relevant configuration values to it


bitnami/kubewatch/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 name: kubewatch
-version: 1.0.15
+version: 1.0.16
```
juan131, line 2: Please bump minor version instead. We're usually doing so when adding new parameters to the chart.

bitnami/kubewatch/README.md
```diff
@@ -77,6 +77,16 @@ The following table lists the configurable parameters of the kubewatch chart and
...
 | `webhook.url`                            | Webhook URL                                                                                                                 | `""`                                                    |
+| `smtp.enabled`                           | Enable SMTP (email) notifications                                                                                           | `false`                                                 |
+| `smtp.to`                                | Destination email address (required)                                                                                        | `""`                                                    |
+| `smtp.from`                              | Source email address (required)                                                                                             | `""`                                                    |
+| `smtp.smarthost`                         | SMTP server address (name:port) (required)                                                                                  | `""`                                                    |
```
juan131, line 83: We could check that mandatory parameters are passed when enabling SMTP. Feel free to implement this but it's a nice to have. You can do it adding these block to the `_helpers.tpl`:

```
{{/*
Compile all warnings into a single message, and call fail.
*/}}
{{- define "kubewatch.validateValues" -}}
{{- $messages := list -}}
{{- $messages := append $messages (include "kubewatch.validateValues.smtp" .) -}}
{{- $messages := without $messages "" -}}
{{- $message := join "\n" $messages -}}

{{- if $message -}}
{{- end -}}
{{- end -}}

{{/* Validate values of Kubewatch - SMTP settings */}}
{{- define "kubewatch.validateValues.smtp" -}}
{{- if and .Values.smtp.enabled (or (empty .Values.smtp.to) (empty .Values.smtp.from) (empty .Values.smtp.smarthost) }}
kubewatch: smtp
    The following parameters are mandatory when enabling STMP:
      - smtp.to
      - smtp.from
      - smtp.smarthost
{{- end -}}
{{- end -}}
```

Then adding the line below at the bottom of the `NOTES.txt`:

```
{{- include "kubewatch.validateValues" . }}
```

mkmik: uh; a lot of work, and the controller already does the validation ; yes it's easier to read errors that happen during helm install as opposed to read a crashlooping controller log; but I'm not willing to block this PR on.

bitnami/kubewatch/Chart.yaml
mkmik: PTAL

bitnami/kubewatch/README.md
juan131: Fair enough

PR #3036, opened 2020-07-06
[bitnami/discourse] #3035 - Add existingSecrets sections to chart configuratoin

### LINKED ISSUE #3035: [bitnami/discourse] Add existingSecret sections to chart config
**Which chart**:
discourse

The chart does not allow to use existing secrets and managing db dependencies (i.e. create new DB vs use existing DBs) is quite complex. Many values for DBs are duplicated in the values.yaml file.

I'd like to remove the externalDatabase / externalRedis sections. I'll absorb the params in the postgresql and redis sections.
I'll also introduce new values that will allow to use existing secrets, both for discourse, postgresql and redis passwords.

LucaPrete:
Allow users to use existing secrets for Discourse, Postgresql and Redis passwords.
It also simplifies Discourse database dependencies management (i.e. how values are set for new vs existing databases).

**Benefits**

* Simplified DB management
* Allows users to use existing secrets

**Applicable issues**

  - fixes #3035

**Additional information**

**Checklist**


joancafom:
Hey @LucaPrete 

Thanks for your contribution!! I see the PR is still in draft mode, could you please request my review when changes are ready?

LucaPrete:
@joancafom for sure! It's still WIP but I should have something ready by today ;)

LucaPrete:
@joancafom I should be done with it, although I can't add you as a reviewer.

Please, be extra careful because there are lots of moving pieces and it's very likely for some mistakes to show up... :)

While working on it and its deployment I've also noticed other weird stuff (i.e. the ingress configuration is quite complex and different from any other I've seen. There are also some related mistakes in the readme - i.e. see certmanager annotations). I'll work on it in a new PR, once this gets merged.

Please, let me know if you need any clarification!

LucaPrete:
@joancafom - unrelated - but it was worth to ask. Not sure if you can also help [here](https://github.com/bitnami/bitnami-docker-discourse/issues/152)

bitnami/discourse/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: discourse
 description: A Helm chart for deploying Discourse to Kubernetes
-version: 0.1.3
+version: 0.2.1
```
joancafom, line 4: This should be bumped to `0.2.0`

bitnami/discourse/README.md
```diff
@@ -110,7 +111,8 @@ The following table lists the configurable parameters of the Discourse chart and
 | `discourse.host`                          | Discourse host to create application URLs (include the port if =/= 80)                | `""`                                                         |
 | `discourse.siteName`                      | Discourse site name                                                                   | `My Site!`                                                   |
 | `discourse.username`                      | Admin user of the application                                                         | `user`                                                       |
-| `discourse.password`                      | Application password (min length of 10 chars)                                         | _random 10 character long alphanumeric string_               |
+| `discourse.password`                      | Application password (min length of 10 chars) - ignored if existingSecret is set to true                                        | _random 10 character long alphanumeric string_               |
```
joancafom, line 114: I think you mean "ignored if `existingSecret` is provided", am I right?

bitnami/discourse/README.md
```diff
@@ -110,7 +111,8 @@ The following table lists the configurable parameters of the Discourse chart and
...
+| `discourse.existingSecret`                      | Name of an exsting Kubernetes secret                                         | nil               |
```
joancafom, line 115: "exsting" --> existing

bitnami/discourse/values.yaml
```diff
@@ -68,6 +68,10 @@ discourse:
   ##
   # password:
 
+  ## Wheter to use an existing secret to store the Discourse passwords
```
joancafom, line 71: I think we can make it clearer by stating something like "Name of an existing secret containing the password (ignores previous `password`)"

bitnami/discourse/values.yaml
```diff
@@ -312,34 +316,46 @@ persistence:
   ##   GKE, AWS & OpenStack)
   ##
   # storageClass: "-"
-  ##
+
```
joancafom, line 319: We usually tend to keep common blocks together like this, and space out the different sections to ease the reading. For example, take a look:

https://github.com/bitnami/charts/blob/ccdebcbc99403900d0248fa22adc17c450c5261a/bitnami/airflow/values.yaml#L237-L248

https://github.com/bitnami/charts/blob/ccdebcbc99403900d0248fa22adc17c450c5261a/bitnami/airflow/values.yaml#L312-L322

bitnami/discourse/values.yaml
```diff
@@ -312,34 +316,46 @@ persistence:
...
   ## If you want to reuse an existing claim, you can pass the name of the PVC using
   ## the existingClaim variable
   # existingClaim: your-claim
-  accessMode: ReadWriteOnce
+  ## To use the portal and to ensure you can scale Discourse you need to provide a
+  ## ReadWriteMany PVC, if you don't have a provisioner for this type of storage
+  ## we recommend that you install the NFS provisioner and map it to a RWO volume
+  ## helm install nfs-server stable/nfs-server-provisioner --set persistence.enabled=true,persistence.size=10Gi
+  ##
+  accessMode: ReadWriteMany
```
joancafom, line 328: You are totally right here, but this Chart already offers a `values-production.yaml` file oriented to production-ready environments, which considers `ReadWriteMany` as the default value for this property in order to scale up. However, sometimes it is difficult to find a provisioner for this type and `values.yaml` takes that into account, easing the deployment when the number of replicas is 1.

bitnami/discourse/values.yaml
```diff
@@ -312,34 +316,46 @@ persistence:
...
   size: 10Gi
 
 ## Kubernetes configuration
 ## For minikube, set this to NodePort, elsewhere use LoadBalancer or ClusterIP
 ##
 service:
   type: LoadBalancer
+
```
joancafom, line 336: Same as before, not a huge issue at all, but we usually tend to keep the components of a block together.

bitnami/discourse/values.yaml
```diff
@@ -435,15 +455,35 @@ tolerations: []
...
 ##
 ## https://github.com/bitnami/charts/blob/master/bitnami/postgresql/values.yaml
 ##
 postgresql:
-  ## Whether to deploy a postgresql server to satisfy the applications database requirements. To use an external database set this to false and configure the externalDatabase parameters
+  ## Whether to deploy a postgresql server to satisfy the applications database requirements.
+  ## To use an external database set this to true.
   ##
-  enabled: true
+  useExistingDatabase: false
```
joancafom, line 466: This is a great idea, but I think we should be conservative in favour of standardization across other Charts available. The common approach is to use `externalX` and `X.enabled` so as to make it easy for users to use different Charts in a similar manner.

E.g, [Airflow](https://github.com/bitnami/charts/blob/ccdebcbc99403900d0248fa22adc17c450c5261a/bitnami/airflow/values.yaml#L548) or [Harbor](https://github.com/bitnami/charts/blob/ccdebcbc99403900d0248fa22adc17c450c5261a/bitnami/harbor/values.yaml#L2237) charts.

LucaPrete:
@joancafom thanks! I'll work on it right away and I'll send back a new patch asap.

bitnami/discourse/Chart.yaml
LucaPrete: done

bitnami/discourse/README.md
LucaPrete: yep :) good catch. Done

bitnami/discourse/README.md
LucaPrete: done

bitnami/discourse/values.yaml
LucaPrete: done

bitnami/discourse/values.yaml
LucaPrete: done

bitnami/discourse/values.yaml
LucaPrete: done

bitnami/discourse/values.yaml
LucaPrete: @joancafom I think the current approach duplicates lots of values and makes the chart management much (much!) more complex. Indeed the database section and the external section share exactly the same values, but the hostname of the database if this is not automatically created. Any chance we can maybe work on the wording, but keep the current logic?
It may be a good chance to re-think other charts as well? I can try to help there as needed :)

joancafom: Hi @LucaPrete !

Thank you very much for your help. We are currently discussing your proposal and thinking about the best way to tackle this, and make it a standard for the rest of charts as well. I will come back to you soon and provide with further info. 😄

Thanks for your feedback!

bitnami/discourse/requirements.yaml
```diff
@@ -2,8 +2,8 @@ dependencies:
   - name: postgresql
     repository: https://charts.bitnami.com/bitnami
     version: 8.X.X
-    condition: postgresql.enabled
+    condition: not postgresql.useExistingDatabase
   - name: redis
     repository: https://charts.bitnami.com/bitnami
     version: 10.X.X
-    condition: redis.enabled
+    condition: not redis.useExistingDatabase
```
dani8art, line 9: please don't change this.

bitnami/discourse/requirements.lock
```diff
@@ -1,9 +1,9 @@
 dependencies:
```
dani8art, line 1: Please, skip this file.

bitnami/discourse/values.yaml
joancafom: Hi @LucaPrete,

We have discussed long about this considering your feedback and reached to the conclusion that we should keep our current approach of splitting into two values, `XXXX` and `externalXXXX`. Although it is true that merging into a single value would reduce the redundancy of information, our experience with previous cases tells us that it would be more difficult for the final to use. Having to separate entries, although redundant, seems to be clear and less confusing, so we should proceed this way in detriment of redundancy.

Nevertheless, we appreciate a lot your feedback and have discussed long this because we consider it extremely valuable.

Best regards!

LucaPrete:
@dani8art @joancafom I've re-worked the patch, as requested.

It now contains the "existingSecrets" mod only. When you have a chance, it should be ready for review.

Let me say, it's a nightmare to develop (and use) this current DB config model. I've spent a couple of hours just to understand all the possible combinations of if in the NOTES.txt, and I'm not even convinced it's still totally correct...I totally understand it's crucial to keep consistency between charts for optimal user experience, but I'd recommend thinking again about it in the future...

joancafom:
Hi @LucaPrete 

Thanks for all your work, I am going to proceed with the revision. 

Regarding the current DB config model, thanks for your feedback. I want to let you know that we took your comments into account and discussed it with the team. We are currently undergoing some other standardizations for the Charts, so we considered including this one as well (we were not biased by the fact that this Chart was different to the rest because we would just simply change the others in pro of standardization 😁 ).

The main reason why we remain to this pattern is ease of use for the final consumer. Your idea of merging values is wonderful, it even would reduce redundancy. However, our experience tells us that having them separated is generally a clearer and more intuitive approach. This way, you have a disaggregated view of the parameters affecting a subchart deployment or an external connection. We actually have a specific [section](https://github.com/bitnami/charts/tree/master/bitnami/discourse#using-an-external-database) in README addressing this fact so that users find it even easier to use.

Nevertheless, your thoughts are always considered and still they brought up some other future discussions such us re-wording parameters.

Thanks very much!

bitnami/discourse/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: discourse
 description: A Helm chart for deploying Discourse to Kubernetes
-version: 0.1.7
+version: 0.1.8
```
joancafom, line 4: This should be bumped to `0.2.0`

LucaPrete:
Jose,

Thank you.
No worries, I totally understand.
It's great to know you guys take into account user inputs.

Looking forward to getting your feedback on the PR!

Il giorno mer 22 lug 2020 alle ore 09:46 Jose Antonio Carmona <
notifications@github.com> ha scritto:

bitnami/discourse/README.md
```diff
@@ -186,8 +187,9 @@ The following table lists the configurable parameters of the Discourse chart and
...
 | `postgresql.enabled`                      | Deploy PostgreSQL container(s)                                                        | `true`                                                       |
 | `postgresql.postgresqlUsername`           | PostgreSQL user to create (used by Discourse)                                         | `bn_discourse`                                               |
-| `postgresql.postgresqlPassword`           | Password for the Dicourse user                                                        | _random 10 character long alphanumeric string_               |
-| `postgresql.postgresqlPostgresPassword`   | Password for the admin user ("postgres")                                              | `bitnami`                                                    |
+| `postgresql.postgresqlPassword`           | Password for the Dicourse user - ignored if existingSecret is set to true                                                       | _random 10 character long alphanumeric string_               |
```
joancafom, line 190: Could you please re-word this?

...ignored if existingSecret _is set to true_ --> ignored if existingSecret _is provided_

Same for the `postgresql.postgresqlPostgresPassword` description, as existingSecret is not a boolean.

bitnami/discourse/README.md
```diff
@@ -203,7 +205,9 @@ The following table lists the configurable parameters of the Discourse chart and
 |-------------------------------------------|---------------------------------------------------------------------------------------|--------------------------------------------------------------|
 | `redis.enabled`                           | Deploy Redis container(s)                                                             | `true`                                                       |
 | `redis.usePassword`                       | Use password authentication                                                           | `false`                                                      |
-| `redis.password`                          | Password for Redis authentication                                                     | `nil`                                                        |
+| `redis.password`                          | Password for Redis authentication  - ignored if existingSecret is set to true                                                    | `my-password`                                                        |
```
joancafom, line 208: The re-wording I mentioned also applies here 😁

Also, `my-password` should be updated to `nil`.

bitnami/discourse/templates/_helpers.tpl
```diff
@@ -120,7 +120,29 @@ imagePullSecrets:
 {{- end -}}
 
 {{/*
-Get the user defined LoadBalancerIP for this release.
+Return the Discourse secret name
+*/}}
+{{- define "discourse.discourseSecretName" -}}
+{{- if not (empty .Values.discourse.existingSecret) }}
+    {{- .Values.discourse.existingSecret }}
```
joancafom, line 127: We can use the `printf` function here just to ensure proper indentation is followed.
```go
{{- printf "%s" .Values.discourse.existingSecret -}}
```

bitnami/discourse/templates/_helpers.tpl
```diff
@@ -120,7 +120,29 @@ imagePullSecrets:
 {{- end -}}
 
 {{/*
-Get the user defined LoadBalancerIP for this release.
+Return the Discourse secret name
+*/}}
+{{- define "discourse.discourseSecretName" -}}
+{{- if not (empty .Values.discourse.existingSecret) }}
```
joancafom, line 126: I think `empty` is used for arrays or collections. It may apply to string as well, but I think the standard approach for checking if `.existingSecret` is provided is by simply:

```go
{{- if  .Values.discourse.existingSecret }}
```

Any nil pointer or string whose length is 0 will evaluate to `false`.

bitnami/discourse/templates/_helpers.tpl
```diff
@@ -120,7 +120,29 @@ imagePullSecrets:
...
+    {{- .Values.discourse.existingSecret }}
+{{- else -}}
+    {{- printf "%s" (include "discourse.fullname" .) -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
+Return true if a secret object should be created
+*/}}
+{{- define "discourse.createSecret" -}}
```
joancafom, line 136: I think the best approach for this is to split the secrets into three, one for each of the main applications in this huge Chart. That would result in a simpler management and a more neat result IMHO.

For example, you can take a look at how EJBCA Chart addresses this problem:

- [externaldb-secrets](https://github.com/bitnami/charts/blob/master/bitnami/ejbca/templates/externaldb-secrets.yaml) : Used to store externaldb credentials
- [secrets](https://github.com/bitnami/charts/blob/master/bitnami/ejbca/templates/secrets.yaml): Used to store EJBCA's own sensitive data.

What's your opinion on this? If you think that is too much I can handle that myself 😀

LucaPrete:
@joancafom changes done! Ready again for review ;)

bitnami/discourse/templates/_helpers.tpl
```diff
@@ -120,7 +120,29 @@ imagePullSecrets:
 {{- end -}}
 
 {{/*
-Get the user defined LoadBalancerIP for this release.
+Return the Discourse secret name
+*/}}
+{{- define "discourse.discourseSecretName" -}}
```
joancafom, line 125: I think it'd be nicer to call this simply `discourse.secretName`. As this chart has multiple secrets coming from different sub-charts, you can take a look at [bitnami/airflow](https://github.com/bitnami/charts/blob/master/bitnami/airflow/templates/_helpers.tpl)  which follows this approach.

https://github.com/bitnami/charts/blob/37e4526071788e0adb2c14d9b427ecf42cc6c64d/bitnami/airflow/templates/_helpers.tpl#L328
https://github.com/bitnami/charts/blob/37e4526071788e0adb2c14d9b427ecf42cc6c64d/bitnami/airflow/templates/_helpers.tpl#L311
https://github.com/bitnami/charts/blob/37e4526071788e0adb2c14d9b427ecf42cc6c64d/bitnami/airflow/templates/_helpers.tpl#L290

bitnami/discourse/templates/_helpers.tpl
```diff
@@ -120,7 +120,29 @@ imagePullSecrets:
...
+    {{- printf "%s" .Values.discourse.existingSecret -}}
+{{- else -}}
+    {{- printf "%s-discourse" (include "discourse.fullname" .) -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
+Return true if a secret object for Discourse should be created
+*/}}
+{{- define "discourse.discourseCreateSecret" -}}
```
joancafom, line 136: This is totally fine, but we usually create these `XXXXX.createSecret` when the condition is complex so as to simplify in further references in the code. Here, I think we can just skip this helper and use:
```yaml
{{- if not .Values.discourse.existingSecret }}
```
in `templates/secrets-discourse.yaml`

bitnami/discourse/templates/_helpers.tpl
```diff
@@ -243,18 +265,35 @@ Return the PostgreSQL User
 {{- end -}}
 
 {{/*
-Return the PostgreSQL Secret
+Return true if a secret object for Postgres should be created
+*/}}
+{{- define "discourse.databaseCreateSecret" -}}
+{{- if or (and .Values.postgresql.enabled .Values.postgresql.existingSecret) (and (not .Values.postgresql.enabled) .Values.externalDatabase.existingSecret) }}
```
joancafom, line 271: Nice! Here the restriction is complex so a helper for whether to create the secret or not comes handy!!

bitnami/discourse/templates/_helpers.tpl
```diff
@@ -276,12 +315,29 @@ Return the Redis Port
...
+    {{- false -}}
+{{- else -}}
+    {{- true -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
+Return the Redis secret name
 */}}
 {{- define "discourse.redisSecretName" -}}
```
joancafom, line 331: As I have mention, I'd rather use `discourse.secretName` rather than `discourse.discourseSecretName`. The chart uses this naming pattern `discourse.redisSecretName`, it is okay if you leave `database` and `redis` as it is, but I'd definitely change the `discourse` one to `discourse.secretName`.

bitnami/discourse/templates/_helpers.tpl
joancafom: Nevertheless, this statement is not OK. Let me exemplify where we should create a secret or not.
Let `pe` be `postgres.enabled`, `es` be `postgres.existingSecret ` and `ees` be `externalDatabase.existingSecret`:

|        | es | not es | ees | not ees |
|--------|----|--------|-----|---------|
| **pe**     | NO | NO     | NO  | NO      |
| **not pe** | NO | NO     | NO  | YES     |

If `postgres.enabled` => Either the subchart or the externalSecret will handle the credentials
If not `postgres.enabled` => You must create a secret only in the case that the external one is not provided.

So the restriction (correct me if I am wrong) should be:
```yaml
if and (not .Values.postgresql.enabled) (not .Values.externalDatabase.existingSecret)
```

bitnami/discourse/templates/_helpers.tpl
```diff
@@ -276,12 +315,29 @@ Return the Redis Port
 {{- end -}}
 
 {{/*
-Return the Redis Secret
+Return true if a secret object for Redis should be created
+*/}}
+{{- define "discourse.redisCreateSecret" -}}
+{{- if or (and .Values.redis.enabled .Values.redis.existingSecret) (and (not .Values.redis.enabled) .Values.externalRedis.existingSecret) }}
```
joancafom, line 321: Similar to the other one.

bitnami/discourse/templates/secrets-database.yaml
```diff
@@ -0,0 +1,19 @@
...
+  labels: {{- include "discourse.labels" . | nindent 4 }}
+  {{- if .Values.commonLabels }}
+  {{- include "discourse.tplValue" ( dict "value" .Values.commonLabels "context" $ ) | nindent 4 }}
+  {{- end }}
+  {{- if .Values.commonAnnotations }}
+  annotations: {{- include "discourse.tplValue" ( dict "value" .Values.commonAnnotations "context" $ ) | nindent 4 }}
+  {{- end }}
+type: Opaque
+data:
+  {{- if and (not .Values.postgresql.enabled) (not .Values.externalDatabase.existingSecret) }}
```
joancafom, line 15: When the condition is updated, we should need to be doing this as the secret will only render in the correct scenarios

bitnami/discourse/templates/secrets-discourse.yaml
```diff
@@ -11,17 +12,11 @@ metadata:
   {{- end }}
 type: Opaque
 data:
+  {{- if not .Values.discourse.existingSecret }}
```
joancafom, line 15: This is not needed, we only create this secret if `.Values.discourse.existingSecret` is not provided.

bitnami/discourse/templates/_helpers.tpl
LucaPrete: Correct, good catch..this become quite confusing since the name of the secret is managed somewhere else...apologize for the confusion

bitnami/discourse/templates/secrets-database.yaml
LucaPrete: I think you meant SHOULD NOT :) Agree. Removed.

LucaPrete:
> Thanks so much again!! I have added a couple more comments, please check them out

@joancafom done! THANK YOU! Changes coming in a sec

bitnami/discourse/README.md
```diff
@@ -186,8 +187,9 @@ The following table lists the configurable parameters of the Discourse chart and
...
+| `postgresql.postgresqlPostgresPassword`   | Password for the admin user ("postgres") - ignored if existingSecret is provided                                             | `bitnami`                                                    |
+| `postgresql.existingSecret`   | Name of an existing Kubernetes secret. The secret must have the following keys configured: `postgresql-postgres-password`, `postgresql-password`, `postgresql-replication-password`                                              | `nil`                                                    |
 | `postgresql.postgresqlDatabase`           | Name of the database to create                                                        | `bitnami_application`                                        |
 | `postgresql.persistence.enabled`          | Enable database persistence using PVC                                                 | `true`                                                       |
 | `externalDatabase.host`                   | Host of the external database                                                         | `""`                                                         |
```
joancafom, line 195: Great! Don't forget to add an entry for `externalDatabase.existingSecret` as well 😁

bitnami/discourse/README.md
```diff
@@ -203,7 +205,9 @@ The following table lists the configurable parameters of the Discourse chart and
...
+| `redis.existingSecret`   | Name of an existing Kubernetes secret. The secret must have the `redis-password` key configured                                              | `nil`                                                           |
+| `redis.existingSecretPasswordKey`   | Name of the key pointing to the password in your Kubernetes secret                                              | `redis-password`                                                          |
 | `redis.cluster.enabled`                   | Whether to use cluster replication                                                    | `false`                                                      |
 | `redis.master.persistence.enabled`        | Enable database persistence using PVC                                                 | `true`                                                       |
 | `externalRedis.host`                      | Host of the external database                                                         | `""`                                                         |
```
joancafom, line 213: Same here regarding `externalRedis.existingSecret` 😁

bitnami/discourse/templates/_helpers.tpl
```diff
@@ -243,18 +254,35 @@ Return the PostgreSQL User
 {{- end -}}
 
 {{/*
-Return the PostgreSQL Secret
+Return true if a secret object for Postgres should be created
+*/}}
+{{- define "discourse.postgresql.createSecret" -}}
+{{- if and (not .Values.postgresql.enabled) (not .Values.externalDatabase.existingSecret) }}
+    {{- false -}}
```
joancafom, line 261: I think here is the other way around. This is `true` and the else statement should be `false`

bitnami/discourse/templates/_helpers.tpl
```diff
@@ -276,12 +304,29 @@ Return the Redis Port
 {{- end -}}
 
 {{/*
-Return the Redis Secret
+Return true if a secret object for Redis should be created
 */}}
-{{- define "discourse.redisSecretName" -}}
+{{- define "discourse.redis.createSecret" -}}
+{{- if and (not .Values.redis.enabled) (not .Values.externalRedis.existingSecret) }}
```
joancafom, line 310: You also have to bear in mind that `redis` may be provided without a password:

```yaml
{{- if and (not .Values.redis.enabled) (not .Values.externalRedis.existingSecret) (.Values.externalRedis. password) }}
```

bitnami/discourse/templates/_helpers.tpl
```diff
@@ -276,12 +304,29 @@ Return the Redis Port
...
+    {{- false -}}
```
joancafom, line 311: Similar to the postgres one 😁

joancafom:
Hey @LucaPrete,

Just to let you know I have implemented some changes on my last two commits. The PR is almost ready, however I would like to give it another revision yet to ensure everything works as expected.

Regards!

LucaPrete:
Jose, sounds good!

Could you please point me to the PRs?

-Luca

Il giorno mer 29 lug 2020 alle ore 17:17 Jose Antonio Carmona <
notifications@github.com> ha scritto:

joancafom:
Sure, my two last commits are c8534fe and 240ec7a respectively.

bitnami/discourse/templates/NOTES.txt
```diff
@@ -133,10 +150,18 @@ This deployment will be incomplete until you configure Discourse with a resolvab
...
+    --set redis.existingSecret={{ include "discourse.redis.secretName" . }} \
+    {{- else if not (empty .Values.redis.password) }}
     --set redis.password=$APP_REDIS_PASSWORD \
-    {{- else if and (not .Values.redis.enabled) (not (empty .Values.externalRedis.password)) }}
-    --set redis.externalRedis=$APP_REDIS_PASSWORD \
+    {{- end }}
+    {{- else if (not .Values.redis.enabled) }}
     --set redis.enabled=false \
+    {{- if .Values.externalRedis.existingSecret }}
+    --set externalRedis.existingSecret={{ include "discourse.redis.secretName" . }} \
```
joancafom, line 161: This is great but not necessary. The reason behind setting specific values when performing an upgrade is because there are certain fields (such as `passwords`) that are autogenerated if the final user does not provide with one. 

Imagine this scenario:

**1st Release** 
redis.password='' ==> The Chart autogenerates a random password, let's call it `alpha`

**2nd Release** 
We perform an upgrade from our previous release. Remember that `redis.password='' ` ==> The Chart will misunderstand this and replace `alpha` by a new random password, let's call it `beta`

Discourse cannot connect to Redis because it thinks `beta` is now the password!

However, if we explicitly set these parameter to fixed values, the release will be performed correctly 😁.

joancafom:
I think we are ready to launch this in our internal CI/CD tool, I'll come back with the results 😄🚀🎊

LucaPrete:
Amazing! Looking forward to it

Il giorno lun 3 ago 2020 alle 17:17 Jose Antonio Carmona <
notifications@github.com> ha scritto:

joancafom:
Hi! @LucaPrete the tests have succeeded! I will proceed with the merge 🎉🎊

LucaPrete:
Nice!

Il giorno mer 5 ago 2020 alle 12:07 Jose Antonio Carmona <
notifications@github.com> ha scritto:

PR #3037, opened 2020-07-06
[bitnami/thanos] add existing serviceaccount support

### LINKED ISSUE #3015: [bitnami/thanos] Allow providing an existing service account for all componenets
**Which chart**:
bitnami/thanos v0.13.0

I'm using IAM roles for service accounts in AWS EKS to give access to the S3 bucket for Thanos componenets. While this work quite nicely when there's a single service account, it's a bit of a problem when you can't control service account name.
Currently when creating the assume policy for the IAM role, I have to consider that there are 4 service accounts I need to give access to and add corresponding rules. Also it implies that I have to set the same annotation to 4 different service accounts when installing the helm chart.

Normally helm charts allow providing name of an existing service account via something like `*.serviceAccount.name` and when this property it's set, the helm chart doesn't create its service account. 

Providing the role via annotation for each service account and tweaking the assume policy, so it accepts several service accounts, however it seems kinda error prone. Creating a single service account and providing it for all components would allow reducing boilerplate.

dani8art:
**Description of the change**

Sometimes we want to use an existing ServiceAccount so we have added some fields to the  new value.yaml to support using a ServiceAccount which already exists in the cluster.

**Benefits**

You could use existig ServiceAccount in your cluster.

**Possible drawbacks**

N/A

**Applicable issues**

  - fixes #3015

**Additional information**

For testing propose you could use:

```bash
## should create service account and use it
$helm template thanos bitnami/thanos \
  --set querier.enabled=true \
  -s templates/querier/deployment.yaml \
  -s templates/querier/serviceaccount.yaml
```

```bash
## should use the common existing service account
helm template thanos bitnami/thanos \
  --set querier.enabled=true,existingServiceAccount=mySAcccount -s templates/querier/deployment.yaml
```

```bash
## should not create a new service account for querier
helm template thanos bitnami/thanos \
  --set querier.enabled=true,existingServiceAccount=mySAcccount -s templates/querier/serviceaccount.yaml
```

```bash
## should override the common existing service account name and use the querier service account.
helm template thanos bitnami/thanos \
  --set querier.enabled=true,existingServiceAccount=mySAcccount,querier.serviceAccount.existingServiceAccount=MyQuerierServiceAccount -s templates/querier/deployment.yaml
```

**Checklist**


Alexx-G:
Thanks a lot @dani8art for putting this together. It'll work perfectly for my use case.

bitnami/thanos/values-production.yaml
```diff
@@ -1119,10 +1138,10 @@ minio:
...
-    password: ""
+    password: ''
   secretKey:
-    password: ""
+    password: ''
 
   ## Default MinIO buckets
   ##
-  defaultBuckets: "thanos"
+  defaultBuckets: 'thanos'
```
joancafom, line 1147: Is there a reason this is changed here, but not in the regular `values.yaml`?

dani8art: Sorry, I formatted only `values-production.yaml`, fixed now! ;)

PR #3050, opened 2020-07-07
[bitnami/rabbitmq] Show the correct secret names in post-install notes

anton-johansson:
**Description of the change**

Prints the correct secret names in the post-install script. Technically shouldn't matter, because you should know your secrets if you created them yourselves, but still. This might avoid some future confusion.

**Benefits**

Less risk of confusion.

**Possible drawbacks**

None.

**Additional information**

:warning: I have not tested this locally, because I couldn't figure out how to package the Helm charts due to the dependency on `common`.

### DIFF
--- a/bitnami/rabbitmq/templates/NOTES.txt
+++ b/bitnami/rabbitmq/templates/NOTES.txt
@@ -4,8 +4,8 @@
 Credentials:
 
     echo "Username      : {{ .Values.auth.username }}"
-    echo "Password      : $(kubectl get secret --namespace {{ .Release.Namespace }} {{ template "rabbitmq.secretPasswordName" . }} -o jsonpath="{.data.rabbitmq-password}" | base64 --decode)"
-    echo "ErLang Cookie : $(kubectl get secret --namespace {{ .Release.Namespace }} {{ template "rabbitmq.secretErlangName" . }} -o jsonpath="{.data.rabbitmq-erlang-cookie}" | base64 --decode)"
+    echo "Password      : $(kubectl get secret --namespace {{ .Release.Namespace }} {{ include "rabbitmq.secretPasswordName" . }} -o jsonpath="{.data.rabbitmq-password}" | base64 --decode)"
+    echo "ErLang Cookie : $(kubectl get secret --namespace {{ .Release.Namespace }} {{ include "rabbitmq.secretErlangName" . }} -o jsonpath="{.data.rabbitmq-erlang-cookie}" | base64 --decode)"
 
 RabbitMQ can be accessed within the cluster on port {{ .Values.service.nodePort }} at {{ include "rabbitmq.fullname" . }}.{{ .Release.Namespace }}.svc.{{ .Values.clustering.k8s_domain }}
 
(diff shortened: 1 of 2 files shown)


bitnami/rabbitmq/templates/NOTES.txt
```diff
@@ -4,8 +4,8 @@
 Credentials:
 
     echo "Username      : {{ .Values.auth.username }}"
-    echo "Password      : $(kubectl get secret --namespace {{ .Release.Namespace }} {{ include "rabbitmq.fullname" . }} -o jsonpath="{.data.rabbitmq-password}" | base64 --decode)"
-    echo "ErLang Cookie : $(kubectl get secret --namespace {{ .Release.Namespace }} {{ include "rabbitmq.fullname" . }} -o jsonpath="{.data.rabbitmq-erlang-cookie}" | base64 --decode)"
+    echo "Password      : $(kubectl get secret --namespace {{ .Release.Namespace }} {{ template "rabbitmq.secretPasswordName" . }} -o jsonpath="{.data.rabbitmq-password}" | base64 --decode)"
+    echo "ErLang Cookie : $(kubectl get secret --namespace {{ .Release.Namespace }} {{ template "rabbitmq.secretErlangName" . }} -o jsonpath="{.data.rabbitmq-erlang-cookie}" | base64 --decode)"
```
andresbono, line 8:
```suggestion
    echo "Password      : $(kubectl get secret --namespace {{ .Release.Namespace }} {{ include "rabbitmq.secretPasswordName" . }} -o jsonpath="{.data.rabbitmq-password}" | base64 --decode)"
    echo "ErLang Cookie : $(kubectl get secret --namespace {{ .Release.Namespace }} {{ include "rabbitmq.secretErlangName" . }} -o jsonpath="{.data.rabbitmq-erlang-cookie}" | base64 --decode)"
```

andresbono:
Thanks so much for your contribution!

> ⚠️ I have not tested this locally, because I couldn't figure out how to package the Helm charts due to the dependency on `common`.

Hi, you can update the dependencies by executing `helm dependency update bitnami/rabbitmq/`.

bitnami/rabbitmq/templates/NOTES.txt
anton-johansson: @andresbono What's the difference between `include` and `template`? I did what was already done:
https://github.com/bitnami/charts/blob/f5f3ad2cb7737cfe6ef3a822c3e70313a5616799/bitnami/rabbitmq/templates/statefulset.yaml#L150

And this seems to work (I just tested it locally). :)

anton-johansson:
That worked a lot better, thanks @andresbono.

bitnami/rabbitmq/templates/NOTES.txt
andresbono: @anton-johansson in this case, both functions are actually equivalent. I just suggested the change to be consistent with the rest of the file.

Both functions behave differently if you use template pipelines (which is not the case). `include` supports it and `template` doesn't:

https://helm.sh/docs/howto/charts_tips_and_tricks/#using-the-include-function

anton-johansson: Thanks for the explanation! I have now changed.

anton-johansson:
I have now rebased with `master` and updated version accordingly.

andresbono:
Thanks for your contribution! :tada:

PR #3051, opened 2020-07-07
[bitnami/thanos] Drop the replica label on alerts for Ruler

hwoarang:
In case we have multiple replicas of Ruler, then we get multiple alerts
from each ruler pod. As such, we need to drop the 'replica' label before
we send the alert to the alertmanager in order for alertmanager to
perform deduplication and only send the alert once.

**Checklist** 


hwoarang:
@rafariossaa Is there a valid reason why you want to get duplicate alerts? If we want to hide this fix behind a flag then maybe there is no reason to add the flag in the first case as uses can use the extraFlags variable anyway.

rafariossaa:
Hi, 
  I am not very sure, I am not a thanos expert, but I guess if this is not the default behavior there should be a reason behind.
  Anyway, I will merge your PR and see if there are more comments on this topic.

rafariossaa:
Could you resolve the conflict ?

hwoarang:
> Could you resolve the conflict ?

@rafariossaa done :) Thanks

The reason this appears to not be enabled by default is that when Ruler replica:1 then there is not need for deduplication and dropping the label has no effect. But for multiple Ruler replicas it appears that you need it if you do not want duplicate alerts.

rafariossaa:
Hi,
 It seems that the chart version is not the right one.
  Could you check this ?

hwoarang:
@rafariossaa I have rebased again

PR #3068, opened 2020-07-08
[bitnami/mongodb] Fix MongoDB service port names

### LINKED ISSUE #3067: [bitnami/mongodb] Invalid port name for mongo services
_Originally posted in https://github.com/bitnami/charts/pull/2918_

Hey @juan131 , we just installed mongo using version 8.0.7 of the mongodb chart and noticed that we aren't able to connect to the deployed mongo using this connection string: `mongodb+srv://mongodb-headless.persistence.svc.cluster.local`

I found out that the reason is, that the port within the `mongodb-headless` service is named `tcp-mongodb` instead of `mongodb` and then I found this change here.

Could you please have a look again at your changes and ensure that connecting to mongo via `srv` still works?
This also affects the other service definitions that you created during your refactorings.

More information about this: 
* https://www.mongodb.com/blog/post/mongodb-3-6-here-to-SRV-you-with-easier-replica-set-connections -> mongo prepends `_mongodb._tcp` to the hostname 
* https://kubernetes.io/docs/concepts/services-networking/service/#dns: k8s is using the port name + protocol in the SRV DNS entry as a prefix

chris922:
Signed-off-by: Christian Bandowski <christian@myvm.de>

**Description of the change**

Changed the port names within the services from `tcp-mongodb` to `mongodb`.

**Benefits**

Connecting via MongoDB SRV connection string will work using the headless service. Right now it doesn't work because the name of the port will be used as a prefix for the required DNS query and mongo is using `mongodb` and not `tcp-mongodb` as the prefix for the port name. More information available in #3067.

**Possible drawbacks**

If someone already referencing the `tcp-mongodb` port by name he could get into troubles.

**Applicable issues**

  - fixes #3067

**Additional information**

**Checklist**

### DIFF
--- a/bitnami/mongodb/Chart.yaml
+++ b/bitnami/mongodb/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: mongodb
-version: 8.0.9
+version: 8.0.10
 appVersion: 4.2.8
 description: NoSQL document-oriented database that stores JSON-like documents with dynamic schemas, simplifying the integration of data in content-driven applications.
 keywords:
--- a/bitnami/mongodb/templates/arbiter/headless-svc.yaml
+++ b/bitnami/mongodb/templates/arbiter/headless-svc.yaml
@@ -13,7 +13,7 @@ spec:
   type: ClusterIP
   clusterIP: None
   ports:
-    - name: tcp-mongodb
+    - name: mongodb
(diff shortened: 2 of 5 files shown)


chris922:
I will close this one in favor of https://github.com/bitnami/charts/pull/3109

PR #3070, opened 2020-07-08
Print offending value in validateValues

dpc:
Hi,

I'm debugging why I'm hitting this issue and is really difficult because
it's a remote CI, and this chart is a dependency of my dependency,
and ... I don't want to waste your time.

My point is - if there is an invalid value that doesn't contain secrets, please print it.
Helm is such a terrible combination of templates and yamls files, with absolutely no
resistance for mistakes like typos, that in a larger setups is really difficult to track
the root cause of problems, and a printout like this at very least confirms if the
value is what we think it should be.

### DIFF
--- a/bitnami/postgresql-ha/Chart.yaml
+++ b/bitnami/postgresql-ha/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: postgresql-ha
-version: 3.4.0
+version: 3.4.1
 appVersion: 11.8.0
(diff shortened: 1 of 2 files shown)


javsalgar:
Hi!

Thank you very much for the contribution! Could you please bump the patch version?

dpc:
Like this?

javsalgar:
That's it! Thank you very much!

PR #3079, opened 2020-07-09
[bitnami/postgres] Transition to Primary/Replica terminology

kav:
**Description of the change**
Replaces master and slave with primary and readReplicas respectively. I've coalesced the respective values so existing configurations will continue to work but have updated all the docs to point to the newer terms.

**Benefits**
Let's use more inclusive language. Honestly it's more clear with primary and read replica terms what the actual difference is. Happy to tweak further of course. Just saw in my comment on the linked issue it perhaps should be standby?

**Possible drawbacks**
We currently use master and slave in pod names and labels and I've transitioned that. This may break custom scripts that are looking for those markers. I've bumped as non-breaking since values will still work fine but this might justify amending that.

I've not changed environment variables or anything that may require work on the image itself. Yet.

**Applicable issues**

  - related https://github.com/bitnami/bitnami-docker-postgresql/issues/175

**Additional information**

**Checklist** 


javsalgar:
Hi,

Thank you very much for the PR. However, I'm still not fully convinced that this may not need a major bump. Could you confirm that there are no issues with immutable fields when upgrading (for example, with a deployed networkpolicy)?

kav:
Oh good point, let me verify. If we need a major bump would you propose keeping the coalesce or moving to new terms completely?

javsalgar:
If we need a major bump I would not maintain the coalesce, but I would also add an extra set of standard parameters to the chart. We are in the process of adding these standardizations to all of our charts so I think that we can hold this PR and integrate everything in one major version bump (instead of having two that would be very painful for users).

kav:
Should we start a branch for the breaking changes, I'll update and PR to that or something else? What's the standard y'all are using here?

javsalgar:
Hi,

I just created this branch: `postgres-chart-v10`. Could you point to the PR to this branch?

javsalgar:
It seems there are some conflicts, could you check them?

kav:
I've merged but the changes in that branch don't lint successfully so I wasn't able to confirm I got a clean merge. Fairly certain but...

javsalgar:
However, before merging, could you fix this?

```
[ERROR] templates/: template: postgresql/templates/statefulset.yaml:35:22: executing "postgresql/templates/statefulset.yaml" at <.Values.master.podLabels>: nil pointer evaluating interface {}.podLabels
```

kav:
Hey @javsalgar looks like the postgres v10 branch is gone and not merged. Do you want me to reopen this?

javsalgar:
Hi! Sorry, there was a remote branch cleanup and it was accidentally deleted. We should restore that branch or re-create it again.

PR #3088, opened 2020-07-10
[bitnami/logstash] Update logstash ingress for newer K8S versions

michaelpporter:
Updates Logstash ingress for Kubernetes 1.14+

Checks Kubernetes version and chooses between "extensions/v1beta1" and "networking.k8s.io/v1beta1"

None known.

apiVersion is limited to two versions.

  - fixes #

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/logstash/templates/ingress.yaml
+++ b/bitnami/logstash/templates/ingress.yaml
@@ -1,5 +1,9 @@
 {{- if .Values.ingress.enabled -}}
+{{- if semverCompare ">=1.14-0" .Capabilities.KubeVersion.GitVersion -}}
+apiVersion: networking.k8s.io/v1beta1
+{{- else -}}
 apiVersion: extensions/v1beta1
+{{- end }}
 kind: Ingress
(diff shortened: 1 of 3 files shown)


bitnami/logstash/templates/ingress.yaml
```diff
@@ -1,5 +1,9 @@
 {{- if .Values.ingress.enabled -}}
+{{- if semverCompare ">=1.14-0" .Capabilities.KubeVersion.GitVersion -}}
+apiVersion: networking.k8s.io/v1beta1
+{{- else -}}
 apiVersion: extensions/v1beta1
+{{- end }}
```
juan131, line 6: We have a macro (`common.capabilities.ingress.apiVersion`) in the common library for this exactly purpose: https://github.com/bitnami/charts/blob/master/bitnami/common/templates/_capabilities.tpl#L16

However, it seems we're not using the common library as dependency on this chart (yet), so it's fine to me to use your approach at this moment.

PR #3093, opened 2020-07-10
bitnami/metallb: add pre-upgrade to helm hooks for memberlist

woodcockjosh:
**Description of the change**
Add pre-upgrade hook to memberlist

**Benefits**
If I add metallb chart as a dependency to another chart with an existing release, when I run `helm upgrade`  memberlist secret will be installed, as opposed to not being installed at all.

**Possible drawbacks**

**Applicable issues**
memberlist secret is not installed at all on upgrade
- https://github.com/helm/helm/issues/8401

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/metallb/Chart.yaml
+++ b/bitnami/metallb/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: metallb
 description: The Metal LB for Kubernetes
-version: 0.1.17
+version: 0.2.0
 appVersion: 0.9.3
 keywords:
   - "load-balancer"
--- a/bitnami/metallb/templates/secret.yaml
+++ b/bitnami/metallb/templates/secret.yaml
@@ -8,6 +8,6 @@ metadata:
   labels: {{- include "metallb.labels" . | nindent 4 }}
     app.kubernetes.io/component: speaker
   annotations:
-    "helm.sh/hook": "pre-install"
+    "helm.sh/hook": "pre-install,pre-upgrade"
     "helm.sh/hook-delete-policy": "before-hook-creation"
 {{- end }}


woodcockjosh:
The alternative fix to this problem is to remove the pre-install hook

dani8art:
Hi @woodcockjosh,

Thank you so much for the contribution!

why do you need it? I mean, since the secret is not removed in an upgrade action it will be always there, won't be?

BTW, I think this changes should be at least a minor bump of the chart version.

woodcockjosh:
@dani8art. I needed it because I added metallb as a dependency to an existing release. when you do this the only way to "install" the new dependency is to run "helm upgrade" in which case that hook will not run and that secret will not be installed at all. I have opened a bug report on helm but @bacongobbler has stated that this is not a bug because we can add the `pre-upgrade` hook. In my opinion the resources with the pre-install hook should still be installed as a pre-install hook if they are not already installed. I'm really not sure why I'm the only person running into this. Are people just installing every chart as a separate release? Anyway let me know if you have any other questions about why we would need this.

dani8art:
Ok, I see. 

Then basically we can remove the pre-hooks because at the end we are going to create it always, aren't we?

The thing is with this change we are going to recreate the `secret.yaml` every time we upgrade the chart and the `secretKey` inside will change every time because it is auto generated 

```yaml
data:
  secretkey: {{ randAlphaNum 256 | b64enc | quote }}
```

Then we should take this into account an document it in the README.md

woodcockjosh:
Yeah that seems weird. I honestly don't know what to do. I'll go with whatever you recommend. I feel like the real solution should be in the helm repo.

dani8art:
Hi @woodcockjosh sorry for my late response, 

We were discussing about how to tackle this. We are going to start a refactor in our charts to avoid errors with passwords/secrets when upgrading so this PR will be done by another in the future using the standards we have finally decided

Thank you so much for your feedback!!

woodcockjosh:
https://github.com/helm/charts/issues/5167

PR #3100, opened 2020-07-13
[bitnami/influxdb] Fix cronjob pod annotations

mkruliv:
Kubernetes doesn't create annotations for the cronjob pod


joancafom:
Hi @mkruliv ,

Thanks for your PR! Could you please bump the chart version according to semver? (in this case, the patch). Also, it would be nice if you could edit your description and fill in our generic template, it makes it easier for users to have an standardized approach in PR and an integrated checklist at the same time.

PR #3107, opened 2020-07-14
[bitnami/drupal] Fix typo in deployment env vars

andresbono:

**Checklist** 

### DIFF
--- a/bitnami/drupal/templates/deployment.yaml
+++ b/bitnami/drupal/templates/deployment.yaml
@@ -72,7 +72,7 @@ spec:
             - name: BITNAMI_DEBUG
               value: {{ ternary "true" "false" .Values.image.debug | quote }}
             - name: ALLOW_EMPTY_PASSWORD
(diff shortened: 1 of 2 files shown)


bitnami/drupal/templates/deployment.yaml
```diff
@@ -72,7 +72,7 @@ spec:
             - name: BITNAMI_DEBUG
               value: {{ ternary "true" "false" .Values.image.debug | quote }}
             - name: ALLOW_EMPTY_PASSWORD
-              value: {{ ternary "yes" "no" .Values. allowEmptyPassword | quote }}
+              value: {{ .Values.allowEmptyPassword | quote }}
```
juan131, line 75: The problem was the space between "Values" and "allowedEmptyPassword". Why do you remove the ternary?

andresbono: `.Values.allowEmptyPassword` is not a boolean, it's a `yes/no` string so it doesn't work with the ternary operator. I'm doing what is done in other standardized charts.

PR #3116, opened 2020-07-14
[bitnami/prometheus-operator] Expose additional ports from Thanos sidecar

hwoarang:
The sidecar is configured to expose the following two ports

- GRPC/10901
- HTTP/10902

However, the service that targets the sidecar does not expose the later
making it impossible to get any metrics from it through traditional
prometheus scraping as these are only available to the HTTP endpoint.
As such, add a new option to allow us to expose additional ports from
the sidecar container.

**Checklist** 
- [x] Title of the PR starts with chart name (e.g. `[bitnami/chart]`hart/kubeapps). This is only a synchronized mirror.


bitnami/prometheus-operator/templates/prometheus/thanos-service.yaml
```diff
@@ -29,6 +29,10 @@ spec:
       {{- if and .Values.prometheus.thanos.service.nodePort (or (eq .Values.prometheus.thanos.service.type "NodePort") (eq .Values.prometheus.thanos.service.type "LoadBalancer")) }}
       nodePort: {{ .Values.prometheus.thanos.service.nodePort }}
       {{- end }}
+    - name: http
+      port: 10902
+      targetPort: http
+      protocol: TCP
```
dani8art, line 35: I would do it customizable as grpc port is with `.Values.prometheus.thanos.service.port`. In fact, you could refactor this values:

```yaml
prometheus:
  thanos:
    service:
      ports:
        grpc: 10901
        http: 10902
```

then use them
```
.Values.prometheus.thanos.service.ports.grpc
.Values.prometheus.thanos.service.ports.http
```

bitnami/prometheus-operator/Chart.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: 0.40.0
 description: The Prometheus Operator for Kubernetes provides easy monitoring definitions for Kubernetes services and deployment and management of Prometheus instances.
 name: prometheus-operator
-version: 0.22.3
+version: 0.22.4
```
dani8art, line 5: I think this should be a minor bump

bitnami/prometheus-operator/templates/prometheus/thanos-service.yaml
hwoarang: @dani8art that would require a migration path for existing users (and as a result a major bump in Chart version). That's mainly the reason why I did not do it like this.

bitnami/prometheus-operator/Chart.yaml
hwoarang: @dani8art if we follow your approach then we will need a major bump as the change is breaking backwards compatibility and users need to perform manual steps to migrate no?

bitnami/prometheus-operator/templates/prometheus/thanos-service.yaml
dani8art: Yes you are right, and what about the `Values.prometheus.thanos.service.extraPorts` approach? please take a look at this https://github.com/bitnami/charts/blob/b0c0732b5ef07d31a84cb64d2770b34fee6ecc57/bitnami/discourse/templates/service.yaml#L36

dani8art: then you could set it by default in values:

```yaml
prometheus:
  thanos:
    service:
      extraPorts:
        - name: http
          port: 10902
          targetPort: http
          protocol: TCP
```

hwoarang: @dani8art that sounds like a very good suggestion. Thanks

hwoarang: It's done

bitnami/prometheus-operator/values-production.yaml
```diff
@@ -692,6 +692,15 @@ prometheus:
...
       annotations: {}
 
+      ## Extra ports to expose from the Thanos sidecar container
+      ##
+      # extraPorts:
+      #   - name: http
+      #     port: 10902
+      #     targetPort: http
+      #     protocol: TCP
+      extraPorts: {}
```
dani8art, line 702: it should be an array, please change it to `[]`

PR #3122, opened 2020-07-15
[bitnami/prometheus-operator] Adjust thanos-sidecar for prom-op no longer using mount subpaths

teozkr:
**Description of the change**

Seems like `storage.disableMountSubPath`'s default value was recently changed to `true`. This is a breaking change, in that it will break setups for users who have manually changed it to `false`. On the other hand, the current state leaves the Thanos sidecar silently unable to upload blocks.

This adapts the sidecar to use the root of the `prometheus-db` volume, just like the prom-op-managed containers.

**Benefits**

This should make the Thanos sidecar able to ship metrics again in its default configuration.

**Possible drawbacks**

This will have the opposite effect for people who have manually re-disabled `disableMountSubPath` via their values.yaml.

**Applicable issues**

  - fixes #

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/prometheus-operator/templates/prometheus/prometheus.yaml
+++ b/bitnami/prometheus-operator/templates/prometheus/prometheus.yaml
@@ -212,6 +212,9 @@ spec:
       volumeMounts:
         - mountPath: /prometheus
           name: prometheus-{{ template "prometheus-operator.prometheus.fullname" . }}-db
+          {{- if not (.Values.prometheus.storageSpec.disableMountSubPath | default true) }}
+          subPath: prometheus-db
+          {{- end }}
     {{- end }}
     {{- if .Values.prometheus.containers }}
     {{- include "prometheus-operator.tplValue" (dict "value" .Values.prometheus.containers "context" $) | nindent 4 }}
--- a/bitnami/prometheus-operator/templates/prometheus/prometheus.yaml
+++ b/bitnami/prometheus-operator/templates/prometheus/prometheus.yaml
@@ -212,7 +212,6 @@ spec:
       volumeMounts:
         - mountPath: /prometheus
           name: prometheus-{{ template "prometheus-operator.prometheus.fullname" . }}-db
-          subPath: prometheus-db
     {{- end }}
     {{- if .Values.prometheus.containers }}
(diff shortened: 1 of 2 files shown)


bitnami/prometheus-operator/templates/prometheus/prometheus.yaml
```diff
@@ -212,7 +212,6 @@ spec:
       volumeMounts:
         - mountPath: /prometheus
           name: prometheus-{{ template "prometheus-operator.prometheus.fullname" . }}-db
-          subPath: prometheus-db
```
dani8art, line 215: shouldn't it be optional depending on `storage.disableMountSubPath`?

teozkr: Yeah.. This was based on the assumption that `storageSpec`'s options are currently only passed through as is, and with no documentation in particular. I can change it to be conditional instead.

That said, `storage.disableMountSubPath` is deprecated and will presumably be removed soon.

teozkr: Oh, fun. So it turns out that the Helm chart still ships an old version of the `Prometheus` CRD that doesn't even have the option (https://github.com/coreos/prometheus-operator/blob/v0.40.0/bundle.yaml#L9274, doesn't exist in https://github.com/bitnami/charts/blob/b0c0732b5ef07d31a84cb64d2770b34fee6ecc57/bitnami/prometheus-operator/crds/crd-prometheus.yaml). So there are no existing users to break, unless they manually installed a different CRD.

dani8art: So, I think the best approach would be, keep it optional depending on `storage.disableMountSubPath` at least until we had upgraded our CRD version, don't you think so?

teozkr: The current CRD version doesn't allow setting `storage.disableMountSubPath` anyway.

teozkr: I'm still not really convinced that this is the correct path, but 362d7c90d makes it configurable.

teozkr:
Rebased

PR #3128, opened 2020-07-15
[bitnami/kafka] Add parameters to configure command and args for Kafka pod

### LINKED ISSUE #923: Awkward way to set Kafka broker.id
Hi, 

By default, broker.id is set to -1 (leading to auto generation by kafka).
If you delete the statefulset and then recreate it (while not recreating zookeeper as well), each pod will get a new (higher) broker id which will not match the broker ids of topic __consumer_offsets.
Then, consumer.subscribe() will hang forever.

Script /run.sh allows to override broker.id by setting environment variable BROKER_ID_COMMAND.
Pod name is <statefulset name>-N while N is 0 to replicas - 1.
I would like to set FIRST_BROKER_ID as an integer.
Then automatically set broker.id for each pod as FIRST_BROKER_ID + N.

The undocumented way to do it is to define the following environment variables:
        - name: FIRST_BROKER_ID
          value: "50"
        - name: BROKER_ID_COMMAND
          value: "if [[ $HOSTNAME =~ .+-([0-9]+)$ ]] ; then echo $[${FIRST_BROKER_ID:-10}+${BASH_REMATCH[1]}] ; else echo -1 ; fi"

It would be great, if BROKER_ID_COMMAND is documented.
Or even better, if the existence of FIRST_BROKER_ID will automatically set the appropriate broker ids without the need to specify BROKER_ID_COMMAND.

Thanks.

vpartington:
**Description of the change**

Adds parameters for the command and args of the `kafka` container. 

**Benefits**

Allows extra commands to be inserted before the standard behaviour of the pod is executed. For example, this can be used to patch the `meta.properties` file to allow for a seamless upgrade to v8.0.0 of the chart. See this issue for a discussion on that:
https://github.com/bitnami/charts/issues/923#issuecomment-654924491

**Possible drawbacks**

None.

**Applicable issues**

This does NOT fix #923. It does allow one to work around it, if one wants/needs to (like me).

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/kafka/Chart.yaml
+++ b/bitnami/kafka/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: kafka
-version: 11.4.0
+version: 11.5.0
 appVersion: 2.5.0
 description: Apache Kafka is a distributed streaming platform.
 keywords:
--- a/bitnami/kafka/README.md
+++ b/bitnami/kafka/README.md
@@ -147,6 +147,8 @@ The following tables lists the configurable parameters of the Kafka chart and th
 | `pdb.create`                                      | Enable/disable a Pod Disruption Budget creation                                                                                   | `false`                                                 |
 | `pdb.minAvailable`                                | Minimum number/percentage of pods that should remain scheduled                                                                    | `nil`                                                   |
 | `pdb.maxUnavailable`                              | Maximum number/percentage of pods that may be made unavailable                                                                    | `1`                                                     |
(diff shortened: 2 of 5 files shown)


bitnami/kafka/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: kafka
-version: 11.4.0
+version: 11.4.1
```
juan131, line 3: Please bump the minor version instead

vpartington: A "minor" comment to bump the "minor" version? ;-)

Done!

PR #3148, opened 2020-07-17
[bitnami/spark] Allow release namespace to be overridden through global parameter

rajan123456:
**Description of the change**

Allow users of the chart to override the release namespace through global parameters

**Benefits**

Users will be able to override the release namespace of the chart by passing a global parameter.
In case this chart is used as a sub-chart in an umbrella chart - the parent chart would be able to override the namespace parameter for this chart without affecting other sub-charts directly.

**Possible drawbacks**

None. The --namespace directive would still take precedence if specified.

**Applicable issues**

NA

**Additional information**

NA

**Checklist** 

### DIFF
--- a/bitnami/spark/Chart.yaml
+++ b/bitnami/spark/Chart.yaml
@@ -1,5 +1,5 @@
 apiVersion: v1
-version: 2.0.0
+version: 2.1.0
 appVersion: 3.0.0
 description: Spark is a fast and general-purpose cluster computing system.
 name: spark
--- a/bitnami/spark/README.md
+++ b/bitnami/spark/README.md
@@ -43,7 +43,13 @@ $ helm delete my-release
 
 The command removes all the Kubernetes components associated with the chart and deletes the release. Use the option `--purge` to delete all persistent volumes too.
 
-## Parameters
+## Global parameters
+
+| Parameter                                 | Description                                                                                                | Default                                                      |
+|-------------------------------------------|------------------------------------------------------------------------------------------------------------|--------------------------------------------------------------|
+| `global.namespaceOverride`                | Global string to override the release namespace                                                            | `nil`                                                        |
+
+## Common Parameters
 
 The following tables lists the configurable parameters of the spark chart and their default values.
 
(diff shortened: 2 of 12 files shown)


javsalgar:
Hi,

Thank you very much for the PR! Could you let us know the use case of this and why the `--namespace` of helm install wouldn't work here?

rajan123456:
> Hi,
> 
> Thank you very much for the PR! Could you let us know the use case of this and why the `--namespace` of helm install wouldn't work here?

I am working on an umbrella chart which depends on multiple sub-charts. I would like to install one of the sub-charts (cert-manager) in a different namespace compared to other resources. By passing the '--namespace' directive as part of the install command, all namespaces are set to the same value.

Also, I noticed some of the other charts in this repo provide this setting (mongodb, kafka, etc.),  hence wanted to extend it to the spark chart as well.

javsalgar:
Hi,

Thank you very much for the explanation. So, from what I understand, if we change `global.namespaceOverride` all the subcharts that use this value would be affected, am I right? Would it make sense to allow it as a non-global parameter as well?

rajan123456:
> Hi,
> 
> Thank you very much for the explanation. So, from what I understand, if we change `global.namespaceOverride` all the subcharts that use this value would be affected, am I right? Would it make sense to allow it as a non-global parameter as well?

Hi,

I am not sure if that would be useful. The global parameter is within the scope of the spark chart, so a parent chart can override the global parameter within the spark sub-chart's context; thereby avoiding affecting other sub-charts.

Ex.

umbrella
|- charts
|- |- spark-x.x.x

values.yaml (umbrella)
spark:
  global:
    namespaceOverride: customNamespace

javsalgar:
Hi,

As far as I know, the global parameter is in the scope of all the charts and sub-charts. That's what we use for example for modifying a repository so our charts work with the Azure or AWS marketplaces.

rajan123456:
Hi,

Sorry for the confusion. I was able to work around this requirement for my use case, so I no longer need this change in the chart.

PR #3155, opened 2020-07-17
[bitnami/rabbitmq] Fix values yaml template

### LINKED ISSUE #3095: New properties on values.yaml & values-production.yaml don't work
**Which chart**:
rabbitmq

**Describe the bug**
Install the helm chart with the proposed values.yaml and values-production.yaml doesn't consider new properties, example:
auth.username, auth.password, auth.erlangCookie, replicaCount.

**To Reproduce**
use the two values | values-production yaml, setting before the username, password and replicaCount, and you notice immediately those parameters keep original / default values.

**Expected behavior**
I expect username, password, erlangCookie and replicaCount are taken from my custom value.yaml 

helm v3.0.2, k8s v3.0.2

- Output of `helm version`:
version.BuildInfo{Version:"v3.0.2", GitCommit:"19e47ee3283ae98139d98460de796c1be1e3975f", GitTreeState:"clean", GoVersion:"go1.13.5"}
```
(paste your output here)
```

Client Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.0", GitCommit:"70132b0f130acc0bed193d9ba59dd186f0e634cf", GitTreeState:"clean", BuildDate:"2019-12-07T21:20:10Z", GoVersion:"go1.13.4", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.8", GitCommit:"ea670c3ed1fab723e191551e68da613f8a9c5667", GitTreeState:"clean", BuildDate:"2020-05-30T04:43:12Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"linux/amd64"}

```
(paste your output here)
```

**Additional context**
this is the values.yaml I use.
[values.txt](https://github.com/bitnami/charts/files/4905968/values.txt)

TheHunter:
**Description of the change**

These changes affects only for values.yaml and values-production.yaml.

**Benefits**

No expert rabbit users would waste time to understand the problem, and the installation could be easier.

**Possible drawbacks**

**Applicable issues**

  - fixes #3095

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/rabbitmq/Chart.yaml
+++ b/bitnami/rabbitmq/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: rabbitmq
-version: 7.5.5
+version: 7.5.6
 appVersion: 3.8.5
 description: Open source message broker software that implements the Advanced Message Queuing Protocol (AMQP)
 keywords:
--- a/bitnami/rabbitmq/values-production.yaml
+++ b/bitnami/rabbitmq/values-production.yaml
@@ -186,7 +186,7 @@ extraContainerPorts: []
 configuration: |-
   ## Username and password
   default_user = {{ .Values.auth.username }}
-  default_pass = CHANGEME
+  default_pass = {{ .Values.auth.password }}
   ## Clustering
   cluster_formation.peer_discovery_backend  = rabbit_peer_discovery_k8s
   cluster_formation.k8s.host = kubernetes.default.svc.{{ .Values.clusterDomain }}
--- a/bitnami/rabbitmq/values.yaml
+++ b/bitnami/rabbitmq/values.yaml
@@ -186,7 +186,7 @@ extraContainerPorts: []
 configuration: |-
   ## Username and password
   default_user = {{ .Values.auth.username }}
-  default_pass = CHANGEME
+  default_pass = {{ .Values.auth.password }}
   ## Clustering
   cluster_formation.peer_discovery_backend  = rabbit_peer_discovery_k8s
   cluster_formation.k8s.host = kubernetes.default.svc.{{ .Values.clusterDomain }}


TheHunter:
Hi all,
could anyone help me for this pull request ?

TheHunter:
Hi @juan131 
your're wellcome.
I don't understand well about RABBIT_PASSWORD env. variable, Is it taken from my custom values.yaml ? 

I know if you don't change this, you cannot set your own password.
image


Anyway, Exists another way to set a password ? 

This changes works, but I don't understand if exists another solution or workaround for this issue.

Please let me know.

juan131:
@TheHunter 

The `RABBITMQ_PASSWORD` is managed by RabbitMQ containers' logic to update the credentials during the initialization. Check the lines below:

- https://github.com/bitnami/bitnami-docker-rabbitmq/blob/master/3.8/debian-10/rootfs/opt/bitnami/scripts/librabbitmq.sh#L530
- https://github.com/bitnami/bitnami-docker-rabbitmq/blob/master/3.8/debian-10/rootfs/opt/bitnami/scripts/librabbitmq.sh#L430

jagguli:
The RABBITMQ_USERNAME and RABBITMQ_PASSWORD variables are being set

something is wrong with the initialization script ...

juan131:
Could you provide more information @jagguli ?? I just installed the chart and it works as expected.. Are you performing an upgrade of an existing release?

jagguli:
This is a fresh install.
I've cleard all existing data, set the auth.username and auth.existingPasswordSecret
I'm using the latest release 3.8.5-debian-10-r40
when I exec into the pod and chekc the variables are set as expected

I dont think change_password will work because the user does not exist, actuall I'm not sure its being called

juan131:
Hi @TheHunter 

Does your existing secret has a key named "rabbitmq-password"? Please note that RabbitMQ will be configured like this when using an existing secret:

```yaml
(...)
          environment:
            (...)
            - name: RABBITMQ_USERNAME
              value: "FOO"
            - name: RABBITMQ_PASSWORD
              valueFrom:
                secretKeyRef:
                  name: "SOME-EXISTING-SECRET"
                  key: rabbitmq-password
```

Therefore, it assumes that your existing secret has this format:

```yaml
apiVersion: v1
kind: Secret
metadata:
  (...)
type: Opaque
data:
  rabbitmq-password: ".........."
```

TheHunter:
**I don't use any secret** for password.. in your example you use a secret, please try to do it without it..

SarjakSShah:
@TheHunter: If you have enabled persistent volume and then try to delete the PVC first and then install this chart again

The scripts mentioned by @juan131 sets the password during initialization but this initialization steps are only performed if persistent volume doesn't already exists

https://github.com/bitnami/bitnami-docker-rabbitmq/blob/master/3.8/debian-10/rootfs/opt/bitnami/scripts/librabbitmq.sh#L530
https://github.com/bitnami/bitnami-docker-rabbitmq/blob/master/3.8/debian-10/rootfs/opt/bitnami/scripts/librabbitmq.sh#L430

Hope that helps

TheHunter:
@SarjakSShah, I've done that, I've created a new deployment without any volumes.. when I've retried, I removed existing volumes.

Can you do the same test I've done ? so, without secret, setting your credentials, and using any template values proposed by this chart capplication.

TheHunter:
I've notice now that my change is done in your yaml file, I'm talking about this:

default_pass = {{ .Values.auth.password }}

So, have you changed or created a different values.yaml ? different than values.yaml or values-production.yaml proposed by this chart ?

dani8art:
Hi @TheHunter 

> Anyway, Exists another way to set a my custom password ? This changes works, but I don't understand if exists another solution or workaround for this issue.

I did the same as @juan131 and I was able to change the password without these changes, please could you try to execute

```
$ helm install rabbitmq bitnami/rabbitmq --set auth.password=my-password
```

thomasLeclaire:
> Hi @TheHunter
> 
> Thanks for your PR!
> 
> This won't make any real difference since that will be overwritten during the container initialization with whatever value is set in the env. variable `RABBITMQ_PASSWORD`. See:
> 
> * https://github.com/bitnami/charts/blob/master/bitnami/rabbitmq/templates/statefulset.yaml#L156
> 
> And.. Since this is a credential, we shouldn't include it in the ConfigMap that's generated with the RabbitMQ configuration file.

Why? it add here some complexity into the deployment with no real added value it seems. 
If we talk about security concerns, the fact to have the password in an en variable or in the conf file sounds the same. 

Moreover, it means we don't have any way to update this password thanks to helm without have to delete the pvc, and therefore the possible data.

juan131:
Hi @TheHunter 

> Why? it add here some complexity into the deployment with no real added value it seems.
If we talk about security concerns, the fact to have the password in an en variable or in the conf file sounds the same.
> Moreover, it means we don't have any way to update this password thanks to helm without have to delete the pvc, and therefore the possible data.

Both using the ConfigMap or the env. variable to set the password, you'll have to use Helm to manage these kind of changes. I mean, you shouldn't edit the ConfigMap manually, but use a "helm upgrade" instead.

On the other hand, with this approach the password is saved on a secret (https://github.com/bitnami/charts/blob/master/bitnami/rabbitmq/templates/secrets.yaml) rather than a ConfigMap (https://github.com/bitnami/charts/blob/master/bitnami/rabbitmq/templates/configuration.yaml), following K8s best practices.

That said, I agree that we could mount these "secret" information in a file that RabbitMQ container reads instead of using an env. variable for security reasons. That's something we already do in other charts (e.g. MariaDB), see https://github.com/bitnami/charts/tree/master/bitnami/mariadb#production-configuration and the "injectSecretsAsVolume" parameters.

PR #2945, opened 2020-06-27
[bitnami/spring-cloud-data-flow] Fixes and improvements

donbeave:
**Description of the change**

1) Added ability to configure Hibernate dialect.
2) Fixed issue with locating `application.yaml` in Kubernetes cluster 
3) Added JDWP support
4) Removed MariaDB from database comments, because the external database can be PostgreSQL as well
5) Fixed some wrong properties (externalDatabase.server -> externalDatabase.dataflow)
6) Support for external RabbitMQ

**Benefits**

Most of these changes related to usage Spring Cloud Data Flow with an external database (for example PostgreSQL). Depends on database vendor and version, need to have the ability to set different Hibernate dialect.

Also dataflow server's and skipper server's `application.yaml` config directly mounted, which fixes some issues on Kubernetes platforms when configmap is not located/loaded by the name. This is how Spring Cloud Data Flow team used in their repository (https://github.com/spring-cloud/spring-cloud-dataflow/blob/master/src/kubernetes/server/server-deployment.yaml, https://github.com/spring-cloud/spring-cloud-dataflow/blob/master/src/kubernetes/skipper/skipper-deployment.yaml). It also written in documentation (https://docs.spring.io/spring-cloud-dataflow/docs/current-SNAPSHOT/reference/htmlsingle/#configuration-kubernetes-general), quote:

> The Spring Cloud Data Flow server for Kubernetes uses the spring-cloud-kubernetes module to process secrets that are mounted under `/etc/secrets`. ConfigMaps must be mounted as `application.yaml` in the `/config` directory that is processed by Spring Boot. To avoid access to the Kubernetes API server the `SPRING_CLOUD_KUBERNETES_CONFIG_ENABLE_API` and `SPRING_CLOUD_KUBERNETES_SECRETS_ENABLE_API` are set to `false`.

**Possible drawbacks**

N/A

**Applicable issues**

https://github.com/bitnami/charts/issues/2346

**Checklist**
- [X] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [X] Variables are documented in the README.md
- [X] Title of the PR starts with chart name (e.g. `[bitnami/chart]`)
- [X] If the chart contains a `values-production.yaml` apart from `values.yaml`, ensure that you implement the changes in both files

PR #3162, opened 2020-07-20
[bitnami/spring-cloud-dataflow] Add parameters to configure dataflow deployer properties and possibility to use existing password secrets

vaivanov:
**Description of the change**

Spring Cloud Dataflow chart improvements:  added nodeSelector and tolerations properties to deployer, provided possibility to use existing password secrets for datasource and RabbitMQ, added RabbitMQ virtual host parameter, renamed mariadb-password to datasource-password

**Benefits**

Improve process of customization how Streams and Tasks are deployed to kubernetes. Better security when we could use existing password secrets for external datasource and external rabbitmq instead of creating it with helm chart.

**Possible drawbacks**

None.

**Applicable issues**

None.

**Additional information**

None.

**Checklist**

### DIFF
--- a/bitnami/spring-cloud-dataflow/Chart.yaml
+++ b/bitnami/spring-cloud-dataflow/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: spring-cloud-dataflow
-version: 0.2.10
+version: 0.3.0
 appVersion: 2.5.3
 description: Spring Cloud Data Flow is a microservices-based toolkit for building streaming and batch data processing pipelines in Cloud Foundry and Kubernetes.
 keywords:
--- a/bitnami/spring-cloud-dataflow/Chart.yaml
+++ b/bitnami/spring-cloud-dataflow/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: spring-cloud-dataflow
-version: 0.2.9
+version: 0.2.10
 appVersion: 2.5.3
 description: Spring Cloud Data Flow is a microservices-based toolkit for building streaming and batch data processing pipelines in Cloud Foundry and Kubernetes.
 keywords:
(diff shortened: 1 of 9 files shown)


bitnami/spring-cloud-dataflow/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: spring-cloud-dataflow
-version: 0.2.9
+version: 0.2.10
```
carrodher, line 3: As there are different additions and not just bug fixing, I would bump the chart version in a minor instead of a patch
```suggestion
version: 0.3.0
```

bitnami/spring-cloud-dataflow/values.yaml
```diff
@@ -785,6 +790,8 @@ externalRabbitmq:
   ##
   username: guest
   password: guest
+  # vhost: /
```
carrodher, line 793: I just saw that the `externalRabbitmq` section is not present in the _values-production.yaml_ while it should appear in both files. It seems it was only implemented in the values.yaml by mistake (see https://github.com/bitnami/charts/pull/2945). Please, taking advantage of those changes, can you copy the whole
```yaml
##
## External RabbitMQ Configuration
##
## All of these values are ignored when rabbitmq.enabled is set to true
##
externalRabbitmq:
  ## Enables or disables external RabbitMQ, can be disabled when Kafka is using
  ##
  enabled: false
  ## RabbitMQ host and port
  ##
  host: localhost
  port: 5672
  ## RabbitMQ username and password, password will be saved in a kubernetes secret
  ##
  username: guest
  password: guest
  # vhost: /
  # existingPasswordSecret: name-of-existing-secret
```
code block to the _values-production.yaml_? On this way, we have unified both files and the new parameters `vhost` and `existingPasswordSecret` are present

vaivanov:
Thanks for you review. Fixes is on way

PR #3167, opened 2020-07-20
[bitnami/rabbitmq] Allow nodes to communicate with each other out of the box  when NetworkPolicy is enabled

machine424:
**Description of the change**

The Chart doesn't allow a cluster nodes to communicate with each other when:
```
networkPolicy:
    enabled: true
    allowExternal: true
```
As the nodes don't get the labels: https://github.com/bitnami/charts/blob/master/bitnami/rabbitmq/templates/networkpolicy.yaml#L26

I didn't want to remove them to avoid the breaking change, I don't know is adding a new `podSelector` is considered as a breaking change (as we change the behaviour), or a bugfix/patch.

**Checklist** 

### DIFF
--- a/bitnami/rabbitmq/templates/networkpolicy.yaml
+++ b/bitnami/rabbitmq/templates/networkpolicy.yaml
@@ -25,6 +25,9 @@ spec:
             matchLabels:
               app: {{ template "rabbitmq.name" . }}
               release: {{ .Release.Name }}
+        - podSelector:
+            matchLabels:
+            {{- include "common.labels.matchLabels" . | nindent 14 }}
         {{- if .Values.networkPolicy.additionalRules }}
         {{- include "common.tplvalues.render" (dict "value" .Values.networkPolicy.additionalRules "context" $) | nindent 8 }}
         {{- end }}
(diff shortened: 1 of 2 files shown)


joancafom:
Hi @machine424 

I might be wrong, but I think the situation you described happens when:

```yaml
networkPolicy:
    enabled: true
    allowExternal: false
```

Note the value for `allowExternal: false`, which requires client labelling for establishing connections. Am I right? 

Regarding the changes themselves, I could not see any other resources using the labels specified in current's `NetworkPolicy`, which were removed in favour of `common` in this PR #2843 . Hence, I guess we can safely remove that rule and replace with the common (this would be a fix, the previous rule was not working properly).

Thanks!

bitnami/rabbitmq/templates/networkpolicy.yaml
```diff
@@ -25,6 +25,9 @@ spec:
             matchLabels:
               app: {{ template "rabbitmq.name" . }}
```
joancafom, line 26: Please, replace this rule by the one you are providing. No other resource is using those matchLabels. 😃

PR #3171, opened 2020-07-21
[bitnami/rabbitmq] Fix prestop hook not applied when rebalance is true

mboutet:
**Description of the change**

Fix prestop hook not applied when rebalance is true

**Benefits**

Prestop hook is now defined when reblance is also enabled

**Possible drawbacks**

None

**Applicable issues**

None

**Additional information**

None

**Checklist** 


bitnami/rabbitmq/templates/statefulset.yaml
```diff
@@ -252,6 +245,12 @@ spec:
                     done
                     rabbitmq-queues rebalance "all"
           {{- end }}
+            preStop:
+              exec:
+                command:
+                  - bash
+                  - -ec
+                  - rabbitmqctl stop_app
```
juan131, line 253: I guess this should block should be inside a "{{ else }}", shouldn't it?? If I'm not wrong you want to avoid running this `rabbitmqctl stop_app` when rebalance is true

mboutet: Since these lifecycle hooks don't run at the same time, I'm not sure how it would cause problem. The preStop hook runs when the pod receives a SIGQUIT (so that rabbitmq is shutdown gracefully) whereas the postStart hook is executed after the pod is ready.

For reference, the rendered portion of the statefulset when `clustering.rebalance=true` and `replicaCount=3` is:
```
[...]
          resources:
            limits: {}
            requests: {}
          lifecycle:
            postStart:
              exec:
                command:
                  - /bin/bash
                  - -ec
                  - |
                    until rabbitmqctl cluster_status >/dev/null; do
                        echo "Waiting for cluster readiness..."
                        sleep 5
                    done
                    rabbitmq-queues rebalance "all"
            preStop:
              exec:
                command:
                  - bash
                  - -ec
                  - rabbitmqctl stop_app
          volumeMounts:
            - name: configuration
              mountPath: /bitnami/rabbitmq/conf
            - name: data
              mountPath: /bitnami/rabbitmq/mnesia
[...]
```

And when `clustering.rebalance=false` and `replicaCount=3`, it is:
```
[...]
          resources:
            limits: {}
            requests: {}
          lifecycle:
            preStop:
              exec:
                command:
                  - bash
                  - -ec
                  - rabbitmqctl stop_app
          volumeMounts:
            - name: configuration
              mountPath: /bitnami/rabbitmq/conf
            - name: data
              mountPath: /bitnami/rabbitmq/mnesia
[...]
```

We see that in both cases the preStop hook is present which, I think, is the expected behaviour.

Before this fix, there were two `spec.template.spec.containers[0].lifecycle` keys in the rendered template when `clustering.rebalance=true` and since `spec.template.spec.containers[0].lifecycle.postStart` was defined after the `spec.template.spec.containers[0].lifecycle.preStop` key in the rendered template, the `spec.template.spec.containers[0].lifecycle.preStop` was overwritten in the yaml.

juan131: Thanks for the detailed explanation

PR #3180, opened 2020-07-21
[bitnami/spring-cloud-dataflow] Added ability to configure volumeMounts, volumes and environmentVariables.

donbeave:
**Description of the change**

Added ability to configure volume mounts, volumes, and environment variables.

**Checklist**


carrodher:
Thanks for your contribution, changes in the code LGTM! Just a question, it would make sense to implement the same changes to the rest of the components like server, skipper, etc? I am not an active user of the application itself, so just asking in case it is useful

donbeave:
@carrodher yeah, you are right, I have added it to the server component as well.

PR #3192, opened 2020-07-22
[bitnami/rabbitmq]: 8.8.0 - adds auth.tls.existingSecretFullChain option

logicbomb421:
**Description of the change**

This change will allow one to specify that the `tls.crt` key in an existing TLS secret contains the full chain, including any Intermediates CAs.

**Benefits**

This is necessary when provisioning AMQPS certificates using jetstack/cert-manager, as the generated TLS secrets do not contain data for `ca.crt` when using an ACME issuer (e.g. Let's Encrypt). See jetstack/cert-manager#1571 for more info.

Without this change, attempting to connect to the AMQPS service backed with a cert-manager-provisioned Let's Encrypt certificate will result in a TLS error, e.g. in NodeJS:

```
{ Error: unable to verify the first certificate
  code: 'UNABLE_TO_VERIFY_LEAF_SIGNATURE' }
```

**Possible drawbacks**

This will default to `false`, so there should be no drawbacks to this addition.

**Checklist** 


joancafom:
Hi @logicbomb421 

Thanks for your PR! From what I understand, this change aims to override `ca_certificate.pem` with the contents of `server_certificate.pem`. I might be wrong, but wouldn't it be easier if for this specific use case the secret with the key `ca.crt` contained the same value as the one used in `tls.crt`?

logicbomb421:
Hu @joancafom 

That would definitely work, however I don't provision these TLS secrets myself -- they are automatically created by cert-manager, which does not include data for the `ca.crt` key. The current workaround is to make this change to the StatefulSet manually after installing the chart, which is less than ideal.

Alternatively, I could copy the `tls.crt` key data to the `ca.crt` key in the TLS secret, however cert-manager handles auto-renewing these before they expire, so the secret would need to be recreated every 60 or so days.

joancafom:
Hi @logicbomb421 

Oh, I see. I do completely understand your concerns regarding this changes now, but to me it seems like the issue itself is not chart but the way cert-manager handles these type of certificates. At the end of the day I might be wrong, but I think we are adapting the Chart to workaround an issue that from my point of view does need to be fixed in the origin. What are your thoughts on this? Do you agree? 😃

Another workaround could be to implement some type of job that periodically checks for renewals and performs the copy and generation of the TLS secret itself.

Best regards, and thanks for your PR!

logicbomb421:
Hi @joancafom,

I agree with you now, this falls outside the scope of this helm chart, and I have a simple workaround that is working for me.

Thanks for your time here!

WolfspiritM:
I'm also having this problem. The secret from cert-manager doesn't contain the ca.crt and without ca.crt this chart won't deploy:
```references non-existent secret key: ca.crt```

From issues I've found cert-manager says that as the tls.crt already contains intermediate certs and they won't inclide ca.crt as they can't retrieve the ca used.
Does Rabbit need the ca.crt if the certificate is signed with a trusted ca?
Can we probably have an option to not provide ca.crt at all as I think it's not required in this case?

https://github.com/jetstack/cert-manager/issues/1571

**Comment by juan131** [comment]
HI @WolfspiritM 

I think we can do sth like that.. We can adapt this PR so instead of using the **tls.crt** to set both **tls.crt** and **ca.crt**, it skips **ca.crt** when `auth.tls.existingSecretFullChain` is `true`.

What do you think about this approach??

**Comment by WolfspiritM** [comment]
@juan131 Sorry for the late response. I forgot about it and stumpled over that problem again as rabbit failed to start cause of the certificate being updated. I think that would be a good idea and would make working with cert-manager way easier. Right now I manually have to copy tls.crt to ca.crt in the secret just to make rabbit happy whenever lets encrypt certs are updated.

**Comment by juan131** [comment]
Hi @WolfspiritM 

> Right now I manually have to copy tls.crt to ca.crt in the secret just to make rabbit happy whenever lets encrypt certs are updated.

Couldn't you use the `sidecars` parameter to add a sidecar container on RabbitMQ pods so it does it automatically for you? I mean to copy `tls.crt` to `ca.crt`

**Comment by WolfspiritM** [comment]
Hi @juan131 
I've not worked with sidecar containers that much yet but I don't think that will work.
Especially as I somehow have to change the secret directly (instead of just copying files around in the pod) as otherwise the rabbit pods won't even come up as they try to mount ca.crt from the secret and fail cause ca.crt doesn't exist in there.

I don't remember the exact error right now but it was something along the line "failed to mount ca.crt".

Maybe the ca.crt part can be made optional in here?
https://github.com/bitnami/charts/blob/93f3434a62c60a279c2a52cb9f0008c439795550/bitnami/rabbitmq/templates/statefulset.yaml#L295-L306

**Comment by juan131** [comment]
Hi @WolfspiritM 

I can see your point. Let's try another approach:

- Manually create a secret (secret A) updated/managed by cert-manager.
- Launch a cron job that, on a daily basis, creates/updates a new secret (secret B) based on the certificates available in secret A, but using the format expected by RabbitMQ.
- Finally, install RabbitMQ using a configuration like then one below:

    ```yaml
    auth:
      tls:
        enabled: true
        existingSecret: SECRET-B
    ```

Does that makes sense?

**Comment by logicbomb421** [comment]
I'm running into this problem again myself.

I'm curious about the pushback here, as this full chain cert is a well known thing cert-manager does. Further, this chart appears to be built with cert-manager usage in mind, based on the `ingress.certManager` option in `values.yaml`.

It also appears the current versions of cert-manager (I'm on 1.1.0) no longer add the empty `ca.crt` key to the provisioned TLS secret, which now causes the StatefulSet containers to fail to create due to `MountVolume.SetUp failed for volume "certs" : references non-existent secret key: ca.crt`.

Now it seems like there are multiple users of this chart experiencing this issue, and so far all we have received are several workaround solutions that don't have the same stability as a first-class chart configuration option. I am trying to build a terraform module to spin up a Kubernetes cluster and all resources we run at my company, and currently this doesn't work since the certificate secret keys aren't correct.

I'm currently using my own fork since it has this fix in it, but I would really like to go back to the official chart. Can I get a solid reason why this can't be added please?

**Comment by juan131** [comment]
Hi @logicbomb421 

Thanks for your feedback.

> Further, this chart appears to be built with cert-manager usage in mind, based on the ingress.certManager option in values.yaml.

Please note that option is to provide certificates when accessing via HTTPS to the RabbitMQ management UI, using the the ingress rule that can be optionally created with the chart.

We're talking about different use cases, since the certificates you're struggling with are intended to encrypt communication between RabbitMQ peers.

> but I would really like to go back to the official chart. Can I get a solid reason why this can't be added please?

I'm glad to discuss better alternatives to make this chart compatible with cert-manager without using "workaround solutions" as you said. Please feel free to send a PR where we can discuss the alternatives to implement this.

**Comment by logicbomb421** [comment]
@juan131 Thank you for the reply.

> Please note that option is to provide certificates when accessing via HTTPS to the RabbitMQ management UI, using the the ingress rule that can be optionally created with the chart.

This is a fair point, however they both are functionally the same, so I'm having trouble understanding why this is a concern. Please correct me if my understanding of the UI/AMQPS certs is incorrect.

> Please feel free to send a PR where we can discuss the alternatives to implement this.

Not trying to be pushy here, but why exactly is _this_ PR not acceptable for this?

**Comment by juan131** [comment]
> Not trying to be pushy here, but why exactly is this PR not acceptable for this?

I'm very sorry @logicbomb421 . I though this was an issue instead of a PR... 🤦 
I guess this is what happens when you review too many issues/prs on a Friday. My apologies.

Reopening the PR and re-reviewing it...

File: bitnami/rabbitmq/README.md [hunk 1]
```diff
@@ -83,6 +83,7 @@ The following table lists the configurable parameters of the RabbitMQ chart and
 | `auth.tls.serverCertificate`              | Server certificate content                                                                                           | `nil`                                                        |
 | `auth.tls.serverKey`                      | Server private key content                                                                                           | `nil`                                                        |
 | `auth.tls.existingSecret`                 | Existing secret with certificate content to RabbitMQ credentials                                                     | `nil`                                                        |
+|`auth.tls.existingSecretFullChain`         | Whether or not the existing secret contains the full chain in the certificate (`tls.crt`). Will be used in place of `ca.cert` if `true`. | `false`|
```
juan131, line 86: Should we document how to use this as a subsection of the `### Enabling TLS support` section? I mean, sth like:
```
#### TLS integration with cert-manager

...
```
And provide some basic instructions to use cert-manager to generate the certificates, and how to use them on the RabbitMQ chart

juan131:
Please remember to rebase your branch from master, and update the PR

bitnami/rabbitmq/README.md
logicbomb421: Good call, added. Since I have only confirmed this issue with Let's Encrypt certificates, I made sure to specify such in the docs.

bitnami/rabbitmq/README.md
```diff
@@ -353,6 +354,12 @@ auth:
 - Setting [auth.tls.failIfNoPeerCert](https://www.rabbitmq.com/ssl.html#peer-verification-configuration) to `false` allows a TLS connection if client fails to provide a certificate.
 - When setting [auth.tls.sslOptionsVerify](https://www.rabbitmq.com/ssl.html#peer-verification-configuration) to `verify_peer`, the node must perform peer verification. When set to `verify_none`, peer verification will be disabled and certificate exchange won't be performed.
 
+#### TLS integration with `cert-manager` (Let's Encrypt certificates)
+
+If using `cert-manager` to provision Let's Encrypt certificates, the `tls.crt` key in the generated TLS secret will contain the full certificate chain. Depending on the version of `cert-manager` in use, there can either be an empty `ca.crt` key, or none at all.
+
+In order to instruct RabbitMQ to look for the CA certificate within the primary certificate, `auth.tls.existingSecretFullChain` can be set to `true`.
```
juan131, line 361: nice!! Thanks so much

juan131 (approved):
Approving the changes!

juan131:
PR merged! Now we need to wait a few hours for our pipeline to detect the commit and trigger the release process to publish the new chart version.

PR #3195, opened 2020-07-23
[bitnami/prometheus-operator] Use the operator template for container name

hwoarang:
Using the top level 'prometheus-operator.name' template can have weird
consequences in some cases. For example, imagine we use the default
value of 'prometheus-operator' as Chart name. In this case, the template
will strip the '-operator' part and the container will end up being
named 'prometheus'. However, the name 'prometheus' is also used by the
prometheus CRD when it creates prometheus instances. As such, it makes
it harder to distinguish between the prometheus-operator container and
those in the actual prometheus instances. It's better to have a clear
identification for the container that belongs to the opearator

**Checklist** 

### DIFF
--- a/bitnami/prometheus-operator/Chart.yaml
+++ b/bitnami/prometheus-operator/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: 0.40.0
 description: The Prometheus Operator for Kubernetes provides easy monitoring definitions for Kubernetes services and deployment and management of Prometheus instances.
 name: prometheus-operator
-version: 0.25.0
+version: 0.25.1
 keywords:
   - prometheus
   - alertmanager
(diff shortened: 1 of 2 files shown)


marcosbc:
Sorry, but I'm a bit confused with this change. If I deploy the chart in the following way:

```console
$ helm install prometheus-operator bitnami/prometheus-operator
```

Then the container does have the `-operator` suffix:

```console
$ kubectl get pods prometheus-operator-operator-686dbd6f9c-dfc59
NAME                                            READY   STATUS    RESTARTS   AGE
prometheus-operator-operator-686dbd6f9c-dfc59   1/1     Running   0          2m4s
$ kubectl get pods prometheus-operator-operator-686dbd6f9c-dfc59 -o jsonpath='{.spec.containers[*].name}'
prometheus-operator
```

Could you let me know if there's anything I'm missing?

hwoarang:
Yes I think you are right. There are some overrides locally and these are likely the root cause of what I am seeing. I am going to close this for now. Apologies for the noise :)

PR #3196, opened 2020-07-23
[bitnami/etcd] Ensure ETCD_DATA_DIR permissions are set to 700

carrodher:
**Description of the change**

Since etcd 3.4.10 (release in process on our side), etcd checks that permissions of the data directory are 700, in order to support upgrades in a smooth way, we are going to set the proper permissions before initializing etcd if needed

The image for the new version already contains the data dir folder created using the proper permission

**Checklist** 

### DIFF
--- a/bitnami/etcd/templates/scripts-configmap.yaml
+++ b/bitnami/etcd/templates/scripts-configmap.yaml
@@ -144,6 +144,12 @@ data:
 {{- end }}
     else
         echo "==> Detected data from previous deployments..." 1>&3 2>&4
+        if [[ $(stat -c "%a" "$ETCD_DATA_DIR") != *700 ]]; then
+            echo "==> Setting data directory permissions to 700 in a recursive way (required in etcd >=3.4.10)" 1>&3 2>&4
+            chmod -R 700 $ETCD_DATA_DIR
+        else
+            echo "==> The data directory is already configured with the proper permissions" 1>&3 2>&4
+        fi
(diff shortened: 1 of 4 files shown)


bitnami/etcd/templates/scripts-configmap.yaml
```diff
@@ -181,6 +181,13 @@ data:
         fi
     fi
 
+    if [[ $(stat -c "%a" "$ETCD_DATA_DIR") != *700 ]]; then
+        echo "==> Setting data directory permissions to 700 in a recursive way (required in etcd >=3.4.10)" 1>&3 2>&4
+        chmod -R 700 $ETCD_DATA_DIR
+    else
+        echo "==> The data directory is already configured with the proper permissions" 1>&3 2>&4
+    fi
+
```
juan131, line 190: Add the block before line 122

carrodher: In that case, we can not ensure that the data directory exists at this point:
```
stat: cannot stat '/bitnami/etcd/data': No such file or directory
==> Setting data directory permissions to 700 in a recursive way (required in etcd >=3.4.10)
chmod: cannot access '/bitnami/etcd/data': No such file or directory
```
we can place the block in the else of `if [[ ! -d "$ETCD_DATA_DIR" ]]; then` if we want to ensure that the data directory is created from a previous deployment or as it's right now that is going to check everything (previous deployments and created in this one)

PR #3222, opened 2020-07-24
[bitnami/rabbitmq] Run secret name through tpl

almahmoud:
**Description of the change**

Runs `auth.existingPasswordSecret` through tpl

**Benefits**

Fixes situations like: https://github.com/galaxyproject/galaxy-helm/pull/152#issuecomment-663510166

**Checklist** 


bitnami/rabbitmq/templates/_helpers.tpl
```diff
@@ -64,7 +64,7 @@ Get the password secret.
 */}}
 {{- define "rabbitmq.secretPasswordName" -}}
     {{- if .Values.auth.existingPasswordSecret -}}
-        {{- printf "%s" .Values.auth.existingPasswordSecret -}}
+        {{- printf "%s" (include "common.tplvalues.render" (dict "value" .Values.auth.existingPasswordSecret "context" .)) -}}
```
juan131, line 67: Since it's a just a string with the secret name, we can make it simpler and avoid using `common.tplvalues.render`:
```suggestion
        {{- printf "%s" (tpl .Values.auth.existingPasswordSecret $) -}}
```

PR #3234, opened 2020-07-27
[bitnami/mongodb-sharded] add podManagementPolicy config for statefulsets to enable parallel startup

oldluke92:
**Description of the change**

- making the statefulsets configurable to enable parallel startup of db pods
- Default to "OrderedReady" to stay backwards compatible to deployed clusters

**Benefits**
- this speeds up db startup and improves resilency against concecutive failures. For example if you got a setup using 3 data replicas and the "middle" one fails because of a synchronisation error, you will not be able to restart/move the 3rd one because the middle one not being ready prevents the third one from being created again.

**Possible drawbacks**

You tell me. Maybe there was a reason for not allowing parallel startup.


bitnami/mongodb-sharded/templates/config-server/config-server-statefulset.yaml
```diff
@@ -11,6 +11,7 @@ spec:
       app.kubernetes.io/component: configsvr
   serviceName: {{ include "mongodb-sharded.fullname" . }}-headless
   replicas: {{ .Values.configsvr.replicas }}
+  podManagementPolicy: {{ $.Values.configsvr.podManagementPolicy }}
```
juan131, line 14: You don't need to use the `$.`. Suggestion:
```
  podManagementPolicy: {{ .Values.configsvr.podManagementPolicy }}
```

bitnami/mongodb-sharded/templates/shard/shard-arbiter-statefulset.yaml
```diff
@@ -11,6 +11,7 @@ spec:
   selector:
     matchLabels: {{- include "mongodb-sharded.matchLabels" $ | nindent 6 }}
       app.kubernetes.io/component: shardsvr-arbiter
+  podManagementPolicy: {{ $.Values.shardsvr.arbiter.podManagementPolicy }}
```
juan131, line 14: You don't need to use the `$.`. Suggestion:
```
  podManagementPolicy: {{ .Values.shardsvr.arbiter.podManagementPolicy }}
```

bitnami/mongodb-sharded/templates/shard/shard-data-statefulset.yaml
```diff
@@ -11,6 +11,7 @@ spec:
   selector:
     matchLabels: {{- include "mongodb-sharded.matchLabels" $ | nindent 6 }}
       app.kubernetes.io/component: shardsvr
+  podManagementPolicy: {{ $.Values.shardsvr.dataNode.podManagementPolicy }}
```
juan131, line 14: You don't need to use the `$.`. Suggestion:
```
  podManagementPolicy: {{ .Values.shardsvr.dataNode.podManagementPolicy }}
```

oldluke92:
Well it seems it doesn't work like that :(

juan131:
You're right
Please revert the suggestions (except in the **config-server-statefulset.yaml** where it actually works)

juan131:
Thanks so much!

oldluke92:
It seems that the bootstrapping scripts for the data replicasets do not support a parallel initial spinup of the cluster nodes. This is a problem, as the problem described in the "benefits" section of this PR might arise (and did, in our deployment ..) when unsing OrderedReady. Shall I open an issue for that? I don't really have the time to search for that (and I would also not know where to start) ....
Unfortunately, the cluster can also not be deployed using OrderedReady and then changed to parallel later - because this is a field that cannot be changed....

PR #3240, opened 2020-07-27
[bitnami/fluentd] Add missing service type for fluentd aggregator

### LINKED ISSUE #3239: [bitnami/fluentd] Aggregator service type in values is not being used in actual service
**Which chart**:
- bitnami/fluentd v1.2.9 (and older)

**Describe the bug**
In the `values.yaml` file it allows users to specify the `type` and `annotations` of the aggregator service as well as setting some LoadBalancer values such as `LoadBalancerIP` and `loadBalancerSourceRanges`

```yaml
aggregator:
  service:
    ## Service type
    ##
    type: ClusterIP

    ## Service ports
    ##
    ports:
      # syslog-udp:
      #   port: 5140
      #   targetPort: syslog-udp
      #   protocol: UDP
      #   nodePort: 31514
      # syslog-tcp:
      #   port: 5140
      #   targetPort: syslog-tcp
      #   protocol: TCP
      #   nodePort: 31514
      # tcp:
      #   port: 24224
      #   targetPort: tcp
      #   protocol: TCP
      http:
        port: 9880
        targetPort: http
        protocol: TCP

    ## loadBalancerIP for the forwarders service (optional, cloud specific)
    ## ref: http://kubernetes.io/docs/user-guide/services/#type-loadbalancer
    ##
    # loadBalancerIP:

    ## Load Balancer sources
    ## https://kubernetes.io/docs/tasks/access-application-cluster/configure-cloud-provider-firewall/#restrict-access-for-loadbalancer-service
    ##
    # loadBalancerSourceRanges:
    #   - 10.10.10.0/24

    ## Set the Cluster IP to use
    ## ref: https://kubernetes.io/docs/concepts/services-networking/service/#choosing-your-own-ip-address
    ##
    # clusterIP: None

    ## Provide any additional annotations which may be required
    ##
    annotations: {}
```

Although on the service specification it is: 
```yaml
spec:
  type: ClusterIP
  clusterIP: None
  ports:
  {{- range $key, $value := .Values.aggregator.service.ports }}
    - name: {{ $key }}
      {{ toYaml $value | nindent 6 }}
  {{- end }}
  selector: {{- include "fluentd.matchLabels" . | nindent 4 }}
    app.kubernetes.io/component: aggregator
{{- end -}}
```

**To Reproduce**
1. On `values.yaml` set the `aggregator.service.type` to `LoadBalancer`
2. On `values.yaml` add an annotation on the `aggregator.service.annotation` value
3. Install the Helm Chart
4. `kubectl get services`
5. See that the service created is of type ClusterIP
6. See that the service created does not have the annotation added

**Expected behavior**
It is expected that the `aggregator` service reflects the settings available on the `values.yaml` file.

Helm Version:
```
version.BuildInfo{Version:"v3.2.4", GitCommit:"0ad800ef43d3b826f31a5ad8dfbb4fe05d143688", GitTreeState:"dirty", GoVersion:"go1.14.3"}
```

Kubernetes Version:
```
Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.4", GitCommit:"c96aede7b5205121079932896c4ad89bb93260af", GitTreeState:"clean", BuildDate:"2020-06-18T02:59:13Z", GoVersion:"go1.14.3", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"16+", GitVersion:"v1.16.9-gke.6", GitCommit:"14bc8ad5f8c245f1240a8e4eab128c4d51bfeffe", GitTreeState:"clean", BuildDate:"2020-05-30T02:07:03Z", GoVersion:"go1.13.9b4", Compiler:"gc", Platform:"linux/amd64"}
```

fsschmitt:
**Description of the change**

- Current `master` version of bitnami chart for `fluentd` has in `values.yaml` described to specify the service type and LoadBalancer specs for the aggregator, although the service does not provide that support to actually use these values;
- This change adds that functionality that was already expected in the `values.yaml` file as it can be seen on issue #3239

**Benefits**

- This will enable users to use fluentd as an aggregator through a LoadBalancer service (e.g. across clusters or zones through Internal LB)

**Possible drawbacks**

- None can be foreseen given that this just extends the functionality, everything else remains the same;

**Applicable issues**
  - fixes #3239

**Checklist**
- [x] Variables are documented in the README.md **(nothing changed)**
- [x] If the chart contains a `values-production.yaml` apart from `values.yaml`, ensure that you implement the changes in both files **(nothing changed)**

### DIFF
--- a/bitnami/fluentd/templates/svc-headless.yaml
+++ b/bitnami/fluentd/templates/svc-headless.yaml
@@ -6,6 +6,9 @@ metadata:
   labels: {{- include "fluentd.labels" . | nindent 4 }}
     app.kubernetes.io/component: aggregator
     app: aggregator
+  {{- if .Values.aggregator.service.annotations }}
+  annotations: {{- include "fluentd.tplValue" (dict "value" .Values.aggregator.service.annotations "context" $) | nindent 4 }}
+  {{- end }}
 spec:
   type: {{ .Values.aggregator.service.type }}
   {{- if and .Values.aggregator.service.loadBalancerIP (eq .Values.aggregator.service.type "LoadBalancer") }}
--- a/bitnami/fluentd/templates/svc-headless.yaml
+++ b/bitnami/fluentd/templates/svc-headless.yaml
@@ -6,23 +6,9 @@ metadata:
   labels: {{- include "fluentd.labels" . | nindent 4 }}
     app.kubernetes.io/component: aggregator
     app: aggregator
-  {{- if .Values.aggregator.service.annotations }}
-  annotations: {{- include "fluentd.tplValue" (dict "value" .Values.aggregator.service.annotations "context" $) | nindent 4 }}
-  {{- end }}
 spec:
-  type: {{ .Values.aggregator.service.type }}
-  {{- if and .Values.aggregator.service.loadBalancerIP (eq .Values.aggregator.service.type "LoadBalancer") }}
-  loadBalancerIP: {{ .Values.aggregator.service.loadBalancerIP }}
-  {{- end }}
-  {{- if and .Values.aggregator.service.loadBalancerSourceRanges (eq .Values.aggregator.service.type "LoadBalancer") }}
-  loadBalancerSourceRanges:
-  {{- with .Values.aggregator.service.loadBalancerSourceRanges }}
-    {{ toYaml . | nindent 4 }}
-  {{- end }}
-  {{- end }}
-  {{- if eq .Values.aggregator.service.type "ClusterIP" }}
+  type: ClusterIP
   clusterIP: None
-  {{- end }}
   ports:
   {{- range $key, $value := .Values.aggregator.service.ports }}
     - name: {{ $key }}
--- a/bitnami/fluentd/templates/svc-headless.yaml
+++ b/bitnami/fluentd/templates/svc-headless.yaml
@@ -7,8 +7,19 @@ metadata:
     app.kubernetes.io/component: aggregator
     app: aggregator
 spec:
-  type: ClusterIP
+  type: {{ .Values.aggregator.service.type }}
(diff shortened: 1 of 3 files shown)


bitnami/fluentd/templates/svc-headless.yaml
```diff
@@ -6,9 +6,23 @@ metadata:
   labels: {{- include "fluentd.labels" . | nindent 4 }}
     app.kubernetes.io/component: aggregator
     app: aggregator
+  {{- if .Values.aggregator.service.annotations }}
```
andresbono, line 9: Does it make sense to keep considering this a headless service if it may have resource allocation? For instance, if you choose `service.type` `LoadBalancer`.

https://kubernetes.io/docs/concepts/services-networking/service/#headless-services

fsschmitt: Totally agree, that once this is no longer a headless service that the name should change. What's your take on it? 

Would you prefer to keep the headless service (to avoid any breaking change), but if a service of type LoadBalancer was defined, then a new service is deployed apart from the headless?

andresbono: > Would you prefer to keep the headless service (to avoid any breaking change), but if a service of type LoadBalancer was defined, then a new service is deployed apart from the headless?

That approach sounds great, @fsschmitt. Will you update the PR?

fsschmitt: Yes, I'll update this afternoon with this approach. Thanks!

fsschmitt: @andresbono, I've updated with the following:
- Reverted any change on the `svc-headless.yaml` file;
- Added (on the same fashion as the forwarder-svc) the `aggregator-svc.yaml` which enables the same level of configuration on a service of its own;

andresbono:

PR #3250, opened 2020-07-28
[bitnami/wordpress] Document MaintenanceMode limitation & add missing parameters

juan131:
Credits to @lwj5 for bringing this up and provide the solution for liveness probes.

**Description of the change**

This PR documents a known limitation when running the WP with the production values related to the maintenance mode. It also uses a better endpoint for liveness probes so the container is not restarted when performing operations such as updating a plugin.

On the other hand, it adds some common useful parameters that were missing in the chart.

**Possible drawbacks**

None

**Checklist**

### DIFF
--- a/bitnami/wordpress/README.md
+++ b/bitnami/wordpress/README.md
@@ -294,7 +294,7 @@ mariadb.master.persistence.storageClass=nfs
 
 When performing admin operations that require activating the maintenance mode (such as updating a plugin or theme), it's activated in only one replica (see: [bug report](https://core.trac.wordpress.org/ticket/50797)). This implies that WP could be attending requests on other replicas while performing admin operations, with unpredictable consequences.
 
-To avoid that, you can manually activate/deactivate the maintenance mode on every replica using the WP CLI. For instance, if you installed WP with replicas you can run the commands below to activate the maintenance mode (the following example assumes that the release name is `wordpress`):
+To avoid that, you can manually activate/deactivate the maintenance mode on every replica using the WP CLI. For instance, if you installed WP with three replicas, you can run the commands below to activate the maintenance mode in all of them (assuming that the release name is `wordpress`):
 
 ```console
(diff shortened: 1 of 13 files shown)


bitnami/wordpress/README.md
```diff
@@ -277,15 +281,27 @@ This chart includes a `values-production.yaml` file where you can find some para
...
+Then, you can deploy WordPress chart using the proper parameters:
 
 ```console
 persistence.storageClass=nfs
 mariadb.master.persistence.storageClass=nfs
 ```
 
+#### Known limitations
+
+When performing admin operations that require activating the maintenance mode (such as updating a plugin or theme), it's activated in only one replica (this is related to some limitation that doesn't allow us to change the maintenance file path to include it in the NFS volume, find more information on [this ticket](https://core.trac.wordpress.org/ticket/50797)). This implies that WP could be attending requests on other replicas while performing admin operations, with unpredictable consequences.
```
marcosbc, line 295: I don't think we need to explain the background here, this should be fine:

```suggestion
When performing admin operations that require activating the maintenance mode (such as updating a plugin or theme), it's activated in only one replica (see: [bug report](https://core.trac.wordpress.org/ticket/50797)). This implies that WP could be attending requests on other replicas while performing admin operations, with unpredictable consequences.
```

bitnami/wordpress/README.md
```diff
@@ -277,15 +281,27 @@ This chart includes a `values-production.yaml` file where you can find some para
...
+
+To avoid that, you can manually activate/deactivate the maintenance mode on every replica using the WP CLI. For instance, if you installed WP with replicas you can run the commands below to activate the maintenance mode (the following example assumes that the release name is `wordpress`):
```
marcosbc, line 297:
```suggestion
To avoid that, you can manually activate/deactivate the maintenance mode on every replica using the WP CLI. For instance, if you installed WP with three replicas, you can run the commands below to activate the maintenance mode in all of them (assuming that the release name is `wordpress`):
```

PR #3261, opened 2020-07-29
[bitnami/zookeeper] Fix probes when using CentOS images

juan131:

**Description of the change**

The option `-q` is not supported on CentOS netcat, while the option `-w` is shared: 

- Debian:

```console
$ nc -h
[v1.10-41.1]
connect to somewhere:	nc [-options] hostname port[s] [ports] ...
listen for inbound:	nc -l -p port [-options] [hostname] [port]
options:
	...
	-q secs			quit after EOF on stdin and delay of secs
	...
	-w secs			timeout for connects and final net reads
```

- CentOS:

```console
$ nc -h
Ncat 7.50 ( https://nmap.org/ncat )
Usage: ncat [options] [hostname] [port]
  ...
  -w, --wait <time>          Connect timeout
  ...
```

We can obtain a similar effect (quitting after N seconds) using the `timeout` command.

**Benefits**

You can replace default image with a "centos based" one.

**Possible drawbacks**

None

**Checklist**

### DIFF
--- a/bitnami/zookeeper/templates/statefulset.yaml
+++ b/bitnami/zookeeper/templates/statefulset.yaml
@@ -1,4 +1,3 @@
-
 apiVersion: apps/v1
 kind: StatefulSet
 metadata:
--- a/bitnami/zookeeper/templates/statefulset.yaml
+++ b/bitnami/zookeeper/templates/statefulset.yaml
@@ -1,3 +1,4 @@
+
 apiVersion: apps/v1
 kind: StatefulSet
 metadata:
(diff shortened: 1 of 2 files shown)


bitnami/zookeeper/templates/statefulset.yaml
```diff
@@ -247,7 +247,7 @@ spec:
           readinessProbe:
             exec:
               {{- if not .Values.service.tls.disable_base_client_port }}
-              command: ['/bin/bash', '-c', 'echo "ruok" | nc -w {{ .Values.readinessProbe.probeCommandTimeout }} -q {{ .Values.readinessProbe.probeCommandTimeout }} localhost {{ .Values.service.port }} | grep imok']
+              command: ['/bin/bash', '-c', 'echo "ruok" | timeout {{ .Values.livenessProbe.probeCommandTimeout }} nc -w {{ .Values.readinessProbe.probeCommandTimeout }} localhost {{ .Values.service.port }} | grep imok']
```
ddcprg, line 250: apologies for the late comment but it seems this is using the wrong property, shouldn't it use `.Values.readinessProbe.probeCommandTimeout` instead of `.Values.livenessProbe.probeCommandTimeout` ?

dani8art: Hi @ddcprg your right, please check #3321

ddcprg: hi @dani8art thanks for fixing this

PR #3265, opened 2020-07-29
[bitnami/nginx] custom targetPorts for service type Loadbalancer

cw-vikas:
**Description of the change**

Have created the values for the targetPort as there were no values for the parameter, updated the service (svc.yaml) to utilize the user-defined targetPort.

**Benefits**

Upon deploying the chart with the service type as LoadBalancer the service target port is mapped to the provisioned Loadbalancer as "http: http" and "https: https". There's no way to map the targetPorts to http or https  "https [listner] to http[targetport]". Henceforth I was facing the limitations with the targetPort so I tried to resolve it in the PR. I have tested it with the Nginx reverse proxy deployment on the EKS and it works without an issue. Have also deployed it with the helmfile.

**Possible drawbacks**

**Applicable issues**

  - fixes #

**Additional information**

**Checklist** 


joancafom:
Hi @cw-vikas 

Thanks for your PR! I was trying to analyse the changes but I do not fully understand the use case in which this might come handy. `targetPort` is just a named shortcut to refer to a specific port in a deployment, in this case:

https://github.com/bitnami/charts/blob/e176b405c249f6ef2c02b99e51197646665c7ad5/bitnami/nginx/templates/svc.yaml#L23-L26

https://github.com/bitnami/charts/blob/e176b405c249f6ef2c02b99e51197646665c7ad5/bitnami/nginx/templates/deployment.yaml#L70-L72

Instead of using the actual value `{{ .Values.containerPort }} `, we will be using `http` to refer to this port. In case you wanted that both `http` and `https` pointed to the same port, wouldn't it make sense to set the properties `containerPort` and `containerTlsPort` to the same value?

If you are trying to implement a redirection from the standard http site to the https, I think a good idea would be to implement this logic in the application level (by using serverBlock).

Thanks and best regards! 😄

cw-vikas:
> to set the properties `containerPort` and `containerTlsPort` to the same value?
Thanks for your review I have tried to achieve the solution as per your recommendation but it didn't work out for me.

I am using the EKS cluster where I have a preference to use the dynamically generated nodePorts for the services so I don't want to manually define the nodePorts for the services. Hence when I deploy the loadbalancer via service on AWS the target ports are set as http|http, and https|https. 

The issue is when I need https access to the service from the loadbalacer, then, in that case, I can't access it with any of the desired protocols, so as a fix I need to manually update the load balancers Instance port for HTTPS lister to the same port as of the HTTP lister.  And then the loadbalancer gets resolved. If I have the flexibility to set the desired versions of the target ports it will be get done within the deployment.

Screen Shot 2020-08-13 at 20 31 05

If you have some other recommendation, please help me to understand

bitnami/nginx/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: nginx
-version: 6.0.2
+version: 6.0.3
```
juan131, line 3: Please bump minor version instead

bitnami/nginx/README.md
```diff
@@ -87,7 +87,9 @@ The following tables lists the configurable parameters of the NGINX Open Source
...
 | `service.httpsPort`                        | Service HTTPS port                                                                           | `443`                                                        |
 | `service.nodePorts.http`                   | Kubernetes http node port                                                                    | `""`                                                         |
-| `service.nodePorts.https`                  | Kubernetes https node port                                                                   | `""`                                                         |
+| `service.nodePorts.https`                  | Kubernetes https node port
+| `service.targetPort.http`                  | Kubernetes http targetPort                                                                   | See `values.yaml`
+| `service.targetPort.https`                 | Kubernetes https targetPort                                                                  | See `values.yaml`
```
juan131, line 92:
```suggestion
| `service.targetPort.http`                  | Kubernetes http targetPort                                                                   | `http`                   |
| `service.targetPort.https`                 | Kubernetes https targetPort                                                                  | `https`                  |
```

bitnami/nginx/values.yaml
```diff
@@ -185,6 +185,13 @@ service:
     http: ""
     https: ""
 
+  ## Target port reference value for the Loadbalancer service types can be specified explicitly.
+  ## Listeners for the Loadbalancer can be custom mapped to the http or https service.
+  ## Example: Mapping the https listener to targetPort http [http: https]
+  targetPort:
+    http: ""
+    https: ""
```
juan131, line 193: The default values should be `http` and `https`:
```suggestion
    http: http
    https: https
```

juan131 (changes requested):

bitnami/nginx/Chart.yaml
cw-vikas: Thanks for the review 
I have bumped the version to 6.1.0 as a minor version

cw-vikas (commented):
All requested changes are applied

PR #3293, opened 2020-08-01
[bitnami/fluentd] Allow creation/configuration of dedicated ServiceAccount for fluentd aggregator

### LINKED ISSUE #3292: [bitnami/fluentd] Aggregator pods don't use created & declared ServiceAccount
**Which chart**:
Fluentd chart, in aggregator configuration

**Describe the bug**
Even with the default top-level `values.yaml` configuration of
```yaml
serviceAccount:
  create: true
```

The aggregator `StatefulSet` definition does not use the created `ServiceAccount` in its pod spec, and instead uses the `default` service account. Given this is a top level configuration in the chart, this is a bit confusion. It feels like you should be able to use

```yaml
aggregator:
  enabled: true
forwarder:
  enabled: false
serviceAccount:
  create: true
rbac:
  create: false
```
(since I don't think the aggregator has a use case for the RBAC permissions being created by the chart?)

**To Reproduce**
- Install the default chart with `aggregator.enabled=true`
- Look at the `serviceAccount` the aggregator pods are running with

**Expected behavior**
The aggregator pods should be running with the service account created when `serviceAccount.create=true`

- Output of `helm version`:
```
```

```
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.9", GitCommit:"a17149e1a189050796ced469dbd78d380f2ed5ef", GitTreeState:"clean", BuildDate:"2020-04-16T11:44:51Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"15+", GitVersion:"v1.15.11-eks-af3caf", GitCommit:"af3caf6136cd355f467083651cc1010a499f59b1", GitTreeState:"clean", BuildDate:"2020-03-27T21:51:36Z", GoVersion:"go1.12.17", Compiler:"gc", Platform:"linux/amd64"}
```

chadlwilson:
**Description of the change**

- Introduces ability to configure the aggregator `serviceAccount` via `.Values.aggregator.serviceAccount.*`
- Migrates the previous forwarder-only configuration of `.Values.serviceAccount` and `.Values.rbac` to `.Values.forwarder.*` values
- Adds validation errors to historical usage of `.Values.serviceAccount` and `.Values.rbac` to help users migrate to the new configuration points.

**Benefits**

There is currently no way to change the service account used by the aggregator's `StatefulSet`.

Since there is no `PodSecurityPolicy` assigned by this chart, having the aggregator run under a deterministic `ServiceAccount` allows wrapper charts to target a PSP at this `ServiceAccount`, as well as any custom RBAC/`ClusterRoleBinding`s they need.

Follow-up work could involve introducing pod/container `securityContext` and PSPs targeting each component.

**Possible drawbacks**

I couldn't find an elegant way to make this change backward compatible. This is because 
* the previous values of `serviceAccount.create` and `rbac.create` were `true`, making user-intention to disable the creation difficult to detect
* the chart supports `ServiceAccount` `name` and `annotations` customisation. Defaulting all these values and merging the annotation values from the old configuration location got messy, fast and made the chart confusing to configure.

I feel it was better to try and move to a more consistent configuration, as used by [the contour chart](https://github.com/bitnami/charts/tree/master/bitnami/contour) and help users get there by failing the template rendering if using the old configuration values.

**Applicable issues**

  - fixes #3292 

**Checklist**

### DIFF
--- a/bitnami/fluentd/templates/_helpers.tpl
+++ b/bitnami/fluentd/templates/_helpers.tpl
@@ -91,13 +91,24 @@ imagePullSecrets:
 {{- end -}}
 
 {{/*
-Create the name of the service account to use
+Create the name of the forwarder service account to use
 */}}
-{{- define "fluentd.serviceAccountName" -}}
-{{- if .Values.serviceAccount.create -}}
-    {{ default (include "fluentd.fullname" .) .Values.serviceAccount.name }}
+{{- define "fluentd.forwarder.serviceAccountName" -}}
+{{- if .Values.forwarder.serviceAccount.create -}}
+    {{ default (printf "%s-forwarder" (include "fluentd.fullname" .)) .Values.forwarder.serviceAccount.name }}
 {{- else -}}
-    {{ default "default" .Values.serviceAccount.name }}
+    {{ default "default" .Values.forwarder.serviceAccount.name }}
+{{- end -}}
+{{- end -}}
+
+{{/*
+Create the name of the aggregator service account to use
+*/}}
+{{- define "fluentd.aggregator.serviceAccountName" -}}
+{{- if .Values.aggregator.serviceAccount.create -}}
+    {{ default (printf "%s-aggregator" (include "fluentd.fullname" .)) .Values.aggregator.serviceAccount.name }}
+{{- else -}}
+    {{ default "default" .Values.aggregator.serviceAccount.name }}
 {{- end -}}
 {{- end -}}
 
@@ -116,6 +127,7 @@ Validate data
 {{- $messages := list -}}
 {{- $messages := append $messages (include "fluentd.validateValues.deployment" .) -}}
 {{- $messages := append $messages (include "fluentd.validateValues.rbac" .) -}}
+{{- $messages := append $messages (include "fluentd.validateValues.serviceAccount" .) -}}
 {{- $messages := without $messages "" -}}
 {{- $message := join "\n" $messages -}}
  {{- if $message -}}
@@ -134,10 +146,34 @@ fluentd:
 
 {{/* Validate values of Fluentd - must create serviceAccount to create enable RBAC */}}
 {{- define "fluentd.validateValues.rbac" -}}
-{{- if and .Values.rbac.create (not .Values.serviceAccount.create) -}}
+{{- if not (typeIs "<nil>" .Values.rbac.create) -}}
 fluentd: rbac.create
-    A ServiceAccount is required ("rbac.create=true" is set)
-    Please create a ServiceAccount (--set serviceAccount.create=true)
+    Top-level rbac configuration has been removed, as it only applied to the forwarder.
+    Please migrate to forwarder.rbac.create
+{{- end -}}
+{{- if and .Values.forwarder.rbac.create (not .Values.forwarder.serviceAccount.create) }}
+fluentd: forwarder.rbac.create
+    A ServiceAccount is required ("forwarder.rbac.create=true" is set)
+    Please create a ServiceAccount (--set serviceAccount.forwarder.create=true)
+{{- end -}}
+{{- end -}}
+
+{{/* Validate values of Fluentd - prefer per component serviceAccounts to top-level definition */}}
+{{- define "fluentd.validateValues.serviceAccount" -}}
+{{- if not (typeIs "<nil>" .Values.serviceAccount.create) -}}
+fluentd: serviceAccount.create:
+    Top-level serviceAccount configuration has been removed, as it only applied to the forwarder.
+    Please migrate to forwarder.serviceAccount.create
+{{- end -}}
+{{- if .Values.serviceAccount.name }}
+fluentd: serviceAccount.name
+    Top-level serviceAccount configuration has been removed, as it only applied to the forwarder.
+    Please migrate to forwarder.serviceAccount.name
+{{- end -}}
+{{- if .Values.serviceAccount.annotations }}
+fluentd: serviceAccount.annotations
+    Top-level serviceAccount configuration has been removed, as it only applied to the forwarder.
+    Please migrate to forwarder.serviceAccount.annotations
 {{- end -}}
 {{- end -}}
 
--- a/bitnami/fluentd/templates/aggregator-statefulset.yaml
+++ b/bitnami/fluentd/templates/aggregator-statefulset.yaml
@@ -29,6 +29,7 @@ spec:
         {{- end }}
     spec:
       {{- include "fluentd.imagePullSecrets" . | nindent 6 }}
+      serviceAccountName: {{ template "fluentd.aggregator.serviceAccountName" . }}
       {{- if .Values.aggregator.securityContext.enabled }}
       securityContext:
         runAsUser: {{ .Values.aggregator.securityContext.runAsUser }}
--- a/bitnami/fluentd/Chart.yaml
+++ b/bitnami/fluentd/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: fluentd
-version: 1.4.1
+version: 2.0.0
 appVersion: 1.11.2
 description: Fluentd is an open source data collector for unified logging layer
 keywords:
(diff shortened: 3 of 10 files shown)


joancafom:
Hi @chadlwilson ,

Thanks for your PR! To mitigate the possible drawbacks, I was actually thinking of using a new `serviceAccount` specifically targeted to the aggregator. This, you would not expect any permission overlapping and at the same time no two different release would be needed. Does it make sense? We already have some charts implementing multiple serviceAccounts (see [Contour](https://github.com/bitnami/charts/blob/master/bitnami/contour/templates/service-accounts.yaml) for example).

What do you think of it? Thanks!

dani8art:
Hi @chadlwilson did you have time to work on this?

chadlwilson:
Maybe in a few days, yes.

However, how do we come to an agreement on the way forward consistent with the original intention? SInce the original property is `serviceAccount.create` (not `forwarder.serviceAccount.create`) my inference was that the intention was for the serviceAccount to be shared by both components in some way.

It will be confusing unless the old property is deprecated and moved with backward compatibility. Is that what I should do?

dani8art:
Hi @chadlwilson thanks for answering,

I think we should keep both and use `serviceAccount.create` as default value. I mean, if I want to create a serviceAccount for all the component I will set just `serviceAccount.create=true` but if I want to only create it for forwarder I will set `serviceAccount.create=false,forwarder.serviceAccount.create=true`. The other way around we would need to set `...create=false` for every component except the forwarder.

Does it make sense?

chadlwilson:
Hi @joancafom and @dani8art - I've updated the PR to create a separate `ServiceAccount` for the `aggregator` in line with the Contour chart - also updated the PR description to reflect the current state of it.

Note that it's not fully backward compatible - it got too messy, and I felt it was better to move forward towards consistency. I bumped the Chart version to `1.4` as it's not a fully breaking change (since users overriding the defaults will be warned via validation and can adjust their values overrides to get exactly the same behaviour as before).

Let me know what you think.

bitnami/fluentd/templates/aggregator-statefulset.yaml
```diff
@@ -29,6 +29,7 @@ spec:
         {{- end }}
     spec:
       {{- include "fluentd.imagePullSecrets" . | nindent 6 }}
+      serviceAccountName: {{ template "fluentd.aggregator.serviceAccountName" . }}
```
marcosbc, line 32: Shouldn't this property be added conditionally, depending on whether service accounts are enabled or not for the aggregator?

marcosbc (commented):
Thanks for the contribution! I've added a comment regarding a small concern, but I couldn't find anything problematic with the rest of the changes.

bitnami/fluentd/templates/service-accounts.yaml
```diff
@@ -0,0 +1,24 @@
+{{- if and .Values.forwarder.serviceAccount.create .Values.forwarder.enabled }}
```
marcosbc, line 1: Another thing is that we use the suffix `serviceaccount` / `serviceaccounts`. For consistency with the other charts, could you also rename it?

bitnami/fluentd/templates/aggregator-statefulset.yaml
chadlwilson: Hi @marcosbc - this is following the same style as  the forwarder does - if the serviceAccounts are not created by the chart or the `serviceAccount.name` is not overridden, the value [off the template here](https://github.com/bitnami/charts/pull/3293/files#diff-f184e19aae77c89089cfe28594781b45R111) will be `default` which is the same as what pods will get out of the box if you don't set the value in the pod spec.

bitnami/fluentd/templates/service-accounts.yaml
chadlwilson: Could you clarify if you're talking about the template file name? 

If so, here I was just following the pattern of the `contour` chart which was suggested as an example and uses this name. But I agree consistency would be better, if we can find the pattern to follow.

However I do see that charts like `elasticsearch` have multiple service accounts in a template named `serviceaccount.yaml` (singular). There don't seem to be any with `serviceaccounts.yaml` (plural, no hyphen) as the name.

bitnami/fluentd/templates/aggregator-statefulset.yaml
chadlwilson: In any case, I have reverted to the original name of `serviceaccount.yaml` as this seems the most consistent with the most number of charts.

bitnami/fluentd/templates/service-accounts.yaml
chadlwilson: In any case, I have reverted to the original name of `serviceaccount.yaml` as this seems the most consistent with the most number of charts.

bitnami/fluentd/templates/aggregator-statefulset.yaml
dani8art: I think this is ok thanks!

dani8art:
Please, could you fix these conflicts with the master branch?

chadlwilson:
> Please, could you fix these conflicts with the master branch?

Done, thx for the reminder.

bitnami/fluentd/templates/_helpers.tpl
```diff
@@ -134,10 +146,34 @@ fluentd:
...
-    A ServiceAccount is required ("rbac.create=true" is set)
-    Please create a ServiceAccount (--set serviceAccount.create=true)
+    Top-level rbac configuration has been removed, as it only applied to the forwarder.
+    Please migrate to forwarder.rbac.create
+{{- end -}}
+{{- if and .Values.forwarder.rbac.create (not .Values.forwarder.serviceAccount.create) -}}
+fluentd: forwarder.rbac.create
+    A ServiceAccount is required ("forwarder.rbac.create=true" is set)
+    Please create a ServiceAccount (--set serviceAccount.forwarder.create=true)
+{{- end -}}
```
dani8art, line 158: We should add the same validation for the aggregator, shouldn't we?

dani8art (changes requested):
Hi, we were thinking about these changes, and finally, we should go for a major change even if we advise throw validation helpers we need to mark this as a braking changing since it can break the way some people are performing their upgrades. 

Please, bump the version properly and add some line explaining how to upgrade to the new version without issues in the upgrading section in the readme, something like [this](https://github.com/bitnami/charts/tree/master/bitnami/ghost#upgrading)

bitnami/fluentd/templates/_helpers.tpl
chadlwilson: The `aggregator` doesn't need/have any RBAC - since it didn't have a serviceaccount prior to this change there was no way for the aggregator to have a special `Role`/`ClusterRole`.

The `forwarder` RBAC allows the forwarder to augment the container logs from a node with kubernetes-oriented metadata (namespace, node, node-etc) via the `fluent-plugin-kubernetes_metadata_filter` which doesn't seem like a relevant use case for the aggregator?

chadlwilson:
> Hi, we were thinking about these changes, and finally, we should go for a major change even if we advise throw validation helpers we need to mark this as a braking changing since it can break the way some people are performing their upgrades.

Fair enough.

That's kind-of why I originally submitted the simpler approach to share the service account with the `forwarder`. There is no way to introduce a dedicated service account for the aggregator that won't be confusing since the original `values` structure used for the forwarder was not really correctly modelled (since it was at top level) and made the defaulting of `.create`, `.name` and `.annotations` very messy/complex to get correct/consistent across all the places that refer to them. 

> Please, bump the version properly and add some line explaining how to upgrade to the new version without issues in the upgrading section in the readme, something like [this](https://github.com/bitnami/charts/tree/master/bitnami/ghost#upgrading)

Will do.

dani8art (approved):
LGTM! thank you so much.

We are going to pass some internal tests to verify all is working properly, after that we will merge it.

PR #3299, opened 2020-08-03
[bitnami/prometheus-operator] Revert "Adjust thanos-sidecar for prom-op no longer using mount subpaths (#3122)"

hwoarang:
This reverts commit ad72b27091f2b7ee29e22c236463c6a8a1b4e6b7.

The change is breaking Thanos due to disableMountSubPath option not
existing to the Prometheus CRD

```
Error: unable to build kubernetes objects from release manifest: error validating "": error validating data:
ValidationError(Prometheus.spec.storage): unknown field "disableMountSubPath" in com.coreos.monitoring.v1.Prometheus.spec.storage
```

Which is actually true since the CRD installed by Bitnami does not have
this option in the spec

https://github.com/bitnami/charts/blob/master/bitnami/prometheus-operator/crds/crd-prometheus.yaml

As such there is no way to set this option to the values.yaml file

However, the message in the original commit is rather confusing as the
option hasn't changed its default value upstream

https://github.com/coreos/prometheus-operator/blob/release-0.40/pkg/prometheus/statefulset.go#L947

The operator will still return 'prometheus-db' subpath by default. Also
this option was not documented at all in the README file so it's not
clear what its intention is.

Having said that, not being able to override it in the values.yaml file,
means that Thanos sidecar is unable to mount the subpath no matter what
and there is no straight way to recover this apart from editing the live
objects.

As such, I suggest we revert this in order to restore functionality as
it was until it's clear what problem we are trying to solve.

**Checklist** 

### DIFF
--- a/bitnami/prometheus-operator/Chart.yaml
+++ b/bitnami/prometheus-operator/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v1
 appVersion: 0.41.0
 description: The Prometheus Operator for Kubernetes provides easy monitoring definitions for Kubernetes services and deployment and management of Prometheus instances.
 name: prometheus-operator
-version: 0.27.1
+version: 0.27.2
 keywords:
   - prometheus
   - alertmanager
--- a/bitnami/prometheus-operator/templates/prometheus/prometheus.yaml
+++ b/bitnami/prometheus-operator/templates/prometheus/prometheus.yaml
@@ -220,9 +220,7 @@ spec:
       volumeMounts:
         - mountPath: /prometheus
           name: prometheus-{{ template "prometheus-operator.prometheus.fullname" . }}-db
-          {{- if not (.Values.prometheus.storageSpec.disableMountSubPath | default true) }}
           subPath: prometheus-db
-          {{- end }}
     {{- end }}
     {{- if .Values.prometheus.containers }}
     {{- include "prometheus-operator.tplValue" (dict "value" .Values.prometheus.containers "context" $) | nindent 4 }}


hwoarang:
@dani8art @teozkr could you please look into this as this is breaking existing deployments by default without an easy way to recover?

teozkr:
> The operator will still return 'prometheus-db' subpath by default. Also
this option was not documented at all in the README file so it's not
clear what its intention is.

Not for new deployments, at least. Maybe that's the crux, maybe it saves which case was used at install-time somewhere?

hwoarang:
@teozkr Prometheus operator 0.40 still returns 'prometheus-db' subpath by default. Why do you think the default option has changed? Could you point me to it? Maybe we need to sync the bitnamid CRDs with the upstream 0.40 ones but right now we are in a mixed state. I suggest we introduce the disableSubMountPath when the CRDs are updated otherwise now existing deployments are broken after an update.

teozkr:
> Why do you think the default option has changed?

Empirical testing when creating the cluster. But I agree that I can't find anywhere that should actually have such an effect in the operator..

hwoarang:
That's strange because we also deploy it on clean clusters and the subpath is mounted like it always used it with the default options :/ This is how we caught the spec violation I posted in the PR message.

I think it makes sense to revert now and bring it back along with a full CRD update from 0.40 (or 0.41) that was released as that option would appear in the CRD spec and we would have to add it anyway.

teozkr:
Trying to repro it now.

hwoarang:
This is how we define our storage (not using the storage spec directly) in the chart

```
    # persistence, Prometheus persistence parameters
    persistence:
      enabled: true
      accessModes:
        - ReadWriteOnce
      size: 500Gi
```

teozkr:
I just tried a clean install of ad72b27091f2b7ee29e22c236463c6a8a1b4e6b7 into Kind (k8s 1.18.6) and got [this STS](https://gist.github.com/teozkr/d0e4b4b6f5ec489c5cd7445e5d92cb5c), no `subPath` in sight.

teozkr:
Oh, that seems to be the difference, heh. So the subpath is only enabled by default when persistence is enabled.

hwoarang:
So are you also able to see the subpath when you enable persistency in your case?

teozkr:
Yes. Patch on the way that should cover both cases.

PR #3310, opened 2020-08-03
[bitnami/consul] improvement consul prometheus probe

ysicing:
**Description of the change**

1. Prometheus Probe is usually enabled on service.
2. Bump consul & consul-exporter major version

**Benefits**

**Possible drawbacks**

**Applicable issues**

  - fixes #

**Additional information**

**Checklist** 


joancafom:
Hi @ysicing 

Thanks for your PR! The Bitnami Consul Chart uses [prometheus/consul_exported](https://github.com/prometheus/consul_exporter) to scrape metrics for Prometheus, and it seems to me that it is correctly configured in order to behave properly:

https://github.com/bitnami/charts/blob/c3fec265e642e3273993323a02b0ae68cba9d7ae/bitnami/consul/templates/statefulset.yaml#L190-L192

Have I misconstrued something?

ysicing:
@joancafom There is no problem with the configuration. I just adjusted the probe position. default probe is configured in StatefulSet, in most cases we default to automatic discovery from service.Or for what reason is it configured in StatefulSet ?

bitnami/consul/templates/consul-service.yaml
```diff
@@ -26,5 +29,10 @@ spec:
     - name: consuldns-udp
       protocol: "UDP"
       port: {{.Values.service.consulDnsPort}}
+{{- if .Values.metrics.enabled }}
+    - name: metrics
+      port: 9104
```
joancafom, line 34: Adding the property `.Values.metrics.service.` would also allow to provide with a parameter named `port` to customize it. Nonetheless, the default `podAnnotations.prometheus.io/port` is set to `9107` instead of `9104`, so be careful.

bitnami/consul/templates/statefulset.yaml
```diff
@@ -18,14 +18,11 @@ spec:
   template:
     metadata:
       labels: {{- include "consul.labels" . | nindent 8 }}
-      {{- if or .Values.podAnnotations (and .Values.metrics.enabled .Values.metrics.podAnnotations) }}
+      {{- if .Values.podAnnotations }}
```
joancafom, line 21: Finally, I would leave this as it is and we can cover both use cases. What do you think?

ysicing: ok. You are right, I will update it later.

PR #3021, opened 2020-07-03
[bitnami/postgresql] Helm label best practices

denniseffing:
**Description of the change**

This PR replaces the labels `app=`, `chart=`, `release=` and `heritage=` with their new counterparts (`app.kubernetes.io/name`, `helm.sh/chart`, `app.kubernetes.io/instance`, `app.kubernetes.io/managed-by`) according to the [Helm label best practices](https://helm.sh/docs/chart_best_practices/labels/).

Please let me know whether this change qualifies as a new major, minor or patch version (as this change may break existing queries based on old labels, see possible drawbacks).

**Benefits**

Please take a look at the related problem mentioned in #2889 

**Possible drawbacks**

Customers who implemented search queries based on existing labels must update those queries after deploying this change.

**Applicable issues**

  - fixes #2889 

**Checklist** 
- [ ] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [X] Title of the PR starts with chart name (e.g. `[bitnami/chart]`)

PR #3322, opened 2020-08-04
[bitnami/postgresql] Add upgrade notes for 9.X.X

carrodher:
Add upgrade notes for the major release in https://github.com/bitnami/charts/pull/3021

I am not bumping the chart version as those changes are going to be reflected in the README and we don't need to release a new version for that.

### DIFF
--- a/bitnami/postgresql/README.md
+++ b/bitnami/postgresql/README.md
@@ -536,9 +536,12 @@ postgresql	default  	3       	2020-08-04 13:42:08.020385884 +0000 UTC	deployed	p
 
 - We can kill the existing pod and the new statefulset is going to create a new one:
 ```console
+$ kubectl delete pod postgresql-postgresql-0
+pod "postgresql-postgresql-0" deleted
+
 $ kubectl get pods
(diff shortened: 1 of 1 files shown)


bitnami/postgresql/README.md
```diff
@@ -492,6 +492,57 @@ $ helm upgrade my-release stable/postgresql \
...
+$ helm ls
+NAME      	NAMESPACE	REVISION	UPDATED                                	STATUS  	CHART           	APP VERSION
+postgresql	default  	3       	2020-08-04 13:42:08.020385884 +0000 UTC	deployed	postgresql-9.1.2	11.8.0
+```
+
+- We can kill the existing pod and the new statefulset is going to create a new one:
+```console
+$ kubectl get pods
+NAME                      READY   STATUS    RESTARTS   AGE
+postgresql-postgresql-0   1/1     Running   0          102s
```
miguelaeh, line 541: I think this pod is the same as before regardng the timestamp, I would add the kill command too.

PR #3323, opened 2020-08-04
[bitnami/thanos] Fix - Move ruler rules to separate folder

Whyeasy:
**Description of the change**

GKE doesn't play nicely with `subPath` in VolumeMounts. Needed to move the `ruler.yml` to a separate folder inside of the conf folder so it wouldn't have issues with the objstore-config. This way I could delete the `subPath` from the deployment.

Error that was generated when deploying current chart/pod:

```shell
OCI runtime create failed: container_linux.go:349: starting container process caused "process_linux.go:449: 
container init caused \"rootfs_linux.go:58: mounting \\\
"/var/lib/kubelet/pods/281a145a-f287-4afa-804d-044855ce9e87/volume-subpaths/ruler-config/ruler/0
\\\" to rootfs 
\\\"/var/lib/docker/overlay2/3579a2bf3b39d7b13f9a4c0c885e38e1d8c07a86c746d061fabe9680befee096/merged
\\\" at \\\"/var/lib/docker/overlay2/3579a2bf3b39d7b13f9a4c0c885e38e1d8c07a86c746d061fabe9680befee096/merged/conf/ruler.yml
\\\" caused 
\\\"not a directory
\\\"\"": unknown: Are you trying to mount a directory onto a file (or vice-versa)? Check if the specified host path exists and is the expected type
```

Removed subPath in all deployment to reduce any further issues.

**Benefits**

Fixed the helm chart for GKE cluster deployments.

**Possible drawbacks**

Might break on other clouds? Don't think it will.

**Applicable issues**

**Additional information**

**Checklist** 
- [x ] Variables are documented in the README.md
- [ x] Title of the PR starts with chart name (e.g. `[bitnami/chart]`)
- [ x] If the chart contains a `values-production.yaml` apart from `values.yaml`, ensure that you implement the changes in both files

### DIFF
--- a/bitnami/thanos/Chart.yaml
+++ b/bitnami/thanos/Chart.yaml
@@ -1,5 +1,5 @@
 apiVersion: v1
-version: 2.0.0
+version: 2.0.1
 appVersion: 0.14.0
 description: Thanos is a highly available metrics system that can be added on top of existing Prometheus deployments, providing a global query view across all Prometheus installations.
 engine: gotpl
--- a/bitnami/thanos/templates/bucketweb/deployment.yaml
+++ b/bitnami/thanos/templates/bucketweb/deployment.yaml
@@ -83,7 +83,6 @@ spec:
               mountPath: /conf
               {{- else }}
               mountPath: /conf/objstore.yml
-              subPath: objstore.yml
               {{- end }}
       volumes:
         - name: objstore-config
--- a/bitnami/thanos/templates/compactor/deployment.yaml
+++ b/bitnami/thanos/templates/compactor/deployment.yaml
@@ -101,7 +101,6 @@ spec:
               mountPath: /conf
               {{- else }}
               mountPath: /conf/objstore.yml
-              subPath: objstore.yml
               {{- end }}
             - name: data
               mountPath: /data
--- a/bitnami/thanos/templates/querier/deployment.yaml
+++ b/bitnami/thanos/templates/querier/deployment.yaml
@@ -110,7 +110,6 @@ spec:
       {{- if or (include "thanos.querier.createSDConfigmap" .) .Values.querier.existingSDConfigmap }}
             - name: sd-config
               mountPath: /conf/servicediscovery.yml
-              subPath: servicediscovery.yml
       {{- end }}
       {{- if .Values.querier.grpcTLS.server.secure }}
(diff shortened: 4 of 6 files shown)


carrodher:
Hi, thanks for creating this PR.

I am not sure if this is a common issue in GKE or something specific in your use case. We are testing the bitnami charts in different clouds like GKE, AKS, TKG, IKS, etc and we didn't face this issue.

Can you provide more information about your use case?:
- What Helm/k8s version are you using?
- Are you able to reproduce the issue with the default parameters or are you modifying the values.yaml or adding some parameters during the installation with `--set`?
- Is the issue appearing in a fresh installation or during an upgrade?

Thanks!

Whyeasy:
Hi,

I'm using `helm3`, K8S master version is `GitVersion:"v1.16.11-gke.5`.

I'm running this chart for quite a while already and had no issues, but I need to enable the ruler now. Which was first disabled. So it's a upgrade, but not upgrading thanos ruler.

I tried to use `ruler.config` and `ruler.existingConfigmap`, both gave the same error. I provide additional values via a extra `values.yaml`. I can't use the default values for the ruler, as it needs atleast one value for providing rules. I didn't try this with a clean chart.

Editing the statefulset once it's deployed with the changes within this PR, it starts and runs fine.

The `objectstore-cfg` I provide with via `existingObjstoreSecret` which doesn't have the `subPath` defined either in the chart.

I use the following components from this chart and these are running fine:
- querier
- bucketweb
- storegateway
- compactor

carrodher:
I am trying to reproduce the issue on my side but I am not facing the error. The subpath should be needed to mount different things in the same volume, I am not sure if removing it we are going to be able to mount the volumes in the different use cases.

Are you using `objstore.yml` as the path for your secret?

Whyeasy:
Indeed it's needed. That's why I moved the rules file to a different folder which fixed it for me. All other components don't have a mount in the same folder. So that's why I removed them as well in the other templates 😄 These were not giving any problems, just out of consistency. 

Yes, I use `objstore.yml` as the path of the secret.

carrodher:
That's good, I just tested it in our internal pipeline and it works with the parameters we are using for testing; apart from that, I did some manual test with other parameters. Thanks for your contribution and the detailed explanations

PR #3371, opened 2020-08-09
[bitnami/postgresql-ha] Fix postgresql-ha chart prometheus annotations

RashadAnsari:
In `postgresql-ha` chart when we want to use the Prometheus exporter, The annotations that we have In values file for exporter don't set.
This PR is for fixing this issue.

### DIFF
--- a/bitnami/postgresql-ha/templates/postgresql/statefulset.yaml
+++ b/bitnami/postgresql-ha/templates/postgresql/statefulset.yaml
@@ -25,8 +25,14 @@ spec:
         {{- if .Values.postgresql.podLabels }}
         {{- include "postgresql-ha.tplValue" (dict "value" .Values.postgresql.podLabels "context" $) | nindent 8 }}
         {{- end }}
+      {{- if or .Values.postgresql.podAnnotations (and .Values.metrics.enabled .Values.metrics.annotations) }}
+      annotations:
       {{- if .Values.postgresql.podAnnotations }}
-      annotations: {{- include "postgresql-ha.tplValue" (dict "value" .Values.postgresql.podAnnotations "context" $) | nindent 8 }}
+      {{- include "postgresql-ha.tplValue" (dict "value" .Values.postgresql.podAnnotations "context" $) | nindent 8 }}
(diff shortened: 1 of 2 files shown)


bitnami/postgresql-ha/templates/postgresql/statefulset.yaml
```diff
@@ -25,8 +25,12 @@ spec:
         {{- if .Values.postgresql.podLabels }}
         {{- include "postgresql-ha.tplValue" (dict "value" .Values.postgresql.podLabels "context" $) | nindent 8 }}
         {{- end }}
+      annotations:
```
carrodher, line 28: This part is going to be added to the deployment even if the podAnnotations and metrics.annotations are empty, can you include this part inside a condition? Something like
```
{{- if or .Values.postgresql.podAnnotations .Values.metrics.annotations }}
annotations:
...
```

RashadAnsari: I test this feature with Kubernetes v1.10.0 and Helm v3.0.1 and if we don't have any `postgresql.podAnnotations` and `metrics.annotations` we don't have any annotations in our Kubernetes deployments.
Do you want something like this or something else?

carrodher: For example, running `helm template`, in the statefulset we can see an empty `annotations` block:
```yaml
# Source: postgresql-ha/templates/postgresql/statefulset.yaml
apiVersion: apps/v1
kind: StatefulSet
...
  template:
    metadata:
      labels:
        app.kubernetes.io/name: postgresql-ha
        helm.sh/chart: postgresql-ha-3.5.4
        app.kubernetes.io/instance: RELEASE-NAME
        app.kubernetes.io/component: postgresql
      annotations:
    spec:
      securityContext:
        fsGroup: 1001
...
```
with the suggested approach, this section shouldn't appear when the annotations are not set

RashadAnsari: Done.

PR #3367, opened 2020-08-08
[bitnami/contour] Add ability to specify ingressClass and restart pods on config change

mkilchhofer:
**Description of the change**
As mentioned in the documentation on [projectcontour.io](https://projectcontour.io/docs/v1.7.0/annotations/), Contour supports watching a specific ingressClass only.

**Benefits**
- Also as mentioned in the docs above:
  > This can be useful while you are migrating from another controller, or if you need multiple instances of Contour.

- Contour pods are automatically restarted after we need to change `configInline`
- While I was already editing this chart I updated some missing settings in `README.md`

**Possible drawbacks**
`-`

**Applicable issues**
`-`

  - fixes #

**Additional information**
`-`

**Checklist** 
- [x] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [x] Variables are documented in the README.md
- [x] Title of the PR starts with chart name (e.g. `[bitnami/chart]`)
- [x] If the chart contains a `values-production.yaml` apart from `values.yaml`, ensure that you implement the changes in both files

:warning: Keep in mind that if you want to make changes to the kubeapps chart, please implement them in the [kubeapps repository](https://github.com/kubeapps/kubeapps/tree/master/chart/kubeapps). This is only a synchronized mirror.

PR #2961, opened 2020-06-29
[bitnami/chart] Add NLB load balancer support, run envoy container as root

geota:
**Description of the change**

Support NLB load balancers for contour and fix crash loop in envoy container due to insufficient permissions. 

**Benefits**

NLB have a number of benefits over ELBs. Including support service endpoints, no need to pre-warm, higher throughput etc.

https://docs.aws.amazon.com/elasticloadbalancing/latest/network/introduction.html

**Possible drawbacks**

NLB support is implemented in a backward-compatible way and defaults to the existing behavior of using a Classic ELB.

Defaulting the envoy container to run as root maybe is unacceptable. If so, I am happy to remove defaulting that and just support passing it in.  

**Applicable issues**

  - fixes #2960
  - fixes #2968

**Additional information**

Relevant docs:
https://projectcontour.io/guides/deploy-aws-nlb/
https://github.com/projectcontour/contour/tree/master/examples/contour#deploying-with-host-networking-enabled-for-envoy

- [X] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [X] Variables are documented in the README.md
- [X] Title of the PR starts with chart name (e.g. `[bitnami/chart]`)
- [X] If the chart contains a `values-production.yaml` apart from `values.yaml`, ensure that you implement the changes in both files

PR #3381, opened 2020-08-10
[bitnami/contour] Sync upstream changes and chart standardization

mkilchhofer:
**Description of the change**
The most important change is the refactoring of the shutdown sidecar:
https://github.com/projectcontour/contour/commit/7cd9f4a6857ba3ca728f27f2b2605629961e9a54

I manually diff'ed the example manifests here: https://github.com/projectcontour/contour/tree/master/examples/contour

**Benefits**
By using this configuration in combination with contour version `1.7.0` the envoys now properly entering into the `DRAINING` state when the pods needs to be terminated.

Additionally this change fixes the failing pipeline issue.

**Possible drawbacks**
`-`

**Applicable issues**
`-`

**Additional information**
`-`

**Checklist** 


bitnami/contour/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: contour
 description: Contour Ingress controller for Kubernetes
-version: 1.2.0
+version: 1.2.1
```
dani8art, line 4: it would be a minor at least. Don't you think so?

bitnami/contour/README.md
```diff
@@ -208,6 +214,18 @@ configInline:
...
+  # - "HTTP/2"
+  # - "HTTP/1.1"
+  #
+  # The following shows the default proxy timeout settings.
+  # timeouts:
+  #   request-timeout: infinity
+  #   connection-idle-timeout: 60s
+  #   stream-idle-timeout: 5m
+  #   max-connection-duration: infinity
+  #   connection-shutdown-grace-period: 5s
```
dani8art, line 228: why do you need those values?

bitnami/contour/templates/daemonset.yaml
```diff
@@ -150,6 +151,7 @@ spec:
         - /config/envoy.json
         - --xds-address={{ template "contour.fullname" . }}
         - --xds-port=8001
+        - --resources-dir=/config/resources
```
dani8art, line 154: does this file exist? how are you creating it?

bitnami/contour/values-production.yaml
```diff
@@ -47,6 +47,12 @@ configInline:
   tls:
   #   minimum TLS version that Contour will negotiate
   #   minimum-protocol-version: "1.1"
+  # Defines the Kubernetes name/namespace matching a secret to use
```
dani8art, line 50: same here

bitnami/contour/values.yaml
```diff
@@ -46,6 +46,12 @@ configInline:
   tls:
   #   minimum TLS version that Contour will negotiate
   #   minimum-protocol-version: "1.1"
+  # Defines the Kubernetes name/namespace matching a secret to use
```
dani8art, line 49: same with values here

bitnami/contour/README.md
mkilchhofer: The chart was merged into the bitnami repo with appVersion `1.5.1`. These settings in the `README.md`, `values.yaml`  and `values-production.yaml` were copied from the upstream kubernetes manifests you find here:
https://github.com/projectcontour/contour/blob/master/examples/contour/01-contour-config.yaml

Is it a bad idea to sync back the development process from the guys at projectcontour.io?

bitnami/contour/templates/daemonset.yaml
mkilchhofer: I tested it like this.
`/config` is an emtpyDir and the config inside here is coming from the initcontainer I think.

```console
root@projectcontour-internal-envoy-vfvth:/# ls -la /config/resources/sds/
total 4
drwxr-x--- 2 1001 root 1024 Aug 10 20:24 .
drwxr-x--- 3 1001 root 1024 Aug 10 20:24 ..
-rw-r--r-- 1 1001 root  186 Aug 10 20:24 xds-tls-certicate.json
-rw-r--r-- 1 1001 root  185 Aug 10 20:24 xds-validation-context.json

root@projectcontour-internal-envoy-vfvth:/# mount |grep \/config
/dev/sda1 on /config type ext4 (rw,relatime)
```

bitnami/contour/values-production.yaml
mkilchhofer: See comment above (https://github.com/bitnami/charts/pull/3381#discussion_r468520032)

bitnami/contour/values.yaml
mkilchhofer: See comment above (https://github.com/bitnami/charts/pull/3381#discussion_r468520032)

bitnami/contour/Chart.yaml
mkilchhofer: Mhhh I am not sure. To be 100% correct, it is a bugfix of the automatical app update from `1.6.x` to `1.7.x` of your bot :smiley: 
https://github.com/bitnami/charts/commit/04732989c72fb1a03aaed8798d0ae24345190475#diff-8d26fcd3a47b928f0e5ca90b7bc1effcR5

bitnami/contour/README.md
dani8art: Ok

mkilchhofer:
> Please, could you fix the errors when deploying the chart? it seems it cannot execute the readiness probe well or doesn't get the proper response.

Hi @dani8art, I would love to fix them, but it is not possible. Please read what your colleague @marcosbc wrote here: https://github.com/bitnami/charts/pull/3367
> As for the tests, we recently enabled PR chart testing so some failures like these are expected until we polish things up. Note that we run more reliable tests in our internal pipeline, and in the case of Contour they're currently succeeding.

.github/workflows/lint-test.yaml
```diff
@@ -27,10 +27,10 @@ jobs:
         run: git fetch --prune --unshallow
 
       - name: Create kind cluster
-        uses: helm/kind-action@v1.0.0-rc.1
+        uses: helm/kind-action@v1.0.0
```
mkilchhofer, line 30: Version `v1.0.0-rc.1` of the kind-action uses kind 1.17.0. When I try to launch a cluster with 1.17.0 on my notebook, coredns is going to CrashLoopBackOff. Therefore there is no valid DNS server in k8s which would help envoy to discover the contour service.

Maybe the real problem for failling pipelines was only the change here:
https://github.com/bitnami/charts/pull/3381/commits/8bab0b3db8b4b3285d4a1d0c3496dfc71edd102c

Should I revert to the RC actions?

mkilchhofer:
Finally the tests are green :green_circle: :100: :partying_face:

.github/workflows/lint-test.yaml
dani8art: yes, please could you revert to RC versions? We would like to manage this in a different PR, please.

mkilchhofer: Done and the tests still are green :partying_face:

mkilchhofer:
Hi @dani8art 
Are my changes ok now? :)

dani8art:
Hi @mkilchhofer we went deeper on these changes and we saw you are introducing and modify CRDs. We would like to consider these changes as a major version and take this opportunity to include some of our standardization and recent common features.

Even though a `helm upgrade` works fine, it will remove a CRD and will update the rest, so instances that are deployed using previous CRDs will not work properly after an upgrade. 

We know that adding those standardization and common features may be out of scope for you but I can help you to manage this. 

We would like to include the next list of features.

- [ ] add `bitnami/common` as a dependency, please find it [here](https://github.com/bitnami/charts/tree/master/bitnami/common) you can also find an example for all the helpers in the common chart [here](https://github.com/bitnami/charts/tree/master/bitnami/spring-cloud-dataflow)
- [ ] `extraVolumes`, `extraVolumeMounts` both contour and envoy, please find an example [here](https://github.com/bitnami/charts/blob/master/bitnami/ghost/values.yaml#L349)
- [ ] `extraEnvVar` both contour and envoy, please find an example [here](https://github.com/bitnami/charts/blob/master/bitnami/ghost/values.yaml#L361)
- [ ] `extraEnvVarsConfigMap` both contour and envoy, please find an example [here](https://github.com/bitnami/charts/blob/master/bitnami/ghost/values.yaml#L365)
- [ ] `extraEnvVarsSecret` both contour and envoy, please find an example [here](https://github.com/bitnami/charts/blob/a8d24643756470d0280fc585b01397358c1c242d/bitnami/ghost/values.yaml#L369)
- [ ] `initContainers` both contour and envoy, please find an example [here](https://github.com/bitnami/charts/blob/a8d24643756470d0280fc585b01397358c1c242d/bitnami/ghost/values.yaml#L345)
- [ ] `service.extraPorts` [here](https://github.com/bitnami/charts/blob/master/bitnami/ghost/values.yaml#L205)

mkilchhofer:
Hi @dani8art 
Sure, I understand your point and I can work on it. The changeset would be even bigger then.

I only want to mention that the guys behind projectcontour (mainly the heptio.com people which are now part of VMware) talks about using the bitnami chart and already prepared their changes:
https://github.com/projectcontour/contour/issues/2050

Do we still need to support helm2 for all the bitnami charts?

dani8art:
Hi!! thank you so much for being handling that.

> Do we still need to support helm2 for all the bitnami charts?

Yes, we need it by now.

bitnami/contour/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v1
 name: contour
 description: Contour Ingress controller for Kubernetes
-version: 1.2.0
+version: 1.3.0
```
juan131, line 4: You should be bumping the major version, right? I mean, if I enable a feature included in the new CRDs during a helm upgrade it will fail complaining that the feature doesn't exist in the previously installed CRDs.

juan131 (changes requested):
Thanks so much for implementing all the suggestions that @dani8art shared with you!! Your PR looks awesome. Just a minor comment

bitnami/contour/Chart.yaml
mkilchhofer: Oh yes I wanted to discuss about that. I've no problem with that. But I think this is not a breaking change. The chart itself does not apply resources which belongs to the CRDs. The CRDs are only used by other applications to have more features than the generic and built-in Ingress-v1 API.

The guys behind projectcontour.io only extend their CRDs but do not remove fields since they went GA with 1.0.0 in november 2019.

Shall I bump it?
Please also read through  this comment here: https://github.com/bitnami/charts/pull/3381#issuecomment-678777617. Eg. when upgrading from chart version `1.1.2` to `1.1.3` (https://github.com/bitnami/charts/commit/04732989c72fb1a03aaed8798d0ae24345190475#diff-b04dc01907aa7a4911696009fecec82f), the upgrade will fail.

juan131: > Oh yes I wanted to discuss about that. I've no problem with that. But I think this is not a breaking change. The chart itself does not apply resources which belongs to the CRDs. The CRDs are only used by other applications to have more features than the generic and built-in Ingress-v1 API.

Oh... You're right.. In that case the `helm upgrade` won't fail.

> Please also read through this comment here: #3381 (comment). Eg. when upgrading from chart version 1.1.2 to 1.1.3 (0473298#diff-b04dc01907aa7a4911696009fecec82f), the upgrade will fail.

Please correct me if I'm wrong but this job should be run just once, shouldn't it? I mean, why should I re-generate the certificates during a helm upgrade? In that case, I guess we need to use helm hooks to ensure the job is only executed during the installation (skipping upgrades). WDYT?

mkilchhofer: The job generates certifcates which are valid for 365days. I am not sure if `envoy` and `contour` hot-reloads the mounted certificates at runtime when the job finishes (and thus re-generation is done).

I propose that we should also use the `latest` image in the JobSpec to ensure chart upgrade won't fail unless we know exactly how to rotate the certificates "on demand".

juan131: > I propose that we should also use the latest image in the JobSpec to ensure chart upgrade won't fail unless we know exactly how to rotate the certificates "on demand".

Then, we need to create a specific image for the job. I mean in the values.yaml, we should have sth like:

```yaml
job: 
  image:
    registry: docker.io
    repository: bitnami/contour
    tag: latest
```

The thing is.. We are currently reusing the image taken from:

```yaml
contour:
  image:
    registry: docker.io
    repository: bitnami/contour
    tag: 1.7.0-debian-10-r0
```

That image (used for the main Contour container) **must use immutable tags**, the reasons can be found on this link:
- https://docs.bitnami.com/tutorials/understand-rolling-tags-containers/

> I am not sure if envoy and contour hot-reloads the mounted certificates at runtime when the job finishes (and thus re-generation is done)

I really doubt that it regenerates the certificates. That said, you could implement a way to "force" re-generating them.

IMHO using hooks is the right way to ensure a certain object is created just once.

mkilchhofer: Mh I am unsure what to implement for now. When we bump the major version to `2.0.0` I would introduce the new structure for the job image just to be flexible in the future. If we do not bump the major version we should not use a new structure and fix it with a hook.

We can also say that this issue is not part of this PR and we simply do not bump the major version.

dani8art:
> > * [ ]   passwords/secrets upgrade helper [here](https://github.com/bitnami/charts/blob/0f2713b0ad1ed1e820b6bdf46b29f9592cf91b44/bitnami/mongodb/templates/NOTES.txt#L192)
> 
> I think this does not make too much sense in this chart. Or what are your ideas?

Yes, you are right, it doesn't make sense.

bitnami/contour/Chart.yaml
dani8art: Hi guys!

Regarding the bumping of the version, as we are including `bitnami/common` it should be a major one. I mean, when adding these changes in other charts we are bumping it to a major.

I think we could use this major bump to ensure "regenerates the certificates" works fine after upgrading the chart.

mkilchhofer: I studied the hooks documentation here https://helm.sh/docs/topics/charts_hooks/.
Thinking about these hooks we have the following problems:
* `pre-install` seems to early for me, configmap etc. are not yet there
* `post-install` otherwise seems to late, because when we use it with the option `--wait` which most CI/CD systems do, we will run into a timeout because envoy and contour cannot start without an existing secret containing TLS secrets.

Maybe we should involve your VMware colleagues from projectcontour for a discussion?

dani8art: let's involve them to hear their opinion on what would be the best solution.

IMHO, if there is no problem updating certs each time we upgrade, I would go for variant 1 using the new structure.

bitnami/contour/templates/daemonset.yaml
```diff
@@ -3,33 +3,34 @@
...
-      labels: {{- include "contour.labels" . | nindent 8 }}
+      labels: {{- include "common.labels.standard" . | nindent 8 }}
         app.kubernetes.io/component: envoy
-    spec: {{- include "envoy.imagePullSecrets" . | nindent 6 }}
+    spec:
+      {{- include "common.images.pullSecrets" ( dict "images" (list .Values.contour.image .Values.envoy.image) "global" .Values.global) | nindent 6 }}
       {{- if .Values.envoy.affinity }}
-      affinity: {{- include "contour.tplValue" (dict "value" .Values.envoy.nodeSelector "context" $) | nindent 8 }}
+      affinity: {{- include "common.tplvalues.render" (dict "value" .Values.envoy.nodeSelector "context" $) | nindent 8 }}
       {{- end }}
```
mkilchhofer, line 25: `affinity` is defined twice

bitnami/contour/templates/deployment.yaml
```diff
@@ -15,37 +15,42 @@ spec:
...
+{{- include "common.tplvalues.render" (dict "value" .Values.contour.nodeSelector "context" $) | indent 8 }}
       {{- end }}
       {{- if .Values.contour.nodeSelector }}
-      nodeSelector: {{- include "contour.tplValue" (dict "value" .Values.contour.nodeSelector "context" $) | nindent 8 }}
+      nodeSelector: {{- include "common.tplvalues.render" (dict "value" .Values.contour.nodeSelector "context" $) | nindent 8 }}
       {{- end }}
       {{- if .Values.contour.affinity }}
-      affinity: {{- include "contour.tplValue" (dict "value" .Values.contour.affinity "context" $) | nindent 8 }}
+      affinity: {{- include "common.tplvalues.render" (dict "value" .Values.contour.affinity "context" $) | nindent 8 }}
       {{- end }}
```
mkilchhofer, line 47: `affinity` is defined twice

mkilchhofer (commented):
`affinity` is defined twice on both DaemonSet and Deployment

bitnami/contour/Chart.yaml
mkilchhofer: By this time I had a working setup with hooks, I would prefer this solution now because we really use the same image for the certificate generation process.
Let's assume we have a running contour installation in a larger company, which cannot easily upgrade the ingress controller for reasons. But then the certs needs to be regenerated. Who knows if we can generate new certificates with eg. contour-1.8 for a running contour-1.6 installation? :)

dani8art: Ok lgtm!

dani8art: Regarding the version, anyway, we should bump it into a major change, please.

bitnami/contour/README.md
```diff
@@ -65,10 +65,11 @@ The following tables lists the configurable parameters of the contour chart and
 | `contour.customResourceDeletePolicy`               | Deletion hook of customResources viah helm hooks (only helm v2)                                        | `nil`                                                   |
 | `contour.nodeSelector`                             | Node labels for contour pod assignment                                                                 | `{}`                                                    |
 | `contour.tolerations`                              | Tolerations for contour pod assignment                                                                 | `[]`                                                    |
+| `contour.antiAffinity`                             | Contour anti-affinity policy (`soft`, `hard` or `""`)                                                  | `soft`                                                  |
```
dani8art, line 68: don't you think we should be more specific here? `contour.antiAffinityPolicy`

bitnami/contour/crds/ingressroutes.yaml
```diff
@@ -1,373 +0,0 @@
-apiVersion: apiextensions.k8s.io/v1beta1
```
dani8art, line 1: I'm wondering what would happen with resources already created based on this CRD.

bitnami/contour/README.md
mkilchhofer: I copied this from the influxdb chart (https://github.com/bitnami/charts/blob/master/bitnami/influxdb/values.yaml#L157).
But yes I also think we should be more specific

bitnami/contour/Chart.yaml
mkilchhofer: Sure, then I will bump it to 2.0.0 and merge the variant2 in.

bitnami/contour/crds/ingressroutes.yaml
mkilchhofer: Helm does not delete CRDs
> There is no support at this time for upgrading or deleting CRDs using Helm. This was an explicit decision after much community discussion due to the danger for unintentional data loss. Furthermore, there is currently no community consensus around how to handle CRDs and their lifecycle. As this evolves, Helm will add support for those use cases.
Ref: https://helm.sh/docs/chart_best_practices/custom_resource_definitions/

mkilchhofer: When helm would delete CRDs, this change would be dangerous

dani8art: you are right!

bitnami/contour/README.md
```diff
@@ -55,7 +55,7 @@ The following tables lists the configurable parameters of the contour chart and
 | `rbac.create`                                      | create the RBAC roles for API accessibility                                                            | `true`                                                  |
```
dani8art, line 55: Could we add a small note at the end in the `Upgrading` section telling all the big changes regarding contour and also adding something like this for the bitnami/common addition?

```
## Upgrading

### To 2.0.0
...
<most important changes here>
...

This version also introduces `bitnami/common`, a [library chart](https://helm.sh/docs/topics/library_charts/#helm) as a dependency. More documentation about this new utility could be found [here](https://github.com/bitnami/charts/tree/master/bitnami/common#bitnami-common-library-chart). Please, make sure that you have updated the chart dependencies before executing any upgrade.
```

bitnami/contour/templates/rbac.yaml
```diff
@@ -3,117 +3,94 @@
...
+  - get
   - update
-  - patch
 ---
 apiVersion: rbac.authorization.k8s.io/v1beta1
 kind: Role
 metadata:
-  name: {{ include "contour.fullname" .}}:contour-certgen
-  labels: {{- include "contour.labels" . | nindent 4 }}
+  name: {{ include "common.names.fullname" .}}:contour-certgen
```
dani8art, line 89: do you know if there is any reason to use `:` instead of `-` as in the rest of the templates? please change it for the rest if there is no strong reason in order to be aligned.

dani8art (changes requested):
Thank you so much for these amazing changes, please take a look at my latest comments.

Could we use these other helpers as well? https://github.com/bitnami/charts/blob/master/bitnami/common/templates/_capabilities.tpl

mkilchhofer:
> Could we use these other helpers as well? https://github.com/bitnami/charts/blob/master/bitnami/common/templates/_capabilities.tpl

We could use the one for the Deployment `apiVersion`. But `batch/v1` for the Job and `apps/v1` for the DaemonSet is not available in the common chart. I suggest to leave this as-is for all 3 types of workload definitions used and do the implentations inside the common at the time the `apiVersion` changes again (eg. when the workload definitions changes to `v2` or so).

bitnami/contour/templates/rbac.yaml
mkilchhofer: Uff I simply do not know :) Maybe we need to ask the initial chart contributor (@Cellebyte ?).

Cellebyte: No it was adapted from an old chart you could change it :)

mkilchhofer: Okay :+1: Already updated it.

bitnami/contour/templates/crd-ingressroutes.yaml
```diff
@@ -1,381 +0,0 @@
-{{- if and .Values.contour.enabled .Values.contour.createCustomResource }}
```
mkilchhofer, line 1: I think I should rename the keyword `createCustomResource` to `installCRDs` like the option of the [kong chart](https://github.com/bitnami/charts/blob/master/bitnami/kong/values.yaml#L142) since these are definitions and not an instance of a definition.

bitnami/contour/README.md
mkilchhofer: Sure. Already pushed it.

bitnami/contour/templates/service.yaml
```diff
@@ -21,11 +24,11 @@ spec:
...
-  labels: {{- include "contour.labels" . | nindent 4 }}
+  name: {{ include "common.names.fullname" . }}-envoy
+  labels: {{- include "common.labels.standard" . | nindent 4 }}
     app.kubernetes.io/component: envoy
     {{- if .Values.envoy.service.labels }}
-    {{- include "contour.tplValue" (dict "value" .Values.envoy.service.labels "context" $) | nindent 4 }}
+    {{- include "common.tplvalues.render" (dict "value" .Values.envoy.service.labels "context" $) | nindent 4 }}
     {{- end }}
   annotations:
     {{- if (ne (get .Values.envoy.service.annotations "service.beta.kubernetes.io/aws-load-balancer-type") "nlb") }}
```
mkilchhofer, line 34: See comment here: https://github.com/bitnami/charts/pull/2961/files#r484070226

> While working on #3381 and testing my changes with helm 2.14 and 2.16 I get this error:
> 
> ```
> helm2 lint .
> 
> ==> Linting .
> [ERROR] templates/: parse error in "contour/templates/service.yaml": template: contour/templates/service.yaml:34: function "get" not defined
> 
> Error: 1 chart(s) linted, 1 chart(s) failed
> ```
> 
> I realized that this is a function introduced in [Sprig v3](https://github.com/Masterminds/sprig/releases/tag/v3.0.0): [Masterminds/sprig#197](https://github.com/Masterminds/sprig/pull/197).
> 
> Since bitnami needs helm v2 compat, I think we should try to solve this using an available function.

Cellebyte:
@mkilchhofer thank you for improving my Chart. :)

bitnami/contour/templates/crd-ingressroutes.yaml
dani8art: Agree!

dani8art:
> > Could we use these other helpers as well? https://github.com/bitnami/charts/blob/master/bitnami/common/templates/_capabilities.tpl
> 
> We could use the one for the Deployment `apiVersion`. But `batch/v1` for the Job and `apps/v1` for the DaemonSet is not available in the common chart. I suggest to leave this as-is for all 3 types of workload definitions used and do the implentations inside the common at the time the `apiVersion` changes again (eg. when the workload definitions changes to `v2` or so).

I would use it for the one that we have implemented in the common, please.

bitnami/contour/templates/service.yaml
dani8art: Sure!! good catch!

mkilchhofer:
Thanks for the review again @dani8art. I will implement the changes :+1: But before I go on, I made a proposal here to refactor the whole chart into subfolders based on components https://github.com/mkilchhofer/bitnami-charts/pull/4/. WDYT? Is it a good idea? If so, I would merge this in before I make changes to not do my changes twice :)

dani8art:
> Thanks for the review again @dani8art. I will implement the changes 👍 But before I go on, I made a proposal here to refactor the whole chart into subfolders based on components [mkilchhofer#4](https://github.com/mkilchhofer/bitnami-charts/pull/4). WDYT? Is it a good idea? If so, I would merge this in before I make changes to not do my changes twice :)
> 
> Update: changes are outside of the restructured files and therefore I updated both this PR and PR [mkilchhofer#4](https://github.com/mkilchhofer/bitnami-charts/pull/4).

Yes I think it would be a good approach LGTM!

mkilchhofer:
> Yes I think it would be a good approach LGTM!
Superb :) I now merged these changes also on this PR here. So from my point of view, we now should be done with all the requested changes, right?

dani8art:
> > Yes I think it would be a good approach LGTM!
> 
> Superb :) I now merged these changes also on this PR here. So from my point of view, we now should be done with all the requested changes, right?

Yes, regarding changes suggested by bitnami team it is all ok. If you are ok with your changes/refactors we could proceed to test it internally in our CI/CD system. Do you agree?

mkilchhofer:
> > > Yes I think it would be a good approach LGTM!
> > 
> > 
> > Superb :) I now merged these changes also on this PR here. So from my point of view, we now should be done with all the requested changes, right?
> 
> Yes, regarding changes suggested by bitnami team it is all ok. If you are ok with your changes/refactors we could proceed to test it internally in our CI/CD system. Do you agree?

Sure. I hope we get green light from your CI/CD system :)

dani8art (approved):
We have got the green light!

@mkilchhofer Thank you so much for this amazing job!

PR #3383, opened 2020-08-11
[bitnami/thanos]Add feature to configure cache settings store gateway.

Whyeasy:
**Description of the change**

With this change the capability is added to configure the index caching within the Store Gateway. More info: https://thanos.io/components/store.md/#index-cache

**Benefits**

Have the option to move the index cache to a `memcached` deployment or to a `caching bucket`.

**Possible drawbacks**

When a existing configmap is provided, it should contain the data with the key `config.yml`. As is done within the generated configmap.

**Checklist** 
- [X] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [X] Variables are documented in the README.md
- [X] Title of the PR starts with chart name (e.g. `[bitnami/chart]`)
- [X] If the chart contains a `values-production.yaml` apart from `values.yaml`, ensure that you implement the changes in both files

:warning: Keep in mind that if you want to make changes to the kubeapps chart, please implement them in the [kubeapps repository](https://github.com/kubeapps/kubeapps/tree/master/chart/kubeapps). This is only a synchronized mirror.

PR #3387, opened 2020-08-11
[bitnami/thanos] Add cache configuration blocks to Thanos store. Index and Bucket

### LINKED ISSUE #3384: [Thanos] Add memcached index and bucket blocks to chart as you do with `objstore`
**Which chart**:
Thanos 2.1.1

I would like to add the configuration for the Memcached stores utilised by the Thanos Store gateway since 0.12.0 for index (GA) and bucket (beta)
https://thanos.io/components/store.md/#memcached-index-cache

but providing the configuration on the Helm chart `extraFlags` 
```
storegateway:
  enabled: true
  extraFlags:
    - |-
      --store.caching-bucket.config=type: MEMCACHED
      config:
        addresses:
        - dns+thanos-memcached:11211
```
returns an error
```
Error: YAML parse error on thanos/templates/storegateway/statefulset.yaml: error converting YAML to JSON: yaml: line 57: mapping values are not allowed in this context
```

Add files to `{{ include "thanos.fullname" . }}-objstore-secret` secret created on the chart

using a conditional when the key is defined

```
data:
  objstore.yml: |-
    {{- include "thanos.tplValue" (dict "value" .Values.objstoreConfig "context" $) | b64enc | nindent 4 }}
```
similarly
```
data:
  objstore.yml: |-
    {{- include "thanos.tplValue" (dict "value" .Values.objstoreConfig "context" $) | b64enc | nindent 4 }}
{{- if .Values.indexCacheConfig }}
  index-cache.yml: |-
    {{- include "thanos.tplValue" (dict "value" .Values.indexCacheConfig "context" $) | b64enc | nindent 4 }}
{{- end }}
{{- if .Values.bucketCacheConfig }}
  bucket-cache.yml: |-
    {{- include "thanos.tplValue" (dict "value" .Values.bucketCacheConfig "context" $) | b64enc | nindent 4 }}
{{- end }}
```

Use Helm and `extraFlags` but since the value of the flag is YAML there are validation errors when this is rendered
I'm pretty sure there is a workaround for this but it wouldn't be as clean as the proposed approach, IMO

**Additional context**
-

carlosjgp:
**Description of the change**
Add configuration files to the existing secret that contains the `objstore.yml` file

**Benefits**
Enabled `index` 
- https://thanos.io/components/store.md/#memcached-index-cache
and `bucket` caching on Thanos store
- https://thanos.io/components/store.md/#caching-bucket

**Possible drawbacks**
either the strategy to externalise this secret 
https://github.com/bitnami/charts/blob/master/bitnami/thanos/values.yaml#L45
```
## Secret with Objstore Configuration
## Note: This will override objstoreConfig
##
# existingObjstoreSecret:
##  optional item list for specifying a custom Secret key. If so, path should be objstore.yml
# existingObjstoreSecretItems: []
```
or the fact that we are keeping configuration files on a secret called `{{ include "thanos.fullname" . }}-objstore-secret` doesn't make much sense...

Not sure what's the object of providing multiple secret items on the external secret if it's not to add more files but they won't be related to the "`objstore`"...?

What I'm trying to highlight here is that I'm happy to create a different secret for the cache instead of keeping it on the same secret as the `objstore` but I'd like to know what's the use case for adding `existingObjstoreSecretItems` that is not the required `objstore.yml`

**Applicable issues**
  - fixes #3384
  - overlaps #3383

**Additional information**
These changes are currently running in our CI cluster

**Checklist**


Whyeasy:
The reason why I added the `existingObjstoreSecretItems ` is because I needed to point to a custom key within my secret. As we have a certain way for creating our secrets where I work. So it's not for adding more secrets, but a way to specify a keyname and a path, `objstore.yml`. So the chart will work with all other ways to provide a `objstore.yml`.

The reason why objstore is a separate secret in this case. It hold sensative data which I won't like to have in my helm values in a repo.

PR #3388, opened 2020-08-11
[bitnami/phpmyadmin]Add allowArbitraryServer environment variable

clementmattera:
**Description of the change**

Add _allowArbitraryServer_ environment variable to be able to modify via **values.yaml** the variable _PHPMYADMIN_ALLOW_ARBITRARY_SERVER_ currently set to "true" in **deployment.yaml** . The default value for _allowArbitraryServer_ remains "true" so the default behavior doesn't change.

**Benefits**

Being able to access _PHPMYADMIN_ALLOW_ARBITRARY_SERVER_ without modifying **deployment.yaml**.

**Checklist** 

### DIFF
--- a/bitnami/phpmyadmin/Chart.yaml
+++ b/bitnami/phpmyadmin/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: phpmyadmin
-version: 6.3.4
+version: 6.4.0
 appVersion: 5.0.2
 description: phpMyAdmin is an mysql administration frontend
 keywords:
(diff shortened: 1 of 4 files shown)


bitnami/phpmyadmin/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: phpmyadmin
-version: 6.3.3
+version: 6.3.4
```
carrodher, line 3: As it is not bug fixing but a new feature, can you bump the minor version instead?
```suggestion
version: 6.4.0
```

PR #3397, opened 2020-08-12
external-dns infoblox provider additional parameter

razvan-ig:
**Description of the change**

https://github.com/kubernetes-sigs/external-dns/blob/master/provider/infoblox/infoblox.go#L46 adds a new parameter (view) required for servers that don't have just a "default" view (matching the Bind 9 syntax).

### DIFF
--- a/bitnami/external-dns/Chart.yaml
+++ b/bitnami/external-dns/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: external-dns
-version: 3.2.7
+version: 3.3.0
 appVersion: 0.7.3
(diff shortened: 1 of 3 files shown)


marcosbc:
BTW, I just checked the available CLI options and it looks like the default for `infoblox-view` is `""` and not `default`. See [this](https://github.com/kubernetes-sigs/external-dns/blob/master/pkg/apis/externaldns/types.go#L354). Could you check if that is the case?

razvan-ig:
> BTW, I just checked the available CLI options and it looks like the default for `infoblox-view` is `""` and not `default`. See [this](https://github.com/kubernetes-sigs/external-dns/blob/master/pkg/apis/externaldns/types.go#L354). Could you check if that is the case?

Sorry for the confusion, on our setup the default view is confusingly named "default"

PR #3399, opened 2020-08-12
[bitnami/apache] Add resources for git container

chifu1234:
**Description of the change**
Currently on git init containers you can not set any resources
**Benefits**
support for resources on git init container

**Checklist** 

### DIFF
--- a/bitnami/apache/values.yaml
+++ b/bitnami/apache/values.yaml
@@ -79,6 +79,7 @@ cloneHtdocsFromGit:
   # repository:
   # branch:
   interval: 60
+  resources: {}
 
 ## Name of a config map with the server static content
 ##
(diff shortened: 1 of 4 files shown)


bitnami/apache/values.yaml
```diff
@@ -38,6 +38,7 @@ git:
   repository: bitnami/git
   tag: 2.28.0-debian-10-r6
   pullPolicy: IfNotPresent
+  resources: {}
```
rafariossaa, line 41: I think this is not the right place to put this line.
Shouldn't it be at in the block of line 78 ?
```
cloneHtdocsFromGit:
  enabled: false
  # repository:
  # branch:
  interval: 60
```

chifu1234: yes your are right. I update the PR.

PR #3404, opened 2020-08-12
[bitnami/rabbitmq] Feature/username secret alignment

peppe77:
**Description of the change**

allows username to be stored in the same secret as password and cookies. same logic as password, if auth.username is not defined then user will be b64 encoded and stored in the same secret as password/cookies are. moreover, it is also possible to user existingSecret that has username.

**Benefits**

app charts no longer needs to hard code 'username' defined via auth.ausername though left outside secret.

**first time contribution to helm/chart - if anything I may have missed and/or not as expected. Please let me know.


carrodher:
Hi, thanks for the PR. As this change should be implemented in all the charts, just to maintain the same logic across all of them and avoid divergences in some common sections like this one, I checked this topic with other members of the team.

The majority idea is that although we understand that it could be a good point having user and password close; at the end store in a secret something that is not a secret is not the best option for us. In the same way, passwords can be randomly generated but usernames are used to be constant. In terms of "secret" as secure storage, a public username is not a security risk at all and it is a common practice, so we don't fully see a clear advantage of storing the username in a secret

peppe77:
Don't quite agree - here is why. It is more convenient user/password under same secret (simplifies our charts) as opposed to hard code it outside. I'd even argue that username is sensitive because both pieces are needed if one wants to do anything. So, as of now, we are giving away half of what someone would need to access the broker. (Let alone this very same behavior exists in the previous rabnitmq charts - prior to Bitnami). So, I don't feel that argument is strong enough to drop this PR. Sure, if you merge, it will mean that eventually same would have to be done across all charts. Isn't that great? Why would you want give username away and have app charts hardcoding username?

mkmik:
The more I think about it the more I agree with @peppe77 , although I still don't think the username should be considered a secret (because in most authentication protocols, it's not treated as such, being sent in clear over the wire etc).

For me the strongest selling point is credentials rotation: if I want to rotate a credential without causing a downtime, I'd first need to create a new user account (user+password pair) on the server, configure the client to use it, then delete the old one from the server.

Pairing up the username+password in one single object, simplifies this rollout (and possible rollback in case of problems).

juan131:
> I still don't think the username should be considered a secret (because in most authentication protocols, it's not treated as such, being sent in clear over the wire etc)

Totally agree on that.

>  It is more convenient user/password under same secret (simplifies our charts) as opposed to hard code it outside.

> For me the strongest selling point is credentials rotation: if I want to rotate a credential without causing a downtime, I'd first need to create a new user account (user+password pair) on the server, configure the client to use it, then delete the old one from the server.

Having both user/password under the same secret **will simplify things exclusively when using a existing secret to manage credentials.** I mean, in case I'm delegating the secrets creation to the chart providing the credentials via values.yaml the complexity is exactly the same one.

That said, I agree that's more compatible with using a 3rd systems to manage the credentials on my cluster, and it simplifies credentials rotation. So, IMHO I think we should go in this direction and move the username(s) to the same secret(s) where we store the passwords.

javsalgar:
>> For me the strongest selling point is credentials rotation: if I want to rotate a credential without causing a downtime, I'd first need to create a new user account (user+password pair) on the server, configure the client to use it, then delete the old one from the server.
> Having both user/password under the same secret will simplify things exclusively when using a existing secret to manage credentials. I mean, in case I'm delegating the secrets creation to the chart providing the credentials via values.yaml the complexity is exactly the same one.

In my opinion, I like the existing secret approach when providing credentials rather than using the values.yaml. In this case, making this easier for this approach is a good selling point to me, so I vote to go on with the unification approach.

Issue #3331, opened 2020-08-05
postgresql : password authentication failed for user "postgres" / User "postgres" has no password assigned.


yaakov-berkovitch:
**Which chart**:
postgresql-9.1.2

**Describe the bug**
Install the postgresql chart and set helm to take the postgres password from k8s secret.
After pod is up and running, tries to login to postgres using psql and it failed

**To Reproduce**
1. `/usr/local/bin//helm install postgresql bitnami/postgresql --version 9.1.2 -f /tmp/postgres-values.yml`
2. Connect to the pod, and execute `psql -U postgres`
   Got the failure: `psql: error: could not connect to server: FATAL:  password authentication failed for user "postgres"`
4. In the log file we can see those errors:
```
2020-08-05 02:43:25 EDT postgres 3539  FATAL:  password authentication failed for user "postgres"
2020-08-05 02:43:25 EDT postgres 3539  DETAIL:  User "postgres" has no password assigned.
        Connection matched pg_hba.conf line 1: "host     all             all             0.0.0.0/0               md5"
```
The values file is as follows:
```
## Postgres image tag
image:
  tag: 12.3.0-debian-10-r75

# ConfigMap with the PostgreSQL configuration
configurationConfigMap: postgresql.conf

## PostgreSQL password using existing secret
existingSecret: postgres-secret

## Mount PostgreSQL secret as a file instead of passing environment variable
usePasswordFile: false

persistence:
  enabled: true
  existingClaim: postgres-data-pvc
  storageClass: "postgres-data-pv"

volumePermissions:
  enabled: true

master:
  podAnnotations:
    sidecar.istio.io/inject: "false"
```
And the k8s secret:
```
apiVersion: v1
data:
  postgresql-password: YWxnb3NlYzE=
  postgresql-postgres-password: YWxnb3NlYzE=
kind: Secret
metadata:
  creationTimestamp: "2020-08-04T13:15:04Z"
  managedFields:
  - apiVersion: v1
    fieldsType: FieldsV1
    fieldsV1:
      f:data:
        .: {}
        f:postgresql-password: {}
        f:postgresql-postgres-password: {}
      f:type: {}
    manager: kubectl
    operation: Update
    time: "2020-08-04T13:15:04Z"
  name: postgres-secret
  namespace: default
  resourceVersion: "1235"
  selfLink: /api/v1/namespaces/default/secrets/postgres-secret
  uid: 94b411a7-c169-441d-876e-40fa0b781030
type: Opaque
```

[...]

PR #3408, opened 2020-08-13
[bitnami/postgresql] Fix Postgres password with externalSecret

joancafom:

**Description of the change**
Fixes a bug that forced to provide with a dummy password for `postgres` user when using the `externalSecret` feature.

**Benefits**
There is no longer need to provide with dummy passwords when using external Secrets.

**Possible drawbacks**

**Applicable issues**

  - Partially solves #3331

**Additional information**
Note that according to the documentation (see [README.md](https://github.com/bitnami/charts/blob/master/bitnami/postgresql/README.md)), when using the `externalSecret` feature the provided secret **must** contain the key for `postgresql-postgres-password` (even if the value is empty), so there is no possibility for this key to be missing.

**Checklist** 


bitnami/postgresql/templates/statefulset.yaml
```diff
@@ -189,7 +189,7 @@ spec:
             - name: POSTGRES_CLUSTER_APP_NAME
               value: {{ .Values.replication.applicationName }}
             {{- end }}
-            {{- if and .Values.postgresqlPostgresPassword (not (eq .Values.postgresqlUsername "postgres")) }}
+            {{- if and (or (not (include "postgresql.createSecret" .)) .Values.postgresqlPostgresPassword) (not (eq .Values.postgresqlUsername "postgres")) }}
```
marcosbc, line 192: These conditions are really hard to read (and understand):

```lua
{{- if and (or (not (include "postgresql.createSecret" .)) .Values.postgresqlPostgresPassword) (not (eq .Values.postgresqlUsername "postgres")) }}
```

"If a new secret does not need to be created or a password is specified, then also check that the username is Postgres"

I think that just by re-ordering this would make it simpler:

```lua
{{- if and (not (eq .Values.postgresqlUsername "postgres")) (or .Values.postgresqlPostgresPassword (not (include "postgresql.createSecret" .))) }}
```

"When the username is postgres, and either a password was specified or a new secret does not need to be created"

I think we could simplify it even more if we improve the `postgresql.createSecret` helper:

```lua
{{/*
Return true if a secret object should be created
*/}}
{{- define "postgresql.useExistingSecret" -}}
{{- if or .Values.global.postgresql.existingSecret .Values.existingSecret -}}
    {{- true -}}
{{- end -}}
{{- end -}}

{{/*
Return true if a secret object should be created
*/}}
{{- define "postgresql.createSecret" -}}
{{- if not (include "postgresql.useExistingSecret" .) -}}
    {{- true -}}
{{- end -}}
{{- end -}}
```

And then:

```lua
{{- if and (not (eq .Values.postgresqlUsername "postgres")) (or .Values.postgresqlPostgresPassword (include "postgresql.useExistingSecret" .)) }}
```

"When the username is postgres, and either a password was specified or an existing secret will be used"

mot256:
Should you not also be updating the docker tag to at least 11.8.0-debian-10-r79 ? See https://github.com/bitnami/bitnami-docker-postgresql/pull/239

marcosbc:
> Should you not also be updating the docker tag to at least 11.8.0-debian-10-r79 ? See [bitnami/bitnami-docker-postgresql#239](https://github.com/bitnami/bitnami-docker-postgresql/pull/239)

Our internal pipeline is responsible for updating the Docker images, e.g. see the related commit for the tag you mentioned: https://github.com/bitnami/charts/commit/47b3f22873a2e622442ae7962d19a07b74b5de91

bitnami/postgresql/templates/statefulset.yaml
joancafom: Thanks very much for your suggestions, changes have been applied 😁

rafariossaa:
LGTM

PR #3454, opened 2020-08-18
aspnet-core - Adding the health ingress

BaskaranSelvaraj:
**Description of the change**

Adding a separate health ingress for checking the health status of the application and it will be use for traffic manager to monitor the health check of the application with specific timeline.

**Benefits**

Created the separate health ingress to monitor the application health & Availability.

**Possible drawbacks**

No drawbacks

**Additional information**

We can't use the same ingress which is available now for the health status. Since i want to use the ingress annotation **nginx.ingress.kubernetes.io/rewrite-target: /health/status** which is going to use only for ingress health host  and not for the application host. So created a separate health ingress.
**Note:** Exactly used the same existing ingress copy as a new health ingress


bitnami/aspnet-core/values.yaml
```diff
@@ -395,6 +395,61 @@ ingress:
...
+  ## If it is not set and you're using cert-manager, this is unneeded, as it will create the secret for you
+  ## If it is not set and you're NOT using cert-manager either, self-signed certificates will be created
+  ## It is also possible to create and manage the certificates outside of this helm chart
+  ## Please see README.md for more information
+  ##
+  secrets: []
+  ## - name: aspnet-core.local-tls
+  ##   key:
+  ##   certificate:
+
```
juan131, line 452: You need to update the **values-production.yaml** too

bitnami/aspnet-core/templates/health-ingress.yaml
```diff
@@ -0,0 +1,51 @@
...
+    {{- end }}
+    {{- range .Values.healthIngress.extraHosts }}
+    - host: {{ .name }}
+      http:
+        paths:
+          - path: {{ default "/" .path }}
+            backend:
+              serviceName: {{ include "aspnet-core.fullname" $ }}
+              servicePort: http
+    {{- end }}
```
juan131, line 39: Do we really need to support N hostnames for the healtcheck endpoint?

bitnami/aspnet-core/templates/health-ingress.yaml
```diff
@@ -0,0 +1,51 @@
...
+  {{- if or .Values.healthIngress.tls .Values.healthIngress.extraTls .Values.healthIngress.hosts }}
+  tls:
+    {{- if .Values.healthIngress.tls }}
+    - hosts:
+        - {{ .Values.healthIngress.hostname }}
+      secretName: {{ printf "%s-tls" .Values.healthIngress.hostname }}
+    {{- end }}
+    {{- if .Values.healthIngress.extraTls }}
+    {{- toYaml .Values.healthIngress.extraTls | nindent 4 }}
+    {{- end }}
```
juan131, line 49: Same question here

bitnami/aspnet-core/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: aspnet-core
-version: 0.1.4
+version: 0.1.5
```
juan131, line 3: You'll have to bump the minor version instead.

juan131 (changes requested):
Thanks so much for your contribution! Please take a look to my comments.

Just curious before moving on, did you try playing with the `ingress.extraHosts[].path` parameter? Isn't that enough?

bitnami/aspnet-core/templates/health-ingress.yaml
BaskaranSelvaraj: Yes, its required and we are using extraHosts to specifically mentioned path.

bitnami/aspnet-core/templates/health-ingress.yaml
BaskaranSelvaraj: in our scenario we have not used it in azure k8s but it may be useful for other environments

bitnami/aspnet-core/values.yaml
BaskaranSelvaraj: we need to  use the different annotations for health ingress. example **nginx.ingress.kubernetes.io/rewrite-target: /health/status** which not able to achieve in same ingress.
Updated the values-production.yaml

bitnami/aspnet-core/Chart.yaml
BaskaranSelvaraj: Changed.

juan131 (approved):
LGTM

PR #3541, opened 2020-08-28
[bitnami/mongodb-sharded] Bump major version due to new major in the app

carrodher:
**Description of the change**

The new MongoDB 4.4 should have bumped the chart version in a major, but it was not the case due to a weird race condition, see https://github.com/bitnami/charts/commit/18311369fa24d3c220ba1d4386d023cd2dd1f923

The upgrade notes were already placed in the README, see https://github.com/bitnami/charts/tree/master/bitnami/mongodb-sharded#to-200 so this PR is only bumping the chart version as the CI/CD should have done automatically


andresbono (approved):
LGTM

PR #2453, opened 2020-04-28
[bitnami/redis] Fix zombie processes in readiness/liveness check

javsalgar:
**Description of the change**

Change the signal of the liveness and readiness check. Also needed to perform several linting changes so our automated scripts would allow pushing the change.

**Benefits**

**Possible drawbacks**

**Applicable issues**

  - fixes https://github.com/bitnami/charts/issues/2441

**Additional information**

**Checklist** 
- [x] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [x] Variables are documented in the README.md
- [x] Title of the PR starts with chart name (e.g. `[bitnami/chart]`)
- [x] If the chart contains a `values-production.yaml` apart from `values.yaml`, ensure that you implement the changes in both files

:warning: Keep in mind that if you want to make changes to the kubeapps chart, please implement them in the [kubeapps repository](https://github.com/kubeapps/kubeapps/tree/master/chart/kubeapps). This is only a synchronized mirror.

PR #3559, opened 2020-08-30
[bitnami/redis] prevent zombie PIDs in redis health checks

### LINKED ISSUE #2441: zombie processes caused by health checks
**Which chart**:
bitnami redis template version 9.0.2 and also 10.5.7

**Describe the bug**
on related docker nodes, you'll find redis-cli zombie processes:
```
# ps -ef | grep defunc
1001       920 29540  0 04:17 ?        00:00:00 [redis-cli] <defunct>
1001      1519 29540  0 Apr22 ?        00:00:00 [redis-cli] <defunct>
1001      9561 14677  0 Apr15 ?        00:00:00 [redis-cli] <defunct>
1001     12851 29540  0 Apr16 ?        00:00:00 [redis-cli] <defunct>
1001     16168 29540  0 Apr17 ?        00:00:00 [redis-cli] <defunct>
1001     22101 29540  0 Apr18 ?        00:00:00 [redis-cli] <defunct>
1001     24210 14677  0 07:58 ?        00:00:00 [redis-cli] <defunct>
1001     24285 14677  0 Apr22 ?        00:00:00 [redis-cli] <defunct>
1001     27874 14677  0 Apr21 ?        00:00:00 [redis-cli] <defunct>
1001     28288 14677  0 Apr16 ?        00:00:00 [redis-cli] <defunct>
```
those zombies seem to get caused from readiness / liveness health checks, when slave or master could not be reached within related timeout.

see also https://github.com/bitnami/bitnami-docker-redis/issues/165

**To Reproduce**

1. simulate connection problem by iptables DROP rule: `iptables -I INPUT -p tcp --dport 6379 -j DROP`

2. login into redis-master or redis-slave container (pod) and execute health check:
```
docker exec -ti redis-master  bash
 
I have no name!@af3cc23c7511:/$ timeout -s 9 1 redis-cli -a 1q2w3e4r --no-auth-warning -h 192.168.99.99 -p 6379 ping
Killed
```

3. look at the OS of docker node:
```
# ps -ef | grep defunc
karli     3681  2105  0 09:42 pts/0    00:00:00 [redis-cli] <defunct>
root      3683  2832  0 09:42 pts/1    00:00:00 grep --color=auto defunc
```

**Expected behavior**
having no zombie processes on docker nodes after some days the docker redis service is running.

- Output of `helm version`:

```
branch": "master",
"commit": "f5f5d10c5255216e757f8bec5651aa8a"
```

```
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.1", GitCommit:"d647ddbd755faf07169599a625faf302ffc34458", GitTreeState:"clean", BuildDate:"2019-10-02T17:01:15Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"linux/amd64"}
```

**Additional context**
after looking inside the health checks mounted from helm chart via config map at `/health` and running some tests, i found a solution:
instead of `timeout -s 9` just use `timeout -s 3` in following scripts:
- ping_liveness_local.sh
- ping_liveness_master.sh
- ping_readiness_local.sh
- ping_readiness_master.sh
which are dynamic generated by https://github.com/bitnami/charts/blob/master/bitnami/redis/templates/health-configmap.yaml
when using kill signal 3 instead of 9, no zombie process will be spawned any more

because the kill signal for the `timeout` command is hard coded, please change it to 3 or replace it with an environment variable for more flexibility.
thx

kabakaev:
**Description of the change**

Though PR #2453 decreased probability of zombie PIDs generation, the problem still exists, see `Steps to reproduce` below.

The same issue probably exists in the [redis-cluster chart](https://github.com/bitnami/charts/blob/07bf067a628615aa8b182da90c0fc733d3a792ff/bitnami/redis-cluster/templates/redis-statefulset.yaml#L183-L190). Let me know if a PR for redis-cluster is needed.

This PR fixes #2441 once and for all.

**Benefits**

Whatever a user may write into readiness or liveness probes, no more zombie PIDs will be accumulated. Default settings will also stop generating zombies randomly.

**Possible drawbacks**

None.

**Applicable issues**

  - fixes #2441.

**Root cause**

Both `livenessProbe` and `readinessProbe` are configured as `exec` of a shell script. Each script has a configurable timeout parameter. The same timeout value is used in the `timeoutSeconds` parameters of the probe spec. This creates a race condition: if the probe script exits longer than the `kubelet` Go runtime schedules the timeout handler, then `kubelet` will fire the probe timeout procedure, leaving behind the probe script zombie PID. From what i see, such race happens more often on nodes with high CPU utilization.

This PR increases the `timeoutSeconds` parameter by 1 second, giving the probe script enough time to finish and let exit code reach `kubelet`.

The second commit also enables `shareProcessNamespace` pod option, which is the universal solution to the problem of zombie PID accumulation. Details are given below.

**Steps to reproduce**

Tested in [kind](https://kind.sigs.k8s.io/docs/user/quick-start/).

1. Install redis chart: `helm upgrade test1 ./bitnami/redis --install --set=cluster.enabled=false,usePassword=false,persistence.enabled=false,metrics.enabled=false,master.livenessProbe.enabled=false`

1. Make sure the pod is running:
    ```bash
    kubectl get po
    #NAME                   READY   STATUS    RESTARTS   AGE
    #test1-redis-master-0   1/1     Running   0          5m1s
    ```
1. Run a `watch ps` inside container:
    ```bash
    kubectl exec -ti test1-redis-master-0 -- bash -c "watch -n1 'ps auxwf'"
    #Every 1.0s: ps auxwf
    #
    #USER       PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
    #1001      2586  0.1  0.0   2988  2328 pts/0    Ss+  19:47   0:00 watch -n1 ps auxwf
    #1001      4446  0.0  0.0   2988   424 pts/0    S+   19:48   0:00  \_ watch -n1 ps auxwf
    #1001      4447  0.0  0.0   2384   764 pts/0    S+   19:48   0:00      \_ sh -c ps auxwf
    #1001      4448  0.0  0.0   7636  2768 pts/0    R+   19:48   0:00          \_ ps auxwf
    #1001         1  0.1  0.0  52792  7752 ?        Ssl  19:46   0:01 redis-server *:6379
    ```
1. Find the redis process in host and stop it:
    ```bash
    # run in another terminal outside any cotnainer namespace
    ps ax | grep -E 'redis-server|COMMAND'
    #USER       PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
    #1001     30042  0.1  0.0  52792  7672 ?        Ssl  19:46   0:01 redis-server *:6379

    kill -STOP 30042
    ```
1. Look back on the `ps watch` inside containter to see the zombie PIDs ("defunct").
1. "Unpause" redis: `kill -CONT 30042`.
1. The zombie PIDs will remain.
    ```
    1001         1  0.1  0.0  52792  7672 ?        Ssl  19:46   0:01 redis-server *:6379
    1001       599  0.0  0.0      0     0 ?        Z    19:48   0:00 [ping_readiness_] <defunct>
    1001       612  0.0  0.0      0     0 ?        Z    19:48   0:00 [ping_liveness_l] <defunct>
    ```

**Why even bother about zombie PIDs?**

Though it may seem not important, but high amount of zombies may lead to all sorts of problems, from inability to make redis snapshot to dead kubernetes daemons. With health probes running every 5 seconds, it's quite easy to reach a [pod PID limit](https://kubernetes.io/blog/2019/04/15/process-id-limiting-for-stability-improvements-in-kubernetes-1.14/).

**Shared process namespace is good**

In kubernetes, `shareProcessNamespace: true` option has a side effect that all containers will run with proper init process.
Specifically, the first process in first container of a pod gets PID=1 (init), which becomes also PID=1 in all other containers. In kubernetes, first pod process is always `/pause` from the pause container (part of network setup). This mighty binary can read exit codes of orphaned processes and thus prevent zombies from exhausting the pod/node PID limit.
Such shared PID namespace mode was enabled in k8s@docker by default for some time. But with introduction of `shareProcessNamespace` option, the shared PID NS was switched back to disabled by default.
Thus, an easy solution would be to enable `shareProcessNamespace` on all pods which use `exec` health check.

Otherwise, in order to avoid leaving zombies behind, each exec script must be written to finish in a guaranteed time shorter as the k8s probe `timeoutSeconds` parameter. This is nearly impossible to do for user-defined health checks.

**Checklist** 


dani8art:
Hi @kabakaev, Thanks for opening this PR! 

I think is a good approach to have less timeout in the probe script than in the probe spec so the probe script could finish safely and we'll avoid the zombie PID's, I've just had a question regarding this: what would be the best time between them? 1 sec would be enough? I mean, It could depend on the CPU usage of the cluster, right? wasn't it good to add a new property to let the user configure it?

Regarding `sharedProcessNamespace`, could it bring back some security issues? we would like to keep them separated if the timeout approach already fixed the PID's problem I would keep it as default in Kubernetes.

kabakaev:
@dani8art, thanks for quick review!

> I think is a good approach to have less timeout in the probe script than in the probe spec. (...) Wasn't it good to add a new property to let the user configure it?

The scrips are [written already in a way to guarantee that the probe ends in exact time specified](https://github.com/bitnami/charts/blob/498056ad16a6e89aa3b7cc231da7467ab5bd3986/bitnami/redis/templates/health-configmap.yaml#L22), maybe only a few milliseconds longer than configured. Thus, it's not the the kubernetes object spec's `timeoutSeconds` parameter which defines exact timeout, but the value passed as first parameter to a health check script. 
In fact, the `timeoutSeconds` in pod spec can have any integer value, provided it is higher than the value passed to the health check script. IMHO, no separate parameter is required, because sense of `timeoutSeconds` is not to set the pod parameters, but to guarantee the timeout is small enough to prevent many simultaneous health checks. 

Moreover, setting the timeout to `timeoutSeconds - 1` will not be transparent. Think of `master.readinessProbe.timeoutSeconds=1` (the default now). I think, the comment in generated yaml is enough, though we may copy it over to all places where `add1()` function is used. HDYT?


> Regarding `sharedProcessNamespace`, could it bring back some security issues?

I'm not aware of any security issues with shared process namespace. In fact, shared PID namespace [was default setting for kubernetes@docker](https://github.com/kubernetes/kubernetes/pull/45236) for a while.

> Regarding 1sharedProcessNamespace1, (...) I would keep it as default in Kubernetes.

I'm fine with that. First commit already fixes the issue for default settings of liveness/reasiness probes. 
Still, it may be helpful to be able to switch on the `master.sharedProcessNamespace` or `slave.sharedProcessNamespace` if customReadinessProbe or customLivenessProbe is used.

I've updated the second commit accordingly.

What about `bitnami/redis-cluster`? Personally, i do not use it. Should we leave it broken?

dani8art:
> What about bitnami/redis-cluster? Personally, i do not use it. Should we leave it broken?

I think we should tackle as well, are you able to open a PR too?

> Still, it may be helpful to be able to switch on the master.sharedProcessNamespace or slave.sharedProcessNamespace if customReadinessProbe or customLivenessProbe is used.

That's right but could you set it `false` by default in values yaml as it is the default behavior in k8s now, please?

Regarding timeout, it is ok to me, let see how it works and if there is no issue with that. In case someone comes with an issue then we will change to a separate parameter. 

Thank you so much for the detailed explanation.

kabakaev:
> could you set it `false` by default in values yaml

I said that it's switched to `false`, but forgot to stage changes before `git commit --amend` yesterday :)

The change is pushed properly now.

kabakaev:
> are you able to open a PR too?

Sure, will do.

bitnami/redis/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: redis
-version: 10.7.17
+version: 10.7.18
```
dani8art, line 3: As we are adding a new parameter to the values, we should bump it into a minor instead of a patch version.

dani8art (approved):
LGTM!
Just a small change. 

Thank you so much!

bitnami/redis/Chart.yaml
kabakaev: @dani8art, done.

PR #3564, opened 2020-08-31
aspnet-core - Adding condition for command and env for more customization

BaskaranSelvaraj:
**Description of the change**

Adding condition for command and env for more customization and its not mandatory for all the cases and scenario.

**Benefits**

more customization condition for command and env which is not mandatory attributes

Possible drawbacks

No drawbacks

**Additional information**

command is preventing to run the container in the k8s and its showing up the command prompt with dotnet. docker image is already build with application dll with entry point so in that case dotnet command prompt is not required.
env variable is not mandatory is always use to let it be customizable and in my its not required to set.


carrodher:
Hi, thanks for creating the PR.

About the command, it is something already supported; using the `command` parameter you should be able to override the default container command; if you don't want to set any command I think the best option is to use the same command that is used as the entrypoint in the Dockerfile.

You can find more information at https://github.com/bitnami/charts/tree/master/bitnami/aspnet-core#deploying-your-custom-aspnet-core-application

About the environment variable I would implement something to just disable this one:
```yaml
## URLs to bind
##
bind:
  enabled: true
  URLs: http://+:8080
```
and then
```
          env:
            {{- if .Values.bind.enabled }}
            - name: ASPNETCORE_URLS
              value: {{ .Values.bindURLs | quote }}
            {{- end }}
            {{- range $key, $value := .Values.extraEnvVars }}
            - name: {{ $key }}
              value: "{{ $value }}"
            {{- end }}
```

BaskaranSelvaraj:
i am good with the approach for the env ( environment variable). common practice for the dockerfile will be always expose the entrypoint. do we really need the command prompt in top of that, because in my case deployment is failing due to the command by overriding the entrypoint.

carrodher:
In the case of the [default image](https://github.com/bitnami/bitnami-docker-aspnet-core/tree/master/3.1/debian-10), the Dockerfile looks like
```Dockerfile
ENTRYPOINT [ "/opt/bitnami/scripts/aspnet-core/entrypoint.sh" ]
CMD [ "/bin/bash" ]
```
where _entrypoint.sh_ is
```bash

#!/bin/bash -e

# shellcheck disable=SC1091

. /opt/bitnami/base/functions
. /opt/bitnami/base/helpers

print_welcome_page

exec "$@"
```
then in the chart, the default command is `dotnet` if not custom commands are specified by setting the `command` or `appFromExternalRepo.startCommand` parameters.

In your use case, can't you use the already available parameters to make the chart works with your custom image?

BaskaranSelvaraj:
Thank you for your help on this. 
i can able to override the existing port and command from my custom docker file to host the application.

PR #3569, opened 2020-08-31
[bitnami/mariadb] Fix annotation for replica statefulset

bergemalm:
**Description of the change**

Currently annotations are treated differently in the master vs slave statefulset template. This also causes issues with generating slave manifest when having annotations not using strict key: value format.

**Benefits**
No discrepancy in templating slave annotations vs master statefulsets and generating slave manifest works when for example using multi-line value.

**Possible drawbacks**

**Applicable issues**
**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/mariadb/templates/slave-statefulset.yaml
+++ b/bitnami/mariadb/templates/slave-statefulset.yaml
@@ -26,9 +26,8 @@ spec:
     {{- end }}
   template:
     metadata:
-      {{- with .Values.slave.annotations }}
-      annotations:
-        {{- toYaml . | nindent 8 }}
+      {{- if .Values.slave.annotations }}
+      annotations: {{- include "mariadb.tplValue" (dict "value" .Values.slave.annotations "context" $) | nindent 8 }}
       {{- end }}
       labels:
         app: {{ template "mariadb.name" . }}
--- a/bitnami/mariadb/templates/slave-statefulset.yaml
+++ b/bitnami/mariadb/templates/slave-statefulset.yaml
@@ -26,11 +26,9 @@ spec:
     {{- end }}
   template:
     metadata:
-      {{- if .Values.slave.annotations }}
+      {{- with .Values.slave.annotations }}
       annotations:
-        {{- range $key, $value := .Values.slave.annotations }}
-        {{ $key }}: {{ $value }}
-        {{- end }}
+        {{- toYaml . | nindent 8 }}
       {{- end }}
       labels:
         app: {{ template "mariadb.name" . }}
--- a/bitnami/mariadb/Chart.yaml
+++ b/bitnami/mariadb/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: mariadb
-version: 7.9.2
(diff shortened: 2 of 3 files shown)


bitnami/mariadb/templates/slave-statefulset.yaml
```diff
@@ -26,11 +26,9 @@ spec:
     {{- end }}
   template:
     metadata:
-      {{- if .Values.slave.annotations }}
+      {{- with .Values.slave.annotations }}
```
miguelaeh, line 29:
```suggestion
      {{- if .Values.slave.annotations }}
```

bitnami/mariadb/templates/slave-statefulset.yaml
```diff
@@ -26,11 +26,9 @@ spec:
...
       annotations:
-        {{- range $key, $value := .Values.slave.annotations }}
-        {{ $key }}: {{ $value }}
-        {{- end }}
+        {{- toYaml . | nindent 8 }}
```
miguelaeh, line 31:
```suggestion
        annotations: {{- include "mariadb.tplValue" (dict "value" .Values.slave.annotations "context" $) | nindent 8 }}
```

miguelaeh (changes requested):
I think it would be nice if you apply the sugestions to use our standard way, also, it would be nice if you ould make the proper change for the master too.
You can check this: https://github.com/bitnami/charts/blob/4e3eefc450b111a70b1ac9a58da0a1ff75a65c38/bitnami/mariadb-galera/templates/statefulset.yaml#L20
that will render whatever is put at podAnnotations: https://github.com/bitnami/charts/blob/4e3eefc450b111a70b1ac9a58da0a1ff75a65c38/bitnami/mariadb-galera/values.yaml#L401

bergemalm:
Hi @miguelaeh, sure I can do that and the same for master _but_ this means that you have to declare annotation as string literal when using for example Vault and secret templating.

Example before change (and only worked for master statefulset before this PR):
```
vault.hashicorp.com/agent-inject-template-my-password: |
      {{- with secret "my/path/to/password" -}}
        {{ .Data.data.password }}
      {{- end }}
```
after change:
```
vault.hashicorp.com/agent-inject-template-my-password: |
      {{`{{- with secret "my/path/to/password" -}}
        {{ .Data.data.password }}
      {{- end }}`}}
```

are we ok with that?

bergemalm:
PR updated!

miguelaeh:
Hi @bergemalm ,
How are you using the secret in your previous comment? Is
```
vault.hashicorp.com/agent-inject-template-my-password: |
      {{- with secret "my/path/to/password" -}}
        {{ .Data.data.password }}
      {{- end }}
```
part of your `values.yaml`?

Was that properly rendered before, I thought if you pass that in the values.yaml it will not be templatized in the statefulset due to the toYaml

bergemalm:
Yes part of `values.yaml` and `toYaml` works without issues. @miguelaeh

miguelaeh (approved):
Thank you for addressing the changes.
LGTM!

PR #3590, opened 2020-09-02
[bitnami/mongodb] Additional URI options of the Prometheus metrics

binboum:
Signed-off-by: Maxence Laude <maxence@laude.pro>

Description of the change

Additional URI options of the Prometheus metrics

Benefits

The method to add TLS options in metrics-server to be evaluated from version 10.

It is necessary to add options via URI now.


bitnami/mongodb/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: mongodb
-version: 9.0.1
+version: 9.0.2
```
javsalgar, line 3: Could you bump the minor version? It is adding a feature

javsalgar (commented):
Thank you very much! Minor comment

bitnami/mongodb/Chart.yaml
binboum: Of course :+1:

PR #3596, opened 2020-09-03
[bitnami/redis-cluster] Service acount name needs to be set on jobs.

SimonOlenik:
**Description of the change**

Service account is mounted also on post install and post update jobs. 

**Benefits**

In very restrictive k8s cluster, default pod security policies can prevent chart from working. Your chart has great support for PSP, RBAC and service account, but unfortunately service account is only set on statefulset. So statefulset is working, but post-install and post-update job fail due to using PSP not created by chart. 

**Possible drawbacks**

/

**Applicable issues**

/

**Additional information**

Service account is now set to jobs the exact way as it is set on statefulset, so there should not be any issues with this minor change. 

**Checklist** 
- [ X] Title of the PR starts with chart name (e.g. `[bitnami/chart]`)
- [ X] If the chart contains a `values-production.yaml` apart from `values.yaml`, ensure that you implement the changes in both files

### DIFF
--- a/bitnami/redis-cluster/Chart.yaml
+++ b/bitnami/redis-cluster/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: redis-cluster
-version: 3.2.2
+version: 3.2.3
 appVersion: 6.0.7
 description: Open source, advanced key-value store. It is often referred to as a data structure server since keys can contain strings, hashes, lists, sets and sorted sets.
 keywords:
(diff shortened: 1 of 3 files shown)


jurkjurk:
this is exactly what I need to be able to deploy redis chart without workarrounds

avenging:
I've been looking into this as I am having and issue as described, this fix would really help so I don't have to modify the chart myself.

dani8art (approved):
LGTM

PR #3599, opened 2020-09-03
[bitnami/thanos] fix thanos ruler alert label drop

ahmednasir91:
**Description of the change**

With the quotes in place the alert drop doesn't work and the label is included in the data sent to alert manager.

**Benefits**

will be able to drop the de-dupe alerts when running thanos ruler in HA

**Checklist**


carrodher (approved):
Thanks for the PR! LGTM

PR #3609, opened 2020-09-04
[bitnami/metallb] Fixed the PrometheusRule creation issue

utkudarilmaz:
The error fixed that cause of PrometheusRule definition creations. One of two condition deleted that is controlling the PrometheusRule creation. The reason for this .Values.prometheus.enabled not defined on values.yaml and not using on rest of charts.

**Benefits**
PrometheusRules can be deployable anymore.

**Possible drawbacks**
N/A

**Applicable issues**
N/A

**Additional information**
N/A

**Checklist** 


dani8art:
Hi @utkudarilmaz why do you need to remove `prometheus.enabled`? if you want to use these prometheus features you just need to add `--set prometheus.enabled=true` to your deploy command.

utkudarilmaz:
Hi @dani8art, yes i know we can pass this attributes with --set or custom writed x but fresh users don't know this value when firstly downloaded charts. 

When users installing charts they will see PrometheusRules didn't created. The users have to look up the charts to understand why PromethuesRules didn't created. 

So we have two options for fix this issue. First one is adding the .Value.prometheus.enabled value to values.yaml. Second option is deleting the if condition from PrometheusRule chart. In my opinion, second option more simple and elegance than first one.

dani8art:
Yes, I agree with adding it to the `values.yaml` so the users can understand how to use it

bitnami/metallb/README.md
```diff
@@ -109,6 +109,7 @@ The following tables lists the configurable parameters of the metallb chart and
 | `existingConfigMap`                                   | Specify an existing configMapName to use. (this is mutually exclusive with the configInline option)  | `nil`                                                   |
 | `configInline`                                        | Specify the config for metallb as a new configMap inline.                                            | `{}` (does not create configMap)                        |
 | `rbac.create`                                         | Specify if an rbac authorization should be created with the necessarry Rolebindings.                 | `true`                                                  |
+| `prometheus.enabled`                                  | Enable for Prometheus alertmanager basic alerts.                                                     | `true`                                                  |
```
dani8art, line 112: We should keep it disabled by default as we have it now.

```suggestion
| `prometheus.enabled`                                  | Enable for Prometheus alertmanager basic alerts.                                                     | `false`                                                  |
```

bitnami/metallb/values.yaml
```diff
@@ -42,6 +42,8 @@ rbac:
   create: true
 
 prometheus:
+  enabled: true
```
dani8art, line 45:
```suggestion
  enabled: false
```

dani8art (changes requested):
Thanks! Please take a look at my suggestions.

dani8art (approved):
Thank you so much! LGTM

dani8art:
Please, could you fix the linter errors?

utkudarilmaz:
@dani8art Thanks, best regards.

Issue #3605, opened 2020-09-04
[bitnami/cassandra]: wrong reference to configMap


nbenaglia:
In Values there's a section for:

```
## Secret with cql script (with sensitive data). Useful for creating a keyspace
## and pre-populating data
##
## initDBSecret:
```

But secret is not mounted, like show here:
https://github.com/bitnami/charts/blob/374e903fce2e28fcf6a89ab61e3867e415a3d1e7/bitnami/cassandra/templates/statefulset.yaml#L259

Configmap is wrong in L259.

juan131:
You're totally right! I just approved/merged you PR! Thanks so much for the fix.

PR #3612, opened 2020-09-05
[bitnami/cassandra ]Fix volume definition for initDB secret

nbenaglia:
As described in #3605, secret is mount uncorrectly.
This PR fixes the chart.


juan131 (approved):

PR #3613, opened 2020-09-05
[bitnami/ghost] Fix invalid deployment when podAnnotations value is set

mbellgb:
**Description of the change**

When the `podAnnotations` value is set on the ghost helm chart, the resulting manifest for the chart's Deployment resource will be invalid, and Kubernetes won't be able to process it.

**Benefits**

Fixes deployment of the Helm chart. Allows for Istio-related annotations to be set.

**Possible drawbacks**

None that I can think of

**Applicable issues**

n/a

**Additional information**

Thanks for maintaining! Very handy chart to have available.

**Checklist** 

### DIFF
--- a/bitnami/ghost/templates/deployment.yaml
+++ b/bitnami/ghost/templates/deployment.yaml
@@ -12,7 +12,7 @@ spec:
     metadata:
       labels: {{- include "common.labels.standard" . | nindent 8 }}
       {{- if .Values.podAnnotations }}
-      annotations: {{- include "common.tplvalues.render" ( dict "value" .Values.podAnnotations "context" $ ) | nindent 6 }}
+      annotations: {{- include "common.tplvalues.render" ( dict "value" .Values.podAnnotations "context" $ ) | nindent 8 }}
       {{- end }}
     spec:
       {{- include "ghost.imagePullSecrets" . | indent 6 }}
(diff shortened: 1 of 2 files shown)


bitnami/ghost/templates/deployment.yaml
```diff
@@ -11,9 +11,9 @@ spec:
   template:
     metadata:
       labels: {{- include "common.labels.standard" . | nindent 8 }}
-    {{- if .Values.podAnnotations }}
-    annotations: {{- include "common.tplvalues.render" ( dict "value" .Values.podAnnotations "context" $ ) | nindent 6 }}
-    {{- end }}
+      {{- if .Values.podAnnotations }}
+      annotations: {{- include "common.tplvalues.render" ( dict "value" .Values.podAnnotations "context" $ ) | nindent 6 }}
```
carrodher, line 15: Moving the annotation to the correct place, the indentation should be also increased (same approach used for labels):
```suggestion
      annotations: {{- include "common.tplvalues.render" ( dict "value" .Values.podAnnotations "context" $ ) | nindent 8 }}
```

carrodher (changes requested):
Hi, thanks for the PR fixing this issue. Just a minor comment

bitnami/ghost/templates/deployment.yaml
mbellgb: whoops, good spot. thank you!

carrodher (approved):
LGTM!

mbellgb:
thanks! :D

PR #3615, opened 2020-09-06
[bitnami/fluentd] Introduce containerSecurityContext to improve security and compatibility with PodSecurityPolicy

chadlwilson:
**Description of the change**

As a follow-up to #3293 and https://github.com/bitnami/bitnami-docker-fluentd/pull/14 this PR

- introduces ability to configure the container [`securityContext`](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/) for both **aggregator** and **forwarder**
- introduces a secure-by-default configuration of both:
  ```yaml
  privileged: false
  capabilities:
    drop: ['ALL']
  ```
- allows users to extend the existing pod `securityContext` with anything valid within K8s API
  - _use case_: improve forward compatibility, rather than only supporting `runAsUser` and `fsGroup` as in the previous configuration, e.g `sysctls`.
- introduces ability to mount `extraVolumes` within the **aggregator**
  - _use case_: supports mounting an `emptyDir` to `/tmp` so you can run with `readOnlyRootFilesystem: true` `securityContext` which restricted PSPs often require

**Benefits**

As above
- improves the "secure-by-default" configuration of the bitnami fluentd installation. I believe this to be particular important for the **forwarder** installation, which runs as `root` **by default**. "Opting in" to drop all special Linux capabilities and prevent the container from running privileged or escalating privileges mitigates the risk of running as `root` further. 
- prepares us for introduction of optional included `PodSecurityPolicy`s with the chart, if we want to go that direction, like some other bitnami charts do. Users of the `forwarder` in particular could possibly benefit from this, given the ability to restrict `hostPath` mounts via such a policy.

**Possible drawbacks**

In my testing, I _can't_ think of or find any use cases where this more locked down configuration causes issues with pod scheduling, pod startup or operation.

I have validated the out-of-the-box configuration for
- aggregator running non-root as fluentd/1001
- forwarder running as root
- forwarder running non-root as fluentd/1001

I have NOT specifically catered for running the forwarder **container** as root, but with **process** `daemonUser=fluentd` (i.e stepping down privileges after container start). I can't understand what the use case would be for this on Kubernetes, unless there is some "start time" scripting uses cases I am not aware of.

Nevertheless, if we want to support this better, it would require the below, as documented in bitnami/bitnami-docker-fluentd#13

```yaml
containerSecurityContext:
  capabilities:
    add: ["CHOWN", "SETUID". "SETGID"]
```

If we feel this is a valid/common configuration I could either add these capabilities automatically via Helm templating (convenient, but possibly seems a bit dodgy) or document them.

**Additional information**

An alternative approach to **part of** this PR would be to not introduce any opinionated `containerSecurityContext` at all, and instead go straight to introducing an optional `PodSecurityPolicy`, however actively dropping unnecessary permissions in a `securityContext` does have some benefits, especially when running in clusters that have a lenient policy available to be used by the pod which might otherwise be selected.

**Checklist** 


bitnami/fluentd/README.md
```diff
@@ -68,6 +68,9 @@ The following tables lists the configurable parameters of the fluentd chart and
...
 | `forwarder.securityContext.runAsUser`           | User ID for forwarder's containers                                                                             | `0`                                                                                                     |
+| `forwarder.securityContext.*`                   | Other pod security context to be included as-is in pod spec                                                    | `{}`                                                                                                    |
+| `forwarder.containerSecurityContext.enabled`    | Enable security context for the forwarder container                                                            | `true`                                                                                                  |
+| `forwarder.containerSecurityContext.*`          | Other container security context to be included as-is in pod spec                                              | `{ "privileged" : false, "allowPrivilegeEscalation": false, "capabilities": { "drop": ["ALL"] } }`      |
```
javsalgar, line 73: I have my concerns with this. If we create a custom values.yaml file like this

```
cat myvalues.yaml

containerSecurityContext:
  capabilities:
    add: ["CHOWN", "SETUID". "SETGID"]
```

The chart would take the default `drop` section as well (because it is in the default values.yaml file). Would that cause conflicts?

chadlwilson: I'm not sure I understand your concern correctly, but with `capabilities` but it is a common pattern that you `drop: ["ALL"]` and then add back only ones you need, so it shouldn't cause a conflict between the `add` and the `drop` when Helm merges them since these are processed by the underlying runtime, e.g Docker/containerd. e.g https://github.com/bitnami/charts/blob/01774ba8cd5984a9db27b0d3c411f398146ccc75/bitnami/nginx-ingress-controller/templates/controller-deployment.yaml#L95-L100

Since the capability "directives" (add/drop) are in dictionaries, Helm will merge fine. e.g

The below works fine and pods start up OK.
```yaml
forwarder:
  daemonUser: fluentd
  daemonGroup: fluentd
  securityContext:
    runAsUser: 1001
    fsGroup: 1001

  containerSecurityContext:
    capabilities:
      add: [ "CHOWN", "SETUID", "SETGID" ]
```
image


But since the capabilities themselves are an array, which are an override with helm, you can also effectively disable the capability dropping entirely if necessary with a custom values file like:

```yaml
forwarder:
  containerSecurityContext:
    capabilities:
      drop: []
```

javsalgar: Great! That's solves my concern

bitnami/fluentd/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: fluentd
-version: 2.0.2
+version: 2.0.3
```
javsalgar, line 3: Please bump the minor version as it is adding a feature

bitnami/fluentd/values-production.yaml
```diff
@@ -13,7 +13,7 @@ global: {}
 image:
   registry: docker.io
   repository: bitnami/fluentd
-  tag: 1.11.2-debian-10-r18
+  tag: 1.11.2-debian-10-r19
```
javsalgar, line 16: Could you explain the reason for bumping this version?

javsalgar (approved):
LGTM! Minor changes

bitnami/fluentd/values-production.yaml
chadlwilson: My apologies - I originally had a comment in the PR request to explain this when it was draft, but removed it. Oops - sorry.

To work correctly for the **forwarder** running as `root` with no special Linux capabilities, this change is dependent on a couple of changes to the start scripts merged in https://github.com/bitnami/bitnami-docker-fluentd/pull/14 - which is only released onto Docker Hub in `-r19` of the Docker image. I was waiting for the `bitnami-bot` automation to update this version in the chart automatically, but that doesn't seem to have happened yet. I'm not sure what cycle/process these bot-tag-updates are made on.

Am I best to wait for that automation to flow through and make this change for me, and we merge after that?

bitnami/fluentd/Chart.yaml
chadlwilson: Done :-)

bitnami/fluentd/values-production.yaml
javsalgar: Ok, thanks for letting me know!

javsalgar (approved):

Issue #3486, opened 2020-08-21
[bitnami/harbor] add support for cabundlesecretname to match upstream harbor chart


paulczar:
The upstream harbor chart has a [value](https://github.com/goharbor/harbor-helm/commit/782a05418a79d4bc997f03224d0948a462f6b7e4) `caBundleSecretName` that enables you to provide a custom ca bundle, this allows you to use endpoints for auth and other things that are TLS, but self signed.

I am using internally hosted keycloak to provide auth, and I want to be able to use self signed TLS certs, porting this value and feature from the upstream chart would allow this.

andresbono:
Hi @paulczar, thanks for opening this issue. Is this what you are looking for?

https://github.com/bitnami/charts/tree/master/bitnami/harbor#persistence-parameters

| Parameter                                                     | Description                                                                                                                                                                                                                                                                                                                                     | Default         |
|---------------------------------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|-----------------|
| `persistence.imageChartStorage.caBundleSecretName`            | Specify the `caBundleSecretName` if the storage service uses a self-signed certificate. The secret must contain keys named `ca.crt` which will be injected into the trust store  of registry's and chartmuseum's containers.                                                                                                                    |                 |

paulczar:
No, this is a harbor wide root setting, not one specific to chart storage.  See the link I posted in the original issue (or here - https://github.com/goharbor/harbor-helm/commit/782a05418a79d4bc997f03224d0948a462f6b7e4 ) for the commit in which it was implemented upstream.

mkuendig:
Same here. Would love to get that parameter. We use mTLS everyhwere and quering LDAP needs that internal CA cert. I have set persistence.imageChartStorage.caBundleSecretName and I still get the LDAP TLS verification error.

andresbono:
Thanks for the clarification @paulczar! So yes, it seems a nice feature to support.

Would you be interested on contributing with a PR to add the new feature @paulczar, @mkuendig?

paulczar:
Oh its actually much worse.  I just put together the PR, and in testing it still doesn't work.  

[...]

PR #3627, opened 2020-09-07
[bitnami/harbor] Fix and standardize internalTLS

### LINKED ISSUE #3489: [bitnami/harbor] Several containers fail to start when internalTLS is enabled
**Which chart**:
Harbor v7.1.0

**Describe the bug**
When enabling internalTLS, several containers fail to start.

The chartmuseum container enters a crashloop with this error: FATAL	open /bitnami/certs/server.crt: no such file or directory
The secret being mounted for ChartMuseum is created, but has no server.crt key.

The jobservice container enters a crashloop with this error: Failed on load rest config err:Get https://harbor-core:443/api/internal/configurations: x509: certificate signed by unknown authority, url:https://harbor-core:443/api/internal/configurations
It does not trust the automatically generated certificate that the core is using.

The harbor-portal container enters a crashloop, hanging at ** Starting NGINX **

The trivy container never starts, hanging at: {"addr":":8443","certificate":"/bitnami/certs/tls.crt","clientCAs":"/bitnami/certs/ca.crt","key":"/bitnami/certs/tls.key","level":"debug","msg":"Starting API server with TLS","time":"2020-08-22T10:32:53Z"}

**To Reproduce**
1. Install harbor with internalTLS enabled.

**Expected behavior**
All containers will start normally when using internalTLS

joancafom:

**Description of the change**
This PR is aim at fixing the issues that caused the Chart to fail when enabling `internalTLS`,  as detailed in #3489 .

The internalTLS makes the internal components to communicate with each other securely. The components that support this feature are:

- Chartmusem
- Clair
- Core
- JobService
- Portal
- Registry
- Trivy

We need to update their respective docker images as in order to support this feature, the internal CA issuing the certificates needs to be trusted. 

Also, paths regarding internal TLS certificates have been standardized to match upstream, as well as the rest of the paths this chart uses do. This new paths are:

-  `/etc/harbor/{tls, ssl}/` --> Certificates and Keys for internal TLS can be found here
-  `/harbor_cust_cert/` --> CAs that need to be trusted should be placed here. We should adapt the containers to make sure they take them into account.
  

**Benefits**
Internal components can make use of a secure connection to communicate. New directory that allows the installation of custom CAs that can be used in #3486.

**Possible drawbacks**
Not really, this feature is not working at the moment.

**Applicable issues**

  - fixes #3489

**Additional information**

**Checklist** 


bitnami/harbor/templates/core/core-cm.yaml
```diff
@@ -13,8 +13,8 @@ metadata:
...
   app.conf: |+
     appname = Harbor
-    runmode = dev
+    runmode = prod
     enablegzip = true
 
-    [dev]
-    httpport = 8080
+    [prod]
+    httpport = {{ include "harbor.core.containerPort" . | quote }}
```
juan131, line 20: Why do we want to change the default run mode?

juan131 (commented):
Thanks for the changes!! Is it WIP yet? I don't see anything that worries me but please test the upgrade path with these changes, we should be extremely careful to avoid introducing backwards incompatibilities.

bitnami/harbor/templates/core/core-cm.yaml
joancafom: I just change this in order to match the upstream repo (which used to use `runmode=dev` as well):

https://github.com/goharbor/harbor-helm/blob/f5767efaa35f835934ce060aa467b3a00af82837/templates/core/core-cm.yaml#L8

joancafom:
Yes @juan131 , there are a couple of things I am still working on:

1. Check for README in case we need to update
2. Make sure upgrading works as expected, so we can go with a Minor instead of a Major
3. Implement changes in the aforementioned containers and wait for their release. Without the new images this PR would not work.

juan131:
Please ping me once I should review the changes again

joancafom:
Ping @juan131 

This is ready to be reviewed again 😁 ! Tested upgrading in Minikube cluster and did not find issues, hence going for a Minor.

juan131 (approved):
LGTM

Issue #3629, opened 2020-09-08
[bitnami/etcd] add loadBalancerSourceRanges (feature request)


rico-bincentive:
**Which chart**:
etcd

**Is your feature request related to a problem? Please describe.**
Must of charts already support `loadBalancerSourceRanges`.
In my scenario, etcd is our configuration management service. It does not allow all of the traffic to access.

**Describe the solution you'd like**
Add `loadBalancerSourceRanges` setting for etcd chart.

**Describe alternatives you've considered**
Would consider trying `service.beta.kubernetes.io/aws-load-balancer-extra-security-groups` annotation on ELB.

**Additional context**
I can help this feature. 👋

javsalgar:
Hi,

Thank you very much for the input! We will work on that but we cannot guarantee an ETA. However, if you want to speed up the process feel free to create a PR and we will review it.

PR #3637, opened 2020-09-09
[bitnami/etcd] add loadBalancerSourceRanges

rico-bincentive:
**Description of the change**

add `loadBalancerSourceRanges` new option.

**Benefits**

Restrict only accessible traffic can access etcd service from the load balancer.

**Possible drawbacks**

no.

**Applicable issues**

  - feature request #3629

**Checklist** 


bitnami/etcd/templates/svc.yaml
```diff
@@ -10,6 +10,12 @@ spec:
   type: {{ .Values.service.type }}
   {{- if and (eq .Values.service.type "LoadBalancer") (not (empty .Values.service.loadBalancerIP)) }}
   loadBalancerIP: {{ .Values.service.loadBalancerIP }}
+  {{- end }}
+  {{- if and (eq .Values.service.type "LoadBalancer") (not (empty .Values.service.loadBalancerSourceRanges)) }}
+  loadBalancerSourceRanges:
+  {{- with .Values.service.loadBalancerSourceRanges }}
```
javsalgar, line 16: Is the `with` instruction necessary? wouldn't using toYaml with the array be enough?

bitnami/etcd/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: etcd
-version: 4.10.1
+version: 4.10.2
```
javsalgar, line 3: As it is adding a feature, please bump the minor version

bitnami/etcd/templates/svc.yaml
rico-bincentive: Thank for the feedback. You are right. But I saw some of chart use this pattern, some of not.
use `with`:
https://github.com/bitnami/charts/blob/master/bitnami/redis/templates/redis-master-svc.yaml#L23-L28
https://github.com/bitnami/charts/blob/master/bitnami/influxdb/templates/influxdb/service.yaml#L16-L21
just use `toYaml`:
https://github.com/bitnami/charts/blob/master/bitnami/external-dns/templates/service.yaml#L19-L21
https://github.com/bitnami/charts/blob/master/bitnami/thanos/templates/compactor/service.yaml#L19-L21

javsalgar: Yes, we are trying to normalize it Could you go for toYaml instead?

rico-bincentive: sure

javsalgar (approved):
LGTM! Thanks

PR #3648, opened 2020-09-10
[bitnami/kube-prometheus] Bump thanos sidecar to 0.15

hwoarang:
Thanos chart was bumped to use the 0.15 version of Thanos so we should
do the same with Prometheus.

**Checklist** 


carrodher:
Hi @hwoarang, thanks for letting us know. This is something done automatically by the Bitnami CI/CD pipeline, at this moment we already released the thanos 01.5 docker image and it should be replaced in the `bitnami/kube-prometheus` chart sooner rather than later with the proper testing process. So unfortunately we need to close the PR and wait for the pipeline. Thanks for the ping, we are going to manually check the automation just in case something is delaying this change

PR #3652, opened 2020-09-10
[bitnami/thanos] Fixes for Ingresses of Querier and Bucketweb in Thanos Chart

Xtigyro:
**Description of the change**

Fixes for the Ingress API objects of Querier and Bucketweb in the Thanos chart.

**Checklist** 

### DIFF
--- a/bitnami/thanos/templates/querier/ingress.yaml
+++ b/bitnami/thanos/templates/querier/ingress.yaml
@@ -2,7 +2,7 @@
 apiVersion: extensions/v1beta1
 kind: Ingress
 metadata:
-  name: {{ include "thanos.fullname" . }}
+  name: {{ include "thanos.fullname" . }}-querier
   labels: {{- include "thanos.labels" . | nindent 4 }}
   annotations:
     {{- if .Values.querier.ingress.certManager }}
@@ -33,7 +33,7 @@ spec:
     {{- end }}
   {{- if or .Values.querier.ingress.tls .Values.querier.ingress.extraTls .Values.querier.ingress.hosts }}
   tls:
-    {{- if .Values.querier.ingress.tls }}
+    {{- if .Values.querier.ingress.secrets }}
     - hosts:
         - {{ .Values.querier.ingress.hostname }}
       secretName: {{ printf "%s-tls" .Values.querier.ingress.hostname }}
--- a/bitnami/thanos/Chart.yaml
+++ b/bitnami/thanos/Chart.yaml
@@ -1,5 +1,5 @@
 apiVersion: v1
-version: 2.3.4
+version: 2.3.5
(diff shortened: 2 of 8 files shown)


javsalgar:
Hi,

Thank you very much for the PR! Do you have a GH issue where you explain the context of the PR? If not, could you add it here?

Xtigyro:
@javsalgar There are missing parts of the Ingress implementation for Querier and Bucketweb which were probably by mistake not included. The PR fixes those.

bitnami/thanos/templates/querier/ingress.yaml
```diff
@@ -2,7 +2,7 @@
 apiVersion: extensions/v1beta1
 kind: Ingress
 metadata:
-  name: {{ include "thanos.fullname" . }}
+  name: {{ include "thanos.fullname" . }}-querier
```
hwoarang, line 5: This is a breaking change (imho) for tools that connect to that ingress so I think we either need to bump the minor version or mention this in the README file.

hwoarang: Also is there a real need to rename the ingress?

Xtigyro: No pain - no gain. There's an Upgrade note in the `README` now and the chart was bumped to `2.4.0`. 🌞

ahmednasir91:
Thanks for this fix @Xtigyro the thanos query ingress name

bitnami/thanos/templates/querier/ingress.yaml
javsalgar: Hi,

Could you elaborate why renaming the ingress would break other tools? I would like to know an example.

hwoarang: @javsalgar no particular example in mind. I was mostly thinking external tooling that treats the object an as API. For example, think of a script that fetches the ingress object and extracts some info from the json like
`kubectl get ing thanos -o json | jq ....` . I am not sure really if we should be treating the objects themselves as part of some API that needs to follow the same semver rules.

Xtigyro: Yeah - no obvious issues with the name change of the resource.

Even so it's still a change of the name of a resource. So should we keep the minor version bumped? I think so.

javsalgar: In my opinion, I think that the major change should focus on the chart itself. I'm not sure until what extent we should guarantee that no tool around would break. I also believe it should be a minor change in this case. However, thanks for the input, I will forward it to the rest of the team for future retrospective

javsalgar (approved):

PR #3663, opened 2020-09-12
[bitnami/nginx-ingress-controller] Make default backend health check path configurable

longwa:
**Description of the change**

This change makes the liveness and readiness probe paths configurable for a custom default backend. 

The default is unchanged (still `/`) so this will not be a breaking change for anyone.

**Benefits**

When using an image such as `quay.io/kubernetes-ingress-controller/custom-error-pages-amd64` (among others), the health check is served at `/healthz` according to what the ingress-controller docs specify. The image can't be deployed with this chart without manually modifying the health check after deployment since `/` will serve a 404 by default (thus failing the health checks).

**Possible drawbacks**

None

**Applicable issues**

None

**Checklist** 

### DIFF
--- a/bitnami/nginx-ingress-controller/Chart.yaml
+++ b/bitnami/nginx-ingress-controller/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: nginx-ingress-controller
-version: 5.5.1
+version: 5.5.2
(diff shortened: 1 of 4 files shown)


rafariossaa (approved):
Hi,
  Thank you very much for your PR! This is very nice to have.

PR #3664, opened 2020-09-13
[bitnami/fluentd] Introduce a bundled PodSecurityPolicy for the fluentd forwarder

chadlwilson:
**Description of the change**

This change
- makes `runAsGroup` explicit for both **aggregator** and **forwarder** (i don't see a reason the aggregator runs in the root group by default)
- introduces a PodSecurityPolicy for the forwarder
- attempts to document some of the discussion from #1905 in the README

The new PodSecurityPolicy
- is **disabled** by default
- allows all major volume types (including `persistentVolumeClaim` - this is debateable)
- whitelists only specific `hostPath`s
- adapts to whether the user is running as `root`
- is consistent with other PSPs in the `bitnami/charts` collection
- ensures all other capabilities are dropped; prevents usage of `hostIPC`, `hostPID`,`hostNetwork`,`hostPorts` etc.
- will cause the `DaemonSet` to rolling update automatically when the PSP is changed

**Benefits**

- Some clusters allow tenants to only create unprivileged pods without a PodSecurityPolicy. Since this chart both requires `hostPaths` (which have their own risks) and runs as root by default, a custom targeted `PodSecurityPolicy` is useful.
- Can serve as a template for users to document the requirements of the chart/software, even if they need to customise it themselves.

**Possible drawbacks**

- It may seem a bit strange to create a PodSecurityPolicy which is dynamic as to whether you are running as root or not.
- Should we recommend it is enabled in `values-production.yaml` as I have currently done? Not sure about this.

**Additional information**

- Currently, the forwarder requires `readOnlyRootFilesystem: false` because the start scripts write into `/tmp`. There is a workaround for this using [the standard `emptyDir` mounting approach that the **memcached** chart uses](https://github.com/bitnami/charts/blob/b483f7c1f1805927f9e3f024a42472b52ec322e9/bitnami/memcached/templates/deployment.yaml#L75-L77), however I am not sure if it would be a good idea to add a default `emptyDir` mount to `/tmp` at this stage.

User workaround:
```yaml
forwarder:
  containerSecurityContext:
    readOnlyRootFilesystem: true
  extraVolumes:
    - name: tmpdir
      emptyDir: {}
  extraVolumeMounts:
    - name: tmpdir
      mountPath: /tmp
```

**Testing**

I have tested a number of scenarios with the default bundled fluentd configuration
- tested chart validation rules for `forwarder.rbac.pspEnabled`
- pods come up and work correctly with default configuration
- pods come up and work correctly when forced to run as `fluentd`/`1001` user with configuration documented in `README`
- pods come up and work correctly with `readOnlyRootFilesystem: true` workaround above

**Checklist** 


bitnami/fluentd/README.md
```diff
@@ -299,6 +308,52 @@ aggregator.extraEnv[1].name=ELASTICSEARCH_PORT
...
+
+The default behaviour is to run as `root` because
+- the forwarder needs to mount `hostPath` volumes from the underlying node to read Docker container (& potentially other) logs
+- in many Kubernetes node distributions, these log files are not readable by anyone other than `root`
+- `fsGroup` doesn't work with `hostPath` volumes to allow the process to run non-root with alternate file system permissions
+
+Since we would like the chart to work out-of-the-box for as many users as possible, the `forwarder` thus runs as root 
+by default. You can read more about the motivation for this at [#1905](https://github.com/bitnami/charts/issues/1905) 
+and [#2323](https://github.com/bitnami/charts/pull/2323), however you should be aware of this, and the risks of running
+root containers in general.
```
dani8art, line 325: Why a new line here?

bitnami/fluentd/README.md
```diff
@@ -299,6 +308,52 @@ aggregator.extraEnv[1].name=ELASTICSEARCH_PORT
...
+it will allow the pod to run as `root` by default, while ensuring as many other privileges as possible are dropped.
+
+#### Running as non-root
+
+You can run as the `fluentd` user/group (non-root) with the below overrides if
+- you have control of the `hostPath` filesystem permissions on your nodes sufficient to allow the fluentd user to read from them
+- don't need to write to the `hostPath`s
+
+Note that if you have enabled the [bundled PodSecurityPolicy](templates/forwarder-psp.yaml), it will adapt to the Chart
+values overrides.
```
dani8art, line 337: no newline needed, right?

dani8art (approved):
Hi @chadlwilson , thank you so much for this PR

LGTM, just minor comments.

bitnami/fluentd/README.md
chadlwilson: Yeah, possibly my bad - I use a loose (i.e inconsistent - eek) 120 char line-wrap indicator in my IDE (except for tables) to improve readability when editing the markdown.

Is there a particular guide for line wrap to use? Perhaps you don't add new-line wraps at all if I am to infer from the existing format?

dani8art: Yes, we don't apply a new-line format by now, we add it when actually it is a new-line.

chadlwilson: OK, no problems. Fixed now.

dani8art (approved):
LGTM!

chadlwilson:
Thanks @dani8art, much appreciated!

If you or the team have any feedback on my open thoughts above I am happy to submit further PRs.

PR #3665, opened 2020-09-14
[bitnami/contour] Change CRDs approach to handle CRD upgrades

georgegoh:
**Description of the change**

Helm does not perform any actions on CRDs when a chart is upgraded. Functionality is desired to upgrade CRDs when the project itself is updated.

**Benefits**

Placing the CRDs in a subchart causes them to be processed when the
contour chart is upgraded.

**Possible drawbacks**

The upgrade process applies only to the CRD definition and does not perform any schema migration of existing objects created under the older CRD.

**Additional information**

There have been discussions about this approach on the upstream project contour - https://github.com/projectcontour/contour/issues/2050#issuecomment-673722738

**Checklist** 


bitnami/contour/charts/contour-crds/Chart.yaml
```diff
@@ -0,0 +1,23 @@
+apiVersion: v1
+name: contour-crds
```
mkilchhofer, line 2: Are you sure, that we need a subchart for this? As far as I understand about helm and CRDs, the real magic about  CRD management belongs to the `/crds/`-Directory in helm3 and the `crd-install`-hook in helm2.

When we do not apply resources with this contour-chart which belongs to the definitions we need to apply before, I don't see the usage of a subchart for the CRDs.

There are Pros and Cons for managing CRDs with helm using the `/crds/`-Directory / `crd-install`-hook. When the CRD is applied by helm using /crds/ or hooks, the CRD does not get deleted if we uninstall the release. Otherwise it gets deleted including all "instances" of this type (cascade delete) which could be dangerous.

dani8art:
Hi,

Taking into account the main goal of this PR which is to be able to update CRDs of the contour chart when performing upgrade actions, we would like to use this PR as a place where to find the best approach for managing CRDs in contour and it might be extended as a bitnami "standard".

Summarizing, we could talk about two proposed options:

- Helm official and supported [best practices](https://helm.sh/docs/chart_best_practices/custom_resource_definitions/)

- CRDs as a subchart

- ~~Same chart but outside `crds/` directory~~, this option is not possible since we need to ensure CRDs are applied before everything.

**Helm official**
This approach would be having a `crds/` directory (since helm2 is going to be deprecated sooner than later)

1. (pro): the CRDs does not get deleted if we uninstall the release
2. (pro): official support from the Helm team, if they introduce a new feature where the updates are available then we will have it prepared.  
3. (cons): the CRDs do not get updated when performing an upgrade action

**Subchart**

1. (pro): the CRDs get updated when performing an upgrade action
2. (cons): the chart version would have to be separately managed 
3. (cons): the CRDs get deleted if we uninstall the release, as a consequence, we will lose all the definitions in the cluster base on this CRDs (cascade delete)
4. (cons): schema migrations between CRDs version must be done manually

Solving cons
- [3] A solution could be to use `pre-install` and `pre-upgrade` hooks for all the templates inside this subchart so they do not get deleted if we uninstall the release. Do you think it could work?

@georgegoh, @mkilchhofer would you mind adding your concerns and expectations?

bitnami/contour/charts/contour-crds/Chart.yaml
dani8art: We could simply move the CRDs outside the `crds/` directory in the same chart, this does not work since we need to ensure CRDs are applied before everything.

mkilchhofer: That is true for charts applying resources which belongs to the CRDs the chart introduces. The prometheus-operator chart is a good example for this.
But the contour chart currently do not apply a resource to the K8s-API which is an "instance" of the CRD it applies. The CRDs are used by other applications where the Ingress-v1-API does not provide enough features. Contour currently does not enter healthy/ready state when the CRDs are absent, but it get healthy at runtime as soon the CRDs exist. I maintain the CRDs outside of helm, so I have full control when new features arrive.

georgegoh: I pulled out the CRDs into a subchart for a couple reasons:

1. Helm does not currently manage CRD lifecycle.
2. Subcharts will always run first. Even if contour does not instantiate CRs as part of its installation, I wanted to codify that order.

Granted there are risks in this approach as you mentioned - and I'm grateful that you raised them. I'll be happy to collaborate to try to get a solution we can all agree on.

georgegoh:
**Helm official**
(`/crds` directory)
1. (con): CRDs remain in the newer version in a chart rollback.

**Subchart**
1. (unknown): What is the CRD behavior(install, update, deletion) if multiple versions of contour are deployed concurrently?

[1] https://github.com/helm/community/blob/f9e06c16d89ccea1bea77c01a6a96ae3b309f823/architecture/crds.md

dani8art:
> 1. (con): CRDs remain in the newer version in a chart rollback.

It would be actually the first version installed always, I mean, in a rollback, in an upgrade and unistall/install operation, right?

> 2. (unknown): What is the CRD behavior(install, update, deletion) if multiple versions of contour are deployed concurrently?

This is a good question, we should play a bit more using the subchart approach an see what is the result of this.

dani8art:
Hi, I've just implemented a small chart and subchart to test some of the questions/solutions to problem [3]. Unfortunately having a subchart doesn't apply the crds to my k8s cluster, in fact, it throws this error.

```
$ helm install crds-tests contour
Error: unable to build kubernetes objects from release manifest: error validating "": error validating data: [ValidationError(CustomResourceDefinition.spec.validation.openAPIV3Schema.properties.status.properties.conditions): unknown field "x-kubernetes-list-map-keys" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1beta1.JSONSchemaProps, ValidationError(CustomResourceDefinition.spec.validation.openAPIV3Schema.properties.status.properties.conditions): unknown field "x-kubernetes-list-type" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1beta1.JSONSchemaProps]
```

If I move crds into a `crds/` folder in the subchart they get created. 

@georgegoh do you have any hint?

mkilchhofer:
> @georgegoh, @mkilchhofer would you mind adding your concerns and expectations?

I've not that much experience with helm and CRDs. When I use a "critical" chart, I always specify `--skip-crds` and apply the needed CRD separately (i.e. manually or via gitops).
- A hacky way to ensure all CRDs are updated correctly during a helm upgrade could be to render the CRDs as a ConfigMap which is mounted into a "kubectl job".
- Another way is to place the CRDs into the "normal" templates directory and apply them as normal resources. I think jetstacks cert-manager works like this. To prevent data loss we could add the `"helm.sh/resource-policy": keep` annotation (configurable through a value)

georgegoh:
> Hi, I've just implemented a small chart and subchart to test some of the questions/solutions to problem [3]. Unfortunately having a subchart doesn't apply the crds to my k8s cluster, in fact, it throws this error.
> 
> ```
> $ helm install crds-tests contour
> Error: unable to build kubernetes objects from release manifest: error validating "": error validating data: [ValidationError(CustomResourceDefinition.spec.validation.openAPIV3Schema.properties.status.properties.conditions): unknown field "x-kubernetes-list-map-keys" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1beta1.JSONSchemaProps, ValidationError(CustomResourceDefinition.spec.validation.openAPIV3Schema.properties.status.properties.conditions): unknown field "x-kubernetes-list-type" in io.k8s.apiextensions-apiserver.pkg.apis.apiextensions.v1beta1.JSONSchemaProps]
> ```
> 
> If I move crds into a `crds/` folder in the subchart they get created.
> 
> @georgegoh do you have any hint?

@dani8art Do you have a copy of your test charts where I can view?

dani8art:
Hi @georgegoh find attached the test charts, 

Thanks!

[contour-tests.tar.gz](https://github.com/bitnami/charts/files/5254580/contour-tests.tar.gz)

dani8art:
No stale

georgegoh:
> Hi @georgegoh find attached the test charts,
> 
> Thanks!
> 
> [contour-tests.tar.gz](https://github.com/bitnami/charts/files/5254580/contour-tests.tar.gz)

Hey @dani8art I'm sorry I took such a long time to get back to you.

I tested your tarball and the CRDs install fine(I assume the rest of the template directory was dummy yamls):

```
bash-5.0$ kubectl get crds
No resources found in projectcontour namespace.
bash-5.0$ helm install test-crd contour -n projectcontour
NAME: test-crd
LAST DEPLOYED: Fri Dec 11 13:03:35 2020
NAMESPACE: projectcontour
STATUS: deployed
REVISION: 1
TEST SUITE: None
NOTES:
1. Get the application URL by running these commands:
  export POD_NAME=$(kubectl get pods --namespace projectcontour -l "app.kubernetes.io/name=contour,app.kubernetes.io/instance=test-crd" -o jsonpath="{.items[0].metadata.name}")
  echo "Visit http://127.0.0.1:8080 to use your application"
  kubectl --namespace projectcontour port-forward $POD_NAME 8080:80
bash-5.0$ kubectl get crds
NAME                                          CREATED AT
httpproxies.projectcontour.io                 2020-12-11T05:03:35Z
tlscertificatedelegations.projectcontour.io   2020-12-11T05:03:35Z
bash-5.0$ helm uninstall test-crd
release "test-crd" uninstalled
bash-5.0$ kubectl get crds
No resources found in projectcontour namespace.
```

Here's my environment:

```
bash-5.0$ kubectl version
Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.3", GitCommit:"2e7996e3e2712684bc73f0dec0200d64eec7fe40", GitTreeState:"clean", BuildDate:"2020-05-21T14:51:23Z", GoVersion:"go1.14.3", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.2", GitCommit:"52c56ce7a8272c798dbc29846288d7cd9fbae032", GitTreeState:"clean", BuildDate:"2020-04-30T20:19:45Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}
```

dani8art:
Oh yeah! I think it was the environment then I had this

```
Client Version: version.Info{Major:"1", Minor:"16+", GitVersion:"v1.16.13-dispatcher", GitCommit:"fd22db44e150011eccc8729db223945384460143", GitTreeState:"clean", BuildDate:"2020-07-24T07:27:52Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"16+", GitVersion:"v1.16.15-gke.4300", GitCommit:"7ed5ddc0e67cb68296994f0b754cec45450d6a64", GitTreeState:"clean", BuildDate:"2020-10-28T09:23:22Z", GoVersion:"go1.13.15b4", Compiler:"gc", Platform:"linux/amd64"}
```

mkilchhofer:
As I now have some more experience with CRDs and Helm I would recommend the following:

As long as contour installation (w/o the operator) does not require to install the CRDs (1) AND place a resource of that kind (2), I would simply move the CRD yamls to the template folder without any hooks.

For the "newly" created contour-operator, this would be different. To create a chart for the operator, you will need such a behavior like this PR aims to achieve.

dani8art (changes requested):
That's great! thank you so much, just a minor comment. Could we move the upgrade notice to this section https://github.com/bitnami/charts/tree/master/bitnami/contour#upgrading, please?

mkilchhofer:
When we proceed with this PR (what I do not reccomend as mentioned above), I have these additional comments:
- Subcart is not defined in `Chart.yaml` as a dependency, please define it with a conditional like When we proceed with this PR (what I do not reccomend as mentioned above), I have these 
  ```yaml
  - name: contour-crds
    tags:
      - bitnami-common
    condition: contour.installCRDs
  ```
- Update README comment regarding `contour.installCRDs`
- Place Info about uninstallin the release in README: `helm uninstall <myrelease>` will lead to data loss now, when CRDs ar managed by helm since we do not use a `keep`-Policy

oliverbaehler:
Hi there

I also want to jump into the discussion before this is merged. Since this might be a standard applied to all bitnami charts. I have to agree with @mkilchhofer, it seems like way too much overhead to have a subchart just for managing the CRDs. imagine having this for each charts which potentially implements CRDs. I would recommend having a resources/ directory next to your templates directory and reading all the files. This approach can also be seen in other charts (e.g. https://github.com/banzaicloud/banzai-charts/blob/master/prometheus-operator-standalone/templates/crds.yaml). Seems much more adjustable and like less overhead than maintain subcharts. Please consider before merging.

Thanks! :)

georgegoh:
@mkilchhofer @oliverbaehler thanks for your comments. I understand where you're coming from and will test them out before any further action is taken here.

dani8art:
Hi all, 

We would like to clarify that sub chart approach would not be taken as the standard for our charts. The aims of this PR was to find a solution that fits the contour specific requirements, and this drive into a discussion of pros and cons about using sub chart that opened the possibility of using it as standard. After a further review, investigation, and tests we decided to not follow it for the rest of the charts in the repository but for contour specifically. 

Sorry if this conclusion was not shared here. @mkilchhofer @oliverbaehler thanks for the feedback we will take it into account and as @georgegoh said we will perform more tests before taking any further action.

Thanks!

georgegoh:
@oliverbaehler I tried the approach you suggested and it's working well in my test:

1. Install with CRDs using a flag(manageCRDs) in values.yaml.
2. Create `HTTPProxy` object and verify access.
3. Modify CRD by adding new property and run `helm upgrade`.
4. Verify existing `HTTPProxy` object still exists and works.
5. Validate that the CRD is indeed updated by visual inspection on `kubectl get -o yaml crd httpproxies.projectcontour.io`

I think this approach is definitely workable.

pablo-ruth:
@georgegoh if you uninstall the Contour release, does it remove CRDs and so delete all HTTPProxies defined in namespaces in the cluster ? Because using "crds" directory managed by Helm doesn't remove them to avoid losing all HTTPProxies objects.

georgegoh:
> @georgegoh if you uninstall the Contour release, does it remove CRDs and so delete all HTTPProxies defined in namespaces in the cluster ? Because using "crds" directory managed by Helm doesn't remove them to avoid losing all HTTPProxies objects.

@pablo-ruth yes it does remove the CRDs. The way to override this is to set the values flag `contour.manageCRDs` to `false` when you install.

dani8art:
Cool, let's try it in our pipeline then we can merge it ;)

Thanks

bitnami/contour/values.yaml
```diff
@@ -178,6 +178,9 @@ contour:
   ## Create Contour CRDs
   installCRDs: true
 
+  ## Manage the creation and upgrade of Contour CRDs.
+  manageCRDs: true
+
```
dani8art, line 183: what is the difference between them?

georgegoh: `manageCRDs` if set to true, will install, upgrade and delete CRDs along with the Helm chart. If the user wishes to manage their own CRDs, then they can set this to false.

It is used in `00-crds.yaml` -  
https://github.com/bitnami/charts/pull/3665/files#diff-32819b0a35e1ac9c445114f4fb1358eebaed346268f33652d4bb068507000803

dani8art: Yes, and what about `installCRDs`?

dani8art (commented):
Internal tests have passed and all is green. It is just pending to resolve the conflicts and the small comment about the duplication of `installCRDs` and `managaCRDs`.

Thanks!

dani8art (approved):

bitnami/contour/values.yaml
georgegoh: `installCRDs` are an artefact from the Helm V2 days, is obsoleted and can be removed.

PR #3699, opened 2020-09-17
[bitnami/wordpress] Allowing usage of existing secrets for WordPress, MariaDB and SMTP

fabriciomariani:
**Description of the change**

This change will make possible the usage of an existing Secret in the Wordpress Chart. The current version is forcing the creation of a secret for every new WordPress deployment, and does not allow the usage of an existing Secret to handle the `externalDatabase` password, forcing users to hardcode the database password into the repo when using Helm with any GitOps CI/CD tool, in case we don't opt to have the Chart create the database.
 
**Benefits**

Another benefit this change provide is allowing us to use Bitnami's [SealedSecrets](https://github.com/bitnami-labs/sealed-secrets), making this Chart more secure.

**Possible drawbacks**

**Applicable issues**

  - No issues have been opened related to this fix.

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/wordpress/Chart.yaml
+++ b/bitnami/wordpress/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: wordpress
-version: 9.5.6
+version: 9.5.7
 appVersion: 5.5.1
 description: Web publishing platform for building blogs and websites.
 icon: https://bitnami.com/assets/stacks/wordpress/img/wordpress-stack-220x234.png
--- a/bitnami/wordpress/Chart.yaml
+++ b/bitnami/wordpress/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: wordpress
-version: 9.5.5
+version: 9.5.6
 appVersion: 5.5.1
 description: Web publishing platform for building blogs and websites.
 icon: https://bitnami.com/assets/stacks/wordpress/img/wordpress-stack-220x234.png
--- a/bitnami/wordpress/Chart.yaml
+++ b/bitnami/wordpress/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: wordpress
-version: 9.5.7
+version: 9.6.6
 appVersion: 5.5.1
 description: Web publishing platform for building blogs and websites.
 icon: https://bitnami.com/assets/stacks/wordpress/img/wordpress-stack-220x234.png
--- a/bitnami/wordpress/Chart.yaml
+++ b/bitnami/wordpress/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: wordpress
-version: 9.6.6
+version: 9.6.0
 appVersion: 5.5.1
 description: Web publishing platform for building blogs and websites.
 icon: https://bitnami.com/assets/stacks/wordpress/img/wordpress-stack-220x234.png
--- a/bitnami/wordpress/README.md
+++ b/bitnami/wordpress/README.md
@@ -80,6 +80,7 @@ The following table lists the configurable parameters of the WordPress chart and
 | `image.debug`                             | Specify if debug logs should be enabled                                               | `false`                                                      |
 | `wordpressSkipInstall`                    | Skip wizard installation when the external db already contains data from a previous WordPress installation [see](https://github.com/bitnami/bitnami-docker-wordpress#connect-wordpress-docker-container-to-an-existing-database) | `false`                                                      |
 | `wordpressUsername`                       | User of the application                                                               | `user`                                                       |
+| `existingSecret`                          | Name of the existing Wordpress Secret Object                                          | `nil`               |
 | `wordpressPassword`                       | Application password                                                                  | _random 10 character long alphanumeric string_               |
 | `wordpressEmail`                          | Admin email                                                                           | `user@example.com`                                           |
 | `wordpressFirstName`                      | First name                                                                            | `FirstName`                                                  |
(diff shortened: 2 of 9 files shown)


bitnami/wordpress/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: wordpress
-version: 9.5.6
+version: 9.5.7
```
juan131, line 3: Please bump minor version instead since you're including a new feature

bitnami/wordpress/templates/_helpers.tpl
```diff
@@ -259,16 +259,40 @@ Return the MariaDB User
...
+{{/*
+Return the SMTP Secret Name
+*/}}
+{{- define "wordpress.smtpSecretName" -}}
+{{- if .Values.smtpExistingSecret }}
+    {{- printf "%s" .Values.smtpExistingSecret -}}
+{{- else -}}
+    {{- printf "%s" (include "wordpress.fullname" .) -}}
+{{- end -}}
+{{- end -}}
```
juan131, line 294: Why do we need a second secret I mean, why not using the same "existing secret" both for SMTP and WP password?

bitnami/wordpress/values-production.yaml
```diff
@@ -458,6 +458,13 @@ externalDatabase:
   ##
   password: ""
 
+  ## Use existing secret (ignores previous password)
+  # existingSecret:
+
+  ## Password key to be retrieved from secret if not `mariadb-password`
+  ##
+  # existingSecretPasswordKey:
```
juan131, line 466: You're not using this anywhere, are you?

bitnami/wordpress/values.yaml
```diff
@@ -50,6 +50,10 @@ commonLabels: {}
 ##
 commonAnnotations: {}
 
+## Use existing secret (does not create the WordPress Secret object)
+## must contain key `wordpress-secret`
+# existingSecret:
```
juan131, line 55: Add a comment warning about the consequences:
```
 ## NOTE: When it's set, the `wordpressPassword` parameters is ignored
```

bitnami/wordpress/values.yaml
```diff
@@ -151,6 +155,11 @@ updateStrategy:
 # smtpPassword:
 # smtpUsername:
 # smtpProtocol:
+## Use an existing secret for the SMTP Password
+## Can be the same secret as existingSecret
+## Must contain the key `smtp-password`
```
juan131, line 160: Same here: warn users about `smtpPassword ` being ignored.

bitnami/wordpress/README.md
```diff
@@ -80,6 +80,7 @@ The following table lists the configurable parameters of the WordPress chart and
 | `image.debug`                             | Specify if debug logs should be enabled                                               | `false`                                                      |
 | `wordpressSkipInstall`                    | Skip wizard installation when the external db already contains data from a previous WordPress installation [see](https://github.com/bitnami/bitnami-docker-wordpress#connect-wordpress-docker-container-to-an-existing-database) | `false`                                                      |
 | `wordpressUsername`                       | User of the application                                                               | `user`                                                       |
+| `existingSecret`                          | Name of the existing Wordpress Secret Object                                          | `nil`               |
```
juan131, line 83:
```suggestion
| `existingSecret`   | Name of the existing Wordpress Secret (it must contain a key named `wordpress-password`). When it's set, `wordpressPassword` is ignored   | `nil`               |
```

juan131 (changes requested):
Thanks so much for your PR!! Please take a look to my comments when you have a chance.

Generic observations:

- I think we can merge the SMTP and WP secret in a single one. 
- The **values.yaml** and the **values-production.yaml** are not synced.

bitnami/wordpress/templates/_helpers.tpl
denniswebb: I worked Fab with this. We took this route simply to provide more flexibility. If somebody decided to use an existing secret for the smtp password, but either wanted to use the automatically generated WordPress password or just don't care if it's specified at install.

You can use 3 separate secrets for each of these, or you could use 2 or 1 even. I used this new version already as a test and actually used a single secret for the WordPress secret, SMTP, and MariaDB instead of separating the db secret from the others.

bitnami/wordpress/Chart.yaml
juan131: This is not done, please check it.

bitnami/wordpress/values-production.yaml
juan131: Please check this too

bitnami/wordpress/templates/secrets.yaml
```diff
@@ -11,11 +12,16 @@ metadata:
...
 type: Opaque
 data:
+  {{- if not .Values.existingSecret }}
   {{- if .Values.wordpressPassword }}
   wordpress-password: {{ .Values.wordpressPassword | b64enc | quote }}
   {{- else }}
   wordpress-password: {{ randAlphaNum 10 | b64enc | quote }}
   {{- end }}
+  {{- end }}
+  {{- if not .Values.smtpExistingSecret }}
```
juan131, line 22: Can you merge this condition with the one below?
```
{{- if and .Values.smtpPassword (not .Values.smtpExistingSecret) }}
```

bitnami/wordpress/templates/_helpers.tpl
juan131: I don't have a strong opinion about this. I guess that using one secret per credentials might be more flexible, although harder to maintain.
Feel free to use 3 secrets.

juan131 (changes requested):
Thanks for addressing some of the suggestions! Could you please address the missing ones? Thanks in advance

fabriciomariani:
Thank you for your review (and patience) @juan131. I have made all the requested changes.

bitnami/wordpress/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: wordpress
-version: 9.5.6
+version: 9.6.6
```
juan131, line 3: This should be `9.6.0`

juan131 (commented):
The PR looks good to me! Please fix the version and we'll be ready to land you changes.

bitnami/wordpress/Chart.yaml
denniswebb: Fixed

juan131 (approved):
LGTM

PR #3687, opened 2020-09-16
[bitnami/postgresql] Add arbitrary securityContext configuration to improve security

djcass44:
**Description of the change**

Adds options to set arbitrary pod and container security context values.

**Benefits**

Improves the applications ability to run in strict security environments where certain privileges must be dropped or configured to a certain specification but aren't done so automatically (e.g. OPA GateKeeper in Kubernetes)

**Possible drawbacks**

`securityContext.runAsUser` was moved to `containerSecurityContext.runAsUser` as the original was only ever used in the container security context, not the pod one. This may be confusing when upgrading, however it maintains the original behaviour of the chart.

**Applicable issues**

  - references #3669

**Additional information**

- Based off #3615

**Checklist** 
- [x] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [x] Variables are documented in the README.md
- [x] Title of the PR starts with chart name (e.g. `[bitnami/chart]`)
- [x] If the chart contains a `values-production.yaml` apart from `values.yaml`, ensure that you implement the changes in both files

:warning: Keep in mind that if you want to make changes to the kubeapps chart, please implement them in the [kubeapps repository](https://github.com/kubeapps/kubeapps/tree/master/chart/kubeapps). This is only a synchronized mirror.

Issue #3669, opened 2020-09-14
Can't fully modify application securityContext


djcass44:
**Which chart**:
mongodb - 9.1.2
postgresql - 9.4.1
postgresql-ha - 3.6.1
Likely others

**Describe the bug**
Charts do not offer a way of properly configuring an applications security context. E.g. the `allowPrivilegeEscalation` flag cannot be set to false. This is a problem with non-mutating policy engines such as OPA GateKeeper which will reject containers which do not set this flag to false.

**To Reproduce**
1. Install OPA GateKeeper with the allow-privilege-escalation PSP (https://github.com/open-policy-agent/gatekeeper-library/tree/master/library/pod-security-policy)
2. Deploy any of the listed charts using standard values
3. `kubectl get events`
4. See that the pod creation failed

**Expected behavior**

The charts should drop privileges that they don't require, or they should be configurable enough for a user to do so.
A rough example:
```yaml
{{- if .Values.securityContext.enabled }}
      securityContext: {{ toYaml .values.securityContext }}
{{- end }}
```

**Version of Helm and Kubernetes**:

- Output of `helm version`:

```
version.BuildInfo{Version:"v3.0.2", GitCommit:"19e47ee3283ae98139d98460de796c1be1e3975f", GitTreeState:"clean", GoVersion:"go1.13.5"}
```

- Output of `kubectl version`:

```
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.0", GitCommit:"e19964183377d0ec2052d1f1fa930c4d7575bd50", GitTreeState:"clean", BuildDate:"2020-08-26T14:30:33Z", GoVersion:"go1.15", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.6", GitCommit:"dff82dc0de47299ab66c83c626e08b245ab19037", GitTreeState:"clean", BuildDate:"2020-07-15T16:51:04Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}
```

**Additional context**

I'm happy to fix this for some of the charts, I'm not certain of the best way forward in terms of retaining backwards compatibility.

marcosbc:
Hi @djcass44, you're right. We recently implemented something similar for Fluentd, see: https://github.com/bitnami/charts/pull/3615. I'm going to open an internal task for handling these changes to all related charts.

[...]

PR #3705, opened 2020-09-18
[bitnami/postgresql-ha] Add arbitrary securityContext configuration to improve security

djcass44:
**Description of the change**

Adds options to set arbitrary pod and container security context values.

**Benefits**

Improves the applications ability to run in strict environments where certain privileges must be dropped or configured to a certain specification but aren't done so automatically (e.g. OPA GateKeeper in Kubernetes).

**Possible drawbacks**

`securityContext.runAsUser` was moved to `containerSecurityContext.runAsUser` as the original was only ever used in the container security context, not the pod one. This may be confusing when upgrading, however it maintains the original behaviour of the chart.

**Applicable issues**

  - references #3669 

**Additional information**

- Practically identical to #3687 which was based off #3615 

**Checklist** 


bitnami/postgresql-ha/templates/pgpool/deployment.yaml
```diff
@@ -43,8 +43,7 @@ spec:
       tolerations: {{- include "postgresql-ha.tplValue" (dict "value" .Values.pgpool.tolerations "context" $) | nindent 8 }}
       {{- end }}
       {{- if .Values.pgpool.securityContext.enabled }}
-      securityContext:
-        fsGroup: {{ .Values.pgpool.securityContext.fsGroup }}
+      securityContext: {{- omit .Values.pgpool.securityContext "enabled" | toYaml | nindent 8 }}
```
miguelaeh, line 46: From my point, this is a confusing way to select the queys (using negative logic).
I think it would be more readable to follow the previous approach.

This comment applies to the rest of case where you used the `omit` function

miguelaeh (changes requested):
Hi @djcass44 ,

bitnami/postgresql-ha/templates/pgpool/deployment.yaml
miguelaeh: Hi @djcass44 ,
I talked about this with the team and we are going to accept it since it seems to be the more customizable way to add the securityContext.

miguelaeh (approved):
LGTM!

PR #3713, opened 2020-09-18
PoC for pods' affinity

juan131:

**Description of the change**

This is a PoC for standardizing the way to configure pods' affinity is exposed on Bitnami charts.

**Benefits**

- Balanced UX for both advanced Kubernetes admins and less experienced.
- Opinionated presets but users can still define their own affinity
- 100% customizable
- No backwards incompatibility

**Possible drawbacks**

- Complexity

**Checklist**

### DIFF
--- a/bitnami/common/templates/_affinities.tpl
+++ b/bitnami/common/templates/_affinities.tpl
@@ -2,9 +2,9 @@
 
 {{/*
 Return a soft nodeAffinity definition 
-{{ include "common.affinities.node.soft" (dict "key" "FOO" "values" (list "BAR" "BAZ")) -}}
+{{ include "common.affinities.nodes.soft" (dict "key" "FOO" "values" (list "BAR" "BAZ")) -}}
 */}}
-{{- define "common.affinities.node.soft" -}}
+{{- define "common.affinities.nodes.soft" -}}
 preferredDuringSchedulingIgnoredDuringExecution:
   - preference:
       matchExpressions:
@@ -19,9 +19,9 @@ preferredDuringSchedulingIgnoredDuringExecution:
 
 {{/*
 Return a hard nodeAffinity definition
-{{ include "common.affinities.node.soft" (dict "key" "FOO" "values" (list "BAR" "BAZ")) -}}
+{{ include "common.affinities.nodes.hard" (dict "key" "FOO" "values" (list "BAR" "BAZ")) -}}
 */}}
-{{- define "common.affinities.node.hard" -}}
+{{- define "common.affinities.nodes.hard" -}}
 requiredDuringSchedulingIgnoredDuringExecution:
   nodeSelectorTerms:
     - matchExpressions:
@@ -35,21 +35,21 @@ requiredDuringSchedulingIgnoredDuringExecution:
 
 {{/*
 Return a nodeAffinity definition
-{{ include "common.affinities.node" (dict "type" "soft" "key" "FOO" "values" (list "BAR" "BAZ")) -}}
+{{ include "common.affinities.nodes" (dict "type" "soft" "key" "FOO" "values" (list "BAR" "BAZ")) -}}
 */}}
-{{- define "common.affinities.node" -}}
+{{- define "common.affinities.nodes" -}}
   {{- if eq .type "soft" }}
-    {{- include "common.affinities.node.soft" . -}}
+    {{- include "common.affinities.nodes.soft" . -}}
   {{- else if eq .type "hard" }}
-    {{- include "common.affinities.node.hard" . -}}
+    {{- include "common.affinities.nodes.hard" . -}}
   {{- end -}}
 {{- end -}}
 
 {{/*
 Return a soft podAffinity/podAntiAffinity definition
-{{ include "common.affinities.pod.soft" (dict "component" "FOO" "context" $) -}}
+{{ include "common.affinities.pods.soft" (dict "component" "FOO" "context" $) -}}
 */}}
-{{- define "common.affinities.pod.soft" -}}
+{{- define "common.affinities.pods.soft" -}}
 {{- $component := default "" .component -}}
 preferredDuringSchedulingIgnoredDuringExecution:
   - podAffinityTerm:
@@ -66,9 +66,9 @@ preferredDuringSchedulingIgnoredDuringExecution:
 
 {{/*
 Return a hard podAffinity/podAntiAffinity definition
-{{ include "common.affinities.pod.hard" (dict "component" "FOO" "context" $) -}}
+{{ include "common.affinities.pods.hard" (dict "component" "FOO" "context" $) -}}
 */}}
-{{- define "common.affinities.pod.hard" -}}
+{{- define "common.affinities.pods.hard" -}}
 {{- $component := default "" .component -}}
 requiredDuringSchedulingIgnoredDuringExecution:
   - labelSelector:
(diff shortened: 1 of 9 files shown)


bitnami/nginx/templates/deployment.yaml
```diff
@@ -35,6 +35,23 @@ spec:
       {{- include "nginx.imagePullSecrets" . | nindent 6 }}
       {{- if .Values.affinity }}
       affinity: {{- include "common.tplvalues.render" (dict "value" .Values.affinity "context" $) | nindent 8 }}
+      {{- else }}
+      affinity:
+        {{- if eq .Values.podAffinityPreset "soft" }}
+        podAffinity: {{- include "common.affinity.pod.soft" (dict "component" "nginx" "context" $) | nindent 10 }}
```
dani8art, line 41: `nginx` has not `component` label, has it?

bitnami/nginx/templates/deployment.yaml
```diff
@@ -35,6 +35,23 @@ spec:
...
+        {{- else if eq .Values.podAffinityPreset "hard" }}
+        podAffinity: {{- include "common.affinity.pod.hard" (dict "component" "nginx" "context" $) | nindent 10 }}
```
dani8art, line 43: wouldn't it be `podAntiAffinity`?

bitnami/nginx/templates/deployment.yaml
```diff
@@ -35,6 +35,23 @@ spec:
...
+        {{- end }}
+        {{- if eq .Values.podAntiAffinityPreset "soft" }}
```
dani8art, line 45: Couldn't we have an extra helper like:

```
{{- define "common.affinities.pods" -}}
  {{- if eq .type "soft" }}
    {{- include "common.affinities.pods.soft" . -}}
  {{- else if eq .type "hard" }}
    {{- include "common.affinities.pods.hard" . -}}
  {{- end -}}
{{- end -}}
```

And then use it like

```
podAffinity: {{- include "common.affinities.pods" (dict "type" .Values.podAntiAffinityPreset "component" "nginx" "context" $) | nindent 10 }} 
```

We would avoid a lot of `if/else` in all the charts, don't you think?

bitnami/nginx/values.yaml
```diff
@@ -155,8 +155,27 @@ podLabels: {}
 ##
 podAnnotations: {}
 
+## Pod affinity preset
+## ref: https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#inter-pod-affinity-and-anti-affinity
+## Allowed values: soft, hard
+##
+podAffinityPreset: nil
```
dani8art, line 162: Would it be used as `nil` or as a `string`?

bitnami/nginx/templates/deployment.yaml
```diff
@@ -35,6 +35,23 @@ spec:
...
+        podAffinity: {{- include "common.affinity.pod.soft" (dict "component" "nginx" "context" $) | nindent 10 }}
+        {{- else if eq .Values.podAntiAffinityPreset "hard" }}
+        podAffinity: {{- include "common.affinity.pod.hard" (dict "component" "nginx" "context" $) | nindent 10 }}
+        {{- end }}
+        {{- if eq .Values.nodeAffinityPreset "soft" }}
+        nodeAffinity: {{- include "common.affinity.node.soft" (dict "key" "FOO" "values" (list "BAR" "BAZ")) | nindent 10 }}
```
dani8art, line 51: Could you add some examples of what could be `FOO` and `BAR`, sorry I can imagine what values could it be?

dani8art (changes requested):
Thank you so much fo this PR, 

This approach looks very promising!! 

Please, take a look at muy comments?

bitnami/nginx/templates/deployment.yaml
juan131: This depends on you k8s nodes labels...

bitnami/nginx/templates/deployment.yaml
```diff
@@ -35,6 +35,17 @@ spec:
       {{- include "nginx.imagePullSecrets" . | nindent 6 }}
       {{- if .Values.affinity }}
       affinity: {{- include "common.tplvalues.render" (dict "value" .Values.affinity "context" $) | nindent 8 }}
+      {{- else }}
+      affinity:
+        {{- if not (empty .Values.podAffinityPreset) }}
```
dani8art, line 40: I would say it is not needed either, right? I mean, if no preset value matches the condition of `common.affinities.pod`, this would be a valid YAML for k8s

```
affinity:
  podAffinity:
  podAntiAffinity:
  nodeAffinity:
```

bitnami/common/templates/_affinities.tpl
```diff
@@ -0,0 +1,94 @@
...
+        values:
+          {{- range .values }}
+          - {{ . }}
+          {{- end }}
+    weight: 1
+{{- end -}}
+
+{{/*
+Return a hard nodeAffinity definition
+{{ include "common.affinities.node.soft" (dict "key" "FOO" "values" (list "BAR" "BAZ")) -}}
```
javsalgar, line 22: hard

bitnami/nginx/templates/deployment.yaml
```diff
@@ -35,6 +35,17 @@ spec:
...
+        podAffinity: {{- include "common.affinities.pod" (dict "type" .Values.podAffinityPreset "context" $) | nindent 10 }}
+        {{- end }}
+        {{- if not (empty .Values.podAntiAffinityPreset) }}
+        podAntiAffinity: {{- include "common.affinities.pod" (dict "type" .Values.podAntiAffinityPreset "context" $) | nindent 10 }}
+        {{- end }}
+        {{- if not (empty .Values.nodeAffinityPreset) }}
+        nodeAffinity: {{- include "common.affinities.node" (dict "type" .Values.nodeAffinityPreset "key" "some-label-key" "values" (list "some-label-value")) | nindent 10 }}
```
javsalgar, line 47: I imagine `some-label-key` should not be there, right?

javsalgar (changes requested):

bitnami/common/templates/_affinities.tpl
```diff
@@ -0,0 +1,94 @@
...
+  {{- else if eq .type "hard" }}
+    {{- include "common.affinities.node.hard" . -}}
+  {{- end -}}
+{{- end -}}
+
+{{/*
+Return a soft podAffinity/podAntiAffinity definition
+{{ include "common.affinities.pod.soft" (dict "component" "FOO" "context" $) -}}
+*/}}
+{{- define "common.affinities.pod.soft" -}}
```
dani8art, line 52: Sorry, I have an OCD here... XD, I have another proposal for the name, could we put `pods` and `nodes` also XD, it would sound like

`common.affinities.pods.soft`

In the common templates, we have `affinities` for `pods` and they could be `soft` and `hard`. 

Also

`common.affinities.pods`

What do you think about it? do you agree?

dani8art (approved):
Thanks!

LGTM, just two small comments.

bitnami/nginx/templates/deployment.yaml
dani8art: +1

bitnami/nginx/templates/deployment.yaml
dani8art: then we shouldn't add a preset for nodeAffinity, right?

juan131: Correct.. Check new implementation

dani8art (approved):
LGTM!

bitnami/common/templates/_affinities.tpl
```diff
@@ -0,0 +1,94 @@
...
+*/}}
+{{- define "common.affinities.nodes.hard" -}}
+requiredDuringSchedulingIgnoredDuringExecution:
+  nodeSelectorTerms:
+    - matchExpressions:
+        key: {{ .key }}
+        operator: In
+        values:
+          {{- range .values }}
+          - {{ . }}
```
javsalgar, line 32: Question: Would any of these values depend on the release name? If so, would it make sense to allow templating? Maybe not here but in the call

juan131: Nope.. I mean, I don't see how node labels can be related with the name of a specific helm release.

javsalgar: Ok!

javsalgar (approved):

dani8art (approved):

PR #3728, opened 2020-09-21
[bitnami/postgresql] Add customLivenessProbe and customReadinessProbe

javsalgar:
**Description of the change**

In case we enable extra logging settings, users may want to change the liveness and readiness probes so it doesn't add too much noise in the logs.

**Benefits**

**Possible drawbacks**

**Applicable issues**

  - fixes #

**Additional information**

**Checklist** 


bitnami/postgresql/templates/statefulset-slaves.yaml
```diff
@@ -286,6 +286,8 @@ spec:
             timeoutSeconds: {{ .Values.livenessProbe.timeoutSeconds }}
             successThreshold: {{ .Values.livenessProbe.successThreshold }}
             failureThreshold: {{ .Values.livenessProbe.failureThreshold }}
+          {{- else if .Values.customLivenessProbe }}
```
miguelaeh, line 289: I think this logic is confusing.
To use the customLiveness Probes the user will need to set `livenessProbe.enabled=false` and then `customLivenessProbe`. I would do it like:
```
{{- if and .Values.livenessProbe.enabled .Values.customLivenessProbe }}
...
{{- else if .Values.livenessProbe.enabled  }}
...
{{- else }}
...
{{- end }}
```

And something similar for the readiness probe

miguelaeh (commented):
Please take a look at my comment

bitnami/postgresql/README.md
```diff
@@ -126,6 +126,8 @@ The following tables lists the configurable parameters of the PostgreSQL chart a
...
 | `postgresqlPghbaRemoveFilters`                | Comma-separated list of patterns to remove from the pg_hba.conf file                                                                                                                                                                                                                                                                                                                                                                                                       | `nil`                                                         |
+| `customLivenessProbe`  | Override default liveness probe                                                                                       | `nil`                                                   |
+| `customReadinessProbe` | Override default readiness probe                                                                                      | `nil`                                                   |
```
miguelaeh, line 130: Please preserve the format like the other lines of the table.

miguelaeh (changes requested):

bitnami/postgresql/templates/statefulset-slaves.yaml
javsalgar: It's already set like that in most of the charts. We have similar logics in other parameters like `externalDatabase`. For consistency I would leave it like the rest.

javsalgar: But I would add it as something to review in the future

miguelaeh: 👍

miguelaeh (approved):
LGTM!

PR #3729, opened 2020-09-21
[bitnami/wordpress] POC removing Helm 2 support

carrodher:

### DIFF
--- a/bitnami/wordpress/Chart.yaml
+++ b/bitnami/wordpress/Chart.yaml
@@ -1,8 +1,17 @@
+annotations:
+  category: CMS
 apiVersion: v2
-name: wordpress
-version: 10.0.0
 appVersion: 5.5.1
+dependencies:
+  - condition: mariadb.enabled
+    name: mariadb
+    repository: https://charts.bitnami.com/bitnami
+    tags:
+      - wordpress-database
+    version: 7.x.x
 description: Web publishing platform for building blogs and websites.
+engine: gotpl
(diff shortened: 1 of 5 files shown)


juan131 (commented):
The changes LGTM, but please document in the README.md (Upgrade section) the consequences about these changes, and possible alternatives to migrate releases from Helm2 to Helm3 (e.g. link to https://docs.bitnami.com/tutorials/resolve-helm2-helm3-post-migration-issues/)

bitnami/wordpress/Chart.yaml
```diff
@@ -1,6 +1,6 @@
-apiVersion: v1
+apiVersion: v2
 name: wordpress
-version: 9.5.6
+version: 10.0.0
```
scottrigby, line 3: I think you're right a MAJOR version bump is appropriate, since this is an API backwards-incompatible change.

In keeping with this, I agree with @juan131 you'll also want to add a `To 10.0.0`  section under [Upgrading](https://github.com/bitnami/charts/blob/master/bitnami/wordpress/README.md#upgrading) in the README, noting the drop of Helm 2 support.

scottrigby (commented):
Really glad to see this work being prepared in advance of 13 November 👏

carrodher:
I am closing this POC and will replicate those changes in the different charts, starting with the Helm Charts without dependencies and then continuing with the Helm Charts with dependencies once the new major of the dependencies are released

PR #3733, opened 2020-09-21
[bitnami/common] fix: evaluate enabled as string

dani8art:

**Description of the change**

To check boolean as string always, previously we were checking boolean as boolean but sometimes they were string and the `if` statement was evaluating `"false"` as `true`. 

To not run these checks when the chart is installed as a subchart. Before, we when we used this helper in a parent chart we were checking twice in the `parentchar/templates/NOTES.txt` and `parentchart/charts/subchart/templates/NOTES.txt` so that sometimes it went to an error.

**Benefits**

- We will fix the error when evaluating the templates.
- We will avoid the errors when using it in a subchart

**Possible drawbacks**

N/A

**Applicable issues**

N/A

**Additional information**

To test it execute the following:

```
### Test case

# install harbor with external database
helm del hb || \
  k delete pvc -l "app.kubernetes.io/name=harbor" || \
  helm dep update bitnami/harbor && \
  helm upgrade --install hb bitnami/harbor \
  --set postgresql.enabled=false \
  --set externalDatabase.password=xxx

# subchart should throw no errors since the postgresql subchart is disabled.
export HARBOR_ADMIN_PASSWORD=$(kubectl get secret --namespace default hb-harbor-core-envvars -o jsonpath="{.data.HARBOR_ADMIN_PASSWORD}" | base64 --decode)

helm dep update bitnami/harbor && helm upgrade --install hb bitnami/harbor \
  --set postgresql.enabled=false \
  --set externalDatabase.password=xxx \
  --set harborAdminPassword=$HARBOR_ADMIN_PASSWORD \
  --set postgresql.postgresqlPassword= \
  --set postgresql.replication.enabled=false \
  --set postgresql.replication.password=

### End test case
```

**Checklist**


bitnami/common/Chart.yaml
```diff
@@ -1,8 +1,8 @@
 apiVersion: v1
 name: common
 # Please make sure that version and appVersion are always the same.
-version: 0.6.2
-appVersion: 0.6.2
+version: 0.6.3
+appVersion: 0.6.3
```
miguelaeh, line 5: Why are you changing the app version?

miguelaeh (commented):
Thank you for the PR!!

bitnami/common/Chart.yaml
dani8art: It is because pipeline requirements for this special chart

miguelaeh (approved):
LGTM!

PR #3744, opened 2020-09-22
[bitnami/chart] postgre #3722: In exsistingSecret: Switched password keys

### LINKED ISSUE #3722: existingSecret: Wrong or unclear readme
**Which chart**:

Latest postgresql chart

The README.md is incorrect or wrong.

Change the readme.

**The problem**

The readme under existing secret states:

> The secret has to contain the keys postgresql-postgres-password which is the password for postgresqlUsername when it is different of postgres, postgresql-password which will override postgresqlPassword

But `postgresql-postgres-password` is the admin password, and not the password of the optionally-created-non-admin-user of `postgresqlUsername`. I think the password-keys should be swapped so it should instead be:

> The secret has to contain the keys postgresql-password which is the password for postgresqlUsername when it is different of postgres, postgresql-postgres-password which will override postgresqlPassword`

I would like someone to verify that I am correct and then I can do a pull request.

/Jesper

alex-jesper:
**Description of the change**
Switched two keys in exsistingSecret, as they had been switched.

**Benefits**
Less confusion for the next guy why fiddles with secrets

**Applicable issues**
  - fixes #3722

**Checklist** 


juan131 (approved):
Thanks for the fix!! LGTM

PR #3800, opened 2020-09-28
[bitnami/postgresql] Allow specifying different resource requests/limits on slaves

vaghetti:
**Description of the change**
Adds the ability to specify a different resource request for the slave replicas

**Benefits**
If your use case uses the slave replicas as read only replicas, but not intending to use them as a replacement for the master replica in case it goes down, it can make sense to allocate less resources to it. With the current behavior of the helm chart, the resource requirements for master and slave must always be the same.

**Possible drawbacks**
None that I can think of. The change is backwards compatible, and can be safely ignored by people who don't need it. 

**Applicable issues**
Couldn't find any

**Checklist** 


vaghetti:
Seems like I messed up something when forking the repo. I'll fix it and re-open the PR.

bitnami/postgresql/README.md
```diff
@@ -182,6 +182,7 @@ The following tables lists the configurable parameters of the PostgreSQL chart a
...
 | `slave.anotations`                            | Map of annotations to add to the statefulsets (postgresql slave)                                                                                                                                                                                                                                                                                                                                                                                                           | `{}`                                                          |
+| `slave.resources`                             | CPU/Memory resource requests/limits override for slaves. Will fallback to `values.resources` of not defined.                                                                                                                                                                                                                                                                                                                                                               | `{}`                                                          |
```
juan131, line 185: Typo: `~of~ not defined` -> `if not defined`

juan131 (changes requested):
Thanks so much for this PR! 

Could you please fix the small typo? Once done, we can go ahead with the merging process.

bitnami/postgresql/README.md
vaghetti: Oops, typo fixed.

juan131 (approved):
LGTM

PR #3441, opened 2020-08-17
[bitnami/dokuwiki] feat: add init-container for permissions

dani8art:
Signed-off-by: darteaga <darteaga@vmware.com>

**Description of the change**

Add an init-container to manage the volume permission at initialization.

**Benefits**

We can use this chart with PVC which has no the proper permissions.

**Possible drawbacks**

N/A

**Applicable issues**

  - fixes #3338

**Additional information**

N/A

**Checklist** 
- [x] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [x] Variables are documented in the README.md
- [x] Title of the PR starts with chart name (e.g. `[bitnami/chart]`)
- [x] If the chart contains a `values-production.yaml` apart from `values.yaml`, ensure that you implement the changes in both files

PR #3823, opened 2020-09-30
[bitnami/dokuwiki] Do not run volumePermissions init container by default

marcosbc:
**Description of the change**

In none of our containers do we enable the volumePermissions init container by default, but in #3441 we added support for it and then also enabled it by default.

However it should not be needed for most scenarios, in fact some Kubernetes clusters would not even support it because it would start a container as a root user, against cluster policies.

Even the README shows `false` being the default value currently, meaning the current setting is wrong.

**Benefits**

* Consistency with other Bitnami charts, as we're never running the volumePermissions init container by default
* Default values would work on clusters with policies against running containers as a root user.

**Possible drawbacks**

* Changing the default values that would result in a new behavior for users that had accustomed to it.

**Applicable issues**

None

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/dokuwiki/Chart.yaml
+++ b/bitnami/dokuwiki/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: dokuwiki
-version: 9.3.0
+version: 9.4.0
 appVersion: 20200729.0.0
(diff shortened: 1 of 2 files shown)


miguelaeh (approved):
LGTM!!

PR #3871, opened 2020-10-02
[bitnami/mediawiki] Fix service selector greediness

kachkaev:
**Description of the change**

This PR makes pod selection in the MediaWiki service less greedy. Instead of just looking at `app` label, `release` label is also taken into the account now.

**Benefits**

The change allows deploying two instances of MediaWiki into a single namespace (in my case these are `main` and `sandbox`). Without the addition label selector, the services are selecting pods across apps, thus randomly showing users different instances of the wiki.

Before:

```txt
service A      service B
    ↓     ↘  ↙     ↓ 
  pod A          pod B
```

After:

```txt
service A      service B
    ↓              ↓ 
  pod A          pod B
```

**Possible drawbacks**

Unknown

**Additional information**

Source of inspiration:

https://github.com/bitnami/charts/blob/f5033e893646592e1ec6b063f229eff030e90f7a/bitnami/mariadb/templates/master-svc.yaml#L44-L47

After manually tweaking the selectors in the deployed services via the k8s dashboard, I can conclude that the suggested patch is working.

**Checklist** 
- ~~Variables are documented in the README.md~~ (N/A)
- ~~If the chart contains a `values-production.yaml` apart from `values.yaml`, ensure that you implement the changes in both files~~ (N/A)

### DIFF
--- a/bitnami/mediawiki/Chart.yaml
+++ b/bitnami/mediawiki/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: mediawiki
-version: 10.0.8
+version: 10.0.9
 appVersion: 1.35.0
 description: Extremely powerful, scalable software and a feature-rich wiki implementation that uses PHP to process and display data stored in a database.
 home: https://github.com/bitnami/charts/tree/master/bitnami/mediawiki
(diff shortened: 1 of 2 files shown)


marcosbc (approved):
Good catch, thanks a lot for the contribution! LGTM

Issue #3864, opened 2020-10-02
Zookeeper namespace override


danielburrell:
**Which chart**:
The name (and version) of the affected chart.
zookeeper current.

**Is your feature request related to a problem? Please describe.**

Our Chart architecture:

`Our application chart -> Stardog chart -> zookeeper chart`.

This shows that Zookeeper is a dependency of stardog, which is a dependency of own application chart. 

We currently deploy our application using `helm install --namespace somenamespace`. As a result of zookeeper's chart using `{{.Release.Namespace}}` zookeeper is deployed to the same namespace. We really benefit from the way helm allows us to declare dependent charts, however we need to be able to place zookeeper in a separate namespace. We're currently able to customize the stardog namespace, but it's currently not possible with zookeeper.

**Describe the solution you'd like**
We would like the zookeeper chart to be modified such that the `{{.Release.Namespace}}` continues to be used as a default, but it is additionally possible to specify a .Values.namespace override. This would be used by our application's `Values.yaml` by setting `zookeeper.namespace` or some similar variable name. This would allow zookeeper to be deployed at the same time as our application but in a different namespace.

**Describe alternatives you've considered**
- We have considered deploying zookeeper as a separate piece of infrastructure with it's own helm install release process in order to allow its own namespace to be defined, but this is undesirable as it adds to the complexity of the release process. As zookeeper is a dependency of our dependency, we would like to distance ourselves from the implementation details of those dependencies in terms of what needs to be deployed. We feel that since helm is designed to provide flexibility to kubernetes manifests, that namespaces are probably a good thing to allow customization for.
- We have also considered forking the chart, but again, we'd rather not do this as we believe in re-use where possible and the additional overhead of keeping the fork up to date would be undesirable.

**Additional context**
This is the chart that actually uses zookeeper (although it doesn't use bitnami currently there is a ticket open to migrate, bitnami being a likely candidate). https://github.com/stardog-union/helm-charts/issues/30 This chart had the same problem around namespace customization, we made a similar request to them and they were happy to make the change (change is still being implemented), but pointed out to us that zookeeper as a dependency would need the same change.

I understand development time is precious, so I'm happy to provide a pull request with the appropriate changes if you confirm that such a feature would be accepted in principle.

miguelaeh:
Hi,
Thank you very much for the explanation of the issue and the PR. We are already taking care of the PR!
Regards.

PR #3881, opened 2020-10-03
[bitnami/zookeeper] allow for overriding namespace in zookeeper helm chart

monicaruttle:
**Description of the change**

Allow for overriding the namespace that the ZooKeeper helm chart resources use.

**Benefits**

When ZooKeeper is included as a dependency of another helm chart, the preferred behaviour may be for it to be released into a different namespace than the parent. This change will allow for the parent chart values to override the namespace.

**Possible drawbacks**

Unknown

**Applicable issues**

  - addresses #3864

**Additional information**

By default behaviour will not change from taking the Release.Namespace value for the namespace.

**Checklist** 

### DIFF
--- a/bitnami/zookeeper/templates/statefulset.yaml
+++ b/bitnami/zookeeper/templates/statefulset.yaml
@@ -2,7 +2,11 @@ apiVersion: apps/v1
 kind: StatefulSet
 metadata:
   name: {{ template "zookeeper.fullname" . }}
+  {{- if .Values.namespace }}
+  namespace: {{ .Values.namespace }}
+  {{- else }}
   namespace: {{ .Release.Namespace }}
+  {{- end }}
   labels: {{- include "zookeeper.labels" . | nindent 4 }}
     app.kubernetes.io/component: zookeeper
     role: zookeeper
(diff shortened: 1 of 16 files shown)


glothriel:
@juan131 I think you may be interested in this one :-)

Personally I like the idea, with some changes:
* I've seen somewhere (charts/stable?) similar approach, but the field there was named `namespaceOverride`, for some reason I like it more
* I'd extract this logic to a template

bitnami/zookeeper/templates/statefulset.yaml
```diff
@@ -2,7 +2,11 @@ apiVersion: apps/v1
 kind: StatefulSet
 metadata:
   name: {{ template "zookeeper.fullname" . }}
```
miguelaeh, line 4: Line 142 needs to be changed too

miguelaeh (changes requested):
Hi @monicaruttle ,
Please take a look at the files carefully, there are some places where the namespace is used that also need to be changed.
Apart from that, I agree with @glothriel in use `namespaceOverride` and a template to select the namespace instead of repeating the logic in all the files.

dani8art (changes requested):
Hi @monicaruttle thanks for applying the changes. 

Please, take a look at my comments.

dani8art (approved):
LGTM

PR #3890, opened 2020-10-05
[bitnami/nats] Check if it's available ingress api 'networking.k8s.io/v1beta1'

w1mvy:
**Description of the change**

Check and use if it's available ingress apiVersion `networking.k8s.io/v1beta1`.

**Benefits**

`extensions/v1beta1` is deprecated in k8s v1.22, so avoid becoming unavailable this chart.

ref: https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/

```
The v1.22 release will stop serving the following deprecated API versions in favor of newer and more stable API versions:

Ingress in the extensions/v1beta1 API version will no longer be served
Migrate to use the networking.k8s.io/v1beta1 API version, available since v1.14. Existing persisted data can be retrieved/updated via the new version.
```

**Possible drawbacks**

- none

**Applicable issues**

- none

**Additional information**

- none

**Checklist** 

### DIFF
--- a/bitnami/nats/templates/ingress.yaml
+++ b/bitnami/nats/templates/ingress.yaml
@@ -1,6 +1,11 @@
 {{- if .Values.ingress.enabled -}}
+{{- $apiVersions := .Capabilities.APIVersions -}}
 {{- range .Values.ingress.hosts }}
+{{- if $apiVersions.Has "networking.k8s.io/v1beta1" }}
+apiVersion: networking.k8s.io/v1beta1
+{{- else }}
 apiVersion: extensions/v1beta1
+{{- end }}
 kind: Ingress
 metadata:
   name: {{ template "nats.fullname" $ }}-monitoring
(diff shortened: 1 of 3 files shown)


bitnami/nats/templates/ingress.yaml
```diff
@@ -1,6 +1,11 @@
 {{- if .Values.ingress.enabled -}}
+{{- $apiVersions := .Capabilities.APIVersions -}}
 {{- range .Values.ingress.hosts }}
+{{- if $apiVersions.Has "networking.k8s.io/v1beta1" }}
+apiVersion: networking.k8s.io/v1beta1
```
javsalgar, line 5: In other charts we use a helper function that returns the proper API version.

```
apiVersion: {{ include "nats.ingressAPIVersion" . }}
```

javsalgar (changes requested):

javsalgar:
Thank you so much for the PR! Small change requested

bitnami/nats/templates/ingress.yaml
w1mvy: Thanks your review, and fix it!

javsalgar (approved):

PR #3778, opened 2020-09-25
[bitnami/*] Affinity based on common presets (ii)

juan131:
Signed-off-by: juan131 <juanariza@vmware.com>

**Description of the change**

This PR adapts the charts below so it uses the common presets added on `bitnami/common` at https://github.com/bitnami/charts/pull/3713.

- bitnami/etcd
- bitnami/external-dns
- bitnami/fluentd
- bitnami/ghost
- bitnami/grafana
- bitnami/harbor
- bitnami/influxdb
- bitnami/jasperreports
- bitnami/jenkins
- bitnami/joomla
- bitnami/kafka
- bitnami/kibana

**Benefits**

- Balanced UX for both advanced Kubernetes admins and less experienced.
- Opinionated presets but users can still define their own affinity
- 100% customizable
- No backwards incompatibility

**Possible drawbacks**

- Complexity

**Checklist**

- [x] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [x] Variables are documented in the README.md
- [x] Title of the PR starts with chart name (e.g. `[bitnami/chart]`)
- [x] If the chart contains a `values-production.yaml` apart from `values.yaml`, ensure that you implement the changes in both files

PR #3746, opened 2020-09-22
[bitnami/*] Affinity based on common presets

juan131:
Signed-off-by: juan131 <juanariza@vmware.com>

**Description of the change**

This PR adapts the charts below so it uses the common presets added on `bitnami/common` at https://github.com/bitnami/charts/pull/3713.

- bitnami/airflow
- bitnami/apache
- bitnami/aspnet-core
- bitnami/cassandra
- bitnami/consul
- bitnami/contour
- bitnami/discourse
- bitnami/dokuwiki
- bitnami/drupal
- bitnami/ejbca
- bitnami/elasticsearch
- bitnami/nginx

**Benefits**

- Balanced UX for both advanced Kubernetes admins and less experienced.
- Opinionated presets but users can still define their own affinity
- 100% customizable
- No backwards incompatibility

**Possible drawbacks**

- Complexity

**Checklist**

- [x] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [x] Variables are documented in the README.md
- [x] Title of the PR starts with chart name (e.g. `[bitnami/chart]`)

PR #3897, opened 2020-10-05
[bitnami/nats] add priority to zones

SagiMedinaCheckmarx:
k8s official docs:
https://kubernetes.io/docs/reference/kubernetes-api/labels-annotations-taints/#topologykubernetesiozone

**Description of the change**
add pod anti-affinity by additional topology key for "soft" anti-affinity value.
I think that you shall consider removing the "anti-affinity" value and use standard "affinity" value for user-defined.

**Benefits**
multi-zone HA.

**Possible drawbacks**

**Applicable issues**

  - fixes #

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/nats/Chart.yaml
+++ b/bitnami/nats/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: nats
-version: 4.5.5
+version: 4.5.6
 appVersion: 2.1.8
 description: An open-source, cloud-native messaging system
 keywords:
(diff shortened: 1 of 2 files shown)


FraPazGal (commented):
Hi @SagiMedinaCheckmarx,

Thank you so much for your PR! It is true that the current `affinity` configuration can be improved. We are in the process of adding more flexibility and allowing affinity customisation in our charts (see #3746), but it would take a while to have those changes in all of them.

In the meantime, I agree that this change would be beneficial. I believe it would be good to also include your change in the [deployment.yaml](https://github.com/bitnami/charts/blob/90ef327131f6fbf9ff92c79be998a0deccd5790e/bitnami/nats/templates/deployment.yaml#L71), what do you think?

juan131:
As @FraPazGal 

We are standardising the way to configure affinity/antiaffinity on PRs like this one: https://github.com/bitnami/charts/pull/3778

So I'll proceed to close this PR since soon there'll be a easy way to setup every custom desired affinity.

PR #3898, opened 2020-10-05
[bitnami/airflow] add service-account and role creation

### LINKED ISSUE #3779: [bitnami/airflow] Add creation or existing of using service account in airflow chart
**Which chart**:
airflow (all versions)

We want to use kubernetes pod operator in airflow which will require us to attach a service account to airflow with proper roles/role binding.

A service account name to be provided, if provided use the service account else have a flag createServiceAccount: true will create the service account and add to all deployment/statefulset manifests.

We can manage roles and role binding separately or that also can be included in this chart.

Alternative I use today, is to attach the role to default service account that is being used by airflow but I feel it is not secure.

iamsudip:
**Description of the change**

Creates airflow service account
Creates role and rolebinding is rbac enabled.
Adds `commonLabels` and `commonAnnotations` to all deployed resources.

Earlier, the airflow pods were using default service account, so `rbac` and `serviceAccount` creation is still optional and defaults to `false`.
`commonLabels` and `commonAnnotations` are default to `{}`

**Benefits**

- `rbac` set to `true` is required to use KubernetesPodOperator.
- `commonLabels` and `commonAnnotations` set organizational conventions.

**Possible drawbacks**

NA

**Applicable issues**

Fixes #3779 

**Additional information**
NA

**Checklist** 
- [x] If the chart contains a `values-production.yaml` apart from `values.yaml`, ensure that you implement the changes inboth files


iamsudip:
@FraPazGal Please have a look and provide suggestions if any.

bitnami/airflow/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: airflow
-version: 6.6.0
+version: 6.6.2
```
FraPazGal, line 3: I would say this is a minor version rather than a patch, what do you think?

bitnami/airflow/README.md
```diff
@@ -107,6 +107,9 @@ The following tables lists the configurable parameters of the Airflow chart and
...
 | `airflow.baseUrl`                                                        | URL used to access to airflow web ui                                                                                                                                                                                           | `nil`                                                        |
+| `airflow.rbac.create`                                                    | Specifies whether RBAC rules should be created                                                                                                                                                                                 | `false`                                                      |
+| `airflow.serviceAccount`                                                 | Specifies whether a ServiceAccount should be created                                                                                                                                                                           | `false`                                                      |
```
FraPazGal, line 111: I would say this variable should be `airflow.serviceAccount.create`. Also, you forgot to add `airflow.serviceAccount.name`.

bitnami/airflow/templates/role.yaml
```diff
@@ -0,0 +1,27 @@
+{{- if .Values.rbac.create -}}
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: Role
+metadata:
+  name: {{ template "airflow.fullname" . }}
+  namespace: {{ .Release.Namespace }}
+  labels: {{- include "airflow.labels" . | nindent 4 }}
+    {{- if .Values.commonLabels }}
```
FraPazGal, line 8: You have added the parameter `commonLabels` to the new templates but it is not present in `values.yaml`. This is a parameter that we are introducing in our charts and I agree it should also be in this one. 

It would be great if you could implement this parameter in all the other templates. To keep using the `nginx-ingress-controller` as an example, you can use #3585 as reference.

bitnami/airflow/templates/role.yaml
```diff
@@ -0,0 +1,27 @@
...
+    {{- include "airflow.tplValue" (dict "value" .Values.commonLabels "context" $) | nindent 4 }}
+    {{- end }}
+  {{- if .Values.commonAnnotations }}
```
FraPazGal, line 11: Same case like with `commonLabels`.

FraPazGal (changes requested):
Hi @iamsudip!

Thanks for your PR, great work! Please take a look at my comments.

bitnami/airflow/Chart.yaml
iamsudip: @FraPazGal I agree.

bitnami/airflow/README.md
iamsudip: Yes, my bad. Will fix, thanks.

bitnami/airflow/templates/role.yaml
iamsudip: @FraPazGal Thanks for the suggestion, will take hints from there and make changes accordingly.

bitnami/airflow/templates/role.yaml
iamsudip: Understood.

iamsudip:
@FraPazGal Made the changes. Have a look.

FraPazGal:
Hi @iamsudip,

It's looking good so far! I'll fully test these changes and get back to you tomorrow.

bitnami/airflow/templates/configmap-dags.yaml
```diff
@@ -4,10 +4,17 @@ kind: ConfigMap
...
+    {{- if .Values.commonLabels }}
+    {{- include "airflow.tplValue" (dict "value" .Values.commonLabels "context" $) | nindent 4 }}
+    {{- end }}
     app.kubernetes.io/name: {{ include "airflow.name" . }}
     helm.sh/chart: {{ include "airflow.chart" . }}
     app.kubernetes.io/instance: {{ .Release.Name }}
     app.kubernetes.io/managed-by: {{ .Release.Service }}
+  {{- if .Values.commonAnnotations }}
+  annotations:
+    {{- include "airflow.tplValue" (dict "value" .Values.commonAnnotations "context" $) | nindent 4 }}
```
FraPazGal, line 16:
```suggestion
    annotations: {{- include "airflow.tplValue" (dict "value" .Values.commonAnnotations "context" $) | nindent 4 }}
```
This could be only one line as you are already using `nindent`. Sorry if it seems like nick-picking but it is just to maintain standarization across our charts.

FraPazGal (commented):
Hi!

Everything is great and working! Please see my comments, only a couple of small changes to finish off this PR!

bitnami/airflow/templates/role.yaml
```diff
@@ -0,0 +1,28 @@
...
+    resources:
+      - pods
+      - pods/log
+      - pods/exec
+      - services
+      - endpoints
+      - configmaps
+      - secrets
+    verbs:
+      - '*'
```
FraPazGal, line 27:
```suggestion
      - get
      - list
      - watch
      - create
      - delete
      - update
      - patch
```

For clarifications sake, please expand to a list the verbs. Also, I'm not sure if KubernetesPodOperator need permissions so broad to function. Could you confirm this?

iamsudip: @FraPazGal, Hmmm... Removed update/patch. Rest all is needed as far as I understood from reading airflow kubernete pod operator code [here](https://github.com/apache/airflow/tree/master/airflow/kubernetes).

I have removed `services/endpoints` from `resources` cause we run spark-on-kubernetes, but for that, we can specify a different service account if needed and that's the right way to do it not sharing airflow service account with spark driver.

iamsudip:
@FraPazGal Have a look now.

juan131 (approved):
Thanks so much for the great PR! Approving your changes!

PR #3902, opened 2020-10-05
[bitnami/etcd] Allow alternate TLS secret filenames

crandles:
**Description of the change**

Allow alternate TLS secret filenames to be used for client and peer authentication.
Use the etcd-headless service in the serviceMonitor definition.
Additionally, allow etcd's serviceMonitor tlsConfig to be configured.

**Benefits**

Integration with certificate-generation tools like cert-manager which may use alternate filenames for certificates.

**Checklist** 


bitnami/etcd/README.md
```diff
@@ -83,10 +83,16 @@ The following tables lists the configurable parameters of the etcd chart and the
...
 | `auth.client.existingSecret`                    | Name of the existing secret containing cert files for client communication.                                                                               | `nil`                                                       |
+| `auth.client.certFilename`                        | Name of the file containing the client certificate.                                                                                                     | `cert.pem`                                                  |
+| `auth.client.certKeyFilename`                     | Name of the file containing the client certificate private key.                                                                                         | `key.pem`                                                   |
+| `auth.client.caFilename`                          | Name of the file containing the client CA certificate. If not specified and `enableAuthentication: true` or `rbac.enabled: true`, the default is is `ca.crt`. | `""`                                                  |
```
miguelaeh, line 88: Please format these line properly as the rest of the table

bitnami/etcd/README.md
```diff
@@ -83,10 +83,16 @@ The following tables lists the configurable parameters of the etcd chart and the
...
 | `auth.peer.existingSecret`                      | Name of the existing secret containing cert files for peer communication.                                                                                 | `nil`                                                       |
+| `auth.peer.certFilename`                        | Name of the file containing the peer certificate.                                                                                                         | `cert.pem`                                                  |
+| `auth.peer.certKeyFilename`                     | Name of the file containing the peer certificate private key.                                                                                             | `key.pem`                                                   |
+| `auth.peer.caFilename`                          | Name of the file containing the peer CA certificate. If not specified and `enableAuthentication: true` or `rbac.enabled: true`, the default is is `ca.crt`. | `""`                                                      |
```
miguelaeh, line 95: This also needs to be formated properly

bitnami/etcd/values.yaml
```diff
@@ -148,6 +158,17 @@ auth:
...
+    ##
+    certFilename: cert.pem
+    ## Name of the file containing the peer certificate private key.
+    ##
+    certKeyFilename: key.pem
+    ## Name of the file containing the peer CA certificate.
+    ## If not specified and `enableAuthentication: true` or `rbac.enabled: true`, the default is is `ca.crt`.
+    ##
+    caFilename: ""
+
```
miguelaeh, line 171: Remove extra empty line

bitnami/etcd/templates/statefulset.yaml
```diff
@@ -157,29 +157,35 @@ spec:
               value: "true"
             {{- else if .Values.auth.client.secureTransport }}
             - name: ETCD_CERT_FILE
-              value: "/opt/bitnami/etcd/certs/client/cert.pem"
+              value: "/opt/bitnami/etcd/certs/client/{{ .Values.auth.client.certFilename }}"
             - name: ETCD_KEY_FILE
-              value: "/opt/bitnami/etcd/certs/client/key.pem"
+              value: "/opt/bitnami/etcd/certs/client/{{ .Values.auth.client.certKeyFilename }}"
```
miguelaeh, line 162: You will also need to change the `_helpers.tpl`, the `NOTES.txt` and the `cronjob.yaml` files.
The same applies to the `cert.pem` file too.

bitnami/etcd/templates/servicemonitor.yaml
```diff
@@ -13,15 +13,28 @@ metadata:
 spec:
   selector:
     matchLabels: {{- include "etcd.matchLabels" . | nindent 6 }}
+      etcd: headless
```
miguelaeh, line 16: Could you explain why this is needed?

miguelaeh (changes requested):

bitnami/etcd/templates/servicemonitor.yaml
crandles: As-is, the serviceMonitor includes both the headless and non-headless service, and ultimately you end up with two sets of the same pods in the target list.

bitnami/etcd/templates/statefulset.yaml
crandles: `_helper.tpl` seems to be for the autoTLS option. I'm not sure it makes sense to override there. Will add to the other sections you mention

bitnami/etcd/templates/servicemonitor.yaml
crandles: What is the non-headless service used for? Maybe it would be better to allow it to be disabled.

bitnami/etcd/README.md
crandles: Fixed

bitnami/etcd/templates/servicemonitor.yaml
miguelaeh: The non-headless service is used to access the etcd pods deployment from the clients

crandles: Should I remove this? I think I can get by with relabelings, with mTLS/host authentication i need a complicated one already (not including pod IPs in my certificate):

```yaml
    relabelings:
    - separator: .
      sourceLabels:
      - __meta_kubernetes_pod_name
      - __meta_kubernetes_endpoints_name
      targetLabel: __pod_endpoint_address
    - separator: ':'
      sourceLabels:
      - __pod_endpoint_address
      - __meta_kubernetes_pod_annotation_prometheus_io_port
      targetLabel: __address__
```

Dropping the extra service isn't very hard, I just wasn't expecting both services to be targeted by the ServiceMonitor

crandles: Here I'm preferring the headless service (only). My reasoning:
- When I view the targets list associated with this ServiceMonitor, I expect one target per pod.
- Without this change: I was seeing two targets per pod (one per each of the two services)

crandles: I've reverted the `headless` service preference -- it diverges from the original intent of the PR.

dani8art: Thanks

dani8art (approved):
LGTM

PR #3869, opened 2020-10-02
[bitnami/drupal] New major version

juan131:
Signed-off-by: juan131 <juanariza@vmware.com>

**Description of the change**

This PR bumps Drupal major version since its dependency (MariaDB) bumped the major version including breaking changes. More info at https://github.com/bitnami/charts/pull/3849

**Benefits**

- Standardization.
- Reduces technical debt.

**Benefits**

**Possible drawbacks**

It breaks backwards compatibility.

**Applicable issues**

None

**Additional information**

**Checklist**

- [x] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [x] Variables are documented in the README.md
- [x] Title of the PR starts with chart name (e.g. `[bitnami/chart]`)
- [x] If the chart contains a `values-production.yaml` apart from `values.yaml`, ensure that you implement the changes in both files

PR #3916, opened 2020-10-06
[bitnami/opencart] MAJOR: Follow non-root approach

marcosbc:
**Description of the change**

We are working on releasing the OpenCart Docker image with a non-root approach. As part of that, we need to update the chart to use the new port numbers used by the container image and related settings.

We're also taking advantage of this PR being a major change by adding all standarizations (i.e. customizable command, extra volumes, extra envs, K8s labels standarization, etc.).

**Benefits**

**Possible drawbacks**

Breaking change with previous chart versions, as the default HTTP/HTTPS ports are changed to 8080/8443 (previously 80/443). The labels may also cause this to be a breaking change.

**Applicable issues**

None

**Additional information**

**Checklist** 


bitnami/opencart/requirements.yaml
```diff
@@ -1,5 +1,8 @@
 dependencies:
-- name: mariadb
-  version: 7.x.x
-  repository: https://charts.bitnami.com/bitnami
-  condition: mariadb.enabled
+  - name: mariadb
+    version: 7.x.x
```
dani8art, line 3: Could we also take the advantage of this PR to upgrade mariadb to `8.x.x`?

bitnami/opencart/templates/NOTES.txt
```diff
@@ -73,31 +75,27 @@ host. To configure OpenCart with the URL of your service:
...
     --set opencartPassword=$APP_PASSWORD,opencartHost=$APP_HOST,service.type={{ .Values.service.type }},mariadb.enabled=false{{- if not (empty .Values.externalDatabase.user) }},externalDatabase.user={{ .Values.externalDatabase.user }}{{- end }}{{- if not (empty .Values.externalDatabase.password) }},externalDatabase.password={{ .Values.externalDatabase.password }}{{- end }}{{- if not (empty .Values.externalDatabase.database) }},externalDatabase.database={{ .Values.externalDatabase.database }}{{- end }},externalDatabase.host=YOUR_EXTERNAL_DATABASE_HOST{{- if .Values.global }}{{- if .Values.global.imagePullSecrets }},global.imagePullSecrets={{ .Values.global.imagePullSecrets }}{{- end }}{{- end }}
 {{- end }}
 
-{{- if and (contains "bitnami/" .Values.image.repository) (not (.Values.image.tag | toString | regexFind "-r\\d+$|sha256:")) }}
-
-WARNING: Rolling tag detected ({{ .Values.image.repository }}:{{ .Values.image.tag }}), please note that it is strongly recommended to avoid using rolling tags in a production environment.
-+info https://docs.bitnami.com/containers/how-to/understand-rolling-tags-containers/
-
-{{- end }}
+{{- include "opencart.checkRollingTags" . }}
```
dani8art, line 101: could you add password validation? example [here](https://github.com/bitnami/charts/blob/0ffded5c0785040097d76d38a20f15b2f4ae1ea4/bitnami/drupal/templates/NOTES.txt#L47)

bitnami/opencart/templates/_helpers.tpl
```diff
@@ -1,35 +1,3 @@
...
-{{- end -}}
-
-{{/*
-Create chart name and version as used by the chart label.
-*/}}
-{{- define "opencart.chart" -}}
-{{- printf "%s-%s" .Chart.Name .Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-" -}}
-{{- end -}}
-
 {{/*
```
dani8art, line 1: we should remove `"opencart.mariadb.fullname"`, right?

bitnami/opencart/templates/deployment.yaml
```diff
@@ -1,162 +1,241 @@
...
+      {{- if .Values.metrics.podAnnotations }}
+      {{- toYaml .Values.metrics.podAnnotations | nindent 8 }}
+      {{- end }}
     spec:
-{{- include "opencart.imagePullSecrets" . | indent 6 }}
+      {{- include "opencart.imagePullSecrets" . | nindent 6 }}
+      {{- if .Values.podSecurityContext.enabled }}
+      securityContext: {{- omit .Values.podSecurityContext "enabled" | toYaml | nindent 8 }}
+      {{- end }}
+      {{- if .Values.affinity }}
```
dani8art, line 41: Could we use the new standard for affinity? https://github.com/bitnami/charts/blob/0ffded5c0785040097d76d38a20f15b2f4ae1ea4/bitnami/drupal/templates/deployment.yaml#L41

bitnami/opencart/templates/deployment.yaml
```diff
@@ -1,162 +1,241 @@
...
+          {{- if .Values.args }}
+          args: {{- include "common.tplvalues.render" (dict "value" .Values.args "context" $) | nindent 12 }}
+          {{- end }}
+          {{- if .Values.containerSecurityContext.enabled }}
+          securityContext: {{- omit .Values.containerSecurityContext "enabled" | toYaml | nindent 12 }}
+          {{- end }}
+          env:
+            - name: BITNAMI_DEBUG
+              value: {{ ternary "true" "false" .Values.image.debug | quote }}
+            - name: ALLOW_EMPTY_PASSWORD
```
dani8art, line 70: I would change this to a boolean in values and make a ternary decision here as we do with bitnami debug.

bitnami/opencart/templates/deployment.yaml
```diff
@@ -1,162 +1,241 @@
...
+              value: {{ .Values.allowEmptyPassword | quote }}
+            - name: APACHE_HTTP_PORT_NUMBER
+              value: {{ .Values.containerPorts.http | quote }}
+            - name: APACHE_HTTPS_PORT_NUMBER
+              value: {{ .Values.containerPorts.https | quote }}
+            {{- if .Values.mariadb.enabled }}
```
dani8art, line 76: I think we could follow this approach 
- deployment https://github.com/bitnami/charts/blob/0ffded5c0785040097d76d38a20f15b2f4ae1ea4/bitnami/drupal/templates/deployment.yaml#L81
- helpers https://github.com/bitnami/charts/blob/0ffded5c0785040097d76d38a20f15b2f4ae1ea4/bitnami/drupal/templates/_helpers.tpl#L53

bitnami/opencart/templates/opencart-pvc.yaml
```diff
@@ -1,18 +0,0 @@
-{{- if .Values.persistence.enabled -}}
```
dani8art, line 1: don't we need it?

dani8art (changes requested):
Great PR!!

Please, take a look at my comments

bitnami/opencart/requirements.yaml
marcosbc: good point, done

bitnami/opencart/templates/NOTES.txt
marcosbc: done

bitnami/opencart/templates/_helpers.tpl
marcosbc: Why? It's still being used. Note that we didn't remove it for Drupal either: https://github.com/bitnami/charts/pull/3869/files#diff-3379ebca1e47d3cacbd757308a8916f4R13

bitnami/opencart/templates/deployment.yaml
marcosbc: Good point, changed

bitnami/opencart/templates/deployment.yaml
marcosbc: Changed to follow the same approach as Drupal

bitnami/opencart/templates/deployment.yaml
marcosbc: I guess you are referring to the `else` section which is missing here. I'll add it.

bitnami/opencart/templates/opencart-pvc.yaml
marcosbc: It was renamed to pvc.yaml

bitnami/opencart/README.md
```diff
@@ -168,6 +247,43 @@ See the [Parameters](#parameters) section to configure the PVC or to disable per
...
+
+> NOTE: It is highly recommened to create a backup of your database before migrating your site. The steps below would be only valid if your application (e.g. any plugins or custom code) is compatible with MariaDB 10.5.x
+Obtain the credentials and the name of the PVC used to hold the MariaDB data on your current release:
+
+```console
+export OPENCART_PASSWORD=$(kubectl get secret --namespace default opencart -o jsonpath="{.data.opencart-password}" | base64 --decode)
+export MARIADB_ROOT_PASSWORD=$(kubectl get secret --namespace default opencart-mariadb -o jsonpath="{.data.mariadb-root-password}" | base64 --decode)
+export MARIADB_PASSWORD=$(kubectl get secret --namespace default opencart-mariadb -o jsonpath="{.data.mariadb-password}" | base64 --decode)
+export MARIADB_PVC=$(kubectl get pvc -l app.kubernetes.io/instance=opencart,app.kubernetes.io/name=mariadb,app.kubernetes.io/component=primary -o jsonpath="{.items[0].metadata.name}")
```
dani8art, line 263: I discovered this https://github.com/bitnami/charts/pull/3869/commits/afde3d88bd2ae251b6958224b4cf3376245db37e just before merge the other PR.

bitnami/opencart/templates/_helpers.tpl
dani8art: Up sorry, I don't know how I could read `opencart.fullname`

dani8art (changes requested):

bitnami/opencart/templates/opencart-pvc.yaml
dani8art: 👍

bitnami/opencart/templates/deployment.yaml
dani8art: Correct!

bitnami/opencart/README.md
```diff
@@ -48,74 +48,147 @@ The command removes all the Kubernetes components associated with the chart and
...
+| `opencartLastName`                   | Last Name                                                                                                             | `Name`                                         |
+| `opencartCookieCheckIP`              | Whether to check the cookie's IP address or not                                                                       | `no`                                           |
+| `opencartCountry`                    | Default country of the store                                                                                          | `us`                                           |
+| `opencartLanguage`                   | Default language of the store (iso code)                                                                              | `en`                                           |
+| `opencartSkipInstall`                | Skip OpenCart installation wizard (`no` / `yes`)                                                                      | `no`                                           |
```
andresbono, line 110: There are a few parameters that aren't used.

bitnami/opencart/README.md
```diff
@@ -48,74 +48,147 @@ The command removes all the Kubernetes components associated with the chart and
...
+| `tolerations`                        | Tolerations for pod assignment                                                                                        | `[]` (The value is evaluated as a template)    |
+| `updateStrategy`                     | Deployment update strategy                                                                                            | `nil`                                          |
+
+### Traffic Exposure Parameters
+
+| Parameter                        | Description                           | Default            |
+|----------------------------------|---------------------------------------|--------------------|
+| `service.type`                   | Kubernetes Service type               | `LoadBalancer`     |
+| `service.port`                   | Service HTTP port                     | `8080`             |
+| `service.httpsPort`              | Service HTTPS port                    | `8443`             |
```
andresbono, line 142: These don't match the values.yaml defaults

bitnami/opencart/README.md
marcosbc: good catch, removing

bitnami/opencart/README.md
marcosbc: I'm removing this stuff since the recommended upgrade method will be to migrate (backup + restore)

bitnami/opencart/README.md
marcosbc: good catch, fixing

bitnami/opencart/templates/_helpers.tpl
```diff
@@ -53,135 +21,110 @@ Note, returns 127.0.0.1 if using ClusterIP.
...
 Gets the host to be used for this application.
 If not using ClusterIP, or if a host or LoadBalancerIP is not defined, the value will be empty.
+When using Ingress, it will be set to the Ingress hostname.
 */}}
 {{- define "opencart.host" -}}
+{{- if .Values.ingress.enabled }}
+{{- $host := (index .Values.ingress.hosts 0).name | default "" -}}
```
dani8art, line 28: I think this could be problematic, this `index .Values.ingress.hosts 0` force `hosts` to have at least 1 element when `ingress.enabled=true` otherwise it will fail. Although we are offering `ingress.hostname, it could not be used since  `hosts` needs to have at least 1 element then:

- if you use `ingress.enabled=true,ingress.hostsname=myhost.com` this helper will fail
- if you use `ingress.enabled=true,ingress.hosts[0].name=myhost.com` this helper will not fail but you will have an entry in your ingress which hosts will be empty.
- if you use `ingress.enabled=true,ingress.hostname=myhost.com,,ingress.hosts[0].name=myhost.com` this helper will not fail but will have a duplicate entry in your ingress.

bitnami/opencart/templates/deployment.yaml
```diff
@@ -1,162 +1,230 @@
...
+          args: {{- include "common.tplvalues.render" (dict "value" .Values.args "context" $) | nindent 12 }}
+          {{- end }}
+          {{- if .Values.containerSecurityContext.enabled }}
+          securityContext: {{- omit .Values.containerSecurityContext "enabled" | toYaml | nindent 12 }}
+          {{- end }}
+          env:
+            - name: BITNAMI_DEBUG
+              value: {{ ternary "true" "false" .Values.image.debug | quote }}
+            - name: ALLOW_EMPTY_PASSWORD
+              value: {{ ternary "true" "false" .Values.allowEmptyPassword | quote }}
```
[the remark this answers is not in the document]
marcosbc: The container accepts both, but you're right, let's use `yes/no` for consistency with docs

bitnami/opencart/templates/svc.yaml
```diff
@@ -28,5 +30,4 @@ spec:
       {{- if (and (eq .Values.service.type "NodePort") (not (empty .Values.service.nodePorts.https)))}}
```
dani8art, line 30: we need to add an else that set `nodePorts: null` when `service.type=ClusterIP|LoadBalancer`, something like https://github.com/bitnami/charts/blob/8ee923ce98e69ef9e4d0f660600d648bda187a0d/bitnami/ghost/templates/svc.yaml#L23 so that we can upgrade the service to change the type

bitnami/opencart/templates/deployment.yaml
```diff
@@ -1,162 +1,230 @@
...
+            - name: OPENCART_DATABASE_NAME
+              value: {{ include "opencart.databaseName" . | quote }}
+            - name: OPENCART_DATABASE_USER
+              value: {{ include "opencart.databaseUser" . | quote }}
+            - name: OPENCART_DATABASE_PASSWORD
+              valueFrom:
+                secretKeyRef:
+                  name: {{ include "opencart.databaseSecretName" . }}
+                  key: mariadb-password
+            {{- $port:=.Values.service.port | toString }}
```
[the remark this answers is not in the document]
marcosbc: For `OPENCART_HOST`:

```yaml
{{- $port:=.Values.service.port | toString }}
- name: OPENCART_HOST
  value: "{{ include "opencart.host" . }}{{- if ne $port "80" }}:{{ .Values.service.port }}{{ end }}"
```

bitnami/opencart/templates/ingress.yaml
```diff
@@ -2,42 +2,40 @@
...
       paths:
       - path: {{ default "/" .path }}
         backend:
-          serviceName: {{ template "opencart.fullname" $ }}
+          serviceName: "{{ include "common.names.fullname" $ }}"
           servicePort: http
   {{- end }}
-  tls:
-  {{- range .Values.ingress.hosts }}
   {{- if .tls }}
```
dani8art, line 38: wouldn't it be 
```suggestion
  {{- if .Values.ingress.tls }}
```

dani8art (changes requested):

bitnami/opencart/templates/ingress.yaml
marcosbc: It would not, because of the `range`

bitnami/opencart/templates/svc.yaml
marcosbc: Good point, changing

bitnami/opencart/templates/_helpers.tpl
marcosbc: It seems like you're right, good catch! I'm changing this

bitnami/opencart/templates/ingress.yaml
andresbono: Same for `{{- if .certManager }}`?

bitnami/opencart/README.md
```diff
@@ -48,74 +48,178 @@ The command removes all the Kubernetes components associated with the chart and
...
+| `opencartHost`                       | OpenCart host to create application URLs (when ingress, it will be ignored)                                           | `nil`                                          |
+| `opencartUsername`                   | User of the application                                                                                               | `user`                                         |
+| `opencartPassword`                   | Application password                                                                                                  | _random 10 character long alphanumeric string_ |
+| `opencartEmail`                      | Admin email                                                                                                           | `user@example.com`                             |
+| `opencartSkipInstall`                | Skip OpenCart installation wizard                                                                                     | false                                          |
```
andresbono, line 107:
```suggestion
| `opencartSkipInstall`                | Skip OpenCart installation wizard                                                                                     | `false`                                        |
```

bitnami/opencart/templates/ingress.yaml
dani8art: Yes, both are out of the range

bitnami/opencart/README.md
marcosbc: fixed

bitnami/opencart/templates/ingress.yaml
marcosbc: I see now, you were right. Fixed now.

andresbono (approved):
LGTM

PR #3918, opened 2020-10-06
[bitnami/elasticsearch] add resources section to sysctl init container

mikolajkosmal:
**Description of the change**
- Add `resources` section for sysctl init container
- Fix `resources`conditional statement for volume-permissions init container

**Benefits**

Allows to specify resource requests and limits for sysctl and volume-permissions init containers

**Checklist** 


carrodher (approved):
Thanks for your contribution! I just added a commit bumping the chart version in a minor instead of the patch because this can be considered a feature addition instead of a bug fix; the rest of the changes LGTM

Once merged the changes our internal test & release pipeline is triggered and, if everything works as expected, the new version should be pushed to the Helm registry after some time.

PR #3929, opened 2020-10-07
[bitnami/redis] drop an option from the README that does not have a default value

beltran-rubo:
**Description of the change**

Drop securityContext.sysctls from README as it does not have a specific default value. Not needed to bump the version.

**Benefits**

Consistency

**Checklist** 


carrodher:
Is this case the only one failing? There is for example this parameter https://github.com/bitnami/charts/blob/17670384503c06a79a0ba8fddc62f1a1282087bf/bitnami/redis/values.yaml#L302 in the values with the same entry in the README table:
| master.schedulerName | Name of an alternate scheduler | nil |
-- | -- | --

is not this one also failing?

beltran-rubo:
It seems not, maybe the validation do not check 'nil' ones

PR #3930, opened 2020-10-07
[bitnami/postgresq-ha] Add enable default option for Volume SecurityContext

beltran-rubo:
**Description of the change**

Add the default value for the option documented in the README. Not needed to bump the version.

**Benefits**

Consistency

**Checklist** 


carrodher:
At this moment, the `volumePermissions.securityContext.enabled` parameter is not being used, so adding it to the values without adding the logic in the statefulset is useless.
In the same way, if we want to add it, the default value should be `true` as the _"volumePermission"_ initContainer should be executed always as root in order to be able to change the permissions, that is why those _"volumePermissions"_ initContainers doesn't have a value to enable/disable the security context, so I don't think it as sense to add this parameter.

For that reason, I think the desired approach here is to remove the line from the README, what can be done without bumping the chart version, any other change in the values or templates should have a version bump associated.

carrodher (approved):
LGTM

Issue #3939, opened 2020-10-07
Spring Cloud Data Flow External Database Issues


tkvangorder:
Chart: bitnami/spring-cloud-dataflow, Version  0.7.4

I am trying to set the provided values such that I can use MS SQL Server as my database to host the task/job schemas.
I first tried using the externalDatabase.* properties but the formatting for the JDBC URL is not correct. For my experiment, I have just deployed an MSSQL Server into my cluster and I am using a service named "mssql-server".

```
externalDatabase:
  host: mssql-server
  port: 1433
  scheme: "sqlserver"
  driver: "com.microsoft.sqlserver.jdbc.SQLServerDriver"
  dataflow:
    user: dataflow
    password: "<mypassword>"
  skipper:
    user: dataflow
    password: "<mypassword>"
  hibernateDialect: "org.hibernate.dialect.SQLServer2012Dialect"

```

Helm correctly applies the chart to my cluster but once I hit the dataflow server, I see an error in the logs: 

```
Caused by: org.springframework.jdbc.support.MetaDataAccessException: Could not get Connection for extracting meta-data; nested exception is org.springframework.jdbc.CannotGetJdbcConnectionException: Failed to obtain JDBC Connection; nested exception is com.microsoft.sqlserver.jdbc.SQLServerException: The port number 1433/dataflow is not valid.
```

It feels like the provided properties for customization are geared towards mariadb and do not work when attempting to use MSSQL Server.

**Expected behavior**
Ideally, I would like to just specify the jdbc url via :  
```
externalDatabase:
  url: jdbc:sqlserver://mssql-server
.
.
.
```

In my second attempt to do this, I attempted to set the JDBC URL via an environment variable: 

```
server:
  extraEnvVars:
  - name: SPRING_DATASOURCE_URL
    value: "jdbc:sqlserver://mssql-server"
  - name: SPRING_DATASOURCE_USERNAME
    value: "dataflow"
  - name: SPRING_DATASOURCE_PASSWORD
    value: "<my password>"
```

Unfortunately, this results in an error when I attempt to apply the chart: 

```
helm install my-release -f ./scdf-settings.yaml bitnami/spring-cloud-dataflow
[...]

PR #3961, opened 2020-10-08
[bitnami/spring-cloud-dataflow] Fixes external database config

tkvangorder:
This adds an "externalDatabase.url" property that gives the user better control when using an external/managed database. Prior to this fix, the combination of schema, host, port, database, and jdbc.parameters were used to derive the JDBC url and the resulting URL was not formatted correctly when using an alternate database such as MS SQL Server.

If a user defines the "url" property, this value will be used instead of composing the "url" from the other properties. I left the existing properties because I didn't want to break backwards compatibility, but I think it would make more sense to just allow a user to always use the "url" property. 

The second fix is to address an issue when adding environment variables via server.extraEnvVars. Prior to this fix, the extra environment variables were resulting in a malformed configuration file when the template was applied.

**Benefits**

This makes it much easier to use an external/managed database.

**Possible drawbacks**

It is a bit confusing having both a "url" property and also having separate properties (scheme, host, port, database, and properties) that essentially can also be used to derive a URL. I would recommend removing the individual properties because the resulting URL does not work for all vendor implementations.

**Applicable issues**

https://github.com/bitnami/charts/issues/3939

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/spring-cloud-dataflow/Chart.yaml
+++ b/bitnami/spring-cloud-dataflow/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: spring-cloud-dataflow
-version: 1.1.0
+version: 1.2.0
 appVersion: 2.6.3
 description: Spring Cloud Data Flow is a microservices-based toolkit for building streaming and batch data processing pipelines in Cloud Foundry and Kubernetes.
 keywords:
--- a/bitnami/spring-cloud-dataflow/README.md
+++ b/bitnami/spring-cloud-dataflow/README.md
@@ -244,12 +244,16 @@ The following tables lists the configurable parameters of the Spring Cloud Data 
 | `mariadb.auth.password`                      | Password for the new user                                                                              | `change-me`                                             |
 | `mariadb.auth.rootPassword`                  | Password for the MariaDB `root` user                                                                   | _random 10 character alphanumeric string_               |
 | `mariadb.initdbScripts`                      | Dictionary of initdb scripts                                                                           | Check `values.yaml` file                                |
+| `externalDatabase.driver`                    | The fully qualified name of the JDBC Driver class                                                      | `""`                                                    |
+| `externalDatabase.scheme`                    | The scheme is a vendor-specific or shared protocol string that follows the "jdbc:" of the URL          | `""`                                                    |
 | `externalDatabase.host`                      | Host of the external database                                                                          | `localhost`                                             |
 | `externalDatabase.port`                      | External database port number                                                                          | `3306`                                                  |
 | `externalDatabase.password`                  | Password for the above username                                                                        | `""`                                                    |
 | `externalDatabase.existingPasswordSecret`    | Existing secret with database password                                                                 | `""`                                                    |
+| `externalDatabase.dataflow.url`              | JDBC URL for dataflow server. Overrides external scheme, host, port, database, and jdbc parameters.    | `""`                                                    |
 | `externalDatabase.dataflow.username`         | Existing username in the external db to be used by Dataflow server                                     | `dataflow`                                              |
 | `externalDatabase.dataflow.database`         | Name of the existing database to be used by Dataflow server                                            | `dataflow`                                              |
+| `externalDatabase.skipper.url`               | JDBC URL for skipper. Overrides external scheme, host, port, database, and jdbc parameters.            | `""`                                                    |
 | `externalDatabase.skipper.username`          | Existing username in the external db to be used by Skipper server                                      | `skipper`                                               |
 | `externalDatabase.skipper.database`          | Name of the existing database to be used by Skipper server                                             | `skipper`                                               |
 | `externalDatabase.hibernateDialect`          | Hibernate Dialect used by Dataflow/Skipper servers                                                     | `""`                                                    |
@@ -376,6 +380,7 @@ Sometimes you may want to have Spring Cloud components connect to an external da
 
 ```console
 mariadb.enabled=false
+externalDatabase.scheme=mariadb
 externalDatabase.host=myexternalhost
 externalDatabase.port=3306
 externalDatabase.password=mypassword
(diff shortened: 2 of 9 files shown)


tkvangorder:
I am quite new to helm, so not sure if everything I have done in here is correct, I appreciate any feedback to make this better.

bitnami/spring-cloud-dataflow/README.md
```diff
@@ -374,17 +375,16 @@ Only one messaging layer can be used at a given time.
...
-externalDatabase.host=myexternalhost
-externalDatabase.port=3306
+externalDatabase.url=jdbc:sqlserver://mssql-server:1433
 externalDatabase.password=mypassword
 externalDatabase.dataflow.user=mydataflowuser
-externalDatabase.dataflow.database=mydataflowdatabase
 externalDatabase.dataflow.user=myskipperuser
-externalDatabase.dataflow.database=myskipperdatabase
+externalDatabase.hibernateDialect=org.hibernate.dialect.SQLServer2012Dialect
+
```
FraPazGal, line 387: I would leave the previous external database definition as we are not bumping a major and most users will still be using the old approach. We could add a note or a new console example using the `externalDatabase.url` parameter instead, what do you think?

bitnami/spring-cloud-dataflow/README.md
```diff
@@ -242,6 +242,7 @@ The following tables lists the configurable parameters of the Spring Cloud Data
 | `mariadb.db.user`                            | Username of new user to create                                                                         | `dataflow`                                              |
 | `mariadb.db.password`                        | Password for the new user                                                                              | `change-me`_                                            |
 | `mariadb.initdbScripts`                      | Dictionary of initdb scripts                                                                           | Check `values.yaml` file                                |
+| `externalDatabase.url`                       | URL for the external database. Overrides external scheme, host, port, database, and jdbc parameters.   | `""`                                                    |
```
FraPazGal, line 245: Just a small petition. I just saw that `externalDatabase.driver` and `externalDatabase.scheme` are not documented in the README. Could you please add them?

FraPazGal (changes requested):
Hi @tkvangorder,

Thanks for the PR! I'm happy to help you! 

The fix for the extra envs works great! I have a couple of suggestions, please take a look at my comments.

Also, please `pull` from [bitnami/charts] repo as a new major of this chart was just released.

bitnami/spring-cloud-dataflow/README.md
tkvangorder: Good suggestion, done.

bitnami/spring-cloud-dataflow/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: spring-cloud-dataflow
-version: 0.7.4
+version: 0.7.5
```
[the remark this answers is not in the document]
tkvangorder: OK, after rebasing I have set this to 1.1.0.

bitnami/spring-cloud-dataflow/values.yaml
```diff
@@ -761,6 +761,12 @@ mariadb:
 ## All of these values are ignored when mariadb.enabled is set to true
 ##
 externalDatabase:
+
+  ## Database JDBC URL
+  ## This provides a mechanism for providing a fully customized JDBC URL rather than having it
+  ## derived from the attributes below. This property, when defined, has precedence over the individual attributes.
+  # url:
```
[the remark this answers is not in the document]
tkvangorder: Fixed

bitnami/spring-cloud-dataflow/templates/_helpers.tpl
```diff
@@ -149,6 +149,17 @@ Return true if a configmap object should be created for Spring Cloud Skipper
...
 {{- end -}}
 
+{{/*
+Return the database URL
+*/}}
+{{- define "scdf.database.url" -}}
+  {{- if .Values.externalDatabase.url }}
+    {{- printf "%s" .Values.externalDatabase.url -}}
+  {{- else -}}
+    {{- printf "" -}}
```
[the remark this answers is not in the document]
tkvangorder: Good idea, this cleaned things up. I had to create two methods for getting the jdbc URL : one for dataflow and then one for skipper because the database name (which is included in the URL when using mariadb) was different. This begs the question of if we should put the "url" property under both dataflow and skipper to be consistent?

bitnami/spring-cloud-dataflow/README.md
tkvangorder: OK, I left the original configuration and then I added an additional section to demonstrate how to use MS SQL Server instead

tkvangorder:
OK, getting closer. A couple of suggestions for the next major release of this template:  

The JDBC SPI specification leaves it to each vendor on the format of their JDBC URL. Just as an example of the variety:
 
https://docs.bmc.com/docs/AtriumOrchestratorPlatform/79/database-urls-and-drivers-592708777.html

Given that, it might be better to drop the individual properties that are used to derive the URL because the format of the URL when using the those properties is really specific to how MariaDB has decided to format the URL and does not work well with other vendors. (I think most folks using an external database that is not MariaDB will be forced to use the URL property anyways)

A second note, there are two properties being added to the data source configuration that should also be optional: 

```
        testOnBorrow: true
        validationQuery: "SELECT 1"
```

New versions of Spring are using the Hikari database connection pool which does not use those two attributes and these can be dropped.  https://stackoverflow.com/questions/53870473/hikari-and-test-on-borrow-option

Thanks for you feedback, I really appreciate it.

bitnami/spring-cloud-dataflow/README.md
```diff
@@ -384,7 +388,20 @@ externalDatabase.dataflow.user=myskipperuser
...
+NOTE: When using the indidual propertes (scheme, host, port, database, an optional jdbcParameters) this chart will format the JDBC URL as `jdbc:{scheme}://{host}:{port}/{database}{jdbcParameters}`. The URL format follows that of the MariaDB database drive but may not work for other database vendors.
+
+To use an alternate database vendor (other than MariaDB) you can use the `externalDatabase.url` to fully define the format of the JDBC URL. If this property is defined, it will take precendence over the individual attributes. As an example of configuring an external MS SQL Server database:
+
+```console
+mariadb.enabled=false
+externalDatabase.url=jdbc:sqlserver://mssql-server:1433
+externalDatabase.password=mypassword
+externalDatabase.dataflow.user=mydataflowuser
+externalDatabase.dataflow.user=myskipperuser
```
FraPazGal, line 400:
```suggestion
externalDatabase.skipper.user=myskipperuser
```

Small typo here.

bitnami/spring-cloud-dataflow/templates/_helpers.tpl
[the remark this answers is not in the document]
FraPazGal: You are right. Since `url` is formed with `externalDatabase.xxx.database`, we should introduce both `externalDatabase .xxx.url`.

FraPazGal:
Thank you for the work and your suggestions too!

I agree that the individual properties are not that useful now that we would have the `url` parameter. Nevertheless, as it is an improvement that can be introduced without breaking compatibility and thus requiring a major, it is recommended to do so. There is also the fact that this chart just received a new major and there is no big changes left to do in the backlog.

We will keep in mind your suggestions for our next major so that we can improve our chart further down the line.

tkvangorder:
I have separated the jdbc url into one for dataflow and one for skipper.

dani8art (approved):
Cool! very cool improvement! thanks!

FraPazGal (approved):
Everything looks good, great job! Thanks again for the PR!

PR #3978, opened 2020-10-11
[bitnami/cassandra] align existingSecret attribute to documentation

### LINKED ISSUE #3979: [bitnami/cassandra] Unable to use existingSecret
**Which chart**:
Cassandra, v6.0.1 (latest)

**Describe the bug**
Unable to use an existingSecret, when this attribute is specified the helm chart is no longer able to render.

**To Reproduce**
1. Set `dbUser.existingSecret` to any secret name of your choice
2. Render helm chart (`helm template`)
3. See error:

```Error: render error in "cassandra/templates/NOTES.txt": template: cassandra/templates/NOTES.txt:1:31: executing "cassandra/templates/NOTES.txt" at <include "common.secrets.key" (dict "existingSecret" .Values.dbUser.existingSecret "key" "cassandra-password")>: error calling include: template: cassandra/charts/common/templates/_secrets.tpl:43:24: executing "common.secrets.key" at <.existingSecret.keyMapping>: can't evaluate field keyMapping in type interface {}```

**Expected behavior**
Able to specify a existing secret, so no random password will get generated

- Output of `helm version`:
```Version.BuildInfo{Version:"v3.3.0", GitCommit:"8a4aeec08d67a7b84472007529e8097ec3742105", GitTreeState:"dirty", GoVersion:"go1.14.7"}```

```
Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.3", GitCommit:"2e7996e3e2712684bc73f0dec0200d64eec7fe40", GitTreeState:"clean", BuildDate:"2020-05-20T12:52:00Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}
```

**Additional context**
I am currently thinking this bug is caused by the different expected definition of `existingSecret` in https://github.com/bitnami/charts/blob/3e7916d0e39fcc88b13e17bab6144ac90aa7603a/bitnami/cassandra/templates/statefulset.yaml#L125 (expects string), and https://github.com/bitnami/charts/blob/3e7916d0e39fcc88b13e17bab6144ac90aa7603a/bitnami/cassandra/templates/NOTES.txt#L1 (expects object with this layout: https://github.com/bitnami/charts/tree/master/bitnami/common#existingsecret )

I tried working on a fix in PR #3978 , but can not get the linter to pass. I don't have an idea how to reference the 'cassandra-password' attribute without using the '-' (dash) character. Thus I'm currently stuck at this point.

### PR DESCRIPTION
**Pull request by ailox**
**Description of the change**

The current layout of existingSecret does not match https://github.com/bitnami/charts/tree/master/bitnami/common#existingsecret, so either the instruction in NOTES.txt or in statefulset.yaml will fail, as they both expect a different layout.

To reproduce the current error:
- Set any `existingSecret` value as simple string
- apply the helm chart
- get error

example error
```
[...] "cassandra/templates/NOTES.txt" at <include "common.secrets.key" (dict "existingSecret" .Values.dbUser.existingSecret "key" "cassandra-password")>: error calling include: template: cassandra/charts/common/templates/_secrets.tpl:43:24: executing "common.secrets.key" at <.existingSecret.keyMapping>: can't evaluate field keyMapping in type interface {}
```

**Benefits**

As an accidental benefit, the user is now able to configure the key inside the secret used for storing the existing Secret, so it no longer has to be 'cassandra-password'

**Possible drawbacks**

It is no longer sufficient to specify a simple String for `existingSecret`, the user now has to adhere to the format laid out in https://github.com/bitnami/charts/tree/master/bitnami/common#existingsecret .

This will break compatibility. On the other hand, this functionality is **broken right now anyways**, so I don't think a major version bump is justified, but I want to leave that decision up to you, the maintainer.

**Applicable issues**

  - fixes #3979

**Additional information**

**Checklist** 
- [x] Chart version bumped in `Chart.yaml`
- [x] linter is happy

I am currently stuck with getting the linter to pass, since the original 'existingSecret' key, 'cassandra-password' contains a dash.

### DIFF
--- a/bitnami/cassandra/values.yaml
+++ b/bitnami/cassandra/values.yaml
@@ -194,6 +194,9 @@ dbUser:
   forcePassword: false
   # password:
   # existingSecret:
+  #  name: mySecret
+  #  keyMapping:
+  #    cassandra-password: myKey
 
 ## ConfigMap with cql scripts. Useful for creating a keyspace
 ## and pre-populating data
--- a/bitnami/cassandra/values.yaml
+++ b/bitnami/cassandra/values.yaml
@@ -193,10 +193,19 @@ dbUser:
   user: cassandra
   forcePassword: false
   # password:
+
+  ## Use an existing secrets which already stores your password data.
+  ## for backwards compatibility, existingSecret can be a simple string,
+  ## referencing the secret by name.
   # existingSecret:
-  #  name: mySecret
-  #  keyMapping:
-  #    cassandra-password: myKey
+  #   ## Name of the existing secret
+  #   ##
+  #   name: mySecret
+  #   ## Key mapping where <key> is the value which the deployment is expecting and
+  #   ## <value> is the name of the key in the existing secret.
+  #   ##
+  #   keyMapping:
+  #     cassandra-password: myCassandraPasswordKey
 
 ## ConfigMap with cql scripts. Useful for creating a keyspace
 ## and pre-populating data
--- a/bitnami/cassandra/Chart.yaml
+++ b/bitnami/cassandra/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: cassandra
-version: 6.0.1
+version: 6.0.2
 appVersion: 3.11.8
 description: Apache Cassandra is a free and open-source distributed database management system designed to handle large amounts of data across many commodity servers, providing high availability with no single point of failure. Cassandra offers robust support for clusters spanning multiple datacenters, with asynchronous masterless replication allowing low latency operations for all clients.
 keywords:
--- a/bitnami/cassandra/templates/NOTES.txt
+++ b/bitnami/cassandra/templates/NOTES.txt
@@ -1,5 +1,5 @@
 {{- $cassandraPasswordKey := ( include "common.secrets.key" (dict "existingSecret" .Values.dbUser.existingSecret "key" "cassandra-password") ) -}}
-{{- $cassandraSecretName := (include "common.names.fullname" .) -}}
+{{- $cassandraSecretName := ( include "common.secrets.name" (dict "existingSecret" .Values.dbUser.existingSecret "context" $) ) -}}
 
 ** Please be patient while the chart is being deployed **
 
--- a/bitnami/cassandra/templates/statefulset.yaml
+++ b/bitnami/cassandra/templates/statefulset.yaml
@@ -122,7 +122,7 @@ spec:
             - name: CASSANDRA_PASSWORD
               valueFrom:
                 secretKeyRef:
-                  name: {{ if .Values.dbUser.existingSecret }}{{ .Values.dbUser.existingSecret }}{{- else }}{{ include "common.names.fullname" . }}{{- end }}
+                  name: {{ include "common.secrets.name" (dict "existingSecret" .Values.dbUser.existingSecret "context" $) }}
                   key: {{ include "common.secrets.key" (dict "existingSecret" .Values.dbUser.existingSecret "key" "cassandra-password") }}
             - name: POD_IP
               valueFrom:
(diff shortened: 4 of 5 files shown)

### CODE REVIEW

**Review by FraPazGal — CHANGES_REQUESTED** [review]
Hi @ailox,

Thanks for the PR! This is indeed a bug introduced in the last major. We have decided to modify a bit how to solve this issue; we would introduce a patch on [bitnami/common] to allow backward compatibility in `common.secrets.name` and `common.secrets.key`. You can see said PR #4006. 

With this approach, we only need to modify one line on `NOTES.txt` from this chart, which means it would only be a patch instead of a new major.

If you want, you could introduce the change and merge it using this PR as we found the bug thanks to you. The change would only require the bump to the patch version and the following change to `NOTES.txt`:

```diff
- {{- $cassandraSecretName := (include "common.names.fullname" .) -}}
+ {{- $cassandraSecretName := ( include "common.secrets.name" (dict "existingSecret" .Values.dbUser.existingSecret "context" $) ) -}}
```

ailox:
@FraPazGal, very happy about this. I took the liberty to make the StatefulSet respect the user specified secret key, since currently it is hardcoded to 'cassandra-password'.

Thought this is well within the scope of this bug, but can also split into a seperate PR.
Cheers!

bitnami/cassandra/values.yaml
```diff
@@ -194,6 +194,9 @@ dbUser:
   forcePassword: false
   # password:
   # existingSecret:
+  #  name: mySecret
+  #  keyMapping:
+  #    cassandra-password: myKey
```
FraPazGal, line 199: I'm not sure we should add these sub-parameters like this. It can be confusing for people that already upgraded to this major, as it seems they are mandatory. A solution will be to add a comment explaining that it accepts both structures (only defining `existingSecret: secretname` and the new standard using common helpers). Please take a look how this is implemented at the bitnami/ghost chart's [values.yalm](https://github.com/bitnami/charts/blob/fafe2a30b1e46d1ed5f7b325e042d09d026ab801/bitnami/ghost/values.yaml#L99) and README.

You will also need to replicate any changes done in `values.yaml` on `values-production.yaml`

FraPazGal (commented):
Hi!

I agree with your change to the statefulset so that it uses common helpers. We could use common for the secret.name. Please take a look at my comments.

Also, if I'm not mistaken, you will still need to use `cassandra-password` as the secret key. The only thing the new structure adds is the possibility to use a single secret to define several `<key>:<value>` pairs. As you coded in `statefulset.yaml` `(dict "existingSecret" .Values.dbUser.existingSecret "key" "cassandra-password")`, the common helper will look for the pair with "cassandra-password" as its key.

ailox:
> Also, if I'm not mistaken, you will still need to use `cassandra-password` as the secret key. The only thing the new structure adds is the possibility to use a single secret to define several `<key>:<value>` pairs. As you coded in `statefulset.yaml` `(dict "existingSecret" .Values.dbUser.existingSecret "key" "cassandra-password")`, the common helper will look for the pair with "cassandra-password" as its key.

I'm not sure if I understood your suggestion. What I did was checking by running
```
helm template cassandra/ \
    --set="dbUser.existingSecret.name=hummus" \
    --set="dbUser.existingSecret.keyMapping.cassandra-password=foo"
```

the resulting StatefulSet will contain the expected section of:
```
[...]
            - name: CASSANDRA_PASSWORD
              valueFrom:
                secretKeyRef:
                  name: hummus
                  key: foo
[...]
```

bitnami/cassandra/values.yaml
FraPazGal: You are right, it would only be in very specific cases where there could be confusion. Nevertheless, the backward compatibility comment takes care of it. Thanks!

FraPazGal (approved):
Ah sorry, I misunderstood you. Yes that is the correct use of the new `existingSecret`.

Well, LGTM! Thank you very much for the PR!

PR #3989, opened 2020-10-12
[bitnami/thanos] Enable thanos TLS client in a modular way #3988

### LINKED ISSUE #3988: [bitnami/thanos] Enable thanos querier client TLS certificates in a modular way
**Which chart**:
thanos

Thanos store TLS client enables the usage of TLS validation only though CA, not requiring a certificate's crt or key, therefore while **highly** recommended, these are not mandatory. 

Due to the fact that currently Thanos only supports a single TLS configuration across all stores (feature requested: https://github.com/thanos-io/thanos/issues/977), it is possible to use TLS with multiple stores through the usage of CA-only validation. This is required for use-cases such as the following:

image

Where, for example, the grpc ingresses are:
- https://thanos-grpc.cluster-a.mycloud.dev
- https://thanos-grpc.cluster-b.mycloud.dev
- https://thanos-grpc.cluster-c.mycloud.dev

Enable modular templating for CA, Cert and Key on Thanos Querier client TLS secrets and extra ags without breaking any backwards compatibility.

**Additional information**
I will be creating the PR with this solution as soon as possible, just wanted to create the issue so it would be tracked.

fsschmitt:
**Description of the change**

This change enables the users to specify only the CA when choosing TLS secure client, in order to enable the communication across multiple ingresses that use the same CA, as Thanos Query only supports a global TLS configuration for its stores.

While previously it was required for the user to specify the client TLS CA, Cert and Key with this change it is possible to only pass CA if the user decides to do so.

**Benefits**

As explained on #3988 this would enable the usage of multiple ingresses as target stores if they are all under the same CA.

**Possible drawbacks**

It does not break compatibility as the user is still able to specify all CA, Cert and Key for its TLS client. This only makes it so it is modular enough for the user to only pass the information needed to its use-case.

**Applicable issues**

  - fixes #3988 

**Checklist** 

cc/ @carrodher, @juan131

### DIFF
--- a/bitnami/thanos/Chart.yaml
+++ b/bitnami/thanos/Chart.yaml
@@ -1,5 +1,5 @@
 apiVersion: v1
-version: 2.4.5
+version: 2.4.6
 appVersion: 0.15.0
 description: Thanos is a highly available metrics system that can be added on top of existing Prometheus deployments, providing a global query view across all Prometheus installations.
 engine: gotpl
--- a/bitnami/thanos/templates/querier/deployment.yaml
+++ b/bitnami/thanos/templates/querier/deployment.yaml
@@ -86,9 +86,15 @@ spec:
             {{- end }}
             {{- if .Values.querier.grpcTLS.client.secure }}
             - --grpc-client-tls-secure
+            {{- if .Values.querier.grpcTLS.client.cert }}
             - --grpc-client-tls-cert=/tls/client/cert.pem
+            {{- end }}
+            {{- if .Values.querier.grpcTLS.client.key }}
             - --grpc-client-tls-key=/tls/client/key.pem
+            {{- end }}
+            {{- if .Values.querier.grpcTLS.client.ca }}
             - --grpc-client-tls-ca=/tls/client/ca.pem
+            {{- end }}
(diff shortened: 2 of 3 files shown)


trynity:
Please add this! Currently just ran into this where I only need to add a client cert and key, since we're using nginx ingresses that actually use a different CA for the server cert, and we're only needing to specify client certs.

FraPazGal (approved):
Hi @fsschmitt,

Thanks you very much for the PR! LGTM!

fsschmitt:
Hi @FraPazGal just to understand the process behind it, if you are aware, how long does it normally take for the Bitnami's helm repo to be updated with the newer version (2.4.6) of the chart?

juan131:
It's already published:

```
$ helm search repo bitnami/thanos
NAME          	CHART VERSION	APP VERSION	DESCRIPTION
bitnami/thanos	2.4.6        	0.15.0     	Thanos is a highly available metrics system tha...
```


It usually takes some hours to be updated since it need to pass some internal tests before the publishing process is completed. Find more information at https://github.com/bitnami/charts/blob/master/CONTRIBUTING.md#pr-approval-and-release-process

PR #4006, opened 2020-10-14
[bitnami/common] Allow backward compatibility for existingSecret

FraPazGal:
**Description of the change**

As described in #3979, there is an error with `existingSecrets` using the previous implementation (only providing the secret name). It has been decided to allow backward compatibility in `common` to avoid having to release a new major of said chart.

**Benefits**

It would allow the use of `common.secrets.name` and `common.secrets.key` with the previous implementation of `existingSecret`.

**Possible drawbacks**

None

**Applicable issues**

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/common/Chart.yaml
+++ b/bitnami/common/Chart.yaml
@@ -2,8 +2,8 @@ annotations:
   category: Infrastructure
 apiVersion: v1
 # Please make sure that version and appVersion are always the same.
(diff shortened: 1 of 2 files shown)


javsalgar (approved):
LTGM. Could you add a small comment in the helper to explain the backwards compatibility?

PR #4019, opened 2020-10-15
Update deployment.yaml fix externaldb secret password use

lupieczonka:
external database is using a different secrets password key

**Description of the change**

**Benefits**

**Possible drawbacks**

**Applicable issues**

  - fixes #

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/prestashop/Chart.yaml
+++ b/bitnami/prestashop/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: prestashop
-version: 11.0.0
+version: 11.0.1
 appVersion: 1.7.6-8
 description: A popular open source ecommerce solution. Professional tools are easily accessible to increase online sales including instant guest checkout, abandoned cart reminders and automated Email marketing.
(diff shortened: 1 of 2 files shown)


bitnami/prestashop/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: prestashop
-version: 11.0.0
+version: 11.0.1
```
FraPazGal, line 3: As far as I know, changing the secret key will be a breaking change, which would mean a new major. I'm not really sure about the necessity of this change to justify a new major.

FraPazGal (changes requested):
Hi @lupieczonka,

Thank you for the PR! Is there an associated issue where I could get a bit of context? It would also help if you could explain a bit the change and its benefits/drawbacks. From what I can see, the change revolves around having a different secret key when using an external database, but I'm not sure if I understood it correctly.

Please take a look at my comment so we can discuss further.

PR #4023, opened 2020-10-15
[bitnami/contour] Contour v1.9

Mauraza:
**Description of the change**

Add some changes have been added to make it compatible with v1.9

**Benefits**

**Possible drawbacks**

**Applicable issues**

  - fixes #

**Additional information**

**Checklist** 


rafariossaa (approved):
LGTM, but it would be nice if @dani8art can review it too.

carrodher:
If those changes are required by a new version I am missing here the new Docker images and also the `appVersion` bump, without those changes we are releasing a chart ready to support 1.9 but without effectively supporting it.

In the same way, regarding the chart version (`MAJOR.MINOR.PATCH`), in this case, it is not `PATCH` as it is not bug fixing, it should be `MINOR` if the upgrade works or `MAJOR` if the backward compatibility is not guaranteed

PR #4040, opened 2020-10-16
Phpbb 1602614583

Ruwanthi7:
remove empty lines


miguelaeh (changes requested):
Hi,
This kind of change should be done by our internal pipeline, I think it is not needed to update the images manually

PR #4042, opened 2020-10-16
[bitnami/contour] fix: automatic update of ingress.status by contour

tricky42:
**Description of the change**
As the helm chart uses contour-envoy as the name for the envoy service (which differs from the default envoy service name 'envoy' which contour expects as default), this change adjusts the default configuration to ensure automatic ingress.status updates work out of the box.

**Benefits**
The automatic updating of the Ingress status will work with the default configuration of the Helm-Chart.

**Possible drawbacks**
none

**Checklist** 
- [x] Applied changes to values to both value yaml files

### DIFF
--- a/bitnami/contour/values-production.yaml
+++ b/bitnami/contour/values-production.yaml
@@ -135,7 +135,7 @@ contour:
   image:
     registry: docker.io
     repository: bitnami/contour
-    tag: 1.8.2-debian-10-r0
+    tag: 1.9.0-debian-10-r0
     ## Specify a imagePullPolicy
     ## Defaults to 'Always' if image tag is 'latest', else set to 'IfNotPresent'
     ## ref: http://kubernetes.io/docs/user-guide/images/#pre-pulling-images
@@ -315,7 +315,7 @@ envoy:
   image:
     registry: docker.io
     repository: bitnami/envoy
-    tag: 1.15.2-debian-10-r0
+    tag: 1.15.2-debian-10-r11
     ## Specify a imagePullPolicy
     ## Defaults to 'Always' if image tag is 'latest', else set to 'IfNotPresent'
     ## ref: http://kubernetes.io/docs/user-guide/images/#pre-pulling-images
--- a/bitnami/contour/values-production.yaml
+++ b/bitnami/contour/values-production.yaml
@@ -59,8 +59,8 @@ configInline:
   leaderelection:
     # configmap-name: leader-elect
     configmap-namespace: '{{ .Release.Namespace }}'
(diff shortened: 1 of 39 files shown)


bitnami/contour/values-production.yaml
```diff
@@ -59,6 +59,10 @@ configInline:
   leaderelection:
     # configmap-name: leader-elect
     configmap-namespace: '{{ .Release.Namespace }}'
+  ## This sets the service name that will be inspected for address details to be applied to Ingress objects.
+  # As by default the envoy services name deployed by this helm chart is 'contour-envoy', this configuration is needed
+  # in order to ensure, that the ingress status is automatically updated by contour.
+  envoy-service-name: contour-envoy
```
joancafom, line 65: This is nice, but I think that the name of the Service can actually differ if, for instance, the property `.Values.fullnameOverride` is provided.

Instead, I think I would be great to use the exact same expression used to generate the name:
https://github.com/bitnami/charts/blob/6bf9dfd96083702704461d268895a1e6416fdde3/bitnami/contour/templates/envoy/service.yaml#L4-L6

```suggestion
  envoy-service-name: '{{ include "common.names.fullname" . }}-envoy'

File: bitnami/contour/values-production.yaml [hunk 2]
```diff
@@ -59,6 +59,10 @@ configInline:
   leaderelection:
     # configmap-name: leader-elect
     configmap-namespace: '{{ .Release.Namespace }}'
+  ## This sets the service name that will be inspected for address details to be applied to Ingress objects.
+  # As by default the envoy services name deployed by this helm chart is 'contour-envoy', this configuration is needed
```
joancafom, line 63: I would change this here to remove any mention to 'contour-envoy' and optionally say that the expression used is the same one to generate the name in the Envoy-Service

https://github.com/bitnami/charts/blob/6bf9dfd96083702704461d268895a1e6416fdde3/bitnami/contour/templates/envoy/service.yaml#L4-L6

joancafom (changes requested):
Thank you so much for your PR! Could you please take at my comments?

bitnami/contour/values-production.yaml
tricky42: Yes, done so.

tricky42:
Implemented the requested changes. I think after fighting a bit to find why Contour is not updating the ingress.status although it was fixed in an earlier version of Contour, I couldn't see this better solution here ;)

tricky42 (commented):
Implemented all requested changes.

tricky42 (commented):
Implemented all requested changes.

joancafom (approved):
LGTM! Thanks for your PR 😁

PR #4044, opened 2020-10-17
[bitnami/wordpress] Make primary path configurable

eherot:
Make the default/main path configurable

This is useful if you are deploying with an Application Load Balancer ingress controller because it uses exact matches, so this value will need to be set to `/*`.

**Checklist** 


javsalgar (approved):
Thanks for the PR! LGTM

PR #3849, opened 2020-10-01
[bitnami/mariadb] New major version

juan131:
Signed-off-by: juan131 <juanariza@vmware.com>

**Description of the change**

This PR refactors the whole chart to follow Helm chart best practices. This includes several breaking changes, so please review it carefully, putting special emphasis to the "Upgrading" section describing all of them.

The most important change are:

- The terms *master* and *slave* have been replaced by the terms *primary* and *secondary*.
- The default MariaDB version was updated from `10.3` to `10.5`.
- Labels are adapted to follow the [Helm charts standard labels](https://helm.sh/docs/chart_best_practices/labels/#standard-labels).

**Benefits**

- Standardization.
- Reduces technical debt.

**Benefits**

**Possible drawbacks**

It breaks backwards compatibility.

**Applicable issues**

None

**Additional information**

Upgrade tests from version `7.x.x` using workaround suggested in the Upgrading Notes:

- Install old chart version:

```console
$ helm install mariadb bitnami/mariadb \
    --set rootUser.password=root_password \
    --set replication.enabled=true \
    --set replication.password=rep_password \
    --set db.user=my_username \
    --set db.name=my_database \
    --set db.password=my_password \
    --version 7.10.3
```

- Add some data once everything is up and running:

```console
$ kubectl run mariadb-client --rm --tty -i --restart='Never' --image  docker.io/bitnami/mariadb:10.3.24-debian-10-r17 --namespace default --command -- bash
I have no name!@mariadb-client:/$ mysql -h mariadb.default.svc.cluster.local -uroot -p my_database
Enter password:
...
MariaDB [my_database]> CREATE OR REPLACE TABLE foo (a int);
Query OK, 0 rows affected (0.032 sec)
MariaDB [my_database]> INSERT INTO foo (a) VALUES (10);
Query OK, 1 row affected (0.018 sec)
```

- Remove chart and slave PVC (we only need the master PVC since slave only replicates data):

```console
$ helm uninstall mariadb && kubectl delete pvc data-mariadb-slave-0
$ kubectl get pvc
NAME                    STATUS   VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS   AGE
data-mariadb-master-0   Bound    pvc-d9b6a420-146a-4fad-94eb-9dbace8cacd4   8Gi        RWO            standard       6m49s
```

- Install new chart version reusing existing PVC for primary node:

```console
$ helm install mariadb bitnami/mariadb \
  --set architecture=replication \
[...]

PR #4045, opened 2020-10-17
[bitnami/redis] Add recommended labels to all resources

margaretgorguissian:
**Description of the change**

Adds a template helper with recommended labels (`app.kubernetes.io/name`, `app.kubernetes.io/instance`, `app.kubernetes.io/version`, `app.kubernetes.io/managed-by` and `helm.sh/chart`) to all Redis resources.

**Benefits**

- This is a best practice/recommended by Kubernetes (https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/)
- Related to the above, makes it easier to manage Redis resources with tooling (eg. kubectl) that relies on shared labels

**Possible drawbacks**

~None AFAIK~ **Edit**: As juan131 notes below, this change breaks backwards compatibility with `matchLabels`. 

**Checklist** 

### DIFF
--- a/bitnami/redis/Chart.yaml
+++ b/bitnami/redis/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: redis
-version: 11.2.1
+version: 11.3.1
 appVersion: 6.0.8
 description: Open source, advanced key-value store. It is often referred to as a data structure server since keys can contain strings, hashes, lists, sets and sorted sets.
 keywords:
--- a/bitnami/redis/templates/_helpers.tpl
+++ b/bitnami/redis/templates/_helpers.tpl
@@ -419,3 +419,16 @@ Usage:
         {{- tpl (.value | toYaml) .context }}
     {{- end }}
 {{- end -}}
+
+{{/*
+Common labels
+*/}}
+{{- define "redis.labels" -}}
+app.kubernetes.io/name: {{ include "redis.name" . }}
+helm.sh/chart: {{ include "redis.chart" . }}
+app.kubernetes.io/instance: {{ .Release.Name }}
+{{- if .Chart.AppVersion }}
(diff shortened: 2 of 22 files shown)


juan131 (changes requested):
Thanks so much for this PR @margaretgorguissian 

Please note that a change affecting the default labels will include backwards incompatibilities since there are some fields that are inmutable such as the "matchLabels" (`sts.spec.selector.matchLabels`). Therefore, including a change such as this will require bumping the major version.

We plan to do include these changes since we are standardizing our helm charts in PRs such as the one below:

- https://github.com/bitnami/charts/pull/3849

So, expect this chart (Redis) to updated soon on our side including this and other changes.

Thanks again!!

margaretgorguissian:
Thank you @juan131 for the context around major versioning and the backwards incompatibility. I am new to K8s and Helm, I wasn't aware that it would break `matchLabels`. 😅 Are you looking for contributors to make other changes like in the Maria DB PR? (eg. updating terminology) Thank you!

juan131:
> Thank you @juan131 for the context around major versioning and the backwards incompatibility. I am new to K8s and Helm, I wasn't aware that it would break matchLabels

You're welcome. It's one of the common pitfalls when dealing with Helm charts, we should always pay attention to upgrades.

> Are you looking for contributors to make other changes like in the Maria DB PR? (eg. updating terminology) Thank you!

We are always open to receive contributions!

That said, we'd like to do all the changes (using common library, using recommended labels, parameters renaming, removing technical debt, etc.) in a single PR, since several of these changes breaks upgradability. This will imply quite a big change and it'll be lead by one of the core maintainers of the project due to its complexity. I hope you understand it.

Then, we'll provide an upgrade path from Redis `11.x.x` to `12.x.x` and perform a lot of tests before the new major is released.

PR #4057, opened 2020-10-20
[bitnami/ghost] Fix ingress when tls is enabled

juan131:

**Description of the change**

Creates the `tls-secret.yaml` object, so the tls secret is generated when the `ingress.secrets` array is set. It also ensure the secretName is set in the Ingress object when it's not indicated in the `ingress.hosts` array.

**Benefits**

Users can set their own certificates fo TLS

**Possible drawbacks**

None

**Checklist** 


dani8art (approved):
LGTM

PR #4079, opened 2020-10-22
[bitnami/harbor] Fix SCANNER_REDIS_URL for trivy-scanner

funkypenguin:
**Description of the change**

The latest harbor-trivy-scanner seems to require the env var `SCANNER_REDIS_URL`, whereas previous versions required `SCANNER_STORE_REDIS_URL` and `SCANNER_JOB_QUEUE_REDIS_URL` (although they were set to the same value)

This change updates the chart v8.x.x for the new env var, which is required to make Trivy work. In the absense of this change, the trivy pod will start, but fail to scan, outputting the following:

```
ERROR: worker.fetch - dial tcp [::1]:6379: connect: connection refused
ERROR: worker.fetch - dial tcp [::1]:6379: connect: connection refused
```

**Benefits**

Trivy works!

**Possible drawbacks**

None

  - fixes #

**Additional information**

**Checklist** 
`values.yaml`, ensure that you implement the changes in both files

### DIFF
--- a/bitnami/harbor/templates/trivy/trivy-secret-envvars.yaml
+++ b/bitnami/harbor/templates/trivy/trivy-secret-envvars.yaml
@@ -15,4 +15,6 @@ type: Opaque
 data:
   SCANNER_TRIVY_GITHUB_TOKEN: {{ .Values.trivy.gitHubToken | default "" | b64enc | quote }}
   SCANNER_REDIS_URL: {{ include "harbor.redisForTrivyAdapter" . | b64enc }}
+  SCANNER_STORE_REDIS_URL: {{ include "harbor.redisForTrivyAdapter" . | b64enc }}
+  SCANNER_JOB_QUEUE_REDIS_URL: {{ include "harbor.redisForTrivyAdapter" . | b64enc }}
 {{- end }}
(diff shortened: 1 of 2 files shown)


bitnami/harbor/templates/trivy/trivy-secret-envvars.yaml
```diff
@@ -14,6 +14,5 @@ metadata:
 type: Opaque
 data:
   SCANNER_TRIVY_GITHUB_TOKEN: {{ .Values.trivy.gitHubToken | default "" | b64enc | quote }}
-  SCANNER_STORE_REDIS_URL: {{ include "harbor.redisForTrivyAdapter" . | b64enc }}
```
joancafom, line 17: The upstream project seems to maintain the three env variables, which may be due to backwards compatibility with previous versions of the trivy container. What about keeping the three of them as well here to match this idea?

https://github.com/goharbor/harbor-helm/blob/f5767efaa35f835934ce060aa467b3a00af82837/templates/trivy/trivy-sts.yaml#L104

joancafom (changes requested):
Thanks for your PR! Could you please take a look at my comment?

bitnami/harbor/templates/trivy/trivy-secret-envvars.yaml
funkypenguin: Good idea, that seems safe(r) ;)

joancafom (approved):
LGTM! Thank you for your changes!

PR #4096, opened 2020-10-23
bitnami/consul Use new replicaCount parameter for expected consul servers

### LINKED ISSUE #4095: Consul: replicaCount is not used in StatefulSet
**Which chart**:

consul v8

**Describe the bug**

In the version 8 of the chart the name of the variable `replicas` was changed to `replicaCount`. The old value is still used for determing how many replicas to expect when starting the Consul server. Currently can provide both values as a valid workaround:

```yaml
replicas: 1
replicaCount: 1
``` 

**Additional context**

riker09:
**Description of the change**

With Chart version 8 the parameter `replicas` was renamed to `replicaCount`. However, the old parameter name was still used for the `CONSUL_BOOTSTRAP_EXPECT` variable. This caused Consul to expect 3 servers (and crash indefinitely), when `replicaCount` is set to `1` or `2`. I believe values higher than `3` would actually be no issue, but I have not tested this.

**Applicable issues**

  - fixes #4095

**Checklist**

### DIFF
--- a/bitnami/consul/Chart.yaml
+++ b/bitnami/consul/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: consul
-version: 8.0.2
+version: 8.0.3
 appVersion: 1.8.4
 description: Highly available and distributed service discovery and key-value store designed with support for the modern data center to make distributed systems and configuration easy.
 home: https://github.com/bitnami/charts/tree/master/bitnami/consul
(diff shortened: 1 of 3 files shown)


dani8art (approved):
Hi @riker09 thank you so much!! LGTM

riker09:
> Hi @riker09 thank you so much!! LGTM

You're welcome. I stumbled accross this and thought this wouldn't take much effort to fix. I was right! 😄

Issue #4105, opened 2020-10-24
[bitnami/drupal] Using Drupal helm chart with external db is not working due to password issue


ruzickap:
**Which chart**:
Drupal 9.1.0

**Describe the bug**
I'm trying to install the Drupal with external DB, but I'm getting the error:

```bash
kubectl describe pods -n drupal drupal-6456b67944-7pbmb
...
...
Error: couldn't find key mariadb-password in Secret drupal/drupal-externaldb
```

**To Reproduce**

Install Drupal:

```bash
helm install --version 9.1.0 --namespace drupal --create-namespace --values - drupal bitnami/drupal << EOF
drupalUsername: myuser
drupalPassword: test12345
drupalEmail: petr.ruzicka@gmail.com
externalDatabase:
  host: test.aaa.com
  user: root
  password: test123456
  database: drupal
mariadb:
  enabled: false
EOF
```

Check the pod:

```bash
$ kubectl get pods -n drupal
NAME                      READY   STATUS                       RESTARTS   AGE
drupal-5d4c745447-kj8vv   0/1     CreateContainerConfigError   0          105s

$ kubectl describe pods -n drupal drupal-5d4c745447-kj8vv
...
Warning  Failed                  12s (x2 over 12s)  kubelet                  Error: couldn't find key mariadb-password in Secret drupal/drupal-externaldb
```

The pod can not start because the `mariadb-password` is not in `drupal/drupal-externaldb`:

```bash
$ kubectl get secret --namespace drupal drupal-externaldb -o yaml | head -14
apiVersion: v1
data:
  db-password: dGVzdDEyMzQ1Ng==
kind: Secret
metadata:
  annotations:
    meta.helm.sh/release-name: drupal
    meta.helm.sh/release-namespace: drupal
  creationTimestamp: "2020-10-24T07:30:37Z"
  labels:
    app.kubernetes.io/instance: drupal
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/name: drupal
    helm.sh/chart: drupal-9.1.0
```

It seems like `mariadb-password` can not be used in case of "External DB" usage where `db-password` needs to be used: 
https://github.com/bitnami/charts/blob/1c9371a689a1d4a6bc80905dc5be023ce2006ce8/bitnami/drupal/templates/deployment.yaml#L162

**Expected behavior**
Drupal pod should start without error and connect to the external db

**Version of Helm and Kubernetes**:

- Output of `helm version`:

```
[...]

PR #4026, opened 2020-10-15
[bitnami/prestashop] fix: wrong database password key

dani8art:
Signed-off-by: darteaga <darteaga@vmware.com>

**Description of the change**

Add a helper to decide between different keys when an externalDatabase is set

**Benefits**

We can use external databases

**Possible drawbacks**

N/A

**Applicable issues**

rel #3949

**Additional information**

N/A

**Checklist** 
- [x] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [x] Variables are documented in the README.md
- [x] Title of the PR starts with chart name (e.g. `[bitnami/chart]`)
- [x] If the chart contains a `values-production.yaml` apart from `values.yaml`, ensure that you implement the changes in both files

PR #4106, opened 2020-10-24
[bitnami/drupal] fix external_db password issue (#4105)

ruzickap:
**Description of the change**

The PR is fixing the problem of having MariaDB password in ` mariadb-password` and "external DB" password in `db-password`.

Details are described here: https://github.com/bitnami/charts/issues/4105

**Benefits**

Fix: https://github.com/bitnami/charts/issues/4105

### DIFF
--- a/bitnami/drupal/Chart.yaml
+++ b/bitnami/drupal/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: drupal
-version: 9.1.1
+version: 9.1.2
 appVersion: 9.0.7
 description: One of the most versatile open source content management systems.
 keywords:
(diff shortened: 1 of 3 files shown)


dani8art (changes requested):
Hi @ruzickap thank you so much for your PR, 

Please, could you follow the same approach that we followed here: https://github.com/bitnami/charts/pull/4026.

We would like to have them as standardized as possible. Thanks!

ruzickap:
Make sense...

Code changed according the https://github.com/bitnami/charts/pull/4026.

dani8art (approved):
Thanks!! LGTM

PR #4147, opened 2020-10-28
[bitnami/etcd] Allow client/peer port name overrides for custom SRV domains

lactose:
**Description of the change**
Currently, the port names for client/peer ports (used in generation of SRV records) are hard coded to "client" and "peer". This adds a field to override that value for custom SRV domain records.

**Benefits**

The user has control of the client/peer port names.

**Possible drawbacks**

N/A with defaults.

**Checklist** 


crandles:
FYI this is to better help use the `etcdsrv.GetClient` function: https://github.com/etcd-io/etcd/blob/2c834459e1aab78a5d5219c7dfe42335fc4b617a/pkg/srv/srv.go#L110-L141

rafariossaa:
Hi,
  Thank you very much for your PR.
  I would like to understand better the reason behind this PR.
  Could you elaborate a bit more ?

lactose:
Sorry for the delay here. The reason we did this is because the port names are used to create the etcd SRV domain record. Without overriding, we did not have control over the SVR record that got created: `_client._tcp.release-name.namespace.svc.cluster.local`

for our purposes, we need the SRV domain record to be customized, e.g.: `_etcd-client._tcp.release-name.namespace.svc.cluster.local`

miguelaeh (approved):
Thank you for the PR!
LGTM!

miguelaeh:
Hi,
I have a question before merging this.
I think this change could modify the domains of the current releases if you make an upgrade. At least the `-ssl` one. Could we add backwards compatibilty for that? I am not sure this could cause issues in other user releases

rafariossaa:
LGTM

PR #4151, opened 2020-10-29
[bitnami/moodle] Moodle metrics port

### LINKED ISSUE #4140: [bitnami/moodle] metrics not working
**Which chart**:
bitnami/moodle

**Describe the bug**
I have installed moodle using kubeapps following the wiki installation process you set in the wiki, 
I have set metrics.enable = true and the load balancer do not expose the port 9117

I did the same in bitnami/wordpress and it's working for bitnami/wordpress, but not working for moodle

**To Reproduce**
Install bitnami moodle with metrics.enable = true

**Expected behavior**
The load balancer shoud expose the 9117 port with apache-exporter metrics

FraPazGal:
**Description of the change**

Currently, the metrics sidecar is not exposed by the service. This PR adds the `metricsPort` variable and expose said port when `metrics.enabled=true`

**Benefits**

**Possible drawbacks**

**Applicable issues**

  - fixes #4140

**Additional information**

**Checklist** 


bitnami/moodle/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: moodle
-version: 9.0.0
+version: 9.0.1
```
juan131, line 3: You're adding a new parameter, so please bump the minor version instead.

bitnami/moodle/templates/svc.yaml
```diff
@@ -41,4 +41,14 @@ spec:
...
+      {{- if .Values.metrics.enabled }}
+    - name: metrics
+      port: {{ .Values.service.metricsPort }}
+      targetPort: metrics
+      {{- if (and (or (eq .Values.service.type "NodePort") (eq .Values.service.type "LoadBalancer")) (not (empty .Values.service.nodePorts.metrics))) }}
+      nodePort: {{ .Values.service.nodePorts.metrics }}
+      {{- else if eq .Values.service.type "ClusterIP" }}
+      nodePort: null
+      {{- end }}
+      {{- end }}
```
juan131, line 53: Would it make sense to create a different service for exposing the metrics?
I mean, let's say that I use a "LoadBalancer" service type because I want to expose my Moodle website. If I use this approach, I will be exposing the metrics to the outside world to... However, if there are two different services, I can expose the http/https ports while keeping the metrics private.

juan131 (changes requested):
Thanks for the PR! Please check my comments

bitnami/moodle/templates/svc.yaml
developerbuendata: Hi @juan131, you are right, the better aproch is creating a sidecar, but in this case is not posible because this chart do not support the extraPorts 
in order to support the extraport we need to add the following code
**svc.yaml**
`{ { - if .Values.service.extraPorts } }
{ { - include "moodle.tplValue" (dict "value" .Values.service.extraPorts "context" $) | nindent 4 } }
{ { - end } }`

**_helper.tpl**
`{{/*
Usage:
{{ include "wordpress.tplValue" ( dict "value" .Values.path.to.the.Value "context" $) }}
*/}}
{{- define "moodle.tplValue" -}}
        {{- tpl .value .context }}
    {{- else }}
    {{- end }}
{{- end -}}
`

FraPazGal: I ended up doing what @juan131 suggested and created a separated service for exposing the metrics; `metrics-svc.yaml`

juan131 (approved):
Thanks!! LGTM

PR #4152, opened 2020-10-29
[bitnami/redis-cluster] Only render yaml keys that are lists when there are actual values for them

EDeijl:
**Description of the change**

I changed the templates of the statefulset and jobs to not render keys (e.g. `volumes`, `volumeMounts` and `envFrom`) in the output yamls if they have no items in their lists.

We are using the output of this chart as an input into Kustomize, and kustomize fails on empty lists that are not explicitly set with `[]`, therefore I removed the keys here if they are empty, as that is an easier solution in go templating than adding a `[]` when there are no items.

**Benefits**

Better integration with tools like Kustomize.

**Possible drawbacks**

It might be a bit less clear that `volumes` etc. can be added to when looking at the output. 

**Applicable issues**

N/A

**Additional information**
N/A

**Checklist** 


bitnami/redis-cluster/templates/redis-statefulset.yaml
```diff
@@ -373,8 +377,8 @@ spec:
           requests:
             storage: {{ .Values.persistence.size | quote }}
         {{- include "redis-cluster.storageClass" . | nindent 8 }}
-        selector:
         {{- if .Values.persistence.matchLabels }}
+        selector:
```
andresbono, line 381: With this change, `matchExpressions:` could be orphan, right? What about the following?

```suggestion
      {{- if or .Values.persistence.matchLabels .Values.persistence.matchExpressions }}
        selector:
        {{- if .Values.persistence.matchLabels }}
```

andresbono: You will need to add an `{{- end }}` below, to close the new conditional.

andresbono:
Thanks for this contribution, @EDeijl!! 🎉  Could you please check my comments?

EDeijl:
@andresbono good catches, I fixed them up

andresbono (approved):
LGTM

andresbono:
Thanks for your contribution and for checking the suggestions. The PR is now merged :tada:

PR #4155, opened 2020-10-29
Updated README files for Apache, MySQL and WordPress charts

vikram-bitnami:
**Description of the change**

Updates the README files with additional links and condenses duplicate content

**Benefits**

Smaller, more compact README

**Possible drawbacks**

None

**Applicable issues**

None

  - fixes #

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/apache/README.md
+++ b/bitnami/apache/README.md
@@ -152,7 +152,7 @@ The Apache chart allows you to deploy a custom web application using one of the 
 - Providing a ConfigMap: Set the `htdocsConfigMap` value to mount a ConfigMap in the Apache htdocs folder.
 - Using an existing PVC: Set the `htdocsPVC` value to mount an PersistentVolumeClaim with the web application content.
 
-Learn more about this in the [chart documentation](https://docs.bitnami.com/kubernetes/infrastructure/apache/get-started/deploy-custom-application/).
+Refer to the [chart documentation](https://docs.bitnami.com/kubernetes/infrastructure/apache/get-started/deploy-custom-application/) for more information.
 
 ### Setting Pod's affinity
 
--- a/bitnami/apache/README.md
+++ b/bitnami/apache/README.md
@@ -144,7 +144,7 @@ It is strongly recommended to use immutable tags in a production environment. Th
 
 Bitnami will release a new chart updating its containers if a new version of the main container, significant changes, or critical vulnerabilities exist.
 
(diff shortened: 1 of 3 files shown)


marcosbc (commented):
These changes LGTM but I would also like to check the published docs before merging this PR. I would also like to get @juan131's point of view regarding these changes.

beltran-rubo (changes requested):

juan131:
I have my concerns about removing sections from the chart’s README(s) to add links to https:///docs.bitnami.com.

There are two reasons for not doing so:

- GitHub users contribute to these README.md(s) enriching them and this model is currently working very well. However, they can't contribute if the docs are at https://docs.bitnami.com
- We currently have everything (docs and charts) in a single repo. This simplifies everything.

I agree we have duplicated docs, but this is not the right solution IMHO. If we want https://docs.bitnami.com to be synced with https://github.com/bitnami/charts, then let's create a `docs/` folder and generate https://docs.bitnami.com helm sections based on markdown files at this folder. This way, uses can contribute and the docs are also open-sourced.

@javsalgar @carrodher I'd like to hear your opinion

carrodher:
I agree with Juan's comment, it is nice to see users contributing and improving the docs and it is something we should keep.
In the same way, changing the context from GitHub to docs.bitnami.com is not a good UX, I think it's better to maintain all the documentation in the same place. With "the same place" I mean the GitHub repository, not necessary in the same README.

As per the above comments, we can discuss the implementation of a system like the one shown in the below picture:

Screenshot 2020-11-12 at 15 49 11

In the `bitnami/chartX/` folder we have the regular _README.md_, containing the basic information such as how to deploy, most important parameters & configurations, changelog. That means we should remove information from the current _README.md_ and point to the new markdown files placed in the docs folder.

Then there is a new `docs/` folder that can contain several subfolders but, at least, a `charts/` folder containing a subfolder for every Helm Chart.
In that per chart folder, we can place those sections previously removed from the regular _README.md_ including information about 2nd-day operations such as TLS, LDAP configuration, differences between architectures, etc.

Apart from that, at the same level as the charts folder, we can create other subfolders for different kind of docs, such as `troubleshooting/`, `common/`, etc. in order to add other content that is interesting but that is not specific to a particular chart, like security guidelines, rolling VS immutable tags, etc.

That is just an idea, let's discuss the different options properly with a meeting.

PR #4157, opened 2020-10-29
[bitnami/postgres] Extra manifests should be top-level

vbehar:
**Description of the change**

The "extra manifests" defined with the `extraDeploy` value are now rendered as top-level manifests, instead of being wrapped inside a List.

**Benefits**

This change allows defining Jobs as Helm hooks, which was not possible before - because Helm only looks at top-level manifests to find hooks annotations.

**Additional information**

- I've bumped the major semver component because it won't render the same way - if people used `extraDeploy`.
- There are a lot of charts in this repo defining extra manifests in the same way - do you prefer 1 PR per chart, or a single PR to change it for all charts at once?

**Checklist**

### DIFF
--- a/bitnami/postgresql/Chart.yaml
+++ b/bitnami/postgresql/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: postgresql
-version: 9.8.7
+version: 10.0.0
 appVersion: 11.9.0
 description: Chart for PostgreSQL, an object-relational database management system (ORDBMS) with an emphasis on extensibility and on standards-compliance.
 keywords:
--- a/bitnami/postgresql/templates/extra-list.yaml
+++ b/bitnami/postgresql/templates/extra-list.yaml
@@ -1,5 +1,4 @@
-{{- if .Values.extraDeploy }}
-apiVersion: v1
-kind: List
-items: {{- include "common.tplvalues.render" (dict "value" .Values.extraDeploy "context" $) | nindent 2 }}
+{{- range .Values.extraDeploy }}
+---
+{{ include "common.tplvalues.render" (dict "value" . "context" $) }}
 {{- end }}


juan131:
Thanks for this great suggestion @vbehar !! I personally have no objection against it. 

@dani8art @javsalgar could you please take a look and share your thoughts?

> I've bumped the major semver component because it won't render the same way - if people used `extraDeploy`.

I wouldn't bump the major version (even though users should adapt the **values.yaml** to upgrade). My main reason to avoid doing so is that this a totally optional parameter (it doesn't affect the application to be deployed, PostgreSQL in this case). We can still bump the minor version, and document the changes in the "Upgrading" section of the **README.md** file.

As alternative, we can adapt the PR to support both the current and the new approach, so we deprecate the current approach once there is a real reason to bump the major version. But let's try not to bump the major version for this.

> There are a lot of charts in this repo defining extra manifests in the same way - do you prefer 1 PR per chart, or a single PR to change it for all charts at once?

I think we can do all the changes in a single PR (it would be much easier)

vbehar:
> even though users should adapt the values.yaml to upgrade

in fact there is no need to change anything in the values.yaml: the `extraDeploy` is still an array of manifests - it's just the rendering which is different

I'll rewrite this PR to apply the change to all charts

dani8art:
I personally think that this a better approach to include extra resources to the chart, I agree with @juan131, this is a great suggestion and I have no objection against it.

Regarding the version, I agree with @juan131 as well, I think we wouldn't need to bump the major due to the fact that  we will have the same result in the end, I mean, we are just adding a new feature in the sense that we are going to be able to add hooks and helm is going to take them into account the rest is going to be the same. If I'm not wrong, if we add a service to this list we will create the same service into the k8s API in both versions.

vbehar:
yes you're right. ok so I'll just bump the patch component. thanks!

vbehar:
@juan131 @dani8art I've created #4161 to apply the change to all charts - so I'm closing this PR

PR #4161, opened 2020-10-30
[bitnami/*] Extra manifests should be top-level

vbehar:
**Description of the change**

The "extra manifests" defined with the extraDeploy value are now rendered as top-level manifests, instead of being wrapped inside a List.

**Benefits**

This change allows defining Jobs as Helm hooks, which was not possible before - because Helm only looks at top-level manifests to find hooks annotations.

**Possible drawbacks**

None: users won't need to change anything in their values.yaml

**Additional information**

This is a follow-up on #4157 but applied to all charts (instead of just postgresql)

**Checklist**

### DIFF
--- a/bitnami/aspnet-core/Chart.yaml
+++ b/bitnami/aspnet-core/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: aspnet-core
-version: 0.3.2
+version: 0.3.3
 appVersion: 3.1.9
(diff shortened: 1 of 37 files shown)


miguelaeh (approved):
Thank you very much for taking the time to change all the charts!
LGTM!

vbehar:
thanks for merging it quickly!

PR #4173, opened 2020-11-01
bitnami/rabbitmq: Setting target port to equal port

guyarb:
**Description of the change**

GCP k8s does not allow port forwarding to the machine, since it does not recognize amqp as target port.
Setting the target port to equal the port of the relevant protocols.

**Benefits**

GCP k8s will allow port forwarding to the machine

**Possible drawbacks**

None

**Applicable issues**

None 

  - fixes https://github.com/bitnami/charts/issues/4174

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/rabbitmq/Chart.yaml
+++ b/bitnami/rabbitmq/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: rabbitmq
-version: 7.7.1
+version: 7.7.2
 appVersion: 3.8.9
 description: Open source message broker software that implements the Advanced Message Queuing Protocol (AMQP)
 keywords:
--- a/bitnami/rabbitmq/templates/svc.yaml
+++ b/bitnami/rabbitmq/templates/svc.yaml
@@ -26,7 +26,7 @@ spec:
   ports:
     - name: amqp
       port: {{ .Values.service.port }}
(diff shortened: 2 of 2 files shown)


miguelaeh (commented):
Hi @guyarb ,
In general the target port has not to be the same as the service port. It can be different since it is the pod's port and the labels used should be correct as they are defined here: Setting target port to equal port

guyarb:
> Hi @guyarb ,
> In general the target port has not to be the same as the service port. It can be different since it is the pod's port and the labels used should be correct as they are defined here: Setting target port to equal port

Hey @miguelaeh, thanks for the quick reply. I'm not sure I understood you correctly. I understand your point the target port might be different, but how do you suggest the fix should be?
What about adding the target ports to the values.yaml and allowing users to override them?

Spittal:
Hey just wanna check in here and mention that the headless service yaml file should be edited as part of this PR with the same fix.

miguelaeh:
>What about adding the target ports to the values.yaml and allowing users to override them?

I agree with this, but, in that case it would be needed to set the container port to the same value to be consitent and the proper env vars to configure the ports in container.

I would like to understand better the issue. Why ports can't be set using their names? Is this a limitation of GCP itself? In that case, I think we would need to check all the charts.

PR #4187, opened 2020-11-02
[bitnami/moodle] Fix docs and examples re. externalDatabase.user

michzimny:
**Description of the change**

Moodle chart: This change fixes documentation and `values.yaml` re. field `externalDatabase.user`. The name of the field is actually `user` (see [here](https://github.com/bitnami/charts/blob/master/bitnami/moodle/templates/_helpers.tpl#L102)) but it's incorrectly named `username` in those places.

**Benefits**

The docs and the example config will reflect the actual implementation.

**Possible drawbacks**

None.

**Applicable issues**

None.

**Additional information**


alemorcuq (changes requested):
Hi, @michzimny. Thanks for fixing this, the change looks good to me. However, can you please increase the version again? Version `9.0.2` is already out with your other fix.

alemorcuq (approved):
Thanks!

PR #4191, opened 2020-11-03
[bitnami/redis] Fix: "sh -c" for volume-permissions initContainer

iwilltry42:
Add the /bin/sh -cx command in front of the "chown" stuff to avoid "file
not found" and similar errors. Similar to how it's done in the Postgres
chart.

Signed-off-by: Thorsten Klein <iwilltry42@gmail.com>

**Description of the change**

Small change adding `/bin/sh -cx` in front of the `chown` commands in the `volume-permissions` initContainers of the Redis chart. Without it, I was running into "File not Found" and similar errors when using `volumePermissions.enabled`, as the command consists of several parts, which otherwise will be treated as a single string.

**Benefits**

No more errors :tada: 

**Possible drawbacks**

None that I can think of.

**Applicable issues**

None.

**Additional information**

None.

**Checklist**

### DIFF
--- a/bitnami/redis/Chart.yaml
+++ b/bitnami/redis/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: redis
-version: 11.3.0
+version: 11.3.1
 appVersion: 6.0.9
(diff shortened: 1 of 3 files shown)


marcosbc (approved):

PR #4195, opened 2020-11-03
[bitnami/redis] Force redis-exporter to use TLS

jjpp:
**Description of the change**

The change adds env variable REDIS_ADDR with value "rediss://localhost:{{ .Values.redisPort }}"  to metrics container of all three stateful sets in case TLS is enabled

**Benefits**

dialer in the redis exporter uses TLS only if the URL has rediss as scheme. By default it will use redis as scheme even if TLS parameters are enabled.

**Possible drawbacks**

The change does not update documentation (not sure if needed either). 
Perhaps it would be more correct to fix the exporter (then again, fixing the exporter will not render this change to be incorrect). 

**Applicable issues**

  - fixes #

**Additional information**

**Checklist** 


miguelaeh (approved):
Hi,
LGTM!

PR #4196, opened 2020-11-03
[bitnami/redis] Makes password argument conditional for TLS as well

jjpp:
**Description of the change**

This change adds a missing condition so that scripts will not attempt to use -a without argument in case of enabled TLS and usePassword: false. 

**Benefits**

Scripts will not fail mysteriously. 
**Additional information**

**Checklist** 


alemorcuq (changes requested):
Thanks for this, @jjpp, LGTM.

Can you increase the version again?

alemorcuq (approved):
Great, thanks!

PR #4201, opened 2020-11-03
[bitnami/redis] Non-special node 0 in stateful sets

jjpp:
**Description of the change**

Scripts in configmap treat node -0 as special -- always being master. This may not be the case for example if the node-0 pod gets deleted because of the underlying host dies and the rest of the cluster still works. 
Code in this patch attempts to figure out if there are any living nodes in the stateful set via DNS. 
It also attempts to fill and update pointers to other living nodes in sentinel configuration file so that the newly spawned node can join in a cluster a bit more reliably. 

**Benefits**

The cluster recovers more reliably from the loss (and respawning) of a node or a few of them.

**Possible drawbacks**

I have so far only tested it using stateless cluster. 

**Applicable issues**

  - fixes #

**Additional information**

This pull request contains the code from https://github.com/bitnami/charts/pull/4196 and should probably be applied after that. 

**Checklist** 


bitnami/redis/templates/configmap-scripts.yaml
```diff
@@ -23,14 +23,16 @@ data:
         fi
     }
 
+    NODES_ALIVE="$(getent hosts "{{ template "redis.fullname" . }}-headless.{{ .Release.Namespace }}.svc.{{ .Values.clusterDomain }}")"
+
     export REDIS_REPLICATION_MODE="slave"
     if [[ $HOSTNAME =~ (.*)-([0-9]+)$ ]]; then
-      if [[ ${BASH_REMATCH[2]} == "0" ]]; then
```
juan131, line 28: I guess you don't need to compare the `$HOSTNAME` variable if you're not using its result.

bitnami/redis/templates/configmap-scripts.yaml
```diff
@@ -23,14 +23,16 @@ data:
...
+      if [[ -z "$NODES_ALIVE" ]]; then
```
juan131, line 30: You can also use the exit code of the `getent hosts` command to decide whether there are alive nodes or not.

bitnami/redis/templates/configmap-scripts.yaml
```diff
@@ -23,14 +23,16 @@ data:
...
         if [[ ! -f /data/redisboot.lock ]]; then
           export REDIS_REPLICATION_MODE="master"
         else
           if is_boolean_yes "$REDIS_TLS_ENABLED"; then
-            sentinel_info_command="redis-cli -a $REDIS_PASSWORD -h {{ template "redis.fullname" . }}-headless.{{ .Release.Namespace }}.svc.{{ .Values.clusterDomain }} -p {{ .Values.sentinel.port }} --tls --cert ${REDIS_TLS_CERT_FILE} --key ${REDIS_TLS_KEY_FILE} --cacert ${REDIS_TLS_CA_FILE} info"
+            sentinel_info_command="redis-cli {{- if .Values.usePassword }} -a $REDIS_PASSWORD {{- end }} -h {{ template "redis.fullname" . }}-headless.{{ .Release.Namespace }}.svc.{{ .Values.clusterDomain }} -p {{ .Values.sentinel.port }} --tls --cert ${REDIS_TLS_CERT_FILE} --key ${REDIS_TLS_KEY_FILE} --cacert ${REDIS_TLS_CA_FILE} info"
```
juan131, line 35: Nice catch

bitnami/redis/templates/configmap-scripts.yaml
```diff
@@ -23,14 +23,16 @@ data:
         fi
     }
 
+    NODES_ALIVE="$(getent hosts "{{ template "redis.fullname" . }}-headless.{{ .Release.Namespace }}.svc.{{ .Values.clusterDomain }}")"
```
juan131, line 26: Use the same approach you used below:
```
    HEADLESS_SERVICE="{{ template "redis.fullname" . }}-headless.{{ .Release.Namespace }}.svc.{{ .Values.clusterDomain }}"
    NODES_ALIVE="$(getent hosts "$HEADLESS_SERVICE")"
```

bitnami/redis/templates/configmap-scripts.yaml
```diff
@@ -208,19 +219,39 @@ data:
...
+
+    for node in $(seq 0 {{ .Values.cluster.slaveCount }}); do
+      NAME="{{ template "redis.fullname" . }}-node-$node"
+      IP="$(getent hosts "$NAME.$HEADLESS_SERVICE" | awk ' {print $1 }')"
+      if [[ "$NAME" != "$HOSTNAME" && -n "$IP" ]]; then
+        sentinel_conf_add "sentinel known-sentinel {{ .Values.sentinel.masterSet }} $IP {{ .Values.sentinel.port }} $(host_id $NAME)"
+        add_replica "$IP"
+      fi
+    done
+    add_replica "$(hostname -i)"
```
juan131, line 254: Could you explain this?

juan131 (commented):
Thanks for the PR! Please check my comments
@rafariossaa @miguelaeh could you please take a look too?

rafariossaa:
Nothing more to add to @juan131 review.

bitnami/redis/templates/configmap-scripts.yaml
jjpp: Absolutely. I guess it remained either because of keeping the diff small(er) or because the initial attempt was to do minimal change. But I will remove the unnecessary layer of if.

bitnami/redis/templates/configmap-scripts.yaml
jjpp: Hm. Good idea. Will apply.

bitnami/redis/templates/configmap-scripts.yaml
jjpp: Ack.

bitnami/redis/templates/configmap-scripts.yaml
jjpp: TL;DR: in case the master is (temporarily) missing, the vanilla sentinels/redis system is not able to survive the changing IP addresses of sentinels/replicas (pods). This code attempts to seed the sentinel with everything that we can deduce from DNS. 

Background: I did try to kill half of the nodes in cluster (deleted the pods either directly or via evicting them from kubernetes nodes). I guess the problem was that when I (among the other pods) killed the acting master and there was not enough nodes to elect a new one and so the new/respawned pods could not connect to master (or node 0) to get information about the remaining nodes of the cluster. Ultimately I arrived at partitoned configurations (say, 2 surviving nodes knew about each other and 2 new ones knew each other but no communication between partitions). So I figured that I should supply information about existing nodes as deducted from DNS.

What the code does:

Function add_replica declares a known replica unless the IP address belongs to the declared/acting master. Without that condition, the sentinels would say to the master that it should be replica of itself (instead of replica of no one) and that will effectively disable that node. 

For-cycle iterates over all the nodes and if a) the name resolves (node is alive) and b) it is not the node itself, it adds the sentinel as known sentinel. It conditionally also adds replica as known replica. 

It finally also adds the node itself as replica (unless it is the master). The last bit is important for the sentinel(s) to be also able to reach that replica in case the master was down when the replica was started.

jjpp: Oh, .. and for that to work, one has to use static ids. I will wrap it into relevant if.

jjpp: Actually, it did work with non-static ids as well but logged that the ids were wrong.

bitnami/redis/templates/configmap-scripts.yaml
jjpp: It appears that this did not work out. It appears that the query will succeed when starting the sentinel because the first pod itself is already registered in DNS. Replaced with 
[[ -z "$(getent ahosts "$HEADLESS_SERVICE" | grep -v "^$(hostname -i) ")" ]]
(list all matching records, remove those that have my own IP).

bitnami/redis/templates/configmap-scripts.yaml
juan131: Thanks so much for the detailed explanation!! I don't see any caveats.
Anyway, I'd like @rafariossaa and @miguelaeh to review it too

juan131 (approved):
Approving your changes! Please @miguelaeh @rafariossaa re-review the latest changes and share your opinion

bitnami/redis/templates/configmap-scripts.yaml
```diff
@@ -159,6 +159,9 @@ data:
 
         replace_in_file "/opt/bitnami/redis-sentinel/etc/sentinel.conf" "^#*\s*${key} .*" "${key} ${value}" false
     }
+    sentinel_conf_add() {
+        (echo; echo "$@") >> "/opt/bitnami/redis-sentinel/etc/sentinel.conf"
```
miguelaeh, line 163:
```suggestion
        echo "\n$@" >> "/opt/bitnami/redis-sentinel/etc/sentinel.conf"
```

Would that create a subshell to add the line to the file? Not sure if that could cause issues. I would this in this way.

miguelaeh (commented):
In general LGTM, but I think we should test this manually properly before merging @juan131 .

bitnami/redis/templates/configmap-scripts.yaml
jjpp: "\n$@" would output literal backslash and n. Adding -e to enable interpretation would also apply to possible escape sequences in parameters. Commited and pushed yet another alternative.

juan131 (approved):
Sorry for the delay @brew !! I've been testing your changes and everything seems to be working, therefore, I'm approving them

haljin:
Hi, Is there a reason that the `12.0.1` version is not available in the chart repository?

juan131:
Hi @haljin 

Sorry for the delay, there was an issue affecting our publishing pipeline yesterday. The issue is resolved and the chart is now published:

```
helm repo update && helm search repo bitnami/redis
...
NAME                 	CHART VERSION	APP VERSION	DESCRIPTION
bitnami/redis        	12.0.1       	6.0.9      	Open source, advanced key-value store. It is of...
```

PR #4211, opened 2020-11-04
[bitnami/thanos] allow overriding service selectors

janlud:
**Description of the change**

Allow overriding of selectors for all thanos components services. If the value labelSelectorsOverride stays empty, current behavior with default labels is used.

**Benefits**

We use ArgoCD to deploy helm chart. ArgoCD by default overrides label `app.kubernetes.io/instance` with its own value to identify resources belonging to the application.
The chart implements this label in service selector and because the deployment `app.kubernetes.io/instance` label will be always changed by ArgoCD, the service will be never able to find correct pods.

**Possible drawbacks**

I believe there are none - I made the overriding of selector optional and the default value is empty - current behavior will stay unless configured.

**Checklist** 


bitnami/thanos/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: v1
-version: 2.5.3
+version: 2.5.4
```
andresbono, line 2: Could you please bump the minor version as this is introducing new input parameters?
```suggestion
version: 2.6.0
```

bitnami/thanos/templates/query-frontend/service.yaml
```diff
@@ -29,6 +29,11 @@ spec:
...
       nodePort: null
       {{- end }}
-  selector: {{- include "thanos.matchLabels" . | nindent 4 }}
-    app.kubernetes.io/component: query-frontend
+  selector:
+    {{- if .Values.queryFrontend.service.labelSelectorsOverride }}
+    {{- include "thanos.tplValue" (dict "value" .Values.queryFrontend.service.labelSelectorsOverride "context" $) | nindent 4 }}
+    {{- else }}
+    {{- include "thanos.matchLabels" . | nindent 4 }}
+    app.kubernetes.io/component: queryFrontend
```
andresbono, line 37: This label should be kept as `query-frontend`, right?

```suggestion
    app.kubernetes.io/component: query-frontend
```

bitnami/thanos/templates/bucketweb/service.yaml
```diff
@@ -28,6 +28,11 @@ spec:
       {{- else if eq .Values.bucketweb.service.type "ClusterIP" }}
       nodePort: null
       {{- end }}
-  selector: {{- include "thanos.matchLabels" . | nindent 4 }}
+  selector:
+    {{- if .Values.bucketweb.service.labelSelectorsOverride }}
+    {{- include "thanos.tplValue" (dict "value" .Values.bucketweb.service.labelSelectorsOverride "context" $) | nindent 4 }}
+    {{- else }}
+    {{- include "thanos.matchLabels" . | nindent 4 }}
     app.kubernetes.io/component: bucketweb
```
andresbono, line 36: Shouldn't `app.kubernetes.io/component` label be kept even if the `labelSelectorsOverride` is provided?

andresbono:
Thanks for your contribution @janlud! Could you please check my comments?

bitnami/thanos/templates/query-frontend/service.yaml
janlud: Apologies, my mistake because of copy pasting. Fixed.

bitnami/thanos/Chart.yaml
janlud: Done

bitnami/thanos/templates/bucketweb/service.yaml
janlud: I added it also there. I am not really sure here - it's not a problem for my use case, but my initial thinking was that either someone wants to take the default chart selector or he or she might want to customize these entirely.

janlud: Should I change the `labelSelectorsOverride` to something without Override like just `labelSelectors` because it doesn't really override all selectors?

andresbono: > my initial thinking was that either someone wants to take the default chart selector or he or she might want to customize these entirely.

You are right @janlud.

andresbono: Oh, you changed it already! Sorry for the misunderstanding. After giving it a second thought, it makes more sense the first implementation that you proposed.

We are going to revert the change on your behalf and merge the PR.

janlud: Yeah, I wanted to speed it up in case you insist on the change. All good! :) Thanks

PR #4233, opened 2020-11-06
[bitnami/mongodb] Metrics container will not take TLS settings when cluster is TLS enabled

franklin432:
**Description of the change**

TLS was recently introduced into the cluster which works fine with the arbiter. However when the metrics is set to true, and the metrics.extraUri flag set to extraUri: "?tls=true&tlsCertificateKeyFile=/certs/mongodb.pem&tlsCAFile=/certs/mongodb-ca-cert",
the container fails to start up using these TLS settings which are the same TLS settings being applied to the MongoDB containers.
This change will slightly modify the mongodb-exporter URI flag as well as change the livenessProbe/readinessProbe to include the TLS settings.

**Benefits**

The ability to have a TLS enabled MongoDB helm chart which will also run the mongodb-exporter metrics alongside. 

**Additional information**

**Checklist** 


franklin432:
@dani8art can someone please check this PR

bitnami/mongodb/requirements.lock
```diff
@@ -1,7 +1,6 @@
 dependencies:
```
dani8art, line 1: Sorry, this file is automatically updated by our CI/CD, please don't update it

bitnami/mongodb/templates/standalone/dep-sts.yaml
```diff
@@ -299,9 +299,9 @@ spec:
           args:
             - |
               {{- if .Values.auth.enabled }}
-              /bin/mongodb_exporter --mongodb.uri mongodb://root:$(echo $MONGODB_ROOT_PASSWORD | sed -r "s/@/%40/g;s/:/%3A/g")@localhost:27017/admin{{ .Values.metrics.extraUri }} {{ .Values.metrics.extraFlags }}
+              /bin/mongodb_exporter --mongodb.uri "mongodb://root:$(echo $MONGODB_ROOT_PASSWORD | sed -r "s/@/%40/g;s/:/%3A/g")@localhost:27017/admin{{ .Values.metrics.extraUri }}" {{ .Values.metrics.extraFlags }}
```
dani8art, line 302: I think we could build an uri base on the tls parameters automatically when `tls.enabled=true` and then append the `.Values.metrics.extraUri`, so that user won't have to be aware to add extra uri parameters when enabling tls, what do you think?

dani8art (changes requested):
Hi @franklin432 thanks for this great fix, I would like to share some thoughts with you, could you take a look at my comments?

bitnami/mongodb/templates/standalone/dep-sts.yaml
franklin432: That could work...how exactly would you suggest we set that up when tls.enabled=true ?

bitnami/mongodb/requirements.lock
franklin432: ok I will change it back

franklin432 (commented):
added changes to the mongodb_exporter mongodb.uri for when TLS is enabled. This will apply the mongodb.uri to automatically contain the TLS parameters when the statefulset has tls.enabled set to true instead of having the user to manually input the settings in the value.yaml file.

bitnami/mongodb/templates/standalone/dep-sts.yaml
franklin432: I applied a change for when tls.enabled=true, let me know what you think.

franklin432 (commented):
changed

bitnami/mongodb/templates/replicaset/statefulset.yaml
```diff
@@ -359,11 +359,15 @@ spec:
             - -ec
           args:
             - |
+              {{- if and .Values.tls.enabled .Values.auth.enabled }}
```
dani8art, line 362: Could we extract some repeated logic to a helper, something like:

```
{{- define "mongodb.mongodb_exporter.uri" -}}
    {{- $uriTlsArgs := ternary "tls=true&tlsCertificateKeyFile=/certs/mongodb.pem&tlsCAFile=/certs/mongodb-ca-cert" "" .Values.tls.enabled -}}
    {{- $uriAuth := ternary "root:$(echo $MONGODB_ROOT_PASSWORD | sed -r \"s/@/%40/g;s/:/%3A/g\")@" "" .Values.auth.enabled -}}

    {{- printf "mongodb://%slocalhost:27017/admin?%s" $uriAuth $uriTlsArgs -}}
{{- end -}}
```

Then use it in all the yamls like 

```
- |
   /bin/mongodb_exporter --mongodb.uri "{{ include "mongodb.mongodb_exporter.uri" . }}" {{ .Values.metrics.extraFlags }}
```

bitnami/mongodb/templates/replicaset/statefulset.yaml
```diff
@@ -372,14 +376,29 @@ spec:
...
+                - -c
+              {{- if .Values.tls.enabled }}
+                - >-
+                  /bin/mongodb_exporter --mongodb.uri "mongodb://root:$(echo $MONGODB_ROOT_PASSWORD | sed -r "s/@/%40/g;s/:/%3A/g")@localhost:27017/admin?tls=true&tlsCertificateKeyFile=/certs/mongodb.pem&tlsCAFile=/certs/mongodb-ca-cert{{ .Values.metrics.extraUri }}" {{ .Values.metrics.extraFlags }}
+                  --test
+              {{- else }}
+                - >-
+                  /bin/mongodb_exporter --mongodb.uri "mongodb://root:$(echo $MONGODB_ROOT_PASSWORD | sed -r "s/@/%40/g;s/:/%3A/g")@localhost:27017/admin{{ .Values.metrics.extraUri }}" {{ .Values.metrics.extraFlags }}
+                  --test
+              {{- end }}
```
dani8art, line 401: Maybe I'm wrong but this command is testing if the `/bin/mongodb_exporter` configuration passed to the command is ok, isn't it?

Instead we should keep the previous liveness and readiness probe, right?

dani8art (changes requested):
Thank you so much for applying all these changes, I have one more suggestion, please take a look at my comment

bitnami/mongodb/templates/replicaset/statefulset.yaml
franklin432: Yes you are correct in saying that command is testing if the /bin/mongodb_exporter config passed to the command is ok or not. 
However the way the previous liveness/readiness probe is setup, its impossible for it to do a check if TLS is enabled because there is no way or format to add TLS settings or anything else to that command. 
```
livenessProbe:
            httpGet:
              path: /metrics
              port: metrics
```
Basically with the previous setup, to perform a probe, the kubelet sends an HTTP GET request to the server that is running in the container and listening on port 9216. If the handler for the server's /metrics path returns a success code, the kubelet considers the container to be alive and healthy. If the handler returns a failure code, the kubelet kills the container and restarts it. Any code greater than or equal to 200 and less than 400 indicates success. Any other code indicates failure.
There is no way (at least to my knowledge) to specify additional strings or TLS parameters for this probe's command

However I did see in a previous chart, they used the format I proposed in which the /bin/mongodb_exporter takes in the mongodb.uri string with --test appended to it to check the mongodb connection status, print the buildInfo() information and then exit. This to me is equivalent to the previous way but also allows us to specify a connection string which can take in our TLS settings for a TLS enabled cluster.

bitnami/mongodb/templates/replicaset/statefulset.yaml
franklin432: Perfect I love this setup. I was trying to do it earlier but was unsure how to setup the helper. 
I just tested it and it works fine. 
And like I mentioned earlier, if I use the previous livenessProbe and readinessProbe command with this new helper with tls enabled it fails. However once I give it the string **"/bin/mongodb_exporter --mongodb.uri "{{ include "mongodb.mongodb_exporter.uri" . }}" {{ .Values.metrics.extraFlags }}
                  --test"** for the probes it works perfect.

juan131 (approved):
Thanks so much!! Approving your changes

PR #4238, opened 2020-11-06
[bitnami/phabricator] SSH for VCS (git) + SVC annotations

### LINKED ISSUE #4204: [bitnami/phabricator] Add option to enable SSH for GIT
**Which chart**: phabricator 9.1.8

I want to use the ssh access to repositories that Phabricator offers.

Add a switch to the helm chart values, that sets the environment variable for the phabricator container `PHABRICATOR_ENABLE_GIT_SSH_REPOSITORY` to `yes` (or similar acceptable positive values). This could be integrated in the existing LoadBalancer definition.

**Additional context**
[Your docker image](https://github.com/bitnami/bitnami-docker-phabricator) already has the capability to run an SSH server for that purpose but there is no option in the helm chart to enable that.

### LINKED ISSUE #4205: [bitnami/phabricator] Add the possibility to add annotations (and labels) to the service
**Which chart**: phabricator 9.1.18

I want to deploy the helm chart to AKS and would like to only make the LoadBalancer available to the private network. Azure offers this feature by adding the annotation `service.beta.kubernetes.io/azure-load-balancer-internal: "true"` to the service ressource.

A switch in values.yaml similar to ingress.annotations for service.

Using ClusterIP in combination with an Ingress Controller would be possible but since I also plan to use the feature mentioned in #4204 I'd prefer to have it in the LoadBalancer in order to not proxy SSH with nginx.

**Additional context**
Adding support for labels seems like a good idea as well.
The AKS feature I'm referring to is documented here: https://docs.microsoft.com/en-us/azure/aks/internal-lb

VincentPost-TomTom:
**Description of the change**

* The environment variable `PHABRICATOR_ENABLE_GIT_SSH_REPOSITORY` has been set to `on` for the phabricator container.
* The template for the service object has been extended to read the values.yaml option service.annotations.

**Benefits**
* SSH can be used for checking out and pushing changes to GIT repositories.
* Annotations can be added if the service object is of type LoadBalancer to e.g. enable Azure Internal LoadBalancer.

**Possible drawbacks**

* The SSH functionality is not a toggle

**Applicable issues**

  - fixes #4204
  - fixes #4205 

**Additional information**

**Checklist** 


rafariossaa (commented):
I think it would be better to have something like ```ssh.enabled``` by default to `false`, and only enable it when it is `true`. This way the new chart version will behave the same way as currently.

Centzilius:
Hi, I just returned from vacation and will implement this change later today.

VincentPost-TomTom:
Hi, today I returned from vacation and have implemented the change now. I kept the variable name in line to the other container variables but I replaced yes/no with true/false.

VincentPost-TomTom:
Regarding what I think about it: I understand that not introducing changes to already deployed helm apps is desirable but I actually wonder why nobody wanted this feature earlier.

andresbono (approved):
LGTM

andresbono:
Thanks for addressing the suggestions, @VincentPost-TomTom!! The PR is now merged 🎉

VincentPost-TomTom:
Thanks for merging

Issue #4182, opened 2020-11-02
[Bitnami/MongoDB] External access with SSL enabled not working


mnmami:
**Which chart**:
Bitnami/MongoDB Chart version 9.1.2, App version 4.4.1

**Describe the bug**
The external access with SSL enabled not working after enabling SSL encryption.

**To Reproduce**
- Install the Chart with external access enabled, for example, using the following settings:
```yaml
architecture: replicaset
replicaCount: 4
externalAccess.enabled: true
externalAccess.service.type: LoadBalancer
externalAccess.service.port: 27017
externalAccess.autoDiscovery.enabled: true
rbac.create: true
```
This will generate a LoadBalancer service for every pod with external IP addresses, say IP1, IP2 ,IP3 ,IP4.
- Generate or simply copy `mongodb.pem` and `mongo-ca-cert` files from one of the generated pods and start mongo shell with the full list of external service IPs, like so:
```bash
mongo --host "IP1,IP2,IP3,IP4" --ssl --sslPEMKeyFile=ssl_keys/mongodb.pem --sslCAFile=ssl_keys/mongodb-ca-cert 
```

You get: `The server certificate does not match the host name. Hostname: {IP1} does not match SAN(s): {the list of CNs configured}` (DNSs list see under **Additional context** below). 

This since because the external IP addresses are only known later after the chart has been installed.

**Expected behavior**
Client (e.g., mongo shell) successfully starts by providing the PEM file and CA file as arguments.

**Version of Helm and Kubernetes**:

- Output of `helm version`:

```
version.BuildInfo{Version:"v3.3.0-rc.1", GitCommit:"5c2dfaad847df2ac8f289d278186d048f446c70c", GitTreeState:"dirty", GoVersion:"go1.14.4"}
```

- Output of `kubectl version`:

```
Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.5", GitCommit:"e6503f8d8f769ace2f338794c914a96fc335df0f", GitTreeState:"clean", BuildDate:"2020-06-26T03:47:41Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}
[...]

PR #4249, opened 2020-11-09
[bitnami/mongodb] External static LoadBalancer IPs added to the SSL certificate SANs

mnmami:
**Description of the change**
When TLS/SSL is enabled, the external access with LoadBalancers stops to work. This is because the automatically assigned LoadBalancer IPs are only known after or when the Chart is being Created. Hence, it is not possible to use those IPs to connect to the server (e.g., via Mongo shell) as they do not match the certificate pre-configured DNS names.

**Benefits**
The above issue can be solved by using _static_ LoadBalancer external IPs. For this, we add the list of external access `LoadBalancerIPs` to the list of SSL certificate SANs (Subject Alternative Names), which are used during the creation of the Statefulset (inside the `initContainer` command part).

**Possible drawbacks**
If the static external IPs change, and thus the respective services, the Statefulset needs to be recreated to regenerate new SSL certificates and therein add the new IP addresses.

**Applicable issues**
This PR should solve #4182 with the above context.

**Checklist** 
- [x] Variables are documented in the README.md (No variables but extra explanation reflecting the change).


bitnami/mongodb/templates/replicaset/statefulset.yaml
```diff
@@ -131,6 +131,11 @@ spec:
               DNS.3 = $my_hostname.$svc.$MY_POD_NAMESPACE.svc.cluster.local
               DNS.4 = localhost
               DNS.5 = 127.0.0.1
+              {{- if .Values.externalAccess.service.loadBalancerIPs }}
```
dani8art, line 134: I like this solution, but this would imply to update the chart with the requested ips to work and wait until the loadBalancerIPs are set to generate the certs, right?

dani8art (changes requested):
Hi @mnmami thank you so much for your contribution, please take a look at my comment.

bitnami/mongodb/templates/replicaset/statefulset.yaml
mnmami: I'm not sure I fully understood you, but the creation of the certificates is independent of the creation of the LoadBalancer services, as the LoadBalancers are created with pre-defined static IPs. One would just as usual wait for both the Statefulset (thus Pods) and the LoadBalancer services to be running with the latter having the appointed static IP addresses as external IP addresses.

dani8art: Yes, that is what I meant, this solution is assuming that we have static IPs, however, this could not happen always, right?

We need to provide a way where we can create this having dynamically assigned IPs. Normally, LoadBalancers can be created to supply dynamic IPs. 

We always try to fix these kinds of issues by doing a two steps installation: [see this](https://github.com/bitnami/charts/blob/master/bitnami/opencart/templates/NOTES.txt)

mnmami: You are right, not always one has/can have/want to have static IPS.  When that's the case, we go with the two-step deployment process you referenced above.

On the other hand, when the static IPs are available*, we should make use of them and simplify the process (to only one step). As in this PR, only a one if-statement and appending the IPs to the cert SAN would be all needed. Luckily the title of this PR is very specific to this case, so it can be implemented then another PR could be made (by someone, who is not me :)) to complement the other situation.

*In fact, in corporations where certificates need to be signed by the internal CA, there must be previously reserved static IP addresses anyway.

juan131: I agree with you @mnmami . We can't assume that static IPS will be used, but when they're used, we should them.

juan131 (approved):
Approving your changes! Thanks so much for the PR

PR #4254, opened 2020-11-09
[bitnami/etcd] Remove unused etcd.dataDir template function

pixiono:
**Description of the change**

I noticed that the etcd.dataDir template function has a strange behavior: The function returns a different path, depending on whether a PV should be used or not. If no PV should be used the path /opt/bitnami/etcd/data is returned, although the emptydir is mounted to the place /bitnami/etcd/data, this has become a problem for me in a private project with etcd Image with strict permissions under /opt.
  
Since I did not find a reason for this behaviour, and even the init-container has the path /bitnami/etcd hardcodet, I took the chance to provide the following PR. I also noticed that in the scripts-configmap.yaml the variable etcdDataDir is set to the value of this function but is never used. Therefore I assume that the function was used here once, but not anymore.

**Benefits**

- Compatibility to more restrictive images/environments 
- Reducing the code and simplifying readability

**Possible drawbacks**
**Applicable issues**
**Additional information**

**Checklist** 


andresbono (commented):
Hi, thanks for opening this PR! Can you also check the `bitnami/etcd/templates/NOTES.txt`file? It also uses the `etcd.dataDir` function.

pixiono:
Hi @andresbono, thanks for your feedback! I have adapted the NOTES.txt.

andresbono (approved):
Thanks for addressing the suggestions!!

PR #4258, opened 2020-11-10
[bitnami/common] Major version. Adapt Chart to apiVersion: v2

carrodher:
**Description of the change**

1. Changes related to the new apiVersion and Helm 2 deprecation:

- Use `type: library` in _Chart.yaml_
- Move dependency information from the _requirements.yaml_ to the _Chart.yaml_ (if any)
- After running `helm dependency update`, a new _Chart.lock_ file is generated containing the same structure used in the previous _requirements.lock_ (if any)
- Update _README.md_

2. How the new `type: library` works:

Without the `type: library`, you can install the common chart (although it doesn't deploy any object):
```console
$ helm install common bitnami/common
NAME: common
LAST DEPLOYED: Tue Nov 10 13:01:40 2020
NAMESPACE: default
STATUS: deployed
REVISION: 1
TEST SUITE: None

$ helm ls
helm ls
NAME                	NAMESPACE	REVISION	UPDATED                                  	STATUS  	CHART              	APP VERSION
common              	default  	1       	2020-11-10 13:01:40.513720878 +0000 UTC  	deployed	common-0.10.0      	0.10.0

$ kubectl get all --all-namespaces | grep common
# Nothing
```

With the addition of `type: library` to the _Chart.yaml_, the Helm Chart is not installed by `helm install` it can be only used as a sub-chart:
```console
$ helm install common .
Error: library charts are not installable
```

**Possible drawbacks**

Helm v2 is no longer supported

### DIFF
--- a/bitnami/common/Chart.yaml
+++ b/bitnami/common/Chart.yaml
@@ -1,11 +1,9 @@
 annotations:
   category: Infrastructure
-apiVersion: v1
+apiVersion: v2
 # Please make sure that version and appVersion are always the same.
-version: 0.10.0
-appVersion: 0.10.0
-description: A Library Helm Chart for grouping common logic between bitnami charts.
-  This chart is not deployable by itself.
+appVersion: 1.0.0
+description: A Library Helm Chart for grouping common logic between bitnami charts. This chart is not deployable by itself.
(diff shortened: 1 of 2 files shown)


andresbono (approved):
LGTM

PR #4307, opened 2020-11-11
[bitnami/thanos] rename querier component to query

### LINKED ISSUE #3368: Consider renaming thanos-querier to thanos-query
**Which chart**:
Thanos 2.0.0

When trying to use the thanos dashboards (like https://github.com/thanos-io/thanos/raw/9b578afb476374b68dd5ac5bfe50bcef70690760/examples/dashboards/overview.json), they all key on `thanos-query`. This means that the dashboards do not work by default (stateless Prometheus clusters are painful because of this)

Rename thanos-querier to thanos-query

mhulscher:
**Description of the change**

Rename the `querier` component to `query`.

**Benefits**

Using `query` is more inline with the official naming of the component.

It is called `query` in the official thanos documentation:
* https://thanos.io/tip/components/query.md/

It is assumed to be called `query` in the thanos example dashboards:
* https://github.com/thanos-io/thanos/tree/master/examples/dashboards

**Possible drawbacks**

Users need to replace `querier` with `query` in their `values.yaml`

**Applicable issues**

fixes #3368 

**Additional information**

N/A

**Checklist**


mhulscher:
Rebased on latest master

FraPazGal (commented):
Hi @mhulscher,

Thank you for the PR! As it is now, this changes would be a major. I think that releasing a new major for this kind of "minor" change should be avoided. A possible solution would be to allow using both `query.xxxx` and `querier.xxxxx` parameters, which would allow for backwards compatibility.

We could do that by adding a small function to `_helpers.tpl` that changes the name of the parameters like so:

```
{{/* Use query/querier for parameters */}}
{{- define "thanos.queryName" -}}
{{- $query := .Values.querier -}}
{{- if and (not .Values.querier) .Values.query -}}
{{- $query := .Values.query -}}
{{- end -}}
{{- end -}}
```

That will solve the current issue and allow us to keep this a minor.

mhulscher:
@FraPazGal , do you have an example of how you would use this function in a manifest to replace, for example, `.Values.query.enabled` ? I've read the helm docs up and down but I can't seem to figure it out.

Also, I understand that you want to keep supporting both `.Values.querier` and `.Values.query`, but are you OK with renaming the Deployment and its supporting objects to `thanos-query` regardless of whether `querier` or `query` is used in `values.yaml` ?

FraPazGal:
Hi @mhulscher, an approximate example would be something like this:

```
{{- $query := include "thanos.queryName" . }}
...
spec:
  replicas: {{ $query.replicaCount }}
...
```

Regarding the resources name change, I am in favour of changing them!

I also talked with my colleague @miguelaeh regarding your other issue and we could use this PR for both changes. It is true that it is a bit more complex for the sake of not bumping the major version, but a new one was released mere days ago. 

We also are planning to release another major in the coming months adding some standarizations across all of our catalog, so when the time comes we will refactor as needed.

mhulscher:
@FraPazGal I spent the better part of this day working on a new proposal.

I added the following snippet to `_helpers.tpl`:
```
{{/*
Return a YAML of either .Values.query or .Values.querier
If .Values.querier is used, we merge in the defaults from .Values.query, giving preference to .Values.querier
*/}}
{{- define "thanos.query.values" -}}
{{- if .Values.querier -}}
    {{- if .Values.query -}}
        {{- mergeOverwrite .Values.query .Values.querier | toYaml -}}
    {{- else -}}
        {{- .Values.querier | toYaml -}}
    {{- end -}}
{{- else -}}
    {{- .Values.query | toYaml -}}
{{- end -}}
{{- end -}}
```

**Several things to note:**

This PR changes the key under which the default values of query lives from `.Values.querier` to `Values.query`. To make sure that users that use `.Values.querier` can continue relying on the default values being used in case they have not explicitly set a key, we perform a merge of both `.Values.querier` and `.Values.query`. Supposedly this must be a `mergeOverwrite` and not a `merge` because of this: https://github.com/Masterminds/sprig/issues/134 . The current function seems to work. I can now arbitrarily change the name of the key `querier` to `query` and back again.

I could not find a different way of returning a dict from a function other than serializing it to and from YAML. You'll see that each template starts with the following line: `{{- $query := (include "thanos.query.values" . | fromYaml) -}}`

Please let me know what you think.

FraPazGal:
Hi @mhulscher,

It is looking really good! I have tested it locally and everything seems to be working. We will need to change a few things for our internal tools to tests the new chart properly. I'll get back to you when we have the all green internally!

FraPazGal (approved):
Everything is working on our end!

Thank you again for the PR!

PR #4312, opened 2020-11-11
[bitnami/harbor] Support Redis Sentinel

hobti01:
**Description of the change**

Update redis implementation to securely use Sentinel where supported by components.

* feat: Support embedded redis with sentinel.
* feat: Support redis idle timeout for components that support it.
* feat: Support chart cache for Harbor core.
* fix: Relocate redis passwords from ConfigMaps to Secrets.
* fix: Use redis username `default` to match the implicit user https://redis.io/commands/auth
* fix: Use `_REDIS_URL_CORE` instead of `_REDIS_URL` for Harbor core component
* chore: Update embedded redis and postgresql charts.

**Benefits**

Redis sentinel can be used with redis password authentication.

**Possible drawbacks**

Two outstanding issues which prevent use of redis+sentinel in all connections:
* https://github.com/helm/chartmuseum/issues/179
* https://github.com/goharbor/harbor/issues/13485
* Redis password authentication with Sentinel authentication does not (seem to) work with Harbor clients

**Applicable issues**

 
**Additional information**

Apologies for the refactoring in `_helper.tpl` but I didn't see a more concise approach.

**Checklist** 

### DIFF
--- a/bitnami/harbor/Chart.yaml
+++ b/bitnami/harbor/Chart.yaml
@@ -34,4 +34,4 @@ sources:
   - https://github.com/bitnami/bitnami-docker-harbor-registry
   - https://github.com/bitnami/bitnami-docker-harbor-registryctl
   - https://goharbor.io/
-version: 9.0.0
+version: 9.0.1
--- a/bitnami/harbor/README.md
+++ b/bitnami/harbor/README.md
@@ -89,6 +89,7 @@ The following tables list the configurable parameters of the Harbor chart and th
 | `proxy.components`                    | The component list that the proxy settings apply to                                                                                                       | core, jobservice, clair                                 |
 | `externalURL`                         | The external URL for Harbor core service                                                                                                                  | `https://core.harbor.domain`                            |
 | `extraDeploy`                         | Array of extra objects to deploy with the release (evaluated as a template).                                                                              | `nil`                                                   |
+| `redisIdleTimeoutSeconds`             | Idle timeout on redis connections that support it                                                                                                         | `30`                                                    |
 
 ### Traffic Exposure Parameters
 
@@ -281,6 +282,7 @@ The following tables list the configurable parameters of the Harbor chart and th
 | `coreImage.pullPolicy`      | Harbor Core image pull policy                                                                                                                                                                                                                                                            | `IfNotPresent`                                          |
 | `coreImage.pullSecrets`     | Specify docker-registry secret names as an array                                                                                                                                                                                                                                         | `[]` (does not add image pull secrets to deployed pods) |
 | `coreImage.debug`           | Specify if debug logs should be enabled                                                                                                                                                                                                                                                  | `false`                                                 |
+| `core.chartCacheDriver`     | The driver type to cache charts: `memory`, `redis`, or `redis_sentinel`. | Detected from `redis` or `externalRedis` configuration. |
 | `core.secretKey`            | The key used for encryption. Must be a string of 16 chars                                                                                                                                                                                                                                | Random 16 character long alphanumeric string            |
 | `core.uaaSecretName`        | If using external UAA auth which has a self signed cert, you can provide a pre-created secret containing it under the key `ca.crt`.                                                                                                                                                      | ``                                                      |
 | `core.tls.existingSecret`                 | Name of a secret with the certificates for internal TLS access. Requires internalTLS.enabled to be set to true. If this values is not set it will be automatically generated                                                     | `nil`                                                   |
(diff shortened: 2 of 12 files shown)


rafariossaa:
Hi, 
  Thanks you very much for adding all this features and fixed.
  I am adding @javsalgar  and @dani8art  to review it too.

hobti01:
Thank you for acknowledging this PR.  I realize this has several changes and may take some time to review.
We are not successful in getting redis+sentinel to work with Harbor and passwords, but we suspect this is because the redis chart does not enable AUTH for sentinel and redis separately: https://github.com/bitnami/charts/blob/master/bitnami/redis/templates/configmap-scripts.yaml#L102

miguelaeh (commented):
Hi,
>redis chart does not enable AUTH for sentinel and redis separately?https://github.com/bitnami/charts/blob/master/bitnami/redis/templates/configmap-scripts.yaml#L102

I just checked and in the Redis Helm Chart we are using the same password for Redis and Sentinel, but the `usePassword` parameter seems to be set separately for Redis and Redis Sentinel.

Apart from that, I can see some code to work around the issues when Sentinel is not supported in the Harbor components. Is there any reason to use Sentinel instead of, for example, a Redis Cluster with a single entrypoint?

hobti01:
We are not interested in the details of how Redis is HA, but only want an HA installation for Harbor.  Harbor 1.x did not support any redis clustering approach.  Harbor 2.x now supports Sentinel, but does not work as expected with authentication.  Part of this is due to the "default" username, and part seems to be due to the sentinel vs node authentication.

We'd be ok to use a cluster with a single entrypoint, but my understanding is that the bitnami/redis chart cannot support multiple databases with "cluster mode" https://github.com/bitnami/charts/blob/master/bitnami/redis/README.md#choose-between-redis-helm-chart-and-redis-cluster-helm-chart
> Supports only one database.

So the only HA option would be with sentinel?  Please tell me I'm wrong and there is already a working HA configuration :D

miguelaeh:
Sadly you are not wrong 😂. The Redis cluster solution supports only one database, so if that does not fit the requirements we must continue with the sentinel approach.

hobti01:
The more we are testing with sentinel, the more issues we are finding.  Even after setting `staticID: true` https://github.com/bitnami/charts/blob/master/bitnami/redis/values.yaml#L90
we continue to have issues with redis, causing failures during user login, caching, and general instability.

Despite the changes we do not think the sentinel implementation is ready for use.  Please let me know how you would like to progress this PR: I can revert the sentinel changes and leave the fixes, or leave all changes to prepare for future support.

miguelaeh:
Hi @hobti01 ,
I am sorry to hear that, but thank you very much for the work you are doing with it!
You could create a separate PR for the fixes in order to merge those changes, and left this PR open for the sentinel changes. Or, if you prefer, you can use this PR for the fixes and open a new one with the sentinel changes. You can do it as you prefer.

hobti01:
OK, I've made a few updates to the documentation so that it is clear that Sentinel is not used in all cases.  As previously documented it is disabled by default, so I suggest keeping this MR to add the configuration for sentinel and the other fixes.

Upstream it looks like the harbor team is deprecating chartmuseum so it's lack of sentinel support should not be a long-term issue.

hobti01:
I have rebased on the new version, but am unable to test with Helm v3.  I hope you will still consider merging these changes.

miguelaeh:
Hi @hobti01 ,
Thank you very much, could you just share in this thread the link to the other PR, please? Just to be able to track them.
Apart from that, I will create an internal task to continue checking this. In the meantime, I will leave this PR open.

hobti01:
This is the only PR which defaults to not using Sentinel.  Even if Redis+Sentinel is configured, chartmuseum will not use it because it does not support it.

miguelaeh:
I thought we were talking about creating a different PR for the fixes not related to sentinel and left this one only for the sentinel changes. I think that will allow us to merge the fixes not related to sentinel.

hobti01:
This MR incorporates
* security fixes such as putting passwords into Secrets
* supported redis+sentinel configuration of Harbor 2.1.1

The default configuration of redis has not been changed and remains a single instance without TLS or password.

miguelaeh:
I created an internal task to check this PR in the future.
We will update this thread once we have news.
Thank you very much for the work you have done on that!!
Regards.

panteparak:
This PR only covers support for Redis Sentinel for build-in deployment of redis right? External Redis configuration should support Redis Sentinel?

miguelaeh:
Hi @panteparak ,
Yes, using an external Redis I think you should be able to use sentinel. You would need to configure the deployment pointing to the external Redis.
In any case, let us know if you find any issues.

carrodher:
Unfortunately, due to other priorities and the limited bandwidth, we have the work on this PR was not continued and there are no plans to work on it in the mid-term. I'm replacing the `on-hold` label with the `stale` one so the PR will be automatically closed if it is not continued during the following days.

PR #4353, opened 2020-11-13
[bitnami/phpBB] Major version. Adapt Chart to apiVersion: v2 and Update MariaDB Dependency

FraPazGal:
**Description of the change**

- Update _README.md_
- Updated MariaDB dependency to the new major.
- Added some missing parameters present on `charts/template`
- Added support for existingSecret for external databases

**Possible drawbacks**

- Helm v2 is no longer supported
- Breaking changes due to MariaDB major bump


bitnami/phpbb/templates/svc.yaml
```diff
@@ -1,8 +1,9 @@
 apiVersion: v1
 kind: Service
 metadata:
-  name: {{ template "phpbb.fullname" . }}
+  name: {{ template "common.names.fullname" . }}
   labels: {{- include "common.labels.standard" . | nindent 4 }}
+    app.kubernetes.io/component: phpbb
```
juan131, line 6: Update the "selector" at the end of this service including the new label:
```diff
  selector: {{- include "common.labels.matchLabels" . | nindent 4 }}
+   app.kubernetes.io/component: phpbb
```

bitnami/phpbb/templates/tls-secrets.yaml
```diff
@@ -5,6 +5,7 @@ kind: Secret
 metadata:
   name: {{ .name }}
   labels: {{- include "common.labels.standard" $ | nindent 4 }}
+    app.kubernetes.io/component: phpbb
     {{- if $.Values.commonLabels }}
     {{- include "common.tplvalues.render" ( dict "value" $.Values.commonLabels "context" $ ) | nindent 4 }}
     {{- end }}
```
juan131, line 11: It'd be great if we could auto-generate a certificate when not provided (see keycloak for instance)

juan131 (changes requested):
Thanks so much for the PR!! Just a couple of comments.

As a general comment:

- Am I missing the Chart.lock file?

PR #4354, opened 2020-11-13
[bitnami/OrangeHRM] Major version. Adapt Chart to apiVersion: v2 and Update MariaDB Dependency

FraPazGal:
**Description of the change**

- Update _README.md_
- Updated MariaDB dependency to the new major.
- Added support for existingSecret for external databases

**Possible drawbacks**

- Helm v2 is no longer supported
- Breaking changes due to MariaDB major bump


bitnami/orangehrm/README.md
```diff
@@ -191,6 +195,79 @@ Find more information about how to deal with common errors related to Bitnami’
 
 ## Upgrading
 
+### To 8.0.0
+
+In this major there were two main changes introduced:
+
+1. Adaptation to Helm v2 EOL
+2. Updated MariaDB dependency version
```
marcosbc, line 203: IMHO we should be using bullet points here

bitnami/orangehrm/README.md
```diff
@@ -191,6 +195,79 @@ Find more information about how to deal with common errors related to Bitnami’
...
+
+Please read the update notes carefully.
+
+**1. Adaptation to Helm v2 EOL**
```
marcosbc, line 207: The "1. ..." is at the same level as the entries inside, maybe we could use h4/h5 instead so it is easier to understand?

bitnami/orangehrm/requirements.yaml
```diff
@@ -1,5 +0,0 @@
-dependencies:
```
marcosbc, line 1: It looks like we are removing this file, is it a mistake?

bitnami/orangehrm/requirements.lock
```diff
@@ -1,6 +0,0 @@
-dependencies:
```
marcosbc, line 1: It looks like we are removing this file, is it a mistake?

marcosbc (changes requested):
We should be adding the common-subchart dependency and making use of it for the new standarizations. It is probably easier if you base on bitnami/template or any other PHP app (i.e. bitnami/drupal) to get such standarizations.

I would recommend the latter, by replacing the entire chart's contents and then replace any drupal (or app name) occurrence with orangehrm, update descriptions and all that. It is the easier process and will take you less than 1 hour (compared to the 5+ hours if you apply all the standarizations manually).

bitnami/orangehrm/requirements.yaml
carrodher: It is desired, in Helm 3 the dependencies are set in the _Chart.yaml_, you can see mariadb as a dependency in that file right now

bitnami/orangehrm/requirements.lock
carrodher: A _Chart.lock_ will be generated once `helm dep update` is executed

carrodher:
> We should be adding the common-subchart dependency and making use of it for the new standarizations. It is probably easier if you base on bitnami/template or any other PHP app (i.e. bitnami/drupal) to get such standarizations.
> 
> I would recommend the latter, by replacing the entire chart's contents and then replace any drupal (or app name) occurrence with orangehrm, update descriptions and all that. It is the easier process and will take you less than 1 hour (compared to the 5+ hours if you apply all the standarizations manually).

The aim of the task is to use the latest MariaDB version that is 2 major versions behind the latest version (currently 7.X.X, while the latest is 9.X.X), which is something that should be fixed ASAP due to different reason. I agree the optimal solution is to take advantage of the major version and include the common standardization but that is going to depend on how long it's taking because it is not only this Helm Chart the one with an old MariaDB version, we need to apply the same changes in around 15 charts.

If we have time we will implement the common changes but, at least, let's create the PRs with the mandatory changes about bumping the MariaDB version and depending on the available capacity we will add other good-to-have changes or just keep what is necessary and postpone the other changes for another time with the adequate estimation and capacity

bitnami/orangehrm/README.md
FraPazGal: I tried it at first but h4 is actually smaller than using bolds and h5 is smaller than the standard font. We also shouldn't use h3 as that would make too big. As a compromise, I used numbers to highlight these lines, though I actually prefer using bullet points.

marcosbc: It just looks weird having multiple bolds with the same size, but up to you

bitnami/orangehrm/requirements.yaml
marcosbc: Ok!

marcosbc (approved):
I see now. Thanks for the context!

PR #4361, opened 2020-11-13
[bitnami/opencart] fixed sessionAffinity

dmytrokosiachenko:
**Description** 
Wrong position of the sessionAffinity variable in the values.yaml.

**Benefits**
Bug will be fixed


alemorcuq (approved):
Good catch, @dmytrokosiachenko. Thanks!

PR #4370, opened 2020-11-16
[bitname/memcached] Only add secret and env vars when auth is enabled

unguiculus:
With this change, the secret for the password is only created if
the password is set. Environment variables for user and passwords
are only added if values are set.

**Checklist**


marcosbc (approved):

PR #4373, opened 2020-11-16
[bitnami/wordpress] Enable using the bitnami/wordpress container custom user post-init.d scripts

pestebogdan:
**Description of the change**

Added `customPostInitScripts` **map variable** which, if not empty, populates a ConfigMap which can then be mounted to  **/docker-entrypoint-init.d/**  (using existing extraVolumes and extraVolumeMounts functionality). Advantages of this are:
  - scripts declared here are run only once, the very first time the pod is created
  - are run as the very last step of **entrypoint.sh** so all the pre-requisites are already initialized (packages, config files created, database initialized)
  - variable value structure that ends up in the ConfigMap allows for making use of a scratch space for better organising your code; since the logic of the code is to only execute files with suffixes .sh, .sql or .php , all other file extensions (or no extension) can be used as helper files for the actual post-init script

ref: https://github.com/bitnami/bitnami-docker-wordpress/tree/master/5/debian-10/rootfs/post-init.d

As an example, using the above functionality I was able to configure Multisite out-of-the-box for a fresh wordpress deployment, but given the abstract nature of the change, any run-once tasks that need be executed right before the start of the process are a good fit.

**Possible drawbacks**

In case the team developing the bitnami/wordpress container image decide to remove the post-init.d functionality


bitnami/wordpress/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: wordpress
-version: 9.9.3
+version: 9.9.4
```
miguelaeh, line 3: This should be a minor bump

bitnami/wordpress/templates/configmap.yaml
```diff
@@ -0,0 +1,8 @@
+{{- if .Values.customPostInitScripts }}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ template "wordpress.fullname" . }}-custominit-configmap
```
miguelaeh, line 5: Some times the length of the name causes issues, I think this name will be too long due to the length fixed part `-custominit-configmap`. Could you check that and in that case, use a short name, please?

bitnami/wordpress/templates/configmap.yaml
```diff
@@ -0,0 +1,8 @@
...
+data:
+{{ toYaml .Values.customPostInitScripts | indent 2}}
```
miguelaeh, line 7: Could you use here the `wordpress.tplValue` function instead of `toYaml` and `nindent` instead of `indent` please?

bitnami/wordpress/values.yaml
```diff
@@ -570,3 +570,23 @@ sidecars: {}
 ##    command: ['sh', '-c', 'copy themes and plugins from git and push to /bitnami/wordpress/wp-content. Should work with extraVolumeMounts and extraVolumes']
 ##
 initContainers: {}
+
```
miguelaeh, line 573: Add this same content to the `values-production.yaml` file

miguelaeh (changes requested):
Hi,

bitnami/wordpress/Chart.yaml
pestebogdan: Ok, so I assume that would mean 9.10.0 right ? changed

bitnami/wordpress/templates/configmap.yaml
pestebogdan: Fair point; changed name to {{ template "wordpress.fullname" . }}-postinit

bitnami/wordpress/templates/configmap.yaml
pestebogdan: Done

bitnami/wordpress/values.yaml
pestebogdan: Done

miguelaeh (approved):
LGTM!

Issue #3675, opened 2020-09-15
Rabbitmq auth admin credentials does not work.


hansehe:
**Which chart**:
Rabbitmq helm chart version 7.6.6.

**Describe the bug**
Setting auth.username and auth.password results in 'login failed' when trying to login to the admin dashboard.
The auto generated credentials works the first time it is deployed. However, you are required to set the same auth credentials to update the helm chart, which leads to 'login failed' again.

In summary, setting auth.username and auth.password does not work.

**To Reproduce**

helm install rabbitmq rabbitmq/rabbitmq -f values.yaml

values.yaml:
```yaml
persistence:
  enabled: false
  size: 8Gi
replicaCount: 1
auth:
  username: user
  password: <password>
  erlangCookie: <cookie>
extraConfiguration: |-
    mqtt.allow_anonymous  = false
    # 24 hours by default
    mqtt.subscription_ttl = 86400000
    mqtt.prefetch         = 10
metrics:
  enabled: true
ingress:
  enabled: true
  hostname: rabbitmq.localhost
  tls: true
  annotations:
    kubernetes.io/ingress.class: nginx
    cert-manager.io/cluster-issuer: letsencrypt
    nginx.ingress.kubernetes.io/force-ssl-redirect: "true"
extraSecrets:
  load-definition:
    load_definition.json: |
      {
          "users": [
              {
                  "name": "amqp",
                  "password": "<password>",
                  "tags": "amqp-client"
              },
              {
                  "name": "mqtt",
                  "password": "<password>",
                  "tags": "mqtt-client"
              }
          ],
          "vhosts": [
              {
                  "name": "amqp"
              },
              {
                  "name": "mqtt"
              }
          ],
          "permissions": [
              {
                  "user": "amqp",
                  "vhost": "amqp",
                  "configure": ".*",
                  "write": ".*",
                  "read": ".*"
              },
              {
                  "user": "mqtt",
                  "vhost": "mqtt",
                  "configure": ".*",
                  "write": ".*",
                  "read": ".*"
              }
          ]
      }
loadDefinition:
  enabled: true
  existingSecret: load-definition
extraConfiguration: |
  load_definitions = /app/load_definition.json
extraPlugins: "rabbitmq_auth_backend_ldap,rabbitmq_mqtt,rabbitmq_web_mqtt"
service:
  extraPorts: 
    - name: mqtt-tcp
      port: 1883
      targetPort: 1883
    - name: mqtt-web
      port: 15675
[...]

PR #4374, opened 2020-11-16
[bitnami/rabbitmq] Fixing support for loading definitions

msvticket:
**Description of the change**

The change allows for RABBITMQ_PASSWORD to not be set and when RABBITMQ_LOAD_DEFINITIONS is set instead the rabbitmq_initialize function won't try to change the password.

**Benefits**

It will again be possible to load definitions.

**Possible drawbacks**

Any user password needs to be explicitly set in the definition file. If not the password will stay as it is in rabbitmq.conf, which by default is CHANGEME when using the helm chart. I therefore expand the [documentation for loading definitions](https://github.com/bitnami/charts/tree/master/bitnami/rabbitmq#load-definitions) with the requirement to include user credentials in the definitions.  

**Applicable issues**

#3675

**Checklist** 

### DIFF
--- a/bitnami/rabbitmq/Chart.yaml
+++ b/bitnami/rabbitmq/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: rabbitmq
-version: 7.7.1
+version: 7.7.2
 appVersion: 3.8.9
 description: Open source message broker software that implements the Advanced Message Queuing Protocol (AMQP)
 keywords:
--- a/bitnami/rabbitmq/README.md
+++ b/bitnami/rabbitmq/README.md
@@ -63,55 +63,55 @@ The following table lists the configurable parameters of the RabbitMQ chart and 
 
 ### RabbitMQ parameters
 
-| Parameter                                 | Description                                                                                                          | Default                                                      |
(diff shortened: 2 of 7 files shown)


while-loop:
This also fixes the issue for me when the pod would exit because of failed to change password followed by a restart w/ an unchanged password.

andresbono:
Thanks so much, @msvticket!! 🎉 

https://github.com/bitnami/bitnami-docker-rabbitmq/pull/139#issuecomment-728128299
> I have a branch prepared to make a PR from. But with the changes for helm 3 I get conflicts. And I can't start using helm 3 now anyway... I have now created a PR with conflicts (bitnami/charts#4374). Maybe someone else wants to fork my repo and finish the work.

We can take care of it. We will create a new PR based on this one. Thanks for your contribution!

c0rvax:
@msvticket we seem to be having an issues with this, which basically keeps password set to `CHANGEME`. The Kube's secret has a randomly generated password, which however does not work to login into the rabbit instance. So the password is generated, but not changed from `CHANGEME`.

We used these commands to install rabbit

```$ helm repo add bitnami https://charts.bitnami.com/bitnami
$ helm install my-release bitnami/rabbitmq```

juan131:
@andresbono @alemorcuq did you create the new PR based on this one?

PR #4123, opened 2020-10-27
[bitnami/postgresql] Fix secret logic for passwords

miguelaeh:
**Description of the change**
The logic to create the Postgres password is on the _helpers.tpl so it is not necessary at the secret

**Benefits**

**Possible drawbacks**

**Applicable issues**

  - fixes #4120

**Additional information**

**Checklist** 
- [x] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [x] Variables are documented in the README.md
- [x] Title of the PR starts with chart name (e.g. `[bitnami/chart]`)
- [x] If the chart contains a `values-production.yaml` apart from `values.yaml`, ensure that you implement the changes in both files

:warning: Keep in mind that if you want to make changes to the kubeapps chart, please implement them in the [kubeapps repository](https://github.com/kubeapps/kubeapps/tree/master/chart/kubeapps). This is only a synchronized mirror.

PR #4380, opened 2020-11-17
[bitnami/postgresql] `postgresql-postgres-password` doesn't work with `global.postgresqlUsername` and without `postgresqlPostgresPassword` #3130

### LINKED ISSUE #3130: `postgresql-postgres-password` doesn't work with `global.postgresqlUsername` and without `postgresqlPostgresPassword`
**Which chart**: postgresql 9.1.1

**Describe the bug**
Including `postgresql-postgres-password` in the `existingSecret` only works when `postgresqlUsername` and `postgresqlPostgresPassword` are set. It doesn't work when `global.postgresqlUsername` is set instead or when the password is only included as `postgresql-postgres-password` in the secret without a value for `postgresqlPostgresPassword` in the values. There's an explicit check for both of those in `charts/bitnami/postgresql/templates/statefulset.yaml`:
```
{{- if and .Values.postgresqlPostgresPassword (not (eq .Values.postgresqlUsername "postgres")) }}
```
Not sure whether actually a bug or expected behavior. If it's expected behavior it might be better documented.

**To Reproduce**
1. set `global.postgresqlUsername` and include `postgresql-postgres-password` in `existingSecret`
2. don't set `postgresqlUsername` and/or `postgresqlPostgresPassword`
3. install chart
4. see that `POSTGRES_POSTGRES_PASSWORD` is missing in the pods `env` definition

**Expected behavior**
- that it doesn't matter whether `global.postgresqlUsername` or `postgresqlUsername` is used
- that `postgresqlPostgresPassword` doesn't need to be set in values but it's value only needs to be included as `postgresql-postgres-password` in the `existingSecret`

- Output of `helm version`:

```
version.BuildInfo{Version:"v3.0.3", GitCommit:"ac925eb7279f4a6955df663a0128044a8a6b7593", GitTreeState:"clean", GoVersion:"go1.13.6"}
```

```
Client Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.3", GitCommit:"5e53fd6bc17c0dec8434817e69b04a25d8ae0ff0", GitTreeState:"clean", BuildDate:"2019-06-06T01:44:30Z", GoVersion:"go1.12.5", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.0+k3s.1", GitCommit:"0f644650f5d8e9f091629f860b342f221c46f6d7", GitTreeState:"clean", BuildDate:"2020-01-06T23:22:38Z", GoVersion:"go1.13.5", Compiler:"gc", Platform:"linux/amd64"}
```

serg-bs:
**Description of the change**
install helm chart with val.yaml
```
global:
  postgresql:
    postgresqlUsername: gg
    postgresqlPassword: gg
    postgresqlPostgresPassword: postgres-password
```
`helm install postgre -f val.yaml . `

You cannot login with user postgres, because in StatefulSet not set variable
- name: POSTGRES_POSTGRES_PASSWORD 

Before that bug was  #4123 

**Possible drawbacks**
If you install with empty parameter and you user is postgres you password will take from not from
postgresql-postgres-password

  - fixes #3130


bitnami/postgresql/templates/NOTES.txt
```diff
@@ -7,7 +7,7 @@ PostgreSQL can be accessed via port {{ template "postgresql.port" . }} on the fo
     {{ template "postgresql.fullname" . }}-read.{{ .Release.Namespace }}.svc.cluster.local - Read only connection
 {{- end }}
 
-{{- if and (not (eq .Values.postgresqlUsername "postgres")) (or .Values.postgresqlPostgresPassword (include "postgresql.useExistingSecret" .)) }}
+{{- if and (not (eq (include "postgresql.username" .) "postgres")) (or (include "postgresql.postgres.password" .) (include "postgresql.useExistingSecret" .)) }}
```
andresbono, line 10: Given the definition of the `postgresql.postgres.password` helper function:

https://github.com/bitnami/charts/blob/9084881e8330e01c49492243c29e144e77d4c530/bitnami/postgresql/templates/_helpers.tpl#L80-L91

Wouldn't it be `(include "postgresql.postgres.password" .)` always `true`?

andresbono (commented):
Thanks for submitting this PR! Could you please check my comment?

bitnami/postgresql/templates/NOTES.txt
serg-bs: You are right.   `{{- if not (eq (include "postgresql.username" .) "postgres" }} `  will be more appropriate.

andresbono:
Hi @serg-bs, after checking the issue again, everything seems to work fine. At least with bitnami/postgresql `10.0.2`, we can't reproduce the issue that you are describing. See the following snippets. The environment variables are present and the connections with both users are successful:

`custom-values.yaml`
```
postgresqlUsername: gg
postgresqlPassword: gg_password
postgresqlPostgresPassword: postgres_password
```
```console
$ kubectl exec -it p-postgresql-0 -- env | grep -E "POSTGRES_USER|POSTGRES_PASSWORD"
POSTGRES_USER=gg
POSTGRES_PASSWORD=gg_password
POSTGRES_POSTGRES_PASSWORD=postgres_password
```

```console
$ kubectl exec -it p-postgresql-0 -- bash -c 'PGPASSWORD=gg_password psql -U gg postgres -c "select 1"'
 ?column?
----------
        1
(1 row)

$ kubectl exec -it p-postgresql-0 -- bash -c 'PGPASSWORD=postgres_password psql -U postgres postgres -c "select 1"'
 ?column?
----------
        1
(1 row)
```

Is it different for you?

serg-bs:
Of course different, usual parameter works fine. It's all about global parameters.
Main problem what  - check for POSTGRES_POSTGRES_PASSWORD is not use global parameters. So POSTGRES_POSTGRES_PASSWORD disapear if .Values.postgresqlUsername not set.

bitnami/postgresql/templates/NOTES.txt
miguelaeh: I am not sure about that, it will be missing the condition to check that the password has been provided instead of auto-generated, that is the intention of the `(or .Values.postgresqlPostgresPassword (include "postgresql.useExistingSecret" .))` section

miguelaeh:
I would like to understand better the issue, could you share with us why do you need to use global parameters?

serg-bs:
I use global parameter cause I want use postgres from subchart. I do it because documentation tell me about possibility. For example, I write common chart for take backup.

bitnami/postgresql/templates/NOTES.txt
serg-bs: I see. you tell me what is happen if we use values.yaml
`global:
  postgresql:
    postgresqlUsername: gg
    postgresqlPassword: gg
`
autogenerated secret - we have no problem, postgresPostgresPassword always exists. was fix #4123. 

Do you see any other error cases?

serg-bs:
By the way. If mysql helm chart will be we looks like postgres with global parameters will be also convinient. Or you think it's not your mainstream?

miguelaeh (commented):
Hi @serg-bs ,
I still have some concerns the changes. It seems that the final aim of this PR is to create global values for the passwords. IMHO I would add those new global values and the necessary conditions to make them take precedence over the current ones.

miguelaeh (approved):
LGTM!

bitnami/postgresql/templates/NOTES.txt
miguelaeh: it seems to be an old code that can be removed

PR #4389, opened 2020-11-17
[bitnami/mariadb][bitnami/common] Fix bad password validation with existingSecret

sereinity:
Fix bad password validation when using an `existingSecret` as it was looking on a bad place.

Strangely, in my situation, the error only occurs on `upgrade`, not on `install`.

The change look for the name of the existing secret in `auth` as describe in the mariadb chart's README and `values.yaml`.

### DIFF
--- a/bitnami/common/Chart.yaml
+++ b/bitnami/common/Chart.yaml
@@ -2,7 +2,7 @@ annotations:
   category: Infrastructure
 apiVersion: v2
 # Please make sure that version and appVersion are always the same.
-appVersion: 1.0.0
+appVersion: 1.0.1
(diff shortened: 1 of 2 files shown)


sereinity:
Tests and lint fail because the future version of `bitnami/common` (required in mariadb's `Chart.lock`) isn't released yet.

FraPazGal (commented):
Hi @sereinity,

Thank you for catching and fixing this! 

Regarding the changes to `bitnami/mariadb`, you don't need to bump MariaDB's chart version or change its `Chart.lock`. When the new version of `common` is released, our internal tools will update the `Chart.lock` files automatically.

sereinity:
Thank you @FraPazGal , commit removed.

FraPazGal (approved):
LGTM!

PR #4433, opened 2020-11-19
[bitnami/mariadb-galera] Allow to set custom admin username

matejhasul:
Signed-off-by: Matej Hasul <matej.hasul@gooddata.com>

**Description of the change**

Do not hardcode username of admin user to `root`. Allow to create custom admin username which defaults to `root`.

Motivation is to support admin (`root`) user credential rotation. If you change password of `root` user in the databsae then probes start to fail. I'm not aware of any other option (share your ideas!) how to do the rotation other than to create new admin user (i.e. `root2`) with new password manually in the database and then upgrade the chart. Problem is that chart does not support custom admin username although entrypoint support it.

**Benefits**
To ease admin user credential rotation. 

**Possible drawbacks**
`¯\_(ツ)_/¯`

Probably this one:
`{{ .Values.rootUser.user }}` is a bit ugly, should be changed to `{{ .Values.admin.user }}`

**Applicable issues**

  - fixes #

**Additional information**

**Checklist** 


FraPazGal:
Hi @matejhasul,

Thanks for the PR, it looks good! As you said, only changing the root password in the DB will fail, you would need to upgrade the chart with `--set rootUser.password` for the probes to work. Nevertheless, I think this is a good addition to the chart!

Regarding the new parameter's name, we could use `.Values.rootUser.username` to avoid possible confusion, but changing the existing parameters to `.Values.admin.xxxx` would be a breaking change. I would leave `.Values.rootUser` so this doesn't constitute a major, what do you think?

matejhasul:
> Regarding the new parameter's name, we could use `.Values.rootUser.username` to avoid possible confusion, but changing the existing parameters to `.Values.admin.xxxx` would be a breaking change. I would leave `.Values.rootUser` so this doesn't constitute a major, what do you think?

I'm ok with that, less work for me :)

matejhasul:
I've squashed the commits.

FraPazGal (approved):
Thank you again for the PR!

LGTM!

PR #4439, opened 2020-11-19
[bitnami/minio] Let chart consumer indicate their own path to load credential files

### LINKED ISSUE #4422: [bitnami/minio] Provide own credentials file
**Which chart**:
bitnami/minio (4.0.2)

I'm trying to get Vault agent injector to provide the credentials used by Minio so that there's a single source of truth for the secrets used in my cluster. (https://www.vaultproject.io/docs/platform/k8s/injector)

When the useCredentialsFile property is set to true, it would be nice if I could indicate my own path to load the access-key and secret-key files from. The minio secret then also doesn't need to be created.

The Vault agent lets you indicate different paths to mount to share the vault secrets with Minio, but the problem is that this is already a mounted path for the secret.

Sleepy-GH:
**Description of the change**

This adds the possibility for chart consumers to indicate where Minio can load their credential files from. See https://github.com/bitnami/charts/issues/4422 for some more context.

**Benefits**

This will enable chart consumers to use i.e. Vault agent injector to specify the credentials.

**Possible drawbacks**

I have made some changes to the values.yaml that aren't backwards compatible.

**Applicable issues**

  - fixes #4422 

**Additional information**

N/A

**Checklist** 


bitnami/minio/Chart.yaml
```diff
@@ -19,4 +19,4 @@ name: minio
 sources:
   - https://github.com/bitnami/bitnami-docker-minio
   - https://min.io
-version: 4.0.2
+version: 4.0.3
```
Sleepy-GH, line 22: To any reviewers, not sure if this is a correct new version for the chart as the values.yaml has some incompatible changes. Any suggestions?

bitnami/minio/README.md
```diff
@@ -84,7 +84,11 @@ The following table lists the configurable parameters of the MinIO chart and the
...
 | `deployment.updateStrategy`          | Deployment update strategy policy                                                                                                                         | `Recreate`                                              |
 | `existingSecret`                     | Existing secret with MinIO credentials                                                                                                                    | `nil`                                                   |
-| `useCredentialsFile`                 | Have the secret mounted as a file instead of env vars                                                                                                     | `false`                                                 |
+| `credentialsFile.enabled`            | Use credential files in stead of env vars                                                                                                     | `false`                                                 |
```
andresbono, line 87:
```suggestion
| `credentialsFile.enabled`            | Use credential files instead of env vars                                                                                                      | `false`                                                 |
```

bitnami/minio/Chart.yaml
andresbono: This change should definitely bump the major or at least the minor.

bitnami/minio/values-production.yaml
```diff
@@ -143,9 +143,14 @@ statefulset:
...
-## Mount MinIO secret as a file instead of passing environment variable
+## Mount MinIO secret as a file instead of passing environment variable. If useSecretValues is true it will mount the values from the minio secret. Otherwise a local file can be used.
 ##
-useCredentialsFile: false
+credentialsFile:
+  enabled: false
+  useSecretValues: true
+  mountPath: /opt/bitnami/minio/secrets/
+  accessKeyFile: access-key
+  secretKeyFile: secret-key
```
andresbono, line 153: Could you extend the description of each parameter a little bit more? Check the rest of the file for examples of the format. You can just use the descriptions from the README.

bitnami/minio/templates/deployment-standalone.yaml
```diff
@@ -74,19 +74,19 @@ spec:
               value: {{ ternary "true" "false" .Values.image.debug | quote }}
             - name: MINIO_FORCE_NEW_KEYS
               value: {{ ternary "yes" "no" .Values.forceNewKeys | quote }}
-            {{- if .Values.useCredentialsFile }}
+            {{- if .Values.credentialsFile.enabled }}
             - name: MINIO_ACCESS_KEY_FILE
-              value: "/opt/bitnami/minio/secrets/access-key"
+              value: {{ .Values.credentialsFile.mountPath }}{{ .Values.credentialsFile.accessKeyFile }}
```
andresbono, line 79: For safety, I would use a `/` between template expansions, in camse mountPath is provided without the final slash.

```suggestion
              value: {{ .Values.credentialsFile.mountPath }}/{{ .Values.credentialsFile.accessKeyFile }}
```

Please update the rest of paths, if you think this suggestion is useful.

andresbono (commented):
Thank you very much for providing this PR!!!

Could you please check my comments? After that, we will give another round of review and we will test it on our side, but it would be great if you also share how are you testing this to have an idea of possible scenarios that we may be missing.

Sleepy-GH:
> Thank you very much for providing this PR!!!
> 
> Could you please check my comments? After that, we will give another round of review and we will test it on our side, but it would be great if you also share how are you testing this to have an idea of possible scenarios that we may be missing.

No problem, thanks for taking the time to review it! I'll go over the suggested changes and implement them.
I've tested it by running the linter and rendering the templates locally. I checked that the paths were correct and that the default values weren't changed from the original situation. Since there were also some minor changes in the if / else logic in the persistentVolumeClaimTemplate in the statefulset.yaml I also checked that the persistence looked alright.

If it gives you some ease of mind I can deploy a set of the rendered templates to a cluster later today and see if it all seems to be working okay.

bitnami/minio/Chart.yaml
Sleepy-GH: Bumped it to 5.0.0

bitnami/minio/templates/deployment-standalone.yaml
Sleepy-GH: Yup, sounds good. I updated all the paths with a dividing /

bitnami/minio/values-production.yaml
Sleepy-GH: I copy pasted the README descriptions into comments above the parameters with formatting like the rest of the file (in both values.yaml and values-production.yaml)

carrodher:
Thanks for creating this PR. As it is a major version we need to double-check everything carefully as well as look for other major features that can be incorporated in order to avoid bumping the major version frequently. No more action on your side right now, we will test & review the changes.

Sleepy-GH:
> Thanks for creating this PR. As it is a major version we need to double-check everything carefully as well as look for other major features that can be incorporated in order to avoid bumping the major version frequently. No more action on your side right now, we will test & review the changes.

Understood, thank you for taking the time :) 
If I can find the time I'll create a similar change for the influxdb chart. Should I open up a separate issue for that?

bitnami/minio/templates/NOTES.txt
```diff
@@ -4,10 +4,18 @@ MinIO can be accessed via port {{ .Values.service.port }} on the following DNS n
 
    {{ include "minio.fullname" . }}.{{ .Release.Namespace }}.svc.{{ .Values.clusterDomain }}
 
+{{- if and .Values.credentialsFile.enabled (not .Values.credentialsFile.useSecretValues) }}
```
javsalgar, line 7: Looking at this, I think we could avoid the major version bump if we allow setting either Values.credentialsFile.enabled or useCredentialsFile. This would make the solution backwards compatible. Maybe we could throw a warning if useCredentialsFile is set. Thoughts?

carrodher: Yes, I agree. If there is a way to avoid the breaking change and the major version bump it would be nice to implement it. Do you know any parameter in another Helm Chart using the proposed logic? In this way, Sleepy-GH can use it as an example to perform the suggested changes.

Sleepy-GH: I'm waiting for an answer on this. I can try and provide my own backwards compatible fix if that's easier?

javsalgar: Hi! In thanos we did something similar. 

https://github.com/bitnami/charts/pull/4307/files#diff-f0d8330197d82de54920f75eac2c5cba46aa5d6918b5ea4fda8d7eb5a82ca3b9R168-R172

The idea would be use an if statement and check whether one value is set or the other. It could be done in a helper funciton.

carrodher:
Unfortunately, due to other priorities and the limited bandwidth, we have the work on this PR was not continued and there are no plans to work on it in the mid-term. I'm replacing the `on-hold` label with the `stale` one so the PR will be automatically closed if it is not continued during the following days.

PR #4451, opened 2020-11-20
[bitnami/ghost] set default DeploymentStrategy to Recreate

morremeyer:
**Description of the change**

This commit enables configuration of the DeploymentStrategy and sets the default DeploymentStrategy to Recreate to match the other default settings.

**Benefits**

The chart can be updated with default settings.

**Possible drawbacks**

Users of ReadWriteMany volumes will need to change the strategy as Recreate is not necessary with RWX volumes.

**Applicable issues**

None.

**Additional information**

Nothing.

**Checklist** 

### DIFF
--- a/bitnami/ghost/README.md
+++ b/bitnami/ghost/README.md
@@ -146,6 +146,7 @@ The following table lists the configurable parameters of the Ghost chart and the
 | `smtpPassword`                              | SMTP password                                                                                                         | `nil`                                                        |
 | `smtpFromAddress`                           | SMTP from address                                                                                                     | `nil`                                                        |
 | `smtpService`                               | SMTP service                                                                                                          | `nil`                                                        |
+| `strategy`                                  | The DeploymentStrategy to use when updating (Recreate or RollingUpdate)                                               | `{ type: Recreate }`                                         |
 | `tolerations`                               | Tolerations for pod assignment                                                                                        | `[]` (The value is evaluated as a template)                  |
 | `updateStrategy`                            | Deployment update strategy                                                                                            | `nil`                                                        |
 
--- a/bitnami/ghost/Chart.yaml
+++ b/bitnami/ghost/Chart.yaml
@@ -33,4 +33,4 @@ name: ghost
 sources:
   - https://github.com/bitnami/bitnami-docker-ghost
   - http://www.ghost.org/
-version: 11.2.1
+version: 12.0.0
--- a/bitnami/ghost/templates/deployment.yaml
+++ b/bitnami/ghost/templates/deployment.yaml
@@ -14,7 +14,8 @@ metadata:
 spec:
   selector:
     matchLabels: {{- include "common.labels.matchLabels" . | nindent 6 }}
-  replicas: {{ .Values.replicaCount }}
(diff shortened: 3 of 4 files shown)


carrodher:
Thanks for creating this PR. As it is a major we need to double-check everything carefully as well as look for other major features that can be incorporated in order to avoid bumping the major version frequently. No more action on your side right now, we will test & review the changes

bitnami/ghost/README.md
```diff
@@ -146,6 +146,7 @@ The following table lists the configurable parameters of the Ghost chart and the
 | `smtpPassword`                              | SMTP password                                                                                                         | `nil`                                                        |
 | `smtpFromAddress`                           | SMTP from address                                                                                                     | `nil`                                                        |
 | `smtpService`                               | SMTP service                                                                                                          | `nil`                                                        |
+| `strategy`                                  | The DeploymentStrategy to use when updating (Recreate or RollingUpdate)                                               | `{ type: Recreate }`                                         |
```
juan131, line 149: Is there any reason why we should use "Recreate" as default value?
```console
$ k explain deployment.spec.strategy.type
KIND:     Deployment
VERSION:  apps/v1

FIELD:    type <string>

DESCRIPTION:
     Type of deployment. Can be "Recreate" or "RollingUpdate". Default is
     RollingUpdate.
```
I would use a different approach to ensure users don't try to scale the solution while using "ReadWriteOnce" PVCs. We can create a validation in the `_helpers.tpl` that shows an error if `persistence.accessMode` is not "ReadWriteMany" when `replicaCount` is `>1`

juan131 (changes requested):

morremeyer:
[not stale]

juan131:
I added the `on-hold` label to avoid the stalebot to mark the PR as stale @morremeyer

bitnami/ghost/README.md
juan131: There are other alternatives that are less disruptive since they don't imply introducing a breaking change. E.g. using affinity presets.

For instance, we could set by default:
```yaml
podAntiAffinityPreset: ""
podAffinityPreset: hard
persistence:
  enabled: true
  accessMode: ReadWriteOnce
  size: 8Gi
```
By using this, the Ghost deployment would include this section:
```yaml
...
    spec:
      affinity:
        podAffinity:
            - labelSelector:
                matchLabels:
                  app.kubernetes.io/name: ghost
                  app.kubernetes.io/instance: ghost
              namespaces:
                - default
              topologyKey: kubernetes.io/hostname
```
Preventing new pods to be scheduled on different K8s nodes so the PVs can be attached to new pods

juan131 (changes requested):

bitnami/ghost/README.md
morremeyer: That’s a good idea, I’ll update the PR once I get around to it. Thanks!

juan131:
@morremeyer did you have a chance to take a look to my comments?

morremeyer:
I didn’t get around to update the MR and it is unneeded now as #5558 already enables this feature.

PR #4453, opened 2020-11-21
[bitnami/metrics-server] Support for extraVolumeMounts and non-privileged metrics readers

chadlwilson:
**Description of the change**

1. Adds support for `extraVolumeMounts` so you can mount a `tmpDir` and thus run with a `readOnlyRootFilesystem`
2. Creates the aggregating `ClusterRole` similar to the off-the-shelf [metrics-server rbac manifest](https://github.com/kubernetes-sigs/metrics-server/blob/master/manifests/base/rbac.yaml#L5) to allow non-admin users to see metrics

**Benefits**

Improved security posture through
- compatibility with unprivileged PSPs that don't allow mutable rootFilesystems
- ability for non-privileged users to still view metrics

**Possible drawbacks**

None known

**Checklist** 

### DIFF
--- a/bitnami/metrics-server/Chart.yaml
+++ b/bitnami/metrics-server/Chart.yaml
@@ -17,4 +17,4 @@ name: metrics-server
 sources:
   - https://github.com/bitnami/bitnami-docker-metrics-server
   - https://github.com/kubernetes-incubator/metrics-server
-version: 5.0.2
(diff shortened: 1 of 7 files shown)


alemorcuq (approved):
Thanks for this, @chadlwilson ! And thanks for implementing the validation as well!

PR #4470, opened 2020-11-24
[bitnami/rabbitmq] Start to use shutdown script waiting for synchronization

### LINKED ISSUE #3931: [bitname/rabbitmq] RollingUpdate taking replication into account?
**Which chart**:
rabbitmq

**Question**
When a the rabbitmq statefulset is updated and the update strategy is rollingUpdate the nodes are deleted and created again  successively. My understanding is that replicated queues then need to be replicated to a node when it comes up. Is there a mechanism so that it has replicated all messages it should before the next node is deleted?

I think this can also be asked like this: Does `rabbitmq-diagnostics -q check_running` report the node as down as long as it is still replicating after being started?

If the answer is no there is a risk that an update leads to loosing data if there are a large amount of messaging in replicated queues in the cluster. This is something that happens when using the rabbitmq-ha chart and I would like to know if this problem has been addressed in the bitnami rabbitmq chart.

I'm sorry if there is another preferred channels for questions like this, but I couldn't find it.

msvticket:
**Description of the change**

Start using script `nodeshutdown.sh` added in bitnami/bitnami-docker-rabbitmq#142 as preStop hook.

**Benefits**

If you have a cluster with replicated queues with many messages using this script as a this will reduce the risk for data loss especially when doing a configuration update that require all nodes/containers to be recreated.

**Possible drawbacks**

Stopping the container might take longer.

**Applicable issues**

  - Fixes #3931

**Checklist** 


bitnami/rabbitmq/templates/statefulset.yaml
```diff
@@ -240,14 +240,11 @@ spec:
...
-                    until rabbitmqctl cluster_status >/dev/null; do
-                        echo "Waiting for cluster readiness..."
-                        sleep 5
-                    done
-                    rabbitmq-queues rebalance "all"
+                  - /opt/bitnami/scripts/rabbitmq/nodeshutdown.sh
+                  - -t
+                  - {{ .Values.terminationGracePeriodSeconds | quote }}
+                  - -d
+                  - {{ ternary "true" "false" .Values.image.debug | quote }}
```
javsalgar, line 247: You mentioned it should be a preStop hook but I see it in the postStart hook. Could you double check this?

bitnami/rabbitmq/Chart.yaml
```diff
@@ -23,4 +23,4 @@ name: rabbitmq
 sources:
   - https://github.com/bitnami/bitnami-docker-rabbitmq
   - https://www.rabbitmq.com
-version: 8.1.1
+version: 8.1.2
```
javsalgar, line 26: It is adding a non-breaking feature so I believe it should be a minor version bump.

javsalgar (changes requested):
Thank you very much! Please check my comments

bitnami/rabbitmq/templates/statefulset.yaml
msvticket: True (I can't use this chart anymore since I can't use helm 3 yet. Unfortunately I made a copy-paste error from the chart I actually use.)

bitnami/rabbitmq/Chart.yaml
msvticket: OK

javsalgar (approved):
LGTM

mltsy:
I love this update - I do think a documentation section on rolling upgrades is sorely needed here to clarify the expected behavior and impact of the available settings.

If you have a classic mirrored queue that is is not setup for automatic syncing, the default termination grace-period of 120 seconds could be way too short.  (It might actually require a manual intervention to sync, depending on the situation)

I *think* the options available in rabbit (ha-sync-mode, specifically) provide enough functionality (you can enable automatic synchronization on a per-queue basis), but we should still document what happens in each case, in the context of this chart.

Should I put in a new issue for this?

miguelaeh:
Hi @mltsy ,
Thank you very much for your suggestion. Would you like to create a PR adding that documentation? We will really appreciate it.

mltsy:
Ah, okay.  I might be able to get something together over the holiday :)

Issue #4344, opened 2020-11-12
[bitnami/thanos] inconsistent caching configuration in values.yaml


mhulscher:
**Which chart**:

bitnami/thanos

**Is your feature request related to a problem? Please describe.**

Recently the query-frontend component has been added to the thanos helm-chart. The query-frontend supports caching query-range results in memcache or in memory. This configuration can be passed in the `values.yaml` file at `queryFrontend.config`.

However, other components that consume configmaps for cache configuration are configured in a different way. Whereas the query caching configuration lives under the `queryFrontend` key, that's not true for the following settings:

| key | desc |
|---|---|
| `objstoreConfig` | rightfully at top-level because it is shared by multiple components (at least storegateway and compactor)
| `indexCacheConfig` | lives at the top-level but is specific to the storegateway 
| `bucketCacheConfig` | lives at the top-level but is specific to the storegateway

Additionally, the next release of thanos seems to support a 2nd caching configuration for the query-frontend component; caching of labels: https://github.com/thanos-io/thanos/pull/3315 . This would make the current key at `queryFrontend.config` confusing. Am I configuring the query-range result cache or the labels cache?

**Describe the solution you'd like**

I am curious to know if there has already been discussions regarding this topic.

I can see at least two options to make the caching configurations more consistent.

1. Move all caching configuration to the top-level of the `values.yaml` and use specific names, for example:
```yaml
indexCacheConfig:
  foo: bar

queryRangeResultCacheConfig:
  baz: qux
```

2. Move all caching configuration under their respective component and use specific key names, for example
```yaml
storegateway:
  indexCacheConfig:
    foo: bar

  bucketCacheConfig:
    baz: qux

queryFrontend:
  queryRangeResultCacheConfig:
    foo: bar

  LabelsResponseCacheConfig:
    baz: qux

```

Regardless of what would happen, this will likely result in a backwards incompatible change in the format of the `values.yaml` file.

If there's a preferred direction then I am willing to help author a PR.

**Describe alternatives you've considered**

N/A

**Additional context**

N/A

miguelaeh:
Hi @mhulscher ,
It seems we already support that, see https://github.com/bitnami/charts/blob/c1b6bf77051f5715151a032584d12b75df34a408/bitnami/thanos/values.yaml#L40

mhulscher:
[...]

PR #4473, opened 2020-11-24
[bitnami/thanos] allow $component.extraFlags to contain multi-line strings

mhulscher:
**Description of the change**

Deserialize `$component.extraFlags` as YAML instead of looping over each item.

**Benefits**

This makes it possible to add multi-line flags from `values.yaml`. Thanos allows configuring caching for components directly from the command-line without requiring a configMap. Users can easily add a new cache configuration without having to wait for that particular cache to be supported by this helm-chart.

**Possible drawbacks**

Can't think of any.

**Applicable issues**

This would make https://github.com/bitnami/charts/issues/4344 less troublesome.

**Additional information**

N/A

**Checklist** 


carrodher:
Thanks for your contribution. I just manually tested the current approach of adding the flags and it was not modified, so the backward compatibility is guaranteed:

- Existing approach:
```console
$ helm template thanos bitnami/thanos --set query.extraFlags[0]=--sync-block-duration=3m --set query.extraFlags[1]=--chunk-pool-size=2GB
...
      containers:
        - name: query
          image: docker.io/bitnami/thanos:0.17.0-scratch-r0
          args:
            - query
            - --log.level=info
            - --grpc-address=0.0.0.0:10901
            - --http-address=0.0.0.0:10902
            - --query.replica-label=replica
            - --sync-block-duration=3m
            - --chunk-pool-size=2GB
```

- New approach:
```console
$ helm template thanos . --set query.extraFlags[0]=--sync-block-duration=3m --set query.extraFlags[1]=--chunk-pool-size=2GB
...
      containers:
        - name: query
          image: docker.io/bitnami/thanos:0.17.0-scratch-r0
          args:
            - query
            - --log.level=info
            - --grpc-address=0.0.0.0:10901
            - --http-address=0.0.0.0:10902
            - --query.replica-label=replica
            - --sync-block-duration=3m
            - --chunk-pool-size=2GB
```
In both cases the flags are added as arguments to the container, in the same way, so the current workflow is not modified. The same happens when modifying directly the _values.yaml_.

Apart from that, I also perform some tests using the multi-line approach by adding
```diff
  ## Extra Flags to passed to Thanos Query
  ##
- extraFlags: []
+ extraFlags:
+   - |
+     --query-range.response-cache-config=
+     type: MEMCACHED
+     config:
+       addresses:
+         - <MEMCACHED_SERVER>:11211
+       timeout: 500ms
+       max_idle_connections: 100
+       max_async_concurrency: 10
+       max_async_buffer_size: 10000
+       max_get_multi_concurrency: 100
+       max_get_multi_batch_size: 0
+       dns_provider_update_interval: 10s
+       expiration: 24h
```
to the _values.yaml_, in that case, it is rendered as follow:
```yaml
      containers:
        - name: query
          image: docker.io/bitnami/thanos:0.17.0-scratch-r0
          args:
            - query
            - --log.level=info
            - --grpc-address=0.0.0.0:10901
            - --http-address=0.0.0.0:10902
            - --query.replica-label=replica
            - |
              --query-range.response-cache-config=
              type: MEMCACHED
              config:
                addresses:
                  - <MEMCACHED_SERVER>:11211
                timeout: 500ms
                max_idle_connections: 100
                max_async_concurrency: 10
                max_async_buffer_size: 10000
                max_get_multi_concurrency: 100
                max_get_multi_batch_size: 0
                dns_provider_update_interval: 10s
                expiration: 24h
```
is the `- |` expected or the line should start directly with the `--query-range`? In that case, can you implement the required changes to obtain the desired result?

mhulscher:
Hi @carrodher ,

The extra pipe symbol is supposed to be there, so it becomes `- |`. This is followed by the multi-line string on a new line. This is how a multi-line string is defined in YAML. It could also be written as `- |-` which would add an additional newline to the end of the multi-line string, but that doesn't change the rendered template in this case.

Just like you, I tested these changes for backwards compatibility and found no issues.

For reference:
* [There's this example from thanos-io/kube-thanos](https://github.com/thanos-io/kube-thanos/blob/master/examples/all/manifests/thanos-query-frontend-deployment.yaml#L51)
* [There's also the YAML spec](https://yaml-multiline.info/)

carrodher:
Sounds good, thanks for the clarification. I am going to trigger our internal test & release pipeline, if everything works as expected those changes should be tested and released during the morning, in the same way, the PR will be merged.

PR #4476, opened 2020-11-24
[bitnami/*] Affinity based on common presets (iii)

juan131:

**Description of the change**

This PR adapts the charts below so it uses the common presets added on `bitnami/common` at https://github.com/bitnami/charts/pull/3713.

- bitnami/kube-prometheus
- bitnami/kube-state-metrics
- bitnami/kubewatch

It also refactors Kubewatch and kube-state-metrics based on best practices since they had a lot of technical debt. These changes implied a major version in the case of Kubewatch.

**Benefits**

- Balanced UX for both advanced Kubernetes admins and less experienced.
- Opinionated presets but users can still define their own affinity
- 100% customizable
- No backwards incompatibility

**Possible drawbacks**

- Complexity

**Checklist**


bitnami/kube-state-metrics/values-production.yaml
```diff
@@ -187,18 +187,52 @@ updateStrategy: {}
...
+  values: []
+
+## Affinity for pod assignment. Evaluated as a template.
 ## Ref: https://kubernetes.io/docs/concepts/configuration/assign-pod-node/#affinity-and-anti-affinity
+## Note: podAffinityPreset, podAntiAffinityPreset, and nodeAffinityPreset will be ignored when it's set
 ##
 affinity: {}
 
-## Node labels for pod assignment
-## Ref: https://kubernetes.io/docs/user-guide/node-selection/
```
miguelaeh, line 196: Just curiosity. I have seen some PR where this is changed from `ref` to `Ref` and others like this one, that makes the opposite.
Maybe we should make it standard

bitnami/kube-prometheus/README.md
```diff
@@ -107,11 +107,14 @@ The following table lists the configurable parameters of the kube-prometheus cha
...
 | `operator.podAnnotations`                             | Pod annotations                                                                                                     | `{}`                                                                         |
-| `operator.nodeAffinity`                               | Node Affinity (this value is evaluated as a template)                                                               | `{}`                                                                         |
-| `operator.podAntiAffinity`                            | Pod anti-affinity policy                                                                                            | `soft`                                                                       |
-| `operator.podAffinity`                                | Affinity, in addition to antiAffinity (this value is evaluated as a template)                                       | `{}`                                                                         |
```
miguelaeh, line 112: It seems these parameters will disappear and could cause missing sections in the templates while rendering after an upgrade.

miguelaeh (commented):
In general LGTM, except for the fact that in the case of Kube-prometheus the backwards compatibility is not totally supported due to parameters that will disappear. The upgrade will not fail, but I would add that note in the README.

miguelaeh (approved):
LGTM!

PR #3850, opened 2020-10-01
[bitnami/wordpress] New major version

juan131:
Signed-off-by: juan131 <juanariza@vmware.com>

**Description of the change**

This PR bumps WP major version since its dependency (MariaDB) bumped the major version including breaking changes. More info at https://github.com/bitnami/charts/pull/3849

It also:

- Bump apiVersion from `v1` to `v2` in **Chart.yaml**
- Fields in **Chart.yaml** file has been ordered alphabetically
- Move dependency information from the **requirements.yaml** to the **Chart.yaml**

**Benefits**

- Standardization.
- Reduces technical debt.

**Possible drawbacks**

- It breaks backwards compatibility.
- Helm v2 is no longer supported

**Applicable issues**

None

**Additional information**

**Checklist**

- [x] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [x] Variables are documented in the README.md
- [x] Title of the PR starts with chart name (e.g. `[bitnami/chart]`)
- [x] If the chart contains a `values-production.yaml` apart from `values.yaml`, ensure that you implement the changes in both files

PR #4487, opened 2020-11-25
[bitnami/wordpress] - Add back previously removed _helper functions to unbreak post-init configmap

pestebogdan:
**Description of the change**

This PR: https://github.com/bitnami/charts/pull/4373 , which brings new functionality from versions >=9.10.0, requires the declaration of 2 helper scripts, which were removed from **_helpers.tpl** as of version 10.0.0, specifically:
  - wordpress.fullname
  - wordpress.tplValue

This PR is meant to ammend the changes of https://github.com/bitnami/charts/pull/3850, specifically adding back these 2 helper functions, as without them, whenever the `customPostInitScripts` variable is populated, helm chart installation fails.

**Possible drawbacks**

None that I can see

### DIFF
--- a/bitnami/wordpress/Chart.yaml
+++ b/bitnami/wordpress/Chart.yaml
@@ -31,4 +31,4 @@ name: wordpress
 sources:
   - https://github.com/bitnami/bitnami-docker-wordpress
   - http://www.wordpress.com/
-version: 10.0.6
+version: 10.0.7
--- a/bitnami/wordpress/templates/_helpers.tpl
+++ b/bitnami/wordpress/templates/_helpers.tpl
@@ -4,23 +4,6 @@
 Create a default fully qualified app name.
 We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
 */}}
-{{- define "wordpress.fullname" -}}
-{{- if .Values.fullnameOverride -}}
-{{- .Values.fullnameOverride | trunc 63 | trimSuffix "-" -}}
-{{- else -}}
-{{- $name := default .Chart.Name .Values.nameOverride -}}
-{{- if contains $name .Release.Name -}}
-{{- .Release.Name | trunc 63 | trimSuffix "-" -}}
-{{- else -}}
-{{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" -}}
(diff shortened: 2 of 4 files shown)


miguelaeh (changes requested):
Hi @pestebogdan ,
Instead of adding those functions back we should be using the following:
For the fullname:
```
{{ include "common.names.fullname" . }}
```

For the tplvalue, something like the following changing the value:
```
{{- include "common.tplvalues.render" (dict "value" .Values.metrics.serviceMonitor.additionalLabels "context" $) | nindent 4 }}
```

The tplValue needs to be changed in once at `templates/configmap.yaml` and twice at `templates/tls-secrets.yaml`. And the fullname needs to be changed once at `templates/configmap.yaml`.

Thank you for reporting the issue!!

pestebogdan:
> Hi @pestebogdan ,
> Instead of adding those functions back we should be using the following:
> For the fullname:
> 
> ```
> {{ include "common.names.fullname" . }}
> ```
> 
> For the tplvalue, something like the following changing the value:
> 
> ```
> {{- include "common.tplvalues.render" (dict "value" .Values.metrics.serviceMonitor.additionalLabels "context" $) | nindent 4 }}
> ```
> 
> The tplValue needs to be changed in once at `templates/configmap.yaml` and twice at `templates/tls-secrets.yaml`. And the fullname needs to be changed once at `templates/configmap.yaml`.
> 
> Thank you for reporting the issue!!

Thanks for the tip @miguelaeh ! I updated the changes with the correct values.

pestebogdan (commented):
Changes updated

miguelaeh (approved):
Thank you very much!!!!
LGTM!

miguelaeh:
Could you check errors with the testing please?

pestebogdan:
Hmm, not really sure why it is failing (some warning regarding a connection to DB) , not sure if it is related to this particular change. Is it possible to run the same test on the master branch to make sure the master is not broken ?

andresbono:
Hi @pestebogdan, thanks for checking the GitHub action. As you mention, master is not passing the tests: See https://github.com/bitnami/charts/pull/4567.

We will merge this PR as it is working properly and we will investigate the issue with the GitHub action.

Thanks for your contribution!

PR #4488, opened 2020-11-25
[bitnami/mongodb] ensure password is handled as string

ryandawsonuk:
fixes https://github.com/helm/charts/issues/23343

The referenced issue went stale but was not addressed. If the password is numeric then b64enc can't encode it. With this change it will be able to encode it. Using toString rather than quote to avoid adding unwanted quotes to the content of the password.

This will also affect other charts using b64enc on a potentially-numeric value, even if the value being entered is entered as a string such as "123"

### DIFF
--- a/bitnami/mongodb/templates/secrets.yaml
+++ b/bitnami/mongodb/templates/secrets.yaml
@@ -11,20 +11,20 @@ data:
   {{- if .Values.auth.rootPassword }}
   mongodb-root-password:  {{ .Values.auth.rootPassword | toString | b64enc | quote }}
   {{- else }}
-  mongodb-root-password: {{ randAlphaNum 10 | toString | b64enc | quote }}
+  mongodb-root-password: {{ randAlphaNum 10 | b64enc | quote }}
   {{- end }}
   {{- if and .Values.auth.username .Values.auth.database }}
   {{- if .Values.auth.password }}
   mongodb-password:  {{ .Values.auth.password | toString | b64enc | quote }}
   {{- else }}
-  mongodb-password: {{ randAlphaNum 10 | toString | b64enc | quote }}
+  mongodb-password: {{ randAlphaNum 10 | b64enc | quote }}
   {{- end }}
   {{- end }}
   {{- if eq .Values.architecture "replicaset" }}
   {{- if .Values.auth.replicaSetKey }}
   mongodb-replica-set-key:  {{ .Values.auth.replicaSetKey | toString | b64enc | quote }}
   {{- else }}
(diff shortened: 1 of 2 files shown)


bitnami/mongodb/templates/secrets.yaml
```diff
@@ -9,22 +9,22 @@ metadata:
 type: Opaque
 data:
   {{- if .Values.auth.rootPassword }}
-  mongodb-root-password:  {{ .Values.auth.rootPassword | b64enc | quote }}
+  mongodb-root-password:  {{ .Values.auth.rootPassword | toString | b64enc | quote }}
```
dani8art, line 12: That's a good catch but I think not the best approach, please, could you check why the linters and tests are failing?

bitnami/mongodb/templates/secrets.yaml
```diff
@@ -9,22 +9,22 @@ metadata:
...
   {{- else }}
-  mongodb-root-password: {{ randAlphaNum 10 | b64enc | quote }}
+  mongodb-root-password: {{ randAlphaNum 10 | toString | b64enc | quote }}
```
dani8art, line 14: For the ones that are auto-generated this is not needed.

dani8art (commented):
Hi, Thanks for the PR,

bitnami/mongodb/templates/secrets.yaml
ryandawsonuk: Failure says `Unable to find image 'quay.io/helmpack/chart-testing:v3.0.0-rc.1' locally`. Looks like it's the test infrastructure failing rather than something related to this change.

bitnami/mongodb/templates/secrets.yaml
ryandawsonuk: Have updated this now.

bitnami/mongodb/templates/secrets.yaml
ryandawsonuk: For context I hit the same problem with a minio chart https://github.com/minio/charts/issues/68

ryandawsonuk: Ok this time checks say `✖︎ mongodb => (version: "10.1.2", path: "bitnami/mongodb") > Chart version not ok. Needs a version bump!`

ryandawsonuk: Passing now

dani8art: Cool! 👍

dani8art (approved):
LGTM, Thanks!

PR #4497, opened 2020-11-26
[bitnami/common] Add mongodb validation template

carrodher:
Add MongoDB validations to bitnami/common chart

PR #4498, opened 2020-11-26
[bitnami/mongodb] Use common validations

carrodher:
**Description of the change**

Continuation of https://github.com/bitnami/charts/pull/4497, we should wait for the common release before merging this PR. Once released the new common, we should run `helm dep update` and check that the new version is included in the Chart.lock


joancafom (approved):
LGTM! Please, wait until https://github.com/bitnami/charts/pull/4497 is published to merge this PR 🙂

Issue #4508, opened 2020-11-26
[rabbitmq] invalid: spec.clusterIP: Invalid value: "": field is immutable


silverbackdan:
**Which chart**:
rabbitmq 8.2.0

**Describe the bug**
When upgrading with Helm v3 the error is `Service "review-feature-qe-dh9bxm-rabbitmq" is invalid: spec.clusterIP: Invalid value: "": field is immutable`

**To Reproduce**
Attempt to upgrade an existing rabbitmq release

**Additional context**
I believe the issue lies in https://github.com/bitnami/charts/blob/master/bitnami/rabbitmq/templates/svc.yaml

If the service type is ClusterIP then I believe that spec.clusterIP needs to be set to `None` for Helm v3.

I think the issue listed here is similar https://github.com/helm/charts/issues/16010 - however looking at svc.yaml it does not appear that we are able to set this with values on this chart. I couldn't see an issue raised in here relating to this so forgive me if I'm missing something.

TIA

silverbackdan:
Also see: https://github.com/helm/charts/issues/20510

miguelaeh:
Hi @silverbackdan ,
I have tried the upgrade and it works fine for me. Which version did you have of the Helm Chart before upgrading to `8.2.0`?

silverbackdan:
Thank you @miguelaeh 

After this information, I deleted the entire namespace and updated Helm in my CI from 3.2.1 to 3.4.1 (when I realised it was using an outdated version). I updated my repositories for other charts to non-deprecated ones. I also updated helm on the server to 3.4.1. I uninstalled the chart I have which uses local templates and a couple of dependencies

Now the initial installation then works well and on every upgrade/re-deploy I'm now finding that every chart/template I'm using has this issue. So I can see it isn't related to this chart specifically, but more of a global issue I'm facing. A little stuck now, but thank you for your time. Much appreciated, and sorry for you having to spend the time to replicate a non-existent issue.

silverbackdan:
🤦‍♂️ I had the `--force` option still in place in this CI script from Helm v2 implementation a while ago. So sorry, thanks again

PR #4509, opened 2020-11-26
[bitnami/rabitmq] Fix: Allow clusterIP config for Helm v3 upgrade support

silverbackdan:
fixes https://github.com/bitnami/charts/issues/4508

Bug fix to support ClusterIP on Helm v3. This allows a user to set the value to `None`


silverbackdan:
Mis-diagnosis. I was using `--force` option for the implementation I was having this trouble with.

PR #4545, opened 2020-11-30
[bitnami/kafka] Use autoDiscovery values for NodePort external access

mkerins:
**Description of the change**

If externalAccess is enabled and the service is NodePort the `EXTERNAL_ACCESS_IP` value is not used when writing the `setup.sh` script.

**Benefits**

Use the autoDiscovery feature when using NodePort for the external access service

**Possible drawbacks**

None

**Applicable issues**

None

**Additional information**

None

**Checklist** 

### DIFF
--- a/bitnami/kafka/templates/scripts-configmap.yaml
+++ b/bitnami/kafka/templates/scripts-configmap.yaml
@@ -84,7 +84,9 @@ data:
     {{- end }}
     export EXTERNAL_ACCESS_PORT={{ .Values.externalAccess.service.port }}
     {{- else if eq .Values.externalAccess.service.type "NodePort" }}
-    {{- if .Values.externalAccess.service.domain }}
+    {{- if .Values.externalAccess.autoDiscovery.enabled }}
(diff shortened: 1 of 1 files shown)


bitnami/kafka/Chart.yaml
```diff
@@ -29,4 +29,4 @@ name: kafka
 sources:
   - https://github.com/bitnami/bitnami-docker-kafka
   - https://kafka.apache.org/
-version: 12.2.1
+version: 12.3.0
```
andresbono, line 32: This seems like a bug fix, right? Should we just bump the patch version?

```suggestion
version: 12.2.2
```

andresbono (commented):
Thanks for your contribution!

bitnami/kafka/Chart.yaml
mkerins: I wasn't sure, it's not really a bug, more of an enhancement but not quite a feature.

andresbono (approved):
LGTM

PR #4549, opened 2020-12-01
[bitnami/fluentd] Added field to append additional FluentD configuration

MissakaI:
**Description of the change**

An additional field is introduced to the `values.yaml` named `forwarder.extraConfig` and `aggregator.extraConfig` to add fluentd configuration string that is appended to the `forwarder-configmap.yaml/data/fluentd.conf` and `aggregator-configmap.yaml/data/fluentd.conf` 

**Benefits**

- Add additional configuration without needing to install new a ConfigMap. Drawback of adding a new ConfigMap using the `forwarder.configMap` and `aggregator.configMap` is the user has to add all the Metrics configuration to the new ConfigMap which is already configured in the existing ConfigMap. And new ConfigMap also will have to setup the forwarding configuration as well which too is available in existing one.
- This allows the user to add custom FluentD configuration to be loaded during the chart installation and reduce the need to change configuration during runtime.

**Possible drawbacks**

**Applicable issues**

**Additional information**

**Checklist** 


bitnami/fluentd/README.md
```diff
@@ -326,6 +367,11 @@ aggregator.extraEnv[1].name=ELASTICSEARCH_PORT
 aggregator.extraEnv[1].value=your-port-here
 ```
 
+#### By changing the value at `aggregator.extraConfig`
+
+An example of how to do this is available in the `values-production.yaml`
+
+
```
andresbono, line 374: Please remove extra new line

```suggestion
```

bitnami/fluentd/README.md
```diff
@@ -237,11 +239,50 @@ This chart includes a `values-production.yaml` file where you can find some para
 + forwarder.rbac.pspEnabled: true
 ```
 
+- Enable forwarding logs to another service like ElasticSearch
```
andresbono, line 242:
```suggestion
- Enable forwarding logs to another service like Elasticsearch
```

bitnami/fluentd/README.md
```diff
@@ -237,11 +239,50 @@ This chart includes a `values-production.yaml` file where you can find some para
...
+-    # Send all logs to the standard output
+-    <match **>
+-      @type stdout
+-    </match>
++  aggregator.extraConfig: |
++   # Send all logs to elasticsearch 
++    <match **>
++      @type elasticsearch
++      host elasticsearch-master-headless.default.svc.cluster.local
++      port 9200
```
andresbono, line 254: Wouldn't it be better if this is not hardcoded? In the configmap.yaml example, we use env-vars. What do you think?

bitnami/fluentd/values-production.yaml
```diff
@@ -358,6 +371,40 @@ aggregator:
...
+  ## ConfigMap created by this chart. For more details on default ConfigMap view
+  ## /templates/aggregator-configmap.yaml
+  ## To learn more about fluend configurations
+  ## https://docs.fluentd.org/configuration/config-file#list-of-directives
+  extraConfig: |
+    # Send all logs to elasticsearch 
+    <match **>
+      @type elasticsearch
+      host elasticsearch-master-headless.default.svc.cluster.local
+      port 9200
```
andresbono, line 384: This won't work, right? You need to have an elasticsearch deployed. What if we leave the example in the README but we keep some default values here that doesn't require dependencies with other charts?

bitnami/fluentd/templates/forwarder-configmap.yaml
```diff
@@ -101,10 +105,12 @@ data:
         flush_interval 5s
       </buffer>
     </match>
-    {{- else }}
+    {{ else }}
     # Send the logs to the standard output
     <match **>
       @type stdout
     </match>
```
andresbono, line 112: In the aggregator, you mention a problem with multiple wide matches:

```
    # Following config will match widely causing any matches that follows invalid. Hence commented out.
```

Isn't the issue the same here?

andresbono (commented):
Thanks for your contribution! Could you please check my comments?

bitnami/fluentd/README.md
MissakaI: Oh i didn't notice that. I will add it. Thank you.

bitnami/fluentd/templates/forwarder-configmap.yaml
MissakaI: I considered it as if the forwarder is not forwarding it would just print all out for debugging purposes. And if the users had something to do with the outputs won't it be done using the aggregator. 
What do you think is better.

MissakaI: Another solution would be to move

        {{ .Values.aggregator.extraConfig | indent 4 }}

it up to line 83. Then default match all won't affect user configuration. 
If this is considered then same can be done in the aggregator as well.

bitnami/fluentd/values-production.yaml
MissakaI: I didn't get you. Do you mean to remove the complete `match` block from the `values-production.yml` and leave it only as a example in the README.

bitnami/fluentd/README.md
```diff
@@ -237,11 +239,50 @@ This chart includes a `values-production.yaml` file where you can find some para
...
 
+- Enable forwarding logs to another service like Elasticsearch
+```diff
+-  aggregator.extraConfig: |
+-    # Send all logs to the standard output
+-    <match **>
+-      @type stdout
+-    </match>
++  aggregator.extraConfig: |
++   # Send all logs to elasticsearch
```
marcosbc, line 250: Isn't the comment wrongly indented here? I.e.:

```suggestion
+    # Send all logs to elasticsearch 
```

bitnami/fluentd/README.md
```diff
@@ -326,6 +367,10 @@ aggregator.extraEnv[1].name=ELASTICSEARCH_PORT
 aggregator.extraEnv[1].value=your-port-here
 ```
 
+#### By changing the value at `aggregator.extraConfig`
+
+An example of how to do this is available in the `values-production.yaml`
```
marcosbc, line 372: We are considering removing the `values-production.yaml` file soon, so this would not be applicable anymore. Could you copy the example here instead?

bitnami/fluentd/templates/forwarder-configmap.yaml
```diff
@@ -101,10 +105,12 @@ data:
         flush_interval 5s
       </buffer>
     </match>
-    {{- else }}
+    {{ else }}
```
marcosbc, line 108: Why removing the "`-`" from "`{{- else }}`"? That will cause the rendered contents to look a bit weird.

bitnami/fluentd/values-production.yaml
marcosbc: Like mentioned by @andresbono, I would not hardcode this example here as it would add an extra dependency on Elasticsearch, which is not defined in `Chart.yaml` (and thus making this chart impossible to deploy with the default contents of values.yaml).

marcosbc (changes requested):
I have added a few comments/requests, please have a look at them whenever you have a chance!

bitnami/fluentd/README.md
MissakaI: Ok.

bitnami/fluentd/templates/forwarder-configmap.yaml
MissakaI: With the `-` the yaml will remove the final `/n` which makes `extraConfig` to start in the same line as `</match>` and causes fluend to throw an error due to unfinished configuration.

bitnami/fluentd/values-production.yaml
MissakaI: Ok. Got it.

andresbono:
Hi @MissakaI, it seems like due to a rebase/merge, this PR is not updated with master. Could you please check it so that the PR only includes the desired changes for `bitnami/fluentd`?

MissakaI:
I will create a new PR. I messed up the history of my repo. Sorry about that.

MissakaI:
I messed up my fork's git history. I will fix it and open a new PR.

PR #4586, opened 2020-12-03
[bitnami/orangehrm] MAJOR: Follow non-root approach

andresbono:
**Description of the change**

We are working on releasing the OrangeHRM Docker image with a non-root approach. As part of that, we need to update the chart to use the new port numbers used by the container image and related settings.

We're also taking advantage of this PR being a major change by adding all standarizations (i.e. customizable command, extra volumes, extra envs, K8s labels standarization, etc.).

**Benefits**

**Possible drawbacks**

Breaking change with previous chart versions, as the default HTTP/HTTPS ports are changed to 8080/8443 (previously 80/443). The labels may also cause this to be a breaking change.

**Applicable issues**

None

**Additional information**

**Checklist** 


bitnami/orangehrm/templates/secrets.yaml
```diff
@@ -1,17 +1,23 @@
+{{- if not .Values.existingSecret }}
```
FraPazGal, line 1: Should we also include in the chart the option to use `smtopExistingSecret`?

bitnami/orangehrm/README.md
```diff
@@ -52,114 +52,181 @@ The following table lists the configurable parameters of the OrangeHRM chart and
...
+| `updateStrategy`                     | Deployment update strategy                                                                                            | `nil`                                          |
+
+### Traffic Exposure Parameters
+
+| Parameter                        | Description                             | Default           |
+|----------------------------------|-----------------------------------------|-------------------|
+| `service.type`                   | Kubernetes Service type                 | `LoadBalancer`    |
+| `service.port`                   | Service HTTP port                       | `80`              |
+| `service.httpsPort`              | Service HTTPS port                      | `443`             |
+| `service.externalTrafficPolicy`  | Enable client source IP preservation    | `Cluster`         |
```
FraPazGal, line 139: There seems to be two parameters missing in the table; `loadBalancerSourceRanges` and `loadBalancerIP`.

bitnami/orangehrm/templates/NOTES.txt
```diff
@@ -41,16 +52,36 @@
...
-
 1. Complete your OrangeHRM deployment by running:
 
   helm upgrade {{ .Release.Name }} --set service.type={{ .Values.service.type }},mariadb.enabled=false,externalDatabase.host=YOUR_EXTERNAL_DATABASE_HOST bitnami/orangehrm
 
 {{- end }}
 
-{{- if and (contains "bitnami/" .Values.image.repository) (not (.Values.image.tag | toString | regexFind "-r\\d+$|sha256:")) }}
+{{- include "common.warnings.rollingTag" .Values.image }}
+{{- include "common.warnings.rollingTag" .Values.metrics.image }}
```
FraPazGal, line 62: Should we include this warning for `volumePermissions.image` and `certificates.image`? If so, we could move this to a small helper to avoid adding too many lines here.

FraPazGal (commented):
Thanks for the PR, it looks great! I added a few comments, could you take a look at them?

bitnami/orangehrm/templates/NOTES.txt
andresbono: `volumePermissions` and `certificates` are using rolling tags (`buster`), we shouldn't be adding the warning for those, right?

FraPazGal (approved):
LGTM!

Issue #3783, opened 2020-09-25
[bitnami/rabbitmq] After deleting release nd then re-installing the pods never get past "Waiting for Mnesia tables"


zenbones:
Chart: 7.6.8

rabbitmq:
  service:
    type: NodePort
  replicaCount: 2
  persistence:
    enabled: true
    storageClass: gp2
    accessMode: ReadWriteOnce
    size: 2Gi
  resources:
    limits:
      cpu: 1000m
      memory: 2048M

Always works the first time. After shut down and restart works about half the time, the other half never restarts due to the waiting error on the db.

andresbono:
Hi @zenbones, where do you find the error? Do you find it in the container logs? Can you share the specific message?

kristinnardal2:
This sounds like an issue with persistent volumes, I've run into this a few times, try deleting the old PVCs and then reinstalling.

andresbono:
It might be the cause of the issue, @kristinnardal2. Thanks for pointing out.

maxisam:
It seems like it can be solved by force boot

https://github.com/helm/charts/pull/14149

This is a very annoying issue. It shows up every now and then.

#3974 #3697  #3074  #2868

javsalgar:
Hi,

Looking at the cases, I see that several times the persistent volumes are the cause of the issue. We test the persistent storage providers of major Kubernetes distributions and we do not see that issue. Could you provide more details about the persisent volumes you are using?

maxisam:
I use longhorn but I don't think that is the problem. It is kinda odd you can't reproduce it because it is a common issue on both helm chart repo. Did you try to shut down the nodes without draining them first? I don't have this issue until I accidently shut down the nodes.

javsalgar:
By shutting down the nodes, you mean the k8s nodes or the rabbitmq nodes?

maxisam:
> By shutting down the nodes, you mean the k8s nodes or the rabbitmq nodes?

K8S. like unexpected shut down

javsalgar:
Hi,

I've been unable to reproduce the issue. Which is your Kubernetes setup so we can try mimicking it.

maxisam:
Well, it is an issue existing in 2 repos and multiple people reported it already. So I am pretty sure it is there. 

I used Rancher to create the cluster and use Longhorn for storage. I have 3 rabbit nodes with HA policy

```json
        "policies": [
          {
            "vhost": "/",
            "name": "HA",
[...]

PR #4598, opened 2020-12-03
[bitnami/rabbitmq] Update readinessProbe to avoid deadlocks

javsalgar:
**Description of the change**

Change readiness probe command as it was causing issues

**Benefits**

**Possible drawbacks**

**Applicable issues**

  - fixes https://github.com/bitnami/charts/issues/3783

**Additional information**

**Checklist** 


mltsy:
This looks right to me, although I have not tested this solution (so this is just from knowing the issue and looking at the documentation)

dani8art (approved):
LGTM, please resolve conflict in `Chart.yaml`

mltsy:
One thing that concerns me about this is what happens in the case of a rolling upgrade of RabbitMQ version.  I guess I don't even know if the chart supports rolling upgrade, or what it does if you try to upgrade the rabbit version!  So I will open a new issue for this.  However, in any case, this setting would *not* support a rolling upgrade with no downtime, since it doesn't wait for the sync to complete before moving on to the next node (which you would want in the case of a rolling upgrade - I think there may be some features needed around specifying an upgrade strategy, that might trigger different rabbitmq-specific actions/checks during the upgrade process)

mltsy:
Update: it looks like a rolling upgrade is the default for this chart, which I *think* means this PR changes the default from one that is safe (waits for mirrors to sync) to one that is unsafe, in the case of durable mirrored queues.  Right?

Example: If you have a classic mirrored queue, and perform a `helm upgrade` it will start a rolling upgrade that doesn't wait for mirrors to re-sync after restarted, and therefore could cause downtime when it finally shuts down the master node for the mirrored queue (assuming none of the mirrors are synced yet) - and could actually result in data-loss if `ha-promote-on-shutdown` is set to `always`.

I think there needs to be a new parameter for this called something like `upgradeOptions.waitForSync` that can be true or false, so that if you're spinning up a cluster.

Although, now that I'm thinking about it, the way it's supposed to work is that the first node to start up is the last node to shut down (and therefore the master node, which doesn't wait for other pods in order to start working).  In the case of this chart, that's always supposed to be rabbitmq-0.  However, in the case of a total cluster failure or something, the chart doesn't know which node was the master.  So I guess the above suggestion (`upgradeOptions.waitForSync`) might still be a useful option, but it should default to true for regular upgrade situations, rather than false as this PR does.

dani8art:
I see your point, so you mean to have two types of probes, one will wait for mirrors and the other one will just ping the node to see if it is working, and then a new property `upgradeOptions.waitForSync` that will decide which one to use.

On the other hand, I think the best default would `true` as you have argued.

Thanks!!

bitnami/rabbitmq/templates/statefulset.yaml
```diff
@@ -228,7 +228,7 @@ spec:
               command:
                 - /bin/bash
                 - -ec
-                - rabbitmq-diagnostics -q check_running
+                - rabbitmq-diagnostics -q ping
```
dani8art, line 231: Add wait for sync options

dani8art (commented):
Please, let's try to integrate the latest suggestions

mltsy:
Wow!  Actually, this issue addresses the syncing issue in a much better way, I think: https://github.com/bitnami/charts/issues/3931

Which means, this PR is probably fine.  The livenessProbe shouldn't have to worry about any kind of syncing, as long as the sync-wait is handled on pod shutdown (where it should really be handled anyway).  Right?  As long as the pod is allowed the time it needs to shutdown!

dani8art:
Waiting for this https://github.com/bitnami/bitnami-docker-rabbitmq/pull/142 to be solved

PR #4629, opened 2020-12-07
[bitnami/mongodb-sharded] Changed mongos kind to statefulset

### LINKED ISSUE #4499: [mongodb-sharded] mongos as statefulset
**Which chart**:
mongodb-sharded

When using more than 1 replicas of mongos and multiple connections via the service we get cursor not found exception. It happens because queries go to different mongos instances. Especially it happens with Spring Framework and Spring Data.

The only way to solve the issue is to pass all mongos uris in connection string. But with deployment kind it's now possible because pod name always different. Would be better to convert mongos to StatefulSet so we can use static routes to mongos like `mongos-0.infra-mongodb-sharded.default.svc.cluster.local`

We tried to set `sessionAffinity: "ClientIP"` for the service but it doesn't solve the issue at all. Sometimes we still get the error.

alexsergin:
**Description of the change**

Changed mongos kind from deployment to Statefulset

**Benefits**

Known addresses of mongos pods allows to pass all of it in connection string when it's impossible to use dynamic addresses. Like with spring data framework.

**Possible drawbacks**

Dunno. I guess it should work same as deployment

**Applicable issues**

  - fixes #4499

**Additional information**

**Checklist**
- [X] Chart version bumped


juan131 (changes requested):
This change will introduce a backwards incompatibility and users won't be able to upgrade their existing releases.

Therefore, in order to accept accept a change like this, it's necessary to:

1. Bump chart major version
1. Document upgrade path in the **README.md**

As an alternative, we can support both deploy Mongos as a deployment a statefulset by adding a new parameter that allows deciding what setup to use.

---

That said, before going on with any of the alternatives, I'd like to understand the reasons why you require Mongos to be deployed using a statefulset. Could you please detail them? Thanks in advance.

alexsergin:
@juan131 I think will be better to support both types of resources to let old users keep staying with deployment and get future upgrades for other components. What do you think?

juan131:
It's fine to me @alexsergin 
We already do sth similar on other charts (e.g. MongoDB - not the sharded one), see:

- https://github.com/bitnami/charts/blob/master/bitnami/mongodb/templates/standalone/dep-sts.yaml

As you can see there's a `useStatefulSet` parameter than you can set to use a StatefulSet instead of a Deployment.

alexsergin:
I've made all changes @juan131. Added `useStatefulSet` parameter for mongos like you've done in mongodb chart.

bitnami/mongodb-sharded/values.yaml
```diff
@@ -741,6 +745,9 @@ mongos:
   ## ref: https://kubernetes.io/docs/tasks/administer-cluster/configure-multiple-schedulers/
   ##
   schedulerName:
+  ## Use StatefulSet instead of Deployment when deploying standalone
```
juan131, line 748: There's no "standalone" mode for MongoDB Sharded. I guess you took this from "MongoDB" chart.

juan131 (changes requested):
Please rebump the version in the Chart.yaml since other PRs were merged bumping it.

bitnami/mongodb-sharded/values.yaml
alexsergin: Thanks. Fixed

alexsergin:
Fixed all and bumped version @juan131

juan131 (approved):
LGTM! Thanks for addressing the suggestions!

maxisam:
@alexsergin I thought we can use headless service for deployment?

alexsergin:
@maxisam You can but you can't access directly to the pod through this service anyway because deployment pods have random names.

maxisam:
@alexsergin  You mean you wanna connect to specific mongos pod?

alexsergin:
@maxisam Yep!

PR #4647, opened 2020-12-08
[bitnami/spark] Fix Prometheus metrics pod annotations

juan131:

**Description of the change**

This PR fixes the Prometheus metrics pod annotations. The metrics endpoint is actually at `/metrics/` instead of `metrics` (the final `/` is required). Therefore, we need to adapt the annotations accordingly.

This PR also removes some parameters that were not used (`metrics.service.XXX`) that are probably a result of copy-pasting from other charts.

**Benefits**

Prometheus can detect the metrics.

**Possible drawbacks**

None

**Checklist**


marcosbc (approved):
LGTM!

PR #4657, opened 2020-12-09
[bitnami/grafana] Change health check path

skinnaj:
**Description of the change**

This change changes the readiness and liveness probe of the deployment to the Grafana health route path.

**Benefits**

When deploying the chart with an ingress, the backends remain in an unhealthy state as the / route, which is currently used for the health checks, returns a 302 status (redirects to /login). When changing the readiness and liveness probe to /api/health the ingress health check gets a 200 status when the instance is healthy.

**Possible drawbacks**

None.

**Checklist**

- [x] Chart version bumped in Chart.yaml according to semver.
- [x]  Variables are documented in the README.md
- [x]  Title of the PR starts with chart name (e.g. [bitnami/chart])
- [x]  If the chart contains a values-production.yaml apart from values.yaml, ensure that you implement the changes in both files

### DIFF
--- a/bitnami/grafana/Chart.lock
+++ b/bitnami/grafana/Chart.lock
@@ -1,6 +1,6 @@
 dependencies:
 - name: common
   repository: https://charts.bitnami.com/bitnami
-  version: 1.0.1
-digest: sha256:e66388f254b9de6470bca6d8c1c565d1c16a5569beef68a7bc99e486e73ccbdb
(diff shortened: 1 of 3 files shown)


juan131 (approved):
LGTM!!
I think we were indeed using `/api/health` in the past as default value for probes in the past. We probably removed that accidentally in some commit.

PR #4687, opened 2020-12-10
[bitnami/mongodb] add existingSecret for tls certificates

benjamin-wright:
**Description of the change**

Adding a `tls.existingSecret` option to allow existing certificates to be passed in to the mongo cluster

**Benefits**

Easier to use custom certs, rather than relying on the certs generated in the helm chart

**Possible drawbacks**

Backwards compatibile, so no obvious issues beyond the minor increase in complexity

**Applicable issues**

NA - can raise one if there's some discussion needed?

**Additional information**

NA

**Checklist** 


deveshk0:
@benjamin-wright Sorry didn't see your PR, I have added a similar one to use custom ca

benjamin-wright:
> @benjamin-wright Sorry didn't see your PR, I have added a similar one to use custom ca https://github.com/bitnami/charts/pull/4690

@deveshk0 No worries, great minds think alike! Both look like orthogonal improvements I think, the existing secret approach is useful for me with sealed secrets, but i can definitely see the benefit of yours for terraform / helm install / etc.

bitnami/mongodb/templates/replicaset/statefulset.yaml
```diff
@@ -450,7 +450,7 @@ spec:
           emptyDir: {}
         - name: certs-volume
           secret:
-            secretName: ca-clients
+            secretName: {{ default "ca-clients" .Values.tls.existingSecret }}
```
juan131, line 453: I just noticed the default secret is hardcoded to "ca-clients", which is clearly wrong. We should use sth like:'
```
  name: {{ include "mongodb.fullname" . }}-metrics
```

Could you please fix it both here and the **secrets-ca.yaml** file (line 9)?

bitnami/mongodb/Chart.yaml
```diff
@@ -26,4 +26,4 @@ name: mongodb
 sources:
   - https://github.com/bitnami/bitnami-docker-mongodb
   - https://mongodb.org
-version: 10.1.4
+version: 10.1.5
```
juan131, line 29: Please bump minor version instead since you're adding a new parameter.

juan131 (changes requested):
Thanks for the contribution!! 

I agree with @benjamin-wright that both this and the solution provided @deveshk0 are valid ones. They are different approaches but I don't see any problem on supporting both.

Please review my comments when you have a chance.

bitnami/mongodb/Chart.yaml
benjamin-wright: Done

bitnami/mongodb/templates/replicaset/statefulset.yaml
benjamin-wright: @juan131 updated, does this look ok? Pulling it out into a secret seemed like the easiest solution

juan131 (approved):
Great! Your PR looks perfect now! Approving it

Issue #4681, opened 2020-12-10
nodeAffinityPreset generates invalid template


ondrej-smola:
**Which chart**:
bitnami/etcd 5.3.0

**Describe the bug**
When nodeAffinityPreset is defined in values.yml 
```
nodeAffinityPreset:
  type: hard
  key: "node-role.ntx.io"
  values: ["etcd"]
```
installation fails with 
```
helm install -f k8s/helm/etcd.yml etcd bitnami/etcd

Error: unable to build kubernetes objects from release manifest: error validating "": error validating data: 
ValidationError(StatefulSet.spec.template.spec.affinity.nodeAffinity.requiredDuringSchedulingIgnoredDuringExecution.nodeSelectorTerms[0].matchExpressions): 
invalid type for io.k8s.api.core.v1.NodeSelectorTerm.matchExpressions: got "map", expected "array"
```

**Version of Helm and Kubernetes**:

- Output of `helm version`:

```
version.BuildInfo{Version:"v3.4.1", GitCommit:"c4e74854886b2efe3321e185578e6db9be0a6e29", GitTreeState:"clean", GoVersion:"go1.14.11"}
```

- Output of `kubectl version`:

```
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.4", GitCommit:"d360454c9bcd1634cf4cc52d1867af5491dc9c5f", GitTreeState:"clean", BuildDate:"2020-11-11T13:17:17Z", GoVersion:"go1.15.2", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.4", GitCommit:"d360454c9bcd1634cf4cc52d1867af5491dc9c5f", GitTreeState:"clean", BuildDate:"2020-11-11T13:09:17Z", GoVersion:"go1.15.2", Compiler:"gc", Platform:"linux/amd64"}
```

**Additional context**
Everything works ok when nodeAffinityPreset section is removed

juan131:
Hi @ondrej-smola 

I was able to reproduce the issue:

```console
$ helm template -f values-etcd.yaml etcd bitnami/etcd -s templates/statefulset.yaml | kubeval
[...]

PR #4692, opened 2020-12-11
[bitnami/common] Fix node affinity templates

juan131:

**Description of the change**

This PR fixes the node affinity templates by ensuring the "matchExpress" object contains an array of objects, instead of an object.

```console
$ k explain sts.spec.template.spec.affinity.nodeAffinity.preferredDuringSchedulingIgnoredDuringExecution.preference.matchExpressions
KIND:     StatefulSet
VERSION:  apps/v1

RESOURCE: matchExpressions <[]Object>

DESCRIPTION:
     A list of node selector requirements by node's labels.

     A node selector requirement is a selector that contains values, a key, and
     an operator that relates the key and values.

FIELDS:
   key	<string> -required-
     The label key that the selector applies to.

   operator	<string> -required-
     Represents a key's relationship to a set of values. Valid operators are In,
     NotIn, Exists, DoesNotExist. Gt, and Lt.

   values	<[]string>
     An array of string values. If the operator is In or NotIn, the values array
     must be non-empty. If the operator is Exists or DoesNotExist, the values
     array must be empty. If the operator is Gt or Lt, the values array must
     have a single element, which will be interpreted as an integer. This array
     is replaced during a strategic merge patch.
```

**Benefits**

Users can set NodeAffinity with presets.

**Possible drawbacks**

None

**Applicable issues**

  - fixes https://github.com/bitnami/charts/issues/4681

**Checklist**

### DIFF
--- a/bitnami/common/Chart.yaml
+++ b/bitnami/common/Chart.yaml
@@ -20,4 +20,4 @@ sources:
   - https://github.com/bitnami/charts
   - http://www.bitnami.com/
 type: library
-version: 1.1.1
+version: 1.1.2
--- a/bitnami/common/templates/_affinities.tpl
+++ b/bitnami/common/templates/_affinities.tpl
@@ -8,12 +8,12 @@ Return a soft nodeAffinity definition
 preferredDuringSchedulingIgnoredDuringExecution:
   - preference:
       matchExpressions:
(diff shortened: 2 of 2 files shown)


miguelaeh (approved):
LGTM!

Issue #4618, opened 2020-12-05
[bitnami/dokuwiki] Multiple ingress hosts not working


landakram:
**Which chart**:
bitnami/dokuwiki
latest

**Describe the bug**

When attempting to use multiple ingress hosts, only the last ingress is created. It appears to be the same problem as https://github.com/bitnami/bitnami-docker-odoo/issues/74#issuecomment-525624494, given that the templates are similar. 

**To Reproduce**

(I am using helm through ansible). Add the following parameters:

```yml
      ingress:
        enabled: true
        hosts:
        - name: "dokuwiki.{{ ingress_server_ip }}.nip.io"
        - name: "dokuwiki.{{ some_other_ip }}.nip.io"
```

After running helm, only one ingress rule is created for `dokuwiki.{{ some_other_ip }}.nip.io`.

**Expected behavior**

Two ingress rules total are created for `dokuwiki.{{ ingress_server_ip }}.nip.io` and `dokuwiki.{{ some_other_ip }}.nip.io`.

**Version of Helm and Kubernetes**:

- Output of `helm version`:

```
version.BuildInfo{Version:"v3.2.1", GitCommit:"fe51cd1e31e6a202cba7dead9552a6d418ded79a", GitTreeState:"clean", GoVersion:"go1.13.10"}
```

- Output of `kubectl version`:

```
Client Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.5+k3s1", GitCommit:"58ebdb2a2ec5318ca40649eb7bd31679cb679f71", GitTreeState:"clean", BuildDate:"2020-05-06T23:43:23Z", GoVersion:"go1.13.8", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.5+k3s1", GitCommit:"58ebdb2a2ec5318ca40649eb7bd31679cb679f71", GitTreeState:"clean", BuildDate:"2020-05-06T23:43:23Z", GoVersion:"go1.13.8", Compiler:"gc", Platform:"linux/amd64"}
```

**Additional context**
N/A

marcosbc:
Hi @landakram, it looks like the `ingress.yaml` template for DokuWiki is wrong as it's appending resource templates with the same `name`, causing the last one to be the only one enabled.

[...]

PR #4695, opened 2020-12-11
[bitnami/owncloud] New major version

juan131:

**Description of the change**

This PR refactors the whole chart to follow Helm chart best practices, and move the ownCloud image to a "non-root" approach.

This includes several breaking changes, so please review it carefully, putting special emphasis to the "Upgrading" section describing all of them.

**Benefits**

- Standardization.
- Reduces technical debt.

**Possible drawbacks**

It breaks backwards compatibility.

**Applicable issues**

None

**Additional information**

-

**Checklist**

### DIFF
--- a/bitnami/owncloud/templates/deployment.yaml
+++ b/bitnami/owncloud/templates/deployment.yaml
@@ -3,6 +3,7 @@ apiVersion: {{ include "common.capabilities.deployment.apiVersion" . }}
 kind: Deployment
 metadata:
   name: {{ template "common.names.fullname" . }}
+  namespace: {{ .Release.Namespace }}
   labels: {{- include "common.labels.standard" . | nindent 4 }}
     {{- if .Values.commonLabels }}
     {{- include "common.tplvalues.render" ( dict "value" .Values.commonLabels "context" $ ) | nindent 4 }}
--- a/bitnami/owncloud/templates/deployment.yaml
+++ b/bitnami/owncloud/templates/deployment.yaml
@@ -66,6 +66,8 @@ spec:
           args: {{- include "common.tplvalues.render" (dict "value" .Values.args "context" $) | nindent 12 }}
           {{- end }}
           env:
+            - name: BITNAMI_DEBUG
+              value: {{ ternary "true" "false" .Values.image.debug | quote }}
             - name: ALLOW_EMPTY_PASSWORD
               value: {{ .Values.allowEmptyPassword | quote }}
             - name: MARIADB_HOST
--- a/bitnami/owncloud/templates/deployment.yaml
+++ b/bitnami/owncloud/templates/deployment.yaml
@@ -1,143 +1,175 @@
 {{- if include "owncloud.host" . -}}
-apiVersion: {{ template "owncloud.deployment.apiVersion" . }}
+apiVersion: {{ include "common.capabilities.deployment.apiVersion" . }}
 kind: Deployment
 metadata:
-  name: {{ template "owncloud.fullname" . }}
-  labels:
-    app.kubernetes.io/name: {{ include "owncloud.fullname" . }}
-    helm.sh/chart: {{ include "owncloud.chart" . }}
-    app.kubernetes.io/instance: {{ .Release.Name | quote }}
-    app.kubernetes.io/managed-by: {{ .Release.Service | quote }}
+  name: {{ template "common.names.fullname" . }}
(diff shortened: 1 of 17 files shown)


bitnami/owncloud/templates/tls-secrets.yaml
```diff
@@ -0,0 +1,43 @@
+{{- if .Values.ingress.enabled }}
```
marcosbc, line 1: I think this won't work for multiple hosts, since multiple ingress entries with the same name would be created (and only the last would be taken into account).

A user reported an issue with this format, see #4618 for the issue and #4707 for the fix. If you copy it from a chart like Drupal it would work.

bitnami/owncloud/templates/deployment.yaml
```diff
@@ -1,143 +1,176 @@
...
+          {{- if .Values.containerSecurityContext.enabled }}
+          securityContext: {{- omit .Values.containerSecurityContext "enabled" | toYaml | nindent 12 }}
+          {{- end }}
+          {{- if .Values.command }}
+          command: {{- include "common.tplvalues.render" (dict "value" .Values.command "context" $) | nindent 12 }}
+          {{- end }}
+          {{- if .Values.args }}
+          args: {{- include "common.tplvalues.render" (dict "value" .Values.args "context" $) | nindent 12 }}
+          {{- end }}
+          env:
```
marcosbc, line 68: I'm missing BITNAMI_DEBUG here when image.debug is set, could you add it? It's pretty useful sometimes.

bitnami/owncloud/templates/tls-secrets.yaml
juan131: I don't think issue applies to this case @marcosbc 
See this example:
```console
$ helm template owncloud bitnami/owncloud \
    --set ingress.enabled=true \
    # Main domain
    --set ingress.tls=true --set ingress.hostname=main.domain \
    # Extra Domains
    --set ingress.extraHosts[0].name=second.domain --set ingress.extraHosts[0].path=/ \
    --set ingress.extraTls[0].hosts[0]=second.domain --set ingress.extraTls[0].secretName=second.domain-tls \
    --set ingress.secrets[0].name=second.domain-tls --set ingress.secrets[0].key="PRIVATE KEY ..." --set ingress.secrets[0].certificate="BEGIN ..." \
    -s templates/ingress.yaml -s templates/tls-secrets.yaml

---
# Source: owncloud/templates/ingress.yaml
apiVersion: networking.k8s.io/v1beta1
kind: Ingress
metadata:
  name: owncloud
  namespace: default
  labels:
    app.kubernetes.io/name: owncloud
    helm.sh/chart: owncloud-10.0.0
    app.kubernetes.io/instance: owncloud
spec:
  rules:
    - host: main.domain
      http:
        paths:
          - path: /
            backend:
              serviceName: owncloud
              servicePort: http
    - host: second.domain
      http:
        paths:
          - path: /
            backend:
              serviceName: owncloud
              servicePort: http
  tls:
    - hosts:
        - main.domain
      secretName: main.domain-tls
    - hosts:
        - second.domain
      secretName: second-secret
---
# Source: owncloud/templates/tls-secrets.yaml
apiVersion: v1
kind: Secret
metadata:
  name: second-secret
  namespace: default
  labels:
    app.kubernetes.io/name: owncloud
    helm.sh/chart: owncloud-10.0.0
    app.kubernetes.io/instance: owncloud
type: kubernetes.io/tls
data:
  tls.crt: "..."
  tls.key: "..."
# Source: owncloud/templates/tls-secrets.yaml
apiVersion: v1
kind: Secret
metadata:
  name: main.domain-tls
  namespace: default
  labels:
    app.kubernetes.io/name: owncloud
    helm.sh/chart: owncloud-10.0.0
    app.kubernetes.io/instance: owncloud
type: kubernetes.io/tls
data:
  tls.crt: "..."
  tls.key: "..."
  ca.crt: "..."
```
As you can see, I'm able to create a default ingress rule + tls configuration for "main" doman, and add an extra one for a second domain.

marcosbc: Ok then, in that case my only concern is regarding consistency, since they are both PHP applications and TLS secrets config should be the same for both. If this is the "correct" approach then we should probably fix the others as well.

marcosbc:
Solved via #4849

PR #4683, opened 2020-12-10
bitnami/kafka add minBrokerId variable

antonmatsiuk:
**Description of the change**

**Benefits**
As described [here](https://github.com/bitnami/charts/issues/4626#issuecomment-740569976) The PR adds `minBrokerId` variable to kafka chart to support stretched clusters in zero-downtime migration scenarios. This allows to set an arbitrary minBrokerId and consequently an arbitrary minimal `broker.id` of kafka brokers in the cluster. This allows to combine brokers from a helm release with other brokers or releases into a single kafka cluster. 

**Possible drawbacks**
no

**Applicable issues**
no 

  - fixes #4626

**Additional information**

**Checklist** 
- [x] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [x] Variables are documented in the README.md
- [x] Title of the PR starts with chart name (e.g. `[bitnami/chart]`)
- [x] If the chart contains a `values-production.yaml` apart from `values.yaml`, ensure that you implement the changes in both files

PR #4700, opened 2020-12-11
bitnami/zookeeper add minServerId variable

### LINKED ISSUE #4626: bitnami/kafka no possibility to set an arbitrary KAFKA_CFG_MIN_BROKER_ID
**Which chart**:
bitnami/kafka

In current kafka chart it is not possible to set an arbitrary `KAFKA_CFG_MIN_BROKER_ID` for a stateful set with multiple pods. The `KAFKA_CFG_BROKER_ID` value is hardcoded via `/scripts/setup.sh` to the ID of the pod in the stateful set. This makes impossible to provision a stretched cluster where two separate helm releases form a single kafka cluster, because BROKER_IDs in the stateful set always start from zero. Such stretched clusters are absolutely needed in zero-downtime migration scenarios.
 
A possibility to set `KAFKA_CFG_MIN_BROKER_ID` variable in the chart. Brokers derive their IDs from the following formula: `KAFKA_CFG_BROKER_ID = KAFKA_CFG_MIN_BROKER_ID + POD_ID`, e.g. if KAFKA_CFG_MIN_BROKER_ID=10, brokers get ID=10,11,12 etc.
If `KAFKA_CFG_MIN_BROKER_ID`variable is set in the chart `setup.sh` script [here](https://github.com/bitnami/charts/blob/master/bitnami/kafka/templates/scripts-configmap.yaml) takes it into account while setting `KAFKA_CFG_BROKER_ID`.

A current alternative is to deactivate `/scripts/setup.sh` in the chart completely, set `KAFKA_CFG_BROKER_ID` in the chart manually and provision a separate stateful set per broker, what leads to huge management overhead.

antonmatsiuk:
**Description of the change**

Similar to the #4626 and #4683 the PR adds `minServerId` variable zookeeper chart to support stretched clusters in zero-downtime migration scenarios. This allows to set an arbitrary `minServerId` and consequently an arbitrary minimal `ZOO_SERVER_ID` (myid) of zookeeper servers in the cluster. This allows to combine zookeeper instances from a helm release with other servers or releases into a single cluster. The PR respects `myid` stored in the persistent volume (if present) and does not overwrite it.

**Benefits**
`ZOO_SERVER_ID` values are decoupled from the pod IDs in a StatefulSet

**Possible drawbacks**
not known

**Applicable issues**
  - partially fixes #4626

**Additional information**

There is an open [issue](https://github.com/bitnami/bitnami-docker-zookeeper/issues/45) with `bitnami-docker-zookeeper` which does not allow to set `SERVER_ID > 9` in `ZOO_SERVERS` variable in the new format (containing IDs), this must be considered in testing. Until the issue is solved, a testing scenario works with e.g. `minServerId: 5` and `replicaCount: 3` 

**Checklist** 


bitnami/zookeeper/Chart.yaml
```diff
@@ -15,4 +15,4 @@ name: zookeeper
 sources:
   - https://github.com/bitnami/bitnami-docker-zookeeper
   - https://zookeeper.apache.org/
-version: 6.0.1
+version: 6.0.2
```
andresbono, line 18: As this includes a new parameter, please bump the minor version.

```suggestion
version: 6.1.0
```

bitnami/zookeeper/templates/statefulset.yaml
```diff
@@ -99,14 +99,20 @@ spec:
             - bash
             - -ec
             - |
-                # Execute entrypoint as usual after obtaining ZOO_SERVER_ID based on POD hostname
-                HOSTNAME=`hostname -s`
-                if [[ $HOSTNAME =~ (.*)-([0-9]+)$ ]]; then
-                  ORD=${BASH_REMATCH[2]}
-                  export ZOO_SERVER_ID=$((ORD+1))
+                # Execute entrypoint as usual after obtaining ZOO_SERVER_ID=
```
andresbono, line 102: Is this a typo?

```suggestion
                # Execute entrypoint as usual after obtaining ZOO_SERVER_ID
```

andresbono (commented):
Please check my comments.

andresbono:
Thank you very much for your contribution! A fix for https://github.com/bitnami/bitnami-docker-zookeeper/issues/45 is on the way.

bitnami/zookeeper/templates/statefulset.yaml
```diff
@@ -137,13 +143,14 @@ spec:
...
               {{- $replicaCount := int .Values.replicaCount }}
+              {{- $minServerId := int .Values.minServerId }}
               {{- $followerPort := int .Values.service.followerPort }}
               {{- $electionPort := int .Values.service.electionPort }}
               {{- $releaseNamespace := include "zookeeper.namespace" . }}
               {{- $zookeeperFullname := include "zookeeper.fullname" . }}
               {{- $zookeeperHeadlessServiceName := printf "%s-%s" $zookeeperFullname "headless" | trunc 63  }}
               {{- $clusterDomain := .Values.clusterDomain }}
-              value: {{ range $i, $e := until $replicaCount }}{{ $zookeeperFullname }}-{{ $e }}.{{ $zookeeperHeadlessServiceName }}.{{ $releaseNamespace }}.svc.{{ $clusterDomain }}:{{ $followerPort }}:{{ $electionPort }} {{ end }}
+              value: {{ range $i, $e := until $replicaCount }}{{ $zookeeperFullname }}-{{ $e }}.{{ $zookeeperHeadlessServiceName }}.{{ $releaseNamespace }}.svc.{{ $clusterDomain }}:{{ $followerPort }}:{{ $electionPort }}::{{ add $e $minServerId }} {{ end }}
```
juan131, line 153: This will be conflictive with the workaround to respect existing ids, right? I mean.. This will assign an id based on `{{ .Values.minServerId }}` regardless there was a previous set at `/bitnami/zookeeper/data/myid`. Therefore, the conflict.
Wouldn't be easier to set `ZOO_SERVERS` without including the id?

juan131 (changes requested):

bitnami/zookeeper/templates/statefulset.yaml
antonmatsiuk: @juan131 
> Wouldn't be easier to set ZOO_SERVERS without including the id?

This unfortunately doesn't work, if `ZOO_SERVERS` are specified without IDs, and IDs are non-standard (0,1,2,..), then  an exception is thrown: 
```
java.lang.RuntimeException: My id 5 not in the peer list	at org.apache.zookeeper.server.quorum.QuorumPeer.start(QuorumPeer.java:1076)	at org.apache.zookeeper.server.quorum.QuorumPeerMain.runFromConfig(QuorumPeerMain.java:227)	at org.apache.zookeeper.server.quorum.QuorumPeerMain.initializeAndRun(QuorumPeerMain.java:136)	at org.apache.zookeeper.server.quorum.QuorumPeerMain.main(QuorumPeerMain.java:90)2020-12-15 11:05:37,258 [myid:5] - ERROR [main:ServiceUtils@42] - Exiting JVM with code 1
```
Alternatively, one could override non-standard `ZOO_SERVERS` via custom `zoo.cfg` or introduce another custom variable which will override standard `ZOO_SERVERS`, This will be anyway required for a stretched cluster scenario to include zookeeper instances from outside the current helm release.

juan131: You can try to define/export `ZOO_SERVERS` in the same script where `ZOO_SERVER_ID` is being defined and apply similar logic to the one you already implemented.
If you do so, you can remove it from the "env" section.

antonmatsiuk: @juan131 setting `ZOO_SERVERS` variable while respecting `myid` from an existing PV will require querying not only local `myid` but also `myid` values of other instances in a StatefulSet, which may be completely uncorrelated to the local one. I don't see an obvious way to do it. To overcome the limitation one can  set `ZOO_SERVERS` variable manually, if `myids` from existing PVs are known and must be preserved.

juan131: You're right.. We'll have to implement dirty workarounds to fix this... I agree with you, let's keep this simple

juan131 (approved):
LGTM

PR #4715, opened 2020-12-13
[bitnami/mongodb] allow selector definitions for volumeClaimTemplates

MatthiasLohr:
**Allow selector definitions for volumeClaimTemplates when a StatefulSet is used**

With this change, chart users can define `selector` configuration for the StatefulSet's `volumeClaimTemplates` definition.

**Benefits**

The change gives more freedom to the users using this chart. Allowing for defining the `selector` in `volumeClaimTemplates` allows for example the re-use of existing `PersistentVolume`s, e.g. when using local volumes.

**Possible drawbacks**

Nothing known.

**Applicable issues**

Nothing known.

**Checklist**

### DIFF
--- a/bitnami/mongodb/Chart.yaml
+++ b/bitnami/mongodb/Chart.yaml
@@ -26,4 +26,4 @@ name: mongodb
 sources:
   - https://github.com/bitnami/bitnami-docker-mongodb
   - https://mongodb.org
-version: 10.2.0
+version: 10.3.0
--- a/bitnami/mongodb/README.md
+++ b/bitnami/mongodb/README.md
@@ -245,13 +245,14 @@ The following tables lists the configurable parameters of the MongoDB chart and 
 
 | Parameter                                 | Description                                                                                                | Default                                                 |
 |-------------------------------------------|------------------------------------------------------------------------------------------------------------|---------------------------------------------------------|
-| `persistence.enabled`                     | Enable MongoDB data persistence using PVC                                                                  | `true`                                                  |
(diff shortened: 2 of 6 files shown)


miguelaeh (approved):
LGTM!

MatthiasLohr:
@miguelaeh, thank you for merging so fast. However, unfortunately, the changes are not online yet since another pull request, also changing the version number to 10.3.0, was merged first. So I created a new pull request just for bumping the version to 10.3.1: #4720. Would be nice if you could take a look there, thanks!

miguelaeh:
Hi @MatthiasLohr ,
Thank you for checking this. I am not sure why we didn't get any conflicts when merging the changes but we have faced this before. I will launch the release jobs manually so the pipeline will update the tags on the chart and bump the version by itself.
Thank you very much for checking this.

MatthiasLohr:
The reason is quite simple: A very isolated change (single line) with the same target content, so it's easy to merge automatically. There would only be a conflict if the two versions to be merged were different.

Thanks for helping!

miguelaeh:
Hi @MatthiasLohr ,
Thanks for that clarification, it has sense.
Also, the chart was published again 19 hours ago, your changes should be included in that new version.
You will need to execute:
```
helm repo update
```

MatthiasLohr:
Yes, in between it started to work, could use the new chart since yesterday evening (CET). Eventually, the Chart version number is increased now to 10.3.1.

miguelaeh:
yes, the pipeline increases the version in every release if it is not updated on GH, that's why I closed the PR.

PR #4478, opened 2020-11-24
[bitnami/redis] wait for new master when using sentinel and the master pod has been restarted/destroyed

dustinrue:
**Description of the change**

In situations where the master pod is destroyed it may be restarted before the remaining pods have elected a new master. This will cause the new pod to never discover the new master because it doesn't know what to connect too. This change modifies the start-sentinel.sh script so that when a pod starts it attempts to connect to the reported master. If the master doesn't actually exist then it will cause the sentinel container to timeout and be restarted. While waiting for this timeout the remaining pods should have selected a new master. If not (because the user as spec'd a longer timeout perhaps) then the container will continue to restart until a master has been selected.

**Benefits**

Resolves the final issue noted in #3700 

**Possible drawbacks**

I am not able to test this change in all scenarios. I was able to test a fresh install as well as deletion of the current master pod. In both cases the cluster was able to reform properly.

**Applicable issues**

  - fixes #3700 

**Additional information**

**Checklist** 
- [X] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [ ] Variables are documented in the README.md
- [ ] Title of the PR starts with chart name (e.g. `[bitnami/chart]`)
- [ ] If the chart contains a `values-production.yaml` apart from `values.yaml`, ensure that you implement the changes in both files

:warning: Keep in mind that if you want to make changes to the kubeapps chart, please implement them in the [kubeapps repository](https://github.com/kubeapps/kubeapps/tree/master/chart/kubeapps). This is only a synchronized mirror.

Issue #3700, opened 2020-09-17
Redis sentinel is not picking right master after master pod restart


mahanam:
NAME: redis-master
REVISION: 1
RELEASED: Thu Sep 17 07:37:59 2020
CHART: redis-11.0.0

**Expected behavior**
It should pick the right master after the pod restart.

Helm version:
```
Client: &version.Version{SemVer:"v2.16.1", GitCommit:"bbdfe5e7803a12bbdf97e94cd847859890cf4050", GitTreeState:"clean"}
Server: &version.Version{SemVer:"v2.16.1", GitCommit:"bbdfe5e7803a12bbdf97e94cd847859890cf4050", GitTreeState:"clean"}
```
kubectl version
```
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.3", GitCommit:"b3cbbae08ec52a7fc73d334838e18d17e8512749", GitTreeState:"clean", BuildDate:"2019-11-13T11:23:11Z", GoVersion:"go1.12.12", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.3", GitCommit:"b3cbbae08ec52a7fc73d334838e18d17e8512749", GitTreeState:"clean", BuildDate:"2019-11-13T11:13:49Z", GoVersion:"go1.12.12", Compiler:"gc", Platform:"linux/amd64"}
```
Redis Sentinel is not picking the right master IP address after the master pod restart in fact it's picking its own old IP address.

1. I have deployed a fresh redis-ha
```
NAME READY STATUS RESTARTS AGE IP NODE NOMINATED NODE READINESS GATES
redis-master-node-0 2/2 Running 0 2m30s 10.233.121.246 localhost
redis-master-node-1 2/2 Running 0 116s 10.233.121.249 localhost
redis-master-node-2 2/2 Running 0 104s 10.233.121.250 localhost
```
2. deleted the master pod 10.233.121.246

3. Now at this point, other slave [10.233.121.250] became a new master as expected

4. but the problem here is the deleted master pod came up with new ip [10.233.121.251] and when i see the logs inside that redis container, its trying for the connect the old master.
```
20:S 17 Sep 2020 07:10:36.234 * Connecting to MASTER 10.233.121.246:6379
20:S 17 Sep 2020 07:10:36.234 * MASTER <-> REPLICA sync started
[...]

PR #4726, opened 2020-12-15
[bitnami/redis] Fix Sentinel Redis with TLS

haljin:
**Description of the change**

Fix the startup of the pods when `sentinel.enabled=true` and `tls.enabled=true`. Removed a duplicate variable that breaks on TLS enabled.

**Benefits**

It actually makes the chart work with TLS.

**Possible drawbacks**

**Applicable issues**

  - fixes #

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/redis/Chart.yaml
+++ b/bitnami/redis/Chart.yaml
@@ -19,4 +19,4 @@ name: redis
 sources:
   - https://github.com/bitnami/bitnami-docker-redis
   - http://redis.io/
-version: 12.2.2
+version: 12.2.3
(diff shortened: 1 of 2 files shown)


marcosbc (changes requested):
Hi, thanks for contributing to our repo.

Those changes were implemented in #4478 (issue #3700) fixing a major issue when restarting a pod.

If these lines are causing TLS to break, then it's probably happening because we didn't think of TLS support in #4478. It is probably failing because of not specifying SSL options, so if you could adapt the existing logic to also support SSL, that would be awesome. Thanks!

haljin:
Ooops, I was convinced I did that, didn't notice the tiny little difference between the command I removed and the one before. Should be good now!

haljin:
Also tested that with our TLS deployment and it's good :+1:

marcosbc (approved):
Awesome! LGTM now, thanks!!!

PR #4728, opened 2020-12-15
[bitnami/kafka] Bring Kafka chart more in line with other charts with volumePermissions

hierynomus:
**Description of the change**
- Added a `volumePermissions.securityContext` to the volumePermissions init container
- Moved the `persistence` and `logPersistence` mountPath to a value instead of hardcoded in statefulset.
- Added a specific `podSecurityContext.enabled` flag. If you use the kafka chart as a Helm dependency it is impossible to unset the `fsGroup` and `runAsUser`, which is needed for an OpenShift deployment.

**Benefits**
The chart is now more in line with other charts and more compatible with an OpenShift deployment

**Possible drawbacks**
None

**Applicable issues**
None

  - fixes #

**Additional information**

**Checklist** 


rafariossaa:
Hi,
  Thanks for sending a PR.

rafariossaa (commented):
Hi,
  If an user changes the mount points for persistence and logs to other paths than the default ones, you need to set some environment variables so the image can use the new mount point.
Please take a look to the environment variables and their values set here: https://github.com/bitnami/bitnami-docker-kafka/blob/103223181054edc057bcbf8f8ec14ae1a2225e98/2/debian-10/rootfs/opt/bitnami/scripts/libkafka.sh#L92 .

bitnami/kafka/templates/statefulset.yaml
```diff
@@ -91,23 +91,20 @@ spec:
         - name: volume-permissions
           image: {{ include "kafka.volumePermissions.image" . }}
           imagePullPolicy: {{ .Values.volumePermissions.image.pullPolicy | quote }}
-          command:
```
rafariossaa, line 94: Why do you remove this block ?

hierynomus: not sure, I think that was accidental when copying the changes I made to our clone of the chart. I'll restore them.

hierynomus:
> Hi,
> If an user changes the mount points for persistence and logs to other paths than the default ones, you need to set some environment variables so the image can use the new mount point.
> Please take a look to the environment variables and their values set here: https://github.com/bitnami/bitnami-docker-kafka/blob/103223181054edc057bcbf8f8ec14ae1a2225e98/2/debian-10/rootfs/opt/bitnami/scripts/libkafka.sh#L92 .

Thanks, I'll add those!

bitnami/kafka/templates/statefulset.yaml
```diff
@@ -94,20 +94,19 @@ spec:
           command:
             - /bin/bash
           args:
-            - -ec
+            - -cx
```
rafariossaa, line 97: Sorry, I missed this one.
Could you set it back to `ec`, `cx` is more for debugging purposes.

rafariossaa (changes requested):

rafariossaa:
@juan131 Could you take a look to this one ?

bitnami/kafka/templates/statefulset.yaml
hierynomus: 👍 Done

bitnami/kafka/templates/statefulset.yaml
```diff
@@ -96,18 +96,17 @@ spec:
...
+          securityContext: {{- .Values.volumePermissions.securityContext | toYaml | nindent 12 }}
           {{- if .Values.volumePermissions.resources }}
           resources: {{- toYaml .Values.volumePermissions.resources | nindent 12 }}
           {{- end }}
           volumeMounts:
             - name: data
-              mountPath: /bitnami/kafka
+              mountPath: {{ .Values.persistence.mountPath }}
+            - name: logs
+              mountPath: {{ .Values.logPersistence.mountPath }}
```
juan131, line 109: Nice catch! Actually the current `chown` had no effect since it wasn't sharing the "logs" folder with the main container.

bitnami/kafka/templates/statefulset.yaml
```diff
@@ -96,18 +96,17 @@ spec:
...
             - -ec
             - |
-              mkdir -p /bitnami/kafka
-              chown -R "{{ .Values.podSecurityContext.runAsUser }}:{{ .Values.podSecurityContext.fsGroup }}" "/bitnami/kafka"
-              mkdir -p /opt/bitnami/kafka/logs
-              chown -R "{{ .Values.podSecurityContext.runAsUser }}:{{ .Values.podSecurityContext.fsGroup }}" "/opt/bitnami/kafka/logs"
-          securityContext:
-            runAsUser: 0
+              chown -R "{{ .Values.podSecurityContext.runAsUser }}:{{ .Values.podSecurityContext.fsGroup }}" "{{ .Values.persistence.mountPath }}"
+              chown -R "{{ .Values.podSecurityContext.runAsUser }}:{{ .Values.podSecurityContext.fsGroup }}" "{{ .Values.logPersistence.mountPath }}"
```
juan131, line 100: I wouldn't remove the `mkdir -p` actions. They're harmless and I've found issues in the past with similar init containers despite volumes being mounted on those specific paths.

bitnami/kafka/templates/statefulset.yaml
hierynomus: True indeed

bitnami/kafka/templates/statefulset.yaml
hierynomus: Actually, if the path does not exist, then the volume isn't mounted (correctly), and the `mkdir` would be hiding the fact that the `chown` doesn't do anything. So adding those back in would mean that it potentially hides a problem with the chart

hierynomus:
@juan131 @rafariossaa See my last comment on Juan's remark on the `mkdir`.

Anything else to do here? Or is it ready to be merged?

rafariossaa:
Hi, 
  There was a new chart release, would you mind bumping the version to solve the conflict ?

hierynomus:
@rafariossaa Done, version is now marked as `12.5.0`. But if you want I could also make that `12.4.3`, I don't know what the preference is there. I think this is slightly larger than just a patch bump, so I incremented the minor version.

hierynomus:
Added 1 more change to make it possible to deploy to OpenShift as a dependency from a parent chart.

rafariossaa (approved):

PR #4758, opened 2020-12-17
[bitnami/memcadhed] Memcached persistence is redundant

kfirfer:
Memcached persistence is redundant due to its designed place all data on RAM

### DIFF
--- a/bitnami/memcached/Chart.yaml
+++ b/bitnami/memcached/Chart.yaml
@@ -1,7 +1,7 @@
 annotations:
   category: Infrastructure
 apiVersion: v2
-appVersion: 1.6.9
+appVersion: 1.7.0
 dependencies:
   - name: common
     repository: https://charts.bitnami.com/bitnami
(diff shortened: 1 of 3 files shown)


bitnami/memcached/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 annotations:
   category: Infrastructure
 apiVersion: v2
-appVersion: 1.6.9
+appVersion: 1.7.0
```
javsalgar, line 4: You need to bump the `version` not the `appVersion`

javsalgar (changes requested):
Thanks for the PR! Please check my comment

bitnami/memcached/Chart.yaml
kfirfer: ahh rights! thanks,  I will change it right away

kfirfer:
@javsalgar changed it to version instead appVersion

javsalgar (approved):

PR #4739, opened 2020-12-16
[bitnami/nginx-ingress-controller] Use correct port for ingress-nginx healthchecks

NeonSludge:
**Description of the change**

`ingress-nginx` uses port `10254` to serve the `/healthz` location.
Starting from version `7.0.0` the `bitnami/nginx-ingress-controller` chart specifies a different port for `ingress-nginx`'s liveness and readiness checks in `values.yaml` and `values-production.yaml`. This prevents the `ingress-nginx` pods from passing the checks if the user doesn't change the default setting.

This PR sets the `livenessProbe.httpGet.port` and `readinessProbe.httpGet.port` settings to match the port used by `ingress-nginx`.

**Benefits**

`ingress-nginx` pods pass liveness and readiness checks after installing the `bitnami/nginx-ingress-controller` chart.

**Checklist** 
- [X] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [X] Title of the PR starts with chart name (e.g. `[bitnami/chart]`)
- [X] If the chart contains a `values-production.yaml` apart from `values.yaml`, ensure that you implement the changes in both files

PR #4759, opened 2020-12-17
[bitnami/nginx-ingress-controller] Fix probes port

carrodher:
**Description of the change**

Changes introduced in PR #4739 are not working in the CI/CD because of the following error:
```
Error: Deployment.apps "nginx-ingress-controller-kz7yrdhin0" is invalid: [spec.template.spec.containers[0].livenessProbe.httpGet.port: Invalid value: "10254": must contain at least one letter or number (a-z, 0-9), spec.template.spec.containers[0].readinessProbe.httpGet.port: Invalid value: "10254": must contain at least one letter or number (a-z, 0-9)]
```

**Applicable issues**

  - fixes #4739

**Checklist** 

### DIFF
--- a/bitnami/nginx-ingress-controller/values.yaml
+++ b/bitnami/nginx-ingress-controller/values.yaml
@@ -240,7 +240,7 @@ livenessProbe:
   enabled: true
   httpGet:
     path: /healthz
-    port: "{{ .Values.containerPorts.metrics }}"
+    port: 10254
     scheme: HTTP
   failureThreshold: 3
   initialDelaySeconds: 10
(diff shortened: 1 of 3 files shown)


bitnami/nginx-ingress-controller/values.yaml
```diff
@@ -240,7 +240,7 @@ livenessProbe:
   enabled: true
   httpGet:
     path: /healthz
-    port: "{{ .Values.containerPorts.metrics }}"
+    port: 10254
```
juan131, line 243: Better approach:
```suggestion
port: metrics
```

juan131 (changes requested):

juan131 (approved):
LGTM

bitnami/nginx-ingress-controller/values.yaml
carrodher: It can't be used as `metrics=10254` is only used when `metrics.enabled=true`

Issue #4651, opened 2020-12-08
[kafka] Add topic provisioning support


sergeyshaykhullin:
**Which chart**:
bitnami/kafka

**Is your feature request related to a problem? Please describe.**
I want to provision kafka cluster with predefined topics (and add more topics later)

**Describe the solution you'd like**
Add post install hook with job

```yaml
annotations:
  helm.sh/hook: post-install,post-upgrade
  helm.sh/hook-delete-policy: hook-succeeded,before-hook-creation
```

**Describe alternatives you've considered**
Add configmap with init container?

**Additional context**
Minio implementation of buckets provisioning https://github.com/minio/charts/blob/fc621a0d562075d9100476f4d266d49312ecbf65/minio/values.yaml#L236

```yaml
# bin/kafka-topics.sh --bootstrap-server localhost:9092 --create --topic my-topic --partitions 1 --replication-factor 1 --config max.message.bytes=64000 --config flush.messages=1

topics:
- name: my-topic
  partitions: 1
  replicationFactor: 1
  # https://kafka.apache.org/documentation/#topicconfigs
  config:
    max.message.bytes: 64000
    flush.messages: 1
```

miguelaeh:
Hi @sergeyshaykhullin ,
I think you could do that in both ways, either creating a hook or using an initContainer.
We are missing the InitContainers parameter in this chart, so the above PR adds it. Have in mind that will be executed in every pod restart so maybe you will need to check if those topics already exist before creating them.

sergeyshaykhullin:
I think that helm hook with job is better, because it will run once on new deploy => values changes. It will run even if topics are not changed, but it is better than execute each time pod restarts

sergeyshaykhullin:
I don't think that pr fixes this issue, it might be a prerequisite

This issue about to add ability to specify topics in helm

miguelaeh:
Hi @sergeyshaykhullin ,
IMHO the above PR provides a fast workaround so it is possible to configure an init container to create those topics when they don't exist. We are a small team and I think the creation of the hook will require some more time and testing. Would you like to create a PR adding that? We will be happy to review it.

sergeyshaykhullin:
@miguelaeh I had some time to implement draft version of topics provisioning

I found a little bit painful to work with kafka chart, because a lot of custom templates and monorepo structure
I will need some help to do it properly

juan131:
> I found a little bit painful to work with kafka chart, because a lot of custom templates and monorepo structure
[...]

PR #4783, opened 2020-12-18
[bitnami/kafka] Add topics provisioning job

sergeyshaykhullin:
**Description of the change**
Add hook job to run kafka-topics on each release

**Benefits**
You can specify which topics should be in kafka and it will create them for you

**Possible drawbacks**
For first iteration kafka topics can't be changed

**Applicable issues**
https://github.com/bitnami/charts/issues/4651

**Additional information**
```yaml
## Kafka provisioning
##
provisioning:
  enabled: true

  image:
    registry: docker.io
    repository: bitnami/kafka
    tag: 2.6.0-debian-10-r106
    ## Specify a imagePullPolicy
    ##
    pullPolicy: IfNotPresent
    ## Optionally specify an array of imagePullSecrets (secrets must be manually created in the namespace)
    ## Example:
    ## pullSecrets:
    ##
    pullSecrets: []

    ## Set to true if you would like to see extra information on logs
    ##
    debug: false

  resources:
    # We usually recommend not to specify default resources and to leave this as a conscious
    # choice for the user. This also increases chances charts run on environments with little
    # resources, such as Minikube. If you do want to specify resources, uncomment the following
    # lines, adjust them as necessary, and remove the curly braces after 'resources:'.
    limits: {}
    #   cpu: 250m
    #   memory: 1Gi
    requests: {}
    #   cpu: 250m
    #   memory: 256Mi

  topics:
  - name: topic-name1
    partitions: 1
    replicationFactor: 2
    # https://kafka.apache.org/documentation/#topicconfigs
    config:
      max.message.bytes: 64000
      flush.messages: 1
  - name: topic-name2
    partitions: 3
    replicationFactor: 4
    # https://kafka.apache.org/documentation/#topicconfigs
    config:
      max.message.bytes: 128000
      flush.messages: 10
```

```yaml
# Source: kafka/templates/kafka-provisioning.yaml
kind: Job
apiVersion: batch/v1
metadata:
  name: test-release-kafka-provisioning
  labels:
    app.kubernetes.io/name: kafka
    helm.sh/chart: kafka-12.5.0
    app.kubernetes.io/instance: test-release
    app.kubernetes.io/component: kafka
  annotations:
    helm.sh/hook: post-install,post-upgrade
    helm.sh/hook-delete-policy: before-hook-creation
spec:
  template:
    metadata:
      labels:
        app.kubernetes.io/name: kafka
        helm.sh/chart: kafka-12.5.0
        app.kubernetes.io/instance: test-release
        app.kubernetes.io/component: kafka
      annotations:
    spec:      
      restartPolicy: OnFailure
      terminationGracePeriodSeconds: 0
      containers:
        - name: kafka-provisioning
          image: docker.io/bitnami/kafka:2.6.0-debian-10-r106
          command:
            - /bin/bash
            - -c
            - >-
              set -e;
              /opt/bitnami/kafka/bin/kafka-topics.sh \
                --create \
                --if-not-exists \
                --bootstrap-server test-release-kafka:9092 \
                --replication-factor 2 \
                --partitions 1 \
                --config flush.messages=1 \
                --config max.message.bytes=64000 \
                --topic topic-name1;
              /opt/bitnami/kafka/bin/kafka-topics.sh \
                --create \
                --if-not-exists \
                --bootstrap-server test-release-kafka:9092 \
                --replication-factor 4 \
                --partitions 3 \
                --config flush.messages=10 \
                --config max.message.bytes=128000 \
                --topic topic-name2;
              echo "Success provisioning";
          env:
            - name: BITNAMI_DEBUG
              value: "false"
          resources:
          limits: {}
          requests: {}
          volumeMounts:
          volumes:
```

**Checklist**
- [x] Test this job is working
- [x] Add labels
- [x] Add annotations
- [x] Add tpl templates


sergeyshaykhullin:
This how it might be look like. I need a feedback to move forward

bitnami/kafka/values-production.yaml
```diff
@@ -745,6 +745,41 @@ rbac:
...
+    ## pullSecrets:
+    ##   - myRegistryKeySecretName
+    ##
+    pullSecrets: []
+
+    ## Set to true if you would like to see extra information on logs
+    ##
+    debug: false
+
+  topics:
```
sergeyshaykhullin, line 774: Fiz

bitnami/kafka/templates/kafka-provisioning.yaml
```diff
@@ -0,0 +1,42 @@
...
+            --if-not-exists \
+            --bootstrap-server {{ $bootstrapServer }} \
+            --replication-factor {{ $topic.replicationFactor }} \
+            --partitions {{ $topic.partitions }} \
+            {{- range $name, $value := $topic.config }}
+            --config {{ $name }}={{ $value }} \
+            {{- end }}
+            --topic {{ $topic.name }};
+        {{- end }}
+        env:
```
sergeyshaykhullin, line 39: I add only required env, do i need some more?

bitnami/kafka/templates/kafka-provisioning.yaml
```diff
@@ -0,0 +1,42 @@
+{{- if .Values.provisioning.enabled }}
+kind: Job
+apiVersion: batch/v1
+metadata:
+  name: {{ include "kafka.fullname" . }}-provisioning
+  labels:
```
sergeyshaykhullin, line 6: Same labels from statefulset?

bitnami/kafka/templates/kafka-provisioning.yaml
```diff
@@ -0,0 +1,42 @@
...
+  annotations:
```
sergeyshaykhullin, line 7: Fo not forget to add additional annotations

bitnami/kafka/templates/kafka-provisioning.yaml
```diff
@@ -0,0 +1,42 @@
...
+    helm.sh/hook: post-install,post-upgrade
+    helm.sh/hook-delete-policy: before-hook-creation
+spec:
+  template:
+    metadata:
+      labels:
```
sergeyshaykhullin, line 13: From statefulset?

bitnami/kafka/templates/kafka-provisioning.yaml
```diff
@@ -0,0 +1,42 @@
...
+      annotations:
```
sergeyshaykhullin, line 14: From statefulset?

sergeyshaykhullin:
I will add values to readme after review, because something might change

rafariossaa:
Hi, @dani8art would you mind taking a look into this?

bitnami/kafka/templates/kafka-provisioning.yaml
```diff
@@ -0,0 +1,91 @@
...
+      terminationGracePeriodSeconds: 0
+      containers:
+        - name: kafka-provisioning
+          image: {{ include "kafka.provisioning.image" . }}
+          imagePullPolicy: {{ .Values.provisioning.image.pullPolicy | quote }}
+          command:
+            - /bin/bash
+            - -c
+            - >-
+            {{- $bootstrapServer := printf "%s:%d" (include "kafka.fullname" .) (.Values.service.port | int64) }}
```
javsalgar, line 38: Would it make sense to add a wait for the server to be ready? We have a binary called wait-for-port that could work for this case, but we would need to add it to the container.

sergeyshaykhullin: I tested this problem, kafka-topics retries if kafka is not ready without exiting the process. But i think adding wait-for-port as init container still a good idea

sergeyshaykhullin: Because if kafka is not running or start slowly kafka-topics will throw

sergeyshaykhullin: @javsalgar I've not found docker image of bitnami wait-for-port in dockerhub, can you help me?

I found Dockerfile in repo, but i think ci is not publishing it

chris3081: for the retry just use restart OnFailure..

bitnami/kafka/templates/kafka-provisioning.yaml
```diff
@@ -0,0 +1,91 @@
...
+            {{- range $topic := .Values.provisioning.topics }}
+              /opt/bitnami/kafka/bin/kafka-topics.sh \
```
chris3081, line 40: There is no check after kafka-topics runs to see the exit code

bitnami/kafka/templates/kafka-provisioning.yaml
sergeyshaykhullin: Hmm? I'm already using it

bitnami/kafka/templates/kafka-provisioning.yaml
sergeyshaykhullin: @chris3081 What do you want to do with explicitly captured status codes? All errors you will see in logs, does exit codes means something for kafka-topics?

Do you want to stop execution on error and restart job? I think that `set -e` will solve it

chris3081 (commented):
yep a set -e   should do the trick

bitnami/kafka/templates/kafka-provisioning.yaml
```diff
@@ -0,0 +1,91 @@
...
+                --create \
+                --if-not-exists \
+                --bootstrap-server {{ $bootstrapServer }} \
+                --replication-factor {{ $topic.replicationFactor }} \
+                --partitions {{ $topic.partitions }} \
+                {{- range $name, $value := $topic.config }}
+                --config {{ $name }}={{ $value }} \
+                {{- end }}
+                --topic {{ $topic.name }};
+            {{- end }}
```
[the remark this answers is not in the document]
sergeyshaykhullin: Can you explain what topic injection failure means? Does `set -e` already fixes this problem?

bitnami/kafka/templates/kafka-provisioning.yaml
rafariossaa: Hi,
  `wait-for-port` is a small binary included in some of our container, for example [here](https://github.com/bitnami/bitnami-docker-zookeeper/blob/8ea99b75582c9565edaad7b499a397fd7dfd9f5d/3/debian-10/Dockerfile#L15)

sergeyshaykhullin: @rafariossaa https://github.com/bitnami/bitnami-docker-kafka/pull/135

bitnami/kafka/templates/kafka-provisioning.yaml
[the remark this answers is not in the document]
chris3081: Sure can, I gave this a run the other night and I had a spelling mistake in my values.yaml which was throwing an error in the pod logs, but the pod was exiting with zero due to the echo at the end exiting with a zero. So the restart on failure never triggered and rescheduled the job. The thinking I have is we can reliably fail until the server is up and we can inject the topics or until the helm deployment timeouts in which case helm should fail because the job isn't completed.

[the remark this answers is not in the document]
sergeyshaykhullin: Hmm, if you want non zero status codes on job error - set -e will propogate it

This job is a helm hook, if it fails - release will fail

sergeyshaykhullin:
```yaml
initContainers:
  - name: wait-for-available-kafka
    image: 'docker.io/bitnami/kafka:2.7.0-debian-10-r10'
    command:
      - /bin/bash
      - '-c'
      - |-
        set -e; wait-for-port \
          --host=test-kafka \
          --state=inuse \
          --timeout=120 \
          9092;
        echo "Kafka is available";
    resources: {}
    volumeMounts:
      - name: default-token-h2m2h
        readOnly: true
        mountPath: /var/run/secrets/kubernetes.io/serviceaccount
    terminationMessagePath: /dev/termination-log
    terminationMessagePolicy: File
    imagePullPolicy: IfNotPresent
```

bitnami/kafka/templates/kafka-provisioning.yaml
rafariossaa: Hi, 
  We had released a kafka image that includes that change.

sergeyshaykhullin: Yes, i added wait-for-port as initContainer to job

FraPazGal:
Hi, great PR!

Just my two cents, would an implementation only using `initContainers` be viable? From what my colleague @miguelaeh and you commented in the related issue it could be another implementation, although one with its own drawbacks. 

I'm only bringing this up given that we are actively trying to keep the use of hooks at a minimum as they don't work if you use only use helm for `helm template` and then installing the chart afterwards. Seeing that your are already using `initContainers`, could we expand their use to the topics creation?

sergeyshaykhullin:
Hmm, initContainers is a prerequisite of main workload => topic creation

If i move topic creation to init containers, that will be in containers? Echo "Success"? It seems weird, but possible ;3 Also i never tested job with initContainers only
InitContainers mostly have to exit succesfully. If them fail - k8s will restart initialization infinelty, because i am using `restartPolicy: OnFailure` for retries
Also InitContainers reporting just PodInitializing status, does it means that we are waiting for kafka or creating topics?

I think that possible solution is to add two serial initContainers for waiting kafka and topic creation, but is it better initContainer+container? Why?

sergeyshaykhullin:
Ohh... I understand you... You mean to use initContainers in kafka pods, i don't think it is a good idea, because:
- What will happen it you create new kafka cluster? You need running kafka to provision topics, kafka will not start before initContainers completed
- Each restart/new node of kafka will try to recreate topics, so pod startup time impact
- If initContainer will fail - kafka will not start. Job is an "optional" thing and not affecting kafka cluster

You can use sidecars in kafka pods to run topic creation in parallel with kafka pods, but it will have "second" and "third" drawbacks

sergeyshaykhullin:
Job is not affecting kafka cluster, runs only on new release

I can add something like `runAsHook: true`. It will fix `helm template` problem, but jobs are immutable, so you will have to delete previous job to run new one or generate unique name for job each release and GC succeeded job periodically

FraPazGal:
I see your point, maybe we have to keep going with hooks. 

> I can add something like runAsHook: true flag to use helm hook annotations conditionally. It will fix helm template problem , but jobs are immutable, so you will have to delete previous job to run new one or generate unique name for job each release and GC succeeded jobs periodically

I'm not sure if I understood it correctly. Are you talking about mimicking a hook behaviour without actually using hook annotations? Having to manually delete the jobs prior a `helm upgrade` seems very error prone. In that case it would be better to not have this functionality available for those users installing the chart with `helm template`.

sergeyshaykhullin:
No, i mean to add hook annotations conditionally. If you want to helm upgrade you will use hooks, if you want to helm template - you will use job as part of chart

FraPazGal:
Okay, thanks for the explanation. That could be a good middle ground, good idea. I'll test this PR internally and come back with my findings!

bitnami/kafka/templates/kafka-provisioning.yaml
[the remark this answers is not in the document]
FraPazGal: Hi @sergeyshaykhullin,

We should add `&&` to `--topic {{ $topic.name }} && ` instead of `;` as there are instances where:  the job installs one topic -> gets an error for the other one -> job restarts -> reruns both creation commands successfully -> the logs only shows `Created 2nd topic` as the first one was already created.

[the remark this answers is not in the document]
sergeyshaykhullin: Hmm, i am not really good at bash, but... Does && fix something?

First operation completed && second failed
Restart
First operation without stdout && second operation completed

I will test does && fix it. I thought that --if-not-exists if... exists prints nothing and exit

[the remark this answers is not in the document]
FraPazGal: It seems it won't solve my corner case, though at least it would restart the job if its the firts topic the one that fails. That way, only when the first topic is installed will the job continue.

bitnami/kafka/templates/kafka-provisioning.yaml
```diff
@@ -0,0 +1,109 @@
...
+apiVersion: batch/v1
+metadata:
+  name: {{ include "kafka.provisioning.fullname" . }}
+  labels: {{- include "common.labels.standard" . | nindent 4 }}
+    app.kubernetes.io/component: kafka
+    {{- if .Values.commonLabels }}
+    {{- include "common.tplvalues.render" ( dict "value" .Values.commonLabels "context" $ ) | nindent 4 }}
+    {{- end }}
+  annotations:
+    {{- if .Values.provisioning.runAsHook }}
```
FraPazGal, line 12: I have tested this `if` and I would say it is not really useful. That is, whether the annotations are present or not, the job installation will behave the same one when installing the chart with `helm template`. I think we can drop this `if` and leave it as it was.

FraPazGal (commented):
It looks good! Just a couple of comment, could you give them a look?

bitnami/kafka/templates/kafka-provisioning.yaml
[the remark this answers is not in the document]
sergeyshaykhullin: To restart job on first (N) failure i used set -e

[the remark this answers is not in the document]
sergeyshaykhullin: For now topics provisioning works sequentially

Do you mean you want to provision topics is parallel and if any failed - try again?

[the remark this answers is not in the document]
FraPazGal: That would be ideal, as the topic's creation will always be reflected on the logs in its entirety.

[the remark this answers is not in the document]
sergeyshaykhullin: No, when you create topics in parallel, you can also lose logs. Even more

Start creating 10 topics
9 created, 1 failed
Restart job, lose 9 topic creation logs
1 topic created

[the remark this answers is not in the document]
FraPazGal: You are right, I completely misunderstood that code. I was thinking of a way to have all the created topics on the job's log. What do you think about adding something like:

```
              echo "Provisioning succeeded.";
              /opt/bitnami/kafka/bin/kafka-topics.sh \
                --bootstrap-server {{ $bootstrapServer }} \
                --list
```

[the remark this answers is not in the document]
sergeyshaykhullin: Can you please provide a use-case for listing all topics? Does kafka-topics.sh supports somehow listing newly created topics only?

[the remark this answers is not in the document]
FraPazGal: It would be just for readability as it would avoid confusion. Although it would list all the existing topics and not only the newly created, it may be a compromise to not having the stdout for those topics created before the pod's last restart.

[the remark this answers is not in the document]
sergeyshaykhullin: I think current approach is good. If you creating topic - you will get message, if no message for topic you want - means it is already created

For readability i think before each kafka-topics run i can add something like 'echo Ensure {topic} created'

If topic exists - next ensure message logged
If topic created - kafka-topics will print message before 'ensure' message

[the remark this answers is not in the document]
FraPazGal: Yes, I think that could be a good compromise. Thank you for your work!

FraPazGal (approved):
Thank you very much for the great PR @sergeyshaykhullin! LGTM!

PR #4790, opened 2020-12-19
[bitnami/contour] Allow extra arguments to be specified, and passed to Contour container

MrSaints:
**Description of the change**

Allow extra arguments to be specified, and passed to Contour container.

It can be specified in the values using the `contour.extraArgs` key. An array must be provided.

**Benefits**

This is useful for enabling the PROXY protocol by specifying
`--use-proxy-protocol` as an extra arg.

See https://projectcontour.io/guides/proxy-proto/ for more information.

It can potentially be used for other options.

**Possible drawbacks**

- The implementation requires hard coding / specifying the YAML indentation
- It does not account for removal of existing arguments in the template (a full arg override option is probably needed for this, but it is out-of-scope for this change)

**Applicable issues**

N/A

**Additional information**

Other options considered:

1. Full arg override: goes against sane defaults, and may be too fiddly
2. Boolean option / toggle: not extensible if we want to cover other non-default options

**Checklist** 


bitnami/contour/templates/contour/deployment.yaml
```diff
@@ -67,6 +67,9 @@ spec:
             {{- if .Values.ingressClass }}
             - --ingress-class-name={{ .Values.ingressClass }}
             {{- end }}
+            {{- if .Values.contour.extraArgs }}
+            {{ toYaml .Values.contour.extraArgs | indent 12 | trim }}
```
dani8art, line 71: Maybe it could be even better using `common.tplvalues.render` helper, don't you think?

dani8art (changes requested):
Hi @MrSaints thank you so much for this PR

Please, take a look at my comment

bitnami/contour/templates/contour/deployment.yaml
MrSaints: Sounds good! Thanks for the feedback.

MrSaints:
Rebased due to upstream changes.

dani8art (approved):
Thanks! LGTM!

PR #4802, opened 2020-12-21
[bitnami/zookeeper] feat: add initContianers

### LINKED ISSUE #4793: [bitnami/zookeeper] extraInitContainers
**Which chart**:

bitnami/zookeeper

When deploying in a cluster with readOnlyRootFilesystem, conf folder is not writable. Mounting an emptyDir volume there will override all config files inside.

Ability to define extraInitContainers would enable copying original config files to a shared emptyDir volume which is mounted to /opt/bitnami/zookeeper/conf to ensure folder is writable with all original config files

Manually mounting original config files as secret but they are not writable (even with defaultMode 0777). Also maintaining external config files is a burden between version upgrades.

**Additional context**

This also makes it consistent with other bitnami charts which allow extraInitContainers in values.yaml

dani8art:

**Description of the change**

- Add `.Values.initContainers`

**Benefits**

We will be able to add extra init container to the statefulset.

**Possible drawbacks**

N/A

**Applicable issues**

  - fixes #4793

**Additional information**

**Checklist** 


bitnami/zookeeper/templates/statefulset.yaml
```diff
@@ -65,8 +65,8 @@ spec:
       {{- if .Values.priorityClassName }}
       priorityClassName: {{ .Values.priorityClassName }}
       {{- end }}
+      initContainers: {{ include "common.tplvalues.render" (dict "value" .Values.initContainers "context" $) | trim | nindent 8 }}
```
carrodher, line 68: In this case, if there are no initContainers specified in the _values.yaml_, an empty `initContainers:` section is going to be rendered in the statefulset, right?

I think what we are doing in other cases is something like
```yaml
{{- if or (and .Values.volumePermissions.enabled .Values.persistence.enabled) .Values.initContainers }}
initContainers:
  {{- if and .Values.volumePermissions.enabled .Values.persistence.enabled }}
  - name: volume-permissions
    ...
  {{- end }}
  {{- include "common.tplvalues.render" ( dict "value" .Values.initContainers "context" $ ) | nindent 8 }}
  {{- end }}
{{- end }}
```

bitnami/zookeeper/values-production.yaml
```diff
@@ -258,6 +257,18 @@ securityContext:
...
+## Example:
+## initContainers:
+##   - name: your-image-name
+##     image: your-image
+##     imagePullPolicy: Always
+##     ports:
+##       - name: portname
+##         containerPort: 1234
+##
+initContainers:
```
carrodher, line 270: Using the previous suggested approach
```suggestion
initContainers: []
```

carrodher (commented):
The suggested approach is used, for example, in kafka.

carrodher (approved):

PR #4820, opened 2020-12-22
[bitnami/redis] Fix Redis sentinel synchronization and user creation permission error

### LINKED ISSUE #4819: [bitnami/redis] Redis with sentinel enabled can have separate masters
**Which chart**:
bitnami/redis:12.2.4

**Describe the bug**
There can be seperate masters with varying slaves(usually none) when the statefulset is starting up or recovering from a failure with no pods running which results in master-slave sync failure. This also causes data loss upon failure/shutdown if `redis` service redirects to a master that has no slaves. Also redis user when cannot be created which outputs a permission error.

**To Reproduce**

1. Watch the endpoints by `kubectl get endpoints -w`.
2. Deploy the helm chart using `helm install redis bitnami/redis -f values-production.yaml`. (Only differences from the original file are `sentinel.enabled=true`, `metrics.serviceMonitor.enabled=true`.
3. Check the pod logs for each StatefulSet node by `kubectl logs redis-node-* -c redis`. ( * can be 0,1,2 ... replicaCount-1)
4. First pod has the logs shown below and some slave pods might also have the same logs which points that `redis-headless` service can't be resolved by the dns at the time and the running node is set as master.
```
useradd: Permission denied.
useradd: cannot lock /etc/passwd; try again later.
chown: invalid user: 'redis'
 Warning: Using a password with '-a' or '-u' option on the command line interface may not be safe.
Could not connect to Redis at redis-headless.default.svc.cluster.local:26379: Name or service not known
I am master
redis 19:46:57.77 INFO  ==> ** Starting Redis **
1:C 22 Dec 2020 19:46:57.779 # oO0OoO0OoO0Oo Redis is starting oO0OoO0OoO0Oo
1:C 22 Dec 2020 19:46:57.779 # Redis version=6.0.9, bits=64, commit=00000000, modified=0, pid=1, just started
1:C 22 Dec 2020 19:46:57.779 # Configuration loaded
1:M 22 Dec 2020 19:46:57.781 * Running mode=standalone, port=6379.
1:M 22 Dec 2020 19:46:57.781 # Server initialized
1:M 22 Dec 2020 19:46:57.782 * Ready to accept connections
```
5. Delete all the slave nodes by `kubectl delete po redis-node-1 redis-node-2` (Include all the slave node names) 
6. After deletion and recreation of the pods, connections to master node as slaves are successfull since headless-service registered it's ip address to the relative Endpoints object which enables slaves to resolve the `redis-headless` service.

**Expected behavior**
A successful startup without having to restart slave nodes.

- Output of `helm version`:

```
version.BuildInfo{Version:"v3.4.2", GitCommit:"23dd3af5e19a02d4f4baa5b2f242645a1a3af629", GitTreeState:"clean", GoVersion:"go1.14.13"}
```

```
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.5", GitCommit:"e338cf2c6d297aa603b50ad3a301f761b4173aa6", GitTreeState:"clean", BuildDate:"2020-12-09T11:18:51Z", GoVersion:"go1.15.2", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.5", GitCommit:"e338cf2c6d297aa603b50ad3a301f761b4173aa6", GitTreeState:"clean", BuildDate:"2020-12-09T11:10:32Z", GoVersion:"go1.15.2", Compiler:"gc", Platform:"linux/amd64"}
```

**Additional context**
This seems like a race condition. [Headless services with selectors](https://kubernetes.io/docs/concepts/services-networking/service/#with-selectors) in kubernetes creates Endpoints records in the API, and modifies the DNS configuration to return records (addresses) that point directly to the Pods backing the Service. This record is created after the pod is ready (readinessProbe is met) unless headless service contains `publishNotReadyAddresses: true` in it's spec. The StatefulSet node pods have `readinessProbe`s as local ping checks for both the `redis` and the `sentinel` containers. Which means that any node that boots up properly can be ready and let the waiting nodes creations continue. The first node to boot up will always be the master since no ip address is registered on the Endpoints object. But the slaves might bootup in a chain even before kubernetes can register the ip for the pod that just got ready which causes dns resolves in slaves to fail and get an empty response. In that case "slaves" also declare themselves as masters which breaks the synchronization. 

Failure for user creation is caused by a missing `nil` check in the helm template. [This line](https://github.com/bitnami/charts/blob/28701368ae0d7c5a08b4439254bf54700f1e8e06/bitnami/redis/templates/configmap-scripts.yaml#L45) casts the variable to integer which results in nil being converted to `0`. Since we want to create a user if we have our user set to `root`, not having a `runAsUser` property in our `values.yaml` causes these line to be added since nil is casted to 0 and the condition is met. But we don't actually have our user as root in the container which results in a permission error when trying to create a user.

berkayoz:
**Description of the change**

Edited `redis-headless` service to contain `publishNotReadyAddresses: true` to make redis nodes discoverable without being ready due to [kubernetes implementation](https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-hostname-and-subdomain-fields). This change caused a bug with the current startup scripts when `redis-headless` service redirects the sentinel connections to the same pod which isn't initialized yet. The `redis` service is used for sentinel/redis connections which only contains the fully initialized and ready pods as the workaround. Removed the `boot.lock` files due to these changes making the need for `boot.lock` files(redisboot.lock, sentinelboot.lock) obsolete. 

Also fixed the issue where not having a `runAsUser` under the `securityContext` definition in our `values.yaml` causes a permission error due to `nil` casting to 0 which adds the code to create a user named `redis` when `runAsUser` is set to root(0). Since we don't actually have our user set as root the commands face a permission error.

**Benefits**

Redis is now booting up, restarting, scaling up/down properly with synchronization successfully. This fix also prevents a potential data loss if a separate master without a slave is resolved by the `redis` service, failure/deletion of this node causes the data to be deleted when it synchronizes with an available master upon startup.

**Possible drawbacks**

I am unsure of the performance effects of using the `redis` service for sentinel/replication communication which also serves the clients.

**Applicable issues**

  - fixes #4819

**Additional information**

Please let me know if there are any bugs/improvements I've missed.

**Checklist** 

### DIFF
--- a/bitnami/redis/Chart.yaml
+++ b/bitnami/redis/Chart.yaml
@@ -19,4 +19,4 @@ name: redis
 sources:
   - https://github.com/bitnami/bitnami-docker-redis
   - http://redis.io/
-version: 12.3.1
+version: 12.3.2
--- a/bitnami/redis/templates/configmap-scripts.yaml
+++ b/bitnami/redis/templates/configmap-scripts.yaml
@@ -24,25 +24,14 @@ data:
     }
 
     HEADLESS_SERVICE="{{ template "redis.fullname" . }}-headless.{{ .Release.Namespace }}.svc.{{ .Values.clusterDomain }}"
+    REDIS_SERVICE="{{ template "redis.fullname" . }}.{{ .Release.Namespace }}.svc.{{ .Values.clusterDomain }}"
 
     export REDIS_REPLICATION_MODE="slave"
     if [[ -z "$(getent ahosts "$HEADLESS_SERVICE" | grep -v "^$(hostname -i) ")" ]]; then
-      if [[ ! -f /data/redisboot.lock ]]; then
-        export REDIS_REPLICATION_MODE="master"
-      else
-        if is_boolean_yes "$REDIS_TLS_ENABLED"; then
-          sentinel_info_command="redis-cli {{- if .Values.usePassword }} -a $REDIS_PASSWORD {{- end }} -h $HEADLESS_SERVICE -p {{ .Values.sentinel.port }} --tls --cert ${REDIS_TLS_CERT_FILE} --key ${REDIS_TLS_KEY_FILE} --cacert ${REDIS_TLS_CA_FILE} info"
-        else
-          sentinel_info_command="redis-cli {{- if .Values.usePassword }} -a $REDIS_PASSWORD {{- end }} -h $HEADLESS_SERVICE -p {{ .Values.sentinel.port }} info"
-        fi
-        if [[ ! ($($sentinel_info_command)) ]]; then
-           export REDIS_REPLICATION_MODE="master"
-           rm /data/redisboot.lock
-        fi
-      fi
+      export REDIS_REPLICATION_MODE="master"
     fi
 
-    {{- if (eq (.Values.securityContext.runAsUser | int) 0) }}
+    {{- if and .Values.securityContext.runAsUser (eq (.Values.securityContext.runAsUser | int) 0) }}
     useradd redis
     chown -R redis {{ .Values.slave.persistence.path }}
     {{- end }}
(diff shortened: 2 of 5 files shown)


dani8art (commented):
Hi @berkayoz thanks for this PR. It looks good to me but I would like @miguelaeh to take a look at this too.

marcosbc:
I think @javsalgar would also be able to check this if @miguelaeh is unavailable

rafariossaa:
Hi,
  I am reviewing the PR, but meanwhile there was a new chart release. 
  Would you like to bump the chart version ?
  Also, I would like to understand this change better. Would you mind to provide some background on why you needed to add `publishNotReadyAddresses=true` ?

berkayoz:
Hello,
The reason for adding `publishNotReadyAdresses=true` is that our containers have their `readinessProbe` as pinging(connecting to) the local redis/sentinel container that just booted up. The problem is that our containers are looking for other redis/sentinel nodes on the headless service to synchronize but these services ip addresses are not published until their `readinessProbe` succeeds. Since `readinessProbe` can only succeed after the containers finished up booting, another pod might start initializing just after the previous pod that got ready and try to synchronize even before that pod can publish its ip address. If we add `publishNotReadyAdresses=true` the pods will publish their ip address as soon as the pod starts up which ensures that ip adresses will be always registered before container initialization.

An example race condition happens like this:
redis-node-0 starts up, can't find any other nodes on the headless service and declares itself master.
redis-node-1 starts as soon as the redis-node-0 got ready and starts initializing and checks the headless service for other nodes. There are 2 possibilities: redis-node-0 could publish its ip address before the check or redis-node-0 could not publish its ip address before the check which results in redis-node-1 declaring itself master too.
Having more than 1 master in a masterset might result in data loss if a master without slaves dies. When that node boots back up it finds services on the headless service which registered their ip addresses way before and starts synchronizing to master node which results in data written to that master without any slaves to be lost.

I also bumped the chart version.

miguelaeh:
Hi guys,
I have been checking the PR and it LGTM.
@berkayoz could you please rebase master and bump the version again?

miguelaeh (approved):

berkayoz:
@miguelaeh  rebased and bumped the version, should be fine now. Thanks.

PR #4831, opened 2020-12-23
[bitnami/*] POC Lookup function implementation

FraPazGal:
Signed-off-by: fdepaz <fdepaz@bitnami.com>

**Description of the change**

This is a POC for the `lookup` function implementation, which solves the issue of the regenerating secret's passwords during `helm upgrade`. Implemented in three charts to allow for an initial study of the change's effectivity as well as to limit possible bugs to a small part of the bitnami/charts catalog.

The change consist of a helpers function in `common` where there are several comprobations made:

1. Whether a secret with the provided name already exists, in which case the function returns the existing password's value for the provided key.
2. If there is no secret, the function checks whether there is a defined value by the user in any of the provided parameters.
3. If there is no secret nor custom user value, the function generates a random password of the defined length and adding `_` (if defined as such).

This implementation also makes some code related to password/secret management in our charts obsolete, such as:

- Removed the `hooks` from MetalLB as they are no longer required to avoid password regeneration.
- Removed OrangeHRM password validations. Now Helm will retrieve the passwords from the secret without user input.

> Note: the slightly complex code of strong password generation comes from the need to have at least one upper case, lower case, symbol and numerical value to have a "strong" password by [OrangeHRM Policy](https://help.orangehrm.com/hc/en-us/articles/360016271693-OrangeHRM-Standard-Password-Policy).

**Benefits**

- Solves the issue with password regeneration on upgrades.
- Cleaner code with only one template in `common`.

**Possible drawbacks**

Keep in mind that Helm is not supposed to contact the Kubernetes API Server during a helm template or a `helm install|update|delete|rollback --dry-run`, so the `lookup` function will return an empty list (i.e. dict) in such a case. For users that may be using gitops or rendering templates to a git repo before applying them to the cluster, it could be a problem if we don't take this into account. 

Having said that, the worst case scenario would be that the secret will just regenerate, just as it is functioning now.

**Applicable issues**

  - fixes #

**Additional information**

**Checklist** 


bitnami/common/templates/_secrets.tpl
```diff
@@ -55,3 +55,48 @@ Params:
 
 {{- printf "%s" $key -}}
 {{- end -}}
+
+{{/*
+Generate secret password or retrieve one if already created.
+
+Usage:
+{{ include "common.secrets.manage.password" (dict "secret" "secret-name" "key" "keyName" "customValues" (list .Values.path.to.the.password1 .Values.path.to.the.password2) "length" "password-length" "strong" false "context" $) }}
```
dani8art, line 63: Could it be `"common.secrets.passwords.manage"`?

bitnami/common/templates/_secrets.tpl
```diff
@@ -55,3 +55,48 @@ Params:
...
+
+Params:
+  - secret - String - Required - Name of the 'Secret' resource where the password is stored.
+  - key - String - Required - Name of the key in the secret.
+  - customValues - Optional - List of parameters that can be used by the user to define the password. Will pick first parameter with a defined value.
```
dani8art, line 68: Why not only `{given|custom|provided}Value` in singular and as `string`? Do you know any case where a password could be set via two different keys?

bitnami/common/templates/_secrets.tpl
```diff
@@ -55,3 +55,48 @@ Params:
...
+  - length - int - Optional - Length of the generated random password
+  - strong - Boolean - Optional - Whether to add "_" symbols to the generated random password
```
dani8art, line 70: Why adding "_" would be strong?

bitnami/common/templates/_secrets.tpl
```diff
@@ -55,3 +55,48 @@ Params:
...
+  {{- $providedPassword = first $passwordList }}
+{{- end }}
+{{- $secret := (lookup "v1" "Secret" $.context.Release.Namespace .secret) }}
+
+{{- if $secret }}
+  {{- if index $secret.data .key }}
+  {{- $password = index $secret.data .key }}
+  {{- end -}}
+{{- else if $providedPassword }}
+  {{- $password = $providedPassword | b64enc | quote }}
```
dani8art, line 89: This could conflict with passwords build with only numbers, could we add an additional pipe like `| toString |` so this issue will be fixed?

bitnami/common/templates/_secrets.tpl
```diff
@@ -55,3 +55,48 @@ Params:
...
+{{- else }}
+  {{- if .strong }}
+    {{- $subStr := list (lower (randAlpha 1)) (randNumeric 1) (upper (randAlpha 1)) | join "_" }}
+    {{- $password = randAscii $passwordLength }}
+    {{- $password = regexReplaceAllLiteral "\\W+" $password "@" | substr 5 $passwordLength }}
+    {{- $password = cat $subStr $password | nospace | shuffle | b64enc | quote }}
```
dani8art, line 95: - Could we `printf` instead of `cat`, please?
- same with `| toString |`

bitnami/common/templates/_secrets.tpl
```diff
@@ -55,3 +55,48 @@ Params:
...
+
+{{- if $secret }}
+  {{- if index $secret.data .key }}
+  {{- $password = index $secret.data .key }}
+  {{- end -}}
+{{- else if $providedPassword }}
+  {{- $password = $providedPassword | b64enc | quote }}
+{{- else }}
+  {{- if .strong }}
+    {{- $subStr := list (lower (randAlpha 1)) (randNumeric 1) (upper (randAlpha 1)) | join "_" }}
```
dani8art, line 92: Could you show an example of what this part of the function does? until `cat` function.

dani8art (changes requested):
Great! this will improve our charts a lot! Please, take a look at my comments.

bitnami/common/templates/_secrets.tpl
FraPazGal: My bad, I didn't change this definition after my latests changes. It should say `Whether to add symbols...`

bitnami/common/templates/_secrets.tpl
FraPazGal: AFAIK, only [bitnami/postgresql](https://github.com/bitnami/charts/blob/86b98162b81dccc144615f0c5bd5313e7e57c905/bitnami/postgresql/templates/_helpers.tpl#L48) and [bitnami/postgresql-ha](https://github.com/bitnami/charts/blob/86b98162b81dccc144615f0c5bd5313e7e57c905/bitnami/postgresql-ha/values.yaml#L5) have that option, but I included it nonetheless.

bitnami/common/templates/_secrets.tpl
FraPazGal: Added both suggestions, though we shouldn't need `| toString |` as we will always have at least 1 uppercase letter, 1 lowercase and 2 `_` symbols.

bitnami/common/templates/_secrets.tpl
FraPazGal: If we want a "strong" password of length `10`, the code would follow this steps:

1. We create a String `subStr` that will have 1 random lowercase letter, 1 number and 1 uppercase letter with the structure `a_1_A`. This is to pass the OrangeHRM password validation a cited in the description.
2. We generate a random string `password` of length 10 that can contain every printable ASCII character.
3. We replace every non letter or number character on `password` and change them for `@`. This is to avoid posible conflicts if the generated strings has characters like `"` or spaces. Then, we only save in the variable the generated string starting at the position `5`. That is, we don't save the characters `0-4`.
4. We concatenate `password` and `subStr`, which would give us a password of length `10`. Then we shuffle the characters to avoid a fixed pattern.

1. `d_4_Z`
2. `59o%p"9KLs`
3. `59o@p@9KLs` -> `@9KLs`
4. `d_4_Z@9KLs` -> `_sK94Z_@Ld`

alemorcuq:
> For users that may be using gitops or rendering templates to a git repo before applying them to the cluster, it could be a problem if we don't take this into account.

What would be the difference for users not using Helm to deploy the chart compared to the previous approach?

FraPazGal:
Replying to @alemorcuq I discovered that I have missed the use-case of a user not using Helm to upgrade the chart. I have included a new helper and restored the password validation on upgrades taking into account the `lookup` function implementation.

As it is now, there shouldn't be any difference between both implementations for anyone using gitops or similar. They will still receive the error of having to `--set` the values on upgrades and the passwords will still be generated on install.

bitnami/common/templates/_secrets.tpl
dani8art: do you mean the global one?

bitnami/common/templates/_secrets.tpl
dani8art: Ok

bitnami/common/templates/_secrets.tpl
dani8art: Cool!

bitnami/common/templates/_secrets.tpl
dani8art: Perfect!

bitnami/orangehrm/templates/NOTES.txt
```diff
@@ -64,7 +64,7 @@ host. To configure OrangeHRM to use and external database host:
 {{- $passwordValidationErrors := list -}}
 {{- $secretName := include "orangehrm.secretName" . -}}
 
-{{- if not .Values.existingSecret -}}
+{{- if and (not .Values.existingSecret) (not (include "common.secrets.exists" (dict "secret" $secretName "context" $))) -}}
```
dani8art, line 67: This will be a very common statement for all the asset, could we find a way of moving it to the common helpers? Maybe in "common.validations.values.single.empty" or "common.errors.upgrade.passwords.empty"

bitnami/common/templates/_secrets.tpl
FraPazGal: Yes. Another option would be to implement a helper in those charts when `lookup` is added and avoid having `customValues ` as a list.

bitnami/orangehrm/templates/NOTES.txt
FraPazGal: I have tried a couple of things but I haven't been able to change neither of those helpers without changing their current behaviour. If we decide to have `customValue` in singular and as a String, I think it would be easy to move the entire validation of passwords to `common`.

dani8art: do you mean to add the validation inside "common.secrets.passwords.manage"?

dani8art: Could we test it? let's see it, I think that could be really good, don't you think?

FraPazGal:
To summarise a bit the latest changes:

- I have updated `customValues-> providedValues` maintaining the list but having it as a `list<String>`.
- I have included the password validation on `common.secrets.passwords.manage` using an auxiliary helper `common.utils.getKeyFromList` to refactor the input of `providedValues` for `common.validations.values.single.empty`.
- I changed `providedValues` to be required instead of optional, as the password validation's helpers need a `valueKey`. I created a new parameter for bitnami/metallb because of this.

For those users installing the charts without `helm` everything will work as before, getting the missing password errors instead of having them automatically obtained by the `lookup` function.

There is only one difference between the new and old password validation implementations. Given that we are validating the passwords one by one, the error messages when upgrading the chart with missing passwords won't show every missing password at once. That is, if we are missing 2 passwords we will get a `export and set password1` message and after solving that error we will get a `export and set password2` instead of the old `export and set password1 and password2` message, but I will say that is a minor inconvenience.

dani8art (approved):
> That is, if we are missing 2 passwords we will get a export and set password1 message and after solving that error we will get a export and set password2 instead of the old export and set password1 and password2 message, but I will say that is a minor inconvenience.

Yes, I think so, and it will be fixed if we move in the future the entire creation of the secret to the common where we will receive a list of passwords to be created and can validate all together.

LGTM! Please, let's test it carefully as it is a big change. Great work here!!

Thanks!

FraPazGal:
- Fixed regexp used to replace ASCII characters when creating a strong password.
- Added subchart support. Using the option "chartName" will make the helper check whether the validated password is part of a subchart and print the export password instructions correctly.

bitnami/common/templates/_secrets.tpl
```diff
@@ -55,3 +55,73 @@ Params:
...
+  {{- end -}}
+{{- else if $providedPasswordValue }}
+  {{- $password = $providedPasswordValue | toString | b64enc | quote }}
+{{- else }}
+
+  {{- if .context.Values.enabled }}
+    {{- $subchart = $chartName }}
+  {{- end -}}
+
+  {{- $requiredPassword := dict "valueKey" $providedPasswordKey "secret" .secret "field" .key "subchart" $subchart "context" $.context -}}
```
dani8art, line 95: I think `"subchart" $subchart` is a boolean

dani8art: Oh sorry, you are implementing a compatible change in "common.validations.value.empty", right?

FraPazGal: Yes, I needed to expand a bit that helper. I have tested it and it seems to be working correctly.

charfa:
@FraPazGal, @dani8art, you talked about testing this change out on the three charts in this PR before rolling it out to other charts in the repo. When do you think it will be mature/stable enough to start doing that?

I could use having this implemented in rabbitmq chart and I can prepare a PR for that if you need support, but my use case is quite limited and I won't be able to test it on all secret management scenarios there.

javsalgar:
Hi,

This will require some further discussion. We need to update the decision documents but the idea with the charts is to not rely on helm functions as much as possible. This is because we have several clients that don't want to use helm but render the templates following a gitops approach. Because of this, we may need to re-think this approach.

charfa:
I see. I thought that the solution proposed here would be fine for those clients, as it's not supposed to impact the behavior when doing a dry-run.

Anyway, I'll keep watching this space for a decision then.

PR #4836, opened 2020-12-25
[bitnami/mariadb] Allow to set priority class for primary and secondary pods

fl42:
**Description of the change**

This PR allows to set priority class for primary and secondary pods.

**Benefits**

Allow to set priority class ([docs](https://kubernetes.io/docs/concepts/configuration/pod-priority-preemption/))

**Possible drawbacks**

No

**Applicable issues**

No

**Additional information**

**Checklist** 


bitnami/mariadb/templates/primary/statefulset.yaml
```diff
@@ -60,7 +60,7 @@ spec:
       tolerations: {{- include "common.tplvalues.render" (dict "value" .Values.primary.tolerations "context" $) | nindent 8 }}
       {{- end }}
       {{- if .Values.priorityClassName }}
```
marcosbc, line 62: It seems like the `if .Values.priorityClassName` line should also be updated.

marcosbc (changes requested):
Thanks for the contribution!

I'm a bit concerned with this being a breaking change, as we'd be renaming parameters.

For now, to keep things working, could you update the logic to make use of `priorityClassName` if set, as long as `primary/secondary.priorityClassName` is not set?

Also, we should probably also update the chart version as a minor version increase (i.e. 9.1.4 -> 9.2.0) due to there being new flags.

bitnami/mariadb/templates/primary/statefulset.yaml
fl42: Thanks for the heads up

fl42:
@marcosbc As the value was not documented (AFAIK) I did not consider this as a breaking change
But your comment makes sense, I fixed accordingly

The priority class will be taken from `.Values.primary.priorityClassName` if it exists, or from `.Values.priorityClassName`, if it exists, or no priorityClassName will be set.
Same for secondary.

How does it sound to you?

marcosbc (approved):
LGTM!

Issue #4791, opened 2020-12-19
bitnami/wordpress: ValidationError(Ingress.spec.rules[0].http.paths[0].backend): unknown field "serviceName" in io.k8s.api.networking.v1.IngressBackend


acim:
**Which chart**:
bitnami/wordpress

**Describe the bug**
diff fails, probably install and upgrade as well.

**To Reproduce**

**Expected behavior**
To show diff and be able to upgrade.

**Version of Helm and Kubernetes**:

- Output of `helm version`:

```
version.BuildInfo{Version:"v3.4.2", GitCommit:"23dd3af5e19a02d4f4baa5b2f242645a1a3af629", GitTreeState:"clean", GoVersion:"go1.14.13"}
```

- Output of `kubectl version`:

```
Client Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.1", GitCommit:"c4d752765b3bbac2237bf87cf0b1c2e307844666", GitTreeState:"clean", BuildDate:"2020-12-18T12:09:25Z", GoVersion:"go1.15.5", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.1", GitCommit:"c4d752765b3bbac2237bf87cf0b1c2e307844666", GitTreeState:"clean", BuildDate:"2020-12-18T12:00:47Z", GoVersion:"go1.15.5", Compiler:"gc", Platform:"linux/amd64"}
```

**Additional context**

```txt
STDERR:
  Error: Failed to render chart: exit status 1: Error: unable to build kubernetes objects from release manifest: error validating "": error validating data: [ValidationError(Ingress.spec.rules[0].http.paths[0].backend): unknown field "serviceName" in io.k8s.api.networking.v1.IngressBackend, ValidationError(Ingress.spec.rules[0].http.paths[0].backend): unknown field "servicePort" in io.k8s.api.networking.v1.IngressBackend, ValidationError(Ingress.spec.rules[1].http.paths[0].backend): unknown field "serviceName" in io.k8s.api.networking.v1.IngressBackend, ValidationError(Ingress.spec.rules[1].http.paths[0].backend): unknown field "servicePort" in io.k8s.api.networking.v1.IngressBackend]
  Error: plugin "diff" exited with error

COMBINED OUTPUT:
[...]

PR #4850, opened 2020-12-29
[bitnami/common] POC of Ingress adaptation (NOT TO BE MERGED)

javsalgar:
**Description of the change**

In order to make Ingress rules compatible between Kubernetes 1.20 and lower, we add a functionality to the `common` library for generating different types of `backend` entries depending on the apiVersion:

   - The apiVersion is generated using `.Capabilities.apiVersion`
   - It is also possible to force a specific apiVersion by using the `ingress.apiVersion` value. This is important for GitOps approaches that may not depend on the K8s client/server versions.
   -  We add a helper function for generating the backend entry, which is different depending on the API Version
   - We add the pathType value (which is set to `ImplementationSpecific` by default), which is compatible with all versions.
Tested using kubeval and found no issues.

If the approach is correct, I will create a PR for bitnami/common and then proceed to update all the ingress entries in the chart.

**Benefits**

Our chart is compatible with both versions

**Possible drawbacks**

No known issues

**Applicable issues**

  - fixes https://github.com/bitnami/charts/issues/4791

**Additional information**

**Checklist** 


bitnami/wordpress/values.yaml
```diff
@@ -450,18 +457,22 @@ ingress:
...
-  ##   path: /
+  ##   - name: wordpress.local
+  ##     path: /
   ##
 
   ## Any additional arbitrary paths that may need to be added to the ingress under the main host.
   ## For example: The ALB ingress controller requires a special rule for handling SSL redirection.
   ## extraPaths:
   ## - path: /*
+  ##   pathType: ImplementationSpecific
```
marcosbc, line 470: Isn't `pathType` / `ImplementationSpecific` something only added very recently (in 1.18)? https://kubernetes.io/blog/2020/04/02/improvements-to-the-ingress-api-in-kubernetes-1.18/

It would break compatibility with this chart and older K8s versions where it was working before, even if you specify an older value to `ingress.apiVersion`.

bitnami/common/templates/_ingress.tpl
```diff
@@ -0,0 +1,28 @@
...
+{{- define "common.ingress.backend" -}}
+{{- $apiVersion := (include "common.capabilities.ingress.apiVersion" .context) -}}
+{{- if (eq $apiVersion "extensions/v1beta1") -}}
+serviceName: {{ .serviceName }}
+servicePort: {{ .servicePort }}
+{{- else -}}
+service:
+  name: {{ .serviceName }}
+  port:
+    {{- if typeIs "string" .servicePort }}
```
marcosbc, line 22: What if the port was specified as a numeric string? I.e. `"8080"` in YAML instead of `8080`. Will `typeIs` work properly for that case?

alemorcuq:
Looks great. My only concern is with the new `pathType` field starting in k8s 1.18. Does it work in previous versions?

bitnami/common/templates/_ingress.tpl
```diff
@@ -26,3 +26,17 @@ service:
...
+{{ include "common.ingress.supportsPathType" . }}
+*/}}
+{{- define "common.ingress.supportsPathType" -}}
+{{- $apiVersion := (include "common.capabilities.ingress.apiVersion" .) -}}
+{{- if (eq $apiVersion "extensions/v1beta1") -}}
+{{- print "false" -}}
+{{- else -}}
+{{- print "true" -}}
+{{- end -}}
+{{- end -}}
```
alemorcuq, line 42: Does this address the `pathType` concern? k8s 1.17 would print `true` but `pathType` was added in k8s 1.18

```
❯ k version
Client Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.9", GitCommit:"4fb7ed12476d57b8437ada90b4f93b17ffaeed99", GitTreeState:"clean", BuildDate:"2020-07-15T16:18:16Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"17+", GitVersion:"v1.17.14-gke.1200", GitCommit:"7c407f5cc8632f9af5a2657f220963aa7f1c46e7", GitTreeState:"clean", BuildDate:"2020-12-01T09:20:59Z", GoVersion:"go1.13.15b4", Compiler:"gc", Platform:"linux/amd64"}


❯ k explain ingress.spec.rules.http.paths.pathType
error: field "pathType" does not exist
```

javsalgar: pathType was added in `networking/v1beta1` if I'm not mistaken. It k8s 1.17 it would use extensions/v1beta1

alemorcuq:
```
❯ k explain ingress.spec.rules.http.paths --api-version='networking.k8s.io/v1beta1'
KIND:     Ingress
VERSION:  networking.k8s.io/v1beta1

RESOURCE: paths <[]Object>

DESCRIPTION:
     A collection of paths that map requests to backends.

     HTTPIngressPath associates a path regex with a backend. Incoming urls
     matching the path are forwarded to the backend.

FIELDS:
   backend	<Object> -required-
     Backend defines the referenced service endpoint to which the traffic will
     be forwarded to.

   path	<string>
     Path is an extended POSIX regex as defined by IEEE Std 1003.1, (i.e this
     follows the egrep/unix syntax, not the perl syntax) matched against the
     path of an incoming request. Currently it can contain characters disallowed
     from the conventional "path" part of a URL as defined by RFC 3986. Paths
     must begin with a '/'. If unspecified, the path defaults to a catch all
     sending traffic to the backend.
     ```

Comment (alemorcuq) [reply bitnami/common/templates/_ingress.tpl:42]: If I'm not missing anything, 1.17 would use `networking/v1beta1`:

```
{{- if .Values.ingress.apiVersion -}}
{{- .Values.ingress.apiVersion -}}
{{- else if semverCompare "<1.14-0" .Capabilities.KubeVersion.Version -}}
{{- print "extensions/v1beta1" -}}
{{- else if semverCompare "<1.19-0" .Capabilities.KubeVersion.Version -}}
{{- print "networking.k8s.io/v1beta1" -}}
{{- else -}}
{{- print "networking.k8s.io/v1" -}}
{{- end -}}
```

File: bitnami/common/templates/_ingress.tpl [hunk 4]
```diff
@@ -0,0 +1,64 @@
...
+
+{{/*
+Generate path entry that is compatible with all Kubernetes API versions.
+
+Usage:
+{{ include "common.ingress.backend" (dict "path" "pathPrefix" "pathType" "pathType" "serviceName" "backendName" "servicePort" "backendPort" "context" $) }}
+
+Params:
+  - path - String. Path inside the service
+  - pathType - String. Path typo
```
alemorcuq, line 11: Typo

bitnami/common/templates/_ingress.tpl
javsalgar: Added extra check

bitnami/common/templates/_ingress.tpl
javsalgar: I wouldn't overcomplicate this. It is us the ones that are setting these values when developing the chart.

bitnami/wordpress/values.yaml
javsalgar: Fixed

bitnami/common/templates/_ingress.tpl
```diff
@@ -0,0 +1,64 @@
...
+{{- end -}}
+
+{{/*
+Print "true" if the API pathType field is supported
+Usage:
+{{ include "common.ingress.supportsPathType" . }}
+*/}}
+{{- define "common.ingress.supportsPathType" -}}
+{{- $apiVersion := (include "common.capabilities.ingress.apiVersion" .) -}}
+{{- if or (eq $apiVersion "extensions/v1beta1") (semverCompare "<1.18-0" .Capabilities.KubeVersion.Version) -}}
```
alemorcuq, line 59: If I use `extensions/v1beta1` with k8s 1.19 it knows about the `pathType`:

```
❯ k version
Client Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.0", GitCommit:"af46c47ce925f4c4ad5cc8d1fca46c7b77d13b38", GitTreeState:"clean", BuildDate:"2020-12-13T19:50:45Z", GoVersion:"go1.15.5", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.4", GitCommit:"d360454c9bcd1634cf4cc52d1867af5491dc9c5f", GitTreeState:"clean", BuildDate:"2020-11-11T13:09:17Z", GoVersion:"go1.15.2", Compiler:"gc", Platform:"linux/amd64"}
```

```
❯ k explain ingress.spec.rules.host.paths.pathType
KIND:     Ingress
VERSION:  extensions/v1beta1

FIELD:    pathType <string>

DESCRIPTION:
     Host is the fully qualified domain name of a network host, as defined by
     RFC 3986. Note the following deviations from the "host" part of the URI as
     defined in RFC 3986: 1. IPs are not allowed. Currently an IngressRuleValue
     can only apply to the IP in the Spec of the parent Ingress.
     2. The `:` delimiter is not respected because ports are not allowed.
     Currently the port of an Ingress is implicitly :80 for http and :443 for
     https. Both these may change in the future. Incoming requests are matched
     against the host before the IngressRuleValue. If the host is unspecified,
     the Ingress routes all traffic based on the specified IngressRuleValue.

     Host can be "precise" which is a domain name without the terminating dot of
     a network host (e.g. "foo.bar.com") or "wildcard", which is a domain name
     prefixed with a single wildcard label (e.g. "*.foo.com"). The wildcard
     character '*' must appear by itself as the first DNS label and matches only
     a single label. You cannot have a wildcard label by itself (e.g. Host ==
     "*"). Requests will be matched against the Host field in the following way:
     1. If Host is precise, the request matches this rule if the http host
     header is equal to Host. 2. If Host is a wildcard, then the request matches
     this rule if the http host header is to equal to the suffix (removing the
     first label) of the wildcard rule.
```

Honestly, I'm probably showing some lack of knowledge here in how things work in k8s.

javsalgar: I see, so it changed in all of the API specs. That's a terrible practice, but if it is supported I will simplify the check

alemorcuq: Thanks!

marcosbc:
LGTM

marcosbc (approved):

alemorcuq (approved):

marcosbc (approved):

Issue #4412, opened 2020-11-18
Unable to change jenkin password after the installtion


mkmrah:
We are facing issue while updating the Jenkin chart values. Note for the testing purpose i have used helm3 CLI on IBM cloud cluster.

1. Jenkin https://charts.bitnami.com/ibm/jenkins-6.0.0.tgzHelm chart installed using the latest helm 3 client (https://github.com/helm/helm/releases/tag/v3.4.1)

2. helm release deployed into Kubernetes cluster and the jenkin server is accessible via the service API as mentioned in the readme document.

3. After the installation, we want to change the jenkin password, so you used "helm release upgrade" (helm upgrade jenkins6b-5045-46 jenkins --reuse-values --set jenkinsPassword=password1212 --debug) command to update the password.

4. There is no error / warning in the console while executing the helm upgrade command.

5. We can see the secret value change in K8 cluster and pod was not re-started.

6. When we use the new credential for jenkin server login then we get the authentication error but old password is working.

7. Tried with --force option and got below error

```
client.go:219: [debug] error updating the resource "jenkins6b-5045-46":
         failed to replace object: PersistentVolumeClaim "jenkins6b-5045-46" is invalid: spec: Forbidden: is immutable after creation except resources.requests for bound claims
client.go:219: [debug] error updating the resource "jenkins6b-5045-46":
         failed to replace object: Service "jenkins6b-5045-46" is invalid: spec.clusterIP: Invalid value: "": field is immutable
```

Looks like POD is using the old secret (it is defined as env_value inside the container) and its not refreshed as part of the helm upgrade. So here is my questions

1. How do we update the password after the installation?

2. If its an issue then any work around to manage this?

FraPazGal:
Hello @mkmrah!

As you said, the jenkins password is passed to the container by an env and used only at installation time. When an upgrade is made, the new pod will reuse the persisted data present in the pvc so the container will skip the installation and won't use the secret anymore:

```
$ kubectl logs jenkins-654d5fb9bc-hpvn9

...
nami    INFO  Initializing jenkins
jenkins INFO  Moving plugins to /opt/bitnami/jenkins/jenkins_home...
jenkins INFO  Restoring persisted data...
jenkins INFO  Waiting Jenkins to be deployed...
jenkins INFO
jenkins INFO  ########################################################################
jenkins INFO   Installation parameters for jenkins:
jenkins INFO     Persisted data and properties have been restored.
jenkins INFO     Any input specified will not take effect.
[...]

PR #4853, opened 2020-12-29
[bitnami/*] Add notice regarding parameters immutability after chart installation

migruiz4:
**Description of the change**

Hi team,

This PR adds a notice in the chart catalog READMEs to make clear that once a chart is installed, using `helm upgrade --set appnamePassword --set appnameUsername` won't change the username or password, as it is something set at installation time. 

The notice has been included after the block of code with the chart's installation example. Using bitnami/jenkins as an example:

-------------------

```console
$ helm install my-release \
  --set jenkinsUser=admin,jenkinsPassword=password \
    bitnami/jenkins
```
The above command sets the Jenkins administrator account username and password to admin and password respectively.

>    NOTE: once installed, neither username nor password can be changed using Helm CLI. If given the capability, please use the in-app features to do so.

-------------------

**Affected charts**

The notice has been added to the following charts:
- airflow
- consul
- discourse
- dokuwiki
- drupal
- ejbca
- etcd 
- ghost
- grafana 
- harbor
- influxdb
- jasperreports
- jenkins
- joomla
- keycloak
- kibana
- magento
- mariadb
- mariadb-galera
- mediawiki
- memcached
- minio
- mongodb
- moodle
- mysql
- nats
- odoo
- opencart
- orangehrm
- osclass
- owncloud
- parse
- phabricator
- phpbb
- postgresql
- postgresql-ha
- prestashop
- rabbitmq
- redis
- redis-cluster
- redmine
- suitecrm
- testlink
- tomcat
- wildfly
- wordpress
- zookeeper

The notification has been added to those applications with username and password

**Related issues**
- fixes https://github.com/bitnami/charts/issues/4412

### DIFF
--- a/bitnami/airflow/README.md
+++ b/bitnami/airflow/README.md
@@ -355,6 +355,8 @@ $ helm install my-release \
 
 The above command sets the credentials to access the Airflow web UI.
 
+> NOTE: Once this chart is deployed, it is not possible to change the application's access credentials, such as usernames or passwords, using Helm. To change these application credentials after deployment, delete any persistent volumes (PVs) used by the chart and re-deploy it, or use the application's built-in administrative tools if available.
+
 Alternatively, a YAML file that specifies the values for the parameters can be provided while installing the chart. For example,
 
 ```console
--- a/bitnami/consul/README.md
+++ b/bitnami/consul/README.md
@@ -216,6 +216,8 @@ $ helm install my-release --set domain=consul-domain,gossipKey=secretkey bitnami
 
 The above command sets the HashiCorp Consul domain to `consul-domain` and sets the gossip key to `secretkey`.
 
+> NOTE: Once this chart is deployed, it is not possible to change the application's access credentials, such as usernames or passwords, using Helm. To change these application credentials after deployment, delete any persistent volumes (PVs) used by the chart and re-deploy it, or use the application's built-in administrative tools if available.
+
 Alternatively, a YAML file that specifies the values for the above parameters can be provided while installing the chart. For example,
 
 ```console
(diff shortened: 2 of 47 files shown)


bitnami/airflow/README.md
```diff
@@ -344,6 +344,8 @@ $ helm install my-release \
 
 The above command sets the credentials to access the Airflow web UI.
 
+> NOTE: once installed, neither username nor password can be changed using Helm CLI. If given the capability, please use the in-app features to do so.
```
marcosbc, line 347: I'd recommend re-writing this as it is a bit confusing:

```
 > NOTE: Once installed for the first time, it is not possible to change username/password credentials. To change the credentials, it is necessary to remove persistent volumes (PV) and re-install.
```

@vikram-bitnami What do you think?

marcosbc (changes requested):

bitnami/airflow/README.md
vikram-bitnami: WDYT of the below? I use `access credentials` because then you can use this message even for apps that use different types of credentials (tokens etc) and I mentioned both the options available for changing post-deployment

```
> NOTE: Once this chart is deployed, it is not possible to change the application's access credentials, such as usernames or passwords, using Helm. To change these application credentials after deployment, delete any persistent volumes (PVs) used by the chart and re-deploy it, or use the application's built-in administrative tools (if available).
```

migruiz4: Thank you for your suggestions, I will rewrite the note.

I'm considering removing the brackets from the last sentence as it seems to fit well and it doesn't need to be a side note.
> Or use the application's built-in administrative tools if available.

marcosbc (approved):
LGTM, thanks!

PR #4863, opened 2020-12-30
[bitnami/redis] Attach existing volume as persistence storage

ronakthamor:
**Description of the change**

If volume is exist, there should be an option to attach using disk name and disk URI. Volumes can be assigned directly to the pod without creating / using existing PersistentVolumeClaim.

It can be used on azure like this (https://docs.microsoft.com/en-us/azure/aks/azure-disk-volume):

    master:
        persistence:
          enabled: true
          volumes:
            - name: data
               azureDisk:
                   kind: Managed
                   diskName: "static-redis"
                   diskURI: "/subscriptions/*****/resourceGroups/storage-rg/providers/Microsoft.Compute/disks/static-redis"

**Benefits**

One can attach existing volume without creating PVC directly to the statefulset.

**Possible drawbacks**

None. Existing feature is still there. This is an improvement.


rafariossaa:
Hi,
  Thanks your very much for you PR.
  There was a new release of the chart, do you mind bumping the version ?.
Thanks forehand.

FraPazGal (approved):
LGTM!

PR #4766, opened 2020-12-17
[bitnami/mongodb-sharded] fix for #4536

alrf:
**Description of the change**

**Benefits**

**Possible drawbacks**

**Applicable issues**

  - fixes #4536

**Additional information**

**Checklist** 
- [x] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [x] Variables are documented in the README.md
- [x] Title of the PR starts with chart name (e.g. `[bitnami/chart]`)
- [x] If the chart contains a `values-production.yaml` apart from `values.yaml`, ensure that you implement the changes in both files

:warning: Keep in mind that if you want to make changes to the kubeapps chart, please implement them in the [kubeapps repository](https://github.com/kubeapps/kubeapps/tree/master/chart/kubeapps). This is only a synchronized mirror.

PR #4876, opened 2020-12-31
Fix probe error messages

Orabig:
**Description of the change**

Default liveness and readiness probes are based on the command `mongo --eval "db.adminCommand('ping')"` which fills the logs with this error message : 
````
ACCESS   [conn165] Unauthorized: not authorized on admin to execute command { endSessions: [ { id: UUID("5e039c6f-fc79-48e1-9f99-73fb8c5b5c25") } ], $db: "admin" }
````

This change adds an additional `--disableImplicitSessions` which removes this annoying error message.

**Benefits**

No more error message that may be difficult to understand and fills the logs for no reason

**Possible drawbacks**

-

**Applicable issues**

-

**Additional information**

- EDIT : I just noticed that there is a typo in the commit messages (bitnami->bitname). If necessary, I can post another PR.

### DIFF
--- a/bitnami/mongodb-sharded/Chart.yaml
+++ b/bitnami/mongodb-sharded/Chart.yaml
@@ -26,4 +26,4 @@ name: mongodb-sharded
 sources:
   - https://github.com/bitnami/bitnami-docker-mongodb-sharded
   - https://mongodb.org
-version: 3.2.7
+version: 3.2.8
--- a/bitnami/mongodb-sharded/templates/config-server/config-server-statefulset.yaml
+++ b/bitnami/mongodb-sharded/templates/config-server/config-server-statefulset.yaml
@@ -193,16 +193,10 @@ spec:
           readinessProbe:
             exec:
               command:
-                - sh
-                - -ec
-                - |-
-                  #!/bin/sh
-                  {{- if $.Values.usePasswordFile }}
-                  export MONGODB_INITIAL_PRIMARY_ROOT_PASSWORD="$(cat "${MONGODB_ROOT_PASSWORD_FILE}")"
-                  {{- else }}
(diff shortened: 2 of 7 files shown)


alemorcuq:
Hi, @Orabig. We recently merged #4766, fixing this error for `mongodb-sharded`. Could you try if the same approach fixes your issue in `mongodb`?

Orabig:
Thanks for the information @alemorcuq : Yes, I just tested the fix in #4766 and it seems to work, but I'm uncertain if this solution is better. Here is why : 
- It uses root and explicit password, while it is not necessary
- It does not fix the issue in some case, because there is a need to add a fallback case ( the `... || mongo --eval "db.adminCommand('ping')"` part). If that fallback is used (when the initial root password has changed for some reason I guess), then the unwanted error message is back.
- It produces 2 additional lines in the logs prompting `"Successfully authenticated as principal root on admin from client"` each time. Thus, we have 5 lines of log instead of 3 lines with this fix. Knowing that the objective was to reduce the log size, this seems paradoxical. 
- It is much more complex (10 big lines instead of 4 easy-to-read ones)

Initially, I only needed to have a fix on the `mongodb` chart, but of course, I also published it in the `mongodb-shard` chart which is very similar, for consistency. So I still believe that whatever solution is chosen, it has to be deployed on both charts. Too bad this hasn't be done in #4766.

alemorcuq:
Thanks for checking, @Orabig. I agree this solution is cleaner, so if you fix the conflicts in the PR and upgrade the version in the `Chart.yaml` file I will merge the PR.

Orabig:
Merge is ready @alemorcuq . Thanks !

alemorcuq (approved):
Thanks!

PR #3788, opened 2020-09-27
[bitnami/kubernetes-event-exporter] Introduce kubernetes-event-exporter

mkilchhofer:
**Description of the change**
Contribute a new helm chart for [opsgenie/kubernetes-event-exporter](https://github.com/opsgenie/kubernetes-event-exporter/)

**Benefits**
Deploy the [opsgenie/kubernetes-event-exporter](https://github.com/opsgenie/kubernetes-event-exporter/) via a Helm Chart

**Possible drawbacks**
`-`

**Applicable issues**
`-`

**Additional information**
More infos about the software can be found here: https://github.com/opsgenie/kubernetes-event-exporter/

**Checklist** 
- [x] Discuss / organize bitnami container image build process
- [x] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [x] Variables are documented in the README.md
- [x] Title of the PR starts with chart name (e.g. `[bitnami/chart]`)
- [x] If the chart contains a `values-production.yaml` apart from `values.yaml`, ensure that you implement the changes in both files

:warning: Keep in mind that if you want to make changes to the kubeapps chart, please implement them in the [kubeapps repository](https://github.com/kubeapps/kubeapps/tree/master/chart/kubeapps). This is only a synchronized mirror.

PR #4881, opened 2021-01-03
[bitnami/contour] Implement a default backend for convenience

mkilchhofer:
**Description of the change**

As I also use nginx-ingress-controller, I always missed the possibility to manage a default backend within contour.
While here, I updated additionals things:
- Refactor parameters into smaller blocks
- Drop the `values-production.yaml` as it is identical with `values.yaml` and there is a comment about you want to remove this in the future:
  > Also, you can delete the values-production.yaml file, as it is something that is in our roadmap to be removed from all of our charts.
  Ref: https://github.com/bitnami/charts/pull/3788#pullrequestreview-516865711

**Benefits**

Add convenience for migrating users from other ingress controllers such as nginx-ingress-controller.

**Possible drawbacks**

Nothing. To use this, you have to Opt-in. It is not enabled by default (at least for now).

**Applicable issues**

`-`

**Additional information**

`-`

**Checklist** 


mkilchhofer:
The upstream nginx-ingress (https://kubernetes.github.io/ingress-nginx/) implements the default backend with a custom container image: 
```yaml
  image:
    repository: k8s.gcr.io/defaultbackend-amd64
    tag: "1.5"
```
Ref: https://github.com/kubernetes/ingress-nginx/blob/master/charts/ingress-nginx/values.yaml#L597-L599

Is there a reason, why the bitnami chart uses a plain nginx for this? It would be nice, if we could use this container image or a bitnami variant of this.

bitnami/contour/README.md
```diff
@@ -153,24 +173,66 @@ The following tables lists the configurable parameters of the contour chart and
...
+| `defaultBackend.resources.limits`          | The resources limits for the Default backend container                                   | `{}`                                                    |
+| `defaultBackend.resources.requests`        | The requested resources for the Default backend container                                | `{}`                                                    |
+| `defaultBackend.leavinessProbe`            | Leaviness probe configuration for Default backend                                        | Check `values.yaml` file                                |
+| `defaultBackend.readinessProbe`            | Readiness probe configuration for Default backend                                        | Check `values.yaml` file                                |
+| `defaultBackend.customLivenessProbe`       | Override default liveness probe                                                          | `nil`                                                   |
+| `defaultBackend.customReadinessProbe`      | Override default readiness probe                                                         | `nil`                                                   |
```
FraPazGal, line 203: This 2 parameters seem to be missing from `values.yaml`.

bitnami/contour/templates/default-backend/ingress.yaml
```diff
@@ -0,0 +1,25 @@
+{{- if .Values.defaultBackend.enabled }}
+apiVersion: {{ include "common.capabilities.ingress.apiVersion" . }}
```
FraPazGal, line 2: This is throwing an error on my end as there is no `.Values.ingress.apiVersion` in `values.yaml`.

FraPazGal (commented):
Hello @mkilchhofer, thanks for the great PR! 

Indeed, dropping the `values-production.yaml` files is in our short-term backlog, so this is one less to change! 

Regarding the use of a custom `defaultbackend` nginx image upstream, I would say that they use a custom one to give a custom html page for `/` instead of the default `404` one to make for an easier customization, but I can't seem to find any other difference between the custom and standard one. Given that our nginx image defaults to a `200` on `/`, I would say that there is not much to gain from changing our approach what do you think? 

Just a couple of comments, could you please take a look?

bitnami/contour/templates/default-backend/ingress.yaml
mkilchhofer: Ahh.. this is common v1.2.0 :)
Fixed - defined the values.

bitnami/contour/README.md
mkilchhofer: Done

mkilchhofer:
Thanks for the review @FraPazGal 

The image `k8s.gcr.io/defaultbackend-amd64:1.5` contains only a single-binary and not a full nginx. But until now I did not find where the k8s guys published the source code.

FraPazGal (approved):
LGTM!

Good catch on the source code! It seems that image was specifically created to work with the default chart configuration of the upstream ingress-nginx. I suppose that when we created our chart we decided to use bitnami/nginx, as creating an extra bitnami container just for the default backend would have been excessive. Nevertheless, thanks for asking and investigating!

Issue #4877, opened 2021-01-01
postgresqlConfiguration causes service to be unconnectable


djhoese:
**Which chart**:
bitnami/postgresql postgresql-9.8.1

**Describe the bug**
I would like to force the size of my shared_buffers. As described in the readme I set `postgresqlConfiguration.sharedBuffers="10MB"` and the helm chart installs correctly. However, when trying to connect to the service it is unconnectable and says that Connection refused. Am I missing something?

**To Reproduce**

1. `helm install -n default --set persistence.enabled=false --set postgresqlConfiguration.sharedBuffers="10MB" test-shared-buffers bitnami/postgresql `
2. `kubectl get pod,svc`

   ```
   NAME                                   READY   STATUS    RESTARTS   AGE
   pod/test-shared-buffers-postgresql-0   1/1     Running   0          12m

   NAME                                      TYPE        CLUSTER-IP     EXTERNAL-IP   PORT(S)    AGE
   kubernetes                                ClusterIP   10.43.0.1      <none>        443/TCP    218d
   test-shared-buffers-postgresql-headless   ClusterIP   None           <none>        5432/TCP   7s
   test-shared-buffers-postgresql            ClusterIP   10.43.218.53   <none>        5432/TCP   7s
   ```
3. `kubectl exec -it test-shared-buffers-postgresql-0 -- bash`
4. Note how only connecting to localhost is successful:

   ```
   I have no name!@test-shared-buffers-postgresql-0:/$ psql postgres postgres
   Password for user postgres:
   I have no name!@test-shared-buffers-postgresql-0:/$ psql postgres postgres -h test-shared-buffers-postgresql
   psql: could not connect to server: Connection refused
           Is the server running on host "test-shared-buffers-postgresql" (10.43.218.53) and accepting
           TCP/IP connections on port 5432?
   I have no name!@test-shared-buffers-postgresql-0:/$ psql postgres postgres -h test-shared-buffers-postgresql-headless
   psql: could not connect to server: Connection refused
           Is the server running on host "test-shared-buffers-postgresql-headless" (10.42.0.155) and accepting
           TCP/IP connections on port 5432?
   I have no name!@test-shared-buffers-postgresql-0:/$ psql postgres postgres -h 10.43.218.53
   psql: could not connect to server: Connection refused
           Is the server running on host "10.43.218.53" and accepting
           TCP/IP connections on port 5432?
[...]

PR #4883, opened 2021-01-04
[bitnami/postgresql] Clarify sharedBuffers example in postgresql README

djhoese:
**Description of the change**

Clarify the postgresql chart README as discussed in #4877. Changes the `sharedBuffers` configuration example to use `postgresqlExtendedConf` which "extends" the configuration file instead of the current example of `postgresqlConfiguration` which **replaces** the configuration.

**Benefits**

The current description was confusing to me as following the example exactly produces a pod/service that is not usable in the same way the default configuration. If this resulted in small differences I'd say maybe it was OK, but currently this results in a postgresql server that can't be connected to outside localhost.

**Possible drawbacks**

Since the change mentions both configuration options I don't see any downsides.

**Applicable issues**

  - related to #4877 

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/postgresql/README.md
+++ b/bitnami/postgresql/README.md
@@ -353,7 +353,7 @@ This helm chart also supports to customize the whole configuration file.
 
 Add your custom file to "files/postgresql.conf" in your working directory. This file will be mounted as configMap to the containers and it will be used for configuring the PostgreSQL server.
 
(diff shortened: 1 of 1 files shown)


djhoese:
I wasn't sure if the README was included as part of the chart and if the chart would need a version bump. Let me know.

miguelaeh (approved):
The README is included in the chart tgz but we don't think it is necessary to bump the version for this kind of changes, just merging this and wait for the next version..

PR #4898, opened 2021-01-06
[bitnami/suitecrm] Fix for smtpPassword and SUITECRM_SMTP_PORT

malagant:
**Description of the change**

When deploying bitnami/suitecrm the pod complains about missing smtp password and port.

**Benefits**

With this fix suitecrm will deploy correctly when using smtp settings.

**Possible drawbacks**

None.

**Applicable issues**

  - fixes #

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/suitecrm/templates/deployment.yaml
+++ b/bitnami/suitecrm/templates/deployment.yaml
@@ -187,11 +187,11 @@ spec:
               value: {{ .Values.suitecrmNotifyAddress | quote }}
             - name: SUITECRM_SMTP_HOST
               value: {{ .Values.suitecrmSmtpHost | quote }}
-            - name: SUITECRM_SMTP_PORT_NUMBER
+            - name: SUITECRM_SMTP_PORT
               value: {{ .Values.suitecrmSmtpPort | quote }}
             - name: SUITECRM_SMTP_USER
               value: {{ .Values.suitecrmSmtpUser | quote }}
-            {{- if .Values.smtpPassword }}
+            {{- if .Values.suitecrmSmtpPassword }}
             - name: SUITECRM_SMTP_PASSWORD
               valueFrom:
                 secretKeyRef:
--- a/bitnami/suitecrm/templates/deployment.yaml
+++ b/bitnami/suitecrm/templates/deployment.yaml
@@ -187,7 +187,7 @@ spec:
               value: {{ .Values.suitecrmNotifyAddress | quote }}
             - name: SUITECRM_SMTP_HOST
               value: {{ .Values.suitecrmSmtpHost | quote }}
-            - name: SUITECRM_SMTP_PORT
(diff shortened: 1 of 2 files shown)


bitnami/suitecrm/templates/deployment.yaml
```diff
@@ -187,11 +187,11 @@ spec:
               value: {{ .Values.suitecrmNotifyAddress | quote }}
             - name: SUITECRM_SMTP_HOST
               value: {{ .Values.suitecrmSmtpHost | quote }}
-            - name: SUITECRM_SMTP_PORT_NUMBER
+            - name: SUITECRM_SMTP_PORT
```
juan131, line 190: The env. variable used to configure SMTP is actually `SUITECRM_SMTP_PORT_NUMBER`, see:

- https://github.com/bitnami/bitnami-docker-suitecrm/blob/master/7/debian-10/rootfs/opt/bitnami/scripts/libsuitecrm.sh#L303

That said, both are aliases so it would work using both formats. In any case, please revert this change

bitnami/suitecrm/templates/deployment.yaml
```diff
@@ -187,11 +187,11 @@ spec:
...
               value: {{ .Values.suitecrmSmtpPort | quote }}
             - name: SUITECRM_SMTP_USER
               value: {{ .Values.suitecrmSmtpUser | quote }}
-            {{- if .Values.smtpPassword }}
+            {{- if .Values.suitecrmSmtpPassword }}
```
juan131, line 194: Nice catch!

juan131 (changes requested):
Thanks so much for this PR! Please check my comments.

bitnami/suitecrm/templates/deployment.yaml
malagant: Sorry, but when I open the link I only see SUITECRM_SMTP_PORT. Am I missing something?

juan131: Sorry, I shared the wrong link... Here we're using `SUITECRM_SMTP_PORT_NUMBER`:
- https://github.com/bitnami/bitnami-docker-suitecrm/blob/master/7/debian-10/rootfs/opt/bitnami/scripts/libsuitecrm.sh#L74
It seems there's some inconsistency around this, I'll fix it in the container image. 
Here you can find they're actually aliases:
- https://github.com/bitnami/bitnami-docker-suitecrm/blob/master/7/debian-10/rootfs/opt/bitnami/scripts/suitecrm-env.sh#L86
But we should use `SUITECRM_SMTP_PORT_NUMBER` for consistency with other Bitnami containers.

malagant: Thanks for the clarification. I Reverted the change. Thank you again.

juan131 (approved):
LGTM

PR #4902, opened 2021-01-07
[bitnami/mongodb] Add helpful information about some variable validity

Orabig:
**Description of the change**

Added a comment explaining when the custom readiness/liveness probe configurations are taken into account

**Benefits**

I had to spend some time reading the template source code to understand why my configuration was not taken into account, so I guess this comment will spare this work to others

**Possible drawbacks**

None

**Applicable issues**

None


andresbono (commented):
Thanks for contributing with these changes!

Could you please increase the `version` field in the `Chart.yaml`?

```diff
-version: 10.3.4
+version: 10.3.5
```

andresbono (approved):
Never mind, we did it to speed things up. Thanks!!

PR #4911, opened 2021-01-07
[bitnami/redis] Fix Redis Sentinel Setup

ricoberger:
**Description of the change**

- This adds support to add additional labels and annotations to the StatefulSet used for the Redis Sentinel setup. It uses the existing values from `slave.statefulset.labels` and `slave.statefulset.annotations`, because we are also using the slave section for the configuration of the persistent volumes.
- This fixes `volumeMounts` for the `redis-data` volume:
  - The `REDIS_DATA_DIR` is set to `slave.persistence.path`, but the volume mount used the hardcoded value `/data` in the Redis container.
  - The Redis Sentinel container used the `master.persistence.path` instead of `slave.persistence.path`, which looks wrong, since the volume is configured using the salve values.

**Benefits**

This uses a consistent style for the configuration of the volume mounts across all containers and allows to add additional labels and annotations for the StatefulSet.

**Possible drawbacks**

Could be a breaking change for users, which are using different values for the master/slave configuration.

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/redis/Chart.yaml
+++ b/bitnami/redis/Chart.yaml
@@ -19,4 +19,4 @@ name: redis
 sources:
   - https://github.com/bitnami/bitnami-docker-redis
   - http://redis.io/
-version: 12.3.2
+version: 12.3.3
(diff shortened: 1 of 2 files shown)


andresbono (approved):
LGTM

andresbono:
Thanks for your contribution, @ricoberger!

PR #4915, opened 2021-01-07
[bitnami/elasticsearch] allow for custom volumeClaimTemplate selectors

MatthiasLohr:
**Description of the change**

The current version of the elasticsearch helm chart only allows for the hard-coded volume selector label name `volume`. Adding the new options `master.persistence.existingVolumeSelector` and `data.persistence.existingVolumeSelector`, it gives users the freedom to define custom selectors using their own labels.

**Benefits**

With this pull request, instead of only having a hardcoded label named `volume` as volume selector, users can define their own custom selectors for `existingVolumeClaimTemplates`.

**Possible drawbacks**

Nothing known.

**Applicable issues**

*N.A.*

**Additional information**

In a future major rewrite of this helm chart (targeting a new major version), it should be considered to improve the structure of settings for existing volumes/claimes. However, this will break downwards compatibility, so this is not a real solution for minor updates.

**Checklist** 


FraPazGal:
Hello @MatthiasLohr,

Thank you so much for the PR! I agree this would be a good addition to our chart, but I have doubts about the parameters' name. If I'm not wrong, we will be adding selectors to the pvc, regardless if one of those is `volume` for an existingVolume. As such, I think `master.persistence.selector` would be closer the actual definition, wouldn't it?

MatthiasLohr:
@FraPazGal, thank you for your comment. I have no strong opinion about the parameter's name. However, what, if not an existing volume, can you select using these selectors? As to my knowledge, that's only matching to existing volumes, or am I wrong here?

FraPazGal:
That is a good point, you are right. I thought this could cause misunderstandings as we always use `persistence.selector` in our charts for this functionality. It really shouldn't, so I would say that standarisations is the only reason I'll change to `.selector`.

MatthiasLohr:
> [...] so I would say that standarisations is the only reason I'll change to `.selector`.

I agree, done.

Doca:
That's exactly what I need!

FraPazGal (approved):
LGTM! Thanks!!

PR #4937, opened 2021-01-11
[bitnami/mongodb] Make readiness probe only validate usable instance

Orabig:
**Description of the change**

Change the readiness probe code to return "true" only if the mongodb instance is PRIMARY or SECONDARY state.

**Benefits**

Before this change, the readiness probe was exactly the same than the liveness (it just detected if the mongodb instance was running). The main risk here is to tell that the member is "ready" while it is RECOVERING. This could lead to data loss, for example in a scenario where there is an helm upgrade (or a kubernetes update) which is rolling out all the members, and that for some reason one or several member need some time to recover their data. If Kubernetes has no way to know that the members are not ready, it can blindly shut down the other members at a fast pace, which could have major consequences

**Possible drawbacks**

None afaik

**Applicable issues**

none

**Additional information**

This commit has been tested on a 4.2 cluster, a 4.2 standalone instance and a 3.6 standalone instance.

**Checklist** 
- (NA) Variables are documented in the README.md
- (NA)If the chart contains a `values-production.yaml` apart from `values.yaml`, ensure that you implement the changes in both files

### DIFF
--- a/bitnami/mongodb/Chart.lock
+++ b/bitnami/mongodb/Chart.lock
@@ -1,6 +1,6 @@
 dependencies:
 - name: common
   repository: https://charts.bitnami.com/bitnami
-  version: 1.1.3
-digest: sha256:3422f8c1139c2f0f0d816e7ee7280705d2010e83c3e2367b28dcb4f7dd911135
-generated: "2020-12-22T11:27:07.137984504Z"
+  version: 1.2.3
+digest: sha256:3fc1fbf3ae204e0121f1e202d6d57f9381f3a45d8821647d1dfe0a475644da0c
+generated: "2021-01-12T13:11:04.410273076Z"
--- a/bitnami/mongodb/Chart.yaml
+++ b/bitnami/mongodb/Chart.yaml
@@ -26,4 +26,4 @@ name: mongodb
 sources:
   - https://github.com/bitnami/bitnami-docker-mongodb
   - https://mongodb.org
-version: 10.3.6
+version: 10.3.7
--- a/bitnami/mongodb/templates/replicaset/statefulset.yaml
+++ b/bitnami/mongodb/templates/replicaset/statefulset.yaml
@@ -312,11 +312,13 @@ spec:
           readinessProbe:
             exec:
               command:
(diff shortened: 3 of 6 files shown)


andresbono:
Hi @Orabig, thanks for your contribution! We are reviewing the PR. Is this PR related to an existing issue that you know of?

Orabig:
No, didn't found any issue here about this point (maybe there will be one when someone loose its data because of this ;-) )

andresbono (approved):
Thanks again for your contribution @Orabig!! We have done some minor style changes and a minor fix for the options when tls is enabled.

hamdifourati:
Hello, the merged solution rely on `db.hello()` command.
Nevertheless this does not consider versions of mongodb where this command is not yet implemented ( prior to  [4.2.10, 3.6.20, 4.7.0, 4.4.2, 4.0.21](https://jira.mongodb.org/browse/SERVER-49989) ).
A solution can be to set a minimum mongodb version ( if possible ) or handling both cases.

Orabig:
Good point ! I wasn't aware of that... :(

Le mar. 23 févr. 2021 à 18:35, Hamdi Fourati <notifications@github.com> a
écrit :

rafariossaa:
@Orabig ,
  Are you providing a PR for handling the versions and the commands ?

Orabig:
Unfortunately, I won't have time for this these days. If nobody did it
before, I'll try to do this asap... :/

rafariossaa:
ok, thanks

Orabig:
> Hello, the merged solution rely on `db.hello()` command.
> Nevertheless this does not consider versions of mongodb where this command is not yet implemented ( prior to [4.2.10, 3.6.20, 4.7.0, 4.4.2, 4.0.21](https://jira.mongodb.org/browse/SERVER-49989) ).
> A solution can be to set a minimum mongodb version ( if possible ) or handling both cases.

Fixed in https://github.com/bitnami/charts/pull/6342

rafariossaa:
Hi @Orabig ,
  Thank you for coming back and addressing this.

PR #4938, opened 2021-01-11
[bitnami/mongodb] Make VolumePermission init container consistent wit…

Orabig:
…h persistence.subPath value

**Description of the change**

When  persistence.subPath value is used, it should be used to limit the scope of the volumePermission init container. It usually means that the user don't want or can't change the rights outside of this path.

**Benefits**

- Can fix initialization error (blocking issue) if there are some read-only folders at the root of the PersistanceVolume
- Allow to only apply the right change where it is needed

**Possible drawbacks**

None. It's a bug fix

**Applicable issues**

None

**Additional information**

There is another PR pending, thus chart version has been bumped. I will resolve the conflict if both PR are accepted.

**Checklist** 
- (NA) Variables are documented in the README.md
- (NA) If the chart contains a `values-production.yaml` apart from `values.yaml`, ensure that you implement the changes in both files

### DIFF
--- a/bitnami/mongodb/Chart.yaml
+++ b/bitnami/mongodb/Chart.yaml
@@ -26,4 +26,4 @@ name: mongodb
 sources:
   - https://github.com/bitnami/bitnami-docker-mongodb
   - https://mongodb.org
-version: 10.3.6
+version: 10.3.7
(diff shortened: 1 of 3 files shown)


rafariossaa (approved):
Thank you very much for your PR !.

PR #4956, opened 2021-01-12
[bitnami/nginx-ingress-controller] Add service.targetPorts as a parameter

### LINKED ISSUE #4905: [bitnami/nginx-ingress-controller] Can't terminate HTTPS traffic on the load balancer level anymore
**Which chart**:

bitnami/nginx-ingress-controller (I'm using 7.0.10, but the affected versions are 5.1.0 and above. The last version without the issue is 5.0.11)

A couple years ago when we set up a cluster we were able to  set up HTTPS Traffic termination on the load balancer level using the following configuration (`values.yaml` file):

```
service:

  targetPorts:
    https: http

  annotations:
    service.beta.kubernetes.io/aws-load-balancer-backend-protocol: http
    service.beta.kubernetes.io/aws-load-balancer-ssl-ports: https
    service.beta.kubernetes.io/aws-load-balancer-ssl-cert: arn:aws:acm:XX:XX:certificate/XXX-XXX-XXX-XXX-XXX
```

Basically, the load balancer was accepting https connection and then were sending the traffic to the backend as http and it worked like a charm for our use case. Then, [this change](https://github.com/bitnami/charts/pull/1660/files#diff-c3ad8b69b984b05b68d54928fcb7d106eb1054edf373c3a917bbe174a4db6c9eL54) happened which removed ability to move https traffic to the backend via http. I'm not sure if it was intentional behavior, but I wasn't able to find an in-place replacement without having to rework how the traffic is being sent on the cluster. I found an [issue](https://github.com/kubernetes/ingress-nginx/issues/5935#issuecomment-668254893) which mentions exactly the same problem and the solution is to change https' port configuration to route traffic to http. Also, I found an [article](https://aws.amazon.com/premiumsupport/knowledge-center/terminate-https-traffic-eks-acm/) from AWS which also mention https => http traffic routing as a valid solution.

Basically, I'd like to get the old behavior back where we were able to configure where the traffic from the https on the load balancer will go. So that the `values.yaml` file allows to configure `targetPort` for https (and maybe http as well, but I'm not sure if there's a use case that).

Haven't figured out the workarounds which would not change the existing traffic flow and would not require rework of the existing services and ingresses.

alemorcuq:
**Description of the change**
Adds the controller's service target ports as a configurable parameter.

**Benefits**
Give more customization to users.
Align `bitnami/nginx-ingress-controller` chart with the upstream chart.

**Possible drawbacks**
None. Default values are set to the originally hardcode values.

**Applicable issues**
  - fixes #4905 
 
**Checklist**

### DIFF
--- a/bitnami/nginx-ingress-controller/Chart.yaml
+++ b/bitnami/nginx-ingress-controller/Chart.yaml
@@ -26,4 +26,4 @@ name: nginx-ingress-controller
 sources:
   - https://github.com/bitnami/bitnami-docker-nginx-ingress-controller
   - https://github.com/kubernetes/ingress-nginx
-version: 7.0.11
+version: 7.1.0
--- a/bitnami/nginx-ingress-controller/README.md
+++ b/bitnami/nginx-ingress-controller/README.md
@@ -190,6 +190,8 @@ The following tables lists the configurable parameters of the Nginx Ingress Cont
 | `service.omitClusterIP`                 | To omit the `ClusterIP` from the controller service                                      | `false`                                                 |
 | `service.ports.http`                    | Controller Service HTTP port                                                             | `80`                                                    |
 | `service.ports.https`                   | Controller Service HTTPS port                                                            | `""`                                                    |
+| `service.targetPorts.http`              | Map the controller service HTTP port                                                     | `http`                                                  |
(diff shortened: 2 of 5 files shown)


FraPazGal (approved):
LGTM!

PR #4959, opened 2021-01-12
[bitnami/kibana] enable elastic tls support if use Open Distro plugin

serg-bs:
**Description of the change**

Open Distro Security plugin require elasticsearch tls support. 
This chages add tls flag for it.

Bitnami Docker images example with OpenDistro Security plugin could be foud [here](https://github.com/serg-bs/docker-elasticsearch-opendistro-openldap) 
Bitnami EFK helm chart example could be found [here](https://github.com/serg-bs/efk/blob/master/values.yaml)

**Checklist** 


bitnami/kibana/Chart.yaml
```diff
@@ -25,4 +25,4 @@ name: kibana
 sources:
   - https://github.com/bitnami/bitnami-docker-kibana
   - https://www.elastic.co/products/kibana
-version: 6.2.3
+version: 6.2.4
```
javsalgar, line 28: Could you bump the minor version? It is adding a non-breaking feature

javsalgar (changes requested):
Thanks for the PR! Minor change

javsalgar (approved):

PR #4859, opened 2020-12-30
[bitnami/common] Make ingress rules compatible with all Kubernetes versions

javsalgar:
**Description of the change**

Based on the discussion in https://github.com/bitnami/charts/pull/4850, this PR improves the common library with the following features

- Allow setting a target Kubernetes version different from the helm or kubectl client (important for GitOps approaches that don't rely on helm or kubectl).
- Generate different ingress backend entries depending on the API version or Kubernetes version, so it is compatible with v1.20 or below.

**Benefits**

More compatibility between k8s versions

**Possible drawbacks**

We'll need to update all the charts to use these functions

**Applicable issues**

  - fixes https://github.com/bitnami/charts/issues/4791
  - fixes https://github.com/bitnami/charts/issues/4840

**Additional information**

**Checklist** 
- [x] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [x] Variables are documented in the README.md
- [x] Title of the PR starts with chart name (e.g. `[bitnami/chart]`)
- [x] If the chart contains a `values-production.yaml` apart from `values.yaml`, ensure that you implement the changes in both files

:warning: Keep in mind that if you want to make changes to the kubeapps chart, please implement them in the [kubeapps repository](https://github.com/kubeapps/kubeapps/tree/master/chart/kubeapps). This is only a synchronized mirror.

PR #4962, opened 2021-01-12
[bitnami/drupal] Major change: Update ingress

javsalgar:
**Description of the change**

This PR updates the common library with the changes from https://github.com/bitnami/charts/pull/4859. 

- Depending on the Kubernetes version it will generate the proper Ingress object
- It allows forcing a Kubernetes version so it is compatible with GitOps approaches

**Benefits**

Charts compatible with all Kubernetes versions
**Possible drawbacks**

n/a
**Applicable issues**

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/drupal/Chart.lock
+++ b/bitnami/drupal/Chart.lock
@@ -4,6 +4,6 @@ dependencies:
   version: 9.2.2
 - name: common
   repository: https://charts.bitnami.com/bitnami
-  version: 1.2.3
-digest: sha256:30ef1ad5b2e5e1c96b063b22e92188a23b3d312b1030e25283b4a156e5f3c456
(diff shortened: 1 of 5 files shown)


marcosbc (changes requested):
For major releases we should also document the change in the Notable Changes section.

I'm also failing to understand why we need to bump the major here. Is there no way to support previous parameter names?

Also, why are we making so many indentation changes to the README? I didn't see any issue with the previous one.

javsalgar:
I double checked and in the end there's only one parameter that we need to add backwards compatibility. Added it and now we don't need to bump the major. About the README, it is a table auto-formatter. It auto-formats it when it is saved. I'd rather use automated tools that manually formatting the table.

marcosbc (approved):
LGTM

PR #4963, opened 2021-01-12
[bitnami/ejbca] Make Ingress compatible with k8s 1.20

javsalgar:
**Description of the change**

This PR updates the common library with the changes from https://github.com/bitnami/charts/pull/4859. 

- Depending on the Kubernetes version it will generate the proper Ingress object
- It allows forcing a Kubernetes version so it is compatible with GitOps approaches

NOTE: bumping WordPress as there was a bug in the latest common library

**Benefits**

Charts compatible with all Kubernetes versions
**Possible drawbacks**

n/a
**Applicable issues**

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/ejbca/values.yaml
+++ b/bitnami/ejbca/values.yaml
@@ -29,6 +29,10 @@ image:
   ## Set to true if you would like to see extra information on logs
   ##
 
+## Force target Kubernetes version (using Helm capabilites if not set)
+##
+kubeVersion:
+
 ## String to partially override ebjca.fullname template (will maintain the release name)
 ##
 # nameOverride:
@@ -282,7 +286,7 @@ service:
   # extraPorts:
 
 ## Configure the ingress resource that allows you to access the
-## WordPress installation. Set up the URL
+## EJBCA installation. Set up the URL
 ## ref: http://kubernetes.io/docs/user-guide/ingress/
 ##
 ingress:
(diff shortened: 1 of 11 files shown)


bitnami/ejbca/values.yaml
```diff
@@ -302,6 +323,52 @@ ingress:
...
+  ##
+
+  ## Any additional arbitrary paths that may need to be added to the ingress under the main host.
+  ## For example: The ALB ingress controller requires a special rule for handling SSL redirection.
+  ## extraPaths:
+  ## - path: /*
+  ##   backend:
+  ##     serviceName: ssl-redirect
+  ##     servicePort: use-annotation
+  ##
```
carrodher, line 346: Not sure if we are adding an empty comment `##` at the end of a block or we usually finish the commented blocks without this

carrodher: Check for example:
```yaml
## String to partially override ebjca.fullname template (will maintain the release name)
##
# nameOverride:

## String to fully override ebjca.fullname template
##
# fullnameOverride:
```
is not
```yaml
## String to partially override ebjca.fullname template (will maintain the release name)
##
# nameOverride:
##

## String to fully override ebjca.fullname template
##
# fullnameOverride:
##
```

carrodher (approved):

PR #4967, opened 2021-01-12
[bitnami/keycloak] fix configMap cert directory permission issues

### LINKED ISSUE #4821: [bitnami/keycloak] cannot create directory '/opt/bitnami/keycloak/certs': Permission denied
**Which chart**:
keycloak 1.1.4

**Describe the bug**
When deploying helm chart for Keycloak with TLS enabled, the keycloak pod fails to start with the following error
```
mkdir: cannot create directory '/opt/bitnami/keycloak/certs': Permission denied
```

**To Reproduce**

1. Enable TLS settings by configuring `answers.yaml` as follows
```
auth.tls.enabled: true
auth.tls.jksSecret: keycloak-jks
auth.tls.keystorePassword: <password>
auth.tls.truststorePassword: <password>
```
2. Deploy helm chart per instructions

**Expected behavior**
Keycloak deploys with custom TLS configuration

- Output of `helm version`:

```
Using Rancher catalog (unsure which version of Helm is under the hood)
```

```
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.8", GitCommit:"ec6eb119b81be488b030e849b9e64fda4caaf33c", GitTreeState:"clean", BuildDate:"2020-03-12T21:00:06Z", GoVersion:"go1.13.8", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.5", GitCommit:"e0fccafd69541e3750d460ba0f9743b90336f24f", GitTreeState:"clean", BuildDate:"2020-04-16T11:35:47Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}
```

**Additional context**
Red Hat Enterprise Linux 7.8
Rancher Kubernetes 17.5.1
Rancher Management 2.4.3

jgallucci32:
Signed-off-by: jgallucci32 <john.gallucci.iv@gmail.com>

**Description of the change**
Fixes an issue with the security context of the container image not allowing certificates to be copied to the `/opt/bitnami/keycloak/certs` directory. Also fixes a mismatch in the configMap for the environment variables to use the correct cert directory (previously was using a non-existent directory). Also simplifies the deployment to use a single TLS certificate rather than having to create one per replica.

**Benefits**
Properly configures Keycloak to use a custom TLS certificate.

**Possible drawbacks**
Simplifying TLS certificate to a single cert changes the previous method of using unique certificates for each replica of a stateful set. This is an edge case and is unknown if it is actually used in Kubernetes deployments. Given this chart did not function when initially created 2 months ago, the risk is low.

**Applicable issues**

  - fixes #4821

**Additional information**

**Checklist** 


jgallucci32:
/hold

While the chart does fix the cert directory issue, the underlying problem of the container using the TLS settings correctly does not get fixed until version `11.0.3-debian-10-r63` of the container. The current version of the container in the chart is `11.0.3-debian-10-r59`. So do I need to bump the container version to `r63` or a more current version to fully fix TLS settings?

FraPazGal:
Hello @jgallucci32,

Looking into this, I think the option for multiple keystores is to ensure that internal cluster secure communication is possible. I think we can readily solve the permissions error when creating the directory with an initContainer and a mounted volume. We could transfer the code done in the container's command to the initContainer where we won't have any permissions-related error, what do you think?

Regarding the container's version used in the chart, our internal tools will update them before merging the PR, so you do not need to change it.

jgallucci32:
@FraPazGal Thanks for note about chart version. I created a separate PR to remove it from the checklist since it's handled by the internal tools https://github.com/bitnami/charts/pull/5014

If all replicas used the same certificate, shouldn't internal TLS communication still function normally? How do we validate this?

If we go down this path 3 volume mounts will be requied. The initContainer will need to mount the configMap and a second shared volume between the primary container.

FraPazGal:
Hello @jgallucci32,

I see my colleague responded to your other PR that the checklist still has value, but thanks for taking the time anyways!

> If all replicas used the same certificate, shouldn't internal TLS communication still function normally? How do we validate this?

I don't think that would be the case as each pod would have a different hostname. I''m not sure if with the default configuration the pods communicate with each other, but they would do so in a cluster configuration. I'll prefer to use 3 volumes like you said; we would solve the current issue without dropping any functionalities and keep the door open for future additions.

jgallucci32:
@FraPazGal I updated the PR to use an initContainer to configure the certs with a shared volume which is mounted to both containers. I also updated the logic of obtaining the ID of the replica to use a Kubernetes variable rather than rely on the Helm chart to inject it during the initial deployment.

bitnami/keycloak/templates/statefulset.yaml
```diff
@@ -59,9 +59,43 @@ spec:
       {{- if .Values.podSecurityContext.enabled }}
       securityContext: {{- omit .Values.podSecurityContext "enabled" | toYaml | nindent 8 }}
       {{- end }}
-      {{- if .Values.initContainers }}
+      {{- if or .Values.initContainers .Values.auth.tls.enabled }}
       initContainers: {{- include "common.tplvalues.render" (dict "value" .Values.initContainers "context" $) | nindent 8 }}
       {{- end }}
+      {{- if .Values.auth.tls.enabled }}
+        - name: init-certs
+          image: busybox
```
FraPazGal, line 67: As per bitnami's policy, we have to use images created and maintained by us in our charts. Could you modify the used image here to `bitnami/minideb`? We should also include the `securityContext` and `resources` parameters for the initContainer and have this as parameters in `values.yaml`. You could use [this other chart](https://github.com/bitnami/charts/blob/638fdadaa907a36b0304d738cb30aef366cc3255/bitnami/dokuwiki/templates/deployment.yaml#L56) as a reference.

bitnami/keycloak/templates/statefulset.yaml
```diff
@@ -59,9 +59,43 @@ spec:
...
+          imagePullPolicy: {{ .Values.image.pullPolicy }}
+          command:
+          - /bin/sh
+          - -c
+          - |-
+            ID="${MY_POD_NAME#${MY_INSTANCE_NAME}-}"
```
FraPazGal, line 73: I'm not sure `$MY_INSTANCE_NAME` is equivalent to `$fullname := include "keycloak.fullname" .`. If you take a look at the helper's logic, there are cases when these 2 envs are different and this would make `$ID` have a wrong value. I think it would be best to leave this as `$fullname`, WDYT?

FraPazGal (commented):
Hi @jgallucci32!

It is looking good! There is a couple of changes needed, could you please take a look at my comments?

bitnami/keycloak/templates/statefulset.yaml
jgallucci32: Got it. I figured as much just didn't know which image to use. Thanks for the reference.

bitnami/keycloak/templates/statefulset.yaml
jgallucci32: I'll revert back to original logic to keep it simple. I was trying to remove a dependency on Helm, but if there is a potential for conflict it's not worth the refactor.

jgallucci32:
@FraPazGal PR updated with your suggested changes.

bitnami/keycloak/templates/statefulset.yaml
```diff
@@ -59,9 +59,45 @@ spec:
       {{- if .Values.podSecurityContext.enabled }}
       securityContext: {{- omit .Values.podSecurityContext "enabled" | toYaml | nindent 8 }}
       {{- end }}
-      {{- if .Values.initContainers }}
+      {{- if or .Values.initContainers .Values.auth.tls.enabled }}
       initContainers: {{- include "common.tplvalues.render" (dict "value" .Values.initContainers "context" $) | nindent 8 }}
       {{- end }}
+      {{- if .Values.auth.tls.enabled }}
+      {{- $fullname := include "keycloak.fullname" . }}
+        - name: init-certs
```
FraPazGal, line 67: There is an `if` missing here:
```suggestion
      initContainers: 
        {{- if .Values.initContainers }}
        {{- include "common.tplvalues.render" (dict "value" .Values.initContainers "context" $) | nindent 8 }}
        {{- end }}
        {{- if .Values.auth.tls.enabled }}
        {{- $fullname := include "keycloak.fullname" . }}
        - name: init-certs
```

bitnami/keycloak/templates/statefulset.yaml
```diff
@@ -59,9 +59,45 @@ spec:
...
+                fieldPath: metadata.name
+          {{- if .Values.auth.tls.resources }}
+          resources: {{- toYaml .Values.auth.tls.resources | nindent 12 }}
+          {{- end }}
+          volumeMounts:
+          - name: certificates
+            mountPath: /certs
+          - name: shared-certs
+            mountPath: /opt/bitnami/keycloak/certs
+      {{- end }}
```
FraPazGal, line 100: Missing `end` form the previous if:

```suggestion
            mountPath: /opt/bitnami/keycloak/certs
        {{- end }}
      {{- end }}
```

bitnami/keycloak/templates/statefulset.yaml
```diff
@@ -59,9 +59,45 @@ spec:
...
+          {{- if .Values.containerSecurityContext.enabled }}
+          securityContext: {{- omit .Values.containerSecurityContext "enabled" | toYaml | nindent 12 }}
+          {{- end }}
+          command:
+          - /bin/sh
+          - -c
+          - |-
+            ID="${MY_POD_NAME#"{{ $fullname }}-"}"
+            if [[ -f "/certs/keycloak.truststore.jks" ]] && [[ -f "/certs/keycloak-${ID}.keystore.jks" ]]; then
+                mkdir -p /opt/bitnami/keycloak/certs
```
FraPazGal, line 79: As we are mounting `shared-certs` to `/opt/bitnami/keycloak/certs`, we no longer need to create the folder with `mkdir`.

bitnami/keycloak/templates/statefulset.yaml
```diff
@@ -59,9 +59,45 @@ spec:
...
+      {{- $fullname := include "keycloak.fullname" . }}
+        - name: init-certs
+          image: {{ include "keycloak.auth.tls.image" . }}
+          imagePullPolicy: {{ .Values.auth.tls.image.pullPolicy | quote }}
+          {{- if .Values.containerSecurityContext.enabled }}
+          securityContext: {{- omit .Values.containerSecurityContext "enabled" | toYaml | nindent 12 }}
+          {{- end }}
+          command:
+          - /bin/sh
+          - -c
```
FraPazGal, line 75: Our `bitnami/minideb` image gives some errors if we use `sh`, please change it to `bin/bash`:

```suggestion
          - /bin/bash
          - -ec
```

bitnami/keycloak/values.yaml
```diff
@@ -92,6 +92,36 @@ auth:
     ## Password to access the truststore when it's password-protected.
     ##
     truststorePassword: ""
+    ## Init container parameters:
```
FraPazGal, line 95: Please add these new parameters to the README.

FraPazGal (commented):
Hello @jgallucci32,

I tested this and there are a few issues left I commented on, minor stuff mostly. 

Also, please bump the chart's version to a new minor.

bitnami/keycloak/templates/statefulset.yaml
jgallucci32: Line 62 handles the logical condition for that `if` statement. Please double check, but your suggested change should not be required.

bitnami/keycloak/templates/statefulset.yaml
jgallucci32: Good catch, I will remove that.

bitnami/keycloak/templates/statefulset.yaml
jgallucci32: Line `100` ends the if statement in line `65` (I double checked and also counted the if statements). One thing I noticed is that some of the bitnami charts have an extra indent under the `container` and `initContainer` sections. If you look closely enough, the entire section is 2 spaces to the right of what actually gets deployed at runtime. I thought about correcting this but left it alone in case that was done on purpose.

bitnami/keycloak/templates/statefulset.yaml
FraPazGal: My bad, I didn't explain myself very well. The `if` I was referring to is `{{- if .Values.initContainers }}`. As it is now, whenever `auth.tls.enabled=true`, the code `{{- include "common.tplvalues.render" (dict "value" .Values.initContainers "context" $) | nindent 8 }}` is going to be executed, which is not something necessary. IMHO, separating this in 2 `ifs` leaves a clearer code and is the default practice in most of our charts.

jgallucci32: Ah ok I see now. You are correct my logic isn't perfect there. I agree two `if` statements needed to keep the extra include from generating.

bitnami/keycloak/values.yaml
jgallucci32: This has been completed.

bitnami/keycloak/templates/statefulset.yaml
```diff
@@ -62,8 +62,46 @@ spec:
...
-      initContainers: {{- include "common.tplvalues.render" (dict "value" .Values.initContainers "context" $) | nindent 8 }}
+      {{- if or .Values.initContainers .Values.auth.tls.enabled }}
+      initContainers: 
+        {{- if .Values.initContainers }}
+        {{- include "common.tplvalues.render" (dict "value" .Values.initContainers "context" $) | nindent 8 }}
+        {{- end }}
+        {{- if .Values.auth.tls.enabled }}
+        {{- $fullname := include "keycloak.fullname" . }}
+        - name: init-certs
+          image: {{ include "keycloak.auth.tls.image" . }}
```
FraPazGal, line 73: This helper is missing from `_helpers.tlp`. We should include the following:

```
{{/*
Return the proper Auth TLS image name
*/}}
{{- define "keycloak.auth.tls.image" -}}
{{- include "common.images.image" (dict "imageRoot" .Values.auth.tls.image "global" .Values.global) -}}
{{- end -}}
```

bitnami/keycloak/Chart.yaml
```diff
@@ -26,4 +26,4 @@ name: keycloak
 sources:
   - https://github.com/bitnami/bitnami-docker-keycloak
   - https://github.com/keycloak/keycloak
-version: 1.4.0
+version: 1.4.1
```
FraPazGal, line 29: I'm not sure whether to have this as a patch or bump the minor version instead. It is true that this is a fix for an existing functionality, but the changes may be extensive enough for a minor, WDYT?

FraPazGal (commented):
Hello @jgallucci32,

Just a couple of final changes before merging, could you please take a look?

bitnami/keycloak/Chart.yaml
jgallucci32: I think minor bump (as-is) since the changes focus solely on the setting `.Values.auth.tls.enabled` being true and getting it to the desired functionality. No new features have been introduced here.

bitnami/keycloak/templates/statefulset.yaml
jgallucci32: Thank you for pointing that out. This has been comitted.

FraPazGal (approved):
LGTM! Thanks for the great PR!

PR #5014, opened 2021-01-13
Remove chart bumping from checklist

jgallucci32:
Removing note from checklist to bump chart version since the internal tools will update the chart version before merging the PR.

**Description of the change**
Remove from the checklist bumping the chart version since internal tools already handle this.

**Benefits**
Prevents user from manually bumping chart in PR which can create a conflict.

**Possible drawbacks**
None known at this time.

**Applicable issues**
n/a

  - fixes #

**Additional information**

**Checklist** 

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -33,7 +33,6 @@
 request, mention that information here.-->
 
 **Checklist** <!-- [Place an '[X]' (no spaces) in all applicable fields. Please remove unrelated fields.] -->
-- [ ] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
 - [ ] Variables are documented in the README.md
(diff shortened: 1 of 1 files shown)


carrodher:
Thanks for creating this PR, but unfortunately it is non applicable due to different reasons:
- The internal tools are not executed before merging all the PRs, they are launched on-demand in certain PRs where we can double-check different use cases (for example a PR changing some comments in the _values.yaml_ is not needed to be launched in the internal CI/CD)
- Apart from that, the internal tools are going to preserve the version bumped by the user avoiding any kind of conflicts
- In the same way, the internal tools are not smart enough to determine if the version should be bumped in a patch, minor or major, this is something the owner and reviewers should manually check

Because of the above reason, I think this item in the checklist is valid and it adds more value than noise (even in some cases it can be not needed to bump the chart version, but it is not the generic case)

jgallucci32:
Ok thanks for clarifying. I'll keep an eye out on future PRs for chart version to see how this plays out.

PR #5020, opened 2021-01-13
[bitnami/mongodb] Enable psp for mongodb

yasn77:
**Description of the change**

This PR adds the ability for a user to use PSP with MongoDB chart.

**Benefits**

For users that have PSP enabled in their clusters, they also need to have a PSP resource for Pods that use volumes. Without it, the user will see errors and permission related messages.

The implementation in this PR, is pretty flexible and can allow a user to create PSP resource best for their environment.

**Possible drawbacks**

By default the PSP creation is disabled. So by default, not changes should be seen by the user. However, if PSP is created, then an understanding of how PSP work would be required.

**Applicable issues**

N/A

**Additional information**

Without this PR, using this chart in clusters that have PSP would require manual creation of the PSP and also updating the Role. Since the Role is managed by this chart (with `rbac.create` set to `true`), that would be a bad idea.

If `rbac.create` is false, then the user would also have to create the Role and RoleBinding.

**Checklist** 

### DIFF
--- a/bitnami/mongodb/templates/psp.yaml
+++ b/bitnami/mongodb/templates/psp.yaml
@@ -11,5 +11,38 @@ metadata:
   annotations: {{- include "common.tplvalues.render" ( dict "value" .Values.commonAnnotations "context" $ ) | nindent 4 }}
   {{- end }}
 spec:
+{{- if .Values.podSecurityPolicy.spec }}
 {{ .Values.podSecurityPolicy.spec | toYaml | indent 2 }}
+{{- else }}
+  allowPrivilegeEscalation: {{ .Values.podSecurityPolicy.allowPrivilegeEscalation }}
+  fsGroup:
+    rule: 'MustRunAs'
+    ranges:
+      - min: {{ .Values.podSecurityContext.fsGroup }}
+        max: {{ .Values.podSecurityContext.fsGroup }}
+  hostIPC: false
+  hostNetwork: false
+  hostPID: false
+  privileged: {{ .Values.podSecurityPolicy.privileged }}
+  readOnlyRootFilesystem: false
+  requiredDropCapabilities:
+    - ALL
+  runAsUser:
+    rule: 'MustRunAs'
+    ranges:
+      - min: {{ .Values.containerSecurityContext.runAsUser }}
+        max: {{ .Values.containerSecurityContext.runAsUser }}
+  seLinux:
+    rule: 'RunAsAny'
+  supplementalGroups:
+    rule: 'MustRunAs'
+    ranges:
+      - min: {{ .Values.containerSecurityContext.runAsUser }}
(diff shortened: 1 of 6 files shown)


alemorcuq (commented):
Thanks for the PR! Although this approach is very customizable, users need to define the whole spec in the values.yaml file. We already have many charts that support PSPs, but we are following a simpler and friendlier approach. Could you take a look at it and share your thoughts?

yasn77:
Hi @alemorcuq,

I've updated the PR with your suggestion. The PSP now follows a similar approach to Redis chart. However, I have left the ability to define a custom spec. 

Hope that's fine.

Thanks 👍🏽

bitnami/mongodb/templates/psp.yaml
```diff
@@ -0,0 +1,48 @@
...
+  labels: {{- include "common.labels.standard" . | nindent 4 }}
+    {{- if .Values.commonLabels }}
+    {{- include "common.tplvalues.render" ( dict "value" .Values.commonLabels "context" $ ) | nindent 4 }}
+    {{- end }}
+  {{- if .Values.commonAnnotations }}
+  annotations: {{- include "common.tplvalues.render" ( dict "value" .Values.commonAnnotations "context" $ ) | nindent 4 }}
+  {{- end }}
+spec:
+{{- if .Values.podSecurityPolicy.spec }}
+{{ .Values.podSecurityPolicy.spec | toYaml | indent 2 }}
```
alemorcuq, line 15: What about this? This will allow users to use templates on the spec definition.

```suggestion
{{- include "common.tplvalues.render" ( dict "value" .Values.podSecurityPolicy.spec "context" $ ) | nindent 2 }}
```

bitnami/mongodb/values-production.yaml
```diff
@@ -844,6 +844,53 @@ rbac:
   ##
   create: false
 
+## PodSecurityPolicy configuration
+## Be sure to also set rbac.create to true, otherwise Role and RoleBinding
```
alemorcuq, line 848: Do you think it makes sense to add a validation to the `_helpers.tpl` file to throw an error if `podSecurityPolicy.create` is `true` but `rbac.create` is not?

alemorcuq (changes requested):
Thanks! Please, take a look at my comments

bitnami/mongodb/templates/psp.yaml
yasn77: Oh nice, I wasn't aware of that helper

bitnami/mongodb/values-production.yaml
yasn77: Thanks, good call... Again, I wasn't aware of that.

I've updated my branch with that suggestion

alemorcuq (approved):
Thank you very much!

PR #5053, opened 2021-01-15
[bitnami/airflow]: add component wise nodeselector support

### LINKED ISSUE #4983: airflow: add support for different nodeSelectors for different airflow components
**Which chart**:
airflow chart all versions.

We need to run workers on different nodes and web/scheduler on different nodes.

Make nodeSelector variable component scoped instead of current global scoped.

Trivial for now, so we are still running all of them at same type of nodes.

**Additional context**

iamsudip:
**Description of the change**

add component wise nodeSelector support  

**Benefits**

Component (web, scheduler, worker, metrics) wise nodeSelector support

**Possible drawbacks**

Removes global nodeSelector (.Value.nodeSelector), this might affect existing chart users of global nodeSelector

**Applicable issues**

  - fixes #4983 

**Additional information**

**Checklist** 


marcosbc (changes requested):
Hi @iamsudip, this would actually be a breaking change since it would be removing support for a supported parameter (and creating new ones instead).

I'd suggest to recover the original `nodeSelector`, supporting it in conjunction with the new ones (`web.nodeSelector` and so), i.e. doing something like this:

```diff
-{{- if .Values.scheduler.nodeSelector }}
+{{- if or .Values.nodeSelector .Values.scheduler.nodeSelector }}
```

That way, this change could be considered as a "minor" change (i.e. 7.1.3 -> 7.2.0). Hope you consider it!

marcosbc (commented):
Thanks for the changes. Could you also add the global nodeSelector back to the README and values.yaml files as well?

iamsudip:
@marcosbc Added it back, have a look now. I guess we should have a patch bump, however, I did minor, shall I revert?

marcosbc (approved):
No worries, it is fine like this. So LGTM, thanks!

PR #5056, opened 2021-01-16
[bitnami/common] fix wrong include reference

Cellebyte:
**Description of the change**

The `_secrets.tpl` template had a wrong reference in it which broke the usage of the template `common.secrets.passwords.manage`. Fixing the reference to be `common.capabilities.supportsHelmVersion` fixed the issue with the template function.

**Benefits**

**Possible drawbacks**
```json
null
```

**Applicable issues**

  - fixes #5033 now properly.

**Additional information**

**Checklist** 
- [x] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [x] Variables are documented in the README.md
- [x] Title of the PR starts with chart name (e.g. `[bitnami/chart]`)
- [x] If the chart contains a `values-production.yaml` apart from `values.yaml`, ensure that you implement the changes in both files

:warning: Keep in mind that if you want to make changes to the kubeapps chart, please implement them in the [kubeapps repository](https://github.com/kubeapps/kubeapps/tree/master/chart/kubeapps). This is only a synchronized mirror.

PR #5057, opened 2021-01-16
[bitnami/common] same behavior with empty string when the secret obje…

Cellebyte:
…ct is used.

When you pass in an empty string as the existingSecret the with statement ensures that the default name is used.
If you pass in the new existingSecret object which is a dict it uses the empty  string and not use the emptyString as reference to use the default string.
Implementing another with statement ensures that the variable is set.

**Description of the change**

**Benefits**

**Possible drawbacks**

**Applicable issues**

  - fixes #

**Additional information**

Please merge after #5056 

**Checklist** 

### DIFF
--- a/bitnami/common/Chart.yaml
+++ b/bitnami/common/Chart.yaml
@@ -20,4 +20,4 @@ sources:
   - https://github.com/bitnami/charts
   - http://www.bitnami.com/
 type: library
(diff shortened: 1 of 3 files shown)


miguelaeh (approved):
LGTM!

Issue #5030, opened 2021-01-14
[bitnami/thanos] add support for mounting existing secret for grpcTLS config in Query


IbraheemAlSaady:
**Which chart**:
Current chart version: 3.3.0

**Is your feature request related to a problem? Please describe.**
My current issue is providing the config under grpcTLS as plain text. there is no option to provide an existing secret

**Describe the solution you'd like**
A solution would be to provide a configuration for defining an existing secret and mount the secret . Something like this

values.yaml sample config
```YAML
query:
  grpcTLS:
    server:
      secure: true
      cert: ""
      key: ""
      ca : ""

      ## Use existing Secret that contains the ca, key, and cert
      existingSecretName: someSecret # <<< This is a new field
      certFile: "/tls/server/cert.pem" <<< This is a new field
      keyFile: "/tls/server/key.pem" <<< This is a new field
      caFile: "/tls/server/ca.pem" <<< This is a new field
    client:
      # same as server
      ...
```
The query templates will be adjusted. 

Please let me know, I can pick this up and have a PR ready

carrodher:
Yes, sure, PRs are more than welcome. There are other places (in this chart or in others) where secrets are being used, you can use pieces of existing code as a base for your PR. The team will be happy to review it! Thanks

IbraheemAlSaady:
@carrodher I have just opened a PR

carrodher:
Thanks for creating the PR, we will check it.

IbraheemAlSaady:
PR is already merged, closing this.

PR #5058, opened 2021-01-16
[bitnami/thanos]: query support for mounting existing secret for gRPC TLS confi…

IbraheemAlSaady:
**Description of the change**
Support for mounting an existing secret that has the TLS configuration for query gRPC

**Benefits**

**Possible drawbacks**

**Applicable issues**

  - #5030 

**Additional information**

**Checklist** 


alemorcuq:
Thanks for the PR, @IbraheemAlSaady. What do you think of adding only the `existingSecret` field and using the default key names? Similar to what we do in the Harbor chart: https://github.com/bitnami/charts/blob/6361aa7a02fd8fd023ad682f8bf49765e881d9ae/bitnami/harbor/values.yaml#L387-L394

IbraheemAlSaady:
@alemorcuq I have thought of that, but I just wanted to add the possibility to add custom keys, but I guess it makes sense just to use an existingSecret field, I'll modify that then

IbraheemAlSaady:
@alemorcuq I just pushed the change

bitnami/thanos/templates/query/deployment.yaml
```diff
@@ -86,21 +86,32 @@ spec:
             - --store={{ . }}
             {{- end }}
             {{- if $query.grpcTLS.server.secure }}
+            {{- if $query.grpcTLS.server.existingSecret }}
+            - --grpc-server-tls-cert=/tls/server/tls.crt
+            - --grpc-server-tls-key=/tls/server/tls.key
+            - --grpc-server-tls-client-ca=/tls/server/ca.crt
```
alemorcuq, line 92: Thanks. I would suggest keeping the same names as in the secret that the chart creates (`cert.pem`, `key.pem`, `ca.pem`) so that we can avoid all this if-else logic. What do you think?

IbraheemAlSaady: @alemorcuq a case that I have for this is that I'm using certificates generated by cert-manager, which have keys similar to the one I provided. I agree though that these are not consistent with the `.pem`, which is already being used by the template. 

I would suggest giving the user the flexibility to specify the key names, similar to what I had in my first commit. The default keys can be the ones mentioned  (`cert.pem`, `key.pem`, `ca.pem`), but the user can modify them in case of relying on certificates generated by cert-manager

dani8art: Agree with to permit the users to specify the keys but respecting the default one.

IbraheemAlSaady: @alemorcuq should I revert back to my first commit?

dani8art: I guess we could use something like this https://github.com/bitnami/charts/pull/5189/files

IbraheemAlSaady: I like the idea, would look something like this

```yaml
existingSecret:
  name: ''
  keyMapping:
    ca-cert: ca.pem # <-- default
    cert: cert.pem # <-- default
    key: key.pem # <-- default 
     
```
     
@alemorcuq @dani8art should I continue on it to reflect the snippet above?

IbraheemAlSaady: Any updates on this? should I proceed with the above snippet?

dani8art: I like the idea! Let's start working on the PR. @alemorcuq will review it later ;)

Thank you so much!

alemorcuq: Yes, @IbraheemAlSaady, thank you very much. Sorry for the late response. @dani8art will take care of reviewing your PR.

IbraheemAlSaady:
@alemorcuq @dani8art PR is ready, let me know if there are any changes required

bitnami/thanos/templates/query/deployment.yaml
```diff
@@ -86,21 +86,32 @@ spec:
             - --store={{ . }}
             {{- end }}
             {{- if $query.grpcTLS.server.secure }}
+            {{- if $query.grpcTLS.server.existingSecret.name }}
```
dani8art, line 89: Sorry, I may not explain well. As we used in the referred PR above, I meant to use https://github.com/bitnami/charts/tree/master/bitnami/common#existingsecret in order to make this compatible with external secrets. I think it could be good to have something like

```
{{- if or $query.grpcTLS.server.existingSecret.name $query.grpcTLS.server.secure }}
- --grpc-server-tls-cert=/tls/server/{{ include "common.secrets.key" (dict "existingSecret" .Values...existingSecret "key" "tlsCert") }}
...
...
``` 

And also do the same for the client (same logic as server). Then in the volume:

```
{{- if or $query.grpcTLS.server.secure $query.grpcTLS.server.existingSecret.name }}
        - name: tls-server
          secret:
            secretName: {{ include "common.secrets.name" (dict "existingSecret" .Values...existingSecret "context" $) }}
{{- end }}
```

dani8art (changes requested):

bitnami/thanos/templates/query/deployment.yaml
IbraheemAlSaady: @dani8art I see. One thing that I'm not getting is what if the current secret is used (the one part of the template)?

IbraheemAlSaady: @dani8art I see. What about not using an existing secret and just setting up the values of ca, cert, and key, which results in a secret being created. doing  `{{ include "common.secrets.name" (dict "existingSecret" .Values...existingSecret "context" $) }}` will not result in setting the correct secret name.
Am I missing anything?

dani8art: We would like to maintain the `existingSecret` feature.

IbraheemAlSaady: @dani8art I don't understand

IbraheemAlSaady: @dani8art Do you mean we should only rely on the `existingSecret` field and remove the secrets created as part of the template?

dani8art: No, I mean you could use this common helper that was created to support both as we do in other solutions:
- helpers code ("common.secrets.name", "common.secrets.key"): https://github.com/bitnami/charts/blob/master/bitnami/common/templates/_secrets.tpl
- a generic example of use: https://github.com/bitnami/charts/tree/master/bitnami/common#existingsecret
-  a specific case: https://github.com/bitnami/charts/pull/5189

My last comment was about refactoring the `if/else` logic to not repeat some pieces of code.

IbraheemAlSaady: I see @dani8art. That means the current secrets' names in the templates need to change. If that's okay, I'll be happy to push a fix, its cleaner with the common helper.

IbraheemAlSaady: @dani8art I just pushed a draft, looks cleaner to me. Thanks for the help. Let me know what you think

dani8art (approved):
WoW, I think it looks great now!

LGTM!

Thanks for the contribution and sorry for the confusion in my comments.

PR #5083, opened 2021-01-19
[bitnami/nginx] add extraInitContainers

serg-bs:
**Description of the change**

Add extraInitContainers parameters

**Benefits**
Put python static into nginx.
I could not build nginx image with static because python collect static script requere database, But I don't want to create any shared volumes with python app.
Run `python manage.py collectstatic` in initcontainer of nginx is ideal scenario. 

**Checklist** 

### DIFF
--- a/bitnami/nginx/Chart.yaml
+++ b/bitnami/nginx/Chart.yaml
@@ -25,4 +25,4 @@ name: nginx
 sources:
   - https://github.com/bitnami/bitnami-docker-nginx
   - http://www.nginx.org
-version: 8.3.0
+version: 8.4.0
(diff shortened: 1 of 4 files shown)


dani8art (changes requested):
Hi @serg-bs Thanks! this is cool.

Just one small comment, could you rename it to `initContainers` as we propose in our standards https://github.com/bitnami/charts/tree/master/template?

serg-bs:
> Hi @serg-bs Thanks! this is cool.
> 
> Just one small comment, could you rename it to `initContainers` as we propose in our standards https://github.com/bitnami/charts/tree/master/template?

No problem. Please Notice, Looks this standart is quite new, because I copy it from another chart

dani8art (approved):
Yep, we are updating our charts to this new standard.

Thanks! LGTM

Issue #5038, opened 2021-01-15
Redmine crashloopbackoff forever


MrAmbiG:
**Which chart**:
redmine 15.0.3

**Describe the bug**
mariadb deployed.
redmine service created.
redmine pod crashloopbackoff error.

**To Reproduce**

describe pod    
```
Events:
  Type     Reason     Age                     From               Message
  ----     ------     ----                    ----               -------
  Normal   Scheduled  8m33s                   default-scheduler  Successfully assigned phabricator/redmine-67f7fc756d-d9n5c to 10.216.55.145
  Normal   Pulling    8m32s                   kubelet            Pulling image "docker.io/bitnami/redmine:4.1.1-debian-10-r189"
  Normal   Pulled     8m11s                   kubelet            Successfully pulled image "docker.io/bitnami/redmine:4.1.1-debian-10-r189"
  Normal   Created    6m31s (x3 over 8m9s)    kubelet            Created container redmine
  Normal   Started    6m31s (x3 over 8m9s)    kubelet            Started container redmine
  Warning  Unhealthy  6m11s (x10 over 8m1s)   kubelet            Readiness probe failed: Get http://IPADDRESS:3000/: dial tcp 10.42.131.148:3000: connect: connection refused
  Normal   Pulled     5m38s (x3 over 7m29s)   kubelet            Container image "docker.io/bitnami/redmine:4.1.1-debian-10-r189" already present on machine
  Warning  BackOff    3m28s (x11 over 6m50s)  kubelet            Back-off restarting failed container
```

**Expected behavior**
Redmine pod runs.

**Version of Helm and Kubernetes**:

- Output of `helm version`:

```
version.BuildInfo{Version:"v3.3.1", GitCommit:"249e5215cde0c3fa72e27eb7a30e8d55c9696144", GitTreeState:"clean", GoVersion:"go1.14.7"}
```

- Output of `kubectl version`:

```
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.2", GitCommit:"f5743093fd1c663cb0cbc89748f730662345d44d", GitTreeState:"clean", BuildDate:"2020-09-16T13:41:02Z", GoVersion:"go1.15", Compiler:"gc", Platform:"linux/amd64"}
[...]

PR #4159, opened 2020-10-29
[bitnami/redmine] MAJOR - Redmine subchart bump, also apiVersion: v2

joancafom:
**Description of the change**

This PR bumps Redmine major version since its dependencies (both PostgreSQL and MariaDB) bumped their major versions including breaking changes. More info at #3021 (PostgreSQL) and (MariaDB) #3849.

It also introduces the bitnami/commons library chart and lots of standardizations found in other solutions. More on this in additional information.

**Benefits**

Up-to-date subcomponents, standardization 

**Possible drawbacks**

Breaks backwards compatibility

**Applicable issues**
  NA

**Additional information**
- [X] commonAnnotations
- [X] commonLabels
- [X] extraVolumes
- [X] extraVolumeMounts
- [X] podAnnotations
- [X] podLabels
- [X] podSecurityContext
- [X] containerSecurityContext
- [X] affinity
- [X] podAffinityPreset
- [X] podAntiaffinityPreset
- [X] nodeAffinityPreset
- [X] sidecars
- [X] initContainers
- [X] extraEnvVars
- [X] extraEnvVarsCM
- [X] extraEnvVarsSecret
- [X] command 
- [X] args 
- [X] resources
- [X] customLivenessProbe
- [X] customReadinessProbe
- [X] existingSecret
- [x] extraDeploy
- [x] priorityClassName

**Checklist** 
- [x] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [x] Variables are documented in the README.md
- [x] Title of the PR starts with chart name (e.g. `[bitnami/chart]`)
- [x] If the chart contains a `values-production.yaml` apart from `values.yaml`, ensure that you implement the changes in both files

:warning: Keep in mind that if you want to make changes to the kubeapps chart, please implement them in the [kubeapps repository](https://github.com/kubeapps/kubeapps/tree/master/chart/kubeapps). This is only a synchronized mirror.

PR #5084, opened 2021-01-19
[bitnami/redmine] Fix Pod/Container SecurityContext(s)

juan131:

**Description of the change**

On [this PR](https://github.com/bitnami/charts/pull/4159) (major version) we modified the default pod/container security context so it treats Redmine container as a "non-root" container. This is a common practice on Bitnami containers, **but Redmine hasn't adopted it yet**. Therefore, we should ensure that, by default, the `runAsUser` and `fsGroup` are set to `0` (root user)

**Benefits**

SecurityContext properly configured

**Possible drawbacks**

None

**Applicable issues**

  - fixes https://github.com/bitnami/charts/issues/5038

**Checklist**

### DIFF
--- a/bitnami/redmine/Chart.yaml
+++ b/bitnami/redmine/Chart.yaml
@@ -36,4 +36,4 @@ name: redmine
 sources:
   - https://github.com/bitnami/bitnami-docker-redmine
   - http://www.redmine.org/
-version: 15.1.0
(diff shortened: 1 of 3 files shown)


andresbono (approved):

joancafom (approved):
LGTM, thanks for your PR!

PR #5141, opened 2021-01-19
[bitnami/harbor] Incorporate more IF statements for absent variables to handle unique use cases.

voor:
**Description of the change**

This adds a few additional if statements into variables that are normally populated for 99.9% of use cases, we just happen to fall into that 0.1% where this would be very beneficial for us.

**Benefits**

We bring in certain variables -- specifically the AWS REGION and AWS Bucket name through `extraEnvVars` options in the chart.  Currently we have a `ytt` overlay that strips these values out of the config maps and secrets, and would like to work towards deprecating any modifications we make to the Bitnami chart in favor of incorporating those hooks directly into the chart.

**Possible drawbacks**

In almost every use case these values are already populated, and in the default values file they have default values, so this should not impact any current chart users.

**Applicable issues**

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/harbor/Chart.yaml
+++ b/bitnami/harbor/Chart.yaml
@@ -34,4 +34,4 @@ sources:
   - https://github.com/bitnami/bitnami-docker-harbor-registry
   - https://github.com/bitnami/bitnami-docker-harbor-registryctl
   - https://goharbor.io/
-version: 9.4.2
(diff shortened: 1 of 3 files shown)


FraPazGal (approved):
LGTM! Thanks for the PR!

PR #5159, opened 2021-01-20
[bitnami/phabricator] NEW MAJOR: migrate to non-root + good practices

juan131:

**Description of the change**

We are working on releasing the Phabricator Docker image with a non-root approach. As part of that, we need to adapt the chart to the new requirements implicit to this new container image. As part of this change we'll also include support for new features added in the container image version.

**Benefits**

- Support for environments such as OpenShift, where a non-root UID is required to run a container.
- Also, follow new Bitnami charts good practices: lots of new customization features (container command, support for security contexts...), support for installing with existing DB, and more.

**Possible drawbacks**

- Backwards incompatibility

**Applicable issues**

None

**Checklist**


bitnami/phabricator/values.yaml
```diff
@@ -86,99 +134,258 @@ phabricatorGitSSH: false
...
+## initContainers:
+##   - name: your-image-name
+##     image: your-image
+##     imagePullPolicy: Always
+##     ports:
+##       - name: portname
+##         containerPort: 1234
+##
+initContainers: {}
+
```
marcosbc, line 318: I'm missing certificates injection support here (see Magento certificates: example in values.yaml)

bitnami/phabricator/values.yaml
```diff
@@ -189,37 +396,65 @@ ingress:
   ##
   enabled: false
 
+  ## Enable TLS configuration for the hostname defined at ingress.hostname parameter
```
marcosbc, line 399: Why is the ingress block inside the service block? Shouldn't it have its own block for itself? See Magento example

bitnami/phabricator/templates/deployment.yaml
```diff
@@ -94,44 +163,53 @@ spec:
...
-            periodSeconds: 5
+          {{- if .Values.livenessProbe.enabled }}
+          livenessProbe: {{- include "common.tplvalues.render" (dict "value" (omit .Values.livenessProbe "enabled") "context" $) | nindent 12 }}
+          {{- else if .Values.customLivenessProbe }}
+          livenessProbe: {{- include "common.tplvalues.render" (dict "value" .Values.customLivenessProbe "context" $) | nindent 12 }}
+          {{- end }}
+          {{- if .Values.readinessProbe.enabled }}
+          readinessProbe: {{- include "common.tplvalues.render" (dict "value" (omit .Values.readinessProbe "enabled") "context" $) | nindent 12 }}
+          {{- else if .Values.customReadinessProbe }}
+          readinessProbe: {{- include "common.tplvalues.render" (dict "value" .Values.customReadinessProbe "context" $) | nindent 12 }}
```
marcosbc, line 197: I have added some comments in the values.yaml regarding this, but it was decided we would define the entire readiness/livenessProbe in the template, and if users wants to customize anything apart from path/initialDelay/..., they should use customReadiness/LivenessProbe

bitnami/phabricator/values.yaml
```diff
@@ -285,65 +517,135 @@ metrics:
...
+      [client]
+      port=3306
+      socket=/opt/bitnami/mariadb/tmp/mysql.sock
+      default-character-set=UTF8
+      plugin_dir=/opt/bitnami/mariadb/plugin
+
+      [manager]
+      port=3306
+      socket=/opt/bitnami/mariadb/tmp/mysql.sock
+      pid-file=/opt/bitnami/mariadb/tmp/mysqld.pid
```
marcosbc, line 627: It would be great if we could rely on envvars/extraFlags for Phabricator specifics, to avoid having to mount the entire file.

However, I understand we already mount the my.cnf by default so maybe it makes sense anyways to do it this way.

marcosbc (changes requested):
Thanks for the contribution! Adding some comments/requests

bitnami/phabricator/templates/externaldb-secrets.yaml
```diff
@@ -1,10 +0,0 @@
-{{- if (not (or .Values.mariadb.enabled .Values.externalDatabase.existingSecret)) }}
```
marcosbc, line 1: Doesn't this file make sense anymore?

marcosbc (commented):
Also missing pv.yaml file

bitnami/phabricator/values.yaml
```diff
@@ -86,99 +134,258 @@ phabricatorGitSSH: false
...
+    path: /auth/
+    port: http
+    httpHeaders:
+      - name: Host
+        value: "{{ include \"phabricator.host\" . }}"
+  initialDelaySeconds: 30
+  periodSeconds: 10
+  timeoutSeconds: 5
+  failureThreshold: 6
+  successThreshold: 1
```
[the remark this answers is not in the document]
juan131: Sorry I cannot find in that thread that we agreed on only adding the specific fields..
Also, IMHO all the parameters included there are likely to be cuztomizable.

bitnami/phabricator/values.yaml
juan131: The ingress block is not inside the service block. Please double-check

bitnami/phabricator/values.yaml
juan131: Do we really need this? In K8s, containers should rely on Ingress to manage TLS encryption.

bitnami/phabricator/values.yaml
juan131: We already mount a config file in MariaDB by default. This just replaces the original one with one that meets PH requirements.
Also, this is specific configuration of MySQL/MariaDB

bitnami/phabricator/templates/externaldb-secrets.yaml
juan131: Ups! Removed by mistake

juan131:
Why would I need a `pv.yaml` @marcosbc ? The README.md clearly says
```
## Prerequisites

- Kubernetes 1.12+
- Helm 3.1.0
- PV provisioner support in the underlying infrastructure
- ReadWriteMany volumes for deployment scaling
```
It's stated that a PV provisioner is required. Using it, I don't see any advantage on defining the PV.

bitnami/phabricator/values.yaml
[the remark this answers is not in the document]
marcosbc: It seems like we discussed internally, just noticed that, sorry. The thing is that we do not want to be 100% flexible with `livenessProbe/readinessProbe`, since that's what `customLivenessProbe/customReadinessProbe` is for.

Check this issue/PR for more info: #2022

Also ccing @javsalgar in case you have questions about this.

bitnami/phabricator/values.yaml
marcosbc: I see, probably I was confused with the indentation or something. Nevermind then.

bitnami/phabricator/values.yaml
marcosbc: Nvm then, this should be fine

bitnami/phabricator/values.yaml
marcosbc: In Redmine it was found useful: #2774. I think it makes sense:

>Insert a custom certificate into the pod, allowing for a trusted
>connection using a loadbalancer, or between the ingress and the pod.

marcosbc:
Regarding the PV, do you think it would make sense to remove it from all Bitnami charts then?

juan131:
> Regarding the PV, do you think it would make sense to remove it from all Bitnami charts then?

I think so.. Idk if this comes from some user contribution, but honestly it's quite a basic thing in a K8s cluster to have a PV Provisioner to guarantee you can use a PVC and rely on the cluster to generate the needed PVs.

bitnami/phabricator/values.yaml
juan131: I'd skip it for now

andresbono (approved):
LGTM

PR #5167, opened 2021-01-20
[bitnami/keycloak] Add keycloak-config-cli job

### LINKED ISSUE #5146: [bitnami/keycloak] Configuration as a service CaaS
**Which chart**:

bitnami keycloak version 1.2.0

Keycloak is very difficult and time consuming service to implement and once implemented maintaining is a hassle. By configuring this service just once will allow for fast setup and more reliability. Currently the configuration of bitnami keycloak sets up the server which will have to be configured by hand. 

My solution suggest to use Configuration as as service CaaS which allow the user to provide a file either json or yaml file with all keycloak configuration as they want it. This means the code will execute exactly each and every time. The way we propose to do this is actually very simple. By enabling this project [Here](https://github.com/adorsys/keycloak-config-cli) as the bitnami keycloak helm chart dependency and passing the configs either as environment variables, providing key/value, or directly passing json or yml file.

**Additional context**

Please Note:

[Example configuration file in json](https://github.com/adorsys/keycloak-config-cli/blob/master/contrib/example-config/moped.json) 

[Extensive list of examples ](https://github.com/adorsys/keycloak-config-cli/tree/master/src/test/resources/import-files) 

[Project dependency](https://github.com/adorsys/keycloak-config-cli/tree/master/contrib/charts/keycloak-config-cli)

iamaverrick:
**Description of the change**

Add a job that executes bitnami/keycloak-config-cli, a Configuration As Code solution.

Testing done: https://github.com/bitnami/charts/pull/5167#issuecomment-791404435

**Applicable issues**

  - fixes #5146

**Checklist** 


migruiz4:
Hi @imaverrick,

After discussing it with the team, we consider it would be more appropriate for `keycloak-config-cli` to be integrated with the `keycloak` chart instead of as an independent chart.

This could be done by moving your templates into the Keycloak chart adding `{{- if .Values.configCli.enabled }}` at the beginning.

For the values, they would be moved into `bitnami/keycloak/values.yml` under `configCli` as follows:

```yaml
configCli:
  enabled: false
  #keycloak-config-cli values
  ...
```

If you have any questions don't hesitate to let us know.

iamaverrick:
His @migruiz4 ,

this sounds good and i can get started i was just following what i was told by @carrodher at [Issue](https://github.com/bitnami/charts/issues/5146)

carrodher:
Yes, this is the process to add any Helm Chart to the Bitnami catalog; but after taking a look at this PR, we think it is better the approach suggested by @migruiz4.

The chart in this PR doesn't meet the requirements above mentioned (chart using BitNami scaffolding and guidelines, container image in the catalog, etc) so once we reviewed the PR, it seems more easy to add the templates directly inside the keycloak chart than support a different chart to be used as a subchart. As it is an already existing chart that was just copied here it is not close to follow the same standards than the rest of Bitnami charts.

In the same way, we think the solution add more value as a feature bundled in the keycloak chart than as a separate chart

iamaverrick:
@carrodher thank you for your prompt response and clarifications. i totally agree with you im extremely excited to see this feature implemented. 

thank you

bitnami/keycloak/templates/job.yaml
```diff
@@ -0,0 +1,81 @@
...
+    {{- end }}
+    spec:
+      {{- with .Values.image.pullSecrets }}
+      imagePullSecrets:
+      {{- toYaml . | nindent 8 }}
+      {{- end }}
+      restartPolicy: Never
+      containers:
+        - name: keycloak-config-cli
+          image: "{{ .Values.image.repository }}:{{ tpl .Values.image.tag $ }}"
```
javsalgar, line 34: This should reference the .Values.KcCasc.image section. Could you create a helper just like we do with statefulset.yaml?

bitnami/keycloak/templates/job.yaml
```diff
@@ -0,0 +1,81 @@
...
+          imagePullPolicy: {{ .Values.image.pullPolicy }}
+          {{- with .Values.resources }}
```
javsalgar, line 36: Use the .Values.KcCasc section. Apply to the rest of the yaml

bitnami/keycloak/templates/job.yaml
```diff
@@ -0,0 +1,81 @@
+---
+{{- if .Values.KcCasC.enabled }}
+apiVersion: batch/v1
```
javsalgar, line 3: Could you try adding the same type of standard values that we can see in statefulset.yaml?

bitnami/keycloak/templates/realms.yaml
```diff
@@ -0,0 +1,17 @@
...
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ template "keycloak-config-cli.fullname" . }}-config-realms
+  labels:
+  {{- include "keycloak-config-cli.labels" . | nindent 4 }}
+data:
+  {{- range $name, $config := .Values.config }}
+    {{- if hasKey $config "file" }}
+  {{ $name }}.json: "{{ tpl ($.Files.Get $config.file) $ | b64enc }}"
```
javsalgar, line 12: We try not using the Files function as it caused issues in the past. We prefer using either the inline, the existingConfig or the extraDeploy value.

javsalgar (commented):
Please check my comments

bitnami/keycloak/values.yaml
```diff
@@ -655,3 +655,78 @@ externalDatabase:
   ## Database name
   ##
   database: bitnami_keycloak
+
+##
+## Below is the implementation of Keycloak Configuration as Code.
+##
+KcCasC:
```
andresbono, line 662: What do you think about using a more descriptive name and following camel-case style? `keycloakConfigAsCode` maybe?

bitnami/keycloak/values.yaml
```diff
@@ -655,3 +655,78 @@ externalDatabase:
...
+  ## Whether to enable Keycloak Configuration as Code.
+  ##
+  enabled: false
+
+  image:
+    repository: docker.io/adorsys/keycloak-config-cli
+    tag: "v{{ .Chart.AppVersion }}-11.0.3"
```
andresbono, line 669: We have released the bitnami/keycloak-config-cli container image, please update the values and test that image on your side. Let us know about the results.

```suggestion
  ## Bitnami keycloak-config-cli image version
  ## ref: https://hub.docker.com/r/bitnami/keycloak-config-cli/tags/
  ##
  image:
    registry: docker.io
    repository: bitnami/keycloak-config-cli
    tag: 3.1.0-debian-10-r0
```

andresbono:
Hi @iamaverrick, we are going test and push the changes that we consider necessary to have this PR merged. You allowed maintainers to edit this PR, so it shouldn't be any issue for us to include new commits.

iamaverrick:
@andresbono yes of course, I’m sorry I had a lot of free time before but became busy with some projects. Some ideas to include because I have used the keycloak config chart is to use the environment variables provided by the main chart in the sub chart as most of the variables will be used anyways. We should only have to provide the configuration files to the subchart. Another thing worth noting is that the sub chart is forcing only json format. We need to be able to provide either json or yml. If you look at the secrets.yml it explicitly says json.

andresbono:
Thanks for the suggestions, @iamaverrick.

> to use the environment variables provided by the main chart in the sub chart as most of the variables will be used anyways. We should only have to provide the configuration files to the subchart.

Correct, that is how this is implemented. 👍 

> Another thing worth noting is that the sub chart is forcing only json format. We need to be able to provide either json or yml. If you look at the secrets.yml it explicitly says json.

Done, see the example in the `values.yaml`: https://github.com/bitnami/charts/blob/a87c6fd8d96d247c00543662cd32d14d6cf4d0ee/bitnami/keycloak/values.yaml#L847-L861

bitnami/keycloak/values.yaml
```diff
@@ -729,3 +729,138 @@ externalDatabase:
...
+  ##     {
+  ##       "realm": "realm1",
+  ##       "clients": []
+  ##     }
+  ##   files/realm2.yaml:
+  ##   realm3.yaml: |
+  ##     realm: realm3
+  ##     clients: []
+  ##
+  config: {}
```
juan131, line 861: Use `configuration` for consistency with the already existing parameter. You'll have to update the job and README.md accordingly.
```suggestion
  ## configuration:
  ##   realm1.json: |
  ##     {
  ##       "realm": "realm1",
  ##       "clients": []
  ##     }
  ##   files/realm2.yaml:
  ##   realm3.yaml: |
  ##     realm: realm3
  ##     clients: []
  ##
  configuration: {}
```

bitnami/keycloak/values.yaml
```diff
@@ -729,3 +729,138 @@ externalDatabase:
...
+
+  ## ConfigMap with keycloak-config-cli configuration
+  ## NOTE: This will override keycloakConfigCli.config
```
juan131, line 864: Following the previous suggestion:
```suggestion
  ## NOTE: This will override keycloakConfigCli.configuration
```

bitnami/keycloak/values.yaml
```diff
@@ -729,3 +729,138 @@ externalDatabase:
...
+    ## Defaults to 'Always' if image tag is 'latest', else set to 'IfNotPresent'
+    ## ref: http://kubernetes.io/docs/user-guide/images/#pre-pulling-images
+    ##
+    pullPolicy: IfNotPresent
+    ## Optionally specify an array of imagePullSecrets.
+    ## Secrets must be manually created in the namespace.
+    ## ref: https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/
+    ##
+    # pullSecrets:
+    #   - myRegistryKeySecretName
```
juan131, line 758:
```suggestion
    ## Optionally specify an array of imagePullSecrets.
    ## Secrets must be manually created in the namespace.
    ## ref: https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/
    ## e.g:
    ## pullSecrets:
    ##   - myRegistryKeySecretName
    ##
    pullSecrets: []
```

bitnami/keycloak/templates/_helpers.tpl
```diff
@@ -31,11 +31,29 @@ Return the proper Keycloak image name
...
+Return the keycloak-config-cli configuration configmap.
+*/}}
+{{- define "keycloak.keycloakConfigCli.configmapName" -}}
+{{- if .Values.keycloakConfigCli.existingConfigmap -}}
+    {{- printf "%s" (tpl .Values.keycloakConfigCli.existingConfigmap $) -}}
+{{- else -}}
+    {{- printf "%s-keycloak-config-cli-configmap" (include "common.names.fullname" .) -}}
+{{- end -}}
+{{- end -}}
+
```
juan131, line 51: I'dd a macro such as the one below:

```
{{/*
Return true if a configmap object should be created for keycloak-config-cli
*/}}
{{- define "keycloak.keycloakConfigCli.createConfigmap" -}}
{{- if and .Values.keycloakConfigCli.enabled .Values.keycloakConfigCli.configuration (not .Values.keycloakConfigCli.existingConfigmap)
    {{- true -}}
{{- end -}}
{{- end -}}
```

Then you can use later in the **keycloak-config-cli-configmap.yaml** and **keycloak-config-cli-job.yaml**

bitnami/keycloak/templates/keycloak-config-cli-configmap.yaml
```diff
@@ -0,0 +1,21 @@
+{{- if and .Values.keycloakConfigCli.enabled .Values.keycloakConfigCli.config (not .Values.keycloakConfigCli.existingConfigmap) }}
```
juan131, line 1: Following the previous suggestion:
```suggestion
{{- if (include "keycloak. keycloakConfigCli .createConfigmap" .) }}
```

bitnami/keycloak/templates/keycloak-config-cli-job.yaml
```diff
@@ -0,0 +1,111 @@
...
+  template:
+    metadata:
+      labels: {{- include "common.labels.standard" . | nindent 8 }}
+        app.kubernetes.io/component: keycloak-config-cli
+        {{- if .Values.keycloakConfigCli.podLabels }}
+        {{- include "common.tplvalues.render" (dict "value" .Values.keycloakConfigCli.podLabels "context" $) | nindent 8 }}
+        {{- end }}
+      {{- if .Values.keycloakConfigCli.podAnnotations }}
+      annotations: {{- include "common.tplvalues.render" (dict "value" .Values.keycloakConfigCli.podAnnotations "context" $) | nindent 8 }}
+      {{- end }}
```
juan131, line 28: Add the checksum to ensure the job is updated in the configuration changes
```suggestion
      annotations:
        {{- if (include "keycloak. keycloakConfigCli.createConfigmap" .) }}
        checksum/configuration: {{ include (print $.Template.BasePath "/keycloak-config-cli-configmap.yaml") . | sha256sum }}
        {{- end }}
        {{- if .Values.keycloakConfigCli.podAnnotations }}
        {{- include "common.tplvalues.render" (dict "value" .Values.keycloakConfigCli.podAnnotations "context" $) | nindent 8 }}
        {{- end }}
```

juan131 (changes requested):
Wow, great PR! We really appreciate it.

Please check my suggestions when you have a chance

andresbono:
# Manual testing done

### Simple realms

```console
$ mkdir -p files && echo '{"realm":"realm2", "clients": []}' >files/realm2.json
$ cat <<EOF | helm install keycloak . -f -
keycloakConfigCli:
  enabled: true

  config:
    realm1.json: |
      {
        "realm": "realm1",
        "clients": []
      }
    files/realm2.json:
    realm3.yaml: |
      realm: realm3
      clients: []
EOF
```
`/auth/admin/master/console/#/realms`
image

## Testing realm from upstream [moped.json](https://github.com/adorsys/keycloak-config-cli/blob/main/contrib/example-config/moped.json)

```console
$ cat <<EOF | helm install keycloak . -f -
keycloakConfigCli:
  enabled: true

  extraEnvVars:
    - name: IMPORT_PATH
      value: "https://raw.githubusercontent.com/adorsys/keycloak-config-cli/v3.1.0/contrib/example-config/moped.json"
EOF
```

`/auth/admin/master/console/#/realms`

image

juan131 (approved):
Thanks for addressing all the suggestions! LGTM

migruiz4 (approved):
Super cool feature!
LGTM, thank you very much for the contribution!

PR #5184, opened 2021-01-22
[bitnami/rabbitmq]: use managerPortName in ingress target port

1995parham:
In the ingress route, the target port is hard-coded and change the service port name disables the ingress route.
As an example the following value changes the service port name for the manager port:

```yml
service:
  managerPortName: stats
```

Then because of the hard-coded value in the ingress route it will be disabled.


javsalgar (approved):
Thank you so much for the PR! LGTM

PR #5189, opened 2021-01-22
[bitnami/keycloak] feat: add existing secret feature

### LINKED ISSUE #5173: [bitnami/keycloak] externalDatabase password could be fetched from existing secret
**Which chart**:
keycloak v1.2.0

when using externalDatabase, normally we already have the secret and configmap which contains all information to connect to that database. However, the current chart doesn't support a way to specify those configmap and secret directly but putting the values into the values.yaml. 

As I can see from the common chart, there is a way to use this existing secret already (https://github.com/bitnami/charts/tree/master/bitnami/common#existingsecret). It would be nice to have this feature.

The second way of doing this is checking in the statefulset if another value (e.g .Values.externalDatabase.passwordFromExistingSecret.[secretName, secretKey]) then it will use this value; otherwise, use the current envVar KEYCLOAK_DATABASE_PASSWORD

**Additional context**
N/A

What do you think which way is better?

dani8art:

**Description of the change**

Add the possibility to use existing secrets for managing passwords

**Benefits**

We can use external tools for managing secrets.

**Possible drawbacks**

N/A

**Applicable issues**

  - fixes #5173

**Additional information**

N/A

**Checklist** 


hieulengoc:
This is incredibly fast, thank you very much! I'll update our pipeline as soon as this get merged.

javsalgar (approved):

hieulengoc:
Hi @dani8art 
There is still one bug for checking the existing password, whenever I run the helm upgrade command, it will ask for the password:

Error: UPGRADE FAILED: template: keycloak/charts/keycloak/templates/NOTES.txt:80:4: executing "keycloak/charts/keycloak/templates/NOTES.txt" at <include "common.errors.upgrade.passwords.empty" (dict "validationErrors" $passwordErrors "context" $)>: error calling include: template: keycloak/charts/keycloak/charts/common/templates/_errors.tpl:18:48: executing "common.errors.upgrade.passwords.empty" at <fail>: error calling fail:
PASSWORDS ERROR: you must provide your current passwords when upgrade

This makes sense when the password is auto-generated by the chart, but it doesn't make sense when we ref the password from our existing secret. And this breaks the pipeline as well. Of course, we can specify anything in the pipeline because it will not be used anyway; however, it doesn't necessarily. Could you please adapt this one as well?

Thanks a lot for this PR.
Best,
Hieu

dani8art:
Hi @hieulengoc that's completely right, I've just created a PR trying to fix it. 

Thanks!

PR #5204, opened 2021-01-23
fix missing .image

emilweth:


carrodher (changes requested):
Thanks for your contribution. The changes LGTM but it is also needed to bump the `version` field in the _Chart.yaml_ even for minor changes like this one

bitnami/redis-cluster/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 annotations:
   category: Database
 apiVersion: v2
-appVersion: 6.0.10
+appVersion: 6.0.11
```
carrodher, line 4: This is the application version (the redis one), the chart version is the `version` field at the end of this file. I am going to edit the file by myself and it should be ready to be landed

emilweth: Yep sorry, I did this too quickly
Thanks!

carrodher (approved):

PR #5213, opened 2021-01-25
[bitnami/postgresql] Add chmod `-R` option in init-chmod-data init-container

tyg03485:
Context.
Database Pod CrashLoopBackOff because wrong permission

After setting `volumePermissions.enabled` true, It still has problem.
because`/bitnami/postgresql/data/data` has permission problem in my case.

FYI) I didn't touch any variable about `persistence.mountPath` since first deploy

### DIFF
--- a/bitnami/postgresql/Chart.lock
+++ b/bitnami/postgresql/Chart.lock
@@ -1,6 +1,6 @@
 dependencies:
 - name: common
   repository: https://charts.bitnami.com/bitnami
-  version: 1.2.3
-digest: sha256:ef82cbc1e15eb074ed05cdec22538cacb27b5b2ad706ba4c5b18ffb1a14f701c
-generated: "2021-01-04T17:59:33.709728131Z"
+  version: 1.3.7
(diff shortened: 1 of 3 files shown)


migruiz4:
Hi @tyg03485 ,

Setting permissions 700 recursively for all the files under the data volume may solve the issue, but I'd say it is not the best way to proceed as you'd be giving additional permissions to all the other files and directories.

If you provide some more details about your case I may be able to help you find an alternative solution to your problem.

tyg03485:
I agree with you.  Changing all permission under the data volume would be dangerous or not intended.
How about just add `{{ .Values.persistence.init-chmod-data.extraCommands }}` for init-container?

When somebody wanna add some command which will be run on init-container, it will be easily to handle it.

migruiz4:
Could you please open a ticket describing your issue?

Any useful information you can provide in there would be helpful: configuration, what type of PVC are you using, error messages, and logs.

We test the chart on all major cloud platforms and we haven't found this issue. I think we should focus on finding the root cause of the permission issue and when solved then add the fix to the chart.

PR #5233, opened 2021-01-26
[bitnami/wordpress] Allow post-init scripts out of the box

### LINKED ISSUE #5231: [bitnami/wordpress] Usage and syntax for customPostInitScripts
Hi,
I'm trying to figure out how customPostInitScripts can be used to run some simple post config in Wordpress like creating additional users or adding plugins and themes to the running container.

First of all I'd like to understand if this is the best way to complete those tasks or maybe you can suggest a better way.
I thought that a user can easily added by adding:

```yaml
customPostInitScripts:
    adduser.sh: |
        #!/bin/bash
        wp user create bob bob@example.com --role=author
```
but apparently the script is not run at all, so I assume I'm missing a piece there.
I'm currently deploying the helm chart through kubeapp, by editing the yaml before the deployment.

carrodher:
**Description of the change**

At this moment the `customPostInitScripts` feature was implemented (https://github.com/bitnami/charts/pull/4373) but without a proper user experience since apart from creating the scripts in the values, it is needed to perform more changes to use the created ConfigMap as a volume mounted in the container.

**Benefits**

This PR improves the UX when using `customPostInitScripts`; with the new changes creating the script should be enough

**Manual test**

- Modify _values.yaml_:
```yaml
customPostInitScripts:
  adduser.sh: |
    #!/bin/bash
    echo 'I am creating a new user!!'
    wp user create bob bob@example.com --role=author
```

- Deploy helm chart:
```console
$ helm install . --generate-name -f values.yaml
```

- Check logs:
```console
$ kubectl logs -f chart-1611650204-wordpress-6b94f8bcb4-nwz26
...
nami    INFO  wordpress successfully initialized
INFO  ==> Starting gosu...
INFO  ==> Executing /docker-entrypoint-init.d/..2021_01_26_08_36_49.469001149/adduser.sh
I am creating a new user!!
Success: Created user 2.
Password: u7vtT&$H4$##WOl9JSM1Wd6b
[Tue Jan 26 08:37:59.221009 2021] [ssl:warn] [pid 88] AH01909: www.example.com:8443:0 server certificate does NOT include an ID which matches the server name
[Tue Jan 26 08:37:59.221639 2021] [ssl:warn] [pid 88] AH01909: www.example.com:8443:0 server certificate does NOT include an ID which matches the server name
[Tue Jan 26 08:37:59.253140 2021] [ssl:warn] [pid 88] AH01909: www.example.com:8443:0 server certificate does NOT include an ID which matches the server name
...
```
the _adduser.sh_ script defined in the _values.yaml_ is executed during the Wordpress initialization.

**Applicable issues**

  - fixes #5231

**Checklist** 


bitnami/wordpress/Chart.yaml
```diff
@@ -31,4 +31,4 @@ name: wordpress
 sources:
   - https://github.com/bitnami/bitnami-docker-wordpress
   - http://www.wordpress.com/
-version: 10.4.7
+version: 10.4.8
```
miguelaeh, line 34: I would bump the minor since you are adding new functionality. Even if the configmap was created before it was not being added to the deployment. so IMHO this PR is the one adding a new feature

miguelaeh (approved):
LGTM!

PR #5244, opened 2021-01-26
[bitnami/airflow] fix: not to use redis when KubernetesExecutor

### LINKED ISSUE #5179: airflow install KubernetesExecuter redis.enable=false not working
**Which chart**:
airflow 7.1.3

**Describe the bug**
When I run helm install to use the KubernetesExecutor with the seetings per the docs, I get the following error.

```
helm -n airflow install airflow bitnami/airflow --set service.type=LoadBalancer --set executor=KubernetesExecutor --set redis.enabled=false --set rbac.create=true --set serviceAccount.create=true

Error: template: airflow/templates/config/secret-external-redis.yaml:15:54: executing "airflow/templates/config/secret-external-redis.yaml" at <b64enc>: invalid value; expected string
```

**To Reproduce**

1. Run 'helm -n airflow install airflow bitnami/airflow --set service.type=LoadBalancer --set executor=KubernetesExecutor --set redis.enabled=false --set rbac.create=true --set serviceAccount.create=true'

**Expected behavior**
I expect airflow to deploy using the KubernetesExecutor

- Output of `helm version`: 

```
(v3.3.0
```

```
Client Version: version.Info{Major:"1", Minor:"18+", GitVersion:"v1.18.2-6+38ac483e736488", GitCommit:"38ac483e736488517dd754156441b89e0b2060e2", GitTreeState:"clean", BuildDate:"2020-07-07T13:54:17Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.5+vmware.1", GitCommit:"1abde2b816bac0da89c6c71360799c681094ca0e", GitTreeState:"clean", BuildDate:"2020-06-29T22:31:51Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}
```

**Additional context**
I believe the issue is with the logic in the template. When redis.enabled is set to false the error occurs. When set to true it errors.

dani8art:

**Description of the change**

Add a condition to not use Redis when the executor is `KubernetesExecutor`

**Benefits**

We can use `KubernetesExecutor`

**Possible drawbacks**

N/A

**Applicable issues**

  - fixes #5179

**Additional information**

N/A

**Checklist** 

### DIFF
--- a/bitnami/airflow/Chart.yaml
+++ b/bitnami/airflow/Chart.yaml
@@ -32,4 +32,4 @@ name: airflow
 sources:
   - https://github.com/bitnami/bitnami-docker-airflow
   - https://airflow.apache.org/
-version: 8.0.0
+version: 8.0.1
--- a/bitnami/airflow/README.md
+++ b/bitnami/airflow/README.md
@@ -47,7 +47,7 @@ The command removes all the Kubernetes components associated with the chart and 
 The following tables lists the configurable parameters of the Airflow chart and their default values.
 
 | Parameter                 | Description                                     | Default                                                 |
-|---------------------------|-------------------------------------------------|---------------------------------------------------------|
+| ------------------------- | ----------------------------------------------- | ------------------------------------------------------- |
 | `global.imageRegistry`    | Global Docker image registry                    | `nil`                                                   |
 | `global.imagePullSecrets` | Global Docker registry secret names as an array | `[]` (does not add image pull secrets to deployed pods) |
 | `global.storageClass`     | Global storage class for dynamic provisioning   | `nil`                                                   |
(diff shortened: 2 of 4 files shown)


miguelaeh (commented):
I have a question, in this case, since we are not going to configure Redis, shoudn't we also change the `redis.enabled` condition to not deploy it?
Also, I think this change could modify the deployment for someone that had that configuration, maybe bumping the minor would be more appropiate even if it is not a new feature. WDYT?

dani8art:
Oh I see, do you mean to have something like `or redis.enabled (not (eq executor "KubernetesExecutor"))`, right?

miguelaeh:
Exactly, could you take a look?

dani8art:
>Condition - The condition field holds one or more YAML paths (delimited by commas). If this path exists in the top parent's values and resolves to a boolean value, the chart will be enabled or disabled based on that boolean value. Only the first valid path found in the list is evaluated and if no paths exist then the condition has no effect.
> Tags - The tags field is a YAML list of labels to associate with this chart. In the top parent's values, all charts with tags can be enabled or disabled by specifying the tag and a boolean value.

Regarding Helm chart docs, I think it is not possible to provide this kind of condition. Said that I think the best approach would be to add some lines to the documentation so the user will know that Redis is not needed when `executor=KubernetesExecutor`.

PR #5263, opened 2021-01-27
Update ingress.yaml

melakonda:
Ingress servicePort is coming as http. I changed it to {{ .Values.service.port }}

**Description of the change**

**Benefits**

**Possible drawbacks**

**Applicable issues**

  - fixes #

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/grafana/templates/ingress.yaml
+++ b/bitnami/grafana/templates/ingress.yaml
@@ -29,7 +29,7 @@ spec:
             {{- if eq "true" (include "common.ingress.supportsPathType" .) }}
             pathType: {{ .Values.ingress.pathType }}
             {{- end }}
-            backend: {{- include "common.ingress.backend" (dict "serviceName" (include "common.names.fullname" .) "servicePort" "http" "context" $)  | nindent 14 }}
+            backend: {{- include "common.ingress.backend" (dict "serviceName" (include "common.names.fullname" .) "servicePort" "{{ .Values.service.port }}" "context" $)  | nindent 14 }}
     {{- end }}
     {{- range .Values.ingress.extraHosts }}
     - host: {{ .name | quote }}
@@ -39,7 +39,7 @@ spec:
             {{- if eq "true" (include "common.ingress.supportsPathType" $) }}
             pathType: {{ default "ImplementationSpecific" .pathType }}
             {{- end }}
-            backend: {{- include "common.ingress.backend" (dict "serviceName" (include "common.names.fullname" $) "servicePort" "http" "context" $) | nindent 14 }}
+            backend: {{- include "common.ingress.backend" (dict "serviceName" (include "common.names.fullname" $) "servicePort" "{{ .Values.service.port }}" "context" $) | nindent 14 }}
     {{- end }}
   {{- if or .Values.ingress.tls .Values.ingress.extraTls }}
   tls:


bitnami/grafana/templates/ingress.yaml
```diff
@@ -29,7 +29,7 @@ spec:
             {{- if eq "true" (include "common.ingress.supportsPathType" .) }}
             pathType: {{ .Values.ingress.pathType }}
             {{- end }}
-            backend: {{- include "common.ingress.backend" (dict "serviceName" (include "common.names.fullname" .) "servicePort" "http" "context" $)  | nindent 14 }}
+            backend: {{- include "common.ingress.backend" (dict "serviceName" (include "common.names.fullname" .) "servicePort" "{{ .Values.service.port }}" "context" $)  | nindent 14 }}
```
javsalgar, line 32: This change is not necessary. Using the named port should be enough. Could you explain the issue you are having?

javsalgar (changes requested):

bitnami/grafana/templates/ingress.yaml
1993Saurabh: Hi @javsalgar ,

Could you please explain why using the port named is enough in the servicePort. Due to this issue the port number in the ingress file is coming like this and sadly I am unable to open the rabbitmq management console:

image

But once I manually fixed this issue in my local machine and redeployed the ingress, it got changed and was able to open the management console after that:

image


So could you please explain why this is not an issue in the Chart itself and there is no way to override this value.

javsalgar: When setting an ingress rule, the named port is enough for Ingress for recognizing the service port. I just deployed it with ingress.enabled=true and could access without problems.

image

1993Saurabh: Hi @javsalgar ,

Sadly this setting is not working with the Azure Application Gateway, you probably tested it with the nginx-controller. If it is then can you please help how can I do the same with Azure App Gateway? 

P.S: Azure App Gateway does not allow text in the Port.

javsalgar: Hi,

If that's the case, we should update all of our Ingress rules. Do you have a link where they confirm that incompatibility with named ports? I'll need it to forward it to the rest of the engineering team for evaluation.

1993Saurabh: See if this link can help : [Link](https://github.com/Azure/application-gateway-kubernetes-ingress/blob/master/docs/annotations.md)

But we definitely need the updation as Azure Application Gateway is right away discarding the named port at the moment.

javsalgar: I opened a ticket so they can confirm https://github.com/Azure/application-gateway-kubernetes-ingress/issues/1139

1993Saurabh: okay thanks

javsalgar: Still no answer. Let's give it some extra days before continuing with this.

javsalgar: I pinged again

1993Saurabh: Yeah, and I opened the same question [here](https://stackoverflow.com/questions/66204759/does-azure-application-gateway-ingress-controller-supports-named-ports) as well

javsalgar: Let's do something to not block you. For the time being, I will accept this change in the chart, but I won't change it in the rest until we have a confirmation from Azure about this. Could you bump the minor version of the chart?

1993Saurabh: @javsalgar could you please tell me the version of the chart which you bumped? We need to freeze the version in our project.

javsalgar: Hi! We didn't bump the version, could you do it in this PR? The minor version would be required.

PR #5293, opened 2021-01-27
[bitnami/spring-cloud-dataflow] External Kafka support

### LINKED ISSUE #5169: [bitnami/spring-cloud-dataflow] Fully support use of external Kafka
**Which chart**:
spring-cloud-dataflow

The spring-cloud-dataflow chart currently supports both internal and external databases, and both internal and external RabbitMQ, but only includes parameters in `values.yaml` for internal Kafka.

The chart should support external Kafka in the same way that it supports external RabbitMQ.

An alternative option would be to configure an external Kafka fully manually using the `extraEnvVars`, but this is more cumbersome and doesn't appear like it would even work based on the templates. Currently in the process of testing this, will update this afterwards.

OmriHarary:
* Provide configuration options for external Kafka and Zookeeper (#5169)

**Description of the change**

Adds values to support configuration of external Kafka in a way similar to the existing support for RabbitMQ

**Benefits**

These changes improve support for external Kafka installations, as well as make the options for using Kafka with the chart more equivalent with those for using RabbitMQ.

**Possible drawbacks**

Due to the fact that there are many potential configuration options for the Kafka binder, and that it can potentially take several brokers makes configuration somewhat different than it is for RabbitMQ.

**Applicable issues**

  - fixes #5169 

**Additional information**

**Checklist** 


OmriHarary:
@juan131 Please take a look when you have a chance. I took a pretty simple approach to keep it as unobtrusive as possible, but let me know if any changes are needed.

juan131 (commented):
Great! Thanks so much for the PR, it looks great! Just a minor suggestion:

Similar to this validation we have in the `_helpers.tpl`:

```
{{/* Validate values of Spring Cloud Dataflow - Messaging System */}}
{{- define "scdf.validateValues.messagingSystem" -}}
{{- if and .Values.kafka.enabled .Values.rabbitmq.enabled -}}
scdf: Messaging System
    You can only use one messaging system.
    Please enable only RabbitMQ or Kafka as messaging system.
{{- end -}}
{{- end -}}
```

We should have a validation to ensure that you're setting `kafka.enabled=true` and `externalKafka=true` simultaneously.

OmriHarary:
@juan131 Just to confirm, do you want validation to pass or fail when both `kafka.enabled=true` and `externalKafka.enabled=true`? In the RabbitMQ case, setting both causes the external to be ignored and only uses the internal, so I figure it should be the same but I just want to make sure I'm not misunderstanding your request.

juan131:
Hi @OmriHarary 

I'd like the validation to block the installation when both settings are set to `true`. The reason: avoid confusion.

Some users will do sth like this:

```console
$ helm install my-release bitnami/spring-cloud-dataflow --set externalKafka=true
```

And they'll be surprised when they realize they're not using the external kafka because they forgot to add `--set kafka.enabled=false`. 

Adding this validation to block the installation, we avoid this kind of misunderstandings.

OmriHarary:
@juan131 Understood, just wanted to make sure. I will be pushing the change shortly.

juan131 (approved):
LGTM

PR #5315, opened 2021-01-28
[bitnami/node] fix invalid default var value in readme

davideperozzi:
**Description of the change**

Just fixes the default value of the `extraEnvVars` variable in the readme

**Benefits**

People don't get confused when they read the README instead of the values.yml

**Possible drawbacks**
N/A

**Additional information**
N/A


carrodher (approved):
Thanks for the fix!

PR #5329, opened 2021-01-28
[bitnami/memcached] Add support for portName

kiuby:
**Description of the change**
Move portName into the values.yaml to let the user override it.

**Benefits**
Compatibility with istio.

**Possible drawbacks**

**Applicable issues**

  - fixes #

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/memcached/Chart.yaml
+++ b/bitnami/memcached/Chart.yaml
@@ -22,4 +22,4 @@ name: memcached
 sources:
   - https://github.com/bitnami/bitnami-docker-memcached
   - http://memcached.org/
-version: 5.5.0
+version: 5.5.1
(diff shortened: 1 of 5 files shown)


migruiz4 (changes requested):
Hi @kiuby,

The PR looks good to me, but could you please use `service.portName` and `metrics.service.portName` instead of  `portName` and `metrics.portName`?

This setting is also included in MongoDB and RabbitMQ charts and it would be nice if they were referenced the same:

https://github.com/bitnami/charts/blob/86aca4dd44191b39737aa7c837a556c95d9bb601/bitnami/rabbitmq/templates/svc.yaml#L29-L31

https://github.com/bitnami/charts/blob/86aca4dd44191b39737aa7c837a556c95d9bb601/bitnami/rabbitmq/values.yaml#L575-L584

kiuby:
Hey @migruiz4,

I didn't put it inside service because it's only concerning the deployment and statefulset object, so i think having it directly at the top level (which represents the main component, so the deployment or the statefulset) makes more sense in this case ! (ie: not the same purpose)

What do you think ? 

And thanks for the fast feedback btw :)

migruiz4 (approved):
Hi @kiuby ,

I'm sorry for the late response and the confusion in my previous review.

I've looked into Istio and I think the PR makes sense as it is.

PR #5333, opened 2021-01-29
[bitnami/Postgresql] Add explicit namespace field on resources

hanzala1234:
**Description of the change**

This PR adds support for explicitly adding a namespace field in resources. when `helm template`  command is run to generate YAML files, it does not add the namespace field which results in resources getting created in the default namespace.

**Benefits**

Add namespace field in manifests generated from `helm template`

**Possible drawbacks**

None known

**Additional information**


rafariossaa (approved):
Thank you very much for adding this.

Issue #5337, opened 2021-01-29
Elasticsearch - Support metrics without a coordinating-only node


Pikabanga:
**Which chart**:
[elasticsearch-14.0.0](https://github.com/bitnami/charts/tree/master/bitnami/elasticsearch)

**Is your feature request related to a problem? Please describe.**

When you deploy a cluster with no coordinating-only nodes (`coordinating.replicas: 0`), the metrics pod fails to start because it's hard coded to query the coordinating-only nodes. It would be nice to be able to have metrics without coordinating-only nodes.

Here is the log output from the metrics pod:

```
level=warn ts=2021-01-29T13:05:45.363072036Z caller=cluster_health.go:270 msg="failed to fetch and decode cluster health" err="failed to get cluster health from http://some-elasticsearch-coordinating-only:9200/_cluster/health: Get http://some-elasticsearch-coordinating-only:9200/_cluster/health: dial tcp 10.233.32.219:9200: connect: connection refused"
```

**Describe the solution you'd like**
An option on the metrics config to point at another node/url. Here's the code in question that would need to become configurable: https://github.com/bitnami/charts/blob/038862f1d35e826784031b77caa88dd3e16f22cf/bitnami/elasticsearch/templates/metrics-deploy.yaml#L37

**Describe alternatives you've considered**
Right now we are deploying a coordinating-only node so we can have metrics, but for the size of the cluster the coordinating-only nodes really just adds extra overhead.

javsalgar:
Hi,

Thank you very much for the feedback. I'll open a task for investigating the issue. I cannot guarantee an ETA as it is not a critical feature, but we will update the ticket as soon as we have more news.

javsalgar:
Hi,

I created this PR to solve the issue https://github.com/bitnami/charts/pull/5364

Pikabanga:
That was fast, thank you very much!

PR #5364, opened 2021-02-01
[bitnami/elasticsearch] Change metrics endpoint when no coordinating nodes are used

javsalgar:
**Description of the change**

In order to slightly improve the performance of the metrics gathering (in the end all the nodes are queries because we set the --es.all flag in elasticserach exporter), we were using the coordinating only nodes as the metrics target for elasticsearch_exporter. It is possible to deploy the chart without coordinating nodes and that was causing issues with the exporter. In this PR we set the master node as fallback in case there are no coordinating nodes available. I considered using a service that points for all the nodes but that caused issues in the past because of incompatibilities with istio (reason why we removed the headless service). 

**Benefits**

Metrics available even though we don't deploy coordinating only nodes.

**Possible drawbacks**

None known

**Applicable issues**

  - fixes https://github.com/bitnami/charts/issues/5337

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/elasticsearch/Chart.yaml
+++ b/bitnami/elasticsearch/Chart.yaml
@@ -25,4 +25,4 @@ name: elasticsearch
 sources:
   - https://github.com/bitnami/bitnami-docker-elasticsearch
   - https://www.elastic.co/products/elasticsearch
-version: 14.1.1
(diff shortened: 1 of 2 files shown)


Mauraza:
LGTM

PR #5370, opened 2021-02-02
[bitnami/nginx] fixes #4394, implements same fix as #3891

belfhi:
Small changes that fix the scope issue of the Nginx Chart,
where the top level variables `.Values` need to be referenced by `$.Values` 
because it's inside  a loop over the `secrets` list/

PR #5373, opened 2021-02-02
[bitnami/several] Fix template issue when using ingress secrets

carrodher:
The conditional logic is being evaluated inside a `range` loop. This means the `.` we're using to access `Values` is not the expected one, as it's overridden for each range iteration evaluation. Instead of that, we should use `$` which references the global scope

Previously fixed in Wordpress, see https://github.com/bitnami/charts/pull/3891 and also reported by another user, see https://github.com/bitnami/charts/pull/5370


bitnami/thanos/templates/bucketweb/tls-secrets.yaml
```diff
@@ -4,7 +4,7 @@ apiVersion: v1
 kind: Secret
 metadata:
   name: {{ include "common.names.fullname" . }}-bucketweb
```
javsalgar, line 6: Also here

bitnami/thanos/templates/query/tls-secrets.yaml
```diff
@@ -5,7 +5,7 @@ apiVersion: v1
 kind: Secret
 metadata:
   name: {{ include "common.names.fullname" . }}-query
```
javsalgar, line 7: Also here

bitnami/thanos/templates/receive/tls-secrets.yaml
```diff
@@ -4,7 +4,7 @@ apiVersion: v1
 kind: Secret
 metadata:
   name: {{ include "common.names.fullname" . }}-receive
```
javsalgar, line 6: Also here

javsalgar (changes requested):

javsalgar (approved):
LGTM

PR #5375, opened 2021-02-02
[bitnami/several] Monthly trademark review

carrodher:
**Description of the change**

Monthly check that the trademark (`(TM)` in yamls and txt, and `<sup>TM</sup>` in markdown) is included in those assets requiring it.

If there are changes only in the README, the chart version is not bumped for that chart. When there are changes in other files different than the README, the chart version was bumped in a patch.


juan131 (approved):
LGTM

PR #5403, opened 2021-02-05
[bitnami/kafka] Separate tls encryption helpers

valdemar-giosg:
**Description of the change**

This change provides two separate helper templates returning `true` if the client connections to Kafka and the inter broker communication connections are encrypted with TLS (i.e. are using tls or mtls).
The new `(include "kafka.client.tlsEncryption" .)` helper template is used in kafka metric exporter deployment to decide if tls command line options should be applied.

**Benefits**

If the kafka release is configured so that both or neither client connections and inter broker connections are using TLS, the metric exporter deployment works as intended.
However, in case when client connections are configured not to use encryption, but broker connections are, the output of `(include "kafka.tlsEncryption" .)` still evaluates to `true`, which results in metric exporter trying to establish connection over TLS, which naturally fails and leads to CrashLoopBackOff in k8s.

With this change, the issue is resolved and the metric exporter uses TLS only if the client connections support it.

**Possible drawbacks**

No possible drawbacks discovered. The chart is made backward compatible and the original `(include "kafka.tlsEncryption" .)` helper template still follows the same logic as before, so will evaluate correctly in all places it is used.

**Checklist** 
- [x] Variables are documented in the README.md **(no new variables included)**

### DIFF
--- a/bitnami/kafka/templates/_helpers.tpl
+++ b/bitnami/kafka/templates/_helpers.tpl
@@ -144,11 +144,30 @@ Return true if authentication via SASL should be configured for inter-broker com
 {{- end -}}
 
 {{/*
+Return true if encryption via TLS for client connections should be configured
+*/}}
+{{- define "kafka.client.tlsEncryption" -}}
+{{- $tlsProtocols := list "tls" "mtls" "sasl_tls" -}}
+{{- if (has .Values.auth.clientProtocol $tlsProtocols) -}}
+    {{- true -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
+Return true if encryption via TLS for inter broker communication connections should be configured
+*/}}
+{{- define "kafka.interBroker.tlsEncryption" -}}
+{{- $tlsProtocols := list "tls" "mtls" "sasl_tls" -}}
+{{- if (has .Values.auth.interBrokerProtocol $tlsProtocols) -}}
+    {{- true -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
 Return true if encryption via TLS should be configured
 */}}
 {{- define "kafka.tlsEncryption" -}}
-{{- $tlsProtocols := list "tls" "mtls" "sasl_tls" -}}
-{{- if or (has .Values.auth.clientProtocol $tlsProtocols) (has .Values.auth.interBrokerProtocol $tlsProtocols) -}}
+{{- if or (include "kafka.client.tlsEncryption" .) (include "kafka.interBroker.tlsEncryption" .) -}}
     {{- true -}}
 {{- end -}}
 {{- end -}}
--- a/bitnami/kafka/templates/kafka-metrics-deployment.yaml
+++ b/bitnami/kafka/templates/kafka-metrics-deployment.yaml
@@ -45,7 +45,7 @@ spec:
               --sasl.username="$SASL_USERNAME" \
               --sasl.password="${sasl_passwords[0]}" \
               {{- end }}
-              {{- if (include "kafka.tlsEncryption" .) }}
+              {{- if (include "kafka.client.tlsEncryption" .) }}
               --tls.enabled \
               {{- if .Values.metrics.kafka.certificatesSecret }}
               --tls.ca-file="/opt/bitnami/kafka-exporter/certs/ca-file" \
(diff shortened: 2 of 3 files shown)


bitnami/kafka/templates/_helpers.tpl
```diff
@@ -143,12 +143,31 @@ Return true if authentication via SASL should be configured for inter-broker com
...
+{{/*
+Return true if encryption via TLS for inter broker communication connections should be configured
+*/}}
+{{- define "kafka.interBroker.tlsEncryption" -}}
+{{- $tlsProtocols := list "tls" "mtls" "sasl_tls" -}}
+{{- if (has .Values.auth.interBrokerProtocol $tlsProtocols) -}}
+    {{- true -}}
+{{- end -}}
+{{- end -}}
+
```
migruiz4, line 165: Instead of repeating code, I'd suggest using a function that received the value as a parameter:
```mustache
{{/*
Return true if variable "protocol" has a valid TLS protocol configured
Usage:
{{ include "kafka.tlsEnabled" ( dict "protocol" .Values.path.to.the.Value ) }}
*/}}
{{- define "kafka.tlsEnabled" -}}
{{- $saslProtocols := list "tls" "mtls" "sasl_tls" -}}
{{- if has .protocol $saslProtocols -}}
    {{- true -}}
{{- end -}}
{{- end -}}
```

bitnami/kafka/templates/kafka-metrics-deployment.yaml
```diff
@@ -45,7 +45,7 @@ spec:
               --sasl.username="$SASL_USERNAME" \
               --sasl.password="${sasl_passwords[0]}" \
               {{- end }}
-              {{- if (include "kafka.tlsEncryption" .) }}
+              {{- if (include "kafka.client.tlsEncryption" .) }}
```
migruiz4, line 48:
```mustache
{{- if (include "kafka.tlsEnabled" ( dict "protocol" .Values.auth.clientProtocol )) -}}
```

bitnami/kafka/templates/kafka-metrics-deployment.yaml
```diff
@@ -73,7 +73,7 @@ spec:
           {{- if .Values.metrics.kafka.resources }}
           resources: {{ toYaml .Values.metrics.kafka.resources | nindent 12 }}
           {{- end }}
-      {{- if and (include "kafka.tlsEncryption" .) .Values.metrics.kafka.certificatesSecret }}
+      {{- if and (include "kafka.client.tlsEncryption" .) .Values.metrics.kafka.certificatesSecret }}
```
migruiz4, line 76:
```mustache
{{- if and (include "kafka.tlsEnabled" ( dict "protocol" .Values.auth.clientProtocol )) .Values.metrics.kafka.certificatesSecret }}
```

bitnami/kafka/templates/_helpers.tpl
```diff
@@ -143,12 +143,31 @@ Return true if authentication via SASL should be configured for inter-broker com
...
 {{/*
 Return true if encryption via TLS should be configured
 */}}
 {{- define "kafka.tlsEncryption" -}}
-{{- $tlsProtocols := list "tls" "mtls" "sasl_tls" -}}
-{{- if or (has .Values.auth.clientProtocol $tlsProtocols) (has .Values.auth.interBrokerProtocol $tlsProtocols) -}}
+{{- if or (include "kafka.client.tlsEncryption" .) (include "kafka.interBroker.tlsEncryption" .) -}}
```
migruiz4, line 170: With the above change, this line would look like this:
```mustache
{{- if or (include "kafka.tlsEnabled" ( dict "protocol" .Values.auth.clientProtocol )) (include "kafka.tlsEnabled" ( dict "protocol" .Values.auth.interBrokerProtocol )) -}}
```

migruiz4 (changes requested):
Hi @valdemar-giosg ,

Your PR looks good to me, but could you please take a look at the change suggested?

valdemar-giosg:
With all the respect, I would decline your suggested change.

That would require including additional parameter to all places in the chart where the legacy "kafka.tlsEncryption" is used. Also, SASL authentication does not depend on TLS neither vice versa, therefore the line

>{{- $saslProtocols := list "tls" "mtls" "sasl_tls" -}}

is invalid.

If the above change is forceful maintainers' decision, please make it. However, I do not favor it as an original author.

migruiz4:
Hi @valdemar-giosg ,

My bad, the `$saslProtocols` was a typo, it was meant to be `$tlsProtocols`.

The suggestion was meant to not repeat code in the functions `"kafka.client.tlsEncryption"` and `"kafka.interBroker.tlsEncryption"` while preserving backward compatibility with `kafka.tlsEncryption` as in the original PR.

But I understand that using parameters does not necessarily simplify things in the template, so I have decided to approve the original PR as it is.

Thank you for your contribution.

migruiz4 (approved):

PR #5426, opened 2021-02-09
Add registered icon to all the MongoDB references

jotamartos:
Modify titles and descriptions and added the disclaimer information in the MongoDB solutions


carrodher (changes requested):
Everything LGTM, but you need to bump the chart version. In the case of bitnami/common, please, take into account chart version and appVersion (both in Chart.yaml) should be set to the same value

carrodher (approved):
Lgtm

PR #5394, opened 2021-02-04
[bitnami/postgresql-ha] Trunc pgpool node hostname instead of showing an error

carrodher:
**Description of the change**

Currently, installing the chart with a "long" name (or long name + long namespace), an error is shown since the leght of the hostname should be less than 128chars:
```
$ helm template aaaaaaaaaaaaabbbbbbbbbbbbcccccccccccddddddddddddeeeee bitnami/postgresql-ha
VALUES VALIDATION:
postgresql-ha: Nodes hostnames
    PostgreSQL nodes hostnames (aaaaaaaaaaaaabbbbbbbbbbbbcccccccccccddddddddddddeeeee-postgresq-00.aaaaaaaaaaaaabbbbbbbbbbbbcccccccccccddddddddddddeeeee-postgresq-headless.default.svc.cluster.local:1234) exceeds the characters limit for Pgpool: 128.
    Consider using a shorter release name or namespace.
```

Something similar happens for the app name but instead of giving an error, we're truncating the name to 63char (it is the DNS limit).

The solution proposed in this PR truncates the hostname in the same way we're truncating the app name.

**Applicable issues**

  - fixes #5387

**Additional information**

Those are the differences of templating the chart using:
```
helm template aaaaaaaaaaaaabbbbbbbbbbbbcccccccccccddddddddddddeeeee .
```
VS
```
helm template artifactory-db bitnami/postgresql-ha
```
```diff
6c6
-  name: artifactory-db-postgresql-ha-pgpool
+  name: aaaaaaaaaaaaabbbbbbbbbbbbcccccccccccddddddddddddeeeee-postgresq
10c10
-    app.kubernetes.io/instance: artifactory-db
+    app.kubernetes.io/instance: aaaaaaaaaaaaabbbbbbbbbbbbcccccccccccddddddddddddeeeee
21c21
-  name: artifactory-db-postgresql-ha-postgresql
+  name: aaaaaaaaaaaaabbbbbbbbbbbbcccccccccccddddddddddddeeeee-postgresq
25c25
-    app.kubernetes.io/instance: artifactory-db
+    app.kubernetes.io/instance: aaaaaaaaaaaaabbbbbbbbbbbbcccccccccccddddddddddddeeeee
37c37
-  name: artifactory-db-postgresql-ha-postgresql-hooks-scripts
+  name: aaaaaaaaaaaaabbbbbbbbbbbbcccccccccccddddddddddddeeeee-postgresq-hooks-scripts
41c41
[...]

PR #5428, opened 2021-02-09
[bitnami/postgresql-ha] Shortened pgpool backend hostnames

### LINKED ISSUE #4923: [bitnami/posgresql-ha] Why include the cluster domain in the cluster node hosts? And what about the namespace?
**Which chart**:
postgresql-ha

I'm having problems with the length limit on the cluster node hosts, as defined [here](https://github.com/bitnami/charts/blob/master/bitnami/postgresql-ha/templates/postgresql/statefulset.yaml#L225) and validated [here](https://github.com/bitnami/charts/blob/master/bitnami/postgresql-ha/templates/_helpers.tpl#L563). I'm frequently confronted with the error message saying that the name is too long.

I wonder why the cluster domain is included in the host names. As far as I know, Kubernetes supports shortened names, e.g. `foo.namespace` instead of `foo.namespace.svc.cluster.local`. Why not use them?

And if the cluster domain can be omitted, would be possible to omit the namespace too? It would only need to be there when the namespaces for postgres and pgpool differ.

If possible, I would suggest to omit the cluster domain and thereby increasing the headroom for the release name and namespace.

And if that's possible, we could possibly omit the namespace too.

I've been shortening release names and I've added an alias (pgha) for this chart in order to make the name less lengthy.

**Additional context**
I'm using this chart as a dependency of an application. The application can be deployed into a review environment, allowing the customer to see the changes we've made. All reviews are deployed to their own namespace, which name included the project name. This leads to hostnames like these: `deployment-pgha-postgresql-00.deployment-pgha-postgresql-headless.project-name-identifier-review-deployment-tdu9h8.svc.cluster.local:1234`.

JeroenVanOort:
**Description of the change**
Shorten the hostnames used by pgpool by omitting the namespace and cluster domain.

**Benefits**
It will be possible to use longer release names.

**Possible drawbacks**
Short hostnames might not be supported by all DNS plugins. However, other charts use short hosts too. For example: e.g. [redis-cluster](https://github.com/bitnami/charts/blob/master/bitnami/redis-cluster/templates/redis-statefulset.yaml#L128).

**Applicable issues**
  - fixes #4923

**Checklist**

### DIFF
--- a/bitnami/postgresql-ha/templates/_helpers.tpl
+++ b/bitnami/postgresql-ha/templates/_helpers.tpl
@@ -580,9 +580,7 @@ Compile all warnings into a single message, and call fail.
 {{- define "postgresql-ha.validateValues.nodesHostnames" -}}
 {{- $postgresqlFullname := include "postgresql-ha.postgresql" . }}
 {{- $postgresqlHeadlessServiceName := printf "%s-headless" (include "postgresql-ha.postgresql" .) }}
-{{- $releaseNamespace := .Release.Namespace }}
-{{- $clusterDomain:= .Values.clusterDomain }}
-{{- $nodeHostname := printf "%s-00.%s.%s.svc.%s:1234" $postgresqlFullname $postgresqlHeadlessServiceName $releaseNamespace $clusterDomain }}
+{{- $nodeHostname := printf "%s-00.%s:1234" $postgresqlFullname $postgresqlHeadlessServiceName }}
 {{- if gt (len $nodeHostname) 128 -}}
 postgresql-ha: Nodes hostnames
     PostgreSQL nodes hostnames ({{ $nodeHostname }}) exceeds the characters limit for Pgpool: 128.
--- a/bitnami/postgresql-ha/templates/_helpers.tpl
+++ b/bitnami/postgresql-ha/templates/_helpers.tpl
@@ -580,7 +580,7 @@ Compile all warnings into a single message, and call fail.
 {{- define "postgresql-ha.validateValues.nodesHostnames" -}}
 {{- $postgresqlFullname := include "postgresql-ha.postgresql" . }}
 {{- $postgresqlHeadlessServiceName := printf "%s-headless" (include "postgresql-ha.postgresql" .) }}
-{{- $nodeHostname := printf "%s-00.%s:1234" $postgresqlFullname $postgresqlHeadlessServiceName }}
+{{- $nodeHostname := printf "%s-00.%s" $postgresqlFullname $postgresqlHeadlessServiceName }}
 {{- if gt (len $nodeHostname) 128 -}}
 postgresql-ha: Nodes hostnames
     PostgreSQL nodes hostnames ({{ $nodeHostname }}) exceeds the characters limit for Pgpool: 128.
--- a/bitnami/postgresql-ha/Chart.yaml
+++ b/bitnami/postgresql-ha/Chart.yaml
@@ -27,4 +27,4 @@ name: postgresql-ha
 sources:
   - https://github.com/bitnami/bitnami-docker-postgresql
   - https://www.postgresql.org/
-version: 7.0.0
+version: 7.1.0
--- a/bitnami/postgresql-ha/templates/pgpool/deployment.yaml
+++ b/bitnami/postgresql-ha/templates/pgpool/deployment.yaml
@@ -96,7 +96,6 @@ spec:
       {{- $postgresqlReplicaCount := int .Values.postgresql.replicaCount }}
       {{- $postgresqlFullname := include "postgresql-ha.postgresql" . }}
       {{- $postgresqlHeadlessServiceName := printf "%s-headless" (include "postgresql-ha.postgresql" .) }}
-      {{- $clusterDomain:= .Values.clusterDomain }}
       containers:
         - name: pgpool
           image: {{ include "postgresql-ha.pgpoolImage" . }}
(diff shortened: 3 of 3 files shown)


TJM:
I was wondering why it was using the long hostname. Is it because it is using the "headless" svc?

JeroenVanOort:
I thought that too, but I tested it and it just as well without the namespace and cluster domain.

Mauraza:
Hi @JeroenVanOort,

thanks for the contribution!, could you check if this PR #5394 solves the same issue?

Mauraza:
@carrodher, could you please take a look at this PR? I saw you have a PR trying to solve the same issue with the name of pgpool?

JeroenVanOort:
@Mauraza Both PR's would help. https://github.com/bitnami/charts/pull/5394 would be the final solution, but this PR would leave more room for the release name and make it easier to reach the goal of that PR.

bitnami/postgresql-ha/templates/_helpers.tpl
```diff
@@ -580,9 +580,7 @@ Compile all warnings into a single message, and call fail.
 {{- define "postgresql-ha.validateValues.nodesHostnames" -}}
 {{- $postgresqlFullname := include "postgresql-ha.postgresql" . }}
 {{- $postgresqlHeadlessServiceName := printf "%s-headless" (include "postgresql-ha.postgresql" .) }}
-{{- $releaseNamespace := .Release.Namespace }}
-{{- $clusterDomain:= .Values.clusterDomain }}
-{{- $nodeHostname := printf "%s-00.%s.%s.svc.%s:1234" $postgresqlFullname $postgresqlHeadlessServiceName $releaseNamespace $clusterDomain }}
+{{- $nodeHostname := printf "%s-00.%s:1234" $postgresqlFullname $postgresqlHeadlessServiceName }}
```
carrodher, line 583: This was a typo in the existing code
```suggestion
{{- $nodeHostname := printf "%s-00.%s:5432" $postgresqlFullname $postgresqlHeadlessServiceName }}
```
It doesn't matter for the validation since the length is going to be the same, but just to fix the typo

carrodher (commented):
Apart from the change in the pgpool deployment, there is a similar env variable (`REPMGR_PARTNER_NODES `) populated with the same approach in https://github.com/bitnami/charts/blob/d99d9be29a43bf8dd27b97bb0a9186dbc92156e6/bitnami/postgresql-ha/templates/postgresql/statefulset.yaml#L228. Do you think the same approach should be applied?

Apart from that, if the backward compatibility is guaranteed it's fine bumping the chart version in a minor but if upgrading from the previous version to this new one is going to cause any kind of issues we should bump the chart version in a major. Have you been able to try an upgrade?

bitnami/postgresql-ha/templates/_helpers.tpl
TJM: Also, for what it's worth, based on my testing, the `:5432` is not part of the hostname length. It is stored in a separate parameter. See the pgpool log output from #5394

JeroenVanOort: I've removed the port from the length check.

JeroenVanOort:
> Apart from the change in the pgpool deployment, there is a similar env variable (`REPMGR_PARTNER_NODES `) populated with the same approach in
> . Do you think the same approach should be applied?

Possibly. My first try was to remove every occurrence of `clusterDomain`, but couldn't get it to work in the time I had. If this is a problem of itself, it might be better to solve it in a separate PR.

> Apart from that, if the backward compatibility is guaranteed it's fine bumping the chart version in a minor but if upgrading from the previous version to this new one is going to cause any kind of issues we should bump the chart version in a major. Have you been able to try an upgrade?

I'm not sure if I can follow you. According to [semver](https://semver.org/) I've made a change in the minor version number. I did this because I think this is a new feature more than it is the solving of a bug.

Mauraza:
Hi @JeroenVanOort,

@carrodher means if you have tried to upgrade a chart running with the changes you have made, I think will be work without any problems. 
Could you do a rebase the master to eliminate the conflicts?

JeroenVanOort:
I just pushed the rebase. I haven't had time to test an upgrade yet. Help is welcome, as I don't think I'll have the time to set up a test any time soon.

Mauraza:
I will test it, I will update the thread when I have more news.

TJM:
Thanks, we found that our "application" fell on its face during a failover operation, so it didn't really matter how long the hostnames were, we removed this HA for a simple single node database (`bitnami/postgresql`), until the application can survive the DB fail-over operation... the restart of the single node database happens faster than the application can restart, so in the end, the "outage" is the same.

Mauraza (changes requested):
Hi @JeroenVanOort, 

I was testing and it seems is working, could you resolve the conflicting file? after that, I will merge the PR.
Thanks for the contribution

JeroenVanOort:
@Mauraza I just did as just asked.

Mauraza:
Hi @JeroenVanOort,

Thanks again for the PR! LGTM!!

Mauraza (approved):

David-Angel:
This change killed us when upgrading to the latest.
pgpool crashing all over due to DNS issues and now we find out the name was shortened on purpose.
And you knew it could cause DNS issues?

Ouch seriously!

vincepower:
This 128 character limit was increased to 254 in pgpool2 a month after the workaround was put in place in this chart.
https://git.postgresql.org/gitweb/?p=pgpool2.git;a=commit;h=2623552275aaf77fea749515ea36b21a8b253ea6

rafariossaa:
Hi,
  Sorry for the delay.
  Thanks for sharing.

vanakamdamapla:
psql: error: connection to server on socket "/opt/bitnami/pgpool/tmp/.s.PGSQL.5432" failed: FATAL: failed to create a backend 0 connection

PR #5430, opened 2021-02-09
bitnami/thanos - Thanos Receive fixes

EladDolev:
**Description of the change**

Thanos Receive fixes
- hashring config now rendered
- `extraFlags` is now being utilized
- missing remote write port was added to the service

**Benefits**

Thanos Receive is now usable

**Additional information**

There are still unimplemented/missing parts for Thanos Receive

**Checklist** 
- [x] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [x] Variables are documented in the README.md
- [x] Title of the PR starts with chart name (e.g. `[bitnami/chart]`)

PR #5438, opened 2021-02-09
[bitnami/thanos] Added remote-write in service and ingress also make srv record discovery possible.

Cellebyte:
**Description of the change**

Added support for the remote-write on ingress.

Fixed a minor bug.

Finally fixed the dnsDiscovery mechanism because the thanos queriers were not able to correctly discover the other components.

```yaml
additionalHeadless: true
```
is now default

**Benefits**

Remote Write can be exposed via an ingress.

**Possible drawbacks**

It's not possible to disable the functionality when the ingress gets enabled.

**Applicable issues**

  - fixes #

**Additional information**

**Checklist** 


alemorcuq (changes requested):
Thank you for yet another contribution, @Cellebyte ! Could you please rebase? The remote write port was already added to the service by another user in #5430, but the Ingress is still missing. The version increase should also be a minor instead of a patch, since you are adding new functionality.

Cellebyte:
@alemorcuq please not merge.
Iam currently trying the following to solve.
When you want to use thanos dns discovery you need a service with "clusterIP: None", but then you are not anymore able to use the ingress.
Should we cover such a use-case because it only affects the `grpc` port of the container.

Cellebyte:
@alemorcuq fixed some bugs coming along the way.
Is the semver version still fine?
Best Regards.

bitnami/thanos/Chart.yaml
```diff
@@ -28,4 +28,4 @@ name: thanos
 sources:
   - https://github.com/bitnami/bitnami-docker-thanos
   - https://thanos.io
-version: 3.8.4
+version: 3.8.5
```
alemorcuq, line 31:
```suggestion
version: 3.9.0
```

bitnami/thanos/templates/query/deployment.yaml
```diff
@@ -80,10 +80,13 @@ spec:
             - --store=dnssrv+_grpc._tcp.{{- include "common.tplvalues.render" ( dict "value" $query.dnsDiscovery.sidecarsService "context" $) -}}.{{- include "common.tplvalues.render"  ( dict "value" $query.dnsDiscovery.sidecarsNamespace "context" $) -}}.svc.{{ .Values.clusterDomain }}
             {{- end }}
             {{- if and .Values.storegateway.enabled $query.dnsDiscovery.enabled }}
-            - --store=dnssrv+_grpc._tcp.{{ include "common.names.fullname" . }}-storegateway.{{ .Release.Namespace }}.svc.{{ .Values.clusterDomain }}
+            - --store=dnssrv+_grpc._tcp.{{ include "common.names.fullname" . }}-storegateway-headless.{{ .Release.Namespace }}.svc.{{ .Values.clusterDomain }}
```
alemorcuq, line 83: The headless service can be disabled. What happens if the headless service is disabled?

Cellebyte: hmm we could add it as a flag.
headless service just makes more sense.

Cellebyte: like that?

romainmeunier:
Can we add a service discovery for the **receive** headless service in this PR ?

Cellebyte:
> Can we add a dns service discovery entry for the **receive** headless service in this PR ?

I just added all as it is before and i could change the default so that it behaves like before and is opt-in

bitnami/thanos/templates/query/deployment.yaml
```diff
@@ -80,10 +80,13 @@ spec:
...
-            - --store=dnssrv+_grpc._tcp.{{ include "common.names.fullname" . }}-storegateway.{{ .Release.Namespace }}.svc.{{ .Values.clusterDomain }}
+            - --store=dnssrv+_grpc._tcp.{{ include "common.names.fullname" . }}-storegateway{{ if .Values.storegateway.service.additionalHeadless }}-headless{{ end }}.{{ .Release.Namespace }}.svc.{{ .Values.clusterDomain }}
             {{- end }}
             {{- if and .Values.ruler.enabled $query.dnsDiscovery.enabled }}
-            - --store=dnssrv+_grpc._tcp.{{ include "common.names.fullname" . }}-ruler.{{ .Release.Namespace }}.svc.{{ .Values.clusterDomain }}
+            - --store=dnssrv+_grpc._tcp.{{ include "common.names.fullname" . }}-ruler{{ if .Values.ruler.service.additionalHeadless }}-headless{{ end }}.{{ .Release.Namespace }}.svc.{{ .Values.clusterDomain }}
+            {{- end }}
+            {{- if and .Values.receive.enabled $query.dnsDiscovery.enabled }}
+            - --store=dnssrv+_grpc._tcp.{{ include "common.names.fullname" . }}-receive{{ if .Values.receive.service.additionalHeadless }}-headless{{ end }}.{{ .Release.Namespace }}.svc.{{ .Values.clusterDomain }}
             {{- end }}
```
carrodher, line 90: This block was implemented (using the previous approach) in a different PR a couple of hours ago, see https://github.com/bitnami/charts/pull/5467.

Can you rebase the changes from the master branch? Thanks

Cellebyte:
messed up.

bitnami/thanos/templates/query/deployment.yaml
Cellebyte: rebased it in  #5475 @carrodher sorry for breaking it here ^^.

PR #5457, opened 2021-02-11
[bitnami/thanos] fix query TLS client in a modular way (#5437)

### LINKED ISSUE #5437: [bitnami/thanos] query do not start if query.grpcTLS.client.secure is set without TLS config
**Which chart**:
bitnami/thanos version >= 3.6.0

**Describe the bug**
The Thanos query refuse to start if no gRPC TLS client config is provided. Before 3.6.0, we could use gRPC TLS client without providing certificates, as described in https://github.com/bitnami/charts/issues/3988.
The PR https://github.com/bitnami/charts/pull/5058 changed this behavior, leading to the following error on query start:

```
level=info ts=2021-02-09T17:15:35.999068898Z caller=client.go:54 msg="enabling client to server TLS"
level=error ts=2021-02-09T17:15:35.999292402Z caller=main.go:130 err="open /tls/client/ca-cert: no such file or directory\nreading client CA\ngithub.com/thanos-io/thanos/pkg/tls.NewClientConfig\n\t/app/pkg/tls/options.go:69\ngithub.com/thanos-io/thanos/pkg/extgrpc.StoreClientGRPCOpts\n\t/app/pkg/extgrpc/client.go:56\nmain.runQuery\n\t/app/cmd/thanos/query.go:262\nmain.registerQuery.func1\n\t/app/cmd/thanos/query.go:161\nmain.main\n\t/app/cmd/thanos/main.go:128\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:204\nruntime.goexit\n\t/usr/local/go/src/runtime/asm_amd64.s:1374\nbuilding gRPC client\nmain.runQuery\n\t/app/cmd/thanos/query.go:264\nmain.registerQuery.func1\n\t/app/cmd/thanos/query.go:161\nmain.main\n\t/app/cmd/thanos/main.go:128\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:204\nruntime.goexit\n\t/usr/local/go/src/runtime/asm_amd64.s:1374\npreparing query command failed\nmain.main\n\t/app/cmd/thanos/main.go:130\nruntime.main\n\t/usr/local/go/src/runtime/proc.go:204\nruntime.goexit\n\t/usr/local/go/src/runtime/asm_amd64.s:1374"
```

**To Reproduce**

1. Install chart 3.5.0 with the following values:
```yaml
query:
  enabled: true
  grpcTLS:
    client:
      secure: true
```
2. Upgrade to version >= 3.6.0

**Expected behavior**
Thanos query should start even if no client TLS config is provided.

Tassatux:
**Description of the change**

Restore TLS client config in a modular way as implemented in PR https://github.com/bitnami/charts/pull/3989 (issue https://github.com/bitnami/charts/issues/3988).

**Benefits**

As explained on #3988 this would enable the usage of multiple ingresses as target stores if they are all under the same CA.

**Possible drawbacks**

It does not break compatibility as the user is still able to specify all CA, Cert and Key for its TLS client. This only makes it so it is modular enough for the user to only pass the information needed to its use-case.

**Applicable issues**

  - fixes #5437

**Checklist*

### DIFF
--- a/bitnami/thanos/templates/query/deployment.yaml
+++ b/bitnami/thanos/templates/query/deployment.yaml
@@ -100,9 +100,15 @@ spec:
             {{- end }}
             {{- if or $query.grpcTLS.client.secure $query.grpcTLS.client.existingSecret }}
             - --grpc-client-tls-secure
+            {{- if or $query.grpcTLS.client.cert $query.grpcTLS.client.existingSecret }}
             - --grpc-client-tls-cert=/tls/client/{{ include "common.secrets.key" (dict "existingSecret" .Values.query.grpcTLS.client.existingSecret "key" "tls-cert") }}
+            {{- end }}
+            {{- if or $query.grpcTLS.client.key $query.grpcTLS.client.existingSecret }}
             - --grpc-client-tls-key=/tls/client/{{ include "common.secrets.key" (dict "existingSecret" .Values.query.grpcTLS.client.existingSecret "key" "tls-key") }}
+            {{- end }}
+            {{- if or $query.grpcTLS.client.ca $query.grpcTLS.client.existingSecret }}
             - --grpc-client-tls-ca=/tls/client/{{ include "common.secrets.key" (dict "existingSecret" .Values.query.grpcTLS.client.existingSecret "key" "ca-cert") }}
+            {{- end }}
             {{- end }}
             {{- if $query.grpcTLS.client.servername }}
             - --grpc-client-server-name={{$query.grpcTLS.client.servername}}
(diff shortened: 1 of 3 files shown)


bitnami/thanos/templates/query/deployment.yaml
```diff
@@ -95,10 +95,16 @@ spec:
             {{- end }}
```
carrodher, line 95: Some lines above (I can't mark them here since they're out of the scope of the PR), there is the same logic but applicable to the server, do you think the modular approach should be applied to the server as well?

carrodher (commented):
Thanks for the fix! Just a minor comment

bitnami/thanos/templates/query/deployment.yaml
Tassatux: Indeed, this make sense for `--grpc-server-tls-client-ca` option, I've pushed a change to apply the same logic.

juan131 (approved):
Thanks so much for the contribution! LGTM

PR #5466, opened 2021-02-11
[bitnami/charts/discourse] Add selector option to pvc

### LINKED ISSUE #5454: [bitnami/discourse] Add support for persistant volume selector to discourse chart pcv.yaml
**Which chart**:
Discourse

Add selector support to the PVC template

In the values.yaml 
```yaml
persistence:
  enabled: true
  ## discourse & sidekiq data Persistent Volume Storage Class
  ## If defined, storageClassName: <storageClass>
  ## If set to "-", storageClassName: "", which disables dynamic provisioning
  ## If undefined (the default) or set to null, no storageClassName spec is
  ##   set, choosing the default provisioner.  (gp2 on AWS, standard on
  ##   GKE, AWS & OpenStack)
  ##
  # storageClass: "-"
  ##
  ## If you want to reuse an existing claim, you can pass the name of the PVC using
  ## the existingClaim variable
  # existingClaim: your-claim
  accessMode: ReadWriteMany
  size: 10Gi
  selector:
    matchLabels:
       <label>:<value>
```
This would generate a selector clause for the PVC.   Similar to what the Postgresql and Redis template do.

ArcticSnowman:
**Description of the change**

This change adds a selector to the PCV so that user can use label matching to select the PV
Also updated the accessMode to ReadWriteMany as it being used by discourse and sidekiq  pods.

**Benefits**

User can target the right PV in a kube that has many PVs

**Possible drawbacks**

If the user does not have a storage provider that supports ReadWriteMany, they will need to create two PVC, one for discourse and one for sidekiq

**Applicable issues**

  - fixes #5454

**Additional information**

**Checklist** 
- [ X ] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [ X ] Variables are documented in the README.md
- [ X ] Title of the PR starts with chart name (e.g. `[bitnami/chart]`)

### DIFF
--- a/bitnami/discourse/Chart.yaml
+++ b/bitnami/discourse/Chart.yaml
@@ -35,4 +35,4 @@ sources:
   - https://github.com/bitnami/bitnami-docker-discourse
   - https://github.com/spinnaker
   - https://www.discourse.org/
-version: 2.2.1
+version: 2.2.2
--- a/bitnami/discourse/Chart.yaml
+++ b/bitnami/discourse/Chart.yaml
@@ -35,4 +35,4 @@ sources:
   - https://github.com/bitnami/bitnami-docker-discourse
   - https://github.com/spinnaker
   - https://www.discourse.org/
-version: 2.2.2
+version: 2.3.0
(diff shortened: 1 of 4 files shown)


carrodher (approved):
Thanks for your contribution, changes LGTM. I just added a new commit bumping the chart version in a minor (instead of a patch) since it is a feature addition.
As there is an issue linked, it is going to be automatically closed once merged.

PR #5481, opened 2021-02-12
[bitnami/keycloak] Allow to use the same JKS keystore

yesteph:
**Description of the change**

If TLS termination is enables at the POD level, allow to use the same JKS keystore file for all the replicas.
Old mode with one shared truststore plus one keystore for each replica is kept and default.

**Benefits**

Simplify the creation of the Secret storing the JKS files.

Instead of a Secret with a key `keycloak.truststore.jks` plus a set of `keycloak-X.keystore.jks` where X is the ID of each Keycloak replica, you can now provide a Secret with two files: `truststore.jks`and `keystore.jks`.

**Possible drawbacks**

It may be a security concern to get a proper keystore per Keycloak replica if the keycloak pods are directly exposed to the end user traffic.
But this is generally not the case as we generally use an ingress or a level-7 load balancer.

**Applicable issues**

N/A

**Additional information**

JKS files key in the Secret (jksSecret) are renamed to match the keys generated by cert-manager when `keystores` are indicated in the Certificate resource.

**Checklist** 


bitnami/keycloak/templates/statefulset.yaml
```diff
@@ -67,7 +67,7 @@ spec:
         {{- if .Values.initContainers }}
         {{- include "common.tplvalues.render" (dict "value" .Values.initContainers "context" $) | nindent 8 }}
         {{- end }}
-        {{- if .Values.auth.tls.enabled }}
+        {{- if and .Values.auth.tls.enabled (not .Values.auth.tls.sharedJKS) }}
```
javsalgar, line 70: Question, don't we want to change the permissions for the shared JKS certificates as well? Wouldn't this cause issues?

javsalgar (commented):
Thank you so much for the PR! One question

bitnami/keycloak/templates/statefulset.yaml
yesteph: Hello @javsalgar 

The files are just read. No special permissions required for that.

bitnami/keycloak/templates/configmap-env-vars.yaml
```diff
@@ -32,6 +32,6 @@ data:
   KEYCLOAK_ENABLE_TLS: {{ ternary "true" "false" .Values.auth.tls.enabled | quote }}
   {{- if .Values.auth.tls.enabled }}
   KEYCLOAK_HTTPS_PORT: {{ .Values.containerPorts.https | quote }}
-  KEYCLOAK_TLS_KEYSTORE_FILE: "/opt/bitnami/keycloak/certs/keycloak.keystore.jks"
-  KEYCLOAK_TLS_TRUSTSTORE_FILE: "/opt/bitnami/keycloak/certs/keycloak.truststore.jks"
+  KEYCLOAK_TLS_KEYSTORE_FILE: {{ ternary "/certs/keystore.jks" "/opt/bitnami/keycloak/certs/keycloak.keystore.jks" .Values.auth.tls.sharedJKS | quote }}
```
javsalgar, line 35: I have my concerns about the paths. It seems to me that, for consistency, the other mount folder should also be `/opt/bitnami/keycloak`. Maybe `/opt/bitnami/keycloak/shared-certs`?

bitnami/keycloak/templates/statefulset.yaml
javsalgar: Ok

PR #5504, opened 2021-02-15
[bitnami/thanos] Fixing local-endpoint to work also with scaling.

Cellebyte:
Fixing the local-endpoint

**Description of the change**

Change the flag to make the container able to discover itself.

**Benefits**

This change makes it possible to scale better in the statefulset.
It autogenerates the hashring config when no config is specified.

```yaml
# config can be specified as yaml

config:
  - endpoints:
    - "localhost:{{ .Values.receive.service.grpc.port }}"
# or as string
config: |
    [
      {
        "endpoints": [
            "127.0.0.1:{{ .Values.receive.service.grpc.port }}"
        ]
      }
    ]
```
**Possible drawbacks**

**Additional information**

> WARNING: 
> The old behavior is only triggered when you run it with `receive.replicaCount: 1` and set `receive.config:` with a value.
> If `""` or `{}` is set as `receive.config` the new behavior is triggered.
> When the flag `receive.service.additionalHeadless: true` is set the template `{{ include "common.tplvalues.render" (dict "value" (include "thanos.receive.config" .) "context" .) | nindent 4 }}` templates against the headless service and not against the default service.

**Checklist** 


Mauraza:
Hi @Cellebyte, 

thanks for the contribution! could you review the checklist and the contributing guide https://github.com/bitnami/charts/blob/master/CONTRIBUTING.md

Cellebyte:
@Mauraza I think I followed the guideline.

bitnami/thanos/Chart.yaml
```diff
@@ -28,4 +28,4 @@ name: thanos
 sources:
   - https://github.com/bitnami/bitnami-docker-thanos
   - https://thanos.io
-version: 3.10.1
+version: 3.10.4
```
juan131, line 31: I think we should bump the minor version. You're including a new parameter, and modifying the expected format of another one.

bitnami/thanos/values.yaml
```diff
@@ -1785,14 +1788,9 @@ receive:
...
-    [
-      {
-        "endpoints": [
-            "127.0.0.1:10901"
-        ]
-      }
-    ]
+  config:
+    - endpoints:
+      - "127.0.0.1:{{ .Values.receive.service.grpc.port }}"
```
juan131, line 1793: I'm a little concerned about this change.
Let's say I have my **values.yaml** already adapted specifying my custom configuration for Thanos Receiver, if I upgrade my release I can find an unexpected behaviour

juan131: Oh, I just saw you covered this possibility in the `thanos.receive.config` macro

bitnami/thanos/templates/receive/statefulset.yaml
```diff
@@ -71,13 +71,17 @@ spec:
...
-            - --receive.replication-factor=1
+            - --receive.replication-factor={{ .Values.receive.replicationFactor }}
             - --objstore.config=$(OBJSTORE_CONFIG)
             - --tsdb.path=/var/thanos/receive
             - --label={{ .Values.receive.replicaLabel }}="$(NAME)"
             - --label=receive="true"
             - --tsdb.retention=15d
-            - --receive.local-endpoint=127.0.0.1:10901
+            {{- if and (eq .Values.receive.replicaCount 1) .Values.receive.config }}
+            - --receive.local-endpoint=127.0.0.1:{{ .Values.receive.service.grpc.port }}
```
juan131, line 81: Why are you using the service port? It should use the container port (regardless what port you open at service level)

bitnami/thanos/templates/receive/statefulset.yaml
```diff
@@ -71,13 +71,17 @@ spec:
...
+            {{- else }}
+            - --receive.local-endpoint=$(NAME).thanos-receive{{ if .Values.receive.service.additionalHeadless }}-headless{{ end }}.$(NAMESPACE).svc.{{ .Values.clusterDomain }}:10901
```
juan131, line 83: In this case, I do see the value of using the service port (instead of hardcoding it)

juan131 (changes requested):
Thanks for the contribution @Cellebyte ! Please take a look to my comments

bitnami/thanos/templates/receive/statefulset.yaml
Cellebyte: So just hardcode I will hardcode it here.

bitnami/thanos/templates/receive/statefulset.yaml
Cellebyte: I will hardcode it here.

bitnami/thanos/values.yaml
Cellebyte: Yep I did. 👍

bitnami/thanos/templates/receive/statefulset.yaml
Cellebyte: I think both should be hardcoded.
Because its the fqdn for the pod itself. It will allways listen on that port unless we make it configurable.

Cellebyte: as this port is not configurable we should not change it.

bitnami/thanos/Chart.yaml
Cellebyte: Done

juan131 (approved):
Great @Cellebyte !! Thanks so much again for the PR
LGTM

bitnami/thanos/templates/receive/statefulset.yaml
```diff
@@ -71,13 +71,17 @@ spec:
...
             - --http-address=0.0.0.0:10902
             - --remote-write.address=0.0.0.0:19291
-            - --receive.replication-factor=1
+            - --receive.replication-factor={{ .Values.receive.replicationFactor }}
             - --objstore.config=$(OBJSTORE_CONFIG)
             - --tsdb.path=/var/thanos/receive
             - --label={{ .Values.receive.replicaLabel }}="$(NAME)"
             - --label=receive="true"
             - --tsdb.retention=15d
+            {{- if and (eq .Values.receive.replicaCount 1) .Values.receive.config }}
```
guilhermef, line 80: @Cellebyte There's an error when running this part:
```
Error: template: thanos/templates/receive/statefulset.yaml:80:24: executing "thanos/templates/receive/statefulset.yaml" at <eq .Values.receive.replicaCount 1>: error calling eq: incompatible types for comparison
helm.go:81: [debug] template: thanos/templates/receive/statefulset.yaml:80:24: executing "thanos/templates/receive/statefulset.yaml" at <eq .Values.receive.replicaCount 1>: error calling eq: incompatible types for comparison
```

Cellebyte: @guilhermef -> #5549

PR #5506, opened 2021-02-15
Fix permissions image in thanos

dbason:
**Description of the change**
Simple bug fix to the volumePermissions helper in the thanos chart


javsalgar (approved):
Thank you so much for the PR! LGTM

Issue #5578, opened 2021-02-22
[bitnami/kafka] Default Kafka-Config values override in values.yaml


sveba:
Why do you need to change the default config values for the kafka brokers in the values.yaml? I mean: https://github.com/bitnami/charts/blob/91dbd11bcdd9284029e5857235b0b7b0ddade1b5/bitnami/kafka/values.yaml#L122-L207

Logs directory is the only value, where I see, that a change is needed. The other props are not required by the docker image and are in some cases impossible to override. Like this one, I would like to change it to default value of `9223372036854775807`. This is not possible, because the value is being parsed to `int`: 
https://github.com/bitnami/charts/blob/91dbd11bcdd9284029e5857235b0b7b0ddade1b5/bitnami/kafka/values.yaml#L131

juan131:
Comments added on your PR @sveba

sveba:
@juan131 
Thank you for your comment.

I would like to discuss the general problem with predefined values.  

1. What is the intension of setting those? 
2. Why not go with default Apache-Kafka values and let the user decide what to change, if needed. The Bitnami-Kafka-Image can be run without setting any ENVVAR

juan131:
Honestly I don't know it @sveba 

I just saw these default values have been there since the very beginning, see:

- https://github.com/bitnami/charts/pull/681/files#diff-eb48889a737011f08ce276531652461851fa335e5988268092b6698e44870cce

@tompizmor do you remember where this comes from or what's the reason to use them?

tompizmor:
I may know the reason...

The first version of the chart (and the kafka container) didn't support configuring all Kafka properties through env vars (that feature was added in #1202) so we only exposed the main properties with the Kafka default values at that time. 

Right now, as we can set any kafka property with `KAFKA_CFG_XXX` env vars, it wouldn't be necessary to maintain those properties configuration in the values.yaml.

Taking the `logFlushIntervalMessages` as an example, currently is set in the _values.yaml_ as:

```
## The number of messages to accept before forcing a flush of data to disk.
##
logFlushIntervalMessages: _10000
```


and it is used from _statefulset.yaml_ as:

```
            - name: KAFKA_CFG_LOG_FLUSH_INTERVAL_MESSAGES
              value: {{ .Values.logFlushIntervalMessages | replace "_" "" | quote }
```


The same could be accomplish if we delete the previous blocks and simply define this extra environment variable in the values.yaml:

```
[...]

PR #5579, opened 2021-02-22
[bitnami/kafka] Fix `long` value for `log.flush.scheduler.interval.ms`

sveba:
**Description of the change**

Fix for `long` value for `log.flush.scheduler.interval.ms`. See #5578

**Checklist**

### DIFF
--- a/bitnami/kafka/Chart.yaml
+++ b/bitnami/kafka/Chart.yaml
@@ -29,4 +29,4 @@ name: kafka
 sources:
   - https://github.com/bitnami/bitnami-docker-kafka
   - https://kafka.apache.org/
-version: 12.9.3
+version: 12.9.4
--- a/bitnami/kafka/README.md
+++ b/bitnami/kafka/README.md
@@ -85,7 +85,7 @@ The following tables lists the configurable parameters of the Kafka chart and th
 | `heapOpts`                                | Kafka's Java Heap size                                                                                                                               | `-Xmx1024m -Xms1024m`                                   |
 | `deleteTopicEnable`                       | Switch to enable topic deletion or not                                                                                                               | `false`                                                 |
 | `autoCreateTopicsEnable`                  | Switch to enable auto creation of topics. Enabling auto creation of topics not recommended for production or similar environments                    | `false`                                                 |
-| `logFlushIntervalMessages`                | The number of messages to accept before forcing a flush of data to disk                                                                              | `10000`                                                 |
+| `logFlushIntervalMessages`                | The number of messages to accept before forcing a flush of data to disk                                                                              | `_10000`                                                 |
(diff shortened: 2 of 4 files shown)


migruiz4 (approved):
Hi @sveba,

Your PR looks good to me, thank you for your contribution.

migruiz4 (changes requested):
Sorry, I noticed the PR is missing a change in the README.md updating the default value for `logFlushIntervalMessages`.

Could you please add the change before merging this PR?

juan131:
@sveba I'm unable to understand what's the problem here.

If we compare the results of using the current approach:

```console
$ helm template kafka bitnami/kafka --set logFlushIntervalMessages=9223372036854775807 -s templates/statefulset.yaml
...
            - name: KAFKA_CFG_LOG_FLUSH_INTERVAL_MESSAGES
              value: "9223372036854775807"
```

With your changes and your approach:

```console
$ helm template kafka bitnami/kafka --set logFlushIntervalMessages=9223372036854775807 -s templates/statefulset.yaml
...
            - name: KAFKA_CFG_LOG_FLUSH_INTERVAL_MESSAGES
              value: "9223372036854775807"
```

I don't see any difference.

sveba:
@juan131 
that is strange. You are right, if you set the value over cli, than helm renders the template right. But if you put it in a values file the output is
`    value: "9.223372036854776e+18"
`

my helm version is 3.4.2

migruiz4:
Hi @juan131 ,

This issue only happens when modifying the `values.yaml`.
```
> cat bitnami/kafka/values.yaml | grep logFlushIntervalMs
logFlushIntervalMs: 9223372036854775807

> helm template bitnami/kafka | grep -A1 KAFKA_CFG_LOG_FLUSH_INTERVAL_MS
            - name: KAFKA_CFG_LOG_FLUSH_INTERVAL_MS
              value: "9.223372036854776e+18"
```

It can be solved by defining the variable as a string using quotes (`"`)
```
> cat bitnami/kafka/values.yaml | grep logFlushIntervalMs
logFlushIntervalMs: "9223372036854775807"

> helm template bitnami/kafka | grep -A1 KAFKA_CFG_LOG_FLUSH_INTERVAL_MS
            - name: KAFKA_CFG_LOG_FLUSH_INTERVAL_MS
              value: "9223372036854775807"
```

But using the `_` we prevent errors by forcing the variable to be a string. 

I think there should be a better alternative for this, but currently, other values are being handled the same way so that is why I'm approving this PR.

migruiz4 (approved):
Thank you for updating this PR with the suggested changes.

juan131:
I guess this some kind of YAML limitation. I wasn't aware of this, so I guess I learned something new today 😃 
Thanks so much @migruiz4 @sveba !!

PR #5602, opened 2021-02-24
bitnami/dataplatform-bp1 addition of observability framework

gkulkarni32:
**Description of the change**

Addition of wavefront observability framework for the dataplatform
Addition of the values.schema.json for the form layout in kubeapps

**Benefits**

we will be able to provide observability framework to the users
we will be able to provide a form layout in kubeapps portal 

**Possible drawbacks**

**Applicable issues**

  - fixes #

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/dataplatform-bp1/Chart.yaml
+++ b/bitnami/dataplatform-bp1/Chart.yaml
@@ -1,7 +1,7 @@
 annotations:
   category: Infrastructure
 apiVersion: v2
-appVersion: 2.0.0
+appVersion: 1.0.0
 dependencies:
   - condition: kafka.enabled
     name: kafka
@@ -58,4 +58,4 @@ sources:
   - https://github.com/bitnami/bitnami-docker-wavefront-proxy
   - https://github.com/wavefrontHQ/wavefront-collector-for-kubernetes
   - https://github.com/wavefrontHQ/wavefront-proxy
-version: 1.0.0
+version: 0.2.0
--- a/bitnami/dataplatform-bp1/Chart.yaml
+++ b/bitnami/dataplatform-bp1/Chart.yaml
@@ -39,7 +39,7 @@ keywords:
   - solr
   - zookeeper
   - wavefront
-  - obesrvability
+  - observability
   - apache
 maintainers:
   - email: containers@bitnami.com
--- a/bitnami/dataplatform-bp1/Chart.yaml
+++ b/bitnami/dataplatform-bp1/Chart.yaml
@@ -22,6 +22,7 @@ dependencies:
   - condition: wavefront.enabled
     name: wavefront
     repository: https://charts.bitnami.com/bitnami
+    version: 1.x.x
   - name: common
     repository: https://charts.bitnami.com/bitnami
     tags:
--- a/bitnami/dataplatform-bp1/Chart.yaml
+++ b/bitnami/dataplatform-bp1/Chart.yaml
@@ -31,6 +31,7 @@ dependencies:
 description: OCTO Data platform Kafka-Spark-Solr Helm Chart
 engine: gotpl
 home: https://github.com/bitnami/charts/tree/master/bitnami/dataplatform-bp1
+icon: https://bitnami.com/assets/stacks/dataplatform-bp1/img/dataplatform-bp1-stack-220x234.png
 keywords:
   - dataplatform
   - kafka
--- a/bitnami/dataplatform-bp1/Chart.yaml
+++ b/bitnami/dataplatform-bp1/Chart.yaml
@@ -1,7 +1,7 @@
 annotations:
   category: Infrastructure
 apiVersion: v2
-appVersion: 1.0.0
+appVersion: 2.0.0
 dependencies:
   - condition: kafka.enabled
     name: kafka
@@ -19,6 +19,9 @@ dependencies:
     name: zookeeper
     repository: https://charts.bitnami.com/bitnami
     version: 6.x.x
+  - condition: wavefront.enabled
+    name: wavefront
+    repository: https://charts.bitnami.com/bitnami
   - name: common
     repository: https://charts.bitnami.com/bitnami
     tags:
@@ -26,13 +29,15 @@ dependencies:
     version: 1.x.x
 description: OCTO Data platform Kafka-Spark-Solr Helm Chart
 engine: gotpl
-home: https://github.com/bitnami/charts/tree/master/bitnami/dataplatform
+home: https://github.com/bitnami/charts/tree/master/bitnami/dataplatform-bp1
 keywords:
   - dataplatform
   - kafka
   - spark
   - solr
   - zookeeper
+  - wavefront
+  - obesrvability
   - apache
 maintainers:
   - email: containers@bitnami.com
@@ -47,4 +52,8 @@ sources:
   - https://github.com/bitnami/bitnami-docker-solr
   - https://zookeeper.apache.org/
   - https://github.com/bitnami/bitnami-docker-zookeeper
+  - https://github.com/bitnami/bitnami-docker-wavefront-kubernetes-collector
+  - https://github.com/bitnami/bitnami-docker-wavefront-proxy
+  - https://github.com/wavefrontHQ/wavefront-collector-for-kubernetes
+  - https://github.com/wavefrontHQ/wavefront-proxy
 version: 0.1.0
--- a/bitnami/dataplatform-bp1/Chart.yaml
+++ b/bitnami/dataplatform-bp1/Chart.yaml
@@ -56,4 +56,4 @@ sources:
   - https://github.com/bitnami/bitnami-docker-wavefront-proxy
   - https://github.com/wavefrontHQ/wavefront-collector-for-kubernetes
   - https://github.com/wavefrontHQ/wavefront-proxy
-version: 0.1.0
+version: 1.0.0
--- a/bitnami/dataplatform-bp1/README.md
+++ b/bitnami/dataplatform-bp1/README.md
@@ -169,6 +169,23 @@ Parameters below are set as per the recommended values, they can be overwritten 
 | `spark.metrics.workerAnnotations`       | Annotations for enabling prometheus to access the metrics endpoint of the worker nodes   | `{prometheus.io/scrape: "true", prometheus.io/path: "/metrics/", prometheus.io/port: "8081"}` |
 | `spark.metrics.resources.requests` | The requested resources for the metrics exporter container  | Spark exporter container resource requests for optimal resource usage size|
 
+### Wavefront chart parameters
+
+| Parameter                                  | Description                                                                                                            | Default                                                 |
+|--------------------------------------------|------------------------------------------------------------------------------------------------------------------------|---------------------------------------------------------|
+| `wavefront.enabled`                        | Switch to enable or disable the Wavefront helm chart                                                                   | `true`                                                  |
+| `wavefront.clusterName`                    | Unique name for the Kubernetes cluster (required)                                                                      | `KUBERNETES_CLUSTER_NAME`                               |
+| `wavefront.wavefront.url`                  | Wavefront URL for your cluster (required)                                                                              | `https://YOUR_CLUSTER.wavefront.com`                    |
+| `wavefront.wavefront.token`                | Wavefront API Token (required)                                                                                         | `YOUR_API_TOKEN`                                        |
+| `wavefront.wavefront.existingSecret`       | Name of an existing secret containing the token                                                                        | `nil`                                                   |
+| `wavefront.collector.discovery.enabled`    | Rules based and Prometheus endpoints auto-discovery                                                                    | `true`                                                  |
+| `wavefront.collector.discovery.enableRuntimeConfigs` | Enable runtime discovery rules                                                                               | `true`                                                 |
+| `wavefront.collector.discovery.config`     | Configuration for rules based auto-discovery                                                                           | Data Platform components pods discovery config                                                   |
+| `wavefront.collector.resources.limits`     | The resources limits for the collector container                                                                       | `{}`                                                    |
+| `wavefront.collector.resources.requests`   | The requested resources for the collector container                                                                    | Collectors pods resource requests for optimal resource usage size                       |
+| `wavefront.proxy.resources.limits`         | The resources limits for the proxy container                                                                           | `{}`                                                    |
+| `wavefront.proxy.resources.requests`       | The requested resources for the proxy container                                                                        | Proxy pods resource resource requests for optimal resource usage size                                |
+
 Specify each parameter using the `--set key=value[,key=value]` argument to `helm install`. For example,
 
 ```console
--- a/bitnami/dataplatform-bp1/README.md
+++ b/bitnami/dataplatform-bp1/README.md
@@ -196,15 +196,19 @@ $ helm install my-release \
 
 The above command deploys the data platform with Kafka with 3 nodes (replicas).
 
-In case you need to deploy the data platform skipping any component, you can specify the 'enabled' parameter using the `--set <component>.enabled=false` argument to `helm install`. For Example,
+In case you need to deploy the data platform with Wavefront Observability Framework for all the applications (Kafka/Spark/Solr) in the data platform, you can specify the 'enabled' parameter using the `--set <component>.metrics.enabled=true` argument to `helm install`. For Solr, the parameter is `solr.exporter.enabled=true` For Example,
 
 ```console
-$ helm install my-release \
-  --set solr.enabled=false \
-  bitnami/dataplatform-bp1
+$ helm install my-release bitnami/dataplatform-bp1 \
+    --set kafka.metrics.kafka.enabled=true \
+    --set kafka.metrics.jmx.enabled=true \    
+    --set spark.metrics.enabled=true \
+    --set solr.exporter.enabled=true \
+    --set wavefront.enabled=true \
+    --set wavefront.clusterName=<K8s-CLUSTER-NAME> \
+    --set wavefront.wavefront.url=https://<YOUR_CLUSTER>.wavefront.com \
+    --set wavefront.wavefront.token=<YOUR_API_TOKEN>
 ```
-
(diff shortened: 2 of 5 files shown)


bitnami/dataplatform-bp1/values.yaml
```diff
@@ -71,6 +72,46 @@ kafka:
                   values:
                     - zookeeper
             topologyKey: "kubernetes.io/hostname"
+  ## Prometheus Exporters / Metrics
+  ##
+  metrics:
+    ## Prometheus Kafka Exporter: exposes complimentary metrics to JMX Exporter
+    ##
+    kafka:
+      enabled: '{{ .Values.wavefront.enabled }}'
```
marcosbc, line 81: Not sure if this is the best / more consistent way for enabling metrics, I'm not a big fan of it. @javsalgar what do you think?

bitnami/dataplatform-bp1/values.yaml
```diff
@@ -161,6 +228,30 @@ solr:
...
+        preferredDuringSchedulingIgnoredDuringExecution:
+         - weight: 50
+           podAffinityTerm:
+             labelSelector:
+               matchExpressions:
+               - key: app.kubernetes.io/name
+                 operator: In
+                 values:
+                   - solr
+             topologyKey: "kubernetes.io/hostname"
```
marcosbc, line 249: Why are we adding these options? Not sure if this is something that should be enabled by default. @javsalgar what do you think?

bitnami/dataplatform-bp1/values.yaml
```diff
@@ -170,3 +261,110 @@ solr:
...
+    ##
+    existingSecret:
+  ## Wavefront Collector is responsible to get all Kubernetes metrics from your cluster.
+  ## It will capture Kubernetes resources metrics available from the kubelets,
+  ## as well as auto-discovery capabilities.
+  ##
+  collector:
+  ## Rules based discovery configuration
+  ## Ref: https://github.com/wavefrontHQ/wavefront-kubernetes-collector/blob/master/docs/discovery.md
+  ##
```
marcosbc, line 288: Could you fix the indentation for this comment?

```suggestion
    ## Rules based discovery configuration
    ## Ref: https://github.com/wavefrontHQ/wavefront-kubernetes-collector/blob/master/docs/discovery.md
    ##
```

marcosbc (commented):
Thanks for the contribution! It mostly looks good. However, I'm requesting @javsalgar to check a couple of things and also to have another check of this PR in case I missed anything.

bitnami/dataplatform-bp1/values.yaml
javsalgar: Actually, is this template rendered in the proper subchart?

bitnami/dataplatform-bp1/Chart.yaml
```diff
@@ -19,20 +19,25 @@ dependencies:
     name: zookeeper
     repository: https://charts.bitnami.com/bitnami
     version: 6.x.x
+  - condition: wavefront.enabled
+    name: wavefront
+    repository: https://charts.bitnami.com/bitnami
```
carrodher, line 24: It is needed to specify here the major branch for the subcharts
```suggestion
  - condition: wavefront.enabled
    name: wavefront
    repository: https://charts.bitnami.com/bitnami
    version: 1.x.x
```

bitnami/dataplatform-bp1/Chart.yaml
```diff
@@ -19,20 +19,25 @@ dependencies:
...
   - name: common
     repository: https://charts.bitnami.com/bitnami
     tags:
       - bitnami-common
     version: 1.x.x
 description: OCTO Data platform Kafka-Spark-Solr Helm Chart
 engine: gotpl
-home: https://github.com/bitnami/charts/tree/master/bitnami/dataplatform
+home: https://github.com/bitnami/charts/tree/master/bitnami/dataplatform-bp1
 keywords:
```
carrodher, line 33: It is also needed to add the icon to be rendered in the different hubs and/or marketplaces:
```suggestion
home: https://github.com/bitnami/charts/tree/master/bitnami/dataplatform-bp1
icon: https://bitnami.com/assets/stacks/dataplatform-bp1/img/dataplatform-bp1-stack-220x234.png
keywords:
```

bitnami/dataplatform-bp1/Chart.yaml
```diff
@@ -47,4 +52,8 @@ sources:
   - https://github.com/bitnami/bitnami-docker-solr
   - https://zookeeper.apache.org/
   - https://github.com/bitnami/bitnami-docker-zookeeper
-version: 0.1.0
+  - https://github.com/bitnami/bitnami-docker-wavefront-kubernetes-collector
+  - https://github.com/bitnami/bitnami-docker-wavefront-proxy
+  - https://github.com/wavefrontHQ/wavefront-collector-for-kubernetes
+  - https://github.com/wavefrontHQ/wavefront-proxy
+version: 1.0.0
```
carrodher, line 59: Is backward compatibility guaranteed? If the changes in this PR are not going to break the upgrades from the previous version, we only need to bump the chart version in a minor instead of a major
```suggestion
version: 0.2.0
```

bitnami/dataplatform-bp1/values.yaml
gkulkarni32: @marcosbc This is basically a solution chart to deploy kafka/spark/solr together. The intention is to provide the pod affinity rules + resource optimization to provide recommended kubernetes node count.

bitnami/dataplatform-bp1/Chart.yaml
```diff
@@ -19,20 +19,25 @@ dependencies:
...
-home: https://github.com/bitnami/charts/tree/master/bitnami/dataplatform
+home: https://github.com/bitnami/charts/tree/master/bitnami/dataplatform-bp1
 keywords:
   - dataplatform
   - kafka
   - spark
   - solr
   - zookeeper
+  - wavefront
+  - obesrvability
```
miguelaeh, line 40: Typo here

bitnami/dataplatform-bp1/Chart.yaml
```diff
@@ -19,20 +19,25 @@ dependencies:
...
+    repository: https://charts.bitnami.com/bitnami
   - name: common
     repository: https://charts.bitnami.com/bitnami
     tags:
       - bitnami-common
     version: 1.x.x
 description: OCTO Data platform Kafka-Spark-Solr Helm Chart
 engine: gotpl
-home: https://github.com/bitnami/charts/tree/master/bitnami/dataplatform
+home: https://github.com/bitnami/charts/tree/master/bitnami/dataplatform-bp1
```
miguelaeh, line 32: I think we can already add the icon link pointing to the bitnami one.

bitnami/dataplatform-bp1/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 annotations:
   category: Infrastructure
 apiVersion: v2
-appVersion: 1.0.0
+appVersion: 2.0.0
```
carrodher, line 4: As there is no standalone application behind this chart, let's use the same app version as the one used as the chart version (see below comment)
```suggestion
appVersion: 0.2.0
```

bitnami/dataplatform-bp1/values.schema.json
```diff
@@ -0,0 +1,81 @@
+{
```
miguelaeh, line 1: I think this should not be necessary. Are you adding it because of certain reason?

bitnami/dataplatform-bp1/README.md
```diff
@@ -143,12 +156,18 @@ Parameters below are set as per the recommended values, they can be overwritten
...
 | `spark.worker.affinity`     | Spark worker affinity for pod assignment         | Spark worker pods Affinity rules set for best possible resiliency (evaluated as a template) |
 | `spark.worker.resources`    | CPU/Memory resource requests/limits for worker   | Spark worker pods resource requests set for optimal resource usage                          |
+| `spark.metrics.enabled`     | Start a side-car prometheus exporter                                                                                           | `false`       |
+| `spark.metrics.masterAnnotations`       | Annotations for enabling prometheus to access the metrics endpoint of the master nodes  | `{prometheus.io/scrape: "true", prometheus.io/path: "/metrics/", prometheus.io/port: "8080"}` |
+| `spark.metrics.workerAnnotations`       | Annotations for enabling prometheus to access the metrics endpoint of the worker nodes   | `{prometheus.io/scrape: "true", prometheus.io/path: "/metrics/", prometheus.io/port: "8081"}` |
+| `spark.metrics.resources.requests` | The requested resources for the metrics exporter container  | Spark exporter container resource requests for optimal resource usage size|
```
carrodher, line 171: The wavefront subcharts parameters should be added to this table

bitnami/dataplatform-bp1/values.yaml
miguelaeh: No, it would not be rendered in the subchart. I think we should use here just a boolean value here.

bitnami/dataplatform-bp1/values.yaml
miguelaeh: AFAIK this is how the OCTO team wants the Helm Chart to be deployed by default. You can also see other affinity rules set in the rest of components

miguelaeh (changes requested):
Thank you for the PR!
Please check the comments

carrodher (commented):
Thanks for the contribution, I just added some comments mainly related to some generic files like Chart.yaml, deps and README

bitnami/dataplatform-bp1/values.yaml
gkulkarni32: yes thats right @miguelaeh

bitnami/dataplatform-bp1/Chart.yaml
gkulkarni32: can you please give me the link, will update

bitnami/dataplatform-bp1/values.schema.json
gkulkarni32: yes, we would like to have a form layout as well in kubeapps for user experience.

bitnami/dataplatform-bp1/Chart.yaml
gkulkarni32: In the first release, the appVersion was 1.0.0 and the chart version was 0.1.0. 
According to us, this is a major release for the data platform chart, hence we went with 2.0.0 and 1.0.0.

bitnami/dataplatform-bp1/README.md
gkulkarni32: done

bitnami/dataplatform-bp1/values.yaml
gkulkarni32: I changed this, please check.

bitnami/dataplatform-bp1/values.yaml
marcosbc: Fine then

bitnami/dataplatform-bp1/values.yaml
marcosbc: @gkulkarni32 Could you change this to use a plain boolean then?

marcosbc:
Added a request to remove the templating from `metrics.enabled` keys, as it would not be rendered properly in the subchart.

bitnami/dataplatform-bp1/Chart.yaml
carrodher: The `appVersion` is not used by Helm to generate the package tarball, it is considered as metadata. This version points to the version of the bundled application in the Helm Chart. For example, in the PostgreSQL Helm chart, this field is `appVersion: 11.11.0` which is the version of the PostgreSQL container used by the Helm Chart.

Then we have the `version` field which is the version of the chart itself (nothing related to the application inside the chart). This version should be modified every time the chart is modified (changes in the values.yaml, notes.txt, k8s manifests, etc). This version follows [semver](https://semver.org/spec/v2.0.0.html). According to semver, the patch version is bumped when the changes are fixing bugs, the minor version when some feature is added and the major version when the backward compatibility is not guaranteed. +info: https://helm.sh/docs/topics/charts/#charts-and-versioning

---------------------------------------

The above explanation is how the version works in Helm as a general rule. In the case of this dataplatform chart, there are some exceptions because there is not any main component bundled in the chart to be used to set the `appVersion` according to it, we can say this chart is not bundling an application so we can consider the chart itself as the product/application, that is why we are suggesting to use the same version in the `appVersion` and `version` field.

In the rest of the charts, the `appVersion` is set according to the upstream maintainers, I mean, if PostgreSQL releases a new version such as 12.4.5 and we update the chart to use it, the `appVersion` is going to reflect this change.

The `version` is not metadata, this field is used to package the chart and it is important to follow the semver directives

bitnami/dataplatform-bp1/Chart.yaml
carrodher: Per the above explanation, this field should be updated using semver, in this case as it is a feature addition and not a backward incompatibility we should bump the `version` in a minor.

gkulkarni32: Updated the version as per the comments. Chart version now is 0.2.0 and appVersion is 1.0.0.

marcosbc (approved):
LGTM! I would also like to get a review from another team member before merging.

bitnami/dataplatform-bp1/Chart.yaml
carrodher: @miguelaeh maybe you have more context about this topic and I'm missing something, what do you think the relationship between `appVersion` and `version` should be?

gkulkarni32: I have made the changes to the versions, please have a look.

carrodher: Thanks for the change, now the `version` was bumped in a minor what is fine, but the `appVersion` field is still pointing to `1.0.0` that was the original value (probably wrong in the first iteration of the Helm chart).

The topic is: `version` is going to be updated at the moment in which there is any change in the chart (even a minor change in a comment), but what about `appVersion`? The purpose of this field is to serve as metadata to show what is the version used by the bundled application, but in this case, there is not a single application. What is your plan to bump `appVersion`? Where should we look to detect that there is a new version? What we usually do in Bitnami in this kind of chart where there is not a single app is set the same value used in `version` and bump both values at the same time.

gkulkarni32: Hi @carrodher Considering we had 1.0.0 as the appversion in the first one, we can let that be. 
The idea is to have different sizes of the data platform clusters.
Example: Current appversion 1.0.0 is related to 'Small' data platform cluster. Next major version i.e. appversion 2.0.0 will be when we add the 'Medium'/'Large' data platform cluster. 
Hope this clarifies.

bitnami/dataplatform-bp1/README.md
```diff
@@ -40,6 +41,8 @@ The "Small" size data platform in default configuration deploys the following:
 3. Solr with 2 nodes using the zookeeper deployed above
 4. Spark with 1 Master and 2 worker nodes
 
+One can optionally deploy the data platform with the wavefront observability framework in which case the wavefront collectors will be set up as a DaemonSet to collect the kubernetes cluster metrics to enable runtime feed into Tanzu Observability service. It will also be pre-configured to scrape the metrics from the prometheus endpoint that each application (Kafka/Spark/Solr) emits the metrics to.
```
miguelaeh, line 44:
```suggestion
The data platform can be optionally deployed with the wavefront observability framework. In that case, the wavefront collectors will be set up as a DaemonSet to collect the Kubernetes cluster metrics to enable runtime feed into the Tanzu Observability service. It will also be pre-configured to scrape the metrics from the Prometheus endpoint that each application (Kafka/Spark/Solr) emits the metrics to.
```

I would edit it a bit on this way

bitnami/dataplatform-bp1/values.yaml
```diff
@@ -131,12 +178,32 @@ spark:
...
+  ## Metrics configuration
+  ##
+  metrics:
+    enabled: false
+    ## Annotations for the Prometheus metrics on master nodes
+    ##
+    masterAnnotations:
+      prometheus.io/scrape: 'true'
+      prometheus.io/path: '/metrics/'
+      prometheus.io/port: '{{ .Values.master.webPort }}'
```
miguelaeh, line 190: I think this will not be templatized in the subchart so we should use the port value.

bitnami/dataplatform-bp1/values.yaml
```diff
@@ -131,12 +178,32 @@ spark:
...
+    ## Annotations for the Prometheus metrics on worker nodes
+    ##
+    workerAnnotations:
+      prometheus.io/scrape: 'true'
+      prometheus.io/path: '/metrics/'
+      prometheus.io/port: '{{ .Values.worker.webPort }}'
```
miguelaeh, line 196: Same here

miguelaeh (changes requested):

bitnami/dataplatform-bp1/values.yaml
gkulkarni32: Thanks for pointing this out. I updated the port to `{{ .values.spark.master.webPort}}`

bitnami/dataplatform-bp1/values.yaml
gkulkarni32: Thanks for pointing this out. I updated the port to `{{ .values.spark.worker.webPort}}`

miguelaeh (approved):
LGTM!

bitnami/dataplatform-bp1/Chart.yaml
carrodher: Ok, as it is something that you are going to handle due to the lack of application when the chart evolves

carrodher (approved):
Last request on my side, can you run `helm dep update` and add the _Chart.lock_ file to the PR? Once done, we will be able to merge the PR

gkulkarni32:
Updated Chart.lock file, Please check from your side.

PR #5606, opened 2021-02-24
[bitnami/postgresql-ha] Add serviceAccount

erpel:
**Description of the change**

This change ports the mechanism to create a serviceAccount or use an existing serviceAccount from the bitnami/postgres chart.

**Benefits**

The postgresql-ha chart can be equipped with a dedicated serviceAccount, enabling the use of serviceAccounts with the required imagePullSecrets, permissions to use PodSecurityPolicies and other reasons to use serviceAccounts other than the default one.

**Possible drawbacks**

Since the feature is disabled by default and copies the already proven mechanism of the related bitnami/postgresql chart, the risk of introducing breaking changes with this is low.

**Checklist** 


bitnami/postgresql-ha/templates/serviceaccount.yaml
```diff
@@ -0,0 +1,12 @@
+{{- if and (.Values.serviceAccount.enabled) (not .Values.serviceAccount.name) }}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  labels:
+  {{- include "common.labels.standard" . | nindent 4 }}
```
miguelaeh, line 6:
```suggestion
  labels:  {{- include "common.labels.standard" . | nindent 4 }}
```

bitnami/postgresql-ha/templates/serviceaccount.yaml
```diff
@@ -0,0 +1,12 @@
...
+  name: {{ template "common.names.fullname" . }}
```
miguelaeh, line 7:
```suggestion
  name: {{ include "common.names.fullname" . }}
```

bitnami/postgresql-ha/templates/serviceaccount.yaml
```diff
@@ -0,0 +1,12 @@
...
+  {{- if .Values.commonAnnotations }}
+  annotations: {{- include "common.tplvalues.render" ( dict "value" .Values.commonAnnotations "context" $ ) | nindent 4 }}
+  {{- end }}
+  namespace: {{ .Release.Namespace }}
```
miguelaeh, line 11:
```suggestion
  namespace: {{ .Release.Namespace | quote }}
```

The quote is needed for certain characters on the namespace

miguelaeh (changes requested):
Thank you very much for the PR!!!
Please take a look to some minor comments I left.

erpel:
Should I just apply these changes to the bitnami/postgresql chart in this PR as well? It seems like a good thing to keep them in sync.

miguelaeh:
Hi @erpel ,
I would create a different one and reference this one. Just to have one PR per chart. It is easier to review and to find the thread in case we need it in the future

miguelaeh (approved):
LGTM!

PR #5611, opened 2021-02-25
Update README.md

robinsomarajv:
Missing the parameter 'server.ingress.tls' which seems to be mandatory and set to be true for enabling SSL with a tls secret on Kuberentes cluster

Updates confined to Readme


juan131 (approved):
Thanks for the fix!

PR #5619, opened 2021-02-25
[bitnami/logstash] Add parameter to customize 'externalTrafficPolicy'

ikus060:
**Description of the change**

Add parameter to customize 'externalTrafficPolicy'

**Benefits**

Allow preserving Source-IP with MetalLB

**Possible drawbacks**

N/A

**Applicable issues**

N/A

**Additional information**

Changes copied from other charts.

**Checklist**


bitnami/logstash/Chart.yaml
```diff
@@ -23,4 +23,4 @@ name: logstash
 sources:
   - https://github.com/bitnami/bitnami-docker-logstash
   - https://www.elastic.co/products/logstash
-version: 3.1.3
+version: 3.1.4
```
migruiz4, line 26: Please bump version to `3.2.0`

bitnami/logstash/README.md
```diff
@@ -101,6 +101,7 @@ The following tables lists the configurable parameters of the Logstash chart and
 | `service.loadBalancerIP`                   | loadBalancerIP if service type is `LoadBalancer`                                                                     | `nil`                                                   |
 | `service.loadBalancerSourceRanges`         | Address that are allowed when service is LoadBalancer                                                                | `[]`                                                    |
 | `service.clusterIP`                        | Static clusterIP or None for headless services                                                                       | `nil`                                                   |
+| `service.externalTrafficPolicy`            | Enable client source IP preservation                                                                                 | `Cluster`                                               |
```
migruiz4, line 104: The description may lead to confusion, as it seems to be a true/false parameter.

I'd suggest replacing with:
```suggestion
| `service.externalTrafficPolicy`            | Sevice external traffic policy, configure to Local to preserve client source IP when using an external loadBalancer.                                                                                 | `""`                                               |```

File: bitnami/logstash/values.yaml [hunk 3]
```diff
@@ -291,6 +291,12 @@ service:
   ## Load Balancer sources
   ## https://kubernetes.io/docs/tasks/access-application-cluster/configure-cloud-provider-firewall/#restrict-access-for-loadbalancer-service
   ##
+
+  ## Enable client source IP preservation
+  ## ref http://kubernetes.io/docs/tasks/access-application-cluster/create-external-load-balancer/#preserving-the-client-source-ip
+  ##
+  externalTrafficPolicy: Cluster
```
migruiz4, line 298: Additionally to the recommendation above, this should set to `""` by default.

bitnami/logstash/templates/svc.yaml
```diff
@@ -17,6 +17,9 @@ spec:
   {{- if and (eq .Values.service.type "ClusterIP") .Values.service.clusterIP }}
   clusterIP: {{ .Values.service.clusterIP }}
   {{- end }}
+  {{- if (or (eq .Values.service.type "LoadBalancer") (eq .Values.service.type "NodePort")) }}
+  externalTrafficPolicy: {{ .Values.service.externalTrafficPolicy | quote }}
+  {{- end }}
```
migruiz4, line 22: The current changes forces `externalTrafficPolicy` to be configured explicitly when using `LoadBalancer` or `NodePort`.

Kubernetes will automatically configure this to `Cluster` by default when missing, so could you please make this setting optional?

That way we will also prevent future issues with this chart if Kubernetes changes its default value.

```suggestion
  {{- if and .Values.service.externalTrafficPolicy (or (eq .Values.service.type "LoadBalancer") (eq .Values.service.type "NodePort")) }}
  externalTrafficPolicy: {{ .Values.service.externalTrafficPolicy | quote }}
  {{- end }}
```

migruiz4 (changes requested):
Hi @ikus060,

`externalTrafficPolicy` is a great addition to have for the Logstash chart.

I've left some change suggestions below, most of them regarding this setting being mandatory in the chart when it is optional in Kubernetes.

bitnami/logstash/templates/svc.yaml
ikus060: @migruiz4 I'm a bit confused. If I'm following you recommendation, this chart will not be aligned with other chart that has this setting already. 
If I keep the default value to `externalTrafficPolicy: Cluster` in values.yaml it's the same behaviour plus it's aligned with other charts.

migruiz4: Hi @ikus060,

You are right, almost all of the charts including the `externalTrafficPolicy` parameter have it configured as `externalTrafficPolicy: Cluster` by default.

Nonetheless, there is no standard for the condition, and I think the `and .Values.service.externalTrafficPolicy` condition to only set the value when the variable has been configured would make a good addition as it would prevent misconfigurations.

I have created an internal task to revisit and discuss this topic in the future and try to align a common configuration for all the charts.

bitnami/logstash/values.yaml
migruiz4: Closing this request according to https://github.com/bitnami/charts/pull/5619#discussion_r583637858

bitnami/logstash/README.md
migruiz4: Edited the description suggestion to keep `Cluster` as the default value.

ikus060:
@migruiz4 I've made the recommended changes. Thanks to review again.

migruiz4 (approved):
Thanks to you for your contribution and for adapting it to the suggestions!

PR #5622, opened 2021-02-25
[bitnami/redis] Add sentinel updated redis master service

### LINKED ISSUE #4082: [bitnami/redis] How to Access Redis Master Node in the Redis Sentinel Deployment outside the K8s Cluster
**Which chart**:
 bitnami/redis

Step 1: Helm install the Redis Sentinel Cluster into the K8s cluster.
Step 2: Expose the Redis Sentinel Service via Ingress to make it accessible outside the cluster (Redis Port 6379, Sentinel 26379).
Step 3: Now I want to write something into Redis, I first connect to sentinel port 26379 to get the Redis Master Node IP and Port. but the returned IP and Port are not accessible outside the cluster.

How we can deploy Redis Sentinel Cluster, and make it accessible outside the cluster?

youngnicks:
**Description of the change**

Add a sentinel managed redis master service.

**Benefits**

Clients do not need sentinel to use redis failover.

Useful for clients outside of the Kubernetes Cluster.

**Possible drawbacks**

Sentinel pod has kubectl installed. Role has list and patch pod permissions.

**Applicable issues**

  - fixes #4082 

**Additional information**

**Checklist** 


bitnami/redis/Chart.yaml
```diff
@@ -25,4 +25,4 @@ name: redis
 sources:
   - https://github.com/bitnami/bitnami-docker-redis
   - http://redis.io/
-version: 12.7.7
+version: 12.7.8
```
carrodher, line 28: As it's a new feature more than a bug fix, I would bump the minor version instead
```suggestion
version: 12.8.0
```

bitnami/redis/README.md
```diff
@@ -261,6 +261,7 @@ The following table lists the configurable parameters of the Redis<sup>TM</sup>
 | `sentinel.cleanDelaySeconds`                          | Delay seconds before issuing the the cleaning in the next node                                                                                      | `5`                                                     |
 | `sentinel.configmap`                                  | Additional Redis<sup>TM</sup> configuration for the sentinel nodes (this value is evaluated as a template)                                          | `nil`                                                   |
 | `sentinel.staticID`                                   | Enable static IDs for sentinel replicas (If disabled IDs will be randomly generated on startup)                                                     | `false`                                                 |
+| `sentinel.service.createMaster`                       | Create master service pointing to the current master                                                                                                | `false`                                                 |
```
carrodher, line 264: Just to maintain the order, I would place this row after the `sentinel.service.loadBalancerIP` to reflect the place where the parameter is appearing in the values.yaml

bitnami/redis/templates/redis-node-statefulset.yaml
```diff
@@ -370,8 +374,19 @@ spec:
...
+        - name: kubectl
+          image: bitnami/kubectl
+          imagePullPolicy: {{ .Values.volumePermissions.image.pullPolicy | quote }}
+          command: ["cp", "/opt/bitnami/kubectl/bin/kubectl", "/kubectl"]
+          securityContext:
+            runAsUser: 0
+          volumeMounts:
+            - name: kubectl
+              mountPath: /kubectl
+        {{- end }}
```
carrodher, line 389: In this case, the initcontainer should be fully templatized (check below other examples). This means the image parameters should be added to the values, create the proper template in the helpers, etc

carrodher (commented):
Thanks for your contribution. I just added some inline comments; apart from that I would like @juan131 and @miguelaeh also taking a look at the PR since they're more familiar with the application itself

miguelaeh (commented):
Hi @youngnicks ,
Thank you very much for this PR and for the extended discussion in the issue with @juan131 .
I have not tested yet, but I think the solution could work as expected. Nevertheless, I have some concerns:
- It will require to use RBAC by default. Was discussed in the issue but I still think we should try to avoid it unless it is totally necessary.
- Labels changes could bring up issues when upgrading the release since it is a satefulset
- Also, in general I think this kind of logic will be really complex to maintain, it seems to be more for an operator than for a helm chart.

youngnicks:
Hi @miguelaeh ,

Thank you for considering this PR.

> It will require to use RBAC by default. Was discussed in the issue but I still think we should try to avoid it unless it is totally necessary.

I don't know how to get around the RBAC requirement as we are making changes to pods.

> Labels changes could bring up issues when upgrading the release since it is a satefulset

I noticed I had issues with the statefulset staying around if I modified the labels defined in the statefulset, `role`. By creating a new label, `isMaster`, I haven't noticed these issues with deletions. I will do some testing to see if there is an issue with upgrades.

> Also, in general I think this kind of logic will be really complex to maintain, it seems to be more for an operator than for a helm chart.

The thought of using an operator has crossed my mind several times. I was hoping to make it work here as your chart already works well. I completely understand if you won't want this logic in your chart though.

If this is something you are wanting, I will work on making the requested changes.

youngnicks:
I tested making a few small updates (changing replica counts & downAfterMilliseconds). Everything worked as it should. Adding the additional isMaster label did not interfere with the upgrade or delete process.

miguelaeh:
Hi @youngnicks ,
thank you very much for testing it.

I feel like this change is a bit over engineered. I don't feel comfortable with changing the labels on the fly. I think it could bring up some other issues, mainly for upgrades.
Apart from that, I also think the proposed workflow will introduce some complexity that will make maintaining the chart really hard.
IMHO, we should either try to find a more simpler approach, and if it is not possible, do not support that specific scenario.

As I said, it is just my opinion, so I would like to hear what the rest of the team think, specially @rafariossaa that debugged the latest issues for this Helm Chart.

youngnicks:
Hi @miguelaeh ,

What upgrade issues are you thinking of? Through my testing, I noticed that if you update a label defined in a Kubernetes file there are issues with upgrades/deletions. Because of this, I created a new label named `isMaster` for Sentinel to use.

As far as over engineered, I am just using a feature of Sentinel to perform a script upon master change. Kubectl is used in the script to update the labels, but if there is a better way of updating pod labels I can implement that method instead. Evertyhing in this PR shouldn't need to be changed unless Sentinel changes their script which I find unlikely, or kubectl needs to be upgraded.

The only other alternative for external Redis access that I see is using a load balancer like HAProxy or NodePort.

juan131 (commented):
I agree with @miguelaeh that updating labels on the fly, while it's a valid solution, is a risky operation. So, my two cents:

- As fair as this feature remains optional and disabled by default, I'm okay with supporting it.
- A section should be included in the README.md explaining how this feature works, its pros and cons, and explicitly mentioning that RBAC rules must be created if RBAC is enabled in the target K8s cluster.

anthosz:
Hi, any plan to rollback & merge it?

carrodher:
Feel free to create a new PR adding those changes on top of the current `main` branch. We will be happy to review the changes and provide feedback.

PR #5625, opened 2021-02-25
[bitnami/common] Add app and version to common.labels.standard

Starefossen:
**Description of the change**

Add `app` and `version` labels that are comonly used by Istio.

**Benefits**

This will properly group and display the application in Istio (and other tools).

**Possible drawbacks**

A few bytes overhead

**Applicable issues**

n/a

**Additional information**

* https://istio.io/latest/docs/ops/deployment/requirements/

> Pods with app and version labels: We recommend adding an explicit app label and version label to the specification of the pods deployed using a Kubernetes Deployment. The app and version labels add contextual information to the metrics and telemetry that Istio collects.
>
>  * The app label: Each deployment should have a distinct app label with a meaningful value. The app label is used to add contextual information in distributed tracing.
>
> * The version label: This label indicates the version of the application corresponding to the particular deployment.

**Checklist** 

### DIFF
--- a/bitnami/common/Chart.yaml
+++ b/bitnami/common/Chart.yaml
@@ -2,7 +2,7 @@ annotations:
   category: Infrastructure
 apiVersion: v2
 # Please make sure that version and appVersion are always the same.
-appVersion: 1.4.1
+appVersion: 1.4.2
 description: A Library Helm Chart for grouping common logic between bitnami charts. This chart is not deployable by itself.
 home: https://github.com/bitnami/charts/tree/master/bitnami/common
 icon: https://bitnami.com/downloads/logos/bitnami-mark.png
@@ -20,4 +20,4 @@ sources:
   - https://github.com/bitnami/charts
   - http://www.bitnami.com/
 type: library
-version: 1.4.1
+version: 1.4.2
--- a/bitnami/common/templates/_labels.tpl
+++ b/bitnami/common/templates/_labels.tpl
@@ -4,6 +4,9 @@ Kubernetes standard labels
 */}}
 {{- define "common.labels.standard" -}}
 app.kubernetes.io/name: {{ include "common.names.name" . }}
+{{- if .Release.appVersion -}}
+app.kubernetes.io/version: {{ .Release.appVersion }}
+{{- end -}}
 helm.sh/chart: {{ include "common.names.chart" . }}
 app.kubernetes.io/instance: {{ .Release.Name }}
 app.kubernetes.io/managed-by: {{ .Release.Service }}


bitnami/common/Chart.yaml
```diff
@@ -2,7 +2,7 @@ annotations:
   category: Infrastructure
 apiVersion: v2
 # Please make sure that version and appVersion are always the same.
-appVersion: 1.4.1
+appVersion: 1.4.2
```
juan131, line 5: Please update the version too.

bitnami/common/templates/_labels.tpl
```diff
@@ -3,7 +3,12 @@
 Kubernetes standard labels
 */}}
 {{- define "common.labels.standard" -}}
+app: {{ include "common.names.name" . }}
```
juan131, line 6: We already included this label for certain charts to be compatible with Istio (see https://github.com/bitnami/charts/pull/2286)
That said, I'd like Istio to move forward to embrace the [default k8s recommended labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/). `app`, `version`, etc. labels are not recommended by k8s since quite a lot time.
I have mixed feelings about adding this since we're duplicating labels.

bitnami/common/templates/_labels.tpl
```diff
@@ -3,7 +3,12 @@
...
 app.kubernetes.io/name: {{ include "common.names.name" . }}
+{{- if .Release.appVersion -}}
+app.kubernetes.io/version: {{ .Release.appVersion }}
+version: {{ .Release.appVersion }}
+{{- end -}}
```
juan131, line 11: We should be extremely cautious when including labels that change everytime the version is bumped (`appVersion` in this case). In the past, this lead into upgradability issues.

juan131 (changes requested):
@javsalgar @carrodher I'd like you to share your thoughts on this.

bitnami/common/templates/_labels.tpl
carrodher: My 2cts is to follow the k8s recommendations as much as possible, if there is a specific asset that needs to follow a different approach due to a widely common use case we can evaluate an exception for this kind of case. But adding non-recommended and duplicated labels to the different k8s objects in this library chart that is going to be applied to the whole catalog doesn't sound like the best option to me.

Do you know why Istio is not using the recommended labels (any open issue or similar)? I'm just thinking out loud, but I am not sure if we should create this (potentially risky) workaround to support something it's not recommended by the core maintainers.

carrodher:
Thanks for the ping, I just added my thoughts in one of the inline comments

bitnami/common/templates/_labels.tpl
Starefossen: It looks like Istio is moving more towards the recommended Kubernetes labels so this might not be as needed as I first though ref. https://github.com/istio/istio/issues/10724

Starefossen:
As per discussion https://github.com/istio/istio/issues/10724 is seams that Istio is moving more and more to adopt the now recommended Kubernetes lables and that the old labels are no longer needed.

migruiz4:
Hi @Starefossen ,

In that case, I'd suggest deleting the duplicated labels `app` and `version` from this PR.

Regarding `app.kubernetes.io/version`, I agree with @juan131 and I would recommend being extremely cautious when using labels that could change between chart versions.

I'd like to perform upgrade tests in as many charts as possible before accepting it because `common.labels.standard` is used in almost every template in this repository.

bitnami/common/Chart.yaml
Starefossen: Done ✅

bitnami/common/templates/_labels.tpl
Starefossen: I have removed the legacy labels.

Starefossen:
Done @migruiz4 ✅

migruiz4:
Hi @Starefossen, @juan131 and @carrodher,

After some investigation, looks like the `common.labels.standard` is only used in the fields `metadata.labels`(All manifests) and `spec.template.metadata.labels`(Statefuls and Deployments).

The only exception I found was:
https://github.com/bitnami/charts/blob/8af317270ab0d665cedad512d9932f4358174770/bitnami/nginx-ingress-controller/templates/controller-poddisruptionbudget.yaml#L15-L24

This will have to be modified before merging this change because `spec.selector` is indeed an [immutable field](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#label-selector-updates), but it seems to be a bug because I think we should instead be using `common.labels.matchLabels`.

So far, upgrade tests changing `metadata.labels` and `spec.template.metadata.labels` were successful, so we can assume they are not immutable.

Additionally, Helm recommends the appVersion field to be quoted so YAML doesn't interpret versions as floating-point values.

bitnami/common/templates/_labels.tpl
```diff
@@ -4,6 +4,9 @@ Kubernetes standard labels
 */}}
 {{- define "common.labels.standard" -}}
 app.kubernetes.io/name: {{ include "common.names.name" . }}
+{{- if .Release.appVersion -}}
+app.kubernetes.io/version: {{ .Release.appVersion }}
+{{- end -}}
```
migruiz4, line 9:
```suggestion
{{- if .Chart.AppVersion }}
app.kubernetes.io/version: {{ .Chart.AppVersion }}
{{- end }}
```

migruiz4 (changes requested):
Please make the changes below, as it is `.Chart.AppVersion` instead of `.Release.appVersion`.

Additionally, the `-}}` at the end will consume the next line, causing issues.

PR #5627, opened 2021-02-26
[bitnami/external-dns] Fix reference to TransIP API secret

WebSpider:
**Description of the change**

Fixes an incorrect reference to the TransIP API secret

**Benefits**

Produce a rendering chart when chosing `provider=transip`

**Possible drawbacks**

Not tested for all other providers

**Applicable issues**

None found

**Additional information**

N/A

**Checklist**
- [X ] Variables are documented in the README.md

### DIFF
--- a/bitnami/external-dns/Chart.yaml
+++ b/bitnami/external-dns/Chart.yaml
@@ -24,4 +24,4 @@ sources:
   - https://github.com/kubernetes-sigs/external-dns
   - https://github.com/bitnami/bitnami-docker-external-dns
   - https://github.com/kubernetes-sigs/external-dns
-version: 4.8.4
+version: 4.8.5
--- a/bitnami/external-dns/templates/deployment.yaml
+++ b/bitnami/external-dns/templates/deployment.yaml
@@ -618,7 +618,7 @@ spec:
         # TransIP volume(s)
         - name: transip-api-key
           secret:
-            name: {{ template "external-dns.fullname" . }}
+            secretName: {{ template "external-dns.fullname" . }}
         {{- end }}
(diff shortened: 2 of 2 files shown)


miguelaeh (changes requested):
Hi @WebSpider ,
From https://kubernetes.io/docs/concepts/storage/volumes/:
> For secrets, the secretName field has been changed to name to be consistent with ConfigMap naming.

So I am afraid we can't accept this PR

WebSpider:
The current version of this chart produces a rendering error when trying to use TransIP as provider
Also, this usage is aligned with https://kubernetes.io/docs/concepts/configuration/secret/#using-secrets and with the way other providers use secrets in this chart. This includes alibaba, aws, azure, coredns and google.

miguelaeh:
Hi @WebSpider ,
I have been trying to debug it. It seems we would need to add some logic to set `name` or `secretName` depending on the Kubernetes version. We will implement something similar to what we do with the Ingress.
Thank you for reporting it

miguelaeh (approved):
Hi @WebSpider,

I rechecked this and the link I shared is for projected volumes, not volumes so I think this PR is correct then.
Sorry for the confusion.
PLease solve the conflicts in the Chart.yaml and we can merge it.
LGTM!

carrodher:
Unfortunately, due to other priorities and the limited bandwidth we have, the work on this PR was not continued and there are no plans to work on it in the mid-term. I'm replacing the `on-hold` label with the `stale` one so the PR will be automatically closed if it is not continued during the following days.

WebSpider:
Thanks for the ping, I will pick this up again

Issue #5091, opened 2021-01-19
Notice: New Helm version prerequisite


FraPazGal:
Bitnami Helm Charts will be bumping the required Helm version from `3.0-beta3+` to `Helm v3.1.0`. This comes from our implementation of a new functionality, `lookup` function, that was introduced on Helm 3.1. This new feature will allow us to improve password's management during upgrades.

You can find more information about our implementation at https://github.com/bitnami/charts/pull/4831.

Babakbabak952:
###

PR #5635, opened 2021-02-28
[bitnami/postgresql] Don't put postgresql-postgres-password in secret when not needed

mkilchhofer:
**Description of the change**

When only using the "postgres" user, the secret gets updated every helm upgrade with a
random string. When using with argocd (and maybe other gitops tools) there are always
changes until you put a dummy value into postgresqlPostgresPassword.

Use the same conditional as in statefulset.yaml:
~~~
{{- if not (eq (include "postgresql.username" .) "postgres")  }}
  # (..)
- name: POSTGRES_POSTGRES_PASSWORD
  valueFrom:
    secretKeyRef:
      name: {{ template "postgresql.secretName" . }}
      key: postgresql-postgres-password
{{- end }}
{{- end }}
~~~

**Benefits**

Accomplish idempotency when only using the "postgres" user.

**Possible drawbacks**

`-`

**Applicable issues**

`-`

**Additional information**

`-`

**Checklist** 

### DIFF
--- a/bitnami/postgresql/Chart.yaml
+++ b/bitnami/postgresql/Chart.yaml
@@ -26,4 +26,4 @@ name: postgresql
 sources:
   - https://github.com/bitnami/bitnami-docker-postgresql
   - https://www.postgresql.org/
-version: 10.3.7
+version: 10.3.8
--- a/bitnami/postgresql/templates/secrets.yaml
+++ b/bitnami/postgresql/templates/secrets.yaml
@@ -11,7 +11,9 @@ metadata:
   namespace: {{ .Release.Namespace }}
 type: Opaque
 data:
+  {{- if not (eq (include "postgresql.username" .) "postgres")  }}
   postgresql-postgres-password: {{ include "postgresql.postgres.password" . | b64enc | quote }}
+  {{- end }}
   postgresql-password: {{ include "postgresql.password" . | b64enc | quote }}
   {{- if .Values.replication.enabled }}
   postgresql-replication-password: {{ include "postgresql.replication.password" . | b64enc | quote }}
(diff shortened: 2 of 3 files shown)


mkilchhofer:
While here, what about extending the conditional to handle `"-"` special?: 
```yaml
{{- if not (or (eq (include "postgresql.username" .) "postgres") (eq (include "postgresql.postgres.password" .) "-")) }}
```
This would support the default behavior of bitnamis docker image and disallows remote login:
> **Note!**
> When `POSTGRESQL_USERNAME` is specified, the `postgres` user is not assigned a password and as a result you cannot login remotely to the PostgreSQL server as the `postgres` user. If you still want to have access with the user `postgres`, please set the `POSTGRESQL_POSTGRES_PASSWORD` environment variable (or the content of the file specified in `POSTGRESQL_POSTGRES_PASSWORD_FILE`).

carrodher:
Taking a look at the code, there is also a similar condition in the read replicas statefulset, see https://github.com/bitnami/charts/blob/master/bitnami/postgresql/templates/statefulset-readreplicas.yaml#L179; but in the same way, is an `and` logic so it doesn't matter for the original purpose.

I think this change is not going to break the backward compatibility since the key in the secret is not going to be used when the condition in the `if` statement is met.

Apart from that, we are investigating if using the `lookup` function can solve this kind of issue (see https://github.com/bitnami/charts/issues/5091), but in the meantime, I would look for a second opinion. What do you think @javsalgar @juan131?

juan131:
Thanks for this contribution! I don't think we'll break anything neither since this key is ignored when only using the "postgres" user. So go ahead @carrodher @mkilchhofer

mkilchhofer:
Ahh, the read statefulset drifted in since this PR here: #4380

Shall I align this?

carrodher (approved):
Thanks! LGTM

PR #5665, opened 2021-03-03
[bitnami/fluentd] Add ingress for the aggregator 

ekrucio:
**Description of the change**

Added means to create Ingress for the Aggregator.

**Benefits**

Adding Ingress to the Aggregator is a good way to expose FluentD to be utilized by external servises. This is an elegant way to reuse your existing FluentD instance to collect and aggregate logs from other places except forwarders.
For example, this addition allowed us to substitute Logstash and use FluentD as a service to accept logs and forward them to ElasticSearch. We preffered FluentD since it's a part of the Cloud Native Computing Foundation and for it's easy and straight forward configuration.

**Possible drawbacks**

Currently only a single service is being exposed. If this change is accepted I can add means to expose multiple services with a single Ingress.

**Applicable issues**

None

**Additional information**

This change has been tested with https://hub.docker.com/r/bitnami/fluentd/latest and no issues were encountered.

**Checklist** 


ekrucio:
It looks like my fork was a bit old, closing this PR since this change has already been implemented.

PR #5670, opened 2021-03-03
[bitnami/minio]: Enable tls in front of container

### LINKED ISSUE #5668: [bitnami/minio]: Support MinIO server with TLS certificates
**Which chart**:
bitnami/minio:6.1.4

I want to enable tls, as described by min.io itself (here)[https://github.com/minio/minio/tree/master/docs/tls/kubernetes], in front of the min.io server to achieve encrypted transfer between my ingress and min.io servers.  
This is currently not possible because the bitnami/minio chart does not support to add custom/more volumes and volumeMounts.

I'd request you to enable the chart to add custom volumes and volumeMounts from a secret to the statefulset.

Alternatively you could introduce a switch to enable/disable the server tls which then statically adds the needed volume/volumemount

awuetz:
**Description of the change**

Introduced new values `tls.enabled` and `tls.secretName`.  
By setting these values the users can enable tls in front of the container. It is assumed, that `tls.secretName` is the name of an existing secret containing the tls certs and keys.
By enabling this, the deployment will attach another volumeMount to the container in the path `/opt/bitnami/minio/certs/` and define another volume `minio-certs` which reference the given secret via name.  

**Benefits**

An easy way to provide tls encryption in transit e.g. between an ingress and the container

**Applicable issues**
  - fixes #5668 

**Checklist** 


bitnami/minio/templates/statefulset.yaml
```diff
@@ -165,6 +165,10 @@ spec:
             - name: minio-credentials
               mountPath: /opt/bitnami/minio/secrets/
             {{- end }}
+            {{- if .Values.tls }}
```
miguelaeh, line 168:
```suggestion
            {{- if .Values.tls.enabled }}
```

bitnami/minio/templates/statefulset.yaml
```diff
@@ -173,6 +177,18 @@ spec:
           secret:
             secretName: {{ include "minio.secretName" . }}
         {{- end }}
+        {{- if .Values.tls }}
```
miguelaeh, line 180:
```suggestion
        {{- if .Values.tls.enabled }}
```

bitnami/minio/Chart.yaml
```diff
@@ -25,4 +25,4 @@ name: minio
 sources:
   - https://github.com/bitnami/bitnami-docker-minio
   - https://min.io
-version: 6.1.13
+version: 6.1.14
```
miguelaeh, line 28:
```suggestion
version: 6.2.0
```
Since we are adding new features you should bump the minor version

bitnami/minio/templates/statefulset.yaml
```diff
@@ -173,6 +177,18 @@ spec:
...
+        - name: minio-certs
+          secret:
+            secretName: {{ .Values.tls.secretName }}
```
miguelaeh, line 183: I would add a verification in the NOTES.txt to check that if TLS is enabled the `secretName` is provided, if not this will fail with a template error.

miguelaeh: Another approach could be to generate certificates if TLS is enabled but a secret was not provided as we do in the Harbor chart

miguelaeh (changes requested):

bitnami/minio/templates/statefulset.yaml
awuetz: > 
> 
> Another approach could be to generate certificates if TLS is enabled but a secret was not provided as we do in the Harbor chart

Yes could be a good idea.

bitnami/minio/templates/NOTES.txt
```diff
@@ -65,5 +65,10 @@ To access the MinIO(R) web UI:
    WARN: MinIO(R) Web UI is disabled.
 {{- end }}
 
+{{- if and (.Values.tls.enabled) (not .Values.tls.secretName) }}
+
+   NOTE: Since tls is enabled, the name of a secret (containing the certificats) has to be provided.
```
miguelaeh, line 70:
```suggestion
   ERROR: Since TLS is enabled, the name of a secret containing the certificates has to be provided.
   {{- fail -}}
```

In this way we make the installation to fail since it will not be correct instead of just warning the issue.

miguelaeh (changes requested):

miguelaeh (approved):
Thank you very much!
LGTM!

PR #5685, opened 2021-03-04
[bitnami/elasticsearch] Simplify sysctl logic

carrodher:
**Description of the change**

In previous commits, the defacto image for initcontainers ([`bitnami/minideb`](https://github.com/bitnami/minideb)) was replaced by [`bitnami/bitnami-shell`](https://github.com/bitnami/bitnami-docker-bitnami-shell). [Here](https://github.com/bitnami/charts-docs/blob/main/DESIGN_DECISIONS.md#2021-02-25-replace-bitnamiminideb-by-bitnamibitnami-shell) you can find more info about this decision.

This new image already contains `procps` (the system package providing `sysctl`) so from this moment it is not needed to use the CM (_configmap-initcontainer.yaml_) containing the bash script that installs `procps` in a different way according to the distro and run the `sysctl` command.

**Benefits**

As `procps`/`sysctl` is preinstalled in the initContainer image, the logic can be simplified to just execute the `sysctl` command, removing the CM and associated volumes.

**Possible drawbacks**

The initContainer image should contain `procps`, in this case, the image used by default for this purpose `bitnami/bitnami-shell` contains it

**Checklist**


Mauraza:
Thanks for the PR! LGTM.

PR #5705, opened 2021-03-05
[bitnami/spark] Added hostNetwork in the spark charts

a-nldisr:
Signed-off-by: Rogier Dikkes <1867776+a-nldisr@users.noreply.github.com>

**Description of the change**

The change adds hostNetwork to the spark worker and master. Gives people the ability to enable hostNetwork if they require host networking to connect to their workloads. 

**Benefits**

Enables hostNetwork in the bitnami spark chart. 

**Possible drawbacks**

The option to not use hostNetwork stays intact and therefore should have no impact. 

**Applicable issues**

None (yet), I can create an issue if requested. 

**Additional information**

The reason why I want to add this feature is that we are currently migrating from DCOS to Kubernetes. We use Spark to connect to various internal datastores which are reachable inside the DCOS cluster. In DCOS there is an ipvs based load balancer that offers internal URI's to connect from spark to the databases. While we are migrating our Spark workloads to Kubernetes we need to keep using the datastores. We therefore made an ipvs load balancer setup that bridges the Kubernetes cluster via IPVS to DCOS we however need to enable hostNetwork in all the pods that need to connect to the DCOS cluster. This feature was missing in the bitnami spark charts. 

**Checklist** 


bitnami/spark/Chart.yaml
```diff
@@ -22,4 +22,4 @@ name: spark
 sources:
   - https://github.com/bitnami/bitnami-docker-spark
   - https://spark.apache.org/
-version: 5.2.2
+version: 5.2.4
```
miguelaeh, line 25:
```suggestion
version: 5.3.0
```

miguelaeh (changes requested):

bitnami/spark/Chart.yaml
a-nldisr: Updated the version as requested.

miguelaeh (approved):
LGTM!

PR #5707, opened 2021-03-05
[bitnami/elasticsearch] use common template helpers

yevgeny-z:
**Description of the change**
Use common template helpers for elasticsearch chart. 

**Benefits**
That allow have short list of helpers for chart and standardised override names of k8s objects.
`fullnameOverride` now work in all `elasticsearch.*.fullname` templates

**Possible drawbacks**
Because `fullnameOverride` in old chart override only `elasticsearch.fullname` template. In this version name of other chart objects (master, ingest, coordinating, data, initScripts, curator) also will be override if this value set

**Applicable issues**
none

**Additional information**
none

**Checklist** 


migruiz4 (approved):
Hi @yevgeny-z ,

Thank you very much for your contribution, your PR looks good to me. This will help to align the Elasticsearch chart with the other charts by using the common template.

There was a change I wasn't completely sure we should be adding, and it is replacing `elasticsearch.curator.name` with `elasticsearch.curator.fullname`, because it could cause problems when upgrading. But a local test proved there shouldn't be any issues. (Old Role, RoleBinding, and podSecurityPolicy were removed and recreated with the new name).

PR #5735, opened 2021-03-09
[bitnami/redis-cluster] Update instruction for nodes in values.yaml

dibosh:
**Type**

Improvement of documentation/instruction to be followed

**What it tries to solve**

It might be a case, where people will override the node count(e.g. `nodes`) and will set it to something less than 3 and will not necessarily understand the inclusion of replica count. More importantly, if the node count does not allow at least 3 master nodes to be created; for example- nodes=3/4 with replicas=1, the cluster creation will fail.

This has to be clearly mentioned to avoid further confusion.


andresbono:
Hi, thanks for providing this PR! We will review it soon.

bitnami/redis-cluster/README.md
```diff
@@ -251,8 +251,8 @@ The following table lists the configurable parameters of the Redis<sup>TM</sup>
...
 | `cluster.init`                                  | Enable the creation of a job that initializes the Redis<sup>TM</sup> Cluster                                                                      | `true`         |
-| `cluster.nodes`                                 | Number of nodes in the Redis<sup>TM</sup> cluster                                                                                                 | `6`            |
-| `cluster.replicas`                              | Number of replicas for every master in the cluster                                                                                                | `1`            |
+| `cluster.nodes`                                 | Total Number of nodes in the Redis<sup>TM</sup> cluster including `replicas`. The cluster has to have at least 3 master nodes, and `nodes = numOfMasterNodes + numOfMasterNodes * replicas`. So `nodes = 6` and `replicas = 1` means 3 master and 3 replica nodes.                                                                                        | `6`            |
```
andresbono, line 254: The descriptions in this table should be short. Do you think that we can change it to something like the following?

```suggestion
| `cluster.nodes`                                 | Total Number of nodes in the Redis<sup>TM</sup> cluster including `replicas`. See the "Cluster topology" section                                  | `6`            |
```

Then, you can consider adding more information in https://github.com/bitnami/charts/tree/master/bitnami/redis-cluster#cluster-topology

bitnami/redis-cluster/values.yaml
```diff
@@ -47,6 +47,13 @@ cluster:
   init: true
   ## Number of Redis(TM) nodes to be deployed
   ##
+  ## Note:
+  ## This is total number of nodes including the replicas. Meaning there will be 3 master and 3 replica
+  ## nodes(as replica count is set to 1 by default, there will be 1 replica per master node).
```
andresbono, line 52:
```suggestion
  ## nodes (as replica count is set to 1 by default, there will be 1 replica per master node).
```

andresbono (commented):
Thanks again for your contribution!!

dibosh:
Hello @andresbono 

Thanks for your time to review this. I have updated according to your suggestions. Please check again. 

Thank you!

andresbono (approved):
LGTM. Thanks for the PR!!

PR #5747, opened 2021-03-10
[bitnami/thanos] introduce logFormat in Thanos components

jleloup:
**Description of the change**

Current Thanos chart does not allow users to set log format: the default `logfmt` format is implicitly used by Thanos.

This Pull Request introduce a set of Helm values to do so in each Thanos component: using `logfmt` as default to prevent any change in behaviour for users that don't want to change this format.

**Benefits**

This change allows users to set a different format for Thanos logs: especially JSON format comes to mind.

**Possible drawbacks**

No drawback identified.

**Applicable issues**

No issue currently opened related to this change.

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/thanos/templates/ruler/statefulset.yaml
+++ b/bitnami/thanos/templates/ruler/statefulset.yaml
@@ -86,6 +86,7 @@ spec:
           args:
             - rule
             - --log.level={{ .Values.ruler.logLevel }}
+            - --log.format={{ .Values.bucketweb.logFormat }}
             - --grpc-address=0.0.0.0:10901
             - --http-address=0.0.0.0:10902
             - --data-dir=/data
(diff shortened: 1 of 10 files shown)


bitnami/thanos/templates/ruler/statefulset.yaml
```diff
@@ -86,6 +86,7 @@ spec:
           args:
             - rule
             - --log.level={{ .Values.ruler.logLevel }}
+            - --log.format={{ .Values.bucketweb.logFormat }}
```
rafariossaa, line 89: I think there is a copy& paste error here, shouldn't this be `.Values.ruler.logFormat` ?

If it is, please review the other places where `logFormat` is used.

rafariossaa (changes requested):
Hi,
  Thank you very much for your PR !.

  I think I have found an issue, could you take a look to it ?

bitnami/thanos/templates/ruler/statefulset.yaml
jleloup: Indeed, I'm caught red handed :)
I'll double check other values.

jleloup: Fixed

rafariossaa (approved):
Thanks !
This LGTM

PR #5751, opened 2021-03-10
[bitnami/jasperserver] fix secrets name reference if externaldb is used

smelchior:
**Description of the change**

Currently the externaldb secret is not referenced correctly:
https://github.com/bitnami/charts/blob/master/bitnami/jasperreports/templates/externaldb-secrets.yaml#L5
```
  name: {{ printf "%s-externaldb" (include "common.names.fullname" .) }}
```
but in https://github.com/bitnami/charts/blob/master/bitnami/jasperreports/templates/_helpers.tpl#L80
```
{{- else -}}
    {{- printf "%s-%s" .Release.Name "externaldb" -}}
{{- end -}}
```

**Additional information**

**Checklist** 


Mauraza (approved):
Hi @smelchior,

Thanks for the PR. LGTM!

PR #5757, opened 2021-03-11
[bitnami/mariadb-galera] Remove percona keyword from mariadb-galera

WMP:
Why  do you use in keyword percona, but this chart install mariadb software?


andresbono (approved):
Thanks for this PR! LGTM

PR #5760, opened 2021-03-11
[bitnami/redis] add enable flag for PSP

guyguy333:
Signed-off-by: Guillaume Delbergue <guillaume.delbergue@hiventive.com>

**Description of the change**

It allows to enable PSP without creating it on Bitnami/Redis chart. 

**Benefits**

If you need to deploy chart multiple times in the same cluster, you need to create PSP once but enable it multiple times. Currently, you can't, you have to create to enable which result in a resource conflict.

**Possible drawbacks**

**Applicable issues**

  - fixes #

**Additional information**

**Checklist** 


bitnami/redis/Chart.yaml
```diff
@@ -25,4 +25,4 @@ name: redis
 sources:
   - https://github.com/bitnami/bitnami-docker-redis
   - http://redis.io/
-version: 12.8.3
+version: 12.8.4
```
alemorcuq, line 28:
```suggestion
version: 12.9.0
```

bitnami/redis/README.md
```diff
@@ -302,6 +302,7 @@ The following table lists the configurable parameters of the Redis<sup>TM</sup>
 | `sysctlImage.mountHostSys`                            | Mount the host `/sys` folder to `/host-sys`                                                                                                         | `false`                                                 |
 | `sysctlImage.resources`                               | sysctlImage Init container CPU/Memory resource requests/limits                                                                                      | {}                                                      |
 | `podSecurityPolicy.create`                            | Specifies whether a PodSecurityPolicy should be created                                                                                             | `false`                                                 |
+| `podSecurityPolicy.enabled`                           | Enable PodSecurityPolicy                                                                                                                            | `false`                                                 |
```
alemorcuq, line 305: Can you move this above `.create`?

Also, could you update the `.create` description to mention that you also need the set `podSecurityPolicy.enabled` to true in order to enable it?

bitnami/redis/values.yaml
```diff
@@ -921,6 +921,9 @@ podSecurityPolicy:
   ## Specifies whether a PodSecurityPolicy should be created
   ##
   create: false
+  ## Enable PodSecurityPolicy
+  ##
+  enabled: false
```
alemorcuq, line 926: Same as before: please, move this above `create:` and update the `create` description to mention that you should also set `podSecurityPolicy.enabled` to true to enable it.

alemorcuq (changes requested):
Thanks for the PR. I agree with you, but I have some suggestions for the implementation. Please, see the following comments.

I also think it would be interesting to add a validation to the `_helpers.tpl` file that throws an error when `podSecurityPolicy.create` is set to `true` but `podSecurityPolicy.enabled` is set to `false`, so that users that are already deploying the chart with the `create` option set to true know that they have to also enable the PSP. What do you think?

guyguy333:
> Thanks for the PR. I agree with you, but I have some suggestions for the implementation. Please, see the following comments.
> 
> I also think it would be interesting to add a validation to the `_helpers.tpl` file that throws an error when `podSecurityPolicy.create` is set to `true` but `podSecurityPolicy.enabled` is set to `false`, so that users that are already deploying the chart with the `create` option set to true know that they have to also enable the PSP. What do you think?

Agree. I did a draft implementation, what do you think ?

juan131:
Sorry for the conflicts @guyguy333. We did a major refactorization of the chart at https://github.com/bitnami/charts/pull/6102.

Could you please rebase you branch from master and address the conflicts? Thank in advance.

guyguy333:
@juan131 I think it should be fine now. Please let me know if I need to fix something else.

bitnami/redis/templates/_helpers.tpl
```diff
@@ -71,6 +71,15 @@ Return the appropriate apiGroup for PodSecurityPolicy.
...
 {{- end -}}
 
+{{/*
+ Check PodSecurityPolicy create.
+ */}}
+ {{- define "podSecurityPolicy.checkCreate" -}}
+ {{- if and (eq .Values.podSecurityPolicy.create true) (eq .Values.podSecurityPolicy.enabled false) }}
+ {{- fail "In order to create PodSecurityPolicy, you also need to enable podSecurityPolicy.enabled field" }}
+ {{- end }}
+ {{- end }}
```
alemorcuq, line 81: We have a standard way to add values validation. Check this line: https://github.com/bitnami/charts/blob/0ade812deb74acdf1b9c79c52bfec735cd285f6c/bitnami/redis/templates/_helpers.tpl#L189

In short, we call all the validations in the same function and them show the errors in the NOTES.txt. Could you adapt the validation to this standard?

alemorcuq (changes requested):
Thanks, @guyguy333.

bitnami/redis/templates/_helpers.tpl
guyguy333: Yes, I did the update. Please let me know if that's fine for you.

bitnami/redis/values.yaml
```diff
@@ -950,9 +950,12 @@ networkPolicy:
 ## ref: https://kubernetes.io/docs/concepts/policy/pod-security-policy/
 ##
 podSecurityPolicy:
-  ## @param podSecurityPolicy.create Specifies whether a PodSecurityPolicy should be created
+  ## @param podSecurityPolicy.create Specifies whether a PodSecurityPolicy should be created. You also need to set podSecurityPolicy.enabled to true
```
juan131, line 953: Let's try to keep the des
```suggestion
  ## @param podSecurityPolicy.create Specifies whether a PodSecurityPolicy should be created (set `podSecurityPolicy.enabled` to `true` to enable it)
```

bitnami/redis/templates/master/psp.yaml
```diff
@@ -1,4 +1,5 @@
 {{- if .Values.podSecurityPolicy.create }}
+{{ template "podSecurityPolicy.checkCreate" . }}
```
juan131, line 2: You don't need to add it here since the `"redis.validateValues"` macro is already making this check.
```suggestion
```

bitnami/redis/templates/_helpers.tpl
```diff
@@ -223,3 +224,12 @@ redis: architecture
     (--set "architecture=replication,sentinel.enabled=true")
 {{- end -}}
 {{- end -}}
+
+{{/* Validate values of Redis(TM) - PodSecurityPolicy create */}}
+{{- define "redis.validateValues.podSecurityPolicy.create" -}}
+{{- if and (eq .Values.podSecurityPolicy.create true) (eq .Values.podSecurityPolicy.enabled false) }}
```
juan131, line 230: You can simplify this:
```suggestion
{{- if and .Values.podSecurityPolicy.create (not .Values.podSecurityPolicy.enabled) }}
```

juan131 (commented):
I just added some comments with small suggestions

guyguy333:
Thanks, I fixed them.

alemorcuq (approved):

juan131 (approved):
LGTM

juan131:
CI/CD tests are failing.. retrying them

PR #5769, opened 2021-03-12
[bitnami/kafka] Always include serviceAccountName on statefulset

ctailor2:
**Description of the change**

The serviceAccountName is always included on the statefulset

**Benefits**

This fixes a bug where users could not specify an existing service account name to use via the following values:
```yaml
serviceAccount:
  create: false
  name: useThisName
```

**Possible drawbacks**

No known limitations

**Applicable issues**

No applicable issues

**Additional information**

This change brings the kafka chart in line with other bitnami charts with respect to specifying the service account, like the following:
https://github.com/bitnami/charts/blob/master/bitnami/cassandra/templates/statefulset.yaml#L47
https://github.com/bitnami/charts/blob/master/bitnami/zookeeper/templates/statefulset.yaml#L46

**Checklist** 

### DIFF
--- a/bitnami/kafka/Chart.yaml
+++ b/bitnami/kafka/Chart.yaml
@@ -29,4 +29,4 @@ name: kafka
 sources:
   - https://github.com/bitnami/bitnami-docker-kafka
   - https://kafka.apache.org/
-version: 12.11.0
+version: 12.11.1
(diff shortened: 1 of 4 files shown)


juan131 (approved):
Thanks so much! Your changes totally make sense!

PR #5786, opened 2021-03-15
[bitnami/etcd] Service: Fix clientPortNameOverride and peerPortNameOverride

desaintmartin:
It seems recent refactor broke it.

**Checklist** 

### DIFF
--- a/bitnami/etcd/Chart.yaml
+++ b/bitnami/etcd/Chart.yaml
@@ -25,4 +25,4 @@ name: etcd
 sources:
   - https://github.com/bitnami/bitnami-docker-etcd
   - https://coreos.com/etcd/
-version: 6.1.0
+version: 6.1.1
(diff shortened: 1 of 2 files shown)


desaintmartin:
Side note: I needed to override client port name in order to have it working with Istio (naming it http-client, see https://github.com/bitnami/charts/blob/master/bitnami/etcd/templates/svc.yaml#L37), which has hard times understanding it is HTTP if we call it from the Pod IP (instead of service IP).
This is always the case for metrics grabbed by Prometheus. Maybe we could rename the default so that it work out of the box for other people ? It took quite some time for me to understand it.

juan131 (approved):
Thanks for the fix @desaintmartin !

Istio isn't something we support by default. That said, we know that it's very popular and some of our charts do follow the `PROTOCOL-NAME` standard (e.g. Kafka or Zookeeper).

I won't change the default name though since that can include some backwards incompatibility issues.

PR #5802, opened 2021-03-16
[bitnami/nats] fix: Quote string values in config file

joancafom:
Signed-off-by: joancafom <jcarmona@vmware.com>

**Description of the change**

When the configuration file for nats holds strings starting with numeric values, the server is unable to parse the information and the pod fails to start:

```
$ helm install --set clusterAuth.password=1234test alpha bitnami/nats

$ kubectl get pods
NAME           READY   STATUS   RESTARTS   AGE
alpha-nats-0   0/1     Error    0          2s

$ kubectl logs alpha-nats-0
nats-server: Parse error on line 29: 'Expected a map value terminator "," or a map terminator "}", but got 'e' instead.'
```

This is an error we have faced in the upcoming version of the application `2.2.0`, but does not seem to fail with the current one. However, this is a known issue and the recommended way of storing such values in the configuration file is by quoting them.

See ref: https://github.com/nats-io/nats-server/issues/716#issuecomment-406679679

This PR implements this recommendation and prevents the chart from failing when supplying strings beginning with numbers.

**Benefits**

The server can interpret all string values correctly and does not fail-exit on start up.

**Possible drawbacks**

None

**Applicable issues**

N/A

**Additional information**

N/A

**Checklist** 

### DIFF
--- a/bitnami/nats/Chart.yaml
+++ b/bitnami/nats/Chart.yaml
@@ -24,4 +24,4 @@ name: nats
 sources:
   - https://github.com/bitnami/bitnami-docker-nats
   - https://nats.io/
-version: 6.2.2
+version: 6.2.3
--- a/bitnami/nats/Chart.yaml
+++ b/bitnami/nats/Chart.yaml
@@ -1,7 +1,7 @@
 annotations:
   category: Infrastructure
 apiVersion: v2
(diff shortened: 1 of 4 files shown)


miguelaeh (approved):
LGTM!

PR #5806, opened 2021-03-17
[bitnami/harbor] Use notary secret for notary ingress

### LINKED ISSUE #5770: [bitnami/harbor] Notary Ingress secret using core ingress secret
**Which chart**:
[bitnami/harbor]

**Describe the bug**
I've been deploying this in an EKS v1.19 cluster running nginx-ingress controller and cert-manager v1.2.0 and have noticed that cert-manager creates two certificates but only a single secret.

Looking into the code, the core ingress secret is also referenced in the notary ingress object. When in this configuration, visiting https://harbor.mydomain.com would warn with a CA invalid prompt. Investigation of the certificate showed that it was from the generic ingress certificate. While visiting https://notary.mydomain.com returned the expected LetsEncrypt certificate.

I changed the below value to `{{ include "harbor.ingress-notary" }}` and cert-manager created two secrets and both domain endpoints return valid LetsEncrypt certificates

https://github.com/bitnami/charts/blob/5fd202d22c8af0546b31f2c11cf89d5e8a5dd408/bitnami/harbor/templates/ingress/ingress.yaml#L122

**To Reproduce**

1. Deploy Harbor into a cluster using nginx-ingress controller and cert-manager
2. Visit both domains. Notice that the harbor domain returns a Kubernetes ingress certificate while notary returns a valid LetsEncrypt certificate

**Expected behavior**
Both domains should return a valid LetsEncrypt certificate

- Output of `helm version`:

```
version.BuildInfo{Version:"v3.2.4", GitCommit:"0ad800ef43d3b826f31a5ad8dfbb4fe05d143688", GitTreeState:"clean", GoVersion:"go1.13.12"}
```

```
Client Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.3", GitCommit:"01849e73f3c86211f05533c2e807736e776fcf29", GitTreeState:"clean", BuildDate:"2021-02-18T12:10:55Z", GoVersion:"go1.15.8", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"19+", GitVersion:"v1.19.6-eks-49a6c0", GitCommit:"49a6c0bf091506e7bafcdb1b142351b69363355a", GitTreeState:"clean", BuildDate:"2020-12-23T22:10:21Z", GoVersion:"go1.15.5", Compiler:"gc", Platform:"linux/amd64"}
```

**Additional context**
I'm still just getting started with Harbor and Notary, so it's possible that this works in some cases, but it does not when using my setup described above. I suppose it is possible that the original context would work if a wildcard cert was provided to the `harbor-ingress` secret.

chrisbsmith:
Forces the Notary Ingress TLS secret to be stored in a separate secret than the Harbor Ingress TLS secret. When they both use the same secret, you would have to use a wildcard cert for this to work. 

**Benefits**

Allows both Harbor and Notary Ingress endpoints to have their own secrets which allows them to present valid TLS certs to clients.

**Possible drawbacks**

No known limitations

**Applicable issues**

  - fixes #5770 

**Checklist** 

### DIFF
--- a/bitnami/harbor/Chart.yaml
+++ b/bitnami/harbor/Chart.yaml
@@ -34,4 +34,4 @@ sources:
   - https://github.com/bitnami/bitnami-docker-harbor-registry
   - https://github.com/bitnami/bitnami-docker-harbor-registryctl
   - https://goharbor.io/
-version: 9.8.1
+version: 9.8.2
--- a/bitnami/harbor/templates/ingress/ingress.yaml
+++ b/bitnami/harbor/templates/ingress/ingress.yaml
@@ -122,7 +122,7 @@ spec:
     {{- else if .Values.service.tls.existingSecret }}
     - secretName: {{ .Values.service.tls.existingSecret }}
     {{- else }}
-    - secretName: {{ include "harbor.ingress" . }}
+    - secretName: {{ include "harbor.ingress-notary" . }}
     {{- end }}
       {{- if .Values.ingress.hosts.notary }}
       hosts:
--- a/bitnami/harbor/templates/ingress/secret.yaml
+++ b/bitnami/harbor/templates/ingress/secret.yaml
@@ -1,10 +1,28 @@
 {{- if eq (include "harbor.autoGenCertForIngress" .) "true" }}
 {{- $ca := genCA "harbor-ca" 365 }}
-{{- $cert := genSignedCert .Values.ingress.hosts.core nil (list .Values.ingress.hosts.core .Values.ingress.hosts.notary) 365 $ca }}
+{{- $cert := genSignedCert .Values.ingress.hosts.core nil (list .Values.ingress.hosts.core ) 365 $ca }}
 apiVersion: v1
 kind: Secret
 metadata:
   name: {{ include "harbor.ingress" . }}
+  labels: {{- include "common.labels.standard" . | nindent 4 }}
+    {{- if .Values.commonLabels }}
+    {{- include "common.tplvalues.render" ( dict "value" .Values.commonLabels "context" $ ) | nindent 4 }}
+    {{- end }}
+  {{- if .Values.commonAnnotations }}
+  annotations: {{- include "common.tplvalues.render" ( dict "value" .Values.commonAnnotations "context" $ ) | nindent 4 }}
+  {{- end }}
+type: kubernetes.io/tls
+data:
+  tls.crt: {{ $cert.Cert | b64enc | quote }}
+  tls.key: {{ $cert.Key | b64enc | quote }}
+  ca.crt: {{ $ca.Cert | b64enc | quote }}
+---
+{{- $cert := genSignedCert .Values.ingress.hosts.notary nil (list .Values.ingress.hosts.notary) 365 $ca }}
+apiVersion: v1
+kind: Secret
+metadata:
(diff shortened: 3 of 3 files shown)


chrisbsmith:
> Hi,
> 
>   I am not very sure on this PR. 
> 
>   There is a [notary secretName](https://github.com/bitnami/charts/tree/master/bitnami/harbor#user-content-notary-parameters) already. 
> 
>   Didn't it work for you ?

Hi, the PR doesn't add a new secret, rather it just uses the already existing notary secret for the notary ingress object. 

The current config uses the harbor secret both both harbor and notary ingress. This just ensures the notary ingress uses the right secret.

rafariossaa:
Sorry, I didn't explain myself very well.
What I wanted to mean is that you can pass the same secret in `notary.secretName` and not changes in code would be necessary. These settings (`xxxx.secretName`) allows to have different secrets, hence certificates, for each of the elements.

Would this work for you ?

chrisbsmith:
I see what you're saying, though I would argue that the current implementation does not use the `[harbor/notary].secretName` for either ingress object.  Instead, a template is used and add a `-ingress` flag to the full name of the object.  

Since I used cert-manager, I just needed to provide a secret name that cert-manager could save the secret to. Both Harbor and Notary cannot share the same secret, hence, my using the notary-ingress template.

rafariossaa:
Hi,
  Could you explain a little further how you are deploying harbor and what you want to achieve ?
  It looks for me that maybe what you need is other ingress to access notary.

chrisbsmith:
Hi @rafariossaa , actually the ingress object at https://github.com/bitnami/charts/blob/master/bitnami/harbor/templates/ingress/ingress.yaml already provides ingress for [harbor](https://github.com/bitnami/charts/blob/0119143eb22a3def9382dde44136c8d03966b313/bitnami/harbor/templates/ingress/ingress.yaml#L25-L98) and [notary](https://github.com/bitnami/charts/blob/0119143eb22a3def9382dde44136c8d03966b313/bitnami/harbor/templates/ingress/ingress.yaml#L100-L147)

They're tucked into the same file, so I missed it at first.

Using these two ingress options to Harbor and to Notary allow me to sign and push my images to the harbor docker registry.

rafariossaa:
Hi,

> Using these two ingress options to Harbor and to Notary allow me to sign and push my images to the harbor docker registry.

I am not very sure if with this you were able to solve the issue and this PR is not longer needed.

rafariossaa:
Ok, I see.

In that case a similar secret to [this](https://github.com/bitnami/charts/blob/master/bitnami/harbor/templates/ingress/secret.yaml) need to be created for notary.

chrisbsmith:
I missed that part because I never did the auto cert generation.  Thanks for pointing it out.  I'll update my PR with these changes and rebase.

rafariossaa:
Hi,
  Don't worry, we will be glad to review and merge it.

chrisbsmith:
hi @rafariossaa , I've updated the PR.  The original code created a single certificate secret with a SAN for both harbor and notary. This worked when you used self signed certs.  My update creates two certificate secrets; one for harbor and one for notary. This fixes the naming collision if multiple certs are needed; for example when using cert-manager.

rafariossaa (approved):
Thank you very much!.
LGTM.

PR #5813, opened 2021-03-17
[bitnami/wavefront] Move collector image to scratch

miguelaeh:
**Description of the change**
As part of the process of moving the wavefront image to be scratch this PR adds the needed changes to find the executable inside the scratch container. We are bumping the minor version in order to not use just a patch for this change.

**Benefits**

**Possible drawbacks**

**Applicable issues**

  - fixes #

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/wavefront/Chart.yaml
+++ b/bitnami/wavefront/Chart.yaml
@@ -30,4 +30,4 @@ sources:
   - https://github.com/bitnami/bitnami-docker-wavefront-proxy
   - https://github.com/wavefrontHQ/wavefront-collector-for-kubernetes
   - https://github.com/wavefrontHQ/wavefront-proxy
-version: 1.4.0
+version: 2.0.0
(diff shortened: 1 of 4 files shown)


bitnami/wavefront/Chart.yaml
```diff
@@ -30,4 +30,4 @@ sources:
   - https://github.com/bitnami/bitnami-docker-wavefront-proxy
   - https://github.com/wavefrontHQ/wavefront-collector-for-kubernetes
   - https://github.com/wavefrontHQ/wavefront-proxy
-version: 1.3.1
```
javsalgar, line 33: I'm not sure sure bumping the minor version is enough. By changing the container to scratch there are some features that will work differently. For example, those using the `.Values.command` section expecting a container with a shell will fail, the same with the custom liveness probes and other parts. Maybe, just because of that, we could consider bumping the major version just in case.

javsalgar (approved):

PR #5575, opened 2021-02-20
[bitnami/metrics-server] Add probes, SecurityContext and command config

mkilchhofer:
**Description of the change**

This change adds the ability to configure:
- liveness and readiness probes
- container and pod level SecurityContext
- command (useful when using custom images)

Probes and containerSecurityContext are turned on by default, as used in upstream (https://github.com/kubernetes-sigs/metrics-server/releases/download/v0.4.2/components.yaml).

**Benefits**

Probes, SecurityContext (pod and container level) and command can be configured.

**Possible drawbacks**

`-`

**Applicable issues**

`-`

**Additional information**

**Checklist** 
- [X] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [X] Variables are documented in the README.md
- [X] Title of the PR starts with chart name (e.g. `[bitnami/chart]`)

:warning: Keep in mind that if you want to make changes to the kubeapps chart, please implement them in the [kubeapps repository](https://github.com/kubeapps/kubeapps/tree/master/chart/kubeapps). This is only a synchronized mirror.

PR #5820, opened 2021-03-17
[Bitnami/metrics-server] Removed unused Symlink

### LINKED ISSUE #5800: Symlink in the chart repository
**Which chart**:
metrics-server

**Describe the bug**
We tried to import the github repository for testing our PR's, but we ran into issues because Rancher gives an error that there is a symlink in the repository. 

**To Reproduce**

1. Setup Rancher
2. Click on 'Apps' in cluster explorer.
3. Try to import the github repository directly instead of using the charts webpage
4. See error: `symlink found at path management-state/git-repo/bitnami-symlink-test/760ab41eef8b4545d286988ac8f6603204ad0112ac874dfd7c6ee131a119d32a/bitnami/metrics-server/ci/ct-values.yaml`

Is this symlink necessary? We can remove it in our tests and creates a bit of overhead when we want to make PR's, but its the only chart that has this so it could be unintended.

a-nldisr:
Signed-off-by: Rogier Dikkes <1867776+a-nldisr@users.noreply.github.com>

**Description of the change**

Removed a symlink that was unused.

Removes CI/CD link, it is a symlink and there is already another file doing the same. 
When we tried to import the GitHub URL directly in Rancher 2.5.6 we ran into the issue that it detected the symlink. Rancher breaks. After investigation we found out that the symlink is not needed in this repo. 

**Benefits**

No longer breaks Rancher Apps imports. When 

**Possible drawbacks**
Unknown

**Applicable issues**

  - fixes #5800

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/metrics-server/ci/ct-values.yaml
+++ b/bitnami/metrics-server/ci/ct-values.yaml
@@ -1,1 +1,0 @@
-values-with-rbac.yaml
--- a/bitnami/metrics-server/Chart.yaml
+++ b/bitnami/metrics-server/Chart.yaml
@@ -23,4 +23,4 @@ name: metrics-server
 sources:
   - https://github.com/bitnami/bitnami-docker-metrics-server
   - https://github.com/kubernetes-incubator/metrics-server
-version: 5.8.0
+version: 5.8.1


andresbono:
Hi @a-nldisr, just to clarify, the symlink is actually used, but it doesn't add much extra value. Please check the following comments for context:

https://github.com/bitnami/charts/pull/5575#issuecomment-782733134
https://github.com/bitnami/charts/pull/5575#issuecomment-784177259

andresbono (approved):

bitnami/metrics-server/ci/ct-values.yaml
```diff
@@ -1 +0,0 @@
-values-with-rbac.yaml
```
andresbono, line 1: What if you copy the file instead of deleting the symlink?

a-nldisr: I will try that soon, based on the answer in issue #5800 I assumed it didn't add any value thus could be removed. Going to copy the file and update the PR next week.

PR #5522, opened 2021-02-17
[bitnami/wordpress] Add support for ingressClassName

migruiz4:
**Description of the change**

This PR modifies `bitnami/wordpress` and `bitnami/common` adding compatibility for `ingressClassName`.
https://kubernetes.io/blog/2020/04/02/improvements-to-the-ingress-api-in-kubernetes-1.18/

Decided to add a common function considering it would be useful to make other charts compatible with this feature in the future.

**Benefits**

Adds compatibility for `ingressClassName` preserving backward compatibility.

**Possible drawbacks**

None known.

**Applicable issues**

  - fixes #5514

**Additional information**

**Checklist** 
- [x] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [x] Variables are documented in the README.md
- [x] Title of the PR starts with chart name (e.g. `[bitnami/chart]`)

:warning: Keep in mind that if you want to make changes to the kubeapps chart, please implement them in the [kubeapps repository](https://github.com/kubeapps/kubeapps/tree/master/chart/kubeapps). This is only a synchronized mirror.

PR #5841, opened 2021-03-18
[bitnami/common] Add ingress.supportsIngressClassname

migruiz4:
**Description of the change**

This PR is complementary to #5522.

**Benefits**

Adds a new helper to check if IngressClassname field is supported or not.

**Possible drawbacks**

None known.

**Checklist** 

### DIFF
--- a/bitnami/common/templates/_ingress.tpl
+++ b/bitnami/common/templates/_ingress.tpl
@@ -40,3 +40,14 @@ Usage:
 {{- print "true" -}}
 {{- end -}}
 {{- end -}}
+
+{{/*
+Returns true if the ingressClassname field is supported
+Usage:
+{{ include "common.ingress.supportsIngressClassname" . }}
+*/}}
+{{- define "common.ingress.supportsIngressClassname" -}}
+{{- if semverCompare ">=1.18-0" (include "common.capabilities.kubeVersion" .) -}}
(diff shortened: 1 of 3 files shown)


bitnami/common/templates/_ingress.tpl
```diff
@@ -40,3 +40,14 @@ Usage:
 {{- print "true" -}}
 {{- end -}}
 {{- end -}}
+
+{{/*
+Returns true if the ingressClassname field is supported
+Usage:
+{{ include "common.ingress.supportsIngressClassname" . }}
+*/}}
+{{- define "common.ingress.supportsIngressClassname" -}}
```
javsalgar, line 49: I would make it in a similar way as the macro above. In order to avoid issues with the functions, I'm more into using strings instead of return `true` or `nothing` (ideally we should return `false` but it won't work)

migruiz4: Indeed, that is the only helper in the whole repository that uses `print "true"`:

```
> grep -R 'print "true"' .                   
./charts/bitnami/common/templates/_ingress.tpl:{{- print "true" -}}
> grep -R 'print "false"' .
./charts/bitnami/common/templates/_ingress.tpl:{{- print "false" -}}
```

PR #5787, opened 2021-03-15
[bitnami/minio] Extracted mountPath to be flexible and set the default value

awuetz:
**Description of the change**
extracted the mountPath variable to be flexible to change the location where the secret will be mounted.

**Benefits**

flexibility for the users

**Checklist** 
- [X] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [X] Variables are documented in the README.md
- [X] Title of the PR starts with chart name (e.g. `[bitnami/chart]`)

:warning: Keep in mind that if you want to make changes to the kubeapps chart, please implement them in the [kubeapps repository](https://github.com/kubeapps/kubeapps/tree/master/chart/kubeapps). This is only a synchronized mirror.

PR #5847, opened 2021-03-19
[bitnami/minio] adopt livenessProbe scheme according to tls.enabled status

awuetz:
**Description of the change**

The livenessProbe scheme is set per default to HTTP. 
If the tls at the container is enabled, the livenessProbe will fail, because it still tries HTTP requests. This change is to reflect the tls settings in the livenessProbe scheme to enalbe HTTPS if tls is used.

**Benefits**

A working livenessProbe without the need to use a customLivenessProbe.

**Additional information**

As workaround you can define customLivenessProbe, but I think the chart should be consistent and deliver a working application.

**Checklist** 


bitnami/minio/Chart.yaml
```diff
@@ -25,4 +25,4 @@ name: minio
 sources:
   - https://github.com/bitnami/bitnami-docker-minio
   - https://min.io
-version: 6.5.0
+version: 6.6.0
```
migruiz4, line 28: Could you please replace it with a PATH version bump? (6.5.1)

We use [semver](https://semver.org/) and this would be a PATCH version as it is a backward-compatible bug fix and not a new functionality.

migruiz4 (changes requested):
Hi @awuetz ,

Thank you for your contribution! Your change looks good to me, but I see your PR includes commits from your previous PR (#5787).

Could you please update your repository and use a new branch for your PR?

awuetz:
@migruiz4 i will create a new clean PR.

PR #5855, opened 2021-03-19
[bitnami/minio] Define the livenessProbe scheme value if tls.enabled, otherwise the c…

awuetz:
**Description of the change**

The livenessProbe scheme is set per default to HTTP. 
If the tls at the container is enabled, the livenessProbe will fail, because it still tries HTTP requests. This change is to reflect the tls settings in the livenessProbe scheme to enalbe HTTPS if tls is used.

**Benefits**

A working livenessProbe without the need to use a customLivenessProbe.

**Additional information**

As workaround you can define customLivenessProbe, but I think the chart should be consistent and deliver a working application.
Additional clean PR as replacement for PR https://github.com/bitnami/charts/pull/5847

**Checklist** 

### DIFF
--- a/bitnami/minio/Chart.yaml
+++ b/bitnami/minio/Chart.yaml
@@ -25,4 +25,4 @@ name: minio
 sources:
   - https://github.com/bitnami/bitnami-docker-minio
   - https://min.io
-version: 6.6.0
+version: 6.6.1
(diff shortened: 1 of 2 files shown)


alemorcuq (approved):
Thanks!

PR #5865, opened 2021-03-22
[bitnami/postgresql] Fix postgresql.conf generation from values

luhahn:
Currently the postgresql chart will fail if postgresqlConfiguration contains strings, since those strings are not contained 
in the configuration file.

```yaml
postgresql:
  postgresqlConfiguration:
    listenAddresses: '*'
```
-> results in:

listen_addresses: *

This can be fixed with this PR.
I've already tested this change on my cluster.

### DIFF
--- a/bitnami/postgresql/templates/configmap.yaml
+++ b/bitnami/postgresql/templates/configmap.yaml
@@ -15,7 +15,11 @@ data:
 {{- else if .Values.postgresqlConfiguration }}
   postgresql.conf: |
 {{- range $key, $value := default dict .Values.postgresqlConfiguration }}
-    {{ $key | snakecase }}={{ $value }}
+    {{- if kindIs "string" $value }}
+    {{ $key | snakecase }} = '{{ $value }}'
+    {{- else }}
+    {{ $key | snakecase }} = {{ $value }}
+    {{- end }}
 {{- end }}
 {{- end }}
 {{- if (.Files.Glob "files/pg_hba.conf") }}
--- a/bitnami/postgresql/templates/extended-config-configmap.yaml
+++ b/bitnami/postgresql/templates/extended-config-configmap.yaml
@@ -16,7 +16,11 @@ data:
 {{ with .Values.postgresqlExtendedConf }}
   override.conf: |
 {{- range $key, $value := . }}
-    {{ $key | snakecase }}={{ $value }}
+    {{- if kindIs "string" $value }}
+    {{ $key | snakecase }} = '{{ $value }}'
+    {{- else }}
+    {{ $key | snakecase }} = {{ $value }}
+    {{- end }}
 {{- end }}
 {{- end }}
 {{- end }}
--- a/bitnami/postgresql/Chart.yaml
+++ b/bitnami/postgresql/Chart.yaml
@@ -26,4 +26,4 @@ name: postgresql
 sources:
   - https://github.com/bitnami/bitnami-docker-postgresql
   - https://www.postgresql.org/
-version: 10.3.13
+version: 10.3.14


Mauraza:
Hi @luhahn,

Thanks for the contribution! Did you try this?

```yaml
postgresqlConfiguration:
  listenAddresses: "'*'"
```

luhahn:
Ok, i just tested your suggestion and yes this would work with the current state of the chart.
However i think, that the solution in this PR would be more intuitive :)

bitnami/postgresql/templates/configmap.yaml
```diff
@@ -15,7 +15,11 @@ data:
 {{- else if .Values.postgresqlConfiguration }}
   postgresql.conf: |
 {{- range $key, $value := default dict .Values.postgresqlConfiguration }}
-    {{ $key | snakecase }}={{ $value }}
+    {{- if kindIs "string" $value }}
+    {{ $key | snakecase }} = '{{ $value }}'
+    {{- else }}
+    {{ $key | snakecase }} = {{ $value }}
+    {{- end }}
```
Mauraza, line 22: What do you think about change this?
change this lines
```yaml
postgresql.conf: |
{{- range $key, $value := default dict .Values.postgresqlConfiguration }}
    {{- if kindIs "string" $value }}
    {{ $key | snakecase }} = '{{ $value }}'
    {{- else }}
    {{ $key | snakecase }} = {{ $value }}
    {{- end }}
```
to 
```yaml
  postgresql.conf: |
    {{- include "common.tplvalues.render" ( dict "value" .Values.postgresqlConfiguration "context" $ ) | nindent 4 }}
```

bitnami/postgresql/templates/extended-config-configmap.yaml
```diff
@@ -16,7 +16,11 @@ data:
 {{ with .Values.postgresqlExtendedConf }}
   override.conf: |
 {{- range $key, $value := . }}
-    {{ $key | snakecase }}={{ $value }}
+    {{- if kindIs "string" $value }}
+    {{ $key | snakecase }} = '{{ $value }}'
+    {{- else }}
+    {{ $key | snakecase }} = {{ $value }}
+    {{- end }}
```
Mauraza, line 23: here will be the same.

Mauraza (commented):
Thanks for the PR. Could you check the comments?

bitnami/postgresql/templates/configmap.yaml
luhahn: Sorry for the late response, had a busy week.
However, correct me if i'm wrong, but the common.tplvalues.render does not include the detection of strings and quoting them?
This would be the exact same problem as before.

luhahn: Ok, I saw that there is some kind of string detection, however this is not sufficient in order to have all values correctly transformed:

```yaml
postgresqlConfiguration:
  maxConnections: 1000
  idleInTransactionSessionTimeout: 2000
  listenAddresses: "*"
  sharedBuffers: 128MB
  dynamicSharedMemoryType: 'posix'
  maxWalSize: 1GB
  minWalSize: 80MB
  logTimezone: 'UTC'
  datestyle: 'iso, mdy'
  timezone: 'UTC'
  defaultTextSearchConfig: 'pg_catalog.english'
  ssl: "on"
  ssl_ca_file: "/opt/ngi/certs/root.crt"
  ssl_cert_file: "/opt/ngi/certs/server.crt"
  ssl_key_file: "/opt/ngi/certs/server.key"
```

results in:

```
  postgresql.conf: |
    datestyle: iso, mdy
    defaultTextSearchConfig: pg_catalog.english
    dynamicSharedMemoryType: posix
    idleInTransactionSessionTimeout: 2000
    listenAddresses: '*'
    logTimezone: UTC
    maxConnections: 1000
    maxWalSize: 1GB
    minWalSize: 80MB
    sharedBuffers: 128MB
    ssl: "on"
    ssl_ca_file: /opt/ngi/certs/root.crt
    ssl_cert_file: /opt/ngi/certs/server.crt
    ssl_key_file: /opt/ngi/certs/server.key
    timezone: UTC
```

marcosbc (approved):
LGTM, thanks!

Issue #5734, opened 2021-03-09
[RabbitMQ] Credential issues after reinstall


mladedav:
**Which chart**:
RabbitMQ 8.11.2
(possibly other charts have similar issues if I understand correctly)

**Is your feature request related to a problem? Please describe.**
Credentials persist between reinstallation of charts (as described in the troubleshooting guide.

When trying to upgrade the chart it notifies developers that they need to supply credentials. I though I was too smart for that and simply reinstalling it would do since I didn't mind losing any data. It took me a few hours to figure out that the original credentials I luckily still had in one of my terminals still worked.

**Describe the solution you'd like**
Adding more explanation to the upgrade error that uninstall/install doesn't work either. I think a bit of credentials explanation (one sentence with a link to the troubleshooting page]) could be also part of NOTES.txt, but I would understand if you wouldn't agree with that.

**Describe alternatives you've considered**
Everyone properly reads documentation and the troubleshooting page.

carrodher:
Thanks for creating this issue, you are referring to [this troubleshooting guide](https://docs.bitnami.com/general/how-to/troubleshoot-helm-chart-issues/#credential-errors-while-upgrading-chart-releases), right? In fact, the specific version is pointed in the link.

We created the guide because it is a known issue, in the same way, the guide appears in the bug report template, see
![Screenshot 2021-03-10 at 11 07 08](https://user-images.githubusercontent.com/13216600/110612735-c4164680-8190-11eb-953c-686561efbbbd.png)

Being said that, it is true that the deployment notes are also a valid place to show this kind of information since it is the first place where the user is going to see after the installation (opening an issue or looking for a guide in Google is a 2nd step). Would you like to create a Pull Request with a POC of what do you expect to see in the notes? We will be happy to review it and provide feedback, as well as adding it to other charts if needed.

Apart from that, we are investigating different alternatives to handle passwords during upgrades, such as using the `lookup` function, to get rid of this kind of issues, see https://github.com/bitnami/charts/pull/4831

PR #5871, opened 2021-03-22
[bitnami/rabbitmq] Credentials during upgrade information in notes

mladedav:
**Description of the change**

Added information about credential issues after upgrade to RabbitMQ and the general error message. I am not completely sure it is accurate as it is or if that error is used in more general settings I am unaware of, if that is the case, I can delete it and merge only on of the two changes.

**Benefits**

Clearer statement of the issue with upgrades and credentials in NOTES.txt and during upgrade.

**Possible drawbacks**

The error message not be used in this way

**Applicable issues**

https://github.com/bitnami/charts/issues/5734

**Additional information**

**Checklist** 

Version bumped both in RabbitMQ and commons


bitnami/rabbitmq/templates/NOTES.txt
```diff
@@ -7,6 +7,9 @@ Credentials:
     echo "Password      : $(kubectl get secret --namespace {{ .Release.Namespace }} {{ include "rabbitmq.secretPasswordName" . }} -o jsonpath="{.data.rabbitmq-password}" | base64 --decode)"
     echo "ErLang Cookie : $(kubectl get secret --namespace {{ .Release.Namespace }} {{ include "rabbitmq.secretErlangName" . }} -o jsonpath="{.data.rabbitmq-erlang-cookie}" | base64 --decode)"
 
+Note that the credentials are saved in persistent volume claims and will not be changed upon upgrade or reinstallation unless the persistent volume claim has been deleted. If this is not the first installation of this chart, the credentials may not be valid.
+More information about the credentials may be found at https://docs.bitnami.com/general/how-to/troubleshoot-helm-chart-issues/#credential-errors-while-upgrading-chart-releases.
```
carrodher, line 11:
```suggestion
Note that the credentials are saved in persistent volume claims and will not be changed upon upgrade or reinstallation unless the persistent volume claim has been deleted. If this is not the first installation of this chart, the credentials may not be valid.
This is applicable when no passwords are set and therefore the random password is autogenerated. In case of using a fixed password, you should specify it when upgrading.
More information about the credentials may be found at https://docs.bitnami.com/general/how-to/troubleshoot-helm-chart-issues/#credential-errors-while-upgrading-chart-releases.
```

carrodher (commented):
Thanks for your contirbution. I just added a suggestion for the rabbitmq notes. Regarding the changes in the common subchart, I prefer to maintain them separately since it is applicable to the whole catalog and it is something to be evaluated more carefully.

mladedav:
@carrodher I added the change you proposed and moved the changes in commons to [another pull request](https://github.com/bitnami/charts/pull/5884).
I have rebased the whole thing so that the history is a bit clearer.

PR #5884, opened 2021-03-23
[bitnami/common] Common credential error

mladedav:
**Description of the change**

Added information about credential issues in the general error message in commons. I am not completely sure it is accurate as it is or if that error is used in more general settings I am unaware of.

**Benefits**

Clearer statement of the issue with upgrades and credentials during upgrade when credentials are not provided.

**Possible drawbacks**

The error message may be used in context where this change would not work

**Applicable issues**

#5734

**Additional information**

Originally part of https://github.com/bitnami/charts/pull/5871

**Checklist** 


bitnami/common/templates/_errors.tpl
```diff
@@ -14,7 +14,10 @@ Required password params:
 {{- define "common.errors.upgrade.passwords.empty" -}}
   {{- $validationErrors := join "" .validationErrors -}}
   {{- if and $validationErrors .context.Release.IsUpgrade -}}
-    {{- $errorString := "\nPASSWORDS ERROR: you must provide your current passwords when upgrade the release%s" -}}
+    {{- $errorString := "\nPASSWORDS ERROR: You must provide your current passwords when upgrading the release." -}}
+    {{- $errorString = print $errorString "\n                 Note that even after reinstallation old credentials may be needed as they may be kept in persistent volume claims." -}}
```
andresbono, line 18: I think the information of the sentence below is not relevant in the context of upgrading. That's information that a user can get from the documentation link. But if the user is doing an upgrade, it is not relevant to receive information about reinstallation.

```suggestion
```

Do you agree?

bitnami/common/templates/_errors.tpl
```diff
@@ -14,7 +14,10 @@ Required password params:
...
+    {{- $errorString = print $errorString "\n                 Furhter information can be obtained at https://docs.bitnami.com/general/how-to/troubleshoot-helm-chart-issues/#credential-errors-while-upgrading-chart-releases" -}}
```
andresbono, line 19:
```suggestion
    {{- $errorString = print $errorString "\n                 Further information can be obtained at https://docs.bitnami.com/general/how-to/troubleshoot-helm-chart-issues/#credential-errors-while-upgrading-chart-releases" -}}
```

andresbono (commented):
Thanks for your contribution! We have merged master into your branch to avoid conflicts with existing changes and we left some review comments. Can you take a look at them?

bitnami/common/templates/_errors.tpl
mladedav: This is from my personal experience. When the chart told me I needed to provide credentials, I decided reinstallation would be simpler since I didn't mind losing any data and getting a new password. I don't think I would have clicked the link in the proposed change without more context because I would have assumed there is a need for credentials only during upgrade. Of course the reality was I couldn't lose any data or have a new password this way anyway.

It is true that the information is provided in the link so I wouldn't mind deleting this, it's just something that I think would have helped me in my specific case and reasoning. I would understand if that is not the expected usual case so just say the word and I will accept the change.

andresbono: If it was useful for you, might be useful for others as well.

andresbono, line 18:
```suggestion
    {{- $errorString = print $errorString "\n                 Note that even after reinstallation, old credentials may be needed as they may be kept in persistent volume claims." -}}
```

andresbono (approved):
LGTM

andresbono:
Thanks again! Your contribution is now merged.

PR #5930, opened 2021-03-26
[bitnami/kafka] HOST_IP env always available if NodePort as external service

sveba:
**Description of the change**
 
`HOST_IP` environment variable is dependent from `externalAccess.autoDiscovery.enabled` This is not right. This is fix for that issue.

**Benefits**

Fix the issue

**Possible drawbacks**

**Applicable issues**

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/postgresql/Chart.lock
+++ b/bitnami/postgresql/Chart.lock
@@ -1,6 +1,6 @@
 dependencies:
 - name: common
   repository: https://charts.bitnami.com/bitnami
-  version: 1.4.1
-digest: sha256:661d04dcbe78cf167e2bc3b7295fad641726962c1e6b2e5aa602f559c6289cef
-generated: "2021-02-24T07:26:18.305585622Z"
+  version: 1.4.2
+digest: sha256:dce0349883107e3ff103f4f17d3af4ad1ea3c7993551b1c28865867d3e53d37c
+generated: "2021-03-30T09:13:28.360322819Z"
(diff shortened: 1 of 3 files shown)


javsalgar:
Hi,

Thanks for the PR! I see no issue attached to the PR so, could you explain the issue in more detail? I see that you are removing a line that queries for the public IP, not sure if that could cause issues.

sveba:
Hi @javsalgar 

I have a setup where I need an external access with `NodePorts`. I don't need autodiscovery, because the `HOST_IP` is being populated over the Spec with `status.hostIP`. The `if .Values.externalAccess.autoDiscovery.enabled` prevents the `HOST_IP` to be popuated. This is not right

javsalgar:
Ok but why removing the curl for external access when autodiscovery is not enabled? Shouldn't it be there for the else branch?

sveba:
Hm, I think I messed up with git rebase. Now I force pushed again

The curl is actually not needed anymore. It just doesn't help in the case of NodePorts. You need the NodeIps in this case and they are being populated over `$HOST_IP`

sveba:
I just don't understand the usecase for the `curl`

sveba:
@javsalgar can you please review again

sveba:
any comment on this? we need this rather quick ;)

javsalgar:
I'm not the author of the feature but I'm very concerned on removing the logic. You say it is not necessary, assuming that the hostIP that it is returning is a public IP, right? If it is a node private IP then the external access would not work. Maybe in your installation the nodeIPs that return are public, but in other clouds these IPs could be private, so the cluster configuration would fail. Am I missing something?

PR #5942, opened 2021-03-29
[bitnami/spark] Add parameter for workers podManagementPolicy

### LINKED ISSUE #5887: Spark Pods Currently Spin Up One At A Time
**Which chart**:
bitnami/spark helm chart

I am currently running a standalone spark cluster deployed with the bitnami/spark helm chart and noticed that during pod creation each pod waits for the one before it to complete its health check prior to being created which causes the cluster spin up time to increase dramatically when deploying a decent number of worker pods.  Is that intended behavior or is this something that I may have mis-configured?  

Ideally all of the pods would spin up at the same time, but if that is not possible, then at least to spin up a bit faster.  Again, if this is something configurable that I have missed I would very much appreciate any adivce that you can offer.

alemorcuq:
**Description of the change**
Add a new parameter `worker.podManagementPolicy` to control the `podManagementPolicy` of the worker's statefulset.

**Benefits**
Users can choose to deploy the worker pods in parallel making the deployment faster.

**Possible drawbacks**
None. The default behavior is the same as it was before (`OrderedReady`)

**Applicable issues**
  - fixes #5887

**Checklist** 


Mauraza (approved):
Thanks for the PR! LGTM!!

PR #5979, opened 2021-03-31
[bitnami/wordpress] Respect deployment's strategy in case replicaCount is not defined

### LINKED ISSUE #5960: Helm doesn't respect the kubernetes deployment's strategy due to presence of replicas field in the chart
**Which chart**:
bitnami/wordpress

**Describe the bug**
When using the helm chart, without explicitly setting replicaCount in values, in conjunction with an autoscaler if you're running more than 1 pod at the time of helm release deployment strategy is not properly executed by the cluster. In this case the precedence takes the `replicas: {{ .Values.replicaCount }}` (https://github.com/bitnami/charts/blob/40075c7cee5651b2b845386e42cd2306ae625294/bitnami/wordpress/templates/deployment.yaml#L18) with default value of 1.

Assuming there are currently 10 running pods a helm release would default all pods to 1 only and then the autoscaler would restore the desired count back over time. 

**To Reproduce**

1. Create a bitnami wordpress helm release from the bitnami/wordpress chart omitting the replicaCount in values but including the following  autocaler and update strategy for example:

autoscaling:
  enabled: true
  minReplicas: 2
  maxReplicas: 11
  targetCPU: 70
  targetMemory: 70

updateStrategy:
  rollingUpdate:
    maxSurge: 1 
    maxUnavailable: 0

2. Trigger a helm upgrade to force a helm chart release.

**Expected behavior**
As part of the release process the deployment update strategy to be respected and only one pod created at a time and once it reach a status of running, ready to remove one of the existing ones and so on.

replicaCount definition in the helm chart should be fully optinal, not just its value

- Output of `helm version`:

```
version.BuildInfo{Version:"v3.5.2", GitCommit:"167aac70832d3a384f65f9745335e9fb40169dc2", GitTreeState:"dirty", GoVersion:"go1.15.7"}
```

```
Client Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.17", GitCommit:"f3abc15296f3a3f54e4ee42e830c61047b13895f", GitTreeState:"clean", BuildDate:"2021-01-13T13:21:12Z", GoVersion:"go1.13.15", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.17", GitCommit:"f3abc15296f3a3f54e4ee42e830c61047b13895f", GitTreeState:"clean", BuildDate:"2021-01-13T13:13:00Z", GoVersion:"go1.13.15", Compiler:"gc", Platform:"linux/amd64"}
```

**Additional context**
For anyone out there having the same issue I'm able to hacky mitigate the issue in CI build with query the deployment's replicas count first then add `--set replicaCount=$replicaCount` to the helm upgrade command and it that fashion the deployment strategy is properly respected.

alvneiayu:
**Description of the change**

This change gives us a support to set up the replicaCount in cases that deployment has a autoscaler enabled. This new implementation helps us to avoid problem during the upgrade.

During the upgrade, the wordpress Chart sets up always the replicaCount and if we do not specific the value by default will be 1. Setting up the value like this during the upgrade, if we have several pods running (because the autoscaler has ordered to create more than one pod), we are going to erase everything except one and then the autoscaler will order again the number of pods needed. 

**Benefits**

Solve unwanted behaviors during the upgrade. 

**Applicable issues**

  - fixes #5960

**Checklist** 

### DIFF
--- a/bitnami/wordpress/Chart.yaml
+++ b/bitnami/wordpress/Chart.yaml
@@ -35,4 +35,4 @@ name: wordpress
 sources:
   - https://github.com/bitnami/bitnami-docker-wordpress
   - http://www.wordpress.com/
-version: 10.7.3
+version: 10.8.0
--- a/bitnami/wordpress/templates/deployment.yaml
+++ b/bitnami/wordpress/templates/deployment.yaml
@@ -15,7 +15,9 @@ spec:
   {{- if .Values.updateStrategy }}
   strategy: {{- toYaml .Values.updateStrategy | nindent 4 }}
   {{- end }}
+  {{- if not .Values.autoscaling.enabled }}
   replicas: {{ .Values.replicaCount }}
+  {{- end }}
   template:
     metadata:
       labels: {{- include "common.labels.standard" . | nindent 8 }}


marcosbc (approved):
LGTM!

PR #5988, opened 2021-04-01
[bitnami/kafka] Add serviceAccount to Kafka metrics deployment and add automountServiceAccountToken to SA

mehta-ankit:
Signed-off-by: Ankit Mehta <ankit.mehta@appian.com>

**Description of the change**

The change adds serviceAccount already used for Kafka statefulset to Kafka metrics deployment as well.
In addition this change adds `automountServiceAccountToken` to the serviceAccount template. This field has a default value of `true` and that is set on the values.yaml file in this PR. This would allow someone to set it to `false` for their use case.

**Benefits**
https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#use-the-default-service-account-to-access-the-api-server
- Can use serviceAccount for kafka exporter pods
- Can control automountServiceAccountToken for SA

**Possible drawbacks**

Can't think of Any 😄 

**Applicable issues**

  - fixes #5980 

**Additional information**

**Checklist** 

- [X] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [X] Variables are documented in the README.md
- [X] Title of the PR starts with chart name (e.g. `[bitnami/chart]`)

:warning: Keep in mind that if you want to make changes to the kubeapps chart, please implement them in the [kubeapps repository](https://github.com/kubeapps/kubeapps/tree/master/chart/kubeapps). This is only a synchronized mirror.

PR #5996, opened 2021-04-02
[bitnami/zookeepr] Add automountServiceAccountToken field to the serviceaccount

mehta-ankit:
Signed-off-by: Ankit Mehta <ankit.mehta@appian.com>

**Description of the change**

This change adds `automountServiceAccountToken` to the serviceAccount template. This field has a default value of true and that is set on the values.yaml file in this PR. This would allow someone to set it to `false` for their use case.
Similar change was done for kafka chart: https://github.com/bitnami/charts/pull/5988

**Benefits**

Allows one to control auto mounting of service account token (which allows pods to talk to the K8s api).
https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#use-the-default-service-account-to-access-the-api-server
https://kubernetes.io/docs/tasks/access-application-cluster/access-cluster/#accessing-the-api-from-a-pod

**Possible drawbacks**
None

**Applicable issues**

  - fixes #

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/zookeeper/Chart.yaml
+++ b/bitnami/zookeeper/Chart.yaml
@@ -21,4 +21,4 @@ name: zookeeper
 sources:
   - https://github.com/bitnami/bitnami-docker-zookeeper
   - https://zookeeper.apache.org/
-version: 6.6.0
+version: 6.7.0
--- a/bitnami/zookeeper/README.md
+++ b/bitnami/zookeeper/README.md
@@ -143,6 +143,7 @@ The following tables lists the configurable parameters of the ZooKeeper chart an
 | `service.publishNotReadyAddresses`       | If the ZooKeeper headless service should publish DNS records for not ready pods                    | `true`                                               |
 | `serviceAccount.create`                  | Enable creation of ServiceAccount for zookeeper pod                                                | `false`                                              |
 | `serviceAccount.name`                    | The name of the service account to use. If not set and `create` is `true`, a name is generated     | Generated using the `common.names.fullname` template |
+`serviceAccount.automountServiceAccountToken` | Enable/Disable automountServiceAccountToken  for Service Account                                             | `true`                                                  |
 | `service.tls.client_enable`              | Enable tls for client connections                                                                  | `false`                                              |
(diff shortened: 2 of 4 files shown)


marcosbc (approved):
LGTM!

mehta-ankit:
@marcosbc do we need to update Chart.lock file in Kafka chart for it to be able to pick 6.7.0 version of zookeeper ??
Currently its pulling 6.5.4 when i do `helm dependency update` and that is the version in chart.lock file: https://github.com/bitnami/charts/blob/master/bitnami/kafka/Chart.lock

marcosbc:
Hi @mehta-ankit, usually our release pipeline may take a few hours to publish the new version, as it needs to pass a battery of tests plus publish in all of the chart platforms that we support.

I just ran `helm repo update` and I can now see 6.7.0 being out, so it should be fine now!

mehta-ankit:
@marcosbc Thanks for the reply.
I tried it again and same thing. To explain i am using a chart which has a dependency on kafka chart (as you are well aare which indirectly has a dependency on zookeeper chart).

When i update kafka version to 12.16.0 and try to do `helm dep update`, the charts that get pulled as `.tgz` do not have zookeeper version 6.7.0 (i checked Chart.yaml for downloaded zookeeper chart and it shows 6.5.4).

I also tried `helm pull bitnami/kafka` and it downloaded 12.16.0 kafka version of the chart (which is the latest one) but when i opened it the zookeeper chart version was 6.5.4 only.
But then after downloading 12.16.0 version of kafka chart  i tried `helm dep update` in that directory, it did downwload zookeeper 6.7.0 version chart.

But i believe this is not feasible. Having kafka as a dependecny should be enough and it should pull latest version of the zookeeper chart based on its dependencies list having Zookeeper version as `6.x.x`
image

I think i am right, but please correct me if i am wrong here. I believe updating the chart.lock file might force kafka chart to bring in the latest dependency for the zookeeper chart.

marcosbc:
Thanks for the details, I see the issue now. In this case as it was a dependency, it seems like it took a bit longer to get released. In any case it seems to be released now.

PR #5998, opened 2021-04-02
[bitnami/mongodb] Add startup probe

### LINKED ISSUE #4979: [bitnami/mongodb] Add support for startup probe
**Which chart**:
Mongodb

I had a mongodb replicaset running on GKE without an explicit cache size or ram limit set. After the replicaset started taking too much ram, I set the limit explicitly. After this, mongodb would not start due to a long running oplog truncation process that would prevent mongodb from getting ready, and thus being killed by the liveness and readiness probes.

Kubernetes official docs suggest using startup probes to protect pods that might have long running initialization processes. Adding startup probe support to the chart will be useful imo.

krtk6160:
**Description of the change**

Add a startupProbe, so that a slow starting container gets enough time to initialize without getting killed.

**Benefits**

Slow starting containers are often killed by the readiness and liveness probes before they've gotten a chance to fully initialize. Startup probes mitigate that.

**Possible drawbacks**

Available only from Kubernetes 1.16 onwards

**Applicable issues**

  - fixes #4979 

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/mongodb/Chart.yaml
+++ b/bitnami/mongodb/Chart.yaml
@@ -26,4 +26,4 @@ name: mongodb
 sources:
   - https://github.com/bitnami/bitnami-docker-mongodb
   - https://mongodb.org
-version: 10.11.3
+version: 10.12.0
--- a/bitnami/mongodb/Chart.yaml
+++ b/bitnami/mongodb/Chart.yaml
@@ -26,4 +26,4 @@ name: mongodb
 sources:
   - https://github.com/bitnami/bitnami-docker-mongodb
   - https://mongodb.org
-version: 10.11.2
+version: 10.11.3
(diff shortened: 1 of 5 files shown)


rafariossaa (approved):
Hi,
  Thanks you very much for providing this !.
  LGTM.

rafariossaa:
Could you provide more info on the scenario you are running mongo ? I am curious about the slowness.

krtk6160:
> Could you provide more info on the scenario you are running mongo ? I am curious about the slowness.

It's a 3 replicaset setup with the wiredTigerCache big enough to accommodate the entire db. I do not remember the exact lines in the logs, but it was a long operation that lasted about (~3-5 mins) and mostly occurred after upgrades.

rafariossaa:
Hi @krtk6160 ,
  Thanks for providing feedback.

Issue #5981, opened 2021-03-31
[bitnami/nginx-ingress-controller] default defaultBackend returns HTTP 200 on /


NMFR:
**Which chart**:
bitnami/nginx-ingress-controller:7.4.8

**Describe the bug**
According to https://kubernetes.github.io/ingress-nginx/user-guide/default-backend/ the defaultBackend should return:
- HTTP 200 for the `/healthz` endpoint.
- HTTP 404 for the `/` endpoint. 
- Not referenced in the docs but other endpoints should return HTTP 404.

The bitnami/nginx-ingress-controller:7.4.8 default defaultBackend returns HTTP 200 for endpoints where it should return HTTP 404.

**To Reproduce**
For example, if we install the nginx-ingress-controller:
```
helm upgrade ingress-controller bitnami/nginx-ingress-controller --install
``` 

and grab the IP of the ingress controller service: 
```
kubectl get service
```

making calls to the IP directly should call the defaultBackend:

`curl`ing `/` should give HTTP 404, but it incorrectly returns HTTP 200:
```
curl http://172.31.0.6/ -I         
HTTP/1.1 200 OK
```

`curl`ing `/index.html` should give HTTP 404, but it incorrectly returns HTTP 200:
```
curl http://172.31.0.6/index.html -I
HTTP/1.1 200 OK
```

`curl`ing `/50x.html` should give HTTP 404, but it incorrectly returns HTTP 200:
```
curl http://172.31.0.6/50x.html -I  
HTTP/1.1 200 OK
```

**Expected behavior**
The default defaultBackend should follow the behaviour described in https://kubernetes.github.io/ingress-nginx/user-guide/default-backend/ .

**Version of Helm and Kubernetes**:

- Output of `helm version`:

```
helm version
version.BuildInfo{Version:"v3.4.0", GitCommit:"7090a89efc8a18f3d8178bf47d2462450349a004", GitTreeState:"dirty", GoVersion:"go1.15.3"}
```

- Output of `kubectl version`:

```
kubectl version
Client Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.2", GitCommit:"faecb196815e248d3ecfb03c680a4507229c2a56", GitTreeState:"clean", BuildDate:"2021-01-14T05:14:17Z", GoVersion:"go1.15.6", Compiler:"gc", Platform:"darwin/amd64"}
[...]

PR #6012, opened 2021-04-05
[bitnami/nginx-ingress-controller] fix default backend status code

NMFR:
**Description of the change**

This PR addresses the issue discussed in https://github.com/bitnami/charts/issues/5981. 
According to https://kubernetes.github.io/ingress-nginx/user-guide/default-backend/ the defaultBackend should return:

- HTTP 200 for the /healthz endpoint.
- HTTP 404 for the / endpoint.
- Not referenced in the docs but other endpoints should return HTTP 404.

The current default defaultBackend returns HTTP 200 for the `/`, `/index.html` and `/50x.html` endpoints. This is incorrect, the endpoints should be returning HTTP 404. 

This PR fixes the issue by adding a new `defaultBackend.serverBlockConfig` helm value with the default value of:

```
location /healthz {
  return 200;
}

location / {
  return 404;
}
```

This value will end up in a `ConfigMap` that will be mounted in the defaultBackend @ `/opt/bitnami/nginx/conf/bitnami/defaultBackend.conf`. This works because the `bitnami/nginx` image's main nginx configuration does a `include "/opt/bitnami/nginx/conf/bitnami/*.conf";` inside the default server block.

**Benefits**

Fixes https://github.com/bitnami/charts/issues/5981. 

**Possible drawbacks**

As discussed in https://github.com/bitnami/charts/issues/5981#issuecomment-811835673 we kind of assume that everyone will be using a `bitnami/nginx` image as a defaultBackend. This kind of breaks the use case of the charts user changing the `defaultBackend.image`.

**Applicable issues**

#5981

**Additional information**

None.

**Checklist** 


bitnami/nginx-ingress-controller/values.yaml
```diff
@@ -558,6 +558,18 @@ defaultBackend:
...
+  ##
+  serverBlockConfig: |-
+    location /healthz {
+      return 200;
+    }
+
+    location / {
+      return 404;
+    }
+
```
alemorcuq, line 572: I think I would leave this object empty in this file and add this default configuration to the configmap instead. Then, if this object is set, we can replace the default configuration with the one supplied by the user.

bitnami/nginx-ingress-controller/templates/default-backend-configmap.yaml
```diff
@@ -0,0 +1,17 @@
...
+  labels: {{- include "common.labels.standard" . | nindent 4 }}
+    app.kubernetes.io/component: default-backend
+    {{- if .Values.commonLabels }}
+    {{- include "common.tplvalues.render" (dict "value" .Values.commonLabels "context" $) | nindent 4 }}
+    {{- end }}
+  {{- if .Values.commonAnnotations }}
+  annotations: {{- include "common.tplvalues.render" ( dict "value" .Values.commonAnnotations "context" $ ) | nindent 4 }}
+  {{- end }}
+data:
+  defaultBackend.conf: {{ toYaml .Values.defaultBackend.serverBlockConfig | indent 2 }}
```
alemorcuq, line 16: Check my comment on the `values.yaml` file. It's also better if we use the `"common.tplvalues.render"` function instead of the `toYaml` function to add support for templating.

alemorcuq (changes requested):
Thanks for the PR, @NMFR. Please, see my comments.

bitnami/nginx-ingress-controller/values.yaml
NMFR: I think it makes more sense for this to be in `values.yaml`:
- If a chart's user want's to know what is the default value of this he will go looking into `values.yaml` first. Moving it elsewhere will decrease the charts "readability". 
- Currently an empty value on `defaultBackend.serverBlockConfig` means that the `ConfigMap` is not created and no mount occurs in the default backend. This might be used by users that are using a custom default backend and do not wish to have an unused mount on their custom image. The proposed change will modify this, the user will no longer be able to control if the mount occurs or not.

If you disagree i will apply your change as you have more context and knowledge on this.

alemorcuq (approved):
Thanks!

PR #6015, opened 2021-04-05
[bitnami/nats] Add tcp- to the beginning of each services port name to make them compatible with Istio

### LINKED ISSUE #5999: [bitnami/nats] client-svc not compatible with Istio
Current chart sets the service port name to `client`, this does not allow istio to configure properly to the port.
https://github.com/bitnami/charts/blob/master/bitnami/nats/templates/client-svc.yaml
Changing the port name to `tcp` fixes this and allows Istio to route traffic to the service over MTLS properly.
From:
```
spec:
  clusterIP: 10.43.72.78
  clusterIPs:
  - 10.43.72.78
  ports:
  - name: client
    port: 4222
    protocol: TCP
    targetPort: client
```
To:
```
spec:
  clusterIP: 10.43.72.78
  clusterIPs:
  - 10.43.72.78
  ports:
  - name: tcp
    port: 4222
    protocol: TCP
    targetPort: client
```

See: https://github.com/nats-io/nats-operator/issues/88
I don't think there is any reason to have the port name match the service name.
It would probably be wise to change all of the services port names to `tcp`
i.e.
* client-svc.yaml
* cluster-svc.yaml
* headless-svc.yaml
* metrics-svc.yaml
* monitoring-svc.yaml

Versions:
Istio 1.8.300
nats: bitnami 6.2.3 (latest as of the time of this writing)

dsargent3220:
**Description of the change**
Alter the port names of each service to make them work with Istio
Add tcp- in front of each port name

**Benefits**
Make automatic routing with Istio and its sidecar for mutual TLS work out of the box .

**Possible drawbacks**
None that I know of.

**Applicable issues**
  - fixes #5999 

**Checklist** 
- [X ] Title of the PR starts with chart name (e.g. `[bitnami/chart]`)

### DIFF
--- a/bitnami/nats/templates/metrics-svc.yaml
+++ b/bitnami/nats/templates/metrics-svc.yaml
@@ -24,7 +24,7 @@ spec:
   loadBalancerIP: {{ .Values.metrics.service.loadBalancerIP }}
   {{- end }}
   ports:
-    - name: metrics
+    - name: tcp-metrics
       port: {{ .Values.metrics.service.port }}
       targetPort: metrics
   selector: {{ include "common.labels.matchLabels" . | nindent 4 }}
--- a/bitnami/nats/templates/monitoring-svc.yaml
+++ b/bitnami/nats/templates/monitoring-svc.yaml
@@ -23,8 +23,8 @@ spec:
   {{- end }}
   ports:
     - port: {{ .Values.monitoring.service.port }}
-      targetPort: tcp-monitoring
-      name: monitoring
+      targetPort: monitoring
+      name: tcp-monitoring
       {{- if and (eq .Values.monitoring.service.type "NodePort") (not (empty .Values.monitoring.service.nodePort)) }}
       nodePort: {{ .Values.monitoring.service.nodePort  }}
       {{- end }}
--- a/bitnami/nats/templates/monitoring-svc.yaml
+++ b/bitnami/nats/templates/monitoring-svc.yaml
@@ -23,7 +23,7 @@ spec:
   {{- end }}
   ports:
     - port: {{ .Values.monitoring.service.port }}
-      targetPort: monitoring
+      targetPort: tcp-monitoring
       name: monitoring
       {{- if and (eq .Values.monitoring.service.type "NodePort") (not (empty .Values.monitoring.service.nodePort)) }}
       nodePort: {{ .Values.monitoring.service.nodePort  }}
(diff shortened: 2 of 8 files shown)


bitnami/nats/templates/monitoring-svc.yaml
```diff
@@ -23,7 +23,7 @@ spec:
   {{- end }}
   ports:
     - port: {{ .Values.monitoring.service.port }}
-      targetPort: monitoring
+      targetPort: tcp-monitoring
```
juan131, line 26: This is used in the `ingress.yaml`. You'll have to adapt it there too:
```diff
            pathType: {{ default "ImplementationSpecific" .pathType }}
            {{- end }}
-           backend: {{- include "common.ingress.backend" (dict "serviceName" (include "common.names.fullname" $) "servicePort" "monitoring" "context" $) | nindent 14 }}
+           backend: {{- include "common.ingress.backend" (dict "serviceName" (include "common.names.fullname" $) "servicePort" "tcp-monitoring" "context" $) | nindent 14 }}
    {{- end }}
  {{- if or .Values.ingress.tls .Values.ingress.extraTls }}
  ```

File: bitnami/nats/templates/metrics-svc.yaml [hunk 2]
```diff
@@ -24,7 +24,7 @@ spec:
   loadBalancerIP: {{ .Values.metrics.service.loadBalancerIP }}
   {{- end }}
   ports:
-    - name: metrics
+    - name: tcp-metrics
```
juan131, line 27: This is used in the `servicemonitor.yaml`. You'll have to adapt it there too:
```diff
spec:
  endpoints:
-   - port: metrics
+   - port: tcp-metrics
      {{- if .Values.metrics.serviceMonitor.interval }}
      interval: {{ .Values.metrics.serviceMonitor.interval }}
      {{- end }}
```

juan131 (changes requested):
Hi @dsargent3220 

Thanks for submitting this great PR!

First of all, our charts are not designed to work with Istio. That said, they can be integrated with Istio since they can be extended adding custom sidecar containers, labels, annotations, etc.

I want to be clear on this because we don't test the integration with Istio in our pipeline and, therefore, we don't officially support it. As a consequence, we cannot guarantee that we will not break the compatibility with Istio in the future since we're not taking into account during the development process any possible consequence affecting this integration.

---

That said, in the past, I did a similar contribution for Elasticsearch, see:

- https://github.com/bitnami/charts/pull/2286

And we also received contributions from users, see:

- https://github.com/bitnami/charts/pull/5679

So, no cons from my side on accepting your contribution. Please take a look to my comments.

juan131 (approved):
Thanks for addressing the suggestions! LGTM

Issue #5975, opened 2021-03-31
Could not access Kafka brokers from outside the cluster using NodePort option


zbeedatm:
**Which chart**:
bitnami/kafka
https://github.com/bitnami/charts/tree/master/bitnami/kafka/#installing-the-chart

**Describe the bug**
Could not accessing Kafka brokers from outside the cluster using NodePort option

**To Reproduce**

Note: using minikube on Windows

We followed the instruction in section "Using NodePort service" in the link: https://github.com/bitnami/charts/tree/master/bitnami/kafka/#installing-the-chart

$ helm install kafka bitnami/kafka --set externalAccess.enabled=true --set externalAccess.service.type=NodePort --set externalAccess.autoDiscovery.enabled=true --set serviceAccount.create=true --set rbac.create=true

$ kubectl get pods
NAME                READY   STATUS    RESTARTS   AGE
kafka-0             1/1     Running   0          5m5s
kafka-zookeeper-0   1/1     Running   0          5m5s

$ kubectl get svc
NAME                       TYPE        CLUSTER-IP      EXTERNAL-IP   PORT(S)                      AGE
kafka                      ClusterIP   10.103.19.3     <none>        9092/TCP                     5m29s
kafka-0-external           NodePort    10.111.165.26   <none>        9094:30860/TCP               5m29s
kafka-headless             ClusterIP   None            <none>        9092/TCP,9093/TCP            5m29s
kafka-zookeeper            ClusterIP   10.97.173.24    <none>        2181/TCP,2888/TCP,3888/TCP   5m29s
kafka-zookeeper-headless   ClusterIP   None            <none>        2181/TCP,2888/TCP,3888/TCP   5m29s
kubernetes                 ClusterIP   10.96.0.1       <none>        443/TCP                      6m8s

Port forwarding (for NodePort - https://minikube.sigs.k8s.io/docs/start/):
$  kubectl port-forward service/kafka-0-external 9094:9094
Forwarding from 127.0.0.1:9094 -> 9094
Forwarding from [::1]:9094 -> 9094

Trying to create topic with python:

`
   from kafka.admin import KafkaAdminClient, NewTopic
   admin_client = KafkaAdminClient(
       bootstrap_servers="127.0.0.1:9094",
       client_id='0',
   )
   topic_list = []
   topic_list.append(NewTopic(name="example_topic1", num_partitions=1, replication_factor=1))
   admin_client.create_to
`

and getting:

> Traceback (most recent call last):
........
[...]

Issue #4240, opened 2020-11-06
Kafka helm chart - unable to access externally


ahmed-adly-khalil:
Hello everyone,
Hope everything is great.

After installing the chart following the doc here: https://github.com/bitnami/charts/tree/master/bitnami/kafka/#installing-the-chart

I have enabled external access using the following config:

```yaml
kafka:
  replicaCount: 1
  externalAccess:
    enabled: true
  service:
    type: NodePort
  autoDiscovery:
    enabled: true
  serviceAccount:
    create: true
  rbac:
    create: true 
```

after installing the chat using the above values, the kafka pod disappeared (terminated) and when trying to connect to the chart from outside, I'm getting the following error:

 ```
WARN [AdminClient clientId=adminclient-1] Connection to node -1 (/192.168.64.22:31020) could not be established. Broker may not be available. (org.apache.kafka.clients.NetworkClient) 
```

Appreciate your support. Thank you

**Comment by carrodher** [comment]
It seems you're using the configuration from [this example](https://github.com/bitnami/charts/tree/master/bitnami/kafka/#using-nodeport-services), right?

> after installing the chat using the above values, the kafka pod disappeared (terminated) and when trying to connect to the chart from outside, I'm getting the following error

If the pod is not running it's normal to see an error reporting that the connection can not be done successfully, we should check what is happening and why the pod is not running.

I tried installing the chart and perform some tests, can you reproduce the same steps? See below:

`helm install kafka bitnami/kafka -f values.yaml` using the following _values.yaml_:
```yaml
 externalAccess:
   enabled: true
 externalAccess:
   service:
     type: NodePort
 externalAccess:
   autoDiscovery:
     enabled: true
 serviceAccount:
   create: true
 rbac:
   create: true
```

Check that the release was created and the pods are up and running:
```console
$ helm ls
NAME                	NAMESPACE	REVISION	UPDATED                                  	STATUS  	CHART              	APP VERSION
kafka               	default  	1       	2020-11-09 13:11:39.75289376 +0000 UTC   	deployed	kafka-11.8.9       	2.6.0

$ kubectl get pods
NAME                                    READY   STATUS             RESTARTS   AGE
kafka-0                                 1/1     Running            0          8m41s
kafka-zookeeper-0                       1/1     Running            0          8m41s
```

Then we can check the installation notes in order to obtain information about how we can access/connect the application:
```console
$ helm get notes kafka
NOTES:
** Please be patient while the chart is being deployed **

[...]

PR #6035, opened 2021-04-06
Upd kafka auto discovery for ext access

tampler:
**Description of the change**
Updated documentation related to Kafka Chart. External access fails currently, which was discussed in  #5975 and #4240

**Benefits**
Access Kafka broker from outside the Kubernetes cluster

**Possible drawbacks**

None

**Applicable issues**
#5975 #4240

**Additional information**
Validated in 

Kube version: 1.20.5
Helm version: 3.5.3

### DIFF
--- a/bitnami/kafka/README.md
+++ b/bitnami/kafka/README.md
@@ -452,6 +452,7 @@ Note: This option requires creating RBAC rules on clusters where RBAC policies a
 externalAccess.enabled=true
 externalAccess.service.type=LoadBalancer
 externalAccess.service.port=9094
+externalAccess.autoDiscovery.enabled=true
 externalAccess.service.loadBalancerIPs[0]='external-ip-1'
 externalAccess.service.loadBalancerIPs[1]='external-ip-2'}
 ```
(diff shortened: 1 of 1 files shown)


bitnami/kafka/README.md
```diff
@@ -452,6 +452,7 @@ Note: This option requires creating RBAC rules on clusters where RBAC policies a
 externalAccess.enabled=true
 externalAccess.service.type=LoadBalancer
 externalAccess.service.port=9094
+externalAccess.autoDiscovery.enabled=true
```
juan131, line 455: Auto-discovery should be only enabled for "Option A". This option implies that the user is manually providing the list of external IPs.

bitnami/kafka/README.md
```diff
@@ -479,6 +480,7 @@ Note: This option requires creating RBAC rules on clusters where RBAC policies a
 ```console
 externalAccess.enabled=true
 externalAccess.service.type=NodePort
+externalAccess.autoDiscovery.enabled=true
```
juan131, line 483: Auto-discovery should be only enabled for "Option A". This option implies that the user is manually providing the list of node ports.

juan131 (changes requested):
Thanks

bitnami/kafka/README.md
tampler: How do I do that ?

PR #4916, opened 2021-01-07
[bitnami/redis] Add Redis Sentinel Exporter

ricoberger:
**Description of the change**

Add support to get Prometheus metrics via the Redis Sentinel Exporter. For that the sentinel configuration contains a new section metrics, which adds an additional sidecar with the Redis Sentinel Exporter, when enabled.

**Benefits**

This allows a user to monitor Redis Sentinel within the Helm chart.

**Applicable issues**

  - closes #4909 

**Additional information**

Currently this uses the following image from Docker Hub: https://hub.docker.com/r/leominov/redis_sentinel_exporter. The corresponding GitHub repository can be found here: https://github.com/leominov/redis_sentinel_exporter.

As mentioned by @marcosbc in #4909 let me know when the Bitnami image is ready, so that I can adjust the values file.

**Checklist** 
- [ x] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [ x] Variables are documented in the README.md
- [ x] Title of the PR starts with chart name (e.g. `[bitnami/chart]`)
- [ ] If the chart contains a `values-production.yaml` apart from `values.yaml`, ensure that you implement the changes in both files

PR #6050, opened 2021-04-07
[bitnami/redis] Fix for NetworkPolicy blocking sentinel metrics exporter

ekampf:
**Description of the change**

When sentinel metrics exporter is enabled and we also use a network policy, we need to allow scraping that port on the network policy.
This fixes the new capability added on PR #4916 

**Benefits**

Allows using sentinel metrics exporter option together with network policy

**Checklist** 


ekampf:
FYI @miguelaeh @migruiz4 @juan131 as you authored\reviewd the original PR adding the Sentinel Reporter...

migruiz4 (approved):
Hi @ekampf,

Thank you for reporting and fixing this!

migruiz4 (changes requested):
Sorry, one small change request, could you please use `{{ .Values.sentinel.metrics.service.port }}` instead?

ekampf:
@migruiz4 done!

migruiz4 (approved):
Awesome!
Thank you for applying the requested change!

PR #6057, opened 2021-04-08
[bitnami/external-dns] Allow the use of a predefined secret by specifying a secretName for the rfc2136 provider

nebulon42:
**Description of the change**
In the external-dns chart it allows to specify a secretName for the rfc2136 DNS provider to store the TSIG similarly to what other providers support already. When a secretName is specified no extra secret is generated but the existing one is used.

**Benefits**
Allows to e.g. use a SealedSecret to specify the TSIG for the rfc2136 provider and with that enables one to check the TSIG into version control without compromising its integrity.

**Checklist** 


javsalgar (approved):
Thanks for the PR! LGTM

PR #6069, opened 2021-04-11
[bitnami/kafka] Fix wrong type for listeners and advertisedListeners vars

Jasstkn:
Signed-off-by: Maria Kotlyarevskaya <jasssstkn@yahoo.com>

**Description of the change**

If I try to specify those variables as lists, I get the following error:
```
20:49:41  STDERR:
20:49:41    Error: unable to build kubernetes objects from release manifest: error validating "": error validating data: ValidationError(StatefulSet.spec.template.spec.containers[0].env[7].value): invalid type for io.k8s.api.core.v1.EnvVar.value: got "array", expected "string"
```
Eventually it comes from following usage in the templates:
```
  - name: KAFKA_CFG_LISTENERS
    {{- if .Values.listeners }}
    value: {{ .Values.listeners }}
...
  - name: KAFKA_CFG_ADVERTISED_LISTENERS
    {{- if .Values.advertisedListeners }}
    value: {{ .Values.advertisedListeners }}
```

Fix wrong type of listeners and advertisedListeners. They are used as a string variables but in `values.yaml` and `README.md` they are declared as list

**Benefits**

Chart is deployed without errors.

**Possible drawbacks**

N/A

**Checklist** 


Jasstkn:
> Great PR! Please pay attention to the following items before merging:
> 
> **Files matching `bitnami/*/values.yaml`:**
> 
> * [ ]  Is the PR adding a new container? Please reviewer, add it to the models (internal process)
> * [ ]  Is the PR adding a new parameter? Please, ensure it’s documented in the README.md
> 
> This is an automatically generated QA checklist based on modified files

Nothing from that list.

juan131 (approved):
Thanks for this contribution!

As you mentioned the Kafka image is expecting a "comma separated" string with the list of listeners or advertised listeners. See:

- https://github.com/bitnami/bitnami-docker-kafka#accessing-kafka-with-internal-and-external-clients

juan131 (changes requested):
Thinking it twice... I think we should maintain the array format since it's more appropriate for this setting. Then, we can do sth like this in the `statefulset.yaml` to adapt the array to a comma separated string:
```yaml
            - name: KAFKA_CFG_LISTENERS
              {{- if .Values.listeners }}
              value: {{ join "," .Values.listeners }}
            ...
            - name: KAFKA_CFG_ADVERTISED_LISTENERS
              {{- if .Values.advertisedListeners }}
              value: {{ join "," .Values.advertisedListeners }}
```

Jasstkn:
> Thinking it twice... I think we should maintain the array format since it's more appropriate for this setting. Then, we can do sth like this in the `statefulset.yaml` to adapt the array to a comma separated string:
> 
> ```yaml
>             - name: KAFKA_CFG_LISTENERS
>               {{- if .Values.listeners }}
>               value: {{ join "," .Values.listeners }}
>             ...
>             - name: KAFKA_CFG_ADVERTISED_LISTENERS
>               {{- if .Values.advertisedListeners }}
>               value: {{ join "," .Values.advertisedListeners }}
> ```

Are you sure? Default value isn't formatted this way. I think it could be even more confusing.

juan131:
> Are you sure?

IMHO, we should use arrays for this kind of settings. This is the native YAML way to deal with lists.

> Default value isn't formatted this way. I think it could be even more confusing.

What do you mean with "default value"?

Jasstkn:
> > Are you sure?
> 
> IMHO, we should use arrays for this kind of settings. This is the native YAML way to deal with lists.
> 
> > Default value isn't formatted this way. I think it could be even more confusing.
> 
> What do you mean with "default value"?

here: https://github.com/bitnami/charts/blob/master/bitnami/kafka/templates/statefulset.yaml#L177-L184

juan131:
> here: https://github.com/bitnami/charts/blob/master/bitnami/kafka/templates/statefulset.yaml#L177-L184

That's the value for the environment variable, but that's different from the "chart value"

Jasstkn:
> > here: https://github.com/bitnami/charts/blob/master/bitnami/kafka/templates/statefulset.yaml#L177-L184
> 
> That's the value for the environment variable, but that's different from the "chart value"

Ok, I got your point. I'll make required changes then.

juan131:
Thanks @Jasstkn !!

Jasstkn:
> Thanks @Jasstkn !!

Added. Please, review

bitnami/kafka/templates/statefulset.yaml
```diff
@@ -210,7 +210,7 @@ spec:
             {{- else }}
             - name: KAFKA_CFG_ADVERTISED_LISTENERS
               {{- if .Values.advertisedListeners }}
-              value: {{ .Values.advertisedListeners }}
+              value: {{ include "kafka.advertisedListeners" . }}
```
juan131, line 213: You'll also have to adapt the line below in the **scripts-configmap.yaml**:

- https://github.com/bitnami/charts/blob/master/bitnami/kafka/templates/scripts-configmap.yaml#L110

bitnami/kafka/templates/statefulset.yaml
```diff
@@ -190,7 +190,7 @@ spec:
             {{- end }}
             - name: KAFKA_CFG_LISTENERS
               {{- if .Values.listeners }}
-              value: {{ .Values.listeners }}
+              value: {{ include "kafka.listeners" . }}
```
juan131, line 193: I'd directly used "join" here and avoid adding new macros to `_helpers.tpl`:
```suggestion
              value: {{ join "," .Values.listeners }}
```

bitnami/kafka/templates/statefulset.yaml
juan131: Same concern about avoiding new macros.

juan131 (changes requested):
Thanks for the applying the suggestions! Please check my comments!
Also, you'll have to rebase your branch from master since other contributions also bumped the chart version.

juan131 (approved):
LGTM

Jasstkn:
> Thanks for the applying the suggestions! Please check my comments!
> Also, you'll have to rebase your branch from master since other contributions also bumped the chart version.

Thank you for your review. I've dropped templates and move join to the statefulset and configmap. If you have any other feedback, I'll be glad to get it.

PR #6072, opened 2021-04-12
[bitnami/*] Fix values.yaml for internal automations

miguelaeh:
**Description of the change**
Standardise `values.yaml` that uses readmenator to the format supported by our internal automations

**Benefits**

**Possible drawbacks**

**Applicable issues**

  - fixes #

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/wordpress/values.yaml
+++ b/bitnami/wordpress/values.yaml
@@ -14,9 +14,6 @@ global:
   ##   - myRegistryKeySecretName
   ##
   imagePullSecrets: []
-  ## E.g.
-  ## storageClass: myStorageClass
-  ##
   storageClass:
 
 ## @section Common parameters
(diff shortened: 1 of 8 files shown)


bitnami/wordpress/values.yaml
```diff
@@ -14,9 +14,6 @@ global:
   ##   - myRegistryKeySecretName
   ##
   imagePullSecrets: []
-  ## E.g.
-  ## storageClass: myStorageClass
-  ##
```
juan131, line 19: Why is this required?

miguelaeh: Our internal processes only match the first occurrence. Commented or not, so if we add both commented and no commented it will duplicate the line

carrodher (approved):
Lgtm

bitnami/wordpress/values.yaml
juan131: Then, this is a limitation to be solved in the internal process, not here.

By doing this, the final product (the chart in this case) is the one "harmed". I know it's harder to fix this in the internal automations but that's not an excuse since the product shouldn't be adapted to fit in the pipeline but the other way around.

PR #5528, opened 2021-02-17
[bitnami/redis] PreStop Hook to Initiate Failover for Sentinel on PodTermination

0x46616c6b:
**Description of the change**

We have encountered problems with Redis Sentinel and Rolling Upgrades. In this case, the Master Election takes several minutes, which ensures that the service does not work correctly during that time.
Therefore, we implemented a `PreStop` hook that checks if the pod is the master and initiates a failover. 

**Benefits**

- Service is useable during a Rolling Upgrade
- Rolling Upgrades are much faster

**Possible drawbacks**

- Covers only the case that the Pod receives a SIGTERM

**Applicable issues**

#5418
#5400

**Checklist**

- [X] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [X] Title of the PR starts with chart name (e.g. `[bitnami/chart]`)

PR #6080, opened 2021-04-12
[bitnami/redis] Improve sentinel prestop hook to prevent service interruption

Gregy:
This improves on #5528 by checking and waiting until the failover is finished on both the redis and the sentinel container termination. This completely eliminates a momentary service interruption during rollouts that happens with just #5528 applied.

**Benefits**

This eliminates the few seconds of downtime that happen with v13.0.1 because the redis container terminates before the sentinel prestop hook finishes running (in k8s containers are being stopped concurrently)

I have also replaced a hard-coded masterSet config value in the prestop script with proper config reference. This problem was mentioned by @srueg in this comment: https://github.com/bitnami/charts/pull/5528#discussion_r611600890

**Possible drawbacks**

This code can make pod deletes (and rollouts) slower by a few seconds.

**Applicable issues**

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/redis/README.md
+++ b/bitnami/redis/README.md
@@ -71,16 +71,16 @@ The command removes all the Kubernetes components associated with the chart and 
 
 ### Common parameters
 
-| Name                            | Description                                        | Value           |
-|---------------------------------|----------------------------------------------------|-----------------|
-| `kubeVersion`                   | Override Kubernetes version                        | `nil`           |
-| `nameOverride`                  | String to partially override common.names.fullname | `nil`           |
-| `fullnameOverride`              | String to fully override common.names.fullname     | `nil`           |
-| `commonLabels`                  | Labels to add to all deployed objects              | `{}`            |
-| `commonAnnotations`             | Annotations to add to all deployed objects         | `{}`            |
-| `clusterDomain`                 | Kubernetes cluster domain name                     | `cluster.local` |
-| `extraDeploy`                   | Array of extra objects to deploy with the release  | `[]`            |
-| `terminationGracePeriodSeconds` | Define termination grace period for all pods       | `30`            |
+| Name                | Description                                        | Value           |
+| ------------------- | -------------------------------------------------- | --------------- |
+| `kubeVersion`       | Override Kubernetes version                        | `nil`           |
+| `nameOverride`      | String to partially override common.names.fullname | `nil`           |
+| `fullnameOverride`  | String to fully override common.names.fullname     | `nil`           |
+| `commonLabels`      | Labels to add to all deployed objects              | `{}`            |
+| `commonAnnotations` | Annotations to add to all deployed objects         | `{}`            |
+| `clusterDomain`     | Kubernetes cluster domain name                     | `cluster.local` |
+| `extraDeploy`       | Array of extra objects to deploy with the release  | `[]`            |
+
 
 ### Redis(TM) Image parameters
 
@@ -180,6 +180,7 @@ The command removes all the Kubernetes components associated with the chart and 
 | `master.service.loadBalancerIP`             | Redis(TM) master service Load Balancer IP                                                        | `nil`           |
 | `master.service.loadBalancerSourceRanges`   | Redis(TM) master service Load Balancer sources                                                   | `[]`            |
 | `master.service.annotations`                | Additional custom annotations for Redis(TM) master service                                       | `{}`            |
+| `master.terminationGracePeriodSeconds`      | Integer setting the termination grace period for the redis-master pods                           | `30`            |
 
 
 ### Redis(TM) replicas configuration parameters
@@ -254,6 +255,7 @@ The command removes all the Kubernetes components associated with the chart and 
 | `replica.service.loadBalancerIP`             | Redis(TM) replicas service Load Balancer IP                                                       | `nil`           |
 | `replica.service.loadBalancerSourceRanges`   | Redis(TM) replicas service Load Balancer sources                                                  | `[]`            |
 | `replica.service.annotations`                | Additional custom annotations for Redis(TM) replicas service                                      | `{}`            |
+| `replica.terminationGracePeriodSeconds`      | Integer setting the termination grace period for the redis-replicas pods                          | `30`            |
 
 
 ### Redis(TM) Sentinel configuration parameters
@@ -310,6 +312,7 @@ The command removes all the Kubernetes components associated with the chart and 
 | `sentinel.service.loadBalancerIP`             | Redis(TM) Sentinel service Load Balancer IP                                                      | `nil`                    |
 | `sentinel.service.loadBalancerSourceRanges`   | Redis(TM) Sentinel service Load Balancer sources                                                 | `[]`                     |
 | `sentinel.service.annotations`                | Additional custom annotations for Redis(TM) Sentinel service                                     | `{}`                     |
+| `sentinel.terminationGracePeriodSeconds`      | Integer setting the termination grace period for the redis-node pods                             | `30`                     |
 
 
 ### Other Parameters
--- a/bitnami/redis/README.md
+++ b/bitnami/redis/README.md
@@ -71,16 +71,16 @@ The command removes all the Kubernetes components associated with the chart and 
 
 ### Common parameters
 
-| Name                | Description                                        | Value           |
-| ------------------- | -------------------------------------------------- | --------------- |
-| `kubeVersion`       | Override Kubernetes version                        | `nil`           |
-| `nameOverride`      | String to partially override common.names.fullname | `nil`           |
-| `fullnameOverride`  | String to fully override common.names.fullname     | `nil`           |
-| `commonLabels`      | Labels to add to all deployed objects              | `{}`            |
-| `commonAnnotations` | Annotations to add to all deployed objects         | `{}`            |
-| `clusterDomain`     | Kubernetes cluster domain name                     | `cluster.local` |
-| `extraDeploy`       | Array of extra objects to deploy with the release  | `[]`            |
-
+| Name                            | Description                                        | Value           |
+|---------------------------------|----------------------------------------------------|-----------------|
+| `kubeVersion`                   | Override Kubernetes version                        | `nil`           |
+| `nameOverride`                  | String to partially override common.names.fullname | `nil`           |
+| `fullnameOverride`              | String to fully override common.names.fullname     | `nil`           |
+| `commonLabels`                  | Labels to add to all deployed objects              | `{}`            |
+| `commonAnnotations`             | Annotations to add to all deployed objects         | `{}`            |
+| `clusterDomain`                 | Kubernetes cluster domain name                     | `cluster.local` |
+| `extraDeploy`                   | Array of extra objects to deploy with the release  | `[]`            |
+| `terminationGracePeriodSeconds` | Define termination grace period for all pods       | `30`            |
 
 ### Redis(TM) Image parameters
 
--- a/bitnami/redis/templates/scripts-configmap.yaml
+++ b/bitnami/redis/templates/scripts-configmap.yaml
@@ -297,7 +297,8 @@ data:
     }
 
     REDIS_SERVICE="{{ include "common.names.fullname" . }}.{{ .Release.Namespace }}.svc.{{ .Values.clusterDomain }}"
-    
+   
+    # redis-cli automatically consumes credentials from the REDISCLI_AUTH variable
     [[ -n "$REDIS_PASSWORD" ]] && export REDISCLI_AUTH="$REDIS_PASSWORD"
     [[ -f "$REDIS_PASSWORD_FILE" ]] && export REDISCLI_AUTH="$(< "${REDIS_PASSWORD_FILE}")"
 
@@ -332,6 +333,8 @@ data:
         REDIS_ROLE=$(run_redis_command role | head -1)
         [[ "$REDIS_ROLE" != "master" ]]
     }
+    
+    # redis-cli automatically consumes credentials from the REDISCLI_AUTH variable
     [[ -n "$REDIS_PASSWORD" ]] && export REDISCLI_AUTH="$REDIS_PASSWORD"
     [[ -f "$REDIS_PASSWORD_FILE" ]] && export REDISCLI_AUTH="$(< "${REDIS_PASSWORD_FILE}")"
 
--- a/bitnami/redis/templates/scripts-configmap.yaml
+++ b/bitnami/redis/templates/scripts-configmap.yaml
@@ -281,26 +281,67 @@ data:
     #!/bin/bash
 
     . /opt/bitnami/scripts/libvalidations.sh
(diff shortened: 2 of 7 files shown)


bitnami/redis/templates/configmap-scripts.yaml
```diff
@@ -294,28 +294,79 @@ data:
...
-      sentinel_info_command="redis-cli {{- if .Values.usePassword }} -a ${REDIS_PASSWORD} {{- end }} -h ${REDIS_SERVICE} -p {{ .Values.sentinel.port }} --tls --cert ${REDIS_SENTINEL_TLS_CERT_FILE} --key ${REDIS_SENTINEL_TLS_KEY_FILE} --cacert ${REDIS_SENTINEL_TLS_CA_FILE} sentinel get-master-addr-by-name {{ .Values.sentinel.masterSet }}"
+    if am_i_the_master; then
+      run_sentinel_command failover "{{ .Values.sentinel.masterSet }}"
+      for i in $(seq 10); do
+        if ! am_i_the_master; then
+          exit 0
+        fi
+        echo "Waiting for failover to finish"
+        sleep 1
+      done
```
miguelaeh, line 330: When the loop finishes, the failover is not necessarily complete. Also, I think the second, if will never be executed since the for, will only be executed when you are the master right?
I would use something like:
``` suggestion
    run_failover () {
        run_sentinel_command failover "{{ .Values.sentinel.masterSet }}"
    }
    if ! retry_while run_failover; then
        error "Failed to run the failover"
        exit 1
    fi
```

bitnami/redis/templates/configmap-scripts.yaml
```diff
@@ -294,28 +294,79 @@ data:
     #!/bin/bash
 
     . /opt/bitnami/scripts/libvalidations.sh
```
miguelaeh, line 296:
```suggestion
    . /opt/bitnami/scripts/libvalidations.sh
    . /opt/bitnami/scripts/libos.sh
```

This will be required for the comment about using `retry_while`

bitnami/redis/templates/configmap-scripts.yaml
```diff
@@ -294,28 +294,79 @@ data:
...
+    #!/bin/bash
 
-    if [[ "$REDIS_MASTER_HOST" == "$(hostname -i)" ]]; then
-      if is_boolean_yes "$REDIS_SENTINEL_TLS_ENABLED"; then
-        redis-cli {{- if .Values.usePassword }} -a "$REDIS_PASSWORD" {{- end }} -h "$REDIS_SERVICE" -p {{ .Values.sentinel.port }} --tls --cert "$REDIS_SENTINEL_TLS_CERT_FILE" --key "$REDIS_SENTINEL_TLS_KEY_FILE" --cacert "$REDIS_SENTINEL_TLS_CA_FILE" sentinel failover mymaster
+    . /opt/bitnami/scripts/libvalidations.sh
+    
+    run_redis_command() {
+      if is_boolean_yes "$REDIS_TLS_ENABLED"; then
+        redis-cli -h 127.0.0.1 -p ${REDIS_TLS_PORT} --tls --cert ${REDIS_TLS_CERT_FILE} --key ${REDIS_TLS_KEY_FILE} --cacert ${REDIS_TLS_CA_FILE} "$@"
```
miguelaeh, line 341:
```suggestion
        redis-cli -h 127.0.0.1 -p "$REDIS_TLS_PORT" --tls --cert "$REDIS_TLS_CERT_FILE" --key "$REDIS_TLS_KEY_FILE" --cacert "$REDIS_TLS_CA_FILE" "$@"
```

bitnami/redis/templates/configmap-scripts.yaml
```diff
@@ -294,28 +294,79 @@ data:
...
+    fi
+
+    if am_i_the_master; then
+      echo "Waiting for sentinel to run failover for up to 10s"
+      for i in $(seq 10); do
+        if ! am_i_the_master; then
+          exit 0
+        fi
+        sleep 1
+      done
```
miguelaeh, line 367: Try to use here a similar approach to the above one using `retry_while`

miguelaeh (changes requested):

bitnami/redis/templates/configmap-scripts.yaml
Gregy: You are correct that the loop can end before we failover successfuly. But there is not much we can do about that. We cannot take longer than terminationGracePeriodSeconds (by default 30s) otherwise we get sigkilled. Also if the failover takes more than 10s you likely have some other problem elsewhere in the cluster that we cannot do anything about in the prestop hook.

The second if will be executed. You are correct that we enter this section of code only if we are master but once we execute the failover command on line 323 the process to move master to other node starts. It usually finishes within a second or two and after that the am_i_the_master function will return false and we can exit the wait loop early to make the pod shutdown as fast as possible.

Gregy: Your suggestion doesn't make much sense to me. It runs the failover command multiple times but that is really undesirable. Because each time you run it you swtich the master again and again. What we want to do is switch the master once (to get it away from the terminating pod) and then wait a bit until the process finishes.

bitnami/redis/templates/configmap-scripts.yaml
Gregy: Is there a list of these prepared functions with descriptions somewhere? I can definitely clean up the loop with it but I am not sure where to find the function definitions/docs.

bitnami/redis/templates/configmap-scripts.yaml
miguelaeh: >It runs the failover command multiple times but that is really undesirable. 

I agree, I didn't notice that when writing the suggestion. Isn't there any way to check if the failover has finished? Maybe checking if the master has changed? If we find the way to do that check we can run the failover once, and use the `retry_while` with that new check command. I am not comfortable with the idea of leaving the sigkill in hands of the termination grace period since it is a configurable parameter and the `10` seconds will be a hardcoded number, IMHO it will become unconsistent.

bitnami/redis/templates/configmap-scripts.yaml
miguelaeh: You can check them here: https://github.com/bitnami/bitnami-docker-redis/tree/master/6.2/debian-10/prebuildfs/opt/bitnami/scripts

bitnami/redis/templates/configmap-scripts.yaml
Gregy: But that is precisely what I am doing on line 325 (`if ! am_i_the_master; then`). This checks if the current pod is still the master or not. If it is not the master anymore we terminate the loop and exit. I agree I could use the retry_while but I would need to know the definiton or some documentation for it so I know how exactly it behaves. Could you please point me to the function definition or documentation for the `retry_while` helper?

> I am not comfortable with the idea of leaving the sigkill in hands of the termination grace period since it is a configurable parameter and the 10 seconds will be a hardcoded number, IMHO it will become inconsistent.
This script works on the "best effort" basis. It tries to migrate the master away for 10s and then gives up and lets the pod terminate. I don't think anything more can be accomplished in the pre-stop hook. You are correct that the 10s is an arbitrary wait time but I am not sure what else to do. I definitely do not want to exceed the termination grace period and that is why I capped the wait to 10s. But if there is a way to get the current value of the termination grace period we could cap this proces to that minus 5s or something like that. That would make sure we exit quickly enough even when someone changes the grace period (but I do not think this chart exposes the grace period as a variable to be changed)

Gregy: Ah, just saw your other comment. I will check the docs and attempt to rework the loop with retry_while. Thanks!

Gregy:
I have reworked the waiting logic to use `retry_while` instead of loops like you suggested @miguelaeh . Can you please check if the new approach is acceptable?

miguelaeh (commented):
Thank you very much for implementing the changes!

> But if there is a way to get the current value of the termination grace period we could cap this proces to that minus 5s or something like that. That would make sure we exit quickly enough even when someone changes the grace period (but I do not think this chart exposes the grace period as a variable to be changed)

About this, we could just add it. It is just needed to add it to the values.yaml, README.md and to the statefulset. Then you will be able to use it just like `{ .Values.terminationGracePeriodSeconds }`.
After that a minor the version must be bumped in a minor instead of a patch.

Gregy:
Thanks for the tip. Implemented.

Gregy:
Another PR (https://github.com/bitnami/charts/pull/6146) containing the hard-coded masterset value fix has been merged. I will remove the first commit from this PR to resolve the conflicts. The rest of this PR should still get merged after successful review.

juan131 (changes requested):
We're doing a major refactoring of the chart at https://github.com/bitnami/charts/pull/6102
I will block this PR until the refactoring is completed.

Gregy:
Juan, could you explain why a work on the next major version should prevent a bugfix PR like this from being merged? This PR doesn't bring any backward compatibility breaks and would be useful to people using v13 of this chart.

juan131:
Hi @Gregy 

The major refactoring PR was merged, could you please rebase your branch from master and update the PR? Then, please pin me and I'll be glad to review it.

Thanks in advance.

Gregy:
Wow that was quick. Thanks for pinging me @juan131

Rebase done. Ready for review.

bitnami/redis/templates/configmap-scripts.yaml
Gregy: I believe this problem is resolved in the latest version of the PR

juan131 (commented):
Sorry, I won't have time to check this properly today
I'll check it tomorrow morning

bitnami/redis/README.md
```diff
@@ -71,16 +71,16 @@ The command removes all the Kubernetes components associated with the chart and
...
+| Name                            | Description                                        | Value           |
+|---------------------------------|----------------------------------------------------|-----------------|
+| `kubeVersion`                   | Override Kubernetes version                        | `nil`           |
+| `nameOverride`                  | String to partially override common.names.fullname | `nil`           |
+| `fullnameOverride`              | String to fully override common.names.fullname     | `nil`           |
+| `commonLabels`                  | Labels to add to all deployed objects              | `{}`            |
+| `commonAnnotations`             | Annotations to add to all deployed objects         | `{}`            |
+| `clusterDomain`                 | Kubernetes cluster domain name                     | `cluster.local` |
+| `extraDeploy`                   | Array of extra objects to deploy with the release  | `[]`            |
+| `terminationGracePeriodSeconds` | Define termination grace period for all pods       | `30`            |
```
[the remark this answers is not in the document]
Gregy: I don't think so. The termination grace period is configured for the whole pod. Not the individual containers. We could theoreticaly make the setting seperate for sentinels, masters and replicas but if sentinels are used no masters and replicas are present and it doesn't make much sense to me to have a different grace period for master vs replica. So I am for keeping the setting global. But I can separate them if you wish.

No, I don't even have access to that repo. I have just modified the file with some help from a markdown table generator.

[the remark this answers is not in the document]
juan131: >  It doesn't make much sense to me to have a different grace period for master vs replica. So I am for keeping the setting global. But I can separate them if you wish.

I agree it's very unlikely you want to set a value for this setting on master nodes different from the one you set on replicas nodes. That said, it seems this is the approach we followed on other charts, and I'd follow it for consistency. See:

- https://github.com/bitnami/charts/blob/master/bitnami/metallb/templates/controller/deployment.yaml#L34
- https://github.com/bitnami/charts/blob/master/bitnami/metallb/templates/speaker/daemonset.yaml#L3

> No, I don't even have access to that repo. I have just modified the file with some help from a markdown table generator.

Crap, i though this was already public. Sorry for the noise.

bitnami/redis/templates/scripts-configmap.yaml
```diff
@@ -281,26 +281,67 @@ data:
...
-            redis-cli {{- if .Values.auth.enabled }} -a "$REDIS_PASSWORD" {{- end }} -h "$REDIS_SERVICE" -p {{ .Values.sentinel.service.sentinelPort }} --tls --cert "$REDIS_SENTINEL_TLS_CERT_FILE" --key "$REDIS_SENTINEL_TLS_KEY_FILE" --cacert "$REDIS_SENTINEL_TLS_CA_FILE" sentinel failover {{ .Values.sentinel.masterSet }}
+    . /opt/bitnami/scripts/libvalidations.sh
+    . /opt/bitnami/scripts/libos.sh
+
+    run_redis_command() {
+        if is_boolean_yes "$REDIS_TLS_ENABLED"; then
+            redis-cli -h 127.0.0.1 -p "$REDIS_TLS_PORT" --tls --cert "$REDIS_TLS_CERT_FILE" --key "$REDIS_TLS_KEY_FILE" --cacert "$REDIS_TLS_CA_FILE" "$@"
         else
-            redis-cli {{- if .Values.auth.enabled }} -a "$REDIS_PASSWORD" {{- end }} -h "$REDIS_SERVICE" -p {{ .Values.sentinel.service.sentinelPort }} sentinel failover {{ .Values.sentinel.masterSet }}
+            redis-cli -h 127.0.0.1 -p ${REDIS_PORT} "$@"
```
juan131, line 328: Why the authentication flags (`-a "$REDIS_PASSWORD"`) are not required here when enabling authentication?

Gregy: We are setting the REDISCLI_AUTH variable instead. redis-cli will use it's contents to authenticate automatically if required. No need for additional switches. 

https://github.com/bitnami/charts/pull/6080/files#diff-8bf2ebfa2494dc7e62169b760334cb7ae0812a59a6a0d11e436b7d040e7e30caR335

bitnami/redis/README.md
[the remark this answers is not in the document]
Gregy: Ok, I have separated the config files.

bitnami/redis/templates/scripts-configmap.yaml
juan131: fair enough

bitnami/redis/templates/sentinel/statefulset.yaml
```diff
@@ -79,6 +79,7 @@ spec:
       {{- if .Values.replica.schedulerName }}
       schedulerName: {{ .Values.replica.schedulerName | quote }}
       {{- end }}
+      terminationGracePeriodSeconds: {{ .Values.terminationGracePeriodSeconds }}
```
juan131, line 82:
```suggestion
      terminationGracePeriodSeconds: {{ .Values.sentinel.terminationGracePeriodSeconds }}
```

Gregy:
Ouch, thanks for catching that!

miguelaeh (approved):
Thank you for all the changes @Gregy ,
LGTM!
We have to wait the internal CI to update the images tags

Gregy:
Awesome. Thank you very much Miguelaeh.

PR #6100, opened 2021-04-14
[bitnami/airflow] Fix tolerations for Airflow Exporter

sgrzemski-ias:
**Description of the change**
Make Airflow Exporter respect the tolerations settings. Currently it is declared in values but it was not implemented in the deployment template for airflow exporter. They are also present in the docs, but not used -.-

**Benefits**
Airflow Exporter respect the tolerations settings and end users can use it in their setups. 

**Possible drawbacks**
None.

**Applicable issues**

  - fixes #

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/airflow/templates/metrics/deployment.yaml
+++ b/bitnami/airflow/templates/metrics/deployment.yaml
@@ -39,6 +39,13 @@ spec:
       nodeSelector: {{- include "common.tplvalues.render" (dict "value" .Values.metrics.nodeSelector "context" $) | nindent 8 }}
       {{- end }}
       {{- end }}
+      {{- if or .Values.tolerations .Values.metrics.tolerations }}
+      {{- if .Values.tolerations }}
+      tolerations: {{- include "common.tplvalues.render" (dict "value" .Values.tolerations "context" $) | nindent 8 }}
+      {{- else }}
+      tolerations: {{- include "common.tplvalues.render" (dict "value" .Values.metrics.tolerations "context" $) | nindent 8 }}
+      {{- end }}
+      {{- end }}
       containers:
         - image: {{ include "airflow.metrics.image" . | quote }}
           name: airflow-exporter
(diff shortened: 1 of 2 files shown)


bitnami/airflow/templates/metrics/deployment.yaml
```diff
@@ -39,6 +39,13 @@ spec:
       nodeSelector: {{- include "common.tplvalues.render" (dict "value" .Values.metrics.nodeSelector "context" $) | nindent 8 }}
       {{- end }}
       {{- end }}
+      {{- if or .Values.tolerations .Values.metrics.tolerations }}
+      {{- if .Values.tolerations }}
+      tolerations: {{- include "common.tplvalues.render" (dict "value" .Values.tolerations "context" $) | nindent 8 }}
+      {{- else }}
+      tolerations: {{- include "common.tplvalues.render" (dict "value" .Values.metrics.tolerations "context" $) | nindent 8 }}
+      {{- end }}
+      {{- end }}
```
alemorcuq, line 48: According to the `values.yaml`, the `metrics.tolerations` parameter takes precedence over the global tolerations parameter.

> Note: this configuration supersedes the global tolerations configuration

So this logic should be the other way around: render `metrics.tolerations` if they are set, default to `tolerations` instead.

alemorcuq (changes requested):
Thanks for the PR! See my comment below.

bitnami/airflow/templates/metrics/deployment.yaml
sgrzemski-ias: Yes, I totally agree with that, but I used the same strategy as is used with nodeselector. Let me fix that then.

sgrzemski-ias: Done. `metrics.tolerations` would be set if defined, if not `tolerations` would be set if defined.

alemorcuq (approved):
Thanks!

PR #6119, opened 2021-04-15
[bitnami/nginx-ingress-controller] - Fixed ignoring "enabled" parameter in .Values.podSecurityContext

adam-prohack:
## Description of the change
I am trying to deploy bitnami/nginx-ingress-controller on openshift platform and I saw that .Values.podSecurityContext.enabled parameter is omitted. I need disable podSecurityContext in order to prevent openshift problems.

## Benefits
You can run helm chart on OpenShift


marcosbc (approved):
LGTM, thanks for the fix!

PR #6122, opened 2021-04-15
[bitnami/thanos] fixes thanos/receive HPA

volgorean:
**Description of the change**
Fixes the thanos/receive HPA which currently looks for a `Deployment` instead of a `StatefulSet` set resulting in only one pod when HPA is enabled.

**Benefits**
Thanos Recieve can now be scaled with an HPA properly

**Possible drawbacks**
more pods = more money 💵 

**Applicable issues**
No existing issue open, the component is fairly new so I may be the first to notice 🤓 

**Checklist** 


miguelaeh (approved):
LGTM

PR #6137, opened 2021-04-17
[bitnami/redis] Add missiing subPath to redis slave

YaoC:
**Description of the change**

The subPath exists in volume-permissions(initContainers) while missing in main container.

**Benefits**

**Possible drawbacks**

**Applicable issues**

  - fixes #

**Additional information**

**Checklist** 


bitnami/redis/templates/redis-slave-statefulset.yaml
```diff
@@ -200,6 +200,7 @@ spec:
             {{- end }}
             - name: redis-data
               mountPath: /data
+              subPath: {{ .Values.slave.persistence.subPath }}
```
marcosbc, line 203: Since `subPath` is empty by default, we should avoid setting if empty, i.e.:

```suggestion
              {{- if .Values.slave.persistence.subPath }}
              subPath: {{ .Values.slave.persistence.subPath }}
              {{- end }}
```

Could you make that change?

marcosbc (changes requested):
Thanks for the contribution! Added a small request, could you look into it?

bitnami/redis/templates/redis-slave-statefulset.yaml
YaoC: Since the default value of subPath in [values.yaml](https://github.com/bitnami/charts/blob/master/bitnami/redis/values.yaml#L744) is `""` and the default value of subPath in [kubernetes](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#volumes-1) is `""` too, maybe there is no need to check again :)

marcosbc: I see your point. I was mostly thinking about a user doing `helm template`. However, checking other existing templates, it seems we are doing it this exact same way, so LGTM.

marcosbc (approved):
Thanks!

marcosbc (changes requested):
Actually, I just noticed that there are conflicts between master due to the new 14.0.0 major. Could you resolve them and update the PR?

YaoC:
> Actually, I just noticed that there are conflicts between master due to the new 14.0.0 major. Could you resolve them and update the PR?

Hi, the conflicts are solved.

marcosbc (approved):
Thanks for fixing!

PR #6167, opened 2021-04-21
[bitnami/nats] Fix monitoring ingress endpoint

benjamin658:
**Description of the change**

Fixed the monitoring ingress endpoint.

**Benefits**

The monitoring page will work properly with nginx ingress.

**Possible drawbacks**

N/A

**Applicable issues**

  - fixes #6166

**Additional information**

N/A

**Checklist** 

- [X] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [ ] Variables are documented in the README.md
- [X] Title of the PR starts with chart name (e.g. `[bitnami/chart]`)

:warning: Keep in mind that if you want to make changes to the kubeapps chart, please implement them in the [kubeapps repository](https://github.com/kubeapps/kubeapps/tree/master/chart/kubeapps). This is only a synchronized mirror.

PR #6177, opened 2021-04-21
[bitnami/nats] Adds auth.timeout value to manage client authentication timeout

alexerm:
**Description of the change**

This change is a minor update of parameters for NATS chart. It is introducing a new parameter `auth.timeout`. This parameter is used in ConfigMap to control how long authentication of the nats client may hang before timeout. 

**Benefits**

Current ConfigMap contains the hardcoded value of 1 second.  With this PR I propose to make it a parameter but keep `1` second as a default value.

Hardcoded value in the current ConfigMap:
```yaml
    authorization {
      {{- if .Values.auth.user }}
      user: {{ .Values.auth.user | quote }}
      password: {{ $authPwd | quote }}
      {{- else if .Values.auth.token }}
      token: {{ .Values.auth.token | quote }}
      {{- end }}
      timeout:  1 
    }
```

After the change:
```yaml
timeout: {{ int .Values.auth.timeout }}
```

New parameter:
```yaml
## Client Authentication
## ref: https://github.com/nats-io/gnatsd#authentication
##
auth:
  enabled: true
  user: nats_client
  # password:
  # token:
  timeout: 1
```

**Possible drawbacks**

I see no downsides to this change.

**Applicable issues**

I haven't found any issues related to this change.

**Checklist** 

### DIFF
--- a/bitnami/nats/Chart.yaml
+++ b/bitnami/nats/Chart.yaml
@@ -24,4 +24,4 @@ name: nats
 sources:
   - https://github.com/bitnami/bitnami-docker-nats
   - https://nats.io/
-version: 6.2.6
+version: 6.3.0
--- a/bitnami/nats/README.md
+++ b/bitnami/nats/README.md
@@ -79,6 +79,7 @@ The following tables lists the configurable parameters of the NATS chart and the
 | `auth.user`                | Client authentication user                                                     | `nats_client`                                           |
 | `auth.password`            | Client authentication password                                                 | `random alhpanumeric string (10)`                       |
 | `auth.token`               | Client authentication token                                                    | `nil`                                                   |
+| `auth.timeout`             | Client authentication timeout (seconds)                                        | `1`                                                     |
 | `clusterAuth.enabled`      | Switch to enable/disable cluster authentication                                | `true`                                                  |
 | `clusterAuth.user`         | Cluster authentication user                                                    | `nats_cluster`                                          |
 | `clusterAuth.password`     | Cluster authentication password                                                | `random alhpanumeric string (10)`                       |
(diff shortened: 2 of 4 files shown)


javsalgar (commented):
Thanks for the PR! It looks great! As it is adding a feature, could you bump the minor version?

alexerm:
Thanks for the feedback! I was a bit hesitant about bumping the minor version first, but it makes total sense.

Bumped version from 6.2.6 to 6.3.0.

javsalgar (approved):

benjamin658:
Hi @javsalgar @alexerm , I also made a PR before this one: https://github.com/bitnami/charts/pull/6167
The weird thing is that the Artifact Hub nats version is still 6.2.5, even if this PR bump version to 6.3.0.

Any idea?

alexerm:
Hi @benjamin658 No idea, I'm a first-time contributor here. I would be happy to see the latest version available on Artifact Hub too. 

@javsalgar Is it possible that some validation doesn't pass on the Artifact Hub side so it rejects the new version?

benjamin658:
ArtifactHub will fetch the index.yaml of the bitnami helm repo, it seems that somehow the CI doesn't repackage and update the index.yaml.

javsalgar:
Yes, we had issues with our packaging CI. We expect them to be fixed already.

benjamin658:
The version is still 6.2.5, any updated?

javsalgar:
I see there are issues in our testing

```
[2021-04-27T01:26:05.937Z] nats-server: Parse error on line 29: 'Expected a map value terminator "," or a map terminator "}", but got 'v' instead.'
```

We're checking it

PR #6178, opened 2021-04-21
[bitnami/mysql] fixing plugin_dir option in my.cnf

torz:
fixing plugin_dir option in my.cnf to point to the correct directory

The current plugin_dir option in my.cnf points in the mysql chart point to a path that does not exists in the docker image. This fix should correct that. Currently I'm using this chart with the following env var:
extraFlags: "--plugin-dir=/opt/bitnami/mysql/lib/plugin"

This a fix for a misconfiguration

NA

  - fixes #

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/mysql/Chart.yaml
+++ b/bitnami/mysql/Chart.yaml
@@ -25,4 +25,4 @@ name: mysql
 sources:
   - https://github.com/bitnami/bitnami-docker-mysql
   - https://mysql.com
-version: 8.5.4
+version: 8.5.5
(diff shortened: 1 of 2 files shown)


juan131 (approved):
Thanks for this fix!! LGTM

Actually the image sets this setting to that directory by default, but we're overwriting the config file with a wrong setting, see:

```
$ docker run --rm -it bitnami/mysql:latest -- grep "plugin_dir" /opt/bitnami/mysql/conf/my.cnf
(...)
plugin_dir=/opt/bitnami/mysql/lib/plugin
```

PR #6208, opened 2021-04-24
[bitnami/postgresql-ha] remove ldap restriction on pgpool probes

syphr42:
**Description of the change**

Remove the requirement that LDAP be disabled to enable pgpool liveness/readiness/startup probes.

**Benefits**

Currently, to enable pgpool probes with LDAP also enabled, you have to dig into the deployment template and copy the probes to custom* versions in your values YAML. This requires reproducing the values from the deployment and hardcoding parts of it. This increases maintenance requirements for the user.

**Possible drawbacks**

I'm not sure why the probes were disabled with LDAP originally as I can't find any explanation in the history. I made this a major version change since it has at least the potential to be breaking. If the configured user and password could not authenticate, I believe other configurations in this chart would fail as well. However, the liveness and readiness probes are enabled by default and if the user was relying on the fact that enabling LDAP disabled the probes (without doing so explicitly), there might be a surprise if this change is not understood first.

**Applicable issues**

N/A

**Additional information**

I tested this with custom probes performing the same action as the default probes on my own deployment of postgresql-ha with LDAP enabled for the past week and I have not seen a problem from enabling these probes. Furthermore, having the probes enabled would have helped me avoid an outage earlier this month when one of the pool members lost connectivity to the backend causing intermittent errors.

**Checklist** 

### DIFF
--- a/bitnami/postgresql-ha/Chart.yaml
+++ b/bitnami/postgresql-ha/Chart.yaml
@@ -27,4 +27,4 @@ name: postgresql-ha
 sources:
   - https://github.com/bitnami/bitnami-docker-postgresql
   - https://www.postgresql.org/
-version: 6.9.1
+version: 7.0.0
--- a/bitnami/postgresql-ha/templates/pgpool/deployment.yaml
+++ b/bitnami/postgresql-ha/templates/pgpool/deployment.yaml
@@ -272,7 +272,7 @@ spec:
             - name: postgresql
               containerPort: 5432
               protocol: TCP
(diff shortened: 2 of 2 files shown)


alvneiayu (approved):
LGTM

alvneiayu:
hi @syphr42

I verified that the livenessProbe, readinessProbe and startupProbe, they are going to use root user and the root credential is not located in the LDAP. So, the changes LGTM.

Thanks for the PR

Álvaro

PR #6227, opened 2021-04-27
[bitnami/redis] format examples as yaml

unixlab:
**Description of the change**
I like if you can copy&paste the examples.
Currently two of the examples in the README are not yaml compliant.
This fix that.

Additionally I removed the quotes from the bool field `tls.enabled` and added quotes for the string field `auth.existingSecret` according to helm best practices.

**Benefits**
Working examples on how to configure TLS.

**Possible drawbacks**
. / .

**Applicable issues**
. / .

**Additional information**
. / .

**Checklist** 

### DIFF
--- a/bitnami/redis/Chart.yaml
+++ b/bitnami/redis/Chart.yaml
@@ -25,4 +25,4 @@ name: redis
 sources:
   - https://github.com/bitnami/bitnami-docker-redis
   - http://redis.io/
-version: 14.1.0
+version: 14.1.1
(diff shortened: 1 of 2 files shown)


carrodher:
Hi, thanks for your contribution. Unfortunately, those examples were not created to be used directly as yaml, in fact `foo.bar: true` doesn't work when you use it in a _values.yaml_ file, you would need to format the values as a real yaml.

The current examples are designed to be used as part of the `helm install` command with the `--set` flag, to use them in the yaml file we recommend editing the _values.yaml_ itself or copy the format from this place.

I'm sorry but this is the standard used in all the charts so I can't accept this PR because of the above reasons. Thanks anyways for your contribution and interest in Bitnami Helm Charts!

PR #6230, opened 2021-04-27
Update README.md

holmesb:
Wrong key name.  Also PVC template requires a list.  String "ReadWriteOnce" produces error "range can't iterate over ReadWriteOnce"


alemorcuq (approved):
Thanks for the PR!

PR #6232, opened 2021-04-27
[bitnami/postgresql] SHM: remove size limit on the emptyDir.

desaintmartin:
The size of this emptyDir in memory flavor will count against your container's memory limit, so no need to have a separate limit.
A lot of parallel queries on huge workloads can be greater than 1GB.

Note that currently, if postgresql uses more than 1GB of /dev/shm, it will get silently evicted by kubelet without meaningful logs (at least in kube 1.19), and the only trace is an event that you can't get when filtering for the pod's events, because the StatefulSet replaces the evicted Pod by the new Pod.

**Checklist** 


Mauraza (changes requested):
Hi @desaintmartin,

Thanks for the PR!!, could you add a variable to use or not the `sizeLimit` or/and variable to change the size? In this case, if someone wants to use `simeLimit` he/she can do it.

desaintmartin:
Done!

Mauraza (approved):
Hi @desaintmartin,

Thanks again for the PR!! LGTM!!

PR #6245, opened 2021-04-28
[bitnami/kubeapps] bump chart version to 6.1.1

antgamdia:
Signed-off-by: antgamdia agamez@vmware.com

**Description of the change**

i) A new param `isKubeappsCluster` is introduced in the values.yaml. This optional param will allow defining the cluster in which Kubeapps is being installed in case it cannot be automatically inferred. For instance, when every cluster uses an APIServiceURL we are not able to know which one is this cluster.
This param is enforced in both Kubeops and the chart template. Therefore, helm will throw an error if the chart does not define properly the cluster in which kubeapps in installed.

ii) Two params are introduced for passing namespaces in the header. The name of the header field and its value are defined by 2 environment variables.

iii) Upgrade the chart dependencies as part of the usual Kubeapps release process.

Additionnaly, 

**Benefits**

i) Kubeapps will work in OIDC scenarios where it is required to set an APIServiceURL for each cluster (even a single one; for instance, using Pinniped 0.7 impersonation proxy in a managed cluster).

ii) It prevents looping over all namespaces in the cluster.

iii) Chart deps up to date

**Possible drawbacks**

N/A

**Applicable issues**

https://github.com/kubeapps/kubeapps/pull/2696
https://github.com/kubeapps/kubeapps/pull/2671

**Additional information**

As a Kubeapps developer, the reason for sending this PR is twofold: i) there is an issue with the permissions of the account we normally use for syncing the changes; ii) we are about to change the release process, so that we (Kubeapps) will be responsible for sending the PR with the proper chart changes.

**Checklist** 

### DIFF
--- a/bitnami/kubeapps/Chart.yaml
+++ b/bitnami/kubeapps/Chart.yaml
@@ -25,4 +25,4 @@ maintainers:
 name: kubeapps
 sources:
   - https://github.com/kubeapps/kubeapps
-version: 6.0.1
+version: 6.1.1
--- a/bitnami/kubeapps/Chart.lock
+++ b/bitnami/kubeapps/Chart.lock
@@ -1,9 +1,9 @@
 dependencies:
 - name: common
   repository: https://charts.bitnami.com/bitnami
-  version: 1.4.1
+  version: 1.4.3
 - name: postgresql
   repository: https://charts.bitnami.com/bitnami
-  version: 10.3.13
-digest: sha256:cf07bec4f0cbca067c8aa1327bb0df11924e3cd19b956b21c3e41513a0c2de16
-generated: "2021-03-23T13:14:24.11473167+01:00"
+  version: 10.3.18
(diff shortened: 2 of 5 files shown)


bitnami/kubeapps/Chart.yaml
```diff
@@ -25,4 +25,4 @@ maintainers:
 name: kubeapps
 sources:
   - https://github.com/kubeapps/kubeapps
-version: 6.0.1
+version: 6.1.1
```
antgamdia, line 28: This is tricky... we released the 6.1.0, but it never got synced due to the permissions issue. Later on, we bumped it to 6.1.1 (which is the current version in Kubeapps master).
In order to avoid semver problems, perhaps we better talk about what to do here (fix the permissions and re-run the sync from our side, split this PR into two, etc.)

juan131 (approved):
Thanks so much for syncing the changes @antgamdia !!! Approving your PR

bitnami/kubeapps/Chart.yaml
juan131: Fair enough!

PR #6260, opened 2021-04-30
[bitnami/rabbitmq] feat: servicemonitor and ingress configuration

AlexisMtr:
**Description of the change**

They are 2 changes in this PR
* allow to change the path used by the ServiceMonitor to be able to scrap `per-object` metrics. 
By default, th path stay `/metrics`, but it can be override by `/metrics/per-object`
* allow to use template for ingress definition (host and tls) to use `.Release.Namespace` for example

**Benefits**

Modifications about service Monitor can be used to scrape more detailed metrics by changing the path to `/metrics/per-object`

Modifications about ingress allow to use helm variables (.Release.Namespace, .Chart.Version, ...) or to use a helper to compute the hostname

**Possible drawbacks**

**Applicable issues**

**Additional information**

**Checklist** 


jBouyoud (approved):

alemorcuq (changes requested):
Thanks for the PR, @AlexisMtr. We have a helper function `common.tplvalues.render` (https://github.com/bitnami/charts/blob/master/bitnami/common/templates/_tplvalues.tpl) that we use instead of `tpl` to allow using templates. Could you use it instead? You can see an example of use in the InfluxDB chart: https://github.com/bitnami/charts/blob/master/bitnami/influxdb/templates/ingress.yaml

bitnami/rabbitmq/templates/ingress.yaml
```diff
@@ -44,8 +44,14 @@ spec:
...
-    {{- if .Values.ingress.extraTls }}
-    {{- toYaml .Values.ingress.extraTls | nindent 4 }}
+    {{- range .Values.ingress.extraTls }}
+    - hosts:
+      {{- range .hosts }}
+      - {{ include "common.tplvalues.render" ( dict "value" . "context" $ )}}
+      {{- end }}
+      {{- with .secretName }}
+      secretName: {{ include "common.tplvalues.render" ( dict "value" . "context" $ ) }}
+      {{- end }}
```
alemorcuq, line 54: Why do you want to breakdown this `extraTls` settings? Isn't it enough with something like:

```
    {{- include "common.tplvalues.render" ( dict "value" .Values.ingress.extraTls "context" $ ) | nindent 4 }}
```

as we do in the rest of the charts? What's your use case for this?

alemorcuq (commented):
Thanks for the changes. I've added a question below.

bitnami/rabbitmq/templates/ingress.yaml
AlexisMtr: It's a mistake, I forgot to revert the change when I have switch from "tpl" to custom "tplvalue" helper. You right, this change is not necessary, I revers to the default

PR #6261, opened 2021-04-30
[bitnami/wordpress] MAJOR: WordPress image refactor

marcosbc:
**Description of the change**

Preventive major bump of WordPress chart + Add support for new features

**Benefits**

Users can enjoy an improved container image :) 

**Possible drawbacks**

Upgrades are not guaranteed to work

**Applicable issues**

  - fixes #

**Additional information**

**Checklist** 


javsalgar (approved):
The PR overall looks good to me. I would clarify that the multisite options will only work on first boot.

bitnami/wordpress/templates/deployment.yaml
```diff
@@ -145,20 +145,36 @@ spec:
...
             - name: WORDPRESS_TABLE_PREFIX
               value: {{ .Values.wordpressTablePrefix | quote }}
             - name: WORDPRESS_SCHEME
               value: {{ .Values.wordpressScheme | quote }}
             - name: WORDPRESS_EXTRA_WP_CONFIG_CONTENT
               value: {{ .Values.wordpressExtraConfigContent | quote }}
+            - name: WORDPRESS_AUTO_UPDATE_LEVEL
+              value: {{ .Values.wordpressAutoUpdateLevel | quote }}
+            - name: WORDPRESS_PLUGINS
+              value: {{ .Values.wordpressPlugins | quote }}
```
juan131, line 163: I'd use an array format for `wordpressPlugins`. Then, you can use:
```suggestion
              value: {{ join "," .Values.wordpressPlugins | quote }}
```

bitnami/wordpress/values.yaml
```diff
@@ -131,6 +131,13 @@ existingWordPressConfigurationSecret:
 ## NOTE: useful if you deploy Memcached for caching database queries or you use an external cache server
 ##
 wordpressConfigureCache: false
+## @param wordpressAutoUpdateLevel Level of auto-updates to allow. Allowed values: `major`, `minor` or `none`.
+##
+wordpressAutoUpdateLevel: none
+## @param wordpressPlugins Plugins to install and activate, comma-separated. Can also be set to `all` or `none`.
+## NOTE: If set to all, only plugins that are already installed will be activated, and if set to none, no plugins will be activated
+##
+wordpressPlugins: none
```
juan131, line 140: I wouldn't use a "comma-separated" format, but an array instead. In addition to this, shouldn't be possible to indicate what plugins to install and what plugins to activate separately?

juan131 (changes requested):
Thanks for the great PR! Please check my comments

bitnami/wordpress/values.yaml
marcosbc: Well, I guess it should be possible to allow for both arrays, comma-separated and strings (like `none` or `all`). I'll update the description.

marcosbc: BTW I don't think adding complexity to the users (with respect to having the option to specify plugins to activate and install separately) is something we should be doing, at least in the first iteration of these changes. I think it is prone to cause more confusion than good.

juan131: > Well, I guess it should be possible to allow for both arrays, comma-separated and strings (like none or all)

It seems this new approach to support both a string and an array works:

```console
$ values="$(cat << EOF
wordpressPlugins:
  - foo
  - bar
EOF
)"
$ helm template wordpress bitnami/wordpress -f <(echo "$values") -s templates/deployment.yaml
(...)
            - name: WORDPRESS_PLUGINS
              value: "foo,bar"
$ helm template wordpress bitnami/wordpress -s templates/deployment.yaml
(...)
            - name: WORDPRESS_PLUGINS
              value: "none"
```

juan131 (approved):
LGTM

PR #6267, opened 2021-04-30
[bitnami/postgresql-ha] - Added session affinity support in pgpool service

adam-prohack:
# Description of the change
When I run lots of microservics instances which connects to same database and same login, pgpool-II coulnd't properly cache session when runs in more than one pod. When pgpool runs in more than one pod, default service load balancing mechanism multiply frontend sessions and in the result multiply backend sessions. 

# Benefits
Better database resources protection and properly fronend session caching.


bitnami/postgresql-ha/Chart.yaml
```diff
@@ -27,4 +27,4 @@ name: postgresql-ha
 sources:
   - https://github.com/bitnami/bitnami-docker-postgresql
   - https://www.postgresql.org/
-version: 7.1.0
+version: 7.1.1
```
carrodher, line 30: As it's a feature addition, we should bump the minor version:
```suggestion
version: 7.2.0
```

carrodher (changes requested):
Thanks for your contribution, I just added some inline comments to follow the same approach used in other charts. Apart from that, it's needed to add the parameter in the values.yaml, under the `service:` block:
```yaml
   ## Control where client requests go, to the same pod or round-robin
   ## Values: ClientIP or None
   ## ref: https://kubernetes.io/docs/user-guide/services/
   ##
   sessionAffinity: "None"
```

bitnami/postgresql-ha/Chart.yaml
```diff
@@ -28,4 +28,3 @@ sources:
   - https://github.com/bitnami/bitnami-docker-postgresql
   - https://www.postgresql.org/
 version: 7.2.0
```
carrodher, line 30: A new version appeared in the meantime
```suggestion
version: 7.3.0
```

adam-prohack:
@carrodher thanks for your comments. All things are fixed. I am waiting for merge PR

carrodher (approved):
LGTM

PR #6286, opened 2021-05-04
[bitnami/common] pull secrets rendering

### LINKED ISSUE #6229: [bitnami/nginx] evaluate pullSecrets as template
**Which chart**:
bitnami/nginx (v. 8.8.4)

I have nginx chart as a dependency of a parent custom chart that creates the docker-registry secrets automatically. 
The parent chart generates the image pull secrets Secret name using the .Release.Name value from helm.
Because neither `.Values.image.pullSecrets` nor `.Values.global.imagePullSecrets` values are evaluated as a template I've no way of declaring the generated Secret name in nginx chart.

I would like to be able to specify templated strings within `.Values.image.pullSecrets` and/or `.Values.global.imagePullSecrets` to support situations where the the Secrets is generated through helm. This can probably be achieved by changing the `bitnami/common` chart to enable this feature across all bitnami charts.

_Note: I'm not sure if creating image pull secrets from Helm is the right way to do it (after all the credentials, using this approach, are written in plain text within chart values). If this feature is not available on purpose due to security concern I would like to know which approach should be used instead (e.g. the Secret should always be created manually before deploying the chart)._

The generated Secret has a static name that does not depends on any Helm value.

DaviGia:
**Description of the change**

Changes the `common.images.pullSecrets` function to render both current context pullSecrets and global imagePullSecrets as templates.

**Benefits**

The user can now specify templated strings as pull secrets. This enables a parent Chart to generate Secrets programmatically and pass the name of the secrets to all other subcharts that use bitnami/common chart.

**Possible drawbacks**

Changes the signature of a function used by many charts within bitnami repository.

**Applicable issues**

  - fixes #6229

**Checklist** 

### DIFF
--- a/bitnami/common/Chart.yaml
+++ b/bitnami/common/Chart.yaml
@@ -2,7 +2,7 @@ annotations:
   category: Infrastructure
 apiVersion: v2
 # Please make sure that version and appVersion are always the same.
-appVersion: 2.1.0
+appVersion: 1.5.0
 description: A Library Helm Chart for grouping common logic between bitnami charts. This chart is not deployable by itself.
 home: https://github.com/bitnami/charts/tree/master/bitnami/common
 icon: https://bitnami.com/downloads/logos/bitnami-mark.png
@@ -20,4 +20,4 @@ sources:
   - https://github.com/bitnami/charts
   - http://www.bitnami.com/
 type: library
-version: 2.1.0
+version: 1.5.0
--- a/bitnami/common/Chart.yaml
+++ b/bitnami/common/Chart.yaml
@@ -2,7 +2,7 @@ annotations:
   category: Infrastructure
 apiVersion: v2
 # Please make sure that version and appVersion are always the same.
-appVersion: 1.4.3
+appVersion: 2.0.0
 description: A Library Helm Chart for grouping common logic between bitnami charts. This chart is not deployable by itself.
 home: https://github.com/bitnami/charts/tree/master/bitnami/common
 icon: https://bitnami.com/downloads/logos/bitnami-mark.png
@@ -20,4 +20,4 @@ sources:
   - https://github.com/bitnami/charts
   - http://www.bitnami.com/
 type: library
-version: 1.4.3
+version: 2.0.0
--- a/bitnami/common/Chart.yaml
+++ b/bitnami/common/Chart.yaml
@@ -2,7 +2,7 @@ annotations:
   category: Infrastructure
 apiVersion: v2
 # Please make sure that version and appVersion are always the same.
-appVersion: 2.0.0
+appVersion: 2.1.0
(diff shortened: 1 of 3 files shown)


bitnami/common/templates/_images.tpl
```diff
@@ -21,20 +21,21 @@ Return the proper image name
 
 {{/*
 Return the proper Docker Image Registry Secret Names
-{{ include "common.images.pullSecrets" ( dict "images" (list .Values.path.to.the.image1, .Values.path.to.the.image2) "global" .Values.global) }}
+{{ include "common.images.pullSecrets" ( dict "images" (list .Values.path.to.the.image1, .Values.path.to.the.image2) "context" $) }}
```
javsalgar, line 24: I'm concerned of the major change that this would imply. Couldn't we make it backwards compatible. My opinion:

Instead of substituting global with context, I would add context as a third parameter. If "context" is set, then you could use tplvalues.render. If context is not set, then use the values as they are. This way, it would be backwards compatible and we wouldn't need to bump the major in all the charts that use common. How does that sound?

javsalgar:
Thank you for the PR! Please check my concern and proposal

bitnami/common/templates/_images.tpl
DaviGia: I really don't know if it's a good practice in Helm to use optional parameters. Moreover global values can be retrieved from the context anyway, so those parameters seems almost duplicates. In this case, why not create another function? Is it any better?

javsalgar: As a backwards compatibliity measure, the idea would be better. Ideally, at some point, we could go changing the calls in several charts, and then in the future get rid of common.images.pullSecrets. Maybe we could call the function `common.images.renderPullSecrets`.

DaviGia: Hi, I've reverted the changes to `common.images.pullSecrets` function and I've created `common.images.renderPullSecrets`.

DaviGia: As of right now I would like to have this functionality into several charts that I use. Let me know if I can update those charts to use this new function with this PR or if I have to create another one.

javsalgar: Hi,

We'd rather have different PRs

bitnami/common/Chart.yaml
```diff
@@ -20,4 +20,4 @@ sources:
   - https://github.com/bitnami/charts
   - http://www.bitnami.com/
 type: library
-version: 1.4.3
+version: 2.1.0
```
javsalgar, line 23: Please bump the minor version, as not it is not adding a breaking change

javsalgar (changes requested):
Small change and we can proceed with the merging

bitnami/common/Chart.yaml
DaviGia: Done

PR #6291, opened 2021-05-05
bitnami/dataplatform-bp1 and bitnami/dataplatform-bp2 charts updated with extra set of affinity rules.

yogesh-reddy:
**Description of the change**
The pod affinity and anti affinity rules were updated for bitnami/dataplatform-bp1 , bitnami/dataplatform-bp2 charts because when we have two deployments from both the blueprints on the same cluster the affinity rules would conflict each other, as two blueprints have some pods in common. To overcome this extra set of affinity rules were added on the release name for both the blueprints.  

**Benefits**
Two blueprints can be added on the same cluster and namespace without conflicts if provided with the sufficient resources.

**Possible drawbacks**

**Applicable issues**

  - fixes #

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/dataplatform-bp1/Chart.yaml
+++ b/bitnami/dataplatform-bp1/Chart.yaml
@@ -60,3 +60,4 @@ sources:
   - https://github.com/wavefrontHQ/wavefront-collector-for-kubernetes
   - https://github.com/wavefrontHQ/wavefront-proxy
 version: 1.0.2
+
--- a/bitnami/dataplatform-bp1/Chart.yaml
+++ b/bitnami/dataplatform-bp1/Chart.yaml
@@ -59,4 +59,4 @@ sources:
   - https://github.com/bitnami/bitnami-docker-wavefront-proxy
   - https://github.com/wavefrontHQ/wavefront-collector-for-kubernetes
   - https://github.com/wavefrontHQ/wavefront-proxy
-version: 1.0.2
+version: 2.0.0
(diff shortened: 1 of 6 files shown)


yogesh-reddy:
@gkulkarni32 Try to review the code .

gkulkarni32:
@miguelaeh Yogesh is from my team at OCTO Cloud Architecture. We need a small change to be done in the affinity and anti-affinity rules to include the release name as well so it doesnt apply the affinity across a different release of the data platform. Please review this from your side.

miguelaeh (changes requested):
Hi @yogesh-reddy ,
Thank you very much for the PRs.
I see you bumped the versions in a different way while the change seems to be the same for both charts.
Apart from that, modifying the default `values.yaml` usually requires bumping the major version since updating a release could end in unexpected behaviors. 
Bumping the major, we need to add to both READMEs a note in the `# Upgrading` section.
Also, I see you are using the `{{ .Release.Name }}` in the affinity rules, I am not sure if this is going to be templatized properly, could you confirm, please?

gkulkarni32:
Thank you @miguelaeh 

We will change the versions to the following major versions in both the charts and add a section in README for upgrade. 

dataplatform-bp1: 2.0.0
dataplatform-bp2: 1.0.0

Thanks
Geeta

miguelaeh:
Hi @gkulkarni32 ,

>Also, I see you are using the {{ .Release.Name }} in the affinity rules, I am not sure if this is going to be templatized properly, could you confirm, please?

What about this? Could you re-check it works please?

miguelaeh (approved):
Thank you very much for addressing all the changes.
LGTM!

PR #6298, opened 2021-05-05
[bitnami/airflow] Take worker.resources in values.yaml into consideration, when using KubernetesExecutor

### LINKED ISSUE #6279: [bitnami/airflow] resources ignored for worker in KubernetesExecutor
**Which chart**:
Airflow

**Which version**:
2.0.2-debian-10-r7

**Describe the bug**
I don't know if this is intentional, but when using KubernetesExecutor ```worker.resources``` in ```values.yaml``` is ignored. 
This is a problem for instance in openshift, when administrators have set up default limits for containers, because then you would regularly run out of memory.

**To Reproduce**
Just start a worker and verify the YAML file that is generated when starting a worker with KubernetesExecutor - you will see that ```worker.resources```  is empty, even if you have defined it in your values.yaml

**Expected behavior**
I would expect resources settings in ```values.yaml``` to be taken into consideration, or at least get a warning that when using KubernetesExecutor it will be ignored.
The fix is easy, if this is not intentional. Just include in ```template/config/configmap.yml``` the resources

```
 containers:
        - name: airflow-worker
          image: {{ include "airflow.workerImage" . }}
          imagePullPolicy: {{ .Values.worker.image.pullPolicy }}
          securityContext: {{- omit .Values.containerSecurityContext "enabled" | toYaml | nindent 12 }}
        {{- end }}
          {{- if .Values.worker.resources }}
          resources: {{- toYaml .Values.worker.resources | nindent 12 }}
          {{- end }}
          env:
```

diegoami:
**Description of the change**

Ensure that `worker.resources` specified in `values.yaml` are taken into consideration also when spinning a new worker

**Benefits**

Limits and requests in `worker.resources` have an effect 

**Possible drawbacks**

None

**Applicable issues**

  - fixes #6279

**Checklist** 


bitnami/airflow/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 annotations:
   category: WorkFlow
 apiVersion: v2
-appVersion: 2.0.2
+appVersion: 2.0.3
```
juan131, line 4: You shouldn't bump the `appVersion` but `version` instaed.

juan131 (changes requested):
Thanks for this fix!! Please take a look to my comment

bitnami/airflow/Chart.yaml
diegoami: Ok, done

juan131 (approved):
Thanks!

PR #6301, opened 2021-05-05
[bitnami/kubeapps] New major version: standardization

juan131:

**Description of the change**

This PR refactors the whole chart to follow Bitnami best practices for Helm charts. This includes some breaking changes, so please review it carefully,

The most important change are:

- Labels are adapted to follow the [Helm charts standard labels](https://helm.sh/docs/chart_best_practices/labels/#standard-labels).
- Restructuring manifests: one folder per component.
- Include several common parameters available in other charts of the catalog.
- Values adapted to follow [readmenator](https://github.com/bitnami-labs/readmenator]) syntax.

**Benefits**

- Standardization.
- Reduces technical debt.

**Possible drawbacks**

It breaks backwards compatibility since the new labels format is not compatible with upgrades. Check the "Upgrading to chart version 7.0.0" section in the **README.md** for more information.

**Applicable issues**

N/A

**Additional information**

N/A

**Checklist** 


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -37,5 +37,3 @@ request, mention that information here.-->
 - [ ] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
 - [ ] Variables are documented in the README.md
 - [ ] Title of the PR starts with chart name (e.g. `[bitnami/chart]`)
-
-:warning: Keep in mind that if you want to make changes to the kubeapps chart, please implement them in the [kubeapps repository](https://github.com/kubeapps/kubeapps/tree/master/chart/kubeapps). This is only a synchronized mirror.
```
absoludity, line 41: I guess in the past this comment was targeted at bitnami folk who may have attempted a PR against this chart... where as now bitnami folk *should* update this chart. But I'm not sure what the best generally. We'll be making changes in our development chart and then syncing them here via a PR for development-related changes. Perhaps we could still say:

:warning: Updates and bug fixes for the chart belong in this repository but functionality for new Kubeapps features should be included in the [development chart in the Kubeapps repository](https://github.com/kubeapps/kubeapps/tree/master/chart/kubeapps).

Maybe it's not necessary, see what you think.

bitnami/kubeapps/Chart.yaml
```diff
@@ -10,7 +10,7 @@ dependencies:
     version: 1.x.x
   - name: postgresql
     repository: https://charts.bitnami.com/bitnami
-    version: '10.X.X'
+    version: 10.X.X
```
absoludity, line 13: Tiny point for consistency: the common chart uses lower-case `x`, where as this one uses upper. Ignore as needed, just my brain noticing inconsistencies :)

bitnami/kubeapps/README.md
```diff
@@ -51,7 +52,417 @@ Once you have installed Kubeapps follow the [Getting Started Guide](https://gith
...
+
+| Name                      | Description                                                                   | Value   |
+| ------------------------- | ----------------------------------------------------------------------------- | ------- |
+| `kubeVersion`             | Override Kubernetes version                                                   | `nil`   |
+| `nameOverride`            | String to partially override common.names.fullname                            | `nil`   |
+| `fullnameOverride`        | String to fully override common.names.fullname                                | `nil`   |
+| `commonLabels`            | Labels to add to all deployed objects                                         | `{}`    |
+| `commonAnnotations`       | Annotations to add to all deployed objects                                    | `{}`    |
+| `extraDeploy`             | Array of extra objects to deploy with the release                             | `[]`    |
+| `allowNamespaceDiscovery` | Allow users to discover available namespaces (only the ones they have access) | `true`  |
```
absoludity, line 74: Looks like "common parameters" is a collection of all the top-level options in the values.yaml that aren't for a specific service? Nice. Some of them might not fit perfectly, but it's not an issue, imo. EDIT: we could move `allowNamespaceDiscovery`, `clusters` and `featureFlags` to "Other parameters"? Don't mind either way.

bitnami/kubeapps/README.md
```diff
@@ -51,7 +52,417 @@ Once you have installed Kubeapps follow the [Getting Started Guide](https://gith
...
+| `frontend.service.nodePort`                      | Node port for HTTP                                                                        | `nil`                   |
+| `frontend.service.clusterIP`                     | Frontend service Cluster IP                                                               | `nil`                   |
+| `frontend.service.loadBalancerIP`                | Frontend service Load Balancer IP                                                         | `nil`                   |
+| `frontend.service.loadBalancerSourceRanges`      | Frontend service Load Balancer sources                                                    | `[]`                    |
+| `frontend.service.externalTrafficPolicy`         | Frontend service external traffic policy                                                  | `Cluster`               |
+| `frontend.service.annotations`                   | Additional custom annotations for frontend service                                        | `{}`                    |
+
+
+### AppRepository parameters
```
absoludity, line 229:
```suggestion
### AppRepository Controller parameters
```
Just to be clear what the parameters are for.

bitnami/kubeapps/README.md
```diff
@@ -51,7 +52,417 @@ Once you have installed Kubeapps follow the [Getting Started Guide](https://gith
...
+| `pinnipedProxy.resources.limits.memory`               | The memory limits for the Pinniped Proxy container                       | `128Mi`                           |
+| `pinnipedProxy.resources.requests.cpu`                | The requested CPU for the Pinniped Proxy container                       | `25m`                             |
+| `pinnipedProxy.resources.requests.memory`             | The requested memory for the Pinniped Proxy container                    | `32Mi`                            |
+
+
+### Other Parameters
+
+| Name                    | Description                                        | Value                   |
+| ----------------------- | -------------------------------------------------- | ----------------------- |
+| `rbac.create`           | Specifies whether RBAC resources should be created | `true`                  |
```
absoludity, line 443: Woah - is that from before all clusters had RBAC enabled? Looks like it was added [here](https://github.com/kubeapps/kubeapps/pull/483). Happy to delete that option?

bitnami/kubeapps/README.md
```diff
@@ -103,7 +514,6 @@ To enable ingress integration, please set `ingress.enabled` to `true`
 
 Most likely you will only want to have one hostname that maps to this Kubeapps installation (use the `ingress.hostname` parameter to set the hostname), however, it is possible to have more than one host. To facilitate this, the `ingress.extraHosts` object is an array.
 
-If you plan to serve Kubeapps under a subpath (eg., `example.com/subpath`), you will have to disable the default path by setting `ingress.hostname=""` and the enter the hostname and path in the extraHost array; for instance: `ingress.extraHosts[0].name="example.com"` and `ingress.extraHosts[0].path="/subpath"`
```
absoludity, line 106: Did you remove this because the settings changed? It was only added 5 months ago (one of @antgamdia 's first) after testing that this works as requested by users, it'd be a shame to remove it if it still works as documented.

bitnami/kubeapps/README.md
```diff
@@ -140,8 +550,8 @@ In the first two cases, it's needed a certificate and a key. We would expect the
   ```
 
 - If you are going to use Helm to manage the certificates based on the parameters, please copy these values into the `certificate` and `key` values for a given `ingress.secrets` entry.
-- In case you are going to manage TLS secrects separately, please know that you can must a TLS secret with name *INGRESS_HOSTNAME-tls* (where *INGRESS_HOSTNAME* is a placeholder to be replaced with the hostname you set using the `ingress.hostname` parameter).
-- To use self-signed certificates created by Helm, set `ingress.tls` to `true` and `ingress.certManager` to `false`.
+- In case you are going to manage TLS secrets separately, please know that you can must a TLS secret with name *INGRESS_HOSTNAME-tls* (where *INGRESS_HOSTNAME* is a placeholder to be replaced with the hostname you set using the `ingress.hostname` parameter).
```
absoludity, line 553:
```suggestion
- In case you are going to manage TLS secrets separately, please know that you must use a TLS secret with name *INGRESS_HOSTNAME-tls* (where *INGRESS_HOSTNAME* is a placeholder to be replaced with the hostname you set using the `ingress.hostname` parameter).
```
There was a second typo :P

bitnami/kubeapps/README.md
absoludity: Ah, nm - I see you removed it because it's got its own FAQ entry below. Great.

bitnami/kubeapps/README.md
```diff
@@ -330,102 +741,61 @@ helm install --name kubeapps --namespace kubeapps bitnami/kubeapps --set rbac.cr
...
+In this release, no breaking changes were included in Kubeapps (version 2.3.2). However, the chart adopted the standardizations included in the rest of the charts in the Bitnami catalog.
 
-```bash
-kubectl delete statefulset -n kubeapps kubeapps-postgresql-master kubeapps-postgresql-slave
-```
+Most of these standardizations simply add new parameters that allow to add more customizations such as adding custom env. variables, volumes or sidecar containers. That said, some of them include breaking changes:
 
-After that you should be able to upgrade Kubeapps as always and the database will be repopulated.
+- Chart labels were adapted to follow the [Helm charts standard labels](https://helm.sh/docs/chart_best_practices/labels/#standard-labels).
+- `securityContext.*` parameters are deprecated in favor of `XXX.podSecurityContext.*` and `XXX.containerSecurityContext.*`, where XXX is placeholder you need to replace with the actual component(s). For instance, to modify the container security context for "kubeops" use `kubeops.podSecurityContext` and `kubeops.containerSecurityContext` parameters.
```
absoludity, line 798: I expected here to read what steps a user will need to take to upgrade? Or if I update the security context options (or haven't used them) will it just work? I don't think the changed labels should break an upgrade (which is not dependent on those labels for deployment/pod relationships etc.)?

bitnami/kubeapps/templates/dashboard/configmap.yaml
```diff
@@ -2,17 +2,24 @@ apiVersion: v1
...
+    {{- include "common.tplvalues.render" ( dict "value" .Values.commonLabels "context" . ) | nindent 4 }}
+    {{- end }}
+  {{- if .Values.commonAnnotations }}
+  annotations: {{- include "common.tplvalues.render" ( dict "value" .Values.commonAnnotations "context" $ ) | nindent 4 }}
+  {{- end }}
 data:
   vhost.conf: |-
     server {
-      listen {{ .Values.dashboard.service.port }};
+      listen {{ .Values.dashboard.containerPort }};
```
absoludity, line 17: I guess this change isn't a breaking change? (Unless I've set service.port to a non-default value in my values.yaml, but then it should error anyway if that's no longer an option).

bitnami/kubeapps/templates/tls-secrets.yaml
```diff
@@ -5,24 +5,36 @@ apiVersion: v1
...
+  {{- end }}
 type: kubernetes.io/tls
 data:
   tls.crt: {{ .certificate | b64enc }}
   tls.key: {{ .key | b64enc }}
 ---
 {{- end }}
 {{- end }}
-{{- if and .Values.ingress.tls (not .Values.ingress.certManager) }}
+{{- if and .Values.ingress.tls .Values.ingress.selfSigned }}
```
absoludity, line 23: Woohoo :)

absoludity (commented):
LGTM, thanks Juan. I've left a few thoughts inline but nothing blocking except wanting to know exactly what breaks here if you install the latest release with the default options, then `upgrade --install` this chart?

Looking forward to adapting our dev chart and env to match the new format. :)

.github/PULL_REQUEST_TEMPLATE.md
juan131: Nice suggestion! What do you think @carrodher ?

bitnami/kubeapps/Chart.yaml
juan131: LOL I loved this one!  I'll apply it

bitnami/kubeapps/README.md
juan131: Love the suggestion. I also had the feeling these ones don't fit here but I wasn't very sure where to include them.

bitnami/kubeapps/README.md
juan131: According to the section below, the chart should work on K8s 1.15+:
```
## Prerequisites

- Kubernetes 1.15+ (tested with Azure Kubernetes Service, Google Kubernetes Engine, minikube and Docker for Desktop Kubernetes)
```
@absoludity can you confirm that RBAC is enabled on every cluster since that version (including custom on-premises clusters)?

bitnami/kubeapps/README.md
juan131: Correct! It was duplicated in the README.md

bitnami/kubeapps/README.md
juan131: The labels do break the upgrade since fields such as `spec.selector.matchLabels` are immutable in a deployment.

> I expected here to read what steps a user will need to take to upgrade?

They're mentioned in the paragraphs above (see "Error while upgrading the Chart")

bitnami/kubeapps/templates/tls-secrets.yaml
juan131: 😄

bitnami/kubeapps/templates/dashboard/configmap.yaml
juan131: Correct!
The idea is basically difference between the port you used in the container, and the port you expose in the service. In most cases both are set to the same value, but this opens the possibility to customize then independently. I know there are people that use random ports at container level for security reasons.

.github/PULL_REQUEST_TEMPLATE.md
carrodher: Since the development workflow will continue to be in the other repository, it makes sense, so we avoid possible divergences

antgamdia: I'm ok with either option; no strong opinion here.

However, since our main focus is to consider Kubeapps as a completely separate thing, perhaps adding this disclaimer in every pull request for the whole bitnami chart catalog is too muuch.
We can have it documented internally if necessary, but I'd rather avoid adding Kubeapps-related information. in the bitnami repo.

bitnami/kubeapps/templates/dashboard/deployment.yaml
```diff
@@ -0,0 +1,153 @@
...
+  {{- end }}
+spec:
+  replicas: {{ .Values.dashboard.replicaCount }}
+  selector:
+    matchLabels: {{- include "common.labels.matchLabels" . | nindent 6 }}
+      app.kubernetes.io/component: dashboard
+  template:
+    metadata:
+      annotations:
+        checksum/config: {{ include (print $.Template.BasePath "/dashboard/configmap.yaml") . | sha256sum }}
```
antgamdia, line 22: Just out of curiosity: why other pods don't have this annotation? (newbie to this label indeed)

bitnami/kubeapps/templates/frontend/deployment.yaml
```diff
@@ -0,0 +1,196 @@
...
+        - name: auth-proxy
+          image: {{ include "common.images.image" (dict "imageRoot" .Values.authProxy.image "global" .Values.global) }}
+          imagePullPolicy: {{ .Values.authProxy.image.pullPolicy | quote }}
+          {{- if .Values.authProxy.containerSecurityContext.enabled }}
+          securityContext: {{- omit .Values.authProxy.containerSecurityContext "enabled" | toYaml | nindent 12 }}
+          {{- end }}
+          args:
+            - --provider={{ required "You must fill \".Values.authProxy.provider\" with the provider. Valid values at https://pusher.github.io/oauth2_proxy/auth-configuration" .Values.authProxy.provider }}
+            - --upstream=http://localhost:8080/
+            - --http-address=0.0.0.0:3000
```
antgamdia, line 118: Shouldn't be using the newish containerPort parameter?

bitnami/kubeapps/templates/frontend/deployment.yaml
```diff
@@ -0,0 +1,196 @@
...
+            {{- end }}
+        {{- if and .Values.authProxy.enabled (not .Values.authProxy.external) }}
+        - name: auth-proxy
+          image: {{ include "common.images.image" (dict "imageRoot" .Values.authProxy.image "global" .Values.global) }}
+          imagePullPolicy: {{ .Values.authProxy.image.pullPolicy | quote }}
+          {{- if .Values.authProxy.containerSecurityContext.enabled }}
+          securityContext: {{- omit .Values.authProxy.containerSecurityContext "enabled" | toYaml | nindent 12 }}
+          {{- end }}
+          args:
+            - --provider={{ required "You must fill \".Values.authProxy.provider\" with the provider. Valid values at https://pusher.github.io/oauth2_proxy/auth-configuration" .Values.authProxy.provider }}
```
antgamdia, line 116: The website changed: 

```suggestion
            - --provider={{ required "You must fill \".Values.authProxy.provider\" with the provider. Valid values at https://oauth2-proxy.github.io/oauth2-proxy/docs/configuration/overview" .Values.authProxy.provider }}
```

bitnami/kubeapps/templates/ingress.yaml
```diff
@@ -33,15 +43,15 @@ spec:
...
-    {{- if .Values.ingress.tls }}
+    {{- if and .Values.ingress.tls (or .Values.ingress.certManager .Values.ingress.selfSigned) }}
     - hosts:
-        - {{ .Values.ingress.hostname }}
+        - {{ .Values.ingress.hostname | quote }}
       secretName: {{ printf "%s-tls" .Values.ingress.hostname }}
     {{- end }}
     {{- if .Values.ingress.extraTls }}
-    {{- include "common.tplvalues.render" ( dict "value" .Values.ingress.extraTls "context" $ ) | nindent 4 }}
+    {{- toYaml .Values.ingress.extraTls | nindent 4 }}
```
antgamdia, line 54: So this change supersedes https://github.com/kubeapps/kubeapps/pull/2731, isn't it?

.github/PULL_REQUEST_TEMPLATE.md
juan131: 🤷‍♂️  no strong opinion here

bitnami/kubeapps/templates/_helpers.tpl
```diff
@@ -49,106 +20,99 @@ We truncate at 63 chars because some Kubernetes name fields are limited to this
...
+{{- printf "%s-internal-db-secret-jobs-cleanup" (include "common.names.fullname" .) | trunc 63 | trimSuffix "-" -}}
 {{- end -}}
 
 {{/*
 Create name for the kubeapps upgrade job
 */}}
 {{- define "kubeapps.kubeapps-jobs-upgrade.fullname" -}}
-{{ include "common.names.fullname" . }}-internal-kubeapps-jobs-upgrade
+{{- printf "%s-internal-kubeapps-jobs-upgrade" (include "common.names.fullname" .) | trunc 63 | trimSuffix "-" -}}
 {{- end -}}
```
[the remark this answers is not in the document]
juan131: I think we can remove both the template and the SA

bitnami/kubeapps/templates/apprepository/deployment.yaml
```diff
@@ -0,0 +1,94 @@
...
+            {{- end }}
+            - --repo-sync-cmd=/asset-syncer
+            - --namespace={{ .Release.Namespace }}
+            {{- if .Values.postgresql.existingSecret }}
+            - --database-secret-name={{ .Values.postgresql.existingSecret }}
+            {{- else }}
+            - --database-secret-name={{ template "kubeapps.postgresql.fullname" . }}
+            {{- end }}
+            - --database-secret-key=postgresql-password
+            - --database-url={{ template "kubeapps.postgresql.fullname" . }}:5432
```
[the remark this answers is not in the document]
juan131: Indeed, we should make it customizable in case someone modified the PostgreSQL subchart with a custom port.

bitnami/kubeapps/templates/apprepository/deployment.yaml
```diff
@@ -0,0 +1,94 @@
...
+            - --database-user=postgres
+            - --database-name=assets
```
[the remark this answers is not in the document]
juan131: I'll adapt it too

bitnami/kubeapps/templates/assetsvc/deployment.yaml
```diff
@@ -0,0 +1,102 @@
...
+          args:
+            - --database-user=postgres
+            - --database-name=assets
+            - --database-url={{ template "kubeapps.postgresql.fullname" . }}-headless:5432
+          env:
+            - name: DB_PASSWORD
+              valueFrom:
+                secretKeyRef:
+                  key: postgresql-password
+                  # TODO
```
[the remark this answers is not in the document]
juan131: Ups!

bitnami/kubeapps/templates/dashboard/configmap.yaml
```diff
@@ -58,6 +65,7 @@ data:
 {{- .Values.dashboard.customComponents | nindent 4 }}
   custom_locale.json: |-
 {{- .Values.dashboard.customLocale | toJson | nindent 4 }}
+  # TODO
```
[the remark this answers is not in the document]
juan131: Ups

bitnami/kubeapps/templates/dashboard/deployment.yaml
juan131: It's assigned only to the ones that mounts a configmap or secret. It's a way to detect changes in the ConfigMap/Secret and update the pods.

bitnami/kubeapps/templates/ingress.yaml
juan131: I'm going to revert it

bitnami/kubeapps/templates/frontend/configmap.yaml
```diff
@@ -22,11 +29,11 @@ data:
...
     proxy_set_header Authorization "Bearer $http_x_forwarded_access_token";
     {{- end }}
-    {{- range .Values.clusters }}
+{{- range .Values.clusters }}
   {{- if .certificateAuthorityData }}
   {{ .name }}-ca.pem: {{ .certificateAuthorityData }}
   {{- end }}
-  {{- end}}
+{{- end}}
   vhost.conf: |-
```
antgamdia, line 37: We have detected a small issue here, Michael sent this PR https://github.com/kubeapps/kubeapps/pull/2768 fixing it.
The point is that the `--pinniped-proxy` flag is not using the helper for getting the full name.

It'd be great to get it also solved here, no?

bitnami/kubeapps/templates/dashboard/deployment.yaml
antgamdia: Woah! That's really useful indeed. Just yesterday we hit an issue caused by a pod not being upgraded after a cm change!

https://kubernetes.slack.com/archives/C9D3TSUG4/p1620343770056800?thread_ts=1620331053.044800&cid=C9D3TSUG4

.github/PULL_REQUEST_TEMPLATE.md
absoludity: > However, since our main focus is to consider Kubeapps as a completely separate thing, perhaps adding this disclaimer in every pull request for the whole bitnami chart catalog is too muuch.
> We can have it documented internally if necessary, but I'd rather avoid adding Kubeapps-related information. in the bitnami repo.

+1, I'd not taken into account that this was a standard disclaimer for all PRs to the bitnami chart catalog. Let's just document it internally.

bitnami/kubeapps/README.md
absoludity: Erm, no? How could I possibly do that? If this is a standard option in all bitnami charts ([looks like it might be](https://github.com/bitnami/charts/search?q=rbac.create&unscoped_q=rbac.create)) - great, I have no problem with it. I had just looked up the commit and thought it was added specifically to Kubeapps for tiller support.

bitnami/kubeapps/templates/ingress.yaml
absoludity: I don't mind either way - integrating them here is fine.

bitnami/kubeapps/templates/frontend/configmap.yaml
absoludity: I've created a PR to merge the change into this branch at https://github.com/juan131/charts/pull/4

bitnami/kubeapps/README.md
juan131: Let's keep it then. I don't think it makes no harm

juan131:
Sorry I messed up it a little bit after updating from master, but I fixed it.

absoludity (approved):
Thanks Juan! I had meant to leave a +1 with my original LGTM comment, but apparently didn't.

antgamdia (approved):
LGTM as well.

Minor thing: do you have a k8s version helper for `CustomResourceDefinition` ?
We current CRD is using the deprecated `apiextensions.k8s.io/v1beta1`  instead of `apiextensions.k8s.io/v1 `

More context: https://github.com/kubeapps/kubeapps/pull/2445

Since we dropped support for <=1.15, I think we can safely change it to `/v1`

```yaml
apiVersion: apiextensions.k8s.io/v1beta1
...
```

Even it'd be great to have some openapi schema of the AppRepository CRD, but we would have to double-check it first, @absoludity :

Something like: 

```yaml
apiVersion: apiextensions.k8s.io/v1
kind: CustomResourceDefinition
metadata:
  name: apprepositories.kubeapps.com
spec:
  group: kubeapps.com
  scope: Namespaced
  names:
    kind: AppRepository
    singular: apprepository
    plural: apprepositories
    shortNames:
      - apprepos
  versions:
    - name: v1alpha1
      served: true
      storage: true
      schema:
        openAPIV3Schema:
          type: object
          properties:
            spec:
              type: object
              properties:
                type:
                  type: string
                url:
                  type: string
                auth:
                  type: object
                  properties:
                    header:
                      type: object
                      properties:
                          secretKeyRef:
                            type: object
                    customCA:
                      type: object
                      properties:
                          secretKeyRef:
                            type: object
                    properties:
                      secretKeyRef:
                        type: object
                resyncRequests:
                  type: integer
                  format: int64
                syncJobPodTemplate:
                  type: object
                  properties:
                    metadata:
                      type: object
                    spec:
                      type: object
                dockerRegistrySecrets:
                  type: array
                  items:
                    type: string
                ociRepositories:
                  type: array
                  items:
                    type: string
                tlsInsecureSkipVerify:
                  type: boolean
```

absoludity:
> LGTM as well.
> 
> Minor thing: do you have a k8s version helper for `CustomResourceDefinition` ?
> We current CRD is using the deprecated `apiextensions.k8s.io/v1beta1` instead of `apiextensions.k8s.io/v1 `
> 
> More context: [kubeapps/kubeapps#2445](https://github.com/kubeapps/kubeapps/pull/2445)
> 
> Since we dropped support for <=1.15, I think we can safely change it to `/v1`

Is it a breaking change? If not, I'm -1 to add it here - I'd rather @juan131 can land this change and we can try out the sync etc. It also gives us a good proposal for our first PR against the upstream chart?

> 
> Even it'd be great to have some openapi schema of the AppRepository CRD, but we would have to double-check it first,

AIUI, the schema is mandatory with v1, not a great-to-have? (which is why I'm -1 on squeezing the change into this PR unless necessary).

> @absoludity :
> 
> Something like:
> 
...

I think your example might be missing `.spec.filterRule`, for example, so if specified, this would be (silently?) pruned when the CR is stored? Let's add the schema carefully outside of this PR. Another option I'd considered is whether we could just add a schema such as `spec: object` in this PR so that we can update to v1 now and we add the proper schema in a subsequent PR, but I'm not sure it's worthwhile.

juan131:
> Minor thing: do you have a k8s version helper for CustomResourceDefinition ?

I guess you're talking about the macro below, right? See:

- https://github.com/bitnami/charts/blob/master/bitnami/common/templates/_capabilities.tpl#L77

antgamdia:
> Is it a breaking change? If not, I'm -1 to add it here - I'd rather .juan131 can land this change and we can try out the sync etc. It also gives us a good proposal for our first PR against the upstream chart? 
(...) I think your example might be missing .spec.filterRule,

Yes, I do agree. We can later send a PR for that specific change in the future. We should have a closer look at the schema. I just copied it from our API docs, 

> I guess you're talking about the macro below .antgamdia , right? See:

Ah, great, but we are not using it right now (https://github.com/juan131/charts/blob/kubeapps-major-version/bitnami/kubeapps/crds/apprepository-crd.yaml). But, well, if it requires adapting the manifest and adding the schema, perhaps we can leave this change out.

> Another option I'd considered is whether we could just add a schema such as spec: object in this PR so that we can update to v1 now and we add the proper schema in a subsequent PR, but I'm not sure it's worthwhile.

I'm ok with this approach, I mean, at least we get rid of the deprecation warning each time we install Kubeapps and the change in our end is just to send the proper crd schema. See what you think @juan131. Feel free to land whatever you decide, I don't want to block it.

absoludity:
1.22 is now around a month away https://www.kubernetes.dev/resources/release/ . Reprioritising accordingly.

PR #6322, opened 2021-05-09
[bitnami/mediawiki] Don't create a PVC when using existing claim

FernFerret:
This prevents a new PVC from being created when the `persistence.existingClaim` is used.

If an existing claim is used, another PVC will be created. This PVC will either remain unused or be provisioned if a default provisioner acts on it. In either case it will not be used by this chart [due to the custom claim being set](https://github.com/bitnami/charts/blob/master/bitnami/mediawiki/templates/deployment.yaml#L197).

**Description of the change**

Remove excess PVC object. This was causing my provisioner to act on the PVC and provision storage that would never be used.

**Benefits**

Reduce cost/clutter of unused PVC object.

**Possible drawbacks**

None that I can think of.

**Checklist** 

### DIFF
--- a/bitnami/mediawiki/Chart.yaml
+++ b/bitnami/mediawiki/Chart.yaml
@@ -32,4 +32,4 @@ name: mediawiki
 sources:
   - https://github.com/bitnami/bitnami-docker-mediawiki
   - http://www.mediawiki.org/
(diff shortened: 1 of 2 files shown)


miguelaeh (approved):
LGTM!

PR #6325, opened 2021-05-10
[bitnami/mariadb-galera] Add checksum to update pods when the configmap is updated

### LINKED ISSUE #3671: [bitnami/mariadb-galera]  Document how to update my.cnf settings in a running cluster
**Which chart**:
mariadb-galera-4.3.3

In a running system, its sometimes necessary to tweak my.cnf settings (for example, to establish/break replication settings). There's no obvious way to with this chart. I'm assuming its something that is possible, but not spelled out in the documentation.

Documentation for how to update the settings in a running cluster. Deleting the chart isn't acceptable because of the downtime involved (and is worse because the scaling down to zero leaves the cluster in an unknown state and has to be recovered via bootstrapping). 

- Editing the files manually, but this requires starting a pod with access to the persistant storage claim, which isn't possible with engines that are ReadWriteOnce (such as Azure's managed-premium storage class)
- Editing the ConfigMap the chart inserts, but that appears to only be evaluated when first configuring the nodes?
- Yelling profanities at my monitor. No response.

This is such a basic need for any database system, I feel like there's got to be something obvious I'm missing (and, presumably others would miss) in the documentation. Or if its not possible, it really needs to be. Half the point of having a multimaster cluster gets lost if you can't do rolling updates.

miguelaeh:
…he configmap

**Description of the change**
I bumped the minor version since even if the parameters have not changed the behaviour of the upgrade will change.
This PR allows upgrading the pods after modifying the config map.

**Benefits**

**Possible drawbacks**

**Applicable issues**

  - fixes #3671

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/mariadb-galera/Chart.yaml
+++ b/bitnami/mariadb-galera/Chart.yaml
@@ -27,4 +27,4 @@ sources:
   - https://github.com/bitnami/bitnami-docker-mariadb-galera
   - https://github.com/prometheus/mysqld_exporter
   - https://mariadb.org
-version: 5.7.1
+version: 5.8.0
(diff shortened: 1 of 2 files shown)


Mauraza (approved):
Thanks for the PR! LGTM!

PR #6348, opened 2021-05-13
[bitnami/common] Node affinity values must be quoted.

wrossmann:
Label values _must_ be strings, but if they are not quoted then
boolean/numerical/etc values can be misinterpreted by kubernetes.

eg:

```
values:
  - "true"
```
**Description of the change**

Ensure that values in `matchExpressions` for `nodeAffinity` rules are quoted as strings.

**Benefits**

Value strings that can be misinterpreted as non-string types, eg: `"true"`, will no longer be.

**Possible drawbacks**

None

**Applicable issues**

#6349 

**Additional information**

Nada

**Checklist** 

### DIFF
--- a/bitnami/common/Chart.yaml
+++ b/bitnami/common/Chart.yaml
@@ -2,7 +2,7 @@ annotations:
   category: Infrastructure
 apiVersion: v2
 # Please make sure that version and appVersion are always the same.
-appVersion: 1.4.3
+appVersion: 1.5.1
 description: A Library Helm Chart for grouping common logic between bitnami charts. This chart is not deployable by itself.
 home: https://github.com/bitnami/charts/tree/master/bitnami/common
 icon: https://bitnami.com/downloads/logos/bitnami-mark.png
@@ -20,4 +20,4 @@ sources:
   - https://github.com/bitnami/charts
   - http://www.bitnami.com/
 type: library
-version: 1.4.3
+version: 1.5.1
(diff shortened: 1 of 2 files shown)


bitnami/common/Chart.yaml
```diff
@@ -20,4 +20,4 @@ sources:
   - https://github.com/bitnami/charts
   - http://www.bitnami.com/
 type: library
-version: 1.4.3
+version: 1.4.4
```
juan131, line 23: This is a special chart and both `appVersion` and `version` should be bumped

bitnami/common/templates/_affinities.tpl
```diff
@@ -12,7 +12,7 @@ preferredDuringSchedulingIgnoredDuringExecution:
           operator: In
           values:
             {{- range .values }}
-            - {{ . }}
+            - {{ quote . }}
```
juan131, line 15: You can also use a pipe:
```suggestion
            - {{ . | quote }}

```

juan131 (commented):
Thanks so much for this!

I'm okay with your changes, but please address the comments I added.

bitnami/common/templates/_affinities.tpl
wrossmann: Ahh that was my initial impulse, but I switch between Jinja and Go templates so much I never recall which is which. `{{ . | quote }}` looked "too Jinja-ish" and the Helm docs suggested `{{ quote . }}`.

I'll change it around to be consistent with the other charts and bump the appversion.

juan131 (approved):
LGTM

PR #6350, opened 2021-05-13
[bitnami/redis] IPv4/IPv6 dualstack compatibility fixes

### LINKED ISSUE #6329: [bitnami/redis] Nodes do not start with ipv6+ipv4 addresses assigned
**Which chart**:
bitnami/redis, 14.1.1

**Describe the bug**
We use multis to assign ipv6+ipv4 interfaces to pods.  Nodes fail to start, continually logging a message of:
`WARN  ==> redis-test-headless.default.svc.cluster.local does not contain the IP of this pod: 2620:aa:0:fef9:1000:cfcf:1:ed78 172.16.237.138`

It looks like this is due to how the grepping for the node ip is set up in start-node.sh.  hostname -i returns in a format of "ipv6_ip ipv4_ip" in the case that both interfaces are configured.  Whereas getent ahosts only shows the ipv4 interface.

Example:
```
I have no name!@redis-test-node-0:/opt/bitnami/scripts/start-scripts$ hostname -i
2620:aa:0:fef9:1000:cfcf:1:6759 172.16.103.103
I have no name!@redis-test-node-0:/opt/bitnami/scripts/start-scripts$ getent ahosts "$HEADLESS_SERVICE"
172.16.103.103  STREAM redis-test-headless.default.svc.cluster.local
172.16.103.103  DGRAM
172.16.103.103  RAW
172.16.103.118  STREAM
172.16.103.118  DGRAM
172.16.103.118  RAW
172.16.78.80    STREAM
172.16.78.80    DGRAM
172.16.78.80    RAW
```

Hence the grep fails, and it never attempts to actually start redis

**To Reproduce**

1. Set up kubernetes in dual-stack mode, ensure that newly created pods are assigned an ipv4 and ipv6 interface
2. execute `helm install redis-test --set "master.persistence.enabled=false,replica.persistence.enabled=false,usePassword=false,sentinel.enabled=true,sentinel.masterSet=mymaster,cluster.slaveCount=3" --version 14.1.1 bitnami/redis`
3. Pod fails to start, kubectl logs will show warning message

**Expected behavior**
redis and sentinel should both start successfully

- Output of `helm version`:

```
version.BuildInfo{Version:"v3.3.0", GitCommit:"8a4aeec08d67a7b84472007529e8097ec3742105", GitTreeState:"dirty", GoVersion:"go1.14.7"}
```

```
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.8", GitCommit:"fd5d41537aee486160ad9b5356a9d82363273721", GitTreeState:"clean", BuildDate:"2021-02-17T12:41:51Z", GoVersion:"go1.15.8", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.8", GitCommit:"fd5d41537aee486160ad9b5356a9d82363273721", GitTreeState:"clean", BuildDate:"2021-02-17T12:33:08Z", GoVersion:"go1.15.8", Compiler:"gc", Platform:"linux/amd64"}
```

aureus1:
**Description of the change**

Fixed some issues within the bitnami/redis chart that were preventing it from working as-is on a dualstack (IPv4/IPv6) kubernetes installation.  Specifically, that nodes were failing to start due to improper ip detection.  After nodes properly started, sentinel was not communicating with other sentinel nodes until the bind directive was removed.

**Benefits**

bitnami/redis chart will work out of the box in a dualstack k8s installation

**Possible drawbacks**

I'm not aware of any drawbacks

**Applicable issues**

  - fixes #6329 

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/redis/Chart.yaml
+++ b/bitnami/redis/Chart.yaml
@@ -25,4 +25,4 @@ name: redis
 sources:
   - https://github.com/bitnami/bitnami-docker-redis
   - http://redis.io/
-version: 14.1.1
+version: 14.1.2
--- a/bitnami/redis/Chart.yaml
+++ b/bitnami/redis/Chart.yaml
@@ -25,4 +25,4 @@ name: redis
 sources:
   - https://github.com/bitnami/bitnami-docker-redis
   - http://redis.io/
-version: 14.1.2
+version: 14.2.0
--- a/bitnami/redis/templates/scripts-configmap.yaml
+++ b/bitnami/redis/templates/scripts-configmap.yaml
@@ -22,8 +22,7 @@ data:
     myip=$(hostname -i)
     
     #If >1 IP, use the first IPv4 address
-    if [[ "$myip" =~ ( ) ]]
-    then
+    if [[ "$myip" = *" "* ]]; then
         myip=$(echo $myip | awk '{if ( match($0,/([0-9]+\.){3}[0-9]+/) ) { print substr($0,RSTART,RLENGTH); } }')
     fi
 
@@ -149,8 +148,7 @@ data:
     myip=$(hostname -i)
     
     #If >1 IP, use the first IPv4 address
-    if [[ "$myip" =~ ( ) ]]
-    then
+    if [[ "$myip" = *" "* ]]; then
         myip=$(echo $myip | awk '{if ( match($0,/([0-9]+\.)([0-9]+\.)([0-9]+\.)[0-9]+/) ) { print substr($0,RSTART,RLENGTH); } }')
     fi
 
--- a/bitnami/redis/templates/scripts-configmap.yaml
+++ b/bitnami/redis/templates/scripts-configmap.yaml
@@ -18,10 +18,15 @@ data:
     . /opt/bitnami/scripts/libos.sh
     . /opt/bitnami/scripts/liblog.sh
     . /opt/bitnami/scripts/libvalidations.sh
+    myip=$(hostname -i)
+    
+    #If two IPs, use 2nd one (IPv4 in dualstack environment)
+    if [[ "$myip" =~ ( ) ]]
+    then
+        myip=$(echo $myip | cut -d ' ' -f 2)
+    fi
 
     not_exists_dns_entry() {
-        myip=$(hostname -i)
-
         if [[ -z "$(getent ahosts "$HEADLESS_SERVICE" | grep "^${myip}" )" ]]; then
             warn "$HEADLESS_SERVICE does not contain the IP of this pod: ${myip}"
             return 1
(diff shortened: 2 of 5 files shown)


bitnami/redis/Chart.yaml
```diff
@@ -25,4 +25,4 @@ name: redis
 sources:
   - https://github.com/bitnami/bitnami-docker-redis
   - http://redis.io/
-version: 14.1.1
+version: 14.1.2
```
marcosbc, line 28: Due to the amount of changes, I would suggest to increase the minor revision

bitnami/redis/templates/scripts-configmap.yaml
```diff
@@ -140,6 +145,14 @@ data:
     . /opt/bitnami/scripts/libvalidations.sh
     . /opt/bitnami/scripts/libfile.sh
 
+    myip=$(hostname -i)
+    
+    #If two IPs, use 2nd one (IPv4 in dualstack environment)
+    if [[ "$myip" =~ ( ) ]]
```
marcosbc, line 151: It is a bit hard to understand this specific line, I would suggest to rewrite it like:

```bash
if [[ "$myip" = *" "* ]]; then
```

bitnami/redis/templates/scripts-configmap.yaml
```diff
@@ -140,6 +145,14 @@ data:
     . /opt/bitnami/scripts/libvalidations.sh
     . /opt/bitnami/scripts/libfile.sh
 
+    myip=$(hostname -i)
+    
+    #If two IPs, use 2nd one (IPv4 in dualstack environment)
```
marcosbc, line 150: Is it 100% certain for the IPv4 address to be the second address in dualstack environments?

marcosbc (changes requested):
Thanks for the contribution! Please have a look at my comments whenever you have a chance.

bitnami/redis/templates/scripts-configmap.yaml
aureus1: As far as I can there's no clear answer to that. "hostname -i" just iterates through h_addr_list in gethostbyname(), and that's documented as "An array of pointers to network addresses for the host (in network byte order), terminated by a null pointer."  (see: https://man7.org/linux/man-pages/man3/gethostbyname.3.html )

Since this is going to dictate how the other comments go, what's the preferred approach here?  As I see it we could either:
- Just assume that the 2nd address is going to be ipv4, in which case we just want to make the cleanup changes
or
- Switch to something a little more heavyweight like an awk script or a bash regex to parse the input line for something that looks like an ipv4 regex?

aureus1: Switched to awk to detect ipv4.  Had to use a slightly different syntax for the sentinel regex since that container is using mawk instead of gnu awk like the redis container does.  mawk doesn't support the {3} syntax in regexes

bitnami/redis/templates/scripts-configmap.yaml
aureus1: done

bitnami/redis/Chart.yaml
aureus1: done

miguelaeh (commented):
Hi @aureus1,
I left a couple of comments, please address them to make the code comments more readable

carrodher (approved):
Thanks for your changes. LGTM, I am launching some tests in our internal test & release pipeline; if everything works as expected, the PR should be automatically merged

PR #6354, opened 2021-05-13
bitnami/rabbitmq Race fix

mikhail-yarosh:
**Description of the change**
Fixing race condition between plugins loading and cluster rebalance

**Benefits**
Obvious...

**Possible drawbacks**
No the best solution

**Additional information**
Faced with issue of postStart: the rabbitmq-queues rebalance "all" exited with code 69 "Discarding message in an old incarnation (2) of this node...".

**Checklist** 


alemorcuq:
Hi, @mikhail-yarosh. Thanks for the PR! Could you explain a bit more why this is needed? Is it possible that `rabbitmqctl cluster_status` is not the right command to check?

PR #6405, opened 2021-05-18
[bitnami/keycloak] Keycloak updated notes.txt

ARUNMOHANRAJ471:
Updated kubectl command to obtain admin password. The name of the secrets are common.names.fullname and env-vars need not be appended.

**Description of the change**

Updated the command in notes.txt to fetch admin password from secrets. The name of the secret is appended with env-vars which is not being created and the name of the secret is as same as the release name.

**Benefits**

Better readability and ease of use

**Checklist** 

### DIFF
--- a/bitnami/keycloak/templates/NOTES.txt
+++ b/bitnami/keycloak/templates/NOTES.txt
@@ -49,7 +49,7 @@ To access Keycloak from outside the cluster execute the following commands:
 
   echo Username: {{ .Values.auth.adminUser }}
 {{- if not .Values.auth.existingSecretPerPassword }}  
-  echo Password: $(kubectl get secret --namespace {{ .Release.Namespace }} {{ printf "%s" (include "common.names.fullname" .) }} -o jsonpath="{.data.admin-password}" | base64 --decode)
+  echo Password: $(kubectl get secret --namespace {{ .Release.Namespace }} {{ include "common.secrets.name" (dict "existingSecret" .Values.auth.existingSecret "context" $) }} -o jsonpath="{.data.admin-password}" | base64 --decode)
 {{- else }}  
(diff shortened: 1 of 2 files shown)


bitnami/keycloak/templates/NOTES.txt
```diff
@@ -49,7 +49,7 @@ To access Keycloak from outside the cluster execute the following commands:
 
   echo Username: {{ .Values.auth.adminUser }}
 {{- if not .Values.auth.existingSecretPerPassword }}  
-  echo Password: $(kubectl get secret --namespace {{ .Release.Namespace }} {{ printf "%s-env-vars" (include "common.names.fullname" .) }} -o jsonpath="{.data.admin-password}" | base64 --decode)
```
migruiz4, line 52: Thanks for catching this issue, I think the entire `printf` is a typo.

The correct secretName would be obtained by using:
```
{{ include "common.secrets.name" (dict "existingSecret" .Values.auth.existingSecret "context" $) }}
```

https://github.com/bitnami/charts/blob/c0544b4be0281932d26d1a201e8ddf1d6fa405ea/bitnami/keycloak/templates/secrets.yaml#L1-L7

migruiz4 (changes requested):

migruiz4:
Hi @ARUNMOHANRAJ471,

Could you please have a look at my comment below?

Additionally, it will be necessary to upgrade the chart version inside the Keycloack Chart.yaml with a new patch version (3.0.1)

ARUNMOHANRAJ471:
Hi @migruiz4 ,

Thanks for your feedback. I have updated the chart version(3.0.1) and the notes as per your suggestion.

migruiz4 (approved):
LGTM!

I updated the version because the tests reported an additional version bump was required

PR #6426, opened 2021-05-20
[bitnami/metrics-server] Unfortunately the apiService creation is still required up to at least k8s version 1.21

tibeer:
Signed-off-by: Tim Beermann <beermann@betacloud-solutions.de>

**Description of the change**

Changed the default value for "apiServer.create" from "false" to "true".

**Benefits**

The creation of an API server is still required.
The documentation stated, that this should not be required
any more since k8s version 1.8, which does not seem to be
the case. Tested on a kubeadm vanilla installation version 1.21.

**Possible drawbacks**

Users using this chart might encounter an unexpected creation of an API
service when they upgrade the chart.

**Applicable issues**

N/A

**Additional information**

Metrics API is still beta ...

**Checklist** 


Mauraza (changes requested):
Hi @tibeer,

Thanks for the PR! The change in the comments looks good, I think we should keep the default behavior. 
Could you please upload the chart version?

tibeer:
Hi @Mauraza,

this is fine for me, if you want to stick with the current default value. I used chart version 7.6.6.
Maybe we should still improve the documentation here?

Mauraza:
Hi @tibeer,

if we don't do that the change never appears in version  7.6.6. so to make the change, even if it is very small, it is necessary to upload the version. 
Could you update the version to 7.6.7?

tibeer:
@Mauraza 

My bad. I also mentioned the wrong version. 7.6.6. is from another chart. The correct version here was  5.8.7. I bumped it to 5.8.8 and reverted the default value. The documentation change is still there though.

Mauraza (approved):
@tibeer Thanks for the PR!! LGTM!!

PR #6431, opened 2021-05-20
[bitnami/elasticsearch] Make `sysctlImage` parameters configurable

matthew-walker-prolucid:
**Description of the change**

Currently a privileged `sysctlImage` is used to set some parameters in the host kernel.  These parameter's values are not currently configurable.  This change allows them to be configured. 

**Benefits**

Those wishing to use a larger `vm.max_map_count` or `fs.file_max` value will be able to do so without forking the chart.

**Possible drawbacks**

None known.

**Applicable issues**

None.

**Additional information**

There should be no change to behavior for existing deployments.

**Checklist** 


bitnami/elasticsearch/Chart.yaml
```diff
@@ -25,4 +25,4 @@ name: elasticsearch
 sources:
   - https://github.com/bitnami/bitnami-docker-elasticsearch
   - https://www.elastic.co/products/elasticsearch
-version: 15.1.6
+version: 15.1.7
```
juan131, line 28: Please bump minor version instead since you're adding new parameters:
```suggestion
version: 15.2.0
```

bitnami/elasticsearch/templates/data-statefulset.yaml
```diff
@@ -71,7 +71,7 @@ spec:
             - /bin/bash
             - -ec
             - |
-              sysctl -w vm.max_map_count=262144 && sysctl -w fs.file-max=65536
+              sysctl -w vm.max_map_count={{ .Values.sysctlImage.vmMaxMapCount }} && sysctl -w fs.file-max={{ .Values.sysctlImage.fsFileMax }}
```
juan131, line 74: Would it make sense to force integer conversion?
```suggestion
              sysctl -w vm.max_map_count={{ int .Values.sysctlImage.vmMaxMapCount }} && sysctl -w fs.file-max={{ int .Values.sysctlImage.fsFileMax }}
```

juan131 (changes requested):
Thanks so much for this contribution @matthew-walker-prolucid !!! Please take a look to my suggestions when you have a chance.

juan131:
@matthew-walker-prolucid just a friendly reminder to check my suggestions when you have a chance

Issue #6434, opened 2021-05-20
support external access to MongoDB primary pod on a replica-set


korenlev:
**Which chart**:
The name (and version) of the affected chart.

**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

miguelaeh:
Hi @korenlev ,
THank you for the PR we are already taking care of it

korenlev:
Hi miguel, u mean the chart already allow service access to the primary ?
How ?

בתאריך יום ו׳, 21 במאי 2021, 11:03, מאת Miguel Ángel Cabrera Miñagorri ‏<
***@***.***>:

miguelaeh:
Hi @korenlev ,
I mean a colleague is already reviewing the PR

PR #6435, opened 2021-05-20
[bitnami/mongodb] support https://github.com/combor/k8s-mongo-labeler-sidecar - a very …

korenlev:
…crucial component for allowing WRITE access to the mongodb replica-set

**support https://github.com/combor/k8s-mongo-labeler-sidecar - a very crucial component for allowing WRITE access to the MongoDB replica-set"**

**External access to the replica-set dynamic primary pod**

**No drawbacks, tested**

**Applicable to common need to allow external access to MongoDB primary pod**

  - fixes # https://github.com/bitnami/charts/issues/6434

**Checklist** 


korenlev:
instead of adding a whole new section to mongodb main values.yaml i have suggested a new values.primary-mongodb-labeler-sidecar.yaml that includes a clear diff (additional on top of the current values.yaml)

alemorcuq:
Hi, @korenlev. I think this is a very specific feature and I feel like it will be better to support an `extraDeploy` to allow users deploy any extra objects they want, like we do in other charts (for example: https://github.com/bitnami/charts/blob/master/bitnami/redis-cluster/values.yaml#L758). What do you think?

korenlev:
Hi alejandro, isn't it common to have a way to connect to the primary as
the only one allowing WRITE to mongodb ?

בתאריך יום ו׳, 21 במאי 2021, 10:11, מאת Alejandro Moreno ‏<
***@***.***>:

korenlev:
Hi alejandro, isn't it common to have a way to connect to the primary as
the only one allowing WRITE to mongodb ?
Currently, on this bitnami chart, any client connecting to it needs to
figure out (periodically, to make sure) who is primary, then if he is
connecting from outaide the cluster, setup his own service (agaib,
periodically when primary pod have moved).
Seems like a great deal to me, making current deployment very hard to use

בתאריך שבת, 22 במאי 2021, 19:45, מאת Koren Lev ***@***.***>:

korenlev:
i believe  you notes: "without using a service that is load balancing
between the pods" - is the whole point, and actually is the missing piece
from the chart.
as there is only 1 PRIMARY POD and only this PRIMARY POD can allow WRITE
access, it needs to have a 'special service' that is 'outside' any
load-balancing , load-balancing is only good for READ from replica-set.
am i wrong ? i can drop it as extraDeploy no worries ...

‫בתאריך יום ג׳, 25 במאי 2021 ב-17:54 מאת ‪Alejandro Moreno‬‏ <‪
***@***.***‬‏>:‬

korenlev:
moving to contribute the extra-deploy[] method suggested ...

korenlev:
done adding as extraDeploy feature - please review

alemorcuq:
> i believe  you notes: "without using a service that is load balancing
> between the pods" - is the whole point, and actually is the missing piece
> from the chart.

As I said earlier, the `externalAccess` feature in the chart creates one service per pod, so you can connect to MongoDB using a string with all the services instead of just one service providing load balancing to the pods. Please, check that feature as I think it solves your problem.

korenlev:
Hi Alejandro,

externalAccess exposes ALL replica pods through services.
This does NOT solve the problem of connecting to the PRIMARY POD of the
replica set, if someone wishes to WRITE only.
"using a string" - maybe you mean using replica URL l- this does NOT solve
the problem either as the load-balancing hits SECONDARY pods once in a
while where only READ is allowed.
This is a big problem to solve in the current chart, i have no idea how one
can use it for WRITE access without this solution.

can you please test and take a closer look?

BR
Koren

‫בתאריך יום ב׳, 31 במאי 2021 ב-17:15 מאת ‪Alejandro Moreno‬‏ <‪
***@***.***‬‏>:‬

juan131:
Hi Koren,

Thanks so much for the contribution!!

As you said, our current approach to expose MongoDB consist on creating a LoadBalancer or NodePort service per MongoDB node (regardless this node is a primary or secondary node). This approach has two downsides:

- Logic must be implemented in the client side to dynamically detect which is the primary node, so "write" actions aren't sent to the secondary nodes. 
- It's very expensive (when using LoadBalancer svc type) since you need to request to the CloudProvider several external IPs which will increase your operational cost.

Ideally, we should have some kind of intelligent proxy between the client and MongoDB nodes with the ability to identify "write" queries and redirect them to the primary node, while load-balancing "read" queries using round-robin. This will be the optimal solution IMO.

The approach the "MongoDB replica set pod labeler" uses is a possible workaround, although not a very "clean" solution IMO. Assigning labels dynamically to the pods and giving RBAC permissions to a sidecar to do so sounds risky to me. That's why I have my concerns regarding officially support it in our chart.

The solution @alemorcuq proposed you about using the **"extraDeploy"** to use the"MongoDB replica set pod labeler" doesn't require so many changes. He simply assumed we already had support this **"extraDeploy"** parameter (which is very common in our charts) but it we hadn't added it to MongoDB yet. Let me illustrate it using an example using the current chart:

- Ensure you have the changes I just submitted at https://github.com/bitnami/charts/pull/6536 (they add support for **"extraDeploy"** parameter)
- Create  a **values.yaml** like the one below:

```yaml
architecture: replicaset
replicaCount: 3
sidecars:
  - name: mongo-labeler
    image: korenlev/k8s-mongo-labeler-sidecar
    imagePullPolicy: Always
    env:
      - name: LABEL_SELECTOR
        value: "app.kubernetes.io/component=mongodb,app.kubernetes.io/instance=mongodb,app.kubernetes.io/name=mongodb"
      - name: NAMESPACE
        value: "default"
      - name: DEBUG
        value: "true"
extraDeploy:
  - apiVersion: v1
    kind: Service
    metadata:
      name: mongodb-primary
      namespace: default
      labels:
        app.kubernetes.io/component: mongodb
        app.kubernetes.io/instance: mongodb
        app.kubernetes.io/name: mongodb
    spec:
      type: LoadBalancer
      externalTrafficPolicy: Cluster
      ports:
        - name: mongodb
          port: 27017
          protocol: TCP
          targetPort: mongodb
      selector:
        app.kubernetes.io/component: mongodb
        app.kubernetes.io/instance: mongodb
        app.kubernetes.io/name: mongodb
        primary: "true"
```

- Then, install the MongoDB chart in the "default" namespace:

```console
$ helm install mongodb bitnami/mongodb -f values.yaml
```

korenlev:
thanks Juan , got it all and i'll be working on it

korenlev:
@juan131 i have merged your changes and added the extraDeploy and needed section , can you verify please

korenlev:
@juan131 i have tested this with your changes (only uncommenting sections added)  in production - results looks fine, primary pod exposed as designed

bitnami/mongodb/README.md
```diff
@@ -531,3 +531,34 @@ Use the workaround below to upgrade from versions previous to 5.0.0. The followi
 ```console
 $ kubectl delete statefulset my-release-mongodb-arbiter my-release-mongodb-primary my-release-mongodb-secondary --cascade=false
 ```
+
+### Add extra deployment options
+
+To add extra deployments (useful for advanced features like sidecars), use the `extraDeploy` property.
+# example here for a 'mongo-labeler' sidecar to find primary pod in replica set and listen on predictable port
```
juan131, line 538: Please note that, by adding this `# ...` paragraph, you're include a H1 header:
```suggestion

In the example below, you can find how to use a example here for a [MongoDB replica set pod labeler sidecar](https://github.com/combor/k8s-mongo-labeler-sidecar) to identify the primary pod and dynamically label it as the primary node:
```

bitnami/mongodb/README.md
```diff
@@ -531,3 +531,34 @@ Use the workaround below to upgrade from versions previous to 5.0.0. The followi
...
+        - name: mongodb-primary
+          port: 30001
+          nodePort: 30001
+          protocol: TCP
+          targetPort: mongodb
+      selector:
+        app.kubernetes.io/component: mongodb
+        app.kubernetes.io/instance: mongodb
+        app.kubernetes.io/name: mongodb
+        primary: "true"
```
juan131, line 564: Close the "yaml" section block with three "`"

bitnami/mongodb/values.yaml
```diff
@@ -443,6 +472,24 @@ initContainers: {}
...
+#- name: mongo-labeler
+#  image: korenlev/k8s-mongo-labeler-sidecar
+#  imagePullPolicy: Always
+#  env:
+#    - name: LABEL_SELECTOR
+#      value: "app.kubernetes.io/component=mongodb,app.kubernetes.io/instance=mongodb,app.kubernetes.io/name=mongodb"
+#    - name: NAMESPACE
+#      value: "the-mongodb-namespace"
+#    - name: DEBUG
+#      value: "true"
```
juan131, line 490: nice!

bitnami/mongodb/values.yaml
```diff
@@ -52,6 +52,35 @@ clusterDomain: cluster.local
...
+#      - name: mongodb
+#        port: 30001
+#        nodePort: 30001
+#        protocol: TCP
+#        targetPort: mongodb
+#    selector:
+#      app.kubernetes.io/component: mongodb
+#      app.kubernetes.io/instance: mongodb
+#      app.kubernetes.io/name: mongodb
+#      primary: "true"
```
juan131, line 81: Nice, thanks!

juan131 (changes requested):
Thanks @korenlev !! You changes LGTM, just a couple of suggestions for the **README.md**

bitnami/mongodb/README.md
juan131: Still pending, right?

juan131:
Please fix the "too many blank lines (1 > 0)  (empty-lines)" errors in the values.yaml that the linter is throwing

bitnami/mongodb/README.md
korenlev: done now , sorry for the delay

juan131 (approved):
LGTM

juan131:
Your changes were landed @korenlev

PR #6449, opened 2021-05-24
 [bitnami/kubeapps] Add missing $ to use the global scope

antgamdia:
**Description of the change**

Currently, adding an `initialRepos` with `authorizationHeader` prevents the chart to be installed. It tries to retrieve `.Values.commonLabels` but, since it's under a `range` it's not in the scope.  

```console
Error: UPGRADE FAILED: template: kubeapps/templates/apprepository/apprepositories-secret.yaml:13:18: executing "kubeapps/templates/apprepository/apprepositories-secret.yaml" at <.Values.commonLabels>: nil pointer evaluating interface {}.commonLabels
```

This PR just adds `$` to use the global scope when necessary.

**Benefits**

Adding an `initialRepos` with `authorizationHeader` will work again:

```yaml
apprepository:
  initialRepos:
    - authorizationHeader: Basic <redacted>
      name: myrepo
      namespace: kubeapps
      ...
```

**Possible drawbacks**

N/A

**Applicable issues**

N/A

**Additional information**

N/A
**Checklist** 

### DIFF
--- a/bitnami/kubeapps/Chart.yaml
+++ b/bitnami/kubeapps/Chart.yaml
@@ -25,4 +25,4 @@ maintainers:
 name: kubeapps
 sources:
   - https://github.com/kubeapps/kubeapps
-version: 7.0.3
+version: 7.0.4
--- a/bitnami/kubeapps/templates/apprepository/apprepositories-secret.yaml
+++ b/bitnami/kubeapps/templates/apprepository/apprepositories-secret.yaml
@@ -10,10 +10,10 @@ metadata:
   namespace: {{ $.Release.Namespace | quote }}
   {{- end }}
   labels: {{- include "common.labels.standard" $ | nindent 4 }}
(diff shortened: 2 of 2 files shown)


alvneiayu:
hi @antgamdia 

did you not have the same problem with the other .Values? I mean, the symbol $ must be in all the .Values in the range, no?

Thanks a lot

Álvaro

antgamdia:
Thanks for having a look at the PR, @alvneiayu 
> did you not have the same problem with the other .Values? I mean, the symbol $ must be in all the .Values in the range, no?

Thanks you pointing this out. I didn't get an error since I wasn't using any commonLabel or commonAnnotations. I've just tried adding some of them and it fails, so I need to add the `$`  as you suggest.

Let me update the PR in a while :)

alvneiayu (approved):
LGTM

PR #6460, opened 2021-05-25
[bitnami/kibana] Allow savedObjects to be properly loaded when rewriteBasePath is specified.

### LINKED ISSUE #6455: kibana savedObjects not processed when server.rewriteBasePath is used
**Which chart**:
kibana  chart 8.1.2

**Describe the bug**

When rewriteBasePath is used the basePath of the API changes.   The saved-objects-configmap.yaml doesn't consider this, so when rewriteBasePath is use, the curl call to load the savedObjects gets a 404 error and fails to load them.

**To Reproduce**

1. Create elastic/helm with rewriteBasePath  & savedObjects config map
```
elasticsearch:
  fullnameOverride: elasticsearch
  global:
    kibanaEnabled: true
    storageClass: "managed-nfs-storage-delete"
  kibana:
    ingress:
      enabled: true
      path: /kibana
      annotations:
        kubernetes.io/ingress.class: *ingressClass
    savedObjects:
      configmap: kibana-config
    configuration:
      server:
        basePath: /kibana
        rewriteBasePath: true
```

2. The kibana log shows a 404 failure loading the savedObjects.

**Expected behavior**

Saved object should be loaded even when rewriteBasePath is specified.

- Output of `helm version`:

```
version.BuildInfo{Version:"v3.5.2", GitCommit:"167aac70832d3a384f65f9745335e9fb40169dc2", GitTreeState:"dirty", GoVersion:"go1.15.7"}

```

```
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.3", GitCommit:"b3cbbae08ec52a7fc73d334838e18d17e8512749", GitTreeState:"clean", BuildDate:"2019-11-13T11:23:11Z", GoVersion:"go1.12.12", Compiler:"gc", Platform:"linux/amd64"}
The connection to the server 127.0.0.1:46716 was refused - did you specify the right host or port?
```

**Additional context**

Line 8 of the saved-objects-confimap.yaml template -- does not adjust the URL when rewriteBasePath  is in effect.

  {{- $savedObjectsUrl := printf "localhost:%d/api/saved_objects/_import" (int .Values.containerPort) }}

This is a follow-on to https://github.com/bitnami/bitnami-docker-kibana/pull/49  -- in which the kibana container has a similar defect.  Once fixing that, I then was able to hit the related behavior in this issue.

jerryharrow:
**Description of the change**

1. Allow savedObjects to be properly loaded when rewriteBasePath is specified.
2. Allow use of an ingress without a known hostname.  It allows the hostname rule to be dropped by use of the "*" value in the chart.   This value is not valid for a hostname input, so it does not introduce any incompatibly  with prior behavior, or ambiguity. 

**Benefits**

Allows savedObjects to be loaded
Allows use of an existing ingress without a known hostname associated.

**Possible drawbacks**

Would have been preferable to not have a default value for ingress.hostname and just omit the hostname rule if not supplied, but  that would be an incompatible change if others are expecting "kibana.local" by default.
Using the value "*" as a sentinel value for this case seems acceptable as that value is not accepted by kubernetes so there is no ambiquity.

**Applicable issues**

  - fixes #6455

**Additional information**

Follow-on to https://github.com/bitnami/bitnami-docker-kibana/pull/49 which fixed the base kibana container to
properly support rewriteBasePath -- this is needed as well to enable the savedObjects support of the kibana chart to similarly work with this attribute.

**Checklist** 


jerryharrow:
fixes #6460

jerryharrow:
Unfortunately 7.12.1-debian-10-r24 doesn't have the the fix yet.  Need to wait for the next tag.

bitnami/kibana/values.yaml
```diff
@@ -14,7 +14,7 @@
 image:
   registry: docker.io
   repository: bitnami/kibana
-  tag: 7.12.1-debian-10-r0
+  tag: 7.12.1-debian-10-r24
```
migruiz4, line 17: Could you please revert this change?

The bitnami-bot will take care of updating the image tag with the latest changes before merging this PR.

migruiz4 (commented):
Hi @jerryharrow,

Your PR looks good to me. As I commented below, it is not necessary to wait until a new Kibana version is released. Simply revert the tag change and the bitnami-bot will ensure the latest tag is used before merging.

Regarding the Ingress rule without `host`, I'd like to add some links with additional information for other users reading this comment.
According to Kubernetes documentation (https://kubernetes.io/docs/reference/kubernetes-api/service-resources/ingress-v1/#IngressSpec):
- When `rule.host` is not specified, traffic will be routed based on `rule.http`.
> If the host is unspecified, the Ingress routes all traffic based on the specified IngressRuleValue.
- `rule.host` does not support using wildcard by itself.
> You cannot have a wildcard label by itself (e.g. Host == "*").

Although, even if Kubernetes support rules without specifying `rule.host`, it is possible that certain Ingress controllers don't support it. For example, the Nginx ingress-controller does not support this feature at the moment:  https://github.com/nginxinc/kubernetes-ingress/issues/209

bitnami/kibana/values.yaml
jerryharrow: Done.   Automation is great!

migruiz4 (approved):
LGTM! Thank you for your contribution

jerryharrow:
> According to Kubernetes documentation (https://kubernetes.io/docs/reference/kubernetes-api/service-resources/ingress-v1/#IngressSpec):

Excellent reference.   For some reason I only found the ingress information in the Kubernetes concepts section.    Surprisingly those don't link to the reference sections.   Now I know where to look to get better details.   Thanks!

> For example, the Nginx ingress-controller does not support this feature at the moment: nginxinc/kubernetes-ingress#209

Fascinating, since we are using Nginx.  We don't have the  nginx ingress controller resource enabled, so that is apparently key.
Thanks for that pointer too -- since no doubt we will some day hit it :-)

migruiz4:
I'm glad you found the information useful @jerryharrow! 

Again, thank you very much for your contributions!

PR #6466, opened 2021-05-26
[bitnami/external-dns] fix: allow to pass OwnerId for non-txt registry type in external-dns chart

yk12OT:
**Description of the change**
To be able to pass `OwnerId` to non-txt registry, such as `aws-sd`.

**Benefits**
It will allow to see the real OwnerId for DNS records rather than `default`.

**Possible drawbacks**
Had to introduce a new boolean variable `forceTxtOwnerId` to make it backward compatible for customers 
who were using `aws-sd` registry type.  Previous behavior of bitnami/external-dns helm chart was 
to ignore `txtOwnerId` value for non-TXT registry types.

**Additional information**
Here is a fragment from `helm template` that shows that OwnerId is not passed down to the pod for registry type `aws-sd`:
```
      serviceAccountName: external-dns-sd
      containers:
        - name: external-dns
          image: "docker.io/bitnami/external-dns:0.8.0-debian-10-r0"
          args:
            # Generic arguments
            - --log-level=info
            - --log-format=text
            - --domain-filter=domain1
            - --domain-filter=domain2
            - --policy=sync
            - --provider=aws-sd
            - --registry=aws-sd
            - --interval=3m
            - --source=service
            - --source=ingress
            # AWS arguments
            - --aws-api-retries=3
            - --aws-batch-change-size=1000
          env:
            # AWS environment variables
            - name: AWS_DEFAULT_REGION
              value: us-west-2
```
The source code for external DNS controller shows that the same variable `cfg.TXTOwnerID` is used for `aws-sd` registry type as for `txt` type:
https://github.com/kubernetes-sigs/external-dns/blob/a5baad26d7f4a7e1ff9ff2fefa6d566fe726a1cd/main.go#L327-L328
```
	var r registry.Registry
	switch cfg.Registry {
	case "noop":
		r, err = registry.NewNoopRegistry(p)
	case "txt":
		r, err = registry.NewTXTRegistry(p, cfg.TXTPrefix, cfg.TXTSuffix, cfg.TXTOwnerID, cfg.TXTCacheInterval, cfg.TXTWildcardReplacement)
	case "aws-sd":
		r, err = registry.NewAWSSDRegistry(p.(*awssd.AWSSDProvider), cfg.TXTOwnerID)
	default:
		log.Fatalf("unknown registry: %s", cfg.Registry)
	}
```

**Checklist** 


rafariossaa:
Hi,
  Thanks for sending this PR.
  In that case, Wouldn't it be better to check if `TXTOwnerID` is `txt` or `aws-sd` ? Because if not, it would be added always includuing for `noop` or other future types.

yk12OT:
hi @rafariossaa, just FYI - I just updated my PR to make it backward compatible with existing `aws-sd` customers, please take another look.

Answering your question about `noop or other future types` - yes, that is exactly the intent - to pass that value to all registry types. Noop would simply ignore it - nothing wrong with that (or we can hard-code `noop` as exception - that is fine too).

The reason to pass the Owner's ID is simple - that is how the backend service (external DSN controller app) figures out if it "owns" that DNS record or not. If yes, it will be able to update/delete when necessary. If "no" then it will ignore DNS record.
It is essential for application to know the Owner's ID regardless of the registry type.

rafariossaa (approved):
Thank you.
This LGTM.

PR #6467, opened 2021-05-26
[bitnami/discourse] adding extraPorts gives invalid YAML error

JustinAvaLabs:
**Description of the change**
adding extraPorts gives error:
YAML parse error on discourse/templates/service.yaml: error converting YAML to JSON: yaml: line 21: mapping values are not allowed in this context

This change will convert input to yaml without going through the weird bitnami common.tplvalues template.

**Benefits**
Will let me add an https port with http redirect to the bitnami discourse helm chart loadbalancer service

**Possible drawbacks**
I might be wrong

**Applicable issues**
none found

**Additional information**

**Checklist** 


bitnami/discourse/templates/service.yaml
```diff
@@ -34,6 +34,6 @@ spec:
       nodePort: null
       {{- end }}
     {{- if .Values.service.extraPorts }}
-    {{- include "common.tplvalues.render" (dict "value" .Values.service.extraPorts "context" $) | indent 4 }}
+    {{- toYaml .Values.ingress.extraPorts | nindent 4 }}
```
miguelaeh, line 37: `.Values.ingress.extraPorts` does not exist in the values.yaml as a parameter

miguelaeh (changes requested):
Hi @JustinAvaLabs ,
The bitnami common functions are our standard to do thing easier when maintaining the charts so we should not remove them.
Also, please complete the checklist on the PR template in order to ensure you followed all the steps like bumping the chart version.

bitnami/discourse/templates/service.yaml
miguelaeh, line 37:
```suggestion
   {{- include "common.tplvalues.render" (dict "value" .Values.service.extraPorts "context" $) | nindent 4 }}
```

This change should be enought

JustinAvaLabs: Forgot the switch "ingress" with "service", but the issue still stands. 

I cannot get the service.extraPorts field to template properly using the bitnami.common.tplvalues function.  

I found   "{{- toYaml .Values.ingress.extraPorts | nindent 4 }}" in the same "bitnami/discourse" chart, in the ingress.yml file for a similar field (ingress.extraTLS).  This field actually works though, so I believe switching this change would fix it.

JustinAvaLabs: Is this solution just to switch indent with nindent?

JustinAvaLabs: UPdated with bumped chart version

miguelaeh: Yes changing indent by nindent  works

miguelaeh: Check my suggestion below, I checked it and seems to work just changing indent by nindent

miguelaeh (commented):
Hi @JustinAvaLabs ,
I replied to your comments, the suggestion above should fix the issue, it is matter of changing indent by nindent

JustinAvaLabs:
@miguelaeh  Thanks for the update/review, I have committed your changes.

miguelaeh (approved):
LGTM!

PR #6478, opened 2021-05-27
[bitnami/zookeeper] Allow the use of the admin server to provide live…

enixon:
…ness and readiness checks

**Description of the change**

Adds an option to the ZooKeeper chart that performs the kubernetes liveness and readiness probes using the ZooKeeper admin server (https://zookeeper.apache.org/doc/current/zookeeperAdmin.html#sc_adminserver). Previously the only options were to disable the probes or to direct them to the generic ZooKeeper client port.

**Benefits**

Unblocks responsible setting of INFO and WARN log levels (current default is ERROR).

Every iteration of the current liveness and readiness checks adds two log lines each (INFO level message on connect, WARN level message on disconnect) because they communicate with the ZooKeeper server on the port that all client traffic uses. The ZooKeeper admin server responds on a different port that avoids those log messages and provides the same guarantees about the state of the service with its response.

Log spam is also generated by the pattern of polling required for ZooKeeper 3.4 but with the addition of the admin server in 3.5 and metrics providers in 3.6 (https://zookeeper.apache.org/doc/current/zookeeperAdmin.html#sc_adminserver_config), this too can be avoided.

**Possible drawbacks**

This adds complication to the config by introducing a new port. The owner of the service can choose to turn off the admin server (jvm flag "zookeeper.admin.enableServer") or change the address that it uses (jvm flag "zookeeper.admin.serverAddress") or shift the port (jvm flag "zookeeper.admin.serverPort") and the chart would continue to try the contact that server on the default host:port. Making the chart responsive to these additional ZooKeeper settings requires a modification of the Docker image.

**Applicable issues**

None

**Additional information**

The “four letter word” feature that currently backs the liveness and readiness probes has been deprecated by Apache ZooKeeper and (https://zookeeper.apache.org/doc/current/zookeeperAdmin.html#sc_4lw).

**Checklist** 

### DIFF
--- a/bitnami/zookeeper/README.md
+++ b/bitnami/zookeeper/README.md
@@ -251,7 +251,9 @@ Bitnami will release a new chart updating its containers if a new version of the
 
 ### Configure log level
 
-You can configure the ZooKeeper log level using the `ZOO_LOG_LEVEL` environment variable. By default, it is set to `ERROR` because of each readiness probe produce an `INFO` message on connection and a `WARN` message on disconnection.
+You can configure the ZooKeeper log level using the `ZOO_LOG_LEVEL` environment variable or the parameter `logLevel`. By default, it is set to `ERROR` because each instance of the liveness probe and the readiness probe produces an `INFO` message on connection and a `WARN` message on disconnection.
+
+To avoid the connection/disconnection messages from the probes, you can enable the use of the ZooKeeper Admin Server for these checks by setting both `livenessProbe.useAdmin` and `readinessProbe.useAdmin` to `true`. If you are polling for ZooKeeper metrics then you must also ensure that you are using the ZooKeeper metrics server (`metrics.enabled` set to `true`) and not the deprecated pattern of polling 'mntr' on the ZooKeeper client port. If these precautions are taken, it is recommended that you set log level to ‘INFO’.
 
 ## Persistence
 
--- a/bitnami/zookeeper/README.md
+++ b/bitnami/zookeeper/README.md
@@ -251,9 +251,32 @@ Bitnami will release a new chart updating its containers if a new version of the
 
 ### Configure log level
 
-You can configure the ZooKeeper log level using the `ZOO_LOG_LEVEL` environment variable or the parameter `logLevel`. By default, it is set to `ERROR` because each instance of the liveness probe and the readiness probe produces an `INFO` message on connection and a `WARN` message on disconnection.
+You can configure the ZooKeeper log level using the `ZOO_LOG_LEVEL` environment variable or the parameter `logLevel`. By default, it is set to `ERROR` because each use of the liveness probe and the readiness probe produces an `INFO` message on connection and a `WARN` message on disconnection, generating a high volume of noise in your logs.
 
-To avoid the connection/disconnection messages from the probes, you can enable the use of the ZooKeeper Admin Server for these checks by setting both `livenessProbe.useAdmin` and `readinessProbe.useAdmin` to `true`. If you are polling for ZooKeeper metrics then you must also ensure that you are using the ZooKeeper metrics server (`metrics.enabled` set to `true`) and not the deprecated pattern of polling 'mntr' on the ZooKeeper client port. If these precautions are taken, it is recommended that you set log level to ‘INFO’.
+In order to remove that log noise so levels can be set to ‘INFO’, two changes must be made.
+
+First, ensure that you are not getting metrics via the deprecated pattern of polling 'mntr' on the ZooKeeper client port. The preferred method of polling for Apache ZooKeeper metrics is the ZooKeeper metrics server. This is supported in this chart when setting `metrics.enabled` to `true`.
+
+Second, to avoid the connection/disconnection messages from the probes, you can set custom values for these checks which directly them to the ZooKeeper Admin Server instead of the client port. An example is given below which preserves default probe behavior of this chart. By default, an Admin Server will be started that listens on localhost at 8080 and if you have specified either of `zookeeper.admin.serverAddress` or `zookeeper.admin.serverPort` then change the values of the probe to match. The Admin Server is on by default and you must not have disabled it (via `zookeeper.admin.enableServer`) if you wish to use it for the probes.
+
+```
+customLivenessProbe:
+  exec:
+    command: ['/bin/bash', '-c', 'curl -s -m 2 http://localhost:8080/commands/ruok | grep ruok']
+  initialDelaySeconds: 30
+  periodSeconds: 10
+  timeoutSeconds: 5
+  successThreshold: 1
+  failureThreshold: 6
+customReadinessProbe:
+  exec:
+    command: ['/bin/bash', '-c', 'curl -s -m 2 http://localhost:8080/commands/ruok | grep error | grep null']
+  initialDelaySeconds: 5
+  periodSeconds: 10
+  timeoutSeconds: 5
+  successThreshold: 1
+  failureThreshold: 6
+```
 
 ## Persistence
 
--- a/bitnami/zookeeper/README.md
+++ b/bitnami/zookeeper/README.md
@@ -257,9 +257,13 @@ In order to remove that log noise so levels can be set to ‘INFO’, two change
 
 First, ensure that you are not getting metrics via the deprecated pattern of polling 'mntr' on the ZooKeeper client port. The preferred method of polling for Apache ZooKeeper metrics is the ZooKeeper metrics server. This is supported in this chart when setting `metrics.enabled` to `true`.
 
-Second, to avoid the connection/disconnection messages from the probes, you can set custom values for these checks which directly them to the ZooKeeper Admin Server instead of the client port. An example is given below which preserves default probe behavior of this chart. By default, an Admin Server will be started that listens on localhost at 8080 and if you have specified either of `zookeeper.admin.serverAddress` or `zookeeper.admin.serverPort` then change the values of the probe to match. The Admin Server is on by default and you must not have disabled it (via `zookeeper.admin.enableServer`) if you wish to use it for the probes.
+Second, to avoid the connection/disconnection messages from the probes, you can set custom values for these checks which direct them to the ZooKeeper Admin Server instead of the client port. By default, an Admin Server will be started that listens on `localhost` at port `8080`. The following is an example of this use of the Admin Server for probes:
 
 ```
+livenessProbe:
+  enabled: false
+readinessProbe:
+  enabled: false
 customLivenessProbe:
   exec:
     command: ['/bin/bash', '-c', 'curl -s -m 2 http://localhost:8080/commands/ruok | grep ruok']
--- a/bitnami/zookeeper/templates/statefulset.yaml
+++ b/bitnami/zookeeper/templates/statefulset.yaml
@@ -404,6 +404,9 @@ spec:
           {{- if .Values.livenessProbe.enabled }}
           livenessProbe:
             exec:
+              {{- if .Values.livenessProbe.useAdmin }}
+              command: ['/bin/bash', '-c', 'curl -s -m {{ .Values.livenessProbe.probeCommandTimeout }} http://localhost:8080/commands/ruok | grep ruok']
+              {{- else }}
               {{- if not .Values.service.disableBaseClientPort }}
               command: ['/bin/bash', '-c', 'echo "ruok" | timeout {{ .Values.livenessProbe.probeCommandTimeout }} nc -w {{ .Values.livenessProbe.probeCommandTimeout }} localhost {{ .Values.service.port }} | grep imok']
               {{- else }}
@@ -412,6 +415,7 @@ spec:
                 {{- else }}
                 command: ['/bin/bash', '-c', 'echo "ruok" | timeout {{ .Values.livenessProbe.probeCommandTimeout }} openssl s_client -quiet -crlf -connect localhost:{{ .Values.service.tlsClientPort }} -cert {{ .Values.service.tls.client_cert_pem_path }} -key {{ .Values.service.tls.client_key_pem_path }} | grep imok']
                 {{- end }}
+              {{- end }}
               {{- end }}
             initialDelaySeconds: {{ .Values.livenessProbe.initialDelaySeconds }}
             periodSeconds: {{ .Values.livenessProbe.periodSeconds }}
@@ -424,6 +428,9 @@ spec:
           {{- if .Values.readinessProbe.enabled }}
           readinessProbe:
             exec:
+              {{- if .Values.readinessProbe.useAdmin }}
+              command: ['/bin/bash', '-c', 'curl -s -m {{ .Values.readinessProbe.probeCommandTimeout }} http://localhost:8080/commands/ruok | grep error | grep null']
+              {{- else }}
               {{- if not .Values.service.disableBaseClientPort }}
               command: ['/bin/bash', '-c', 'echo "ruok" | timeout {{ .Values.readinessProbe.probeCommandTimeout }} nc -w {{ .Values.readinessProbe.probeCommandTimeout }} localhost {{ .Values.service.port }} | grep imok']
               {{- else }}
@@ -432,6 +439,7 @@ spec:
                 {{- else }}
                 command: ['/bin/bash', '-c', 'echo "ruok" | timeout {{ .Values.livenessProbe.probeCommandTimeout }} openssl s_client -quiet -crlf -connect localhost:{{ .Values.service.tlsClientPort }} -cert {{ .Values.service.tls.client_cert_pem_path }} -key {{ .Values.service.tls.client_key_pem_path }} | grep imok']
                 {{- end }}
+              {{- end }}
               {{- end }}
             initialDelaySeconds: {{ .Values.readinessProbe.initialDelaySeconds }}
             periodSeconds: {{ .Values.readinessProbe.periodSeconds }}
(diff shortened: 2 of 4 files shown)


bitnami/zookeeper/values.yaml
```diff
@@ -398,6 +398,7 @@ livenessProbe:
   failureThreshold: 6
   successThreshold: 1
   probeCommandTimeout: 2
+  useAdmin: false
```
enixon, line 401: I'm indifferent to the value name 'useAdmin' and am happy to change as needed.

While I did add two separate 'useAdmin' toggles in this commit, it was to try to match the existing convention and I remain convinced that a single toggle would be safer (so liveness and readiness are measured in the same fashion).

bitnami/zookeeper/templates/statefulset.yaml
```diff
@@ -268,13 +272,17 @@ spec:
           {{- if .Values.readinessProbe.enabled }}
           readinessProbe:
             exec:
-              {{- if not .Values.service.tls.disable_base_client_port }}
-              command: ['/bin/bash', '-c', 'echo "ruok" | timeout {{ .Values.readinessProbe.probeCommandTimeout }} nc -w {{ .Values.readinessProbe.probeCommandTimeout }} localhost {{ .Values.service.port }} | grep imok']
+              {{- if .Values.readinessProbe.useAdmin }}
+              command: ['/bin/bash', '-c', 'curl -s -m {{ .Values.readinessProbe.probeCommandTimeout }} http://localhost:8080/commands/ruok | grep error | grep null']
```
enixon, line 276: Expected output from this check when the server can host client traffic:
```
$ curl -s http://localhost:8080/commands/ruok
{
  "command" : "ruok",
  "error" : null
}
```

Output from this check with the server cannot host client traffic but the jvm is still running:
```
$ curl -s http://localhost:8080/commands/ruok
{
  "command" : "ruok",
  "error" : "This ZooKeeper instance is not currently serving requests"
}
```

bitnami/zookeeper/templates/statefulset.yaml
```diff
@@ -250,13 +250,17 @@ spec:
           {{- if .Values.livenessProbe.enabled }}
           livenessProbe:
             exec:
-              {{- if not .Values.service.tls.disable_base_client_port }}
-              command: ['/bin/bash', '-c', 'echo "ruok" | timeout {{ .Values.livenessProbe.probeCommandTimeout }} nc -w {{ .Values.livenessProbe.probeCommandTimeout }} localhost {{ .Values.service.port }} | grep imok']
+              {{- if .Values.livenessProbe.useAdmin }}
+              command: ['/bin/bash', '-c', 'curl -s -m {{ .Values.livenessProbe.probeCommandTimeout }} http://localhost:8080/commands/ruok']
```
enixon, line 254: This changes the behavior of the liveness command. Previously liveness and readiness were measured the same - whether the server responds to client traffic. There are times in the normal lifecycle of a ZooKeeper ensemble where a healthy server will not be responsive to clients (e.g. during a leader election). The new form of the check will ensure a positive liveness result during such times and a negative readiness result.

enixon (commented):
The simplest way to test the change is to start an ensemble with override `--set logLevel=INFO` and observe the log spam that results.

Then start the ensemble with overrides `--set logLevel=INFO --set livenessProbe.useAdmin=true,readinessProbe.useAdmin=true` and observe the lack of log spam.

This change has been tested with `replicaCount` values of 1 and 3.

alemorcuq (changes requested):
Thanks for the PR, @enixon! It would be very nice to allow for some extra configuration in the liveness and readiness probes! In other charts we are adding the `customLivenessProbe` and `customReadinessProbe` parameters to allow a complete customization of probes. Can you implement the same here?

Here's an example in the Kafka chart: https://github.com/bitnami/charts/blob/02fc6fabf0cfd0dba0ace6f4430c9d8c5d424209/bitnami/kafka/templates/statefulset.yaml#L340

enixon:
@alemorcuq not a problem! Custom probe support added. Testing can be facilitated using the following values.

```
replicaCount: 3

livenessProbe:
  enabled: false
  useAdmin: true
readinessProbe:
  enabled: false
  useAdmin: true

customLivenessProbe:
  exec:
    command: ['/bin/bash', '-c', 'curl -s -m 2 http://localhost:8080/commands/ruok | grep error']
  initialDelaySeconds: 30
  periodSeconds: 10
  timeoutSeconds: 5
  successThreshold: 1
  failureThreshold: 6
customReadinessProbe:
  exec:
    command: ['/bin/bash', '-c', 'curl -s -m 2 http://localhost:8080/commands/ruok | grep error']
  initialDelaySeconds: 5
  periodSeconds: 10
  timeoutSeconds: 5
  successThreshold: 1
  failureThreshold: 6
```

alemorcuq (changes requested):
Thank you! And sorry for the misunderstanding, I meant to add the custom probe support instead of the `useAdmin` approach to keep things more in line with the rest of charts.

enixon:
@alemorcuq , having the custom probe implementation upstreamed would suffice for my needs so I’m happy to create a PR for just that change.

Before I do, I would like to at least speak for my initial change. Many users of the chart will not have the knowledge of ZooKeeper internals required in order to put together a solution that unblocks logging at the lower levels - there’s value in having it laid out for them here with easy parameters to set.

Just comment back with which path you’d prefer and I’ll get it done.

alemorcuq:
I think it's important to maintain a standard way of configuring things across our many charts and I don't think that `useAdmin` feature you mention fits there. The custom probes parameters were added so that we can have a nice default while also providing users all the configuration potential they need.

However, I do think you are making a valid point regarding the probe itself. It may be possible that the `useAdmin` probe you are suggesting is a better default probe than the one currently implemented in the chart. That would be something different to discuss in another issue.

enixon:
@alemorcuq , I've moved the custom probe support over to #6547. 

This one can stand alone as a change to the default recommended ZooKeeper liveness and readiness probes and be reviewed at bitnami's leisure.

juan131 (changes requested):
Let's block this PR until https://github.com/bitnami/charts/pull/6547 is solved

bitnami/zookeeper/templates/statefulset.yaml
```diff
@@ -250,13 +250,17 @@ spec:
...
                 {{- else }}
-                command: ['/bin/bash', '-c', 'echo "ruok" | timeout {{ .Values.livenessProbe.probeCommandTimeout }} openssl s_client -quiet -crlf -connect localhost:{{ .Values.service.tls.client_port }} -cert {{ .Values.service.tls.client_cert_pem_path }} -key {{ .Values.service.tls.client_key_pem_path }} | grep imok']
+                  {{- if not .Values.service.tls.client_enable }}
+                  command: ['/bin/bash', '-c', 'echo "ruok" | timeout {{ .Values.livenessProbe.probeCommandTimeout }} openssl s_client -quiet -crlf -connect localhost:{{ .Values.service.tls.client_port }} | grep imok']
+                  {{- else }}
+                  command: ['/bin/bash', '-c', 'echo "ruok" | timeout {{ .Values.livenessProbe.probeCommandTimeout }} openssl s_client -quiet -crlf -connect localhost:{{ .Values.service.tls.client_port }} -cert {{ .Values.service.tls.client_cert_pem_path }} -key {{ .Values.service.tls.client_key_pem_path }} | grep imok']
+                  {{- end }}
```
juan131, line 263: Why is this block indented two extra spaces?

bitnami/zookeeper/values.yaml
juan131: I'd rather keep two separate parameters. It gives the user more freedom to customize the probes.

bitnami/zookeeper/templates/statefulset.yaml
juan131: Please do not hardcode the port to 8080. The port where Zookeeper is listening for connections is dynamic (see the block below in the environment section), thus we cannot hardcode it here.
```yaml
            - name: ZOO_PORT_NUMBER
              value: {{ .Values.service.port | quote }}
```

juan131 (commented):
We can work again on this PR since the other changes were merged @enixon !
Please check my comments

bitnami/zookeeper/templates/statefulset.yaml
enixon: 8080 is the default value for `zookeeper.admin.serverPort`. Nothing in the chart or the underlaying Docker container config allows this port to be directly set (outside of ad hoc use of the 'jvmFlags' value).

'ZOO_PORT_NUMBER' corresponds to the `clientPort` value set in the conf file supplied to the ZooKeeper instance.

bitnami/zookeeper/templates/statefulset.yaml
enixon: Looks like an oversight on my part, will change it.

enixon: Actually not an oversight - I'm not sure the right convention to use here.

It looks like each nested if/else in the original chart added a layer of indentation. Since I added a new one with `{{- if .Values.livenessProbe.useAdmin }}`, I added another two spaces of indentation. Should I strip that out?

juan131: Yes strip that out, otherwise the YAML won't be correctly formatted.

enixon:
Rebased to pick up the latest changes and I've added a second commit to address the pre-existing indentation issue pointed out by the reviewers.

juan131 (approved):
LGTM

Let's wait for @alemorcuq to share his thoughts since I recall he had some concerns regarding the use of `xxxProbe.useAdmin` parameters.

alemorcuq:
I personally don't like the idea of complicating the probes adding a toggle to switch between two kinds of probes. IMHO, charts should have two things:

1. Good, default liveness and readiness probes
2. An option to completely customize the probes (which is provided now thanks to @enixon's PR #6547)

With these two things users can deploy a Zookeeper cluster that works by default but also tweak it to add a preferred probe (like the one suggested in this PR).

On the probe itself, here's what ZooKeeper says:

> New in 3.5.3: Four Letter Words need to be explicitly white listed before using. Please refer 4lw.commands.whitelist described in cluster configuration section for details. **Moving forward, Four Letter Words will be deprecated, please use AdminServer instead**.

ZooKeeper is recommending the use of the AdminServer instead of the FLW as those will be deprecated in the future. To me, the question is: should we change the default probes to use the AdminServer? If so, then we should replace the current probes with those using the AdminServer.

But I'm also not 100% sure on changing the probe to this one. According to the [documentation](https://zookeeper.apache.org/doc/current/zookeeperAdmin.html#sc_adminserver):

> ruok : No-op command, check if the server is running. A response does not necessarily indicate that the server has joined the quorum, **just that the admin server is active and bound to the specified port**. No new fields returned.

A response indicates the admin server is active, but does it say anything about the client server? Can the admin server be in an "ok" status and not the client server? The current probe is checking the client port.

I'd like to hear your thoughts on this.

enixon:
So there are two questions here that I'll respond to.

enixon:
So there are two questions here that I'll respond to.

(1) What guarantees on server health does a check on the admin server provide?

In short, the same guarantees.

The current check directly queries the client port. Upon receipt, ZooKeeper sets up a standard client connection object and uses it to immediately send back the right response without adding the request into any of the usual processing queues. In this way it tests whether connections are capable of being established but not whether they can be processed.

The admin server command 'ruok' provides the same guarantee with a different method. It checks for the existence of the `ZooKeeperServer` object that is responsible for managing the client connection factory and processing the incoming messages into the appropriate internal queues. If the object is not healthy or not existent then a response with `"error": "This ZooKeeper instance is not currently serving requests"` will be returned. The admin server is otherwise independent of the main client connection processing so load on it does not create load on the data serving paths and, importantly for this diff, does not run into any of the logging statements that track client connections.

(2) What should the default health check be given the direction that Apache ZooKeeper has gone these last few years?

I very much think that the admin server path is the correct way to monitor the health of an Apache ZooKeeper server. That's how I would choose to manage mine.

I do question whether the admin server path is the correct way to monitor the health of bitnami ZooKeeper - that's a question for you to answer. The issue is that the underlaying Docker image that you use has no notion of admin server. By default, ZooKeeper will start with an admin server so everyone is opted in but there are no controls set up. So every change that the operator would like to make will be done through the ad hoc jvm opts option and these helm charts will have no built in ability to respond (as you point out with the path setting). Making Docker changes is outside the scope of this PR and you have to decide whether the benefits of the approach outlined here are worth it (actual INFO logging, right out the box!) balanced against the possibility of complicating some deployments.

alemorcuq:
Thank you very much for sharing your thoughts. I think you raised two very good points:

1. It could be nice for users to have INFO logging by default.
2. Using the admin server as the default probe is risky as users could be disabling it.

However, I still think `useAdmin` is redundant and not the way to go. Instead, documentation is probably the way to go for this. You are already a nice section in the README about configuring the log level. How about adapting that section to use `customLivenessProbe` and `customReadinessProbe` to change the probes? I think that would be the best solution.

enixon:
Removed the bits that would directly insert admin server queries into the liveness and readiness probes.

Extended the documentation to instruct the user how and why to change the probes themselves.

Changed the version bump from a minor to a patch.

bitnami/zookeeper/README.md
```diff
@@ -251,7 +251,32 @@ Bitnami will release a new chart updating its containers if a new version of the
...
+
+In order to remove that log noise so levels can be set to ‘INFO’, two changes must be made.
+
+First, ensure that you are not getting metrics via the deprecated pattern of polling 'mntr' on the ZooKeeper client port. The preferred method of polling for Apache ZooKeeper metrics is the ZooKeeper metrics server. This is supported in this chart when setting `metrics.enabled` to `true`.
+
+Second, to avoid the connection/disconnection messages from the probes, you can set custom values for these checks which directly them to the ZooKeeper Admin Server instead of the client port. An example is given below which preserves default probe behavior of this chart. By default, an Admin Server will be started that listens on localhost at 8080 and if you have specified either of `zookeeper.admin.serverAddress` or `zookeeper.admin.serverPort` then change the values of the probe to match. The Admin Server is on by default and you must not have disabled it (via `zookeeper.admin.enableServer`) if you wish to use it for the probes.
+
+```
```
alemorcuq, line 262: We need to disable the default probes if we want the custom probes to kick in:

```
livenessProbe:
  enabled: false
readinessProbe:
  enabled: false
```

bitnami/zookeeper/README.md
```diff
@@ -251,7 +251,32 @@ Bitnami will release a new chart updating its containers if a new version of the
...
+
+In order to remove that log noise so levels can be set to ‘INFO’, two changes must be made.
+
+First, ensure that you are not getting metrics via the deprecated pattern of polling 'mntr' on the ZooKeeper client port. The preferred method of polling for Apache ZooKeeper metrics is the ZooKeeper metrics server. This is supported in this chart when setting `metrics.enabled` to `true`.
+
+Second, to avoid the connection/disconnection messages from the probes, you can set custom values for these checks which directly them to the ZooKeeper Admin Server instead of the client port. An example is given below which preserves default probe behavior of this chart. By default, an Admin Server will be started that listens on localhost at 8080 and if you have specified either of `zookeeper.admin.serverAddress` or `zookeeper.admin.serverPort` then change the values of the probe to match. The Admin Server is on by default and you must not have disabled it (via `zookeeper.admin.enableServer`) if you wish to use it for the probes.
```
alemorcuq, line 260: I think we can omit the bit about the Java properties to avoid confusion for less experienced users. By default

```suggestion
Second, to avoid the connection/disconnection messages from the probes, you can set custom values for these checks which direct them to the ZooKeeper Admin Server instead of the client port. By default, an Admin Server will be started that listens on `localhost` at port `8080`. The following is an example of this use of the Admin Server for probes:
```

bitnami/zookeeper/README.md
```diff
@@ -251,7 +251,32 @@ Bitnami will release a new chart updating its containers if a new version of the
...
+
+```
+customLivenessProbe:
+  exec:
+    command: ['/bin/bash', '-c', 'curl -s -m 2 http://localhost:8080/commands/ruok | grep ruok']
```
alemorcuq, line 265: Why are we grepping different things in the probes?

alemorcuq (changes requested):
Thanks! Please, see my suggestions below

bitnami/zookeeper/README.md
enixon: Here's the expected output when the server can host client traffic:
```
$ curl -s http://localhost:8080/commands/ruok
{
  "command" : "ruok",
  "error" : null
}
```
And the output when the server cannot host client traffic but the jvm is still running:
```
$ curl -s http://localhost:8080/commands/ruok
{
  "command" : "ruok",
  "error" : "This ZooKeeper instance is not currently serving requests"
}
```

For liveness, I want to make sure the admin server is alive and responding coherently. That will speak to the overall health of the JVM and does not change its return value when the server/ensemble is undergoing leader election.

For readiness, I want to ensure that the client port is accepting new requests. So we grep for the correct message - namely "null" errors.

enixon:
Lint failed waited on the chart to start up a single replica deploy - `FailedScheduling  46s (x4 over 5m1s)  default-scheduler  0/1 nodes are available: 1 node(s) had taints that the pod didn't tolerate.`

If there's a way to re-run it, please do.

bitnami/zookeeper/README.md
alemorcuq: Thanks for the explanation!

enixon:
Thanks for the discussion, everyone!

juan131:
Thank you for bringing this up!

PR #6481, opened 2021-05-27
[bitnami/external-dns] Include readinessProbe and livenessProbe just in case that there are defined

### LINKED ISSUE #6471: [external-dns] Ability to disable liveness check
**Which chart**:

Chart name: external-dns
version: v4.9.1 (But even the latest one in master has the same issue)

Both the `readiness` and `liveness` probe are currently same:

```
          readinessProbe: 
            failureThreshold: 6
            httpGet:
              path: /healthz
              port: http
            initialDelaySeconds: 5
            periodSeconds: 10
            successThreshold: 1
            timeoutSeconds: 5
          livenessProbe: 
            failureThreshold: 2
            httpGet:
              path: /healthz
              port: http
            initialDelaySeconds: 10
            periodSeconds: 10
            successThreshold: 1
            timeoutSeconds: 5
```

Using a static tool like [kube-score](https://github.com/zegl/kube-score) throws this error:

```
apps/v1/Deployment external-dns in my-namespace                       💥
    [CRITICAL] Pod Probes
        · Container has the same readiness and liveness probe
            Using the same probe for liveness and readiness is very likely
            dangerous. Generally it's better to avoid the livenessProbe than
            re-using the readinessProbe.
            More information: https://github.com/zegl/kube-score/blob/master/README_PROBES.md
```

The liveness probe it currently has isn't correct. Since currently the `external-dns` pod runs as PID 1, it would be best to leave the default behaviour of the k8s controller: If the PID 1 process exits, just restart the container.

With the current helm chart, there is no way to disable the liveness probe to fall back to the default behaviour of the k8s controller. One workaround is to use something like `exec: command` pattern inside the liveness probe to check if the application is alive. But I would like to avoid that and would like the ability in the chart to disable liveness check so that I can fallback to the kubernetes implementation behaviour.

There are various alternatives recommened by kube-score itself: https://github.com/zegl/kube-score/blob/master/README_PROBES.md#livenessprobe

**Additional context**

No additional context, but can provide if something is required.

alvneiayu:
**Description of the change**

Include a support to remove the livenessProbes and readinessProbes in case that the user doesn't want to use them.

**Applicable issues**

  - fixes #6471

**Checklist** 

### DIFF
--- a/bitnami/external-dns/templates/deployment.yaml
+++ b/bitnami/external-dns/templates/deployment.yaml
@@ -498,11 +498,11 @@ spec:
           ports:
             - name: http
               containerPort: {{ .Values.service.port }}
-          {{- if .Values.readinessProbe }}
-          readinessProbe: {{ toYaml .Values.readinessProbe | nindent 12 }}
+          {{- if .Values.readinessProbe.enabled }}
+          readinessProbe: {{- omit .Values.readinessProbe "enabled" | toYaml | nindent 12 }}
           {{- end }}
-          {{- if .Values.livenessProbe }}
-          livenessProbe: {{ toYaml .Values.livenessProbe | nindent 12 }}
+          {{- if .Values.livenessProbe.enabled }}
+          livenessProbe: {{- omit .Values.livenessProbe "enabled" | toYaml | nindent 12 }}
           {{- end }}
           {{- if .Values.securityContext }}
           securityContext: {{ toYaml .Values.securityContext | nindent 12 }}
--- a/bitnami/external-dns/templates/deployment.yaml
+++ b/bitnami/external-dns/templates/deployment.yaml
@@ -498,8 +498,12 @@ spec:
           ports:
             - name: http
               containerPort: {{ .Values.service.port }}
+          {{- if .Values.readinessProbe }}
           readinessProbe: {{ toYaml .Values.readinessProbe | nindent 12 }}
+          {{- end }}
+          {{- if .Values.livenessProbe }}
           livenessProbe: {{ toYaml .Values.livenessProbe | nindent 12 }}
+          {{- end }}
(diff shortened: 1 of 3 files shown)


bitnami/external-dns/templates/deployment.yaml
```diff
@@ -496,8 +496,12 @@ spec:
           ports:
             - name: http
               containerPort: {{ .Values.service.port }}
+          {{- if .Values.readinessProbe }}
```
Mauraza, line 499: Could you add a variable enabled and change the line to `{{- if .Values.readinessProbe.enabled }}`?

bitnami/external-dns/templates/deployment.yaml
```diff
@@ -496,8 +496,12 @@ spec:
...
           readinessProbe: {{ toYaml .Values.readinessProbe | nindent 12 }}
+          {{- end }}
+          {{- if .Values.livenessProbe }}
```
Mauraza, line 502: Could you do the same thing with `livenessProbe`?

Mauraza (changes requested):
Hi @alvneiayu,

Thanks for the PR!, could you check the comments?

bitnami/external-dns/templates/deployment.yaml
alvneiayu: done, thanks @Mauraza

bitnami/external-dns/templates/deployment.yaml
alvneiayu: done, thanks @Mauraza

Mauraza (approved):
Hi @alvneiayu,

Thanks again for the PR! LGTM!, could you resolve the lint problem with the version?

PR #6528, opened 2021-06-01
[bitnami/rabbitmq] Add support for autogenerated certs

migruiz4:
**Description of the change**

Adds support for autoGenerated certificates to the bitnami/rabbitmq chart.

**Possible drawbacks**

None known.

**Checklist** 


carrodher (approved):
LGTM, just a minor comment

bitnami/rabbitmq/templates/tls-secrets.yaml
```diff
@@ -40,5 +40,46 @@ data:
...
+{{- $altNames := list (printf "*.%s.%s.svc.%s" $serviceName $releaseNamespace $clusterDomain) (printf "%s.%s.svc.%s" $serviceName $releaseNamespace $clusterDomain) $fullname }}
+{{- $crt := genSignedCert $fullname nil $altNames 365 $ca }}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ include "rabbitmq.fullname" . }}-crt
+  labels: {{- include "common.labels.standard" . | nindent 4 }}
+    {{- if .Values.commonLabels }}
+    {{- include "common.tplvalues.render" ( dict "value" .Values.commonLabels "context" $ ) | nindent 4 }}
+    {{- end }}
```
juan131, line 79: Add `.Values.commonAnnotations` block

bitnami/rabbitmq/templates/_helpers.tpl
```diff
@@ -221,3 +221,12 @@ rabbitmq: ingress.tls
...
 {{- end -}}
+
+{{/*
+Return true if a TLS credentials secret object should be created
+*/}}
+{{- define "rabbitmq.createTlsSecret" -}}
+{{- if and .Values.auth.tls.enabled .Values.auth.tls.autoGenerated (not .Values.auth.tls.existingSecret) }}
+    {{- true -}}
+{{- end -}}
+{{- end -}}
```
juan131, line 232: This is not true. You're creating the TLS secret under other circumstances too, please review this

juan131 (changes requested):
Thanks for the PR! Please review my comments

bitnami/rabbitmq/templates/_helpers.tpl
migruiz4: Hi @juan131,

I'm not entirely sure of what you mean, so I will explain my point of view to make sure we are on the same page.

`rabbitmq.createTlsSecret` determines if a TLS secret containing the autogenerated certs should be created. (Maybe I should update the helper description)

This helper is referenced only once, and it is at `tls-secret.yaml` in the following logic:

```
{{- if and .Values.auth.tls.enabled (not .Values.auth.tls.existingSecret) (not .Values.auth.tls.autoGenerated )}}

Use certificates provided as plain-text in values.yaml

{{- else if (include "rabbitmq.createTlsSecret" . )}}

Create autogenerated certs

{{ -end }}
```

I had my doubts if we should continue to support providing the certificates as plain-text, but decided to keep the backward compatibility.

So, users of RabbitMQ have 3 methods to provide TLS certificates:
- existingSecret
- Plaintext (using `caCertificate`, `serverCertificate` and `serverKey`)
- Autogenerated certificates

juan131: Macro description says: "Return true if a TLS credentials secret object should be created"

BUT a TLS secret is also created in the "Plaintext" case. So, two possibilities:

- Rename the macro and description so it clearly states that returns true if a TLS credentials secret object should be created **WITH AUTOGENERATED CERTS.**
- Include the "plaintext" possibility in this "if"

migruiz4: I decided to apply the second option, and have the `if` also include the plaintext scenario.

I also modified the secret generated, to not have duplicated code.

juan131 (approved):
LGTM

PR #6529, opened 2021-06-01
[bitnami/redis] Add support for autogenerated certs

migruiz4:
**Description of the change**

Adds support for autoGenerated certificates to the bitnami/redis chart.

**Possible drawbacks**

None known.

**Checklist** 


bitnami/redis/templates/tls-secret.yaml
```diff
@@ -0,0 +1,24 @@
...
+  labels: {{- include "common.labels.standard" . | nindent 4 }}
+    {{- if .Values.commonLabels }}
+    {{- include "common.tplvalues.render" ( dict "value" .Values.commonLabels "context" $ ) | nindent 4 }}
+    {{- end }}
+type: kubernetes.io/tls
+data:
+  ca.crt: {{ $ca.Cert | b64enc | quote }}
+  tls.crt: {{ $crt.Cert | b64enc | quote }}
+  tls.key: {{ $crt.Key | b64enc | quote }}
+{{- end }}
```
rafariossaa, line 24: Missing `\n`

bitnami/redis/README.md
```diff
@@ -337,7 +337,8 @@ The command removes all the Kubernetes components associated with the chart and
 | `pdb.maxUnavailable`                    | Max number of pods that can be unavailable after the eviction                                                       | `nil`   |
 | `tls.enabled`                           | Enable TLS traffic                                                                                                  | `false` |
 | `tls.authClients`                       | Require clients to authenticate                                                                                     | `true`  |
-| `tls.certificatesSecret`                | Then name of the existing secret that contains the TLS certificates                                                 | `nil`   |
+| `tls.autoGenerated`                     | Generate automatically self-signed TLS certificates                                                                 | `nil`   |
+| `tls.existingSecret`                    | The name of the existing secret that contains the TLS certificates                                                  | `nil`   |
```
rafariossaa, line 341: I think some kind of testing should be added to check whether both are set at the same time or not.
Also, for `autogenerated`, it should be indicated that it is a boolean value.

bitnami/redis/templates/_helpers.tpl
```diff
@@ -82,25 +82,58 @@ Return the appropriate apiVersion for PodSecurityPolicy.
...
+
+{{/*
+Return the secret containing Redis TLS certificates
+*/}}
+{{- define "redis.tlsSecretName" -}}
+{{- $secretName := coalesce .Values.tls.existingSecret .Values.tls.certificatesSecret -}}
+{{- if $secretName -}}
+    {{- printf "%s" (tpl $secretName $) -}}
+{{- else -}}
+    {{- printf "%s-crt" (include "common.names.fullname" .) -}}
```
rafariossaa, line 102: I am not very sure when this would happen.

rafariossaa (changes requested):

bitnami/redis/README.md
migruiz4: I have modified the helper "redis.createTlsSecret".
Although `autoGenerated=true`, it will be ignored if an `existingSecret`or `certificatesSecret`  is provided.

```
{{/*
Return true if a TLS secret object should be created
*/}}
{{- define "redis.createTlsSecret" -}}
{{- if and .Values.tls.enabled .Values.tls.autoGenerated (and (not .Values.tls.existingSecret) (not .Values.tls.certificatesSecret)) }}
    {{- true -}}
{{- end -}}
{{- end -}}
```

bitnami/redis/templates/_helpers.tpl
migruiz4: If `.Values.tls.existingSecret` or `.Values.tls.certificatesSecret` are provided, it will be used as the TLS secret by default. Having `existingSecret` priority over `certificatesSecret` because of the [coalesce](https://helm.sh/docs/chart_template_guide/function_list/#coalesce).

If no existing secret is provided, then it will use `$fullname-crt`, which is the name of the auto-generated secret (Check `tls-secret.yaml`)

bitnami/redis/README.md
migruiz4: > I think some kind of testing should be added to check whether both are set at the same time or not.

Additionally, I have added a validation check to trigger an error when `tls.enabled=true`, `tls.autogenerated=false`, and no existing secret is provided.

migruiz4: Testing output
```console
$helm install redis bitnami/redis --set tls.enabled=true,auth.tls.autoGenerated=false,tls.certCAFilename=/this/test,tls.authClients=false,tls.certFilename=/this/test,tls.certKeyFilename=/this/tes
Error: template: redis/templates/NOTES.txt:147:4: executing "redis/templates/NOTES.txt" at <include "redis.validateValues" .>: error calling include: template: redis/templates/_helpers.tpl:234:51: executing "redis.validateValues" at <fail>: error calling fail: 
VALUES VALIDATION:

redis: tls.enabled
    In order to enable TLS, you also need to provide
    an existing secret containing the TLS certificates or
    enable auto-generated certificates.
```

rafariossaa: You should indicate in the README the default vault for `tls.autoGenerated`

rafariossaa (approved):
LGTM

PR #6533, opened 2021-06-01
[bitnami/minio] Add support for autogenerated certs

migruiz4:
**Description of the change**

Adds support for autoGenerated certificates to the bitnami/minio chart.

**Possible drawbacks**

None known.

**Checklist** 


bitnami/minio/values.yaml
```diff
@@ -135,6 +135,10 @@ tls:
   enabled: false
   ## Name of an existing secret holding the certificate information
   ##
+  existingSecret:
+  ## DEPRECATED. Use `tls.existingSecret` instead.
```
Mauraza, line 139: I think this is confusing, could you explain a little more?

Mauraza (commented):
Hi @migruiz4,

Thanks for the PR!, could you check the comment?

bitnami/minio/templates/_helpers.tpl
```diff
@@ -214,7 +214,7 @@ minio: total drives
 Validate values of MinIO(R) - TLS secret must provided if TLS is enabled
 */}}
 {{- define "minio.validateValues.tls" -}}
-{{- if and .Values.tls.enabled (not .Values.tls.secretName) }}
+{{- if and .Values.tls.enabled (not .Values.tls.secretName) (not .Values.tls.existingSecret) (not .Values.tls.autoGenerated) }}
```
juan131, line 217: You should remove this condition (`(not .Values.tls.secretName)`), right? Or are you keeping it to have backwards compatibility?

bitnami/minio/templates/gateway/deployment.yaml
```diff
@@ -171,6 +175,18 @@ spec:
...
+        - name: minio-certs
+          secret:
+            secretName: {{ include "minio.tlsSecretName" . }}
+            items:
+            - key: tls.crt
+              path: public.crt
+            - key: tls.key
+              path: private.key
+            - key: ca.crt
+              path: CAs/public.crt
```
juan131, line 188: Why this path? Second question: should this block be optional?

bitnami/minio/templates/tls-secrets.yaml
```diff
@@ -40,5 +40,29 @@ data:
...
+{{- $headlessServiceName := printf "%s-headless" (include "common.names.fullname" .) }}
+{{- $altNames := list (printf "*.%s.%s.svc.%s" $serviceName $releaseNamespace $clusterDomain) (printf "%s.%s.svc.%s" $serviceName $releaseNamespace $clusterDomain) (printf "*.%s.%s.svc.%s" $headlessServiceName $releaseNamespace $clusterDomain) (printf "%s.%s.svc.%s" $headlessServiceName $releaseNamespace $clusterDomain) $fullname }}{{- $crt := genSignedCert $fullname nil $altNames 365 $ca }}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ include "common.names.fullname" . }}-crt
+  labels: {{- include "common.labels.standard" . | nindent 4 }}
+    {{- if .Values.commonLabels }}
+    {{- include "common.tplvalues.render" ( dict "value" .Values.commonLabels "context" $ ) | nindent 4 }}
+    {{- end }}
```
juan131, line 62: Missing the `.Values.commonAnnotations` block

bitnami/minio/templates/tls-secrets.yaml
```diff
@@ -40,5 +40,29 @@ data:
...
   tls.key: {{ $cert.Key | b64enc | quote }}
   ca.crt: {{ $ca.Cert | b64enc | quote }}
+---
+{{- end }}
 {{- end }}
+{{- if (include "minio.createTlsSecret" .) }}
+{{- $ca := genCA "minio-ca" 365 }}
+{{- $releaseNamespace := .Release.Namespace }}
+{{- $clusterDomain := .Values.clusterDomain }}
+
```
juan131, line 50: Remove this empty line

bitnami/minio/values.yaml
juan131: nice!

bitnami/minio/templates/_helpers.tpl
juan131: Responding myself: yes, backwards compatibility

juan131 (changes requested):
Thanks for the great PR! Please check my comments

bitnami/minio/values.yaml
migruiz4: Hi @Mauraza! 

In addition to adding the autogenerated certificates feature to the charts, I'm also trying to standardize the settings when backward compatibility is possible.

In this case, I'm deprecating `.Values.tls.secretName` and adding the new standard value `.Values.tls.existingSecret` included in other charts.

I considered `existingSecret` would fit better because it is already being used for authentication settings in most of the charts (`auth.existingSecret`), and it would be appropriate considering we now provide two options:
- Use an 'existingSecret' containing the certificates manually generated.
- Autogenerate the certificates secret if `tls.autoGenerated=true`.

Otherwise, a variable name such 'secretName' could be confused with an input variable used, for example, to specify the name of the autogenerated secret.

bitnami/minio/templates/_helpers.tpl
migruiz4: Yes, `Values.tls.secretName` need to be preserved for backward compatibility.

bitnami/minio/templates/gateway/deployment.yaml
migruiz4: Also, the names are not optional.

According to MinIO documentation, the relative path is built using `--certs-dir` (https://docs.min.io/docs/how-to-secure-access-to-minio-server-with-tls.html)

So, renaming from Kubernetes/tls key structure to MinIO's mandatory name policy:
- `ca.crt -> {certs-dir}/CAs/public.crt`
- `tls.crt -> {certs-dir}/public.crt`
- `tls.key -> {certs-dir}/private.key`

juan131 (approved):
LGTM!

bitnami/minio/templates/gateway/deployment.yaml
juan131: 👍

PR #6556, opened 2021-06-04
[bitnami/redis] Added HPA for redis

### LINKED ISSUE #6551: [bitnami/magento], [bitnami/redis] Add HorizontalPodAutoscaler to chart
Edit: added Redis to feature suggestion as well.

**Which chart**:
bitnami/magento, bitnami/redis, latest

It'd be good to be able to enable a HPA for Magento + Elasticsearch, and Redis and not having to create and manage one manually.

Being able to enable autoscaling and set settings via helm chart values would be good. Something like https://github.com/kubernetes/ingress-nginx/blob/9e89951ec76045f69058a6bd68692715364627d5/charts/ingress-nginx/values.yaml#L318

Creating manually, but doing this means it is managed separately which is not ideal.

mahgo:
**Description of the change**

Added Horizontal Pod Autoscaler for the Redis chart

**Benefits**
Being able to enable a HPA when using the chart.

**Applicable issues**

  - fixes #6551

**Additional information**
I was not able to test this out - for some reason the pods were already getting into a CrashLoopBackOff before I even changed anything, is anyone else having this problem?

**Checklist** 


bitnami/redis/values.yaml
```diff
@@ -1000,6 +1008,14 @@ pdb:
   ## @param pdb.maxUnavailable Max number of pods that can be unavailable after the eviction
   ##
   maxUnavailable:
+## Autoscaling configuration
```
alvneiayu, line 1011: Could you move all of those values inside the replica section, please?. I mean, the autoscaling is going to work only for the replicas and right now the values are in a section shared with the master and the replicas.

alvneiayu (changes requested):
Hi @mahgo

I reviewed the patch. Everything looks good to me but I think that it is better to move the autoscaling values inside the replica section.

Again, thanks a lot for your time and your work.

Álvaro

bitnami/redis/values.yaml
mahgo: Done

bitnami/redis/templates/replicas/hpa.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if .Values.autoscaling.enabled }}
+{{- if .Values.replicate.autoscaling.enabled }}
```
alvneiayu, line 1: {{- if .Values.replica.autoscaling.enabled }}

alvneiayu (commented):
hi @mahgo 

I saw that you are having some linter errors. You have a Typo using "replicate" instead of "replica".

Thanks a lot for your time.

Álvaro

bitnami/redis/templates/replicas/hpa.yaml
mahgo: Thanks, fixed.

alvneiayu (approved):
LGTM

kalali:
If sentinel is used the SS name is derived using ` {{ printf "%s-node" (include "common.names.fullname" .) }}` while when there is no sentinel the SS name is derived using ` {{ printf "%s-replicas" (include "common.names.fullname" .) }}`. The hpa needs to take that into account.

alvneiayu:
hi @kalali 

Thanks a lot for notifying this problem. We are open to receive PRs and we will really happy to review it.

Thanks a lot again for your time and your help

Álvaro

PR #6559, opened 2021-06-04
[bitnami/magento] Added HorizontalPodAutoscaler to Magento

### LINKED ISSUE #6551: [bitnami/magento], [bitnami/redis] Add HorizontalPodAutoscaler to chart

mahgo:
**Description of the change**
Added HorizontalPodAutoscaler to the Magento chart.

**Benefits**
 Being able to enable a HPA when using the chart.

**Possible drawbacks**

**Applicable issues**

  - fixes #6551

**Checklist** 


carrodher (changes requested):
Thanks for your contribution.
I just added a couple of commits:
 - Unifying the format of the resources in the values.yaml for the metric container
 - Updating the README table to reflect the changes in the format

Something I can't modify since the file is not part of the PR is the `deplyoment.yaml`; in this file the `resources:` call should be modify to use the new approach:
```diff
### Around line 249
           {{- if .Values.resources }}
-          resources: {{- include "common.tplvalues.render" (dict "value" .Values.resources "context" $) | nindent 12 }}
+          resources: {{- toYaml .Values.resources | nindent 12 }}
           {{- end }}

### Around line 278
-          resources: {{- include "common.tplvalues.render" (dict "value" .Values.metrics.resources "context" $) | nindent 12 }}
+          resources: {{- toYaml .Values.metrics.resources | nindent 12 }}
```

This is the format used by the volumePermission init container (see line 78) that is already using the new format:
```yaml
           {{- if .Values.volumePermissions.resources }}
           resources: {{- toYaml .Values.volumePermissions.resources | nindent 12 }}
           {{- end }}
```
```yaml
   ## Init containers' resource requests and limits
   ## ref: http://kubernetes.io/docs/user-guide/compute-resources/
   ##
   resources:
     ## We usually recommend not to specify default resources and to leave this as a conscious
     ## choice for the user. This also increases chances charts run on environments with little
     ## resources, such as Minikube. If you do want to specify resources, uncomment the following
     ## lines, adjust them as necessary, and remove the curly braces after 'resources:'.
     ##
     limits: {}
     ##   cpu: 100m
     ##   memory: 128Mi
     ##
     requests: {}
     ##   cpu: 100m
     ##   memory: 128Mi
     ##
```

Apart from that, I am suggesting you bump the chart version into a minor instead of a patch since those changes are a feature addition instead of bug fixing.

Thank you again for your contribution!

mahgo:
Hi @carrodher,

This is now updated.

carrodher (approved):
Thanks! Changes LGTM. I just launched the internal test & release process, see https://github.com/bitnami/charts/blob/master/CONTRIBUTING.md#pr-approval-and-release-process

PR #6564, opened 2021-06-04
[bitnami/elasticsearch] Correct default data.replicas in the documentation

yariksheptykin:
According to `values.yaml` the default value for `data.replicas` is `2` not `3`.

See: https://github.com/bitnami/charts/blob/master/bitnami/elasticsearch/values.yaml#L553


juan131 (approved):
Thanks for the fix! LGTM

PR #6575, opened 2021-06-04
[bitnami/grafana] Allow use of ingress without a hostname

jerryharrow:
**Description of the change**

Allow use of an ingress without a known hostname. It allows the hostname rule to be dropped by use of the "*" value in the chart. This value is not valid for a hostname input, so it does not introduce any incompatibly with prior behavior, or ambiguity.

**Benefits**

Allows use of an existing ingress without a known hostname associated.   In our use case the ingress has only an IP address no name, so there is no value that can be specified to enable it to match.   This allows to expose grafana on that nginx ingress.

**Possible drawbacks**

Would have been preferable to not have a default value for ingress.hostname and just omit the hostname rule if not supplied, but that would be an incompatible change if others are expecting "grafana.local" by default.
Using the value "*" as a sentinel value for this case seems acceptable as that value is not accepted by kubernetes so there is no ambiguity.

**Applicable issues**

  - n/a

**Additional information**

Same capability as provided to [bitnami/kibana] chart in https://github.com/bitnami/charts/pull/6460
Quoting a reference by @migruiz4  from that PR.

According to Kubernetes documentation (https://kubernetes.io/docs/reference/kubernetes-api/service-resources/ingress-v1/#IngressSpec):

- When rule.host is not specified, traffic will be routed based on rule.http.

`If the host is unspecified, the Ingress routes all traffic based on the specified IngressRuleValue.
`

- rule.host does not support using wildcard by itself.

`You cannot have a wildcard label by itself (e.g. Host == "*").
`
Although, even if Kubernetes support rules without specifying rule.host, it is possible that certain Ingress controllers don't support it. 

**Checklist** 

- [x ] Variables are documented in the README.md


juan131 (approved):
Thanks so much for the contribution @jerryharrow ! I'm already familiar with this contribution due to the work you did at Kibana and I don't see any downsides on support it.
LGTM

PR #6577, opened 2021-06-04
[bitnami/harbor]Add missing 'skipverify' s3 storage driver option

gnanderson:
* add to template
  * add to values.yaml (with ref to storage driver docs)
  * add to readme

**Description of the change**

Updates Harbor chart with missing s3 TLS skip verify storage option

**Benefits**

Our internal cluster runs Harbor and Minio in the same cluster, we currently use Kustomize post render on the chart to add `skipverify` and obviously this would avoid having to do that

**Possible drawbacks**

Someone might use this in production - I don't want to police that behaviour though ;)

**Checklist** 


bitnami/harbor/Chart.yaml
```diff
@@ -34,4 +34,4 @@ sources:
   - https://github.com/bitnami/bitnami-docker-harbor-registry
   - https://github.com/bitnami/bitnami-docker-harbor-registryctl
   - https://goharbor.io/
-version: 10.0.6
+version: 10.0.7
```
carrodher, line 37:
```suggestion
version: 10.1.0
```

carrodher (commented):
Thanks for your contribution! Changes LGTM, just a minor suggestion about the version; since it is a feature addition instead of a bug fix, I would bump the minor version instead of the patch one

gnanderson:
@carrodher thanks for the suggestion :+1:

carrodher (approved):
LGTM, thanks. I just launched the internal test & release process, if everything works as expected the PR should be automatically merged and the chart released in some hours.
[Here](https://docs.bitnami.com/kubernetes/faq/get-started/understand-charts-release-process/) you can find more info about this process.

PR #6589, opened 2021-06-07
[bitnami/common] extend common.labels.matchLabels with .Values.extraMatchLabels

### LINKED ISSUE #6588: [bitnami/common] extend common.labels.matchLabels with .Values.commonMatchLabels
**Which chart**: redis, `v14.3.4`

When some charts (e.g. redis, etc.) are used multiple times as sub-charts (with different aliases) in an umbrella chart, matching labels confusing pods from these sub-charts.

It is possible to introduce additional set of matching labels that can be supplied through values files, that can remove this confusion.

kuznero:
**Description of the change**

Make it possible to supply `.Values.extraMatchLabels` that become part of matching labels but supplied from values files.

**Benefits**

Makes it possible to re-use same chart (e.g. redis) as sub-chart multiple times without confusing pods through same matching labels.

**Applicable issues**

  - fixes #6588

**Additional information**

**Checklist** 


kuznero:
PR is ready for review

miguelaeh (commented):
Hi @kuznero ,
Thank you very much for the PR.
I have two comments about it. If I understood the use case properly, which is deploy the same chart as a dependency more than once, `commonMatchLabels` would not be an appropriate name from my point since they are actually not common but intended to be different match labels provided manually.
Also, I think that we would need to update the values.yaml and READMEs of the rest of charts, if not, no one will know about this new parameter.

kuznero:
Hi @miguelaeh, thanks for the quick feedback. What would be your recommendation for the name that I should change it to? The following options come to mind:

- `customMatchLabels`
- `additionalMatchLabels`
- `extraMatchLabels`

kuznero:
I also updated the `README.md`, please let me know if you think that is not sufficient.

kuznero:
Hi @miguelaeh, I fixed the name and also updated the README. It is ready to be reviewed again.

bitnami/common/README.md
```diff
@@ -87,6 +87,8 @@ The following table lists the helpers available in the library which are scoped
 | `common.labels.standard`    | Return Kubernetes standard labels                    | `.` Chart context |
 | `common.labels.matchLabels` | Return the proper Docker Image Registry Secret Names | `.` Chart context |
 
+`common.labels.matchLabels` might include extra labels if necessary from `.Values.extraMatchLabels`. This might be necessary when a chart is used multiple times in an umbrella chart in order to ensure different stateful sets or deployments are not using overlapping match labels.
```
miguelaeh, line 90: We can't set this here since we are implementing the autogeneration of the READMEs and this message will disappear.

miguelaeh (commented):
Hi @kuznero ,
After thinking about it, I think that it would be better to add the 
```
{{- if .Values.extraMatchLabels }}
{{- range $key, $value := .Values.extraMatchLabels }}
{{- end }}
{{- end }}
```

to each chart where it is needed. It would be the way to take into account the parameters in order to add them to the respective values.yaml and also to the READMES of the charts where it can be set. I feel that adding this change at the common level will hide the parameters for the rest of users.

bitnami/common/README.md
kuznero: Hi @miguelaeh, I assumed this is a very generic thing that might be required by most charts when used multiple times in an umbrella chart. But I understand the concern. Will update this PR shortly to be on a specific chart level (Redis) instead.

kuznero:
Hi @miguelaeh, I localized the change on [bitnami/redis] chart where I needed it in the first place, and removed any changes to [bitnami/common] as you suggested. Please review.

kuznero:
Hi @miguelaeh, it is getting a bit hairy when done on a chart's level. What is missing in the current implementation is `common.affinities.pods` template that in turn is using `common.labels.matchLabels`. Thus, current approach in this PR is not going to work for affinities when I would really need to use `redis.labels.matchLabels` re-defined on a chart's level. And if I will find a way to override all references to `common.labels.matchLabels` it will be a lot more difficult to maintain going forward.

Thus I would ask if it makes sense to still try to change `common.labels.matchLabels` in `common` to include `.Values.extraMatchLabels` instead of pushing this logic down onto the chart's level?

miguelaeh:
Hi @kuznero ,Thank you very much for the changes. I think in that case we should try to add a new parameter to the affinity function in order to pass the `extraMatchLabels`. i.e. add the parameter in the common chart and the logic to set them if they are passed, and use them when necessary in the redis chart.
An important point doing that is to verify that does not break any backwards compatibility and it continues working in the rest of charts that are not passing that parameter without issues.

kuznero:
> 
> 
> Hi @kuznero ,Thank you very much for the changes. I think in that case we should try to add a new parameter to the affinity function in order to pass the `extraMatchLabels`. i.e. add the parameter in the common chart and the logic to set them if they are passed, and use them when necessary in the redis chart.
> An important point doing that is to verify that does not break any backwards compatibility and it continues working in the rest of charts that are not passing that parameter without issues.

Hi @miguelaeh, I reverted it to the minimum necessary changes in common. Affinity related things addressed this way automatically as it simply includes `common.labels.matchLabels` there. The only question I have at this point is how to document it such that it becomes clear to the consumers of common chart as well as other charts.

miguelaeh:
Hi @kuznero ,
I think I maybe did not explain properly what I was trying to say. I was not talking about reverting to the initial change but to add a new parameter to the affinity function in the common chart, to the `dict` section. So, in the Redis chart, when calling the affinity functions, you will need to pass the extraMatchLabels to the function call.
Have in mind, the common chart is thought to be used as a function library so we don't like the idea of adding default parameters in the common chart. What we should do is to add the capability to pass the parameters when needed to the functions.

kuznero:
Hi @miguelaeh, sorry the confusion, and thanks for clarifying the guiding principles behind [bitnami/common] chart. I hope I am following your guidelines better this time. In this PR there is a change to [bitnami/common] that adds parameters to affinity templates as dictionary entries.

And this is how it will look in [bitnami/redis] (on the following PR, after new common will be accepted) when I will implement it there: [kuznero/redis-matching-labels](https://github.com/kuznero/charts/commit/5d92836761037972b806e8303dd71cdbd784e7d5) branch.

bitnami/common/templates/_affinities.tpl
```diff
@@ -47,17 +47,21 @@ Return a nodeAffinity definition
 
 {{/*
 Return a soft podAffinity/podAntiAffinity definition
-{{ include "common.affinities.pods.soft" (dict "component" "FOO" "context" $) -}}
+{{ include "common.affinities.pods.soft" (dict "component" "FOO" "extraMatchLabels" .Values.extraMatchLabels "context" $) -}}
```
miguelaeh, line 50: Hi @kuznero ,
Thank you very much for all the changes!

Did you check that this will continue working with the rest of charts that are not passing the `extraMatchLabels`?

kuznero: Yes, I set it to empty `dict` by default. Checked it with `redis` chart.

miguelaeh (approved):
Thank you very much for all the changes!
LGTM!

PR #6592, opened 2021-06-07
[bitnami/discourse] Fix URLs for emojis and attachments

### LINKED ISSUE #6572: [bitnami/discourse] 127.0.0.1 being used as hostname in attachment previews
**Which chart**:
discourse 4.0.0

**Describe the bug**
We used chart v2.1.1 for months with no problems. After upgrading to 4.0.0, when creating posts with image uploads, the preview URLs are created by Discourse with a hostname of 127.0.0.1. Both `discourse.host` and `discourse.hostname` are set in the values YAML.

For example:

```
<img src="https://127.0.0.1/secure-media-uploads/optimized/2X/5/[redacted].png" alt="[redacted]" data-base62-sha1="[redacted]" width="500" height="500" class="d-lazyload d-lazyload-hidden">
```

**To Reproduce**

As above.

**Expected behavior**

The preview URLs are created as before, with the correct hostname.

v1.21.1

marcosbc:
**Description of the change**

In version 4.0.0 we introduced major changes to the chart. However we missed a minor thing that caused Sidekiq to continue using old configuration that would break any generated URLs.

Before the Sidekiq container needed to wait for the Discourse container to be ready (listening on port 3000). However that logic was improved and now instead it is waiting for DB migrations to be completed, to consider the DB ready to start.

Note that Sidekiq is used for things such as generating URLs for attachments so if it uses the wrong URLs, wrong ones will be generated. Now, URLs from the [`templates/configmaps.yaml` template](https://github.com/bitnami/charts/blob/master/bitnami/discourse/templates/configmaps.yaml) will be used instead.

**Benefits**

Attachments show proper URL.

**Possible drawbacks**

None.

**Applicable issues**

  - fixes #6572
  - fixes https://github.com/bitnami/bitnami-docker-discourse/issues/130

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/discourse/Chart.yaml
+++ b/bitnami/discourse/Chart.yaml
@@ -35,4 +35,4 @@ sources:
   - https://github.com/bitnami/bitnami-docker-discourse
   - https://github.com/spinnaker
   - https://www.discourse.org/
-version: 4.0.0
+version: 4.0.1
--- a/bitnami/discourse/templates/deployment.yaml
+++ b/bitnami/discourse/templates/deployment.yaml
@@ -157,10 +157,6 @@ spec:
           env:
             - name: BITNAMI_DEBUG
               value: {{ ternary "true" "false" .Values.image.debug | quote }}
-            - name: DISCOURSE_HOST
-              value: "127.0.0.1"
-            - name: DISCOURSE_PORT
-              value: "3000"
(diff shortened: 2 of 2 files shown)


marcosbc:
Tested locally and it works.

juan131 (approved):
LGTM! Thanks for the quick fix @marcosbc

PR #6594, opened 2021-06-07
[bitnami/grafana operator] Make the securityContexts more customizable.

Cellebyte:
**Description of the change**

**Benefits**

Make `containerSecurityContext` and `podSecurityContext` more user friendly.

**Possible drawbacks**

**Applicable issues**

  - fixes #

**Additional information**

**Checklist** 


bitnami/grafana-operator/values.yaml
```diff
@@ -102,6 +102,9 @@ operator:
     enabled: true
     runAsUser: 1001
     runAsGroup: 1001
+    capabilities:
+      drop:
+      - ALL
```
pablogalegoc, line 107: Why are we dropping all capabilities when the containers are running by default as non-root and also setting the `allowPrivilegeEscalation` to false?

pablogalegoc (commented):
Hi @Cellebyte, thanks for your contribution! Check out the comment I left when you have time

bitnami/grafana-operator/values.yaml
Cellebyte: I just added it to comply with our policies which ask for that. It is not needed and if I should not add it we can drop it from the `securityContext`

pablogalegoc: Then in this case I would remove it just to keep the PR only related to the configurability improvements. Thank you very much for your work!

Cellebyte: Should i bump then only the patch release?
This is not changing the behaviour of the `values.yaml` interface

pablogalegoc: Yes, just bump the patch version. Thank you!

PR #6616, opened 2021-06-09
[bitnami/redis-cluster] Replace cluster creator job by cluster update job in pdb selector

mboutet:
**Description of the change**

Replace cluster creator job by cluster update job in pdb selector as the creator job was removed some versions ago. The initialization is now done directly inside the first pod of the statefulset.

**Benefits**

Less unused code is in the chart.

**Possible drawbacks**

None I can think of.

**Applicable issues**

None.

**Additional information**

None.

**Checklist** 

### DIFF
--- a/bitnami/redis-cluster/Chart.yaml
+++ b/bitnami/redis-cluster/Chart.yaml
@@ -23,4 +23,4 @@ name: redis-cluster
 sources:
   - https://github.com/bitnami/bitnami-docker-redis
   - http://redis.io/
-version: 6.1.4
+version: 6.1.5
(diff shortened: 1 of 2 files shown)


migruiz4:
Hi @mboutet,

Thank you very much for your contribution! Your PR looks good to me. 

I'm wondering if we should include the `*-cluster-update` job in the PDB exclusion list. 
Adding @miguelaeh to the discussion.

mboutet:
I don't use the update cluster job, but it would make sense to have it. Otherwise, the PDB will take into account the job. Waiting for your inputs.

miguelaeh:
Hi guys,
I agree that job must be excluded from the list of the PDB. The job must be run only once until marked as completed.

migruiz4:
Thank you @miguelaeh for your input!

@mboutet would you mind modifying this PR to instead replace the excluded job with `*-cluster-update`?

Your contribution looks good to me and I would have no problem merging as it is, but I think it would be better if we could fix two issues in one change.

mboutet:
@migruiz4, done!

migruiz4:
Great! Thank you very much for your contribution @mboutet.

migruiz4 (approved):

PR #6639, opened 2021-06-11
[bitnami/mongodb] Standardize certificates autogeneration

migruiz4:
**Description of the change**

Standardize the certificate autogeneration to use the same settings as the other charts.

**Benefits**

All the charts with TLS should have similar settings.

**Possible drawbacks**

IMPORTANT: Certificates won't be autogenerated by default, users will now have to set `tls.autoGenerated=true` for that to happen. Otherwise, a warning asking to provide a caCert or a caKey will be displayed using `required`.

**Checklist** 


Mauraza (approved):
Hi @migruiz4,

Thanks for the PR!, LGTM!!

PR #6645, opened 2021-06-11
[bitnami/solr] Add support for autogenerated certs

alvneiayu:
**Description of the change**

The autogenerated certs support allow us to generate automatically certificates to test or to verify that the TLS/SSL is working properly.

**Benefits**

Make easy to test SSL scenarios.

**Checklist** 


bitnami/solr/values.yaml
```diff
@@ -157,6 +160,39 @@ tls:
...
+      ## lines, adjust them as necessary, and remove the curly braces after 'resources:'.
+      ##
+      limits: {}
+      ##   cpu: 100m
+      ##   memory: 128Mi
+      ##
+      requests: {}
+      ##   cpu: 100m
+      ##   memory: 128Mi
+      ##
```
migruiz4, line 194: Please check the references to these values.

They are always referred to as `tls.resources`, but here are defined as `tls.image.resources`.

alvneiayu: good catch. Thanks @migruiz4

pablogalegoc (approved):
There are conflicts with the `Chart.yaml` but other than that LGTM!

pablogalegoc:
LGTM!

PR #6646, opened 2021-06-11
[bitnami/nginx] Custom Scrape port for Exporter

oliverbaehler:
**Description of the change**

Added new Value `$.Values.metrics.port`. If it's set a custom port can be configured which is the used by the exporter to scrape the status. The Default value is no longer hard coded to `8080` but takes as default the value of `$.Values.containerPorts.http`.

**Benefits**

When Changing the Port, the scraper currently no longer works, since the port is configured statically so you have to define the entire sidecar yourself. With this change the port value is considered and the possibility is given, to overwrite the port to any value.

**Possible drawbacks**

Don't see any yet. 

**Applicable issues**

Just did the pr with issue.

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/nginx/Chart.yaml
+++ b/bitnami/nginx/Chart.yaml
@@ -25,4 +25,4 @@ name: nginx
 sources:
   - https://github.com/bitnami/bitnami-docker-nginx
   - http://www.nginx.org
-version: 9.1.0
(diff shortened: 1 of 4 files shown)


javsalgar (approved):
Thank you very much for the PR! LGTM

PR #6652, opened 2021-06-14
[bitnami/discourse] Add SMTP variables

kxesd:
**Description of the change**

This PR adds all the SMTP environment variables as values in the helm chart.

**Benefits**

Allows configuring SMTP environment variables in the helm chart.

**Additional information**

Previously `bitnami/discourse/templates/secrets-discourse.yaml` would only ever have a single value, `discourse-password`, however sometimes it may now have both (`discourse-password` and `smtp-password`) or (one of them) or (none at all). I've changed the logic and it seems to handle all of these cases fine.

NOTE: maybe there needs to be warning added around specifying either `discourse.smtp.password` OR `discourse.smtp.existingSecret`. Currently if you specify both `discourse.smtp.existingSecret` will take priority.

**Checklist** 


bitnami/discourse/templates/secrets-discourse.yaml
```diff
@@ -12,9 +12,12 @@ metadata:
...
 type: Opaque
 data:
-  {{- if .Values.discourse.password }}
+  {{- if and (.Values.discourse.password) (not .Values.discourse.existingSecret) }}
   discourse-password: {{ .Values.discourse.password | b64enc | quote }}
-  {{- else }}
+  {{- else if not .Values.discourse.existingSecret }}
   discourse-password: {{ randAlphaNum 10 | b64enc | quote }}
   {{- end }}
+  {{- if and (.Values.discourse.smtp.password) (not .Values.discourse.smtp.existingSecret) }}
```
alemorcuq, line 20: We should probably check for `smtp.enabled` here as well.

bitnami/discourse/Chart.yaml
```diff
@@ -35,4 +35,4 @@ sources:
   - https://github.com/bitnami/bitnami-docker-discourse
   - https://github.com/spinnaker
   - https://www.discourse.org/
-version: 4.1.1
+version: 4.1.3
```
alemorcuq, line 38: Version should be a new minor since we are adding a new feature to the chart.

alemorcuq (changes requested):
Thanks for the PR, @kxesd. Please, check my comments.

kxesd:
Thanks @alemorcuq. I've made the changes as per your comments.

alemorcuq (approved):
Thanks!

PR #6657, opened 2021-06-14
[bitnami/external-dns] update azure private dns provider

dza89:
Signed-off-by: daan <dhapotter@gmail.com>

**Description of the change**

From 8.0.0 Azure Private DNS only works with the same config file as the Azure Provider.
As stated here in the docs:
https://github.com/kubernetes-sigs/external-dns/blob/master/docs/tutorials/azure-private-dns.md#deploy-externaldns
(Follow the steps for azure-dns provider to create a configuration file.)

and this PR:
https://github.com/kubernetes-sigs/external-dns/pull/1912

**Additional information**
Only tested MSI

**Checklist** 

- [n/a] Variables are documented in the README.md


bitnami/external-dns/Chart.yaml
```diff
@@ -24,4 +24,4 @@ sources:
   - https://github.com/kubernetes-sigs/external-dns
   - https://github.com/bitnami/bitnami-docker-external-dns
   - https://github.com/kubernetes-sigs/external-dns
-version: 5.1.0
+version: 5.1.1
```
pablogalegoc, line 27: New line needed here to pass the lint check

bitnami/external-dns/values.yaml
```diff
@@ -144,8 +144,6 @@ azure:
   ## When a secret to load azure.json is not specified,
   ## the host's /etc/kubernetes/azure.json will be used
   ##
-  ## Deprecated: please use tenantId, subscriptionId, aadClientId and aadClientSecret values instead.
-  ##
```
pablogalegoc, line 148: Why remove the deprecation notice?

pablogalegoc (commented):
Hi @dza89! Thanks for your contribution! I left some comments:

bitnami/external-dns/Chart.yaml
dza89: I removed the new line because it failed on the lint check, let me reimplement it.

bitnami/external-dns/values.yaml
dza89: What exactly is deprecated here? The secret? That's actually now the only way to go.

bitnami/external-dns/Chart.yaml
pablogalegoc: This is strange, Github says this file has 28 lines in the [UI view](https://github.com/dza89/charts-4/blob/update-azure-private-dns/bitnami/external-dns/Chart.yaml) but if you download it raw it has 29, so the lint check fails. It may be that your IDE is messing with the endlines, you can verify locally if the file passes the lint check by running `ct lint --config .github/ct-lint.yaml` at the repo root directory

bitnami/external-dns/values.yaml
pablogalegoc: I think it means that it is preferred to set these values so the credentials are generated from them:
https://github.com/bitnami/charts/blob/master/bitnami/external-dns/templates/_helpers.tpl#L216

bitnami/external-dns/Chart.yaml
dza89: I tried it in the github web ide, shouldn't try that anymore it seems. Fixed.

bitnami/external-dns/values.yaml
dza89: Ah ok. i've reinstated them.

pablogalegoc:
Thanks @dza89! LGTM!

PR #6665, opened 2021-06-14
[bitnami/mariadb-galera] Change mountpath of custom config to /bitnami/mariadb/conf/my_custom.cnf

### LINKED ISSUE #6651: [bitnami/mariadb-galera] mariadbConfiguration in default values.yaml not honoured
The default values.yaml file exposes a `mariadbConfiguration` section, which is mounted within the container to `/bitnami/conf/my.cnf`, which is never read. Mariadb runs within the `/opt/bitnami/mariadb/` folders, and the config files it uses come from `/opt/bitnami/mariadb/conf/my.cnf`

Seems like this change was introduced in #6567, but the default mariadb datadir/home directory didn't change, causing this issue.

CrazyByDefault:
**Description of the change**

Changes path to which custom config is mounted, in order for it to be picked up by the mariadb image's `mysql_initialize` function.

**Benefits**

Fixes broken custom config defined in values.yaml

**Possible drawbacks**

None, the new config mount would be following the standards defined in the [base image repo](https://github.com/bitnami/bitnami-docker-mariadb-galera)

**Applicable issues**

  - fixes #6651 

**Additional information**

From the bitnami/bitnami-docker-mariadb-galera repo, upon inspecting both `libmariadbgalera.sh` and `mariadb-env.sh` in all the versions, two things are evident -
  - All config files are in a directory of the format `${BASE_DIR}/mariadb/conf` as opposed to just `${BASE_DIR}/conf`, and hence moved the custom conf path in the statefulset to `/bitnami/mariadb/conf` as opposed to `/bitnami/conf`
  - All *custom* config files picked up by the `mysql_initialize` function are of the form `my_custom.cnf`, not `my.cnf`

The relevant script is available in - [libmariadbgalera.sh:1126](https://github.com/bitnami/bitnami-docker-mariadb-galera/blob/master/10.5/debian-10/rootfs/opt/bitnami/scripts/libmariadbgalera.sh#L1541) and [libmariadbgalera.sh:1085](https://github.com/bitnami/bitnami-docker-mariadb-galera/blob/master/10.5/debian-10/rootfs/opt/bitnami/scripts/libmariadbgalera.sh#L1085)

**Checklist** 

### DIFF
--- a/bitnami/mariadb-galera/templates/statefulset.yaml
+++ b/bitnami/mariadb-galera/templates/statefulset.yaml
@@ -277,8 +277,8 @@ spec:
             {{- end }}
             {{- if or (.Files.Glob "files/my.cnf") .Values.mariadbConfiguration .Values.configurationConfigMap }}
             - name: mariadb-galera-config
-              mountPath: /bitnami/conf/my.cnf
-              subPath: my.cnf
+              mountPath: /bitnami/mariadb/conf/my_custom.cnf
+              subPath: my_custom.cnf
             {{- end }}
             {{- if .Values.tls.enabled }}
             - name: mariadb-galera-certificates
--- a/bitnami/mariadb-galera/Chart.yaml
+++ b/bitnami/mariadb-galera/Chart.yaml
@@ -27,4 +27,4 @@ sources:
   - https://github.com/bitnami/bitnami-docker-mariadb-galera
   - https://github.com/prometheus/mysqld_exporter
   - https://mariadb.org
-version: 5.10.0
+version: 5.10.1
--- a/bitnami/mariadb-galera/templates/configmap.yaml
+++ b/bitnami/mariadb-galera/templates/configmap.yaml
@@ -8,7 +8,7 @@ data:
 {{- if (.Files.Glob "files/my.cnf") }}
 {{ (.Files.Glob "files/my.cnf").AsConfig | indent 2 }}
 {{- else if .Values.mariadbConfiguration }}
-  my.cnf: |
+  my_custom.cnf: |
 {{ .Values.mariadbConfiguration | indent 4 }}
 {{- end }}
 {{ end }}


bitnami/mariadb-galera/templates/statefulset.yaml
```diff
@@ -277,8 +277,8 @@ spec:
             {{- end }}
             {{- if or (.Files.Glob "files/my.cnf") .Values.mariadbConfiguration .Values.configurationConfigMap }}
             - name: mariadb-galera-config
-              mountPath: /bitnami/conf/my.cnf
-              subPath: my.cnf
+              mountPath: /bitnami/mariadb/conf/my_custom.cnf
```
Mauraza, line 280: Could you try with` /bitnami/conf/my_custom.cnf`?

Mauraza (commented):
Hi @CrazyByDefault,

Thanks for the PR!, could you check the comments?

bitnami/mariadb-galera/templates/statefulset.yaml
CrazyByDefault: Will give it a go @Mauraza, but the mysql_initialize script only looks for `my_custom.cnf` files within `${base_dir}/mariadb/conf/`. lemme see if it changes anything

CrazyByDefault: @Mauraza Nope, doesn't work - 

conf mounted at `/bitnami/conf/my_custom.cnf`

image


The main config being used by mariadb is - `/opt/bitnami/mariadb/conf/my.cnf`, and it has the default config - 

image


Ideally, the custom config should be copied to `/opt/bitnami/mariabd/conf/bitnami/my_custom.cnf` to be picked up the `!include` statement at the bottom. But, the folder empty - 

image


Looks like `/bitnami/mariadb/conf/my_custom.cnf` is the only one that works.

Mauraza:
Hi @CrazyByDefault,

We have a small problem with the new revision of the container when it will be released should solve the problem. 
We will update this thread when this happens.

Mauraza:
Hi @CrazyByDefault,

The new revision of the chart was just released. Could you check if that resolves the issue?

CrazyByDefault:
@Mauraza this seems to be fixed in the latest version! thanks a lot, that was quick :D

PR #6684, opened 2021-06-16
[bitnami/redis] make it possible to use redis chart multiple times in an umbrella chart

### LINKED ISSUE #6683: [bitnami/redis] make it possible to use redis chart multiple times in an umbrella chart
**Which chart**: bitnami/redis, v14.4.0

When bitnami/redis chart is used multiple times in an umbrella chart `matchLabels` are confused.

This PR adds support for `.Values.extraMatchLabels` which allows to specify additional match labels per sub-chart in an umbrella chart. This way it will be possible to ensure that `matchLabels` are unique per redis sub-chart.

**Additional context**
In order to be able to completely extend `matchLabels` in bitnami/redis chart it was necessary to extend `common.affinities.pods.*` templates to allow for additional labels. This change was merged with #6589.

kuznero:
**Description of the change**

This PR adds support for .Values.extraMatchLabels which allows to specify additional match labels per sub-chart in an umbrella chart. This way it will be possible to ensure that matchLabels are unique per redis sub-chart.

**Benefits**

**Possible drawbacks**

**Applicable issues**

  - fixes #6683

**Additional information**

In order to be able to completely extend matchLabels in bitnami/redis chart it was necessary to extend common.affinities.pods.* templates to allow for additional labels. This change was merged with #6589.

**Checklist** 


javsalgar:
Hi,

Thank you so much for the PR! After thinking about it, I think it makes sense for this change to be in the bitnami/common chart instead. Modifying the `common.labels.matchLabels` function. Plus, about the `common.labels.standard`, I think you could use the `commonLabels` value that most of our charts have.

kuznero:
> 
> 
> Hi,
> 
> Thank you so much for the PR! After thinking about it, I think it makes sense for this change to be in the bitnami/common chart instead. Modifying the `common.labels.matchLabels` function. Plus, about the `common.labels.standard`, I think you could use the `commonLabels` value that most of our charts have.

Hi @javsalgar, in my previous attempt I have tried changing `common.labels.matchLabels` to directly load from `.Values.extraMatchLabels` in case it is present. But after discussing it with @miguelaeh, I went for just extending redis chart. Any chance we can discuss it again together and agree on what approach is best?

I can see that from the maintenance perspective I would have extended `common.labels.matchLabels` directly with `.Values.extraMatchLabels` and just document it such that it is clear that it can be done in all the bitnami charts. Otherwise, similar thing will need to be done for many (if not all) of the bitnami charts.

javsalgar:
@miguelaeh my main concern with this is that we start deviating again after achieving some standardization. For example here, we would be going back to `redis.labels` helpers just like we had in the past. My take would be in, even though it may take a bit more time, in adding the support in the common chart and once we have all the annotated values from @pablogalegoc , I imagine it won't hurt much adding the `extraMatchLabels` value in all the charts. Thoughts?

kuznero:
Please keep it open, I plan to get back to it this week.

rafariossaa:
Hi @kuznero 
  Did you have the chance to work on this ?

kuznero:
> 
> 
> Hi @kuznero
> Did you have the chance to work on this ?

Not yet, but plan to get back to it during August hopefully.

rafariossaa:
Hi,
  In that case, feel free to reopen this when needed.

PR #6734, opened 2021-06-23
[bitnami/mysql] add networkpolicy for primary (#6720)

### LINKED ISSUE #6720: NetworkPolicy for mysql
**Which chart**:
mysql

Similar to the [postgresql chart's](https://github.com/bitnami/charts/blob/master/bitnami/postgresql/templates/networkpolicy.yaml), would it be possible to implement a policy for this chart too?

Let me know, as I'm happy to attempt creating a merge request ;)

Thanks!

bynare:
**Description of the change**

This changes adds the ability to enable and configure a Network Policy on the primary.

**Benefits**

Better security.

**Possible drawbacks**
n/a

**Applicable issues**

  - fixes #6720

**Checklist** 


bitnami/mysql/templates/networkpolicy.yaml
```diff
@@ -0,0 +1,40 @@
...
+            matchLabels:
+              {{ template "common.names.fullname" . }}-client: "true"
+          {{- if .Values.primary.networkPolicy.explicitNamespacesSelector }}
+          namespaceSelector:
+{{ toYaml .Values.primary.networkPolicy.explicitNamespacesSelector | indent 12 }}
+          {{- end }}
+        - podSelector:
+            matchLabels:
+            {{- include "common.labels.matchLabels" . | nindent 14 }}
+              role: read
```
migruiz4, line 33: The MySQL replicas do not have a label `role: read`. Instead they have `app.kubernetes.io/component: secondary`.

migruiz4 (changes requested):
Hi @bynare,

I have suggested some changes, could you please have a look?

bynare:
Hey @migruiz4, 

Thanks for the review! :)

I've amended the things you've pointed out - thanks!

migruiz4:
Thank you very much for applying the suggestions!

There is one more thing I'd like to discuss with you. Providing a NetworkPolicy that only applies to the primary nodes seems incomplete.

For example, without a Network policy for secondary replicas, metrics wouldn't be accessible for all the pods.

We should cover both the primary node and secondary replicas with a single networkPolicy. Rename `primary.networkPolicy.*` as `networkPolicy.*` and remove "app.kubernetes.io/component: primary" from the networkPolicy podSelector.

Or otherwise, add a networkPolicy for the secondary replicas (`secondary.networkPolicy.*`) and rename the template `networkpolicy.yaml` to `primary/networkpolicy.yaml`.

Both approaches would be fine for me, feel free to let me know what you think.

bynare:
All good @migruiz4 - I was in two minds initially too: whether to include secondary in this.

> We should cover both the primary node and secondary replicas with a single networkPolicy. Rename primary.networkPolicy.* as networkPolicy.* and remove "app.kubernetes.io/component: primary" from the networkPolicy podSelector.

^ I've made changes to go with this option you suggested ;)

Thanks again for reviewing and bearing with me!

migruiz4 (approved):
Looks good to me!

PR #6762, opened 2021-06-25
[bitnami/zookeeper] Rework security context

abestel:
**Description of the change**

In the zookeeper charts, thhe security context was shared between pod and container, and not all options could be overridden:
- only the `fsGroup` was used in the pod securityContext
- `runAsUser` was forced to `0` in the initContainer securityContext
- only the `runAsUser` was used in the container securityContext

**Benefits**

It is now possible to fully override the securityContext of both the pod and containers.
This is also more in line with the charts template provided in the repository.

**Possible drawbacks**

I am not sure why the `runAsUser` was force to `0` in the initContainer... there may be a regression there since the user is now `1001` by default.

**Applicable issues**

 None that I know of.

**Additional information**

/

**Checklist** 

### DIFF
--- a/bitnami/zookeeper/Chart.yaml
+++ b/bitnami/zookeeper/Chart.yaml
@@ -21,4 +21,4 @@ name: zookeeper
 sources:
   - https://github.com/bitnami/bitnami-docker-zookeeper
   - https://zookeeper.apache.org/
-version: 7.0.4
+version: 7.0.5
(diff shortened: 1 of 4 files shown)


alemorcuq:
Thanks, @abestel. This is something that should indeed be reworked, but it will probably mean a new major of the chart as it would break backwards compatibility. Let me discuss with the team before doing a proper review of the PR.

carrodher:
I'm closing this PR since it seems those changes were already implemented at https://github.com/bitnami/charts/pull/8318

PR #6776, opened 2021-06-27
[bitnami/thanos] Add Prometheus alerts for Thanos components

francoisminaud:
**Description of the change**

One missing feature of the Bitnami Thanos chart is the Prometheus alerting.
The [Thanos mixin alerts](https://github.com/thanos-io/thanos/tree/main/mixin) were adapted to the Bitnami chart. 

Most alerts will be added to the Prometheus Rule CRD when the specific (compactor, query ...) is enabled.

**Benefits**

When enabled, you will get alerts when something is going wrong with your Thanos components.
Another benefit is that all alerts are configurable and can be changed to fit the user's needs.

**Possible drawbacks**

**Applicable issues**

  - fixes #

**Additional information**

This was tested with fullNameOverride values enabled, which is how we use the Bitnami charts.

**Checklist** 


francoisminaud:
> Variables are documented in the README.md 

Aren't those automatically generated by a tool when formatting the yaml files accordingly?

bitnami/thanos/values.yaml
```diff
@@ -2582,6 +2582,566 @@ metrics:
...
+    ## Namespace in which the PrometheusRule CRD is created
+    ##
+    namespace: monitoring
+    ## List of rules for each component
+    ##
+    rules:
+      ##
+      ## Compactor Prometheus Rules
+      ##
+      compactor:
```
alvneiayu, line 2598: I think that maybe it is a good idea to use strings, something like:
```
rules:
  compactor: | 
```
and use the common.tplvalues.render. Same with query, receive, etc

alvneiayu (commented):
hi @francoisminaud

My sincere apologies but you need to include it manually. We are working in a solution to try to help with this.

Álvaro

alvneiayu (changes requested):
hi @francoisminaud 

First of all, thanks a lot for your PR and your hard work. Looking the patch, I think that maybe it is a good idea to use strings as I suggested in the values.yaml.
Including this, I think that the other parts LGTM.

Again, thanks a lot for your PR

Álvaro

bitnami/thanos/values.yaml
francoisminaud: Hi @alvneiayu 
This was updated, thank you for your review. 
Happy to contribute to bitnami charts, we've been using them a lot.

François

bitnami/thanos/values.yaml
```diff
@@ -2582,6 +2582,567 @@ metrics:
...
+  ## PrometheusRule CRD configuration
+  ##
+  prometheusRule:
+    enabled: false
+    ## Namespace in which the PrometheusRule CRD is created
+    ##
+    namespace: monitoring
+    ## List of rules for each component
+    ##
+    rules:
```
alvneiayu, line 2594: This is something that I noticed taking a look other charts. We are not including the rules by default, ex:

https://github.com/bitnami/charts/blob/77df34ca6871e06c7bcaa51291b90afe964ec323/bitnami/redis/values.yaml#L1259

Please, could you comment them? I mean, this is something that the users must to modify to configure them. So, I suggest to comment them. Thanks again, you are really close

alvneiayu (changes requested):
Hi @francoisminaud

First of all, thanks a lot for the PR and you are really close.

I included a comment inside the code. Moreover, could you fix the conflict with the chart.yaml?

Thanks a lot

Álvaro

alvneiayu:
hi @francoisminaud 

I created a PR moving the rules shared by you to our chart documentation. We are going to have a really big block of comments inside the values.yaml, if we comment all the prometheus rules. I decide to move to there and we can remove it. Let me know if you think that it is ok in your side.

Thanks a lot

Álvaro

bitnami/thanos/values.yaml
francoisminaud: Indeed they are not included by default!

francoisminaud:
Hi @alvneiayu,

I agree that the values file is really going to bloat if we keep it there. 
Adding them to the documentation will make the values.yaml more readable and to be honest the most important is being able to enable some alerts in the Chart, HOW it is done is not that important!
Thanks for your time.

François

bitnami/thanos/values.yaml
alvneiayu: sorry @francoisminaud, I used a bad word (English is not my mother language). If you take a look in other charts, we try to include new features trying to avoid specific configurations (sometimes we can do it and sometimes not). If you take a look for example Redis or Mariadb-galera, you can see that we are including the prometheusRules feature but we are not setting up any rules. We don't want to establish specific rules for the users. You did a really hard work and we don't want to lose it, for that I decided to move your rules to the documentations repo. Please, remove the rules to keep following the same way that we did in other charts. 

Thanks a lot again for your really hard work and again sorry.

francoisminaud:
Hi @alvneiayu 
I could not do it sooner, does that work for you? I used the postgresql chart as example.
The values file does not feature all the alerts, just added one as example in comments.
Cheers,
François

alvneiayu (approved):
LGTM

alvneiayu:
hi @francoisminaud 

Thanks a lot for your PR again.

Álvaro

francoisminaud:
@alvneiayu can you please confirm that the documentation was added? I cannot see it

alvneiayu:
hi @francoisminaud 

I created the PR and our internal document department require me several changes and I need to apply them. I am still working on it. I hope to finish it as soon as possible.

Thanks a lot for your patience and thanks for your help. I will ping you when the change is finally merged.

Álvaro

PR #6818, opened 2021-07-01
[bitnami/odoo] MAJOR: Odoo image refactor

marcosbc:
**Description of the change**

Update chart to be fully compatible with new refactored Bitnami Redmine container image. Also take advantage of major version change to perform some standarizations.

**Benefits**

Users can enjoy an improved container image :)

**Possible drawbacks**

Upgrades are not guaranteed to work

**Applicable issues**

  - fixes #

**Additional information**

**Checklist** 


bitnami/odoo/templates/deployment.yaml
```diff
@@ -24,8 +31,14 @@ spec:
       {{- end }}
     spec:
       {{- include "odoo.imagePullSecrets" . | nindent 6 }}
+      {{- if .Values.schedulerName }}
+      schedulerName: {{ .Values.schedulerName | quote }}
+      {{- end }}
+      serviceAccountName: {{ .Values.serviceAccountName }}
```
juan131, line 37: We're using `serviceAccount.name` and `serviceAccount.create` parameter in other charts which allow the user to decide whether to create/user a custom service account to run the application or not. I think we should follow the same approach here.

bitnami/odoo/templates/deployment.yaml
```diff
@@ -42,11 +55,31 @@ spec:
       tolerations: {{- include "common.tplvalues.render" (dict "value" .Values.tolerations "context" $) | nindent 8 }}
       {{- end }}
       {{- if .Values.podSecurityContext.enabled }}
-      securityContext:
-        fsGroup: {{ .Values.podSecurityContext.fsGroup }}
+      securityContext: {{- omit .Values.podSecurityContext "enabled" | toYaml | nindent 8 }}
       {{- end }}
-      {{- if .Values.initContainers }}
-      initContainers: {{- include "common.tplvalues.render" (dict "value" .Values.initContainers "context" $) | nindent 8 }}
+      {{- if or (and .Values.volumePermissions.enabled .Values.persistence.enabled) (.Values.initContainers) }}
```
juan131, line 60: Simpler:
```suggestion
      {{- if or (and .Values.volumePermissions.enabled .Values.persistence.enabled) .Values.initContainers }}

```

bitnami/odoo/templates/deployment.yaml
```diff
@@ -120,30 +187,14 @@ spec:
...
-          livenessProbe:
-            httpGet:
-              path: /
-              port: http
-            initialDelaySeconds: {{ .Values.livenessProbe.initialDelaySeconds }}
-            periodSeconds: {{ .Values.livenessProbe.periodSeconds }}
-            timeoutSeconds: {{ .Values.livenessProbe.timeoutSeconds }}
-            successThreshold: {{ .Values.livenessProbe.successThreshold }}
-            failureThreshold: {{ .Values.livenessProbe.failureThreshold }}
+          livenessProbe: {{- omit .Values.livenessProbe "enabled" | toYaml | nindent 12 }}
```
juan131, line 192: I though we agreed on hardcoding the probe method (`httpGet` in this case) and exposing only the threshold/seconds parameters to have opinionated probes. Keeping the possibility of using `customXXXProbe` for those users interested on customizing them.

bitnami/odoo/templates/tls-secrets.yaml
```diff
@@ -18,3 +20,25 @@ data:
 ---
 {{- end }}
 {{- end }}
+{{- if and .Values.ingress.tls (not .Values.ingress.certManager) }}
```
juan131, line 23: We're changing this approach and introducing a parameter `ingress.selfSigned` for users interested on having self signed certificates to make it explicit. 
Check Jenkins, Kubeapps or RabbitMQ that are already using this approach.

juan131 (changes requested):
Thanks for the PR @marcosbc !!! Please check my comments when you have a chance.

bitnami/odoo/templates/tls-secrets.yaml
marcosbc: OK, applied.

bitnami/odoo/templates/deployment.yaml
marcosbc: OK, reverted.

bitnami/odoo/templates/deployment.yaml
marcosbc: Ok, applied

bitnami/odoo/templates/_helpers.tpl
```diff
@@ -67,3 +117,10 @@ Odoo credential secret name
 {{- define "odoo.secretName" -}}
 {{- coalesce .Values.existingSecret (include "odoo.fullname" .) -}}
 {{- end -}}
+
+{{/*
+Return the SMTP Secret Name
+*/}}
+{{- define "odoo.smtpSecretName" -}}
+{{- coalesce .Values.smtpExistingSecret (include "common.names.fullname" .) -}}
+{{- end -}}
```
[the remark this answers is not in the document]
marcosbc: For now, I would avoid this in order not to breaking existing scenarios (i.e. a user that deployed with an existing database but without specifying the port number, and using the default value).

bitnami/odoo/ci/values-hpa-pdb.yaml
```diff
@@ -0,0 +1,4 @@
+autoscaling:
+  enabled: true
+pdb:
+  create: true
```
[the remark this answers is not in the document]
marcosbc: I added it but I get errors when pushing, i.e.:

```
$ helm template --values bitnami/odoo/ci/values-metrics-and-ingress.yaml bitnami/odoo | kubeval --ignore-missing-schemas --strict --openshift failed.
...
templates/ingress.yaml contains an invalid Ingress - service: Additional property service is not allowed
```

So I am going to skip this.

marcosbc:
Ready for review again, thanks

bitnami/odoo/ci/values-hpa-pdb.yaml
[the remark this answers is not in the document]
juan131: That's normal and it's due to some fields we're including in the **ingress.yaml** that **kubeval** doesn't recognize. If that's the only thing the GitHooks actions complain about, send the changes with `--no-verify` flag.
@javsalgar idk if this is something we want to revisit

juan131 (approved):
LGTM

bitnami/odoo/ci/values-hpa-pdb.yaml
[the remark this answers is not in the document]
marcosbc: For now I'd avoid including such change because it is making the actual push process fail, and I'd not like a user to face such an issue when trying to contribute (it looks like something is wrong with the chart).

PR #6840, opened 2021-07-05
[bitnami/*] Avoid creating PVC when unneeded

juan131:

**Description of the change**

There are several apps that requires the external hostname (or IP) to be passed during the chart installation to configure the app properly.

Currently, we're not creating the **deployment** when we detect this happens (while installing other manifests such as the database ones), and show a warning in the **NOTES.txt** explaining the steps to follow to complete the installation. This is done to avoid users to face weird issues and `Crashloopbackoff` errors.

When you install the chart using `--wait` flag, it waits for every resource to be ready. While we're not creating the  **deployment** manifests as I mention before, we are creating the associated PVC manifests. These PVC(s) remain as "Pending" since the pods that are supposed to use them are never created, therefore the `helm install --wait ...` fails.

**Benefits**

`helm install --wait ...` works properly even if the external hostname is not provided.

**Possible drawbacks**

None

**Applicable issues**

N/A

**Additional information**

N/A

**Checklist** 

### DIFF
--- a/bitnami/discourse/Chart.yaml
+++ b/bitnami/discourse/Chart.yaml
@@ -35,4 +35,4 @@ sources:
   - https://github.com/bitnami/bitnami-docker-discourse
   - https://github.com/spinnaker
   - https://www.discourse.org/
-version: 4.2.1
+version: 4.2.2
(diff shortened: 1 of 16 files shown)


gongomgra (approved):
LGTM

PR #6846, opened 2021-07-06
[bitnami/nginx] Use tpl for staticSiteConfigmap and staticSitePVC

lwj5:
**Description of the change**

Allow using templating in `staticSiteConfigmap` and `staticSitePVC`.

**Benefits**

If Nginx is a sub-chart of a chart, we can dynamically define the PVC to mount for Nginx to serve static content.

**Possible drawbacks**

Nil

**Applicable issues**

  - fixes #

**Additional information**

Currently, the chart already allows tpl for server block config map so this enhancement will add to that to make it fully functional as a sub chart.

**Checklist** 

### DIFF
--- a/bitnami/nginx/Chart.yaml
+++ b/bitnami/nginx/Chart.yaml
@@ -25,4 +25,4 @@ name: nginx
 sources:
   - https://github.com/bitnami/bitnami-docker-nginx
   - http://www.nginx.org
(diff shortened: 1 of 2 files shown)


rafariossaa (approved):
Hi, 
  Thanks for adding this.
   LGTM.

PR #6860, opened 2021-07-06
[bitnami/redis] Specify master replicas to be 1

erebe:
**Description of the change**

Specify explicitly the default replicas (1) for Redis master instead of leaving it implicit

**Benefits**

  + If we omit to specify the replicas it is defaulted to 1.
    By omitting the value, an issue arise if an operator scale (down/up)
    the replicaset and after try to re-apply the helm chart in order to
    revert operations. In the case the replicas value is omitted
    helm/kube will not scale back the replicaset to its original default
    value.

  + This PR intends to paliate to the above issue. It just adds
    replicas in the chart, so Helm/Kube will re-adjust/scale the
    replicaset even if changed by an operator.

**Possible drawbacks**

Replicaset will be re-scaled every time the chart is applied if not at the default value, but that's the desired behavior

**Checklist** 

### DIFF
--- a/bitnami/redis/Chart.yaml
+++ b/bitnami/redis/Chart.yaml
@@ -25,4 +25,4 @@ name: redis
 sources:
   - https://github.com/bitnami/bitnami-docker-redis
   - http://redis.io/
-version: 14.6.4
+version: 14.6.5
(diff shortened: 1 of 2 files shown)


marcosbc (approved):
LGTM, thanks for the fix!

PR #6864, opened 2021-07-06
[bitnami/kube-state-metrics] Common labels and annotations

### LINKED ISSUE #6821: [bitnami/common] Extra common labels
Hi,
Is it possible to have a common way to declare extra labels for all manifest of a chart ?

Most of the chart use template "common.labels.standard" to declare theirs labels. This template do not take parameter from values file yet. And it's not possible to add custom labels.

In kube-prometheus we call "kube-prometheus.labels" wich call "common.labels.standard" and add custom label from .Values.global.labels

```
{{/*
Common Labels
*/}}
{{- define "kube-prometheus.labels" -}}
{{ include "common.labels.standard" . }}
{{- if .Values.global.labels }}
{{ toYaml .Values.global.labels }}
{{- end }}
{{- end -}}
```

I want to be able to add custom labels for many charts but i'm lazy and i want to factor.

Is it possible to modify the common library like this or is not so easy ?
```
{{/*
Kubernetes standard labels
*/}}
{{- define "common.labels.standard" -}}
app.kubernetes.io/name: {{ include "common.names.name" . }}
helm.sh/chart: {{ include "common.names.chart" . }}
{{- if .Values.global.labels }}
{{ toYaml .Values.global.labels }}
{{- end }}
{{- end -}}
```

I can do the job if it's a good solution.
Regards,
Etienne.

EtienneBarbier:
**Description of the change**

Addition of commonLabels et commonAnnotations field, to be able to apply this two elements to all manifests of the chart.

**Benefits**

We can now apply Labels and annotations for all k8s ressources deployed by this chart.

**Possible drawbacks**

None

**Applicable issues**

  - fixes #6821 

**Checklist** 


bitnami/kube-state-metrics/values.yaml
```diff
@@ -8,11 +8,6 @@ global:
   #   - myRegistryKeySecretName
   # storageClass: myStorageClass
 
-  ## Additional labels to apply to all resources
-  ##
-  labels: {}
-  #   foo: bar
-
```
alemorcuq, line 15: Can you remove this from the README as well?

alemorcuq (changes requested):
Thanks for the PR! Please, check my comment

bitnami/kube-state-metrics/values.yaml
EtienneBarbier: Sure !

PR #6870, opened 2021-07-06
[bitnami/redis] Fix issue with IP detection when mawk is used

aureus1:
**Description of the change**

The IP detection logic for dualstack environments used two different awk scripts to detect ipv4 IP - one for sentinel (used mawk) and one for redis (used gnu awk in prior versions).  Current installation has mawk on both environments.  As the mawk version works on gnu awk, using it in both containers prevents further issues from occuring.

**Benefits**
IPv4 detection is not tied to specifc awk implementations.  Container changes will not impact configmap script functionality.

**Possible drawbacks**
None

**Applicable issues**

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/redis/Chart.yaml
+++ b/bitnami/redis/Chart.yaml
@@ -25,4 +25,4 @@ name: redis
 sources:
   - https://github.com/bitnami/bitnami-docker-redis
   - http://redis.io/
-version: 14.6.5
+version: 14.6.6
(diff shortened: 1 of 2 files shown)


pablogalegoc (approved):
Thanks for your contribution @aureus1! I see this is related to https://github.com/bitnami/charts/pull/6350, so I'll leave the reference here to have more context. LGTM!

PR #6881, opened 2021-07-07
Upgrade action versions

alemorcuq:
**Description of the change**
Upgrade `kind-action` and `chart-testing-action` to their latest versions. This should fix the issues we are having with the `test` job.


carrodher (approved):
Lgtm

PR #6882, opened 2021-07-07
[bitnami/mongodb] Specify standalone replicas to be 1

erebe:
**Description of the change**

Specify explicitly the default replicas (1) for standalone mongo instead of leaving it implicit

P.s: Found the same issue than yesterday, PR for Redis master

**Benefits**

  + If we omit to specify the replicas it is defaulted to 1.
    By omitting the value, an issue arise if an operator scale (down/up)
    the replicaset and after try to re-apply the helm chart in order to
    revert operations. In the case the replicas value is omitted
    helm/kube will not scale back the replicaset to its original default
    value.

  + This PR intends to paliate to the above issue. It just adds
    replicas in the chart, so Helm/Kube will re-adjust/scale the
    replicaset even if changed by an operator.

**Possible drawbacks**

Replicaset will be re-scaled every time the chart is applied if not at the default value, but that's the desired behavior

**Checklist** 

### DIFF
--- a/bitnami/mongodb/Chart.yaml
+++ b/bitnami/mongodb/Chart.yaml
@@ -26,4 +26,4 @@ name: mongodb
 sources:
   - https://github.com/bitnami/bitnami-docker-mongodb
   - https://mongodb.org
-version: 10.21.0
+version: 10.21.1
--- a/bitnami/mongodb/templates/standalone/dep-sts.yaml
+++ b/bitnami/mongodb/templates/standalone/dep-sts.yaml
@@ -22,6 +22,7 @@ metadata:
     {{- end }}
   {{- end }}
 spec:
+  replicas: 1
   {{- if .Values.useStatefulSet }}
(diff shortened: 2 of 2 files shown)


Mauraza:
Hi @erebe,

Thanks for the PR!! Could you explain this a little more?  Did you change the number of replicas of the standalone? 
It will be nice if you can add steps to reproduce the issue.

erebe:
Hello,
In essence, it is the same issue as https://github.com/bitnami/charts/pull/6860

For reproducing : 
  1. Do a helm install of the mongo chart
  2. Do a kubectl scale --replicas 0 on the statefulset
  3. Do a helm upgrade to revert your change
  4. Notice that the helm upgrade is not scaling back the replica to 1, and let it to 0

Basically, if we omit to specify replicas in the statefulset at the installation it will take the default value (1) but is never to be updated after the install.

Mauraza (approved):
Hi @erebe,

Thanks for the explanation and the PR! LGTM!

PR #6883, opened 2021-07-07
[bitnami/*] Adapt values.yaml of kubernetes-event-exporter, kubewatch, Logstash charts

pablogalegoc:
**Description of the change**

Add metadata to `values.yaml` in order to be used by [readme-generator](https://github.com/bitnami-labs/readme-generator-for-helm) to automatically generate the "Parameters" section of the `README.md`.

General rules followed:
- If `values.yaml` doesn't have a description for a key in a comment → Use `README.md` description
- If `values.yaml` does have a description for a key in a comment but the `README.md` description provides more info → Use `README.md` description
- Else → Leave the `values.yaml` description

**Benefits**

Less manual work and consistency between `values.yaml` and `README.md`.

**Additional information**

Commented keys are not picked by readme-generator so some had to be uncommented in order to appear in the README, however none of them affect the default installation of the chart, verified comparing the output before and after the changes of
```bash
helm install chart . -f values.yaml --dry-run --debug
```

**Checklist** 


bitnami/kubernetes-event-exporter/values.yaml
```diff
@@ -1,93 +1,121 @@
...
+        ## layout:
+        ##   message: "{{ .Message }}"
+        ##   reason: "{{ .Reason }}"
+        ##   type: "{{ .Type }}"
+        ##   count: "{{ .Count }}"
+        ##   kind: "{{ .InvolvedObject.Kind }}"
+        ##   name: "{{ .InvolvedObject.Name }}"
+        ##   namespace: "{{ .Namespace }}"
+        ##   component: "{{ .Source.Component }}"
+        ##   host: "{{ .Source.Host }}"
```
marcosbc, line 85: Since this is a code example, why are we using two `#` formatting before? Usually for code examples, we only include one. I.e.:


```
  ## Example:
  ##
  # pullSecrets:
  #   - myRegistryKeySecretName
```

marcosbc (commented):
Thanks for these improvements! I have added a small concern with the group of these changes, however, let me know if that is how it should be expected.

bitnami/kubernetes-event-exporter/values.yaml
pablogalegoc: I slipped up in this case, to follow the new format it should be like in 9421ede, just like the example you point out is now
```
  ## Example:
  ## pullSecrets:
  ##
  pullSecrets: []
```
Thanks!

marcosbc (approved):
LGTM!

PR #6902, opened 2021-07-09
[bitnami/rabbitmq] Fix default empty "initContainers" object

SimonHarrisonSH:
**Description of the change**

Currently `initContainers` is an empty map.
- "containers" suggests is should be a list
- `statefulset.yaml` has a single-entry list
- when adding an initContainer in the values.yaml, a render warning is issued:
   " Merging destination map for chart 'rabbitmq'. Overwriting table item 'initContainers', with non table value: [map[...."
- Replace default value with empty list

**Benefits**

Removes the helm render error.

**Possible drawbacks**

None 

**Applicable issues**

 None

**Additional information**

**Checklist** 


marcosbc (approved):
LGTM, thanks!

PR #6958, opened 2021-07-15
[bitnami/discourse] Avoid to include ports in DISCOURSE_HOST

### LINKED ISSUE #6932: Discourse with SSL Offloading via Load Balancer
**Which chart**:
Discourse 4.2.3

**Describe the bug**
Long story short I found that the helm chart adds a port to the `DISCOURSE_HOST` in the configmap if the port is not `80`. In this case I am using 443 for SSL offloading on a AWS NLB. This breaks avatar images from working after you upload them. If I remove the port from the configmap directly on the cluster that fixes the issue. But any images uploaded before the change can't be used again. Even spent time trying to delete them from the file system and clearing assets and running precompile again. Spent way to much time on this. The problem is unless you change the SHA of an image you can never upload that image again. Even if I reverted the change but then images don't load. 

https://github.com/bitnami/charts/blob/131fa45f5fdf59adb3ddb00ca440143eff889260/bitnami/discourse/templates/configmaps.yaml#L14

Line 14 is where it adds the port in. From what I can gather there shouldn't be a port there at all, this value is just the hostname of the instance of discourse. 

I think removing that port should work. I can do a PR tomorrow if that won't break anything else?

**To Reproduce**

Use port 443 
1. deploy discourse
2. upload a logo, avatar, etc.
3. Edit the configmap to remove the extra port in DISCOURSE_HOST
4. Restart pod
5. Previous images don't show so clear and reupload and they still won't work
6. Change the image or upload a different image and those work fine

**Expected behavior**
Image/Avatars work out of the box, including profile menu icons etc. 

Kubernetes 1.19 on EKS

- Output of `helm version`:

```
version.BuildInfo{Version:"v3.1.1", GitCommit:"afe70585407b420d0097d07b21c47dc511525ac8", GitTreeState:"clean", GoVersion:"go1.13.8"}
```

```
Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.2", GitCommit:"52c56ce7a8272c798dbc29846288d7cd9fbae032", GitTreeState:"clean", BuildDate:"2020-04-16T23:34:25Z", GoVersion:"go1.14.2", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"19+", GitVersion:"v1.19.8-eks-96780e", GitCommit:"96780e1b30acbf0a52c38b6030d7853e575bcdf3", GitTreeState:"clean", BuildDate:"2021-03-10T21:32:29Z", GoVersion:"go1.15.8", Compiler:"gc", Platform:"linux/amd64"}
```

**Additional context**
Using AWS Network load balancer with ACM on port 443

alvneiayu:
**Description of the change**

This patch modifies the way that we are including the port inside the DISCOURSE_HOST in our chart. After an internal investigation, it seems that Discourse doesn't support to modify the default 80 or 443 ports:

https://meta.discourse.org/t/how-to-specify-a-different-port-not-80-during-installation/47859/10

This patch modifies the DISCOURSE_HOST included in our configmap to set up just the host without the port.

**Applicable issues**

  - fixes #6932

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/discourse/Chart.yaml
+++ b/bitnami/discourse/Chart.yaml
@@ -35,4 +35,4 @@ sources:
   - https://github.com/bitnami/bitnami-docker-discourse
   - https://github.com/spinnaker
   - https://www.discourse.org/
-version: 4.2.3
+version: 4.2.4
--- a/bitnami/discourse/templates/configmaps.yaml
+++ b/bitnami/discourse/templates/configmaps.yaml
@@ -11,7 +11,7 @@ metadata:
   {{- end }}
 data:
   {{- $port := .Values.service.port | toString }}
-  DISCOURSE_HOST: "{{ include "discourse.host" . }}{{- if (and (ne $port "80") (ne $port "443")) }}:{{ .Values.service.port }}{{ end }}"
+  DISCOURSE_HOST: "{{ include "discourse.host" . }}"
   DISCOURSE_SKIP_INSTALL: {{ ternary "yes" "no" .Values.discourse.skipInstall | quote }}
   DISCOURSE_SITE_NAME: {{ .Values.discourse.siteName | quote }}
   DISCOURSE_USERNAME: {{ .Values.discourse.username | quote }}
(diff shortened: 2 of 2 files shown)


rafariossaa (approved):
LGTM

PR #6966, opened 2021-07-15
[bitnami/airflow] Remove nil values

migruiz4:
**Description of the change**

Removes `nil` values from `bitnami/airflow/values.yaml`.

**Checklist** 

### DIFF
--- a/bitnami/airflow/values.yaml
+++ b/bitnami/airflow/values.yaml
@@ -132,11 +132,11 @@ rbac:
 ## ref: https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/
 ## @param serviceAccount.annotations Annotations for service account. Evaluated as a template. Only used if `create` is `true`.
 ## @param serviceAccount.create Specifies whether a ServiceAccount should be created
+## @param serviceAccount.name Name of the service account to use. If not set and create is true, a name is generated using the fullname template.
 ##
 serviceAccount:
   annotations: {}
   create: false
-  ## The name of the service account to use. If not set and create is true, a name is generated using the fullname template.
   name: ""
 ## @param sidecars List of sidecar containers to be added to web, worker and scheduler pods
 ## Example:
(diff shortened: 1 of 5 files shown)


bitnami/airflow/values.yaml
```diff
@@ -148,7 +148,7 @@ serviceAccount:
 ##       - name: portname
 ##         containerPort: 1234
 ##
-sidecars:
+sidecars: []
```
carrodher, line 151:
```suggestion
sidecars: {}
```

bitnami/airflow/values.yaml
```diff
@@ -350,7 +350,7 @@ web:
   ##       - name: portname
   ##         containerPort: 1234
   ##
-  initContainers:
+  initContainers: []
```
carrodher, line 353:
```suggestion
  initContainers: {}
```

carrodher (commented):
By agreement in the template used for new charts, we are using `{}` instead of `[]`. See https://github.com/bitnami/charts/blob/master/template/CHART_NAME/values.yaml

migruiz4:
Hi @carrodher,

In that case, I think it is something we should fix in the template.

In the example, you can see initContainers is an "Array of objects", but not an object itself.
```yaml
  ## initContainers:
  ##  - name: your-image-name
  ##    image: your-image
  ##    imagePullPolicy: Always
  ##    command: ['sh', '-c', 'echo "hello world"']
```

carrodher (commented):
Can you bump the chart version in the Chart.yaml file? In the same way, the README table should be regenerated by running the automated tool.

And just in case, can you double-check there are no single quotes? If yes, please, replace them from `''` to `""`. The current changes LGTM

carrodher:
There are some `''`:
```console
$ ag ": ''" airflow/values.yaml
97:  key: ''
98:  type: ''
112:podAffinityPreset: ''
358:  priorityClassName: ''
507:  priorityClassName: ''
705:  priorityClassName: ''
852:  bindpw: ''
865:    CAcertificateSecret: ''
869:    CAcertificateFilename: ''
```

carrodher (approved):
LGTM

PR #6967, opened 2021-07-15
[bitnami/postgresql] disable automounting of service account token

jonkerj:
**Description of the change**
It disables the automounting of the service account token in the pod. PostgreSQL does not need this. This commit disables it by default, but leaves it configurable if anyone needs to use it.

**Benefits**
By disabling the automount, potential attackers cannot access the Kubernetes API on behalf/through the pod.

**Possible drawbacks**
If anyone is injecting extra sidecar containers doing API access using a mounted token, they will need to explicitly enable the token automount. 

**Applicable issues**
If raising an issue is required or helpful, please tell me and I'd be glad to do so.

**Additional information**

**Checklist** 


bitnami/postgresql/Chart.yaml
```diff
@@ -26,4 +26,4 @@ name: postgresql
 sources:
   - https://github.com/bitnami/bitnami-docker-postgresql
   - https://www.postgresql.org/
-version: 10.5.3
+version: 10.5.4
```
javsalgar, line 29: As it is adding a non breaking feature, could you bump the minor version?

javsalgar (changes requested):
Thank you so much! Small change

bitnami/postgresql/Chart.yaml
jonkerj: Sure no prob. Pushing now

javsalgar (approved):

PR #6968, opened 2021-07-15
[bitnami/nginx]: disable automount of SA token

jonkerj:
**Description of the change**
It disables the automounting of the service account token in the pod. Nginx does not need this. This commit disables it by default, but leaves it configurable if anyone needs to use it.

**Benefits**
By disabling the automount, potential attackers cannot access the Kubernetes API on behalf/through the pod.

**Possible drawbacks**
In the unlikely event that anyone is using some sidecar to access the Kubernetes API, relying on the mounted SA token, it will break with this commit.

**Applicable issues**
If raising an issue is required or helpful, please tell me and I'd be glad to do so.

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/nginx/templates/deployment.yaml
+++ b/bitnami/nginx/templates/deployment.yaml
@@ -33,6 +33,7 @@ spec:
       {{- end }}
     spec:
       {{- include "nginx.imagePullSecrets" . | nindent 6 }}
+      automountServiceAccountToken: {{ .Values.serviceAccount.autoMount }}
       serviceAccountName: {{ template "nginx.serviceAccountName" . }}
       {{- if .Values.hostAliases }}
       hostAliases: {{- include "common.tplvalues.render" (dict "value" .Values.hostAliases "context" $) | nindent 8 }}
--- a/bitnami/nginx/values.yaml
+++ b/bitnami/nginx/values.yaml
@@ -274,6 +274,9 @@ serviceAccount:
   ## Only used if `create` is `true`.
   ##
   annotations: {}
+  ## @param serviceAccount.autoMount Auto-mount the service account token in the pod
+  ##
+  autoMount: false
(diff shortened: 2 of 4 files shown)


bitnami/nginx/templates/deployment.yaml
```diff
@@ -33,6 +33,7 @@ spec:
       {{- end }}
     spec:
       {{- include "nginx.imagePullSecrets" . | nindent 6 }}
+      automountServiceAccountToken: false
```
pablogalegoc, line 36: I would use the same approach as in #6967 and add a new parameter in the `ServiceAccount` section of the `values.yaml`, then bump the minor version of the chart. What do you think?

pablogalegoc (commented):
Hi @jonkerj! Thanks for your contribution! I left one comment 🙂

bitnami/nginx/templates/deployment.yaml
jonkerj: Sound good, I will amend this

bitnami/nginx/values.yaml
```diff
@@ -274,6 +274,9 @@ serviceAccount:
   ## Only used if `create` is `true`.
   ##
   annotations: {}
+  ## @param serviceAccount.autoMount Auto-mount the service account token in the pod
+  ##
```
pablogalegoc, line 278: There's a trailing space here that the linter will not like 😅
```suggestion
  ##
```

pablogalegoc (commented):
There's a conflict on the chart version and left a comment on a linting issue, but other than that LGTM! Thank you very much for your work @jonkerj 🙂

bitnami/nginx/values.yaml
jonkerj: Fixed both, thanks for looking into it.

Issue #6934, opened 2021-07-13
[bitnami/external-dns] Add support for regex domain filter & exclusion


cikupin:
**Which chart**:
external-dns

**Is your feature request related to a problem? Please describe.**
Currently we cannot use regex-domain-filter on external-dns chart version 5.1.4. Meanwhile it's already supported on external-dns app version 0.8.0. 

- https://github.com/kubernetes-sigs/external-dns/blob/v0.8.0/main.go#L150
- https://github.com/kubernetes-sigs/external-dns/blob/v0.8.0/pkg/apis/externaldns/types.go#L373

**Describe the solution you'd like**
Please add regex domain filter parameters in the helm chart

```
regexDomainFilter: "your-regex-pattern"
regexDomainExclusion: "your-regex-pattern"
```

pablogalegoc:
Hi @cikupin!

This would be a good addition to the asset! Would you be open to contributing this change via a PR? We'd be glad to review and help with the release 🙂

cikupin:
@pablogalegoc sure, I'll submit the PR when it's finished

cikupin:
Hi @pablogalegoc I've just added its PR above

pablogalegoc:
Thank you very much @cikupin!

PR #6972, opened 2021-07-16
[bitnami/external-dns] Add  regexDomainFilter and regexDomainExclusion parameters

cikupin:
**Description of the change**

Currently we cannot use regex-domain-filter on external-dns chart version 5.1.4. Meanwhile it's already supported on external-dns app version 0.8.0. 

- https://github.com/kubernetes-sigs/external-dns/blob/v0.8.0/main.go#L150
- https://github.com/kubernetes-sigs/external-dns/blob/v0.8.0/pkg/apis/externaldns/types.go#L373

So that I add regexDomainFilter and regexDomainExclusion parameters to values.yaml and deployment.yaml.

**Benefits**

Users can use external dns if they want to detect domain by regex pattern instead of defining each one-by-one as an array on domainFilters.

**Possible drawbacks**

If `regexDomainFilter` is specified, `domainFilters` and `excludeDomains` will be ignored.

**Applicable issues**

- Proposed feature in #6934 

**Checklist** 


javsalgar (approved):
Thanks!! LGTM

PR #6976, opened 2021-07-16
[bitnami/nginx] unset pdb.maxUnavailable in default values

marcuz:
**Description of the change**

Unset `pdb.maxUnavailable` in default `values.yaml`.

**Benefits**

Resolves conflict introduced in https://github.com/bitnami/charts/commit/97543ce59612ccd6da1f68bee37693e32d9f47d6:

```hcl
│ Error: cannot patch "[REDACTED]" with kind PodDisruptionBudget: PodDisruptionBudget.policy "[REDACTED]" is invalid: spec: Invalid value: policy.PodDisruptionBudgetSpec{MinAvailable:(*intstr.IntOrString)(0xc00f5be560), Selector:(*v1.LabelSelector)(0xc00f5be580), MaxUnavailable:(*intstr.IntOrString)(0xc00f5be5c0)}: minAvailable and maxUnavailable cannot be both set
```

**Possible drawbacks**

N/A

**Applicable issues**

N/A

**Additional information**

Reproducible with Terraform `v1.0.2` Helm provider `2.2.0`.

**Checklist** 

### DIFF
--- a/bitnami/nginx/values.yaml
+++ b/bitnami/nginx/values.yaml
@@ -300,7 +300,7 @@ pdb:
   minAvailable: 1
   ## @param pdb.maxUnavailable Max number of pods that can be unavailable after the eviction
   ##
-  ## maxUnavailable: 1
+  maxUnavailable:
 
 ## @section Custom NGINX application parameters
 
--- a/bitnami/nginx/values.yaml
+++ b/bitnami/nginx/values.yaml
@@ -300,7 +300,7 @@ pdb:
   minAvailable: 1
   ## @param pdb.maxUnavailable Max number of pods that can be unavailable after the eviction
   ##
-  maxUnavailable:
+  maxUnavailable: 0
 
 ## @section Custom NGINX application parameters
 
--- a/bitnami/nginx/values.yaml
+++ b/bitnami/nginx/values.yaml
@@ -300,7 +300,7 @@ pdb:
   minAvailable: 1
   ## @param pdb.maxUnavailable Max number of pods that can be unavailable after the eviction
   ##
-  maxUnavailable: 1
+  ## maxUnavailable: 1
 
 ## @section Custom NGINX application parameters
 
--- a/bitnami/nginx/Chart.yaml
+++ b/bitnami/nginx/Chart.yaml
@@ -25,4 +25,4 @@ name: nginx
 sources:
   - https://github.com/bitnami/bitnami-docker-nginx
   - http://www.nginx.org
-version: 9.3.7
+version: 9.3.8


bitnami/nginx/values.yaml
```diff
@@ -300,7 +300,7 @@ pdb:
   minAvailable: 1
   ## @param pdb.maxUnavailable Max number of pods that can be unavailable after the eviction
   ##
-  maxUnavailable: 1
+  ## maxUnavailable: 1
```
pablogalegoc, line 303: We are transitioning to make every parameter of the `values.yaml` explicit:
```suggestion
  maxUnavailable:
```
In 97543ce I uncommented this parameter and inadvertently left the example value instead of leaving it `null` 🤦

pablogalegoc (commented):
Thanks for catching this @marcuz! I left one comment

bitnami/nginx/values.yaml
marcuz: > We are transitioning to make every parameter of the `values.yaml` explicit:

Oh sorry I have missed that!

marcuz:
@pablogalegoc I guess we need to take care of this too since we allow `null`:

https://github.com/bitnami/charts/blob/master/bitnami/nginx/values.schema.json#L89-L91

What's the correct approach there?

pablogalegoc:
Oh I see, will have to set it to `maxUnavailable: 0` then, like in
https://github.com/bitnami/charts/blob/7f5b0ed1942b9d2f4c109ecef3e130df61ea6a38/bitnami/mongodb-sharded/values.yaml#L705-L710
Sorry about that!

marcuz:
> Oh I see, will have to set it to `maxUnavailable: 0` then, like in
> 
> https://github.com/bitnami/charts/blob/7f5b0ed1942b9d2f4c109ecef3e130df61ea6a38/bitnami/mongodb-sharded/values.yaml#L705-L710

Wouldn't this break it again with the initial error "minAvailable and maxUnavailable cannot be both set"? So either we change the schema to allow `null` or comment `maxUnavailable` like it was initially. :thinking:

pablogalegoc:
In both cases (setting `maxUnavailable: 0` or leaving it null) the output of `kubectl get poddisruptionbudgets` is:
```bash
NAME    MIN AVAILABLE   MAX UNAVAILABLE   ALLOWED DISRUPTIONS   AGE
nginx   1               N/A               0                     15s
```

marcuz:
> In both cases (setting `maxUnavailable: 0` or leaving it null) the output of `kubectl get poddisruptionbudgets` is:
> 
> ```shell
> NAME    MIN AVAILABLE   MAX UNAVAILABLE   ALLOWED DISRUPTIONS   AGE
> nginx   1               N/A               0                     15s
> ```

You are right, that works, apologies. Should we go with `minAvailable: 1` and `maxUnavailable: 0`? It doesn't look like there is much consistency in the other charts.

pablogalegoc:
We'd go with `minAvailable: 1` and `maxUnavailable: 0`, now that almost every `values.yaml` follows this new convention we'll try to reduce the default `null` values to the minimum as a next step. Thank you!

marcuz:
@pablogalegoc it should be good to merge. Thank you! :raised_hands:

PR #6982, opened 2021-07-16
[kibana] change healthcheck endpoint to /api/status

dungdm93:
Currently, kibana healthcheck using `/app/kibana` which regularly lead to 503 or 429 (Too many requests).
So I suggest that change healthcheck endpoint to `/api/status` to avoid this issue.
Screenshot from 2021-07-16 19-55-50


dungdm93:
@alvneiayu Yeah, I already change chart version from 8.1.9 (current) to `8.1.10`

juan131 (approved):
LGTM

Issue #6510, opened 2021-05-30
[postgresql-ha] Upgrade fails if postgresql.existingSecret is used


krichter722:
**Which chart**:
bitnami/postgresql-ha

**Describe the bug**
If PostgreSQL server and reprmgr credentials are managed in a secret with the expected keys outside of the chart and that secret is passed through the chart value `postgresql.existingSecret` a chart is supposed to be upgradable afaik. However, `helm upgrade` fails due to `'postgresql.password' must not be empty`.

I could not find reliable information in the documentation, especially not in [Credential errors while upgrading chart releases](https://docs.bitnami.com/general/how-to/troubleshoot-helm-chart-issues/#credential-errors-while-upgrading-chart-releases) so I opened https://github.com/bitnami/charts-docs/issues/38 where I got the information that maintaining credentials in a secret not managed, but only used by the chart, is a valid upgrade path.

If I have to specify `postgresql.password` for the upgrade this contradicts the information I got from the linked issue comment. And it makes the chart unnecessarily hard to use with tools like Ansible. And it's simply inconvenient that one has to specify the password in both the secret and again as a chart value for upgrades.

The behavior seems to be the same for `bitnami/postgresql`.

**To Reproduce**
In form of a GitLab CI script which allows to reproduce on GKE easily:

```
variables:
  CLUSTER_NAME: "bitnami-psql-upgrade-$CI_COMMIT_SHORT_SHA-$CI_PIPELINE_IID"
  CLUSTER_VERSION: "1.19.9-gke.1900"
  CLUSTER_MASCHINE_TYPE: e2-standard-4
  CLUSTER_ZONE: us-central1-c
  HELM_VERSION: "3.5.4"

stages:
  - setup-cluster
  - setup-operations
  - tear-down-cluster

setup-cluster:
  image: gcr.io/google.com/cloudsdktool/cloud-sdk:340.0.0
  stage: setup-cluster
  script:
    - gcloud config set project $PROJECT_ID
    - gcloud auth activate-service-account --key-file "$GOOGLE_KEY"
    - gcloud config set compute/zone $CLUSTER_ZONE
    - gcloud beta container clusters create $CLUSTER_NAME --machine-type=$CLUSTER_MASCHINE_TYPE --cluster-version=$CLUSTER_VERSION --no-enable-stackdriver-kubernetes --enable-autoupgrade

setup-cluster-operations:
  image: gcr.io/google.com/cloudsdktool/cloud-sdk:340.0.0
[...]

PR #6986, opened 2021-07-17
[all] Unit tests

krichter722:
**Description of the change**
I got started with unit tests for charts only recently and would like to suggest them for the project. The scope is simple: test the templates - one file or multiple - as few as possible, as many as necessary, but of course not the entire chart and don't do anything on a k8s cluster.

**Benefits**
There're value schemes, integration tests, the helm test command, but they're not performing tasks that one might see as unit tests. You can't reach any sane test coverage with integration tests - they have their purpose, of course, but it's a different one. Unit tests simply are a different concept that has seen few attention for helm charts. They're part of quality software engineering where everything needs a test - in theory - and in theory there's no difference between theory and practice - you know the drill.

**Possible drawbacks**

More time and resources are needed to the build. Creating unit tests afterwards is painful. You should be convinced and agree on using helm unit tests in the core dev team, otherwise it's just something that you'll rollback after a few weeks.

**Applicable issues**

This might avoid issues like https://github.com/bitnami/charts/issues/6510 and a lot more (haven't been providing feedback on this one since I wanted to add coverage with unit tests).

**Additional information**
I'm promoting this kind of testing, therefore some parts of the text are copied.

I'm already running the tests on your CI. It doesn't cost much since they're blindingly fast. Please ping me if you'd like me to disable them.

**Checklist** 

### DIFF
--- a/.github/ct-install.yaml
+++ b/.github/ct-install.yaml
@@ -16,3 +16,6 @@ excluded-charts:
   - dataplatform-bp1
   - dataplatform-bp2
 remote: origin
+additional-commands:
+  - helm unittest --helm3 --strict --file tests/*.yaml {{ .Path }}
+  - helm unittest --helm3 --strict --file tests/**/*.yaml {{ .Path }}
--- a/.github/workflows/lint-test.yaml
+++ b/.github/workflows/lint-test.yaml
@@ -30,6 +30,11 @@ jobs:
       - name: Set up chart-testing
         uses: helm/chart-testing-action@v2.1.0
 
+      - name: Install helm unittest plugin
+        run: |
+          helm env
+          helm plugin install https://github.com/quintush/helm-unittest --version 0.2.6
+
       - name: Create kind cluster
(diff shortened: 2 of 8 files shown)


javsalgar:
Hi! Thank you so much for the proposal. This is something that should be discussed internally to see if we should adopt this approach. We already perform different types of testing so we should check if it makes sense to add (and also maintain) this type of testing.

krichter722:
not stale, awaiting input or close by devs

javsalgar:
After checking with the team, right now we wouldn't have the proper bandwidth to maintain an extra test suite for all the charts we have. Here you can see our current release process: https://docs.bitnami.com/kubernetes/faq/get-started/understand-charts-release-process/. We currently test all the charts in all major Kubernetes distributions, as well as different testing scenarios (TLS enabled, LDAP, Networkpolicies), so we ensure that the charts work in different situations.



We really appreciate the suggestion and we thank you for using and contributing to this repository. I proceed to close the ticket.

PR #6990, opened 2021-07-19
[bitnami/postgresql] Add option to create services for each read replica

### LINKED ISSUE #6954: [bitnam/postgresql] Allow ability to create services for each read-replica
**Which chart**:
[bitnam/postgresql]

My program manages round-robin of read-replicas by itself. Currently, being forced to use one service, I end up using only as many read-replicas as I have pods, if I get lucky and the requests go to different replicas at all.

the ability to create services for each read-replica

HA-postgresql may be an alternative, but I would still encounter the situation where I dont have individual connections, relying on pg-pool configuration.

deefdragon:
**Description of the change**
Add option to create services for each read replica (default disabled)
Add option to create a service that connects to all read replicas. (the existing read-replica service, default enabled)

**Benefits**
Allows programs that manage connection to replicas manually to do so.
Also allows for the disabling of the service that connects to all read-replicas if not needed

**Possible drawbacks**
Read-Replica unique services do not have options for LoadBalancers, Node Ports, or Setting ClusterIP.

**Applicable issues**
  - fixes #6954

**Checklist** 


javsalgar:
Thank you for the PR! We will look into it

deefdragon:
Overlooked the pattern on the values. Was not aware that was how they were updated. Just Pushed a fix. 
Might be worth denoting the proper way to update the readme on the checklist and the contribution guidelines.

bitnami/postgresql/templates/svc-read-set.yaml
```diff
@@ -0,0 +1,43 @@
+{{- if and .Values.replication.enabled .Values.replication.uniqueServices }}
+{{- $serviceAnnotations := coalesce .Values.readReplicas.service.annotations .Values.service.annotations -}}
+{{- $serviceType := coalesce .Values.readReplicas.service.type .Values.service.type -}}
+{{- $serviceLoadBalancerIP := coalesce .Values.readReplicas.service.loadBalancerIP .Values.service.loadBalancerIP -}}
```
javsalgar, line 4: Remove the variables that are not used

javsalgar (changes requested):
Thank you! Small change requested

bitnami/postgresql/templates/svc-read-set.yaml
deefdragon: resolved

javsalgar (approved):
Thank you! LGTM

deefdragon:
I committed a test file, and it looks like I need to merge in anyway. will be fixed shortly.

PR #7000, opened 2021-07-20
[bitnami/several] Fix default values and regenerate README (III)

alvneiayu:
**Description of the change**

Fix default values in values.yaml and regenerate README using the automated tool

**Checklist** 


bitnami/mysql/values.yaml
```diff
@@ -178,15 +178,15 @@ primary:
   ## @param primary.existingConfiguration Name of existing ConfigMap with MySQL Primary configuration.
   ## NOTE: When it's set the 'configuration' parameter is ignored
   ##
-  existingConfiguration:
+  existingConfiguration: {}
```
carrodher, line 181:
```suggestion
  existingConfiguration: ""
```

bitnami/mysql/values.yaml
```diff
@@ -492,15 +492,15 @@ secondary:
   ## @param secondary.existingConfiguration Name of existing ConfigMap with MySQL Secondary configuration.
   ## NOTE: When it's set the 'configuration' parameter is ignored
   ##
-  existingConfiguration:
+  existingConfiguration: {}
```
carrodher, line 495:
```suggestion
  existingConfiguration: ""
```

bitnami/mysql/values.yaml
```diff
@@ -984,24 +984,24 @@ metrics:
...
     ##
-    relabellings:
+    relabellings: []
     ## @param metrics.serviceMonitor.honorLabels Specify honorLabels parameter to add the scrape endpoint
     ##
     honorLabels: false
     ## @param metrics.serviceMonitor.release Specify the release for ServiceMonitor. Sometimes it should be custom for prometheus operator to work
     ##
-    release:
+    release: {}
```
carrodher, line 1004:
```suggestion
    release: ""
```

carrodher (commented):
Some minor changes requested, please, run the tool to update the README once changed the values (or perform the minor changes manually)

bitnami/minio/values.yaml
```diff
@@ -656,16 +656,16 @@ metrics:
...
     ##
-    relabellings:
+    relabellings: []
     ## @param metrics.serviceMonitor.honorLabels Specify honorLabels parameter to add the scrape endpoint
     ##
     honorLabels: false
     ## @param metrics.serviceMonitor.release Specify the release for ServiceMonitor. Sometimes it should be custom for prometheus operator to work
     ##
-    release:
+    release: {}
```
carrodher, line 668: We can safely remove this parameter as it is not used

carrodher:
You need to update the READMEs after the changes in eb8d3b5e32fc2c37bdeabaa91e3e774f649e4a4a

alvneiayu:
Yes yes, no worries. I had yesterday a problem with the script. Everything's under control. Thanks @carrodher

bitnami/memcached/README.md
```diff
@@ -78,19 +78,19 @@ The command removes all the Kubernetes components associated with the chart and
...
 | `command`                                     | Default container command (useful when using custom images)                                     | `[]`                         |
 | `arguments`                                   | Default container args (useful when using custom images)                                        | `[]`                         |
 | `extraEnv`                                    | Additional env vars to pass                                                                     | `[]`                         |
 | `podDisruptionBudget.create`                  | Specifies whether a Pod disruption budget should be created                                     | `false`                      |
-| `podDisruptionBudget.minAvailable`            | Minimum number of pods that need to be available                                                | `nil`                        |
+| `podDisruptionBudget.minAvailable`            | Minimum number of pods that need to be available                                                | `""`                         |
```
javsalgar, line 88: This actually should be an integer, is it something we can define in our metadata?

javsalgar (commented):
Small comment

bitnami/memcached/README.md
alvneiayu: hi @javsalgar 

The possible empty values that we have in readme-generator are "", {} or []. Other option is set up this value to 0.

https://github.com/bitnami/charts/pull/6976#issuecomment-881399315
https://github.com/bitnami/charts/pull/6976#issuecomment-881478391

Right now, the 0 is like a nil but I think that this is not a good practice. If for any reason, they change this approach (0 is nil) we are going to have a 0 instead of nil.

What do you think?

Álvaro

carrodher: I think adding a `0` to `minAvailable` or `maxAvailable` when it is not going to be used can be confused for users. Although it's true those parameters are integers instead of a string, adding a value that is not true and the disable it in the template by adding some logic it's something weird.

The _values.yaml_ may be less readable for users, in the same way, the auto-generated table from the README is going to show `0` as the value but this is not totally true since this `0` is going to act as a "false" boolean.

My 2cts are continuing using an empty string as default parameter and re-evaluate this topic with the team, if needed we can do a batch replacement with `sed` once supported integers in the readme-generator, put a `0` or whatever is the decided approach.

carrodher (approved):

PR #6906, opened 2021-07-09
[bitnami/postgresql-ha] Add diagnostic mode

javsalgar:
**Description of the change**

The following PR adds support for a new feature called Diagnostic mode. This mode deploys all the containers disabling all probes, lifecycleshooks and overriding the default command with one provided via values.yaml. This is useful for users wanting to debug and fix faulty installations or unexpected conflicts.

**Benefits**

Allows debugging and fixing issues.

**Possible drawbacks**

none known

**Applicable issues**

  - fixes #

**Additional information**

**Checklist** 

- [ ] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [ ] Variables are documented in the README.md
- [ ] Title of the PR starts with chart name (e.g. [bitnami/<name_of_the_chart>])

PR #7012, opened 2021-07-21
[bitnami/several] Add diagnostic mode

carrodher:
**Description of the change**

The following PR adds support for a new feature called Diagnostic mode. This mode deploys all the containers disabling all probes, lifecycleshooks, and overriding the default command/args with ones provided via _values.yaml_. This is useful for users wanting to debug and fix faulty installations or unexpected conflicts.

Please, note this PR is not "designing" the solution, just replicating what is already implemented in #6906 and #6936; in the same way, it is documented in https://github.com/bitnami/charts-docs/pull/54

**Assets included in this batch**
- [X] Cassandra
- [x] Consul
- [x] EJBCA
- [x] Elasticsearch
- [x] Etcd
- [x] Fluentd
- [x] Grafana Tempo
- [x] HAProxy
- [x] InfluxDB
- [x] Kafka
- [x] Kong
- [x] Logstash
- [x] MariaDB
- [x] MongoDB Sharded
- [x] MySQL
- [x] NATS
- [x] PostgreSQL
- [x] RabbitMQ
- [x] Redis Cluster
- [x] Spark
- [x] Zookeeper
- [x] WordPress

In the same way, the diagnostic mode was included in the scaffolding for new charts

--------------
I manually tested that all of them are properly rendered with `helm template foo . --set diagnosticMode.enabled=true,metrics.enabled=true`. In the same way, taking advantage of the previous command output, I checked that the `commands` and `args` are modified; also probes and lifecycles are removed


bitnami/elasticsearch/templates/metrics-deploy.yaml
```diff
@@ -48,9 +52,11 @@ spec:
             {{- if .Values.metrics.extraArgs }}
             {{- toYaml .Values.metrics.extraArgs | nindent 12 }}
             {{- end }}
+          {{- end }}
           ports:
             - name: metrics
               containerPort: 9114
+          {{- if not .Values.diagnosticMode.enabled }}
```
rafariossaa, line 59: Maybe the user want to have metrics enabled in diagnosticMode.
What do you think?

carrodher: The problem is that we can't guarantee the metric container is going to work if the service in the main container is not up and running. The idea of this mode is to provide a way for users to access the different containers. Being said that, it can be improved in the future based on the feedback provided by the users or internally, this is just a first iteration to see how it works and if it's useful

rafariossaa: OK

rafariossaa (commented):
So far, this LGTM

bitnami/nats/templates/configmap.yaml
```diff
@@ -31,9 +31,9 @@ data:
     {{- end }}
 
     # Logging options
-    debug: {{ .Values.debug.enabled }}
-    trace: {{ .Values.debug.trace }}
-    logtime: {{ .Values.debug.logtime }}
+    debug: {{ ternary "true" "false" (or .Values.debug.enabled .Values.diagnosticMode.enabled) }}
+    trace: {{ ternary "true" "false" (or .Values.debug.trace .Values.diagnosticMode.enabled) }}
+    logtime: {{ ternary "true" "false" (or .Values.debug.logtime .Values.diagnosticMode.enabled) }}
```
rafariossaa, line 36: Do we want to enable all this at the same time ?

rafariossaa (commented):
This look good to me. 
Please check a small comment on nats

bitnami/nats/templates/configmap.yaml
carrodher: Not a strong opinion, just adding as much verbosity as possible in this diagnostic mode, but to be honest I don't know what this means for NATS and if it is going to be more information than the desired one

javsalgar (approved):

PR #7032, opened 2021-07-22
[bitnami/mongodb] doc: use existingSecret from Notes

Startouf:
**Description of the change**

`auth.existingSecret` is successfully taken into consideration when starting the DB
but the notes do not reflect this to grab the password from k8s

below is an example output

```
To get the root password run:

    export MONGODB_ROOT_PASSWORD=$(kubectl get secret --namespace [namespace] mongodb
```

should become

```
To get the root password run:

    export MONGODB_ROOT_PASSWORD=$(kubectl get secret --namespace [namespace] [secret-name]
```
**Benefits**

Better debugging/access to pod

**Checklist** 


bitnami/mongodb/templates/NOTES.txt
```diff
@@ -90,7 +90,7 @@ MongoDB(R) can be accessed on the following DNS name(s) and ports from within yo
 
 To get the root password run:
 
-    export MONGODB_ROOT_PASSWORD=$(kubectl get secret --namespace {{ template "mongodb.namespace" . }} {{ template "mongodb.fullname" . }} -o jsonpath="{.data.mongodb-root-password}" | base64 --decode)
+    export MONGODB_ROOT_PASSWORD=$(kubectl get secret --namespace {{ template "mongodb.namespace" . }} {{ .Values.auth.existingSecret || template "mongodb.fullname" . }} -o jsonpath="{.data.mongodb-root-password}" | base64 --decode)
```
pablogalegoc, line 93: Hi @Startouf! Thanks for raising this improvement! This approach throws a parse error, however:
```
Error: parse error at (mongodb/templates/NOTES.txt:93): unexpected "|" in command
```
I would suggest an if/else approach like in
https://github.com/bitnami/charts/blob/07481c06d6f082595e8c9bc6151899ec10627f36/bitnami/mongodb/templates/NOTES.txt#L110-L114

Startouf: I found out a helper already existed (and it is the same across many charts) so I just used it in the notes instead

pablogalegoc:
@Startouf we'll need a version bump to `10.23.10`, other than that LGTM! Thanks!

PR #7033, opened 2021-07-22
[bitnami/mongodb-sharded] Fix wrong index in servicePerReplica context

lucchmielowski:
**Description of the change**

This PR fixes an issue with the `$index` that keeps the value `1` after the first 2 iterations because of the way the precedence is done in Helm when using `merge`, causing things like : 

```yaml
metadata:
  name: RELEASE-NAME-mongodb-sharded-3
  labels: 
    app.kubernetes.io/name: mongodb-sharded
  annotations:
    external-dns.alpha.kubernetes.io/hostname: mongo-1.test # The .index used here is wrong
```

**Benefits**

**Possible drawbacks**

**Applicable issues**

**Additional information**

I thought it would better to create a copy of the object instead of mutating it throughout the loops, but using `mergeOverwrite` would work too 🤷

**Checklist** 

### DIFF
--- a/bitnami/mongodb-sharded/Chart.yaml
+++ b/bitnami/mongodb-sharded/Chart.yaml
@@ -26,4 +26,4 @@ name: mongodb-sharded
 sources:
   - https://github.com/bitnami/bitnami-docker-mongodb-sharded
   - https://mongodb.org
-version: 3.7.4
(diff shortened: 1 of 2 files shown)


javsalgar (approved):
Thanks for the PR! LGTM

Issue #6775, opened 2021-06-27
Kafka - Datadog intergrations issues


asaf-upstream:
**chart**:
kafka - 12.9.0

**Describe the bug**
I suspect the DD-Kafka integration is not working using the chart.
Tried to use the following guide:
https://docs.datadoghq.com/agent/guide/autodiscovery-with-jmx/?tab=containeragent

WIthout success.
We run the kafka chart on 'kafka' namespace, and DD chart on 'monitoring'.

Tried to add the following podAnnotations:
      ad.datadoghq.com/jmx-exporter.check_names: '["kafka"]'
      ad.datadoghq.com/jmx-exporter.init_configs: '[{"is_jmx": true, "collect_default_metrics": true}]'
      ad.datadoghq.com/jmx-exporter.instances: '[{"host": "%%host%%","port":"5556","namespace":"kafka"}]'
      ad.datadoghq.com/kafka.logs: '[{"source":"kafka","service":"kafka"}]'

      ad.datadoghq.com/kafka.check_names: '["kafka"]'
      ad.datadoghq.com/kafka.init_configs: '[{"is_jmx": true, "collect_default_metrics": true}]'
      ad.datadoghq.com/kafka.instances: '[{"host": "%%host%%","port":"5556","namespace":"kafka"}]'
      ad.datadoghq.com/kafka.logs: '[{"source":"kafka","service":"kafka"}]'

      - name: POD_IP
        valueFrom:
          fieldRef:
            fieldPath: status.podIP
      - name: JAVA_OPTS
        value: >-
          -Dcom.sun.management.jmxremote
          -Dcom.sun.management.jmxremote.authenticate=false
          -Dcom.sun.management.jmxremote.ssl=false
          -Dcom.sun.management.jmxremote.local.only=false
          -Dcom.sun.management.jmxremote.port=5556
          -Dcom.sun.management.jmxremote.rmi.port=5556
          -Djava.rmi.server.hostname=$(POD_IP) \ 0.0.0.0\ 127.0.0.1

Output of the DD agent proves no data is collected:
========
JMXFetch
========

  Information
  ==================
  Initialized checks
  ==================
    no checks

  Failed checks
  =============
    no checks

If I use 'kubectl port-forward' on the kafka pod, I can see the metrics on jconsole.

**Expected behavior**
Monitoring on DD should be presented.

- Output of `helm version`:
[...]

PR #7046, opened 2021-07-23
[bitnami/kafka] adding datadog integration to bitnami kafka chart

soudaburger:
**Description of the change**

This change adds additional customization to the Kafka Helm chart to support Datadog Autodiscovery.

Problem:  Default chart configuration hard-codes `metrics` named port, so the service can't be monitored via autodiscovery because the backend named port points to the prometheus exporter, not the ad-hoc jmx exporter.

Solution:
1. Added `metrics.jmx.port` which would modify the named `metrics` port, if needed.  Defaults to `5556`.
2. Added `metrics.jmx.prometheus.port` which should never really need to be changed, but wanted to explicitly call it out to help others.  Defaults to `5556`.
3. Add instructions in README for configuring the Datadog Integration with Autodiscovery and this helm chart.

Used the conversation in https://github.com/bitnami/charts/issues/6775 as a starting point.

**Checklist** 
- [ X] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [ X] Title of the PR starts with chart name (e.g. [bitnami/<name_of_the_chart>])


bitnami/kafka/Chart.yaml
```diff
@@ -29,4 +29,4 @@ name: kafka
 sources:
   - https://github.com/bitnami/bitnami-docker-kafka
   - https://kafka.apache.org/
-version: 13.0.4
+version: 13.0.5
```
carrodher, line 32: As it's a new feature more than a fix, can you bump the minor version instead of the patch?
```suggestion
version: 13.1.0
```

bitnami/kafka/README.md
```diff
@@ -385,6 +387,37 @@ It is strongly recommended to use immutable tags in a production environment. Th
 
 Bitnami will release a new chart updating its containers if a new version of the main container, significant changes, or critical vulnerabilities exist.
 
+### Datadog Autodiscovery Integration
```
carrodher, line 390: Can you add here a paragraph with a basic explanation apart from the code block?

carrodher (commented):
Thanks for your contribution, I just added a couple of comments

carrodher (commented):
Thanks for the changes, I have the last question, there is a configuration file defined around line 1137 in the values.yaml that is using the JMX port 5555 instead of 5556 that is something you're changing in the statefulset (line 272). Can this topic produce any issues? See https://github.com/bitnami/charts/blob/6340bff66f93c8c797bda3ca0842e4bf770059f1/bitnami/kafka/values.yaml#L1137

soudaburger:
> Thanks for the changes, I have the last question, there is a configuration file defined around line 1137 in the values.yaml that is using the JMX port 5555 instead of 5556 that is something you're changing in the statefulset (line 272). Can this topic produce any issues? See
> 
> https://github.com/bitnami/charts/blob/6340bff66f93c8c797bda3ca0842e4bf770059f1/bitnami/kafka/values.yaml#L1137

So, I haven't noticed any changes with regards to that line.  I don't have any way, that I know of, to test that part of the JMX functionality.  It might simply be that that port should match whatever we're setting in the `values.yaml`, but honestly not sure.  Perhaps someone else could chime in?

soudaburger:
Well, I feel like an idiot now.  After the last comment by @carrodher, I started poking at the RMI configuration that already exists the kafka values.yaml.  Between that and the JMX_PORT, I already had what I needed, just didn't know it.

To recap what you need in order to get metrics into Datadog (now):
```yaml
# Enable JMX Metrics
metrics:
  jmx:
    enabled: true
# Configure pod annotations
podAnnotations:
  ad.datadoghq.com/jmx-exporter.check_names: '["kafka"]'
  ad.datadoghq.com/jmx-exporter.init_configs: '[{"is_jmx": true, "collect_default_metrics": true}]'
  ad.datadoghq.com/jmx-exporter.instances: '[{"host": "%%host%%","port":"5555","namespace":"kafka"}]'
```

The key point is that you can NOT customize that `5555` port, because that's tied to `JMX_PORT` in the statefulset (hard-coded) and the metrics.jmx.config RMI URL, which you wouldn't be able to parameterize anyways.

Thanks for the assist! This can be closed.

bitnami/kafka/README.md
```diff
@@ -311,10 +311,12 @@ The command removes all the Kubernetes components associated with the chart and
 | `metrics.jmx.image.tag`                          | JMX exporter image tag (immutable tags are recommended)                                                                          | `0.16.0-debian-10-r7`    |
 | `metrics.jmx.image.pullPolicy`                   | JMX exporter image pull policy                                                                                                   | `IfNotPresent`           |
 | `metrics.jmx.image.pullSecrets`                  | Specify docker-registry secret names as an array                                                                                 | `[]`                     |
+| `metrics.jmx.port`                               | Specify port for named `metrics` port.                                                                                           | `5556`                   |
```
carrodher, line 314: Now the new value should be updated in the README as well
```suggestion
| `metrics.jmx.port`                               | Specify port for named `metrics` port.                                                                                           | `5555`                   |
```

carrodher:
Changes LGTM, although I would like to see the opinion from another teammate. What do you think about this PR @juan131?

soudaburger:
> Changes LGTM, although I would like to see the opinion from another teammate. What do you think about this PR @juan131?

My two cents is that this PR is no longer needed.

carrodher:
Yeah, I just saw your last comment, is it needed to make the port customizable in the statefulset? Maybe we can use the PR just to implement that change (if it is still needed for your use case, of course)

soudaburger:
No, it's not necessary.  While it would be nice, you'd still have to make manual changes to the `values.yaml` anyways to reconcile the RMI configuration.  So, i think the small code snipped above that I pasted is all that's needed and should be sufficient for people.

carrodher:
Perfect, thanks for confirming and for sharing the code! I'm closing the PR then, thanks again

PR #7049, opened 2021-07-24
[bitnami/thanos] Add hash and time base partion for thanos store

### LINKED ISSUE #4417: [bitnami/thanos] add support for sharding in storegateway
**Which chart**:

bitnami/thanos

By default, the thanos-storegateway will advertise that it can serve all timeseries and all blocks in the object store (ie S3-bucket). As such, you can not scale the number of stores and register them all with thanos-query. Thanos-query will complain about multiple stores that advertise the same data with non-unique labels. The storegateway can be put behind a loadbalancer, such as a kubernetes Service (of type ClusterIP), but the fact that communication happens over long lived gRPC connections means that, most of the time, there is no guarantee that each storegateway replica will serve an equal amount of requests.

This is a known issue and [documented on the thanos website](https://thanos.io/tip/thanos/sharding.md/) and can be solved by introducing sharding (aka partitioning) in the storegateway. Sharding allows each replica of the storegateway to only serve some blocks but not others. A request that spans multiple blocks will then automatically be fanned out to multiple stores.

Introduce the configuration of optional time-based or hash-based partitioning (each nth-block). Default the store to use a single partition. The storegateway could be converted to a deployment with an index suffix, like so:

|desc|deployment name|
|---|---|
|first partition|thanos-storegateway-0
|second partition|thanos-storegateway-1
|third partition|thanos-storegateway-2

Store could also remain a statefulset, although I am not sure why it is a statefulset at the moment.

**Introducing a partition suffix to the store statefulset/deployment will cause it to be re-created. This could be considered a backwards incompatible change.**

Use the thanos chart from banzai-cloud, which already supports sharding. The banzai chart seems less actively maintained.
https://github.com/banzaicloud/banzai-charts/blob/master/thanos/values.yaml#L174

**Additional context**

|desc|link|
|---|---|
| Official documentation | https://thanos.io/tip/thanos/sharding.md/
| CNCF slack discussion in #thanos | https://cloud-native.slack.com/archives/CK5RSSC10/p1604923594031700
| Example implementation | https://github.com/observatorium/deployments/blob/bf1304b0d7bce2ae3fefa80412bb358f9aa176fb/environments/openshift/manifests/observatorium-template.yaml#L1514-L1521

psandhu79:
**Description of the change**

This adds hash and time based partion for the storegateway as documented here
https://thanos.io/tip/components/store.md/#time-based-partitioning
https://thanos.io/tip/thanos/sharding.md/#relabelling
 

**Benefits**

This will allow the thanos query to talk to multiple thanos store gateways which contain the specified timeseries databases

**Applicable issues**

  - fixes #4417
  - previous pr #6909

**Checklist** 


alemorcuq:
Thanks for the PR. Give me some time to test it and I will review it.

bitnami/thanos/README.md
```diff
@@ -851,6 +851,29 @@ You can specify the Ruler configuration using the `ruler.config` parameter.
 
 In addition, you can also set an external ConfigMap with the configuration file. This is done by setting the `ruler.existingConfigmap` parameter. Note that this will override the previous option.
 
+### Store time partions
+
+Thanos store supports partition based on time.
+
+Setting time partitions will create n number of store deployment based on the number of items in the list. Each item must contain min and max time for querying in the supported format (see details here See details at https://thanos.io/tip/components/store.md/#time-based-partioning ).
```
juan131, line 858: This link is not correctly formatted:
```suggestion
Setting time partitions will create N number of store deployment based on the number of items in the list. Each item must contain min and max time for querying in the supported format (see details at [Thanos documentation](https://thanos.io/tip/components/store.md/#time-based-partioning)).
```

bitnami/thanos/README.md
```diff
@@ -851,6 +851,29 @@ You can specify the Ruler configuration using the `ruler.config` parameter.
...
+
+Leaving this empty list ([]) will create a single store for all data.
```
juan131, line 860:
```suggestion
> Note: leaving this empty list ([]) will create a single store for all data.
```

bitnami/thanos/README.md
```diff
@@ -851,6 +851,29 @@ You can specify the Ruler configuration using the `ruler.config` parameter.
...
+
+Example - This will create 3 stores:
```
juan131, line 862:
```suggestion
For instance, to use 3 stores you can use a **values.yaml** like the one below:
```

bitnami/thanos/values.yaml
```diff
@@ -1834,6 +1847,18 @@ storegateway:
     ## @param storegateway.pdb.maxUnavailable Maximum number/percentage of pods that may be made unavailable
     ##
     maxUnavailable: ""
+  ## Sharded parameters
+  ## @param storegateway.sharded.enabled Enable sharding for Thanos Store Gateway
+  ## @param storegateway.sharded.hashPartioning.shards Setting hashPartioning will create multiple store deployments based on the number of shards specified using the hashmod of the blocks
+  ## @param storegateway.sharded.timePartioning [array] Setting time timePartioning will create multipale store deployments based on the number of partions
```
juan131, line 1853: Fix a couple of typos:
```suggestion
  ## @param storegateway.sharded.hashPartioning.shards Setting hashPartioning will create multiple store statefulsets based on the number of shards specified using the hashmod of the blocks
  ## @param storegateway.sharded.timePartioning [array] Setting time timePartioning will create multiple store deployments based on the number of partitions
```

bitnami/thanos/templates/storegateway/statefulset-sharded.yaml
```diff
@@ -0,0 +1,257 @@
...
+{{- else }}
+  {{- $shards = len .Values.storegateway.sharded.timePartioning }}
+  {{- $timePartioning = true }}
+{{- end }}
+
+{{- range $index, $_ := until $shards }}
+apiVersion: {{ include "common.capabilities.statefulset.apiVersion" $ }}
+kind: StatefulSet
+metadata:
+  name: {{ include "common.names.fullname" $ }}-storegateway-{{ $index }}
```
juan131, line 19: We should truncate this to avoid exceeding the 63 characters DNS limit:
```suggestion
  name: {{ printf "%s-storegateway-%s" (include "common.names.fullname" $) $index | trunc 63 | trimSuffix "-" }}
```

bitnami/thanos/templates/storegateway/statefulset-sharded.yaml
```diff
@@ -0,0 +1,257 @@
...
+  labels: {{- include "common.labels.standard" $ | nindent 4 }}
+    app.kubernetes.io/component: storegateway
+    {{- if $.Values.commonLabels }}
+    {{- include "common.tplvalues.render" ( dict "value" .Values.commonLabels "context" $ ) | nindent 4 }}
```
juan131, line 23:
```suggestion
    {{- include "common.tplvalues.render" ( dict "value" $.Values.commonLabels "context" $ ) | nindent 4 }}
```

bitnami/thanos/templates/storegateway/statefulset-sharded.yaml
```diff
@@ -0,0 +1,257 @@
...
+    {{- end }}
+spec:
+  replicas: {{ $.Values.storegateway.replicaCount }}
+  podManagementPolicy: {{ $.Values.storegateway.podManagementPolicy }}
+  serviceName: {{ include "common.names.fullname" $ }}-storegateway-headless
```
juan131, line 28: Same here. Truncate:
```suggestion
  serviceName: {{ printf "%s-storegateway-headless-%s" (include "common.names.fullname" $) $index | trunc 63 | trimSuffix "-" }}
```

bitnami/thanos/templates/storegateway/statefulset-sharded.yaml
```diff
@@ -0,0 +1,257 @@
...
+    metadata:
+      labels: {{- include "common.labels.standard" $ | nindent 8 }}
+        app.kubernetes.io/component: storegateway
+        shard: {{ $index | quote }}
+        {{- if $.Values.commonLabels }}
+        {{- include "common.tplvalues.render" ( dict "value" .Values.commonLabels "context" $ ) | nindent 8 }}
+        {{- end }}
+        {{- if $.Values.storegateway.podLabels }}
+        {{- include "common.tplvalues.render" (dict "value" .Values.storegateway.podLabels "context" $) | nindent 8 }}
+        {{- end }}
```
juan131, line 48:
```suggestion
        {{- if $.Values.commonLabels }}
        {{- include "common.tplvalues.render" ( dict "value" $.Values.commonLabels "context" $ ) | nindent 8 }}
        {{- end }}
        {{- if $.Values.storegateway.podLabels }}
        {{- include "common.tplvalues.render" (dict "value" $.Values.storegateway.podLabels "context" $) | nindent 8 }}
        {{- end }}
```

juan131 (changes requested):
Thanks so much for this great contribution! We really appreciate it!!

I added some minor corrections.

I wonder if we could merge `templates/storegateway/statefulset-sharded.yaml` and `templates/storegateway/statefulset.yaml` into a single file.. Or do you think it will be very complex due to a big quantity of `if/else/range` blocks?

psandhu79:
> Thanks so much for this great contribution! We really appreciate it!!
> 
> I added some minor corrections.
> 
> I wonder if we could merge `templates/storegateway/statefulset-sharded.yaml` and `templates/storegateway/statefulset.yaml` into a single file.. Or do you think it will be very complex due to a big quantity of `if/else/range` blocks?

In the original PR i combined it in to one file , and was asked to split into two files

bitnami/thanos/templates/storegateway/statefulset-sharded.yaml
psandhu79: would this not break, the service does not have an index

alemorcuq:
Thanks for the PR, @psandhu79! It seems there are a lot of typos here (partions instead of partitions). Could you check them?

mhulscher:
thanos-query uses a persistent grpc connection to the stores. If **all** stores/shards are behind a single (ClusterIP) Service, then each thanos-query will connect to only 1 store, which only advertises the blocks from its shard. Thanos-query will not know where to find blocks from the other shards. I think that each shard should have its own ClusterIP Service, and that thanos-query should connect to each of these services. WDYT?

psandhu79:
Does the headless service not solve this problem ?

mhulscher:
I don't believe a headless service will be enough. Thanos-query fans out queries to ALL connected stores. It then deduplicates the results based on a replica label. For example thanos-query will fan out a query to all thanos-sidecars inside a cluster, and then deduplicate the result based on the promeheus_replica label. However, stores do not return results with a unique externalLabel, like the thanos-sidecar does so there is no externalLabel to deduplicate results by.

This also makes sense; we don't want thanos-query to fan out a query to ALL thanos-stores, because the thanos-stores all look at the same S3-bucket. If thanos-query did fan out across all stores, then all stores would execute the same query and download the same blocks from S3, that's not something we want.

If the stores are sharded, then they advertise to thanos-query which blocks they can be queried for. Again, if thanos-query directly connects to all replicas of a shard, then there will be duplicate results returned.

If we create a ClusterIP Service per shard, then thanos-query will end up connecting to a single replica of each store shard. If the connection to a store is broken, query will connect to the service and find another replica of the same shard.
In the end each thanos-query will connect to 1 replica of each shard; which is what we want.

psandhu79:
Ok , I will need help in how to implement this

psandhu79:
@migruiz4 Should this service shard be in a different file too ?

bitnami/thanos/templates/storegateway/statefulset-sharded.yaml
juan131: Then we need to figure out how to trim the name while keeping the index. For instance:
```
  {{- $svcNamePrefix :=  printf "%s-storegateway-headless"  (include "common.names.fullname" $) | trunc 61 | trimSuffix "-" }}
  serviceName: {{ printf "%s-%s" $svcNamePrefix }}
```

juan131:
> In the original PR i combined it in to one file , and was asked to split into two files

All right, I guess it can be very confusing (specially for reviewers) if we put everything together in a single template. Let's go with the 2 templates approach.

> @migruiz4 Should this service shard be in a different file too ?

There would be N services, right? I mean, one service per "store statefulset", right? If that's the case, I vote for having a separate template to define these "shard services".

juan131:
By the way, I see a lot of references to `partions` (typo) instead of `partitions` as @alemorcuq mentioned before:

> Thanks for the PR, @psandhu79! It seems there are a lot of typos here (partions instead of partitions). Could you check them?

Please double-check that.

psandhu79:
> By the way, I see a lot of references to `partions` (typo) instead of `partitions` as @alemorcuq mentioned before:
> 
> > Thanks for the PR, @psandhu79! It seems there are a lot of typos here (partions instead of partitions). Could you check them?
> 
> Please double-check that.

Should i also change Partioning to partitioning

psandhu79:
> 
> > @migruiz4 Should this service shard be in a different file too ?
> 
> There would be N services, right? I mean, one service per "store statefulset", right? If that's the case, I vote for having a separate template to define these "shard services".

What should I do about clusterIP,loadBalancerIP and nodePort i would have to make these a range, i guess which will be a breaking change

juan131:
> Should i also change Partioning to partitioning

Yes

> What should I do about clusterIP,loadBalancerIP and nodePort i would have to make these a range, i guess which will be a breaking change

Yes, if we're having an array of services, then these properties should be arrays (only when "sharding" is used). Check how we did it with properties such as `externalAccess.service.loadBalancerIPs` in the [Kafka chart](https://github.com/bitnami/charts/tree/master/bitnami/kafka).

psandhu79:
> > Should i also change Partioning to partitioning
> 
> Yes
> 
> > What should I do about clusterIP,loadBalancerIP and nodePort i would have to make these a range, i guess which will be a breaking change
> 
> Yes, if we're having an array of services, then these properties should be arrays (only when "sharding" is used). Check how we did it with properties such as `externalAccess.service.loadBalancerIPs` in the [Kafka chart](https://github.com/bitnami/charts/tree/master/bitnami/kafka).

In this case externalAcess properties are only in one template, do you want me to create new properties just for the sharded service.

for e.g for the normal service
Values.storegateway.service.clusterIP 

for the sharded service
Values.storegateway.sharded.service.clusterIP 

so i do not affect the existing service

bitnami/thanos/README.md
```diff
@@ -851,6 +851,29 @@ You can specify the Ruler configuration using the `ruler.config` parameter.
 
 In addition, you can also set an external ConfigMap with the configuration file. This is done by setting the `ruler.existingConfigmap` parameter. Note that this will override the previous option.
 
+### Store time partitions
+
+Thanos store supports partion based on time.
+
+Setting time partitions will create N number of store deployment based on the number of items in the list. Each item must contain min and max time for querying in the supported format (see details at [Thanos documentation](https://thanos.io/tip/components/store.md/#time-based-partitioning)).
```
juan131, line 858:
```suggestion
Setting time partitions will create N number of store statefulsets based on the number of items in the `timePartitioning` list. Each item must contain the min and max time for querying in the supported format (find more details at [Thanos documentation](https://thanos.io/tip/components/store.md/#time-based-partitioning)).
```

bitnami/thanos/README.md
```diff
@@ -851,6 +851,29 @@ You can specify the Ruler configuration using the `ruler.config` parameter.
...
+
+> Note: leaving this empty list ([]) will create a single store for all data.
```
juan131, line 860:
```suggestion
> Note: leaving the `timePartitioning` list empty (`[]`) will create a single store for all data.
```

bitnami/thanos/templates/storegateway/service-sharded.yaml
```diff
@@ -0,0 +1,67 @@
...
+    app.kubernetes.io/component: storegateway
+    {{- if and .Values.metrics.enabled .Values.metrics.serviceMonitor.enabled }}
+    prometheus-operator/monitor: 'true'
+    {{- end }}
+    {{- if .Values.commonLabels }}
+    {{- include "common.tplvalues.render" ( dict "value" .Values.commonLabels "context" $ ) | nindent 4 }}
+    {{- end }}
+  {{- if .Values.storegateway.service.annotations }}
+  annotations: {{- include "common.tplvalues.render" (dict "value" .Values.storegateway.service.annotations "context" $) | nindent 4 }}
+  {{- end }}
```
juan131, line 26: Be careful here.. The context changes since you're in a loop:
```suggestion
  name: {{ include "common.names.fullname" $ }}-storegateway
  labels: {{- include "common.labels.standard" $ | nindent 4 }}
    app.kubernetes.io/component: storegateway
    {{- if and $.Values.metrics.enabled $.Values.metrics.serviceMonitor.enabled }}
    prometheus-operator/monitor: 'true'
    {{- end }}
    {{- if $.Values.commonLabels }}
    {{- include "common.tplvalues.render" ( dict "value" $.Values.commonLabels "context" $ ) | nindent 4 }}
    {{- end }}
  {{- if $.Values.storegateway.service.annotations }}
  annotations: {{- include "common.tplvalues.render" (dict "value" $.Values.storegateway.service.annotations "context" $) | nindent 4 }}
  {{- end }}
```
This applies to the whole file.

bitnami/thanos/templates/storegateway/statefulset-sharded.yaml
juan131: Please review this

bitnami/thanos/values.yaml
```diff
@@ -1834,6 +1847,18 @@ storegateway:
...
+  ## @param storegateway.sharded.hashPartitioning.shards Setting hashPartitioning will create multiple store statefulsets based on the number of shards specified using the hashmod of the blocks
+  ## @param storegateway.sharded.timePartitioning [array] Setting time timePartitioning will create multiple store deployments based on the number of partitions
+  ##
+  sharded:
+    enabled: false
+    hashPartitioning:
+      shards: ""
+    timePartitioning:
+      - min: ""
+        max: ""
```
juan131, line 1861: I'm missing the `storegateway.sharded.service` parameters, right?

juan131 (changes requested):
> In this case externalAcess properties are only in one template, do you want me to create new properties just for the sharded service.
>
> for the normal service `storegateway.service.clusterIP`
> 
> for the sharded service `storegateway.sharded.service.clusterIPs`
>
> so i do not affect the existing service
>
> I have submitted an update on where i am heading

That sounds great to me.

bitnami/thanos/values.yaml
psandhu79: yes i will add these , my last push was just to show the direction i was heading

psandhu79: pushed

bitnami/thanos/templates/storegateway/service-sharded.yaml
```diff
@@ -0,0 +1,68 @@
...
+    {{- include "common.tplvalues.render" ( dict "value" $.Values.commonLabels "context" $ ) | nindent 4 }}
+    {{- end }}
+  {{- if $.Values.storegateway.service.annotations }}
+  annotations: {{- include "common.tplvalues.render" (dict "value" $.Values.storegateway.service.annotations "context" $) | nindent 4 }}
+  {{- end }}
+spec:
+  type: {{ $.Values.storegateway.service.type }}
+  {{- if not (empty $.Values.storegateway.sharded.service.clusterIPs) }}
+  clusterIP: {{ index $.Values.storegateway.sharded.service.clusterIPs $index }}
+  {{- end }}
```
juan131, line 31: I guess we can find an issue if the `storegateway.sharded.service.clusterIPs` array length is different from `storegateway.sharded.timepartitioning` value or `storegateway.sharded.hashpartitioning.shards` array length. Sth similar will happen with `loadBalancerIPs`, `nodePorts` or `loadBalancerSourceRanges`.

I would include a new validation in the `thanos.validateValues` macro to show an error when this happens.

- https://github.com/bitnami/charts/blob/master/bitnami/thanos/templates/_helpers.tpl#L184

bitnami/thanos/values.yaml
```diff
@@ -1834,6 +1847,45 @@ storegateway:
...
+  ## Example:
+  ## nodePorts:
+  ##   - 30001
+  ##   - 30002
+  ## @param storegateway.sharded.service.grpc.nodePorts Array of grpc node ports used for Store Gateway service. Length must be the same as the number of shards
+  ## Example:
+  ## nodePorts:
+  ##   - 30011
+  ##   - 30012
+  ##
```
juan131, line 1874: All this metadata is missing in the **README.md**, you can use our [readme generator](https://github.com/bitnami-labs/readme-generator-for-helm) tool to automatically update it based on this metadata for you.

bitnami/thanos/templates/storegateway/statefulset-sharded.yaml
```diff
@@ -0,0 +1,257 @@
...
+  name: {{ printf "%s-storegateway-%s" (include "common.names.fullname" $) $index | trunc 63 | trimSuffix "-" }}
+  labels: {{- include "common.labels.standard" $ | nindent 4 }}
+    app.kubernetes.io/component: storegateway
+    {{- if $.Values.commonLabels }}
+    {{- include "common.tplvalues.render" ( dict "value" $.Values.commonLabels "context" $ ) | nindent 4 }}
+    {{- end }}
+spec:
+  replicas: {{ $.Values.storegateway.replicaCount }}
+  podManagementPolicy: {{ $.Values.storegateway.podManagementPolicy }}
+  serviceName: {{ printf "%s-storegateway-%s" (include "common.names.fullname" $) $index | trunc 63 | trimSuffix "-" }}
```
juan131, line 28:
```suggestion
  {{- $svcNamePrefix := printf "%s-storegateway" (include "common.names.fullname" $) | trunc 61 | trimSuffix "-" }}
  serviceName: {{ printf "%s-%s" $svcNamePrefix }}
```

bitnami/thanos/templates/storegateway/service-sharded.yaml
```diff
@@ -0,0 +1,68 @@
...
+  {{- $shards = int .Values.storegateway.sharded.hashpartitioning.shards }}
+{{- else }}
+  {{- $shards = len .Values.storegateway.sharded.timepartitioning }}
+{{- end }}
+
+{{- range $index, $_ := until $shards }}
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ printf "%s-storegateway-%s" (include "common.names.fullname" $) $index | trunc 63 | trimSuffix "-" }}
```
juan131, line 15:
```suggestion
  {{- $svcNamePrefix := printf "%s-storegateway" (include "common.names.fullname" $) | trunc 61 | trimSuffix "-" }}
  name: {{ printf "%s-%s" $svcNamePrefix }}
```

juan131 (changes requested):
Some minor changes and we're ready to go

bitnami/thanos/Chart.yaml
```diff
@@ -28,4 +28,4 @@ name: thanos
 sources:
   - https://github.com/bitnami/bitnami-docker-thanos
   - https://thanos.io
-version: 5.2.7
+version: 5.3.0
```
juan131, line 31: Add a new line character at the end of the file to avoid the linter to complain:
```suggestion
version: 5.3.0

```

bitnami/thanos/templates/_helpers.tpl
```diff
@@ -186,6 +186,10 @@ Compile all warnings into a single message, and call fail.
 {{- $messages := append $messages (include "thanos.validateValues.objstore" .) -}}
 {{- $messages := append $messages (include "thanos.validateValues.ruler.alertmanagers" .) -}}
 {{- $messages := append $messages (include "thanos.validateValues.ruler.config" .) -}}
+{{- $messages := append $messages (include "thanos.validateValues.sharded.service.clusterIPs" .) -}}
+{{- $messages := append $messages (include "thanos.validateValues.sharded.service.loadBalancerIPs" .) -}}
+{{- $messages := append $messages (include "thanos.validateValues.sharded.service.http.nodePorts" .) -}}
+{{- $messages := append $messages (include "thanos.validateValues.sharded.service.grpc.nodePorts" .) -}}
```
juan131, line 192: We can merge everything in a single validation:
```suggestion
{{- $messages := append $messages (include "thanos.validateValues.sharded.service" .) -}}
```

bitnami/thanos/templates/_helpers.tpl
```diff
@@ -238,6 +242,67 @@ thanos: ruler configuration
...
+
+{{/* Validate values of Thanos - number of sharded service grpc.nodePorts */}}
+{{- define "thanos.validateValues.sharded.service.grpc.nodePorts" -}}
+{{- if and .Values.storegateway.sharded.enabled (not (empty .Values.storegateway.sharded.service.grpc.nodePorts) ) -}}
+{{- if eq "false" (include "thanos.validateProperty" (dict "property" $.Values.storegateway.sharded.service.grpc.nodePorts "context" $) ) }}
+thanos: storegateway sharded service grpc.nodePorts
+    The number of shards does not match the number of grpc.nodePorts $.Values.storegateway.sharded.service.grpc.nodePorts
+{{- end -}}
+{{- end -}}
+{{- end -}}
```
juan131, line 283: Again, let's merge everything in a single validation:
```suggestion
{{/* Validate values of Thanos - number of sharded service properties */}}
{{- define "thanos.validateValues.sharded.service" .) -}}
{{- if and .Values.storegateway.sharded.enabled (not (empty .Values.storegateway.sharded.service.clusterIPs) ) -}}
{{- if eq "false" (include "thanos.validateValues.storegateway.sharded.length" (dict "property" $.Values.storegateway.sharded.service.clusterIPs "context" $) ) }}
thanos: storegateway.sharded.service.clusterIPs
    The number of shards does not match the number of ClusterIPs $.Values.storegateway.sharded.service.clusterIPs
{{- end -}}
{{- end -}}
{{- if and .Values.storegateway.sharded.enabled (not (empty .Values.storegateway.sharded.service.loadBalancerIPs) ) -}}
{{- if eq "false" (include "thanos.validateValues.storegateway.sharded.length" (dict "property" $.Values.storegateway.sharded.service.loadBalancerIPs "context" $) ) }}
thanos: storegateway.sharded.service.loadBalancerIPs
    The number of shards does not match the number of loadBalancerIPs $.Values.storegateway.sharded.service.loadBalancerIPs
{{- end -}}
{{- end -}}
{{- if and .Values.storegateway.sharded.enabled (not (empty .Values.storegateway.sharded.service.http.nodePorts) ) -}}
{{- if eq "false" (include "thanos.validateValues.storegateway.sharded.length" (dict "property" $.Values.storegateway.sharded.service.http.nodePorts "context" $) ) }}
thanos: storegateway.sharded.service.http.nodePorts
    The number of shards does not match the number of http.nodePorts $.Values.storegateway.sharded.service.http.nodePorts
{{- end -}}
{{- end -}}
{{- if and .Values.storegateway.sharded.enabled (not (empty .Values.storegateway.sharded.service.grpc.nodePorts) ) -}}
{{- if eq "false" (include "thanos.validateValues.storegateway.sharded.length" (dict "property" $.Values.storegateway.sharded.service.grpc.nodePorts "context" $) ) }}
thanos: storegateway.sharded.service.grpc.nodePorts
    The number of shards does not match the number of grpc.nodePorts $.Values.storegateway.sharded.service.grpc.nodePorts
{{- end -}}
{{- end -}}
{{- end -}}
```

bitnami/thanos/templates/_helpers.tpl
```diff
@@ -238,6 +242,67 @@ thanos: ruler configuration
...
+
+false
+{{- end }}
+
+
+
+{{- end }}
+
+
+
```
juan131, line 305: Let's use a more meaningful name:
```suggestion
{{- define "thanos.validateValues.storegateway.sharded.length" -}}
{{/* Get number of shards */}}
{{- $shards := int 0 }}
{{- if .context.Values.storegateway.sharded.hashPartitioning.shards }}
  {{- $shards = int .context.Values.storegateway.sharded.hashPartitioning.shards }}
{{- else }}
  {{- $shards = len .context.Values.storegateway.sharded.timePartitioning }}
{{- end }}
{{- $propertyLength := (len .property) -}}
{{/* Validate property */}}
{{- if ne $shards $propertyLength -}}
false
{{- end }}
{{- end }}
```

juan131 (approved):
Adding some final suggestions for simplicity! And approving the changes

juan131:
The `bitnami-bot` will land your changes once our internal tests pass successfully! Thanks so much (again) for this great contribution.

PR #7055, opened 2021-07-26
[bitnami/postgresql] Add lifecycleHooks value to the postgresql chart

ddi-wtag:
**Description of the change**

This PR allows a user to pass [lifecycle hooks](https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/#container-hooks) to the PostgreSQL container.

Sample `values.yaml`:

```yaml
# This would allow the PostgreSQL container to do a fast shutdown instead of a smart shutdown
# 
# See https://www.postgresql.org/docs/current/server-shutdown.html for more info
lifecycleHooks:
  preStop:
    exec:
      command: 
      - pg_ctl
      - stop
```

**Benefits**

It could be for instance used to specify any script before we start/shutdown the PostgreSQL server.

**Possible drawbacks**

N/A

**Applicable issues**

N/A

**Additional information**

N/A

**Checklist** 


alvneiayu:
hi @ddi-wtag 

First of all, thanks a lot for your PR. I need to request you some changes. Right now, we are including lifecycle in other charts but we are following a different way:

https://github.com/bitnami/charts/blob/20e4cad2aad257a4ebc277877651c4a22d9ab783/bitnami/redis/templates/sentinel/statefulset.yaml#L88

https://github.com/bitnami/charts/blob/20e4cad2aad257a4ebc277877651c4a22d9ab783/bitnami/redis/values.yaml#L329

We don't have this support Postgresql yet. I think that including this support, you will be able to include the hook like you are doing. So, could you help us to include the support like we are doing?

Thanks a lot

Álvaro

ddi-wtag:
Hi @alvneiayu,

Thanks a lot for your feedback! I have changed the PR to use the `lifecycleHooks` value as proposed.

alvneiayu (approved):
LGTM. Thanks a lot for your PR and you work.

bitnami/postgresql/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 annotations:
   category: Database
 apiVersion: v2
-appVersion: 11.12.0
+appVersion: 11.13.0
```
alvneiayu, line 4: Sorry, just one thing. This is not the version that you need to change it. Please leave the version to 11.12.0. You need to modify the version version in the end of the file. This is the reason that you are having linter errors

alvneiayu (changes requested):
Just one request to fix the linter test errors that you are having in this change.

bitnami/postgresql/Chart.yaml
ddi-wtag: Good to know, thanks! I reverted the `appVersion` and changed the `version` accordingly.

alvneiayu (approved):
Now, yes. Thanks a lot again for your work.

PR #7060, opened 2021-07-26
Kibana chart should have service account

### LINKED ISSUE #7059: Kibana chart should have service account
**Which chart**:

Kibana latest

In order to give service accounts the fine-grained access control that they need and not just give the default service account permissions, a service account should be defined for a deployment's pods.  In the case of kibana, in some environments it needs the anyuid security context constraint.  We may not want to provide this to default, plus doing so may be blocked by auditing + infosec.

Like the elasticsearch chart, for the kibana chart there should be a `serviceAccount` section with two key-values: name (string), and create (bool).  If not specified, the default service account may be used.

I think adding the above should be straightforward enough.  We're using the kibana subchart with `kibanaEnabled: true` from the elasticsearch chart, so things ShouldJustWork.  One alternative is not to use the kibana subchart and go with the official kibana chart from elasticsearch.

**Additional context**

We're installing on openshift and aiding openshift customers to do so.  They have infosec requirements that we do not control.

jeff-hammel-leapyear:
**Description of the change**

This allows the specification of a service account for Kibana.  See also https://github.com/bitnami/charts/issues/7059

**Benefits**

You should now be able to specify security contexts to a particular service account, not just defaults.

**Possible drawbacks**

None that I can think of.

**Applicable issues**

  - fixes #7059 

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/kibana/templates/deployment.yaml
+++ b/bitnami/kibana/templates/deployment.yaml
@@ -73,6 +73,7 @@ spec:
           securityContext:
             runAsUser: {{ .Values.securityContext.runAsUser }}
           {{- end }}
+          serviceAccountName: {{ .Values.serviceAccount.name }}
           env:
             - name: KIBANA_PORT_NUMBER
               value: {{ .Values.containerPort | quote }}
--- a/bitnami/kibana/templates/serviceaccount.yaml
+++ b/bitnami/kibana/templates/serviceaccount.yaml
@@ -1,0 +1,6 @@
+{{- if.Values.serviceAccount.create }}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name:  {{ required .Values.serviceAccount.name }}
+{{- end }}
--- a/bitnami/kibana/values.yaml
+++ b/bitnami/kibana/values.yaml
@@ -321,6 +321,11 @@ ingress:
   ##   certificate:
   ##
   secrets: []
+## @param serviceAccount.name Name of serviceAccount
+## @param serviceAccount.create Enable creation of ServiceAccount for Kibana
+serviceAccount:
+  name: default
+  create: false
 ## @param containerPort Port to expose at container level
 ##
 containerPort: 5601


jeff-hammel-leapyear:
I'm not going to have a chance to look at this soon.  Could someone please comment on what needs to be done to get this PR in a state?  Alternately, if someone could shepherd this in, I would be grateful.

bitnami/kibana/templates/serviceaccount.yaml
```diff
@@ -0,0 +1,6 @@
+{{- if.Values.serviceAccount.create }}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name:  {{ required .Values.serviceAccount.name }}
+{{- end }}
```
miguelaeh, line 6: Could you please base on the following service account in order to standardize it please? https://github.com/bitnami/charts/blob/master/bitnami/redis/templates/serviceaccount.yaml

bitnami/kibana/templates/deployment.yaml
```diff
@@ -73,6 +73,7 @@ spec:
           securityContext:
             runAsUser: {{ .Values.securityContext.runAsUser }}
           {{- end }}
+          serviceAccountName: {{ .Values.serviceAccount.name }}
```
miguelaeh, line 76: We use a template in order to render the proper serviceAccount name, you can base on the following: https://github.com/bitnami/charts/blob/c10c7f4015d1ed0bf52ed10f9cd0cc6fb7d94910/bitnami/redis/templates/_helpers.tpl#L141

miguelaeh (changes requested):
Hi @jeff-hammel-leapyear ,
Thank you very much for the change.
Please take a look at my comments, also don't forget to bump the chart version. There is a checklist on the PR comment that you should complete.

james-pollard-leapyear:
@miguelaeh Hi, Jeff's teammate here. He's on vacation for a while and I imagine you can understand that this change isn't a priority for us. Please consider his change a PoC and take lead on shepherding this change through.

Thanks for getting back to us!

Mauraza:
Hi @james-pollard-leapyear,

I just create a new PR with the change.

james-pollard-leapyear:
> I just create a new PR with the change.

Excellent - thank you! feel free to close out our PR here

juan131:
Implemented at https://github.com/bitnami/charts/pull/7211

james-pollard-leapyear:
Thanks a bunch!

PR #7061, opened 2021-07-27
[bitnami/keycloak] Adding a post initContainer value to allow init containers to be added to the chart which are run after tls certificate creation

geoffo-dev:
**Description of the change**

The change has been made to allow the execution of init containers after the init container which creates the tls certificates.

**Benefits**

This would allow for manipulation of the keystore/truststore once it has been created - this would be particularly useful for adding additional trusted certificates without having to generate all the certificates.

``` yaml
initContainersPost:
  - name: init-custom-certs
    image: docker.io/bitnami/keycloak:14.0.0-debian-10-r0
    imagePullPolicy: Always
    command:
      - /bin/bash
      - -ec
      - |-
        yes | keytool -import -alias ad-cert -file "/custom-certs/ad.crt" \
            -keystore "/opt/bitnami/keycloak/certs/keycloak.truststore.jks" \
            -storepass $KEYCLOAK_TLS_TRUSTSTORE_PASSWORD -noprompt
    volumeMounts:
      - name: shared-certs
        mountPath: /opt/bitnami/keycloak/certs
      - name: custom-certs
        mountPath: /custom-certs
    env:
      - name: KEYCLOAK_TLS_TRUSTSTORE_PASSWORD
        valueFrom: 
          secretKeyRef: 
            name: ${site_name}
            key: tls-truestore-password
```

This would be an example of what could be used to add additional certificates to the the store

**Possible drawbacks**

Not that pretty...

**Applicable issues**

N/A

**Additional information**

I am not sure whether it would be easier to just add the InitContainers value to after the TLS generation instead - I cannot think of a reason why it would need to be after... 

**Checklist** 


miguelaeh (changes requested):
Hi @geoffo-dev ,
I think we should move the `initContainer` renderization block after the `init-certs` init container instead of adding duplicated parameters

geoffo-dev:
Hey @miguelaeh,

Sounds good to me - I wasn't sure if there was a reason as to why it was after. Will make the amendments now.

Many Thanks
Geoff

geoffo-dev:
@miguelaeh - done :-)

miguelaeh (changes requested):
Please bump the chart version

geoffo-dev:
@miguelaeh - done :-)

miguelaeh (approved):
LGTM!

PR #7099, opened 2021-07-29
[bitnami/contour] Sync CRDs with contour release 1.18

artificial-aidan:
**Description of the change**

Updates the Contour CRDs to match the recently release Contour 1.18

**Benefits**

Adds `ingressClassName` field

**Checklist** 

### DIFF
--- a/bitnami/contour/Chart.lock
+++ b/bitnami/contour/Chart.lock
@@ -1,6 +1,6 @@
 dependencies:
 - name: common
   repository: https://charts.bitnami.com/bitnami
-  version: 1.7.0
-digest: sha256:6786bc7fdfc6e4038fd28819c2b07339f232282dbaeab21de8064aaa6814a8d1
-generated: "2021-07-07T06:53:45.969134789Z"
+  version: 1.7.1
+digest: sha256:d05549cd2eb5b99a49655221b8efd09927cc48daca3fa9f19af0257a11e5260f
+generated: "2021-08-02T17:24:22.28379482Z"
--- a/bitnami/contour/Chart.yaml
+++ b/bitnami/contour/Chart.yaml
@@ -27,4 +27,4 @@ sources:
   - https://github.com/envoyproxy/envoy
   - https://github.com/bitnami/bitnami-docker-contour
   - https://projectcontour.io
-version: 5.0.3
+version: 5.0.4
--- a/bitnami/contour/resources/extensionservices.yaml
+++ b/bitnami/contour/resources/extensionservices.yaml
@@ -159,8 +159,8 @@ spec:
                   service's certificate
                 properties:
                   caSecret:
-                    description: Name of the Kubernetes secret be used to validate
-                      the certificate presented by the backend
+                    description: Name or namespaced name of the Kubernetes secret
+                      used to validate the certificate presented by the backend
                     type: string
                   subjectName:
                     description: Key which is expected to be present in the 'subjectAltName'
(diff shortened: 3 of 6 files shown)


artificial-aidan:
In reality this should probably be done by the same bot that updates the application versions.

VA-PLE:
@artificial-aidan Fix the file bitnami/contour/resources/tlscertificatedeligations.yaml there should be no changes

carrodher:
> In reality this should probably be done by the same bot that updates the application versions.

Currently, there are two separate processes, the automation detect the new version of the upstream application (contour in this case) and then release the container image, update the Helm chart with the new container and also the subchart and other containers (if any), then the automated pipeline test the chart on top of different clusters such as GKE, AKS, IKS, TKG, etc and perform the release with the following changes (https://github.com/bitnami/charts/commit/b506bb4f6b421d29964abcd68edbd21954d5416c).

This is how the chart is automatically updated, then the changes in the chart templates should be done manually, there is not automated way to implement changes into the templates, this is done by Bitnami team members or community users, but it is not possible to automatically change the source code of the charts

carrodher (commented):
Thanks for the PR, can you take a look at the comment from @VA-PLE about the tlscertificatedeligations?

artificial-aidan:
> > In reality this should probably be done by the same bot that updates the application versions.
> 
> Currently, there are two separate processes, the automation detect the new version of the upstream application (contour in this case) and then release the container image, update the Helm chart with the new container and also the subchart and other containers (if any), then the automated pipeline test the chart on top of different clusters such as GKE, AKS, IKS, TKG, etc and perform the release with the following changes (https://github.com/bitnami/charts/commit/b506bb4f6b421d29964abcd68edbd21954d5416c).
> 
> This is how the chart is automatically updated, then the changes in the chart templates should be done manually, there is not automated way to implement changes into the templates, this is done by Bitnami team members or community users, but it is not possible to automatically change the source code of the charts

That makes sense. But what would have happened if the released version of contour had relied upon changes to the CRDs to function, in this case it was just additive, but it seems like there could be a case where it could break things.

carrodher:
> > > In reality this should probably be done by the same bot that updates the application versions.
> > 
> > 
> > Currently, there are two separate processes, the automation detect the new version of the upstream application (contour in this case) and then release the container image, update the Helm chart with the new container and also the subchart and other containers (if any), then the automated pipeline test the chart on top of different clusters such as GKE, AKS, IKS, TKG, etc and perform the release with the following changes ([b506bb4](https://github.com/bitnami/charts/commit/b506bb4f6b421d29964abcd68edbd21954d5416c)).
> > This is how the chart is automatically updated, then the changes in the chart templates should be done manually, there is not automated way to implement changes into the templates, this is done by Bitnami team members or community users, but it is not possible to automatically change the source code of the charts
> 
> That makes sense. But what would have happened if the released version of contour had relied upon changes to the CRDs to function, in this case it was just additive, but it seems like there could be a case where it could break things.

If those breaking changes are something that is detected by our tests, the release would be blocked, the "problem" here is that the Helm charts have a lot of parameters and it is not possible to test all the combinations. We deploy the Helm chart with a subset of parameters, if the installation and tests with those parameters fail, the chart is not released (code not pushed to GH nor chart .tgz pushed to the chart registry), but it is not 100% safe, maybe a new version change something that is not covered by the tests, in that case, we will need to release a new version fixing the issue if it is reported by a user.

VA-PLE:
@artificial-aidan can you fix your request?

artificial-aidan:
> @artificial-aidan can you fix your request?

Yes I will get to it today, had a hectic week last week

carrodher (approved):
LGTM

PR #7101, opened 2021-07-29
[bitnami/ postgresql-ha] 7.8.4 Update mountPath owner

Nigel-Caughey:
**Description of the change**
Change the owner of mountPath to 1001/1001, not just subfolders when volumePermissions.enabled

**Benefits**
k3s v1.21.3 ( https://github.com/k3s-io/k3s/releases/tag/v1.21.3%2Bk3s1 )
Changes the permission of local storage pods ( https://github.com/k3s-io/k3s/pull/3548 )
This changes it back, so postgres can access the pvc
The change has the same functionality as postgresql ( which does a chown vs minpath of 0 in this patch ).

**Possible drawbacks**
Not tested with an external pvc


bitnami/postgresql-ha/templates/postgresql/statefulset.yaml
```diff
@@ -106,7 +106,7 @@ spec:
             - |
               mkdir -p {{ .Values.persistence.mountPath }}/conf {{ .Values.persistence.mountPath }}/data {{ .Values.persistence.mountPath }}/lock
               chmod 700 {{ .Values.persistence.mountPath }}/conf {{ .Values.persistence.mountPath }}/data {{ .Values.persistence.mountPath }}/lock
-              find {{ .Values.persistence.mountPath }} -mindepth 1 -maxdepth 1 -not -name ".snapshot" -not -name "lost+found" | \
+              find {{ .Values.persistence.mountPath }} -mindepth 0 -maxdepth 1 -not -name ".snapshot" -not -name "lost+found" | \
```
alemorcuq, line 109:
```suggestion
              chown {{ .Values.postgresql.containerSecurityContext.runAsUser }}:{{ .Values.postgresql.securityContext.fsGroup }} {{ .Values.persistence.mountPath }}
              find {{ .Values.persistence.mountPath }} -mindepth 1 -maxdepth 1 -not -name ".snapshot" -not -name "lost+found" | \
```

alemorcuq: Your solution would change the ownership of some directories we are excluding. Instead, let's only change the ownership of `persistence.mounPath` in a non-recursive way.

alemorcuq (changes requested):
Thanks for the PR!

Issue #5181, opened 2021-01-22
[bitnami/redis] All pods in CrashLoopBackOff status after rebooting worker nodes


wantdrink:
**Which chart**:
bitnami/redis redis-12.6.2 app version 6.0.10

**Describe the bug**
After rebooting all of the K8S worker nodes, the redis pods are always in CrashLoopBackOff status.

**To Reproduce**

1. Set redis image with 5.0.10-debian-10-r81, sentinel image with 5.0.10-debian-10-r78.
2. cluster = true, slaveCount=3
3. helm install mycache bitnami/redis  -n test -f values.yaml
4. Test and everything works fine. Reboot all K8S worker nodes.
5. 3 redis pods are always in CrashLoopBackOff  status
NAME                READY   STATUS             RESTARTS   AGE
mycache-redis-node-0   1/3     CrashLoopBackOff   19         30m
mycache-redis-node-1   1/3     CrashLoopBackOff   19         30m
mycache-redis-node-2   1/3     CrashLoopBackOff   19         30m

**Version of Helm and Kubernetes**:

- Output of `helm version`:

```
version.BuildInfo{Version:"v3.2.4", GitCommit:"0ad800ef43d3b826f31a5ad8dfbb4fe05d143688", GitTreeState:"clean", GoVersion:"go1.13.12"}
```

- Output of `kubectl version`:

```
Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.5", GitCommit:"e6503f8d8f769ace2f338794c914a96fc335df0f", GitTreeState:"clean", BuildDate:"2020-06-26T03:47:41Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.5", GitCommit:"e6503f8d8f769ace2f338794c914a96fc335df0f", GitTreeState:"clean", BuildDate:"2020-06-26T03:39:24Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}
```

**Additional context**

> kubectl describe pod mycache-redis-node-x:

```
Events:
  Type     Reason                  Age                    From                  Message
  ----     ------                  ----                   ----                  -------
  Normal   Scheduled               <unknown>              default-scheduler     Successfully assigned test/mycache-redis-node-0 to app3.jh.com
[...]

PR #7124, opened 2021-08-03
[bitnami/redis]: fix CrashLoopBackOff with sentinel enabled

rsecob:
**Description of the change**

After upgrading to the latest Redis chart version from `12.0.1` we noticed that sentinel behaviour was broken.

When sentinel was enabled there was a check that would ensure a new master was elected before starting the redis & sentinel processes, otherwise, it would exit with status code 1, causing workers to stay in a CrashLoopBackOff while the remaining workers were unable to elect a new master.

We removed this check in both the`start-node` and `start-sentinel` scripts as it will eventually lead to this CrashLoopBackOff state.

We should not enforce a master check before starting the processes. In a configuration of 3 sentinel replicas with a quorum of 2, if we lose 2 nodes at the same time (including the master), the last remaining node won't be able to elect a new master while the two others will keep failing as no master is available.

However, if we let the nodes fully restart, they will join the quorum and participate in the election.

The check has been added with https://github.com/bitnami/charts/pull/4478 claiming that when the previous master node was restarted it will still see the previous master on bootup. But this behaviour should be expected because if your node restarts faster than your `sentinel.downAfterMilliseconds` value, the other nodes might not have started a failover just yet. The newly connected node will be able to participate in the election regardless. 

**Benefits**
- Fixes the CrashLoopBackOff state that some of us have experienced when using sentinel mode

**Possible drawbacks**
- The newly created node might still see the previous master as the current master, but as mentioned above, if the quorum is reached, a new master will be elected and this new node will participate in the election. How fast a new master will be elected will depend on your `sentinel.downAfterMilliseconds` and `sentinel.failoverTimeout` config. 

**Applicable issues**

  - fixes https://github.com/bitnami/charts/issues/5181

**Additional information**
It is possible (and I have seen it happen) that when not using `sentinel.staticID = true` (which was a feature introduced by me https://github.com/helm/charts/pull/19059) sentinels are not able to elect a new master after a few restarts. The reason being is that every time a new worker is created its sentinel will be given a random ID which will be added to the list of sentinel peers. If a node has been running for longer than others, they might keep the list of old sentinel instances in their list of peers which compromises the election process. I see that in the `start-sentinel` script the command `sentinel reset "*"` is sent to all sentinels on the network which should technically alleviate this problem, but somehow I still had issues when running without `sentinel.staticID = true`.

My opinion is that `staticID = true` should be the default behaviour because I don't see how it could work without in k8s, though that is a different discussion. In the meantime, all configs using `staticID = true` should see their CrashLoopBackOff resolved.

**Checklist** 

### DIFF
--- a/bitnami/redis/Chart.yaml
+++ b/bitnami/redis/Chart.yaml
@@ -25,4 +25,4 @@ name: redis
 sources:
   - https://github.com/bitnami/bitnami-docker-redis
   - http://redis.io/
-version: 14.8.9
+version: 14.8.10
--- a/bitnami/redis/templates/scripts-configmap.yaml
+++ b/bitnami/redis/templates/scripts-configmap.yaml
@@ -70,22 +70,6 @@ data:
         REDIS_SENTINEL_INFO=($($sentinel_info_command))
         REDIS_MASTER_HOST=${REDIS_SENTINEL_INFO[0]}
         REDIS_MASTER_PORT_NUMBER=${REDIS_SENTINEL_INFO[1]}
-
-        # Immediately attempt to connect to the reported master. If it doesn't exist the connection attempt will either hang
-        # or fail with "port unreachable" and give no data. The liveness check will then timeout waiting for the redis
-        # container to be ready and restart the it. By then the new master will likely have been elected
-        if is_boolean_yes "$REDIS_TLS_ENABLED"; then
-            sentinel_info_command="redis-cli {{- if .Values.auth.enabled }} -a $REDIS_PASSWORD {{- end }} -h $REDIS_MASTER_HOST -p {{ .Values.sentinel.service.sentinelPort }} --tls --cert ${REDIS_TLS_CERT_FILE} --key ${REDIS_TLS_KEY_FILE} --cacert ${REDIS_TLS_CA_FILE} sentinel get-master-addr-by-name {{ .Values.sentinel.masterSet }}"
-        else
-            sentinel_info_command="redis-cli {{- if .Values.auth.enabled }} -a $REDIS_PASSWORD {{- end }} -h $REDIS_MASTER_HOST -p {{ .Values.sentinel.service.sentinelPort }} sentinel get-master-addr-by-name {{ .Values.sentinel.masterSet }}"
-        fi
-
-        if [[ ! ($($sentinel_info_command)) ]]; then
-            # master doesn't actually exist, this probably means the remaining pods haven't elected a new one yet
-            # and are reporting the old one still. Once this happens the container will get stuck and never see the new
-            # master. We stop here to allow the container to not pass the liveness check and be restarted.
-            exit 1
-        fi
     fi
 
     if [[ ! -f /opt/bitnami/redis/etc/redis.conf ]];then
@@ -219,22 +203,6 @@ data:
         REDIS_SENTINEL_INFO=($($sentinel_info_command))
         REDIS_MASTER_HOST=${REDIS_SENTINEL_INFO[0]}
         REDIS_MASTER_PORT_NUMBER=${REDIS_SENTINEL_INFO[1]}
-
-        # Immediately attempt to connect to the reported master. If it doesn't exist the connection attempt will either hang
-        # or fail with "port unreachable" and give no data. The liveness check will then timeout waiting for the sentinel
-        # container to be ready and restart the it. By then the new master will likely have been elected
-        if is_boolean_yes "$REDIS_SENTINEL_TLS_ENABLED"; then
-            sentinel_info_command="redis-cli {{- if .Values.auth.enabled }} -a $REDIS_PASSWORD {{- end }} -h $REDIS_MASTER_HOST -p {{ .Values.sentinel.service.sentinelPort }} --tls --cert ${REDIS_SENTINEL_TLS_CERT_FILE} --key ${REDIS_SENTINEL_TLS_KEY_FILE} --cacert ${REDIS_SENTINEL_TLS_CA_FILE} sentinel get-master-addr-by-name {{ .Values.sentinel.masterSet }}"
-        else
-            sentinel_info_command="redis-cli {{- if .Values.auth.enabled }} -a $REDIS_PASSWORD {{- end }} -h $REDIS_MASTER_HOST -p {{ .Values.sentinel.service.sentinelPort }} sentinel get-master-addr-by-name {{ .Values.sentinel.masterSet }}"
-        fi
-
-        if [[ ! ($($sentinel_info_command)) ]]; then
-            # master doesn't actually exist, this probably means the remaining pods haven't elected a new one yet
-            # and are reporting the old one still. Once this happens the container will get stuck and never see the new
-            # master. We stop here to allow the container to not pass the liveness check and be restarted.
-            exit 1
-        fi
     fi
 
     # Clean sentineles from the current sentinel nodes after failover completed.


miguelaeh (commented):
Hi @rsecob ,
Thank you very much for the detailed description of the issue and for performing the changes.

> The newly created node might still see the previous master as the current master, but as mentioned above, if the quorum is reached, a new master will be elected and this new node will participate in the election. How fast a new master will be elected will depend on your `sentinel.downAfterMilliseconds` and `sentinel.failoverTimeout` config.

About this, I have the doubt, if your failoverTimeout and downAfterMilliseconds is relatively high, would that end on a cluster downtime? In that case, maybe we should configure them with a proper value in this same PR.

rsecob:
Hi @miguelaeh,

Thank you for reviewing this PR! 

> > The newly created node might still see the previous master as the current master, but as mentioned above, if the quorum is reached, a new master will be elected and this new node will participate in the election. How fast a new master will be elected will depend on your `sentinel.downAfterMilliseconds` and `sentinel.failoverTimeout` config.
> 
> About this, I have the doubt, if your failoverTimeout and downAfterMilliseconds is relatively high, would that end on a cluster downtime? In that case, maybe we should configure them with a proper value in this same PR.

I would say it is almost impossible to avoid a cluster downtime regardless if you use the sentinel mode or not, but I guess it depends on your cloud provider & if your k8s node pools scale up and down constantly depending on your workloads.

The `sentinel.downAfterMilliseconds` is 60secs and `sentinel.failoverTimeout` is 18secs by default. When a master goes does, it will take 1min for the other sentinels to acknowledge their master is down and start the election process, almost immediately after, a new master should be elected. In our config, we reduced this time to 5secs but I would say 1min is an ok default value, I don't know what is your opinion on this?

When a master goes down, a cluster downtime is unavoidable but a master will be elected eventually. Whereas with this check that this PR removes, in many situations, it could lead to a forever cluster downtime.

Most redis clients automatically try to reconnect to the new master (often with a back-off strategy). The default is 1min as mentioned above, it could be faster if we changed the default value. It is up to you guys, but I would say 1min is fine, maybe 30 secs?

miguelaeh (approved):
Hi @rsecob ,
Thank you very much for the explanation again.
The PR LGTM, but I would like to hear the @rafariossaa opinion about it. I think this logic was probably added to avoid sending traffic to pods that are not actually part of the Redis cluster. I think deleting it there will be some time where some running pods will not be actually part of the Redis cluster, and, if they receive traffic, the responses could be inconsistent.

rafariossaa:
Hi,
  I am not very sure about this part:
  ```
We should not enforce a master check before starting the processes. In a configuration of 3 sentinel replicas with a quorum of 2, if we lose 2 nodes at the same time (including the master), the last remaining node won't be able to elect a new master while the two others will keep failing as no master is available.```
  When there is only one node left, I could not be automatically be a new master because the lack of quorum. If it is, it won't be able to differentiate between 2 node loss and if it is the only node isolated and there are 2 nodes running.

dannylee-:
We ran into this issue and this was the fix we were planning to do (+ seeing if we can use hostnames instead of IP addresses).  Thank you, @rsecob, for this PR.

rsecob:
@dannylee-

> We ran into this issue and this was the fix we were planning to do (+ seeing if we can use hostnames instead of IP addresses). Thank you, @rsecob, for this PR.
 
Nice ! Why would you want to use hostnames for?

@miguelaeh @rafariossaa Should I go ahead and merge?

dannylee-:
@rsecob A portion of the start-sentinel script seems to be deleting/creating/replacing entries in the sentinel.conf, due to the fact that restarted kubernetes pods will get a new IP address.  In the new 6.2 version of Redis, they [introduced a way to add hostnames instead of IP addresses](https://redis.io/topics/sentinel#ip-addresses-and-dns-names).  So if we used hostnames instead of IPs, we wouldn't need to do all this logic to regenerate these sentinel entries.

Note, it has nothing to do with this fix and I think we should go ahead with this PR.

rsecob:
> @rsecob A portion of the start-sentinel script seems to be deleting/creating/replacing entries in the sentinel.conf, due to the fact that restarted kubernetes pods will get a new IP address. In the new 6.2 version of Redis, they [introduced a way to add hostnames instead of IP addresses](https://redis.io/topics/sentinel#ip-addresses-and-dns-names). So if we used hostnames instead of IPs, we wouldn't need to do all this logic to regenerate these sentinel entries.
> 
> Note, it has nothing to do with this fix and I think we should go ahead with this PR.

Oh yes, that is a game-changer.
This is awesome news, looking forward to seeing this implemented, let me know if you need help establishing this in the charts I'd be happy to help.

Maybe I should wait for @rafariossaa approval, he did not seem convinced about this PR.

rafariossaa:
Hi,
  Sorry for the delay.
 
  Regarding this:

> That would be awesome if it worked like that and indeed my PR would not be required. However, as per the Sentinel documentation quotes:
>
>one of the Sentinels need to be elected leader for the failover and be authorized to proceed. This only happens with the vote of the majority of the Sentinel processes.

  Sorry if I didn't explain myself very well. That is what I meant, you need quorum (a set of nodes that agree) usually you need half of the nodes plus one (`n/2 + 1`, where `n` if the number of nodes). 
  So if you lose 2 nodes from a 3 nodes cluster:
> We should not enforce a master check before starting the processes. In a configuration of 3 sentinel replicas with a quorum of 2, if we lose 2 nodes at the same time (including the master), the last remaining node won't be able to elect a new master while the two others will keep failing as no master is available.
  there is not quorum enough to automatically recover.

rsecob:
Hello @rafariossaa,

Thank you for your reply.

Sorry, I misunderstood your answer. Indeed like you say we need the majority of sentinels to elect a new master.

The master check that this PR removes prevented the new nodes from joining the quorum and participate in the election. Hence why we end up having a CrashLoopBackOff in the case where we lose 2 replicas at the same time (including the master replica) with a setting of 3 sentinels.

Should we go ahead and merge?

rsecob:
@miguelaeh

I had to bump the chart's version as it was updated by another commit, can we have another approval please 🙏

rafariossaa:
> The master check that this PR removes prevented the new nodes from joining the quorum and participate in the election. Hence why we end up having a CrashLoopBackOff in the case where we lose 2 replicas at the same time (including the master replica) with a setting of 3 sentinels.

You need to have a new master elected before other nodes can join, if there is not master there is not cluster to join to.
Let say you have 5 nodes, `node-0` being the master. At some point, there is a network issue that splits the cluster like this:
set 1: `node-0 + node-1` and set 2: `node-2 + node-3 + node-4`. Also, you are adding more nodes (node-5 + node-6). 
The new nodes need to be added to the cluster. Where these nodes should go in set 1 or in set 2?
At one hand, node-0 is master, but at other hand it is in a set that doesn't contains the majority of the nodes. Set 2, as it contains the majority of the nodes, needs to elect a new master. In the case that the new nodes are added to set-1 you will end with two separated cluster (splited brain).

Once that in the cluster a split is detected, no other operations are allowed until a new master is elected, the new nodes can not help in the election of a new master.

dannylee-:
How does a newly added redis sentinel instance know who the master is?  From what I'm seeing in the sentinel.conf, it automatically assumes the redis master is pod-0?

rafariossaa:
The `REDIS_MASTER_HOST` is set by [this code block](https://github.com/bitnami/charts/blob/bfb3b88277b1830b23b1f6b32bb2b776f3b76884/bitnami/redis/templates/scripts-configmap.yaml#L210).

rsecob:
> You need to have a new master elected before other nodes can join, if there is not master there is not cluster to join to.
> Let say you have 5 nodes, `node-0` being the master. At some point, there is a network issue that splits the cluster like this:
> set 1: `node-0 + node-1` and set 2: `node-2 + node-3 + node-4`. Also, you are adding more nodes (node-5 + node-6).
> The new nodes need to be added to the cluster. Where these nodes should go in set 1 or in set 2?
> At one hand, node-0 is master, but at other hand it is in a set that doesn't contains the majority of the nodes. Set 2, as it contains the majority of the nodes, needs to elect a new master. In the case that the new nodes are added to set-1 you will end with two separated cluster (splited brain).
> 
> Once that in the cluster a split is detected, no other operations are allowed until a new master is elected, the new nodes can not help in the election of a new master.

I disagree with your statement that says "You need to have a new master elected before other nodes can join". It just doesn't work in the case where we lose enough nodes at the same time that the quorum for election is not met. In the case of 3 nodes, as mentioned before, losing 2 nodes will lead to a CrashLoopBackOff state. Losing many nodes at the same time might seem like an edge case (which it is) and might seem like it's not very likely to happen, but believe me, from experience, it can happen and it will.

We want to have a resilient sentinel system that allows for self-recovery in all cases. Which is not the case right now.

Let me give you an example so you can see it fail for yourself:

> values.yaml:
```yaml
auth:
  enabled: false

sentinel:
  enabled: true
```

Deploy and let it dispatch the three workers

> helm install -f values.yaml test bitnami/redis 
```
NAME                READY   STATUS    RESTARTS   AGE     IP           NODE                                          NOMINATED NODE   READINESS GATES
test-redis-node-0   2/2     Running   0          2m49s   10.8.11.36   gke-staging-zonal-preemptible-f583984b-091k   <none>           <none>
test-redis-node-1   2/2     Running   0          2m18s   10.8.3.40    gke-staging-zonal-preemptible-f583984b-8h9f   <none>           <none>
test-redis-node-2   2/2     Running   0          83s     10.8.10.30   gke-staging-zonal-preemptible-f583984b-zlbt   <none>           <none>
```

At this point the master is `test-redis-node-0`, lets force kill 2 nodes:
> kubectl delete pods test-redis-node-0 test-redis-node-1

And now we have 1 node that is failing & looping. Waiting for a master to be elected. The last remaining node can not elect & promote a new master by itself.

```
NAME                READY   STATUS             RESTARTS   AGE     IP           NODE                                          NOMINATED NODE   READINESS GATES
test-redis-node-0   0/2     CrashLoopBackOff   11         8m51s   10.8.3.42    gke-staging-zonal-preemptible-f583984b-8h9f   <none>           <none>
test-redis-node-2   2/2     Running            0          11m     10.8.10.31   gke-staging-zonal-preemptible-f583984b-zlbt   <none>           <none>
```

We made this PR cause we wanted to do a quick fix for this specific case but then realized that the sentinel implementation is much more fragile than we thought it was.

I don't think it makes sense we handle two different behaviours for sentinel, those who use `staticID` and those who do not use it. When one makes a PR and fixes a behaviour for one or the other, it might not account for the other functionality. I think we should harmonize this behaviour so everyone expects the same resiliency.

We're gonna close this PR and make another one that we believe will fix the issue above and bring more stability to the system.

rafariossaa:
Hi,
  Thanks for the explanation.
  I am closing this PR, and let's continue in the new PR.

PR #7128, opened 2021-08-03
[bitnami/external-dns] rfc2136 provider can now be configured to use Kerberos authentication

### LINKED ISSUE #7127: [bitnami/external-dns] Add Kerberos support
**Which chart**:
external-dns 5.2.2

The application supports configuring the rfc2136 provider with authentication over Kerberos, but no equivalent option is possible in the Helm chart.

Add variables to the values.yaml to allow configuration over rfc3645 (Kerberos)

Configure the application by hand and stop using the Helm chart

**Additional context**
PR forthcoming...

pipeworks-asmith:
**Description of the change**

Adds configuration options to support rfc3645 as described at the bottom of this doc in the parent project https://github.com/kubernetes-sigs/external-dns/blob/master/docs/tutorials/rfc2136.md

**Benefits**

Allows users interfacing with Windows DNS to authenticate over Kerberos

**Possible drawbacks**

Some of the conditionals in the template have gotten rather unwieldy, but it seemed more painful to refactor (and would introduce a breaking change) than to add another "or" in the chain.

**Applicable issues**

  - fixes #7127

**Additional information**

Tested and working on an internal Windows DNS server over Kerberos. Thanks for making this chart so easy to fork and change!

**Checklist** 


bitnami/external-dns/templates/configmap.yaml
```diff
@@ -1,10 +1,16 @@
...
   name: {{ template "external-dns.fullname" . }}
   labels: {{ include "external-dns.labels" . | nindent 4 }}
 data:
+{{- if .Values.designate.customCA.enabled }}
   {{ .Values.designate.customCA.filename }}: |
 {{ .Values.designate.customCA.content | indent 4 }}
 {{- end }}
+{{- if .Values.rfc2136.rfc3645Enabled }}
+  krb5.conf: |
+{{ required "rfc2136.kerberosConfig must be set if rfc2136.3645Enabled is `true`" .Values.rfc2136.kerberosConfig | indent 4 }}
```
alemorcuq, line 14: We usually perform this kind of validations in the `_helpers.tpl` file.

bitnami/external-dns/templates/configmap.yaml
```diff
@@ -1,10 +1,16 @@
-{{- if and (eq .Values.provider "designate") .Values.designate.customCA.enabled }}
+{{- if and (eq .Values.provider "designate") .Values.designate.customCA.enabled | or (and (eq .Values.provider "rfc2136") .Values.rfc2136.rfc3645Enabled) }}
```
alemorcuq, line 1: It may make sense to do something in the `_helpers.tpl` file like we do with the `createSecret` helper.

bitnami/external-dns/templates/deployment.yaml
```diff
@@ -220,14 +220,19 @@ spec:
             - --rfc2136-port={{ .Values.rfc2136.port }}
             - --rfc2136-zone={{ required "rfc2136.zone must be supplied for provider 'rfc2136'" .Values.rfc2136.zone }}
             - --rfc2136-min-ttl={{ .Values.rfc2136.minTTL }}
-              {{- if .Values.rfc2136.tsigKeyname }}
-            - --rfc2136-tsig-secret-alg={{ .Values.rfc2136.tsigSecretAlg }}
-            - --rfc2136-tsig-keyname={{ .Values.rfc2136.tsigKeyname }}
-                {{- if .Values.rfc2136.tsigAxfr }}
+              {{- if .Values.rfc2136.tsigAxfr }}
```
alemorcuq, line 223: Previously we were adding this flag if `tsigKeyname` was also set and now we are adding it regardless of `tsigKeyname`. Why the change?

bitnami/external-dns/templates/deployment.yaml
```diff
@@ -220,14 +220,19 @@ spec:
...
             - --rfc2136-tsig-axfr
-                {{- end }}
+              {{- end }}
+              {{- if .Values.rfc2136.rfc3645Enabled }}
+            - --rfc2136-gss-tsig
+            - --rfc2136-kerberos-realm={{ required "rfc2136.kerberosRealm must be supplied for provider `rfc2136` if `rfc2136.rfc3645Enabled` is true." .Values.rfc2136.kerberosRealm }}
```
alemorcuq, line 228: The `values.yaml` file says the realm is optional, but it shows as required here. Which one is correct?

If it's required, this validation fits the `_helpers.tpl` file better.

bitnami/external-dns/templates/deployment.yaml
```diff
@@ -446,14 +451,27 @@ spec:
                   key: infoblox_wapi_password
             {{- end }}
             {{- end }}
-            {{- if or .Values.rfc2136.tsigSecret .Values.rfc2136.secretName }}
+            {{- if or .Values.rfc2136.secretName .Values.rfc2136.tsigSecret (and .Values.rfc2136.kerberosUsername .Values.rfc2136.kerberosPassword) }}
```
alemorcuq, line 454: I'm not sure I like this condition:

```
(and .Values.rfc2136.kerberosUsername .Values.rfc2136.kerberosPassword)
```

If I'm not mistaken, this means users can't add the username and password keys to a secret they own and use it via `rfc2136.secretName`. I feel like that should be an option. What do you think?

bitnami/external-dns/templates/deployment.yaml
```diff
@@ -618,6 +641,12 @@ spec:
             path: {{ .Values.designate.customCAHostPath }}
         {{- end }}
         {{- end }}
+        {{- if and (eq .Values.provider "rfc2136") .Values.rfc2136.rfc3645Enabled }}
+        # Kerberos config volume
+        - name: krb5config
+          configMap:
+            name: {{ template "external-dns.fullname" . }}
+        {{- end }}
```
alemorcuq, line 649: We should probably specify only the key and path we want from the Config Map, just as we do with the CA one.

bitnami/external-dns/templates/deployment.yaml
```diff
@@ -618,6 +641,12 @@ spec:
             path: {{ .Values.designate.customCAHostPath }}
         {{- end }}
         {{- end }}
+        {{- if and (eq .Values.provider "rfc2136") .Values.rfc2136.rfc3645Enabled }}
+        # Kerberos config volume
```
alemorcuq, line 645: No need for this.

alemorcuq (changes requested):
Thanks for the PR! I've left some comments, please check them.

bitnami/external-dns/templates/deployment.yaml
pipeworks-asmith: As documented on the external-dns project itself https://github.com/kubernetes-sigs/external-dns/blob/master/docs/tutorials/rfc2136.md `--rfc2136-tsig-axfr` is needed to delete records, even if our authentication method is Kerberos rather than a tsig key.

bitnami/external-dns/templates/deployment.yaml
pipeworks-asmith: It is required if rfc3645Enabled is true -- is there a better way to express that?

bitnami/external-dns/templates/deployment.yaml
pipeworks-asmith: You're misreading the conditional, which is:

.Values.rfc2136.secretName OR .Values.rfc2136.tsigSecret OR (.Values.rfc2136.kerberosUsername AND .Values.rfc2136.kerberosPassword)

Users can specify the secret name, and that is indeed my exact use case

bitnami/external-dns/templates/deployment.yaml
pipeworks-asmith: Done in latest commits

bitnami/external-dns/templates/deployment.yaml
pipeworks-asmith: Done in latest commits

bitnami/external-dns/templates/configmap.yaml
pipeworks-asmith: Done in latest commits

bitnami/external-dns/templates/deployment.yaml
alemorcuq: You are right!

bitnami/external-dns/templates/configmap.yaml
alemorcuq: This doesn't show as done. It should go to the `_helpers.tpl` file, just as the one you did for the Kerberos realm.

bitnami/external-dns/templates/deployment.yaml
alemorcuq: You are right. Ignore my comment.

alemorcuq (changes requested):
Thanks for the changes! I think there's one missing still.

alemorcuq (approved):
Thanks! Changes LGTM, but I'd like @juan131 to review this as well.

juan131 (approved):
LGTM

PR #7141, opened 2021-08-05
[bitnami/influxdb] Fix ingress path

### LINKED ISSUE #7126: [Support]: influxdb with ingress
Hello I am trying to setup HA influxdb with ingress but getting error when accessing the application
# Installation of ingress controller:
helm install nginx-release --set defaultBackend.enabled=false bitnami/nginx-ingress-controller
# kubectl get svc 
nginx-release-nginx-ingress-controller        LoadBalancer   172.20.24.179    afb8d3df47a424cc590a0fa256a0737e-613715148.ap-south-1.elb.amazonaws.com   80:30467/TCP,443:30479/TCP   33m

# Installation of influxdb HA :
helm install greenkoinfluxhelm-release \
--set influxdb.service.sessionAffinity="ClientIP" \
--set  architecture="high-availability"  \
--set ingress.enabled=true \
--set ingress.hostname="afb8d3df47a424cc590a0fa256a0737e-613715148.ap-south-1.elb.amazonaws.com" \
--set auth.admin.username==##### \
--set auth.admin.password=##### \
--set relay.replicaCount=3 \
--set auth.admin.bucket="adminbucket"  \
--set auth.user.bucket="userbucket"  \
--set auth.admin.org="="#####" \
--set auth.user.org="#####" \
--set auth.readUser.username="#####" \
--set auth.readUser.password="#####" \
--set auth.writeUser.username="#####" \
--set auth.writeUser.password="#####" \
--set influxdb.replicaCount="3" \
--set metrics.enabled="true" \
--set influxdb.service.type="ClusterIP" \
--set relay.service.type="ClusterIP" \
--set metrics.service.type="ClusterIP" \
bitnami/influxdb
# error
image

rafariossaa:
**Description of the change**

Fix default ingress path

**Benefits**

**Possible drawbacks**

**Applicable issues**

  - fixes #7126

**Additional information**

**Checklist** 

- [-] Variables are documented in the README.md


juan131 (changes requested):
It seems you need to rebase from master since the linter is failing @rafariossaa 

Regarding replacing the default path.. I'd like to understand more about this.. I mean, I guess the problem is the Ingress Controller is not treating the path as a regular expression (therefore it's not considering the "wildcard" as "any character").

Starting with K8s 1.18, there's a new field in the Ingress rule called "pathType" (see line 42) and you can play with it to treat the path one way or another.

- hhttps://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#httpingresspath-v1-networking-k8s-io

Could you take a look to this?

rafariossaa:
Updated the chart version

There is not need to treat this different that in other charts. In any case, for paths `/` would be the same as `/(.*)`, so there is not need for regular expressions.

juan131 (approved):
LGTM! It might makes sense to use `Prefix` as default "pathType" but let's keep it as "ImplementationSpecific" by default.

MatthiasHertel:
@juan131 @rafariossaa 

please fix that in chart version `1.1.9` too.

juan131:
Hi @MatthiasHertel 

I'm afraid we cannot fix previous versions of the chart. What is preventing you to upgrade to the latest version?

MatthiasHertel:
Hi @juan131 

first, thank you for quick reply.

we use ntopng and as backend influxdb 1.x from this chart here, unfortunately ntopng do not support influxdb 2.x yet.
see issue:
https://github.com/ntop/ntopng/issues/3764

i guess influx 1.x is still in use, it would be very nice if it still maintained in bitnami charts.

kind regards
matthias

juan131:
Hi @MatthiasHertel 

Why not using the current chart replacing the image with the `1.x.` latest image release? You can obtain it from:

- https://github.com/bitnami/bitnami-docker-influxdb#supported-tags-and-respective-dockerfile-links

I can't guarantee that it works, but it's worth giving a try installing the chart using the **values.yaml** below:

```yaml
image:
  tag: 1.8.5-debian-10-r90
```

PR #7148, opened 2021-08-05
[bitnami/thanos] Thanos receiver dualmode

whitleykeith:
**Description of the change**

This change enables users of this chart to leverage the new deployment model of Thanos Receiver implemented in [Thanos v0.22](https://github.com/thanos-io/thanos/releases/tag/v0.22.0) by adding a `recevier.mode` to the `values.yaml` and additional configurations for the `distributor` deployment. 

**Benefits**

See [this](https://github.com/thanos-io/thanos/blob/release-0.22/docs/proposals-accepted/202012-receive-split.md) for benefits of the receiver dual-mode implementation. 

**Possible drawbacks**

Turning on the dual-mode implementation will reroute the receive service to point to the distributor. This means that there would be no service in front of the ingestor receive pods (i.e. the StatefulSet), although the headless service will still exist. This shouldn't have any impact on users correctly using the dual-mode setup as all remote_write calls need to be made to the distributor. 

**Additional information**

Small nit but I noticed that there are Github Actions to add the new values to the README on push. I updated the values.yaml with the appropriate comments, do I need to manually add it in the README or will it be auto generated on merge? 

**Checklist** 

### DIFF
--- a/bitnami/thanos/README.md
+++ b/bitnami/thanos/README.md
@@ -1005,6 +1005,10 @@ Find more information about how to deal with common errors related to Bitnami’
 
 ## Upgrading
 
+### To 5.4.0
+
+This version introduces support for the receiver dual-mode implementation for Thanos [v0.22+](https://github.com/thanos-io/thanos/releases/tag/v0.22.0)
+
 ### To 5.3.0
 
 This version introduces hash and time partitioning for the store gateway.
--- a/bitnami/thanos/templates/receive/distributor.yaml
+++ b/bitnami/thanos/templates/receive/distributor.yaml
@@ -1,62 +1,60 @@
-{{- if .Values.receive.enabled }}
-apiVersion: apps/v1
-kind: StatefulSet
+{{- if and .Values.receive.enabled ( eq .Values.receive.mode "dual-mode" )}}
+apiVersion: {{ include "common.capabilities.deployment.apiVersion" . }}
+kind: Deployment
 metadata:
-  name: {{ include "common.names.fullname" . }}-receive
+  name: {{ include "common.names.fullname" . }}-receive-distributor
   labels: {{- include "common.labels.standard" . | nindent 4 }}
-    app.kubernetes.io/component: receive
+    app.kubernetes.io/component: receive-distributor
     {{- if .Values.commonLabels }}
     {{- include "common.tplvalues.render" ( dict "value" .Values.commonLabels "context" $ ) | nindent 4 }}
     {{- end }}
 spec:
-  replicas: {{ .Values.receive.replicaCount }}
-  podManagementPolicy: {{ .Values.receive.podManagementPolicy }}
-  serviceName: {{ include "common.names.fullname" . }}-receive-headless
-  updateStrategy:
-    type: {{ .Values.receive.updateStrategyType }}
-    {{- if (eq "OnDelete" .Values.receive.updateStrategyType) }}
+  replicas: {{ .Values.receive.distributor.replicaCount }}
+  strategy:
+    type: {{ .Values.receive.distributor.strategyType }}
+    {{- if (eq "Recreate" .Values.receive.distributor.strategyType) }}
     rollingUpdate: null
     {{- end }}
   selector:
     matchLabels: {{- include "common.labels.matchLabels" . | nindent 6 }}
-      app.kubernetes.io/component: receive
+      app.kubernetes.io/component: receive-distributor
   template:
     metadata:
       labels: {{- include "common.labels.standard" . | nindent 8 }}
-        app.kubernetes.io/component: receive
+        app.kubernetes.io/component: receive-distributor
         {{- if .Values.commonLabels }}
         {{- include "common.tplvalues.render" ( dict "value" .Values.commonLabels "context" $ ) | nindent 8 }}
         {{- end }}
-        {{- if .Values.receive.podLabels }}
-        {{- include "common.tplvalues.render" (dict "value" .Values.receive.podLabels "context" $) | nindent 8 }}
+        {{- if .Values.receive.distributor.podLabels }}
+        {{- include "common.tplvalues.render" (dict "value" .Values.receive.distributor.podLabels "context" $) | nindent 8 }}
         {{- end }}
       annotations:
         checksum/ojbstore-configuration: {{ include (print $.Template.BasePath "/objstore-secret.yaml") . | sha256sum }}
         {{- if (include "thanos.receive.createConfigmap" .) }}
         checksum/receive-configuration: {{ include (print $.Template.BasePath "/receive/configmap.yaml") . | sha256sum }}
         {{- end }}
-        {{- if .Values.receive.podAnnotations }}
-        {{- include "common.tplvalues.render" (dict "value" .Values.receive.podAnnotations "context" $) | nindent 8 }}
+        {{- if .Values.receive.distributor.podAnnotations }}
+        {{- include "common.tplvalues.render" (dict "value" .Values.receive.distributor.podAnnotations "context" $) | nindent 8 }}
         {{- end }}
     spec:
       {{- include "thanos.imagePullSecrets" . | nindent 6 }}
       serviceAccount: {{ include "thanos.serviceaccount.name" (dict "component" "receive" "context" $) }}
-      {{- if .Values.receive.hostAliases }}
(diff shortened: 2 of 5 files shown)


bitnami/thanos/README.md
```diff
@@ -946,6 +946,9 @@ Find more information about how to deal with common errors related to Bitnami’
 
 ## Upgrading
 
+### To 5.3.0
+
+This version introduces support for the receiver dual-mode implementation for Thanos [v0.22+](https://github.com/thanos-io/thanos/releases/tag/v0.22.0)
```
miguelaeh, line 951: Add an empty line after this one

bitnami/thanos/templates/receive/distributor.yaml
```diff
@@ -0,0 +1,153 @@
...
+      labels: {{- include "common.labels.standard" . | nindent 8 }}
+        app.kubernetes.io/component: receive-distributor
+        {{- if .Values.commonLabels }}
+        {{- include "common.tplvalues.render" ( dict "value" .Values.commonLabels "context" $ ) | nindent 8 }}
+        {{- end }}
+        {{- if .Values.receive.distributor.podLabels }}
+        {{- include "common.tplvalues.render" (dict "value" .Values.receive.distributor.podLabels "context" $) | nindent 8 }}
+        {{- end }}
+      annotations:
+        checksum/ojbstore-configuration: {{ include (print $.Template.BasePath "/objstore-secret.yaml") . | sha256sum }}
```
miguelaeh, line 32: there is a typo here 
```suggestion
        checksum/objstore-configuration: {{ include (print $.Template.BasePath "/objstore-secret.yaml") . | sha256sum }}
```

bitnami/thanos/templates/receive/distributor.yaml
```diff
@@ -0,0 +1,153 @@
...
+          image: {{ include "thanos.image" . }}
+          imagePullPolicy: {{ .Values.image.pullPolicy | quote }}
+          {{- if .Values.receive.containerSecurityContext.enabled }}
+          securityContext: {{- omit .Values.receive.containerSecurityContext "enabled" | toYaml | nindent 12 }}
+          {{- end }}
+          args:
+            - receive
+            - --log.level={{ .Values.receive.logLevel }}
+            - --log.format={{ .Values.receive.logFormat }}
+            - --grpc-address=0.0.0.0:10901
```
miguelaeh, line 79: We should not hardcode the ports

bitnami/thanos/templates/receive/distributor.yaml
```diff
@@ -0,0 +1,153 @@
...
+            - name: OBJSTORE_CONFIG
+              valueFrom:
+                secretKeyRef:
+                  key: objstore.yml
+                  name: {{ include "thanos.objstoreSecretName" . }}
+            {{- if .Values.receive.distributor.extraEnv }}
+            {{- toYaml .Values.receive.distributor.extraEnv | nindent 12 }}
+            {{- end }}
+          ports:
+            - containerPort: 10901
```
miguelaeh, line 106: We usually make the ports configurable

bitnami/thanos/values.yaml
```diff
@@ -2238,6 +2238,94 @@ receive:
...
+    ## resources, such as Minikube. If you do want to specify resources, uncomment the following
+    ## lines, adjust them as necessary, and remove the curly braces after 'resources:'.
+    ## @param receive.distributor.resources.limits The resources limits for the Thanos Receive container
+    ## @param receive.distributor.resources.requests The requested resources for the Thanos Receive container
+    ##
+    resources:
+    ## Example:
+    ## limits:
+    ##    cpu: 100m
+    ##    memory: 128Mi
```
miguelaeh, line 2259:
```suggestion
    ##    memory: 128Mi
    ##
```

bitnami/thanos/values.yaml
```diff
@@ -2238,6 +2238,94 @@ receive:
...
+    limits: {}
+    ## Examples:
+    ## requests:
+    ##    cpu: 100m
+    ##    memory: 128Mi
```
miguelaeh, line 2264:
```suggestion
    ##    memory: 128Mi
    ##
```

bitnami/thanos/values.yaml
```diff
@@ -2657,6 +2745,7 @@ receive:
     ##
     pathType: ImplementationSpecific
 
+
```
miguelaeh, line 2748: Remove extra line

miguelaeh (changes requested):
Hi @whitleykeith ,
Please take a look at the comments

whitleykeith:
@miguelaeh sorry for the delay in fixing this, just pushed an update with the requested changes

miguelaeh (commented):
Thank you very much for addressing the changes, it looks good to me but I a left a couple of comments, please check them when you have the time.

bitnami/thanos/templates/receive/statefulset.yaml
```diff
@@ -92,7 +92,9 @@ spec:
             {{- else }}
             - --receive.local-endpoint=$(NAME).{{ include "common.names.fullname" . }}-receive-headless.$(NAMESPACE).svc.{{ .Values.clusterDomain }}:10901
             {{- end }}
+            {{- if eq .Values.receive.mode "standalone" }}
```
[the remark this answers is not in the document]
whitleykeith: So the Stateful Set should always have `receive.local-endpoint` set in either mode since in both modes the stateful set is what's storing the TSDB blocks. The condition for the local endpoint flag is essentially whether or not to create the distributor deployment. 

The flags in question behave like this within these two manifests:

* StatefulSet: Either has both `--receive.hashrings` and `--receive.local-endpoint` set in Standalone or just `--receive.local-endpoint` in dual-mode. 
* Distributor: Isn't deployed in Standalone or deployed with  just `--receive.hashrings`  in dual-mode

bitnami/thanos/templates/receive/service.yaml
```diff
@@ -58,6 +58,10 @@ spec:
     {{- include "common.tplvalues.render" (dict "value" .Values.receive.service.labelSelectorsOverride "context" $) | nindent 4 }}
     {{- else }}
     {{- include "common.labels.matchLabels" . | nindent 4 }}
+    {{- if eq .Values.receive.mode "dual-mode"}}
+    app.kubernetes.io/component: receive-distributor
+    {{ else }}
     app.kubernetes.io/component: receive
+    {{ end }}
```
miguelaeh, line 65: Adding this here, when the `receive.mode` is `dual-mode`, there will not be any service to connect to the receive stateful set pod. Is that expected? Maybe a new service for the distributor would have more sense?

bitnami/thanos/values.yaml
```diff
@@ -2290,6 +2290,97 @@ receive:
...
+    ## @param receive.distributor.resources.limits The resources limits for the Thanos Receive container
+    ## @param receive.distributor.resources.requests The requested resources for the Thanos Receive container
+    ##
+    resources:
+    ## Example:
+    ## limits:
+    ##    cpu: 100m
+    ##    memory: 128Mi
+    ##
+    limits: {}
```
ethernoy, line 2314: `limits` and `requests` here seem to be properties of `resources` and hence should have one more level of indentation

bitnami/thanos/templates/receive/service.yaml
whitleykeith: The reasoning behind that was because the distributor handles all routing to the receive pods, and it leverages the headless service for that. Functionally a service hitting the stateful set doesn't do much if the distributor is enabled, and I felt it would be more confusing to have another service (esp. if both services would have a remote_write port)

bitnami/thanos/values.yaml
whitleykeith: yep good catch, fixed

whitleykeith:
> Hi! I left some other comments, also I have a couple of questions:
> 
> 1. Why are we calling the new mode `dual-mode` if it is called `distributor mode` in the docs you linked ?
> 2. You shared the proposal document, is this already supported by thanos or is it just something for the future?

This was implemented in [Thanos v0.22](https://github.com/thanos-io/thanos/releases/tag/v0.22.0). In the release notes for that version, they referred to this feature as `dual-mode` which is why I called it that instead of the distributor mode as it seemed dual-mode was the "official" name

miguelaeh (approved):
LGTM!

PR #7149, opened 2021-08-05
[bitnami/keycloak] Include keystoreFilename and truststoreFilename keys

### LINKED ISSUE #7086: [bitnami/keycloak] Make the TLS key/crt file names configurable
**Which chart**:
bitnami/keycloak: version 4.0.3

I would like to configure TLS with existing certs, but I am finding the helm chart does not easily support certs created outside of the keycloak helm chart. I am using Cert-Manager to create certificates (and keystores/truststores). The current keycloak helm chart requires that the `.Values.auth.tls.existingSecret` have secret keys in the form of `keycloak-${ID}.key`, `keycloak-${ID}.crt`, and `ca.crt`. The `ca.crt` is standard, so that isn't an issue. However, I'm unable to get cert manager to create secrets with custom `keycloak-${ID}.key` and `keycloak-${ID}.crt` names. Further, cert manager will create a `keystore.jks` and a `truststore.jks` file, but these names are not configurable either.

Here is the relevant template: https://github.com/bitnami/charts/blob/38438c70afa090a5ac1733af724abfe7eb5f4fd9/bitnami/keycloak/templates/statefulset.yaml#L86-L114

Additionally, it's relevant to point out that because the `/cert` directory is mounted with limited file system permissions, it is not possible to create a keystore/truststore in an earlier init container and then have those stores used by keycloak.

In some cases, I would also like to use the host's default trusted ca certificate store, as the ca certs are for well know issuers.

I am new to this helm chart, so I don't have a strong sense of what the "right" solution should be. Below, I share some ideas.

It could be possible to add a conditional block to the above `initContainer` where end users can supply the names of the `.key` and `.crt` files via the `.Values` file to use when creating the `truststore.jks` and the `keystore.jks`.

Similarly, the truststore and keystore file names could be configurable values supplied by end users.

alvneiayu:
**Description of the change**

Include truststoreFilename and keystoreFilename for setting up custom truststore and keystore filenames.

**Benefits**

Make it simple to set up a keystore and truststore filename for special circumstances like to use the same keystore and truststore generated by cert-manager

**Possible drawbacks**

Those keys can be used just to use the same keystore and truststore in all the pods. Using this keys, you can not set up multiples keystore as we are doing without using those keys. 

**Applicable issues**

  - fixes #7086

**Checklist** 


michaeljmarshall:
@alvneiayu and @javsalgar - I pulled down this branch and ran some basic tests against it on minikube. I can confirm this change works as expected for my use case. When this is merged and released, I plan to pull the update into the DataStax Pulsar Helm chart: https://github.com/datastax/pulsar-helm-chart.

Just in case anyone would find my test values helpful, here they are:

```yaml
keycloak:
  enabled: true
  auth:
    tls:
      enabled: true
      jksSecret: "keycloak-pulsar-tls"
      # These must be the same because the cert manager will use the keystorePassword to protect both
      # the keystore.jks and the truststore.jks.
      truststorePassword: "kU4Fv3Qb8zfe-BK7drdhsX.pmau9FR"
      keystorePassword: "kU4Fv3Qb8zfe-BK7drdhsX.pmau9FR"
      truststoreFilename: "truststore.jks"
      keystoreFilename: "keystore.jks"
```

And here is the sample `Certificate` resource definition that Cert Manager used to create the secret that was then mounted in to the keycloak pod.

```yaml
apiVersion: cert-manager.io/v1
kind: Certificate
metadata:
  name: pulsar-keycloak-tls
  namespace: default
spec:
  dnsNames:
  - test-keycloak.default.svc.cluster.local
  - test-keycloak.default
  - test-keycloak
  issuerRef:
    name: pulsar-ca-issuer
  keystores:
    jks:
      create: true
      passwordSecretRef:
        key: tls-keystore-password
        name: test-keycloak
  secretName: keycloak-pulsar-tls
```

rafariossaa:
Hi @michaeljmarshall 
  Thanks for your feedback.
  I am merging this.

PR #7150, opened 2021-08-05
[bitnami/rabbitmq] add service.portEnabled

mattmichal:
**Description of the change**

Add a parameter that disables the AMQP port on the Service and the Pod. This parameter is only considered if the `auth.tls.enabled` parameter is set to true so that one AMQP port is always enabled. NOTES.txt template has been updated to display the proper port.

**Benefits**

Service and Pod port mappings accurately reflect the desired behavior: if TLS is enabled, clients should not be able to connect without TLS.

**Possible drawbacks**

RMQ server will still be listening on the AMQP port unless `listeners.tcp = none` is given in `additionalConfiguration`.

**Checklist** 


alemorcuq (approved):
Thanks for the PR, @mattmichal. This looks good to me. However, I will also ask @javsalgar to review this.

javsalgar (approved):

PR #7166, opened 2021-08-07
Fix template replicaCount

fengsi:
**Description of the change**
There should be a typo around `replicaCount` in the template.  The `%%MAIN_OBJECT_BLOCK%%` placeholder was missing in `template/CHART_NAME/templates/deployment.yaml` for some reason, though it's documented in `template/CHART_NAME/values.yaml`.


pablogalegoc (approved):
Hi @fengsi! Thank you for your contribution, LGTM!

PR #7170, opened 2021-08-09
[bitnami/mysql] Improve default values for probes

juan131:

**Description of the change**

We're enabling the startup probes by default in MySQL. However, we're still using huge values for `initialDelaySeconds` properties in the probes which doesn't make sense at all. Instead, we should a combination `failureThreshold * periodSeconds` long enough to cover the worse case startup time.

More info at:

- https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-startup-probes

**Benefits**

The MySQL installation is ready much faster.

**Possible drawbacks**

None

**Applicable issues**

N/A

**Additional information**

N/A

**Checklist** 

### DIFF
--- a/bitnami/mysql/Chart.yaml
+++ b/bitnami/mysql/Chart.yaml
@@ -25,4 +25,4 @@ name: mysql
 sources:
   - https://github.com/bitnami/bitnami-docker-mysql
   - https://mysql.com
-version: 8.8.2
+version: 8.8.3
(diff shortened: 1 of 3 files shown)


Mauraza (approved):
Hi @juan131, 

Huge thanks for the PR! LGTM!!

PR #7172, opened 2021-08-09
Fix template image section

fengsi:
**Description of the change**

Similar to #7166, there were some obvious typos in the templates.


carrodher (approved):
Thanks for the fixes!

PR #7177, opened 2021-08-10
[bitnami/haproxy] Adding condition to serviceaccount

GGre4:
**Description of the change**

Adding serviceAccount.create = false in the values was being ignored and creating a new service account each time a new deployment was made.  

- Added condition to the service account template


bitnami/haproxy/templates/service-account.yaml
```diff
@@ -1,13 +1,21 @@
+{{- if .Values.serviceAccount.create }}
```
pablogalegoc, line 1: I would keep just this condition and remove the unrelated changes

bitnami/haproxy/templates/service-account.yaml
```diff
@@ -1,13 +1,21 @@
...
+  {{- if or .Values.commonAnnotations .Values.serviceAccount.annotations }}
+  annotations:
+    {{- if or .Values.commonAnnotations }}
+    {{- include "common.tplvalues.render" ( dict "value" .Values.commonAnnotations "context" $ ) | nindent 4 }}
+    {{- end }}
+    {{- if .Values.serviceAccount.annotations }}
+    {{- include "common.tplvalues.render" ( dict "value" .Values.serviceAccount.annotations "context" $ ) | nindent 4 }}
+    {{- end }}
   {{- end }}
+{{- end }}
```
pablogalegoc, line 21: New line need here

pablogalegoc (changes requested):
Hi @GGre4! Thanks for your contribution! I left some comments below. Also remember that a patch version bump of the chart will be needed to pass the lint test 🙂

bitnami/haproxy/templates/service-account.yaml
GGre4: Hi @pablogalegoc Thanks for your feedback, first time contributing to a project. 

I have updated the PR accordingly

pablogalegoc: Hey @GGre4, that's awesome! This kind of little changes are very useful and help building confidence.
I asked to keep the changes just to what's intended for the PR because then it's easier to find them and reason about them 🙂

bitnami/haproxy/templates/service-account.yaml
pablogalegoc: @GGre4 remember to add a new line here. If you are using the Github editor it sometimes messes with the endlines, it is better to change it locally and push

pablogalegoc: Yep, that's it

pablogalegoc (approved):
Thank you @GGre4! LGTM!

PR #7193, opened 2021-08-11
[bitnami/wordpress] Add "extraContainerPorts" parameter

mamiu:
**Description of the change**

Add a separate parameter `extraContainerPorts` to expose custom ports of the
WordPress container.
This is especially useful if a custom image is used.

Furthermore there's already an `extraContainerPorts` parameter for the WordPress
service, so this change completes the current configuration options.

**Benefits**

The chart doesn't has to be forked and customized but easily configured with the `extraContainerPorts` variable.

**Possible drawbacks**

Unknown (probably none).

**Checklist** 


migruiz4 (changes requested):
Hi @mamiu,

Could you please rename this feature as `extraContainerPorts` as we do in `bitnami/rabbitmq`?

https://github.com/bitnami/charts/blob/22f189d36c70da0f46733fd5da548f6ea6ec5d28/bitnami/rabbitmq/templates/statefulset.yaml#L233-L235

That way we also can avoid confusion with `service.extraPorts`, which is present in most of the charts including `bitnami/wordpress`.

mamiu:
@migruiz4 Thanks for your review. Did all the changes you asked for.

migruiz4 (approved):
Thank you very much for your contribution and prompt response!

Looks good to me!

mamiu:
Sure thing. Thanks for merging it!

PR #5814, opened 2021-03-17
[bitnami/*] Allow git ssh connections

andresbono:
**Description of the change**

Enable git ssh connections by loading the new git [entrypoint.sh](https://github.com/bitnami/bitnami-docker-git/blob/3c86ff8e824b2716e77afc3e27856bb12152d84e/2/debian-10/rootfs/opt/bitnami/scripts/git/entrypoint.sh) and allowing extra volumes to be mounted.

**Applicable issues**

  - fixes #3043
  - fixes https://github.com/bitnami/bitnami-docker-git/issues/2

**Additional information**

**Checklist** 
- [x] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [x] Variables are documented in the README.md
- [x] Title of the PR starts with chart name (e.g. `[bitnami/chart]`)

PR #5188, opened 2021-01-22
[bitnami/postgresql-ha] Remove support to load conf/scripts from 'files/' directory

juan131:
Signed-off-by: juan131 <juanariza@vmware.com>

**Description of the change**

This PR:

- Removes support to load configuration and init scripts from `files/` directory
- Treats `postgresql.configuration` & `postgresql.extendedConf` parameters as plain configuration (instead of doing weird `camelCase` to `sneak_case` transformation). 

**Benefits**

Avoid confusion and have a most intuitive approach for loading custom configuration & scripts.

**Possible drawbacks**

Users can no longer load their configuration from the `files/` directory.

**Applicable issues**

  - fixes https://github.com/bitnami/charts/issues/5075

**Additional information**

Adding custom configuration files or initialization scripts by placing them under the "files" directory is very confusing for users since they do not usually clone the repo nor they fetch the charts to their working directories.
On the other hand, doing a `camelCase` to `sneak_case` transformation to pass custom configuration for PostgreSQL makes no sense at all.

**Checklist**

- [x] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [x] Variables are documented in the README.md
- [x] Title of the PR starts with chart name (e.g. `[bitnami/chart]`)

PR #7199, opened 2021-08-12
[bitnami/pytorch] support for files/**

chenkins:
**Description of the change**

Recursively add files under `files` folder, not only top-level files.

**Benefits**

Flexibility of recursively copying folder structure into containers.

**Possible drawbacks**

**Applicable issues**

None,

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/pytorch/Chart.yaml
+++ b/bitnami/pytorch/Chart.yaml
@@ -24,4 +24,4 @@ name: pytorch
 sources:
   - https://github.com/bitnami/bitnami-docker-pytorch
   - http://pytorch.org/
-version: 2.3.12
+version: 2.4.0
--- a/bitnami/pytorch/README.md
+++ b/bitnami/pytorch/README.md
@@ -170,14 +170,15 @@ Bitnami will release a new chart updating its containers if a new version of the
 The PyTorch chart supports three different ways to load your files. In order of priority, they are:
 
   1. Existing config map
-  2. Files under the `files` directory
+  2. Files under the `files` directory (including subfolders)
   3. Cloning a git repository
+
 
 This means that if you specify a config map with your files, it won't look for the `files/` directory nor the git repository.
 
 In order to use use an existing config map, set the `configMap=my-config-map` parameter.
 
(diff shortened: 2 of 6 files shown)


juan131 (commented):
Hi @chenkins 

Before start reviewing, I'd like to ask you about what's the use case you want to cover with this contribution.

We're trying to deprecate the possibility of loading configuration and scripts from the `files/` directory since it's a confusing method for many users who don't recall they actually need to clone the repo or fetch the chart before installing it. See for instance PRs like the one below:

- https://github.com/bitnami/charts/pull/5188

chenkins:
Hi @juan131 

Thanks for your message.

Our use case is having code in non-public git repository with sub-folder (or even only local files in an interactive development scenario that needs to be tested in a deployment) - the proper way for the first scenario would probably be along the lines of https://www.jannikarndt.de/blog/2018/03/ssh_key_as_kubernetes_secret/, but would require some more effort, I guess?

Oh, there it is: https://github.com/bitnami/charts/pull/5814/files. Documentation could be improved by adding an example on how to inject the ssh key from a secret into `extraVolumes` in `values.yaml` and also pointing out in `README.md`  `git.extraVolumeMounts` can be used for ssh.

I think it's still good to have `files/**` for a "quick-and-dirty" scenario - how about adding a warning to the files approach - "appropriate for a dev scenario only, we recommend using git clone approach"?

In addition,  there seems to be caveat with this `files/**`: updating the configmap seems not trigger an update of the k8s deployment/statefulset, need to check this again, though.

juan131:
Hi @chenkins 

> Documentation could be improved by adding an example on how to inject the ssh key from a secret into extraVolumes in values.yaml and also pointing out in README.md git.extraVolumeMounts can be used for ssh.

Did you take a look to the section below of the README.md file?

- https://github.com/bitnami/charts/tree/master/bitnami/pytorch#loading-your-files

> I think it's still good to have files/** for a "quick-and-dirty" scenario - how about adding a warning to the files approach - "appropriate for a dev scenario only, we recommend using git clone approach"?

Our plan is to deprecate the possibility of loading configuration and scripts from the `files/` directory. We know it's a useful for development purposes but as I mentioned before it's confusing and it's not suitable for production environments.

PR #7211, opened 2021-08-12
[bitnami/kibana] add serviceaccout

### LINKED ISSUE #7059: Kibana chart should have service account

Mauraza:
**Description of the change**

Add service account to Kibana 
**Benefits**

**Possible drawbacks**

**Applicable issues**

  - fixes #7059

**Additional information**

This continues working of https://github.com/bitnami/charts/pull/7060

**Checklist** 

### DIFF
--- a/bitnami/kibana/templates/deployment.yaml
+++ b/bitnami/kibana/templates/deployment.yaml
@@ -73,6 +73,7 @@ spec:
           securityContext:
             runAsUser: {{ .Values.securityContext.runAsUser }}
           {{- end }}
+          serviceAccountName: {{ .Values.serviceAccount.name }}
           env:
             - name: KIBANA_PORT_NUMBER
               value: {{ .Values.containerPort | quote }}
--- a/bitnami/kibana/templates/deployment.yaml
+++ b/bitnami/kibana/templates/deployment.yaml
@@ -73,7 +73,7 @@ spec:
           securityContext:
             runAsUser: {{ .Values.securityContext.runAsUser }}
           {{- end }}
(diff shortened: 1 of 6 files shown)


bitnami/kibana/templates/deployment.yaml
```diff
@@ -73,6 +73,7 @@ spec:
           securityContext:
             runAsUser: {{ .Values.securityContext.runAsUser }}
           {{- end }}
+          serviceAccountName: {{ .Values.serviceAccount.name }}
```
juan131, line 76: Use the macro here:
```suggestion
          serviceAccountName: {{ include "kibana.serviceAccountName" . }}
```

juan131 (approved):
LGTM

jeff-hammel-leapyear:
Thank you very much!

PR #7213, opened 2021-08-12
[bitnami/external-dns] Adds namespace entries to several templates

### LINKED ISSUE #7210: [external-dns] Resources are applied to a wrong namespace
**Which chart**:
The name (and version) of the affected chart
external-dns

**Describe the bug**
When using helm template command, the namespace metadata associated to namespaced resources is not consistent. For example, the deployment has set the namespace, but configmaps and secret don't, so the apply of these manifests could result in pods deployed on a different namespace than other resources. 

I think all namespaced objects should have set the namespace metadata to avoid this situation. 

**To Reproduce**

1. Execute helm template to generate k8s manifests using the option --namespace to indicate a specific one.
2. Apply these manifests with kubectl (but with a default namespace configured with a different namespace). 
3. You can see that the deployment has been deployed to the namespace indicated to helm, but other resources as configmaps or secrets have been deployed to the kubectl default namespace. 

**Expected behavior**
All resources should be deployed to the namespace specified via helm. 

- Output of `helm version`:
```
version.BuildInfo{Version:"v3.1.1", GitCommit:"afe70585407b420d0097d07b21c47dc511525ac8", GitTreeState:"clean", GoVersion:"go1.13.8"}
```

```
Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.8", GitCommit:"9f2892aab98fe339f3bd70e3c470144299398ace", GitTreeState:"clean", BuildDate:"2020-08-14T11:09:22Z", GoVersion:"go1.14.7", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.2", GitCommit:"092fbfbf53427de67cac1e9fa54aaa09a28371d7", GitTreeState:"clean", BuildDate:"2021-06-17T16:37:00Z", GoVersion:"go1.16.5", Compiler:"gc", Platform:"linux/amd64"}```

**Additional context**
Add any other context about the problem here.

pipeworks-asmith:
**Description of the change**

Adds a metadata.namespace entry to a handful of namespace-scoped resources. These would work before when using `helm install -n <namespace>`, but when generating the manifest from `helm template -n <namespace>` those resources would be placed in default, instead.

**Benefits**

Allows users to generate correct manifests from Helm

**Possible drawbacks**

If users relied on the bad behavior, it is now gone

**Applicable issues**

  - fixes #7210 

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/external-dns/Chart.yaml
+++ b/bitnami/external-dns/Chart.yaml
@@ -24,4 +24,4 @@ sources:
   - https://github.com/kubernetes-sigs/external-dns
   - https://github.com/bitnami/bitnami-docker-external-dns
   - https://github.com/kubernetes-sigs/external-dns
-version: 5.4.0
+version: 5.4.1
--- a/bitnami/external-dns/templates/configmap.yaml
+++ b/bitnami/external-dns/templates/configmap.yaml
@@ -3,6 +3,7 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   name: {{ template "external-dns.fullname" . }}
+  namespace: {{ .Release.Namespace }}
   labels: {{ include "external-dns.labels" . | nindent 4 }}
 data:
 {{- if .Values.designate.customCA.enabled }}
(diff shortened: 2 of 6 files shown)


carrodher (approved):
LGTM, thanks!

Issue #5948, opened 2021-03-29
Mongo Credentials using credential file injected using vault.


gau31rav:
Which chart:
bitnami/mongodb 

Is your feature request related to a problem? Please describe.
We are trying to get Vault agent injector to provide the credentials used by Mongo so that there's a single source of truth for the secrets used in cluster. (https://www.vaultproject.io/docs/platform/k8s/injector)

Describe the solution you'd like
When the CredentialsFile is avaiable, it would be nice if mongo reads the credentials from the file mounted by Vault and then pass it as the required env variables. The format of credentials files can be made generic.

Same request has been made for other charts earlier.
https://github.com/bitnami/charts/issues/4422

miguelaeh:
Hi @gau31rav ,
Would you like to create the PR adding that feature?

PR #7221, opened 2021-08-13
Mongo Credentials using credential file injected using vault. #5948

gau31rav:
Which chart:
bitnami/mongodb

Is your feature request related to a problem? Please describe.
We are trying to get the Vault agent injector to provide the credentials used by Mongo so that there's a single source of truth for the secrets used in the cluster. (https://www.vaultproject.io/docs/platform/k8s/injector)

Describe the solution you'd like
When the CredentialsFile is available, it would be nice if mongo reads the credentials from the file mounted by Vault and then pass it as the required env variables. The format of credentials files can be made generic.

The same request has been made for other charts earlier.
#4422

This PR address reading credentials from the vault.
https://github.com/bitnami/charts/issues/5948

### DIFF
--- a/bitnami/common/templates/validations/_mongodb.tpl
+++ b/bitnami/common/templates/validations/_mongodb.tpl
@@ -19,29 +19,33 @@ Params:
   {{- $valueKeyPassword := printf "%s.password" $authPrefix -}}
   {{- $valueKeyReplicaSetKey := printf "%s.replicaSetKey" $authPrefix -}}
   {{- $valueKeyAuthEnabled := printf "%s.enabled" $authPrefix -}}
+  {{- $bypassUpgradeValidation := include "common.mongodb.values.bypassUpgradeValidation" . -}}
 
   {{- $authEnabled := include "common.utils.getValueFromKey" (dict "key" $valueKeyAuthEnabled "context" .context) -}}
 
-  {{- if and (not $existingSecret) (eq $enabled "true") (eq $authEnabled "true") -}}
-    {{- $requiredPasswords := list -}}
+  {{- if eq $bypassUpgradeValidation "false" -}}
 
-    {{- $requiredRootPassword := dict "valueKey" $valueKeyRootPassword "secret" .secret "field" "mongodb-root-password" -}}
-    {{- $requiredPasswords = append $requiredPasswords $requiredRootPassword -}}
+    {{- if and (not $existingSecret) (eq $enabled "true") (eq $authEnabled "true") -}}
+      {{- $requiredPasswords := list -}}
 
-    {{- $valueUsername := include "common.utils.getValueFromKey" (dict "key" $valueKeyUsername "context" .context) }}
-    {{- $valueDatabase := include "common.utils.getValueFromKey" (dict "key" $valueKeyDatabase "context" .context) }}
-    {{- if and $valueUsername $valueDatabase -}}
-        {{- $requiredPassword := dict "valueKey" $valueKeyPassword "secret" .secret "field" "mongodb-password" -}}
-        {{- $requiredPasswords = append $requiredPasswords $requiredPassword -}}
+      {{- $requiredRootPassword := dict "valueKey" $valueKeyRootPassword "secret" .secret "field" "mongodb-root-password" -}}
+      {{- $requiredPasswords = append $requiredPasswords $requiredRootPassword -}}
+
+      {{- $valueUsername := include "common.utils.getValueFromKey" (dict "key" $valueKeyUsername "context" .context) }}
+      {{- $valueDatabase := include "common.utils.getValueFromKey" (dict "key" $valueKeyDatabase "context" .context) }}
+      {{- if and $valueUsername $valueDatabase -}}
+          {{- $requiredPassword := dict "valueKey" $valueKeyPassword "secret" .secret "field" "mongodb-password" -}}
+          {{- $requiredPasswords = append $requiredPasswords $requiredPassword -}}
+      {{- end -}}
+
+      {{- if (eq $architecture "replicaset") -}}
+          {{- $requiredReplicaSetKey := dict "valueKey" $valueKeyReplicaSetKey "secret" .secret "field" "mongodb-replica-set-key" -}}
+          {{- $requiredPasswords = append $requiredPasswords $requiredReplicaSetKey -}}
+      {{- end -}}
+
+      {{- include "common.validations.values.multiple.empty" (dict "required" $requiredPasswords "context" .context) -}}
+
     {{- end -}}
-
-    {{- if (eq $architecture "replicaset") -}}
-        {{- $requiredReplicaSetKey := dict "valueKey" $valueKeyReplicaSetKey "secret" .secret "field" "mongodb-replica-set-key" -}}
-        {{- $requiredPasswords = append $requiredPasswords $requiredReplicaSetKey -}}
-    {{- end -}}
-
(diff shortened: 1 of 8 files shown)


bitnami/mongodb/values.yaml
```diff
@@ -143,6 +146,14 @@ auth:
...
   ## @param auth.database MongoDB&reg; custom database
   ##
+  # username: username
+  # password: password
+  # database: database
+  ## By default custom user role is readWrite to database configured above
+  # userRole: readWrite
+  #vaultCredentials: /tmp/credentialFile
+  ## Key used for replica set authentication
+  ## Ignored when mongodb.architecture=standalone
```
juan131, line 156: Did you accidentally add these comments? I mean, the unique new parameter I can see here is `auth. vaultCredentials`

bitnami/mongodb/values.yaml
```diff
@@ -123,6 +123,9 @@ schedulerName: ""
 ## @param architecture MongoDB&reg; architecture (`standalone` or `replicaset`)
 ##
 architecture: standalone
+
+bypassUpgradeValidation: false
+
```
juan131, line 128: Can't we implement the logic to detect when to bypass the validations instead? I mean, we can simply detect that Vault is being used to manage the credentials and skip the validations.

bitnami/mongodb/templates/standalone/scripts-configmap.yaml
```diff
@@ -0,0 +1,26 @@
...
+        declare $line; 
+        done < "$VAULT_CREDENTIAL_FILE"
+        export MONGODB_ROOT_PASSWORD="$MONGODB_ROOT_PASSWORD"
+        export MONGODB_USERNAME="$MONGODB_USERNAME"
+        export MONGODB_PASSWORD="$MONGODB_PASSWORD"
+        export MONGODB_DATABASE="$MONGODB_DATABASE"
+        export MONGODB_USER_ROLE="${MONGODB_USER_ROLE:-readWrite}"
+    fi
+    exec /opt/bitnami/scripts/mongodb/entrypoint.sh /opt/bitnami/scripts/mongodb/run.sh
+{{- end }}
```
juan131, line 26: This logic should be managed by the container. I mean, if we need to support a way to retrieve the credentials from a file (instead of the env. variables `MONGODB_ROOT_PASSWORD`, `MONGODB_USERNAME`, `MONGODB_PASSWORD`, etc.) that logic should be implemented at container level.

bitnami/mongodb/templates/standalone/dep-sts.yaml
```diff
@@ -180,18 +180,25 @@ spec:
...
+              value: {{ ternary "true" "false" .Values.image.debug | quote }}
+            {{- if (and .Values.auth.enabled (not .Values.auth.vaultCredentials)) }}
             {{- if .Values.auth.username }}
             - name: MONGODB_USERNAME
               value: {{ .Values.auth.username | quote }}
             {{- end }}
+            {{- if .Values.auth.userRole }}
+            - name: MONGODB_USER_ROLE
+              value: {{ .Values.auth.userRole | quote }}
+            {{- end }}
```
juan131, line 201: I don't think this env. variable is being taken into account.

bitnami/common/templates/validations/_mongodb.tpl
```diff
@@ -106,3 +110,16 @@ Params:
...
+Params:
+  - subchart - Boolean - Optional. Whether to bypassUpgradeValidation for password environment variables. Default: false
+*/}}
+{{- define "common.mongodb.values.bypassUpgradeValidation" -}}
+  {{- if .subchart -}}
+    {{- .context.Values.mongodb.bypassUpgradeValidation -}}
+  {{- else -}}
+    {{- .context.Values.bypassUpgradeValidation -}}
+  {{- end -}}
+{{- end -}}
```
juan131, line 125: Apart from using Vault to manage the credentials, under what other circumstances are you interested on bypassing the validations? If Vault is the unique case, then let's skip the validations when `auth.vaultCredentials` is set

juan131 (changes requested):
Thanks so much for this contribution @gau31rav !

I would split the effort of supporting credentials from Vault in two parts:

- Support retrieving the credentials from a mounted file (instead of using env. vars) in the [Bitnami MongoDB container](https://github.com/bitnami/bitnami-docker-mongodb).
- Support mounting a file with the credentials from a Vault secret in the MongoDB container (here in the MongoDB chart)

Other observations:

- Please update the version in the `Chart.yaml` 
- We have a tool named [readme-generator-for-helm](https://github.com/bitnami-labs/readme-generator-for-helm) that automatically updates the **README.md** parameters' tables based on the **values.yaml**. Please use it to update the tables with the new parameters information.

bitnami/mongodb/values.yaml
gau31rav: This is for the upgrade use case. When we try to upgrade it from the existing deployment where the credentials are already saved as Mongo Env Vars.

bitnami/mongodb/templates/standalone/dep-sts.yaml
gau31rav: This is to add another fix. When the custom user created in a container can have a custom role.  As of now, the mongo container is creating a custom user with only read-write access to one database. With this parameter, we can assign a custom role to the Mongo custom user. Please refer the below PR.
https://github.com/bitnami/bitnami-docker-mongodb/pull/281

bitnami/common/templates/validations/_mongodb.tpl
gau31rav: This is added to support the upgrade use case, will revisit it and check if the any other use case needs to be covered.

bitnami/mongodb/templates/standalone/scripts-configmap.yaml
gau31rav: There was a challenge implementing it container level. The details are not on top of my mind, will find the details and share. Will revisit and try to find a way if we can address those.

bitnami/mongodb/values.yaml
gau31rav: vaultCredentials is the configuration holding the file path which is mounted by the vault. This is added as a comment to provide an example, on how to use it.

juan131: Then we should make it very clear that's an example. Following the format
```
## @param {{PARAM_NAME}} {{PARAM_DESCRIPTION}}
## e.g:
## {{PARAM_NAME}}:
##   ...
##   ...
##
```
You can use this example as reference: https://github.com/bitnami/charts/blob/master/bitnami/mongodb/values.yaml#L1000

bitnami/mongodb/values.yaml
juan131: I don't find the parameter useful yet. I mean, regardless I'm upgrading from an existing deployment with credentials as env. vars or already using Vault, why can't I simply assume that validations should be skipped when Vault is enabled?

bitnami/common/templates/validations/_mongodb.tpl
juan131: Answered above.

juan131:
I just added some new comments @gau31rav 

I'm so sorry that I lost the track of this PR @gau31rav ! I apologize for that.

juan131:
@gau31rav did you have a chance to take a look to my comments?

PR #7223, opened 2021-08-13
Fix template logic

fengsi:
**Description of the change**

The base template is very much broken unless these below are fixed:

1. Fix logic about `diagnosticMode`
2. Remove non-existing `persistence` section (should have been ported from other charts while not used here – but why breaking the base template?)
3. Add missing quotes around `%%COMPONENT_NAME%%`


carrodher:
Thanks for your contribution, changes LGTM but I just found the same issues in _daemonset.yaml_ and _statefulset.yaml_; would you like to apply the same changes in those files as well? Thanks!

fengsi:
> Thanks for your contribution, changes LGTM but I just found the same issues in _daemonset.yaml_ and _statefulset.yaml_; would you like to apply the same changes in those files as well? Thanks!

Good catch!  Just pushed. 👍

carrodher:
Thanks for addressing the issue in the different files!

carrodher (approved):

PR #7238, opened 2021-08-16
[bitnami/kubernetes-event-exporter] Add extraEnvVars support

nickgryg:
**Description of the change**

`extraEnvVars` variable is described in README/documentation but isn't supported in chart itself.

**Benefits**

We can inject ENV variables into containers from now on.

**Possible drawbacks**

None.

**Checklist** 

### DIFF
--- a/bitnami/kubernetes-event-exporter/templates/deployment.yaml
+++ b/bitnami/kubernetes-event-exporter/templates/deployment.yaml
@@ -58,9 +58,19 @@ spec:
           imagePullPolicy: {{ .Values.image.pullPolicy }}
           args:
             - -conf=/data/config.yaml
-          {{- if .Values.extraEnvVars }}
-          env:
-            {{- include "common.tplvalues.render" (dict "value" .Values.extraEnvVars "context" $) | nindent 12 }}
+          {{- if .Values.extraEnv }}
+          env: {{- include "common.tplvalues.render" (dict "value" .Values.extraEnv "context" $) | nindent 12 }}
+          {{- end }}
+          {{- if or .Values.extraEnvVarsCM .Values.extraEnvVarsSecret }}
+          envFrom:
+            {{- if .Values.extraEnvVarsCM }}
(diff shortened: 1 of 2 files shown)


bitnami/kubernetes-event-exporter/templates/deployment.yaml
```diff
@@ -58,6 +58,10 @@ spec:
           imagePullPolicy: {{ .Values.image.pullPolicy }}
           args:
             - -conf=/data/config.yaml
+          {{- if .Values.extraEnvVars }}
+          env:
+            {{- include "common.tplvalues.render" (dict "value" .Values.extraEnvVars "context" $) | nindent 12 }}
+          {{- end }}
```
juan131, line 64: We should follow the same approach used on other charts and allow custom env. variables from existing secrets and configmaps also:
```suggestion
          {{- if .Values.extraEnv }}
          env: {{- include "common.tplvalues.render" (dict "value" .Values.extraEnv "context" $) | nindent 12 }}
          {{- end }}
          {{- if or .Values.extraEnvVarsCM .Values.extraEnvVarsSecret }}
          envFrom:
            {{- if .Values.extraEnvVarsCM }}
            - configMapRef:
                name: {{ include "common.tplvalues.render" (dict "value" .Values.extraEnvVarsCM "context" $) }}
            {{- end }}
            {{- if .Values.extraEnvVarsSecret }}
            - secretRef:
                name: {{ include "common.tplvalues.render" (dict "value" .Values.extraEnvVarsSecret "context" $) }}
            {{- end }}
          {{- end }}
```

juan131 (changes requested):
Thanks for the fix!! It seems the values exist but they're not used anywhere, see:

- https://github.com/bitnami/charts/blob/master/bitnami/kubernetes-event-exporter/values.yaml#L201

Please check my comments to support other alternatives too.

juan131 (approved):
LGTM

Issue #7242, opened 2021-08-17
[bitnami/testlink]  Ingress is broken


pgorlewar:
Helm upgrade failed: error validating "": error validating data: ValidationError(Ingress.spec.rules[0].http.paths[0].backend): unknown field "service" in io.k8s.api.extensions.v1beta1.IngressBackend

pablogalegoc:
Thanks for flagging this @pgorlewar! I see you've already opened a PR to fix this issue so a member of the team will review it

PR #7245, opened 2021-08-17
bitnami/testlink-7242-Update-ingress.yaml-to-fix-bad-ingress-behavior

pgorlewar:
**Description of the change**

Added {{ include "common.capabilities.ingress.apiVersion" . }} instead of hardcoded extensions/v1beta1

**Benefits**

The Ingresses of this chart will work in more cluster / work at all.

**Possible drawbacks**

Hopefully none. Because it is keeping track off in the bitnami/common chart.

**Applicable issues**

[bitnami/testlink] Ingress is broken #7242


marcosbc (approved):

PR #7246, opened 2021-08-17
[bitnami/redis] Fix clustering when service port != container port

javsalgar:
**Description of the change**

The clustering was configured using the headless service (that will return the pod IPs) with the service IPs (which is not present in the containers). This was causing conflicts when changing the service port. This PR fixes this. 

**Benefits**

**Possible drawbacks**

**Applicable issues**

  - fixes #

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/redis/templates/scripts-configmap.yaml
+++ b/bitnami/redis/templates/scripts-configmap.yaml
@@ -20,7 +20,7 @@ data:
     . /opt/bitnami/scripts/libvalidations.sh
 
     myip=$(hostname -i)
-    
+
     # If there are more than one IP, use the first IPv4 address
     if [[ "$myip" = *" "* ]]; then
         myip=$(echo $myip | awk '{if ( match($0,/([0-9]+\.)([0-9]+\.)([0-9]+\.)[0-9]+/) ) { print substr($0,RSTART,RLENGTH); } }')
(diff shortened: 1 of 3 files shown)


marcosbc (approved):
LGTM!

bitnami/redis/templates/scripts-configmap.yaml
```diff
@@ -241,20 +241,19 @@ data:
...
         sleep {{ .Values.sentinel.cleanDelaySeconds }}
     done
     info "Sentinels clean up done"
 
-
     sentinel_conf_set "sentinel monitor" "{{ .Values.sentinel.masterSet }} "$REDIS_MASTER_HOST" "$REDIS_MASTER_PORT_NUMBER" {{ .Values.sentinel.quorum }}"
 
     add_replica() {
         if [[ "$1" != "$REDIS_MASTER_HOST" ]]; then
-            sentinel_conf_add "sentinel known-replica {{ .Values.sentinel.masterSet }} $1 {{ .Values.sentinel.service.port }}"
```
dannylee-, line 257: Is this wrong?  Shouldn't this be .Values.sentinel.service.port?

dannylee-: Or maybe it should .Values.replica.containerPort

marcosbc: Previously, `sentinelPort` was set to `26379` by default, using the same value as `sentinel.containerPort`. In that sense, I don't think this has changed value. Are you finding any particular issues?

dannylee-: The port that is set in the "sentinel known-replica" used to be 6379.  After I got this change, the port changed to 26379, which caused a bunch of weird issues (i.e. somehow in a 3 pod setup, the sentinel container in one of the redis pods thought there were 4 sentinels somehow).  Once I changed it back to 6379, everything worked as expected.

PR #7250, opened 2021-08-17
[bitnami/airflow] Add Extra Environment Variables to Metrics Deployment

### LINKED ISSUE #6820: Airflow Missing Postgresql SSL Mode Option
**Which chart**:
Airflow (10.2.2)

**Describe the bug**
The option to set Postgres SSL Mode via a value is missing.  Alternatively, we tried using the Global Additional Env Vars, but the Metrics container does not receive them.

**To Reproduce**

1. Use external Postgres with SSL Mode Require (i.e. Azure Postgresql).
2. Add `AIRFLOW_PROMETHEUS_POSTGRES_SSL_MODE` to `extraEnvVars`.
3. Deploy Helm Chart.
4. Look at the `airflow-exporter` pod's logs
5. See constant errors about SSL mode.

**Expected behavior**
Best case, the helpers would generate an `airflow.sslMode` value that the Metric Exporter Deployment would use.  Alternatively, the Metric Exporter should receive the global env vars that the other pods are receiving.

- Output of `helm version`:

```
Using Azure DevOps hosted agent: Helm 3.6.1
```

```
Client: Using Azure DevOps hosted ageint: Kubectl 1.20.1-5-g76a04fc

Server Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.7", GitCommit:"4a4913e790a033f1101af2f0fb7be189b106e3ae", GitTreeState:"clean", BuildDate:"2021-05-25T17:37:03Z", GoVersion:"go1.15.12", Compiler:"gc", Platform:"linux/amd64"}
```

**Additional context**
Hosted in AKS using Azure Postgresql for external Airflow Database.

treyhendon:
**Description of the change**

Add Extra Environment Variables to the Airflow Metrics Deployment.

**Benefits**

Additional configuration of the deployment container.  This allows environment variables like `AIRFLOW_PROMETHEUS_POSTGRES_SSL_MODE` to be injected to the Airflow Metrics Deployment.

**Possible drawbacks**

None foreseen.

**Applicable issues**

  - fixes #6820 

**Additional information**

This does not include `.Values.extraEnvVars` like scheduler, web, and worker as those variables are not relevant to metrics.

This is a quick fix to get additional environment variables into the metrics deployment.  As my original conversation mentioned, ultimately, it would great to have a value (i.e. `externalDatabase.sslMode`) and helper (`airflow.database.sslMode`) that added the appropriate Postgres environment variables in the necessary places.

**Checklist** 

### DIFF
--- a/bitnami/airflow/Chart.yaml
+++ b/bitnami/airflow/Chart.yaml
@@ -32,4 +32,4 @@ name: airflow
 sources:
   - https://github.com/bitnami/bitnami-docker-airflow
   - https://airflow.apache.org/
-version: 10.2.8
+version: 10.2.9
--- a/bitnami/airflow/Chart.yaml
+++ b/bitnami/airflow/Chart.yaml
@@ -32,4 +32,4 @@ name: airflow
 sources:
   - https://github.com/bitnami/bitnami-docker-airflow
   - https://airflow.apache.org/
-version: 10.2.9
+version: 10.3.0
--- a/bitnami/airflow/values.yaml
+++ b/bitnami/airflow/values.yaml
@@ -1051,6 +1051,15 @@ metrics:
     scrapeTimeout: 10s
     # ServiceMonitor Labels
     labels: {}
+  ## @param metrics.extraEnvVars Array containing extra environment variables
+  ##
+  extraEnvVars: []
+  ## @param metrics.extraEnvVarsCM ConfigMap containing extra environment variables
+  ##
+  extraEnvVarsCM: ""
+  ## @param metrics.extraEnvVarsSecret Secret containing extra environment variables (in case of sensitive data)
+  ##
+  extraEnvVarsSecret: ""
   ## @param metrics.resources Metrics exporter resource requests and limits
   ## ref: http://kubernetes.io/docs/user-guide/compute-resources/
   ##
--- a/bitnami/airflow/README.md
+++ b/bitnami/airflow/README.md
@@ -336,27 +336,27 @@ The command removes all the Kubernetes components associated with the chart and 
 
 ### Airflow metrics parameters
 
-| Name                                   | Description                                                 | Value                         |
-| -------------------------------------- | ----------------------------------------------------------- | ----------------------------- |
-| `metrics.enabled`                      | Start a side-car prometheus exporter                        | `false`                       |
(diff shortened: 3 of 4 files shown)


bitnami/airflow/values.yaml
```diff
@@ -1051,6 +1051,15 @@ metrics:
     scrapeTimeout: 10s
     # ServiceMonitor Labels
     labels: {}
+  ## @param metrics.extraEnvVars Array containing extra environment variables
+  ##
+  extraEnvVars: []
+  ## @param metrics.extraEnvVarsCM ConfigMap containing extra environment variables
+  ##
+  extraEnvVarsCM: ""
+  ## @param metrics.extraEnvVarsSecret Secret containing extra environment variables (in case of sensitive data)
```
marcosbc, line 1060: The descriptions are tied with the ones in the `README.md`. We have a tool that is used to sync them from time to time, we just havent completed the integration with GitHub yet.

To avoid it causing any changes to the README.md afterwards, could you make the following changes so that it doesn't overwrite some of these changes afterwards?

* Update the descriptions so they match
* Update either `README.md` or `values.yaml` so that they are added in the same location. E.g. in both files after `metrics.resources`.

bitnami/airflow/Chart.yaml
```diff
@@ -32,4 +32,4 @@ name: airflow
 sources:
   - https://github.com/bitnami/bitnami-docker-airflow
   - https://airflow.apache.org/
-version: 10.2.8
+version: 10.2.9
```
marcosbc, line 35: In this case, since it is adding a new feature (new env. vars) we should increase the minor version, e.g. `10.2.8 -> 10.3.0`.

marcosbc (changes requested):
Thanks for the contribution! Could you have a look at my comments?

bitnami/airflow/Chart.yaml
treyhendon: Ah, good call.  I wasn't sure if this was an accidental oversite in the original and would be considered a bug fix.  Will do!

bitnami/airflow/values.yaml
treyhendon: No problem, makes sense.  I probably copied the readme from web and values from scheduler or something like that.

marcosbc (approved):
LGTM! Thanks!

PR #7253, opened 2021-08-17
[bitnami/grafana-operator] Fix missing RBAC rules in case of cluster-wide installation

yardenshoham:
For the operator to be able to scan all namespaces, it must have these rules.

**Description of the change**

Added more rules to RBAC ClusterRole.

**Benefits**

It will allow the operator to listen to all namespaces.

**Possible drawbacks**

None.

**Applicable issues**

None.

**Checklist** 

### DIFF
--- a/bitnami/grafana-operator/Chart.yaml
+++ b/bitnami/grafana-operator/Chart.yaml
@@ -25,4 +25,4 @@ name: grafana-operator
 sources:
   - https://github.com/integr8ly/grafana-operator
   - https://github.com/bitnami/bitnami-docker-grafana-operator
-version: 1.0.1
+version: 1.0.2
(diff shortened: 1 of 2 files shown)


marcosbc:
Thanks for the PR!

It doesn't seems like the [official Helm chart for grafana-operator](https://github.com/grafana-operator/grafana-operator) includes any of these rules: ([link](https://github.com/grafana-operator/grafana-operator/blob/master/deploy/roles/role.yaml))

We usually try to avoid deviating too much from official solutions, so they behave similar and also don't cause any unexpected issues. Do you think this PR would make sense to the official chart as well?

yardenshoham:
What do you mean? The rules you linked are the exact same rules I included in the PR

marcosbc (approved):
@yardenshoham I see now, basically everything from `Role*` resource definitions were copied to `ClusterRole*` definitions. I compared them and they match, like you mentioned, so LGTM!

Issue #7279, opened 2021-08-19
[bitnami/keycloak] Database password for Postgress defaults to "some-password"; Should be random


michaeljmarshall:
**Which chart**:
Bitnami/keycloak 4.2.0 (current, at time of issue creation)

**Is your feature request related to a problem? Please describe.**
The keycloak helm chart has a default password (`some-password`) for the postgres database.

It's possible that this value hasn't been changed to prevent breaking changes for users. I'm not familiar enough with how this password is used to speak authoritatively on whether or not this password can safely be changed during an upgrade. However, as a new consumer of this chart, I wouldn't expect to see a default password for my keycloak database.

**Describe the solution you'd like**
I'd like to see a random password (or at least be forced to provide my own password) when deploying keycloak. The documentation already suggests that the password will be random:

https://github.com/bitnami/charts/blob/cf14b39f0cfd52c68481f024703d7c80bd45a4ce/bitnami/keycloak/values.yaml#L869-L873

**Describe alternatives you've considered**
Alternatively, I could override the default. However, I am writing a chart that uses this one as a dependency. Thus, I would prefer for this chart to just use a random password.

**To Observe Behavior**
Steps to observe the behavior:

1. Template the keycloak helm chart without overriding any values.
2. Inspect the result of the `templates/secrets.yaml`.
3. Decode `database-password` to see that it is `some-password`.

**Additional context**
It seems to me that the default value for `.Values.postgresql.postgresqlPassword` should be `""`, as it is for other passwords in the keycloak chart. Then, because the `database-password` is generated with a function (shown below), I'd modify the function as follows:

```tpl
{{/*
Return the Database encrypted password
*/}}
{{- define "keycloak.databaseEncryptedPassword" -}}
{{- if .Values.postgresql.enabled }}
    {{- (.Values.postgresql.postgresqlPassword | default (randAlphaNum 10)) | b64enc | quote -}}
{{- else -}}
    {{- .Values.externalDatabase.password | b64enc | quote -}}
{{- end -}}
{{- end -}}
```
This behavior would match the behavior for other secrets in the keycloak helm chart: https://github.com/bitnami/charts/blob/cf14b39f0cfd52c68481f024703d7c80bd45a4ce/bitnami/keycloak/templates/secrets.yaml#L27-L33

If my solution is accepted, I'd be happy to contribute a patch with the above logic. Thanks!

michaeljmarshall:
[...]

PR #7287, opened 2021-08-20
[bitnami/keycloak] Default postgresql.postgresqlPassword to a random …

michaeljmarshall:
…10-character string

**Description of the change**
This change and its motivation is described in https://github.com/bitnami/charts/issues/7279. Essentially, it sets the `postgresql.postgresqlPassword` to `""` in the values file and then updates the `keycloak.databaseEncryptedPassword` function so that the default value of `""` is mapped to a random 10-character alphanumeric string.

**Benefits**
The postgres password will not default to the same password for every installation. This improves the chart's security.

**Possible drawbacks**
This is likely a breaking change. However, I'm not familiar enough with postgres to know for sure. My commit bumps the major version. I'd appreciate insight from the chart maintainers to confirm that this requires a major version bump. If it is only a minor bump, I can update the commit.

**Applicable issues**

  - fixes https://github.com/bitnami/charts/issues/7279

**Additional information**
I tested this commit by running the following from within the keycloak directory for this repository:

```
$ helm template test . | grep database
  database-password: "Z1Vpb1VWdFcxZw=="
                  key: database-password
$ echo Z1Vpb1VWdFcxZw== | base64 --decode
gUioUVtW1g
```

Before this change, the same test results in the following:

```
$ helm template test . | grep database
  database-password: "c29tZS1wYXNzd29yZA=="
                  key: database-password
$ echo c29tZS1wYXNzd29yZA== | base64 --decode
some-password
```

**Checklist** 

### DIFF
--- a/bitnami/keycloak/Chart.yaml
+++ b/bitnami/keycloak/Chart.yaml
@@ -26,4 +26,4 @@ name: keycloak
 sources:
   - https://github.com/bitnami/bitnami-docker-keycloak
   - https://github.com/keycloak/keycloak
-version: 5.0.1
+version: 6.0.0
--- a/bitnami/keycloak/Chart.yaml
+++ b/bitnami/keycloak/Chart.yaml
@@ -26,4 +26,4 @@ name: keycloak
 sources:
   - https://github.com/bitnami/bitnami-docker-keycloak
   - https://github.com/keycloak/keycloak
-version: 6.0.0
+version: 5.0.2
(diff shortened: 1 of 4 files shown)


juan131 (changes requested):
Thanks for the PR @michaeljmarshall !!

I have my doubts bumping the major version. Apparently it can look like a breaking change but let's analyze the 2 possible scenarios:

- Users installing the chart from scratch: unaffected. They will have a new random password to connect to the PostgreSQL database instead of `some-password` and that's all.
- Users upgrading existing releases. They will suffer issues IF they do not indicate the password to use for PostgreSQL in their `values.yaml` since the password parameter will acquire a different value. However, we have validations to block upgrades when this happens in the **NOTES.txt**, see:

- https://github.com/bitnami/charts/blob/master/bitnami/keycloak/templates/NOTES.txt#L76

Therefore, I think we could consider this change a simple "patch" change.

michaeljmarshall:
@juan131 - thanks for your review. I was unaware of the validation check. I see how this is not a breaking change. I've added a commit only bumping the patch version.

miguelaeh (approved):

LGTM!!

PR #7311, opened 2021-08-25
[bitnami/mongodb] Allow configuration of credentials for the metrics exporter

### LINKED ISSUE #7310: [bitnami/mongodb] Allow metrics user to be configurable
**Which chart**:
mongodb:10.23.13

As of now the user which the metrics exporter uses is hardcoded as `root` if the authentication is enabled. We don't have a `root` user enabled in our setup and therefore can't use the chart out of the box. It would be great if one could manually configure credentials for the metrics exporter to use.

Values to configure a username and password for the metrics exporter to use.

Workaround: Create a user called `root` in the MongoDB.

**Additional context**
I implemented a solution and will open a PR for it.

tielou:
**Description of the change**

As of now, the metrics exporter always authenticates as `root` when authentication is enabled. This change allows a configurable username/password to specify for the metrics exporter. If none are specified, it will fallback to default (`root`).

**Benefits**

One can use a different user than `root` for the metrics exporter.

**Possible drawbacks**

none afaik

**Applicable issues**

  - fixes #7310

**Additional information**

**Checklist** 


bitnami/mongodb/templates/_helpers.tpl
```diff
@@ -346,9 +346,13 @@ Validate values of MongoDB&reg; exporter URI string - auth.enabled and/or tls.en
 */}}
 {{- define "mongodb.mongodb_exporter.uri" -}}
     {{- $uriTlsArgs := ternary "tls=true&tlsCertificateKeyFile=/certs/mongodb.pem&tlsCAFile=/certs/mongodb-ca-cert" "" .Values.tls.enabled -}}
-    {{- $uriAuth := ternary "root:$(echo $MONGODB_ROOT_PASSWORD | sed -r \"s/@/%40/g;s/:/%3A/g\")@" "" .Values.auth.enabled -}}
-
-    {{- printf "mongodb://%slocalhost:27017/admin?%s" $uriAuth $uriTlsArgs -}}
+    {{- if and .Values.metrics.username .Values.metrics.password }}
```
yilmi, line 349: What if metrics.password is empty? We may just enter the root user branch.

bitnami/mongodb/templates/hidden/statefulset.yaml
```diff
@@ -398,11 +398,21 @@ spec:
           {{- end }}
           env:
           {{- if .Values.auth.enabled }}
+          {{- if not (or .Values.metrics.username .Values.metrics.password ) }}
```
yilmi, line 401: To avoid confusion we could just check for metrics.username instead. For example, if someone leaves metrics.password with a value but has metrics.username empty, that could cause some issue.

bitnami/mongodb/templates/replicaset/statefulset.yaml
```diff
@@ -432,11 +432,21 @@ spec:
           {{- end }}
           env:
           {{- if .Values.auth.enabled }}
+          {{- if not (or .Values.metrics.username .Values.metrics.password ) }}
```
yilmi, line 435: Same here, if someone leaves metrics.password with a value but has metrics.username empty, that could cause some issue. We could just check for metrics.username instead.

bitnami/mongodb/templates/standalone/dep-sts.yaml
```diff
@@ -366,11 +366,21 @@ spec:
               /bin/mongodb_exporter --web.listen-address ":{{ .Values.metrics.containerPort }}" --mongodb.uri "{{ include "mongodb.mongodb_exporter.uri" . }}" {{ .Values.metrics.extraFlags }}
           env:
           {{- if .Values.auth.enabled }}
+          {{- if not (or .Values.metrics.username .Values.metrics.password ) }}
```
yilmi, line 369: Same here, if someone leaves metrics.password with a value but has metrics.username empty, that could cause some issue. We could just check for metrics.username instead.

bitnami/mongodb/templates/hidden/statefulset.yaml
tielou: I removed `.Values.metrics.password` from the query

bitnami/mongodb/templates/standalone/dep-sts.yaml
tielou: I removed `.Values.metrics.password` from the query

bitnami/mongodb/templates/replicaset/statefulset.yaml
tielou: I removed `.Values.metrics.password` from the query

bitnami/mongodb/templates/_helpers.tpl
tielou: I don't see how it would work with an empty password? The whole logic would break, since the creation of the secret is just executed when the password is defined and then it would try mount a non-existing variable. 
So maybe the best solution would be to trigger a warning that both have to be defined?

javsalgar: I imagine that mounting the secret should be done only if the password is defined. I think it is potentially possible to access without password, if I'm not mistaken.

tielou: Hi @javsalgar 
I just gave it a try. When I try to create a MongoDB user without a password I get the following error:
```
> db.getSiblingDB("admin").createUser({
...     user: "testmetrics",
...     pwd: "",
...     roles: [
...         { role: "clusterMonitor", db: "admin" },
...         { role: "read", db: "local" }
...     ]
... })
uncaught exception: Error: couldn't add user: User passwords must not be empty :
_getErrorWithCode@src/mongo/shell/utils.js:25:13
DB.prototype.createUser@src/mongo/shell/db.js:1386:11
@(shell):1:1
```
I think passwordless admin access is possible when you turn off authentication but I don't see a way how it can be possible for other users.

javsalgar: Then it's fine by me to force the password for the metrics, which should be auto-generated just like the rest of passwords if not set.

tielou: I added the automatic password generation if `.Values.metrics.username` is set but `.Values.metrics.password` is not. Together with the changes in the image this user should get automatically created now if defined. https://github.com/bitnami/bitnami-docker-mongodb/pull/285

javsalgar:
From what I see, the container still does not have these variables available, right? Another option would be to create a configmap is mounted as an init script and adds the metrics user. That would make things work without adding the `_METRICS_` variables to the mongodb container.

tielou:
Yes you are right, we would need to merge this first: https://github.com/bitnami/bitnami-docker-mongodb/pull/285
Or like you mentioned with an init container but I think the cleaner way would be with the updated image since the other users get created in the image as well. 
But if you want me to change it to an initContainer let me know, then I will implement it and add it to the PR.
I also just fixed the linter.

yilmi:
@tielou, a container release with your changes should happen tonight, could you test your chart with it and let us know if everything works as expected?

tielou:
Hi @yilmi 
thank you very much for merging. I tested with the following image: `4.4.8-debian-10-r46`.
I tested the following cases:
- `.Values.metrics.username` and `.Values.metrics.password` set on a fresh install. The user gets created with the appropriate privileges in the MongoDB. The metrics exporter connects successfully.
- `.Values.metrics.username` set on a fresh install. The user gets created with a random password and with the appropriate privileges in the MongoDB. The metrics exporter connects successfully.
- `.Values.metrics.username` set on a fresh install with `.Values.auth.enabled` set to `false`. No user gets created in the MongoDB. The metrics exporter connects successfully.
- `.Values.metrics.username` and `.Values.metrics.password` set on an existing install with the corresponding user present in the database. The metrics exporter connects successfully.
So unless I forgot a case, I think we should be good to go. From what I see the image version gets updated through the automation?

yilmi:
Yes the bot updates the image version automatically when a new release is available. Tested on my end and looks good as well, thank you!

PR #7333, opened 2021-08-26
[bitnami/redis]: hotfix, regenerate sentinel config at each boot-up

rsecob:
**Description of the change**

After upgrading to 15.0.0 we realized there was an issue with sentinel. If the service is restarted within its running container, the sentinel config remained unchanged. The script is supposed to prune the known sentinels and replicas and regenerate them when fetching the available hosts on the network. This part of the script was not working as expected which caused this error:

```
>>> 'sentinel known-sentinel rt-collector-master 10.8.5.30 26379 ecac44470a643ce5d853d166ad8abc2135e4bdbf'
Duplicate runid for sentinel.
```

Our proposed solution is to override the existing configuration _(if it exists)_ with the original configuration at each bootup. This ensures all replicas start with the expected configuration file.

**Benefits**

- Fixes CrashBackLoop state if the service is restarted in a running container.

**Checklist** 

### DIFF
--- a/bitnami/redis/Chart.yaml
+++ b/bitnami/redis/Chart.yaml
@@ -25,4 +25,4 @@ name: redis
 sources:
   - https://github.com/bitnami/bitnami-docker-redis
   - http://redis.io/
-version: 15.0.3
+version: 15.0.4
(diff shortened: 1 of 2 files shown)


yilmi:
Thanks for the details, let me check that and get back to you quickly!

yilmi:
lgtm, one note, I had to recreate the pods. Thanks again @rsecob for this contribution!

PR #7354, opened 2021-08-31
[bitnami/mongodb] Simplify tls certs logic

carrodher:
**Description of the change**

**Benefits**

There is some logic that was duplicated in the `if` and `else` statements, while only the part related to the `$ca` was different in the conditional. With those changes, the duplicated parts are placed after the conditional so they are executed in the same way avoiding the duplication

**Checklist** 


carrodher:
Nevermind, it is required because of the scope of the variables

PR #7365, opened 2021-09-01
[bitnami/redis] Fix HPA condition for sentinel and replicas

### LINKED ISSUE #7360: [bitnami/redis] sentinel hpa is now always created, besides that it is not used (15.2.X)
the last commit for redis always enabled the hpa: https://github.com/bitnami/charts/commit/a34880b435c1406b9654da25cb7e4adcd360331c#diff-315db354f46568e44ac1247c85cf17e873c3e75966985515cb4790e625ee5d49 however this is bad since hpa now is configured but will always fail since no `targetCPU` or `targetMemory` is configured.
autoscaling should be disabled if either both of these are `""` or if it is explictly disabled.
cluster tooling does not like hpa failures.

pablogalegoc:
**Description of the change**

Fixes the HPA resource creation condition for sentinel and replicas.

**Applicable issues**

  - fixes #7360 

**Checklist** 

### DIFF
--- a/bitnami/redis/Chart.yaml
+++ b/bitnami/redis/Chart.yaml
@@ -25,4 +25,4 @@ name: redis
 sources:
   - https://github.com/bitnami/bitnami-docker-redis
   - http://redis.io/
-version: 15.2.0
+version: 15.2.1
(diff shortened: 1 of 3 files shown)


Mauraza (approved):
Hi @pablogalegoc,

Thanks for the PR!! LGTM!

PR #7370, opened 2021-09-01
[bitnami/redis] Update apiVersion of PDB to policy/v1 on 1.21+ cluster (via common.capabilities.policy.apiVersion)

tomislater:
According to https://kubernetes.io/docs/reference/using-api/deprecation-guide/ we can start using `policy/v1` in Kubernetes `1.21`.

Without this change you can observe such things in logs:

```
W0901 11:28:35.264440      91 warnings.go:70] policy/v1beta1 PodDisruptionBudget is deprecated in v1.21+, unavailable in v1.25+; use policy/v1 PodDisruptionBudget
```

**Checklist** 


tomislater:
Hmm, maybe I should apply this manifest only on 1.21+ clusters? I think I can do that in Helm.

Mauraza:
Hi @tomislater,

> Hmm, maybe I should apply this manifest only on 1.21+ clusters? I think I can do that in Helm.

I think that it's a good idea you can do something similar to: 
https://github.com/bitnami/charts/blob/6ec60956956f243967a5cdfee515c484627a4de3/bitnami/redis/templates/_helpers.tpl#L45-L54

Please remember to upgrade the **version** of the chart.

Mauraza (approved):
Hi @tomislater,

Thanks for the contribution!! LGTM!

PR #7381, opened 2021-09-02
[bitnami/contour] Fix ingressClass

### LINKED ISSUE #7372: [bitnami/contour] Contour ingressClass values.yaml confusion
**Chart**
[[bitnami/contour](https://github.com/bitnami/charts/tree/master/bitnami/contour)]

**Describe the bug**
The values spec has the [`ingressClass`](https://github.com/bitnami/charts/blob/f01935dbde4e3bd9e38106e264e201372ccac710/bitnami/contour/values.yaml#L264
) under `contour` key but the [template implementation](https://github.com/bitnami/charts/blob/f01935dbde4e3bd9e38106e264e201372ccac710/bitnami/contour/templates/contour/deployment.yaml#L71) expects the `ingressClass` at top level -> `{{- if .Values.ingressClass }}`

**To Reproduce**
Install the chart with values as specified in the `values.yaml`:
```yaml
contour:
  ingressClass: contourcustom
```
The parameter is ignored and the deployment misses the `--ingress-class-name` arg.

The deployment should use the specs stated in `values.yaml` or the `values.yaml` should be updated to avoid confusion.

Thanks!
Let me know if there are additional details needed.

alvneiayu:
**Description of the change**

Issue reported by an user. The ingressClass key now is generated checking the ingressClass in the top of the keys defined in values.yaml instead of contour.ingressClass (properly key definition path).

**Benefits**

Fix ingressClass definition in our templates.

**Applicable issues**

  - fixes #7372

**Checklist** 

### DIFF
--- a/bitnami/contour/Chart.yaml
+++ b/bitnami/contour/Chart.yaml
@@ -27,4 +27,4 @@ sources:
   - https://github.com/envoyproxy/envoy
   - https://github.com/bitnami/bitnami-docker-contour
   - https://projectcontour.io
-version: 5.5.0
+version: 5.5.1
--- a/bitnami/contour/templates/contour/deployment.yaml
+++ b/bitnami/contour/templates/contour/deployment.yaml
@@ -68,8 +68,8 @@ spec:
             - --contour-cert-file=/certs/tls.crt
             - --contour-key-file=/certs/tls.key
             - --config-path=/config/contour.yaml
-            {{- if .Values.ingressClass }}
-            - --ingress-class-name={{ .Values.ingressClass }}
+            {{- if .Values.contour.ingressClass }}
+            - --ingress-class-name={{ .Values.contour.ingressClass }}
             {{- end }}
             {{- if .Values.contour.extraArgs }}
             {{- include "common.tplvalues.render" (dict "value" .Values.contour.extraArgs "context" $) | nindent 12 }}


alemorcuq (approved):
LGTM

MrSaints:
I have reason to believe this is a backwards incompatible change, and should deserve a major version bump.

As noted in the docs:

> Contour uses its configured IngressClass name to filter Ingresses. If the --ingress-class-name flag is provided, Contour will only accept Ingress resources that exactly match the specified IngressClass name via annotation or spec field, with the value in the annotation taking precedence. If the flag is not passed to contour serve Contour will accept any Ingress resource that specifies the IngressClass name contour in annotation or spec fields or does not specify one at all.

https://projectcontour.io/docs/main/config/ingress/#ingressclass-and-ingressclass-name

Since this has been "fixed", for pre-existing users of this chart who did not include the annotation, they will effectively have an "outage" due to the ingress not being picked up. I may be misunderstanding and looking at a red herring though, but this was the case for us. Setting:

```
contour:
    ingressClass: false
```

Remediated the issue.

alvneiayu:
hi @MrSaints

I think that you are right. I looked the issue as a bug, my bad. Let me investigate a little bit and I will try to find the best solution.

Thanks a lot for your time and help.

Álvaro

MrSaints:
No worries @alvneiayu.

FWIW, it is indeed a bug. But I think this bug can pretty much be considered a feature for pre-existing users who did not include the annotation.

I think an appropriate fix may be to set the default `ingressClass` to `false` because it is a setting which will work for current, and pre-existing users, even those using the annotation. If I'm not mistaken, this is optional, and unset by default in the Operator and example manifests, so it should be line. Those needing to run multiple instances can then opt in by overriding the default value.

alvneiayu:
hi @MrSaints 

Let me investigate this. Maybe it is a really good option.

Thanks a lot

Álvaro

alvneiayu:
hi @MrSaints 

We are working in several changes in Contour. We are going to release a new major with those changes and also this one. Just to let you know that.

If I can help you with something else and again thanks a lot for your help.

Álvaro

PR #7384, opened 2021-09-02
[bitnami/solr] : Workaround/Fix for GCE Ingress

deas:
**Description of the change**
Renaming `targetPort`  value.

**Benefits**
Provide a "painless" solution for Container Native Loadbalancing on GKE before [kubernetes/ingress-gce/pull/1462](https://github.com/kubernetes/ingress-gce/pull/1462) lands.

**Possible drawbacks**
External references to old values. If this is a concern, one could still introduce a helm value defaulting to old values while allowing those affected to override using `http`.

**Applicable issues**

n/a

**Additional information**

**Checklist** 

- [x] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/) - hopefully. ;)

### DIFF
--- a/bitnami/solr/Chart.yaml
+++ b/bitnami/solr/Chart.yaml
@@ -27,4 +27,4 @@ name: solr
 sources:
   - https://github.com/bitnami/bitnami-docker-solr
   - https://lucene.apache.org/solr/
-version: 2.0.4
+version: 2.0.5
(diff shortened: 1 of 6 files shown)


miguelaeh (commented):
Hi @deas ,
Thank you very much for the PR, I don't fully understand why the port name must be changed checking the referenced PR, could you provide more information, please?
I would like to fully understand why we should change the port name since if this is an issue it would be needed to change more charts.

deas:
Hi @miguelaeh ,

Have a look at [the bottom of this comment](https://github.com/kubernetes/ingress-gce/pull/1462#issuecomment-906654998). Google Cloud Ingress won't work unless the `targetPort` name has value `http`. If we have other charts referencing container ports by their current names, I would suggest we introduce a variable holding the current value so we don't break things. This still allows those affected by the Google Cloud issue to override. I'll update the patch if this is ok for you.

miguelaeh (approved):
LGTM!

PR #7386, opened 2021-09-02
[bitnami/redmine] Missing env var in cronjob.yaml

Vinche59:
**Description of the change**

mail-receiver-configmap.yaml reference $REDMINE_DATABASE_TYPE variable but this variable is never defined in cronjob.yaml

**Benefits**

**Possible drawbacks**

**Applicable issues**

  - fixes #

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/redmine/templates/cronjob.yaml
+++ b/bitnami/redmine/templates/cronjob.yaml
@@ -84,6 +84,8 @@ spec:
                 {{- end }}
               resources: {{- toYaml .Values.mailReceiver.resources | nindent 16 }}
               env:
+                - name: REDMINE_DATABASE_TYPE
+                  value: {{ .Values.databaseType | quote }}
                 {{- if eq .Values.databaseType "mariadb" }}
                 - name: REDMINE_MAIL_RECEIVER_DB_MYSQL
                 {{- else }}
--- a/bitnami/redmine/Chart.yaml
+++ b/bitnami/redmine/Chart.yaml
@@ -36,4 +36,4 @@ name: redmine
 sources:
   - https://github.com/bitnami/bitnami-docker-redmine
   - http://www.redmine.org/
-version: 17.0.6
+version: 17.0.7


Mauraza:
Hi @Vinche59,

Did you have any issues?

I deployed a Redmine with `mail-receiver-configmap`  and everything goes well. The environment variables of this configmap are:
```bash
    Environment:
      REDMINE_MAIL_RECEIVER_DB_MYSQL:                 redmain-mariadb
      REDMINE_MAIL_RECEIVER_DB_DATABASE:              my_database
      REDMINE_MAIL_RECEIVER_DB_USERNAME:              bn_redmine
      REDMINE_MAIL_RECEIVER_DB_PASSWORD:              <set to the key 'mariadb-password' in secret 'redmain-mariadb'>  
```

Vinche59:
Hello @Mauraza 

Actually deployment works correctly in my env but when the job execute it failed each time with an error : 
```
"No database settings given. Exiting"
```
In the [receive.sh](https://github.com/bitnami/charts/blob/master/bitnami/redmine/templates/mail-receiver-configmap.yaml) configmap generated, this error is generated because the script can't find ```$REDMINE_DATABASE_TYPE``` env var and I didn't find where this variable should be setted thus this pull request.

But maybe I missed something during the deployment of the helm chart ?

bitnami/redmine/templates/cronjob.yaml
```diff
@@ -84,6 +84,12 @@ spec:
                 {{- end }}
               resources: {{- toYaml .Values.mailReceiver.resources | nindent 16 }}
               env:
+                - name: REDMINE_DATABASE_TYPE
+                {{- if eq .Values.databaseType "mariadb" }}
+                  value: "mariadb"
+                {{- else }}
+                  value: "postgresql"
+                {{- end }}
```
Mauraza, line 92: You can use the same statement as in deployment.yaml
```suggestion
             - name: REDMINE_DATABASE_TYPE
                value: {{ .Values.databaseType | quote }}
```

Mauraza (commented):
Hi @Vinche59,

Could you check the comment?

bitnami/redmine/templates/cronjob.yaml
Vinche59: Hello @Mauraza,

Indeed, it's nicer that way :-)
Modifications done.

Mauraza (approved):
Hi @Vinche59,

PR #7402, opened 2021-09-03
bitnami/thanos: Switch deployment strategy to 'Recreate'

knatsakis:
**Description of the change**

This PR changes the default deployment strategy of Thanos compactor Deployment from `RollingUpdate` to `Recreate`.

**Benefits**

The Thanos compactor Deployment contains two volume mounts by default (https://github.com/bitnami/charts/blob/master/bitnami/thanos/templates/compactor/deployment.yaml#L134). This makes the `RollingUpdate` strategy unsuitable as a new pod cannot be created as long as the old pod is running. The `RollingUpdate` strategy results in

```
Multi-Attach error for volume "pvc-0a8b96f5-249f-4708-8c7d-51c2538d647d" Volume is already used by pod(s) thanos-compactor-7d4dbdf4f9-kgt9j
```

errors. A `Recreate` deployment strategy, waits for the old pod to terminate before starting a new one, avoiding the attempt to mount the PVCs twice.

**Possible drawbacks**

None that I am aware of. Replicas are hardcoded to 1 (https://github.com/bitnami/charts/blob/master/bitnami/thanos/templates/compactor/deployment.yaml#L12), so two thanos-compactor pods are not supposed to run at the same time anyway.

**Applicable issues**

None that I am aware of.

**Additional information**

None.

**Checklist**

### DIFF
--- a/bitnami/thanos/Chart.yaml
+++ b/bitnami/thanos/Chart.yaml
@@ -28,4 +28,4 @@ name: thanos
 sources:
   - https://github.com/bitnami/bitnami-docker-thanos
   - https://thanos.io
-version: 6.0.10
+version: 6.0.11
--- a/bitnami/thanos/values.yaml
+++ b/bitnami/thanos/values.yaml
@@ -1357,7 +1357,7 @@ compactor:
   extraFlags: []
   ## @param compactor.strategyType Deployment Strategy Type, can be set to RollingUpdate or Recreate by default
   ##
-  strategyType: RollingUpdate
+  strategyType: Recreate
   ## @param compactor.podAffinityPreset Thanos Compactor pod affinity preset
   ## ref: https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#inter-pod-affinity-and-anti-affinity
   ##


pablogalegoc:
Hi @knatsakis!

First of all, sorry for the late response. We had an outage on our support-tracking service and some issues and PR's were lost.

The error you point is related to the [access mode](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#access-modes) of the PV provisioned ([ReadWriteOnce](https://github.com/bitnami/charts/blob/master/bitnami/thanos/values.yaml#L1592-L1593) in this case). With ReadWriteOnce if the pod that is going to substitute the old one is scheduled in another node it will throw that error. IMO I'd leave the `RollingUpdate` as default because it is the strategy that avoids downtime and rely on a `ReadWriteMany` access mode or a [nodeSelector](https://github.com/bitnami/charts/blob/master/bitnami/thanos/values.yaml#L1369) to schedule both pod's in the same node. WDYT?

knatsakis:
Hi @pablogalegoc 

> IMO I'd leave the `RollingUpdate` as default because it is the strategy that avoids downtime and rely on a `ReadWriteMany` access mode or a [nodeSelector](https://github.com/bitnami/charts/blob/master/bitnami/thanos/values.yaml#L1369) to schedule both pod's in the same node. WDYT?

I don't think that it's a good idea to allow two compactor pods running at the same time. [Thanos' Documentation](https://thanos.io/v0.21/components/compact.md/#warning-only-one-instance-has-to-run-against-single-stream-of-blocks-in-single-object-storage) explicitly mentions:

> Because there is no safe locking mechanism for all object storage provides, currently, you need to ensure on your own that only single Compactor is running against single stream of blocks on single bucket. Running more can result with Overlap Issues that has to be resolved manually.

pablogalegoc:
Thanks for the reference @knatsakis, I wasn't aware of that. LGTM then!

PR #7417, opened 2021-09-06
[bitnami/redis] Do not set StatefulSet.replicas if autoscaling is enabled

pierophp:
**Description of the change**

If autoscaling is enabled do not set StatefulSet.replicas

**Benefits**

Currently Helm shows as OutOfSync when autoscaling is turned on, when there is actually no change to apply

image

**Possible drawbacks**
I think there is no drawback

**Applicable issues**

  - fixes #

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/redis/Chart.yaml
+++ b/bitnami/redis/Chart.yaml
@@ -25,4 +25,4 @@ name: redis
 sources:
   - https://github.com/bitnami/bitnami-docker-redis
   - http://redis.io/
-version: 15.3.1
+version: 15.3.2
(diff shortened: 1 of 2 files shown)


Mauraza:
Hi @pierophp,

Thanks for the PR! Could you add more information about the issue?

Mauraza (approved):
Hi  @pierophp,

The value for default for replicas is 3:
https://github.com/bitnami/charts/blob/d2b0bde4dfe37c6acb40b884f3e1f0f2d307fb26/bitnami/redis/values.yaml#L441-L444
But the definition of replicas will cancel the effect of the HPA for a while. 
LGTM!

PR #7422, opened 2021-09-07
[bitnami/redis] Prefix Redis Metrics port name correctly

upodroid:
**Description of the change**

Change metrics portName from tcp-metrics to http-metrics

**Benefits**

I use Istio in my cluster and portName prefix is very important. https://istio.io/latest/docs/ops/configuration/traffic-management/protocol-selection/

**Possible drawbacks**

None

**Checklist** 


randradas:
Hello @upodroid thank you for your PR. why should it be called http-metrics instead of tcp when it is about protocol tcp?

upodroid:
Metrics are scraped over http/https. It is better to have a variable there that can set port name.

```
[2021-09-08T09:15:12.092Z] "GET /metrics HTTP/1.1" 200 - via_upstream - "-" 0 4716 7 6 "-" "Prometheus/2.28.1" "714af6c2-2cbd-9a02-8376-b1da507a33c9" "10.42.9.57:9121" "10.42.9.57:9121" inbound|9121|| 127.0.0.6:43473 10.42.9.57:9121 10.42.9.16:46660 - default
```

randradas:
Hi @upodroid, let me clarify something first. Our charts do not support Istio officially, it means that our internal system do not test whether the chart works with Istio or not so, in the future, another change could brake it. Being said that, the change looks good to me. I have also tested the change in our internal CI/CD tools and everything works properly.

The only thing that I would need from you is to rebase from master in order to bump the version properly and then pass every GH action. Once it is done I will merge the change. Your contribution is really appreciated.

Thank you!

upodroid:
I rebased and bumped the version now.

dannylee-:
Might want to add this to the upgrade guide - https://github.com/bitnami/charts/tree/master/bitnami/redis#upgrading

We were using the tcp-metrics port name.

marcosbc:
@dannylee- Would you be up for contributing the fix in the documentation as a PR?

PR #7426, opened 2021-09-07
[bitnami/apache] Add static analysis test for Apache

pablogalegoc:
**Description of the change**

We are adding the [Trivy scan for misconfigurations](https://github.com/aquasecurity/trivy#demo-misconfiguration-detection-iac-files). This is a PoC in one chart for the Github Action to run the checks of the Trivy default ruleset.

**Benefits**

Validation of configuration changes and public visibility of the checks we are running and skipping for the default installation of the Apache chart.

**Additional information**

**Checklist** 

### DIFF
--- a/.github/workflows/static-analysis-test.yaml
+++ b/.github/workflows/static-analysis-test.yaml
@@ -28,6 +28,6 @@ jobs:
           helm template apache bitnami/apache/ -f bitnami/apache/values.yaml > test-resources/generated-manifests.yaml
 
       - name: Run Trivy vulnerability scanner in IaC mode
-        uses: docker://bitnami/trivy:latest
+        uses: docker://bitnami/trivy:0
         with:
           args: config --exit-code 1 --ignorefile test-resources/.trivyignore test-resources/
--- a/.github/workflows/static-analysis-test.yaml
+++ b/.github/workflows/static-analysis-test.yaml
@@ -11,7 +11,7 @@ on:
 jobs:
   static-analysis-test:
     name: Static analysis test
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     steps:
       - name: Checkout Bitnami charts code
         uses: actions/checkout@v2
--- a/.github/workflows/static-analysis-test.yaml
+++ b/.github/workflows/static-analysis-test.yaml
@@ -1,0 +1,33 @@
+name: Static analysis test
+on:
+  pull_request:
+    branches:
+      - master
+    # These hardcoded paths are set for the PoC on the Apache chart
+    paths:
+      - "bitnami/apache/templates/*"
+      - "bitnami/apache/values.yaml"
+
+jobs:
+  static-analysis-test:
+    name: Static analysis test
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout Bitnami charts code
+        uses: actions/checkout@v2
+
+      - name: Prepare test resources
+        run: |
+          mkdir test-resources
+          cp bitnami/apache/ci/static-analysis/.trivyignore test-resources/
+
+      - name: Helm template
+        run: |
+          helm repo add bitnami https://charts.bitnami.com/bitnami
+          helm dependency build bitnami/apache/
+          helm template apache bitnami/apache/ -f bitnami/apache/values.yaml > test-resources/generated-manifests.yaml
+
+      - name: Run Trivy vulnerability scanner in IaC mode
+        uses: docker://bitnami/trivy:latest
+        with:
+          args: config --exit-code 1 --ignorefile test-resources/.trivyignore test-resources/
--- a/bitnami/apache/ci/static-analysis/.trivyignore
+++ b/bitnami/apache/ci/static-analysis/.trivyignore
@@ -1,0 +1,18 @@
+# Default capabilities not dropped: avd.aquasec.com/appshield/ksv003
+KSV003
+# CPU not limited: avd.aquasec.com/appshield/ksv011
+KSV011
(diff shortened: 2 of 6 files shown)


bitnami/apache/ci/static-analysis/.trivyignore
```diff
@@ -0,0 +1,18 @@
+# Default capabilities not dropped: avd.aquasec.com/appshield/ksv003
```
pablogalegoc, line 1: I've created a new directory under `apache/ci/static-analysis` to hold not just the `.trivyignore` but future custom policy files.

.github/workflows/static-analysis-test.yaml
```diff
@@ -0,0 +1,33 @@
...
+          cp bitnami/apache/ci/static-analysis/.trivyignore test-resources/
+
+      - name: Helm template
+        run: |
+          helm repo add bitnami https://charts.bitnami.com/bitnami
+          helm dependency build bitnami/apache/
+          helm template apache bitnami/apache/ -f bitnami/apache/values.yaml > test-resources/generated-manifests.yaml
+
+      - name: Run Trivy vulnerability scanner in IaC mode
+        uses: docker://bitnami/trivy:latest
```
pablogalegoc, line 31: IMO this is a good case for using `latest` instead of a fixed one to always use the latest checks and features

pablogalegoc:
Failing [static-analysis-test](https://github.com/bitnami/charts/runs/3533575342) check before 3ab977b

javsalgar:
I think it is a good idea. We should create a proposal for the rules to check, and perform an analysis in our catalog to ensure that all of them pass the tests before enabling the action.

.github/workflows/static-analysis-test.yaml
```diff
@@ -0,0 +1,33 @@
+name: Static analysis test
+on:
+  pull_request:
+    branches:
+      - master
+    # These hardcoded paths are set for the PoC on the Apache chart
+    paths:
+      - "bitnami/apache/templates/*"
+      - "bitnami/apache/values.yaml"
```
juan131, line 9: How do you plan to solve the problem of running the "Static analysis test" only for the charts that a specific PR modifies?

.github/workflows/static-analysis-test.yaml
juan131: You may want to stick to the latest version of a specific branch (e.g. `bitnami/trivy:0`). Otherwise, these tests will potentially break if Aqua release a new major version.

.github/workflows/static-analysis-test.yaml
```diff
@@ -0,0 +1,33 @@
...
+
+jobs:
+  static-analysis-test:
+    name: Static analysis test
+    runs-on: ubuntu-20.04
```
juan131, line 14: Why not using latest as we do in the rest of workflows?
```suggestion
    runs-on: ubuntu-latest
```

.github/workflows/static-analysis-test.yaml
```diff
@@ -0,0 +1,33 @@
...
+      - name: Prepare test resources
+        run: |
+          mkdir test-resources
+          cp bitnami/apache/ci/static-analysis/.trivyignore test-resources/
+
+      - name: Helm template
+        run: |
+          helm repo add bitnami https://charts.bitnami.com/bitnami
+          helm dependency build bitnami/apache/
+          helm template apache bitnami/apache/ -f bitnami/apache/values.yaml > test-resources/generated-manifests.yaml
```
juan131, line 28: Do we plan to support other values? E.g. the ones under `bitnami/XXX/ci` directory.

juan131 (commented):
This is an amazing proposal @pablogalegoc ! Please check my comments when you have a chance

.github/workflows/static-analysis-test.yaml
pablogalegoc: We'll have to do something similar to the action for readme-generator: https://github.com/bitnami/charts/blob/master/.github/workflows/generate-chart-readme.yml#L41-L51
TBH it's a hack for a functionality that should be provided by Github and be accesible via something like
`${{ github.event.paths }}`
We can go with it or explore other options like https://github.com/dorny/paths-filter

.github/workflows/static-analysis-test.yaml
pablogalegoc: Agree!

.github/workflows/static-analysis-test.yaml
pablogalegoc: We'll have to use the same hacks as for readme-generator. I looked for better workarounds at the time but didn't find anything. This can be a good time for a double check.

pablogalegoc: This is something we should do 100%, otherwise we are only verifying the default installation. I've demoed a quick script to the team that runs the scan for the "bare chart" to have a baseline and with some variations that add, for example, the `volumePermissions` container, the PostgreSQL subchart, etc. This gives us a better picture but we'll have to be careful to not bloat with warnings.

.github/workflows/static-analysis-test.yaml
pablogalegoc: The only concern would be if they stop adding Helm to the latest images, but it is highly unlikely so I agree we can use `latest` instead

.github/workflows/static-analysis-test.yaml
juan131: Well.. At lease we have some workarounds for this problem

pablogalegoc:
Closing since this PR was just a PoC to showcase the SAST tool in CI

PR #7435, opened 2021-09-08
[bitnami/harbor] Fix harbor-ingress with GCE controller

### LINKED ISSUE #7203: Harbor with GCE controller pointnig to ClusterIP port service
There is problem with harbor chart installation on GKE. I set up controller type for "gce" and all other options for ingresses. Everything is installing properly but ingress for "core" want to connect to portal service which is in type of "ClusterIP" instead of "NodePort". Here is ingress error:

```
Warning  Translate  4s                loadbalancer-controller  Translation failed: invalid ingress spec: service "harbor/harbor-portal" is type "ClusterIP", expected "NodePort" or "LoadBalancer"
```
And here is portal svc template:

```
apiVersion: v1
kind: Service
metadata:
  name: {{ include "harbor.portal" . }}
    {{- end }}
  {{- end }}
spec:
  ports:
    - name: http
      port: {{ include "harbor.portal.servicePort" . }}
      targetPort: http
    app.kubernetes.io/component: portal
```

I think there should be added such "if" like in other services or ingress should be pointing to core service which is using NodePort with "gce" controller:

```
  {{- if  (eq .Values.ingress.controller "gce") }}
  type: NodePort
  {{- end }}
  ```
  
EDIT:

I checked ingress template and it looks like ingress is pointing to both services so for "gce" portal service should have this "if" added too like in core and notary service:

image

### PR DESCRIPTION
**Pull request by pablogalegoc**
**Description of the change**

When ingress is enabled with `gce` as controller it causes `harbor-ingress` to get stuck in:
```
  Type     Reason     Age                    From                     Message
  ----     ------     ----                   ----                     -------
  Normal   Sync       2m39s (x2 over 2m40s)  loadbalancer-controller  Scheduled for sync
  Warning  Translate  10s (x14 over 2m39s)   loadbalancer-controller  Translation failed: invalid ingress spec: service "default/harbor-portal" is type "ClusterIP", expected "NodePort" or "LoadBalancer"
```

Harbor portal service needs the same condition found in other components:
https://github.com/bitnami/charts/blob/4986338b63815cdd926a27589204c50aa5e039e1/bitnami/harbor/templates/core/core-svc.yaml#L14-L16

**Applicable issues**

  - fixes #7203

**Additional information**

**Checklist** 


randradas:
LGTM! do we need to test it in dp-testing?

randradas:
Some conflicts here to be resolved, probably you need to rebase master and update the version.

pablogalegoc:
@randradas conflicts solved and re-run the changes in the pipeline since it was a major release what caused the conflicts. Everything seems fine!

javsalgar (approved):

PR #7477, opened 2021-09-14
[bitnami/rabbitmq] Fix typo on values.yaml

Kludex:


carrodher (changes requested):
Thanks for the fix, it is needed to bump the chart version (in the _Chart.yaml_ file). A patch bump is enough in this case.

carrodher (approved):
LGTM

PR #7484, opened 2021-09-14
[bitnami/airflow] fixes the ERR_TOO_MANY_REDIRECTS error that happen when try to login using Airflow's LDAP integration

### LINKED ISSUE #7010: [bitnami/airflow] : Infinite redirect loop on opening Airflow Home Page
When we open the Login page in the browser we are getting too many redirects error. I have used LDAP for authentication

image

image

Note : I am using ingress and I have already added proxy fix in extraEnvVars in values.yaml 

image

randradas:
**Description of the change**

This PR fixes the ERR_TOO_MANY_REDIRECTS error that happen when try to login using Airflow's LDAP integration as method for authentication. This change introduce new needed env variables, in order to work these variables have to be configured properly and aligned with the used OpenLDAP scheme. By default they have been prepared for a basic 'dc=example,dc=org' that uses an 'Admins' and 'All' groups. Before this change a proper configuration couldn't be done.

**Benefits**

Airflow can be integrated with OpenLDAP properly

**Possible drawbacks**

There is not known drawbacks

**Applicable issues**

  - fixes #7010

**Additional information**

**Checklist** 


randradas:
Working on README and bumping up chart version

randradas:
Here the dp-testing job: https://jenkins-dp-testing.nami.run/job/Testing/job/randradas-airflow/job/docker-image/job/airflow/job/2/job/debian-10/3/console

bitnami/airflow/README.md
```diff
@@ -19,7 +19,7 @@ Bitnami charts can be used with [Kubeapps](https://kubeapps.com/) for deployment
 - Kubernetes 1.12+
```
carrodher, line 19: Please, take a look at the _README.md_ there are several code blocks out of place and some weird characters.
If you are not adding any content, just updating the table with the new parameters you should use [the `readme-generator` tool](https://github.com/bitnami-labs/readme-generator-for-helm) in order to generate fill the tables that appear in the _README.md_ using the metadata from the _values.yaml_

bitnami/airflow/templates/web/deployment.yaml
```diff
@@ -105,20 +105,28 @@ spec:
...
+            - name: AIRFLOW_LDAP_ROLES_MAPPING
+              value: {{ .Values.ldap.rolesMapping | quote }}
+            - name: AIRFLOW_LDAP_ROLES_SYNC_AT_LOGIN
+              value: {{ .Values.ldap.rolesSyncAtLogin | quote }}
             - name: AIRFLOW_LDAP_USE_TLS
               value: {{ ternary "True" "False" .Values.ldap.tls.enabled | quote }}
             {{- if .Values.ldap.tls.enabled }}
             - name: AIRFLOW_LDAP_ALLOW_SELF_SIGNED
-              value: {{ .Values.ldap.tls.allowSelfSigned }}
+              value: {{ .Values.ldap.tls.allowSelfSigned | quote }}
```
carrodher, line 129: Remove trailing spaces
```suggestion
              value: {{ .Values.ldap.tls.allowSelfSigned | quote }}
```

bitnami/airflow/templates/web/deployment.yaml
```diff
@@ -105,20 +105,28 @@ spec:
               value: {{ .Values.ldap.uri }}
```
carrodher, line 105: I think it is not needed to quote all the env. variables, in fact, the already existing ones are defined as a string in the _values.yaml_ and are not quoted here. We should follow the same approach for all of them

bitnami/airflow/values.yaml
```diff
@@ -840,17 +840,27 @@ git:
...
-  uri: 'ldap://ldap_server:389'
-  base: 'ou=example,o=org'
-  binddn: 'cn=user,ou=example,o=org'
-  bindpw: ""
-  uidField: 'uid'
+  uri: 'ldap://ldap_server:1389'
+  base: 'dc=example,dc=org'
+  uidField: 'cn'
+  binddn: 'cn=admin,dc=example,dc=org'
+  bindpw: ''
```
carrodher, line 858: Is there any reason why those default parameters were modified? For example, why is the port different now?

If there is a user that installed the existing version of the chart with `ldap.enabled=true`, is a simple `helm upgrade` going to work with those changes or his LDAP configuration is going to be broken because it is different right now?

carrodher (changes requested):
I just write some comments, apart from that you need to ensure the _values.yaml_ is updated pointing to the image tag that contains the changes (see https://github.com/bitnami/bitnami-docker-airflow/commit/9dc2799c6212f10b49b679cc3e23a5900e07710d)

randradas:
Thank you for the review, regarding the new docker image is already reference in `values.yaml`, check [this](https://github.com/bitnami/charts/pull/7484/files#diff-84ba898d0c90b3758720dd33b065c9ee6715eae86e656c7b957b16401cd56ca6R206).

I will check the rest of comments.

bitnami/airflow/values.yaml
randradas: The previous existing parameters do not work properly with OpenLDAP, let me explain it:
* They were not enough in order to get an Airflow+OpenLDAP integration working. Whatever values you used here it produced the ERR_TOO_MANY_REDIRECTS error.
* _uri port_: we can modify it to 389 because this is the OpenLDAP default. No problem with that. should I proceed?
* _base_: since every user is going to use a different OpenLDAP scheme we should provide default values aligned with a basic OpenLDAP example, typically 'example.org'. According to the X.500 Directory Specification 'example.org' dn should be 'dc=example,dc=org' since they are 'domain components', an OpenLDAP search is going to look for the object inside the 'org' domain component, and then inside 'example' domain component. In this case it suggests that there is a 'admin' object inside the 'example' domain component. Each part of a DNS domain is called DC in OpenLDAP and they should be used as that in an object DN (distinguished name) like 'cn=admin,dc=example,dc=org'. In fact what is wrong is the current existing value since it is not fit neither the OpenLDAP standard nor the typical 'dn' value expected by an OpenLDAP administrator.
* Regarding upgrades: current Airflow+OpenLDAP integrations are not working since the current Chart is not configuring a proper webserver_config.py file. This is because the current values are not enough like explained in the first point.

bitnami/airflow/templates/web/deployment.yaml
randradas: I reviewed these values and I will remove those that are not needed (I can remove a couple of them I think). The ones that remain are needed to avoid being "evaluated" as YAML types (which raises the error below) so they are passed straight as strings to the container which will use them to prepare the webserver_config.yaml file.

Error:
`v1.EnvVar.Value: ReadString: expects " or n, but found t, error found in #10 byte of ...|,"value":true},{"nam|..., bigger context ...|{"name":"AIRFLOW_LDAP_USER_REGISTRATION","value":true},{"name":"AIRFLOW_LDAP_USER_REGISTRATION_ROLE"|...`

bitnami/airflow/templates/web/deployment.yaml
randradas: fixed

bitnami/airflow/templates/web/deployment.yaml
carrodher: Perfect, it's fine with that, just adding it when it is strictly necessary 👍

bitnami/airflow/Chart.yaml
```diff
@@ -32,4 +32,4 @@ name: airflow
 sources:
   - https://github.com/bitnami/bitnami-docker-airflow
   - https://airflow.apache.org/
-version: 10.3.3
+version: 10.3.4
```
carrodher, line 35: A new version was released (`11.0.0`) you need to rebase/fix the conflict and then bump to `11.0.1`

bitnami/airflow/Chart.lock
```diff
@@ -9,4 +9,4 @@ dependencies:
   repository: https://charts.bitnami.com/bitnami
   version: 14.8.11
 digest: sha256:41449a4ca6faecac5fe27c0abd9cc6c8cc495418b7ed92362659ea545f8f439d
-generated: "2021-09-02T22:01:47.392733506Z"
+generated: "2021-08-27T17:51:48.282848+02:00"
```
carrodher, line 12: You can safely remove the changes in the _Chart.lock_ from this PR, this file is going to be automatically updated by the test & release pipeline

bitnami/airflow/values.yaml
carrodher: Understood, it makes sense; thanks. My 2cts: maintain those values that don't need to be modified to make it works like the port, the rest of parameters whatever value that works is fine

bitnami/airflow/Chart.lock
randradas: oks

bitnami/airflow/templates/web/deployment.yaml
randradas: thank you for your review it helped to improve it!

bitnami/airflow/values.yaml
randradas: ok port fixed! thank you @carrodher !

bitnami/airflow/Chart.yaml
randradas: ok!

bitnami/airflow/README.md
randradas: working on that, seems I have some problems with the tool. I'll be back as soon they are fixed.

randradas: fixed1

carrodher (approved):
LGTM

PR #7487, opened 2021-09-15
[bitnami/wordpress] Update Liveness / Readiness Probes to use Wordpress Scheme

### LINKED ISSUE #7459: wordpressScheme=https causes wordpress to hang when waiting for ready status
**Which chart**:
wordpress/5.8.1

**Describe the bug**
Setting value of wordpressScheme=https causes wordpress to hang when waiting for ready status

**To Reproduce**
I used the following in my values.yaml when performing helm install:
```
wordpressScheme: https
ingress:
  ingressClassName: nginx
  enabled: true
  hostname: www.travisloyd.xyz
  tls: true
  annotations:
    nginx.ingress.kubernetes.io/proxy-body-size: 1024m
    cert-manager.io/cluster-issuer: letsencrypt
```

**Expected behavior**
I want my WP_HOME and WP_SITEURL to be https://www.travisloyd.xyz, rather than http://www.travisloyd.xyz and expected wordpressScheme to produce this result.

- Output of `helm version`:
```
version.BuildInfo{Version:"v3.6.3", GitCommit:"d506314abfb5d21419df8c7e7e68012379db2354", GitTreeState:"clean", GoVersion:"go1.16.5"}
```

```
Client Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.3", GitCommit:"ca643a4d1f7bfe34773c74f79527be4afd95bf39", GitTreeState:"clean", BuildDate:"2021-07-15T21:04:39Z", GoVersion:"go1.16.6", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"22", GitVersion:"v1.22.1", GitCommit:"632ed300f2c34f6d6d15ca4cef3d3c7073412212", GitTreeState:"clean", BuildDate:"2021-08-19T15:39:34Z", GoVersion:"go1.16.7", Compiler:"gc", Platform:"linux/amd64"}
```

**Additional context**
1. If I leave out wordpressScheme everything works as expected but WP_HOME & WP_SITEURL are set with http instead of https. (this appears under tools -> site health status)
```
You are accessing this website using HTTPS, but your WordPress Address and Site Address are not set up to use HTTPS by default.

HTTPS is already supported for your website.

However, your WordPress Address is currently controlled by a PHP constant and therefore cannot be updated. You need to edit your wp-config.php and remove or update the definitions of WP_HOME and WP_SITEURL.
```

2. I'm using a third party tool to interact with my wp site and its using these values to try and connect.  Since http is specified instead of https it fails to connect.

itsecureadmin:
**Description of the change**

This change modifies the deployment.yaml to set the port and scheme of both the livenessProbe and readinessProbe to match the Wordpress scheme.

**Benefits**

This will eliminate the requirement to implement a workaround when setting the wordpress scheme to https and using an SSL terminating ingress.

**Possible drawbacks**

**Applicable issues**

  - fixes #7459

**Additional information**

**Checklist** 


migruiz4:
Hi @itsecureadmin,

I'm sorry but I can not accept your approach, as the current livenessProbe and readinessProbe work as intended.

What `livenessProbe: {{- omit .Values.livenessProbe "enabled" | toYaml | nindent 12 }}` does is, it will set the livenessProbe using the `livenessProbe` object in `values.yaml` omitting the 'enabled' field.

To apply the proposed fix of setting automatically the port and scheme, it would be necessary to do the following:

Configure livenessProbe and readinessProbe to use `common.tplvalues.render` instead of `toYaml`:
```
```

Then, modify the values.yaml:
```
livenessProbe:
  enabled: true
  httpGet:
    path: /wp-admin/install.php
    port: '{{ .Values.wordpressScheme }}'
    scheme: '{{ .Values.wordpressScheme | upper }}'
    ## If using an HTTPS-terminating load-balancer, the probes may need to behave
    ## like the balancer to prevent HTTP 302 responses. According to the Kubernetes
    ## docs, 302 should be considered "successful", but this issue on GitHub
    ## (https://github.com/kubernetes/kubernetes/issues/47893) shows that it isn't.
    ## E.g.
    ## httpHeaders:
    ## - name: X-Forwarded-Proto
    ##   value: https
    ##
    httpHeaders: []
  initialDelaySeconds: 120
  periodSeconds: 10
  timeoutSeconds: 5
  failureThreshold: 6
  successThreshold: 1
```

The main difference between `toYaml` and `tplvalues.render` is that render (which uses helm `tpl` function) will evaluate the content of the object before writing it, while toYaml will write them as they are.

itsecureadmin:
Thanks @migruiz4, I have refactored the PR to reflect your suggestions.  I appreciate the feedback.

migruiz4 (approved):
LGTM! Thank you for your contribution!

Issue #7115, opened 2021-08-02
[mongodb]: TLS transport and certificate from valid root certificate authority


Andrei-Stepanov:
Hello.

How do I enable TLS transport?
I do have cert/key.
Cert is issued by DigiCert root CA.
This cert is valid all over the Internet.

I enable option: 

`tls.enabled: true`

Then init-container `generate-tls-certs` will start no matter what: 

https://github.com/bitnami/charts/blob/a9074525f83c97df584fb7ebfe408d57cbb06b22/bitnami/mongodb/templates/standalone/dep-sts.yaml#L118

Please explain, why this init container starts all the time? What the purpose of this container when certificate comes from root CA?

Let's go to official documentation :  [--sslCAFile](https://docs.mongodb.com/manual/reference/program/mongo/#std-label-mongo-shell-tls) :

` if --tlsCAFile or net.tls.CAFile (or their aliases --sslCAFile or ssl.CAFile) is not specified, the system-wide CA certificate store will be used when connecting to an TLS/SSL-enabled server.`

==> `--tlsCAFile` can be skipped at all, however we have:

https://github.com/bitnami/charts/blob/a9074525f83c97df584fb7ebfe408d57cbb06b22/bitnami/mongodb/templates/standalone/dep-sts.yaml#L237

`--tlsCAFile=/certs/mongodb-ca-cert` why? Why this string is present if I have world-wide trusted cert file?

Moreover, mongodb helm chart just doesn't allow specify existing cert + key. 

PS. 
Using function: `buildCustomCert` at is very strange: https://github.com/bitnami/charts/blob/a9074525f83c97df584fb7ebfe408d57cbb06b22/bitnami/mongodb/templates/secrets-ca.yaml#L23

Why do you require to do base64 enc of PEM format? This is not functional. Yaml supports:

```
cert: |
    -----BEGIN CERTIFICATE-----
    MIIIFTCCBv2gAwIBAgIQCra4Kst7
    ....
    -----END CERTIFICATE-----
```

The helm comment for values says:  https://github.com/bitnami/charts/blob/a9074525f83c97df584fb7ebfe408d57cbb06b22/bitnami/mongodb/values.yaml#L168

`@param tls.caCert Custom CA certificated (base64 encoded)` 
How do I know that you want base64 encoded PEM certificate, that already inside has base64 encode certificate?
What is purpose of encoding PEM-files?
:-(

miguelaeh:
Hi @Andrei-Stepanov ,

[...]

PR #7503, opened 2021-09-16
[bitnami/mopngodb] Add condition for auto generated certs

miguelaeh:
**Description of the change**

**Benefits**

**Possible drawbacks**

**Applicable issues**

  - fixes #

**Additional information**

**Checklist** 


migruiz4:
Could you please share the associated issue?

The MongoDB case is a bit particular because it already had support for the autogenerated certificates feature before we added it to other charts, so we simply adapted it to use the same values.

One of those things that differentiate MongoDB from other charts, is that it has an init container that allows users to provide their CA and it would generate the node certificates. 

Adding the `if` condition is pointless because the init container is not related to `tls.autogenerated=true` (which uses [helm functions](https://helm.sh/docs/chart_template_guide/function_list/#gensignedcert) to generate the certs), we should either do a complete refactor dropping the feature, or leave it as it is unless it generates further issues.

miguelaeh:
This is the issue: https://github.com/bitnami/charts/issues/7115 I can also see more issues related looking for certificates

PR #7505, opened 2021-09-16
[bitnami/redis] Chart standardized

migruiz4:
**Description of the change**

This PR standardizes the bitnami/redis chart to be in line with the rest of the catalog.

**Checklist** 


alvneiayu (approved):
LGTM

juan131:
Please ping me when you rebase your branch and the PR is ready to be reviewed @migruiz4

bitnami/redis/values.yaml
```diff
@@ -852,11 +868,26 @@ sentinel:
...
+  ## @param sentinel.startupProbe.failureThreshold Failure threshold for startupProbe
+  ## @param sentinel.startupProbe.successThreshold Success threshold for startupProbe
+  ##
+  startupProbe:
+    enabled: false
+    initialDelaySeconds: 20
+    periodSeconds: 5
+    timeoutSeconds: 5
+    successThreshold: 1
+    failureThreshold: 5
```
juan131, line 890: These parameters aren't used anywhere

bitnami/redis/values.yaml
```diff
@@ -885,6 +916,9 @@ sentinel:
     timeoutSeconds: 1
     successThreshold: 1
     failureThreshold: 5
+  ## @param sentinel.customStartupProbe Custom startupProbe that overrides the default one
+  ##
+  customStartupProbe: {}
```
juan131, line 921: Neither this one

juan131 (changes requested):
There are new parameters that you didn't include in the manifests

bitnami/redis/templates/master/statefulset.yaml
```diff
@@ -166,6 +166,22 @@ spec:
               containerPort: {{ .Values.master.containerPorts.redis }}
           {{- if not .Values.diagnosticMode.enabled }}
           {{- if .Values.master.livenessProbe.enabled }}
+          startupProbe:
+            initialDelaySeconds: {{ .Values.master.startupProbe.initialDelaySeconds }}
+            periodSeconds: {{ .Values.master.startupProbe.periodSeconds }}
+            # One second longer than command timeout should prevent generation of zombie processes.
+            timeoutSeconds: {{ add1 .Values.master.startupProbe.timeoutSeconds }}
+            successThreshold: {{ .Values.master.startupProbe.successThreshold }}
+            failureThreshold: {{ .Values.master.startupProbe.failureThreshold }}
```
juan131, line 175: The previous condition is wrong (line 168):
```diff
-{{- if .Values.master.livenessProbe.enabled }}
+{{- if .Values.master.startupProbe.enabled }}
```
Also.. use "omit":
```suggestion
          startupProbe: {{- include "common.tplvalues.render" (dict "value" (omit .Values.master.startupProbe "enabled") "context" $) | nindent 12 }}
```

juan131 (changes requested):

bitnami/redis/values.yaml
migruiz4: They are used here:
https://github.com/bitnami/charts/blob/85085f8232988fc7aeaa3a86393638db44bbb19f/bitnami/redis/templates/sentinel/statefulset.yaml#L348-L362

bitnami/redis/values.yaml
migruiz4: Check reply above

bitnami/redis/templates/master/statefulset.yaml
migruiz4: What do you recommend me to do with this? 
```
            # One second longer than command timeout should prevent generation of zombie processes.
            timeoutSeconds: {{ add1 .Values.master.startupProbe.timeoutSeconds }}
```
Not sure what this workaround is intended for, but it would be lost if we use this method

migruiz4:
Tests are failing, but I think it is caused by this PR: https://github.com/bitnami/charts/pull/8641

bitnami/redis/templates/master/statefulset.yaml
juan131: Okay.. you can keep it that for the readiness/liveness probes, but I don't find it useful for the startup probe.
The startup probe shouldn't be using `ping_liveness_local.sh` nor `ping_readiness_local.sh` script, instead a simple "tcpSocket" probe should be enough. The goal of a startup probe is not checking the app's ready to accept connections or alive but just ensuring it's been started.

juan131 (changes requested):

juan131 (approved):
LGTM

PR #4940, opened 2021-01-11
[bitnami/etcd] Reuse startFromSnapshot volume in the disasterRecovery mode

carrodher:
**Description of the change**

Regardless of data volume, there are different combinations for **disasterRecovery** and/or **startFromSnapshot**:
- **Case 1**: disasterRecovery:white_check_mark: startFromSnapshot:red_circle: -> No volumes in play
- **Case 2**: disasterRecovery:white_check_mark: startFromSnapshot:red_circle: -> The volume is created on-demand or provided by users. This case does not change (Total 1 volume)
- **Case 3**: disasterRecovery:red_circle: startFromSnapshot:white_check_mark: -> Users always give us the volume. This case does not change (Total 1 volume)

The complex case that we want to improve in this PR:
- **Case 4**: disasterRecovery:white_check_mark: startFromSnapshot:white_check_mark:

_**Right now**_: _disasterRecovery_ volume is created or an existing one is used. _startFromSnapshot_ give us the volume (Total 2 volumes)
**_Desired_**: _disasterRecovery_ if the users give it to us, it is used and apart from that, the one provided for _startFromSnapshot_ (Total 2 volumes). But if users don't give us the _disasterRecovery_ volume, the one they give us in _startFromSnapshot_ is used instead of creating a new one (Total 1 volume)

So only in the case of
> `disasterRecovery.enabled = true` && `startFromSnapshot.enabled = true` && ! `disasterRecovery.pvc.existingClaim`

it is when the volume passed as `startFromSnapshot.existingClaim` would be used to make the function of the one that we would create for the _disasterRecovery_

-----------------------------------------------------------
Manual tests:
- **Case 2.1**, no changes with the new behavior
```console
$ helm template bitnami/etcd --set disasterRecovery.enabled=true > old.yaml
$ helm template . --set disasterRecovery.enabled=true > new.yaml
```
```diff
$ colordiff old.yaml new.yaml
14c14
-   etcd-root-password: "SzdpUkZyc3BKcA=="
+   etcd-root-password: "SHJDRjF1TXJFeQ=="
```
- **Case 2.2**, no changes with the new behavior
```console
$ helm template bitnami/etcd --set disasterRecovery.enabled=true --set disasterRecovery.pvc.existingClaim="foo" > old.yaml
$ helm template . --set disasterRecovery.enabled=true --set disasterRecovery.pvc.existingClaim="foo" > new.yaml
```
```diff
$ colordiff old.yaml new.yaml
14c14
[...]

Issue #6251, opened 2021-04-29
[bitnami/etcd] Pods not recovering - Can't update initialClusterState


Lavaburn:
**Which chart**: bitnami/etcd 6.2.3

**Describe the bug**
Pods are crashing when changing initialClusterState from "new" to "existing".

```
etcd 10:44:26.03 INFO ==> Updating member in existing cluster
Error: bad member ID arg (strconv.ParseUint: parsing "": invalid syntax), expecting ID in Hex
```

**To Reproduce**
1. Deploy this chart with initialClusterState: "new" (extra values below)
2. Update this chart with new values: initialClusterState: "existing"

**Expected behavior**
When leaving the initialClusterState: "new" (or not setting it), pods crashing are not recovering. It is expected that setting initialClusterState to existing would resolve this.

When updating the initialClusterState to "existing", the pod should rejoin the cluster and be able to recover from pod crash.

**Version of Helm and Kubernetes**:

- Output of `helm version`:
```
version.BuildInfo{Version:"v3.5.3", GitCommit:"041ce5a2c17a58be0fcd5f5e16fb3e7e95fea622", GitTreeState:"dirty", GoVersion:"go1.15.8"}
```
- Output of `kubectl version`:
```
Client Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.5", GitCommit:"6b1d87acf3c8253c123756b9e61dac642678305f", GitTreeState:"clean", BuildDate:"2021-03-18T01:10:43Z", GoVersion:"go1.15.8", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.10", GitCommit:"f3add640dbcd4f3c33a7749f38baaac0b3fe810d", GitTreeState:"clean", BuildDate:"2020-05-20T13:51:56Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}
```

**Additional context**

values.yaml:
```
replicaCount: 3

auth:
  rbac:
    enabled: false

persistence:
  enabled: true
```

jaspermarcus:
I believe we could be running into something similar in our deployments on helm chart version `6.2.0`. We periodically deploy changes via `helm upgrade --install`, and hence `ETCD_INITIAL_CLUSTER_STATE` also transitioned for us at some point from `new` to `existing`. 

[...]

PR #5682, opened 2021-03-04
[bitnami/etcd] Major version: refactoring

juan131:
Signed-off-by: juan131 <juanariza@vmware.com>

**Description of the change**

This version introduces several features and performance improvements:

- The statefulset can now be scaled using `kubectl scale` command. Using `helm upgrade` to recalculate available endpoints is no longer needed.
- The scripts used for bootstrapping, runtime reconfiguration, and disaster recovery have been refactored and moved to the etcd container (see https://github.com/bitnami/bitnami-docker-etcd/pull/13) with two purposes: removing technical debt & improving the stability.
- Several parameters were reorganized to simplify the structure and follow the same standard used on other Bitnami charts
- New parameters/features were added
- Metrics integration with Prometheus Operator does no longer use a ServiceMonitor object, but a PodMonitor instead.

Find detailed info in the "Upgrade" section of the README.md

Also, some docs are moved to the Bitnami Docs System including new guides that explain how the clustering works. See https://github.com/bitnami/charts-docs/pull/11

**Benefits**

- Standardization
- Remove tech debt
- Better stability (🤞 )

**Possible drawbacks**

None AFAIK

**Applicable issues**

- fixes https://github.com/bitnami/charts/issues/1513
- fixes https://github.com/bitnami/charts/issues/5360
- fixes https://github.com/bitnami/charts/issues/5443
- fixes https://github.com/bitnami/bitnami-docker-etcd/issues/10
- fixes https://github.com/bitnami/bitnami-docker-etcd/issues/7

**Additional information**

Still testing some corner cases, thus the "draft" status.

**Checklist**

- [x] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [x] Variables are documented in the README.md
- [x] Title of the PR starts with chart name (e.g. `[bitnami/chart]`)

PR #7511, opened 2021-09-16
[bitnami/etcd] feat: allow reuse startFromSnapshot volume in the disasterRecovery

relzhong:
add support to allow reuse startFromSnapshot volume in the disasterRecovery. fix breaking change to relate [pr](https://github.com/bitnami/charts/pull/4940)
relate pr in bitnami-docker-etcd [pr](https://github.com/bitnami/bitnami-docker-etcd/pull/22)

**Benefits**

**Possible drawbacks**

**Applicable issues**

  - fixes #

**Additional information**

**Checklist** 


yilmi:
Let's wait for the container changes from https://github.com/bitnami/bitnami-docker-etcd/pull/22 are reviewed an merge before working on this one 👍

yilmi:
@relzhong reviewing your change and the description from https://github.com/bitnami/charts/pull/4940, I don't see what problem you're trying to solve.

Could you please provide more details in your issue, such as Benefits, Possible Drawbacks and more details to explain why the current logic doesn't work for you?

Thanks,
Yass

relzhong:
Hi @yilmi ,
In current disasterRecovery phase, we need to starting a new cluster using an existing snapshot, so the precedure will like this:
1. create cluster and add a disasterrecovery pvc.
```
  --disasterRecovery.pvc.existingClaim=pvc1
```
2. after first snapshot, we get snapshot1, then we uninstall the chart. and recreate using start from snapshot
```
  --startFromSnapshot.existingClaim=pvc1 --startFromSnapshot.snapshotFilename=snapshot1
```
if we also need disasterRecovery, we should create pvc2, and set params to 
```
--disasterRecovery.pvc.existingClaim=pvc2 --startFromSnapshot.existingClaim=pvc1 --startFromSnapshot.snapshotFilename=snapshot1 
```
the reason why create pvc2 is beacuse setting ETCD_INIT_SNAPSHOT_DIR to a fix /init-snapshot.

with this pr, we can just modify to ETCD_INIT_SNAPSHOT_DIR to /snapshots. so we just need one pvc pvc1. and bring back cluster will be this 
```
--disasterRecovery.pvc.enabled=true --startFromSnapshot.existingClaim=pvc1 --startFromSnapshot.snapshotFilename=snapshot1 
```
The advantage is we need just one pvc and no need to switch pvc between pvc1 and pv2 on every disaster recovery.

Thanks.

yilmi:
Hi @relzhong, thanks for providing more details!

I read through the PR you mention in your description, and I believe that your desired state would be achieved by the case 4.2 described from https://github.com/bitnami/charts/pull/4940

> So only in the case of
>
> `disasterRecovery.enabled = true` && `startFromSnapshot.enabled = true` && ! `disasterRecovery.pvc.existingClaim`
>
> it is when the volume passed as `startFromSnapshot.existingClaim` would be used to make the function of the one that we would create for the disasterRecovery

The way I read it, if you set:
- `disasterRecovery.enabled=true`
- `startFromSnapshot.enabled=true`
- `startFromSnapshot.existingClaim="xxxx"`

Without setting `disasterRecovery.enabled`, you should get the behaviour you want.

relzhong:
In #4940, the logic had been merge to master. However, after the [refactor](https://github.com/bitnami/charts/pull/5682) of both etcd chart and  bitnami-docker-etcd, the logic had been move from chart to docker-etcd. The [script](https://github.com/bitnami/charts/blob/1fb5d2c9a0473f74f11233b59b784982df34a9e3/bitnami/etcd/templates/scripts-configmap.yaml) changed in #4940 cannot be found in current master branch. But the script in bitnami-docker-etcd has a fixed directory /init-snapshot. So i modify both project to achieve this.

yilmi:
Hi @relzhong, thanks for the pointers. Let me take some time to test and look into this before getting back to you. I should be able to do it by the end of the week or early next week. 👍

juan131 (changes requested):
I agree with @yilmi , the idea off [this PR ](https://github.com/bitnami/charts/pull/4940) was reusing the same PV both for "start from snapshot" and "disaster recovery features"... But it's true that we messed up this when we moved the logic to the container as you pointed...

juan131:
Potential fix for https://github.com/bitnami/charts/issues/6251

relzhong:
fixed @juan131

bitnami/etcd/templates/statefulset.yaml
```diff
@@ -217,6 +217,10 @@ spec:
             - name: ETCD_INIT_SNAPSHOT_FILENAME
               value: {{ .Values.startFromSnapshot.snapshotFilename | quote }}
             {{- end }}
+            {{- if and .Values.disasterRecovery.enabled .Values.startFromSnapshot.enabled (not .Values.disasterRecovery.pvc.existingClaim) }}
+            - name: ETCD_INIT_SNAPSHOTS_DIR
+              value: "/snapshots"
```
juan131, line 222:
```suggestion
            - name: ETCD_INIT_SNAPSHOTS_DIR
              value: {{ ternary "/snapshots" "/init-snapshot" (and .Values.disasterRecovery.enabled (not .Values.disasterRecovery.pvc.existingClaim)) | quote }}
```

juan131: What do you think about this proposal? This way we always align `ETCD_INIT_SNAPSHOTS_DIR` with the "volumeMounts" used in the chart (regardless the default value used in the image).

relzhong: Sure, it's better solution

juan131 (approved):
LGTM

juan131:
You can always build your custom image. To do so, clone the https://github.com/bitnami/bitnami-docker-etcd repo, browse to the directory `3/debian-10`, and apply the changes below in the Dockerfile:
 
```diff
 FROM docker.io/bitnami/minideb:buster
 (...)
 RUN . /opt/bitnami/scripts/libcomponent.sh && component_unpack "gosu" "1.14.0-0" --checksum 3e6fc37ca073b10a73a804d39c2f0c028947a1a596382a4f8ebe43dfbaa3a25e
-RUN . /opt/bitnami/scripts/libcomponent.sh && component_unpack "etcd" "3.5.1-0" --checksum 912653e5c0879fd27c8bb0327bea171d7ae504611ec0a9c8fcc2ec6184766a63
+RUN . /opt/bitnami/scripts/libcomponent.sh && component_unpack "etcd" "3.4.16-0" --checksum 87ffa91cbaf5ad953cae045b8dfb4cb4f12c66da499bd332d2b4acf2f01941f0
 RUN chmod g+rwX /opt/bitnami
(...)
 ENV BITNAMI_APP_NAME="etcd" \
-    BITNAMI_IMAGE_VERSION="3.5.1-debian-10-r4" \
+    BITNAMI_IMAGE_VERSION="3.4.16-debian-10-r0" \
     ETCDCTL_API="3"
(...)
 CMD [ "/opt/bitnami/scripts/etcd/run.sh" ]
```

Then, use `docker build ...` to build your custom etcd image that uses `3.4.16` but including the latest scripts.

PR #7520, opened 2021-09-16
[bitnami/redis] Use timeoutSeconds to timeout health checks

stuartpb:
The timeout wrapper in health checks was added in helm/charts#11355
to work around Docker/containerd not respecting timeouts in probes
(cf. kubernetes/kubernetes#58925). The upstream issue has been fixed
since Kubernetes 1.20 (kubernetes/kubernetes#94115), and this wrapper
causes degraded behavior (ie. any failure in the wrapped command only
gets reported as "The monitored command dumped core", without details
for the specific failure), so the original behavior should be restored.

**Checklist** 

### DIFF
--- a/bitnami/redis/Chart.yaml
+++ b/bitnami/redis/Chart.yaml
@@ -25,4 +25,4 @@ name: redis
 sources:
   - https://github.com/bitnami/bitnami-docker-redis
   - http://redis.io/
-version: 15.3.2
+version: 15.3.3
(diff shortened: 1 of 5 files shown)


Mauraza:
Hi @stuartpb,

Thanks for the PR!. Although this has been fixed in version 1.20 of Kubernetes, we continue to support older versions so we still can not accept this PR.

PR #7547, opened 2021-09-20
[bitnami/argo-cd] Calculate password Mtime using always UTC

miguelaeh:
**Description of the change**

**Benefits**

**Possible drawbacks**

**Applicable issues**

  - fixes #

**Additional information**

**Checklist** 


alemorcuq (commented):
Could you add some context on why this change?

miguelaeh:
Basically, you need to calculate the time in the same timezone used by the Argo CD application, if not it will interpret the password has changed and you will be unable to login

alemorcuq:
> Basically, you need to calculate the time in the same timezone used by the Argo CD application, if not it will interpret the password has changed and you will be unable to login

I assume Argo CD is configured to use UTC by default, right? Can users change this? What would happen if a user changes the timezone?

miguelaeh:
The timezone cannot be changed, the sync windows are hardcoded to UTC: https://github.com/argoproj/argo-cd/blob/bb88a1c975b43a94da34c4516deaec44f32ac790/pkg/apis/application/v1alpha1/types.go#L1698

alemorcuq (approved):

PR #7554, opened 2021-09-21
[bitnami/argo-workflows] Add new Argo Workflows Helm Chart

miguelaeh:
**Description of the change**

Add a new Helm Chart for Argo Workflows

**Benefits**

**Possible drawbacks**

**Applicable issues**

  - fixes #

**Additional information**

**Checklist** 

- [x] Title of the PR starts with chart name (e.g. [bitnami/<name_of_the_chart>]


bitnami/argo-workflows/values.yaml
```diff
@@ -0,0 +1,1025 @@
...
+  ## @param controller.persistence.mysql.enabled Enable MySQL persistence
+  ## @param controller.persistence.mysql.host Host to use mysql persistence
+  ## @param controller.persistence.mysql.port Port for Postgresql
+  ## @param controller.persistence.mysql.database Postgresql database
+  ## @param controller.persistence.mysql.tableName Postgresql table name
+  ## @param controller.persistence.mysql.userNameSecret.name Name of the secret containing the mysql username
+  ## @param controller.persistence.mysql.userNameSecret.key Key of the secret containing the mysql username
+  ## @param controller.persistence.mysql.passwordSecret.name Name of the secret containing the mysql password
+  ## @param controller.persistence.mysql.passwordSecret.key Key of the secret containing the mysql password
+  ##
```
javsalgar, line 533: The following values seem to be missing in the values.yaml file

```
Key controller.extraArgs not found
Key controller.instanceID.explicitInstanceID not found
Key executor.securityContext not found
Key mysql.nameOverride not found
Key postgresql.nameOverride not found
Key server.auth.sso.rbac.secretWhitelist not found
Key server.extraArgs not found
```

bitnami/argo-workflows/templates/server/service.yaml
```diff
@@ -0,0 +1,31 @@
...
+  {{- end }}
+  {{- if eq .Values.server.service.type "LoadBalancer" }}
+  loadBalancerSourceRanges: {{ .Values.server.service.loadBalancerSourceRanges }}
+  {{- end }}
+  {{- if (and (eq .Values.server.service.type "LoadBalancer") (not (empty .Values.server.service.loadBalancerIP))) }}
+  loadBalancerIP: {{ .Values.server.service.loadBalancerIP }}
+  {{- end }}
+  ports:
+    - name: http
+      port: {{ .Values.server.service.port }}
```
javsalgar, line 28: We are switching to a more standard approach, which would be `.Values.server.service.ports.http`

Even though there is only one port, this would avoid major bumps when adding new ports to the template, as well as being consistent with all charts. This is part of the new standardization.

bitnami/argo-workflows/templates/server/service.yaml
```diff
@@ -0,0 +1,31 @@
...
+metadata:
+  name: {{ include "argo-workflows.server.fullname" . }}
+  namespace: {{ .Release.Namespace | quote }}
+  labels: {{- include "common.labels.standard" . | nindent 4 }}
+    app.kubernetes.io/part-of: argo-workflows
+    app.kubernetes.io/component: server
+    {{- if .Values.commonLabels }}
+    {{- include "common.tplvalues.render" ( dict "value" .Values.commonLabels "context" $ ) | nindent 4 }}
+    {{- end }}
+  {{- if .Values.commonAnnotations }}
```
javsalgar, line 12: Missing service.annotations

bitnami/argo-workflows/templates/server/service.yaml
```diff
@@ -0,0 +1,31 @@
...
+  annotations: {{- include "common.tplvalues.render" ( dict "value" .Values.commonAnnotations "context" $ ) | nindent 4 }}
+  {{- end }}
+spec:
+  type: {{ .Values.server.service.type }}
+  {{- if (or (eq .Values.server.service.type "LoadBalancer") (eq .Values.server.service.type "NodePort")) }}
+  externalTrafficPolicy: {{ .Values.server.service.externalTrafficPolicy | quote }}
```
javsalgar, line 18: Missing these:

```
Value controller.service.nodePorts.http not found in template
Value controller.service.clusterIP not found in template
Value controller.service.clusterIP not found in template
```

bitnami/argo-workflows/templates/controller/service.yaml
```diff
@@ -0,0 +1,43 @@
...
+  ports:
+  {{- if .Values.controller.metrics.enabled }}
+  - name: {{ .Values.controller.metrics.servicePortName }}
+    port: {{ .Values.controller.metrics.servicePort }}
+    protocol: TCP
+    targetPort: {{ .Values.controller.metrics.port }}
+  {{- end }}
+  {{- if .Values.controller.telemetry.enabled }}
+  - name: {{ .Values.controller.telemetry.servicePortName }}
+    port: {{ .Values.controller.telemetry.servicePort }}
```
javsalgar, line 36: This is a case where we want to go to the standard

```
controller.service.ports.telemetry
controller.service.ports.metrics
```

As this is part of the application itself and part of the same service, I would move the port definition from the controller.metrics to controller.service

bitnami/argo-workflows/templates/controller/service.yaml
```diff
@@ -0,0 +1,43 @@
...
+  {{- end }}
+  ports:
+  {{- if .Values.controller.metrics.enabled }}
+  - name: {{ .Values.controller.metrics.servicePortName }}
+    port: {{ .Values.controller.metrics.servicePort }}
+    protocol: TCP
+    targetPort: {{ .Values.controller.metrics.port }}
+  {{- end }}
+  {{- if .Values.controller.telemetry.enabled }}
+  - name: {{ .Values.controller.telemetry.servicePortName }}
```
javsalgar, line 35: Not sure if I see any value in defining the name of the service in the values.yaml file. I imagine this comes from upstream but I don't see why we would want that instead of hardcoding it like the rest.

bitnami/argo-workflows/templates/controller/service.yaml
```diff
@@ -0,0 +1,43 @@
...
+  {{ if eq .Values.controller.service.type "LoadBalancer" }}
+  loadBalancerSourceRanges: {{ .Values.controller.service.loadBalancerSourceRanges }}
+  {{ end }}
+  {{- if (and (eq .Values.controller.service.type "LoadBalancer") (not (empty .Values.controller.service.loadBalancerIP))) }}
+  loadBalancerIP: {{ .Values.controller.service.loadBalancerIP }}
+  {{- end }}
+  ports:
+  {{- if .Values.controller.metrics.enabled }}
+  - name: {{ .Values.controller.metrics.servicePortName }}
+    port: {{ .Values.controller.metrics.servicePort }}
```
javsalgar, line 30: Missing these 

```
Value controller.service.nodePorts.metrics not found in template
Value controller.service.nodePorts.telemetry not found in template
Value controller.service.clusterIP not found in template
Value controller.service.annotations not found in template
```

bitnami/argo-workflows/Chart.yaml
```diff
@@ -0,0 +1,38 @@
+annotations:
+  category: Infrastructure
+apiVersion: v2
+appVersion: 3.1.10
+dependencies:
+  - condition: controller.persistence.postgresql.enabled
+    name: postgresql
+    repository: https://charts.bitnami.com/bitnami
+    version: 10.x.x
+  - condition: controller.persistence.mysql.enabled
```
javsalgar, line 10: Use the same standard as with other charts. mysql.enabled and postgresql.enabled.

I see that it is using mysql and most of ours are using mariadb. Any reason for this?

bitnami/argo-workflows/values.yaml
```diff
@@ -0,0 +1,1025 @@
...
+    ## Secrets must be manually created in the namespace.
+    ## ref: https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/
+    ## e.g:
+    ## pullSecrets:
+    ##   - myRegistryKeySecretName
+    ##
+    pullSecrets: []
+    ## Enable debug mode
+    ##
+    debug: false
```
javsalgar, line 439: This is not used

bitnami/argo-workflows/templates/server/service-account.yaml
```diff
@@ -0,0 +1,13 @@
...
+  name: {{ include "argo-workflows.server.serviceAccountName" . }}
+  namespace: {{ .Release.Namespace | quote }}
+  labels: {{- include "common.labels.standard" . | nindent 4 }}
+    app.kubernetes.io/component: server
+    {{- if .Values.commonLabels }}
+    {{- include "common.tplvalues.render" ( dict "value" .Values.commonLabels "context" $ ) | nindent 4 }}
+    {{- end }}
+  {{- if .Values.commonAnnotations }}
+  annotations: {{- include "common.tplvalues.render" ( dict "value" .Values.commonAnnotations "context" $ ) | nindent 4 }}
+  {{- end }}
```
javsalgar, line 13: Missing the automountServiceAccountToken (I think that's the name), which is now part of the standards

bitnami/argo-workflows/templates/server/deployment.yaml
```diff
@@ -0,0 +1,171 @@
...
+            - --auth-mode
+            - {{ .Values.server.auth.mode }}
+            {{- end }}
+            {{- if .Values.server.extraArgs }}
+            {{- include "common.tplvalues.render" (dict "value" .Values.server.extraArgs "context" $) | nindent 12 }}
+            {{- end }}
+            {{- end }}
+          ports:
+            - name: web
+              containerPort: {{ .Values.server.ports.container }}
```
javsalgar, line 94: Move to the new standard

.Values.server.containerPorts.web (or http, name it as you prefer)

bitnami/argo-workflows/templates/server/service.yaml
```diff
@@ -0,0 +1,31 @@
...
+  loadBalancerSourceRanges: {{ .Values.server.service.loadBalancerSourceRanges }}
+  {{- end }}
+  {{- if (and (eq .Values.server.service.type "LoadBalancer") (not (empty .Values.server.service.loadBalancerIP))) }}
+  loadBalancerIP: {{ .Values.server.service.loadBalancerIP }}
+  {{- end }}
+  ports:
+    - name: http
+      port: {{ .Values.server.service.port }}
+      targetPort: web
+  selector: {{- include "common.labels.matchLabels" . | nindent 4 }}
```
javsalgar, line 30: Missing extraPorts, which is part of the new standards. Apply to the other services

bitnami/argo-workflows/templates/controller/configmap.yaml
```diff
@@ -0,0 +1,91 @@
+apiVersion: v1
```
javsalgar, line 1: Missing the feature to add an existing configmap

bitnami/argo-workflows/templates/controller/configmap.yaml
```diff
@@ -0,0 +1,91 @@
...
+    app.kubernetes.io/part-of: argo-workflows
+    app.kubernetes.io/component: controller
+    {{- if .Values.commonLabels }}
+    {{- include "common.tplvalues.render" ( dict "value" .Values.commonLabels "context" $ ) | nindent 4 }}
+    {{- end }}
+  {{- if .Values.commonAnnotations }}
+  annotations: {{- include "common.tplvalues.render" ( dict "value" .Values.commonAnnotations "context" $ ) | nindent 4 }}
+  {{- end }}
+data:
+  config: |
```
javsalgar, line 16: I'd go more for the JupyterHub approach and move this template to the values.yaml file, so users can easily edit it without having to create a configmap

bitnami/argo-workflows/templates/server/clusterrolebinding.yaml
```diff
@@ -0,0 +1,58 @@
+{{- if and .Values.server.enabled .Values.server.serviceAccount.create -}}
```
javsalgar, line 1: I'd add an rbac.create value just like the other charts

javsalgar: Also, I dont' see why we need serviceAccount.create to be true. We could be using an existing serviceAccount, that's what the `argo-workflows.server.serviceAccountName` is here for

bitnami/argo-workflows/templates/server/clusterrolebinding.yaml
```diff
@@ -0,0 +1,58 @@
...
+apiVersion: rbac.authorization.k8s.io/v1
+{{- if .Values.singleNamespace }}
```
javsalgar, line 3: I'd rename it to `rbac.singleNamespace`

bitnami/argo-workflows/values.yaml
```diff
@@ -0,0 +1,1025 @@
...
+  ports:
+    container: 6060
+
+  ## Controller persistence parameters. It can also be connected to an existing postgresql. Check
+  ## @param controller.persistence.enabled Enabled controller persistence
+  ## @param controller.persistence.connectionPool.maxIdleConns Msximun number of postgresql connections
+  ## @param controller.persistence.connectionPool.maxOpenConns Maximun number of open connections
+  ## @param controller.persistence.nodeStatusOffLoad Save the entire workflow into etcd and DB
+  ## @param controller.persistence.archive Enable archiving old workflows
+  ## @param controller.persistence.postgresql.enabled Enable Postgresql persistence
```
javsalgar, line 515: I see this is actually part of the configuration file. I don't like mixing parts of the chart values with the application configuration (it is very confusing in terms of maintenance). Plus, we should find a similar user experience when dealing with sub-charts. Feel free to take a look to the JupyterHub case.

bitnami/argo-workflows/values.yaml
```diff
@@ -0,0 +1,1025 @@
...
+    archive: false
+    postgresql:
+      enabled: false
+      host: '{{ include "argo-workflows.postgresql.fullname" . }}'
+      port: 5432
+      database: postgres
+      tableName: argo_workflows
+      ## the database secrets must be in the same namespace of the controller
+      userNameSecret:
+        name: argo-postgres-config
```
javsalgar, line 551: Hardcoding this would be a bit problematic, because what about the case where the user specifies its own secret or it is using the `externalDatabase` section of the values (which, by the way, I think it is not here yet, right?)

bitnami/argo-workflows/values.yaml
```diff
@@ -0,0 +1,1025 @@
...
+        key: username
+      passwordSecret:
+        name: argo-postgres-config
+        key: password
+    mysql:
+      enabled: false
+      host: '{{ include "argo-workflows.mysql.fullname" . }}'
+      port: 3306
```
javsalgar, line 559: This should not be hardcoded, check Jupyterhub for example

javsalgar (changes requested):
Thank you! Changes requested

bitnami/argo-workflows/Chart.yaml
miguelaeh: The reason about using MySQL instead of MariaDB is that the upstream ones references MySQL not MariaDB, but in any case, this is a feature I added that is not in the upstream one, so I guess we can configure it however we want.

bitnami/argo-workflows/values.yaml
miguelaeh: I added the dependencies in order to configure MySQL or postgresql automatically. That was not supported by upstream, and it is totally optional to use a database, so maybe what we should do is to only support external database  letting the user to configure all the parameters in that case.
I need to re-evaluate this after applying the changes to not provide config parameters as chart parameters.

bitnami/argo-workflows/Chart.yaml
miguelaeh: I moved to the standard making a big refactor of the configuration, about using mariadb instead of MySQL the change should be trivial, but there is no reference to mariadb in the docs:
```
configure a Postgres or MySQL (>= 5.7.8) database under 
```

bitnami/argo-workflows/templates/server/ingress.yaml
```diff
@@ -0,0 +1,69 @@
...
+  namespace: {{ .Release.Namespace | quote }}
+  labels: {{- include "common.labels.standard" . | nindent 4 }}
+    app.kubernetes.io/component: server
+    app.kubernetes.io/part-of: argo-workflows
+    {{- if .Values.commonLabels }}
+    {{- include "common.tplvalues.render" ( dict "value" .Values.commonLabels "context" $ ) | nindent 4 }}
+    {{- end }}
+  {{- if or .Values.ingress.annotations .Values.commonAnnotations .Values.ingress.certManager }}
+  annotations:
+    ingress.kubernetes.io/rewrite-target: /$2
```
javsalgar, line 15: Could you add a small comment explaining this? Not sure if the upstream one provides an explanation

bitnami/argo-workflows/values.yaml
```diff
@@ -0,0 +1,1075 @@
...
+    ## @param server.service.type server service type
+    ##
+    type: ClusterIP
+    ## @param server.service.ports.http server service HTTP port
+    ##
+    ports:
+      http: 80
+    ## @param server.service.httpsPort server service HTTPS port
+    ##
+    httpsPort: 443
```
javsalgar, line 382: move to ports.https. Is this actually being used?

bitnami/argo-workflows/templates/server/ingress.yaml
```diff
@@ -0,0 +1,69 @@
+{{- if .Values.ingress.enabled }}
```
javsalgar, line 1: Missing the ingress-tls.yaml file for the autogenerated secrets

bitnami/argo-workflows/values.yaml
```diff
@@ -0,0 +1,1075 @@
...
+
+## Postgresql subchart configuration
+## @param postgresql.enabled Enable PostgreSQL subchart and controller persistence using PostgreSQL
+## @param postgresql.port PostgreSQL port
+## @param postgresql.postgresqlUsername PostgreSQL username
+## @param postgresql.postgresqlDatabase PortgreSQL database name
+##
+postgresql:
+  enabled: false
+  port: 5432
```
javsalgar, line 1030: This should be postgresql.service.port. postgresql.port is not used at all.

bitnami/argo-workflows/values.yaml
```diff
@@ -0,0 +1,1075 @@
...
+    ## Uncomment to create containerRuntimeExecutors
+    ## Ref: https://argoproj.github.io/argo-workflows/workflow-executors/
+    ## containerRuntimeExecutors: {}
+    ## How many workflows can be running at the same time
+    parallelism:
+    ## Maximun number of workflows running in a namespace
+    namespaceParallelism:
+    {{- if or .Values.executor.resources .Values.executor.extraEnvVars .Values.executor.podSecurityContext }}
+    executor:
+      {{- if .Values.executor.resources }}
```
javsalgar, line 551: Can you set the executor imagePolicy here? Not sure if these are elements added to the executor container.

bitnami/argo-workflows/values.yaml
```diff
@@ -0,0 +1,1075 @@
...
+commonLabels: {}
+## @param commonAnnotations Annotations to add to all deployed objects
+##
+commonAnnotations: {}
+## @param clusterDomain Kubernetes cluster domain name
+##
+clusterDomain: cluster.local
+## @param extraDeploy Array of extra objects to deploy with the release
+##
+extraDeploy: []
```
javsalgar, line 40: Missing extra-list.yaml

javsalgar (changes requested):
Thank you for all the refactor! Please check these

bitnami/argo-workflows/templates/controller/deployment.yaml
```diff
@@ -0,0 +1,170 @@
...
+            {{- end }}
+            {{- if .Values.controller.extraEnvVarsSecret }}
+            - secretRef:
+                name: {{ include "common.tplvalues.render" (dict "value" .Values.controller.extraEnvVarsSecret "context" $) }}
+            {{- end }}
+          ports:
+            - name: metrics
+              containerPort: {{ .Values.controller.metrics.port }}
+            - name: controller
+              containerPort: {{ .Values.controller.containerPorts.controller }}
```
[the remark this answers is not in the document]
miguelaeh: I was not able to find the binary parameter or the configuration option in the configmap to configure the port, checking the source code it seems to be hardcoded: https://github.com/argoproj/argo-workflows/blob/b35aabe86be9fa5db80299cebcfb29c32be21047/cmd/workflow-controller/main.go#L116 while for the server, it is not hardcoded: https://github.com/argoproj/argo-workflows/blob/b35aabe86be9fa5db80299cebcfb29c32be21047/cmd/argo/commands/server.go#L176

bitnami/argo-workflows/values.yaml
miguelaeh: added to the CLI

bitnami/argo-workflows/values.yaml
javsalgar: This is

bitnami/argo-workflows/values.yaml
```diff
@@ -0,0 +1,1073 @@
...
+  containerSecurityContext:
+    enabled: true
+    runAsUser: 1001
+
+  ## Server container port
+  ## @param controller.containerPorts.controller Argo Workflows controller container port
+  ## @param controller.containerPorts.metrics Port to expose controller metrics
+  ## @param controller.containerPorts.telemetry Port to expose controller telemetry
+  containerPorts:
+    controller: 6060
```
javsalgar, line 514: I would not make this value configurable if it is hardcoded. Add a reference in the YAML with the link to the code.

bitnami/argo-workflows/values.yaml
```diff
@@ -0,0 +1,1073 @@
...
+    enabled: true
+    fsGroup: 1001
+  ## Configure Container Security Context
+  ## ref: https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-pod
+  ## @param controller.containerSecurityContext.enabled Enabled controller containers' Security Context
+  ## @param controller.containerSecurityContext.runAsUser Set controller containers' Security Context runAsUser
+  ##
+  containerSecurityContext:
+    enabled: true
+    runAsUser: 1001
```
javsalgar, line 507: Missing the value `runAsNonRoot`

bitnami/argo-workflows/values.yaml
```diff
@@ -0,0 +1,1073 @@
...
+  ## Configure extra options for controller containers' liveness and readiness probes
+  ## ref: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/#configure-probes
+  ## @param controller.livenessProbe.enabled Enable livenessProbe on controller nodes
+  ## @param controller.livenessProbe.initialDelaySeconds Initial delay seconds for livenessProbe
+  ## @param controller.livenessProbe.periodSeconds Period seconds for livenessProbe
+  ## @param controller.livenessProbe.timeoutSeconds Timeout seconds for livenessProbe
+  ## @param controller.livenessProbe.failureThreshold Failure threshold for livenessProbe
+  ## @param controller.livenessProbe.successThreshold Success threshold for livenessProbe
+  ##
+  livenessProbe:
```
javsalgar, line 457: Missing the startupProbe. Disable it by default

javsalgar: Same in the other component

bitnami/argo-workflows/values.yaml
```diff
@@ -0,0 +1,1073 @@
...
+## @param externalDatabase.type Either postgresql or mysql
+##
+externalDatabase:
+  enabled: false
+  ## Either "postgres" or  "mysql"
+  ##
+  type: ""
+  host: localhost
+  port: 3306
+  user: bn_wordpress
```
javsalgar, line 1067: Careful, you are using externalDatabase.username in the template

bitnami/argo-workflows/values.yaml
```diff
@@ -0,0 +1,1073 @@
...
+  ##
+  args: []
+  ## @param controller.hostAliases controller pods host aliases
+  ## https://kubernetes.io/docs/concepts/services-networking/add-entries-to-pod-etc-hosts-with-host-aliases/
+  ##
+  hostAliases: []
+  ## @param controller.podLabels Extra labels for controller pods
+  ## ref: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/
+  ##
+  podLabels: {}
```
javsalgar, line 710: Missing other standard values

controller.schedulerName
controller.topologySpreadConstraints

bitnami/argo-workflows/templates/server/deployment.yaml
```diff
@@ -0,0 +1,174 @@
...
+  strategy: {{- toYaml .Values.server.updateStrategy | nindent 4 }}
+  {{- end }}
+  selector:
+    matchLabels: {{- include "common.labels.matchLabels" . | nindent 6 }}
+      app.kubernetes.io/component: server
+  template:
+    metadata:
+      annotations:
+        checksum/config: {{ include (print $.Template.BasePath "/configmap.yaml") . | sha256sum }}
+        {{- if .Values.server.podAnnotations }}
```
javsalgar, line 28: Missing commonLabels and commonAnnotations here

bitnami/argo-workflows/templates/server/service-account.yaml
```diff
@@ -0,0 +1,16 @@
+{{- if .Values.server.serviceAccount.create }}
```
javsalgar, line 1: Move this service account one level below because it is used by both

bitnami/argo-workflows/templates/_helpers.tpl
```diff
@@ -0,0 +1,243 @@
...
+*/}}
+{{- define "argo-workflows.controller.database.username" -}}
+{{- if .Values.postgresql.enabled -}}
+{{- .Values.postgresql.postgresqlUsername -}}
+{{- end -}}
+{{- if .Values.mysql.enabled -}}
+{{- .Values.mysql.auth.username -}}
+{{- end -}}
+{{- if .Values.externalDatabase.enabled -}}
+{{- .Values.externalDatabase.username -}}
```
javsalgar, line 125: Careful, in the values you are calling it user

bitnami/argo-workflows/templates/_helpers.tpl
```diff
@@ -0,0 +1,243 @@
...
+{{- define "argo-workflows.controller.database.password.secret" -}}
+{{- if .Values.postgresql.enabled -}}
+{{- include "argo-workflows.postgresql.fullname" . -}}
+{{- end -}}
+{{- if .Values.mysql.enabled -}}
+{{- include "argo-workflows.mysql.fullname" . -}}
+{{- end -}}
+{{- if .Values.externalDatabase.enabled -}}
+{{- if .Values.externalDatabase.existingSecret -}}
+{{- .Values.externalDatabase.existingSecret -}}
```
javsalgar, line 148: Allow rendering for this existingSecret, users may want to use an umbrella chart where they create the secret and then use Argo as a subchart (this practice is done by charts like Gitlab)

bitnami/argo-workflows/templates/_helpers.tpl
```diff
@@ -0,0 +1,243 @@
...
+*/}}
+{{- define "argo-workflows.controller.database.password.secret.key" -}}
+{{- if .Values.postgresql.enabled -}}
+{{- printf "%s" "postgresql-password" -}}
+{{- end -}}
+{{- if .Values.mysql.enabled -}}
+{{- printf "%s" "mysql-password" -}}
+{{- end -}}
+{{- if .Values.externalDatabase.enabled -}}
+{{- printf "%s" "database-password" -}}
```
javsalgar, line 166: In the case of externalDatabase, there are users that have requested a lot to allow a custom key. Could you add a value for that?

bitnami/argo-workflows/templates/controller/aggregate-cluster-roles.yaml
```diff
@@ -0,0 +1,113 @@
+{{- if and .Values.createAggregateRoles .Values.controller.rbac.create }}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: {{ include "common.names.fullname" . }}-view
+  namespace: {{ .Release.Namespace | quote }}
+  labels: {{- include "common.labels.standard" . | nindent 4 }}
+    rbac.authorization.k8s.io/aggregate-to-view: "true"
```
javsalgar, line 8: Missing these labels here

```
    app.kubernetes.io/part-of: argo-workflows
    app.kubernetes.io/component: controller
```

bitnami/argo-workflows/templates/controller/database-secret.yaml
```diff
@@ -0,0 +1,21 @@
+{{- if (include "argo-workflows.controller.persistence.enabled" .) }}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ include "argo-workflows.controller.database.username.secret" . }}
+  namespace: {{ .Release.Namespace | quote }}
+  labels: {{- include "common.labels.standard" . | nindent 4 }}
+    app.kubernetes.io/component: controller
```
javsalgar, line 8: Missing the 

```
    app.kubernetes.io/part-of: argo-workflows
```

label

javsalgar: Please check the labels in all components

bitnami/argo-workflows/values.yaml
miguelaeh: ups, a copy paste error

bitnami/argo-workflows/templates/_helpers.tpl
miguelaeh: Have in mind that for an external database we will create the secret, this is not an existing secret key, but the key that is used in the secret we create with the externalDatabase password

bitnami/argo-workflows/templates/controller/deployment.yaml
```diff
@@ -0,0 +1,199 @@
...
+        podAntiAffinity: {{- include "common.affinities.pods" (dict "type" .Values.controller.podAntiAffinityPreset "component" "controller" "context" $) | nindent 10 }}
+        nodeAffinity: {{- include "common.affinities.nodes" (dict "type" .Values.controller.nodeAffinityPreset.type "key" .Values.controller.nodeAffinityPreset.key "values" .Values.controller.nodeAffinityPreset.values) | nindent 10 }}
+      {{- end }}
+      {{- if .Values.controller.nodeSelector }}
+      nodeSelector: {{- include "common.tplvalues.render" ( dict "value" .Values.controller.nodeSelector "context" $) | nindent 8 }}
+      {{- end }}
+      {{- if .Values.controller.tolerations }}
+      tolerations: {{- include "common.tplvalues.render" (dict "value" .Values.controller.tolerations "context" .) | nindent 8 }}
+      {{- end }}
+      {{- if .Values.controller.spreadConstraints }}
```
javsalgar, line 64: topologySpreadConstraints

bitnami/argo-workflows/values.yaml
```diff
@@ -0,0 +1,1138 @@
...
+    enabled: true
+    fsGroup: 1001
+  ## Configure Container Security Context
+  ## ref: https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-pod
+  ## @param server.containerSecurityContext.enabled Enabled server containers' Security Context
+  ## @param server.containerSecurityContext.runAsUser Set server containers' Security Context runAsUser
+  ##
+  containerSecurityContext:
+    enabled: true
+    runAsUser: 1001
```
javsalgar, line 164: runAsNonRoot missing here

javsalgar (approved):

PR #7563, opened 2021-09-21
[bitnami/mariadb] Chart standarized

migruiz4:
**Description of the change**

This PR standardizes the bitnami/mariadb chart to be in line with the rest of the catalog.

**Checklist**


bitnami/mariadb/templates/primary/svc.yaml
```diff
@@ -19,24 +19,32 @@ metadata:
...
+  {{- if .Values.primary.service.sessionAffinity }}
+  sessionAffinity: {{ .Values.primary.service.sessionAffinity }}
+  {{- end }}
+  {{- if .Values.primary.service.sessionAffinityConfig }}
+  sessionAffinityConfig: {{- include "common.tplvalues.render" (dict "value" .Values.primary.service.sessionAffinityConfig "context" $) | nindent 4 }}
   {{- end }}
   ports:
     - name: mysql
-      port: {{ .Values.primary.service.port }}
+      port: {{ .Values.primary.service.ports.mysql }}
```
juan131, line 43: Why this standardisation? This is the UNIQUE port we're exposing under the `primary.service` object. Therefore, there's no need for a "ports" object.

bitnami/mariadb/templates/primary/svc.yaml
```diff
@@ -19,24 +19,32 @@ metadata:
...
+  {{- if and .Values.primary.service.clusterIP (eq .Values.primary.service.type "ClusterIP") }}
   clusterIP: {{ .Values.primary.service.clusterIP }}
   {{- end }}
-  {{- if and .Values.primary.service.loadBalancerIP (eq .Values.primary.service.type "LoadBalancer") }}
+  {{- if (or (eq .Values.primary.service.type "LoadBalancer") (eq .Values.primary.service.type "NodePort")) }}
+  externalTrafficPolicy: {{ .Values.primary.service.externalTrafficPolicy | quote }}
+  {{- end }}
+  {{ if eq .Values.primary.service.type "LoadBalancer" }}
+  loadBalancerSourceRanges: {{ .Values.primary.service.loadBalancerSourceRanges }}
+  {{ end }}
```
juan131, line 30: This change is not correct. What we had previously was more accurate:
```suggestion
  {{- if and (eq .Values.primary.service.type "LoadBalancer") .Values.primary.service.loadBalancerSourceRanges }}
  loadBalancerSourceRanges: {{ .Values.primary.service.loadBalancerSourceRanges }}
  {{ end }}
```

bitnami/mariadb/templates/NOTES.txt
```diff
@@ -26,9 +26,9 @@ Tip:
 
 Services:
 
-  echo Primary: {{ include "mariadb.primary.fullname" . }}.{{ .Release.Namespace }}.svc.{{ .Values.clusterDomain }}:{{ .Values.primary.service.port }}
+  echo Primary: {{ include "mariadb.primary.fullname" . }}.{{ .Release.Namespace }}.svc.{{ .Values.clusterDomain }}:{{ .Values.primary.service.ports.mysql }}
 {{- if eq .Values.architecture "replication" }}
-  echo Secondary: {{ include "mariadb.secondary.fullname" . }}.{{ .Release.Namespace }}.svc.{{ .Values.clusterDomain }}:{{ .Values.secondary.service.port }}
+  echo Secondary: {{ include "mariadb.secondary.fullname" . }}.{{ .Release.Namespace }}.svc.{{ .Values.clusterDomain }}:{{ .Values.secondary.service.ports.mysql }}
```
juan131, line 31: I'm strongly against this change. It breaks backwards compatibility and there's no need for it all.

bitnami/mariadb/templates/secondary/statefulset.yaml
```diff
@@ -127,7 +129,7 @@ spec:
             - name: MARIADB_MASTER_HOST
               value: {{ include "mariadb.primary.fullname" . }}
             - name: MARIADB_MASTER_PORT_NUMBER
-              value: {{ .Values.primary.service.port | quote }}
+              value: {{ .Values.primary.service.ports.mysql | quote }}
```
juan131, line 132: Same comment here

bitnami/mariadb/templates/secondary/svc.yaml
```diff
@@ -21,23 +21,32 @@ metadata:
...
   ports:
     - name: mysql
-      port: {{ .Values.secondary.service.port }}
+      port: {{ .Values.secondary.service.ports.mysql }}
       protocol: TCP
       targetPort: mysql
-      {{- if (and (or (eq .Values.secondary.service.type "NodePort") (eq .Values.secondary.service.type "LoadBalancer")) .Values.secondary.service.nodePort) }}
-      nodePort: {{ .Values.secondary.service.nodePort }}
+      {{- if (and (or (eq .Values.secondary.service.type "NodePort") (eq .Values.secondary.service.type "LoadBalancer")) .Values.secondary.service.nodePorts.mysql) }}
+      nodePort: {{ .Values.secondary.service.nodePorts.mysql }}
```
juan131, line 49: Same here.

juan131 (changes requested):

migruiz4:
Made the `nodePort` and `service.port` changes backward-compatible

bitnami/mariadb/templates/primary/svc.yaml
migruiz4: I'm sorry, but I copied this exactly as in the Template chart:
https://github.com/bitnami/charts/blob/e34ee5c5aeafabe15cbf0d2b0fecb18483bc6415/template/CHART_NAME/templates/service.yaml#L19-L21

Should I apply the suggested change anyway? We are configuring `loadBalancerSourceRanges` as in this PR in many other charts.

juan131: Then the template is wrong @migruiz4 😄 :
```console
$ k explain service.spec.loadBalancerSourceRanges
...
If specified and supported by the platform, this will restrict traffic
through the cloud-provider load-balancer will be restricted to the
specified client IPs
```
It doesn't make sense to indicate this setting UNLESS the service type is actually "LoadBalancer"

bitnami/mariadb/templates/secondary/svc.yaml
```diff
@@ -21,23 +21,32 @@ metadata:
...
+  {{- if and .Values.secondary.service.clusterIP (eq .Values.secondary.service.type "ClusterIP") }}
   clusterIP: {{ .Values.secondary.service.clusterIP }}
   {{- end }}
-  {{- if and .Values.secondary.service.loadBalancerIP (eq .Values.secondary.service.type "LoadBalancer") }}
+  {{- if (or (eq .Values.secondary.service.type "LoadBalancer") (eq .Values.secondary.service.type "NodePort")) }}
+  externalTrafficPolicy: {{ .Values.secondary.service.externalTrafficPolicy | quote }}
+  {{- end }}
+  {{ if eq .Values.secondary.service.type "LoadBalancer" }}
+  loadBalancerSourceRanges: {{ .Values.secondary.service.loadBalancerSourceRanges }}
+  {{ end }}
```
juan131, line 32: Same comment here

bitnami/mariadb/templates/serviceaccount.yaml
```diff
@@ -15,4 +15,5 @@ metadata:
     {{- if .Values.commonAnnotations }}
     {{- include "common.tplvalues.render" ( dict "value" .Values.commonAnnotations "context" $ ) | nindent 4 }}
     {{- end }}
+automountServiceAccountToken: {{ .Values.serviceAccount.automountServiceAccountToken }}
```
juan131, line 18: AFAIK, none of the containers in this chart access the API server. Then, I don't see why it's necessary to expose this parameter, it's useless.

juan131 (changes requested):
Thanks so much for adding backwards compatibility with "coalesce" for the new ports approach. I guess that we can add also this to `pdb.create`/`pdb.enabled` and this way there's no need to force a major, right @migruiz4 ?

bitnami/mariadb/templates/serviceaccount.yaml
juan131: Forget about this comment @migruiz4 ! I was talking about this point with @javsalgar and it could be useful if users add custom initcontainer/sidecars that talk with the K8s API.
That said, you're currently setting it to `true` by default, please do the opposite and set it to `false`. I mean, there's no reason to mount the token in the container UNLESS you need access the to the API

juan131 (approved):
LGTM! Awesome work @migruiz4

Issue #7541, opened 2021-09-20
[bitnami/spring-cloud-data-flow] - Provide values configuration to skip database tables creation on startup


AbhijitBhattacharya:
**Which chart**:
bitnami/spring-cloud-data-flow

**Is your feature request related to a problem? Please describe.**
When we have external Database already having dataflow/skipper schema and tables created, then while re-installing spring-cloud-dataflow again in local minikube cluster using externalDatabase, values configuration is required to skip Dataflow/Skipper Database Table creation on startup.

**Describe the solution you'd like**
Following Spring property can be used in ConfigMap of server for dataflow & skipper, which skips database tables creation
`spring.flyway.enabled=false`

**Describe alternatives you've considered**
I had to update the above spring property directly in ConfigMap yaml file for dataflow, after install.

**Additional context**

marcosbc:
Hi @AbhijitBhattacharya, this seems like a good improvement. Would you be up for contributing the change via a PR? You can find the sources of the chart in the [`templates`](https://github.com/bitnami/charts/tree/master/bitnami/spring-cloud-dataflow/templates) folder in the GitHub repo. We'd be glad to help getting your changes released into the chart.

AbhijitBhattacharya:
> Hi @AbhijitBhattacharya, this seems like a good improvement. Would you be up for contributing the change via a PR? You can find the sources of the chart in the [`templates`](https://github.com/bitnami/charts/tree/master/bitnami/spring-cloud-dataflow/templates) folder in the GitHub repo. We'd be glad to help getting your changes released into the chart.

Submitted PR: - https://github.com/bitnami/charts/pull/7570

AbhijitBhattacharya:
PR has been merged. Closing the issue now.

PR #7570, opened 2021-09-21
[bitnami/spring-cloud-dataflow] - 7541 Add spring.flyway.enabled configuration

AbhijitBhattacharya:
**Description of the change**

Adding Values property in ConfigMap of server for dataflow & skipper, which can skip running database tables creation scripts on startup.
`spring.flyway.enabled=false`

**Benefits**

We can now startup dataflow and skipper deployments on Kubernetes cluster using helm, where they are connecting to Database connection already having the dataflow and skipper related metadata tables. 

**Possible drawbacks**

**Applicable issues**

  - fixes #https://github.com/bitnami/charts/issues/7541

**Additional information**

Verified settings work on local minikube cluster while using external Postgres database where Dataflow and Skipper schema and tables are already present.

**Checklist** 


bitnami/spring-cloud-dataflow/templates/skipper/configmap.yaml
```diff
@@ -96,4 +96,6 @@ data:
         {{- end }}
         testOnBorrow: true
         validationQuery: "SELECT 1"
-{{ end }}
+      flyway:
+        enabled: {{ .Values.flyway.enabled }}
+  {{ end }}
```
carrodher, line 101: The indentation of the `{{ end }}` statement was modified
```suggestion
      flyway:
        enabled: {{ .Values.flyway.enabled }}
{{ end }}
```

bitnami/spring-cloud-dataflow/values.yaml
```diff
@@ -1092,6 +1092,14 @@ mariadb:
...
       GRANT ALL ON skipper.* to 'skipper'@'%';
       FLUSH PRIVILEGES;
+## Flyway Configuration
+## All database creation scripts are ignored on startup when flyway.enabled is set to false
+## This feature can be used in scenario, where Database tables are already present in Mariadb or ExternalDatabase.
+##
+flyway:
+  ## @param flyway.enabled Enable/disable flyway
+  ##
+  enabled: true
```
carrodher, line 1102:
```suggestion
## Flyway Configuration
## @param flyway.enabled Enable/disable flyway running Dataflow and Skipper Database creation scripts on startup
## All database creation scripts are ignored on startup when flyway.enabled is set to false
## This feature can be used in scenario, where Database tables are already present in Mariadb or ExternalDatabase.
##
flyway:
  enabled: true
```

Please note the table in the README is generated by using this metadata. In my suggestion, I added the sentence you used in the README with the parameter metadata

bitnami/spring-cloud-dataflow/README.md
```diff
@@ -322,6 +322,7 @@ helm uninstall my-release
 | `mariadb.auth.forcePassword`              | Force users to specify required passwords in the database                                           | `false`      |
 | `mariadb.auth.usePasswordFiles`           | Mount credentials as a file instead of using an environment variable                                | `false`      |
 | `mariadb.initdbScripts`                   | Specify dictionary of scripts to be run at first boot                                               | `{}`         |
+| `flyway.enabled`                          | Enable/disable flyway running Dataflow and Skipper Database creation scripts on startup             | `true`       |
```
carrodher, line 325: If I'm not wrong, the default behavior (the one used now), is the equivalent to having this parameter set to `false`. I think it's safer to use `false` as the default value and allow users to modify it when it's needed.

bitnami/spring-cloud-dataflow/Chart.yaml
```diff
@@ -39,4 +39,4 @@ sources:
   - https://github.com/bitnami/bitnami-docker-spring-cloud-dataflow
   - https://github.com/bitnami/bitnami-docker-spring-cloud-skipper
   - https://dataflow.spring.io/
-version: 4.0.3
+version: 4.0.5
```
carrodher, line 42: Since it's a feature addition instead of bug fixing, I would increase the minor version instead of the patch one
```suggestion
version: 4.1.0
```

carrodher (commented):
Thanks for your suggestion, please take a look at my comments

bitnami/spring-cloud-dataflow/README.md
AbhijitBhattacharya: I am thinking with respect of first time users using helm install Spring-Cloud-Dataflow, who have not overridden values.yaml.
I think it is better to create schema on MariaDB StatefulSet, when server starts up.
Only when user starts using pre-created database, flyway.enabled can then be set as `false` in values override yaml file.

carrodher: Thanks, I will check the upgrade from previous chart versions (`4.0.X` -> `4.1.0`); if everything works fine, the PR will be merged.

PR #7582, opened 2021-09-22

pablogalegoc:
**Description of the change**

**Benefits**

**Possible drawbacks**

**Checklist** 


bitnami/airflow/templates/_helpers.tpl
```diff
@@ -460,3 +460,14 @@ Ref: https://github.com/bitnami/charts/pull/6096#issuecomment-856499047
...
+{{/*
+Returns whether the deprecated ingress.*.certManager boolean parameter is used or not
+This helper is needed in order to guarantee backwards-compatibility
+Ref: https://cert-manager.io/docs/usage/ingress/#supported-annotations
+*/}}
+{{- define "airflow.ingress.certManager.isLegacy" -}}
+{{- if typeIs "bool" .Values.ingress.certManager }}
+  true
+{{- else -}}{{- end -}}
+{{- end -}}
```
carrodher, line 473:
```suggestion
{{- if typeIs "bool" .Values.ingress.certManager }}
    {{- true -}}
{{- end -}}
{{- end -}}
```

carrodher (commented):
Minor suggestion. I think this is a good solution in order to make the change backward compatible. As discussed, adding it to the common can be the proper solution to reuse the code, I'm sure we can find a "category" to add this template or just create a miscellany one

pablogalegoc:
Per @juan131 suggestion I'm going to explore removing the parameter and relying on the default `ingress.annotations` for this. Pros: We don't add new logic to handle backwards compatibility, we treat the `cert-manager` annotations as any other ingress annotation.

pablogalegoc:
@carrodher done! If you think this is also a valid approach I'll just use this same PR to update the other instances of `ingress.certManager`

carrodher:
For me it's fine! Just a minor comment, I am not sure if the README was generated by using the `readme-generator-tool` since it seems the table is not properly aligned (in terms of markdown, in the UI it looks fine).

Using the automated tool it should be aligned and you can see if the metadata in the _values.yaml_ was properly set

pablogalegoc:
Yep! It was modified with `readme-generator`. I think it's the UI in Github's review section that is messing something, in [raw format](https://raw.githubusercontent.com/bitnami/charts/003c086225de52bb46c2b53eb05cb6c9cb585757/bitnami/airflow/README.md) you can see it is aligned

pablogalegoc:
@carrodher this is a huge PR but basically is just:
- Deprecate `ingress.certManager` by commenting it
- Add a reference to the cert-manager annotations in the `ingress.annotations` documentation
- Add a mention to cert-manager in the `ingress.annotations` metadata so it appears in the README

juan131 (changes requested):
I'm okay with the changes... That said, these 3 guides from Chart docs should be also updated @pablogalegoc 

- https://github.com/bitnami/charts-docs/blob/main/charts/cert-manager/_secure_ingress_resources.md.erb
- https://github.com/bitnami/charts-docs/blob/main/common/_enable_tls_ingress.md.erb
- https://github.com/bitnami/charts-docs/blob/main/charts/kubeapps/_expose_service.md.erb

I'll block this PR until there's another PR updating the chart docs.

---

Also, I remember from a private conversation that you were planning to create a "named template" to manage an existing condition in the ingress rule related to this `ingress.certManager`  parameter.. This was supposed to be done in the `_helpers.tpl` for Ghost, Kubeapps, Odoo, RabbitMQ and Redmine. Is this happening in other  PR?

pablogalegoc:
Thanks @juan131! Yes, this is just the first batch of the easier ones. On the JIRA issue I'm keeping track of the progress. I'll open another PR similar to this for the remaining ones and another for Ghost, Kubeapps, Odoo, RabbitMQ and Redmine. This PR is already huge, I didn't want to make it worse 😅

pablogalegoc:
The `readme-generator` CI check is failing for unrelated reasons, addressed at https://github.com/bitnami/charts/pull/7649

juan131 (approved):
LGTM

carrodher (approved):
LGTM

AntonOfTheWoods:
It looks like I am not getting a certificate request after the update. Should the following be generating a `certificaterequest`? 

```
helm get values harbor -n harbor --all 
...
ingress:
  annotations:
    cert-manager.io/cluster-issuer: letsencrypt-prod
    ingress.kubernetes.io/proxy-body-size: "0"
    ingress.kubernetes.io/ssl-redirect: "true"
    nginx.ingress.kubernetes.io/proxy-body-size: "0"
    nginx.ingress.kubernetes.io/ssl-redirect: "true"
  apiVersion: ""
  controller: default
  enabled: true
  hosts:
    core: registry.my.doma.in
    notary: notary.harbor.domain
...
```

If not, can someone point me to some docs that might help out (the `readme` is not explicit enough for me unfortunately)? `cert-manager.io/cluster-issuer: letsencrypt-prod` seems to work in other charts' ingress so I'm a little unclear as to the issue here. Thanks.

Issue #7460, opened 2021-09-10
metrics-server feature request: make caBundle and insecureSkipTLSVerify parameterized in the chart


techstep:
**Which chart**:
metrics-server 5.9.3

**Is your feature request related to a problem? Please describe.**
As a user of metrics-server, I want to be able to properly use certificates between metrics-server and the API service to be checked. In the `metrics-api-service.yaml` template, `caBundle` is not set, and `insecureSkipTLSVerify` is hardcoded to `true`. Hence, no matter if the certificate is valid, it will be ignored. Since we are running this on our own environment where we can generate certificates, this would be good to set to `false`.

Along those lines, a parameter that allows us to set the `caBundle` would be great, so that we can pass in the entire chain.

**Describe the solution you'd like**
Within the `metrics-api-service.yaml` template, I would like to see (for example) `{{ .Values.apiService.insecureSkipTLSVerify | default 'true' }}`, and `{{ .Values.apiService.caBundle | default '' }}` or something similar. Parameters that map have the default behavior as things stand now, but offer the flexibility of change.

**Describe alternatives you've considered**
We could maintain a separate fork of the charts, but that would be difficult to the extent that we use a number of other charts unmodified. We have also looked at continuing to ignore certificates, but we would like to avoid doing that and take a more secure stance.

migruiz4:
Hi @techstep,

Thanks for your suggestions. It would be a great addition to the `metrics-server` chart. Would you like to contribute with a PR for adding those parameters?

We will be very happy to review it and help with any questions you may have.

I would also like to recommend you to have a look at our contribution guidelines [here](https://github.com/bitnami/charts/blob/master/CONTRIBUTING.md)

techstep:
Thank you for responding!

I'll make a PR soon, as things allow. And thanks for pointing me to the guidelines!

migruiz4:
Thank you for the PR @techstep!

As PR #7590 was merged, I will close this issue.

PR #7590, opened 2021-09-22
[bitnami/metrics-server]: add insecureSkipTLSVerify and caBundle

techstep:
**Description of the change**

This PR adds support for `insecureSkipTLSVerify` and `caBundle` in the metrics-server chart. 

**Benefits**

This will allow users to be able to add their own certificates between metrics-server and APIService. Previously, `insecureSkipTLSVerify` was hardcoded as `true`, even though there are situations in which it makes sense to require verification.

**Possible drawbacks**

None known. The defaults remain the same as before, so existing templates should require no change.

**Applicable issues**

- Issue #7460 

**Additional information**

Because the default for `insecureSkipTLSVerify` was `true`, in the `values.yaml` file, `false` needs to be quoted (i.e., `"false"`), or else it will render as `true`. (It is a long-standing issue: https://github.com/helm/helm/issues/3308)

**Checklist** 


bitnami/metrics-server/README.md
```diff
@@ -155,10 +157,18 @@ It is strongly recommended to use immutable tags in a production environment. Th
 
 Bitnami will release a new chart updating its containers if a new version of the main container, significant changes, or critical vulnerabilities exist.
 
+<<<<<<< Updated upstream
```
alemorcuq, line 160: Delete this.

bitnami/metrics-server/README.md
```diff
@@ -155,10 +157,18 @@ It is strongly recommended to use immutable tags in a production environment. Th
...
 ### Enable RBAC security
+### Enable security for Metrics Server.
+
+#### Configuring RBAC
```
alemorcuq, line 164: I would remove this to leave there only the instructions for RBAC.

bitnami/metrics-server/README.md
```diff
@@ -155,10 +157,18 @@ It is strongly recommended to use immutable tags in a production environment. Th
...
 
 In order to enable Role-Based Access Control (RBAC) for Metrics Server, use the following parameter: `rbac.create=true`.
 
+#### Configuring certificates
```
alemorcuq, line 168: And change this to:

```suggestion
### Configure certificates
```

alemorcuq (changes requested):
Thanks for the PR! Please, see my suggestions below.

alemorcuq (approved):

PR #7610, opened 2021-09-23
[bitnami/postgresql] Make the provisioning of headless service optional

nachin:
**Description of the change**
Adding a flag in the service section of the values file to allow users to disable the creation of the headless service.  Disabling the creation of the headless service will cause the statefulsets to use the regular service that is provisioned by the chart.

The default value for the flag is set to `true` so this will be backwards compatible.

**Benefits**
 For our use case where this is mainly being provisioned in sandbox environments for testing we have noticed a negative impact on the performance of istio when a lot of headless services are being created  [more info here](https://github.com/istio/istio/issues/33969).

**Checklist** 


juan131:
Hi @nachin 

I'd like to have more context about why you need these changes.

Using headless services is the unique way to configure A records in the DNS that point directly to the POD IP(s). I mean, it's the only way to guarantee you have a hostname associated to each PostgreSQL primary or read replicas node.

See:  https://kubernetes.io/docs/concepts/services-networking/service/#headless-services

Otherwise, how can you guarantee that you can address a specific node (without using its POD IP)?

---

Disclaimer:

> Bitnami charts are not designed to work with Istio. That said, they can be integrated with Istio since they can be extended adding custom sidecar containers, labels, annotations, etc.
> 
> I want to be clear on this because we don't test the integration with Istio in our pipeline and, therefore, we don't officially support it. As a consequence, we cannot guarantee that we will not break the compatibility with Istio in the future since we're not taking into account during the development process any possible consequence affecting this integration.

nachin:
HI @juan131 
We have a staging cluster that also hosts dev and test deployments of our apps that use Bitnami charts in place of AWS manage datastores so their data layer is as isolated as possible and easy to create/destroy. There can be many of these deployments at a time, and the result using Bitnami charts is hundreds of headless services. After having a variety of scaling, resource and network issues we realized that Istio is not capable of handling this number of headless services. This also isn't  because we try to use Istio to configure virtualservices or anything for the Bitnami datastores, but simply being on the cluster causes Istio to consider them part of the mesh and there is not a way for us to ignore the headless services.

> Hi @nachin
> 
> I'd like to have more context about why you need these changes.
> 
> Using headless services is the unique way to configure A records in the DNS that point directly to the POD IP(s). I mean, it's the only way to guarantee you have a hostname associated to each PostgreSQL primary or read replicas node.
> 
> See: https://kubernetes.io/docs/concepts/services-networking/service/#headless-services
> 
> Otherwise, how can you guarantee that you can address a specific node (without using its POD IP)?
> 
> Disclaimer:
> 
> > Bitnami charts are not designed to work with Istio. That said, they can be integrated with Istio since they can be extended adding custom sidecar containers, labels, annotations, etc.
> > I want to be clear on this because we don't test the integration with Istio in our pipeline and, therefore, we don't officially support it. As a consequence, we cannot guarantee that we will not break the compatibility with Istio in the future since we're not taking into account during the development process any possible consequence affecting this integration.

HI @juan131 
We have a staging cluster that also hosts dev and test deployments of our apps that use Bitnami charts in place of AWS manage datastores so their data layer is as isolated as possible and easy to create/destroy. There can be many of these deployments at a time, and the result using Bitnami charts is hundreds of headless services. After having a variety of scaling, resource and network issues we realized that Istio is not capable of handling this number of headless services. This also isn't  because we try to use Istio to configure virtualservices or anything for the Bitnami datastores, but simply being on the cluster causes Istio to consider them part of the mesh and there is not a way for us to ignore the headless services.

juan131:
Hi @nachin 

Thanks for providing the whole context!

I have my concerns of accepting this approach if the unique use case of disabling headless services is making the chart compatible with Istio. These are the two reasons:

1. Bitnami charts do not officially support Istio (for the reasons I mentioned in my previous response).
2. This is a limitation on the Istio side: Headless services are a K8s native object.

By the way, in the blog below, a workaround is suggested to make Istio compatible with headless services (although it's note very scalable):

- https://medium.com/airy-science/making-istio-work-with-kubernetes-statefulset-and-headless-services-d5725c8efcc9

PR #7627, opened 2021-09-27
bitnami/mongodb - Add extra dns names for auto tls.

Hkyx:
## Resume:
 - Add extra dns name option, solve a issue to use x509 auth certificate when auto generate the certificate

**Description of the change**

 - Added conditional for tls.extraDnsName. If exist iterate over a for each to add the differents dns names.

**Benefits**

 - solve a issue to use x509 auth certificate when auto generate the certificate. The differents clients get theyr certificate from the CA but when it's not a refered namespace, svc or others specific dns names to use, the pod can't identify itself with the provided certificate.

**Possible drawbacks**

not really a limitation but we don't control what strange values a user can put inside that.

**Applicable issues**

N/A

**Checklist** 


bitnami/mongodb/Chart.yaml
```diff
@@ -26,4 +26,4 @@ name: mongodb
 sources:
   - https://github.com/bitnami/bitnami-docker-mongodb
   - https://mongodb.org
-version: 10.26.4
+version: 10.26.5
```
Mauraza, line 29: Is a new feature so I think is better:
```suggestion
version: 10.27.0
```

bitnami/mongodb/values.yaml
```diff
@@ -179,6 +179,9 @@ tls:
   ## @param tls.image.tag Init container TLS certs setup image tag (immutable tags are recommended)
   ## @param tls.image.pullPolicy Init container TLS certs setup image pull policy
   ##
+  extraDnsNames: []
+  ## Add extra dns names to the CA, can solve x509 auth issue for pod clients
+  ##
```
Mauraza, line 184: It's necessary to add the metadata to generate de tables of the README.md

```suggestion
  extraDnsNames: []
  ## @param tls.extraDnsNames Add extra dns names to the CA, can solve x509 auth issue for pod clients
  ##
```

Mauraza (changes requested):
Hi @Hkyx,

Thanks for the PR! Could you check the comments?

bitnami/mongodb/values.yaml
Hkyx: Ho ok. never used that before :)

Hkyx:
@Mauraza   thanks for the corrections, PR updated

bitnami/mongodb/templates/standalone/dep-sts.yaml
```diff
@@ -158,6 +158,11 @@ spec:
               DNS.3 = $my_hostname.$svc.$MY_POD_NAMESPACE.svc.cluster.local
               DNS.4 = localhost
               DNS.5 = 127.0.0.1
+              {{- if .Values.tls.extraDnsNames }}
+              {{- range $index, $dnsName := .Values.extraDnsNames }}
+              - {{ $dnsName }}
+              {{- end }}
+              {{- end }}
```
Mauraza, line 165: Testing the solution I have realized that you missed the 'tls'  in 
`{{- range $index, $dnsName := .Values.extraDnsNames }}`
following `DNS.2 = $my_hostname` maybe this makes more sense, what do you think?
```suggestion
             {{- if .Values.tls.extraDnsNames }}
              {{- range $key, $dnsName := .Values.tls.extraDnsNames }}
              {{ $key }} = {{ $dnsName }}
              {{- end }}
              {{- end }}
```

The values would be something like this:
```yaml
  extraDnsNames: 
    "DNS.6": "$my_host"
    "DNS.7": "$test"
```

Hkyx: ho miss my mistake, didn't updated my code but you're right !

For 
> DNS.2 = $my_hostname maybe this makes more sense, what do you think?

I would say it's up to your preference. I prefer just sending a list, smaller to put in the values.yaml but I understand that for the debug it's may be more easy

Mauraza: If you check the actual appears something like this:
https://github.com/bitnami/charts/blob/2afda168725556251dc543ed81cee8b66b331d44/bitnami/mongodb/templates/standalone/dep-sts.yaml#L156-L160
So If you want to add more lines in `[alt_names]` I think that will be similar to the others.
Anyway, it would be nice to add an example in the values.yaml,

Hkyx: understood, didn't put the DNSx:  in my test and it was working but for uniformity I approved the commit

bitnami/mongodb/templates/standalone/dep-sts.yaml
```diff
@@ -158,6 +158,11 @@ spec:
               DNS.3 = $my_hostname.$svc.$MY_POD_NAMESPACE.svc.cluster.local
               DNS.4 = localhost
               DNS.5 = 127.0.0.1
+             {{- if .Values.tls.extraDnsNames }}
```
Mauraza, line 161: Sorry, could you fix the indentation?

bitnami/mongodb/values.yaml
```diff
@@ -179,6 +179,9 @@ tls:
   ## @param tls.image.tag Init container TLS certs setup image tag (immutable tags are recommended)
   ## @param tls.image.pullPolicy Init container TLS certs setup image pull policy
   ##
+  extraDnsNames: []
+  ## @param tls.extraDnsNames Add extra dns names to the CA, can solve x509 auth issue for pod clients
+  ##
```
Mauraza, line 184: Could you add an example to clarify how to use:
```yaml
  extraDnsNames: 
    "DNS.6": "$my_host"
    "DNS.7": "$test"
```

Mauraza (commented):
Hi @Hkyx,

A few small changes and it would be ready

bitnami/mongodb/values.yaml
```diff
@@ -179,6 +179,11 @@ tls:
   ## @param tls.image.tag Init container TLS certs setup image tag (immutable tags are recommended)
   ## @param tls.image.pullPolicy Init container TLS certs setup image pull policy
   ##
+  extraDnsNames: 
+    "DNS.6": "$my_host"
+    "DNS.7": "$test"
+  ## @param tls.extraDnsNames Add extra dns names to the CA, can solve x509 auth issue for pod clients
+  ##
```
Mauraza, line 186:
```suggestion
  ## @param tls.extraDnsNames Add extra dns names to the CA, can solve x509 auth issue for pod clients
  ##
  ## example:
  ##     "DNS.6": "$my_host"
  ##    "DNS.7": "$test"
```

Mauraza (approved):
Hi @Hkyx,

Thanks for the contribution!! LGTM!

PR #4642, opened 2020-12-08
[bitnami/mongodb] Register targetPod in global context

lerminou:
To allow accessing the current targetPod in the annotations

**Description of the change**
We register a new variable with the current index of pod for annotations

**Benefits**
servcie annotations can now retrieve the targetPod for example for a DNS
`external-dns.alpha.kubernetes.io/hostname: '{{ .targetPod }}.example.com'`

**Possible drawbacks**

**Applicable issues**

  - fixes #https://github.com/bitnami/charts/issues/4639

**Additional information**

**Checklist** 
- [X] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [X] Variables are documented in the README.md
- [X] Title of the PR starts with chart name (e.g. `[bitnami/chart]`)
- [X] If the chart contains a `values-production.yaml` apart from `values.yaml`, ensure that you implement the changes in both files

:warning: Keep in mind that if you want to make changes to the kubeapps chart, please implement them in the [kubeapps repository](https://github.com/kubeapps/kubeapps/tree/master/chart/kubeapps). This is only a synchronized mirror.

PR #7635, opened 2021-09-27
[bitnami/redis-cluster] possibility to use targetPod on labels and annotations at externalAccess

houstondapaz:
**Description of the change**

As it is possible to do at mongodb chart, the targetPod variable is usefull to create custom annotations and labels to specific pod, this simple change just add this possibility 

**Benefits**

Create labels and annotations to specific instance of redis

**Possible drawbacks**

I can't see any drawbacks

**Applicable issues**

An simple appplication is to use external-dns in to create an dns entry to specific instance, with annotation
```external-dns.alpha.kubernetes.io/hostname: "{{`{{ .targetPod }}`}}.example.com```

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/redis-cluster/Chart.yaml
+++ b/bitnami/redis-cluster/Chart.yaml
@@ -23,4 +23,4 @@ name: redis-cluster
 sources:
   - https://github.com/bitnami/bitnami-docker-redis
   - http://redis.io/
-version: 6.3.7
+version: 6.3.8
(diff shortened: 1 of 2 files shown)


pablogalegoc (approved):
Thanks @houstondapaz! For reference I'll drop here that this is the same change as in https://github.com/bitnami/charts/pull/4642

PR #7637, opened 2021-09-27
[bitnami/jupyterhub] Add Prometheus monitoring with serviceMonitors for hub and proxy components

captnbp:
**Description of the change**

Add serviceMonitors for hub and proxy components

**Benefits**

Enabled Prometheus scraping if the Prometheus is managed using Prometheus operator

**Possible drawbacks**

**Applicable issues**

**Additional information**

**Checklist** 

- [ x] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [ x] Variables are documented in the README.md
- [ x] Title of the PR starts with chart name (e.g. [bitnami/<name_of_the_chart>])


captnbp:
Hi @yilmi,

What I can see here is the deployement is quite long, and reaches the helm --wait timeout (5min). But the deployement is still successful even if we reach the helm timeout for everything to be ready.
On my test environment it takes about 4 minutes to finish de deployement (images are already pulled).
Most of the time is spent in the `check_postgresql_connection` function within the initContainer of the hub component. This function relies on libs froms the postgresql image.

I can see that all previous checks on this chart failed for the same reason as for this PR.

I think the best way to fix this would be to enhance libs from the postgresql image to make the check fail faster if postgresql is not yet ready, or to rewrite the script used in the wait-for-db initContainer. But this is not related to this PR.

yilmi:
Thanks for looking into it and apologies for the delay. I should be able to get back to you tomorrow with a review 👍

yilmi:
Hi @captnbp, apologies for the delay, I just reviewed and tested your changes. Before I merge, could you fix the linter issue and also, increase the minor version number instead of the patch one as you're adding a new feature.

Thanks again for your contribution, I'll be able to merge it as soon as you can get these two minor changes in! 👍

captnbp:
Hi @yilmi,
I finally understood the real issue : kind was not able to provide an external IP on the Load Balancer service. This is normal, there is no MetalLB to assign IPs to LB services.
On my test env I had MetalLB installed, that's why I didn't see the issue.

yilmi:
Hi @captnbp, thanks a mil for this contribution!

PR #7646, opened 2021-09-29

pablogalegoc:
**Description of the change**

**Benefits**

**Possible drawbacks**

**Additional info**

  - Relates to #7582
  - This PR won't be merged until there's another PR updating the chart docs (https://github.com/bitnami/charts/pull/7582#pullrequestreview-763076526).

**Checklist** 


carrodher:
Phabricator was deprecated, you can safely remove it from this PR

juan131 (approved):
LGTM

carrodher (approved):
LGTM

PR #7656, opened 2021-09-30

pablogalegoc:
**Description of the change**

In cases where `ingress.certManager` was used as part of a condition to trigger a behaviour it has been changed by a macro that looks for one of the two strictly required cert-manager annotations: `cert-manager.io/cluster-issuer` or `cert-manager.io/issuer`.

**Benefits**

**Possible drawbacks**

**Additional info**

  - Relates to #7582
  - This PR won't be merged until there's another PR updating the chart docs (https://github.com/bitnami/charts/pull/7582#pullrequestreview-763076526).

**Checklist** 


bitnami/ghost/values.yaml
```diff
@@ -400,9 +407,10 @@ ingress:
   ##   - Relay on Helm to create self-signed certificates by setting `ingress.selfSigned=true`
```
juan131, line 407: Check the previous line (406). My suggestion:
```diff
-  ##   - Relay on cert-manager to create it by setting `ingress.certManager=true`
+  ##   - Relay on cert-manager to create it by setting the corresponding annotations
```

bitnami/ghost/values.yaml
```diff
@@ -388,8 +388,15 @@ ingress:
...
-  ## NOTE: If `ingress.certManager=true`, annotation `kubernetes.io/tls-acme: "true"` will automatically be added
+  ## @param ingress.annotations Additional annotations for the Ingress resource. To enable certificate autogeneration, place here your cert-manager annotations.
+  ## For a full list of possible ingress annotations, please see
+  ## ref: https://github.com/kubernetes/ingress-nginx/blob/master/docs/user-guide/nginx-configuration/annotations.md
+  ## Use this parameter to set the required annotations for cert-manager, see
+  ## ref: https://cert-manager.io/docs/usage/ingress/#supported-annotations
+  ##
+  ## Examples:
+  ## kubernetes.io/ingress.class: nginx
+  ## cert-manager.io/cluster-issuer: cluster-issuer-name
```
juan131, line 399: I personally prefer the examples to include the parameter name:
```suggestion
  ## e.g:
  ## annotations:
  ##   kubernetes.io/ingress.class: nginx
  ##   cert-manager.io/cluster-issuer: cluster-issuer-name
```

bitnami/ghost/templates/_helpers.tpl
```diff
@@ -146,3 +146,13 @@ ghost: database
...
+
+{{/*
+Return true if cert-manager required annotations for TLS signed certificates are set in the Ingress annotations
+Ref: https://cert-manager.io/docs/usage/ingress/#supported-annotations
+*/}}
+{{- define "ghost.ingress.certManagerRequest" -}}
+{{ if or (hasKey . "cert-manager.io/cluster-issuer") (hasKey . "cert-manager.io/issuer") }}
+    {{- true -}}
+{{- end -}}
+{{- end -}}
```
juan131, line 158: Awesome!

juan131 (changes requested):
Thanks so mucho for this @pablogalegoc 

Please check my comments! The comments I added to Ghost can be applied to the rest of charts

bitnami/ghost/values.yaml
pablogalegoc: Done in e0c3a42!

bitnami/ghost/values.yaml
pablogalegoc: Done in bacc9c7!

bitnami/ghost/templates/_helpers.tpl
pablogalegoc: Thanks!

juan131 (approved):
LGTM

Issue #3919, opened 2020-10-06
Support for IngressClass Object in newer versions of k8s


Cellebyte:
**Which charts**:
[bitnami/nginx-ingress-controller]
[bitnami/contour]
[bitnami/kong]

**Is your feature request related to a problem? Please describe.**

Using the chart it is only possible to differentiate between two ingresses via an Annotation.
Newer Kubernetes versions support differentiating ingress definitions by IngressClass.

[https://kubernetes.io/docs/concepts/services-networking/ingress/#ingress-class](https://kubernetes.io/docs/concepts/services-networking/ingress/#ingress-class)

**Describe the solution you'd like**
Implement either a flag in the helm chart which enables new or old behaviour
or make it dependend on the running Kubernetes version which way the ingressClass gets deployed.

As Annotation or as Object.

andresbono:
Thanks for raising this new feature to our attention, it looks very interesting although we haven't explored it in depth yet. Are you already using IngressClass manifests in your clusters?

Cellebyte:
No I'm not I wanted to first support it in the Helm Charts.
But I could test how they behave and start on implementing them in the helm charts.

andresbono:
That would be great, @Cellebyte ❤️ Thanks again for your all your help and contributions to this project!

Cellebyte:
@andresbono I don't think it should be stale :)

andresbono:
Setting <kbd>on-hold</kbd> label to avoid noise from the stale bot.

muffl0n:
Would be great to also have it in chart "nginx".

discostur:
@andresbono IngressClass support is also missing on [bitnami/kibana] ... just wanted to deploy it on a newer k8s cluster but it doesn't work;

carrodher:
Thanks for reporting it. Would you like to create a PR adding it?

discostur:
@carrodher looks like this PR should fix it: https://github.com/bitnami/charts/pull/8536  ;)

carrodher:
Unfortunately, this issue was created a long time ago and although there is an internal task to fix it, it was not prioritized as something to address in the short/mid term. It's not a technical reason but something related to the capacity since we're a small team.

Being said that, contributions via PRs are more than welcome in both repositories (containers and charts). Just in case you would like to contribute.

During this time, there are several releases of this asset and it's possible the issue has gone as part of other changes. If that's not the case and you are still experiencing this issue, please feel free to reopen it and we will re-evaluate it.

PR #7668, opened 2021-10-01
[bitnami/contour] Add IngressClass resource creation support

devholic:
**Description of the change**

Add IngressClass resource creation support

**Benefits**

Let users to create `IngressClass` resource, which is useful for running multiple ingress controllers in single cluster.

**Possible drawbacks**

IngressClass is introduced at [v1.18.0](https://github.com/kubernetes/kubernetes/pull/88509), but this chart cannot be used for clusters before v1.19.0 since I hardcoded `apiVersion` to `networking.k8s.io/v1`.

> How about making `apiVersion` configurable? (through `contour.ingressClass.apiVersion`)

**Applicable issues**

- This PR can resolves part of https://github.com/bitnami/charts/issues/3919

**Additional information**

- [Ingress Class](https://kubernetes.io/docs/concepts/services-networking/ingress/#ingress-class)

**Checklist**


bitnami/contour/values.yaml
```diff
@@ -259,9 +259,13 @@ contour:
...
-  ## @param contour.ingressClass Name of the ingress class to route through this controller
-  ##
-  ingressClass: contour
+  ## @param contour.ingressClass.name Name of the ingress class to route through this controller
+  ## @param contour.ingressClass.create Whether to create or not the IngressClass resource
+  ## @param contour.ingressClass.default Mark IngressClass resource as default for cluster
+  ingressClass:
+    name: contour
+    create: true
+    default: true
```
juan131, line 268: This change introduce a backwards incompatibility since you're modifying the **values.yaml** structure. Therefore, a user trying to upgrade reusing a previous **values.yaml** that was setting the `contour.ingressClass` will find some errors.
There's a way to make the changes compatible which consists on playing with the `kindIs`/`typeIs` functions and adapt the manifests depending on the parameter type (string vs object)

juan131 (changes requested):
Thanks so much for this contribution! We really appreciate it!
Please take a look to my comments with concerns about upgradability.

bitnami/contour/values.yaml
devholic: Thanks for your review and sorry for late response 🙏 
I added type checking on Deployment/IngressClass manifest, but I'm not sure about how to document multiple type support.

Temporarily, I add comment to `values.yaml` like below: https://github.com/bitnami/charts/blob/422dddf24a4883d7e7030651969535028166852c/bitnami/contour/values.yaml#L269-L271

WDYT?

bitnami/contour/templates/contour/deployment.yaml
```diff
@@ -68,8 +68,13 @@ spec:
             - --contour-cert-file=/certs/tls.crt
             - --contour-key-file=/certs/tls.key
             - --config-path=/config/contour.yaml
-            {{- if .Values.contour.ingressClass }}
-            - --ingress-class-name={{ .Values.contour.ingressClass }}
+            {{- $ingressClass := .Values.contour.ingressClass }}
+            {{- if and (kindIs "string" $ingressClass) ($ingressClass) }}
```
juan131, line 72:
```suggestion
            {{- if and (kindIs "string" $ingressClass) (not (empty $ingressClass)) }}
```

bitnami/contour/templates/contour/deployment.yaml
```diff
@@ -68,8 +68,13 @@ spec:
...
+            - --ingress-class-name={{ $ingressClass }}
+            {{- else if (kindIs "map" $ingressClass) }}
+              {{- if $ingressClass.name }}
```
juan131, line 75:
```suggestion
              {{- if not (empty $ingressClass.name) }}
```

bitnami/contour/templates/contour/ingressclass.yaml
```diff
@@ -0,0 +1,16 @@
+{{ $ingressClass := .Values.contour.ingressClass }}
+{{- if (kindIs "map" $ingressClass) }}
```
juan131, line 2: Are the parenthesis needed?
```suggestion
{{- if kindIs "map" $ingressClass }}
```

bitnami/contour/templates/contour/ingressclass.yaml
```diff
@@ -0,0 +1,16 @@
...
+{{- if $ingressClass.create }}
+---
+apiVersion: networking.k8s.io/v1
+kind: IngressClass
+metadata:
+  name: {{ $ingressClass.name }}
+  annotations:
+    ingressclass.kubernetes.io/is-default-class: {{ $ingressClass.default | quote }}
```
juan131, line 10: We should include the namespace:
```suggestion
  namespace: {{ .Release.Namespace | quote }}
  annotations:
    ingressclass.kubernetes.io/is-default-class: {{ $ingressClass.default | quote }}
```

juan131: Also.. should we support custom annotations?

bitnami/contour/templates/contour/ingressclass.yaml
```diff
@@ -0,0 +1,16 @@
...
+  labels: {{- include "common.labels.standard" . | nindent 4 }}
+    app.kubernetes.io/component: contour
+spec:
+  controller: projectcontour.io/{{ .Release.Namespace }}/{{ include "common.names.fullname" . }}-contour
```
juan131, line 14:
```suggestion
  controller: {{ printf "projectcontour.io/%s/%s-contour" .Release.Namespace (include "common.names.fullname" .) }}
```

bitnami/contour/values.yaml
juan131: I'd add a warning so users understand this "string" approach is deprecated. Also, by our conventions all the comments describing a parameter be placed BEFORE the parameter. Sth like this:
```
  ## @param contour.ingressClass.name Name of the ingress class to route through this controller
  ## @param contour.ingressClass.create Whether to create or not the IngressClass resource
  ## @param contour.ingressClass.default Mark IngressClass resource as default for cluster
  ##
  ## DEPRECATED: Use a map instead
  ## You can use the the 'contour.ingressClass' as a string to indicate the ingress
  ## class name. This will skip the creation of an IngressClass resource
  ## e.g:
  ## ingressClass: contour
  ##
  ingressClass:
    name: contour
    create: true
    default: true
```

juan131 (changes requested):
Thanks so much for implementing the suggestion! Awesome work!
I added some extra suggestions. After that, I think we're ready to merge your PR.

bitnami/contour/templates/contour/ingressclass.yaml
devholic: > We should include the namespace:

Since IngressClass is Cluster resource, we don't have to include namespace 🙂

> Also.. should we support custom annotations?

On configuration perspective, I think we don't have to support custom annotations 🤔. (As far as I know, `ingressclass.kubernetes.io/is-default-class` is the only valid option for now.)

Namespace scoped parameters (`.spec. parameters`) might be a viable option, but I think it would be better to keep as-is because contour doesn't support it.

bitnami/contour/templates/default-backend/ingress.yaml
```diff
@@ -10,11 +10,7 @@ metadata:
     {{- include "common.tplvalues.render" (dict "value" .Values.commonLabels "context" $) | nindent 4 }}
     {{- end }}
   annotations:
-    {{- if .Values.ingressClass }}
-    kubernetes.io/ingress.class: {{ .Values.ingressClass }}
-    {{- else }}
-    kubernetes.io/ingress.class: contour
-    {{- end }}
+    kubernetes.io/ingress.class: {{ include "contour.ingressClassName" . }}
```
devholic, line 13: I found extra field which uses `ingressClass` name while self-reviewing 🤔

Not only `contour` was used as default value on `default-backend`, but also contour [assumes `contour` as default ingress class name](https://github.com/projectcontour/contour/blob/5de4f21c92df9069c0fcb3998636da9f48db377d/internal/ingressclass/ingressclass.go#L23-L26). I think it would be better if we set default ingress class name explicitly - https://github.com/bitnami/charts/pull/7668/commits/ac129e427f07d394f73e5b870bfc1b2198516646, WDYT?

bitnami/contour/templates/_helpers.tpl
```diff
@@ -93,3 +93,15 @@ contour: envoy.kind
...
+{{- define "contour.ingressClassName" -}}
+{{- $ingressClass := .Values.contour.ingressClass }}
+{{- if kindIs "string" $ingressClass -}}
+    {{ default "contour" $ingressClass }}
+{{- else if kindIs "map" $ingressClass -}}
+    {{ default "contour" $ingressClass.name }}
+{{- else -}}
+    contour
+{{- end -}}
+{{- end -}}
```
juan131, line 107: Nice!

juan131 (approved):
Thanks so much!! I love the new approach

bitnami/contour/templates/contour/ingressclass.yaml
juan131: Fair enough!!

juan131:
@devholic due to current issues ongoing in our internal pipeline, please expect some delay in the release of this version of the chart in the Bitnami charts repository.

devholic:
Thanks for your review! 🙌

juan131:
@devholic the chart is now available:

```console
$ helm repo update bitnami && helm search repo bitnami/contour
Hang tight while we grab the latest from your chart repositories...
...Successfully got an update from the "bitnami" chart repository
Update Complete. ⎈Happy Helming!⎈
NAME           	CHART VERSION	APP VERSION	DESCRIPTION
bitnami/contour	5.7.0        	1.18.2     	Contour Ingress controller for Kubernetes
```

PR #7688, opened 2021-10-02
bitnami/mongodb labeler sidecar and labeler service fix

korenlev:
no need to match by 'instance' name, as this is dynamically changing per namespace

**Description of the change**

**Benefits**

**Possible drawbacks**

**Applicable issues**

  - fixes #

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/mongodb/Chart.yaml
+++ b/bitnami/mongodb/Chart.yaml
@@ -26,4 +26,4 @@ name: mongodb
 sources:
   - https://github.com/bitnami/bitnami-docker-mongodb
   - https://mongodb.org
-version: 10.26.4
+version: 10.26.5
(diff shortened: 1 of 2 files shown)


korenlev:
pass pr checks

miguelaeh (changes requested):
Hi @korenlev ,
The changed code is an example, I don't see why it is incorrect

korenlev:
Hi miguel,

If namespace change, the other mongo templates are creating pod names with
mongo-[ns] and service names mongo-[ns] so matching by it is hard, better
to not trust it and only match by simpler attributes

בתאריך יום ב׳, 4 באוק׳ 2021, 11:06, מאת Miguel Ángel Cabrera Miñagorri ‏<
***@***.***>:

miguelaeh:
Hi @korenlev ,
I understand your concern, but the ideal is to set all the labels in order to match only the resources of the release. The comments are an example that users need to edit, they are not though to be used just by uncommenting them. Approving this PR could cause labels confusions since there could be a mismatch of the resources

korenlev:
ok

‫בתאריך יום ג׳, 5 באוק׳ 2021 ב-11:48 מאת ‪Miguel Ángel Cabrera Miñagorri‬‏
***@***.***‬‏>:‬

PR #7700, opened 2021-10-05
Feature/support my pod id for external access

bigtable2006:
**Which issue this PR resolve for?**
When expose the Kafka Service on multiple clusters environment. There are some system that need the EXTERNAL access url must be used the pod id (MY_POD_ID environment param that inject by K8S) instead of HOST_ID.

But the origin source only support using the HOST_ID through enable configuration:

```
externalAccess.service.useHostIPs = true
```
There are no way to use MY_POD_ID

This PR help to support to use MY_POD_ID for external access through configuration
```
externalAccess.service.usePodIPs = true
```


randradas:
Let me run some internal checks before merging :)

bitnami/kafka/templates/scripts-configmap.yaml
```diff
@@ -91,7 +91,9 @@ data:
     {{- end }}
     export EXTERNAL_ACCESS_PORT={{ .Values.externalAccess.service.port }}
     {{- else if eq .Values.externalAccess.service.type "NodePort" }}
-    {{- if or .Values.externalAccess.service.useHostIPs .Values.externalAccess.autoDiscovery.enabled }}
+    {{- if and .Values.externalAccess.service.usePodIPs .Values.externalAccess.autoDiscovery.enabled }}
+    export EXTERNAL_ACCESS_IP="${MY_POD_IP}"
+    {{- else if or .Values.externalAccess.service.useHostIPs .Values.externalAccess.autoDiscovery.enabled }}
     export EXTERNAL_ACCESS_IP="${HOST_IP}"
     {{- else if .Values.externalAccess.service.domain }}
```
bigtable2006, line 98: I have updated the logic here. The previous updated,

```
    {{- if or .Values.externalAccess.service.useHostIPs .Values.externalAccess.autoDiscovery.enabled }}
    export EXTERNAL_ACCESS_IP="${HOST_IP}"
    {{- else if and .Values.externalAccess.service.usePodIPs .Values.externalAccess.autoDiscovery.enabled }}
    export EXTERNAL_ACCESS_IP="${MY_POD_IP}"
```
In case, `.Values.externalAccess.autoDiscovery.enabled ` is true, the logic will never go to the `else if` statement because it is using `or` condition.

It should to change the logic as below to make sure that in case we enable usePodIPs, it will do that. The new logic
```
    {{- if and .Values.externalAccess.service.usePodIPs .Values.externalAccess.autoDiscovery.enabled }}
    export EXTERNAL_ACCESS_IP="${MY_POD_IP}"
    {{- else if or .Values.externalAccess.service.useHostIPs .Values.externalAccess.autoDiscovery.enabled }}
    export EXTERNAL_ACCESS_IP="${HOST_IP}"
```
It only use the MY_POD_IP when both flags: usePodIPs, autoDiscovery.enabled is true. For other cases, it will use HOST_IP.

randradas:
I'm still running internal tests, I found an error but I need to get deeper into it I could not be related with your change.

randradas:
LGTM!

PR #7759, opened 2021-10-10
[bitnami/kafka] Add option to use custom service account

nflaig:
**Description of the change**

The changes allows to use a custom service account for Kafka and Kafka Exporter pods.

**Benefits**

Right now there is only one service account for both the Kafka pods and the Kafka Exporter pods. This leads to limitations if the service accounts need to be different. An example case would be if you are using Consul operator, the serivce account name needs to match the service name in Consul. If you want to register Kafka and Kafka Exporter in Consul you need two different service accounts.

Another benefit is when you just want to use your own service account instead of the one created by the chart which allows for more flexiblity.

**Possible drawbacks**

Change is backward compatible, there are no drawbacks.

**Checklist** 


nflaig:
@yilmi Thanks for the feedback and review again

So basically you mean we should add another/separate service account for the metrics exporter? I was considering this but not sure if it can be considered backward compatible but definitely a good option becasue as you mentioned it would be more consistent

yilmi:
@nflaig no worries, thanks again for this contribution, I hope my feedback didn't put you off! 

Yes, I think it would make sense to follow the principle of least privilege as we require different permissions (fwiw, I didn't check the permissions part but I trust you know more than I do here). 

Looking around, we do this elsewhere, you can look at ArgoCD or MetalLB charts. We do define service account templates per pod under the folder names after the pods. In this case don't bother changing the folder structure, a file prefixed in a way that makes it identifiable is enough.

Thanks again for your help!

PR #7781, opened 2021-10-12
[bitnami/mongodb] Call tpl function on mongo existing secret value

### LINKED ISSUE #7770: Support templatized existingSecret MongoDB value
**Which chart**:
mongodb: 10.27.2 (latest)

Some charts, for example RabbitMQ, allow passing in templatized values for secret name.  The MongoDB chart does not.

See the relevant helper for MongoDB:
https://github.com/bitnami/charts/blob/master/bitnami/mongodb/templates/_helpers.tpl#L157

Compared to the helper for RabbitMQ:
https://github.com/bitnami/charts/blob/master/bitnami/rabbitmq/templates/_helpers.tpl#L65

This would also improve consistency across different charts.

Adjust the MongoDB chart helper to use the `tpl` function on the existing secret name.

I would like to be able to do something like this in my values.yaml:
```
mongodb:
    auth:
      existingSecret: "{{ .Release.Name }}-mongodb"
```

The only alternative is hard-coded values.

**Additional context**

bpfoster:
**Description of the change**
Calls `tpl` function on `auth.existingSecret` so that you can do things like `{{ .Release.Name }}-mongodb-secret`.  This follows the same handling as in the RabbitMQ chart for existing secrets.

**Benefits**

**Possible drawbacks**

**Applicable issues**

  - fixes #7770

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/aspnet-core/Chart.lock
+++ b/bitnami/aspnet-core/Chart.lock
@@ -1,6 +1,6 @@
 dependencies:
 - name: common
   repository: https://charts.bitnami.com/bitnami
-  version: 1.9.0
-digest: sha256:6d608d323ac01a7950ba64fa7caf4169a5d9d33e442c99e23e123caaf303b6b9
-generated: "2021-09-14T19:08:06.666017763Z"
+  version: 1.10.0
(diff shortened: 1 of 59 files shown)


randradas:
LGTM!

PR #7799, opened 2021-10-13
[bitnami/rabbitmq] with more secure guest user

StefanIGit:
**Description of the change**

Hi, I stumbled about this.
If .Values.loadDefinition.enabled is true then 2 other issues are created.

### Smaller one:
There is a fallback to defaults user: guest pw: guest and it is created, 
on first start with empty DB >see https://www.rabbitmq.com/access-control.html#default-state
I did not find a way to prevent that.
```
Default:
default_user = guest
default_pass = guest
```
## Big one 
The guest is not bound to loopback so its potentially open to the internet with default credentials,
kind of uncool.
https://github.com/bitnami/charts/blob/11f8b800f66d48c3177f8d11c819540c405ab405/bitnami/rabbitmq/values.yaml#L291

From https://www.rabbitmq.com/configure.html#config-file 
```
# guest uses well known
# credentials and can only
# log in from localhost
# by default
loopback_users.guest = true
```
 This should be configurable setting like .Values.guestLoopbackOnly: "true" 

**Benefits**

Not open to the internet with default credentials

**Possible drawbacks**

I can imagine this is for convenience while testing/learning.

**Applicable issues**

**Additional information**

**Checklist** 


juan131:
@StefanIGit friendly reminder to check my review comments. Thanks in advance.

PR #6206, opened 2021-04-24
[bitnami/postgres-ha] Enable init-chmod-data when using extendedConf

Librazy:
**Description of the change**

Fix https://github.com/bitnami/charts/issues/6197 by enabling init-chmod-data initContainer when `postgresql.extendedConf` is not empty.

**Benefits**

`postgresql.extendedConf` works out of box and won't fails surprisingly with `'/bitnami/postgresql/conf/postgresql.conf': Permission denied` 

**Possible drawbacks**

An initContainer increase startup time and may requires a extra image. I don't know whether there is a better way to fix #1961 without initContainers.

**Applicable issues**

  - fixes #6197

**Additional information**

N/A

**Checklist** 

- [X] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [X] Variables are documented in the README.md
- [X] Title of the PR starts with chart name (e.g. `[bitnami/chart]`)

:warning: Keep in mind that if you want to make changes to the kubeapps chart, please implement them in the [kubeapps repository](https://github.com/kubeapps/kubeapps/tree/master/chart/kubeapps). This is only a synchronized mirror.

PR #7802, opened 2021-10-13
Do not enforce init containers if volumePermissions is disabled

### LINKED ISSUE #7803: volumePermissions are enforced if postgresql.extendedConfCM is used
**Which chart**:
postgresql-ha-7.11.0

**Describe the bug**
The use of the parameter postgresql.extendedConfCM forces the creation of a initContainer with the name 'init-chmod-data'. If no securityContext is set for the init container it will runAsUser: 0. The initContainer will set the ownership for data mounts like the config map to 1001 as default. Both values could be configured but for openshift it can't be used as default and it can't(shouldn't) be done without a additional permission setting. The error message is added below. IMHO: The condition should not force the use of the init container if the volumePermissions are disabled because it is not needed all the time. In this simple case the config is simply readable from the upcoming container without the init container.

e.g. result without init container:

```
...
> 1001150000@postgresql-ha-postgresql-1:/$ ls -alrt  /bitnami/postgresql/conf/conf.d/postgresql.conf 
> lrwxrwxrwx. 1 root root 22 Oct 13 17:04 /bitnami/postgresql/conf/conf.d/postgresql.conf -> ..data/postgresql.conf
> 1001150000@postgresql-ha-postgresql-1:/$ ls -alrt  /bitnami/postgresql/conf/conf.d/..data/postgresql.conf 
> -rw-r--r--. 1 root 1001150000 346 Oct 13 17:04 /bitnami/postgresql/conf/conf.d/..data/postgresql.conf
...
```

### Error Message
```
...
create Pod postgresql-ha-postgresql-1 in StatefulSet postgresql-ha-postgresql failed error: pods "postgresql-ha-postgresql-1" is forbidden: unable to validate against any security context constraint: [spec.initContainers[0].securityContext.runAsUser: Invalid value: 1001: must be in the ranges: [1001150000, 1001159999]]
...
```

**To Reproduce**
1. Use Openshift
2. Set postgresql.extendedConfCM to a non empty value
3. Wait for the upcomming statefulset

**Expected behavior**
IMHO: If i set the volumePermissions.enabled to false the statefulset should be come up without init container and the error above.

- Output of `helm version`:

```
version.BuildInfo{Version:"v3.5.4", GitCommit:"1b5edb69df3d3a08df77c9902dc17af864ff05d1", GitTreeState:"clean", GoVersion:"go1.15.11"}

```

- Output of `oc version`:

```
oc v3.11.0+0cbc58b
kubernetes v1.11.0+d4cacc0
features: Basic-Auth GSSAPI Kerberos SPNEGO

Server https://api************
kubernetes v1.20.0+bbbc079
```

**Additional context**
More details at #7802

sagiru:
[bitnami/postgresql-ha]

**Description of the change**
There is no need to use a init container with a special security context
for chmod's if a extendedConf* is used and volumePermissions are
disabled. If the init container is forced the deployment of the
statefulset will fail on e.g. openshift with a error message like below.

...pods "postgresql-ha-postgresql-1" is forbidden: unable to validate
against any security context constraint:
[spec.initContainers[0].securityContext.runAsUser: Invalid value: 0:
must be in the ranges:...

**Benefits**

The statefulsets could used on openshift without additional changes if volumePermissions are disabled and extendedConfCM is used.

**Possible drawbacks**

**Applicable issues**

  - fixes #7803

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/postgresql-ha/Chart.yaml
+++ b/bitnami/postgresql-ha/Chart.yaml
@@ -27,4 +27,4 @@ name: postgresql-ha
 sources:
   - https://github.com/bitnami/bitnami-docker-postgresql
   - https://www.postgresql.org/
-version: 7.11.0
+version: 7.11.1
--- a/bitnami/postgresql-ha/templates/postgresql/statefulset.yaml
+++ b/bitnami/postgresql-ha/templates/postgresql/statefulset.yaml
@@ -99,7 +99,7 @@ spec:
             - name: postgresql-certificates
               mountPath: /opt/bitnami/postgresql/certs
       {{- end }}
-      {{- if and .Values.persistence.enabled (or (or (not (empty .Values.postgresql.extendedConf)) (not (empty .Values.postgresql.extendedConfCM)) ) .Values.volumePermissions.enabled) }}
+      {{- if and .Values.persistence.enabled (and (or (not (empty .Values.postgresql.extendedConf)) (not (empty .Values.postgresql.extendedConfCM)) ) .Values.volumePermissions.enabled) }}
         - name: init-chmod-data
           image: {{ include "postgresql-ha.volumePermissionsImage" . }}
           imagePullPolicy: {{ .Values.volumePermissionsImage.pullPolicy | quote }}


migruiz4:
Hi @sagiru,

I'm sorry but, checking at the original PR #6206, the volumePermissions initContainer is required when using `extendedConf` as explained here: https://github.com/bitnami/charts/pull/6206#pullrequestreview-647910293

The current condition is a bit messy, but overall defines the following logic:
```
.Values.persistence.enabled AND ( .Values.volumePermissions.enabled OR ".Values.postgresql.extendedConf not empty" OR ".Values.postgresql.extendedConfCM not empty")
```

While your proposed logic would be:
```
.Values.persistence.enabled AND  .Values.volumePermissions.enabled AND (".Values.postgresql.extendedConf not empty" OR ".Values.postgresql.extendedConfCM not empty")
```
Enforcing `extendedConf` to be provided in order to enable the volumePermissions container.

An alternative would be setting the condition to `.Values.persistence.enabled AND  .Values.volumePermissions.enabled` and adding a warning the chart `Notes.txt`. But not sure if it would make sense considering the chart would fail if permissions are not fixed after setting `extendedConf`.

abhi1693:
@migruiz4 I did some research and found this gem

https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#configure-volume-permission-and-ownership-change-policy-for-pods

I have not tried it but looks like it does what our custom `init-chmod` container does.

abhi1693:
@migruiz4 Can you please review my PR wherein I have resolved the issue using PSP and RBAC

migruiz4:
I will close this PR because, as I previously mentioned, the usage of `extendedConf` does indeed require to run the volumePermissions init container because of what is mentioned [here](https://github.com/bitnami/charts/pull/6206#pullrequestreview-647910293)).

As a workaround, the feature added #7837 may solve the problem with privilege-restricted clusters.

PR #7837, opened 2021-10-18
[bitnami/postgresql-ha] Adds PSP and RBAC

### LINKED ISSUE #7803: volumePermissions are enforced if postgresql.extendedConfCM is used

abhi1693:
**Description of the change**
I have added `PodSecurityPolicy` and `RBAC` configurations which are needed by the `init-chmod` container to fix the permissions. Without these, in a restricted cluster, the operation fails.

**Benefits**

For people running a hardened cluster or using Openshift, this should help resolve the issues with fixing the permissions on the volume.

**Possible drawbacks**

**Applicable issues**

  - fixes #7803 

**Additional information**

Another PR #7802  was opened which suggested disabling the volume permissions, since, that is not the correct way, I have added the missing configuration to make it work.

**Checklist** 


abhi1693:
@marcosbc Could you please merge the change and roll out a release for the chart?

PR #7845, opened 2021-10-18
Improved documentation for TLS of the Grafana chart.

Panzki:
**Description of the change**

This PR extends the documentation for the Grafana charts TLS settings.

**Benefits**

It's clearer how to configure working TLS for the Grafana chart

**Possible drawbacks**

None.

**Applicable issues**

none

**Additional information**
When viewing the diff check the rendered Markdown. The raw diff is pretty cluttered, due to formatting changes.

**Checklist** 


Panzki:
Hi,
let me explain the inconvenience we had. We're using cert-manager with the `cert-manager.io/cluster-issuer` annotation. Since the `ingress.certManager` value is not documented we did not know about it. This caused the problem, that the certificate provided by cert-manager was not used and therefore HTTPS did not work properly.
I think either the documentation should mention how to properly configure HTTPS or the template for the ingress should altered.

https://github.com/bitnami/charts/blob/416cfd0acc9a49cc10223f52eb5fa5a703d9aaed/bitnami/grafana/templates/ingress.yaml#L53-L58

If the `and` in the if statement would be removed the `ingress.certManager` annotation would not be needed anymore. Is there any particular reason why it is not sufficient to just let the user set `ingress.tls=true`? Why the extra condition?

PR #7847, opened 2021-10-18
[bitnami/elasticsearch] Do not set StatefulSet.replicas if autoscaling is enabled

pierophp:
**Description of the change**

If autoscaling is enabled do not set StatefulSet.replicas

**Benefits**

When using HPA the elasticsearch chart is still setting the statefulset replicas to the default value of 3. Setting the replicas and enabling the HPA conflicts with each other, this PR only sets replicas if HPA is not enabled. It applies to master, data and coordinating HPA.
It also changes how ELASTICSEARCH_TOTAL_NODES, ELASTICSEARCH_CLUSTER_MASTER_HOSTS and ELASTICSEARCH_MINIMUM_MASTER_NODES is set (considering autoscaling).

**Possible drawbacks**

I think there is no drawback

**Applicable issues**

N/A

**Additional information**
The same as this PR:
https://github.com/bitnami/charts/pull/7417

**Checklist** 

### DIFF
--- a/bitnami/elasticsearch/Chart.yaml
+++ b/bitnami/elasticsearch/Chart.yaml
@@ -25,4 +25,4 @@ name: elasticsearch
 sources:
   - https://github.com/bitnami/bitnami-docker-elasticsearch
   - https://www.elastic.co/products/elasticsearch
-version: 17.1.1
+version: 17.1.2
--- a/bitnami/elasticsearch/templates/coordinating-statefulset.yaml
+++ b/bitnami/elasticsearch/templates/coordinating-statefulset.yaml
@@ -16,7 +16,9 @@ spec:
     matchLabels: {{- include "common.labels.matchLabels" . | nindent 6 }}
       app.kubernetes.io/component: coordinating-only
   podManagementPolicy: Parallel
+  {{- if not .Values.coordinating.autoscaling.enabled }}
   replicas: {{ .Values.coordinating.replicas }}
+  {{- end }}
   serviceName: {{ template "elasticsearch.coordinating.fullname" . }}
   template:
     metadata:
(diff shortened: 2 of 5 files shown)


carrodher:
As reported by @mprusinski in this issue (https://github.com/bitnami/charts/issues/8132), the changes implemented in this PR are not working when deploying the chart with the following params:
```yaml
ingest:
  enabled: true
  replicas: 1
```
```console
$ helm template elastic bitnami/elasticsearch --set ingest.enabled=true --set ingest.replicas=1
Error: template: elasticsearch/templates/ingest-statefulset.yaml:118:101: executing "elasticsearch/templates/ingest-statefulset.yaml" at <eq .Values.master.autoscaling.enabled>: error calling eq: missing argument for comparison
```

There is an `(eq .Values.master.autoscaling.enabled)` statement repeated 3 times at _templates/ingest-statefulset.yaml_ which only have 1 argument, while 2 arguments are needed for the comparison:

https://github.com/bitnami/charts/blob/63a63c2cc227622d64cc450567448346600b613e/bitnami/elasticsearch/templates/ingest-statefulset.yaml#L118

https://github.com/bitnami/charts/blob/63a63c2cc227622d64cc450567448346600b613e/bitnami/elasticsearch/templates/ingest-statefulset.yaml#L121

https://github.com/bitnami/charts/blob/63a63c2cc227622d64cc450567448346600b613e/bitnami/elasticsearch/templates/ingest-statefulset.yaml#L124

@Mauraza can you please take a look?
@pierophp is this solution working for you? Probably you're using other parameters, isn't you?

pierophp:
Hi @carrodher I will look into this

pierophp:
@Mauraza  The PR is great! I didn't test this specific scenario.

Something that I think would help is having some sort of unit tests for the charts, what do you think?

carrodher:
There are some tests executed in the internal Bitnami test & release pipeline, but it is difficult to deploy and test the chart using a combination of all the different values supported in the _values.yaml_, so some scenarios are not covered and this kind of corner case can appear.

[Here](https://docs.bitnami.com/tutorials/bitnami-best-practices-hardening-charts/#bitnami-release-process-and-tests) you can find more info about the test & release process.

PR #7848, opened 2021-10-18
[bitnami/nginx-ingress-controller] Truncate ingress-nginx-controller metrics service name 

aiceball:
Current metrics service name can exceed the 63 char limit.

**Description of the change**

Add truncation and trim to fix that.

**Benefits**

Stay within the 63 char limit

**Possible drawbacks**

There may be a more elegant solution to keep the `-metrics` identification, like further truncating the passed full name instead to fit the 63 char limit with `-metrics` included, I can apply that if it's preferred.

**Additional information**

**Checklist** 


migruiz4:
Hi @aiceball,

The current solution looks good to me as we are using the same approach in other charts.

PR #7863, opened 2021-10-20
[bitnami/keycloack] Fix externalDatabase password

### LINKED ISSUE #7794: [bitnami/keycloak] Keycloak will not start if externalDatabase.existingSecret is set and internal postgres is disabled
**Which chart**:
keycloak 5.1.2

**Describe the bug**
With following values:
```yaml
keycloak:
  postgresql:
    enabled: false
  externalDatabase:
    existingSecret: iam-keycloak-env
```

Both **env** and **envFrom** for PSQL password are generated in StatefulSet:
```yaml
    - name: KEYCLOAK_DATABASE_PASSWORD
      valueFrom:
        secretKeyRef:
          key: postgresql-password
          name: iam-keycloak
    envFrom:
    - configMapRef:
        name: iam-keycloak-env-vars
    - secretRef:
        name: iam-keycloak-env
```

But **iam-keycloak** doesn't contain **postgresql-password** due to following check:
```gotpl
  {{- if and (not .Values.postgresql.enabled) (not .Values.externalDatabase.existingSecret) }}
  postgresql-password: {{ .Values.externalDatabase.password | b64enc | quote }}
  {{- end }}
```

Because of this keycloak pods will never start.

**To Reproduce**

1. Install chart with values from bug description

**Expected behavior**
**KEYCLOAK_DATABASE_PASSWORD** should not be in StatefulSet if **externalDatabase.existingSecret** is set.

miguelaeh:
**Description of the change**

**Benefits**

**Possible drawbacks**

**Applicable issues**

  - fixes #7794

**Additional information**

**Checklist** 


jack1902:
This was originally removed in https://github.com/bitnami/charts/pull/7319

miguelaeh:
Hi @jack1902 ,
On that PR the `externalDatabase.existingSecret` was not taken into account.
Nevertheless, I think that it would be better to be consistent with the rest of the chart and require the `postgresql-password` in the existing secret instead of adding this condition back which would require adding it as env var to the secret.

Issue #7817, opened 2021-10-14
[Bitnami/nginx-ingress-controller] IngressClass not created by chart


balonik:
**Which chart**:
`nginx-ingress-controller` version `8.0.6`

**Describe the bug**
NGINX version `1.0.4` used in the Helm Chart requires `IngressClass` to be present, otherwise following messages are logged:
```
I1014 14:37:31.991722       1 store.go:367] "Ignoring ingress because of error while validating ingress class" ingress="apps/nginx-class" error="no object matching key \"nginx\" in local store"
```

**To Reproduce**

1. Deploy Nginx Ingress Controller using default chart values
`helm install nginx-ingress bitnami/nginx-ingress-controller`

2. Create `Ingress` resource
```yaml
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: nginx-class
spec:
  ingressClassName: nginx
  defaultBackend:
    service:
      name: nginx
      port:
        number: 80
```

3. observe controller logs

**Expected behavior**
Chart creates the missing `ingressClass` resource.

**Version of Helm and Kubernetes**:

- Output of `helm version`:

```
version.BuildInfo{Version:"v3.7.1", GitCommit:"1d11fcb5d3f3bf00dbe6fe31b8412839a96b3dc4", GitTreeState:"clean", GoVersion:"go1.16.9"}
```

- Output of `kubectl version`:

```
Client Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.10", GitCommit:"8152330a2b6ca3621196e62966ef761b8f5a61bb", GitTreeState:"clean", BuildDate:"2021-08-11T18:06:15Z", GoVersion:"go1.15.15", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.5+k3s2", GitCommit:"724ef700bab896ff252a75e2be996d5f4ff1b842", GitTreeState:"clean", BuildDate:"2021-10-05T19:59:14Z", GoVersion:"go1.16.8", Compiler:"gc", Platform:"linux/amd64"}
```

**Additional context**

worldofgeese:
This release broke production for us.

balonik:
@carrodher thanks for your reply.
annotations are deprecated and we are not using them anymore. The new `ingressClass` should be used.
[...]

PR #7867, opened 2021-10-20
[bitnami/nginx-ingress-controller] add support for IngressClass

balonik:
**Description of the change**

Adds option to create `IngressClass` resource which is required in NGINX Ingress Controller 1.0.x. Ref [docs](https://kubernetes.github.io/ingress-nginx/#what-is-an-ingressclass-and-why-is-it-important-for-users-of-ingress-nginx-controller-now)
Properly handles setting of custom `IngressClass` (i.e. in multi-ingress deployments)
Adds a note how to set custom `kubernetes.io/ingress.class` annotation in NGINX Ingress Controller 1.0.x

**Benefits**

It will be possible to use of `IngressClass` instead of deprecated `kubernetes.io/ingress.class` annotation.

**Possible drawbacks**

Only deprecated annotation `kubernetes.io/ingress.class` will be supported by this chart.
Users would need to figure out themselves how to set custom annotation, because the `ingressClass` parameter in `values.yaml` is broken. 

**Applicable issues**

  - fixes #https://github.com/bitnami/charts/issues/7817

**Additional information**

**Checklist** 


rafariossaa:
Hi,
  Could you rebase your PR ?

PR #7870, opened 2021-10-20
[bitnami/grafana] fix variables in readme

yksflip:
**Description of the change**

update README.md to changes made in [#7733](https://github.com/bitnami/charts/pull/7733)


carrodher:
Thanks for letting us know. I think in this case the issue is not in the README but in the _values.yaml_. In the PR you mentioned the chart was standardized but those values were not moved under the `grafana:` object
@miguelaeh can you take a look?

yksflip:
Hey,
thanks for your comments! I think  #7881 is then the right way to go :)

PR #7970, opened 2021-10-29
[bitnami/apache] Fix clone htdocs when restart

### LINKED ISSUE #7966: [apache] Init container fails to clone the git repo again
**Which chart**:
Apache

**Describe the bug**
I clone htdocs from Github and use `emptyDir` as the volume type.
It works at the first time but once the pod is restarted, then the initContainer `git-clone-repository` failed with the following error.
```
fatal: destination path '/app' already exists and is not an empty directory.
```
So I can see that the initContainer clones the git repo in `/app` path. But that is already there because that path is emptyDir volume.

**To Reproduce**
Rollout, scale down and up again deployments are not the cases. Also to delete the pod is not the case because those actions will delete the emptyDir. EmptyDir is removed once the pod is removed from the node.

Only when the same pod is restarted, we get the issue. This can be reproduced when memory usage of the pod is over, OOM killed.

**Expected behavior**
It would be good to remove all contents in the `/app` before the clone.
```
        initContainers:
        - command:
          - /bin/bash
          - -ec
          - |
            cd /app
            rm -rf ..?* .[!.]* *
            git clone ****
```

Not related to helm or kubectl version.
Thanks.

miguelaeh:
**Description of the change**

**Benefits**

**Possible drawbacks**

**Applicable issues**

  - fixes #7966

**Additional information**

**Checklist** 


alemorcuq (approved):
LGTM

bitnami/apache/templates/deployment.yaml
```diff
@@ -63,7 +63,8 @@ spec:
             - -ec
             - |
               [[ -f "/opt/bitnami/scripts/git/entrypoint.sh" ]] && source "/opt/bitnami/scripts/git/entrypoint.sh"
-              git clone {{ .Values.cloneHtdocsFromGit.repository }} --branch {{ .Values.cloneHtdocsFromGit.branch }} /app
+              git clone {{ .Values.cloneHtdocsFromGit.repository }} --branch {{ .Values.cloneHtdocsFromGit.branch }} /tmp/repo
+              [[ "$?" -eq 0 ]] && rm -rf /app/* && mv /tmp/repo/* /app/
```
cuttingedge1109, line 67: This doesn't remove `.*` file/directories.
How about this?

miguelaeh: Hi @cuttingedge1109 ,
AFAIK the `/*` will remove either files or directories

cuttingedge1109: @miguelaeh well ofc, it will remove dirs and files but not dirs with the name `.*`, typically `.git` dir is not  deleted with that command.
So i suggest this one.
```
rm -rf ..?* .[!.]* *
```

PR #7979, opened 2021-10-30
[bitnami/fluentd] Image configuration overrides

majimenez-stratio:
**Description of the change**

In some situations, it could be useful to specify different docker images for forwarders and aggregators. I added new values to override image configuration in forwarder and aggregator sections. If none are specified, the image configuration will be taken from the image/global sections as it was before, so this should be backwards compatible.

**Benefits**

For example, when using your own custom images, if you have different plugins for forwarders and aggregators, this PR will be handy.

**Possible drawbacks**

None that I know of

**Applicable issues**

None

**Additional information**

**Checklist** 


migruiz4 (approved):
Thank you very much for the contribution @majimenez-stratio! Looks good to me!

PR #7982, opened 2021-10-31
[bitnami/postgresql-ha] Fix service port type for metrics

abhi1693:
**Description of the change**

In my previous PR, I had assumed the service port to be always a string. But, that's not always the case and many people pass it as an integer. However, in our tests, we found that the helm template parses it as a float64. This causes issues in parsing the port for metrics.

**Benefits**

Whether the service port is passed as a string, int64 or float64 the changes properly typecast into the template to avoid any issues

**Possible drawbacks**

NA

**Applicable issues**

  - fixes #

**Additional information**

**Checklist** 


marcosbc (approved):
LGTM!

PR #7989, opened 2021-11-01
[bitnami/apache] Add tls secrets to match values example

YKWeyer:
**Description of the change**

The MR adds a `tls-secrets.yaml` template to output the corresponding values set in `ingress.secrets` as documented in the `values.yaml` defaults.

**Benefits**

The secrets set in `ingress.secrets` are now deployed along with the chart instead of being ignored.

**Possible drawbacks**

In cases where users set values for `ingress.secrets`, noticed nothing was deployed and created the secrets manually without updating their `values.yaml` files, they might encounter a conflict during the next deployment.

The proper way to resolve the conflict would be to ensure the values `ingress.secrets` are the latest valid ones, remove the manually created secrets, and redeploy everything using Helm.

**Additional information**

I used `template/CHART_NAME/templates/tls-secret.yaml` as a reference. It required the introduction of an `ingress.selfSigned` parameter (defaults to `false`).

I personally think this should be a minor version since there *might* be conflicts with manually deployed secrets, but I am open to discussion.

**Checklist** 


javsalgar (approved):

YKWeyer:
Hi @javsalgar ,

Thank you for merging :pray: 

You were so quick I didn't get the chance to bump the chart version after https://github.com/bitnami/charts/pull/7989 was merged, and now it seems the changes are not included in `8.9.0`. How can we fix that?

javsalgar:
Hi,

I believe we could bump the version again so all the changes are included. Could you bump it?

YKWeyer:
Apparently a patch was released in between ([`8.9.1`](https://github.com/bitnami/charts/commit/aecbb3b1e7370f81962d8c373757fe470a6c7048#diff-80270ccc2f47ca2d118d21d91d55fe4c0e204f40bf977380b8e982a315e567aeR29)), and the TLS fixes are now part of it.

Since no new parameter was introduced, and the TLS secret is now deployed as expected, I guess a patch release could make sense. Bumping it to minor now might only make it more confusing, since `8.9.1` and `8.10.0` would then be identical.

javsalgar:
Ok, so I guess we should leave it as it is.

PR #7990, opened 2021-11-01
[bitnami/apache] Add Autoscaling and Disruption Budget options

YKWeyer:
**Description of the change**

Introduce `autoscaling` and `pdb` options in the values file to create `HorizontalPodAutoscaler` and `PodDisruptionBudget` objects

**Benefits**

Allows Apache deployment to scale dynamically and limit the number of concurrent disruptions.

**Possible drawbacks**

**Applicable issues**

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/apache/templates/hpa.yaml
+++ b/bitnami/apache/templates/hpa.yaml
@@ -13,7 +13,7 @@ metadata:
   {{- end }}
 spec:
   scaleTargetRef:
-    apiVersion: apps/v1
+    apiVersion: {{ include "common.capabilities.deployment.apiVersion" . }}
     kind: Deployment
     name: {{ template "common.names.fullname" . }}
   minReplicas: {{ .Values.autoscaling.minReplicas }}
--- a/bitnami/apache/templates/hpa.yaml
+++ b/bitnami/apache/templates/hpa.yaml
@@ -1,0 +1,34 @@
+{{- if .Values.autoscaling.enabled }}
+apiVersion: {{ .Values.autoscaling.apiVersion }}
+kind: HorizontalPodAutoscaler
+metadata:
+  name: {{ template "common.names.fullname" . }}
+  namespace: {{ .Release.Namespace | quote }}
+  labels: {{- include "common.labels.standard" . | nindent 4 }}
+    {{- if .Values.commonLabels }}
+    {{- include "common.tplvalues.render" ( dict "value" .Values.commonLabels "context" $ ) | nindent 4 }}
+    {{- end }}
+  {{- if .Values.commonAnnotations }}
+  annotations: {{- include "common.tplvalues.render" ( dict "value" .Values.commonAnnotations "context" $ ) | nindent 4 }}
+  {{- end }}
+spec:
+  scaleTargetRef:
+    apiVersion: apps/v1
+    kind: Deployment
+    name: {{ template "common.names.fullname" . }}
+  minReplicas: {{ .Values.autoscaling.minReplicas }}
+  maxReplicas: {{ .Values.autoscaling.maxReplicas }}
+  metrics:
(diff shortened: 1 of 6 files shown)


bitnami/apache/templates/hpa.yaml
```diff
@@ -0,0 +1,34 @@
...
+  labels: {{- include "common.labels.standard" . | nindent 4 }}
+    {{- if .Values.commonLabels }}
+    {{- include "common.tplvalues.render" ( dict "value" .Values.commonLabels "context" $ ) | nindent 4 }}
+    {{- end }}
+  {{- if .Values.commonAnnotations }}
+  annotations: {{- include "common.tplvalues.render" ( dict "value" .Values.commonAnnotations "context" $ ) | nindent 4 }}
+  {{- end }}
+spec:
+  scaleTargetRef:
+    apiVersion: apps/v1
```
juan131, line 16: Do no hardcode the apiVersion, instead use our "common.capabilities.deployment.apiVersion" macro.
```suggestion
    apiVersion: {{ include "common.capabilities.deployment.apiVersion" . }}
```

bitnami/apache/values.yaml
```diff
@@ -385,6 +385,34 @@ ingress:
...
+## @param autoscaling.enabled Enable autoscaling for Apache deployment
+## @param autoscaling.apiVersion API Version of the HPA object (for compatibility with Openshift)
+## @param autoscaling.minReplicas Minimum number of replicas to scale back
+## @param autoscaling.maxReplicas Maximum number of replicas to scale out
+## @param autoscaling.targetCPU Target CPU utilization percentage
+## @param autoscaling.targetMemory Target Memory utilization percentage
+##
+autoscaling:
+  enabled: false
+  apiVersion: autoscaling/v2beta1
```
juan131, line 410: This info should not come from a parameter. There are two options here:
- We hardcode the apiVersion to `autoscaling/v2beta1` in the manifest.
- We create a macro in the "common" chart that returns the appropriate version for HPA based on the Kubernetes version (see https://github.com/bitnami/charts/blob/master/bitnami/common/templates/_capabilities.tpl)

juan131 (changes requested):
Thanks so much for this great PR!! We really appreciate it!!

Could you please take a look to my comments when you have a chance?

bitnami/apache/templates/hpa.yaml
YKWeyer: :sweat_smile: this was my first idea, but then I noticed the `apiVersion` was hardcoded [in the Deployment](https://github.com/bitnami/charts/blob/85eeac7ca20ec1dfc5330b70babac15a81c79003/bitnami/apache/templates/deployment.yaml#L1) we're targeting. I am concerned using two different mechanism to set that value might lead to edge cases where the HPA doesn't target the deployment at all.

What do you think? Would you prefer to keep it hardcoded everywhere, or should I change the deployment to use the `include` logic as well?

bitnami/apache/values.yaml
YKWeyer: Okay, I copied that bit from the `bitnami/kong` chart because it was more flexible than hardcoding it, and I didn't want start fiddling with the `bitnami/common` chart.

If you're happy with hardcoding it, this is what I'll do (as it is currently done in most bitnami charts using autoscaling) :+1:

bitnami/apache/templates/hpa.yaml
juan131: Yes, updating the deployment is the right approach.

bitnami/apache/values.yaml
juan131: Fair enough

juan131 (approved):
LGTM

PR #8017, opened 2021-11-04
Update readme to correct parameter name.

cancastilho:
It seems **tomcatUser** is not a valid parameter name according to the parameter table showed in this README file.
This commit replaces **tomcatUser** to **tomcatUsername**.


carrodher (approved):
Thanks for the fix, LGTM

PR #8018, opened 2021-11-04
[bitnami/rabbitmq] PodDisruptionBudget: support apiVersion policy/v1

elchenberg:
**Description of the change**

Use policy/v1 PodDisruptionBudget if available.

**Benefits**

Avoids warnings like `policy/v1beta1 PodDisruptionBudget is deprecated in Kubernetes v1.21+, unavailable in v1.25+`

**Possible drawbacks**

**Applicable issues**

  - fixes #

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/rabbitmq/templates/pdb.yaml
+++ b/bitnami/rabbitmq/templates/pdb.yaml
@@ -1,7 +1,7 @@
 {{- if .Values.pdb.create }}
 {{- if (.Capabilities.APIVersions.Has "policy/v1") }}
 apiVersion: policy/v1
-{{- else -}}
+{{- else }}
 apiVersion: policy/v1beta1
 {{- end }}
 kind: PodDisruptionBudget
--- a/bitnami/rabbitmq/templates/pdb.yaml
+++ b/bitnami/rabbitmq/templates/pdb.yaml
@@ -1,9 +1,5 @@
 {{- if .Values.pdb.create }}
-{{- if (.Capabilities.APIVersions.Has "policy/v1") }}
-apiVersion: policy/v1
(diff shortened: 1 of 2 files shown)


miguelaeh (approved):
LGTM!
Thanks!

bitnami/rabbitmq/templates/pdb.yaml
```diff
@@ -1,5 +1,9 @@
 {{- if .Values.pdb.create }}
+{{- if (.Capabilities.APIVersions.Has "policy/v1") }}
+apiVersion: policy/v1
+{{- else -}}
```
miguelaeh, line 4:
```suggestion
{{- else }}
```

miguelaeh: Just this minor change, to avoid break the template

elchenberg:
@miguelaeh, sorry for the delay. I applied your suggestion.

bitnami/rabbitmq/templates/pdb.yaml
```diff
@@ -1,5 +1,9 @@
 {{- if .Values.pdb.create }}
+{{- if (.Capabilities.APIVersions.Has "policy/v1") }}
+apiVersion: policy/v1
+{{- else }}
 apiVersion: policy/v1beta1
+{{- end }}
```
miguelaeh, line 6:
```suggestion
apiVersion: {{ include "common.capabilities.policy.apiVersion" . }}
```

I think we can reduce the logic to this since there is a function in the common dependency for that

miguelaeh (changes requested):

PR #8022, opened 2021-11-04
bitnami/rabbitmq-operator] Updating default securityContext.

michalschott:
**Description of the change**

Updating default securityContext.

**Benefits**

Increased deployment security out of the box.

**Checklist** 


pablogalegoc:
Hi @michalschott! Thanks for your suggestion, we are looking into introducing these security parameters in all of our charts in the future but some of them (capabilities and specially `readOnlyRootFilesystem`) I think will require extensive testing from our part to make sure we are not breaking anything. In the case of `readOnlyRootFilesystem` may even need changes in the container images.

michalschott:
So far it runs like that in my clusters :)

pablogalegoc:
Thanks for sharing that! However, IMHO we should include these parameters as part of a wider effort in more charts and leverage our internal testing tools to at least have a minimum assurance. It's in our plans to do it in the not-so-distant future 🙂

PR #8045, opened 2021-11-07
[bitnami/kafka] Add separate service account for kafka exporter

nflaig:
**Description of the change**

This PR adds a separate service account and rbac for Kafka exporter.

**Benefits**

Right now there is only one service account for both the Kafka pods and the Kafka Exporter pods. This leads to limitations if the service accounts need to be different. An example case would be if you are using Consul operator, the service account name needs to match the service name in Consul. If you want to register Kafka and Kafka Exporter in Consul you need two different service accounts.

Also the kafka exporter should be considered a separate service and therefore the service accounts should be separated.

**Possible drawbacks**

There are no changes for users that use the default settings but there might be changes required for users that explicitly disabled the service account before.

**Additional information**

Implements the changes discussed in #7759.

**Checklist** 


bitnami/kafka/values.yaml
```diff
@@ -942,6 +942,21 @@ metrics:
       ##   - myRegistryKeySecretName
       ##
       pullSecrets: []
+    ## Kafka exporter pods ServiceAccount
+    ## ref: https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/
+    ##
+    serviceAccount:
+      ## @param metrics.kafka.serviceAccount.create Enable creation of ServiceAccount for Kafka exporter pods
+      ##
+      create: true
```
marcosbc, line 951: I would recommend to change the default to `false`, as this could cause problems. Any concerns against that?

marcosbc (changes requested):
Thanks for the PR! Please have a look at my comment

bitnami/kafka/templates/kafka-metrics-role.yaml
```diff
@@ -0,0 +1,23 @@
+{{- if .Values.rbac.create -}}
```
juan131, line 1: This condition is incomplete
```suggestion
{{- if and .Values.rbac.create .Values.metrics.kafka.enabled Values.metrics.kafka.serviceAccount.create  }}
```

bitnami/kafka/templates/kafka-metrics-rolebinding.yaml
```diff
@@ -0,0 +1,22 @@
+{{- if and .Values.metrics.kafka.serviceAccount.create .Values.rbac.create }}
```
juan131, line 1: Same here:
```suggestion
{{- if and .Values.rbac.create .Values.metrics.kafka.enabled Values.metrics.kafka.serviceAccount.create  }}
```

bitnami/kafka/templates/kafka-metrics-serviceaccount.yaml
```diff
@@ -0,0 +1,15 @@
+{{- if .Values.metrics.kafka.serviceAccount.create }}
```
juan131, line 1: There's no sense on creating the SA if we're not enabling Kafka metrics:
```suggestion
{{- if and .Values.metrics.kafka.enabled Values.metrics.kafka.serviceAccount.create  }}
```

bitnami/kafka/values.yaml
juan131: I agree. As far as RBAC isn't enabled, it shouldn't make much a difference

bitnami/kafka/templates/kafka-metrics-role.yaml
```diff
@@ -0,0 +1,23 @@
...
+  {{- end }}
+rules:
+  - apiGroups:
+      - ""
+    resources:
+      - services
+    verbs:
+      - get
+      - list
+      - watch
```
juan131, line 22: Could you please justify why these permissions are required?

juan131 (changes requested):
Thanks for the PR, please check my comments

bitnami/kafka/templates/kafka-metrics-role.yaml
nflaig: Those are the same permissions as previously applied to the kafka exporter by using the same role/service account as kafka. I am not exactly sure what permissions are required here but I am happy to reduce it further if you have better insights on what is actually required

bitnami/kafka/templates/kafka-metrics-role.yaml
nflaig: Good catch I totally missed that

bitnami/kafka/templates/kafka-metrics-rolebinding.yaml
nflaig: fixed

bitnami/kafka/templates/kafka-metrics-serviceaccount.yaml
nflaig: fixed

nflaig:
Thanks guys for the review! I fixed the issues mentioned in the comments

bitnami/kafka/templates/kafka-metrics-role.yaml
juan131: These permissions were set for a very specific purpose: loadbalancer/nodeports auto-discovery when enabling [external access](https://github.com/bitnami/charts/tree/master/bitnami/kafka#accessing-kafka-brokers-from-outside-the-cluster). In the `auto-discovery.sh` script you can find some commands that use `kubectl` to access the K8s API and retrieve information about existing services:

- https://github.com/bitnami/charts/blob/master/bitnami/kafka/templates/scripts-configmap.yaml#L45

However, I don't see any reason to add these permissions to the Kafka metrics pods.

nflaig: Yes, it does not make sense to add the to the kafka exporter pods. The question then is do we need a role at all? Mostly the requirement is that the kafka exporter pods have separate service account and not use the same as the kafka pods

juan131: > do we need a role at all?

I don't think so. I think it's safe to remove the Role/RoleBinding for this use case.

juan131 (approved):
LGTM! Thanks for implementing the changes

bitnami/kafka/templates/_helpers.tpl
```diff
@@ -60,6 +60,24 @@ Return the proper image name (for the init container volume-permissions image)
...
+{{- define "kafka.metrics.kafka.name" -}}
+  {{- printf "%s-exporter" (include "kafka.fullname" .) | trunc 63 | trimSuffix "-" }}
+{{- end -}}
+
+{{/*
+ Create the name of the service account to use for Kafka exporer pods
+ */}}
+{{- define "kafka.metrics.kafka.serviceAccountName" -}}
+{{- if .Values.metrics.kafka.serviceAccount.create -}}
+    {{ default (include "kafka.metrics.kafka.name" .) .Values.metrics.kafka.serviceAccount.name }}
```
juan131, line 75:
```suggestion
    {{ default (include "kafka.metrics.kafka.fullname" .) .Values.metrics.kafka.serviceAccount.name }}
```

bitnami/kafka/templates/kafka-metrics-deployment.yaml
```diff
@@ -7,7 +7,7 @@
 apiVersion: {{ include "common.capabilities.deployment.apiVersion" . }}
 kind: Deployment
 metadata:
-  name: {{ template "kafka.fullname" . }}-exporter
+  name: {{ template "kafka.metrics.kafka.name" . }}
```
juan131, line 10:
```suggestion
  name: {{ template "kafka.metrics.kafka.fullname" . }}
```

juan131 (commented):
Super minor change

bitnami/kafka/templates/_helpers.tpl
```diff
@@ -60,6 +60,25 @@ Return the proper image name (for the init container volume-permissions image)
 {{ include "common.images.image" (dict "imageRoot" .Values.volumePermissions.image "global" .Values.global) }}
 {{- end -}}
 
+{{/*
+Create a default fully qualified Kafka exporter name.
+We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
+*/}
```
juan131, line 66:
```suggestion
*/}}
```

juan131 (commented):
Typo

PR #8046, opened 2021-11-07
[bitnami/contour] Allow deploying without explicit IngressClass

### LINKED ISSUE #8041: [contour] Adding --ingress-class-name to the contour depeloyment was a critical breaking change
By adding `--ingress-class-name` to the contour depeloyment, contour then uses this value to filter Ingress and HttpProxy objects and thereby ignores object which where included before. Resulting in a potenially disruptive behavior.
 `--ingress-class-name` was added in 176f1891ca4a78b2d1ba8bf146c4a4f67ef59221

https://github.com/bitnami/charts/blob/1829119eceba799634b1a8fdf8ff1d204aeec6ac/bitnami/contour/templates/contour/deployment.yaml#L71

devholic:
**Description of the change**

As @Legion2 mentioned in #8041, adding `--ingress-class-name` to controller argument breaks previous deployments, since it enables controller to filter Ingresses and HTTPProxies.

**Benefits**

Users can deploy contour as a default (implicit) cluster ingress controller.

**Possible drawbacks**

None

**Applicable issues**

- Fixes #8041
- Related to https://github.com/projectcontour/contour/issues/4134

**Additional information**

Since there was a possible breaking change in #7668, I think it makes sense to applying this patch to `v5.x`.
Could anyone give an opinion? 🙇 

**Checklist** 

### DIFF
--- a/bitnami/contour/templates/contour/deployment.yaml
+++ b/bitnami/contour/templates/contour/deployment.yaml
@@ -68,7 +68,7 @@ spec:
             - --contour-cert-file=/certs/tls.crt
             - --contour-key-file=/certs/tls.key
             - --config-path=/config/contour.yaml
-            {{- if gt (len (include "contour.ingressClassName" .)) 0 }}
+            {{- if (include "contour.isIngressClassNameDefined" .) }}
             - --ingress-class-name={{ include "contour.ingressClassName" . }}
             {{- end }}
             {{- if .Values.contour.extraArgs }}
--- a/bitnami/contour/templates/contour/deployment.yaml
+++ b/bitnami/contour/templates/contour/deployment.yaml
@@ -68,7 +68,9 @@ spec:
             - --contour-cert-file=/certs/tls.crt
             - --contour-key-file=/certs/tls.key
             - --config-path=/config/contour.yaml
+            {{- if gt (len (include "contour.ingressClassName" .)) 0 }}
             - --ingress-class-name={{ include "contour.ingressClassName" . }}
+            {{- end }}
             {{- if .Values.contour.extraArgs }}
             {{- include "common.tplvalues.render" (dict "value" .Values.contour.extraArgs "context" $) | nindent 12 }}
             {{- end }}
--- a/bitnami/contour/templates/default-backend/ingress.yaml
+++ b/bitnami/contour/templates/default-backend/ingress.yaml
@@ -10,7 +10,7 @@ metadata:
     {{- include "common.tplvalues.render" (dict "value" .Values.commonLabels "context" $) | nindent 4 }}
     {{- end }}
   annotations:
-    kubernetes.io/ingress.class: {{ include "contour.ingressClassName" . }}
+    kubernetes.io/ingress.class: {{ default "contour" (include "contour.ingressClassName" .) }}
     {{- if .Values.commonAnnotations }}
     {{- include "common.tplvalues.render" ( dict "value" .Values.commonAnnotations "context" $ ) | nindent 4 }}
     {{- end }}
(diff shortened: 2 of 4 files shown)


bitnami/contour/templates/contour/deployment.yaml
```diff
@@ -68,7 +68,9 @@ spec:
             - --contour-cert-file=/certs/tls.crt
             - --contour-key-file=/certs/tls.key
             - --config-path=/config/contour.yaml
+            {{- if gt (len (include "contour.ingressClassName" .)) 0 }}
```
carrodher, line 71: I'm not sure about this condition, can you check if the `--ingress-class-name` parameter should be added using a different approach? Like checking any value present in the _values.yaml_ or something like that, but this `gt 0` is not something we are used to using for this kind of check, it's only used for numeric values:
```console
ag 'if gt' */templates/*.yaml
elasticsearch/templates/metrics-deploy.yaml
44:            {{- if gt (int .Values.coordinating.replicas) 0 }}

etcd/templates/statefulset.yaml
178:            {{- if gt $replicaCount 1 }}

zookeeper/templates/poddisruptionbudget.yaml
2:{{- if gt $replicaCount 1 }}
```

carrodher (commented):
Thanks for your contribution! Can you please check the inline comments?

bitnami/contour/templates/contour/deployment.yaml
devholic: Thanks for your review! 🙂, and I also think this part is ambiguous.

Since we allow multiple types on `.Values.contour.ingressClass`, we need to check type before if value is defined or not. But I think this may decreases readability and reusability.

How about defining new boolean variable named `isIngressClassNameDefined` and use it?

```tpl
{{/* Whether the name of the ingress class is defined or not */}}
{{- define "contour.isIngressClassNameDefined" -}}
{{- $ingressClass := .Values.contour.ingressClass -}}
{{- if kindIs "string" $ingressClass -}}
    true
{{- else if and (kindIs "map" $ingressClass) (or $ingressClass.name $ingressClass.create) -}}
    true
{{- end -}}
{{- end -}}
```

```tpl
{{- if (include "contour.isIngressClassNameDefined" .) }}
- --ingress-class-name={{ include "contour.ingressClassName" . }}
{{- end }}
```

bitnami/contour/templates/default-backend/ingress.yaml
```diff
@@ -10,7 +10,7 @@ metadata:
     {{- include "common.tplvalues.render" (dict "value" .Values.commonLabels "context" $) | nindent 4 }}
     {{- end }}
   annotations:
-    kubernetes.io/ingress.class: {{ include "contour.ingressClassName" . }}
+    kubernetes.io/ingress.class: {{ default "contour" (include "contour.ingressClassName" .) }}
```
juan131, line 13: Shouldn't the default value come from `include "contour.ingressClassName"`?

bitnami/contour/templates/contour/deployment.yaml
juan131: I love this new approach you're suggesting, go ahead with it please

juan131 (commented):
Thanks for this fix!

bitnami/contour/templates/default-backend/ingress.yaml
devholic: I reverted this after applying https://github.com/bitnami/charts/pull/8046#discussion_r744552904 🙂

carrodher (approved):
Thanks! Changes LGTM and I will merge the PR. Unfortunately, we are working on the release of the new Contour version (`1.19.0` -> `1.19.1`), because of that, our internal test & release pipeline needs to be adapted for the new version. A delay in the publication of the Helm chart in the charts repository is expected.

PR #8069, opened 2021-11-09
[bitnami/contour] Set default value for ingressClass.default

devholic:
**Description of the change**

Currently, applying contour chart without `ingressClass.default` value will fail, because `ingressclass.kubernetes.io/is-default-class` will not have any value.

```yaml
kind: IngressClass
metadata:
  name: contour
  annotations:
    ingressclass.kubernetes.io/is-default-class: 
```

This PR fixes this issue by setting default value for `ingressClass.default`.

**Benefits**

This will work as a fallback if user doesn't define `ingressClass.default` value.

**Possible drawbacks**

None

**Applicable issues**

None

**Additional information**

None

**Checklist** 


pablogalegoc (commented):
Thanks @devholic! IMHO having the value already established in the `values.yaml` should be good enough, but better safe than sorry. Could you resolve the conflicts please?

For reference, this PR relates to #7668 and #8046

devholic:
@pablogalegoc I resolved the conflict 🙂

pablogalegoc:
Thanks!

PR #8085, opened 2021-11-10
[bitnami/mongodb] Fix service name in TLS init container

ricoberger:
**Description of the change**

This PR fixes a bug in the MongoDB Helm chart, where TLS doesn't work, when a user specified a value for `service.nameOverride` in the values file. When this property is set, the generated certificates are not working because, the certificate does not match the host name:

```
2021-11-10T08:18:39.984+0000 E  NETWORK  [Replication] The server certificate does not match the host name. Hostname: mongodb-rs-1.mongodb-rs.default.svc.cluster.local does not match SAN(s): mongodb-rs-headless, mongodb-rs-1, mongodb-rs-1.mongodb-rs-headless.default.svc.cluster.local, localhost, 127.0.0.1, CN: mongodb-rs-1
```

```yaml
service:
  nameOverride: mongodb-rs
```

**Benefits**

This allows users to use TLS when `service.nameOverride` was set.

**Possible drawbacks**

I think there shouldn't be any drawbacks, because the service name should still be correct when `service.nameOverride` wasn't set.

**Checklist** 

### DIFF
--- a/bitnami/mongodb/Chart.yaml
+++ b/bitnami/mongodb/Chart.yaml
@@ -26,4 +26,4 @@ name: mongodb
 sources:
   - https://github.com/bitnami/bitnami-docker-mongodb
   - https://mongodb.org
-version: 10.29.1
+version: 10.29.2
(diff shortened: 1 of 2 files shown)


marcosbc (approved):

PR #8091, opened 2021-11-10
[bitnami/keycloak] automounting of serviceAccountToken

KuroObi:
It enables controll over the automounting of the service account token in the pod.
By disabling the automount, potential
attackers cannot access the Kubernetes API on behalf/through the pod.

This commit enables it by default, but leaves it configurable if anyone
needs to use it.

**Description of the change**

This change gives a user finer control over security aspects if it is needed.

**Benefits**

Disabling the automounting reduces an attack vector inside the k8s cluster against the kube-api.

**Possible drawbacks**

An inexperienced user could select a keycloak clustering method that relies on contact with the kube-api, which would be not possible without a mounted AccountToken.

**Checklist** 


bitnami/keycloak/templates/statefulset.yaml
```diff
@@ -41,6 +41,7 @@ spec:
         {{- end }}
     spec:
       serviceAccountName: {{ template "keycloak.serviceAccountName" . }}
+      automountServiceAccountToken: {{ .Values.automountServiceAccountToken }}
```
migruiz4, line 44: The value used is wrong value, it should be `.Values.serviceAccount.automountServiceAccountToken`
```suggestion
      automountServiceAccountToken: {{ .Values.serviceAccount.automountServiceAccountToken }}
```

bitnami/keycloak/values.yaml
```diff
@@ -788,6 +788,9 @@ serviceAccount:
   ## If not set and create is true, a name is generated using the fullname template
   ##
   name: ""
+  ## @param serviceAccount.automountServiceAccountToken Auto-mount the service account token in the pod
+  ##
+  automountServiceAccountToken: true
```
migruiz4, line 793: Please set the default value to `false`.

As you mentioned in your description, it is safer to not automount the service account token, and we are trying to follow a 'secure by default' approach.

>By disabling the automount, potential attackers cannot access the Kubernetes API on behalf/through the pod.

migruiz4 (changes requested):

In addition to the below comments, please be aware that you are adding `automountServiceAccountToken` to the Keycloak Statefulset where it will only apply to the keycloak pod. 
I think it would make sense to add it at `serviceaccount.yaml` instead.

bitnami/keycloak/templates/statefulset.yaml
KuroObi: I've added the option to dis-/en-able it in the pod and the ServiceAccount.
This adds the possibility to disable the service Account and also disable automounting the default Service Account token.

migruiz4:
Hi @KuroObi,

Thank you for applying the proposed changes but I would like to ask you to drop the value `automountServiceAccountToken` (pod level) from your PR, for this chart to be aligned with the other charts that use `serviceAccount.automountServiceAccountToken`.

We are aware that setting `automountServiceAccountToken` granularly for each pod would be a different feature, but our charts do not support it yet and we may decide on a different approach for it in the future.

KuroObi:
> Hi @KuroObi,
> 
> Thank you for applying the proposed changes but I would like to ask you to drop the value `automountServiceAccountToken` (pod level) from your PR, for this chart to be aligned with the other charts that use `serviceAccount.automountServiceAccountToken`.
> 
> We are aware that setting `automountServiceAccountToken` granularly for each pod would be a different feature, but our charts do not support it yet and we may decide on a different approach for it in the future.

I have removed the pod level adjustability and aligned the PR to the other Charts as requested.

migruiz4 (approved):
Thank you for your contribution! Looks good to me!

Issue #6823, opened 2021-07-01
[bitnami/keycloak] keycloakConfigCli parameter is not working in 3.x.x chart version


ruzickap:
**Which chart**:
bitnami/keycloak

**Describe the bug**

`keycloakConfigCli` parameter stopped working when using helm chart version 3.x.x
 
```yaml
keycloakConfigCli:
  enabled: true
  configuration:
    myrealm.yaml: |
      realm: myrealm
      enabled: true
      displayName: My Realm
```

This is working fine for helm chart version 2.x.x ...

**To Reproduce**

I used the current keycloak helm chart with the `keycloakConfigCli`:

```yaml
keycloakConfigCli:
  enabled: true
  configuration:
    myrealm.yaml: |
      realm: myrealm
      enabled: true
      displayName: My Realm
```

The `keycloak-keycloak-config-cli` pods are failing:

```text
kubectl get pods -n keycloak
NAME                                 READY   STATUS    RESTARTS   AGE
keycloak-0                           1/1     Running   0          75s
keycloak-keycloak-config-cli-htwj2   0/1     Error     0          11s
keycloak-keycloak-config-cli-rwkr6   0/1     Error     0          75s
keycloak-postgresql-0                1/1     Running   0          75s
```

Here is the log:

```text
kubectl logs -n keycloak keycloak-keycloak-config-cli-htwj2
2021-07-01 18:19:33.272  INFO 1 --- [           main] d.a.k.config.KeycloakConfigApplication   : Starting KeycloakConfigApplication v3.4.0 using Java 11.0.11 on keycloak-keycloak-config-cli-htwj2 with PID 1 (/opt/bitnami/keycloak-config-cli/keycloak-config-cli-12.0.4.jar started by ? in /opt/bitnami/keycloak-config-cli)
2021-07-01 18:19:33.275  INFO 1 --- [           main] d.a.k.config.KeycloakConfigApplication   : No active profile set, falling back to default profiles: default
2021-07-01 18:19:34.074  INFO 1 --- [           main] d.a.k.config.KeycloakConfigApplication   : Started KeycloakConfigApplication in 1.474 seconds (JVM running for 2.079)
2021-07-01 18:19:34.834  INFO 1 --- [           main] d.a.k.config.KeycloakConfigRunner        : Importing file '/config/myrealm.yaml'
[...]

PR #8107, opened 2021-11-11
[bitnami/keycloak] Run correct configCli jar for version of keycloak

jack1902:
**Description of the change**

By default the docker image used within the keycloak chart invokes a symlinked copy of the keycloakConfigCli jar file. The symlinked jar may or may not be compatible with the version of keycloak you are using. This is unpredictable, but is easy to fix by being more specific on the `command` that you want to run as multiple `jars` are available within the container.

**Benefits**

**Possible drawbacks**

Removes an annoying "out of the box" bug that resorted to me raising an issue on the keycloak-config-cli image produced by bitnami which is something that has been brought up previously but I couldn't find it.

**Applicable issues**

  - fixes #https://github.com/bitnami/bitnami-docker-keycloak-config-cli/issues/2
  - fixes #https://github.com/bitnami/charts/issues/6823

**Additional information**

See https://githubmemory.com/repo/bitnami/bitnami-docker-keycloak/issues/9 for the explanation around the JAR fix.

I originally removed the render piece as I really don't understand what is going on via the nested common subchart. If an explanation of how I can use the common subchart along with a command override that would be great 👍 

**Checklist** 

### DIFF
--- a/bitnami/keycloak/values.yaml
+++ b/bitnami/keycloak/values.yaml
@@ -313,7 +313,10 @@ keycloakConfigCli:
     helm.sh/hook-weight: "5"
   ## @param keycloakConfigCli.command Command for running the container (set to default if not set). Use array form
   ##
-  command: []
+  command:
+    - java
+    - -jar
+    - '{{- printf "/opt/bitnami/keycloak-config-cli/keycloak-config-cli-%s.jar" .Chart.AppVersion }}'
   ## @param keycloakConfigCli.args Args for running the container (set to default if not set). Use array form
   ##
   args: []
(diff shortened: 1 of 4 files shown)


bitnami/keycloak/values.yaml
```diff
@@ -313,7 +313,10 @@ keycloakConfigCli:
     helm.sh/hook-weight: "5"
   ## @param keycloakConfigCli.command Command for running the container (set to default if not set). Use array form
   ##
-  command: []
+  command:
```
javsalgar, line 316: The idea of `command` is to override the default one. If what you are proposing should be the default command, then I would move it directly to the yaml, so it looks like this.

```
          {{- if .Values.keycloakConfigCli.command }}
          command: {{- include "common.tplvalues.render" (dict "value" .Values.keycloakConfigCli.command "context" $) | nindent 12 }}
          {{- else }}
              - java
              - -jar
              - '{{- printf "/opt/bitnami/keycloak-config-cli/keycloak-config-cli-%s.jar" .Chart.AppVersion }}'
          {{- end }}
```

javsalgar (changes requested):

javsalgar (approved):
Awesome! LGTM

Issue #6118, opened 2021-04-15
[bitnami/keycloak] Support multiple database vendors for externalDatabase


m33li5:
**Which chart**:
[bitnami/keycloak] 2.4.3

**Is your feature request related to a problem? Please describe.**
Currently the `bitnami/keycloak` chart only supports PostgreSQL as the database, but the upstream `jboss/keycloak` image [supports choosing between multiple database vendors](https://github.com/keycloak/keycloak-containers/blob/master/server/README.md#database).

**Describe the solution you'd like**
Add an option `externalDatabase.vendor` which supports choosing between the different database vendors as it does in the upstream image.

**Describe alternatives you've considered**
For example, Codecentric's helm chart [supports something similar](https://github.com/codecentric/helm-charts/tree/master/charts/keycloak#database-setup).

**Additional context**
It would great to deploy `bitnami/keycloak` chart with `bitnami/mariadb-galera` as the database for a HA setup.

juan131:
Hi @m33li5 

Thanks so much for this great suggestion!

I'll open an internal task to evaluate adding support for other database on Keycloak. We'll keep you updated about

carrodher:
Thanks for the PR, unfortunately, there are several changes to be performed in our internal tooling and our test & release pipeline in order to integrate a new component as part of this container.
Due to our capacity and other priorities the team has, we won't be able to do this effort in the mid-term.

JuryA:
> Thanks for the PR, unfortunately, there are several changes to be performed in our internal tooling and our test & release pipeline in order to integrate a new component as part of this container.
> 
> Due to our capacity and other priorities the team has, we won't be able to do this effort in the mid-term.

No offense, I’m just trying to understand... but what do you mean by “...several changes to be performed...”? We patched the Bitnami Keycloak chart (over-mounted entry point shell script with the same one but with removed restriction for PostgreSQL) and it works like a charm! We did several tests and have dozens of development and testing environments where Oracle, MariaDB or H2 is used without any problem (I except that it won't go as good FYI)... Is there any other reason why it’s blocked? (Especially if you decline contributed PRs?) Thank your explanation.

PR #8108, opened 2021-11-11
[bitnami/keycloak] Feature/keycloak external databases

jack1902:
**Description of the change**

Attempting to add additional Database vendor support to the chart. I might be overlooking a few things but this seems to be ok having looked at the open issue and referenced links.

**Benefits**

Additional external database vendors can be used with the bitnami keycloak chart which should increase the number of users of the keycloak chart.

**Possible drawbacks**

Changes MIGHT need to be made to the bitnami keycloak image produced in order to support this (may need to add `KEYCLOAK_DB_VENDOR` or overwrite `KEYCLOAK_DATABASE_HOST` which is hardcoded to `postgresql`

https://github.com/bitnami/bitnami-docker-keycloak/blob/2943b693d5a3e122d32731383f45bda3dab6c5b8/15/debian-10/rootfs/opt/bitnami/scripts/keycloak-env.sh#L141

**Applicable issues**

  - fixes #https://github.com/bitnami/charts/issues/6118

**Additional information**

From the issue linked, it references: https://github.com/keycloak/keycloak-containers/blob/main/server/README.md#database

**Checklist** 


bitnami/keycloak/values.yaml
```diff
@@ -937,3 +937,6 @@ externalDatabase:
   ## @param externalDatabase.existingSecret Use an existing secret file with the external PostgreSQL credentials
   ##
   existingSecret: ""
+  ## @param externalDatabase.vendor Database vendor
+  ##
+  vendor: "postgres"
```
juan131, line 942: Please add a new line character at the end of file, as suggested by the linter

bitnami/keycloak/templates/statefulset.yaml
```diff
@@ -204,6 +204,8 @@ spec:
                   name: {{ $globalSecretName }}
                   key: {{ include "common.secrets.key" (dict "existingSecret" .Values.auth.existingSecret "key" "management-password") }}
                 {{- end }}
+            - name: DB_VENDOR
+              value: {{ .Values.externalDatabase.vendor }}
```
juan131, line 208: I don't think this is enough
This env. variable (`DB_VENDOR`) isn't taken into consideration by the container init logic that will configure Keycloak to use PostgreSQL regardless this value. See:

- https://github.com/bitnami/bitnami-docker-keycloak/blob/master/15/debian-10/rootfs/opt/bitnami/scripts/libkeycloak.sh#L102
- https://github.com/bitnami/bitnami-docker-keycloak/blob/master/15/debian-10/rootfs/opt/bitnami/scripts/libkeycloak.sh#L108
- https://github.com/bitnami/bitnami-docker-keycloak/blob/master/15/debian-10/rootfs/opt/bitnami/scripts/libkeycloak.sh#L367

bitnami/keycloak/templates/secrets.yaml
```diff
@@ -25,7 +25,7 @@ data:
   management-password: {{ randAlphaNum 10 | b64enc | quote }}
   {{- end }}
   {{- if and (not .Values.postgresql.enabled) (not .Values.externalDatabase.existingSecret) }}
-  postgresql-password: {{ .Values.externalDatabase.password | b64enc | quote }}
+  database-password: {{ .Values.externalDatabase.password | b64enc | quote }}
```
juan131, line 28: This breaks backwards compatibility (forcing you to bump the major) but I think it's something we can workaround. Check my suggestion:
```suggestion
  {{/*
  We use a different key for PostgreSQL for backwards compatibility, see https://github.com/bitnami/charts/pull/8108
  TO BE DEPRECATED in 6.0.0
  */}}
  {{- if eq .Values.externalDatabase.vendor "postgres" }}
  postgresql-password: {{ .Values.externalDatabase.password | b64enc | quote }}
  {{- else }}
  database-password: {{ .Values.externalDatabase.password | b64enc | quote }}
  {{- end }}
```

bitnami/keycloak/templates/statefulset.yaml
```diff
@@ -212,7 +214,7 @@ spec:
                   key: {{ include "common.secrets.key" (dict "existingSecret" .Values.auth.existingSecretPerPassword "key" "databasePassword") }}
                 {{- else }}
                   name: {{ include "keycloak.databaseSecretName" . }}
-                  key: "postgresql-password"
+                  key: "database-password"
```
juan131, line 217: Following my previous comment:
```suggestion
                  {{/*
                  We use a different key for PostgreSQL for backwards compatibility, see https://github.com/bitnami/charts/pull/8108
  TO BE DEPRECATED in 6.0.0
                  */}}
                  {{- if eq .Values.externalDatabase.vendor "postgres" }}
                  key: "postgresql-password"
                  {{- else }}
                  key: "database-password"
                  {{- end }}
```

juan131 (changes requested):
Thanks so much for the contribution!! Please take a look to my comments when you have a chance!

bitnami/keycloak/templates/statefulset.yaml
jack1902: perhaps for consistency with other env-vars within the container, I'll switch to `KEYCLOAK_DATABASE_VENDOR`? It feels weird seeing DB_VENDOR up against DATABASE_HOST etc inside the container. I'll add the logic as a name change will be a simple find and replace if we want to go back to DB_VENDOR

jack1902: Looking into it a bit more, it doesn't appear to be as simple as using that env-var.

I can see a bunch of .XML files in https://github.com/keycloak/keycloak-containers/tree/main/server/tools which i believe are set via the flags in the scripts from bitnami.

I added some bits for DB_VENDOR and KEYCLOAK_DB_VENDOR similar to the other vars but will pass KEYCLOAK_DB_VENDOR via the chart as its like the rest.

In order to make that var useful though additional things will need to be passed. Prehaps getting a base in to support postgres and something else by default would enable others to expand upon the externalDatabases as and when they want that?

bitnami/keycloak/Chart.yaml
```diff
@@ -26,4 +26,4 @@ name: keycloak
 sources:
   - https://github.com/bitnami/bitnami-docker-keycloak
   - https://github.com/keycloak/keycloak
-version: 5.1.6
+version: 6.0.0
```
juan131, line 29: The major version bump isn't necessary now that you are using a backwards compatible approach.

bitnami/keycloak/templates/statefulset.yaml
juan131: > I'll switch to KEYCLOAK_DATABASE_VENDOR? It feels weird seeing DB_VENDOR up against DATABASE_HOST etc inside the container

Yes please. The images that we expose to configure Keycloak base on users preferences should be prefixed with `KEYCLOAK_`.

> Looking into it a bit more, it doesn't appear to be as simple as using that env-var.

I expected that. A resilient logic to automate the configuration of certain apps can be challenging.

> In order to make that var useful though additional things will need to be passed

What thing are you referring to?

jack1902: It's around the configuration, just setting the DB_VENDOR alone won't be enough to connect to a specific DB for example. Looking at the links it appears we need to update the image to actually make use of the variable and configure Keycloak appropriately

migruiz4:
Hi @jack1902,

Thank you very much for your contribution and applying suggested changes! Your PR looks good to me, but I will add the label `on-hold` until the `bitnami/keycloak` image supports multiple database vendors (`KEYCLOAK_DB_VENDOR`/`KEYCLOAK_DATABASE_VENDOR`) so we can properly test this feature.

For reference, please check @juan131 's comment pointing to some parts of the container logic where 'Postgresql' is hardcoded:
> This env. variable (DB_VENDOR) isn't taken into consideration by the container init logic that will configure Keycloak to use PostgreSQL regardless this value. See:
>
>-    https://github.com/bitnami/bitnami-docker-keycloak/blob/master/15/debian-10/rootfs/opt/bitnami/scripts/libkeycloak.sh#L102
>-    https://github.com/bitnami/bitnami-docker-keycloak/blob/master/15/debian-10/rootfs/opt/bitnami/scripts/libkeycloak.sh#L108
>-    https://github.com/bitnami/bitnami-docker-keycloak/blob/master/15/debian-10/rootfs/opt/bitnami/scripts/libkeycloak.sh#L367

jack1902:
@migruiz4 am i able to get any assistance on this for the base image to carry out the needed functions to support additional databases? (even if we went down a path of getting the initial bits in and then can drip-feed in additional database vendors as needed)

migruiz4:
Hi @jack1902,

I have created an internal task to add support for different database vendors in Keycloak. But, since we are a small team and have to deal with other priorities, I can not give an ETA for when will the new feature be available.

owit-infra:
Hi team, is there an update on this ?

jack1902:
I'm going to abandon this PR as I don't forsee it being feature complete for quite sometime and the number of merge-conflicts are only going to increase overtime.

If someone wishes to pick this up, feel free to use this PR as a reference point.

bitnami/keycloak/templates/statefulset.yaml
JuryA: It works with the small patch (just a few lines) to the entry point shell script in the docker container. We use Bitnami Keycloak with Oracle, MariaDB, and H2 since the end of October 2022 in our Dev and Test environments without any issues...

PR #8110, opened 2021-11-12
[bitnami/haproxy] Add service.labels input

ryanwholey:
**Description of the change**

Add service.labels input to the bitnami/haproxy chart to allow custom labels to be passed to the service.

**Benefits**

Currently the only way to accomplish a user specified label on the HAProxy service is to use `commonLabels` which also adds the label to all other Kubernetes resources created in the chart. This addition allows for a more intuitive solution.

**Possible drawbacks**

- Adds more inputs to the chart

**Applicable issues**

**Additional information**

My use case is that I'm using a label selector in our external-dns controller to control which resources it creates records for. Creation is opt-in via a specific label. My current workaround for this haproxy chart is to use the commonLabels value to add the label to  each resource in the installation.

**Checklist** 

### DIFF
--- a/bitnami/haproxy/Chart.yaml
+++ b/bitnami/haproxy/Chart.yaml
@@ -23,4 +23,4 @@ name: haproxy
 sources:
   - https://github.com/bitnami/bitnami-docker-haproxy
   - https://github.com/haproxytech/haproxy
-version: 0.2.16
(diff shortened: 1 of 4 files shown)


carrodher (approved):
Thanks for the PR. Changes LGTM

PR #8116, opened 2021-11-12
[bitnami/*] Add network policies - third batch

pablogalegoc:
**Description of the change**

Relates to `[bitnami/drupal] Adds networkPolicy` #7857

Added the same rules for the following webapp charts: OpenCart, OrangeHRM, Osclass and ownCloud.

**Checklist** 


bitnami/orangehrm/templates/networkpolicy-backend-ingress.yaml
```diff
@@ -0,0 +1,27 @@
...
+  {{- if .Values.commonAnnotations }}
+  annotations: {{- include "common.tplvalues.render" ( dict "value" .Values.commonAnnotations "context" $ ) | nindent 4 }}
+  {{- end }}
+spec:
+  podSelector:
+    matchLabels:
+      {{- if .Values.networkPolicy.ingressRules.customBackendSelector }}
+      {{- include "common.tplvalues.render" (dict "value" .Values.networkPolicy.ingressRules.customBackendSelector "context" $) | nindent 6 }}
+      {{- else }}
+      app.kubernetes.io/name: mariadb
```
miguelaeh, line 19: I think thisis wrong

bitnami/opencart/values.yaml
```diff
@@ -670,3 +670,92 @@ certificates:
...
+    ##   label: ingress
+    ##
+    podSelector: {}
+    ## e.g:
+    ## namespaceSelector:
+    ##   label: ingress
+    ##
+    namespaceSelector: {}
+  ## @param networkPolicy.ingressRules.backendOnlyAccessibleByFrontend Enable ingress rule that makes the backend (mariadb) only accessible by OpenCart's pods.
+  ## @param networkPolicy.ingressRules.customBackendSelector [object] Backend selector labels. These labels will be used to identify the backend pods.
```
miguelaeh, line 716: do not set `[object]` or `[array]` for empty object, it is automatically recognized, please change it in all the files

bitnami/opencart/templates/networkpolicy-backend-ingress.yaml
```diff
@@ -0,0 +1,27 @@
...
+  {{- if .Values.commonAnnotations }}
+  annotations: {{- include "common.tplvalues.render" ( dict "value" .Values.commonAnnotations "context" $ ) | nindent 4 }}
+  {{- end }}
+spec:
+  podSelector:
+    matchLabels:
+      {{- if .Values.networkPolicy.ingressRules.customBackendSelector }}
+      {{- include "common.tplvalues.render" (dict "value" .Values.networkPolicy.ingressRules.customBackendSelector "context" $) | nindent 6 }}
+      {{- else }}
+      app.kubernetes.io/name: mariadb
```
miguelaeh, line 19: Check this in all the files

bitnami/osclass/values.yaml
```diff
@@ -830,3 +830,92 @@ metrics:
...
+      ##   label: access
+      ##
+      podSelector: {}
+    ## custom ingress rules
+    ## e.g:
+    ## customRules:
+    ##   - from:
+    ##       - namespaceSelector:
+    ##           matchLabels:
+    ##             label: example
```
miguelaeh, line 906:
```suggestion
    ##             label: example
    ##
```

bitnami/osclass/values.yaml
```diff
@@ -830,3 +830,92 @@ metrics:
...
+  egressRules:
+    # Deny connections to external. This is not compatible with an external database.
+    denyConnectionsToExternal: false
+    ## Additional custom egress rules
+    ## e.g:
+    ## customRules:
+    ##   - to:
+    ##       - namespaceSelector:
+    ##           matchLabels:
+    ##             label: example
```
miguelaeh, line 920:
```suggestion
    ##             label: example
    ##
```

miguelaeh: Please check the values.yaml files, the standard comment ends with an empty `##` line before the parameter

miguelaeh (changes requested):

bitnami/osclass/values.yaml
pablogalegoc: Fixed in cc9d389 👌

bitnami/osclass/values.yaml
pablogalegoc: Same, fixed in cc9d389

bitnami/opencart/values.yaml
pablogalegoc: Done in c9b5a1a

bitnami/orangehrm/templates/networkpolicy-backend-ingress.yaml
pablogalegoc: In these webapps charts we want to make the backend (mariadb in these four cases) only accessible by the frontend. We are hardcoding `app.kubernetes.io/name: mariadb` to select the mariadb pods because, unless mariadb is removed from the [dependencies](https://github.com/bitnami/charts/blob/17fd9b9819a1091ff83b9b466976a664f13b932b/bitnami/orangehrm/Chart.yaml#L5-L7), it will always be there.

bitnami/opencart/templates/networkpolicy-backend-ingress.yaml
pablogalegoc: Responded in the comment above

bitnami/orangehrm/templates/networkpolicy-backend-ingress.yaml
miguelaeh: What would happen if `mariadb.enabled` is `false` ? Or if you use an already existing deployment of mariadb instead of the dependency?

pablogalegoc: If `mariadb.enabled=false` and:
- You don't provide an already existing deployment of mariadb → Deployment fails, so here we are covered
- You provide an already existing deployment of mariadb → Then you have to use `customBackendSelector` and the `app.kubernetes.io/name: mariadb` is not added
https://github.com/bitnami/charts/blob/f203bf54e61927b999bd8391e3997f3d5b6b169d/bitnami/orangehrm/values.yaml#L737-L741

miguelaeh: 👍

miguelaeh (approved):
LGTM!

PR #8147, opened 2021-11-16
[bitnami/kube-prometheus] Add support for IngressClassName

### LINKED ISSUE #8142: Add ingressClass options to the kube-prometheus ingresses
**Which chart**:
kube-prometheus:6.3.1, possibly all Bitnami charts

You usually specify which ingress controller an ingress should use via an annotation like  `kubernetes.io/ingress.class: nginx` on the ingress. In Kubernetes 1.22, such annotations like are now deprecated in favor of a `.spec.ingressClass` field. Cf https://kubernetes.io/blog/2021/07/14/upcoming-changes-in-kubernetes-1-22/ ingress part.
Newer versions of controllers like `nginx-ingress` would by default give 404 error if `ingressClass` isn't set. https://kubernetes.github.io/ingress-nginx/#what-is-an-ingressclass-and-why-is-it-important-for-users-of-ingress-nginx-controller-now , you would need the flag '--watch-ingress-without-class' set to true (default is false).

Add  `.Values.XXX.ingress.ingressClass` parameters, check ` .capabilities.ingress.apiVersion` and set the `ingressClass` in the templates if using `networking.k8s.io/v1`. `XXX` would be `prometheus`, `thanos`, `alertmanager`.
Even better, test that `kubernetes.io/ingress.class` and `ingressClass` are not both set, since this isn't allowed.

**Context**
Other separate charts are incorporating this (contour, nginx-ingress).

migruiz4:
**Description of the change**

This PR adds the Ingress field 'IngressClassName' available in Kubernetes version 1.18+.

**Benefits**

Adds support for IngressClassName.

**Possible drawbacks**

None known.

**Applicable issues**

  - fixes #8142

**Checklist** 


miguelaeh (approved):
thank you for the PR!!!
LGTM!

Issue #8167, opened 2021-11-17
reorder metrics in HPA so CI/CD tools dont fail


farconada:
**Which chart**:
wordpress and may be other

**Describe the bug**
With HorizontalPodAutoscaler place memory metric before CPU so CI/CD tools dont fail
that this is an issue for GitOps-style synchronization of HPA objects.

**To Reproduce**
Use wordpress chart with argoCD and autoscaling enabled

**Additional context**
This is a known bug related to kubernetes, ArgoCD and other tools with just need to reorder the declaration of the metrics.

https://github.com/jaegertracing/helm-charts/pull/258
https://github.com/argoproj/argo-cd/issues/1079
https://github.com/kubernetes/kubernetes/issues/74099

Mauraza:
Hi @farconada,

I see you have already created another issue https://github.com/bitnami/charts/issues/8165 about this, rigth?

farconada:
I've done the PR by just updated the api version and a bit more but indeed with #8179 HPA could be more than monitor cpu/mem (not done with this PR)

Mauraza:
Hi @farconada,

Thanks for the contribution ❤️ !! we review it!

Issue #8165, opened 2021-11-17
Upgrade autoscaling to autoscaling/v2beta2


farconada:
**Which chart**:
wordpress

**Is your feature request related to a problem? Please describe.**
v2beta1 is deprecated in 1.19 and you should use v2beta2 going forward.
Also, v2beta2 added the new api field spec.behavior in 1.18 which allows you to define how fast or slow pods are scaled up and down.
autoscaling/V2beta2 allows users to autoscale based on custom metrics. It allow autoscaling based on metrics coming from outside of Kubernetes

**Describe the solution you'd like**
Upgrade autoscaling to autoscaling/v2beta2

miguelaeh:
Hi @farconada ,
I guess you refer to the HPA. Would you like to create a PR?

PR #8179, opened 2021-11-18
[bitnami/wordpress] update HPA to autoscaling/v2beta2 and reordering of metrics (issues #…

farconada:
…8165 and 8167)

https://github.com/bitnami/charts/issues/8165
https://github.com/bitnami/charts/issues/8167
as suggestion from @miguelaeh

**Description of the change**

**Benefits**

**Possible drawbacks**

**Applicable issues**

  - fixes #

**Additional information**

**Checklist** 


bitnami/wordpress/templates/hpa.yaml
```diff
@@ -1,5 +1,5 @@
 {{- if .Values.autoscaling.enabled }}
-apiVersion: autoscaling/v2beta1
+apiVersion: autoscaling/v2beta2
```
alemorcuq, line 2: It would be great to still support older k8s versions. To achieve this we have some helpers in the `bitnami/common` chart to set the `apiVersion` depending on which k8s version is used. See: https://github.com/bitnami/charts/blob/master/bitnami/common/templates/_capabilities.tpl

A new helper for the autoscaling version should be added to the `bitnami/common` chart that could then be used here. Would you mind opening a PR for this?

bitnami/wordpress/templates/hpa.yaml
```diff
@@ -19,16 +19,21 @@ spec:
...
       resource:
-        name: memory
-        targetAverageUtilization: {{ .Values.autoscaling.targetMemory  }}
+        name: cpu
+        target:
+            type: Utilization
+            averageUtilization: {{ .Values.autoscaling.targetCPU }}
     {{- end }}
 {{- end }}
+
```
alemorcuq, line 39: Delete this.

alemorcuq (changes requested):
Thanks for the PR! Please, don't forget to increase the chart version.

javsalgar:
Hi,

I agree that the best option would be to add a PR in the bitnami/common chart and add a a reference to the function instead of hardcoding the API version here (it would likely provide compatibility issues). Would you like to work on that? If not, we will do it.

farconada:
> Hi,
> 
> I agree that the best option would be to add a PR in the bitnami/common chart and add a a reference to the function instead of hardcoding the API version here (it would likely provide compatibility issues). Would you like to work on that? If not, we will do it.

Ok, I'll try

farconada (commented):
hpa with conditional api version

farconada:
all ok?

javsalgar:
Could you bump the common chart version as well?

javsalgar:
Could you check the merge issues and the results from the tests?

javsalgar:
After checking it, I think there might be problems of doing the same releases at the same time. Could you split the PR into two? First we release the common chart and then the wordpress chart.

PR #8183, opened 2021-11-18
[bitnami/harbor] Allow setting secret values from arbitrary Kubernetes secrets

### LINKED ISSUE #2010: [bitnami/harbor] allow using existing secrets (therefore also SealedSecrets) instead of generating them based on plain text values
**Which chart**:

harbor

**Description**

As of now,  it's not possible to use the chart (without some major workarounds) in a GitOps pipeline - `values.yaml` has passwords / secrets in plain text. Obviously, nobody wants plain text passwords committed to version control.

If the chart was to be modified in such a way that it's possible to use values from existing secret(s) instead of the plain text parameters, it would be possible to use solutions like [Sealed Secrets](https://github.com/bitnami-labs/sealed-secrets) to securely manage passwords in a GitOps friendly and secure way.

It should be possible to use existing secrets instead of the plain text values for these parameters:

| Parameter                                                    |
| ------------------------------------------------------------ |
| **General**                                                  |
| `harborAdminPassword`                                        |
| `secretkey`                                                  |
| **Persistence**                                              |
| `persistence.imageChartStorage.azure.accountkey`             |
| `persistence.imageChartStorage.gcs.encodedkey`               |
| `persistence.imageChartStorage.s3.accesskey`                 |
| `persistence.imageChartStorage.s3.secretkey`                 |
| `persistence.imageChartStorage.swift.password`               |
| `persistence.imageChartStorage.oss.accesskeysecret`          |
| **Core**                                                     |
| `core.secret`                                                |
| **Jobservice**                                               |
| `jobservice.secret`                                          |
| **Registry**                                                 |
| `registry.secret`                                            |
| **PostgreSQL**                                               |
| `externalDatabase.password`                                  |
| **Redis**                                                    |
| `externalRedis.password`                                     |

There's a [PR](https://github.com/goharbor/harbor-helm/pull/199) for the `harbor-helm` repo that partially implements this functionality.

BronzeDeer:
**Description of the change**

This pull request allows setting each secret value to be read from a Kubernetes secret instead of being specified as plaintext value or randomly generated on each chart inflation.

For each value `foo.bar` a new value `foo.barSecret` is introduced. `barSecret` is a dictionary with two keys: `name` and `key`. `key` is set to a reasonable default, while `name` is set to an empty string. If `name` is set to a non-empty string, the normal generation of the value is suppressed and the corresponding environment variable is directly read from the secret key indicated by `name` and `key`

**Benefits**

This makes the helm chart suitable for a GitOps Approach with a tool like ArgoCD or Flux. In such an approach the complete   configuration for infrastructure is stored in git, with the final plaintext secret value only manifesting as a Kubernetes Secret after deployment. (Provided by e.g. a SealedSecret or ExternalSecret). For values that do not need to be used outside of the cluster but which need to be stable over multiple inflations of the template, a tool like Carvel SecretGen Controller can be used to provide such a random, but stable value in a Kubernetes Secret declaratively.

**Possible drawbacks**

- the redis password is specified in url for GC, Trivy, Core, and Clair; a more thorough change to the way the chart works might be needed here. If the components cannot accept the password in a separate environment variable, sidecars might be needed to support constructing the url at runtime. Something similar is true of the postgres credentials for notary, here a json file would have to be templated at run time.
- the `foo.barSecret` pattern creates some akwardness next to values like `core.secret` and `core.secretName`, however in the interest of consistency, the pattern has been applied equally.

**Applicable issues**

  - fixes #2010

**Additional information**

To aid in reviewing, I have created a single commit per value, however they should probably squashed before the merge to limit commit log pollution

**Checklist** 


BronzeDeer:
Am I correct that I will not manually need to change the README and can instead use the generator after all changes werde made?

BronzeDeer:
Fixed some small typos/oversights pointed out by the linter

pablogalegoc:
Hi @BronzeDeer! Thanks for your contribution!

There are some parts about this that I would like to discuss with the team before starting to ask you for changes. My main concerns are:
- IMHO this approach adds a bit too much complexity to the chart for the new feature. We will allow mixing envvars from secrets created by the chart with envvars from external secrets.
- In other charts we have the convention of adding an "existingSecret" option. It is less flexible but more simple. Basically

BronzeDeer:
Missed two instances in which secrets were mounted instead of being used as env var.

P.S.: @pablogalegoc, Now that you have engaged with the change, do you prefer I squash those kind of fixups early or late?

pablogalegoc:
After discussing with the team, we think the approach should be using the "existingSecret" convention. Basically, we condition the creation of the built-in secret to the absence of a declared existing one. You can see an example of such condition here:
https://github.com/bitnami/charts/blob/afcf09ef8adc6167b11d65978758c5d2bbcabf3a/bitnami/harbor/templates/internal/internal-crt-secret.yaml#L4

We are aware that it implies that it implies that the external secret has to follow the same structure that the built-in one, but we think it is a small price to pay for simplicity. It has been working for us, but if you think it falls short for your use case, let us know your thoughts. 

> Now that you have engaged with the change, do you prefer I squash those kind of fixups early or late?

Don't worry about the commits, in the end we always squash and merge.

davgia:
Any update on this? This would be very useful.

Roia5:
Would also like an update on why (and if) this was abandoned.
I am attempting to use S3 in a GitOps manner (ArgoCD). I cannot set my access key in plain text in the values.yaml file due to security.
Couldn't find any workaround to setting a Kubernetes Secret with the access key.

PR #8180, opened 2021-11-18
[bitnami/etcd] ETCD_LISTEN_* to use container ports instead of svc ports

juan131:
Signed-off-by: juan131 <juanariza@vmware.com>

**Description of the change**

We're using `.Values.service.*` parameters to set the the _ETCD_LISTEN_*_  environment variables. This is conceptually wrong since these parameters refer to the etcd service port instead of the container ports. We should use `.Values.containerPorts.*` parameters to customize container ports.

**Benefits**

Correct use of `.Values.service.*` and `.Values.containerPorts.*` parameters.

**Possible drawbacks**

None

**Applicable issues**

N/A

**Additional information**

N/A

**Checklist** 

- [x] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [x] Variables are documented in the README.md
- [x] Title of the PR starts with chart name (e.g. [bitnami/<name_of_the_chart>])

PR #8189, opened 2021-11-19
[bitnami/etcd] Fix etcd when using a custom svc port

juan131:

**Description of the change**

We recently adapted the chat so you can use a port different from 2379 as the "service port". However, this isn't working as expected since the headless svc should've been adapted to use the container port too (since this svc is used to generate unique FQDN to access specific etcd pods).

Also, the `ETCD_ADVERTISE_CLIENT_URLS` env. var has been adapted so it advertised both the URL to access etcd through the service and the URL to access etcd by addressing the pod's hostname directly.

**Benefits**

User can set a custom svc port

**Possible drawbacks**

None

**Applicable issues**

N/A

**Additional information**

Follow up from https://github.com/bitnami/charts/pull/8180

**Checklist** 

### DIFF
--- a/bitnami/etcd/Chart.yaml
+++ b/bitnami/etcd/Chart.yaml
@@ -24,4 +24,4 @@ name: etcd
 sources:
   - https://github.com/bitnami/bitnami-docker-etcd
   - https://coreos.com/etcd/
-version: 6.10.2
+version: 6.10.3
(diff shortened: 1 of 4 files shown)


miguelaeh (approved):
LGTM!

FraPazGal (approved):

PR #8192, opened 2021-11-19
[bitnami/tomcat] Add networkpolicy support

bbaudelet:
**Description of the change**

Add networkpolicy support on tomcat chart. 
Networkpolicy control access to containerPort and containerExtraPorts, if metrics is enabled flow are not filtered and open from any source.

If the chart is deployed on a K8S cluster where default rule is deny all, we need to deployed a networkpolicy but it can be tedious. we want to have a chart allign with other ones to integrate a default netwokpolicy.

**Possible drawbacks**

By default network policy is disable, but when it's enabled some change must be done to be work fine.
the good label and sometime the right namespaces selector must be added to each pod, want to reach the tomcat.

**Applicable issues**

  - fixes #

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/tomcat/Chart.yaml
+++ b/bitnami/tomcat/Chart.yaml
@@ -26,4 +26,4 @@ name: tomcat
 sources:
   - https://github.com/bitnami/bitnami-docker-tomcat
   - http://tomcat.apache.org
-version: 9.5.4
+version: 9.6.0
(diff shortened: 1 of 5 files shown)


Mauraza (approved):
Hi @bbaudelet,

Thanks for the PR! LGTM!

PR #8199, opened 2021-11-19
[bitnami/grafana] Add .Values.commonAnnotations to more templates

wileyj:
**Description of the change**

This PR adds`.Values.commonAnnotations` to more templates in the grafana chart. 
https://github.com/bitnami/charts/tree/master/bitnami/nginx-ingress-controller was used as a model

**Benefits**

Use case: adding an annotation like [reloader](https://github.com/stakater/Reloader) becomes possible with this change, where a deployment of grafana will restart the pods if there is a detected change in a defined configmap/secret. 

**Possible drawbacks**

Similar to the nginx-ingress-controller chart above, it can add annotations to resources where they're not needed, but has no impact on the resource created. 

**Checklist** 

### DIFF
--- a/bitnami/grafana/Chart.yaml
+++ b/bitnami/grafana/Chart.yaml
@@ -24,4 +24,4 @@ name: grafana
 sources:
   - https://github.com/bitnami/bitnami-docker-grafana
   - https://grafana.com/
-version: 7.2.2
+version: 7.2.3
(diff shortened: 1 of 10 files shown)


migruiz4 (approved):
Thank you very much for your contribution! Looks good to me!

PR #8228, opened 2021-11-24
[bitnami/spring-cloud-dataflow] Add new variable .Values.server.configuration.defaultSpringApplicationJSON

yadkov:
**Description of the change**

As a system engineer I would like to inject custom configuration through environment variable SPRING_APPLICATION_JSON (e.g. private maven repo). Currently this is not possible because the value of this environment variable is hardcoded.

New variable server.configuration.defaultSpringApplicationJSON has been introduced. The default value is true.

When server.configuration.defaultSpringApplicationJSON is set to true, the default values for the environment variable SPRING_APPLICATION_JSON are set:
```
{
  "maven": {
    "local-repository": null,
    "remote-repositories": {
      "repo1": {
        "url": "https://repo.spring.io/libs-snapshot"
      }
    }
  }
}
```
When server.configuration.defaultSpringApplicationJSON is set to false the default configuration is omitted.

**Benefits**
When server.configuration.defaultSpringApplicationJSON is set to true there will be no change in the helm configuration.

When server.configuration.defaultSpringApplicationJSON is set to false custom values for SPRING_APPLICATION_JSON could be set up through config map or secret.

**Possible drawbacks**

When server.configuration.defaultSpringApplicationJSON is set to false the configuration described in .Values.server.proxy.* won't be taken into account. The user should manually configure the proxy in the custom values for SPRING_APPLICATION_JSON.

**Checklist** 


bitnami/spring-cloud-dataflow/Chart.yaml
```diff
@@ -39,4 +39,4 @@ sources:
   - https://github.com/bitnami/bitnami-docker-spring-cloud-dataflow
   - https://github.com/bitnami/bitnami-docker-spring-cloud-skipper
   - https://dataflow.spring.io/
-version: 4.1.6
+version: 4.1.7
```
rafariossaa, line 42: As you are introducing a new parameter, instead of the patch version the minor version should be increased. Hence this should be `4.2.0`

rafariossaa (changes requested):
Hi,
  Thanks you very much for providing this functionality to the chart.

bitnami/spring-cloud-dataflow/Chart.yaml
yadkov: Hi @rafariossaa 

Thanks for the feedback - fixed.

Issue #8172, opened 2021-11-17
[thanos] CA should be optional when doing client TLS


ArchiFleKs:
**Which chart**:
thanos: > 8.0

**Describe the bug**

I did not had any issue with client TLS and mTLS when running version <8.

specifying CA used to be optionnal [here](https://github.com/bitnami/charts/blob/89d9d7df53e89f8657c907a13b9c4fdb144f9512/bitnami/thanos/templates/query/tls-client-secret.yaml#L24) and [now it set as empty]() when unspecified but the [flag is still in used](https://github.com/bitnami/charts/blob/master/bitnami/thanos/templates/query/deployment.yaml#L144) and point to an empty file specified in the secret.

This behavior seems to break mTLS, I'm using an Nginx ingress secure by Let's encrypt on the other hand, and it was working fine without specifying the CA, when I remove the flag, my TLS querier are working as expected.

I think specifying a CA should be optional like before.

**Expected behavior**
TLS client is working without having to specify CA, only Key and Cert

**Version of Helm and Kubernetes**:

Kubernetes 1.21 and terraform-provider-helm 2.4.1

javsalgar:
Hi,

I agree that, if it is possible to make it work without CA, it should continue as it is. As you discovered the issue, would you like to submit a PR with the fix? If not, we will work on it.

ArchiFleKs:
> Hi,
> 
> I agree that, if it is possible to make it work without CA, it should continue as it is. As you discovered the issue, would you like to submit a PR with the fix? If not, we will work on it.

I'll try to make a PR shortly

javsalgar:
Thank you!!

rlees85:
Its a shame this is closed as stale. I have the same problem and am tempted to submit a PR to work around it. I cannot think of a 'clean' approach to do this though.

Maybe something like a `systemCa` flag under `client` that is default to `false` (to keep existing behaviour). Then we can set it to `true` to skip adding the extra command line for the client CA certificate.

javsalgar:
Hi!

If you had a PR in mind, do not hestitate to send it and we can iterate on it :D

PR #8230, opened 2021-11-24
[bitnami/thanos] Fix autogenerated TLS key values

quadeare:
**Description of the change**

This change just fix TLS autogenerated key values on secret files. For now, when we use autogenerated TLS, some pods won't start because TLS key doesn't exist.

**Benefits**

Autogenerated TLS is able to work now.

**Possible drawbacks**

Maybe a new PR is mandatory to fix CA issue : https://github.com/bitnami/charts/issues/8172

**Checklist** 

- [ X] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [ X] Title of the PR starts with chart name (e.g. [bitnami/<name_of_the_chart>])

### DIFF
--- a/bitnami/thanos/Chart.yaml
+++ b/bitnami/thanos/Chart.yaml
@@ -28,4 +28,4 @@ name: thanos
 sources:
   - https://github.com/bitnami/bitnami-docker-thanos
   - https://thanos.io
-version: 8.1.0
+version: 8.1.1
--- a/bitnami/thanos/Chart.yaml
+++ b/bitnami/thanos/Chart.yaml
@@ -28,4 +28,4 @@ name: thanos
 sources:
   - https://github.com/bitnami/bitnami-docker-thanos
   - https://thanos.io
-version: 8.1.1
+version: 8.1.3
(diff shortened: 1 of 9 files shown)


Mauraza (commented):
Hi  @quadeare,

Thanks for the PR.
The rest of the charts are using `tls.crt`, `tls.key` and `ca.crt`,  could you do the change using these instead of `tls-cert`, `tls-key`, and `ca-cert`?

quadeare:
Hi @Mauraza,

I just push my latest modifications with `tls.crt`, `tls.key` and `ca.crt` keys.
It seems to be good on my side, can you take a look ?

Than you so much ;)

miguelaeh (changes requested):
Hi guys,
this change is not so trivial as it seems. This will imply a backward incompatibility for people using `existingSecrets`.
Please reconsider the version bump.
@quadeare  could you please elaborate on why this change is needed?

quadeare:
Hi @miguelaeh,

Autogenerated TLS isn't working well because `tls-cert`, `tls-key`, and `ca-cert` or not found.
Naively, I thought that by changing all the keys it would work properly.

Since `autogenerated` does not work i use `existingSecrets` and it works well.

miguelaeh:
Hi @quadeare ,
It seems I was wrong, it is currently implemented in two different ways. It is using `tls-` in the case of `existingSecrets` and `tls.` in the case of autogenerated secret.
But in both cases, the config map is created by the chart, so I think changing it will be automatically upgraded.
Honestly, I think this must be changed to use a single approach in both cases, but I would like to ensure this won't break the `helm upgrade`.

PR #8236, opened 2021-11-24
[bitnami/kafka] allowPlaintextListener is now documented properly

JeremyTheocharis:
**Description of the change**

If you look the first time at `allowPlaintextListener` you get the impression that it is activated (because it is on true). However, because it is defined like this:

https://github.com/bitnami/charts/blob/c4e5c940d480211dd6cc50abaea33d488e8cde1d/bitnami/kafka/templates/statefulset.yaml#L228

it is actually to "false" / "no". I changed "true" to "no".

**Benefits**

The end user knows that:
1. the default behaviour is to disallow Plaintext Listener
2. ne needs to change it to yes

**Possible drawbacks**

no drawbacks known

**Applicable issues**

no applicable issues

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/kafka/Chart.yaml
+++ b/bitnami/kafka/Chart.yaml
@@ -29,4 +29,4 @@ name: kafka
 sources:
   - https://github.com/bitnami/bitnami-docker-kafka
   - https://kafka.apache.org/
-version: 14.4.1
+version: 14.4.2
--- a/bitnami/kafka/README.md
+++ b/bitnami/kafka/README.md
@@ -147,7 +147,7 @@ The command removes all the Kubernetes components associated with the chart and 
 | `listeners`                                | The address(es) the socket server listens on. Auto-calculated it's set to an empty array                                                             | `[]`                                |
 | `advertisedListeners`                      | The address(es) (hostname:port) the broker will advertise to producers and consumers. Auto-calculated it's set to an empty array                     | `[]`                                |
 | `listenerSecurityProtocolMap`              | The protocol->listener mapping. Auto-calculated it's set to nil                                                                                      | `""`                                |
-| `allowPlaintextListener`                   | Allow to use the PLAINTEXT listener                                                                                                                  | `true`                              |
+| `allowPlaintextListener`                   | Allow to use the PLAINTEXT listener                                                                                                                  | `no`                              |
 | `interBrokerListenerName`                  | The listener that the brokers should communicate on                                                                                                  | `INTERNAL`                          |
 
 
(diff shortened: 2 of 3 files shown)


carrodher (changes requested):
Thanks for your PR, although it seems the current behavior is the desired one.

By default, when `allowPlaintextListener` is not modified or set to `true` (default value) the `ALLOW_PLAINTEXT_LISTENER` environment variable is set to "yes"
```console
$ helm template kafka bitnami/kafka | grep ALLOW_PLAINTEXT_LISTENER -A 1
            - name: ALLOW_PLAINTEXT_LISTENER
              value: "yes"
```
on the other hand, when `allowPlaintextListener` is set to `false` the value of the `ALLOW_PLAINTEXT_LISTENER` env. variable is "no":
```console
$ helm template kafka bitnami/kafka --set allowPlaintextListener=false | grep ALLOW_PLAINTEXT_LISTENER -A 1
            - name: ALLOW_PLAINTEXT_LISTENER
              value: "no"
```

This is how the `ternary` operator works: 

> The ternary function takes two values, and a test value. If the test value is true, the first value will be returned. If the test value is empty, the second value will be returned. This is similar to the ternary operator in C and other programming languages

JeremyTheocharis:
> Thanks for your PR, although it seems the current behavior is the desired one.
> 
> By default, when `allowPlaintextListener` is not modified or set to `true` (default value) the `ALLOW_PLAINTEXT_LISTENER` environment variable is set to "yes"
> 
> ```
> $ helm template kafka bitnami/kafka | grep ALLOW_PLAINTEXT_LISTENER -A 1
>             - name: ALLOW_PLAINTEXT_LISTENER
>               value: "yes"
> ```
> 
> on the other hand, when `allowPlaintextListener` is set to `false` the value of the `ALLOW_PLAINTEXT_LISTENER` env. variable is "no":
> 
> ```
> $ helm template kafka bitnami/kafka --set allowPlaintextListener=false | grep ALLOW_PLAINTEXT_LISTENER -A 1
>             - name: ALLOW_PLAINTEXT_LISTENER
>               value: "no"
> ```
> 
> This is how the `ternary` operator works:
> 
> > The ternary function takes two values, and a test value. If the test value is true, the first value will be returned. If the test value is empty, the second value will be returned. This is similar to the ternary operator in C and other programming languages

Thank you for your reply and explaination! Then we need to continue debugging to find why we recieve the error issue:
https://github.com/united-manufacturing-hub/united-manufacturing-hub/issues/697

WIll update as soon as I have more information.

PR #8242, opened 2021-11-25
[bitnami/elasticsearch] Extended security context

mFranz82:
**Description of the change**

The current version of this chart does not allow custom *securityContext* settings. Only *fsGroup* as well as *runAsUser*  could be defined individual.  

This change now introduces two new properties:

- ``` [master/data/ingest/coordinating].podSecurityContext ``` 
- ``` [master/data/ingest/coordinating].containerSecurityContext ``` 

which are used to replace and extend the currently used:

- ``` [master/data/ingest/coordinating].securityContext ``` 

Both are used to override the corresponding ```securityContext``` and are **disabled** by default. ```fsGroup```and ```runAsUser``` is set to the expected default values. Both properties accept custom yaml.

**Note:** Remove ``` securityContext ```when bumping this chart to 18.0.0 !

**Benefits**

No limitations anymore according to securityContext settings 

**Possible drawbacks**

As these properties are disabled by default nothing would happen if a chart update occurs. 

**Applicable issues**

**Additional information**

**Checklist** 


marcosbc (changes requested):
Thanks for the contribution! In this case I think it would make sense to standarize `podSecurityContext` and `containerSecurityContext` like it is used in other charts.

In other charts we are allowing users to add any option to `podSecurityContext` and `containerSecurityContext`. And then, in the deployment/statefulset, we add the logic below:

```
{{- end }}
```

That way, there isn't any need to add any extra parameter and it is also a lot simpler for users.

mFranz82:
> Thanks for the contribution! In this case I think it would make sense to standarize `podSecurityContext` and `containerSecurityContext` like it is used in other charts.
> 
> In other charts we are allowing users to add any option to `podSecurityContext` and `containerSecurityContext`. And then, in the deployment/statefulset, we add the logic below:
> 
> ```
> {{- if .Values.podSecurityContext.enabled }}
> securityContext: {{- omit .Values.podSecurityContext "enabled" | toYaml | nindent 8 }}
> {{- end }}
> ```
> 
> That way, there isn't any need to add any extra parameter and it is also a lot simpler for users.

Thanks for your review!
I like your suggestion 👍

miguelaeh (commented):
Hi @mFranz82 ,
Changing the default values of the Chart would require a major change. And, the major change will require a major change of any chart that ships this one as a dependency.

mFranz82:
> Hi @mFranz82 ,
> Changing the default values of the Chart would require a major change. And, the major change will require a major change of any chart that ships this one as a dependency.

Hi @miguelaeh! Sounds understandable to me 👍

mFranz82:
> > Hi @mFranz82 ,
> > Changing the default values of the Chart would require a major change. And, the major change will require a major change of any chart that ships this one as a dependency.
> 
> Hi @miguelaeh! Sounds understandable to me 👍

Looks like the following chars require elasticsearch: **dataplatform-bp2** and **magento**

Should I create a new pull request for each chart?

miguelaeh:
Hi @mFranz82 ,
Let me check with the team if anyone has plans to create a major version for elasticsearch.
We usually group several changes in this kind of major bumps to avoid bumping the major version too many times in several charts.

mFranz82:
> Hi @mFranz82 , Let me check with the team if anyone has plans to create a major version for elasticsearch. We usually group several changes in this kind of major bumps to avoid bumping the major version too many times in several charts.
> 
> The other possibility is to create a backward-compatible layer to be deleted in the next major version. I mean, support both parameters, with the new ones taking precedence over the old ones. I think this would be simpler in this case

I think so too. This was my idea when using ```extraPodSecurityContext``` in the first version of this PR. 
I'll try to implement a nice way and keep you updated

mFranz82:
> Hi @mFranz82 , Let me check with the team if anyone has plans to create a major version for elasticsearch. We usually group several changes in this kind of major bumps to avoid bumping the major version too many times in several charts.
> 
> The other possibility is to create a backward-compatible layer to be deleted in the next major version. I mean, support both parameters, with the new ones taking precedence over the old ones. I think this would be simpler in this case

Hi @miguelaeh 
Can you please have a look?

bitnami/elasticsearch/templates/coordinating-statefulset.yaml
```diff
@@ -65,9 +65,14 @@ spec:
...
       serviceAccountName: {{ template "elasticsearch.coordinating.serviceAccountName" . }}
-      {{- if .Values.coordinating.podSecurityContext.enabled }}
+      {{- if or .Values.coordinating.podSecurityContext.enabled .Values.coordinating.securityContext.enabled }}
       securityContext:
+        {{- if .Values.coordinating.securityContext.enabled }}
+        fsGroup: {{ .Values.coordinating.securityContext.fsGroup }}
+        {{- end }}
+        {{- if .Values.coordinating.podSecurityContext.enabled }}
         {{- omit .Values.coordinating.podSecurityContext "enabled" | toYaml | nindent 8 }}
+        {{- end }}
```
miguelaeh: I would make the `posdSecurityContext` to take precedence over the `securityContext`.
If not, it could duplicate fields.
I mean, if `podSecurityContext` is is set take it, if not, check the `securityContext`. While leaving the new `podSecurityContext` disabled by default, so it will continue working as in the past by default.
This applies to the rest of the files too.

miguelaeh (commented):
thank you very much for the fast implementation

bitnami/elasticsearch/templates/coordinating-statefulset.yaml
mFranz82: Right. 
What about this?
```
          {{- if or .Values.master.containerSecurityContext.enabled .Values.master.securityContext.enabled }}
          securityContext:
            {{- if .Values.master.containerSecurityContext.enabled }}
            {{- omit .Values.master.containerSecurityContext "enabled" | toYaml | nindent 12 }}
            {{ else }}
            runAsUser: {{ .Values.master.securityContext.runAsUser }}
            {{- end }}
          {{- end }}
         ```

miguelaeh: 👍

miguelaeh (approved):
LGTM!

PR #8246, opened 2021-11-26
[bitnami/ghost] Fix host when using ingress with external TLS termination

mwoelk:
**Description of the change**

Gives `ghostHost` precedence over `ingress.hostname` when defining the `ghost.host` variable helper.

**Benefits**

This way, the host can be set independently / override the hostname from ingress. 
This is necessary when using external TLS termination. In this case, we need to provide a `https://`-prefix for the host which is not possible through an ingress hostname.

**Possible drawbacks**

none

**Applicable issues**

none

**Additional information**

none

**Checklist** 

### DIFF
--- a/bitnami/ghost/Chart.yaml
+++ b/bitnami/ghost/Chart.yaml
@@ -33,4 +33,4 @@ name: ghost
 sources:
   - https://github.com/bitnami/bitnami-docker-ghost
   - http://www.ghost.org/
-version: 15.2.1
+version: 15.1.3
--- a/bitnami/ghost/Chart.yaml
+++ b/bitnami/ghost/Chart.yaml
@@ -33,4 +33,4 @@ name: ghost
 sources:
   - https://github.com/bitnami/bitnami-docker-ghost
   - http://www.ghost.org/
-version: 15.1.2
+version: 15.2.1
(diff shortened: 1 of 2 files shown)


mwoelk:
Bumped the minor version instead of patch version because the precedence change might require a check to see if `ghostHost` is set correctly if both `ghostHost` and `ingress.hostname` are provided.

mwoelk:
Problem is that this messes with the live/readiness check. As they are performed via http but Ghost redirects to https.
Maybe it's a solve to change live/readiness to a tcp check on the http port?

bitnami/ghost/Chart.yaml
```diff
@@ -33,4 +33,4 @@ name: ghost
 sources:
   - https://github.com/bitnami/bitnami-docker-ghost
   - http://www.ghost.org/
-version: 15.1.2
+version: 15.2.1
```
miguelaeh, line 36:
```suggestion
version: 15.1.3
```

miguelaeh (changes requested):
Thank you for the PR!

miguelaeh (changes requested):
Hi, 
It seems you need to bump the version again after rebasing master.

PR #8251, opened 2021-11-26
[bitnami/grafana] fix volumePermissions command

### LINKED ISSUE #8231: [bitnami/grafana] Fix volumePermissions initContainer
**Which chart**:
bitnami/grafana  (8.2.5)

~~**Is your feature request related to a problem? Please describe.**~~
**Describe the bug**
Since the Bitnami containers (by default, but rightfully so) don't run as root, containers can have trouble writing to mounted volumes. The recommended approach is to set the `securityContext` [[0]](https://docs.bitnami.com/kubernetes/faq/configuration/use-non-root/) [[1]](https://github.com/bitnami/charts/issues/5568#issuecomment-897419540). However, there are cases where this may not work, which is why many Bitnami charts have implemented a `volumePermissions` template, creating an (root) initContainer to write the appropriate file permissions [[2]](https://docs.bitnami.com/general/how-to/troubleshoot-helm-chart-issues/#solve-issue-222).

Two issues are present:
1. The volumePermissions template looks up `.Values.containerSecurityContext.runAsUser` and `.Values.podSecurityContext.fsGroup` but these do not exist in the Grafana values.yaml.
2. The initContainer tries to `chown` the grafana directory but fails if it does not exist.

~~**Describe the solution you'd like**~~
~~Please also add the `volumePermissions`  to the Grafana chart. It can probably be copied from other charts like [mongodb](https://github.com/bitnami/charts/blob/master/bitnami/mongodb/templates/replicaset/statefulset.yaml#L92) and [influxdb](https://github.com/bitnami/charts/blob/5c9f6e3e46921dcc2aead5676a8b9dcb8e62ece8/bitnami/influxdb/templates/influxdb/deployment-standalone.yaml#L58).~~

~~**Describe alternatives you've considered**~~
~~Only using `securityContext` is not sufficient as not all platforms and storage classes support it (which is what I am running into).~~

toonalbers:
**Description of the change**

Fix volumePermissions initContainer command. It now creates the directory if it does not exist (like the other charts), and uses correct lookup values.

**Possible drawbacks**

Originally the code looked at `.Values.containerSecurityContext.runAsUser` and `.Values.podSecurityContext.fsGroup` but these do not exist in the Grafana values.yaml. I have now redirected the code to `.Values.grafana.{container,pod}SecurityContext`. A neater approach might be to move the settings from `.grafana` to the root.

Another difference between this chart and others, is that `grafana:podSecurityContext:runAsNonRoot` is also applied to the initContainer. So in my case at least this had to be turned off, and I have documented this.

I'm not very satisfied with this behavior, but I also don't want to change the existing chart too much myself.

**Applicable issues**

  - fixes #8231

**Checklist** 


nc-ruth:
Hi, what is the status of this?
We are experiencing issues which gets resolved by this PR.

bitnami/grafana/templates/deployment.yaml
```diff
@@ -79,7 +79,12 @@ spec:
           image: {{ include "volumePermissions.image" . }}
           imagePullPolicy: {{ .Values.volumePermissions.image.pullPolicy | quote }}
           command:
-            - chown -R {{ .Values.containerSecurityContext.runAsUser }}:{{ .Values.podSecurityContext.fsGroup }} /bitnami/grafana
+            - /bin/bash
+            - '-ec'
+          args:
+            - |
+              mkdir -p /bitnami/grafana
```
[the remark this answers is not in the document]
toonalbers: It should be mounted, but without this change we get this error:
```
failed to create shim: OCI runtime create failed:
container_linux.go:380: starting container process caused: exec:
"chown -R 1001:1001 /bitnami/grafana": stat chown -R 1001:1001
/bitnami/grafana: no such file or directory: unknown
```

I just experimented some more, and it also works without `mkdir -p`, however for some reason it does not work directly as a `command`, it needs to be executed in bash to avoid the error. I can't think of any reason for that.

bitnami/grafana/templates/deployment.yaml
```diff
@@ -79,7 +79,12 @@ spec:
...
+              chown -R {{ .Values.grafana.containerSecurityContext.runAsUser }}:{{ .Values.grafana.podSecurityContext.fsGroup }} /bitnami/grafana
```
juan131, line 87: We should skip some hidden files that comes with the mounted volumes:
```suggestion
              find /bitnami/grafana -mindepth 1 -maxdepth 1 -not -name ".snapshot" -not -name "lost+found" | xargs chown -R {{ .Values.grafana.containerSecurityContext.runAsUser}}:{{ .Values.grafana.podSecurityContext.fsGroup }}
```

bitnami/grafana/templates/deployment.yaml
```diff
@@ -79,7 +79,12 @@ spec:
           image: {{ include "volumePermissions.image" . }}
           imagePullPolicy: {{ .Values.volumePermissions.image.pullPolicy | quote }}
           command:
-            - chown -R {{ .Values.containerSecurityContext.runAsUser }}:{{ .Values.podSecurityContext.fsGroup }} /bitnami/grafana
+            - /bin/bash
+            - '-ec'
+          args:
+            - |
```
juan131, line 85: Actually the `-ec` are part of the args:
```suggestion
          args:
            - -ec
            - |
```

juan131 (approved):
LGTM

bitnami/grafana/templates/deployment.yaml
toonalbers: Yes good point! I suppose this could also be changed in the other Bitnami charts.

PR #8261, opened 2021-11-29
[bitnami/rabbitmq-cluster-operator] Fix reference to metrics ports in values.yaml and metrics service

ekeih:
**Description of the change**

With the addition of the messaging-topology-operator the path to the metrics
ports changed. This change was still missing in the values.yaml, causing a
"nil pointer evaluating interface" error when rendering the chart with enabled
metrics.

One of the metrics services was also still using the old path.

**Benefits**

The chart renders correctly when enabling metrics.

**Possible drawbacks**

If the user uses a different name for the port instead of the default `http` or a `nodePort`, they will also have to adjust the annotation. They have to to that anyway though, as for example a hardcoded value would also need manual adjustment by the user.

**Applicable issues**

None.

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/rabbitmq-cluster-operator/Chart.yaml
+++ b/bitnami/rabbitmq-cluster-operator/Chart.yaml
@@ -25,4 +25,4 @@ name: rabbitmq-cluster-operator
 sources:
   - https://github.com/bitnami/bitnami-docker-rabbitmq-cluster-operator
   - https://github.com/rabbitmq/cluster-operator
-version: 2.0.1
+version: 2.0.2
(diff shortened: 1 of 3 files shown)


javsalgar (approved):
Thank you so much! LGTM

ekeih:
@javsalgar I think this needs another version bump to .3? 🤔 Originally I put .2 in here and .3 in https://github.com/bitnami/charts/pull/8264, but @carrodher already merged https://github.com/bitnami/charts/pull/8264 and set the version there to .2, so if I suspect that this merge here might not trigger a new release? 🤔

javsalgar:
Hi,

Let me force a bump so our system updates the chart.

PR #8272, opened 2021-11-30
[bitnami/zookeeper] remove common annotations from volumeClaimTemplates

itscaro:
**Description of the change**

I think common annotations should not be used in volumeClaimTemplates, as in tomcat or kafka.
In my case, I add pipeline info to common annotations, these metadata are variables so between 2 pipelines, sts cannot be patched due to volumeClaimTemplates' annotation changes.

**Benefits**

**Possible drawbacks**

**Applicable issues**

  - fixes #

**Additional information**

**Checklist** 


itscaro:
Should we remove commonLabels too?

carrodher:
Thanks for creating this PR, I am not sure if this change has sense because the `commonAnnotations` are defined as "Annotations to add to all deployed objects". Maybe you need to use a different approach for your use case instead of using the `commonAnnotations` as they are defined nowadays.

itscaro:
I'm not quite in line with you. I don't think that volumeClaimTemplate should receive common annotations, in many other charts, it's is the case, and another key persistence.annotations is reserved to add annotations to volumeClaimTemplate. I would like to add metadata about the pipeline, the triggerer into deployed objects, but not necessarily auto-provisionned objects like PVs.

https://github.com/bitnami/charts/blob/master/bitnami/kafka/templates/statefulset.yaml#L530-L535
https://github.com/bitnami/charts/blob/master/bitnami/tomcat/templates/statefulset.yaml#L31-L33
https://github.com/bitnami/charts/blob/master/bitnami/postgresql/templates/statefulset.yaml#L616-L624

javsalgar:
Hi,

As @carrodher said, I have my concerns with this. In principle, I would like to have common annotations to all objects and also the auto-generated. At the same time, I know that StatefulSets are a bit painful because of the immutable fields. I can easily see a conflict here. In this case, for consistency reasons (even though there are charts were commonAnnotations are not set we will add them when applying standardizations) I would recommend to not use commonAnnotations and use the individual annotation sections. 

Another option that comes to mind would be allowing some sort of exception list to the commonAnnotations, but I'm not sure how this would be implemented.

itscaro:
As I understand, standardizations mean charts with only `persistence.annotations` for now in volumeClaimTemplate will have `commonAnnotations` added together?

This information helps me preparing the breaking changes, since all my charts with statefulsets still use commonAnnotations to add metadata, I just make an exception to zookeeper to use podAnnotations which is not ideal because the annotations are not at top-level at Statefulset

juan131:
> As I understand, standardizations mean charts with only persistence.annotations for now in volumeClaimTemplate will have commonAnnotations added together?

Yes

> This information helps me preparing the breaking changes, since all my charts with statefulsets still use commonAnnotations to add metadata, I just make an exception to zookeeper to use podAnnotations which is not ideal because the annotations are not at top-level at Statefulset.

I think we should improve the commonLabels/commonAnnotations definitions in the values metadata to make it very clear that these affect to every object (top-level and auto-generated ones). This will help to prevent confusion.

itscaro:
https://github.com/bitnami/charts/blob/f8238a4/bitnami/zookeeper/templates/statefulset.yaml#L491-L501
https://github.com/bitnami/charts/blob/f8238a4/bitnami/kafka/templates/statefulset.yaml#L573-L578

@carrodher @javsalgar @juan131 Nearly one year after, there is still difference between charts at VolumeClaimTemplate. Could we align zookeeper with other charts' approche of using `persistence.annotations` and `persistence.labels` in order to eliminate this difference?

carrodher:
Thanks for reporting this issue. Would you like to contribute by creating a PR to solve the issue? The Bitnami team will be happy to review it and provide feedback. [Here](https://github.com/bitnami/charts/blob/main/CONTRIBUTING.md) you can find the contributing guidelines.

PR #8285, opened 2021-12-01
[bitnami/kafka] Add value zookeeperChrootPath

yevgeny-z:
**Description of the change**

Configure zookeeper chroot path to allow use many instance of kafka in one namespace (network).

**Benefits**

Correct work many instance of kafka and zookeeper in one namespace

**Possible drawbacks**

no

**Applicable issues**

no

**Additional information**

Additional docs: https://docs.confluent.io/platform/current/installation/docker/config-reference.html#required-crest-long-settings

**Checklist** 


bitnami/kafka/values.yaml
```diff
@@ -198,6 +198,9 @@ socketSendBufferBytes: 102400
 ## @param zookeeperConnectionTimeoutMs Timeout in ms for connecting to Zookeeper
 ##
 zookeeperConnectionTimeoutMs: 6000
+## @param zookeeperChrootPath Path which puts data under some path in the global ZooKeeper namespace
+##
+zookeeperChrootPath: ""
```
miguelaeh, line 203: Could you improve the paramter description? Why is it called `chroot`?

bitnami/kafka/templates/statefulset.yaml
```diff
@@ -177,7 +177,7 @@ spec:
                   fieldPath: metadata.name
             - name: KAFKA_CFG_ZOOKEEPER_CONNECT
               {{- if .Values.zookeeper.enabled }}
-              value: {{ include "kafka.zookeeper.fullname" . | quote }}
+              value: "{{ include "kafka.zookeeper.fullname" . }}{{ printf "%s" (tpl .Values.zookeeperChrootPath .) }}"
```
miguelaeh, line 180: take into account that this is an URL. The port is missing in this way and will probably be appended after the added endpoint.
Also, use the `printf` to render the full string instead of concatenating both directly.

miguelaeh (changes requested):

bitnami/kafka/values.yaml
yevgeny-z: This term and description from kafka documentation: https://kafka.apache.org/documentation/#brokerconfigs_zookeeper.connect

> The server can also have a ZooKeeper chroot path as part of its ZooKeeper connection string which puts its data under some path in the global ZooKeeper namespace. For example to give a chroot path of `/chroot/path` you would give the connection string as `hostname1:port1,hostname2:port2,hostname3:port3/chroot/path`.

I can add an example how it's in the official doc.

bitnami/kafka/templates/statefulset.yaml
yevgeny-z: Current chart don't use port in this string. I can add it in other PR.

printf for full string is committed. Or better to move this logic in function?

bitnami/kafka/values.yaml
juan131: I thinks it's enough if we add this references in the metadata:
```suggestion
## @param zookeeperChrootPath Path which puts data under some path in the global ZooKeeper namespace
## ref: https://kafka.apache.org/documentation/#brokerconfigs_zookeeper.connect
##
zookeeperChrootPath: ""
```

bitnami/kafka/templates/statefulset.yaml
```diff
@@ -177,7 +177,7 @@ spec:
                   fieldPath: metadata.name
             - name: KAFKA_CFG_ZOOKEEPER_CONNECT
               {{- if .Values.zookeeper.enabled }}
-              value: {{ include "kafka.zookeeper.fullname" . | quote }}
+              value: "{{ printf "%s%s" (include "kafka.zookeeper.fullname" .) (tpl .Values.zookeeperChrootPath .) }}"
```
juan131, line 180: You don't need the quotes, use `quote` instead. Also, wouldn't it be necessary to add a forward slash between the hostname and the chroot path?
```suggestion
              value: {{ printf "%s/%s" (include "kafka.zookeeper.fullname" .) (tpl .Values.zookeeperChrootPath .) | quote }}
```

juan131 (changes requested):

bitnami/kafka/templates/statefulset.yaml
yevgeny-z: Quotes was changed but I don't think that necessary to add a forward slash. Because it creates unexpected behavior for path started without forward slash.

juan131 (approved):
LGTM

PR #8306, opened 2021-12-03
[bitnami/keycloak] Disable http(s) service mapping

wassy92x:
**Description of the change**

Added possibility to disable http(s) mapping by setting service.port / service.httpsPort to 0.
The default value 80 and 443 will remain.

**Benefits**

This change is necessary if you like to use the exposed service with nginx-ingress-controller.
Nginx-Ingress will only work, if just one port is set inside the service.

**Possible drawbacks**

None

**Applicable issues**

None

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/keycloak/Chart.yaml
+++ b/bitnami/keycloak/Chart.yaml
@@ -26,4 +26,4 @@ name: keycloak
 sources:
   - https://github.com/bitnami/bitnami-docker-keycloak
   - https://github.com/keycloak/keycloak
-version: 5.2.3
+version: 5.3.0
--- a/bitnami/keycloak/README.md
+++ b/bitnami/keycloak/README.md
@@ -213,8 +213,8 @@ The command removes all the Kubernetes components associated with the chart and 
 | Name                               | Description                                                                                                                      | Value                    |
 | ---------------------------------- | -------------------------------------------------------------------------------------------------------------------------------- | ------------------------ |
 | `service.type`                     | Kubernetes service type                                                                                                          | `LoadBalancer`           |
-| `service.port`                     | Service HTTP port                                                                                                                | `80`                     |
-| `service.httpsPort`                | HTTPS Port                                                                                                                       | `443`                    |
+| `service.port`                     | Service HTTP port. 0 will disable http mapping.                                                                                  | `80`                     |
+| `service.httpsPort`                | HTTPS Port. 0 will disable https mapping.                                                                                        | `443`                    |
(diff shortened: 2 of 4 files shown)


javsalgar:
Hi,

Could you let us know the issue that shows that nginx ingress controller will only work with one port? This sounds like a bug in nginx ingress controller itself.

wassy92x:
I found out that it is no issue with nginx ingress controller.
It's an issue of octant (an web-interface for kubernetes clusters).
Otant shows an error, that the port of the service is invalid.
It seems like the issue is solved in the newest version of octant, so you can decide if you like to merge my pull request or not.

bitnami/keycloak/values.yaml
```diff
@@ -630,9 +630,11 @@ service:
   ##
   type: LoadBalancer
   ## @param service.port Service HTTP port
+  ## 0 will disable http-mapping
```
miguelaeh, line 633: If you want this to appear in the README it has to be added in the line above. Same for the next one

bitnami/keycloak/Chart.yaml
```diff
@@ -26,4 +26,4 @@ name: keycloak
 sources:
   - https://github.com/bitnami/bitnami-docker-keycloak
   - https://github.com/keycloak/keycloak
-version: 5.2.3
+version: 5.3.0
```
miguelaeh, line 29:
```suggestion
version: 5.2.4
```

bitnami/keycloak/README.md
```diff
@@ -213,8 +213,8 @@ The command removes all the Kubernetes components associated with the chart and
...
 | ---------------------------------- | -------------------------------------------------------------------------------------------------------------------------------- | ------------------------ |
 | `service.type`                     | Kubernetes service type                                                                                                          | `LoadBalancer`           |
-| `service.port`                     | Service HTTP port                                                                                                                | `80`                     |
-| `service.httpsPort`                | HTTPS Port                                                                                                                       | `443`                    |
+| `service.port`                     | Service HTTP port. 0 will disable http mapping.                                                                                  | `80`                     |
+| `service.httpsPort`                | HTTPS Port. 0 will disable https mapping.                                                                                        | `443`                    |
```
miguelaeh, line 217: Please, do not edit this manually, use https://github.com/bitnami-labs/readme-generator-for-helm

miguelaeh (changes requested):
thank you for the PR!
I don't see the real sense of the PR, but l added some comments.

>It seems like the issue is solved in the newest version of octant, so you can decide if you like to merge my pull request or not.

If it was an issue with octant I think we can just close this since there are no issues with that logic.

PR #8317, opened 2021-12-05
[bitnami/influxdb] Adds rclone based backup providers

### LINKED ISSUE #8163: [bitnami/influxdb] Backup using rclone
**Which chart**:
bitnami/influxdb

The current implementation for backups only allow google or azure as the providers. This limits us as we prefer to use either mega.nz or minio for backups and restoration.

Use [rclone](https://rclone.org/) as this has options to use about 15+ different providers

NA

**Additional context**
NA

abhi1693:
**Description of the change**

- Removed existing google and azure based providers
- Added the following providers based on rclone
  - Azure Blob Storage
  - Google Cloud Storage
  - Mega.nz
  - Generic S3

**Benefits**

Adds integration possibilities of 20+ providers and simplifies the upload.sh script. Also enables support for uploading to more than one provider easily

**Possible drawbacks**

Breaking change as previous existing providers have been removed

**Applicable issues**

  - fixes #8163 

**Additional information**

**Checklist** 


abhi1693:
@carrodher Please review

bitnami/influxdb/templates/influxdb/configmap-backup.yaml
```diff
@@ -60,19 +60,59 @@ data:
...
     #!/bin/sh
 
     set -e
 
-    gcloud auth activate-service-account --key-file /var/secrets/google/{{ .Values.backup.uploadProviders.google.secretKey }}
-    gsutil -m rsync -r -d {{ .Values.backup.directory }}/ {{ .Values.backup.uploadProviders.google.bucketName }}
-  upload-azure.sh: |-
-    #!/bin/sh
+    # Google Cloud Storage
+    if [[ {{ .Values.backup.uploadProviders.googlecloudstorage.enabled }} == "true" ]]; then
```
javsalgar, line 71: I would generate the proper bash script only if the google cloud storage is enabled, so instead of adding that check in the if instruction I would just render the bash code for Google only if the value is true

javsalgar (changes requested):
Thanks for the PR! Requested changes

bitnami/influxdb/templates/influxdb/configmap-backup.yaml
abhi1693: I was trying to avoid mounting different scripts for different providers to keep the overall code clean and easy to handle. Having a single script with all the checks does exactly. If you notice, I have added other provider related code in the same script.

abhi1693:
@javsalgar Any updates here?

bitnami/influxdb/templates/influxdb/configmap-backup.yaml
juan131: The recommendation by @javsalgar doesn't propose using different scripts, but rendering different bash code blocks depending on the cloud provider. The resulting script that will be added to the ConfigMap will be simpler using this approach.

bitnami/influxdb/values.yaml
```diff
@@ -960,77 +986,63 @@ backup:
...
+      clientId: ""
+      clientSecret: ""
+      projectNum: ""
+      objectACL: ""
+      bucketACL: ""
+      location: ""
+      storageClass: ""
+      token: ""
+      authUrl: ""
+      tokenUrl: ""
```
juan131, line 1024: You're using all these parameters to construct the `uploads.sh` script where all these credentials will be in plaintext. This script is included in the ConfigMap, therefore it's not secure at all.
I'd suggest to create a YAML file **in a Secret** including all these GCS details. This'll help you to simplify the values structure also, because you'll be able to use an approach like the below in the Secret:
```yaml
data:
  {{- if .Values.backup.uploadProviders.googlecloudstorage.enabled }}
  gcs.yaml: {{- omit .Values.backup.uploadProviders.googlecloudstorage "enabled" | toYaml | b64enc | quote }}
  {{- end }}
```
Using this approach any parameter under the `backup.uploadProviders.googlecloudstorage` map (but the "enabled" one) will be included. This way, you can remove any non-mandatory property from this map in the **values.yaml** structure.

juan131: This recommendation can be extended to other cloud providers.

juan131 (changes requested):
Thanks so much @abhi1693 !! Please check my comments when you have a chance.

bitnami/influxdb/values.yaml
abhi1693: Thanks for the review. I'll try to get this done

abhi1693:
It'll take some time to get back to this PR. I'll try to spend some time later next week.

PR #8320, opened 2021-12-06
[bitnami/postgresql] lookup existing secrets before generating random secrets

yisiqi:
**Description of the change**

**Benefits**

**Possible drawbacks**

**Applicable issues**

  - fixes #

**Additional information**

**Checklist** 


yisiqi:
@carrodher would you please take a look and help to approve the workflow

javsalgar:
Hi,

I have concerns with this approach because, in the past, we found compatibility issues with different versions of helm, as well as the fact that lookup requires that a k8s cluster exists. Would this work with a, for example, helm template approach?

yisiqi:
I tried it in another similar scene and it worked well。

_"Remember, Helm should not contact the Kubernetes API server during the helm template or helm install|update|delete|rollback --dry-run, so the lookup function will return an empty list (ie dict) in this case ."_ 
--Quoted from the official helm document
https://helm.sh/docs/chart_template_guide/functions_and_pipelines/

carrodher:
> I tried it in another similar scene and it worked well。
> 
> _"Remember, Helm should not contact the Kubernetes API server during the helm template or helm install|update|delete|rollback --dry-run, so the lookup function will return an empty list (ie dict) in this case ."_ --Quoted from the official helm document https://helm.sh/docs/chart_template_guide/functions_and_pipelines/

In that case, the `else` statement is not going to be executed:
```
{{- $obj := (lookup "v1" "Secret" .Namespace .Name).data -}}
{{- if $obj }}
{{- index $obj .Key | b64dec -}}
{{- else -}}
{{- randAlphaNum $len -}}
{{- end -}}
```
`$obj` is going to be assigned to `foo.data` if _lookup_ returns any value or just `.data` if _lookup_ returns an empty list.

The problem is that we want to ensure the charts work by running `helm template` and then deploying the rendered k8s manifest by using `kubectl`. In this scenario, I'm not sure if this piece of code is going to work

yisiqi:
> The problem is that we want to ensure the charts work by running helm template and then deploying the rendered k8s manifest by using kubectl. In this scenario, I'm not sure if this piece of code is going to work

I haven't thought of such a scenario before, but when I pre-render the K8s manifest, it can generate random passwords normally. 

I assume that in this scenario, the user will not use helm to upgrade the release after deploying the generated manifest file, so it is no problem to modify the manifest file and update it through `kubectl`.

The only risk is that when there is a `secret` with the same name under the target `namespace`, the existing `secret` data may be overwritten. But this is an inherent risk of using `kubectl` to maintain, and it should have nothing to do with whether you use `lookup` or not


```
╭─yisiqi@yisiqideMacBook-Pro ~/Workspaces/github.com/yisiqi/bitnami-charts ‹lookup-existed-secrets›
╰─$ helm template postgresql-stable ./bitnami/postgresql
---
# Source: postgresql/templates/secrets.yaml
apiVersion: v1
kind: Secret
metadata:
  name: postgresql-stable
  labels:
    app.kubernetes.io/name: postgresql
    helm.sh/chart: postgresql-10.13.12
    app.kubernetes.io/instance: postgresql-stable
  namespace: default
type: Opaque
data:
  postgresql-password: "dDFHckZXa2FHNg=="
---
# Source: postgresql/templates/svc-headless.yaml
apiVersion: v1
kind: Service
metadata:
  name: postgresql-stable-headless
  labels:
    app.kubernetes.io/name: postgresql
    helm.sh/chart: postgresql-10.13.12
    app.kubernetes.io/instance: postgresql-stable
...
```

bitnami/postgresql/templates/_helpers.tpl
```diff
@@ -49,6 +49,20 @@ Return the proper Docker Image Registry Secret Names
 {{ include "common.images.pullSecrets" (dict "images" (list .Values.image .Values.metrics.image .Values.volumePermissions.image) "global" .Values.global) }}
 {{- end -}}
 
+{{/*
+Returns a secret if it already in Kubernetes, otherwise it creates
+it randomly.
+*/}}
+{{- define "getOrCreateSecret" }}
```
juan131, line 56:
```suggestion
{{/*
Returns the available value for certain key in an existing secret (if it exists),
otherwise it generates a random value.
*/}}
{{- define "getValueFromSecret" }}
```

bitnami/postgresql/templates/_helpers.tpl
```diff
@@ -58,7 +72,7 @@ Return PostgreSQL postgres user password
 {{- else if .Values.postgresqlPostgresPassword -}}
     {{- .Values.postgresqlPostgresPassword -}}
 {{- else -}}
-    {{- randAlphaNum 10 -}}
+    {{- include "getOrCreateSecret" (dict "Namespace" .Release.Namespace "Name" (include "common.names.fullname" .) "Length" 10 "Key" "postgresql-postgres-password")  -}}
```
juan131, line 75:
```suggestion
    {{- include "getValueFromSecret" (dict "Namespace" .Release.Namespace "Name" (include "common.names.fullname" .) "Length" 10 "Key" "postgresql-postgres-password")  -}}
```

bitnami/postgresql/templates/_helpers.tpl
```diff
@@ -71,7 +85,7 @@ Return PostgreSQL password
 {{- else if .Values.postgresqlPassword -}}
     {{- .Values.postgresqlPassword -}}
 {{- else -}}
-    {{- randAlphaNum 10 -}}
+    {{- include "getOrCreateSecret" (dict "Namespace" .Release.Namespace "Name" (include "common.names.fullname" .) "Length" 10 "Key" "postgresql-password")  -}}
```
juan131, line 88:
```suggestion
    {{- include "getValueFromSecret" (dict "Namespace" .Release.Namespace "Name" (include "common.names.fullname" .) "Length" 10 "Key" "postgresql-password")  -}}
```

bitnami/postgresql/templates/_helpers.tpl
```diff
@@ -84,7 +98,7 @@ Return PostgreSQL replication password
 {{- else if .Values.replication.password -}}
     {{- .Values.replication.password -}}
 {{- else -}}
-    {{- randAlphaNum 10 -}}
+    {{- include "getOrCreateSecret" (dict "Namespace" .Release.Namespace "Name" (include "common.names.fullname" .) "Length" 10 "Key" "postgresql-replication-password")  -}}
```
juan131, line 101:
```suggestion
    {{- include "getValueFromSecret" (dict "Namespace" .Release.Namespace "Name" (include "common.names.fullname" .) "Length" 10 "Key" "postgresql-replication-password")  -}}
```

juan131:
Hi @yisiqi 

> I haven't thought of such a scenario before, but when I pre-render the K8s manifest, it can generate random passwords normally.

So you can safely run `helm template ... | kubectl apply -f -` locally and it won't be affected by your changes, right? I guess it makes sense

juan131 (changes requested):
Improve named template name & description to be more accurate.

juan131 (approved):
LGTM

miguelaeh:
Hi guys, it seems this process was repeated for several charts. Why was that new function added to each chart instead of common?

PR #8330, opened 2021-12-07
[bitnami/keycloak] fix keycloak add values if is external db

hoplitis:
**Description of the change**

This adds environment variables when using an external database. Part of my values file:
```yaml
postgresql:
  enabled: False
externalDatabase:
  host: external-database.example.io
  port: 5432
```
When I use it, I see that it does not set values for the external database. Part of pod logs:
```log
keycloak 06:19:07.75 INFO  ==> Trying to connect to PostgreSQL server postgresql...
cannot resolve host "postgresql": lookup postgresql on 172.20.0.10:53: no such host
cannot resolve host "postgresql": lookup postgresql on 172.20.0.10:53: no such host
cannot resolve host "postgresql": lookup postgresql on 172.20.0.10:53: no such host
```
**Benefits**
This makes it possible to use an external database

**Possible drawbacks**
Knowing the logic of the chart in more detail, you can make the changes better

**Checklist** 

### DIFF
--- a/bitnami/keycloak/Chart.lock
+++ b/bitnami/keycloak/Chart.lock
@@ -1,9 +1,9 @@
 dependencies:
 - name: common
   repository: https://charts.bitnami.com/bitnami
-  version: 1.8.0
+  version: 1.10.3
 - name: postgresql
   repository: https://charts.bitnami.com/bitnami
-  version: 10.9.5
-digest: sha256:8bd7414347de5ca50108cb48c76ab4b2b3ba37103aa7863d7d3d6144a94393b8
-generated: "2021-09-10T16:53:34.527900169Z"
+  version: 10.13.11
+digest: sha256:b8753330b5c641cf86de7e24f5f4069b1989e04dfbe8e11f9639cc894bd58bad
+generated: "2021-12-07T10:56:14.326957+03:00"
--- a/bitnami/keycloak/templates/configmap-env-vars.yaml
+++ b/bitnami/keycloak/templates/configmap-env-vars.yaml
@@ -24,6 +24,12 @@ data:
   KEYCLOAK_DATABASE_NAME: {{ include "keycloak.databaseName" . | quote }}
   KEYCLOAK_DATABASE_USER: {{ include "keycloak.databaseUser" . | quote }}
   {{- end }}
+  {{- if (not .Values.postgresql.enabled) }}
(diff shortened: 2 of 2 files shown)


bitnami/keycloak/templates/configmap-env-vars.yaml
```diff
@@ -24,6 +24,12 @@ data:
   KEYCLOAK_DATABASE_NAME: {{ include "keycloak.databaseName" . | quote }}
   KEYCLOAK_DATABASE_USER: {{ include "keycloak.databaseUser" . | quote }}
   {{- end }}
+  {{- if (not .Values.postgresql.enabled) }}
+  KEYCLOAK_DATABASE_HOST: {{ .Values.externalDatabase.host | quote }}
+  KEYCLOAK_DATABASE_PORT: {{ .Values.externalDatabase.port | quote }}
+  KEYCLOAK_DATABASE_NAME: {{ .Values.externalDatabase.database | quote }}
+  KEYCLOAK_DATABASE_USER: {{ .Values.externalDatabase.user | quote }}
+  {{- end }}
```
juan131, line 32: The named templates defined in the `_helpers.tpl` already returns you the right host/port/database/user depending on whether you're using an external database or not. In case you want to follow with this approach, I'd simply remove the condition in line 21.

bitnami/keycloak/Chart.lock
```diff
@@ -1,9 +1,9 @@
...
-  version: 1.8.0
+  version: 1.10.3
 - name: postgresql
   repository: https://charts.bitnami.com/bitnami
-  version: 10.9.5
-digest: sha256:8bd7414347de5ca50108cb48c76ab4b2b3ba37103aa7863d7d3d6144a94393b8
-generated: "2021-09-10T16:53:34.527900169Z"
+  version: 10.13.11
+digest: sha256:b8753330b5c641cf86de7e24f5f4069b1989e04dfbe8e11f9639cc894bd58bad
+generated: "2021-12-07T10:56:14.326957+03:00"
```
juan131, line 9: This file is automatically update by our pipeline, it's not required that you update it. Instead, please bump the patch version in the `Chart.yaml` file.

juan131 (changes requested):
Thanks for the contribution! Please check my comments when you have a chance.

PR #8332, opened 2021-12-07
[bitnami/apache] Adds a switch to enable/disable the git-repo-syncer sidecar container…

jmConan:
… to save resources on a deployment with no regular updates from the repo.

**Description of the change**

Simple adds an option to disable the sidecar container responsible for the subsequent git pulls within the lifetime of the apache pods. 

**Benefits**

If the deployment is *not* required to update regularly we don't need to start the extra container. Saves Resources on the Cluster.

**Possible drawbacks**

None that I know of.

**Additional information**

If the option is unused the default value is true, so no changes to existing deployments are required. 

**Checklist** 

### DIFF
--- a/bitnami/apache/values.yaml
+++ b/bitnami/apache/values.yaml
@@ -139,6 +139,7 @@ extraPodSpec: {}
 ## @param cloneHtdocsFromGit.enabled Get the server static content from a git repository
 ## @param cloneHtdocsFromGit.repository Repository to clone static content from
 ## @param cloneHtdocsFromGit.branch Branch inside the git repository
+## @param cloneHtdocsFromGit.enableAutoRefresh Enables an automatic git pull with a sidecar container
 ## @param cloneHtdocsFromGit.interval Interval for sidecar container pull from the repository
 ## @param cloneHtdocsFromGit.resources Init container git resource requests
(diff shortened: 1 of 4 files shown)


bitnami/apache/values.yaml
```diff
@@ -139,6 +139,7 @@ extraPodSpec: {}
 ## @param cloneHtdocsFromGit.enabled Get the server static content from a git repository
 ## @param cloneHtdocsFromGit.repository Repository to clone static content from
 ## @param cloneHtdocsFromGit.branch Branch inside the git repository
+## @param cloneHtdocsFromGit.enableAutoRefresh Enables an automatic git pull with a sidecar container
```
juan131, line 142: We need to update the README.md to reflect this new parameter you're introducing. To do so, please use the [readme-generator](https://github.com/bitnami-labs/readme-generator-for-helm) tool that automatically updates the README.md for you based in this metadata.

juan131 (changes requested):
Thanks so much for this contribution! We really appreciate it.

bitnami/apache/values.yaml
jmConan: hope this is fixed now. was under the impression the readme update is handled by the build pipeline.

juan131 (approved):
LGTM

PR #8351, opened 2021-12-09
[bitnami/thanos] Fix Thanos Receiver service port name

v1ctorrhs:
**Description of the change**

Currently the Thanos receiver Ingress points to a service port `remote-write` and the actual port name is called `remote`. 
Of course this introduces issues and traffic is not received corretly.

**Benefits**
Tha thanos receiver getting the traffic destined to it.

**Possible drawbacks**

None
**Applicable issues**

  - fixes #

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/thanos/Chart.yaml
+++ b/bitnami/thanos/Chart.yaml
@@ -28,4 +28,4 @@ name: thanos
 sources:
   - https://github.com/bitnami/bitnami-docker-thanos
   - https://thanos.io
-version: 8.2.0
+version: 8.2.1
--- a/bitnami/thanos/templates/receive/service.yaml
+++ b/bitnami/thanos/templates/receive/service.yaml
@@ -52,7 +52,7 @@ spec:
     - port: {{ if .Values.receive.service.remoteWrite }}{{ coalesce .Values.receive.service.ports.remote .Values.receive.service.remoteWrite.port }}{{ else }}{{ .Values.receive.service.ports.remote }}{{ end }}
       targetPort: remote-write
       protocol: TCP
-      name: remote
+      name: remote-write
       {{- if and (or (eq .Values.receive.service.type "NodePort") (eq .Values.receive.service.type "LoadBalancer")) .Values.receive.service.nodePorts.remote }}
       nodePort: {{ .Values.receive.service.nodePorts.remote }}
       {{- else if eq .Values.receive.service.type "ClusterIP" }}


javsalgar (approved):

javsalgar (changes requested):
Hi,

This seems to be incorrect, right? According to the description, the port name should be remote and it is being changed to remote-write

v1ctorrhs:
> Hi,
> 
> This seems to be incorrect, right? According to the description, the port name should be remote and it is being changed to remote-write

The ingress is looking for a service port name `remote-write` and currently it is called `remote`.

```         
- path: /api/v1/receive
   {{- if eq "true" (include "common.ingress.supportsPathType" .) }}
   pathType: {{ .Values.receive.ingress.pathType }}
   {{- end }}
   backend: {{- include "common.ingress.backend" (dict "serviceName" (printf "%s-%s"  (include "common.names.fullname" .) "receive") "servicePort" "remote-write" "context" $)  | nindent 14 }}
```

Therefore this changes the service port name from `remote` to `remote-write` so the ingress can match accordingly. 
I'm not sure what change you are requesting

v1ctorrhs:
@javsalgar  Any update here? Thanks

marcosbc:
Hi @v1ctorrhs, wouldn't it make more sense to fix the Ingress configuration? Since it is working for LoadBalancer and other configurations - See [ingress.yaml#L32](https://github.com/bitnami/charts/blob/master/bitnami/thanos/templates/receive/ingress.yaml#L32)

v1ctorrhs:
@marcosbc I thought i'd update the service since that name is already used in the [statefulset#L180](https://github.com/bitnami/charts/blob/master/bitnami/thanos/templates/receive/statefulset.yaml#L160) so it makes more sense to keep it consistent. 

just like the example you gave with [ingress.yaml#L38](https://github.com/bitnami/charts/blob/master/bitnami/thanos/templates/receive/ingress.yaml#L38) - the containerPort name is also `http` in [statefulset.yaml#L157](https://github.com/bitnami/charts/blob/master/bitnami/thanos/templates/receive/statefulset.yaml#L157) as well as in the service [service.yaml#L37](https://github.com/bitnami/charts/blob/master/bitnami/thanos/templates/receive/service.yaml#L37)

Can we please merge this, i'm trying to use it in production. I'm not trying to start a discussion, it's your chart - whatever you want to call it i'm fine with that as long as it works - which it currently doesn't.

v1ctorrhs:
It seems like this got fixed in b8a5eb44a1b6abaa4440d5dc5dc0e923358c028e

PR #8382, opened 2021-12-13
[bitnami/postgresql-ha] Fix default value for pgpool.configuration

### LINKED ISSUE #8375: [postgresql-ha] pgpool.configuration syntax errors silently discarded
**Which chart**:
postgresql-ha version 8.0.4

**Describe the bug**
Syntax errors in the field pgpool.configuration cause the pgpool pod to fail without any error, and one must set BITNAMI_DEBUG to 'true' for this to be observable. This is compounded by the fact that the README lists the default value of this field to be an empty object ("{}"), which seems to indicate that the field should be a map of configuration fields to values, where inspecting the values.yaml file reveals it is actually meant to be a string.

**To Reproduce**
1. Set pgpool.configuration to an invalid value (non-string, or a string containing syntax errors)

**Expected behavior**
1. A concise explanation of the error to be reported to stdout/stderr (i.e. visible in kubectl logs) without needing to enable any debugging features.
2. The README to correctly state (or imply through its default value) that the pgpool.configuration field is, in fact, a string, which will have the helm tpl function applied to it (not terribly relevant to pgpool configuration files, but good to note nonetheless)

Irrelevant

**Additional context**
None

juan131:

**Description of the change**

The `pgpool.configuration` parameter expects a string instead of a map.

**Benefits**

README.md and default values are aligned with the expected format.

**Possible drawbacks**

None

**Applicable issues**

  - fixes #8375

**Additional information**

N/A

**Checklist** 

### DIFF
--- a/bitnami/postgresql-ha/Chart.yaml
+++ b/bitnami/postgresql-ha/Chart.yaml
@@ -27,4 +27,4 @@ name: postgresql-ha
 sources:
   - https://github.com/bitnami/bitnami-docker-postgresql
   - https://www.postgresql.org/
-version: 8.1.0
+version: 8.1.1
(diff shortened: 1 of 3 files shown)


marcosbc (approved):
LGTM

PR #8392, opened 2021-12-13
[bitnami/grafana] change http host for image renderer to enable IPv6

ioachim-hub:
Signed-off-by: Ioachim Lihor <ioachim.lihor@radcom.com>

**Description of the change**
I changed HTTP_HOST from image renderer from `0.0.0.0` to `::`.

**Benefits**
This change will allow communication on IPv6

**Possible drawbacks**
There are no drawbacks

**Additional information**

**Checklist** 


marcosbc (changes requested):
Thanks for the contribution @ioachim-hub! My main concern with this is that it could probably break the chart if IPv6 is not supported by the node. Maybe instead it would be worth creating a new parameter, `enableIPv6`. What do you think?

nicolaerosia:
@marcosbc please let us know if you need more details, thanks!

marcosbc (approved):
Thanks for checking it! I will test this PR in our release pipeline, if the tests succeed it will be automatically merged.

If it fails, it is likely that these changes won't work in non-IPv6 environments, but let's cross the fingers and see.

PR #8393, opened 2021-12-13
[bitnami/mariadb] change mariadb bind-address to enable IPv6

ioachim-hub:
Signed-off-by: Ioachim Lihor <ioachim.lihor@radcom.com>

**Description of the change**
Change `bind-address` from values.yaml from `0.0.0.0` to `::`

**Benefits**
With this change service could listen on IPv6

**Possible drawbacks**
No possible drawbacks

**Checklist** 
- [X] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [X] Variables are documented in the README.md
- [X] Title of the PR starts with chart name (e.g. [bitnami/<name_of_the_chart>])

PR #8395, opened 2021-12-13
[bitnami/mariadb-galera] change mariadb-galera bind-adrs to enable IPv6

ioachim-hub:
Signed-off-by: Ioachim Lihor <ioachim.lihor@radcom.com>

**Description of the change**
Change bind-address from values.yaml from `0.0.0.0` to `::`

**Benefits**
With this change service could listen on IPv6

**Possible drawbacks**
No possible drawbacks

**Applicable issues**

**Checklist** 

### DIFF
--- a/bitnami/mariadb-galera/values.yaml
+++ b/bitnami/mariadb-galera/values.yaml
@@ -319,7 +319,7 @@ mariadbConfiguration: |-
   tmpdir=/opt/bitnami/mariadb/tmp
   socket=/opt/bitnami/mariadb/tmp/mysql.sock
   pid_file=/opt/bitnami/mariadb/tmp/mysqld.pid
-  bind_address=0.0.0.0
+  bind_address=::
 
   ## Character set
   ##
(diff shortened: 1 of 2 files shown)


bitnami/mariadb-galera/values.yaml
```diff
@@ -319,7 +319,7 @@ mariadbConfiguration: |-
   tmpdir=/opt/bitnami/mariadb/tmp
   socket=/opt/bitnami/mariadb/tmp/mysql.sock
   pid_file=/opt/bitnami/mariadb/tmp/mysqld.pid
-  bind_address=0.0.0.0
+  bind_address=::
```
alemorcuq, line 322: What would happen if IPv6 is not available?

ioachim-hub: I'm testing in minikube which does not support ipv6.
here's proof:

ioachim-hub: @alemorcuq

carrodher:
This PR is related to https://github.com/bitnami/charts/pull/8393

bitnami/mariadb-galera/values.yaml
carrodher: If the address is `::`, the server accepts TCP/IP connections on all server host IPv4 and IPv6 interfaces. Use this address to permit both IPv4 and IPv6 connections on all server interfaces.

ioachim-hub:
This IR is also related to #8392

alemorcuq (approved):
Thanks!

PR #8398, opened 2021-12-13
[bitnami/postgresql-ha] Default pgpool reserved_connections to 1

meln5674:
**Description of the change**

Default to setting reserved_connections to 1 in pgpool.conf.

See [here](https://www.pgpool.net/docs/latest/en/html/runtime-config-connection.html#GUC-RESERVED-CONNECTIONS) for documentation on this property.

**Benefits**

When this property is set to 0 (as is default), pgpool blocks excess connections. This results in the readiness/liveness probes for pgpool failing with a timeout and no errors in the logs. When this property is set to 1, pgpool instead rejects the connection with an informative error message in the pod events, as well as the logs. This makes the problem much more obvious, as well as allows for log monitoring to scan for this message and alert if found.

**Possible drawbacks**

As the name implies, one fewer connection is allowed, but users can increase numInitChildren by 1 to compensate.

**Applicable issues**

See [This comment](https://github.com/bitnami/charts/issues/6094#issuecomment-991395399) for context

**Additional information**

This depends on [This PR](https://github.com/bitnami/bitnami-docker-pgpool/pull/73)

**Checklist** 

### DIFF
--- a/bitnami/postgresql-ha/templates/pgpool/deployment.yaml
+++ b/bitnami/postgresql-ha/templates/pgpool/deployment.yaml
@@ -223,7 +223,6 @@ spec:
             - name: PGPOOL_NUM_INIT_CHILDREN
               value: {{ .Values.pgpool.numInitChildren | quote }}
             {{- end }}
-            # .Values.pgpool.reservedConnections: {{ .Values.pgpool.reservedConnections }}
             {{- if .Values.pgpool.reservedConnections }}
             - name: PGPOOL_RESERVED_CONNECTIONS
               value: '{{ .Values.pgpool.reservedConnections }}'
--- a/bitnami/postgresql-ha/templates/pgpool/deployment.yaml
+++ b/bitnami/postgresql-ha/templates/pgpool/deployment.yaml
@@ -223,6 +223,11 @@ spec:
             - name: PGPOOL_NUM_INIT_CHILDREN
               value: {{ .Values.pgpool.numInitChildren | quote }}
             {{- end }}
+            # .Values.pgpool.reservedConnections: {{ .Values.pgpool.reservedConnections }}
+            {{- if .Values.pgpool.reservedConnections }}
+            - name: PGPOOL_RESERVED_CONNECTIONS
+              value: '{{ .Values.pgpool.reservedConnections }}'
+            {{- end }}
             {{- if .Values.pgpool.maxPool }}
             - name: PGPOOL_MAX_POOL
               value: {{ .Values.pgpool.maxPool | quote }}
--- a/bitnami/postgresql-ha/values.yaml
+++ b/bitnami/postgresql-ha/values.yaml
@@ -863,10 +863,12 @@ pgpool:
   ## ref: https://github.com/bitnami/bitnami-docker-pgpool#configuration
   ##
   numInitChildren: ""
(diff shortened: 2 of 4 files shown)


bitnami/postgresql-ha/values.yaml
```diff
@@ -863,6 +863,12 @@ pgpool:
   ## ref: https://github.com/bitnami/bitnami-docker-pgpool#configuration
   ##
   numInitChildren: ""
+  ## @param pgpool.reservedConnections When this parameter is set to 1 or greater, incoming connections from clients
+  ## are not accepted with error message "Sorry, too many clients already", rather than blocked if the number of current
+  ## connections from clients is more than (num_init_children - reserved_connections).
+  ## ref: https://github.com/bitnami/bitnami-docker-pgpool#configuration
+  reservedConnections: 1
+
```
carrodher, line 871: This metadata is used to automatically generate the table present in the README using https://github.com/bitnami-labs/readme-generator-for-helm

As the description is very long, can you please split it into a simple first line and then add more clarifications or links if you want in the next lines?
```suggestion
  ## @param pgpool.reservedConnections When this parameter is set to 1 or greater, incoming connections from clients
  ## are not accepted with error message "Sorry, too many clients already", rather than blocked if the number of current
  ## connections from clients is more than (num_init_children - reserved_connections).
  ## ref: https://github.com/bitnami/bitnami-docker-pgpool#configuration
  ##
  reservedConnections: 1
```
Apart from that I just provide a suggestion for removing the blank line and adding the `##` before the value itself as in the rest of parameters

carrodher (commented):
Thanks for your PR, we will wait until the changes in the container are available, in the meantime, can you check my comment and generate the README table?

bitnami/postgresql-ha/values.yaml
meln5674: Understood. I assumed the REAMDE generation was taken care of by CI or something. I used that description as numInitChildren had taken its description verbatim from the pgpool docs site, so I assumed that was the policy. I kept the full description below, with a shorter one on the first line to be shown in the readme. Let me know if this is what you were looking for.

bitnami/postgresql-ha/templates/pgpool/deployment.yaml
```diff
@@ -223,6 +223,11 @@ spec:
             - name: PGPOOL_NUM_INIT_CHILDREN
               value: {{ .Values.pgpool.numInitChildren | quote }}
             {{- end }}
+            # .Values.pgpool.reservedConnections: {{ .Values.pgpool.reservedConnections }}
```
carrodher, line 226: I think this is a leftover, right?
```suggestion
```

bitnami/postgresql-ha/values.yaml
carrodher: Thanks, now it looks much better.
Yes, there is some automation in place to generate this table, although at this very moment we are iterating over the process to improve it since it is a recent addition.

carrodher (commented):
Apart from the comment related to the leftover, changes LGTM. I will add the `on-hold` label to prevent the "Stale bot" from closing the PR while the changes in the container are applied.
Thanks!

juan131 (approved):
LGTM

Issue #8390, opened 2021-12-13
Memcached HA always out of sync in ArgoCD


chap-dr:
**Which chart**:
memcached 5.15.10

**Describe the bug**
While this is not a bug directly related to this chart, while the root cause is fixed, it would be nice to have a fix here. 

The bug related to HPA resources always being reported as out of sync, due to the HPA flipping entries in the list. 

ArgoCD has a note about it here: https://argo-cd-docs.readthedocs.io/en/latest/user-guide/diffing/ which links to this 2 year old issue: https://github.com/kubernetes/kubernetes/issues/74099

Lastly the statefulset's replica field should match minReplicas value set in hpa, or a note in values should mention to match these two values. 

<img width="986" alt="image" src="https://user-images.githubusercontent.com/25221263/145837665-3a91e75c-b2f3-4460-b978-8043dc9c102b.png">

**To Reproduce**
1. Generate valid kubernetes manifests, using helm. 
2. Sync it using ArgoC
3. Sync will succeede, but shortly after report out of sync on 

**Expected behavior**
Expecting chart to produce yaml that will not get reported out of sync by argo. 

**Additional context**
While I am aware that this is not direclty related to this chart, it would be a minor change (which I am willing to contribute / make if you like) to make make it work with the HPA funkyness.

Mauraza:
Hi @chap-dr,

We are open to contributions, if you want, we will be happy to review it.

chap-dr:
> Hi @chap-dr,
> 
> We are open to contributions, if you want, we will be happy to review it.

Done and done. 😄

Mauraza:
Hi @chap-dr,

Thanks for the contributions!! We will review it

carrodher:
Since the associated PR is already merged I think we can safely close this issue

PR #8408, opened 2021-12-14
[bitnami/memcached] Update to prevent argocd from always  reporting out of sync

chap-dr:
**Description of the change**
Minor change that makes prevents this chart from always being out of sync when deployed using ArgoCD. 
More info here: https://github.com/bitnami/charts/issues/8390#issuecomment-993336875

**Benefits**
Chart can be used with ArgoCD

**Possible drawbacks**
None. 
**Applicable issues**
https://github.com/bitnami/charts/issues/8390#issuecomment-993336875

**Checklist** 


marcosbc (approved):
LGTM! Thanks for the fix!

PR #8411, opened 2021-12-14
[bitnami/rabbitmq-cluster-operator] Use consistent labeling and label selectors

QuingKhaos:
**Description of the change**

Add `app.kubernetes.io/component` and `app.kubernetes.io/part-of` labels where they were missing and remove the `app.kubernetes.io/part-of` label from a few label selectors.

**Benefits**

Labels on all manifests are now consistent throughout the chart. The `app.kubernetes.io/part-of` label has no additional benefit or difference on label selectors, so I removed it to make them all consistent.

**Applicable issues**

Due to the inconsistencies of labeling and label selectors, the messaging topology operator metrics service and service monitor didn't match and no metrics were scraped for it.

**Checklist** 

### DIFF
--- a/bitnami/rabbitmq-cluster-operator/templates/messaging-topology-operator/metrics-service.yaml
+++ b/bitnami/rabbitmq-cluster-operator/templates/messaging-topology-operator/metrics-service.yaml
@@ -49,5 +49,4 @@ spec:
     {{- end }}
   selector: {{- include "common.labels.matchLabels" . | nindent 4 }}
     app.kubernetes.io/component: messaging-topology-operator
-    app.kubernetes.io/part-of: rabbitmq
 {{- end }}
--- a/bitnami/rabbitmq-cluster-operator/Chart.yaml
+++ b/bitnami/rabbitmq-cluster-operator/Chart.yaml
@@ -25,4 +25,4 @@ name: rabbitmq-cluster-operator
 sources:
   - https://github.com/bitnami/bitnami-docker-rabbitmq-cluster-operator
   - https://github.com/rabbitmq/cluster-operator
-version: 2.0.7
+version: 2.0.8
--- a/bitnami/rabbitmq-cluster-operator/templates/cluster-operator/servicemonitor.yaml
+++ b/bitnami/rabbitmq-cluster-operator/templates/cluster-operator/servicemonitor.yaml
@@ -4,6 +4,8 @@ kind: ServiceMonitor
 metadata:
   name: {{ template "rmqco.clusterOperator.fullname" . }}
   labels: {{- include "common.labels.standard" . | nindent 4 }}
+    app.kubernetes.io/component: rabbitmq-operator
+    app.kubernetes.io/part-of: rabbitmq
     {{- if .Values.clusterOperator.metrics.serviceMonitor.additionalLabels }}
     {{- include "common.tplvalues.render" (dict "value" .Values.clusterOperator.metrics.serviceMonitor.additionalLabels "context" $) | nindent 4 }}
     {{- end }}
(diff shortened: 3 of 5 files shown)


miguelaeh (commented):
Hi @QuingKhaos ,
Did you test the upgrade from previous versions work? It is common that after changing labels the upgrade starts to fail

QuingKhaos:
Hi @miguelaeh, not with an `helm update`. I use `helm template` in my GitOps workflow and the GitLab Kubernetes Agent pruned the old service monitor objects.

Would need to setup a quick test scenario with `helm install` and `helm update`

QuingKhaos:
Hi @miguelaeh, not directly with a `helm update`. I use `helm template` in my GitOps workflow and directly applying the changes with `kubectl apply` or the GitLab Kubernetes Agent did no encounter any problems.

Would need to setup a quick test scenario with `helm install` and `helm update`

IMHO it should not break with `helm update` too, as no deployment pod label selectors were changes, which would break during an update if modified.

bitnami/rabbitmq-cluster-operator/templates/messaging-topology-operator/metrics-service.yaml
```diff
@@ -49,5 +49,4 @@ spec:
     {{- end }}
   selector: {{- include "common.labels.matchLabels" . | nindent 4 }}
     app.kubernetes.io/component: messaging-topology-operator
-    app.kubernetes.io/part-of: rabbitmq
```
miguelaeh, line 52: Is there any reason to be deleting this? I won't change this. I think the PR could have the same level of effectiveness just adding the ` app.kubernetes.io/component: messaging-topology-operator` when needed.

QuingKhaos: Consistency with the deployment label selector and overall consistency with chart conventions. Usually you add only the component additionally to the `common.labels.matchLabels` if needed to differ between different components.

QuingKhaos: If it was for me, I would remove the `app.kubernetes.io/part-of: rabbitmq` from the chart completely, as this is something I would only set as a chart user via `commonLabels`, so I can group the resources from the chart how I want :)

juan131 (changes requested):
Hi @QuingKhaos 

Thanks for the PR! Please correct if I'm wrong but you're modifying some fields that are supposed to be immutable, therefore you're breaking backwards compatibility, am I right?

See https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#selector:

> In API version apps/v1, .spec.selector and .metadata.labels do not default to .spec.template.metadata.labels if not set. So they must be set explicitly. Also note that .spec.selector is immutable after creation of the Deployment in apps/v1.

QuingKhaos:
Hi @juan131, sorry for the big delay, needed vacation time happened.

Exactly, some fields are supposed to be immutable. For this reason I didn't modify `.spec.selector` of the Deployment. I adapted the selectors in the services and monitors to match the ones from the Deployment. The selectors in the services and monitors are mutable.

Can you reopen this PR?

juan131:
> Exactly, some fields are supposed to be immutable. For this reason I didn't modify .spec.selector of the Deployment. I adapted the selectors in the services and monitors to match the ones from the Deployment. The selectors in the services and monitors are mutable.

Fair enough.. I didn't check it correctly the other time, sorry for that.

> Can you reopen this PR?

@QuingKhaos I think it's you the one that needs to reopen the PR since you're the author. Please ping me in you find issues.

QuingKhaos:
Hi @juan131, excuse me for my irregular responses, I don't have any regular working hours 😅 

I was not able to reopen the PR, so I re-created it as a new one in #8784

PR #8412, opened 2021-12-14
[bitnami/rabbitmq-cluster-operator] Fix messaging-topology-operator pod affinities

QuingKhaos:
**Description of the change**

Use the correct component name for pod affinities on the `message-topology-operator` deployment.

**Benefits**

When deploying the operators in HA, the `message-topology-operator` will now spread over different nodes to each other instead of having and anti-affinity by default to the `rabbitmq-cluster-operator`.

**Possible drawbacks**

This change will update the default pod anti-affinity of the `messaging-topology-operator` deployment. If anyone has adapted the affinities due to this bug, they can remove their changes.

If anyone is using the pod affinity preset to co-locate the `messaging-topology-operator` with the `rabbitmq-cluster-operator` pods, this will not be possible anymore. But IMHO it's not the intended case.

**Additional information**

RFC: Should the component name of the affinity/anti-affinity presets be configurable in the case of having multiple components in one chart?

**Checklist** 

### DIFF
--- a/bitnami/rabbitmq-cluster-operator/Chart.yaml
+++ b/bitnami/rabbitmq-cluster-operator/Chart.yaml
@@ -25,4 +25,4 @@ name: rabbitmq-cluster-operator
 sources:
   - https://github.com/bitnami/bitnami-docker-rabbitmq-cluster-operator
   - https://github.com/rabbitmq/cluster-operator
-version: 2.0.6
+version: 2.0.7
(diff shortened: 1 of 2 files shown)


alemorcuq (changes requested):
Thanks for the PR! Could you please rebase and increase the version again?

QuingKhaos:
@alemorcuq rebased and all green 🚀

alemorcuq (approved):

PR #8413, opened 2021-12-14
[bitnami/rabbitmq-cluster-operator] ServiceMonitor port must be Service port name

QuingKhaos:
**Description of the change**

Changes the port of the `ServiceMonitor` for the messaging topology operator to `http` which matches the port name in the matching service. `.spec.endpoints[].targetPort` would have be set to `http-metrics` if used, but `.spec.endpoints[].port` refers to the name of the port defined in the service.

**Applicable issues**

Together with #8411 the messaging topology operator metrics were not scraped by the Prometheus operator.

**Checklist** 

### DIFF
--- a/bitnami/rabbitmq-cluster-operator/templates/messaging-topology-operator/servicemonitor.yaml
+++ b/bitnami/rabbitmq-cluster-operator/templates/messaging-topology-operator/servicemonitor.yaml
@@ -29,7 +29,7 @@ spec:
     matchNames:
       - {{ .Release.Namespace }}
   endpoints:
-    - port: http-metrics
+    - port: http
       {{- if .Values.msgTopologyOperator.metrics.serviceMonitor.interval }}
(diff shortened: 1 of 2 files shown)


bitnami/rabbitmq-cluster-operator/templates/messaging-topology-operator/servicemonitor.yaml
```diff
@@ -29,7 +29,7 @@ spec:
     matchNames:
       - {{ .Release.Namespace }}
   endpoints:
-    - port: http-metrics
+    - port: http
```
juan131, line 32: You're completely right, see:
- https://github.com/bitnami/charts/blob/master/bitnami/rabbitmq-cluster-operator/templates/messaging-topology-operator/metrics-service.yaml#L38

juan131 (approved):
Thanks for the fix! LGTM

PR #8432, opened 2021-12-16
bitnami/rabbitmq: README copywriting updates

michaelklishin:
RabbitMQ has been multi-protocol since 2010 or so (when STOMP support was introduced).
The core team would strongly prefer to use this language as "AMQP" these days refers to more
than one protocol for historical reasons.


miguelaeh (approved):
LGTM!

PR #8443, opened 2021-12-17
[bitnami/mariadb-galera] Revert enabling of IPv6

carrodher:
Reverts #8395, because it is failing in the internal CI/CD:
```
                1) MariaDB remote authentication
                     Password authentication is required through 10.49.49.5:
                   Expected "Access denied for user 'root'@'10.49.49.5'" but got "'ERROR 2002 (HY000): Can't connect to server on '10.49.49.5' (115)'"
                2) MariaDB remote authentication
                     Password authentication is required through 127.0.0.1:
                   Expected "Access denied for user 'root'@'127.0.0.1'" but got "'ERROR 2002 (HY000): Can't connect to server on '127.0.0.1' (115)'"
```

### DIFF
--- a/bitnami/mariadb-galera/Chart.yaml
+++ b/bitnami/mariadb-galera/Chart.yaml
@@ -27,4 +27,4 @@ sources:
   - https://github.com/bitnami/bitnami-docker-mariadb-galera
   - https://github.com/prometheus/mysqld_exporter
   - https://mariadb.org
-version: 6.0.9
+version: 6.0.10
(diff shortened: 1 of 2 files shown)


nicolaerosia:
can you share what the CI/CD is doing in this particular case? how it tries to connect?

carrodher:
> MariaDB remote authentication

Yes, sure. The command executed is the following one
```
mysql -uroot -h IP
```
where IP is `127.0.0.1` in the first execution and the machine IP in the second one.

In both cases, it is expected the command to return something like
```
Access denied for user 'root'@'127.0.0.1'
```
(same for the machine IP).

But with those changes, the above command is returning
```
'ERROR 2002 (HY000): Can't connect to server on '127.0.0.1' (115)'
```
(same for the machine IP).

PR #8458, opened 2021-12-20
[bitnami/kafka] Improved compatibility with Kubernetes 1.21+ and Openshift

hierynomus:
**Description of the change**
There are 2 changes to the kafka chart:
- Ensure that the chart is Kubernetes 1.21+ compatible by making sure that the PodDisruptionBudget uses the `policy/v1` apiVersion if it is available.
- Can set a special `auto` value to the `volumePermissions.securityContext.runAsUser` to make it OpenShift compatible as that does not like the `0` (root) user.

**Benefits**
Compatibility with newer Kubernetes versions and OpenShift deployments

**Possible drawbacks**
Fully backwards compatible so no drawbacks expected.>

**Applicable issues**
N/A

**Additional information**
N/A

**Checklist** 

### DIFF
--- a/bitnami/kafka/templates/poddisruptionbudget.yaml
+++ b/bitnami/kafka/templates/poddisruptionbudget.yaml
@@ -1,6 +1,10 @@
 {{- $replicaCount := int .Values.replicaCount }}
 {{- if and .Values.pdb.create (gt $replicaCount 1) }}
+{{- if .Capabilities.APIVersions.Has "policy/v1/PodDisruptionBudget" }}
+apiVersion: policy/v1
+{{- else }}
 apiVersion: policy/v1beta1
+{{- end }}
 kind: PodDisruptionBudget
 metadata:
   name: {{ template "kafka.fullname" . }}
--- a/bitnami/kafka/templates/poddisruptionbudget.yaml
+++ b/bitnami/kafka/templates/poddisruptionbudget.yaml
@@ -1,10 +1,6 @@
 {{- $replicaCount := int .Values.replicaCount }}
 {{- if and .Values.pdb.create (gt $replicaCount 1) }}
-{{- if .Capabilities.APIVersions.Has "policy/v1/PodDisruptionBudget" }}
-apiVersion: policy/v1
-{{- else }}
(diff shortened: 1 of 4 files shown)


bitnami/kafka/values.yaml
```diff
@@ -902,7 +902,7 @@ volumePermissions:
   ##
   enabled: false
   ## The security context for the volumePermissions init container
-  ## @param volumePermissions.securityContext.runAsUser User ID for the container
+  ## @param volumePermissions.securityContext.runAsUser User ID for the container. Can be set to "auto". "auto" is especially useful for OpenShift which has scc with dynamic userids (and 0 is not allowed).
```
juan131, line 905: Split the comment:
```suggestion
  ## @param volumePermissions.securityContext.runAsUser User ID for the container.
  ## Can be set to "auto". "auto" is especially useful for OpenShift which has scc with dynamic user ids (and 0 is not allowed).
```

bitnami/kafka/templates/poddisruptionbudget.yaml
```diff
@@ -1,6 +1,10 @@
 {{- $replicaCount := int .Values.replicaCount }}
 {{- if and .Values.pdb.create (gt $replicaCount 1) }}
+{{- if .Capabilities.APIVersions.Has "policy/v1/PodDisruptionBudget" }}
+apiVersion: policy/v1
+{{- else }}
 apiVersion: policy/v1beta1
+{{- end }}
```
juan131, line 7: We have a [named templated](https://github.com/bitnami/charts/blob/master/bitnami/common/templates/_capabilities.tpl#L21) available in the `common` chart that you can use for this:
```suggestion
apiVersion: {{ include "common.capabilities.policy.apiVersion" . }}
```

juan131 (changes requested):
Thanks so much for this great PR @hierynomus !! Please check my comments when you have a chance.

hierynomus:
@juan131 Thanks for the review. I've committed your suggested changes.

juan131 (approved):
LGTM

Issue #8450, opened 2021-12-17
[bitnami/contour] cannot make ingress non-default


EugenMayer:
I'am deploy to contour instances (into different namespaces): `contour-internal` and `contour-external`

It works out pretty well. The only issue is with the ingressClasses - the internal I set to be the default

```yaml
contour:
  ingressClass:
    name: contour-internal
    create: true
    default: true
  manageCRDs: true
```

while the external I set to be not a default ingressClass

```yaml
contour:
  ingressClass:
    name: contour-internal
    create: true
    default: false
  manageCRDs: false
```

When deploying that and trying to create an ingress router without setting a specific ingressClassName I get the error:

```
Internal error occurred: 2 default IngressClasses were found, only 1 allowed
```

Am I using it the wrong way? Thanks

javsalgar:
Hi,

Could you detail where this error appears? Is it when deploying the YAML? I see that you created a ticket upstream, is this an error upstream?

javsalgar:
Hi,

Just a note to let you know that I was able to reproduce the issue. I believe the problem is here.

```
    ingressclass.kubernetes.io/is-default-class: {{ default true $ingressClass.default | quote }}
```

I think it should be

```
{{- if $ingressClass.default }}
    ingressclass.kubernetes.io/is-default-class: "true"
{{- end }}

```

As you spotted the issue, would you like to submit a PR with the change?

EugenMayer:
PR has been merged and the issue is solved - so the issue can be closed IMHO

PR #8478, opened 2021-12-22
[bitnami/contour] Fix applying default ingress label even if not default #8450

EugenMayer:
Even if `ingressClass.default` is set to false, the actual label will be applied and make the ingress be a default ingress. See #8450

Credits go all to @javsalgar !


javsalgar (approved):
Awesome! LGTM

PR #8486, opened 2021-12-23
[bitnami/redis] lookup existing secrets before generating random secrets

yisiqi:
**Description of the change**

Related to: 
- #8320
- #8495
- #8499
- #8486
- #8504

**Benefits**

**Possible drawbacks**

**Applicable issues**

  - fixes #

**Additional information**

**Checklist** 


rafariossaa (commented):
Hi,
  Thanks for sending this.
  Not very sure what is intended here and how it differs from using `auth.existingSecret`. If you want to use a secret for the password you can use that parameter.

yisiqi:
Hi @rafariossaa 

After trying a lot of Charts deployment, I found that using `auth.existingSecret`  requires additional resource creation . It's repetitive and error-prone. 

But the most important is that if you want to use the generated random secret, the existing charts cannot keep the secret unchanged after the upgrade.

Lookup can balance this problem well. If you want to use a randomly generated secret, lookup can check whether the secret already exists before the upgrade. 

I used the same solution in bitnami/charts/postgresql, and it worked well, making random secret generation a truly usable option.

rafariossaa:
Hi,
  I took a look to #8320, and this is similar.
  LGTM.

PR #8495, opened 2021-12-24
[bitnami/cassandra] lookup existing secrets before generating random secrets

yisiqi:
**Description of the change**

Related to: 
- #8320
- #8495
- #8499
- #8486
- #8504

**Benefits**

**Possible drawbacks**

**Applicable issues**

  - fixes #

**Additional information**

**Checklist** 


bitnami/cassandra/templates/_helpers.tpl
```diff
@@ -180,3 +180,75 @@ Return true if a TLS credentials secret object should be created
 {{- end -}}
 {{- end -}}
 
+{{/*
+Returns the available value for certain key in an existing secret (if it exists),
+otherwise it generates a random value.
+*/}}
+{{- define "getValueFromSecret" }}
+    {{- $len := (default 16 .Length) | int -}}
+    {{- $obj := (lookup "v1" "Secret" .Namespace .Name).data -}}
```
javsalgar, line 189: I have my concerns with this approach, as it forces a cluster to be available. In the past using helm template caused errors, could you confirm that this approach is also compatible with other deployment approaches like helm template + kubectl?

yisiqi: Sure, I can use this on an offline laptop to successfully execute the `helm template`. It will generate a random secret, as if there is no existing one. 

According to the official helm document：

> Helm is not supposed to contact the Kubernetes API Server during a helm template or a helm install|update|delete|rollback --dry-run, so the lookup function will return an empty list (i.e. dict) in such a case.

So, Yes, It will compatible with `helm template` + `kubectl` and any other deployment approaches.

juan131 (approved):
LGTM, we recently did sth similar with PostgreSQL, see https://github.com/bitnami/charts/pull/8320
I won't merge until @javsalgar take a second look in any case

javsalgar (approved):
I recall that in the past helm would crash if lookup doesn't exist. If now this is solved I have no concerns in merging this. LGTM

PR #8509, opened 2021-12-27
[bitnami/phpbb] fix volumemount name for volumePermissions

### LINKED ISSUE #5401: [bitnami/phpbb] volumePermissions paramter
The volumePermissions does not work as the name on the volume is invalid. I belive it should be named _phpbb-data_

https://github.com/bitnami/charts/blob/74f2fd8f163cb4a9aa48bae1cb2a674ce4de2ae4/bitnami/phpbb/templates/deployment.yaml#L80

`Error: UPGRADE FAILED: cannot patch "fopw-phpbb" with kind Deployment: Deployment.apps "fopw-phpbb" is invalid: [spec.template.spec.initContainers[0].volumeMounts[0].name: Not found: "data", spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{"app.kubernetes.io/instance":"fopw", "app.kubernetes.io/name":"phpbb"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable]

`

arcalys:
**Description of the change**

The volume mount name is incorrect in the phpbb deployment template when using volumePermissions. It references a "data" mount instead of "phpbb-data". This causes the following error:

```
[...] is invalid: spec.template.spec.initContainers[0].volumeMounts[0].name: Not found: "data"
```
**Applicable issues**

- Fixes #5401

**Checklist** 

### DIFF
--- a/bitnami/phpbb/Chart.yaml
+++ b/bitnami/phpbb/Chart.yaml
@@ -26,4 +26,4 @@ name: phpbb
 sources:
   - https://github.com/bitnami/bitnami-docker-phpbb
   - https://www.phpbb.com/
-version: 10.2.2
+version: 10.2.3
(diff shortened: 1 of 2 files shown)


carrodher (approved):
You're totally right, thanks for this fix! LGTM

PR #8543, opened 2021-12-31
[bitnami/ghost] Add secret password key parameter

jgillich:
**Description of the change**

Add secret password key parameter

**Benefits**

We will be able to use secret keys other than `mariadb-password`

**Possible drawbacks**

None

**Checklist** 

### DIFF
--- a/bitnami/ghost/values.yaml
+++ b/bitnami/ghost/values.yaml
@@ -592,6 +592,7 @@ mariadb:
   ## @param mariadb.auth.username MariaDB custom user name
   ## @param mariadb.auth.password MariaDB custom user password
   ## @param mariadb.auth.existingSecret Existing secret with MariaDB credentials
+  ## @param mariadb.auth.existingSecretPasswordKey Password key to be retrieved from existing secret
   ## ref: https://github.com/bitnami/bitnami-docker-mariadb#setting-the-root-password-on-first-run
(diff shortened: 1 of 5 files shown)


alemorcuq:
Thanks for opening this PR! Could you please rebase? I think the branch was created from an old commit looking at some of the changes.

jgillich:
@alemorcuq Rebased (although it was based on a commit from the same day)

bitnami/ghost/values.yaml
```diff
@@ -601,6 +602,7 @@ mariadb:
     username: bn_ghost
     password: ""
     existingSecret: ""
+    existingSecretPasswordKey: "mariadb-password"
```
alemorcuq, line 605: Values under the `mariadb` section are values that exist in the MariaDB chart. We should add this value to the MariaDB chart before adding it to Ghost.

alemorcuq (changes requested):
The PR looks good, but we should add `existingSecretPasswordKey` to the MariaDB chart before adding it to this chart. Would you like to send a PR?

bitnami/ghost/values.yaml
jgillich: @alemorcuq Good catch. PR for mariadb chart is up #8602

jgillich:
Rebased to resolve conflict

jgillich:
I've come to realize that Helm doesn't offer the flexibility that I need so I've stopped using it. Anyone who needs this config option is free to take these changes.

PR #8562, opened 2022-01-04
[bitnami/several] Adapt templating format

carrodher:
**Description of the change**

According to https://helm.sh/docs/chart_best_practices/templates/#formatting-templates
- Correct ✅: 
```
{{ .foo }}
{{ print "foo" }}
{{- print "bar" -}}
```
- Incorrect 🔴:
```
{{.foo}}
{{print "foo"}}
{{-print "bar"-}}
```

This PR adapts the existing code to follow the above directive.

**Benefits**

Bitnami Helm charts following the Helm recommendations

**Possible drawbacks**

There are a ton of changes, not all of them were tested, something can be broken 🤞 

**Additional information**

 The chart version was bumped in a minor to reduce the chances of conflicts since a lot of charts are being modified


miguelaeh (approved):
LGTM

dalbani:
Sorry if it sounds like a rant, but I find it strange that all those charts got a `MINOR` version bump, whereas a `PATCH` looked more appropriate.
According to, hum, the Helm chart best practices referring to Semver 2:
```
Given a version number MAJOR.MINOR.PATCH, increment the:

MAJOR version when you make incompatible API changes,
MINOR version when you add functionality in a backwards compatible manner, and
PATCH version when you make backwards compatible bug fixes.
```
Correct me if I'm wrong but I don't this PR added any functionality. It was more like cosmetic changes, isn't it?

It's no big deal in practice, but as an end-user, seeing a `MINOR` version bump means I kind of "expect" new functionalities.
Well, too bad, I suppose I already had my Christmas presents :wink:

carrodher:
Yes, I was aware of that so I added a note in the PR description:
> The chart version was bumped in a minor to reduce the chances of conflicts since a lot of charts are being modified

As charts are automatically updated by our test & release pipeline (apart from the PRs) there a lot of chances to coincide in the version if it is a patch since it is the version usually used by the pipeline for automatic updates

dalbani:
OK, very clear, thanks for the feedback 👍

PR #8570, opened 2022-01-04
[bitnami/redis] Enables Redis to utilize external-dns

jtslear:
**Description of the change**

* By way of using `external-dns` we can assign a DNS entry for each Pod
* Doing so allows anyone who's got the ability to talk to the IP's for which Redis utilize, external to the cluster, to become members of the Redis cluster
* This is helpful during Redis migrations where backing up and restoring are destructive, but instead, we can enable both Redis in Kubernetes and Redis on VM's to talk to each other, perform application configuration changes, and turn down the VM's.
  * Theoretically vice versa as well
* Note that this **REQUIRES** the person leveraging this option to have `external-dns` in a working fashion prior to this being usable.

**Checklist** 


bitnami/redis/templates/headless-svc.yaml
```diff
@@ -7,8 +7,12 @@ metadata:
     {{- if .Values.commonLabels }}
     {{- include "common.tplvalues.render" ( dict "value" .Values.commonLabels "context" $ ) | nindent 4 }}
     {{- end }}
+  annotations:
   {{- if .Values.commonAnnotations }}
-  annotations: {{- include "common.tplvalues.render" ( dict "value" .Values.commonAnnotations "context" $ ) | nindent 4 }}
+  {{- include "common.tplvalues.render" ( dict "value" .Values.commonAnnotations "context" $ ) | nindent 4 }}
+  {{- end }}
```
alemorcuq, line 13: I believe this is a typo.

bitnami/redis/templates/headless-svc.yaml
```diff
@@ -7,8 +7,12 @@ metadata:
...
+  {{- if .Values.externalDNS.enabled }}
+    external-dns.alpha.kubernetes.io/hostname: {{ printf "%s.%s" (include "common.names.fullname" .) .Values.externalDNS.suffix }}
```
alemorcuq, line 15: `{{ printf "%s.%s" (include "common.names.fullname" .) .Values.externalDNS.suffix }}` is used quite a bit on this PR, I think it's worth to add a helper to include it.

bitnami/redis/values.yaml
```diff
@@ -23,6 +23,12 @@ global:
 ## @section Common parameters
 ##
 
+## @param externalDNS Boolean to indicate the desire to utilize `external-dns`
+##
+externalDNS:
+  enabled: false
+  # suffix: ""
+
```
alemorcuq, line 31: I'm not sure this is a good way to name the parameters, as users could assume that this also deploys a installation of the `bitnami/external-dns` chart.

alemorcuq (changes requested):
Thanks for the PR, @jtslear. I've left some comments, but since I'm not very familiar with the technology, I'd like another colleague to review this as well. I'll ping him on Monday.

bitnami/redis/values.yaml
jtslear: I would agree.  I've renamed it to `useExternalDNS`.  Though I'm not thrilled about that name either.  Open to further suggestions.

bitnami/redis/templates/headless-svc.yaml
jtslear: How so? :thinking:

jtslear:
@alemorcuq I could use some guidance on the test, `generate-chart-readme`.  I've added the values in place, but the job claims I did not and also I added one that does not exist.  I'm not sure what I need to do to fix this.  I created the values in the README manually, alongside an example configuration to assist persons whom may want to leverage this configuration option.

bitnami/redis/README.md
```diff
@@ -462,6 +466,38 @@ Bitnami will release a new chart updating its containers if a new version of the
...
+  enabled: true
+  suffix: prod.example.org
+  additionalAnnotations:
+    ttl: 10
+```
+
+On a cluster where the name of the installation is `a`, the hostname of a Pod
+is configured like so: `a-redis-node-0.a-redis.prod.example.org`.  The IP of
+that FQDN will match that of the associated Pod.  This modifies the following
+parameters of the Redis/Sentinel configuration using this new FQDN:
```
WarheadsSE, line 487:
```suggestion
On a cluster where the name of the Helm release is `a`, the hostname of a Pod
is generated as: `a-redis-node-0.a-redis.prod.example.org`.  The IP of
that FQDN will match that of the associated Pod.  This modifies the following
parameters of the Redis/Sentinel configuration using this new FQDN:
```

Tweaking to Helm nomenclature, and slightly tweaking the language.

bitnami/redis/templates/headless-svc.yaml
WarheadsSE: This `{{- end }}` should be here. This is the case because `annotations` itself was moved outside of the `if` block statement to support the possibility of no `commonAnnotations`, but present `useExternalDNS` configuration.

bitnami/redis/templates/headless-svc.yaml
```diff
@@ -7,8 +7,15 @@ metadata:
...
+  annotations:
   {{- if .Values.commonAnnotations }}
-  annotations: {{- include "common.tplvalues.render" ( dict "value" .Values.commonAnnotations "context" $ ) | nindent 4 }}
+  {{- include "common.tplvalues.render" ( dict "value" .Values.commonAnnotations "context" $ ) | nindent 4 }}
+  {{- end }}
+  {{- if .Values.useExternalDNS.enabled }}
+    {{ .Values.useExternalDNS.annotationKey }}hostname: {{ include "redis.externalDNS.suffix" . }}
+    {{- range $key, $val := .Values.useExternalDNS.additionalAnnotations }}
+    {{ $.Values.useExternalDNS.annotationKey }}{{ $key }}: {{ $val | quote }}
+    {{- end }}
```
WarheadsSE, line 18:
```suggestion
    {{- include "common.tplvalues.render" ( dict "value" .Values.commonAnnotations "context" $ ) | nindent 4 }}
  {{- end }}
  {{- if .Values.useExternalDNS.enabled }}
    {{ .Values.useExternalDNS.annotationKey }}hostname: {{ include "redis.externalDNS.suffix" . }}
    {{- range $key, $val := .Values.useExternalDNS.additionalAnnotations }}
    {{ $.Values.useExternalDNS.annotationKey }}{{ $key }}: {{ $val | quote }}
    {{- end }}
```

We can notch the indentation of the `commonAnnoations` render. Ideally, we could do the same for the `additionalAnnotations` , however, the spaces currently matter within the `range` loop because of how this is composed within.

What do you think of making this into a single template for all of these annotations, say `redis.externalDNS.annotations`, and instead simply gating this here?

jtslear: @WarheadsSE thanks for the suggestion, I proceeded to do exactly this in ba304f9c5

alemorcuq:
> @alemorcuq I could use some guidance on the test, `generate-chart-readme`. I've added the values in place, but the job claims I did not and also I added one that does not exist. I'm not sure what I need to do to fix this. I created the values in the README manually, alongside an example configuration to assist persons whom may want to leverage this configuration option.

It is missing the `@param` metadata in the `values.yaml`. The README table can be generated from that metadata. More on this tool here: https://github.com/bitnami/charts/blob/master/CONTRIBUTING.md#documentation-requirements

@juan131, could you take a look at this PR?

jtslear:
@alemorcuq Readme is fixed, thank you much!  This is now ready for review.

bitnami/redis/README.md
```diff
@@ -462,6 +466,38 @@ Bitnami will release a new chart updating its containers if a new version of the
...
+* `known-sentinel`
+* `known-replica`
+* `announce-ip`
+
+:warning: This requires a working installation of [`external-dns`] to be fully
+functional. :warning:
+
+See the documentation of [`external-dns`] for additional configuration options.
+
+[`external-dns`]: https://github.com/kubernetes-sigs/external-dns
```
juan131, line 499:
```suggestion
See the [official ExternalDNS documentation](https://github.com/kubernetes-sigs/external-dns) for additional configuration options.
```

bitnami/redis/README.md
```diff
@@ -462,6 +466,38 @@ Bitnami will release a new chart updating its containers if a new version of the
 
 To modify the application version used in this chart, specify a different version of the image using the `image.tag` parameter and/or a different repository using the `image.repository` parameter. Refer to the [chart documentation for more information on these parameters and how to use them with images from a private registry](https://docs.bitnami.com/kubernetes/infrastructure/redis/configuration/change-image-version/).
 
+### External DNS
+
+This chart is equipped to allow leveraging the [`external-dns`] project.
+Doing so will enable ExternalDNS to publish the FQDN for each instance,
+in the format of `<pod-name>.<release-name>.<dns-suffix>`.
+Example, when using the following configuration:
```
juan131, line 474:
```suggestion
This chart is equipped to allow leveraging the ExternalDNS project. Doing so will enable ExternalDNS to publish the FQDN for each instance, in the format of `<pod-name>.<release-name>.<dns-suffix>`.
Example, when using the following configuration:
```

bitnami/redis/README.md
```diff
@@ -462,6 +466,38 @@ Bitnami will release a new chart updating its containers if a new version of the
...
+  enabled: true
+  suffix: prod.example.org
+  additionalAnnotations:
+    ttl: 10
+```
+
+On a cluster where the name of the Helm release is `a`, the hostname of a Pod
+is generated as: `a-redis-node-0.a-redis.prod.example.org`.  The IP of
+that FQDN will match that of the associated Pod.  This modifies the following
+parameters of the Redis/Sentinel configuration using this new FQDN:
```
juan131, line 487:
```suggestion
On a cluster where the name of the Helm release is `a`, the hostname of a Pod is generated as: `a-redis-node-0.a-redis.prod.example.org`. The IP of that FQDN will match that of the associated Pod. This modifies the following parameters of the Redis/Sentinel configuration using this new FQDN:
```

bitnami/redis/README.md
```diff
@@ -462,6 +466,38 @@ Bitnami will release a new chart updating its containers if a new version of the
...
+
+* `replica-announce-ip`
+* `known-sentinel`
+* `known-replica`
+* `announce-ip`
+
+:warning: This requires a working installation of [`external-dns`] to be fully
+functional. :warning:
```
juan131, line 495:
``` suggestion
:warning: This requires a working installation of [`external-dns`] to be fully functional. :warning:
```

bitnami/redis/values.yaml
```diff
@@ -23,6 +23,17 @@ global:
...
+## @param useExternalDNS.enabled Enable various syntax that would enable external-dns to work.  Note this requires a working installation of [`external-dns`] to be usable.
+## @param useExternalDNS.additionalAnnotations Extra annotations to be utilized when [`external-dns`] is enabled.
+## @param useExternalDNS.annotationKey The annotation key utilized when [`external-dns`] is enabled.
+## @param useExternalDNS.suffix The DNS suffix utilized when [`external-dns`] is enabled.  Note that we prepend the suffix with the full name of the release.
+##
+useExternalDNS:
+  enabled: false
+  suffix: ""
+  annotationKey: external-dns.alpha.kubernetes.io/
+  additionalAnnotations: {}
```
juan131, line 35: This block shouldn't be under the "Common parameters" section, but in its own section instead at the bottom of the **values.yaml**

juan131 (changes requested):
Thanks for your PR @jtslear ! Please check my comments when you have a chance.

alemorcuq:
Sorry for the late reply. Let's try to get this merged, I'm rerunning the jobs.

jtslear:
@alemorcuq I suspect that https://github.com/bitnami/charts/issues/8662 is preventing the tests from being successful.  The behavior is slightly different, so I'm not 100% confident: Reference log output: https://github.com/bitnami/charts/runs/4842365585?check_suite_focus=true#step:6:1081

alemorcuq:
A PR has been merged as a result of the task you mentioned, so let's check again.

dudizimber:
I'm encountering a situation where I'm using External DNS but did not define a suffix, and now the sentinel has an incomplete `full_hostname`.
The `full_hostname` is being constructed as `${hostname}.redis-service.`.

Wouldn't the correct behavior be to continue using the namespace or headless service if I didn't pass a `redis.externalDNS.suffix` value?

PR #8574, opened 2022-01-05
[bitnami/redis] Adds runAsGroup to containerSecurityContext of master, replica and sentinel containers

baflo:
**Description of the change**

This PR adds a default value for `runAsGroup` in the master, replica and sentinel containers of the `redis` chart. It also adds documentation for this. It does __not__ change the chart templates themselves.

**Benefits**

The `volume-permissions` init containers of the redis instances applies user __and__ group to all files in the `/data` folder. However, the users in the redis/sentinel/master containers by default have `gid = 0`, which sometimes results in the files groups being changed to `0` again.

The benefit of this change is that the file group association is maintained as `1001` by default.

**Possible drawbacks**

Because the default `GID` of the users of the master/replica/sentinel containers is changed from `0` to `1001` by this PR, there might be changed behaviour in runtime.

**Applicable issues**

  - may be related to #152

**Additional information**

**Checklist** 

### DIFF
--- a/bitnami/redis/Chart.yaml
+++ b/bitnami/redis/Chart.yaml
@@ -1,7 +1,7 @@
 annotations:
   category: Database
 apiVersion: v2
-appVersion: 6.3.0
+appVersion: 6.2.9
 dependencies:
   - name: common
     repository: https://charts.bitnami.com/bitnami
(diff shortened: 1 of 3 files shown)


alemorcuq (changes requested):
Thanks for the PR, @baflo. However, as you note, setting the container user group to 1001 by default could produce runtime errors as the container is prepared to use the `root` group by default.

PR #7631, opened 2021-09-27
[bitnami/prestashop] Chart standardized

migruiz4:
**Description of the change**

This PR standardizes the bitnami/prestashop chart to be in line with the rest of the catalog.

**Checklist**

- [X] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [X] Variables are documented in the README.md
- [X] Title of the PR starts with chart name (e.g. [bitnami/<name_of_the_chart>])

PR #7558, opened 2021-09-21
[bitnami/drupal] Chart standardized

migruiz4:
**Description of the change**

This PR standardizes the bitnami/drupal chart to be in line with the rest of the catalog.

**Checklist**

- [X] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [X] Variables are documented in the README.md
- [X] Title of the PR starts with chart name (e.g. [bitnami/<name_of_the_chart>])

PR #8583, opened 2022-01-06
[bitnami/rabbitmq] Support for relabling and aligned service monitor with other charts

nc-kfl:
**Description of the change**

Adds support for Prometheus relabeling of labels. Furthermore it refactores the configuration of label and metrics relabeling to align it with other Bitnami charts.

**Benefits**

Support for Prometheus relabeling

**Possible drawbacks**

Existing use of Metrics Relabeling will not work before it has been reconfigured to use the new metricRelabelings parameter.

**Checklist** 

- [X] Is the PR adding a new parameter? Please, ensure it’s documented in the README.md


bitnami/rabbitmq/Chart.yaml
```diff
@@ -23,4 +23,4 @@ name: rabbitmq
 sources:
   - https://github.com/bitnami/bitnami-docker-rabbitmq
   - https://www.rabbitmq.com
-version: 8.25.1
+version: 9.0.0
```
alemorcuq, line 26:
```suggestion
version: 8.26.0
```

Let's do a minor instead.

bitnami/rabbitmq/templates/servicemonitor.yaml
```diff
@@ -27,8 +27,11 @@ spec:
       {{- if .Values.metrics.serviceMonitor.honorLabels }}
       honorLabels: {{ .Values.metrics.serviceMonitor.honorLabels }}
       {{- end }}
-      {{- if .Values.metrics.serviceMonitor.relabellings }}
-      metricRelabelings: {{- toYaml .Values.metrics.serviceMonitor.relabellings | nindent 6 }}
+      {{- if .Values.metrics.serviceMonitor.relabelings }}
+      relabelings: {{- toYaml .Values.metrics.serviceMonitor.relabelings | nindent 6 }}
+      {{- end }}
+      {{- if .Values.metrics.serviceMonitor.metricRelabelings }}
+      metricRelabelings: {{- toYaml .Values.metrics.serviceMonitor.metricRelabelings | nindent 6 }}
```
alemorcuq, line 34: We can take advantage of this `relabellings` typo to avoid a new major. My proposal is to both the old `relabellings` value and the new `relabelings` and `metricRelabelings`.

If `relabellings` is set, the behavior will be the same as the current behavior: `metricRelabelings` will be set to `relabellings`. If, on the other hand, `relabellings` is not set, we can use the new values and set `relabelings` and `metricRelabelings` using the new values.

The description of the old `relabellings` value should be updated to say that the setting is deprecated and will be removed in the next major. 

Do you like this approach?

alemorcuq (changes requested):
Thanks for the PR, @nc-kfl! We are in the process of making changes to our charts to align them in a standardized way (see #7558 or #7631 for some examples). Because of this, we would like to avoid releasing a new major until all the changes are implemented.

We can refactor this to be a minor instead, please check my comments.

bitnami/rabbitmq/templates/servicemonitor.yaml
nc-kfl: I like this approach really much, I have used your suggestion and refactored it into a minor release.

bitnami/rabbitmq/Chart.yaml
nc-kfl: Refactored it into a minor release.

bitnami/rabbitmq/values.yaml
```diff
@@ -1019,9 +1019,15 @@ metrics:
     ## scrapeTimeout: 30s
     ##
     scrapeTimeout: ""
-    ## @param metrics.serviceMonitor.relabellings Specify Metric Relabellings to add to the scrape endpoint
+    ## @param metrics.serviceMonitor.relabellings MetricsRelabelConfigs to apply to samples before ingestion. DEPRECATED: Will be removed in next major.
     ##
     relabellings: []
+    ## @param metrics.serviceMonitor.relabelings RelabelConfigs to apply to samples before scraping. DEPRECATED: Will be removed in next major.
```
alemorcuq, line 1025: This won't be deprecated.

alemorcuq (changes requested):
Could you please update the README with the new values and descriptions? We have a tool that generates it automatically: https://github.com/bitnami/charts/blob/master/CONTRIBUTING.md#documentation-requirements

nc-kfl:
> Could you please update the README with the new values and descriptions? We have a tool that generates it automatically: https://github.com/bitnami/charts/blob/master/CONTRIBUTING.md#documentation-requirements

Thanks. I simply forgot it :) The readme is updated now.

alemorcuq (approved):
Great, thanks!

PR #8623, opened 2022-01-11
[bitnami/zookeeper] Chart standardized

juan131:

**Description of the change**

This PR standardizes the `bitnami/zookeeper` chart to be in line with the rest of the catalog.

**Benefits**

Standardization

**Possible drawbacks**

Changes the values structure forcing a major bump, check the README.md for more information.

**Applicable issues**

N/A

**Additional information**

N/A

**Checklist** 


bitnami/zookeeper/templates/_helpers.tpl
```diff
@@ -41,40 +52,92 @@ Check if there are rolling tags in the images
...
+*/}}
+{{- define "getValueFromSecret" }}
+    {{- $len := (default 16 .Length) | int -}}
+    {{- $obj := (lookup "v1" "Secret" .Namespace .Name).data -}}
+    {{- if $obj }}
+        {{- index $obj .Key | b64dec -}}
+    {{- else -}}
+        {{- randAlphaNum $len -}}
+    {{- end -}}
+{{- end }}
```
miguelaeh, line 86: I have seen this function replicated in several chart standardization, IMHO we should add it to the common chart.

bitnami/zookeeper/templates/_helpers.tpl
```diff
@@ -107,7 +170,16 @@ Return the name of the secret containing the Keystore and truststore password
 {{- end -}}
 
 {{/*
-Return the secret containing Zookeper client TLS certificates
+Return true if a TLS secret should be created for ZooKeeper client
+*/}}
+{{- define "zookeeper.client.createTlsSecret" -}}
+{{- if and .Values.tls.client.enabled .Values.tls.client.autoGenerated (not .Values.tls.client.existingSecret) }}
```
miguelaeh, line 176: Just to double check, looks like this old function used `tls.quorum` instead of `tls.client` is that change expected? Because I can see we are maintaining both objets

bitnami/zookeeper/values.yaml
```diff
@@ -108,175 +133,196 @@ snapCount: 100000
...
-## @param podSecurityContext.enabled Enable security context (ZooKeeper master pod)
-## @param podSecurityContext.fsGroup Group ID for the container (ZooKeeper master pod)
+customStartupProbe: {}
+## @param lifecycleHooks for the ZooKeeper container(s) to automate configuration before or after startup
+##
+lifecycleHooks: {}
+## ZooKeeper resource requests and limits
+## ref: https://kubernetes.io/docs/user-guide/compute-resources/
+## @param resources.limits The resources limits for the ZooKeeper containers
+## @param resources.requests [object] The requested resources for the ZooKeeper containers
```
miguelaeh, line 289: This object is small enough to appear in the README
```suggestion
## @param resources.requests The requested resources for the ZooKeeper containers
```

miguelaeh (commented):
Thank you for the PR!
Please check the comments

bitnami/zookeeper/templates/_helpers.tpl
juan131: Totally agree! I'll do that in a separate contribution.

bitnami/zookeeper/templates/_helpers.tpl
juan131: I think the GitHub UI confused you. There are two named templates: "zookeeper.quorum.createTlsSecret" and "zookeeper.client.createTlsSecret". The former uses `tls.quorum`, while the latter uses `tls.client`

bitnami/zookeeper/templates/secrets.yaml
```diff
@@ -32,15 +33,11 @@ metadata:
...
 type: Opaque
 data:
-  {{- if or .Values.tls.client.keystorePassword .Values.tls.client.autoGenerated }}
-  keystore-password: {{ (.Values.tls.client.keystorePassword | default (randAlphaNum 10)) | b64enc | quote }}
-  {{- end }}
-  {{- if or .Values.tls.client.truststorePassword .Values.tls.client.autoGenerated }}
-  truststore-password: {{ (.Values.tls.client.truststorePassword | default (randAlphaNum 10))| b64enc | quote }}
-  {{- end }}
+  keystore-password: {{ default (randAlphaNum 10) .Values.tls.client.keystorePassword | b64enc | quote }}
+  truststore-password: {{ default (randAlphaNum 10) .Values.tls.client.truststorePassword | b64enc | quote }}
```
migruiz4, line 37: Why remove the conditionals? It is not common and I think you need to do it intentionally, but as far as I know, Truststore and Keystore can be created without a password.

migruiz4 (changes requested):
Please have a look at my comments below.

I've noticed the following missing values:
- `serviceAccount.annotations`.
- `pdb.minAvailable`.
- `metrics.serviceMonitor.honorLabels`
- `metrics.serviceMonitor.jobLabel`

In addition, please rename `pdb.enabled` as `pdb.create`.

bitnami/zookeeper/templates/secrets.yaml
juan131: The 1st conditional didn't make much sense if random password can be created when it's not provided.
On the other hand, the second conditional didn't make sense, it's intended to auto-generate the TLS secrets instead of the associated passwords.

bitnami/zookeeper/templates/_helpers.tpl
miguelaeh: 👍

bitnami/zookeeper/templates/_helpers.tpl
miguelaeh: 👍

miguelaeh (approved):
LGTM!

migruiz4 (approved):
LGTM! Thank you for adding the missing values

PR #8631, opened 2022-01-12
[bitnami/spring-cloud-dataflow] Remove dollar prompt from command examples in README

onobc:
**Description**
This removes the `$` (dollar prompt) from several copyable command examples in the `bitnami/spring-cloud-dataflow` README to allow the user to copy and used them directly without editing them to remove the `$`. This is part of an improvement effort on the Spring Cloud Dataflow Microsite to make these copyable commands easier to use. The coupling between the 2 repos is due to the fact that the microsite [includes the bitnami README](https://github.com/spring-io/dataflow.spring.io/blob/main/content/documentation/pages/1-installation/3-kubernetes/2-helm.md?plain=1#L42) via 
```
<!--TEMPLATE:https://raw.githubusercontent.com/bitnami/charts/master/bitnami/spring-cloud-dataflow/README.md-->
```

**Benefits**
Without this change, the microsite will have 4 (out of ~50) inconsistent command examples.

**Possible drawbacks**
Since this only changes the `bitnami/spring-cloud-dataflow` README, it leaves the other charts inconsistent in this area (they still have the `$ ` on the copyable command examples). 

**Applicable issues**
Supports https://github.com/spring-io/dataflow.spring.io/issues/280

>  ℹ️ ❓ I just referenced the existing dataflow.spring.io ticket via its full path. I was not sure if I should create another bitnami/charts specific issue. Please let me know if that is the case and I will amend. 

**Checklist** 

### DIFF
--- a/bitnami/spring-cloud-dataflow/README.md
+++ b/bitnami/spring-cloud-dataflow/README.md
@@ -718,7 +718,7 @@ To upgrade to `1.0.0`, you will need to reuse the PVC used to hold the MariaDB d
 
 Obtain the credentials and the name of the PVC used to hold the MariaDB data on your current release:
 
-```console
+```bash
 export MARIADB_ROOT_PASSWORD=$(kubectl get secret --namespace default dataflow-mariadb -o jsonpath="{.data.mariadb-root-password}" | base64 --decode)
 export MARIADB_PASSWORD=$(kubectl get secret --namespace default dataflow-mariadb -o jsonpath="{.data.mariadb-password}" | base64 --decode)
 export MARIADB_PVC=$(kubectl get pvc -l app=mariadb,component=master,release=dataflow -o jsonpath="{.items[0].metadata.name}")
@@ -728,7 +728,7 @@ export RABBITMQ_ERLANG_COOKIE=$(kubectl get secret --namespace default dataflow-
 
 Upgrade your release (maintaining the version) disabling MariaDB and scaling Data Flow replicas to 0:
 
-```console
+```bash
 helm upgrade dataflow bitnami/spring-cloud-dataflow --version 0.7.4 \
   --set server.replicaCount=0 \
   --set skipper.replicaCount=0 \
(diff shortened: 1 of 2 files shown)


bitnami/spring-cloud-dataflow/README.md
```diff
@@ -729,7 +729,7 @@ export RABBITMQ_ERLANG_COOKIE=$(kubectl get secret --namespace default dataflow-
 Upgrade your release (maintaining the version) disabling MariaDB and scaling Data Flow replicas to 0:
 
 ```console
-$ helm upgrade dataflow bitnami/spring-cloud-dataflow --version 0.7.4 \
+helm upgrade dataflow bitnami/spring-cloud-dataflow --version 0.7.4 \
```
juan131, line 732: If we remove the dollar, then we need to change the code block language:
```suggestion
```bash
helm upgrade dataflow bitnami/spring-cloud-dataflow --version 0.7.4 \
```

bono007: Thanks for the feedback @juan131 . I have adjusted to use `bash` for the code block language. Yep, you can tell a difference when viewing it for sure. Good catch.

bitnami/spring-cloud-dataflow/README.md
```diff
@@ -750,8 +750,8 @@ $ helm upgrade dataflow bitnami/spring-cloud-dataflow \
 
 You should see the lines below in MariaDB container logs:
 
-```console
-$ kubectl logs $(kubectl get pods -l app.kubernetes.io/instance=dataflow,app.kubernetes.io/name=mariadb,app.kubernetes.io/component=primary -o jsonpath="{.items[0].metadata.name}")
+```bash
+kubectl logs $(kubectl get pods -l app.kubernetes.io/instance=dataflow,app.kubernetes.io/name=mariadb,app.kubernetes.io/component=primary -o jsonpath="{.items[0].metadata.name}")
```
juan131, line 754: I would revert the changes in this particular case. The rest of code blocks are just telling the user the command to use, however this one is showing an example including the command's output. For this particular case, using `console` + `$` is the right choice since the formatting is more appropriate.

bitnami/spring-cloud-dataflow/README.md
juan131: Thanks!

bitnami/spring-cloud-dataflow/README.md
bono007: Agreed. Adjusted.

juan131 (approved):
LGTM!! It seems the CI/CD it's complaining since you didn't sign your comments. This is a new mandatory requirement that we recently introduced, find more info below:

- https://github.com/bitnami/charts/blob/master/CONTRIBUTING.md#sign-your-work

PR #8672, opened 2022-01-14
[bitnami/nginx] Fix issue with ServiceMonitors

marcellippmann:
This fixes an issue with ServiceMonitors. If  `.Values.metrics.serviceMonitor.enabled` is true and `.Values.metrics.serviceMonitor.additionalLabels` is the empty object, no valid YAML is produced.


miguelaeh (approved):
LGTM!

PR #8767, opened 2022-01-24
[bitnami/redis][bitnami/redis-cluster] Make probes MASTERDOWN aware and include timeout message

### LINKED ISSUE #8016: [bitnami/redis] Readiness probe not checking MASTERDOWN state when using replica-serve-stale-data yes
**Which chart**:
redis (likely all versions, take 15.5.4 as reference)

**Describe the bug**
See https://raw.githubusercontent.com/redis/redis/6.2/redis.conf
```
# When a replica loses its connection with the master, or when the replication
# is still in progress, the replica can act in two different ways:
#
# 1) if replica-serve-stale-data is set to 'yes' (the default) the replica will
#    still reply to client requests, possibly with out of date data, or the
#    data set may just be empty if this is the first synchronization.
#
# 2) If replica-serve-stale-data is set to 'no' the replica will reply with
#    an error "SYNC with master in progress" to all commands except:
#    INFO, REPLICAOF, AUTH, PING, SHUTDOWN, REPLCONF, ROLE, CONFIG, SUBSCRIBE,
#    UNSUBSCRIBE, PSUBSCRIBE, PUNSUBSCRIBE, PUBLISH, PUBSUB, COMMAND, POST,
#    HOST and LATENCY.
#
replica-serve-stale-data yes
```

Unfortunately the documentation currently mentions the wrong message "SYNC with master in progress".
redis.conf file was fixed with correct message on unstable: https://github.com/redis/redis/commit/d50967559251496a9efaaafbf4a29038a6726f9e
Recent and old versions of Redis actually return
`MASTERDOWN Link with MASTER is down and replica-serve-stale-data is set to 'no'`
and that means for us it's not Ready.

**To Reproduce**
Configure a redis replica with:
`replica-serve-stale-data no`

Kill master Pod that feeds this replica

Try to execute any read command on replica like GET [keyname], for example.
It will return `MASTERDOWN Link with MASTER is down and replica-serve-stale-data is set to 'no'`
But readiness probe here won't identify that through PING

**Expected behavior**
It should be possible for readiness probe to take down this unavailable replica.

**Additional context**
This is similar to what is already done when LOADING status is seen. The complication here is that according to documentation PING won't return MASTERDOWN, but succeed instead.
An alternative would be to use a dummy lightweight command to detect MASTERDOWN + LOADING at the same time

(I'm asking on redis repo if this can be fixed for future versions)

eduardobr:
**Description of the change**

Redis 7 will reply `PING` to a Replica with `MASTERDOWN` state when using `replica-serve-stale-data no` and master is disconnected from replica. See PR: https://github.com/redis/redis/pull/9757
This PR adds support to that in a way that Liveness probes will still be successful on `MASTERDOWN` and Readiness probes will fail.
For versions prior to 7, this change has no effect.

Additional: return proper description when probe command times out. It's currently returning empty string to Kubernetes, making it harder to know what happened.
Time out could happen in two levels: Kubernetes call and our own internal `timeout` command, we fix it for our `timeout` command. For Kubernetes it was recently fixed and should be in latest version: https://github.com/kubernetes/kubernetes/issues/106111

See https://github.com/bitnami/charts/issues/8016

**Benefits**

MASTERDOWN must be considered a temporary situation and shouldn't cause a Liveness probe to fail. It just means replica is not ready, but it's running and has potential to come back and keep serving.

**Possible drawbacks**

None identified

**Applicable issues**

  - fixes #8016

**Additional information**

**Checklist**


fmulero:
LGTM

fmulero:
Thanks @eduardobr for your PR.  It will be available in our repositories soon

PR #8796, opened 2022-01-26
[bitnami/kube-prometheus] Standardize Ingress TLS configuration

### LINKED ISSUE #8779: Issue in Kube Prometheus Chart with TLS config for ingress
**Which chart**:

kube-prometheus

**Describe the bug**

There are 3 ingress objects that can be created via the helm chart:
1. prometheus
2. thanos
3. alert manager
  
The Helm chart has a different mechanism for configuring the tls for these ingress resources which makes the usage horrible!  
For Thanos it is done correctly by enabling the user to use cert manager via annotations or to add a tls section of yaml pointing to an existing secret.  
For prometheus and Alert Manager you cant point it at an existing secret making the use of a wildcard cert nearly impossible (would require copying the same cert details into the values.yaml multiple times which is not a good idea as multiple secrets will be created with the same cert and when the cert expires updating it will be horrific as it will exist in too many places.).
  
The chart needs to be standardized as currently i need to:
1. create a secret in advance for thanos
2. Get my TLS Key and certs and place those in plain text in a values yaml for prometheus
3. Get my TLS Key and certs and place those in plain text in a values yaml for Alert Manager

Is this a known issue and is it being worked on?

**To Reproduce**

Deploy with TLS for all components with pre created certs
**Expected behavior**

The same UX for ingress TLS config for all components of a single chart

**Additional context**

vrabbi:
**Description of the change**
This Change standardizes the ingress configuration for the three ingress objects created via this chart.
it makes it possible to utilize pre existing secrets like what the thanos configuration allowed also for prometheus and alert manager.

**Benefits**
Standardization and support for a very common use case where secrets are already present in the cluster with the TLS certs for the ingress objects.

**Possible drawbacks**
may be viewed as a breaking change as we have changed the values.yaml structure for ingress objects

**Applicable issues**

  - fixes #8779 

**Additional information**

**Checklist**


bitnami/kube-prometheus/Chart.yaml
```diff
@@ -34,4 +34,4 @@ sources:
   - https://github.com/bitnami/bitnami-docker-prometheus
   - https://github.com/bitnami/bitnami-docker-alertmanager
   - https://github.com/prometheus-operator/kube-prometheus
-version: 6.6.3
+version: 6.7.3
```
carrodher, line 37: Version should be bumped to `6.7.0`

carrodher (commented):
Thanks for creating this PR! I just added a minor comment about the version.

In the same way, please, note it is needed to sign the commits. Your signature certifies that you wrote the patch or otherwise have the right to contribute the material. You can find more info about this topic, in [the “Sign your work” section from the contributing guidelines](https://github.com/bitnami/bitnami-docker-ASSET/blob/master/CONTRIBUTING.md#sign-your-work).

vrabbi:
i will open a clean new PR

PR #8813, opened 2022-01-27
[bitnami/kafka] add additional kafka broker configuration

### LINKED ISSUE #8637: [IMPROVEMENT] Support of replica.fetch.max.bytes configuration in parameters
**Which chart**:

The Kafka chart on 14.9.0 version.

I would like to have a parameter to override the [replica.fetch.max.bytes](https://kafka.apache.org/documentation/#brokerconfigs_replica.fetch.max.bytes) configuration. Currently, the only way to do it is to have an extra environment variable to set it.

The environment variable fulfills my use case, I just wanted to have everything set with the chart's parameters. This is a minor improvement.

```yml
replicaFetchMaxBytes: 10485760
```

```yml
extraEnvVars:
  - name: KAFKA_CFG_REPLICA_FETCH_MAX_BYTES
    value: "10485760"
```

### LINKED ISSUE #8638: [IMPROVEMENT] Support of min.insync.replicas configuration in parameters
**Which chart**:

The Kafka chart on 14.9.0 version.

I would like to have a parameter to override the [min.insync.replicas](https://kafka.apache.org/documentation/#brokerconfigs_min.insync.replicas) configuration. Currently, the only way to do it is to have an extra environment variable to set it.

The environment variable fulfills my use case, I just wanted to have everything set with the chart's parameters. This is a minor improvement.

```yml
minInsyncReplicas: 1
```

```yml
extraEnvVars:
  - name: KAFKA_CFG_MIN_INSYNC_REPLICAS
    value: "1"
```

jaudiger:
**Description of the change**

Introduce four new Kafka broker parameters:

- unclean.leader.election.enable
- min.insync.replicas
- replica.fetch.min.bytes
- replica.fetch.max.bytes

**Benefits**

Having more flexibility when creating a cluster.

**Possible drawbacks**

None

**Applicable issues**

- Fixes #8637
- Fixes #8638

**Additional information**

None

**Checklist**


juan131 (commented):
Thanks so much for this contribution @jaudiger 

I have mixed feelings about this PR to be honest. We already support [a lot of parameters to configure Kafka environment variables](https://github.com/bitnami/charts/tree/master/bitnami/kafka#kafka-parameters) (too many IMHO) and I'd like to avoid adding more complexity to the chart. Wouldn't it be simpler to add all this extra configuration using the `extraEnvVars` parameter? See:

- https://github.com/bitnami/charts/blob/master/bitnami/kafka/values.yaml#L367

jaudiger:
Hi @juan131,

That's currently the approach I'm using when using this chart (see my issues as reference). But in my opinion, it looks not as clean as if the values were set using parameters from the chart.

For me, the chart should be an abstraction of the service. And this abstraction should handle cases where for example the name of a configuration is updated (as it has already happened during the past).

Using the `extraEnvVars` parameter is a solution that forwards the maintenance to the end-user. Having to use this parameter feels like having something that is not completely handled by the chart.

But I tend to agree with your argument. The question is: where the complexity should be set and where the maintenance should be done? If more parameters are added to the chart, it'll tend to complexify it and raise its maintenance.

fmulero:
Hi @jaudiger I appreciate your contribution but to be perfectly blunt I am not happy to support these changes. It is not giving  any extra value to the users and it will require more maintenance from our side. Imagine what will happen if tomorrow Kafka changes the default value for `KAFKA_CFG_REPLICA_FETCH_MAX_BYTES`, we will need to revisit this code when we currently have the `extraEnvVars ` covering the same need, at the end we will remove it.

jaudiger:
Any update on this one? I understand the arguments here, but the problem seems to be already there with all the others parameters.

But, if you prefer to not add more parameters. Just close this MR, and the two related issues.

fmulero:
Yes you right, we have a lot of `passthrough` variables and we know that they are giving us some headaches. We are trying to avoid just wrapping contiainers, we want to add some value to our solutions, for example validations for complex configurations, waits  for dependencies, simplify the configuration, align it  with other chart/container, etc

PR #8828, opened 2022-01-28
Fix link

sinacek:


carrodher (approved):
LGTM! Thanks for the fix, I just removed the version bump since it is not needed at all for changes in the README (although the automation is going to complain 😉 )

PR #8834, opened 2022-01-30
[bitnami/kube-prometheus] Standardize Ingress in Chart

### LINKED ISSUE #8779: Issue in Kube Prometheus Chart with TLS config for ingress

vrabbi:
**Description of the change**
This Change standardizes the ingress configuration for the three ingress objects created via this chart.
it makes it possible to utilize pre existing secrets like what the thanos configuration allowed also for prometheus and alert manager.

**Benefits**
Standardization and support for a very common use case where secrets are already present in the cluster with the TLS certs for the ingress objects.

**Possible drawbacks**
may be viewed as a breaking change as we have changed the values.yaml structure for ingress objects

**Applicable issues**

  - fixes #8779 

**Additional information**

**Checklist**


vrabbi:
@carrodher this is a clean new PR which was easier to do then cleanup the old one and deal with merge conflicts. same changes as before but have addressed the issue of versioning of the chart and followed the DCO requirement.

migruiz4:
Hi @vrabbi,

I'm sorry but we can not accept these changes, as the Ingress for Kube Prometheus and Alert manager are already standardized to be in line with the rest of the charts in our catalog.

All of the configurations that would be done with the suggested changes can be done by using the existing values `ingress.extraTls` (equivalent to using `ingress.tls` in your PR) and `ingress.extraHosts` (equivalent to `ingress.hosts`):

https://github.com/bitnami/charts/blob/96f9d6b7d6536b0d8fb9144859b8929aad870977/bitnami/kube-prometheus/templates/alertmanager/ingress.yaml#L33-L42
https://github.com/bitnami/charts/blob/96f9d6b7d6536b0d8fb9144859b8929aad870977/bitnami/kube-prometheus/templates/alertmanager/ingress.yaml#L50-L52

The additional code that is being removed by this PR, is intended to make Ingress configuration easier for users looking for more simple deployments.

Indeed, we have plans to standardize the `thanos.ingress` to include the missing features, present in Alert Manager and Kube Prometheus Ingress templates.

If you have any questions about how to move from `ingress.tls` to `ingress.extraTls`, feel free to let us know.

vrabbi:
i had not seen the extraTLS before which seems it should work however the description is very misleading on the variable as it is not for additional hostnames rather it replaces tls section from my understanding of the chart

migruiz4:
I'm sorry the description for `extraTls` was not clear enough. 
Maybe something more simple like "Extra TLS settings for the ingress" would be clearer and not mislead users to confusion?

In any case, I will close this PR since it no longer applies. Thank you for your understanding.

vrabbi:
Or something like manualTls as it isnt extra its either or from my understanding. ExtraTls even in the name implies it is additional to tls not another option instead of tls

PR #8861, opened 2022-02-01
[bitnami/common] Improve "common.secrets.passwords.manage" helper

juan131:

**Description of the change**

The current **"common.secrets.passwords.manage"** is a helper to manage secret passwords for a given secret/key. It basically does two things:

- If the given secret exists (release upgrade), it reuses the secret password that was previously set in the existing secret for the given key.
- If the given secret doesn't exist (new release installation), it creates a random secret password if the corresponding parameter is not set.
- If the given secret doesn't exist (new release installation), it creates a secret password based on a parameter if that parameter is set.

This help work flawlessly but there's a scenario it doesn't take into account:

- what happens if the given secret exists but the given key does not? E.g. upgrading between two chart versions where a key was renamed from `foo-pass` to `bar-pass`. 
 
This PR tries to answer this question by throwing a proper error. 

**Benefits**

The **"common.secrets.passwords.manage"** helper manages existing secrets that don't contain an expected key properly.

**Possible drawbacks**

None

**Applicable issues**

N/A

**Additional information**

N/A

**Checklist**

### DIFF
--- a/bitnami/common/Chart.yaml
+++ b/bitnami/common/Chart.yaml
@@ -2,7 +2,7 @@ annotations:
   category: Infrastructure
 apiVersion: v2
 # Please make sure that version and appVersion are always the same.
-appVersion: 1.11.0
+appVersion: 1.11.1
 description: A Library Helm Chart for grouping common logic between bitnami charts. This chart is not deployable by itself.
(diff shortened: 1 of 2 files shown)


migruiz4 (approved):
LGTM

PR #8864, opened 2022-02-01
[bitnami/node] Added possibility to overwrite npm install-command

jwleon:
**Description of the change**

This pull request adds a Node parameter to overwrite the default install command (`npm install`) with a custom one.

**Benefits**

There are deloy scenarios, in which `npm install` is not desired upon initialization - for example, when a complete and tested "fixed package", including node_modules, is put together through CI workflows and saved in a "distribution"-repository. In those cases, it is desired to prevent npm from changing node_modules, package.json or package-lock.json by accident. I thought it would be better to give the option to overwrite the install-command instead of just making it optional.

**Possible drawbacks**

This feature should be non-breaking, since the default option is the same as the hard-coded command before the change. In case someone had, for unknown-reasons, `installCommand` in their helm charts before, the deployments might be affected by this change. I guess that is rather unlikely since `installCommand` was not used before.

**Applicable issues**

None

**Additional information**

None

**Checklist**

### DIFF
--- a/bitnami/node/Chart.yaml
+++ b/bitnami/node/Chart.yaml
@@ -28,4 +28,4 @@ name: node
 sources:
   - https://github.com/bitnami/bitnami-docker-node
   - http://nodejs.org/
-version: 16.2.5
+version: 16.3.0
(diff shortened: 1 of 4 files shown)


fmulero (approved):
Thanks @jwleon. It looks good to me

PR #8886, opened 2022-02-03
Changed CRLF to LF

stefan-jonasson:
The file was committed with CRLF (all other templates use LF).


carrodher (approved):
Thanks, changes LGTM

Issue #8856, opened 2022-02-01
bitnami/kafka: tls not working properly with existingSecrets


richardporras:
bitnami/kafka: 15.0.1

`/scripts/setup.sh` is looking for the keystore `/certs-${ID}/kafka-keystore.jks`. In the chart documentation, this file is spelled with a dot. 
```
kubectl create secret generic kafka-jks-0 --from-file=kafka.truststore.jks=./kafka.truststore.jks --from-file=**kafka.keystore.jks**=./kafka-0.keystore.jks
```
If I create the secret as described in the documentation I got the error:
```
Couldn't find the expected Java Key Stores (JKS) files! They are mandatory when encryption via TLS is enabled.
```
This is because `/scripts/setup.sh` is expecting `kafka-keystore.jks` and not `kafka.keystore.jks`. 

In my opinion `/scripts/setup.sh` should be modified in the kafka docker image.
Instead of:
```
JKS_KEYSTORE="/certs-${ID}/kafka-keystore.jks"
```
it should be:
```
JKS_KEYSTORE="/certs-${ID}/kafka.keystore.jks"
```

carrodher:
Thanks for reporting this issue. Would you like to contribute by creating a PR to solve the issue? The Bitnami team will be happy to review it and provide feedback.

Issue #8660, opened 2022-01-13
 Bitnami Kafka: Couldn't find the expected Java Key Stores (JKS) files!


pierrecoppolino:
Hello,

I find a little mistake in file /bitnami/kafka/templates/scripts-configmap.yaml ligne 123   JKS_KEYSTORE="/certs-${ID}/kafka-keystore.jks"  => should be   JKS_KEYSTORE="/certs-${ID}/kafka.keystore.jks"

Regards 
Pierre

carrodher:
Thanks for reporting this issue. Would you like to contribute by creating a PR to solve the issue? The Bitnami team will be happy to review it and provide feedback.

Issue #3539, opened 2020-08-27
PLAINTEXT for Internal and SSL for external clients


etavene:
**Which chart**:
kafka charts 

**Describe the bug**

I want to configure internal kafka consumers  with PLAINTEXT and External consumers with SSL. Is this supported in current kafka charts ?  

listener.security.protocol.map=INTERNAL:PLAINTEXT,CLIENT:PLAINTEXT,EXTERNAL:SSL

I am able to configure both either SSL or PLAINTEXT

javsalgar:
Hi,

I think that the chart value `listenerSecurityProtocolMap` should do the trick. Please let us know if this works for you

etavene:
@javsalgar Thanks for your reply. The said option didnt work. 

If I mention auth.clientProtocol as PLAINTEXT,  all the information related to  ssl.keystore  ssl.truststore missing in 
/opt/bitnami/kafka/config/server.properties file. 

So, I changed the settings as shown below

auth:
  clientProtocol: tls
  interBrokerProtocol: plaintext
  jksSecret: kafka-jks2
listeners: []
advertisedListeners: []
listenerSecurityProtocolMap: "INTERNAL:PLAINTEXT,CLIENT:PLAINTEXT,EXTERNAL:SSL" 

The kafka log shows " Configuring Kafka for client communications with PLAINTEXT authentication"
So the server.properties doesn't have any entries related to truststore and keystore. 

The libkafka.sh file shows, only if protocol is SSL, then only ssl related information is available in server.properites

 if [[ "$protocol" = "SASL_SSL" ]] || [[ "$protocol" = "SSL" ]]; then
            kafka_configure_ssl

kafka_configure_ssl() 
    kafka_server_conf_set ssl.keystore.location "$KAFKA_CONF_DIR"/certs/kafka.keystore.jks

so the libkafka.sh is only considering $client_protocol as either SSL or PLAINTEXT, it is not considering external client protocol.

javsalgar:
Thank you very much for the input. One last thing, did you try injecting your custom configuration file? If the configuration is injected then all the auto-configuration steps would be skipped.

etavene:
Yes, I injected configuration, as you told all auto-configuration steps were skipped and i was able to acheive what I wanted.  But if we skip auto-configuration, parameters such as KAFKA_CFG_ADVERTISED_LISTENER and KAFKA_CFG_BROKER_ID  are hard to manage per instance in a statefulset configuration.

[...]

PR #8902, opened 2022-02-04
[bitnami/kafka] Add external clients setting

stefan-jonasson:
**Description of the change**

Added a new setting `auth.externalClientProtocol` to configure the protocol for external clients. 

The new setting will default to the setting of internal clients so this change is backward compatible. 

Aslo fixed the "spelling" error of `kafka.keystore .jks` as described in issues: #8660 and #8856 

**Benefits**

With this change the internal/external clients could have protocols configured for their listeners, for example `PLAINTEXT` for internal clients and `SSL` for external. There are several reasons (simplicity/performance) for why you would like to have different settings for external clients. 

This change requires at least kafka docker image version `3.1.0-debian-10-r8` to work. The same effect could be achieved by manually specifying `listenerSecurityProtocolMap` but by doing so other internal clients such as `kafka-exporter` could get an incorrect configuration and not work as expected. 

**Possible drawbacks**

None

**Applicable issues**

Fixes:
 - #3539
 - #8660 
 - #8856  

**Checklist**


bitnami/kafka/templates/_helpers.tpl
```diff
@@ -157,6 +157,23 @@ Return true if encryption via TLS for client connections should be configured
 {{- end -}}
 {{- end -}}
 
+{{/*
+Return the configured value for the external client protocol, defaults to the same value as clientProtocol
```
juan131, line 161: This isn't true: it's returning a boolean. If you want to return the 1st non-empty value, use coalesce:
```
   {{- coalesce .Values.auth.externalClientProtocol .Values.auth.clientProtocol -}}
```

bitnami/kafka/templates/scripts-configmap.yaml
```diff
@@ -121,7 +121,7 @@ data:
     {{- if eq .Values.auth.tls.type "jks" }}
     {{- if not (empty .Values.auth.tls.existingSecrets) }}
     JKS_TRUSTSTORE={{ printf "/%s/%s" (ternary "certs-${ID}" "truststore" (empty $jksTruststoreSecret)) (default "kafka.truststore.jks" $jksTruststore) | quote }}
-    JKS_KEYSTORE="/certs-${ID}/kafka-keystore.jks"
+    JKS_KEYSTORE="/certs-${ID}/kafka.keystore.jks"
```
juan131, line 124: Thanks!

bitnami/kafka/values.yaml
```diff
@@ -225,14 +225,16 @@ auth:
   ## | sasl      | Yes (via SASL)                | No                 |
   ## | sasl_tls  | Yes (via SASL)                | Yes                |
   ## @param auth.clientProtocol Authentication protocol for communications with clients. Allowed protocols: `plaintext`, `tls`, `mtls`, `sasl` and `sasl_tls`
+  ## @param auth.externalClientProtocol Authentication protocol for communications with external clients. Defaults to value of `auth.clientProtocol`. Allowed protocols: `plaintext`, `tls`, `mtls`, `sasl` and `sasl_tls`
   ## @param auth.interBrokerProtocol Authentication protocol for inter-broker communications. Allowed protocols: `plaintext`, `tls`, `mtls`, `sasl` and `sasl_tls`
   ##
   clientProtocol: plaintext
+  externalClientProtocol: ""
```
juan131, line 232: This parameter should have a default protocol (as the rest)

bitnami/kafka/ci/values-with-external-tls.yaml
```diff
@@ -0,0 +1,11 @@
+# Test values file for generating hidden section of the yaml
+# and check that the rendering is correct
+replicaCount: 3
```
juan131, line 3: You should also enable external access, right?
```suggestion
replicaCount: 3
externalAccess:
  enabled: false
```

bitnami/kafka/Chart.yaml
```diff
@@ -29,4 +29,4 @@ name: kafka
 sources:
   - https://github.com/bitnami/bitnami-docker-kafka
   - https://kafka.apache.org/
-version: 15.0.4
+version: 15.0.5
```
juan131, line 32: You're including a new parameter, then you need to bump the minor version
```suggestion
version: 15.1.0
```

juan131 (changes requested):
Thank you so much for this awesome contribution @stefan-jonasson !! Please check my comments when you have a chance.

bitnami/kafka/templates/_helpers.tpl
stefan-jonasson: @juan131 As I understand it `or` is the same or as in go-templates and it's returning the first value that evaluates as boolean true.  See: https://pkg.go.dev/text/template#hdr-Functions. I opted for this because I wanted it to work with empty strings as well as undefined/null values. But coalesce might be better if it works with empty strings?

bitnami/kafka/values.yaml
stefan-jonasson: It's intentionally set to "" so that it will default to the configured value of clientProtocol. If we specify it here it will break a lot of existing configs.

bitnami/kafka/ci/values-with-external-tls.yaml
```diff
@@ -0,0 +1,15 @@
...
+  tls:
+    type: jks
+    existingSecrets:
+      - kafka-secret-0
+      - kafka-secret-1
+      - kafka-secret-2
+externalAccess:
+  enabled: true
+  autoDiscovery:
+    enabled: true
```
juan131, line 15: Please address the "No newline at end of file" warning

bitnami/kafka/templates/_helpers.tpl
juan131: It might work using `or` because of this particular implementations of `or` in go-templates. That said, I find it confusing (so I guess it can be confusing for others too). I'd stick to the `coalesce` function for consistency with the res of the catalog, see:
- https://masterminds.github.io/sprig/defaults.html#coalesce

bitnami/kafka/values.yaml
juan131: I can see your point... Let's add a warning comment to avoid this to be broken in the future:
```suggestion
  # Note: empty by default for backwards compatibility reasons, find more information at
  # https://github.com/bitnami/charts/pull/8902/
  externalClientProtocol: ""
```

bitnami/kafka/templates/_helpers.tpl
stefan-jonasson: Ah, didn't realise that coalesce worked with "empty". I'll change it!

juan131 (approved):
LGTM

PR #8904, opened 2022-02-04
[bitnami/external-dns]fix: add commonAnnotations to deployment

Mattes83:
**Description of the change**
It would be great to be able to add annotations for the main deployment.
Our use case is that we automatically inspect all components running in the cluster to give advice about best practices etc. To ignore some rules per object we need to be able to set annotations.

There is already a property called commonAnnotations, but it is not applied to the deployment

**Checklist**

- [ x] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [ x] Variables are documented in the values.yaml and added to the README.md using (readme-generator-for-helm)[https://github.com/bitnami-labs/readme-generator-for-helm]
- [ x] All commits signed off and in agreement of [Developer Certificate of Origin (DCO)](https://github.com/bitnami/charts/blob/master/CONTRIBUTING.md#sign-your-work)

### DIFF
--- a/bitnami/external-dns/Chart.yaml
+++ b/bitnami/external-dns/Chart.yaml
@@ -24,4 +24,4 @@ sources:
   - https://github.com/kubernetes-sigs/external-dns
   - https://github.com/bitnami/bitnami-docker-external-dns
   - https://github.com/kubernetes-sigs/external-dns
-version: 6.1.3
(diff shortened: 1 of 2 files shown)


javsalgar (approved):
LGTM, Thanks!

PR #8950, opened 2022-02-09
[bitnami/kafka] Include new parameter to allow setting external names for load balanced external listeners

### LINKED ISSUE #8875: Kafka external access does not work with TLS enabled
**Which chart**:

kafka-14.1.0

**Describe the bug**

We have configured external access to kafka (type loadBalancer) with client protocol SASL_TLS enabled. Access from clients does not work, it gives the following error:

```
ERROR [Consumer clientId=consumer-console-consumer-68764-1, groupId=console-consumer-68764] Connection to node 2147483647 (x.x.x.x/x.x.x.x:9094) failed authentication due to: SSL handshake failed (org.apache.kafka.clients.NetworkClient)
ERROR Error processing message, terminating consumer process:  (kafka.tools.ConsoleConsumer$)
org.apache.kafka.common.errors.SslAuthenticationException: SSL handshake failed
Caused by: javax.net.ssl.SSLHandshakeException: No subject alternative names matching IP address x.x.x.x found

```

**To Reproduce**

1. generate TLS cert for kafka server with three hostnames in SAN (DNS records already exist)
Add the key, cert and CA cert in a kubernetes secret object 'kafka-pem'. 

2. install kafka with the following values.yaml:

```yaml
kafka:
  replicaCount: 3 
  persistence: 
    enabled: true
  readinessProbe:
    initialDelaySeconds: 60
  livenessProbe:
    initialDelaySeconds: 60     
  externalAccess:
    enabled: true
    service:
      type: LoadBalancer
      port: 9094
      loadBalancerIPs: 
        - x.x.x.x
        - y.y.y.y
        - z.z.z.z
    
  auth:
    clientProtocol: sasl_tls
    interBrokerProtocol: plaintext

    tls:    
      type: pem
      existingSecret: kafka-pem
      
    sasl:
      jaas:
        existingSecret: kafka-secrets    
        interBrokerUser: brokeruser
        zookeeperUser: zookeeperuser
        clientUsers:
          - kafka-user

  allowPlaintextListener: true
  deleteTopicEnable: true
  autoCreateTopicsEnable: true 

  zookeeper:
    allowAnonymousLogin: false
    replicaCount: 3
    auth:
     enabled: true
     existingSecret: kafka-zookeeper-secrets
     clientUser: zookeeperuser
     serverUsers: zookeeperuser

```

3. set up a client using trusting the same CA cert

4. run kafka shell commands in client, e.g.
`kafka-console-producer.sh --producer.config /config/client-secret --bootstrap-server kafka-0.mydomain.com:9094 --topic test`

**Expected behavior**

The clients should be able to connect to the server, create topics, etc.

**Further investigations**

kafka server.properties file contains only the IP address of the kafka node in advertised listeners:
```
I have no name!@kafka-0:/$ cat /opt/bitnami/kafka/config/server.properties |grep -i advertised
advertised.listeners=INTERNAL://kafka-0.kafka-headless.kafka.svc.cluster.local:9093,CLIENT://kafka-0.kafka-headless.kafka.svc.cluster.local:9092,EXTERNAL://x.x.x.x:9094

```
Which causes the problem, as it advertises only the IP address to the client, and that does not match the SAN hostnames in the server certificate. It seems the chart does not have any configuration option to provide kafka with the correct hostnames used for external listener.

We also tested by adding the external host IPs to the certificate SAN list, and it worked fine. However, defining IP addresses in SAN is really not optimal, as in the end we'd like to use a wildcard certificate. Is there any way - using the chart parameters - to do this properly. Can we make kafka to advertise hostnames instead of IP addresses?

fmulero:
**Description of the change**
Include parameter `externalAccess.service.loadBalancerNames` to allow external names be exposed by kafka brokers

**Benefits**

This change allows the use of FQDN in the advertised listeners shared by kafka brokers. This is useful when external access and TLS are enabled and the certificates are provisioned. In that scenario Kafka brokers should advertise external names to the client, and this should match with the subject or the SAN hostnames in the server certificate. See #8875 

**Possible drawbacks**

Not identified

**Applicable issues**

  - fixes #8875

**Checklist**


bitnami/kafka/templates/scripts-configmap.yaml
```diff
@@ -88,7 +88,7 @@ data:
     {{- if .Values.externalAccess.autoDiscovery.enabled }}
     export EXTERNAL_ACCESS_IP="$(<${SHARED_FILE})"
     {{- else }}
-    export EXTERNAL_ACCESS_IP=$(echo '{{ .Values.externalAccess.service.loadBalancerIPs }}' | tr -d '[]' | cut -d ' ' -f "$(($ID + 1))")
+    export EXTERNAL_ACCESS_IP=$(echo '{{ .Values.externalAccess.service.loadBalancerNames | default .Values.externalAccess.service.loadBalancerIPs }}' | tr -d '[]' | cut -d ' ' -f "$(($ID + 1))")
```
juan131, line 91: Two comments:
- Maybe we need to rename the env. var `EXTERNAL_ACCESS_IP` since it's not very accurate now.
- `externalAccess.service.loadBalancerNames` takes precedence over `.Values.externalAccess.service.loadBalancerIPs`, right? You should state that in the **values.yaml**

bitnami/kafka/Chart.yaml
```diff
@@ -29,4 +29,4 @@ name: kafka
 sources:
   - https://github.com/bitnami/bitnami-docker-kafka
   - https://kafka.apache.org/
-version: 15.1.0
+version: 15.1.1
```
juan131, line 32: You're adding a new parameter, therefore you should bump the minor version:
```suggestion
version: 15.2.0
```

juan131 (changes requested):
Thanks for this PR @fmulero ! Please check my comments.

bitnami/kafka/templates/scripts-configmap.yaml
fmulero: Yes you right, is not accurate accurate using `_IP` suffix in this case. To be honest I didn't change the name of the variable because for NodePort external access the was followed, but I can do it, it will be more readable.
No problem about the documentation

bitnami/kafka/templates/NOTES.txt
```diff
@@ -78,6 +79,16 @@ IPs for Kafka brokers. To complete your deployment follow the steps below:
...
+{{- if and .Values.externalAccess.enabled (not .Values.externalAccess.autoDiscovery.enabled) (gt $loadBalancerNameListLength 0) (not (eq $replicaCount $loadBalancerNameListLength)) (eq .Values.externalAccess.service.type "LoadBalancer") }}
+---------------------------------------------------------------------------------------------
+ WARNING
+
+    If "externalAccess.service.loadBalancerNames" is specified it should have the same number
+    of names as replicas has been deployed. This deployment could be incomplete, some of the
+    brokers could be not reachable for outside of K8S
+
+---------------------------------------------------------------------------------------------
+{{- end}}
```
juan131, line 91: This shouldn't be handled here but using a "validateValues" helper, check the one below as reference:
- https://github.com/bitnami/charts/blob/master/bitnami/kafka/templates/_helpers.tpl#L359

bitnami/kafka/templates/scripts-configmap.yaml
juan131: Thanks for renaming it! Please add a comment in the **values.yaml** with the precedence criteria.

bitnami/kafka/templates/NOTES.txt
fmulero: Initially I didn't include it because that implies a failure in the installation and for a similar `externalAccess.service.loadBalancerIPs` validation the approach was warn the user, I had my doubts and I decided to use the same approach as in `externalAccess.service.loadBalancerIPs` validation. But I agree, raise a failure could be more accurate in this case

bitnami/kafka/templates/scripts-configmap.yaml
fmulero: I thought I included it 😅 , I'll write it in a clearer way

fmulero:
Due to a mistake I created a lot of noise in this PR. I've just created a new one #8983

PR #9004, opened 2022-02-14
Remove unused actions

carrodher:
### Review checklist
This action is not checking anything related to the Helm chart itself, just generating an automated comment if changes in the _values.yaml_ are applied so the reviewer can double-check those action items and apply the changes in the app-catalogue.

Screenshot 2022-02-08 at 17 43 04

We are removing this action because:
 - We (as reviewers) are not paying attention to this comment
 - It's weird for users
 - The validation of new containers added as part of Helm charts should be done as part of the automated test & release pipeline

### Generate README file

This GH action is running the [readme-generator](https://github.com/bitnami-labs/readme-generator-for-helm) *but it's not updating the README file*. It is just a "dry run" execution.

This tool can be configured to 1) update the readme (`-v`/`--values` flag) and/or 2) check the metadata (`-m`/`--metadata` flag).
In the first case, which is the one used in the GH action, it is needed to commit the changes in the README, otherwise, the execution doesn't have sense.

In the pipeline, at the chart-package step, we are executing the tool with the proper flags to check the metadata, the job will fail if there is an issue in the metadata. 

There is an internal task to update the README file in a proper way from the test & release pipeline which is the one updating the files for every automated release.


javsalgar (approved):

juan131 (approved):
LGTM

PR #9011, opened 2022-02-14
[bitnami/kubernetes-event-exporter] Chart standardised

migruiz4:
**Description of the change**

This PR standardizes the bitnami/kubernetes-event-exporter chart to be in line with the rest of the catalog.

**Checklist**


bitnami/kubernetes-event-exporter/templates/deployment.yaml
```diff
@@ -72,19 +102,33 @@ spec:
...
+          {{- if .Values.lifecycleHooks }}
+          lifecycle: {{- include "common.tplvalues.render" (dict "value" .Values.lifecycleHooks "context" $) | nindent 12 }}
+          {{- end }}
+          {{- if .Values.startupProbe.enabled }}
+          startupProbe: {{- omit .Values.startupProbe "enabled" | toYaml | nindent 12 }}
+          {{- else if .Values.customStartupProbe }}
+          startupProbe: {{- include "common.tplvalues.render" (dict "value" .Values.customStartupProbe "context" $) | nindent 12 }}
+          {{- end }}
+          {{- if .Values.livenessProbe.enabled }}
+          livenessProbe: {{- omit .Values.livenessProbe "enabled" | toYaml | nindent 12 }}
```
javsalgar, line 115: Which is the check for the liveness probe?

bitnami/kubernetes-event-exporter/templates/deployment.yaml
```diff
@@ -72,19 +102,33 @@ spec:
...
           {{- end }}
+          resources: {{- toYaml .Values.resources | nindent 12 }}
           volumeMounts:
             - mountPath: /data
               name: cfg
-      {{- if .Values.podSecurityContext.enabled }}
-      securityContext: {{- omit .Values.podSecurityContext "enabled" | toYaml | nindent 8 }}
-      {{- end }}
-      serviceAccountName: {{ include "kubernetes-event-exporter.serviceAccountName" . }}
+        {{- if .Values.sidecars }}
```
javsalgar, line 129: Missing extraVolumeMounts and extraVolumes

javsalgar (changes requested):
Thank you! Apart from the comments above, I see that the file rbac-leader-election.yaml is not following our YAML format standards (for example indentation), please check it.

Also, in the NOTES.txt, I would also add a note on how to check the status of the pod, similar to what we do in Kubewatch

```

To verify that kubewatch has started, run:

  kubectl get deploy -w --namespace {{ .Release.Namespace }} {{ template "common.names.fullname" . }}

{{- include "kubewatch.checkRollingTags" . }}
```

Also, the rollingTags check is missing.

bitnami/kubernetes-event-exporter/templates/deployment.yaml
migruiz4: Nice catch! Both values existed in the `values.yaml`, but I didn't notice they had no implementation.
Fixing it.

bitnami/kubernetes-event-exporter/templates/deployment.yaml
migruiz4: All probes are disabled by default.

I was thinking about not adding probes at all since the chart doesn't even expose a containerPort. But then I saw we already do this in the `bitnami/kubewatch` chart, which I'm also standardising:
https://github.com/bitnami/charts/blob/dd28dccabfa9d9e917a9e12a6b8d887837f0cd87/bitnami/kubewatch/templates/deployment.yaml#L84-L93

Since it is interpreted as a template, users could define their probe command under `livenessProbe`, but imho it doesn't make sense to have both `livenessProbe` and `customLivenessProbe`.

What do you think would fit in both cases?

javsalgar (approved):

bitnami/kubernetes-event-exporter/templates/deployment.yaml
```diff
@@ -72,20 +102,40 @@ spec:
...
           {{- end }}
-          resources: {{- toYaml .Values.resources | nindent 12 }}
-          {{- if .Values.containerSecurityContext.enabled }}
-          # yamllint disable rule:indentation
-          securityContext: {{- omit .Values.containerSecurityContext "enabled" | toYaml | nindent 12 }}
-          # yamllint enable rule:indentation
+          {{- if not .Values.diagnosticMode.enabled }}
+          {{- if .Values.lifecycleHooks }}
+          lifecycle: {{- include "common.tplvalues.render" (dict "value" .Values.lifecycleHooks "context" $) | nindent 12 }}
+          {{- end }}
```
juan131, line 108: Why would you exclude the hook when enabling diagnostic mode?
```suggestion
          {{- if .Values.lifecycleHooks }}
          lifecycle: {{- include "common.tplvalues.render" (dict "value" .Values.lifecycleHooks "context" $) | nindent 12 }}
          {{- end }}
          {{- if not .Values.diagnosticMode.enabled }}
```

bitnami/kubernetes-event-exporter/templates/deployment.yaml
juan131: I would keep only `customXXX` removing `livenessProbe`, `readinessProbe` and `startupProbe` parameters in this case.

juan131 (changes requested):
Thanks for the PR @migruiz4 ! Please check my comments

bitnami/kubernetes-event-exporter/templates/deployment.yaml
migruiz4: Not sure if it is intentional or a typo, but I have recently seen that other charts disable `lifecycleHooks` when `diagnosticMode` is enabled. Those charts are Redis, Oauth2-proxy, Postgresql-ha, and Contour.

bitnami/kubernetes-event-exporter/templates/deployment.yaml
migruiz4: I will then update `kubewatch` PR too.

bitnami/kubernetes-event-exporter/templates/deployment.yaml
juan131: I'd say it's a typo

juan131 (approved):
LGTM

bitnami/kubernetes-event-exporter/values.yaml
```diff
@@ -195,16 +251,27 @@ nodeAffinityPreset:
...
 ## extraEnvVars:
 ##  - name: MY_ENV_VAR
 ##    value: env_var_value
 ##
 extraEnvVars: []
-## @param extraEnvVarsConfigMap ConfigMap containing extra env vars to be added to all containers
+## @param extraEnvVarsCM ConfigMap containing extra env vars to be added to all containers
 ##
-extraEnvVarsConfigMap: ""
+extraEnvVarsCM: ""
```
mkilchhofer, line 274: May I ask why such changes are allowed within a minor chart bump?

mkilchhofer: Ah.. only fixed in values.yaml and was already there as `extraEnvVarsCM` inside the deployment. 👌

migruiz4: Hi @mkilchhofer, thank you for asking!

The value `extraEnvVarsConfigMap` referenced in the `values.yaml` was indeed a typo. The implemented value was `extraEnvVarsCM` as you can see in version  `1.3.2`:
https://github.com/bitnami/charts/blob/8cad18aed9966a6f0208e5ad6cee46cb217f47ab/bitnami/kubernetes-event-exporter/templates/deployment.yaml#L66-L69

So this isn't really a non-backward compatible change, but a bug fix.

PR #9016, opened 2022-02-15
[bitnami/postgresql] Allow enabling TLS without volume-permissions

### LINKED ISSUE #8990: [bitnami/postgresql] Volume permission container should not be required to use TLS
### Name and Version

bitnami/postgresql 11.0.2

It should be possible to enable TLS for PostgreSQL without having to enable the volume permission container, because I'd like to avoid containers that run as root.

Rework the certificate file permission handling to not require the volume permission container.

I've temporarily removed `runAsNonRoot: true` from the pod security context to allow the volume permission container to run, but I'd much prefer disabling it.

juan131:

**Description of the change**

It shouldn't be necessary to run a "root" container to adapt volume permissions to enable TLS. This PR address that.

**Benefits**

TLS without using a "root" container.

**Possible drawbacks**

None

**Applicable issues**

  - fixes #8990

**Additional information**

N/A

**Checklist**

### DIFF
--- a/bitnami/postgresql/Chart.yaml
+++ b/bitnami/postgresql/Chart.yaml
@@ -26,4 +26,4 @@ name: postgresql
 sources:
   - https://github.com/bitnami/bitnami-docker-postgresql
   - https://www.postgresql.org/
-version: 11.0.4
+version: 11.0.5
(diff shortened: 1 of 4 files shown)


bitnami/postgresql/Chart.yaml
```diff
@@ -26,4 +26,4 @@ name: postgresql
 sources:
   - https://github.com/bitnami/bitnami-docker-postgresql
   - https://www.postgresql.org/
-version: 11.0.4
+version: 11.0.5
```
marcosbc, line 29: Given that this is implementing a new feature, we should at least increase the minor version.

marcosbc (approved):
LGTM with comment

bitnami/postgresql/Chart.yaml
juan131: I don't think it's a new feature but removing a blocking condition that made no sense.

Issue #7464, opened 2021-09-10
[bitnami/postgresql-ha] pre-stop.sh script fails to wait for new primary node.


GuoqiangZ:
**Which chart**:
postgresql-ha

**Describe the bug**
As commented in the pre-stop.sh script, when the primary node terminates it should wait for the new primary to come up. 

https://github.com/bitnami/charts/blob/a5fee544b20ceb8cd6131d511e221a74e2a89998/bitnami/postgresql-ha/templates/postgresql/hooks-scripts-configmap.yaml#L61-L72

But during the experiments, the primary node leaves before the new primary coming up.

The root cause is in the following line 

https://github.com/bitnami/charts/blob/a5fee544b20ceb8cd6131d511e221a74e2a89998/bitnami/postgresql-ha/templates/postgresql/hooks-scripts-configmap.yaml#L61

If you take a look into `repmgr_get_primary_node ` in [librepmgr.sh](https://github.com/bitnami/bitnami-docker-postgresql-repmgr/blob/4c84b71236e429435fbe5e7dc66379a76171e7a7/11/debian-10/rootfs/opt/bitnami/scripts/librepmgr.sh#L199-L238), you can find that, when executed in primary node, the returned primary node is `"" <port>`,  so https://github.com/bitnami/charts/blob/a5fee544b20ceb8cd6131d511e221a74e2a89998/bitnami/postgresql-ha/templates/postgresql/hooks-scripts-configmap.yaml#L67 always returns `false`. This results in primary node never waiting for the new one and leaving the cluster immediately. This occasionally leads to split brain issue from our observation.

**To Reproduce**

1. Prepare configuration files for 3 replicas of Postgres node.
2. Append following lines in `pre-stop.sh` and set env `BITNAMI_DEBUG` to be `true` for postgresql container.
```
info "Primary node is $primary_node"
info "Primary host is $primary_host"

# Output logs into a FailedPreStopHook event
exit 1
```
3. Apply configuration and wait for all 3 nodes coming up. Then terminate the primary node.
4. Find logs in FailedPreStopHook event.

```
pre-stop-hook 23:26:43.80 DEBUG ==> Configuring libnss_wrapper...
pre-stop-hook 23:26:43.81 INFO  ==> Querying all partner nodes for common upstream node...
[...]

Issue #7163, opened 2021-08-06
Both nodes becomes primary causing stale/old data in the Database


DevvG:
**Which chart**:
postgresql-ha 11.10.0

Postgres image - bitnami/postgresql-repmgr:13.1.0-debian-10-r63

**Describe the bug**
In a two pod ha setup, sometime both pods becomes primary in postgres-ha. As a result, data is not getting synchronized and the new primary shows old data. Also data in the new primary never syncs with new standby. 

Following is the result from the command '/opt/bitnami/scripts/postgresql-repmgr/entrypoint.sh repmgr -f /opt/bitnami/repmgr/conf/repmgr.conf node check'

POD-0

Node "xxxx-postgres-postgresql-ha-postgresql-0":
        Server role: OK (node is primary)
        Replication lag: OK (N/A - node is primary)
        WAL archiving: OK (0 pending archive ready files)
        Upstream connection: OK (N/A - node is primary)
        Downstream servers: OK (this node has no downstream nodes)
        Replication slots: OK (node has no physical replication slots)
        Missing physical replication slots: OK (node has no missing physical replication slots)
        Configured data directory: OK (configured "data_directory" is "/bitnami/postgresql/data")

POD-1

WARNING: node "xxxx-postgres-postgresql-ha-postgresql-0" not found in "pg_stat_replication"
Node "ems5g-postgres-postgresql-ha-postgresql-1":
        Server role: OK (node is primary)
        Replication lag: OK (N/A - node is primary)
        WAL archiving: OK (0 pending archive ready files)
        Upstream connection: OK (N/A - node is primary)
        Downstream servers: CRITICAL (1 of 1 downstream nodes not attached; missing: xxxx-postgres-postgresql-ha-postgresql-0 (ID: 1000))
        Replication slots: CRITICAL (1 of 1 physical replication slots are inactive)
        Missing physical replication slots: OK (node has no missing physical replication slots)
        Configured data directory: OK (configured "data_directory" is "/bitnami/postgresql/data")

**To Reproduce**

1. Install the chart in a kubernetes cluster. [Vmware Tanzu Kubernetes Grid is used here]
2. Push data at regular interval. Timeframe for exact trigger is not clear but usually after 7 days.
3. Fetch data from an application on regular basis. If this issue occurs, old or inconsistent data will be shown.

**Expected behavior**
There should be no inconsistency in data even after switchover and there should be always one primary node at any instance.

**Version of Helm and Kubernetes**:

[...]

Issue #6555, opened 2021-06-03
[bitnami/postgres-ha] - Having multiple primaries after killing main process on primary pod


giladneiger:
**Which chart**:
bitnami/postgres-ha:7.6.0

**Describe the bug**
After killing (for HA testing purposes) the main process (pid 1) there are 2 primary replicas in the cluster. Moreover, we are getting 2 different results from 2 different replicas.

replica 0:
![image](https://user-images.githubusercontent.com/40278137/120675317-a01b8600-c49d-11eb-9481-8bc4698fad02.png)

replica 1:
![image](https://user-images.githubusercontent.com/40278137/120673443-d9eb8d00-c49b-11eb-833e-ef4ab6d6dfd4.png)

**To Reproduce**

1. We used the following image tags:
        - bitnami/pgpool:4.2.3-debian-10-r2
        - bitnami/postgresql-repmgr:12.6.0-debian-10-r85
2. Install helm chart with postgres replicas count set to 3
3. Run the following command on the primary postgres pod: `kubectl exec -i postgresql-0 -- kill -2 1`
4. `kubectl exec` to a postgres replica and run the following command: `/opt/bitnami/scripts/postgresql-repmgr/entrypoint.sh repmgr -f /opt/bitnami/repmgr/conf/repmgr.conf cluster show`. You will probably see multiple primaries. If not - try running the command again on a different replica (1 or 2). It's inconsistent.
5. This error is inconsistent. Sometimes it happens and sometimes not.

**Expected behavior**
We expected to have 1 primary replica. And also expected to have the same response from the replicas.

**Version of Helm and Kubernetes**:

- Output of `helm version`:

```
version.BuildInfo{Version:"v3.2.1", GitCommit:"fe51cd1e31e6a202cba7dead9552a6d418ded79a", GitTreeState:"clean", GoVersion:"go1.13.10"}
```

- Output of `kubectl version`:

```
Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.0", GitCommit:"9e991415386e4cf155a24b1da15becaa390438d8", GitTreeState:"clean", BuildDate:"2020-03-25T14:58:59Z", GoVersion:"go1.13.8", Compiler:"gc", Platform:"windows/amd64"}
[...]

Issue #5589, opened 2021-02-22
[postgresql-ha]: Default pgpool configuration does not follow primary after failover ?


somejfn:
**Which chart**:
bitnami/postgresql-ha 6.5.1   

Installed on K8s 1.19 with:
```
helm install pgkeycloak bitnami/postgresql-ha \
  --set postgresqlImage.tag=11.10.0-debian-10-r78  \
  --set pgpoolImage.tag=4.2.1-debian-10-r37  \
  --set metricsImage.tag=0.8.0-debian-10-r346 \
  --set metrics.enabled=true \
  --set postgresql.pdb.create=true \
  --set pgpool.replicaCount=2 \
  --set pgpool.customUsersSecret=pgpool-passwd \
  --set pgpool.pdb.create=true
```

A test keycloak instance is the connecting application,  installed with

```
helm install keycloak bitnami/keycloak \
  --set image.tag=12.0.2-debian-10-r6  \
  --set metrics.enabled=true \
  --set postgresql.enabled=false \
  --set externalDatabase.host=pgkeycloak-postgresql-ha-pgpool.default.svc \
  --set replicaCount=2 \
  --set serviceDiscovery.enabled=true \
  --set rbac.create=true \
  --set externalDatabase.user=keycloak \
  --set externalDatabase.password=password \
  --set externalDatabase.database=keycloak \
  --set pdb.create=true
```

See above external DB is set to the pgpool svc

**Describe the bug**

Following a primary PG pod restart (i.e. because of a K8s node is drained), and this is triggering the standby to take the role of primary, then the client application is apparently still getting connected to the now read-only instance and create/update/delete SQL statements are failing.

Looking at the generated /opt/bitnami/pgpool/conf/pgpool.conf ,  I was surprised to see what these parameters were set to:

```
follow_primary_command = ''
failover_command = 'echo ">>> Failover - that will initialize new primary node search!"'
failback_command = ''
```

Im new to pgpool....  but shouldn't these point to a script so that backend connections follow the new primary ?  

**To Reproduce**

Create a 2 node HA cluster as above
Connect a keycloak instance to the pgpool svc
Drain the k8s node  running the primary PG instance
Try an action cause a create statement in keycloak (such as creating an internal user)

**Expected behavior**

Client applications should be shielded from PG instance failover/failback by default configuration.

**Version of Helm and Kubernetes**:

- Output of `helm version`:

```
[...]

PR #9035, opened 2022-02-16
[bitnami/postgresql-ha] Mitigate brain split situations

rafariossaa:
Signed-off-by: Rafael Rios Saavedra <rafael.rios.saavedra@gmail.com>

**Description of the change**

- Change default settings for Reconnect. This will make to fail sooner in the case of primary lost
- Fix the preStop logic. It was not detecting the case of being the primary node hence it was not waiting for a new primary to be promoted before stopping. 

**Benefits**

This should make much harder to get into a brain split situation. 
In any case, the best way to reduce the risk is to have more nodes that can decide on which is the new primary node. Three nodes is the very minimum for a cluster, so if you get the node failing continuously (eg. chaos testing) there are chances that in a moment in time there is not primary to join to.

**Possible drawbacks**

**Applicable issues**

This should help with: 
- #6555
- #5589 
- #7464 
- #7163
- #2610

**Additional information**

**Checklist**

- [-] Variables are documented in the values.yaml and added to the README.md using (readme-generator-for-helm)[https://github.com/bitnami-labs/readme-generator-for-helm]


miguelaeh (commented):
Thank you @rafariossaa !!!

Strictly talking, changing the default parameters would be a major bump of the chart version, since updating could unexpectedly change the default behavior of already installed releases.

Checking the changes, it seems these shouldn't cause any kind of unexpected behavior but just to confirm with you, is there any scenario where these changes could affect someone that deployed the chart with some custom configuration?

vishrantgupta:
Hi @rafariossaa thanks for the PR.

> This should make much harder to get into a brain split situation.

Which scenarios are not being covered by this PR which will still lead to the split-brain issue?

> In any case, the best way to reduce the risk is to have more nodes that can decide on which is the new primary node. > Three nodes is the very minimum for a cluster, so if you get the node failing continuously (eg. chaos testing) there are chances that in a moment in time there is not primary to join to.

Even with the 3 node postgres cluster the split-brain issue is reproducible.

The preStop hook will not be called when the kubernetes node suddenly dies or when the postgres pod is killed and it was leading to split-brain issue.

rafariossaa:
> @rafariossaa your fork is behind bitnami:master and the GH Action is failing
> 
> image

Thanks. 
I have rebased my branch.

rafariossaa:
> Checking the changes, it seems these shouldn't cause any kind of unexpected behavior but just to confirm with you, is there any scenario where these changes could affect someone that deployed the chart with some custom configuration?

There is no one that came to my mind. But, of course, I can not guarantee 100% that for any user with any custom configuration.

vishrantgupta:
> not able to reproduce it in a consistent way. 3 nodes are the base minimum

This might be true that the issue is not reproducible in a consistent way, but it does occur and this adds a possibility of data loss. Have you tried the steps that you performed with the 3 node cluster?

> Sure, for this the decrease in the other setting should help.

Which "other setting" you are referring to?

miguelaeh (commented):
>There is no one that comes to my mind, but I can not guarantee 100% that for any user with any custom configuration. There is a myriad of combinations of parameters and scenarios.

Then, should we consider bumping the major?

rafariossaa:
@vishrantgupta 
> This might be true that the issue is not reproducible in a consistent way, but it does occur and this adds a possibility of data loss. Have you tried the steps that you performed with the 3 node cluster?

I not telling that this not might occur. I am asking for a way to reproduce the issue in my side so I can try to look for a better solution. Also, any proposals are welcome.
The minimum size I tried is the one that the chart deploys by default, 3 nodes in this case. I was not able to reproduce it with this PR, 
You need to check you production scenario and how resilient you need to be to add nodes accordingly.
Have into account that postgresql, repmgr and pgpool are not designed for a k8s scenario, so what we have here is a translation of what is done in a scenario with VMs or baremetal servers. 



> 
> > Sure, for this the decrease in the other setting should help.
> 
> Which "other setting" you are referring to?

javsalgar:
I don't believe that two changes in the values.yaml should be considered a major. I expect that production workloads would not rely on default values but in configured and managed values. If there is no a value structure change I don't believe it deserves a major bump.

miguelaeh (approved):
LGTM!

vishrantgupta:
> Have into account that postgresql, repmgr and pgpool are not designed for a k8s scenario, so what we have here is a translation of what is done in a scenario with VMs or baremetal servers.

Could you please include this in the Postgresql chart readme?

juan131:
Hi @vishrantgupta 

Thank you for your feedback! Please free to contribute with a PR improving the README.md or the [Bitnami documentation](https://docs.bitnami.com/kubernetes/infrastructure/postgresql-ha) which is based on the repository below:

- https://github.com/bitnami/charts-docs/tree/main/charts/postgresql-ha

PR #9072, opened 2022-02-18
[bitnami/kafka] Enable SSL for kafka privisioning job

yohboy:

**Description of the change**
Add new auth.tls parameters to provide a client certificate to the provisioning job. It allows the provisioning job to work with SSL when brokers ask for required SSL authentication from clients. Currently, it is not possible.

**Benefits**
Can use the provisioning job when clientProtocol asks for SSL.

**Possible drawbacks**
I removed a configmap containing a custom `server.properties` mounted in the provisioning job shared with brokers. I'm not sure why we would have to mount server properties in a client job that may cause security concerns? The file seems not used at all. Maybe adding an `existingConfigmap` related to the provisioning job would be more interesting?
Same about broker certificates. All broker certificates (from secrets or autoGenerated) are mounted into the provisioning job. They seem not used, and it looks like they only can cause security concerns?

**Applicable issues**
None.

**Additional information**
I used a configmap to dynamically generate and mount the `client.properties` file. I know you are already generating some files (server.properties, consumer.properties, producer.properties, ...) directly in the container based on environment variables. I don't know if you agree with this way.

**Checklist**


migruiz4 (changes requested):
Hi @yohboy,

Thank you very much for your contribution, your PR looks good to me, but there are a few changes I would like to request regarding the Key/Keystore/Truststore passwords.

1. When using `provisioning.auth.tls.password`, the password is not inside a secret, therefore exposed in the pod manifest.
2. When using JKS certificates, both Keystore and Truststore will have to share the same password.

My suggestion would be to define a secret with the keys `key-password`, `keystore-password` and `truststore-password`, a helper that decided which secretName use (either the `existingSecret` or the one created with the passwords provided), then define a couple of additional variables `KAFKA_KEY_PASSWORD`, `KAFKA_TRUSTSTORE_PASSWORD` and `KAFKA_KEYSTORE_PASSWORD`.

yohboy:
Hi @migruiz4,

Thank you for your feedback!

1. I agree, using `provisioning.auth.tls.password` exposes the password in the pod manifest. I added this parameter because it was done this way for the Kafka brokers (`auth.tls.password`). I can remove `provisioning.auth.tls.password`, definitively not a good practice.

2. Again, I agree with this point. There is no granularity currently, but I followed the Kafka brokers' way. I can add this granularity for sure, as long as we provide some parameters to configure the keys.

 So my proposal is:
- Remove `provisioning.auth.tls.password` and only keep `provisioning.auth.tls.existingSecret`.
- Add additionnal environment variables as you suggested: `KAFKA_CLIENT_KEY_PASSWORD`, `KAFKA_CLIENT_KEYSTORE_PASSWORD` and `KAFKA_CLIENT_TRUSTSTORE_PASSWORD`.
- Use the `provisioning.auth.tls.existingSecret` to provide all passwords with 3 different keys defaulted to `key-password`, `keystore-password` and `truststore-password`.
- Add 3 more parameters to customize above keys: `provisioning.auth.tls.keyPassword`, `provisioning.auth.tls.keystorePassword` and `provisioning.auth.tls.truststorePassword`.

What do you think?

migruiz4:
Sorry for the late response @yohboy.

This sounds great to me:
   - Remove provisioning.auth.tls.password and only keep provisioning.auth.tls.existingSecret.
   - Add additionnal environment variables as you suggested: KAFKA_CLIENT_KEY_PASSWORD, KAFKA_CLIENT_KEYSTORE_PASSWORD and KAFKA_CLIENT_TRUSTSTORE_PASSWORD.
   - Use the provisioning.auth.tls.existingSecret to provide all passwords with 3 different keys defaulted to key-password, keystore-password and truststore-password.

On the other side, this would require more efforts to be done properly:
   - Add 3 more parameters to customize above keys: provisioning.auth.tls.keyPassword, provisioning.auth.tls.keystorePassword and provisioning.auth.tls.truststorePassword.
   - Create a secret containing these keys when `existingSecret` is not provided
   - Create a helper that decided whether to use the secretName `existingSecret` or the name of the secret generated by the chart.

If these 3 points sound too complicated, I can help you implement the changes once the first three points are done.

yohboy:
Hi @migruiz4,

I implemented the 4 first points (not yet pushed).  
However, I have some concerns with the 2 lasts. I do not see the point of creating a secret containing `provisioning.auth.tls.keyPassword`, `provisioning.auth.tls.keystorePassword` and `provisioning.auth.tls.truststorePassword` keys when `existingSecret` is not provided. What is the purpose? These variables only specify the key of a secret, not the value. Then what value needs to be added to the secret? Random ones? It will never match the password from the certificates. I do not understand the purpose, could you develop, please?

migruiz4:
Sorry for the misunderstanding. 

The variables `provisioning.auth.tls.keyPassword`, `provisioning.auth.tls.keystorePassword` and `provisioning.auth.tls.truststorePassword` are not a reference to the secret's key, but a variable that users could use to provide the password without it being exposed plain-text in the manifest.

So, the idea is: 'User provides a values.yam containing `keystorePassword` instead of existingSecret' -> 'The chart creates a secret containing that value' -> 'The generated secret is referenced in the deployment'.

But, since I see that it is not how the `auth.tls.password` is handled, we can omit them at the moment and force the usage of `existingSecret`. In a future iteration, we will add the tls secret for the provisioning job and Kafka.

yohboy:
Oh ok, actually, `provisioning.auth.tls.keyPassword`, `provisioning.auth.tls.keystorePassword` and `provisioning.auth.tls.truststorePassword` ARE references to the secret's keys. That explains the misunderstanding! I need a way to configure these keys. Mainly because cert-manager asks for the same passwords (for keystore and truststore generation) without any option to customize the secret's key (and I am sharing a single secret). So my idea was to provide all of these keys AND `existingSecret`.   
However, I understand your concern about creating a secret based on passwords given in plaintext in the chart. It may be interesting to provide this feature also for Kafka brokers and deprecate `auth.tls.password`, so maybe in a future PR as you suggested?

bitnami/kafka/templates/_helpers.tpl
```diff
@@ -252,6 +252,13 @@ Return the Kafka configuration configmap
 {{- end -}}
 {{- end -}}
 
+{{/*
+Return the Kafka client configuration configmap
+*/}}
+{{- define "kafka.client.configmapName" -}}
+    {{- printf "%s-client-configuration" (include "common.names.fullname" .) -}}
+{{- end -}}
+
```
migruiz4, line 261: Helper used to choose between the passwordsSecretName or the secret generated using user-provided plaintext passwords.
```suggestion

{{/*
Returns the secret name for the Kafka Provisioning client
*/}}
{{- define "kafka.client.passwordsSecretName" -}}
{{- if .Values.provisioning.auth.tls.passwordsSecret -}}
    {{- printf "%s" (tpl .Values.provisioning.auth.tls.passwordsSecret $) -}}
{{- else -}}
    {{- printf "%s-client-secret" (include "common.names.fullname" .) -}}
{{- end -}}
{{- end -}}

```

bitnami/kafka/values.yaml
```diff
@@ -1366,6 +1366,48 @@ provisioning:
...
+      caCert: ca.crt
+      ## @param provisioning.auth.tls.keystore The secret key from the certificatesSecret if 'keystore' key different from the default (keystore.jks)
+      ##
+      keystore: keystore.jks
+      ## @param provisioning.auth.tls.truststore The secret key from the certificatesSecret if 'truststore' key different from the default (truststore.jks)
+      ##
+      truststore: truststore.jks
+      ## @param provisioning.auth.tls.existingSecret Name of the secret containing passwords to access the JKS files or PEM key when they are password-protected.
+      ##
+      existingSecret: ""
```
migruiz4, line 1401: To be aligned with other charts, I would appreciate it if you could rename it as passwordsSecrets
```suggestion
      ## @param provisioning.auth.tls.passwordsSecret Name of the secret containing passwords to access the JKS files or PEM key when they are password-protected.
      ## It should contain two keys called "keystore-password" and "truststore-password", or "key-password" if using a password-protected PEM key.
      ##
      passwordsSecret: ""
```

bitnami/kafka/values.yaml
```diff
@@ -1366,6 +1366,48 @@ provisioning:
...
+      ## @param provisioning.auth.tls.keyPassword The secret key from the existingSecret if 'keyPassword' key different from the default (key-password)
+      ##
+      keyPassword: key-password
+      ## @param provisioning.auth.tls.keystorePassword The secret key from the existingSecret if 'param' key different from the default (keystore-password)
+      ##
+      keystorePassword: keystore-password
+      ## @param provisioning.auth.tls.truststorePassword The secret key from the existingSecret if 'truststorePassword' key different from the default (truststore-password)
+      ##
+      truststorePassword: truststore-password
```
migruiz4, line 1410:
```suggestion
      ## @param provisioning.auth.tls.keyPassword Password to access the password-protected PEM key if necessary. Ignored if 'passwordsSecret' is provided.
      ##
      keyPassword: ""
      ## @param provisioning.auth.tls.keystorePassword Password to access the JKS keystore. Ignored if 'passwordsSecret' is provided.
      ##
      keystorePassword: ""
      ## @param provisioning.auth.tls.truststorePassword Password to access the JKS truststore. Ignored if 'passwordsSecret' is provided.
      ##
      truststorePassword: ""
```

bitnami/kafka/templates/kafka-provisioning.yaml
```diff
@@ -97,41 +101,49 @@ spec:
...
+              valueFrom:
+                secretKeyRef:
+                  name: {{ .Values.provisioning.auth.tls.existingSecret }}
+                  key: {{ .Values.provisioning.auth.tls.keystorePassword }}
+            - name: KAFKA_CLIENT_TRUSTSTORE_PASSWORD
+              valueFrom:
+                secretKeyRef:
+                  name: {{ .Values.provisioning.auth.tls.existingSecret }}
+                  key: {{ .Values.provisioning.auth.tls.truststorePassword }}
+            {{- end }}
```
migruiz4, line 128:
```suggestion
            {{- if (include "kafka.client.tlsEncryption" .) }}
            {{- if (eq .Values.provisioning.auth.tls.type "pem") }}
            - name: KAFKA_CLIENT_KEY_PASSWORD
              valueFrom:
                secretKeyRef:
                  name: {{ template "kafka.client.passwordsSecretName" . }}
                  key: key-password
            {{- else }}
            - name: KAFKA_CLIENT_KEYSTORE_PASSWORD
              valueFrom:
                secretKeyRef:
                  name: {{ template "kafka.client.passwordsSecretName" . }}
                  key: keystore-password
            - name: KAFKA_CLIENT_TRUSTSTORE_PASSWORD
              valueFrom:
                secretKeyRef:
                  name: {{ template "kafka.client.passwordsSecretName" . }}
                  key: truststore-password
            {{- end }}
            {{- end }}
```

migruiz4 (changes requested):
Hi @yohboy,

Please have a look at my suggestions below.

In addition, the following template needs to be added (e.g `kafka-provisioning-secret.yaml`):
```
{{- if and .Values.provisioning.enabled (include "kafka.client.tlsEncryption" .) (not .Values.provisioning.auth.tls.passwordsSecret)  }}
apiVersion: v1
kind: Secret
metadata:
  name: {{ template "kafka.client.passwordsSecretName" . }}
    {{- end }}
  {{- end }}
type: Opaque
data:
  {{- if .Values.provisioning.auth.tls.keystorePassword }}
  truststore-password: {{ default "" .Values.provisioning.auth.tls.keystorePassword | b64enc | quote }}
  {{- end }}
  {{- if .Values.provisioning.auth.tls.truststorePassword }}
  keystore-password: {{ default "" .Values.provisioning.auth.tls.truststorePassword | b64enc | quote }}
  {{- end }}
  {{- if .Values.provisioning.auth.tls.keyPassword }}
  key-password: {{ default "" .Values.provisioning.auth.tls.keyPassword | b64enc | quote }}
  {{- end }}
{{- end }}
```

yohboy:
Hi @migruiz4!

Thank you very much for your feedback and your help regarding this PR.
I checked your suggestions, it seems pretty fine, I just have 2 concerns about some changes.  

The first one is about removing the possibility to customize keys from the `passwordSecrets` when provided. To be honest, I really need this feature, so I will apply your suggestions, but I propose to also add the feature. It means 3 more parameters, let's say `provisioning.auth.tls.keyPasswordSecretKey`, `provisioning.auth.tls.keystorePasswordSecretKey` and `provisioning.auth.tls.truststorePasswordSecretKey`. Let me know if you think of different names? Without this customization feature, it can not be used with `cert-manager` easily.

The second concern is about the env var `KAFKA_CLIENT_KEY_PASSWORD`. You changed the configuration to set it only if `provisioning.auth.tls.type` is equal to `pem`, but a certificate key inside a JKS Keystore can be password protected. So I think this variable can be used with JKS type too. However, I agree it requires some changes to not add it if the certificate key is not password protected when using JKS Keystore (or even `pem` type). It may be difficult to handle though. Let me know what you think, please.

migruiz4:
Hi again @yohboy,

> The first one is about removing the possibility to customize keys from the passwordSecrets when provided. To be honest, I really need this feature, so I will apply your suggestions, but I propose to also add the feature. It means 3 more parameters, let's say provisioning.auth.tls.keyPasswordSecretKey, provisioning.auth.tls.keystorePasswordSecretKey and provisioning.auth.tls.truststorePasswordSecretKey. Let me know if you think of different names? Without this customization feature, it can not be used with cert-manager easily.

Sounds good to me!

> The second concern is about the env var KAFKA_CLIENT_KEY_PASSWORD. You changed the configuration to set it only if provisioning.auth.tls.type is equal to pem, but a certificate key inside a JKS Keystore can be password protected. So I think this variable can be used with JKS type too. However, I agree it requires some changes to not add it if the certificate key is not password protected when using JKS Keystore (or even pem type). It may be difficult to handle though. Let me know what you think, please. 

I forgot about the case of having a password-protected key inside a password-protected keystone. Honestly, it would depend on how Kafka reacts to having this configuration:
```
    ssl.key.password=""
    ssl.keystore.password=""
    ssl.truststore.password=""
```
If it works for non-password protected keys/keystores, then we can go on with the `default ""`, if not, maybe we will need to make some additional changes.

I'm sorry I'm a bit short of time to give this PR the attention I want to. Tomorrow I will test this feature with a couple of PEM/JKS scenarios with and without passwords and get back to you.

yohboy:
Hi @migruiz4,
From my tests, if the pem is not password protected, whatever the value in `ssl.key.password="FOO"`, it will just ignore. I didn't test when both JKS and PEM are password protected though. Let me know if you need anything to help you. I'm waiting for your feedback! Thanks

migruiz4:
Hi @yohboy, and sorry for the late response.

After some testing, I found out that the current approach is not compatible with PEM certificates at all, since     `ssl.keystore.key`, `ssl.keystore.certificate.chain`, and `ssl.truststore.certificates` are not configured with a path, they have to be configured with the **content** of the PEM certificate. 

I managed to make it work, and also fix the issue when no password are set, but it requires some major changes to the current approach.

Please check my suggestions in the following comment.

migruiz4:
Hi @yohboy, and sorry for the late response.

After some testing, I found out that the current approach is not compatible with PEM certificates at all, since     `ssl.keystore.key`, `ssl.keystore.certificate.chain`, and `ssl.truststore.certificates` are not configured with a path, they have to be configured with the **content** of the PEM certificate. 

I managed to make it work, and also fix the issue when no password are set, but it requires some major changes to the current approach.

Please check my suggestions in the following comment.

bitnami/kafka/templates/client-configmap.yaml
```diff
@@ -0,0 +1,33 @@
+{{- if .Values.provisioning.enabled }}
```
migruiz4, line 1: This file has to be removed, since client.properties will now be created during the container initialization.

bitnami/kafka/templates/kafka-provisioning.yaml
```diff
@@ -1,5 +1,6 @@
 {{- if .Values.provisioning.enabled }}
 {{- $replicaCount := int .Values.replicaCount }}
+{{- $clientPropertiesPath := "/opt/bitnami/kafka/config/client.properties" }}
```
migruiz4, line 3:
```suggestion
```

migruiz4, line 3:
```suggestion
```

bitnami/kafka/templates/kafka-provisioning.yaml
```diff
@@ -85,6 +86,9 @@ spec:
           args:
             - -ec
             - |
+              sed -e 's/KAFKA_CLIENT_KEY_PASSWORD/$(KAFKA_CLIENT_KEY_PASSWORD)/g' "{{ .Values.persistence.mountPath }}/config/client.properties" > "{{ $clientPropertiesPath }}"
+              sed -i -e 's/KAFKA_CLIENT_KEYSTORE_PASSWORD/$(KAFKA_CLIENT_KEYSTORE_PASSWORD)/g' "{{ $clientPropertiesPath }}"
+              sed -i -e 's/KAFKA_CLIENT_TRUSTSTORE_PASSWORD/$(KAFKA_CLIENT_TRUSTSTORE_PASSWORD)/g' "{{ $clientPropertiesPath }}"
```
migruiz4, line 91: Create `client.properties` similar to how `server.properties` is configured (https://github.com/bitnami/bitnami-docker-kafka/blob/6156e8acc59fbe7acb581607e417fde448037c90/3/debian-10/rootfs/opt/bitnami/scripts/libkafka.sh#L477-L502)
```suggestion
              . /opt/bitnami/scripts/libkafka.sh
              export CLIENT_CONF="${CLIENT_CONF:-/opt/bitnami/kafka/config/client.properties}"
              if [ ! -f "$CLIENT_CONF" ]; then
                touch $CLIENT_CONF

                kafka_common_conf_set "$CLIENT_CONF" security.protocol {{ include "kafka.listenerType" ( dict "protocol" .Values.auth.clientProtocol ) | quote }}
                kafka_common_conf_set "$CLIENT_CONF" ssl.keystore.type {{ upper .Values.provisioning.auth.tls.type | quote }}
                kafka_common_conf_set "$CLIENT_CONF" ssl.truststore.type {{ upper .Values.provisioning.auth.tls.type | quote }}
                ! is_empty_value "$KAFKA_CLIENT_KEY_PASSWORD" && kafka_common_conf_set "$CLIENT_CONF" ssl.key.password "$KAFKA_CLIENT_KEY_PASSWORD"
                {{- if eq (upper .Values.provisioning.auth.tls.type) "PEM" }}
                file_to_multiline_property() {
                    awk 'NR > 1{print line" \\"}{line=$0;}END{print $0" "}' <"${1:?missing file}"
                }
                kafka_common_conf_set "$CLIENT_CONF" ssl.keystore.key "$(file_to_multiline_property "/certs/{{ .Values.provisioning.auth.tls.key }}")"
                kafka_common_conf_set "$CLIENT_CONF" ssl.keystore.certificate.chain "$(file_to_multiline_property "/certs/{{ .Values.provisioning.auth.tls.caCert }}")"
                kafka_common_conf_set "$CLIENT_CONF" ssl.truststore.certificates "$(file_to_multiline_property "/certs/{{ .Values.provisioning.auth.tls.cert }}")"
                {{- else if eq (upper .Values.provisioning.auth.tls.type) "JKS" }}
                kafka_common_conf_set "$CLIENT_CONF" ssl.keystore.location "/certs/{{ .Values.provisioning.auth.tls.keystore }}"
                kafka_common_conf_set "$CLIENT_CONF" ssl.truststore.location "/certs/{{ .Values.provisioning.auth.tls.truststore }}"
                ! is_empty_value "$KAFKA_CLIENT_KEYSTORE_PASSWORD" && kafka_common_conf_set "$CLIENT_CONF" ssl.keystore.password "$KAFKA_CLIENT_KEYSTORE_PASSWORD"
                ! is_empty_value "$KAFKA_CLIENT_TRUSTSTORE_PASSWORD" && kafka_common_conf_set "$CLIENT_CONF" ssl.truststore.password "$KAFKA_CLIENT_TRUSTSTORE_PASSWORD"
                {{- end }}
              fi
```

bitnami/kafka/templates/kafka-provisioning.yaml
```diff
@@ -97,41 +101,49 @@ spec:
                 {{- range $name, $value := $topic.config }}
                 --config {{ $name }}={{ $value }} \
                 {{- end }}
+                --command-config {{ $clientPropertiesPath }} \
```
migruiz4, line 104:
```suggestion
                --command-config $CLIENT_CONF \
```

bitnami/kafka/templates/kafka-provisioning.yaml
```diff
@@ -97,41 +101,49 @@ spec:
...
           {{- end }}
           volumeMounts:
-            {{- if or .Values.config .Values.existingConfigmap }}
-            - name: kafka-config
-              mountPath: {{ .Values.persistence.mountPath }}/config/server.properties
-              subPath: server.properties
-            {{- end }}
+            - name: kafka-client-config
+              mountPath: {{ .Values.persistence.mountPath }}/config/client.properties
+              subPath: client.properties
```
migruiz4, line 135:
```suggestion
```

bitnami/kafka/templates/kafka-provisioning.yaml
```diff
@@ -140,32 +152,21 @@ spec:
         {{- include "common.tplvalues.render" (dict "value" .Values.provisioning.sidecars "context" $) | nindent 8 }}
         {{- end }}
       volumes:
-        {{- if or .Values.config .Values.existingConfigmap }}
-        - name: kafka-config
+        - name: kafka-client-config
           configMap:
-            name: {{ include "kafka.configmapName" . }}
-        {{- end }}
+            name: {{ include "kafka.client.configmapName" . }}
```
migruiz4, line 157: Delete `kafka-client-config` (Github UI won't let me suggest here for some reason)

bitnami/kafka/templates/_helpers.tpl
```diff
@@ -252,6 +252,24 @@ Return the Kafka configuration configmap
 {{- end -}}
 {{- end -}}
 
+{{/*
+Return the Kafka client configuration configmap
+*/}}
+{{- define "kafka.client.configmapName" -}}
+    {{- printf "%s-client-configuration" (include "common.names.fullname" .) -}}
+{{- end -}}
```
migruiz4, line 260: No longer needed

```suggestion
```

bitnami/kafka/templates/kafka-provisioning-secret.yaml
```diff
@@ -0,0 +1,24 @@
...
+data:
+  {{- if .Values.provisioning.auth.tls.keystorePassword }}
+  truststore-password: {{ default "" .Values.provisioning.auth.tls.keystorePassword | b64enc | quote }}
+  {{- end }}
+  {{- if .Values.provisioning.auth.tls.truststorePassword }}
+  keystore-password: {{ default "" .Values.provisioning.auth.tls.truststorePassword | b64enc | quote }}
+  {{- end }}
+  {{- if .Values.provisioning.auth.tls.keyPassword }}
+  key-password: {{ default "" .Values.provisioning.auth.tls.keyPassword | b64enc | quote }}
+  {{- end }}
```
migruiz4, line 23:
```suggestion
  truststore-password: {{ default "" .Values.provisioning.auth.tls.keystorePassword | b64enc | quote }}
  keystore-password: {{ default "" .Values.provisioning.auth.tls.truststorePassword | b64enc | quote }}
  key-password: {{ default "" .Values.provisioning.auth.tls.keyPassword | b64enc | quote }}
```

Remove conditionals. Otherwise, this section will cause the pod to fail because of "Secret missing key" error.

```

            {{- if (include "kafka.client.tlsEncryption" .) }}
            - name: KAFKA_CLIENT_KEY_PASSWORD
              valueFrom:
                secretKeyRef:
                  name: {{ template "kafka.client.passwordsSecretName" . }}
                  key: {{ .Values.provisioning.auth.tls.keyPasswordSecretKey }}
            - name: KAFKA_CLIENT_KEYSTORE_PASSWORD
              valueFrom:
                secretKeyRef:
                  name: {{ template "kafka.client.passwordsSecretName" . }}
                  key: {{ .Values.provisioning.auth.tls.keystorePasswordSecretKey }}
            - name: KAFKA_CLIENT_TRUSTSTORE_PASSWORD
              valueFrom:
                secretKeyRef:
                  name: {{ template "kafka.client.passwordsSecretName" . }}
                  key: {{ .Values.provisioning.auth.tls.truststorePasswordSecretKey }}
            {{- end }}
```

Since they default to "", the container logic won't take them under consideration if empty.

migruiz4 (changes requested):

yohboy:
Hi  @migruiz4 

Thank you again for your work and feedback!

I applied all of your changes. I tested some basic cases and it works well. I just noticed a lot of WARN logs from provisioning when there is no tlsEncryption enabled (for example: `WARN The configuration 'ssl.keystore.type' was supplied but isn't a known config. (org.apache.kafka.clients.admin.AdminClientConfig)`). I pushed the last commit to not add these options when there is no tlsEncryption enabled.

> After some testing, I found out that the current approach is not compatible with PEM certificates at all, since `ssl.keystore.key`, `ssl.keystore.certificate.chain`, and `ssl.truststore.certificates` are not configured with a path, they have to be configured with the **content** of the PEM certificate.

Good catch, I really didn't notice this.

Thanks a lot.

migruiz4 (approved):
> I pushed the last commit to not add these options when there is no tlsEncryption enabled.

Nice catch! I focused on testing TLS cases and I forgot about testing the TLS disabled scenario.

Thank you very much. I appreciate your contribution, help, and patience.

Looks good to me!

migruiz4:
> I pushed the last commit to not add these options when there is no tlsEncryption enabled.

Nice catch! I focused on testing TLS cases and I forgot about testing the TLS disabled scenario.

Thank you very much. I appreciate your contribution, help, and patience.

Looks good to me!

yohboy:
I'm happy we make it works, thanks for your help!

migruiz4:
Merged! Sorry, I accidentally closed and reopened and had to rerun tests

Issue #9146, opened 2022-02-21
wrong file path in ConfigMap for configure-w3-total-cache.sh


webdz9r:
### Name and Version

Bitnami/wordpress 13.0.9

### What steps will reproduce the bug?

Enable w3 cache

wordpressConfigureCache: true

### Are you using any custom parameters or values?

nope

### What is the expected behavior?

Needs to point to the right path `chmod a-w /bitnami/wordpress/wp-config.php`

 

### What do you see instead?

It's missing the root path `chmod a-w bitnami/wordpress/wp-config.php`

Needs to be `chmod a-w /bitnami/wordpress/wp-config.php`

### Additional information

simple one char fix please

javsalgar:
Hi!

Thank you so much for the PR! The team will take a look :D

javsalgar:
Hi! Did you have time to check the suggestions from the PR?

webdz9r:
I wasn't aware of any suggestions.

PR #9147, opened 2022-02-21
bitnami/wordpress - simple fix w3 cache post init wp-config save path

webdz9r:
**Description of the change**

Fixes the bash script created by the ConfigMap that enables w3 cache. Simple change to point to the right file uses the same path in line 22 

**Benefits**

script actually works and doesn't error out and makes the wp-config.php locked again 

**Possible drawbacks**

NA

**Applicable issues**

fixes issue reported #9146

**Additional information**

Nope pretty straight forward. I was using kustomize to patch this but decided to submit a proper PR and log the issue 

First PR so I might be doing it wrong 

**Checklist**

### DIFF
--- a/bitnami/wordpress/templates/postinit-configmap.yaml
+++ b/bitnami/wordpress/templates/postinit-configmap.yaml
@@ -36,7 +36,7 @@ data:
     wp total-cache flush all
 
     # Revoke permissions to edit wp-config.php
-    chmod a-w bitnami/wordpress/wp-config.php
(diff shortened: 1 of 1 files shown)


juan131 (changes requested):
Thanks so much for your PR @webdz9r 

Please note that it's mandatory to bump the chart version in the Chart.yaml and sign your work in order to contribute to the repository. Find more information about this in the link below: 

- https://github.com/bitnami/charts/blob/master/CONTRIBUTING.md

PR #9151, opened 2022-02-22
bitnami/zookeeper Allow users to provide tls.key/tls.crt/ca.crt in a secret

jesusaurus:
**Description of the change**

Alter template logic so that the resulting init-certs.sh script accepts `tls.key`, `tls.crt`, and `ca.crt` from user-provider secret.

**Benefits**

Enables downstream users to generate certificates similar to how they are generated when `autoGenerated` in enabled, and in the same format used by some zookeeper clients.

**Possible drawbacks**

Changes the behavior of the chart if a user provides a secret with both jks files and tls key/crt files.

**Checklist**


marcosbc (changes requested):
Thanks for the contribution! Since this seems to be adding a new feature, we should be increasing the minor (i.e. to 8.1.0). Could you change that?

jesusaurus:
Done. And thanks for the quick review.

marcosbc (approved):
LGTM!

PR #9169, opened 2022-02-22
[bitnami/redis] Removing timeout on Sentinel command

### LINKED ISSUE #8851: [bitnami/redis] Sometimes All Three Pods are Marked as Master After Startup
**Which chart**:

bitnami/redis version: 16.2.1

**Describe the bug**

When the call `get-master-addr-by-name mymaster` to retrieve the master takes more than the timeout of 5 seconds, all nodes decide to become "master"
https://github.com/bitnami/charts/blob/master/bitnami/redis/templates/scripts-configmap.yaml#L101-L107

**To Reproduce**

1. Create a redis cluster using: ` helm upgrade --install --dependency-update redis -n redis-test --create-namespace -f redis-values.yaml ~/repo/charts/bitnami/redis/ --debug`

This is the contents of `redis-values.yaml`:
```image:
  registry: docker.io
  repository: bitnami/redis
  tag: 6.2.6-debian-10-r103

replica:
  resources:
    limits:
      cpu: 200m
      memory: 100Mi
    requests:
      cpu: 200m
      memory: 100Mi

sentinel:
  enabled: true
  image:
    registry: docker.io
    repository: bitnami/redis-sentinel
    tag: 6.2.6-debian-10-r100
  downAfterMilliseconds: 3000
  failoverTimeout: 3000
  resources:
    limits:
      cpu: 100m
      memory: 50Mi
    requests:
      cpu: 100m
      memory: 50Mi
  extraVolumeMounts:
  - name: tmp
    mountPath: "/tmp"
```

2. Once the redis pods are up, you can see that all 3 are marked as master.

We noticed that in our cluster, the current `timeout 5` was short enough that `get_sentinel_master_info` would always return a non zero exit code for all nodes [here](https://github.com/bitnami/charts/blob/master/bitnami/redis/templates/scripts-configmap.yaml#L101-L102), therefore setting them all up to be "master". We noticed that just increasing that timeout to 10 fixes the problem for us.

**Expected behavior**

Only one master should exist. If the command to [get master info](https://github.com/bitnami/charts/blob/master/bitnami/redis/templates/scripts-configmap.yaml#L81-L90) timed out it doesn't necessarily mean that there is no master so the logic under https://github.com/bitnami/charts/blob/master/bitnami/redis/templates/scripts-configmap.yaml#L101-L102 could be improved so it handles a timeout different than receiving null from the `get-master-addr-by-name` call, or the timeout could be larger

**Additional context**

We have tried to reproduce in a local minikube or kind but since we don't have any other resources than redis we couldn't reproduce the issue. But we were able to see this consistently on our development AWS cluster. We also noticed that this timeout was added under [this pull request](https://github.com/bitnami/charts/pull/8476) and it was noted that this could be a potential issue.

jose-galvez:
**Description of the change**

This change removes a timeout when trying to retrieve the information on who is the Master from Sentinel

**Benefits**

Currently, if the command times out it is been treated as if there was no Master elected. If there's no master elected, the pod calling this command assumes the Master role, which means sometimes multiple pods could call themselves "master" as explained on https://github.com/bitnami/charts/issues/8851

**Possible drawbacks**

If this command takes too long to return the startup could freeze, but this probably indicates a network problem or an issue with Sentinel

**Applicable issues**

  - fixes #8851 

**Additional information**

**Checklist**

### DIFF
--- a/bitnami/redis/Chart.yaml
+++ b/bitnami/redis/Chart.yaml
@@ -24,4 +24,4 @@ maintainers:
 name: redis
 sources:
   - https://github.com/bitnami/bitnami-docker-redis
-version: 16.4.0
+version: 16.4.1
--- a/bitnami/redis/templates/scripts-configmap.yaml
+++ b/bitnami/redis/templates/scripts-configmap.yaml
@@ -86,9 +86,9 @@ data:
 
     get_sentinel_master_info() {
         if is_boolean_yes "$REDIS_SENTINEL_TLS_ENABLED"; then
-            sentinel_info_command="{{- if and .Values.auth.enabled .Values.auth.sentinel }}REDISCLI_AUTH="\$REDIS_PASSWORD" {{ end }}timeout 5 redis-cli -h $REDIS_SERVICE -p $SENTINEL_SERVICE_PORT --tls --cert ${REDIS_SENTINEL_TLS_CERT_FILE} --key ${REDIS_SENTINEL_TLS_KEY_FILE} --cacert ${REDIS_SENTINEL_TLS_CA_FILE} sentinel get-master-addr-by-name {{ .Values.sentinel.masterSet }}"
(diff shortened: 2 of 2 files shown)


alemorcuq (approved):
Thanks!

PR #9176, opened 2022-02-23
[bitnami/thanos] Add a service for each thanos-receive pod

arnecls:
**Description of the change**

This Change adds one Kubernetes service resource per thanos-receive pod, based on the `replicationFactor` setting.
This allows the pods belonging to the stateful set to be referenced via DNS (A record).

**Benefits**

When using thanos-receive in dual mode, it is currently hard/impossible to define a hashring config up-front, as the IPs of the pods are not known at deployment time.
Adding a service for each pod allows to address the pods via their name in the hashring config, allowing a deployment that works without later adjustments to the hashring config. For example:

```yaml
receive:
  replicationFactor: 2
  config:
    - endpoints:
      - "thanos-thanos-receive-0:10901"
      - "thanos-thanos-receive-1:10901"
```

**Possible drawbacks**

The name of the pods is assumed to be `<name-of-the-stateful-set>-<replica>`. It is also assumed, that a label `statefulset.kubernetes.io/pod-name` is set for these pods. This is in line with the [Kubernetes documentation](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#pod-name-label).
If this ever changes, the service resources need to be adjusted.

**Applicable issues**

I could not find any.

**Checklist**


arnecls:
I don't have access to the error logs mentioned by `Early feedback / Package, Lint and Trivy scan`.
Is this a general error pipeline related error, or is it related to this PR?

miguelaeh (changes requested):
Thanks for the PR!

I have some concerns. I don't see the need of creating a service per pod since by default the hostname of each pod is, as you stated `<statefuleset>-<podIndex>`. Adding a service per each pod could have sense if you want to access each pod independently from outside the k8s cluster, in which case the services would need to be LoadBalancer, not ClusterIP.

Could you please describe what are you trying to solve with this please?

arnecls:
> Could you please describe what are you trying to solve with this please?

When I originally set up thanos-receive in dual mode, the thanos-receive-distributor was crashing on startup, using the hashring config I posted above.

> I don't see the need of creating a service per pod since by default the hostname of each pod is, as you stated <statefuleset>-<podIndex>.

This is true, but this name does not resolve inside other pods.
I created a container inside the same namespace and tried to telnet into `thanos-receive-0` (which is the name in my specific case), which gave this error

```
> telnet thanos-receive-0 10901
telnet: could not resolve thanos-receive-0/10901: Name or service not known
```

I.e. the name `thanos-receive-0` does not resolve. I could also verify this using `nslookup`.
Hence the need for the two services, which create a proper A record.

miguelaeh:
It is possible to access those hostnames from inside the pods if you use the FQDN of the pods, which is created by adding the service and namespace. You can check it [here](https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#a-aaaa-records-1).

arnecls:
@miguelaeh based on your feedback I set

```
receive:
  service:
    additionalHeadless: true
```

This makes it possible to access the pods via - in my case - e.g. `thanos-receive-0.thanos-receive-headless`.
This should IMO be added to the documentation, because it is very easy to miss.
Without the additional flag, the headless service will not be created, and without looking up the schema in the documentation you just posted, I would have no clue on how to do this.

Thank you.
Closing this PR, as there's better way's to achieve what I want.

PR #9180, opened 2022-02-23
[bitnami/elasticsearch] Fix named port in ingress

### LINKED ISSUE #9159: [bitnami/elasticsearch] Ingress is incorrectly rendered when Elasticsearch security (X-Pack) is enabled
### Name and Version

bitnami/elasticsearch 7.17.0

When deploying the chart with the below values.yaml file, ingress doesn't work because the ingress resource is rendered incorrectly and the resulting ingress spec points to an incorrect port.

Values.yaml:

```yaml
master:
  ingress:
    enabled: true
    tls: true
    hostname: elasticsearch.example.domain
    extraTls:
      - hosts:
          - elasticsearch.example.domain
        secretName: my-elasticsearch-tls-cert

# X-Pack security parameters
security:
  enabled: true
  # Password for the 'elastic' user
  elasticPassword: "VMware1!"
  tls:
    restEncryption: false
    autoGenerated: true
```

When security.enabled=true, ingress should use port name `https` rather than `http`. 

The resulting ingress spec should be:

```yaml
...
spec:
  rules:
  - host: elasticsearch.example.domain
    http:
      paths:
      - backend:
          service:
            name: elasticsearch-master
            port:
              name: https
        path: /
...
```

### What do you see instead?

Currently, the resulting ingress spec is:

```yaml
...
spec:
  rules:
  - host: elasticsearch.example.domain
    http:
      paths:
      - backend:
          service:
            name: elasticsearch-master
            port:
              name: http
        path: /
...
```

### Additional information

The above occurs because when security.enabled=true, the service port name is `https`.

```yaml
apiVersion: v1
kind: Service
...
spec:
...
  ports:
  - name: https
    port: 9200
    protocol: TCP
    targetPort: https
  - name: tcp-transport
    port: 9300
    protocol: TCP
    targetPort: transport
...
```

Your service helm template uses the `elasticsearch.httpPortName` to set http or https for the Kubernetes service:
https://github.com/bitnami/charts/blob/master/bitnami/elasticsearch/templates/_helpers.tpl#L356#L362

However, for ingress, it is hard-coded:
https://github.com/bitnami/charts/blob/master/bitnami/elasticsearch/templates/ingress.yaml#L29

The same  `elasticsearch.httpPortName` function should be used by the ingress template as well, to prevent such inconsistencies.

As a workaround, I tried to use the `master.ingress.extraHosts` parameter, to point the ingress to the correct port, but apparently this parameter is not used anywhere in your helm templates... Which is a different/additional issue.

I ended up having to create the ingress resource separately for now.

fmulero:
Signed-off-by: Fran Mulero <fmulero@vmware.com>

**Description of the change**
Fix incorrect port in ingress definition. If we enabled security, `https` is set as name for most of the exposed ports but ingress is using always http. The porpouse of the change is use the same logic in the ingress definition.

**Possible drawbacks**

Not identified

**Applicable issues**

  - fixes #9159

**Checklist**

### DIFF
--- a/bitnami/elasticsearch/Chart.yaml
+++ b/bitnami/elasticsearch/Chart.yaml
@@ -25,4 +25,4 @@ name: elasticsearch
 sources:
   - https://github.com/bitnami/bitnami-docker-elasticsearch
   - https://www.elastic.co/products/elasticsearch
-version: 17.9.3
+version: 17.9.4
--- a/bitnami/elasticsearch/templates/ingress.yaml
+++ b/bitnami/elasticsearch/templates/ingress.yaml
@@ -26,7 +26,7 @@ spec:
             {{- if eq "true" (include "common.ingress.supportsPathType" .) }}
             pathType: {{ .Values.master.ingress.pathType }}
             {{- end }}
-            backend: {{- include "common.ingress.backend" (dict "serviceName" (include "elasticsearch.master.fullname" .) "servicePort" "http" "context" $)  | nindent 14 }}
+            backend: {{- include "common.ingress.backend" (dict "serviceName" (include "elasticsearch.master.fullname" .) "servicePort" (include "elasticsearch.httpPortName" .) "context" $)  | nindent 14 }}
       {{- if ne .Values.master.ingress.hostname "*" }}
       host: {{ .Values.master.ingress.hostname }}
       {{- end }}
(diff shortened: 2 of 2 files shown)


carrodher (approved):
LGTM

PR #9202, opened 2022-02-24
[bitnami/influxdb] influx CLI command issues

jcalvarado1965:
Signed-off-by: Juan Alvarado <jcalvarado1965@gmail.com>

**Description of the change**

The influx CLI commands used to generate the backup files were broken (the --host parameter appeared before the actual commands which caused the affected commands to error out). Maybe this used to work in an earlier version of the CLI but it is definitely broken now. Now the commands have the right form.

**Benefits**

The format of the commands is correct.

**Possible drawbacks**

The backup procedure is still somewhat broken for azure, because the azure CLI tries to install azcopy when the az storage blob sync command is issued, and this fails for non-root users. This can be addressed in a future change.

**Applicable issues**

I did not enter an issue for this. Not sure if there is one.

**Additional information**

**Checklist**


marcosbc:
Hi @jcalvarado1965, for legal reasons, I'm afraid we will be unable to merge this request for now. If we need to use the `azcopy` command it should be packaged in the `azure-cli` container image. Therefore, we'd encourage you to open a new issue in the respective `bitnami-docker-azure-cli` repository to add the package.

(Something else that could be tried in the `azure-cli` container image could be to grant write privileges for non-root users (`g+rwX /bin`) so that the `az storage` command does not fail.)

jcalvarado1965:
Thanks for the feedback. I can take out the azcopy changes and open the the issue on azure cli, but can I leave the fixes for the influx CLI commands? As it stands they are broken.

marcosbc:
Sure! That would be great

jcalvarado1965:
Ok I've pruned back the release so only the influx commands are fixed. Let me know if there is anything else. Thanks!

marcosbc (approved):
LGTM! Thanks!

PR #8827, opened 2022-01-28
[bitnami/postgresql] Chart standardization

juan131:
Signed-off-by: juan131 <juanariza@vmware.com>

**Description of the change**

This PR standardizes the `bitnami/postgresql` chart to be in line with the rest of the catalog.

**Benefits**

- Standardization.
- Removing technical debt.

**Possible drawbacks**

This PR **dramatically** changes the values structure including several backwards incompatibilities and forcing a major bump , check the Upgrading Notes for more information.

**Applicable issues**

N/A

**Additional information**

N/A

**Checklist** 

- [X] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [X] Variables are documented in the README.md
- [X] Title of the PR starts with chart name (e.g. [bitnami/<name_of_the_chart>])
- [x] All commits signed off and in agreement of [Developer Certificate of Origin (DCO)](https://github.com/bitnami/charts/blob/master/CONTRIBUTING.md#sign-your-work)

PR #9204, opened 2022-02-25
[bitnami/kubeapps] No longer deploy helm packaging dependencies when helm not enabled.

absoludity:
**Description of the change**

Updates so that if helm is not enabled as a packaging system, then the dependencies of helm are also not enabled - including postgresql, the app-repository controller and associated resources.

I'd initially thought I'd only be able to do this in our development chart as I thought it'd require changes to the dashboard to avoid errors when trying to view app repositories, but it just displays empty data when checking AppRepositories:

app-repos-helm

We've got a separate issue to update the dashboard in https://github.com/kubeapps/kubeapps/issues/4340 to provide something more useful (such as docs for helm/flux).

As a result, when deployed (in dev), the pods are as follows (when flux selected):

```bash
k -n kubeapps get po
NAME                                              READY   STATUS    RESTARTS   AGE
kubeapps-644785b4d-pffjx                          1/1     Running   0          2m14s
kubeapps-internal-dashboard-69db97db89-752l9      1/1     Running   0          2m14s
kubeapps-internal-kubeappsapis-65d7cfbb5c-l4j57   1/1     Running   0          2m14s
kubeapps-internal-kubeops-565b4755c6-krhh7        1/1     Running   0          2m14s
kubeapps-redis-master-0                           1/1     Running   2          2d2h
kubeapps-redis-replicas-0                         1/1     Running   3          2d2h
```

Really looking forward to removing kubeops too. With just carvel, there's no redis currently either.

**Benefits**

Unnecessary resources not used.

**Possible drawbacks**

**Applicable issues**

  - fixes #

**Additional information**

**Checklist**


absoludity:
Actually, I realised afterward that the reason I didn't see any errors was because the AppRepository CRD isn't deleted... if you delete it you do get an error when browsing app repositories, as expected:

apprepo-no-crd

But the reality of Helm charts means that CRDs are always installed (no templating / conditions there). So the situation only arises if you manually delete the CRD (and even then, isn't a bad outcome for now). Nonetheless, I'll be adding a change in our repo to the UI so that we don't try to display AppRepositories if Helm is not enabled.

antgamdia (approved):
Great to see we're deploying Kubeapps' resources in accordance with the actual user's needs.

I haven't personally tried, but what would happen if a user just enables `carvel` and, later on, upgrades Kubeapps adding `helm`?
I mean, I don't know if the helpers (for checking the Postgres' password) we have when upgrading are gonna fail.

Also, I've sent https://github.com/bitnami/charts/pull/9209, aimed at unblocking the release (hopefully). Depending on which of these two gets merged before, we'll have to change the chart version.
(that's why the linters are failing in this PR)

absoludity:
> Great to see we're deploying Kubeapps' resources in accordance with the actual user's needs.
> 
> I haven't personally tried, but what would happen if a user just enables `carvel` and, later on, upgrades Kubeapps adding `helm`? I mean, I don't know if the helpers (for checking the Postgres' password) we have when upgrading are gonna fail.

Worked fine in my dev environment, but there I'm explicitly setting a secret and passing the secret name. I rechecked in my dev environment without the secret (ie. with the following diff:
```diff
--- a/docs/user/manifests/kubeapps-local-dev-values.yaml
+++ b/docs/user/manifests/kubeapps-local-dev-values.yaml
@@ -15,7 +15,7 @@ kubeops:
 postgresql:
   replication:
     enabled: false
-  existingSecret: postgresql-db
+  # existingSecret: postgresql-db
 kubeappsapis:
   replicaCount: 1
 ingress:
@@ -30,3 +30,8 @@ ingress:
     nginx.ingress.kubernetes.io/proxy-read-timeout: "600"
     # Required for ingress-nginx 1.0.0 for a valid ingress.
     kubernetes.io/ingress.class: nginx
+packaging:
+  helm:
+    enabled: true
+  carvel:
+    enabled: false
```
) and you're spot on Antonio:

```
Error: UPGRADE FAILED: execution error at (kubeapps/templates/NOTES.txt:88:4): 

    'postgresql.postgresqlPassword' must not be empty, please add '--set postgresql.postgresqlPassword=$POSTGRESQL_PASSWORD' to the command. To get the current value:

        export POSTGRESQL_PASSWORD=$(kubectl get secret --namespace "kubeapps" kubeapps-postgresql -o jsonpath="{.data.postgresql-password}" | base64 --decode)

```

Nor do I see a way to avoid it - it appears to be an assumption of the [bitnami common library](https://github.com/bitnami/charts/blob/master/bitnami/common/templates/_errors.tpl) that if you are upgrading the main chart, then you are only ever upgrading the subcharts. I've mentioned in the past, my preference would be that Kubeapps doesn't ever generate random passwords for users on installation, but rather provides instructions for users to create a password in a secret (but that would make it different to the rest of the catalog, so perhaps not an option).

So the current state of this change is that it works fine if you use a secret resource, but fails if you rely on the functionality provided by the bitnami common. Not sure if it's worth landing despite this (it happens in a specific circumstance that we can perhaps document), or we separately update the kubeapps chart to not use random passwords, or I create an issue against bitnami charts for conditional dependencies with password requirements during upgrades.

> 
> Also, I've sent #9209, aimed at unblocking the release (hopefully). Depending on which of these two gets merged before, we'll have to change the chart version. (that's why the linters are failing in this PR)

and there was another auto-chart update (nice), so bumped the patch version again.

bitnami/kubeapps/Chart.yaml
```diff
@@ -11,6 +11,7 @@ dependencies:
   - name: postgresql
     repository: https://charts.bitnami.com/bitnami
     version: 10.x.x
```
carrodher, line 13: Please note the bitnami/postgresql version was updated in a major at https://github.com/bitnami/charts/pull/8827. Since this PR is also related to PostgreSQL, can you adapt this PR in order to use the latest version? 
```suggestion
    version: 11.x.x
```

absoludity:
I'm planning to circle back to this tomorrow (Friday) and have one last check if we can avoid the validation error when enabling postgres during an upgrade, but either way, I think it should be good to land (once I fix the DCO and conflicts etc.). Thanks for your patience!

bitnami/kubeapps/Chart.yaml
absoludity: Hmm, I'd be keen to avoid a major update of a dependency if it also means updating our own chart's major version as part of this change? Do you mind if we do that with our next release? I can update it in our dev chart already so it'll get synced when we release.

absoludity:
> Nor do I see a way to avoid it - it appears to be an assumption of the [bitnami common library](https://github.com/bitnami/charts/blob/master/bitnami/common/templates/_errors.tpl) that if you are upgrading the main chart, then you are only ever upgrading the subcharts.

It turns out that the postgres (and redis) password validation checks that we have in the Kubeapps chart are  - I think (?) - now unnecessary? I've removed them in the latest commit here, and tested the chart, but would be keen for @carrodher to double-check my reasoning for me :)

So the postgresql chart [always includes the secret yaml as part of its manifest](https://github.com/bitnami/charts/blob/master/bitnami/postgresql/templates/secrets.yaml) (as long as you have not specified an existing secret name etc.), using the bitnami [common.secrets.passwords.manage](https://github.com/bitnami/charts/blob/master/bitnami/common/templates/_secrets.tpl#L61-L123) which uses the Helm `lookup` feature to *check* if a secret already exists (in the cluster) with the expected name and uses that value.

So if a user upgrades kubeapps (with helm/postgresql enabled), there is no issue - the password no longer needs to be fetched and set explicitly on the command-line as it used to.

The redis chart is similar, although it doesn't use the common.secrets.passwords.manage helper, but does [effectively the same thing with its helper to get the redis password](https://github.com/bitnami/charts/blob/master/bitnami/redis/templates/_helpers.tpl#L184-L209), and similarly [always includes the secret yaml as part of its manifest](https://github.com/bitnami/charts/blob/master/bitnami/redis/templates/secret.yaml#L16) (as long as you have not specified an existing secret name etc.

So as far as I can tell, we only had these validation checks in the NOTES.txt for historical reasons, when the postgres (and redis?) chart did not use the `lookup` function and so relied on the user including the password on the CLI during upgrade?

Does that make sense @carrodher , or am I missing something?

If you agree, then this is good to land. Thanks either way!

antgamdia:
Apart from the DCO and latest changes' conflicts, I'd +1 this PR from a Kubeapps' point of view. It's great to avoid wasting user's resources if they are not needed.

Just to double-check the changes on our side, I've sent [this PR ](https://github.com/kubeapps/kubeapps/pull/4397) that would trigger the Kubeapps CI to ensure everything is OK. If so, let's land once @carrodher confirms your `lookup` assumptions.

absoludity:
> Apart from the DCO and latest changes' conflicts,

Both fixed.

> I'd +1 this PR from a Kubeapps' point of view. It's great to avoid wasting user's resources if they are not needed.
> 
> Just to double-check the changes on our side, I've sent [this PR ](https://github.com/kubeapps/kubeapps/pull/4397) that would trigger the Kubeapps CI to ensure everything is OK.

Good idea - that's passed too.

> If so, let's land once @carrodher confirms your `lookup` assumptions.

Yep. The one question left for me is that, if my understanding of the current postgres and redis is correct (with their use of `lookup`), then I'm not sure how we can use those chart dependencies with carvel?

absoludity:
@antgamdia If you're happy with the changes, perhaps leave a +1 so @carrodher has that info (before this gets outdated again :) )

antgamdia (dismissed):

bitnami/kubeapps/Chart.yaml
carrodher: No problem, I already created an internal task to bump PostgreSQL and Redis versions in a major in a separate PR, but unfortunately, it is something that can't wait until the next release of the app itself since our internal test & release pipeline need to use the latest version of the dependencies to be protected against a possible security issue affecting one of these assets

absoludity:
Bumped the chart version again after the latest auto-release.

absoludity:
> Bumped the chart version again after the latest auto-release.

And again :)

PR #9209, opened 2022-02-25
[bitnami/kubeapps] Add nullable annotation in a param

antgamdia:
**Description of the change**

Once https://github.com/bitnami-labs/readme-generator-for-helm/issues/31 was solved, this PR adds this new `nullable` annotation for letting the readme-generator know that `null` is a valid value for a specific param.

**Benefits**

Generating a schema or running the readme-generator tool will work again for this chart.

**Possible drawbacks**

N/A

**Applicable issues**

N/A

**Additional information**

Thanks for the effort and time in solving this edge case, @miguelaeh. Really appreciate it!

**Checklist**


bitnami/kubeapps/values.yaml
```diff
@@ -1656,7 +1656,7 @@ kubeappsapis:
           ## @param kubeappsapis.pluginConfig.kappController.packages.v1alpha1.defaultUpgradePolicy Default upgrade policy generating version constraints
           ## enum: [ "major", "minor", "patch", "none" ]
           defaultUpgradePolicy: none
-          ## @param kubeappsapis.pluginConfig.kappController.packages.v1alpha1.defaultPrereleasesVersionSelection Default policy for allowing prereleases containing one of the identifiers
+          ## @param kubeappsapis.pluginConfig.kappController.packages.v1alpha1.defaultPrereleasesVersionSelection [nullable] Default policy for allowing prereleases containing one of the identifiers
```
miguelaeh, line 1659: If this field should be a string, you would need to add:
```suggestion
          ## @param kubeappsapis.pluginConfig.kappController.packages.v1alpha1.defaultPrereleasesVersionSelection [string,nullable] Default policy for allowing prereleases containing one of the identifiers
```

in order to show type string into the schema

antgamdia: This param can be:

````yaml
defaultPrereleasesVersionSelection: null
defaultPrereleasesVersionSelection: []
defaultPrereleasesVersionSelection: ["foo", "bar"]
````

So, I'm not sure we would want to show `type: string` in the schema, perhaps

```suggestion
          ## @param kubeappsapis.pluginConfig.kappController.packages.v1alpha1.defaultPrereleasesVersionSelection [array, nullable] Default policy for allowing prereleases containing one of the identifiers
```

miguelaeh: sure, then you should add `array` where I put `string`. You assumption is correct, just do not add spaces (I am not sure right now if there is a trim into the code for that

antgamdia: I've added it, but it results in an invalid openapi schema: the required `items` property for arrays is not being generated.

#### With `@param .... [nullable]`, valid schema, but not representing the reality

````jsonc
"defaultPrereleasesVersionSelection": {
    "type": "object", 
    "description": "Default policy for allowing prereleases containing one of the identifiers",
    "default": "null",
    "nullable": true
},
````

#### With `@param .... [array, nullable]`, invalid schema

````jsonc
"defaultPrereleasesVersionSelection": {
    "type": "array", 
     // missing 'items' property
    "description": "Default policy for allowing prereleases containing one of the identifiers",
    "default": "null",
    "nullable": true
},
````

#### Expected valid OpenAPI 3.0.X schema:

_Note that we are explicitly not targeting openapi 3.1, otherwise, the `nullable` prop would be removed, and the `type` object would become an array. [Useful SO post clarifying the differences](https://stackoverflow.com/questions/48111459/how-to-define-a-property-that-can-be-string-or-null-in-openapi-swagger)._

````jsonc
"defaultPrereleasesVersionSelection": {
    "type": "string",
    "items": {
      "type": "string" // how can we infer this value? 
    },
    "description": "Default policy for allowing prereleases containing one of the identifiers",
    "default": null,
    "nullable": true
  }
````

Comment (antgamdia) [reply bitnami/kubeapps/values.yaml:1659]: So, rather than generating an invalid schema, I'd go with leaving it as is, until the `items` can be added/inferred in this edge case. WDYT?

Comment (miguelaeh) [reply bitnami/kubeapps/values.yaml:1659]: Looks like another corner case for arrays. Thank you for sharing it.
> >So, rather than generating an invalid schema, I'd go with leaving it as is,

Currently, if you don't specify any modifier for setting the type, it will be object by default. I think if you add just `[nullable]` the `type` will be set to `object` into the schema. Is that what you mean?

Comment (antgamdia) [reply bitnami/kubeapps/values.yaml:1659]: Yes, as we are not using the schema currently, semantic correctness is not a priority, so having `type: object` is something bearable until this edge case gets sorted out.
I mean, sacrificing semantic over syntactic correctness is preferred in this case.

Do you want me to file an issue in the readme's repo? Can we merge this PR then?

File: bitnami/kubeapps/README.md [hunk 2]
```diff
@@ -28,7 +28,7 @@ With Kubeapps you can:
 
 - Customize deployments through an intuitive, form-based user interface
 - Inspect, upgrade and delete applications installed in the cluster
-- Browse and deploy [Helm](https://github.com/helm/helm) charts from public or private chart repositories (including [VMware Marketplace���������](https://marketplace.cloud.vmware.com) and [Bitnami Application Catalog](https://bitnami.com/application-catalog))
+- Browse and deploy [Helm](https://github.com/helm/helm) charts from public or private chart repositories (including [VMware Marketplace™](https://marketplace.cloud.vmware.com) and [Bitnami Application Catalog](https://bitnami.com/application-catalog))
```
carrodher, line 31:
```suggestion
- Browse and deploy [Helm](https://github.com/helm/helm) charts from public or private chart repositories (including [VMware Marketplaces&trade;](https://marketplace.cloud.vmware.com) and [Bitnami Application Catalog](https://bitnami.com/application-catalog))
```

bitnami/kubeapps/values.yaml
antgamdia: As discussed offline, this corner case is going to be resolved soon in the readme-generator tool, therefore, let me update the PR to also include the [array] modifier.
Thanks!

miguelaeh: I agree, let's proceed with this PR, I will try to fix t

miguelaeh (dismissed):
LGTM

PR #9210, opened 2022-02-25
[bitnami/wildfly] Chart standardisation

migruiz4:
**Description of the change**

This PR standardizes the bitnami/wildfly chart to be in line with the rest of the catalog.

**Checklist**


bitnami/wildfly/templates/NOTES.txt
```diff
@@ -50,19 +50,19 @@ APP VERSION: {{ .Chart.AppVersion }}
         Watch the status with: 'kubectl get svc --namespace {{ .Release.Namespace }} -w {{ template "common.names.fullname" . }}'
 
   export SERVICE_IP=$(kubectl get svc --namespace {{ .Release.Namespace }} {{ template "common.names.fullname" . }} --template "{{ "{{ range (index .status.loadBalancer.ingress 0) }}{{ . }}{{ end }}" }}")
-  {{- $port:=.Values.service.port | toString }}
-  echo "WildFly URL:            http://$SERVICE_IP{{- if ne $port "80" }}:{{ .Values.service.port }}{{ end }}/
+  {{- $port:=(coalesce .Values.service.ports.http .Values.service.port) | toString }}
+  echo "WildFly URL:            http://$SERVICE_IP{{- if ne $port "80" }}:{{ coalesce .Values.service.ports.http .Values.service.port }}{{ end }}/
```
juan131, line 54: You're repeating this "coalesce" a couple of times. Why don't you create a variable at the top of the file and use that instead?

juan131 (dismissed):
LGTM! Just a minor comment

bitnami/wildfly/templates/NOTES.txt
migruiz4: I'm using coalesce because I didn't want to force a major just because of `.Values.service.port` being deprecated, but this should be removed in the next major.

Instead of making the change more elegant, I think it would be better to add a reminder at `values.yaml`, would you agree?

juan131: I think we should do both:
- Using a variable is just for avoiding duplicated code. I don't see any reason why we shouldn't do that.
- Adding a "TODO" in the values.yaml in the "to-be-deprecated" parameter description would help too.

migruiz4: Suggestions applied 👍

juan131 (dismissed):

PR #9242, opened 2022-03-01
[bitnami/kafka] Improve SASL handling and support for service mesh

### LINKED ISSUE #9026: Kafka add provisioning.extraCommands value
### Name and Version

bitnami/kafka 15.1.0

Customizing the provisioning script is currently only possible via provisioning.command and provisioning.args, which results in overwriting the built-in provisioning script for topics. Thus even a minor addition to the script will require re-implementing it.

In my case the addition I wish to make is writing an ACL initialization immediately after topic creation, no changes to the topic creation needed.

Add provisioning.extraCommands: [] to the values and add each command in it immediately after the rest of the provisioning script. This would remove the need for re-writing and lower the bar for testing additions to the provisioning script.

_No response_

sidju:
**Description of the change**
Add system user as separate entry in jaas-secret.
Make metrics exporter not print SASL_USER_PASSWORD to log on startup.
Move provisioning wait-for-available-kafka into main container,
since init containers aren't wrapped by service mesh and break with such.
Add utility to provisioning container. A ready SASL properties file and
$COMMAND_CONFIG which can be inlined with the kafka tools to conf SASL
if needed and does nothing otherwise.
Add environment values to provisioning job.

**Benefits**
This makes the provisioner work in a strict service mesh, since it otherwise will never be able to reach the kafka service (which would then be inside the mesh).
It also improves the flexibility in the provisioning script, making it easier to extend and adding support for SASL to it.

**Possible drawbacks**
Changing the metrics exporter command and the values.yaml file risk incompatibilities for some users.

**Applicable issues**
  - fixes #9026

**Additional information**

**Checklist**


javsalgar:
Hi!

Could you fix the merge issues?

sidju:
Yes. Just got a bit confused with the 'package, lint and trivy' failure without details.

Fixed the conflict and should be ready for merge now.

bitnami/kafka/templates/kafka-metrics-deployment.yaml
```diff
@@ -77,43 +77,46 @@ spec:
...
           {{- if .Values.diagnosticMode.enabled }}
           command: {{- include "common.tplvalues.render" (dict "value" .Values.diagnosticMode.command "context" $) | nindent 12 }}
-          {{- else if .Values.metrics.command }}
-          command: {{- include "common.tplvalues.render" (dict "value" .Values.metrics.command "context" $) | nindent 12 }}
+          {{- else if .Values.metrics.kafka.command }}
+          command: {{- include "common.tplvalues.render" (dict "value" .Values.metrics.kafka.command "context" $) | nindent 12 }}
           {{- else }}
           command:
-            - kafka_exporter
+            - bash
```
alemorcuq, line 84: Why the change?

bitnami/kafka/templates/kafka-provisioning.yaml
```diff
@@ -38,30 +38,6 @@ spec:
...
-                --state=inuse \
-                --timeout=120 \
-                {{ .Values.service.ports.client | int64 }};
-              echo "Kafka is available";
-          {{- if .Values.provisioning.resources }}
-          resources: {{- toYaml .Values.provisioning.resources | nindent 12 }}
-          {{- end }}
-        {{- if .Values.provisioning.initContainers }}
-        {{- include "common.tplvalues.render" ( dict "value" .Values.provisioning.initContainers "context" $ ) | nindent 8 }}
-        {{- end }}
```
alemorcuq, line 64: I'm not sure about this. Init containers are a native Kubernetes feature and one of their intended uses cases is waiting for some conditions to be met before starting the applications themselves. We do this in many of our charts.

alemorcuq (changes requested):
Thanks for the PR! Could you please check my comments?

bitnami/kafka/templates/kafka-metrics-deployment.yaml
sidju: The original form resulted in printing out potential SASL auth environment variables into the log at startup, which I consider to be a fatal flaw. This change allows bash to interpret the environment variables at runtime, allowing it to work nearly identically without printing out the values of the SASL variables.

bitnami/kafka/templates/kafka-provisioning.yaml
sidju: I realize that this is a bit of a difficult question between init containers and sidecar containers.

The problem, as indicated in the PR, is that an init container that expects fully functioning networking cannot co-exist with a sidecar that is required for functioning networking (envoy service mesh sidecar in my case). By moving the wait for port into the main container the sidecar enables it to work as intended, which I find to be a good solution without serious downsides.

The alternative would be to add an option to remove that init container, since it completely blocks the provisioning pod if used with a service mesh sidecar. But in that case I'd also like to add provisioning.preScript, so that its functionality can be supplanted by a custom script.

sidju: (As long as I can remove the init container in some way I can use the chart, but I'd like to not need to overwrite your topic provisioning script to add a wait in the main container)

alemorcuq: I think that would be a great compromise.

sidju: It has now been applied.

sidju:
A bit of a tricky merge with the SSL addition you just merged. Read through it an extra time, because a lot needed to be changed.

alemorcuq:
I think you mean #9072? Thanks, I'll take a look at everything.

sidju:
Any progress?

alemorcuq:
Sorry, @sidju. We lost track of the ticket. Can you rebase? Recently a new major version of the chart was released upgrading the Zookeeper dependency, but I don't think it affects this PR.

sidju:
There. Put in a slightly better type and some documentation in the values file, to save future users from the string wrangling I just went through.

alemorcuq (approved):
Thanks!

PR #9255, opened 2022-03-02
[bitnami/postgresql-ha] Added fencing mechanism for situation where failed primary occurred.  

TK-Khaw:
**Description of the change**

Added parameters in the Helm charts for enabling the fencing mechanism of failed primary as proposed in this [PR](https://github.com/bitnami/bitnami-docker-postgresql-repmgr/pull/112) 

As depicted in the PR above, repmgr provides callback mechanism when monitoring child nodes of a primary nodes. Issues had been raised multiple times whenever multiple primaries occurred and  required manual intervention to address the issue, such as in #6555. This PR, as well as the one send to bitnami-docker-postgresql-repmgr, which is the image this Helm chart is based on, attempts to solve this issue by killing off the failed primary (that may had arose possibly due to network issue, causing a new primary be elected) and thus resolving the split-brain issue. 

**Benefits**

Albeit there can be momentary split-brain, this PR ensures that the false primary will be killed off after specific amount of time and thus enable the cluster to self-correct without manual intervention.

**Additional information**

This PR is closely tied to this [PR](https://github.com/bitnami/bitnami-docker-postgresql-repmgr/pull/112) on [bitnami-docker-postgresql-repmgr](https://github.com/bitnami/bitnami-docker-postgresql-repmgr). The image had been linked to the proposed new version on the bitnami-docker-postgresql-repmgr repository. Should that PR be rejected, this PR will need to be rejected as well.

**Checklist**


javsalgar:
Hi! Thank you so much for the PRs! Please let us know when the PR adding the feature in the container get merged

TK-Khaw:
Sure, will notify you again :)

bitnami/postgresql-ha/templates/postgresql/statefulset.yaml
```diff
@@ -321,6 +321,22 @@ spec:
...
+              value: {{ .Values.postgresql.repmgrChildNodesCheckInterval | quote }}
+            {{- end }}
+            {{- if .Values.postgresql.repmgrChildNodesConnectedMinCount }}
+            - name: REPMGR_CHILD_NODES_CONNECTED_MIN_COUNT
+              value: {{ .Values.postgresql.repmgrChildNodesConnectedMinCount | quote }}
+            {{- end }}
+            {{- if .Values.postgresql.repmgrChildNodesDisconnectTimeout }}
+            - name: REPMGR_CHILD_NODES_DISCONNECT_TIMEOUT
+              value: {{ .Values.postgresql.repmgrChildNodesDisconnectTimeout | quote }}
+            {{- end }}
```
rafariossaa, line 339: I am missing all this new parameters in the `values.yaml`.

rafariossaa (changes requested):

bitnami/postgresql-ha/templates/postgresql/statefulset.yaml
TK-Khaw: hi.. not sure if I follow... these are optional parameters, you can set it to certain value depending on your needs, I have documented their usage in `README.md` for postgresql-ha. 
When you need them however you can set it in the `values.yaml` file. 

Here's a brief description of their usage:

`REPMGR_FENCE_OLD_PRIMARY` : set this to `true` if you want to have the cluster killing the old/false primary upon its return into the network, realizing that its children are no longer following it. Otherwise, you can leave it unset. Essentially a switch for the feature introduced in this PR.

`REPMGR_CHILD_NODES_CHECK_INTERVAL` : set this to certain interval in integer seconds, say 5, and that will be the interval that `repmgr` on the primary node will use to poll for the child nodes currently following the primary node. Default by `repmgr` is 5 seconds.

`REPMGR_CHILD_NODES_CONNECTED_MIN_COUNT` : this is the minimum number of child nodes following the primary node required to consider the said primary node as current primary. Should the number of child nodes found to fall below this value, the primary node will be considered as false primary (since by binary reasoning, a child node not following this primary node must have found a new primary elsewhere, hinting at a new primary elected in the absence of this primary node) . Default is 1, meaning by default, `repmgr` will only consider the node a false primary after all of its child node had been found disconnected (found a new primary).

`REPMGR_CHILD_NODES_DISCONNECT_TIMEOUT` : time to wait in seconds once the number of child nodes connected fell below the minimum count required, before `repmgr` call the command set in `child_nodes_disconnect_command` in `repmgr.conf`, which is to kill the false primary in this PR. Default is 30 seconds.   


A more detail documentation of parameters used can be found [here](https://repmgr.org/docs/current/repmgrd-primary-child-disconnection.html).

Hopefully that helps. :)

javsalgar: What he means is that all values that are used in the chart template should be documented in the values.yaml file. The table in the README.md is auto-generated from the information in the values.yaml file

javsalgar (commented):
Commented

rafariossaa:
Hi @TK-Khaw ,
  Sorry, but it seems you need to rebase you chages.

  I've been doing some testing with this PR and the changes proposed in https://github.com/bitnami/bitnami-docker-postgresql-repmgr/pull/112 , and it worked for me.

TK-Khaw:
Hi, I've merged here as well.

bitnami/postgresql-ha/templates/postgresql/statefulset.yaml
```diff
@@ -327,6 +327,20 @@ spec:
             - name: REPMGR_DATABASE
               value: {{ (include "postgresql-ha.repmgrDatabase" .) | quote }}
             {{- end }}
+            - name: REPMGR_FENCE_OLD_PRIMARY
+              value: {{ ternary "true" "false" .Values.postgresql.repmgrFenceOldPrimary | quote }}
```
rafariossaa, line 331: You would need to change this accordingly with my comments on https://github.com/bitnami/bitnami-docker-postgresql-repmgr/pull/112 .

rafariossaa (changes requested):

bitnami/postgresql-ha/templates/postgresql/statefulset.yaml
TK-Khaw: Hi I have changed the ternary for this accordingly as well

rafariossaa (dismissed):
LGTM

PR #9259, opened 2022-03-02
[bitnami/grafana] Add Grafana tests

superaleks:
**Description of the change**

Added the Grafana tests, following the test strategy.  
The tests added here are functional and integration tests. 

**Test result**
Please find the newest test results [here](https://github.com/superaleks/charts/runs/5450122669?check_suite_focus=true).

- [GKE](https://cp.production.k.bromelia.bitnami.net/v1/execution-graphs/ed1f356c-e90a-42a8-a8b8-5f0b9ecb88d2/report) - PASS
- [AKS](https://cp.production.k.bromelia.bitnami.net/v1/execution-graphs/0a6a8066-5079-44d2-a6ee-becfb878f2a2/report) - PASS
- [TKG](https://cp.production.k.bromelia.bitnami.net/v1/execution-graphs/a8dcdbab-34f9-4b38-b2b5-b624a18569a5/report) - PASS

**Benefits**
Whenever someone makes a change on this chart, these test can be executed. They can verify that the change didn't break the chart. 

**Possible drawbacks**

As always with UI tests, there is a possibility of flakiess. If we notice this, we might add more API tests instead of the ones that are based on UI interactions.

**Checklist**


.vib/grafana/cypress/cypress.json
```diff
@@ -0,0 +1,17 @@
...
+    "env": {
+      "username": "test_admin",
+      "password": "ComplicatedPassword123!4",
+      "smtp_name": "test_smtp",
+      "smtp_host": "smtp",
+      "smtp_address": "test@smtp.com",
+      "smtp_user": "smtp_user",
+      "panel_title": "Test panel",
+      "panel_description": "Test panel description",
+      "dashboard_title": "Test dashboard",
```
joancafom, line 12: I'd like to know your thoughts on this. To me, these are test data (hence, fixtures) instead of environment variables. IMHO, environment variables should be reserved to values that may change due to external factors (e.g. an SMTP account) or whose values extremely make the test output change (e.g. the username we log in to perform the actions).

These data here (ass well as `test_email_address` and `invite_user_name`) do simply look like regular mockup data for tests that would perfectly fit in the concept of fixture.

.vib/grafana/cypress/cypress/fixtures/example.json
```diff
@@ -0,0 +1,5 @@
+{
+  "name": "Using fixtures to represent data",
+  "email": "hello@cypress.io",
+  "body": "Fixtures are a great way to mock data for responses to routes"
+}
```
joancafom, line 5: I think you missed to removed this autogenerated file.

.vib/grafana/cypress/cypress/integration/grafana_test_suite.js
```diff
@@ -0,0 +1,146 @@
...
+  deleteDashboard,
+  deleteAPIKey,
+  clickDeleteButton,
+  verifySuccesOfAction
+} from './utils'
+
+it('allows user to log out', () => {
+  cy.login();
+  cy.visit('/logout');
+  cy.get('.login-content-box').should('exist');
```
joancafom, line 12: I think `cy.get(...)` already checks whether an element in the DOM exists or not. Hence `should('exist')` should be kind of redundant here. 

Instead (if it applies) you can go with `should('be.visible');`

.vib/grafana/cypress/cypress/support/commands.js
```diff
@@ -0,0 +1,25 @@
...
+    username = Cypress.env("username"),
+    password = Cypress.env("password")
+) => {
+    cy.clearCookies();
+    cy.visit('/login')
+    cy.get('[name="user"]').type(username).should('have.value',username);
+    cy.get('input#current-password').type(password).should('have.value',password);
+    cy.get('[aria-label="Login button"]').click();
+    cy.get('div').contains('Invalid username').should('not.exist');
+});
```
joancafom, line 25: I personally would avoid using assertions in custom commands (unless they are meant to actually assert something, like `checkXXXX`). If a given functionality needs to be verified (in this case, the login), I'd better create a specific test case for that.

.vib/grafana/cypress/cypress/integration/grafana_test_suite.js
```diff
@@ -0,0 +1,146 @@
...
+})
+
+it('allows creating a dashboard with a panel', () => {
+  const CLOSE_DIALOGUE_BUTTON = 'button[aria-label="Close dialogue"]';
```
joancafom, line 16: This is nice, but I'm having a hard time trying to tell when to use a constant or directly write the selector in the command. 

I'd say that it is fine to use constants when you will be reusing them through different lines, but this time it seems it is only used once.

.vib/grafana/cypress/cypress/integration/grafana_test_suite.js
```diff
@@ -0,0 +1,146 @@
...
+
+  cy.login();
+  cy.visit('dashboard/new');
+  cy.get('button').contains('Add a new panel')
+    .should('be.visible').click();
+  cy.get('[value="Panel Title"]').clear().type(Cypress.env('panel_title'));
+  cy.get('textarea#description-text-area')
+    .should('exist').type(Cypress.env('panel_description'));
```
joancafom, line 24: Same thoughts wrt to `exist`

.vib/grafana/cypress/cypress/integration/grafana_test_suite.js
```diff
@@ -0,0 +1,146 @@
...
+  const IMPORT_DASHBOARD_BUTTON = '[data-testid="data-testid-import-dashboard-submit"]';
+  
+  cy.login();
+  cy.visit("dashboard/import");
+  cy.get('label').contains('Upload JSON').click();
+  cy.get('input[type=file]').selectFile('cypress/fixtures/test-dashboard.json',
+  {force:true});
+  cy.get(IMPORT_DASHBOARD_BUTTON).click();
+  cy.visit("dashboards");
+  cy.get('div').contains('New test dashboard')
```
joancafom, line 48: As the `test-dashboard.json` is a JSON file, could it be possible to parse it as a fixture and use `fixture.title` instead of hardcoding the test value?

.vib/grafana/cypress/cypress/integration/grafana_test_suite.js
```diff
@@ -0,0 +1,146 @@
...
+    .should('exist').click();
+  deleteDashboard();
+})
+
+it('allows inviting a user', () => {
+  cy.login();
+  cy.visit('/org/users/invite');
+  cy.get('[name="loginOrEmail"]').clear().type(Cypress.env('test_email_address'));
+  cy.get('[name="name"]').clear().type(Cypress.env('invite_user_name'));
```
joancafom, line 57: I usually also check in this kind of command that the field actually has the expected text (`.should('have.value',XXXX);`. Do you think it is worth it?

.vib/grafana/cypress/cypress.json
```diff
@@ -0,0 +1,17 @@
...
+      "rule_name": "Test rule",
```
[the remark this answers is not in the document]
superaleks: Well, it is being used in the test spec.  Calling the `Cypress.env` takes values from this file.

.vib/grafana/cypress/cypress/integration/grafana_test_suite.js
superaleks: I wanted to add this constant for readability purposes. This way, when the time comes to change this locator due to an upgrade, we can easily find it. WDYT?

.vib/grafana/vib-pipeline.json
```diff
@@ -0,0 +1,35 @@
+{
```
beltran-rubo, line 1: this is not needed, the 'early-feedback' action already covers 'package', 'lint' and 'trivy' for all the Helm charts.

.vib/grafana/cypress/cypress.json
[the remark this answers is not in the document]
superaleks: You are right, this specific parameter wasn't being used. I thought you were talking about the file. Anyways, I changed the approach. See the other comment about the purpose of environment files.

.vib/grafana/cypress/cypress.json
superaleks: I see your point. I have changed this file to contain only the necessary environment data like `baseUrl`, `username`, `password`.
I agree that this data doesn't belong here, the `fixtures` folder would be better.

.vib/grafana/cypress/cypress/support/commands.js
superaleks: Good point!
I've moved this assertion to the relevant test case.

.vib/grafana/cypress/cypress/integration/grafana_test_suite.js
superaleks: Before I read your comment I've applied the following solution to this problem:
` cy.get('[data-testid="data-testid-import-dashboard-title"]').clear().type(${DASHBOARD_TITLE}.${random});`
This will replace the hardcoded title with a randomised one.
I've tested this by executing the test suite many times in a row, I confirm this works.
I think both your approach and mine and valid and achieve the same result.

.vib/grafana/vib-pipeline.json
superaleks: Removed.

.vib/grafana/cypress/cypress/integration/grafana_test_suite.js
superaleks: I read the Cypress documentation and I see this makes sense, especially because of [this Cypress feature](git push origin HEAD). I'll do it!

.vib/grafana/cypress/cypress/integration/grafana_test_suite.js
superaleks: You are right, it is better to maintain uniformity for now. I've made the changes in the spec. 
We will be using the constants directly unless there is a need for reuse of the same locator.

.vib/grafana/cypress/cypress/integration/grafana_test_suite.js
superaleks: Now that we are using dynamic data, we cannot assert expecting the exact value, since it will change all the time. This can remain as a good practice whenever feasible.

.vib/grafana/cypress/cypress/integration/grafana_test_suite.js
superaleks: Fixed!

.vib/grafana/cypress/cypress/integration/grafana_test_suite.js
```diff
@@ -0,0 +1,155 @@
...
+
+  cy.login();
+  cy.visit("/datasources/new");
+  cy.get('input[placeholder="Filter by name or type"]')
+    .type(DATASOURCE);
+  cy.get('button[aria-label= "Add data source ' + DATASOURCE + '"]').click();
+  cy.contains('div','Datasource added')
+    .should('be.visible');
+    cy.visit("/datasources");
+    cy.contains('a',DATASOURCE).click({force:true});
```
javsalgar, line 101: What's the reason behind `force: true`?

.vib/grafana/cypress/cypress/integration/grafana_test_suite.js
```diff
@@ -0,0 +1,155 @@
...
+})
+
+it('checks if it is possible to create and delete a data source', () => {
+  const DATASOURCE = 'Prometheus';
+
+  cy.login();
+  cy.visit("/datasources/new");
+  cy.get('input[placeholder="Filter by name or type"]')
+    .type(DATASOURCE);
+  cy.get('button[aria-label= "Add data source ' + DATASOURCE + '"]').click();
```
javsalgar, line 97: Does the space between the `=` and `"Add` affect?

javsalgar, line 97: I see that you are using `=`. Any reason for not using `*=` for example? It would be more resilient to UI changes

Example

```
'button[aria-label*=' + DATASOURCE + '"]'
```

superaleks: I didn't know that! Let me give it a try.

superaleks: Didn't have any effect, but I removed it anyway.

superaleks: I've applied this improvement to the locators. Thanks!

.vib/grafana/cypress/cypress/integration/grafana_test_suite.js
```diff
@@ -0,0 +1,155 @@
...
+    cy.contains('a',DATASOURCE).click({force:true});
+    cy.contains('button','Delete').click();
+    cy.contains('div','Are you sure you want to delete')
+      .should('be.visible');
+    cy.get('button[aria-label*="Confirm Modal Danger Button"]').should('be.visible').click();
+    verifySuccesOfAction();
+})
+
+it('checks if an API key can be added and deleted', () => {
+  const API_KEY_NAME = "test_api_key";
```
javsalgar, line 110: This one is using `"` instead of `'`.

.vib/grafana/cypress/cypress/integration/grafana_test_suite.js
```diff
@@ -0,0 +1,155 @@
...
+  verifySuccesOfAction();
+})
+
+it('checks if smtp is configured', () => {
+  const SMTP_ADDRESS = 'test@smtp.com';
+  const SMTP_NAME = 'test_smtp';
+  const SMTP_USER = 'smtp_user';
+
+  cy.login();
+  cy.visit("admin/settings");
```
javsalgar, line 74: This one is using `"` instead of `'`.

In our repos we `eslint` to detect this kind of issues

.vib/grafana/cypress/cypress/integration/grafana_test_suite.js
```diff
@@ -0,0 +1,155 @@
...
+})
+
+it('checks if it is possible to create and delete a data source', () => {
+  const DATASOURCE = 'Prometheus';
+
+  cy.login();
+  cy.visit("/datasources/new");
+  cy.get('input[placeholder*="Filter by name or type"]')
+    .type(DATASOURCE);
+  cy.get('button[aria-label*="Add data source ' + DATASOURCE + '"]').click();
```
javsalgar, line 97: I see in you are using string concatenation with `+` while in others (line 48 for example) you are using template strings with backticks. 

We normally use the second option (backticks instead of `+`). For consistency, could you change the `+` for backticks?

.vib/grafana/cypress/cypress/integration/grafana_test_suite.js
```diff
@@ -0,0 +1,155 @@
...
+  cy.get('button[aria-label*="Close dialogue"]').click();
+  cy.get('button[aria-label*="Delete API key"]').click();
+  cy.contains('span','Delete').click();
+})
+
+it('checks admin settings endpoint',() => {
+  cy.request({
+    method: 'GET',
+    url: '/api/admin/settings',
+    form: true,
```
javsalgar, line 129: Trailing space

.vib/grafana/cypress/cypress/integration/grafana_test_suite.js
```diff
@@ -0,0 +1,155 @@
...
+    expect(response.status).to.eq(200);
+    expect(response.body).to.include.all.keys('DEFAULT','alerting', 'analytics','security','smtp');
+  })
+})
+
+it('checks the usage preview endpoint',() => {
+    cy.request({
+    method: 'GET',
+    url: 'api/admin/usage-report-preview',
+    form: true,
```
javsalgar, line 145: Trailing space

.vib/grafana/cypress/cypress/integration/grafana_test_suite.js
```diff
@@ -0,0 +1,155 @@
...
+    .should('be.visible');
+  cy.get('[name*="title"]').clear().type(`${DASHBOARD_TITLE}.${random}`);
+  cy.get('[type*="submit"]')
+    .should('be.visible').click();
+  verifySuccesOfAction();
+})
+
+it('checks if it is possible to upload a dashboard as JSON file', () => {
+  const DASHBOARD_TITLE = "New Test dashboard";
+
```
javsalgar, line 42: Trailing space

.vib/grafana/cypress/cypress/integration/grafana_test_suite.js
```diff
@@ -0,0 +1,155 @@
...
+  cy.contains('button','Add a new panel')
+    .should('be.visible').click();
+  cy.get('[value*="Panel Title"]').clear().type(`${PANEL_TITLE}.${random}`);
+  cy.get('textarea#description-text-area')
+    .should('exist').type(`${PANEL_DESCRIPTION}.${random}`);
+
+  cy.contains('button','Apply')
+    .should('be.visible').click();
+
+  cy.contains('.panel-title',PANEL_TITLE).should('be.visible');
```
javsalgar, line 30: normally there is one space between arguments
```
  cy.contains('.panel-title', PANEL_TITLE).should('be.visible');

```

javsalgar (changes requested):
Thank you! Cosmetic changes required

.vib/grafana/cypress/cypress/integration/grafana_test_suite.js
superaleks: Clicking on an element even if it's not actionable. For this case, the link appears to be overlayed with a `<div>` element. More about forcing [in Cypress docs](https://docs.cypress.io/api/commands/click#Assertions).

.vib/grafana/cypress/cypress/integration/grafana_test_suite.js
superaleks: Fixed.

.vib/grafana/cypress/cypress/integration/grafana_test_suite.js
superaleks: OK, fixed!

.vib/grafana/cypress/cypress/integration/grafana_test_suite.js
superaleks: Where exactly? I see nothing in line 42.

javsalgar (dismissed):

javsalgar (approved):

Issue #8878, opened 2022-02-02
Abnormal behavior of 'Redis + Sentinel' after server reboots


ping2kpm:
**Which chart:  https://github.com/bitnami/charts/tree/master/bitnami/redis**:

```
[root@server1 redis]# helm ls -n redis
NAME    NAMESPACE       REVISION        UPDATED                                 STATUS          CHART           APP VERSION
redis   redis           1               2022-02-02 12:20:46.113567561 +0000 UTC deployed        redis-16.1.0    6.2.6
```

**Describe the bug**

**To Reproduce**

I have used above mentioned chart, and configured (Redis + Sentinel) successfully, however whenever I reboot my server it gives different different errors.

POD:
```
[root@server1  redis]# kubectl get pods,svc,ep,sts -n redis
NAME                                      READY   STATUS    RESTARTS      AGE
pod/redis-node-2                          3/3     Running   9 (44m ago)   4h44m
pod/redis-node-0                          3/3     Running   9 (44m ago)   4h45m
pod/redis-node-1                          3/3     Running   9 (44m ago)   4h44m

NAME                      TYPE        CLUSTER-IP      EXTERNAL-IP   PORT(S)                                        AGE
service/redis-headless    ClusterIP   None            <none>        6379/TCP,26379/TCP                             4h45m
service/redis-metrics     ClusterIP   10.43.40.195    <none>        9121/TCP                                       4h45m
service/redis             ClusterIP   10.43.162.52    <none>        6379/TCP,26379/TCP                             4h45m

NAME                        ENDPOINTS                                                        AGE
endpoints/redis-headless    10.42.0.235:6379,10.42.0.236:6379,10.42.0.237:6379 + 3 more...   4h45m
endpoints/redis             10.42.0.235:6379,10.42.0.236:6379,10.42.0.237:6379 + 3 more...   4h45m
endpoints/redis-metrics     10.42.0.235:9121,10.42.0.236:9121,10.42.0.237:9121               4h45m

NAME                          READY   AGE
statefulset.apps/redis-node   3/3     4h45m
[root@server1 redis]#
```

# Scenario 1:
In the below case the salves are not connected after server reboot.
```
[root@server1 redis]# kubectl -n redis exec -it redis-node-0 redis -- bash
Defaulted container "redis" out of: redis, sentinel, metrics
I have no name!@redis-node-0:/$ redis-cli
127.0.0.1:6379> auth passwd
OK
127.0.0.1:6379> info replication
 Replication
[...]

PR #9282, opened 2022-03-03
[bitnami/redis] feat: :sparkles: Add experimental persistence to sentinel configuration

javsalgar:

**Description of the change**

Currenty, the cluster is configured as follows:

- Contact existing Sentinel instances so they point which the current master is
- Case 1: No Sentinel instances available. Assume we are the first node and therefore configure as master
- Case 2: We get a response from the sentinel and therefore we configure ourselves based on the response.

This worked in most of the cases because StatefulSets by default are expected to be launched in an ordered manner. However, the case reported in this issue is when a reboot of the whole cluster is done.  For example:

```
minikube stop
minikube start
```

In this case, Kubernetes will start all the nodes of the StatefulSet at the same time, making the logic above to not work and causing that all nodes believe that are masters.

A way to workaround this is to allow the persistence of the cluster state, this would be done by persisting the {{sentinel.conf}} file, and therefore the logic would look as follows.

- Contact existing sentinel instances so they point which the current master is.
- Case 1 (lowest priority): No sentinel instances are available and no persisted {{sentinel.conf}} is found. Then assume we're master.
- Case 2 (medium priority): No sentinel instances are available but a persisted {{sentinel.conf}} is found. Configure the instance based on the information of that {{sentinel.conf}} file.
- Case 3 (maximum priority): We get a response from the sentinel and therefore we configure ourselves based on the response.

This solution would cover the case of a full cluster restart. This would not cover 100% of the cases but at least we're reducing the number of corner cases.  In order to not disrupt the existing user experince, we created this PR with the feature disabled by default, so the community can provide feedback on it.
**Benefits**

**Possible drawbacks**

**Applicable issues**

  - fixes https://github.com/bitnami/charts/issues/8878

**Additional information**

**Checklist**


bitnami/redis/templates/scripts-configmap.yaml
```diff
@@ -304,6 +333,15 @@ data:
       REDIS_MASTER_PORT_NUMBER="${REDIS_EXTERNAL_MASTER_PORT}"
     fi
 
+    cp /opt/bitnami/redis-sentinel/mounted-etc/sentinel.conf /opt/bitnami/redis-sentinel/etc/sentinel.conf
```
migruiz4, line 336: Both `start-node.sh` (redis container) and `start-sentinel.sh` (sentinel container) are executed in parallel, and `/opt/bitnami/redis-sentinel/etc/sentinel.conf` is mounted in both containers. 

This overwrite could potentially affect `start-node.sh` since we can't ensure lines 112-119 are executed before this, shouldn't we add some way to control that?

migruiz4 (changes requested):
Hi! 
Please check my comments below

bitnami/redis/templates/scripts-configmap.yaml
javsalgar: Added a lock and a loop just in case

migruiz4 (approved):
LGTM!

bitnami/redis/templates/scripts-configmap.yaml
```diff
@@ -104,37 +104,67 @@ data:
...
+            info "Configuring the node as replica"
+            export REDIS_REPLICATION_MODE="slave"
+            REDIS_MASTER_HOST=${master_in_persisted_conf}
+            REDIS_MASTER_PORT_NUMBER=${master_port_in_persisted_conf}
+        fi
+    else
         # Fetches current master's host and port
         REDIS_SENTINEL_INFO=($(get_sentinel_master_info))
-        info "printing REDIS_SENTINEL_INFO=(${REDIS_SENTINEL_INFO[0]},${REDIS_SENTINEL_INFO[1]})"
+        info "REDIS_SENTINEL_INFO=(${REDIS_SENTINEL_INFO[0]},${REDIS_SENTINEL_INFO[1]})"
```
rafariossaa, line 137: I would suggest to add a more descriptive info in the log, maybe something like:
```suggestion
        info "Current master: REDIS_SENTINEL_INFO=(${REDIS_SENTINEL_INFO[0]},${REDIS_SENTINEL_INFO[1]})"
```

rafariossaa (approved):
A small suggestion in case you want to consider it, but it LGTM in anycase.

PR #9301, opened 2022-03-04
[bitnami/minio] Fix minio provisioning

### LINKED ISSUE #8928: minio: provisioning job will fail on rerun when policies are added to users
### Name and Version

bitnami/minio 10.1.6

1. install with user provisioning
2. upgrade chart in any way, so the provisioning pod is rerun

```yaml
provisioning:
  enabled: true
  users:
    - username: testuser
      password: testpass
      disabled: false
      policies:
        - readwrite       # this fails on second run of provisioning container
```

Upgrading the Chart should be possible with provisioning enabled.

### What do you see instead?

Chart upgrade fails because the provisioning Pod does fail with:

```
mc: <ERROR> Unable to update the policy: policy `readwrite` already exists.
```

### Additional information

The policies seem to be rendered out as `mc admin policy update provisioning readwrite user=testuser`. However this command is not idempotent and fails with `mc: <ERROR> Unable to update the policy: policy `readwrite` already exists.` when the policy is already assigned to the user.

Using `mc admin policy set provisioning readwrite,policy2,policy3 user=testuser` should work multiple times, but this will remove all other policies. (in my usecase this would be desirable, but might depend on expectations, though).

### LINKED ISSUE #9191: Unable to update the policy: policy `consoleAdmin` already exists
### Name and Version

bitnamy/minio:10.1.11

1. deploy helm chart on K8s v1.22.5
2. Config:
```
mode: standalone

auth:
  rootUser: "owieru8baic5Ahli"
  rootPassword: "Raer1Oozecael5ee"

provisioning:
  enabled: true
  users:
    - username: erwin567
      password: xxx
      disabled: false
      policies:
        - consoleAdmin
        - diagnostics
  groups:
    - name: erwin
      disabled: false
      members:
        - erwin567
      policies:
        - readwrite
  buckets:
    - name: erwin
      region: eu-europe
      quota:
        type: hard
        size: 10GiB
  config:
    - name: region
      options:
        name: eu-europe
```
3. minio is deployed
4. provision pod is crash:
```
robobee@andrea-node-3:~$ kminio get pods
NAME                          READY   STATUS    RESTARTS     AGE
minio-57956bf5c7-rpdpc        1/1     Running   0            28s
minio-provisioning--1-fftq6   0/1     Error     1 (8s ago)   26s
robobee@andrea-node-3:~$ p=minio-provisioning--1-fftq6
robobee@andrea-node-3:~$ klog kminio $p
Start Minio provisioning
Added `provisioning` successfully.
Added user `erwin567` successfully.
mc: <ERROR> Unable to update the policy: policy `consoleAdmin` already exists.

```

_No response_

_No response_

### What do you see instead?

Minio is deployed and the users and groups are provisioned.

### Additional information

I'm new to Minio, so maybe I do something wrong in the config?

ComputerBaer:
**Description of the change**

- Fixed policy update on existing deployments. The `mc admin policy update` command throws an error if the policy is already set. The provisioner now loads the current policies of a group or user and only updates unset policies.
- Added `setPolicies` option to users and groups. When true, the policies will overwrite/replace all policies of the user, when false (default) the policies will be added to the current policies.
- Disabled versioning in standalone and gateway mode. Versioning is only supported in distributed mode, in standalone or gateway mode an error is thrown. Even if the command to disable versioning is executed.
- Update the config first, so that e.g. the new region can be used on buckets. A restart of MinIO is required after changed it, so the provisioner restarts it.

**Benefits**

MinIO provisioning works in standalone mode and can use updated config values. The provisioning script can be used to manage users, groups and buckets in an active deployment. This makes it easier to keep active deployments in sync with your charts.

**Possible drawbacks**

None

**Applicable issues**

  - fixes #8928
  - fixes #9191

**Additional information**

None

**Checklist**


bitnami/minio/templates/provisioning-job.yaml
```diff
@@ -69,29 +69,57 @@ spec:
...
               {{- range $policy := $group.policies }}
-              mc admin policy update {{ $minioAlias }} {{ $policy }} group={{ $group.name }};
+              addPolicy group {{ $group.name }} {{ $policy }};
+              {{- end }}
               {{- end }}
               {{- $groupStatus := ternary ("disable") ("enable") (and (not (empty $group.disabled)) $group.disabled) }}
               mc admin group {{ $groupStatus }} {{ $minioAlias }} {{ $group.name }};
               {{- end }}
 
+              {{- $isDistributedMode := and (eq .Values.mode "distributed") (not .Values.gateway.enabled) }}
```
alemorcuq, line 122: Can you move this closer to where it's actually used?

alemorcuq (changes requested):
Thanks for the PR! Please, check my comment below.

bitnami/minio/templates/provisioning-job.yaml
ComputerBaer: Only if I add another variable that contains `.Values`, because `.` is overwritten in `range` and `.Values` does not exist there. Or is there a trick I don't know?

alemorcuq: Fair point

alemorcuq (approved):

alemorcuq (changes requested):
Please solve the conflicts

alemorcuq (approved):
Thanks!

PR #9317, opened 2022-03-07
[bitnami/logstash] Adding service account config for workload identity integration

luizferreira094:
…y integration

Signed-off-by: Luiz Ferreira <luiz.ferreira@idwall.co>

**Description of the change**

This config enables the logstash chart to integrate with workload identity of GCP, useful when you need to use a cloud permission on your config, like, buckets, pubsub, etc..

**Benefits**

This change implements a cleaner config of permissions when installing on a GKE cluster, with that you don't need to create a JSON Key for the logstash service acount with the cloud permission and put it inside the pod, you can config that easily using workload identity(https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity)

**Possible drawbacks**

This is a simply configuration, doesn't affect anything related to how things work right now.

**Applicable issues**

  - fixes #

**Additional information**

I created this PR because i was integrating this logstash chart on my cluster and i needed a special permission for using a custom plugin of pubsub(https://www.elastic.co/guide/en/logstash/current/plugins-inputs-google_pubsub.html), so, instead of creating a new service account on GCP project, download de jsonkey, and put it in a volume of pod, i created this config for easily set this permission using workload identity, using that, all i need to do is add an annotation on service account inside the cluster, pointing to the service account created on GCP project. 

**Checklist**

### DIFF
--- a/bitnami/logstash/templates/serviceaccount.yaml
+++ b/bitnami/logstash/templates/serviceaccount.yaml
@@ -1,0 +1,20 @@
+{{- if .Values.serviceAccount.create -}}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ include "logstash.serviceAccountName" . }}
+  labels: {{- include "common.labels.standard" . | nindent 4 }}
+    {{- if .Values.commonLabels }}
+    {{- include "common.tplvalues.render" ( dict "value" .Values.commonLabels "context" $ ) | nindent 4 }}
+    {{- end }}
+  {{- if or .Values.commonAnnotations .Values.serviceAccount.annotations }}
+  annotations:
+    {{- if .Values.serviceAccount.annotations }}
+    {{- include "common.tplvalues.render" ( dict "value" .Values.serviceAccount.annotations "context" $ ) | nindent 4 }}
+    {{- end }}
+    {{- if .Values.commonAnnotations }}
+    {{- include "common.tplvalues.render" ( dict "value" .Values.commonAnnotations "context" $ ) | nindent 4 }}
+    {{- end }}
+  {{- end }}
+automountServiceAccountToken: {{ .Values.serviceAccount.automountServiceAccountToken }}
+{{- end -}}
--- a/bitnami/logstash/templates/serviceaccount.yaml
+++ b/bitnami/logstash/templates/serviceaccount.yaml
@@ -2,7 +2,8 @@
 apiVersion: v1
 kind: ServiceAccount
 metadata:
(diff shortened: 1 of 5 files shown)


bitnami/logstash/templates/serviceaccount.yaml
```diff
@@ -0,0 +1,20 @@
+{{- if .Values.serviceAccount.create -}}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ include "logstash.serviceAccountName" . }}
```
carrodher, line 5:
```suggestion
  name: {{ template "logstash.serviceAccountName" . }}
  namespace: {{ .Release.Namespace | quote }}
```

carrodher (commented):
Thanks for your contribution, I just added some suggestions, can you please take a look?

bitnami/logstash/templates/serviceaccount.yaml
luizferreira094: done

bitnami/logstash/values.yaml
```diff
@@ -162,6 +162,13 @@ extraVolumes: []
 ##     readOnly: true
 ##
 extraVolumeMounts: []
+
+## @param serviceAccount.create If true, create a service account object on Kubernetes.
+# serviceAccount:
+#   create: false
+serviceAccount: 
+  create: false
```
carrodher, line 170: There are some values used in the template but not defined in the values, see
```suggestion
 ## ServiceAccount for Logstash
 ## ref: https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/
 ##
 serviceAccount:
   ## @param serviceAccount.create Enable creation of ServiceAccount for Logstash pods
   ##
   create: true
   ## @param serviceAccount.name The name of the service account to use. If not set and `create` is `true`, a name is generated
   ## If not set and create is true, a name is generated using the logstash.serviceAccountName template
   ##
   name: ""
   ## @param serviceAccount.automountServiceAccountToken Allows automount of ServiceAccountToken on the serviceAccount created
   ## Can be set to false if pods using this serviceAccount do not need to use K8s API
   ##
   automountServiceAccountToken: true
   ## @param serviceAccount.annotations Additional custom annotations for the ServiceAccount
   ##
   annotations: {}
```

luizferreira094: yeah you're right! thanks for the suggestion!

carrodher (dismissed):
LGTM, thanks

PR #9334, opened 2022-03-08
[bitnami/external-dns] Add support for OCI

mschmidt291:
Signed-off-by: Max Schmidt <ma.schmidt@reply.de>

**Description of the change**

This PR adds the OCI provider for Oracle Cloud platform as a usable provider.

**Benefits**

Oracle Cloud can be used

**Possible drawbacks**

Hopefully none ;)

**Checklist**


mschmidt291:
Just seeing that the README metadata check is failing. Looking at the error produced by the check, it seems like it's parsing the README wrong(?). The README has all the fields in it. Maybe it is because of refactoring of markdown syntax?

javsalgar:
Hi,

It seems that the issue is here:

```
  ## @param oci.passphrase When using the OCI provider and your privateKey has a passphrase, put it in here. (optional)
  ##
  privateKeyPassphrase: ""
```
Notice that you set `oci.passphrase` but the value is `privateKeyPassphrase`. This is what the error reports:

```
ERROR: Missing metadata for key: oci.privateKeyPassphrase
ERROR: Metadata provided for non existing key: oci.passphrase

```

mschmidt291:
> Hi,
> 
> It seems that the issue is here:
> 
> ```
>   ## @param oci.passphrase When using the OCI provider and your privateKey has a passphrase, put it in here. (optional)
>   ##
>   privateKeyPassphrase: ""
> ```
> 
> Notice that you set `oci.passphrase` but the value is `privateKeyPassphrase`. This is what the error reports:
> 
> ```
> ERROR: Missing metadata for key: oci.privateKeyPassphrase
> ERROR: Metadata provided for non existing key: oci.passphrase
> ```

Totally missed that one.. Fixed.

fmulero (approved):
Thanks a lot @mschmidt291 for contributing. All your changes looks good to me. Well done!

rouke-broersma:
Just wanted to note that from our perspective OCI is not the most descriptive key in the context of helm charts. At first we thought this update added support for OCI helm charts. We would have preferred if it was called oracle or something, but I guess this was already decided upstream at some point.

fmulero:
Thanks a lot @rouke-broersma for sharing that misunderstood, we appreciate your feedback. We've taken note to try to avoid this issues in the future

mschmidt291:
> Just wanted to note that from our perspective OCI is not the most descriptive key in the context of helm charts. At first we thought this update added support for OCI helm charts. We would have preferred if it was called oracle or something, but I guess this was already decided upstream at some point.

Just to clarify: You are probably right - I chose OCI instead of Oracle, because the --providers flag for the external-dns tool also has "--provider=oci". So it is consistent.

rouke-broersma:
I agree that it is best to keep consistency with the upstream project, just an unfortunate name collision.

bitnami/external-dns/values.yaml
```diff
@@ -423,11 +423,40 @@ linode:
...
+  ##
+  userOCID: ""
+  ## @param oci.compartmentOCID When using the OCI provider, specify your Compartment OCID where your DNS Zone is located in.
+  ##
+  compartmentOCID: ""
+  ## @param oci.privateKey When using the OCI provider, paste in your RSA private key file for the Oracle API
+  ##
+  privateKey: |
+    -----BEGIN RSA PRIVATE KEY-----
+    -----END RSA PRIVATE KEY-----
```
stevo-f3, line 452: @mschmidt291 would it make sense to remove this default, document is somehow somewhere else? this default key value is unusable/invalid anyway and in my fork it's triggering different security scanners, reporting that sensitive/secret data is attempting to be stored in the repository; forks are a must since availability of OSS chart copy is not guaranteed (see https://github.com/bitnami/charts/issues/10539) and retrieving them from internet from runtime environment is not good practice.

IvanI3:
Please, add support for "--oci-auth-instance-principal" flag to exclude mount of oci.yaml secret from chart and use instance principal instead

PR #9343, opened 2022-03-09
[bitnami/wordpress] WordPress test update to follow Test Strategy

superaleks:
**Description of the change**

Added a dynamic data generator function, used to create a title of the post. Improved some of the locators to avoid flakiness and use a smarter locator strategy. 

**Test results**
Executed successfully 3 different times in a row, total 9 executions. Results of the last one: 

- [GKE](https://cp.production.k.bromelia.bitnami.net/v1/execution-graphs/457a76fb-8eb3-42b9-831c-6f582e62d32b/report) - PASS
- [AKS](https://cp.production.k.bromelia.bitnami.net/v1/execution-graphs/f89dcb60-c8cb-4dfe-8e9d-568d1e8aa655/report) - PASS
- [TKG](https://cp.production.k.bromelia.bitnami.net/v1/execution-graphs/217fa52e-9a39-4364-8f6c-9cd8713367de/report) - PASS

**Benefits**
Less flaky tests that can be run many times in a row and don't generate duplicate data. 

**Possible drawbacks**
As always with UI tests, risk of uncovered flakiness. 

**Checklist**

- [X] Test results added - In progress

### DIFF
--- a/.vib/wordpress/cypress/cypress/integration/wordpress_spec.js
+++ b/.vib/wordpress/cypress/cypress/integration/wordpress_spec.js
@@ -36,8 +36,8 @@ it('disallows login to an invalid user', () => {
   cy.clearCookies();
   cy.visit('/wp-login.php')
   cy.fixture('user').then((user) => {
-    cy.get('#user_login').should('not.be.disabled').type(user.username);
-    cy.get('#user_pass').should('not.be.disabled').type(user.password);
+    cy.get('#user_login').should('not.be.disabled').type(user.username).should('have.value', user.username);
+    cy.get('#user_pass').should('not.be.disabled').type(user.password).should('have.value', user.password);
   })
   cy.get('#wp-submit').click();
   cy.fixture('user').then((user) => {
@@ -80,7 +80,7 @@ it('checks if admin can create a post', () => {
   cy.contains('button[type="button"]',"Publish").click();
   cy.get('h1[aria-label="Add title"]').should('be.visible').clear().type(`Test Hello World!${random}`);
   cy.get('.editor-post-save-draft').should('be.visible').click();
-  cy.get('.editor-post-saved-state').should('exist');
+  cy.get('.editor-post-saved-state').should('be.visible').and('have.text','Saved');
 
 })
 
@@ -112,9 +112,7 @@ it('allows the upload of a file',() => {
 
 it('allows to log out', () => {
   cy.login();
-  cy.get("#wp-admin-bar-logout .ab-item").should('exist').click({
-    force: true
-  });
+  cy.visit('wp-login.php?loggedout=true&wp_lang=en_US');
   cy.contains('.message','logged out');
 })
 
--- a/.vib/wordpress/cypress/cypress/integration/wordpress_spec.js
+++ b/.vib/wordpress/cypress/cypress/integration/wordpress_spec.js
@@ -1,12 +1,16 @@
 /// <reference types="cypress" />
 
+import {
+  random,
+} from './utils'
+
 it('contains the sample blogpost and a comment', () => {
   cy.visit('/');
   cy.get('.wp-block-query').should('exist');
-  cy.get('.wp-block-post-title a').click();
+  cy.contains('.wp-block-post-title a','Hello World').click();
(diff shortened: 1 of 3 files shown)


.vib/wordpress/cypress/cypress/integration/wordpress_spec.js
```diff
@@ -68,13 +72,16 @@ it('checks if admin can edit a site', () => {
...
-  cy.get('#post-1 .row-title').should('exist').click();
-  cy.url().should('include', '/post.php?post=1&action=edit');
-  cy.get('#editor').should('exist');
+  cy.visit('/wp-admin/post-new.php');
+  cy.get('.components-modal__header > .components-button').click();
+  cy.get('#editor').should('be.visible');
+  cy.contains('button[type="button"]',"Publish").click();
+  cy.get('h1[aria-label="Add title"]').should('be.visible').clear().type(`Test Hello World!${random}`);
+  cy.get('.editor-post-save-draft').should('be.visible').click();
+  cy.get('.editor-post-saved-state').should('exist');
```
joancafom, line 83: I think this assertion suits better:

```suggestion
  cy.get('.editor-post-saved-state').should('be.visible');
```
`cy.get()` will already throw an error if the specified element does not exist in the DOM. Additionally, the element may exist but be hidden, so I think this might be more helpful.

.vib/wordpress/cypress/cypress/integration/wordpress_spec.js
```diff
@@ -96,17 +103,18 @@ it('checks the value of auto update status', () => {
...
+  cy.contains("[aria-labelledby]",'Select Files').should('be.visible');
   cy.get('input[type=file]').selectFile('cypress/fixtures/images/test_image.jpeg',
     {force:true});
   cy.get('.attachment').should('be.visible');
 })
 
 it('allows to log out', () => {
   cy.login();
-  cy.get("#wp-admin-bar-logout .ab-item").click({
+  cy.get("#wp-admin-bar-logout .ab-item").should('exist').click({
```
joancafom, line 115: Same thoughts here about `exists`. I personally prefer `be.visible` over `exist` in most cases.

.vib/wordpress/cypress/cypress/support/commands.js
```diff
@@ -20,7 +20,7 @@ Cypress.Commands.add("login", (
...
   cy.clearCookies();
   cy.visit('/wp-login.php')
-  cy.get('#user_login').type(username);
-  cy.get('#user_pass').type(password);
-  cy.get('#wp-submit').click();
-});
\ No newline at end of file
+  cy.get('#user_login').should('not.be.disabled').type(username);
+  cy.get('#user_pass').should('not.be.disabled').type(password);
+  cy.get('#wp-submit').should('be.visible').click();
```
joancafom, line 25: Hmm I think that we should not be performing assertion to check that a given functionality works in `commands.js`, but rather create a test case if that is what we intend to. 

In fact, we are already verifying the same in:
https://github.com/bitnami/charts/blob/bea8ff264dc62115a606a3ea4c8b23efd9275834/.vib/wordpress/cypress/cypress/integration/wordpress_spec.js#L39-L40

.vib/wordpress/cypress/cypress/integration/wordpress_spec.js
superaleks: I made this change because I noticed the `be.visible` assertion failing randomly and adding some flakiness... Let me think about a third way.

.vib/wordpress/cypress/cypress/integration/wordpress_spec.js
superaleks: See comment above. I'll try to see a better way.

.vib/wordpress/cypress/cypress/support/commands.js
superaleks: I agree, but I'm afraid this is needed not for testing purposes but flakiness avoidance. I've seen the tests fail because the input field can't be typed into because it has been marked as disabled. This check helps with this case.

joancafom (approved):
LGTM

PR #9356, opened 2022-03-09
[bitnami/redmine] Chart standardized

juan131:

**Description of the change**

This PR standardizes the `bitnami/redmine` chart to be in line with the rest of the catalog.

**Benefits**

Standardization.

**Possible drawbacks**

None. No major bump is required since breaking changes were added support for deprecated values using `coalesce`.

**Applicable issues**

N/A

**Additional information**

N/A

**Checklist**

- [x] Variables are documented in the values.yaml and added to the README.md using (readme-generator-for-helm)[https://github.com/bitnami-labs/readme-generator-for-helm]


bitnami/redmine/templates/deployment.yaml
```diff
@@ -174,46 +179,36 @@ spec:
             {{- end }}
           ports:
             - name: http
-              containerPort: {{ .Values.containerPort }}
+              containerPort: {{ coalesce .Values.containerPorts.http .Values.containerPort }}
+          {{- if not .Values.diagnosticMode.enabled }}
           {{- if .Values.livenessProbe.enabled }}
-          livenessProbe:
+          livenessProbe: {{- omit .Values.livenessProbe "enabled" "path" | toYaml | nindent 12 }}
```
migruiz4, line 185: Why are you no longer using?
```
{{- include "common.tplvalues.render" (dict "value" (omit .Values.livenessProbe "enabled" "path") "context" $) | nindent 12 }}
```

bitnami/redmine/templates/deployment.yaml
```diff
@@ -233,7 +228,7 @@ spec:
         - name: redmine-data
           {{- if .Values.persistence.enabled }}
           persistentVolumeClaim:
-            claimName: {{ .Values.persistence.existingClaim | default (include "common.names.fullname" .) }}
+            claimName: {{ (tpl .Values.persistence.existingClaim $) | default (include "common.names.fullname" .) }}
```
migruiz4, line 231:
```suggestion
            claimName: {{ default (include "common.names.fullname" .) (tpl .Values.persistence.existingClaim $) }}
```

bitnami/redmine/templates/svc.yaml
```diff
@@ -18,30 +19,33 @@ metadata:
   {{- end }}
 spec:
   type: {{ .Values.service.type }}
-  sessionAffinity: {{ default "None" .Values.service.sessionAffinity }}
+  sessionAffinity: {{ .Values.service.sessionAffinity }}
```
migruiz4, line 22: If removed `default "None"`, shouldn't this be inside a conditional?

bitnami/redmine/templates/svc.yaml
```diff
@@ -18,30 +19,33 @@ metadata:
...
   {{- end }}
   {{- if (or (eq .Values.service.type "LoadBalancer") (eq .Values.service.type "NodePort")) }}
   externalTrafficPolicy: {{ .Values.service.externalTrafficPolicy | quote }}
   {{- end }}
-  {{- if (and (eq .Values.service.type "LoadBalancer") .Values.service.loadBalancerSourceRanges) }}
-  loadBalancerSourceRanges: {{- toYaml . | nindent 4 }}
+  {{- if eq .Values.service.type "LoadBalancer" }}
+  {{- with .Values.service.loadBalancerSourceRanges }}
+  loadBalancerSourceRanges: {{ .Values.service.loadBalancerSourceRanges }}
   {{- end }}
```
migruiz4, line 32: Why the change from
```
  {{- if and (eq .Values.service.type "LoadBalancer") .Values.service.loadBalancerSourceRanges }}
  {{- end }}
```

bitnami/redmine/values.yaml
```diff
@@ -396,18 +422,11 @@ ingress:
...
   ##
   enabled: false
-  ## @param ingress.className Set the ingressClassName which will be used for this ingress
-  ## The newer ingressClassName field on Ingresses is a replacement for that "kubernetes.io/ingress.class" annotation (Kubernetes 1.18+).
-  ## When using this filed you need to create the IngressClass resource with the corresponding name.
+  ## @param ingress.className IngressClass that will be be used to implement the Ingress (Kubernetes 1.18+)
+  ## This is supported in Kubernetes 1.18+ and required if you have more than one IngressClass marked as the default for your cluster .
+  ## ref: https://kubernetes.io/blog/2020/04/02/improvements-to-the-ingress-api-in-kubernetes-1.18/
   ##
   className: ""
```
migruiz4, line 429: Could you please add support for `ingressClassName` using `coalesce`? Don't want to bump a major, but would like the charts to at least support the same values.

bitnami/redmine/values.yaml
```diff
@@ -812,30 +757,25 @@ mailReceiver:
   ## @param mailReceiver.allowOverride Defines if email content is allowed to set attributes values. Values is a comma separated list of attributes or `all` to allow all attributes
   ##
   allowOverride: ""
+  ## @param mailReceiver.command Override default container command (useful when using custom images)
+  ##
+  command: ["bash"]
+  ## @param mailReceiver.args Override default container args (useful when using custom images)
+  ##
+  args: ["-c", "/cj/receive.sh"]
```
migruiz4, line 765: If we are setting here `["bash"]` and `["-c", "/cj/receive.sh"]`,  then these values aren't really meant to override default as their description says.

I think we should move these inside the Deployment template, and leave them empty `[""]` for users who'd like to overwrite them.

migruiz4 (changes requested):
Please check my comments below.

In addition, I would appreciate it if you could also have a look at the `cronjob.yaml`.

bitnami/redmine/templates/deployment.yaml
juan131: It's exactly the same thing

bitnami/redmine/templates/svc.yaml
juan131: Nope. `service.sessionAffinity` is set by default in the values.yaml

migruiz4 (dismissed):

Issue #7382, opened 2021-09-02
[bitnami/mongodb] Option for using emptyDir with medium: memory


MikaelElkiaer:
**Which chart**:
mongodb

**Is your feature request related to a problem? Please describe.**
I am using mongodb as a cache running in memory only.

**Describe the solution you'd like**
I would like to set the emptyDir which is used in case of `persistence: false` to `medium: memory`.

**Describe alternatives you've considered**
Well, using something else than mongo. :)

**Additional context**

MikaelElkiaer:
This is not really useful unless #5752 gets fixed.

pablogalegoc:
Hi @MikaelElkiaer! Sorry for the late response, we had an outage on our support service last week and lost some cases.
Indeed the issue you point is going to be blocking for this. Those kind of reworks in the Docker images may take time. I'm going to mention your use case in the task and see if it can get some traction but, as my teammate said, we can't give an ETA, sorry.

MikaelElkiaer:
No problem.

MikaelElkiaer:
Actually, I don't think this is an image change - correct me if I am wrong.
It is a mater of switching from using emptyDir in memory, as opposed to the node's filesystem.
Unless I am wrong, it should be a matter of changing the template to allow switching between default and memory when having persistence to false (i. e. using emptyDir).

pablogalegoc:
IMO we'll have to change the logic of the initialization scripts in the image. If `pod-0` loses its data, for example by having set `emptyDir` but being rescheduled on another cluster node, the MongoDB cluster won't be able to recover because when `pod-0` is restored, it will be again configured as a Primary replica and then reconnected to the cluster, which has already assigned a new Primary replica automatically.

MikaelElkiaer:
Isn't that a separate problem? Running emptyDir will cause that problem, no matter if it is memory or local fs.

I am already running a single node replica set due to that problem, so in my case it won't matter.

pablogalegoc:
Ah, my bad, I misunderstood what you wanted here once I started to dive into the issue linked above. I'm sorry 🤦 
> Running emptyDir will cause that problem, no matter if it is memory or local fs.

Yes, you are 100% right. However we usually don't provide this as a configurable parameter. We use it in special cases like [this](https://github.com/bitnami/charts/blob/master/bitnami/postgresql-ha/README.md#use-a-volume-for-devshm) or for ssl certs:
[...]

Issue #6741, opened 2021-06-23
[bitnami/mongodb] restarted "primary" pod cannot re-join the Replica Set


bmv-ce:
**Which chart**:
bitnami/mongodb
version: 10.19.0

**Describe the bug**
Restarted "primary" pod is unable to re-join the replica set, failing with "replica set IDs do not match, ours: <>; remote node's: <>"

**To Reproduce**

1. Create a new Replica Set with 1 Primary and 2 Secondary nodes (note that a persistent volume is used, as I think it may be related to the problem):
>   helm install identification-service-rs0-mongodb bitnami/mongodb \
>	--set metrics.enabled=true \
>	--set arbiter.enabled=false \
>	--set persistence.size=100Gi \
>	--set persistence.mountPath=/mongodb \
>	--set persistence.storageClass=glusterfs-storage \
>	--set architecture=replicaset \
>	--set auth.enabled=true \
>	--set auth.rootPassword=XXXXX \
>	--set auth.username=myuser \
>	--set auth.password=XXXXXX \
>	--set auth.database=testdb \
>	--set auth.replicaSetKey=XXXXX \
>	--set replicaCount=3 \
>	--set resources.limits.cpu=1.0,resources.limits.memory=2Gi,resources.requests.cpu=0.5,resources.requests.memory=1Gi

2. As a result - 3 pods are successfully started, see each other and belong to the same replica set:

> ❯ kubectl get pods
> NAME                                   READY   STATUS    RESTARTS   AGE
> identification-service-rs0-mongodb-0   2/2     Running   0          151m
> identification-service-rs0-mongodb-1   2/2     Running   0          3h15m
> identification-service-rs0-mongodb-2   2/2     Running   0          3h49m

3. Replica Set status is correct:

>{
>	"set" : "rs0",
>	"date" : ISODate("2021-06-23T10:07:55.710Z"),
>	"myState" : 1,
>	"term" : NumberLong(2),
>	"syncSourceHost" : "",
>	"syncSourceId" : -1,
>	"heartbeatIntervalMillis" : NumberLong(2000),
>	"majorityVoteCount" : 2,
>	"writeMajorityCount" : 2,
>	"votingMembersCount" : 3,
>	"writableVotingMembersCount" : 3,
>	"optimes" : {
>		"lastCommittedOpTime" : {
>			"ts" : Timestamp(1624442872, 1),
>			"t" : NumberLong(2)
>		},
[...]

Issue #5752, opened 2021-03-10
[mongo] Redeploying designated initial primary pod after node (and storage) failure caused desync of ReplicaSet


mklokocka:
**Which chart**:
bitnami/mongodb 9.3.1

**Describe the bug**
We are using mongodb set up using bitnami helm chart. We are using the `replicaset` architecture with 3 replicas. Since there are high volumes of data going through the DB, we wanted to avoid network overhead and we are using local storage of hosts for persistence (using https://github.com/rancher/local-path-provisioner). We are running on 3 different nodes. The node running our "initial primary pod" `mongodb-0` failed permanently, and we were forced to recreate it. Note, that the replica set itself was still running with 2 replicas without problems, with `mongodb-0` still being present in the configuration.

However, since the new pod started without any present data, it started to initialize the whole replica set anew, even when there were two pods still up and running with no problems. This meant that the whole replica set started failing because of the mismatched configurations. In the end we were forced to edit the `setup.sh` to force the initial primary pod to point to `mongodb-1`, to allow the newly started `mongodb-0` to be correctly synced with the replica set.

This happens due to the scripts present in the bitnami mongo image, most notably the part [here](https://github.com/bitnami/bitnami-docker-mongodb/blob/master/4.2/debian-10/rootfs/opt/bitnami/scripts/libmongodb.sh#L1038-L1056). As far as I can understand, since there are no db data files in the host (since it is a new node altogether), it reinitializes the whole replica set because the pod considers itself as a primary node due to [this](https://github.com/bitnami/charts/blob/master/bitnami/mongodb/templates/replicaset/scripts-configmap.yaml#L76).

**To Reproduce**
First of all, our `values.yaml`:
```yaml
# Docs: https://github.com/bitnami/charts/tree/master/bitnami/mongodb

# Image setting, replaced by custom image for NUMA compatibility with kustomize.
image:
  tag: 4.2.10-debian-10-r12

# Mongo DB params.
architecture: replicaset

# Helm chart has messy naming conventions for secrets, gotta handle this in script with sed.
auth:
  enabled: true
  existingSecret: <secrets>

replicaSetName: <rs-name>

extraEnvVarsSecret: <secrets>

initdbScriptsConfigMap: mongodb-custom-scripts

# Sets up:
# wiredTigerCacheSizeGB - to properly set internal cache size in containerized environment
# taskExecutorPoolSize - to avoid MongoDB starving CPU from other containers running on the same node.
extraFlags:
[...]

PR #9387, opened 2022-03-11
[bitnami/mongodb] feat: :sparkles: Check replicaset status when persistence is not detected

javsalgar:

**Description of the change**

Currently, when a MongoDB node crashes, it will make use of the existing persistence to re-join a replica set. However, if the PVC is lost (or persistence is disabled), the node will try recreating the replica set from scratch. For covering this use case, we improved the initialization configmap with the following:

- Check if there's persistence
-  In case there's persistence, continue with the default logic
- In case there's no persistence, try querying the replicaset for an existing master
- In case the primary is the same restarting node, configure as master.
- In case the primary is another node, configure as replica of this new primary.

This should mitigate the case of MongoDB as caches and cases where local persistence is used.

**Benefits**

**Possible drawbacks**

**Applicable issues**

- Fixes https://github.com/bitnami/charts/issues/7382
- Fixes https://github.com/bitnami/charts/issues/6741
- Fixes https://github.com/bitnami/charts/issues/5752

**Additional information**

**Checklist**

### DIFF
--- a/bitnami/mongodb/templates/replicaset/scripts-configmap.yaml
+++ b/bitnami/mongodb/templates/replicaset/scripts-configmap.yaml
@@ -114,16 +114,17 @@ data:
       {{- range $e, $i := until $replicaCount }}
       {{- $mongoList = append $mongoList (printf "%s-%d.%s-headless.%s.svc.%s:%d" $fullname $i $fullname $releaseNamespace $clusterDomain $portNumber) }}
       {{- end }}
-      current_primary=$(mongo admin --host "{{ join "," $mongoList }}" {{- if .Values.auth.enabled }} --authenticationDatabase admin -u root -p $MONGODB_ROOT_PASSWORD{{- end }} --eval 'db.runCommand("ismaster")' | awk -F\" '/primary/ {print $4}')
+      current_primary=$(mongo admin --host "{{ join "," $mongoList }}" {{- if .Values.auth.enabled }} --authenticationDatabase admin -u root -p $MONGODB_ROOT_PASSWORD{{- end }}{{-if .Values.tls.enabled}} --tls --tlsCertificateKeyFile=/certs/mongodb.pem --tlsCAFile=/certs/mongodb-ca-cert{{- end }} --eval 'db.runCommand("ismaster")' | awk -F\" '/primary/ {print $4}')
+
       if ! is_empty_value "$current_primary"; then
         info "Detected existing primary: ${current_primary}"
       fi
     fi
 
-    if { ! is_empty_value "$current_primary"; } && [[ "$MONGODB_ADVERTISED_HOSTNAME:$MONGODB_ADVERTISED_PORT_NUMBER" == "$current_primary" ]]; then
+    if ! is_empty_value "$current_primary" && [[ "$MONGODB_ADVERTISED_HOSTNAME:$MONGODB_ADVERTISED_PORT_NUMBER" == "$current_primary" ]]; then
         info "Advertised name matches current primary, configuring node as a primary"
         export MONGODB_REPLICA_SET_MODE="primary"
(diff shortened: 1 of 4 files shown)


bitnami/mongodb/templates/replicaset/scripts-configmap.yaml
```diff
@@ -88,18 +91,56 @@ data:
...
+      {{- end }}
+      current_primary=$(mongo admin --host "{{ join "," $mongoList }}" {{- if .Values.auth.enabled }} --authenticationDatabase admin -u root -p $MONGODB_ROOT_PASSWORD{{- end }} --eval 'db.runCommand("ismaster")' | awk -F\" '/primary/ {print $4}')
+      if ! is_empty_value "$current_primary"; then
+        info "Detected existing primary: ${current_primary}"
+      fi
+    fi
 
-    if [[ "$MY_POD_NAME" = "{{ $fullname }}-0" ]]; then
-        echo "Pod name matches initial primary pod name, configuring node as a primary"
+    if { ! is_empty_value "$current_primary"; } && [[ "$MONGODB_ADVERTISED_HOSTNAME:$MONGODB_ADVERTISED_PORT_NUMBER" == "$current_primary" ]]; then
```
rafariossaa, line 123: I think this could be written as:
```suggestion
    if ! is_empty_value "$current_primary" && [[ "$MONGODB_ADVERTISED_HOSTNAME:$MONGODB_ADVERTISED_PORT_NUMBER" == "$current_primary" ]]; then
```

bitnami/mongodb/templates/replicaset/scripts-configmap.yaml
```diff
@@ -88,18 +91,56 @@ data:
...
+        info "Advertised name matches current primary, configuring node as a primary"
+        export MONGODB_REPLICA_SET_MODE="primary"
+    elif { ! is_empty_value "$current_primary"; } && [[ "$MONGODB_ADVERTISED_HOSTNAME:$MONGODB_ADVERTISED_PORT_NUMBER" != "$current_primary" ]]; then
```
rafariossaa, line 126: Same as above.

bitnami/mongodb/templates/replicaset/scripts-configmap.yaml
```diff
@@ -88,18 +91,56 @@ data:
...
+    elif [[ "$MY_POD_NAME" = "{{ $fullname }}-0" ]]; then
+        info "Pod name matches initial primary pod name, configuring node as a primary"
         export MONGODB_REPLICA_SET_MODE="primary"
     else
-        echo "Pod name doesn't match initial primary pod name, configuring node as a secondary"
+        info "Pod name doesn't match initial primary pod name, configuring node as a secondary"
         export MONGODB_REPLICA_SET_MODE="secondary"
+        export MONGODB_INITIAL_PRIMARY_PORT_NUMBER="$MONGODB_PORT_NUMBER"
+    fi
+
```
rafariossaa, line 140: I am not very familiar with the mongo node role decision, but,  what would happen (or should happen) if `$current_primary` is empty for whatever reason ?  (eg. error when running the command, or there is not primary or ...).

rafariossaa (changes requested):

bitnami/mongodb/templates/replicaset/scripts-configmap.yaml
javsalgar: If current_primary is empty, it will assume the default behavior implemented up until now. It will decide if it is primary or secondary based on the node number.

Note that this logic is only for the case that the PVC is lost, so currently this would not cover the case where the node loses the PVC and is unable to communicate with the rest of the cluster. If that case happens, the user will need to launch the cluster in diagnosticMode and fix it.

rafariossaa:
LGTM

rafariossaa (approved):

PR #9422, opened 2022-03-15
[bitnami/kafka] Allow PEM chain to be used for TLS

### LINKED ISSUE #9378: [bitnami/kafka] Make certificates work with cert-manager
### Name and Version

bitnami/kafka:3.1.0-debian-10-r31

`auth.tls.existingSecrets` currently does not work "out of the box" with cert-manager because the `Secret` resource made by cert-manager does not include a `ca.crt` field, which causes the `auto-discovery.sh` script to fail [here](https://github.com/bitnami/charts/blob/3e7c06fdf291753b37603cc27276b53bd173deec/bitnami/kafka/templates/scripts-configmap.yaml#L141). 

There are many solutions to this problem, but I don't know which will fit best into the Chart. It is also not clear to me whether this CA cert is required. Is there not a central collection of trusted root certificate authorities already in the container? 

Here are some ideas:

1. If there is already a collection of trusted roots, make the `Secret` field `ca.crt` optional and verify the connections against the existing root certs (which could lead to failures if the cert is signed by a non-public CA).
2. Include an environment variable like `TRUSTED_CA_CERTIFICATES` which contains a PEM list file, and split this into multiple files (using something like `csplit certchain.pem '/\-*END CERTIFICATE\-*/+1' '{*}'`) and add these to the central collection of trusted roots.
3. Following the mantra of `auth.tls.existingSecrets`, have a `auth.tls.existingCASecrets` that points to a separate secret that contains a PEM of the CA certificate. Make this optional if there is already trusted roots somewhere.

There are many more ways to solve this, so just listing these to start discussion. I am happy to open a PR with the changes once a solution has been decided on.

There are really only workaround, which generally revolve around editing the script in the `ConfigMap` or adding the CA to the `ca.crt` field in the `Secret` made by cert-manager.

rrileyca:
**Description of the change**

This PR adds the `auth.tls.pemChainIncluded` property in `values.yaml`. This boolean flag allows a `Secret` resource to exclude the `ca.crt` field and use the `tls.crt` field as a PEM chain. The initContainer will extract the intermediate authority certificates from the chain and use those for the truststore.

**Benefits**

Specifically, this allows `cert-manager` to create `Secrets` from an ACME issuer that does not provide the Root certificate authority. 

**Possible drawbacks**

I cannot think of any, this is an optional opt-in.

**Applicable issues**

  - fixes #9378

**Additional information**

If using a certificate automation tool (like cert-manager) which would create certificates with short expiry dates, this will not automatically restart the pod, which runs the risk of certificates expiring at runtime. This was always a problem, but it would be exacerbated by the short lifetime of ACME certificates.

**Checklist**


marcosbc (dismissed):
LGTM! Thanks!

marcosbc:
@rrileyca I'm unable to merge the PR due to the conflicts. It seems like that is why the PR was not closed automatically when I approved it. Could you fix them?

rrileyca:
I re-based and force pushed, resolving the conflicts.

rrileyca:
Sorry had some of the git conflict text in the commit somehow... Something is weird with my IDE. Should be better now!

marcosbc (approved):

PR #9440, opened 2022-03-16
[bitnami/nginx] Switch metric list order in nginx

jakuboskera:
To circumvent issues https://github.com/kubernetes/kubernetes/issues/74099 causes with GitOps and using this chart. 

Same issue like [here](https://github.com/bitnami/charts/pull/7849).

**Description of the change**

Switched resource order in HPA.

**Benefits**

**Possible drawbacks**

Bypasses issue https://github.com/kubernetes/kubernetes/issues/74099 that can cause problems when using GitOps like ArgoCD. As the impact is minimal I believe this is the optimal solution for this problem.

**Applicable issues**

**Additional information**

**Checklist**

- [x] Chart version bumped in Chart.yaml according to [semver](http://semver.org/).

### DIFF
--- a/bitnami/nginx/Chart.yaml
+++ b/bitnami/nginx/Chart.yaml
@@ -25,4 +25,4 @@ name: nginx
 sources:
   - https://github.com/bitnami/bitnami-docker-nginx
   - https://www.nginx.org
-version: 9.9.3
+version: 9.9.4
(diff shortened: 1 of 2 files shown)


miguelaeh (approved):
LGTM!

jakuboskera:
@miguelaeh I can see the new version of helm chart nginx 9.9.4 in repo [bitnami chart](https://charts.bitnami.com/bitnami), but even in this new version, the order of metrics in `hpa.yaml` is still old. It seems that the changes didn't take an effect.

jakuboskera:
seems to be fixed in version 9.9.5

PR #9442, opened 2022-03-16
fix: wrong access_key of objstoreConfig

00arthur00:
Signed-off-by: arthur <yang_yapo@126.com>
we should use amin access_key,otherwise compator and store gateway will not run


rafariossaa:
Hi,
  Could you take a look to the automatic testing ? You need to increase the chart version.

PR #9451, opened 2022-03-17
Quote POSTGRESQL_INITSCRIPTS_PASSWORD

### LINKED ISSUE #9445: Can't use numeric password as initdb.password
### Name and Version

bitnami/postgresql 11.1.7

Try to deploy chart with initdb.password=12345678

### What do you see instead?

```
Error from server (BadRequest): error when creating "deploy.yaml": StatefulSet in version "v1" cannot be handled as a StatefulSet: json: cannot unmarshal number into Go struct field EnvVar.spec.template.spec.containers.env.value of type string
```

### Additional information

The password value is not quoted in manifest:

```yaml
            {{- if .Values.primary.initdb.password }}
            - name: POSTGRESQL_INITSCRIPTS_PASSWORD
              value: {{ .Values.primary.initdb.password }}
            {{- end }}
```

EBCEEB:
... to be able to use numeric password

Fixes #9445

### DIFF
--- a/bitnami/postgresql/templates/primary/statefulset.yaml
+++ b/bitnami/postgresql/templates/primary/statefulset.yaml
@@ -276,7 +276,7 @@ spec:
             {{- end }}
             {{- if .Values.primary.initdb.password }}
             - name: POSTGRESQL_INITSCRIPTS_PASSWORD
-              value: {{ .Values.primary.initdb.password }}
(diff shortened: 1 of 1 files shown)


carrodher:

In the same way, since a new Helm chart release is going to be needed, the chart version needs to be bumped in the _Chart.yaml_ file

PR #9453, opened 2022-03-17
[bitnami/postgresql] Quote POSTGRESQL_INITSCRIPTS_PASSWORD

### LINKED ISSUE #9445: Can't use numeric password as initdb.password

EBCEEB:
... to be able to use numeric initdb.password

Fixes #9445

Signed-off-by: Alexander Evseev <aevseev@gmail.com>

### DIFF
--- a/bitnami/postgresql/Chart.yaml
+++ b/bitnami/postgresql/Chart.yaml
@@ -26,4 +26,4 @@ name: postgresql
 sources:
   - https://github.com/bitnami/bitnami-docker-postgresql
   - https://www.postgresql.org/
-version: 11.1.8
+version: 11.1.9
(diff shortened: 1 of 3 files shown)


fmulero (dismissed):
Thank you so much @EBCEEB ! It looks great!

PR #9487, opened 2022-03-19
[bitnami/argo-cd] Fix nil pointer evaluating interface {}.server 

grachevko:


marcosbc (approved):
LGTM! Thanks for the fix!

PR #9495, opened 2022-03-21
[bitnami/solr] changes tls.existingSecret to tls.certificatesSecretName in _helpers.tpl

### LINKED ISSUE #9462: [bitnami/solr] tls.existingSecret
### Name and Version

bitnami/solr

1. I tried to configure TLS (`values.yaml`):
```
tls:
  enabled: true
  certificatesSecretName: solr-tls-cert
  passwordsSecretName: solr-pkcs12-password-secret
```
2. `helm template . ` gives me the following error:
```
VALUES VALIDATION:

solr: tls.enabled
    In order to enable TLS, you also need to provide
    an existing secret containing the Keystore and Truststore or
    enable auto-generated certificates.
```

The Parameter `tls.existingSecret` is not documented (I think that's the only problem):
https://github.com/bitnami/charts/blob/4e3e2e24823338390171d533c49807788da881a2/bitnami/solr/templates/_helpers.tpl#L173

I expect that my the provided certificates should be enough when I set `enabled: true` and `autoGenerated: false` (which is the default).

DirkDittmarMonday:
It seems to me that there is a conflict with the names.
Checking `tls.existingSecret` in `_helpers.tpl` seems to be wrong.

Signed-off-by: Dirk Dittmar <dirk.dittmar@monday-consulting.com>

**Applicable issues**

fixes #9462

**Checklist**


miguelaeh (approved):
LGTM!

Thank you very much for fixing this bug!

PR #9512, opened 2022-03-22
Update template to the latest standardisations

miguelaeh:
**Description of the change**

**Benefits**

**Possible drawbacks**

**Applicable issues**

  - fixes #

**Additional information**

**Checklist**

### DIFF
--- a/template/CHART_NAME/templates/deployment.yaml
+++ b/template/CHART_NAME/templates/deployment.yaml
@@ -55,11 +55,14 @@ spec:
       {{- if .Values.%%MAIN_OBJECT_BLOCK%%.schedulerName }}
       schedulerName: {{ .Values.%%MAIN_OBJECT_BLOCK%%.schedulerName | quote }}
       {{- end }}
+      {{- if .Values.topologySpreadConstraints }}
+      topologySpreadConstraints: {{- include "common.tplvalues.render" (dict "value" .Values.%%MAIN_OBJECT_BLOCK%%.topologySpreadConstraints "context" .) | nindent 8 }}
+      {{- end }}
       {{- if .Values.%%MAIN_OBJECT_BLOCK%%.podSecurityContext.enabled }}
       securityContext: {{- omit .Values.%%MAIN_OBJECT_BLOCK%%.podSecurityContext "enabled" | toYaml | nindent 8 }}
       {{- end }}
       initContainers:
-        {{- if .Values.volumePermissions.enabled }}
+        {{- if and .Values.volumePermissions.enabled .Values.persistence.enabled }}
         - name: volume-permissions
           image: {{ include "%%TEMPLATE_NAME%%.volumePermissions.image" . }}
           imagePullPolicy: {{ .Values.volumePermissions.image.pullPolicy | quote }}
@@ -71,7 +74,7 @@ spec:
           {{- end }}
           volumeMounts:
             - name: foo
-              mountPath: bar
+              mountPath: /bitnami/%%MAIN_OBJECT_BLOCK%%
         {{- end }}
         {{- if .Values.%%MAIN_OBJECT_BLOCK%%.initContainers }}
           {{- include "common.tplvalues.render" (dict "value" .Values.%%MAIN_OBJECT_BLOCK%%.initContainers "context" $) | nindent 8 }}
@@ -143,7 +146,7 @@ spec:
           {{- end }}
           volumeMounts:
             - name: foo
-              mountPath: bar
+              mountPath: /bitnami/%%MAIN_OBJECT_BLOCK%%
           {{- if .Values.%%MAIN_OBJECT_BLOCK%%.extraVolumeMounts }}
           {{- include "common.tplvalues.render" (dict "value" .Values.%%MAIN_OBJECT_BLOCK%%.extraVolumeMounts "context" $) | nindent 12 }}
           {{- end }}
@@ -152,8 +155,12 @@ spec:
         {{- end }}
       volumes:
         - name: foo
-          hostPath:
-            path: bar
+        {{- if .Values.persistence.enabled }}
+          persistentVolumeClaim:
+            claimName: {{ default (include "common.names.fullname" .) .Values.persistence.existingClaim }}
+        {{- else }}
+          emptyDir: {}
+        {{- end }}
         {{- if .Values.%%MAIN_OBJECT_BLOCK%%.extraVolumes }}
         {{- include "common.tplvalues.render" (dict "value" .Values.%%MAIN_OBJECT_BLOCK%%.extraVolumes "context" $) | nindent 8 }}
         {{- end }}
--- a/template/CHART_NAME/templates/deployment.yaml
+++ b/template/CHART_NAME/templates/deployment.yaml
@@ -61,6 +61,9 @@ spec:
       {{- if .Values.%%MAIN_OBJECT_BLOCK%%.podSecurityContext.enabled }}
       securityContext: {{- omit .Values.%%MAIN_OBJECT_BLOCK%%.podSecurityContext "enabled" | toYaml | nindent 8 }}
       {{- end }}
+      {{- if .Values.%%MAIN_OBJECT_BLOCK%%.terminationGracePeriodSeconds }}
+      terminationGracePeriodSeconds: {{ .Values.%%MAIN_OBJECT_BLOCK%%.terminationGracePeriodSeconds }}
+      {{- end }}
       initContainers:
         {{- if and .Values.volumePermissions.enabled .Values.persistence.enabled }}
         - name: volume-permissions
--- a/template/CHART_NAME/templates/deployment.yaml
+++ b/template/CHART_NAME/templates/deployment.yaml
@@ -55,7 +55,7 @@ spec:
       {{- if .Values.%%MAIN_OBJECT_BLOCK%%.schedulerName }}
       schedulerName: {{ .Values.%%MAIN_OBJECT_BLOCK%%.schedulerName | quote }}
       {{- end }}
-      {{- if .Values.topologySpreadConstraints }}
+      {{- if .Values.%%MAIN_OBJECT_BLOCK%%.topologySpreadConstraints }}
       topologySpreadConstraints: {{- include "common.tplvalues.render" (dict "value" .Values.%%MAIN_OBJECT_BLOCK%%.topologySpreadConstraints "context" .) | nindent 8 }}
       {{- end }}
       {{- if .Values.%%MAIN_OBJECT_BLOCK%%.podSecurityContext.enabled }}
@@ -74,7 +74,8 @@ spec:
           {{- end }}
           volumeMounts:
             - name: foo
-              mountPath: /bitnami/%%MAIN_OBJECT_BLOCK%%
+              mountPath: {{ .Values.persistence.path }}
+              subPath: {{ .Values.persistence.subPath }}
         {{- end }}
         {{- if .Values.%%MAIN_OBJECT_BLOCK%%.initContainers }}
           {{- include "common.tplvalues.render" (dict "value" .Values.%%MAIN_OBJECT_BLOCK%%.initContainers "context" $) | nindent 8 }}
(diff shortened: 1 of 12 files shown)


template/CHART_NAME/templates/deployment.yaml
```diff
@@ -55,11 +55,14 @@ spec:
       {{- if .Values.%%MAIN_OBJECT_BLOCK%%.schedulerName }}
       schedulerName: {{ .Values.%%MAIN_OBJECT_BLOCK%%.schedulerName | quote }}
       {{- end }}
+      {{- if .Values.topologySpreadConstraints }}
```
migruiz4, line 58:
```suggestion
      {{- if .Values.%%MAIN_OBJECT_BLOCK%%.topologySpreadConstraints }}
```

template/CHART_NAME/templates/role.yaml
```diff
@@ -0,0 +1,25 @@
...
+  {{- if .Values.commonAnnotations }}
+  annotations: {{- include "common.tplvalues.render" ( dict "value" .Values.commonAnnotations "context" $ ) | nindent 4 }}
+  {{- end }}
+rules:
+  - apiGroups: [""]
+    resources: ["pods", "persistentvolumeclaims"]
+    verbs: ["get", "watch", "list", "create", "delete"]
+  - apiGroups: [""]       # "" indicates the core API group
+    resources: ["events"]
+    verbs: ["get", "watch", "list"]
```
migruiz4, line 21: Replace these rules with a placeholder.

template/CHART_NAME/templates/pvc.yaml
```diff
@@ -0,0 +1,36 @@
...
+    {{- include "common.tplvalues.render" ( dict "value" .Values.persistence.annotations "context" $ ) | nindent 4 }}
+    {{- end }}
+    {{- if .Values.commonAnnotations }}
+    {{- include "common.tplvalues.render" ( dict "value" .Values.commonAnnotations "context" $ ) | nindent 4 }}
+    {{- end }}
+  {{- end }}
+spec:
+  {{- if .Values.persistence.hostPath }}
+  storageClassName: ""
+  {{- end }}
```
migruiz4, line 23: I see several issues with this:

1- `persistence.hostPath` is only used if `pv.yaml` is also created.
2- This section would conflict with `%%MAIN_OBJECT_BLOCK%%.storageClass`.
3- Duplicated, since setting `persistence.storageClass: "-"` has the same effect. Instead we should add a note next to `hostPath` mentioning that setting `storageClass: "-"` is required.

template/CHART_NAME/templates/pvc.yaml
```diff
@@ -0,0 +1,36 @@
...
+  {{- range .Values.persistence.accessModes }}
+    - {{ . | quote }}
+  {{- end }}
+  {{- else }}
+    - {{ .Values.persistence.accessMode | quote }}
+  {{- end }}
+  resources:
+    requests:
+      storage: {{ .Values.persistence.size | quote }}
+  {{- include "%%MAIN_OBJECT_BLOCK%%.storageClass" . | nindent 2 }}
```
migruiz4, line 35: This helper does not exist at `_helpers.tpl`

template/CHART_NAME/templates/pvc.yaml
```diff
@@ -0,0 +1,36 @@
...
+  {{- end }}
+  accessModes:
+  {{- if not (empty .Values.persistence.accessModes) }}
+  {{- range .Values.persistence.accessModes }}
+    - {{ . | quote }}
+  {{- end }}
+  {{- else }}
+    - {{ .Values.persistence.accessMode | quote }}
+  {{- end }}
+  resources:
```
migruiz4, line 32: `accessMode` is deprecated, new charts should only support `accessModes: []`. No need to have two values with the same functionality.

The usage of this if/else was to avoid bumping a major release when possible.

template/CHART_NAME/templates/pvc.yaml
```diff
@@ -0,0 +1,36 @@
+{{- if and .Values.persistence.enabled (not .Values.persistence.existingClaim) -}}
```
migruiz4, line 1: Suggestion:
```
{{- if and .Values.persistence.enabled (not .Values.persistence.existingClaim) }}
kind: PersistentVolumeClaim
apiVersion: v1
metadata:
  name: {{ template "common.names.fullname" . }}
    app.kubernetes.io/component: %%COMPONENT_NAME%%
    {{- end }}
  {{- if or .Values.persistence.annotations .Values.commonAnnotations }}
  annotations:
    {{- if .Values.persistence.annotations }}
    {{- include "common.tplvalues.render" ( dict "value" .Values.persistence.annotations "context" $ ) | nindent 4 }}
    {{- end }}
    {{- end }}
  {{- end }}
spec:
  accessModes:
  {{- range .Values.persistence.accessModes }}
    - {{ . | quote }}
  {{- end }}
  resources:
    requests:
      storage: {{ .Values.persistence.size | quote }}
  {{- include "common.storage.class" (dict "persistence" .Values.persistence "global" .Values.global) | nindent 2 }}
{{- end }}

```

template/CHART_NAME/values.yaml
```diff
@@ -445,6 +497,55 @@ ingress:
...
+persistence:
+  ## @param persistence.enabled Enable persistence using Persistent Volume Claims
+  ##
+  enabled: true
+  ## @param persistence.path Path to mount the volume at.
+  ##
+  path: /bitnami/%%MAIN_OBJECT_BLOCK%%/data
+  ## @param persistence.subPath The subdirectory of the volume to mount to, useful in dev environments and one PV for multiple services
+  ##
+  subPath: ""
```
migruiz4, line 512: `persistence.subPath` was never implemented

template/CHART_NAME/values.yaml
```diff
@@ -445,6 +497,55 @@ ingress:
...
+## Enable persistence using Persistent Volume Claims
+## ref: https://kubernetes.io/docs/user-guide/persistent-volumes/
+##
+persistence:
+  ## @param persistence.enabled Enable persistence using Persistent Volume Claims
+  ##
+  enabled: true
+  ## @param persistence.path Path to mount the volume at.
+  ##
+  path: /bitnami/%%MAIN_OBJECT_BLOCK%%/data
```
migruiz4, line 509: `persistence.path` is not implemented either. 
In addition, I think both `path` and `subPath` cover the same functionality, but I think `subPath` is already present in more charts.

template/CHART_NAME/values.yaml
```diff
@@ -445,6 +497,55 @@ ingress:
...
+  ## @param persistence.size Size of data volume
+  ##
+  size: 8Gi
+  ## @param persistence.matchLabels Persistent Volume selectors
+  ## https://kubernetes.io/docs/concepts/storage/persistent-volumes/#selector
+  ##
+  matchLabels: {}
+  ## @param persistence.matchExpressions matchExpressions Persistent Volume selectors
+  ##
+  matchExpressions: {}
```
migruiz4, line 537: In any case, these two would be substituted by `persistence.selector` which is not implemented either.

template/CHART_NAME/values.yaml
```diff
@@ -445,6 +497,55 @@ ingress:
...
+  ##
+  existingClaim: ""
+  ## @param persistence.selector Selector to match an existing Persistent Volume for WordPress data PVC
+  ## If set, the PVC can't have a PV dynamically provisioned for it
+  ## E.g.
+  ## selector:
+  ##   matchLabels:
+  ##     app: my-app
+  ##
+  selector: {}
```
migruiz4, line 548: Not implemented. If we decide to add `persistence.selector` as a default in this template, I would suggest adding these two values to `pvc.yaml`:
```
  {{- if .Values.persistence.selector }}
  selector: {{- include "common.tplvalues.render" (dict "value" .Values.persistence.selector "context" $) | nindent 4 }}
  {{- end }}
  {{- if .Values.persistence.dataSource }}
  dataSource: {{- include "common.tplvalues.render" (dict "value" .Values.persistence.dataSource "context" $) | nindent 4 }}
  {{- end }}
```

template/CHART_NAME/values.yaml
```diff
@@ -445,6 +497,55 @@ ingress:
   ##
   secrets: []
```
migruiz4, line 499: Be careful, I think `persistence` would fall under the "Traffic Exposure" section.

migruiz4 (changes requested):
Please have a look at my comments below

template/CHART_NAME/templates/service.yaml
```diff
@@ -13,6 +13,12 @@ metadata:
...
   type: {{ .Values.service.type }}
+  {{- if .Values.service.sessionAffinity }}
+  sessionAffinity: {{ .Values.service.sessionAffinity }}
+  {{- end }}
+  {{- if .Values.service.sessionAffinityConfig }}
+  sessionAffinityConfig: {{- include "common.tplvalues.render" (dict "value" .Values.service.sessionAffinityConfig "context" $) | nindent 4 }}
+  {{- end }}
   {{- if or (eq .Values.service.type "LoadBalancer") (eq .Values.service.type "NodePort") }}
   externalTrafficPolicy: {{ .Values.service.externalTrafficPolicy | quote }}
   {{- end }}
```
javsalgar, line 24: Don't hardcode the port in the service, add reference to `service.ports.foo`

template/CHART_NAME/templates/servicemonitor.yaml
```diff
@@ -0,0 +1,50 @@
...
+      {{- if .Values.metrics.serviceMonitor.honorLabels }}
+      honorLabels: {{ .Values.metrics.serviceMonitor.honorLabels }}
+      {{- end }}
+      {{- if .Values.metrics.serviceMonitor.metricRelabelings }}
+      metricRelabelings: {{- include "common.tplvalues.render" ( dict "value" .Values.metrics.serviceMonitor.metricRelabelings "context" $) | nindent 8 }}
+      {{- end }}
+      {{- if .Values.metrics.serviceMonitor.relabelings }}
+      relabelings: {{- include "common.tplvalues.render" ( dict "value" .Values.metrics.serviceMonitor.relabelings "context" $) | nindent 8 }}
+      {{- end }}
+  namespaceSelector:
```
javsalgar, line 50: I'm missing something here

javsalgar (changes requested):

template/CHART_NAME/values.yaml
miguelaeh: They are not the same functionaility. The subpath is useful when mounting files inside directories while the path when mounting directories. I just forgot to add the invocation in the deployment/statefulset/damoenset. Fixing it.

migruiz4 (changes requested):
Please check my comments.

In addition, could you please implement the value `%%MAIN_OBJECT_BLOCK%%.terminationGracePeriodSeconds`?

template/CHART_NAME/values.yaml
```diff
@@ -250,6 +250,12 @@ diagnosticMode:
     ## Can be set to RollingUpdate or OnDelete
```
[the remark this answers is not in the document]
miguelaeh: Sorry, I accidentally deleted the commit that contained it, just recover it

template/CHART_NAME/templates/deployment.yaml
```diff
@@ -71,7 +74,8 @@ spec:
           {{- end }}
           volumeMounts:
             - name: foo
-              mountPath: bar
+              mountPath: {{ .Values.persistence.path }}
+              subPath: {{ .Values.persistence.subPath }}
```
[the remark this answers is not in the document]
miguelaeh: Well, note this is a template, specific cases would be handled while deploying an actual chart.

template/CHART_NAME/values.yaml
miguelaeh: Not sure what you mean

migruiz4: The values metadata divides the values.yaml using `@section`.

Currently, the `persistence.*` values are located under `## @section Traffic Exposure Parameters`
We usually have them under their own section `@section Persistence parameters`

migruiz4 (changes requested):
Two pending requests from my side:
* Add support for `%%MAIN_OBJECT_BLOCK%%.terminationGracePeriodSeconds`.
* Add `@section Persistence Parameters` on top of `persistence.*` values.

Apart from that, looks good to me

migruiz4 (approved):
LGTM!

PR #9529, opened 2022-03-23
[bitnami/harbor] Fix updateStrategy in all components

### LINKED ISSUE #9515: [bitnami/harbor] rollingUpdate may not be specified when strategy `type` is 'Recreate'
### Name and Version

bitnami/harbor 12.1.3

Installing the bitnami/harbor chart with `jobservice.updateStrategy.type` set to `Recreate` causes the following error in Kubernetes:

```shell
The Deployment "foo-harbor-jobservice" is invalid: spec.strategy.rollingUpdate: Forbidden: may not be specified when strategy `type` is 'Recreate'
```

This is because `rollingUpdate` is set to `{}` [here](https://github.com/bitnami/charts/blob/5b159a70ec71d4a59280be1cdb8091660502e51e/bitnami/harbor/values.yaml#L1627) and the entire `jobservice.updateStrategy` [is included in the template](https://github.com/bitnami/charts/blob/5b159a70ec71d4a59280be1cdb8091660502e51e/bitnami/harbor/templates/jobservice/jobservice-dpl.yaml#L20). It would be a much better user experience if `rollingUpdate` was only included if `type=RollingUpdate`.

## MWE

Here's a minimal "working" example of this bug:
 
```shell
helm template foo bitnami/harbor --version 12.1.3 --set 'jobservice.updateStrategy.type=Recreate' | grep '# Source: harbor/templates/jobservice/jobservice-dpl.yaml' -A 123 | kubectl apply --dry-run=server -f -
```

`--set 'jobservice.updateStrategy.type=Recreate'`

(this ought to be true for all `*.updateStrategy.type` fields, but I have only tried it on `jobservice`)

Setting `*.updateStrategy.type=Recreate` should be valid without having to do other things to make sure it works.

### What do you see instead?

The Kubernetes server returns the following error:

```
The Deployment "foo-harbor-jobservice" is invalid: spec.strategy.rollingUpdate: Forbidden: may not be specified when strategy `type` is 'Recreate'
```

## Workaround

This problem can be worked around by setting `jobservice.updateStrategy.rollingUpdate=null`

## `helm version`

```shell
version.BuildInfo{Version:"v3.8.1", GitCommit:"5cb9af4b1b271d11d7a97a71df3ac337dd94ad37", GitTreeState:"clean", GoVersion:"go1.17.5"}
```

## `kubectl version`

```shell
Client Version: version.Info{Major:"1", Minor:"22", GitVersion:"v1.22.8", GitCommit:"7061dbbf75f9f82e8ab21f9be7e8ffcaae8e0d44", GitTreeState:"clean", BuildDate:"2022-03-17T03:51:50Z", GoVersion:"go1.16.15", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.2", GitCommit:"802eff1fe87ad2dd737ebbe891f30500b88beb00", GitTreeState:"clean", BuildDate:"2021-11-15T08:35:41Z", GoVersion:"go1.16.5", Compiler:"gc", Platform:"linux/amd64"}
```

miguelaeh:
**Description of the change**

**Benefits**

**Possible drawbacks**

**Applicable issues**

  - fixes #9515

**Additional information**

**Checklist**


lindhe:
I'm working on some feedback. Just putting this comment here to let you know not to merge yet. :)

rafariossaa (approved):
LGTM

lindhe:
@rafariossaa Note that this PR is in conflict with #9533 !

We need do consider both in order to know if this one looks good or not. :)

rafariossaa:
Hi,
  Thanks
  Maybe @miguelaeh wants to take a look to that .

bitnami/harbor/templates/clair/clair-dpl.yaml
```diff
@@ -18,7 +18,13 @@ spec:
...
       app.kubernetes.io/component: clair
   {{- if .Values.clair.updateStrategy }}
-  strategy: {{- toYaml .Values.clair.updateStrategy | nindent 4 }}
+  strategy:
+    type: {{ .Values.clair.updateStrategy.type }}
+    {{- if (eq "Recreate" .Values.clair.updateStrategy.type) }}
+    rollingUpdate: null
+    {{- else }}
+    rollingUpdate: {{ toYaml .Values.clair.updateStrategy.rollingUpdate | nindent 6 }}
+    {{- end }}
```
rafariossaa, line 27: To be compatible with #9533 , I think all these block could be like this:
```suggestion
  strategy:
    type: {{ .Values.clair.updateStrategy.type }}
    {{- if (eq "Recreate" .Values.clair.updateStrategy.type) }}
    rollingUpdate: {{ toYaml .Values.clair.updateStrategy.rollingUpdate | nindent 6 }}
    {{- end }}
```
What do you think ?

miguelaeh: I think it was already compatible with that. When the type is `Recreate` what I added initially will set the `rollingUpdate` to `null`. If we apply your suggestion it won't be compatible

miguelaeh:
@lindhe The alternative PR is not acceptable IMHO. Removing features should never be the solution for a bug.
This PR has that into account and when `Recreate` is set, the `rollingUpdate` is set to `null`, so there shouldn't be issues with that

lindhe:
It's not removing any features, just the documentation for it. Since the entire `*.updateStrategy` object is included in its entirety, users may still set the `*.updateStrategy.rollingUpdate` field (or anything else, for that matter).

And as I mentioned in #9533, we can just set `updateStrategy` fields to `null` instead of removing it. That resolves the bug and does not require any additional logic in the templates. Then we don't remove the documentation either.

bitnami/harbor/templates/clair/clair-dpl.yaml
lindhe: There's literally no reason for this if-statement at all if we just set `rollingUpdate` to `null` in `values.yaml`.

miguelaeh: We must support the `rollingUpdate` configuration.
`null` is not the default value for any type. Setting it to `null` in the `values.yaml` should not be an alternative, nor deleting the parameter.

miguelaeh:
Hi @lindhe ,

https://github.com/bitnami/charts/pull/9533, directly deletes the `rollingUpdate` parameter. We want to support that parameter and we need to set the parameters to the default value for their type. I mean, `""` for strings, `{}` for objects, etc.
Setting it to null IMHO is not a good option and can cause problems when getting schemas from the values.yaml or when performing charts deployiments with other tools.
The if added here has that intention, we continue supporting the `rollingUpdate` and we still make it work with  the `Recreate` strategy

rafariossaa:
Hi,
 @miguelaeh, I agree with you.

SimonAlling:
I personally don't feel completely comfortable with having 10(!) duplicates of the conditional logic added in this PR.

Also, it may be confusing to a user specifying something like

```yaml
updateStrategy:
  type: Recreate
  rollingUpdate: { … }
```

to just have the `rollingUpdate` part silently ignored instead of getting a clear error message prompting removal of that pointless/misleading dead-code part of the configuration.

That being said, I can't (yet) fully comprehend all implications of setting `updateStrategy.rollingUpdate` to `null` (explicitly or not) in `values.yaml`, with respect to schemas etc.

I do agree that the CI-enforced loss of documentation caused by not specifying `updateStrategy.rollingUpdate` _at all_ (in contrast to _explicitly_ setting it to `null`) should be avoided. #9533 should therefore, in my opinion, be updated accordingly so that it does

```diff
-    rollingUpdate: {}
+    rollingUpdate: null
```

instead of

```diff
-    rollingUpdate: {}
```

before a final decision is made on it or this PR.

bitnami/harbor/templates/clair/clair-dpl.yaml
SimonAlling: At the very least, I think the conditional logic should be rewritten to better reflect the Kubernetes API documentation:

> Present only if DeploymentStrategyType = RollingUpdate.

In my opinion, that should be interpreted/expressed in our template logic as

> if the type is `RollingUpdate`, then specify `rollingUpdate`

rather than

> if the type is `Recreate`, then do not specify `rollingUpdate`; otherwise, specify `rollingUpdate`

That should make it more robust with respect to any future changes to the Kubernetes API, and it's closer to what we *mean*™.

So I would flip the logic around and remove the `else` branch for clarity and simplicity:

```suggestion
  strategy:
    type: {{ .Values.clair.updateStrategy.type }}
    {{- if (eq "RollingUpdate" .Values.clair.updateStrategy.type) }}
    rollingUpdate: {{ toYaml .Values.clair.updateStrategy.rollingUpdate | nindent 6 }}
    {{- end }}
```

SimonAlling: > `null` is not the default value for any type.

I think a clarification on this arguably hard-to-falsify claim might be interesting and helpful to understanding the issue at hand. What type space are we talking about, for starters?

* YAML types? Do they even have default values? Please also note that [`null` is a type according to the specification](https://yaml.org/spec/1.2.2/#24-tags), as well as the only value of that type.
* Go types? `null` is not a Go literal, but `nil` is the default (zero) value of all pointer types.
* Some JSON-schema type space? What is the connection between it and our template files?
* Some more abstract type space? How are its default values defined?

miguelaeh:
> o just have the rollingUpdate part silently ignored instead of getting a clear error message prompting removal of that pointless/misleading dead-code part of the configuration.
> That being said, I can't (yet) fully comprehend all implications of setting updateStrategy.rollingUpdate to null (explicitly or not) in values.yaml, with respect to schemas etc.
> I do agree that the CI-enforced loss of documentation caused by not specifying updateStrategy.rollingUpdate at all (in contrast to explicitly setting it to null) should be avoided. https://github.com/bitnami/charts/pull/9533 should therefore, in my opinion, be updated accordingly so that it does

About this, it does not have sense in terms of Kubernetes to specify a `Recreate` strategy type along with a RollingUpdate configuration, so it is perfectly normal to ignore the rolling update configuration if the strategy is set to `Recreate`. https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#strategy
We can add a warning if that's your main concern.

> type UpdateStrategy =
  | {
      type: "Recreate"
      rollingUpdate?: never
    }
  | {
      type: "RollingUpdate"
      rollingUpdate: {
        maxSurge: number | string
        maxUnavailable: number | string
      }
    }

About your second comment, I don catch what you are trying to say. When I am talking about the types I mean the types like object, array, string, etc. `RollingUpdate`/`Recreate` is not a YAML type of course.
The YAML element under `updateStrategy.rollingUpdate` is of type `object`, and the `updateStrategy.type` is of type `string`

PR #9584, opened 2022-03-25
[bitnami/contour] Add behindSslProxy flag

### LINKED ISSUE #9582: [contour] No option to have the envoy service point external https to internal http, which is required for NLB use case
### Name and Version

bitnami/contour 7.4.4

In this article: https://projectcontour.io/guides/deploy-aws-tls-nlb/

The project contour team instructs how to set up contour behind a SSL terminating AWS network load balancer.

However looking at the example it defines an envoy service like this:

```yaml
apiVersion: v1
kind: Service
metadata:
  name: envoy
  namespace: projectcontour
  annotations:
    service.beta.kubernetes.io/aws-load-balancer-type: external
    service.beta.kubernetes.io/aws-load-balancer-nlb-target-type: ip
    service.beta.kubernetes.io/aws-load-balancer-scheme: internet-facing
    service.beta.kubernetes.io/aws-load-balancer-ssl-cert: "arn:aws:acm:us-east-2:185309785115:certificate/7610ed7d-5a81-4ea2-a18a-7ba1606cca3e"
    service.beta.kubernetes.io/aws-load-balancer-ssl-ports: "443"
spec:
  externalTrafficPolicy: Local
  ports:
  - port: 443
    targetPort: 80
    name: http
    protocol: TCP
  selector:
    app: envoy
  type: LoadBalancer
```

You can see they have port 443 pointing to port 80. This is because the NLB is terminating the SSL in this example.

However in the helm chart, there's no option to support this use case. the HTTPS port always points to the HTTPS target port.

There should be an option to have the HTTPS port point to the HTTP target port, in order to support this use case.

_No response_

That there would be an option to support this common use case, and that it wouldn't be absolutely mandatory to have contour handle HTTPS decryption.

### What do you see instead?

There's no option for this. The code looks like this:

```yaml
    - name: https
      port: {{ .Values.envoy.service.ports.https }}
      protocol: TCP
      targetPort: {{ .Values.envoy.containerPorts.https }}
      {{- if and (or (eq .Values.envoy.service.type "NodePort") (eq .Values.envoy.service.type "LoadBalancer")) (not (empty .Values.envoy.service.nodePorts.https)) }}
      nodePort: {{ .Values.envoy.service.nodePorts.https }}
      {{- else if eq .Values.envoy.service.type "ClusterIP" }}
      nodePort: null
```

Which shows that this can't be configured in the way instructed in the article.

### Additional information

_No response_

roberto-civitas:
Fixes #9582 

**Description of the change**

Enables support for SSL termination at the loadbalancer level

**Benefits**

**Possible drawbacks**

**Applicable issues**

  - fixes #9582 

**Additional information**

**Checklist**


alemorcuq (changes requested):
Thanks for the PR! I'm not sure about the `behindSslProxy` parameter, I think it's not obvious what it does from the name. What do you think about adding a `targetPorts` parameter than can be configured? Other charts, like `bitnami/nginx-ingress-controller` use it. Could you take a look?

Also, please check the DCO, increase the chart version and add the new parameter to the README.md. Thanks!

alemorcuq (changes requested):
Please, add the new parameters to the README and increase the chart version in a minor.

paul-civitas:
@alemorcuq how does it look now? Seems all checks now pass.

bitnami/contour/values.yaml
```diff
@@ -667,6 +667,11 @@ envoy:
   ## Envoy Service properties
   ##
   service:
+    ## @param envoy.service.targetPorts [object] Map the controller service HTTP/HTTPS port
+    ##
+    targetPorts:
+      http: http
```
paul-civitas, line 673: Shouldn't these be numerical values?

paul-civitas (changes requested):

paul-civitas:
@roberto-civitas 

Could you check over this again I don't think this is doing what you had in mind. Seems as it it would break existing deployments.

paul-civitas:
Oh I understand now. The "http" and "https" are references to the port definitions in the deployment spec here:

https://github.com/bitnami/charts/blob/master/bitnami/contour/templates/envoy/deployment.yaml#L178

And here:

https://github.com/bitnami/charts/blob/master/bitnami/contour/templates/envoy/deployment.yaml#L187

That's a brilliant solution. It didn't occur to me to do it that way.

bitnami/contour/values.yaml
paul-civitas: Nevermind! See: https://github.com/bitnami/charts/pull/9584#issuecomment-1087872902

alemorcuq (approved):
Thanks!

PR #9586, opened 2022-03-26
[bitnami/postgresql-ha] Update to use projected secrets rather than initContainers for mounting tls certs.

### LINKED ISSUE #9560: [postgresql-ha] Using tls with pgpool results in having to roll the pgpool pods everytime the cert renews
### Name and Version

bitnami/postgresql-ha:8.6.4

Running cert-manager to make pgpool tls certificates means that the tls certificates will get auto-renewed. When the auto-renew occurs, the app will pick up this new tls cert but pgpool will be stuck on the old one.

pgpool:
  tls:
    enabled: true
    

Pgpool should be able to dynamically get an updated tls certificate.

### What do you see instead?

The pgpool container uses an initContainer to copy the tls certificate to a new volume and then mounts this new volume.

Instead, just mount the volume directly. If the concern is the owner and the permissions (as that is the main thing the initContainer script does), then use [projected volumes](https://kubernetes.io/docs/concepts/storage/projected-volumes/)

### Additional information

_No response_

ProNibs:
**Description of the change**

Changes bitnami/postgresql-ha's pgpool from using an initContainer to inject tls secrets to using kubernetes projected volumes. 

**Benefits**

If pgpool.tls.enabled=true, pgpool.tls.autoGenerated=false, and pgpool.tls.certificatesSecret is set (along with certFilename, certKeyFilename, and certCAFilename), if that certificateSecret is updated in place, pgpool will automatically get the new updates without requiring a restart of the pod.

This also locks down the ownership of these files by making them owned by root and the default 1001 user the pgpool process runs on can only read them.

**Possible drawbacks**

I'm not good at testing, so possible that even if the underlying tls ca.crt,tls.cert,tls.key are updated, the pgpool process wouldn't read them on new requests.

**Applicable issues**

  - fixes #9560

**Additional information**

When exec'ing into the container, the files are read+write by owner and read by group. The owner is root and the group is the default 1001 user.
It used to be that the owner was 1001, group was 1001, read+write by owner, read by group, read by other.

**Checklist**


fmulero (commented):
Hi @ProNibs  Thanks a lot for sharing this PR. I think you did a good contribution but  I have some concerns/doubts about the change I'd like to share.

I could reproduce the issue #9560 but unfortunately I don't see how this change could fix it. I can see how the certificates changes in the container but pgpool is not aware about the change so we need to restart the container to work with the new certificates. These are the steps I followed:

1. Create my own certs (in my case, manually with my own CA)
2. Create a secret with my certs
3. Deploy postgresql-ha (`helm install postgresql-ha bitnami/postgresql-ha -f values.yaml`) from your branch with this values.yaml:
```yaml
pgpool:
  tls:
    enabled: true
    autoGenerated: false
    certificatesSecret: fmulero-crt
    certFilename: tls.crt
    certKeyFilename: tls.key
    certCAFilename: ca.crt
```
4. Run following command to check the certificate:
```bash
kubectl exec deploy/postgresql-ha-pgpool -- openssl s_client -starttls postgres -connect localhost:5432 | openssl x509 -text -noout
```
5 . Change the certificate and wait some minutes
6. Run again previous command to check certificate

pgpool is exposing the same certificate that we have at the begining, we need to restart it to take the new ones.

Apart of that I have one doubt. Why don't you change the postgres statefulset? is there any  reason for that?

fmulero:
Thanks a lot @ProNibs. I'd suggest to work first in the issue about having the certificates exposed in pgpool without manual intervention, if we don't find a way to do it maybe is not worth to apply the changes

ProNibs:
Yeah, the PR as it is does the certificates @fmulero. 
In my opinion, might be worth the trouble as it makes the file ownership a little better as well as removes initContainers which "should" increase the startup time for pgpool.

fmulero:
I know this PR would reduce complexity and it would remove the initContainers, but my main concern is that I couldn't confirm that the following statement is true:
> certificateSecret is updated in place, pgpool will automatically get the new updates without requiring a restart of the pod.

In my tests, I needed to restart the pods to take the new certificates. Please correct me if I am wrong

ProNibs:
@fmulero Yeah, I got the same results as you as pgpool won't pick up the new certificates even if the underlying files change 😞 

Sounds like we agree it's still valuable to merge it in for easing the complexity and the security aspects, so guess the question is should we rename the PR? Do I make a new one? Not sure on the process here.

javsalgar:
Hi,

I think it would still be valuable to merge this PR. However, we would like to ensure that no issues are found in major Kubernetes distributions, as some of them may have unexpected compatibility issues with projected volumes (our automated tests should help with that).  In order to get context from the current testing, where did you test the PR?

ProNibs:
Hi,

I just tested them in a kind cluster. However, I'm open to learning how to properly test for other environments.

bitnami/postgresql-ha/templates/pgpool/deployment.yaml
```diff
@@ -63,32 +63,6 @@ spec:
       serviceAccountName: {{ default (include "common.names.fullname" . ) .Values.serviceAccount.name }}
       {{- end }}
       initContainers:
-      {{- if .Values.pgpool.tls.enabled }}
```
fmulero, line 66: The chart is configured to use Kubernetes Security Context to automatically change the ownership of the volumes. However, this feature does not fully work in all Kubernetes distributions or you can face some restrictions when you are using it. Also, our images run as non-root by default, so is very important to have the right permissions.

This piece of code allow users tune the permissions in those scenarios

bitnami/postgresql-ha/templates/pgpool/deployment.yaml
```diff
@@ -403,11 +378,12 @@ spec:
...
-        - name: raw-certificates
-          secret:
-            secretName: {{ include "postgresql-ha.tlsSecretName" . }}
         - name: pgpool-certificates
-          emptyDir: {}
+          projected:
+            defaultMode: 0600
+            sources:
+              - secret:
+                  name: {{ include "postgresql-ha.tlsSecretName" . }}
```
fmulero, line 386: There is no real need to use a projected volume in this case. I think this piece of code has the same behaviour (I did not test):
```suggestion
          secret:
            defaultMode: 0600
            secretName: {{ include "postgresql-ha.tlsSecretName" . }}
```

fmulero (commented):
Sorry @ProNibs about my very late response. I've been revisiting your PR and I have some concerns and I am not fully happy about this PR. Please have a look to my comments

PR #9596, opened 2022-03-28
Add get started with bitnami charts using sealos on bare metal or VM

fanux:
Many users using bitnami on bare metal or VMS sealos is a command line tool to install kubernetes HA cluster.


rafariossaa:
Thanks for your PR.
The links linked into the README points how to get started with the Bitnami charts and the guide you pointed does not document the process to that.

PR #9609, opened 2022-03-29
[bitnami/nginx] possibility to add lifecycle hooks

Adrian43211234:
**Description of the change**

Add the possibility to add lifecycle hooks: https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/
e.g. https://news.ycombinator.com/item?id=30771149  it prevents ingress routes to send requests to the terminating nginx pods which would else result in 502 errors.

**Benefits**
Following snippet provides a 100% uptime during a deployment restart of nginx with 2 replicas. With this configuration, the termination of the nginx pod gets delayed until the ingress controller removed that container as endpoint.
```
lifecycle:
  preStop:
    exec:
      command: ["/bin/sleep", "20"]
```
tested with 

`siege -v -H "Authorization: Bearer XXXXXXX" <URL> -c 2`

**Possible drawbacks**

Implementation of `{{- toYaml .Values.lifecycle | nindent 12 }}` fits best for different use-cases from my point of view but also could end in syntax errors if configured incorrectly.

**Applicable issues**

  - fixes #

**Additional information**

**Checklist**


Adrian43211234:
?

FraPazGal (commented):
Hello @Adrian43211234,

We're sorry about the delay, it seems this PR got lost due to an error in our support workflow. We'll check it in case we missed anything else. Thanks for the patience!

Regarding the change itself, could you please name the parameter as `lifecycleHooks`. This is to maintain the same [template naming](https://github.com/bitnami/charts/blob/4a94de4eafacc50019cc702ccf1c0bb688ec0937/template/CHART_NAME/templates/deployment.yaml#L150) throughout our catalog.

Adrian43211234:
Hmm... I believe I signed the commits, what is the problem now?

bitnami/nginx/templates/deployment.yaml
```diff
@@ -173,6 +173,9 @@ spec:
             - name: https
               containerPort: {{ .Values.containerPorts.https }}
             {{- end }}
+          {{- if .Values.lifecycleHooks }}
+          lifecycle: {{- toYaml .Values.lifecycleHooks | nindent 12 }}
```
FraPazGal, line 177: I saw you mentioned using `toYaml` here and its possible drawbacks. Could you please check if using our default `common.tplvalues.render` function will work better here?

```
{{- define "common.tplvalues.render" -}}
        {{- tpl .value .context }}
    {{- else }}
    {{- end }}
{{- end -}}
```

Referring again to our scaffolding templates, we normally use:

```suggestion
          lifecycle: {{- include "common.tplvalues.render" (dict "value" .Values.lifecycleHooks "context" $) | nindent 12 }}
```

FraPazGal (commented):
Looking at the [action's summary](https://github.com/bitnami/charts/pull/9609/checks?check_run_id=6004063623), it seems it was expecting a different email to be used to sign-off one of the commits. Could it be related to some changes in your GH emails? Nevertheless, seeing as every commit is correctly I'll override the blocker when merging the PR.

I added one last comment to the PR before approving it.

bitnami/nginx/templates/deployment.yaml
Adrian43211234: yes it does work with the default as well

grafik

FraPazGal (approved):
Thanks for the PR, LGTM!

PR #9635, opened 2022-03-31
[bitnami/thanos] Add Thanos Helm Chart tests

superaleks:
**Description of the change**

**Test results**: 

9 consecutive succesfull executions of each service:

- [Query Frontend](https://github.com/superaleks/charts/runs/5864425000?check_suite_focus=true)
- [Bucketweb](https://github.com/superaleks/charts/runs/5864010452?check_suite_focus=true) 

Adding the tests for the Thanos Helm chart. 

**Benefits**

When someone makes a change in the Helm Chart, they will be able to check if their change has broken the Helm chart by running these automated tests.

**Possible drawbacks**

With UI tests, there is always a possibility for flakiness. 

**Checklist**


.github/workflows/vib-thanos-bucketweb.yaml
```diff
@@ -0,0 +1,36 @@
...
+      fail-fast: false
+    name: K8s verify ${{ matrix.target-platform}}
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          ref: ${{github.event.pull_request.head.ref}}
+          repository: ${{github.event.pull_request.head.repo.full_name}}
+      - uses: vmware-labs/vmware-image-builder-action@main
+        with:
+          pipeline: thanos/bucketweb/vib-platform-verify.json
```
joancafom, line 34: I think there should only exist one pipeline definition, as there is no strong reason I can see for having them separated. This way, you would only need a single Kubernetes cluster provisioning 😁

joancafom, line 34:
```suggestion
          pipeline: thanos/vib-platform-verify.json
```

.github/workflows/vib-thanos-queryfrontend.yaml
```diff
@@ -0,0 +1,36 @@
+name: 'vib'
```
joancafom, line 1: Using a single pipeline you may be able to delete this file.

.vib/thanos/bucketweb/cypress/cypress/integration/thanos_bucketweb_spec.js
```diff
@@ -0,0 +1,21 @@
+/// <reference types="cypress" />
+
+it('allows to see runtime & build information', () => {
+    cy.visit('/');
+    cy.contains('.dropdown-toggle', 'Status').click();
+    cy.contains('[class="dropdown-item"]', 'Runtime & Build Information').click();
+    cy.contains('Runtime Information').should('be.visible');
```
joancafom, line 7:
```suggestion
    cy.contains('Runtime Information');
```

Do you need this `should('be.visible');` assertion here?

.vib/thanos/bucketweb/cypress/cypress/integration/thanos_bucketweb_spec.js
```diff
@@ -0,0 +1,21 @@
...
+    cy.get('h2').contains('Build Information').should('be.visible');
```
joancafom, line 8:
```suggestion
    cy.get('h2').contains('Build Information');
```

Do you need this `should('be.visible');` assertion here?

joancafom, line 8:
```suggestion
    cy.contains('h2', 'Build Information').should('be.visible');
```

I think you can chain the selectors here. Ref: [doc](https://docs.cypress.io/api/commands/contains#Specify-a-selector-to-return-a-specific-element)

.vib/thanos/bucketweb/cypress/cypress/integration/thanos_bucketweb_spec.js
```diff
@@ -0,0 +1,21 @@
...
+})
```
joancafom, line 9: You are checking whether the headers appear, but not asserting that the actual information appears (e.g. the actual version is **displayed**). We have faced some issues in the past in which the information was missing and a user filed a new issue to let us know.

Maybe it is worth it adding it. That does not imply that you need to check the version matches (you can do that in GOSS), but at least that some info is rendered and it meets certain regex.

.vib/thanos/bucketweb/cypress/cypress/integration/thanos_bucketweb_spec.js
```diff
@@ -0,0 +1,21 @@
...
+it('allows to see the Blocks page', () => {
+    cy.visit('/');
+    cy.get('.nav-link').contains('Blocks').click();
+    cy.contains('No blocks found.');
+})
+
+it('allows changing of the UI', () => {
+    cy.visit('/');
+    cy.contains('[class="nav-link"]', 'Classic UI').should('be.visible').click();
+    cy.contains('.nav-link', 'New UI').should('exist');
```
joancafom, line 20: Hmm I'm not really sure how Bucketweb UI works behind the scenes, but this might not guarantee to be working on a second run. If the UI is switched to the Classic one and preserved for the second run, this will fail.

.vib/thanos/queryfrontend/cypress/cypress/integration/thanos_queryfrontend_spec.js
```diff
@@ -0,0 +1,43 @@
+/// <reference types="cypress" />
+
+it('allows to see runtime & build information', () => {
+  cy.visit('/');
+  cy.contains('.dropdown-toggle', 'Status').click();
+  cy.contains('[class="dropdown-item"]', 'Runtime & Build Information').click();
+  cy.contains('Runtime Information').should('be.visible');
+  cy.get('h2').contains('Build Information').should('be.visible');
```
joancafom, line 8: Same comment here

.vib/thanos/queryfrontend/cypress/cypress/integration/thanos_queryfrontend_spec.js
```diff
@@ -0,0 +1,43 @@
...
+})
+
+it('allows the execution of a query', () => {
+  cy.visit('/graph');
+  cy.get('.alert').contains('No data queried yet').should('be.visible');
+  cy.get('.cm-line').type('scalar');
+  cy.contains('.cm-completionLabel', 'scalar').click();
```
joancafom, line 15: I think you can use a query that actually returns something, like `vector(3000)`.

.vib/thanos/queryfrontend/cypress/cypress/integration/thanos_queryfrontend_spec.js
```diff
@@ -0,0 +1,43 @@
...
+  cy.visit('/');
+  cy.contains('.dropdown-toggle', 'Status').click();
+  cy.contains('[class="dropdown-item"]', 'Runtime & Build Information').click();
+  cy.contains('Runtime Information').should('be.visible');
+  cy.get('h2').contains('Build Information').should('be.visible');
+})
+
+it('allows the execution of a query', () => {
+  cy.visit('/graph');
+  cy.get('.alert').contains('No data queried yet').should('be.visible');
```
joancafom, line 13:
```suggestion
  cy.contains('.alert', 'No data queried yet');
```
Same comment about chaining and `be.visible`

.vib/thanos/queryfrontend/cypress/cypress/integration/thanos_queryfrontend_spec.js
joancafom: Also, I would use a constant for the query.

.vib/thanos/queryfrontend/cypress/cypress/integration/thanos_queryfrontend_spec.js
```diff
@@ -0,0 +1,43 @@
...
+  cy.contains('.cm-completionLabel', 'scalar').click();
+  cy.get('.execute-btn').click();
+  cy.get('.alert').contains('Empty query result').should('be.visible');
+})
+
+it('allows listing all installed stores', () => {
+  cy.visit('/');
+  cy.contains('[class="nav-link"]', 'Stores').click();
+  cy.get('[data-testid="endpoint"]').should('be.visible');
+  cy.get('[data-testid="health"]').should('be.visible');
```
joancafom, line 24: You should also check as well that the Store is reachable (the status is OK)

.vib/thanos/queryfrontend/cypress/cypress/integration/thanos_queryfrontend_spec.js
```diff
@@ -0,0 +1,43 @@
...
+  cy.visit('/');
+  cy.contains('button', 'Add Panel').should('be.visible').click();
+  cy.get('.execute-btn').should('have.length', 2);
+})
+
+it('can switch from dark to light mode', () => {
+  cy.visit('/');
+  cy.get('[title="Use light theme"]').should('not.be.disabled').click();
+  cy.get('[class="bootstrap-dark"').should('not.exist');
+})
```
joancafom, line 43: IMHO, I don't think this test adds relevant info on whether the asset was properly deployed or not.

.vib/thanos/bucketweb/cypress/cypress/integration/thanos_bucketweb_spec.js
joancafom: Also, I believe the `exist` is not needed here.

.github/workflows/vib-thanos-bucketweb.yaml
```diff
@@ -0,0 +1,36 @@
...
+        env:
+          TARGET_PLATFORM: ${{ matrix.target-platform-id }}
```
FraPazGal, line 36: As explained in https://github.com/bitnami/charts/pull/9637#discussion_r839459981, we should start using `VIB_ENV_TARGET_PLATFORM` as the default env name.

FraPazGal (commented):
I added a minor comment to the existing review. I agree in particular with using a single pipeline. Having 2 of them means using two `M4` clusters at the same time, which seems like a lot of resources for a single deployment on each. Besides that, given the new queue mode implemented in VIB, we'll be taking 2 slots with each Thanos PR. I think we should try to avoid this unless we have no choice.

superaleks (commented):
Trying to cope with testing 2 services and the CP limitation of ports, I've encountered a Cypress issue. I've raised the issue to the [Cypress project](https://github.com/cypress-io/cypress/issues/20915).
I've agreed with the reviewers to keep the pipelines separate. I will resolve all of the comments which refer to this change.

.vib/thanos/queryfrontend/cypress/cypress/integration/thanos_queryfrontend_spec.js
superaleks: Removed this. It was added before I added the second service test. I didn't have the minimum number of tests.

superaleks:
> I added a minor comment to the existing review. I agree in particular with using a single pipeline. Having 2 of them means using two `M4` clusters at the same time, which seems like a lot of resources for a single deployment on each. Besides that, given the new queue mode implemented in VIB, we'll be taking 2 slots with each Thanos PR. I think we should try to avoid this unless we have no choice.

Please take a look at [this comment](https://github.com/bitnami/charts/pull/9635#pullrequestreview-931835614).
Although I agree with you, the pipelines will need to stay separate for the time being.

.vib/thanos/queryfrontend/cypress/cypress/integration/thanos_queryfrontend_spec.js
superaleks: Screenshot 2022-04-05 at 15 32 23
This is checking for the OK icon, which means the test is already covering that.

superaleks: Oh, now I see! 
Fixed it.

.vib/thanos/queryfrontend/cypress/cypress/integration/thanos_queryfrontend_spec.js
```diff
@@ -0,0 +1,36 @@
...
+  cy.contains('.alert', QUERY_ALERT);
+  cy.get('.cm-line').type(QUERY_KEYWORD, '{enter}').type(`(${QUERY_VALUE})`).click();
+  cy.get('.execute-btn').click();
+  cy.contains('.tab-content', QUERY_VALUE);
+})
+
+it('allows listing all installed stores', () => {
+  cy.visit('/');
+  cy.contains('[class="nav-link"]', 'Stores').click();
+  cy.get('[data-testid="endpoint"]').should('be.visible');
```
joancafom, line 27: Is this `be.visible` needed?

.vib/thanos/queryfrontend/cypress/cypress/integration/thanos_queryfrontend_spec.js
```diff
@@ -0,0 +1,36 @@
...
+  cy.get('[data-testid="health"]').contains('UP');
+  cy.get('.badge-success').should('exist');
+})
+
+it('allows adding a graph', () => {
+  cy.visit('/');
+  cy.contains('button', 'Add Panel').should('be.visible').click();
```
joancafom, line 34: Same here about `be.visible`

FraPazGal (dismissed):
LGTM, though I think you missed using the formatter in a couple of minor files (cypress.json or index.js). Please check them if possible.

.vib/thanos/bucketweb/cypress/cypress/integration/thanos_bucketweb_spec.js
superaleks: I've applied additional assertions for the described cases.

joancafom (dismissed):
LGTM, please apply the Formatter to the files before merging 😁

joancafom (dismissed):

FraPazGal (dismissed):
LGTM

joancafom (approved):

PR #9647, opened 2022-03-31
[bitnami/nginx-intel] Remove nginx-ldap-auth-daemon dependency for Bitnami Nginx Intel Chart

mdhont:
Signed-off-by: michield <michield@vmware.com>

Description of the change

Remove nginx-ldap-auth-daemon dependency from Bitnami Nginx-intel Chart.


carrodher (commented):
I would add a note in the Upgrade section describing what is the reason behind a major version bump. Since this is a new chart, probably you will need to create the whole section. You can check https://github.com/bitnami/charts/tree/master/bitnami/nginx#upgrading as example

carrodher (approved):
LGTM

bitnami/nginx-intel/templates/deployment.yaml
```diff
@@ -193,76 +193,21 @@ spec:
           resources: {{- toYaml .Values.resources | nindent 12 }}
           {{- end }}
           volumeMounts:
-            {{- if or .Values.serverBlock .Values.existingServerBlockConfigmap .Values.ldapDaemon.enabled }}
+            {{- if or .Values.serverBlock .Values.existingServerBlockConfigmap }}
             - name: nginx-server-block-paths
               mountPath: /opt/bitnami/nginx/conf/server_blocks
```
joancafom, line 198: The `nginx-server-block-paths` was introduced as a consequence of LDAP addition. This means that it should also be removed and update the logic accordingly.

Refer to https://github.com/bitnami/charts/pull/2608/files#r427295257 for further details.

bitnami/nginx-intel/templates/server-block-configmap.yaml
```diff
@@ -5,7 +5,6 @@ metadata:
   labels: {{- include "common.labels.standard" . | nindent 4 }}
 data:
   server-blocks-paths.conf: |-
-    include  "/opt/bitnami/nginx/conf/server_blocks/ldap/*.conf";
     include  "/opt/bitnami/nginx/conf/server_blocks/common/*.conf";
```
joancafom, line 8: This should be removed

joancafom (changes requested):
Great PR! I added a couple of comments, could you please review them?

joancafom (dismissed):
LGTM!

PR #9661, opened 2022-04-01
[bitnami/cassandra] Enable/Disable automountServiceAccountToken for STS

jmederer:
Signed-off-by: Jan Mederer <jan@mederer.it>

**Description of the change**
Added Paramater in the STS for automountServiceAccountToken. 
This is necessary for restricted environments

**Benefits**

**Possible drawbacks**

**Applicable issues**

  - fixes #

**Additional information**

**Checklist**

### DIFF
--- a/bitnami/cassandra/Chart.yaml
+++ b/bitnami/cassandra/Chart.yaml
@@ -23,4 +23,4 @@ name: cassandra
 sources:
   - https://github.com/bitnami/bitnami-docker-cassandra
   - http://cassandra.apache.org
-version: 9.1.14
+version: 9.2.0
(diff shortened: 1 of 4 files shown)


FraPazGal:
Hello @jmederer,

I have some doubts about the need for a new value here. I may be missing some context but isn't the parameter `serviceAccount.automountServiceAccountToken` enough here? Looking at the chart's templates, we only have the statefulset deploying pods so configuring this option at the serviceAccount level should suffice. 

As an alternative, we could duplicate the use of `serviceAccount.automountServiceAccountToken` to be also in the sts template, WDYT?

jmederer:
Hi @FraPazGal ,
i think we can reject this PR.
In my tests last week i mixed up some logs.
I believe there is no need for a token of the cassandra ServiceAccount. 
We disabled it with the default chart Option - and everything looks fine 

Im sorry for this useless PR.

best regards
Jan

Issue #9278, opened 2022-03-03
[bitnami/argo-cd] Missing default value with "repoServer:containerPorts:metrics" & Confused the metrics port number between server and repoServer


iwanchiyou:
### Name and Version

bitnami/argo-cd 3.0.11

### What steps will reproduce the bug?

```shell
helm pull bitnami/argo-cd --version 3.0.11
tar xf argo-cd-3.0.11.tgz
cd argo-cd
helm install  argocd . --set repoServer.metrics.enabled=true
```

### Are you using any custom parameters or values?

```
--set repoServer.metrics.enabled=true
```

### What is the expected behavior?

```yaml
containerPorts:
  repoServer: 8081
  metrics: ""
```
to
```yaml
containerPorts:
  repoServer: 8081
  metrics: 8084
```
&
```yaml
server:
  metrics:
    service:
      port: 8084
```
to
```yaml
server:
  metrics:
    service:
      port: 8083
```

### What do you see instead?

```
Error: unable to build kubernetes objects from release manifest: error validating "": error validating data: ValidationError(Deployment.spec.template.spec.containers[0].ports[1]): missing required field "containerPort" in io.k8s.api.core.v1.ContainerPort
```

javsalgar:
Hi!

Thank you so much for reporting! Would you like to contribute with a PR fixing the issue?

PR #9662, opened 2022-04-01
[bitnami/argo-cd] Fix issue enabling repoServer.metrics

dgomezleon:
Signed-off-by: David Gomez <dgomezleon@vmware.com>

Fix issue with containerPort when enabling repoServer.metrics #9278


bitnami/argo-cd/Chart.yaml
```diff
@@ -30,4 +30,4 @@ sources:
   - https://github.com/argoproj/argo-cd/
   - https://github.com/bitnami/bitnami-docker-dex
   - https://github.com/dexidp/dex
-version: 3.1.10
+version: 3.2.0
```
joancafom, line 33: I think the new version should only bump the `patch`, as this does not include a new feature but a bugfix.

joancafom (commented):
Thanks for your PR, could you please take a look at my comment?

joancafom (approved):
LGTM!

PR #9679, opened 2022-04-04
[bitnami/zookeeper] customize secrets keys to use SSL with cert-manager

yohboy:

**Description of the change**
This PR mainly adds support for using SSL with cert-manager. It proposes 8 more parameters to customize passwords' and certificates' secrets keys:
- `tls.client.truststorePasswordSecretKey`
- `tls.client.keystorePasswordSecretKey`
- `tls.quorum.truststorePasswordSecretKey`
- `tls.quorum.keystorePasswordSecretKey`
- `tls.client.truststoreSecretKey`
- `tls.client.keystoreSecretKey`
- `tls.quorum.truststoreSecretKey`
- `tls.quorum.keystoreSecretKey`

It also proposes a few enhancements:
- Remove the use of `log4jProp` in `statefulset.yaml`: this parameter does not exist?
- Add `ZOO_TLS_PORT_NUMBER` env var if `containerPorts.tls` is used: fix a bug in case the user customize this port.
- Add `ZOO_TLS_QUORUM_CLIENT_AUTH` and `ZOO_TLS_CLIENT_AUTH` env var, respectively depending on `tls.quorum.auth` and `tls.client.auth`. Can be set to "none", "want" and "need" (default to "need"). Also add a check on these parameters (value must be in the list).

**Benefits**
This chart can now use cert-manager to issue secrets for SSL management.  

**Possible drawbacks**
None.

**Applicable issues**
None.

**Additional information**
None.

**Checklist**


bitnami/zookeeper/templates/_helpers.tpl
```diff
@@ -261,3 +265,55 @@ zookeeper: tls.quorum.enabled
...
+*/}}
+{{- define "zookeeper.validateValues.quorum.tls.auth" -}}
+{{- if and .Values.tls.quorum.enabled .Values.tls.quorum.auth }}
+{{- $tlsAuths := list "none" "want" "need" -}}
+{{- if not (has .Values.tls.quorum.auth $tlsAuths) }}
+zookeeper: tls.quorum.auth
+    tls.quorum.auth must be one of these values: "none", "want" or "need".
+{{- end -}}
+{{- end -}}
+{{- end -}}
```
joancafom, line 280: This is sweet, but we are already validating it in the docker image 😁 

https://github.com/bitnami/bitnami-docker-zookeeper/blob/208bb77b98f32d461637f853c6e18f6d7dd04677/3.8/debian-10/rootfs/opt/bitnami/scripts/libzookeeper.sh#L103-L104

I think that we can safely remove this provided that we have a note in the `values.yaml` file about the possible values.

bitnami/zookeeper/templates/secrets.yaml
```diff
@@ -33,8 +33,8 @@ metadata:
   {{- end }}
 type: Opaque
 data:
-  keystore-password: {{ default (randAlphaNum 10) .Values.tls.client.keystorePassword | b64enc | quote }}
-  truststore-password: {{ default (randAlphaNum 10) .Values.tls.client.truststorePassword | b64enc | quote }}
+{{ .Values.tls.client.keystorePasswordSecretKey | indent 2 -}}: {{ default (randAlphaNum 10) .Values.tls.client.keystorePassword | b64enc | quote }}
+{{ .Values.tls.client.truststorePasswordSecretKey | indent 2 -}}: {{ default (randAlphaNum 10) .Values.tls.client.truststorePassword | b64enc | quote }}
```
joancafom, line 37: I would restrict the `truststorePasswordSecretKey`/`existingSecretTruststoreKey` exclusively to the name of the key to be used in case an existingSecret is provided. Else, the default values are `keystore-password` and `truststore-password` (in other words, leave this file unchanged).

bitnami/zookeeper/templates/scripts-configmap.yaml
```diff
@@ -37,9 +37,9 @@ data:
         exit 1
     fi
     {{- else }}
-    elif [[ -f "/certs/client/zookeeper.truststore.jks" ]] && [[ -f "/certs/client/zookeeper.keystore.jks" ]]; then
-        cp "/certs/client/zookeeper.truststore.jks" "/opt/bitnami/zookeeper/config/certs/client/zookeeper.truststore.jks"
-        cp "/certs/client/zookeeper.keystore.jks" "/opt/bitnami/zookeeper/config/certs/client/zookeeper.keystore.jks"
+    elif [[ -f {{ printf "/certs/client/%s" .Values.tls.client.truststoreSecretKey | quote }} ]] && [[ -f {{ printf "/certs/client/%s" .Values.tls.client.keystoreSecretKey | quote }} ]]; then
+        cp {{ printf "/certs/client/%s" .Values.tls.client.truststoreSecretKey | quote }} "/opt/bitnami/zookeeper/config/certs/client/zookeeper.truststore.jks"
+        cp {{ printf "/certs/client/%s" .Values.tls.client.keystoreSecretKey | quote }} "/opt/bitnami/zookeeper/config/certs/client/zookeeper.keystore.jks"
```
joancafom, line 42: This is nice, but I would use a common approach found in other charts, which is to define a new helper function that will return the right name:
https://github.com/bitnami/charts/blob/c118f228f5c932f789a31782d992986b64508f70/bitnami/redis/templates/_helpers.tpl#L172-L181

joancafom (changes requested):
Thank you so much for your PR! Could you please take a look at my comments?

bitnami/zookeeper/templates/_helpers.tpl
yohboy: Yes, I already saw this check in the container. The only difference by checking it in helm too is that the chart will fail at applying instead of failing when the container will try to boot. Thus, we can detect the misconfiguration sooner. But yes, the result is the same, an error will be caught. I can remove it if it is redundant :)

bitnami/zookeeper/templates/secrets.yaml
yohboy: Ok I fixed it and I added some helper functions to determine if we need to use `truststorePasswordSecretKey`/`existingSecretTruststoreKey` or `keystore-password`/`truststore-password`.

bitnami/zookeeper/values.yaml
```diff
@@ -750,13 +750,22 @@ tls:
...
     autoGenerated: false
     ## @param tls.client.existingSecret Name of the existing secret containing the TLS certificates for ZooKeeper client communications
     ##
     existingSecret: ""
+    ## @param tls.client.keystoreSecretKey The secret key from the existingSecret if 'keystoreSecretKey' key different from the default (zookeeper.keystore.jks)
+    ##
+    keystoreSecretKey: "zookeeper.keystore.jks"
+    ## @param tls.client.truststoreSecretKey The secret key from the existingSecret if 'truststoreSecretKey' key different from the default (zookeeper.truststore.jks)
+    ##
+    truststoreSecretKey: "zookeeper.truststore.jks"
```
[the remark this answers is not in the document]
yohboy: Fixed!

bitnami/zookeeper/values.yaml
```diff
@@ -750,13 +750,22 @@ tls:
...
     ## @param tls.client.autoGenerated Generate automatically self-signed TLS certificates for ZooKeeper client communications
     ## Currently only supports PEM certificates
     ##
     autoGenerated: false
     ## @param tls.client.existingSecret Name of the existing secret containing the TLS certificates for ZooKeeper client communications
     ##
     existingSecret: ""
+    ## @param tls.client.keystoreSecretKey The secret key from the existingSecret if 'keystoreSecretKey' key different from the default (zookeeper.keystore.jks)
+    ##
+    keystoreSecretKey: "zookeeper.keystore.jks"
```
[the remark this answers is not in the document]
yohboy: Fixed!

yohboy:
Hi @joancafom,
Thank you very much for your review!
I began to fix some points, but I would need some clarifications for other points! I let you some comments.
Waiting for your feedback before continuing :)

bitnami/zookeeper/templates/_helpers.tpl
joancafom: To me, we are safe by documenting the possible values in the `values.yaml`/`README.md` and performing the check in the container. After all, the container will show the validation error when trying to boot, so this validation is kind of redundant IMHO :)

bitnami/zookeeper/values.yaml
```diff
@@ -766,6 +775,12 @@ tls:
     ## @param tls.client.passwordsSecretName Existing secret containing Keystore and truststore passwords
     ##
     passwordsSecretName: ""
+    ## @param tls.client.keystorePasswordSecretKey The secret key from the passwordsSecretName if 'keystorePasswordSecretKey' key different from the default (keystore-password)
+    ##
+    keystorePasswordSecretKey: "keystore-password"
+    ## @param tls.client.truststorePasswordSecretKey The secret key from the passwordsSecretName if 'truststorePasswordSecretKey' key different from the default (truststore-password)
+    ##
+    truststorePasswordSecretKey: "truststore-password"
```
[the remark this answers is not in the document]
joancafom: >By the way, note that having the possibility to set both keystore-password and truststore-password for zookeeper can lead to a mismatch error because zookeeper REQUIRES the same keystore and truststore password. But this is another debate.

I am not really sure about this, in fact, I believe that this chart is setting two different random passwords if the TLS configuration is autogenerated instead of provided:

https://github.com/bitnami/charts/blob/7fb947bbd62c1e5d4a91b0378d65c66cf0c672c0/bitnami/zookeeper/templates/secrets.yaml#L55-L56

> I think this approach would be true if cert-manager allows a different password for keystore and for truststore. But to the best of my knowledge, we can configure only one secret through passwordSecretRef with only one key to set both passwords, so we have to share the same password

You are right! But then, I believe this is really a Cert-Manager limitation 😁 . One should follow the "hacky" approach   because Cert-Manager itself does not allow setting two different encryption passwords. 

Having said that, I believe that even if it a limitation of Cert-Manager itself, having the ability to configure the keys for the provided secret may come handy (plus, would allow to skip the "hacky approach - though you'd still need to point **to the same key** if you are using Cert-Manager-). However, I would rename these two to something more tightly related to `passwordsSecretName` (just as we did with `existingSecret):

- `keystorePasswordSecretKey` --> `passwordsSecretKeystoreKey`
- `truststorePasswordSecretKey` --> `passwordsSecretTruststoreKey`

WDYT?

joancafom:
Hi @yohboy 

Thank you very much for your feedback, I've replied to the threads with some more thoughts 😁

bitnami/zookeeper/templates/_helpers.tpl
yohboy: yes, it makes sense! I removed the check

bitnami/zookeeper/values.yaml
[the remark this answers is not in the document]
yohboy: > I am not really sure about this, in fact, I believe that this chart is setting two different random passwords if the TLS configuration is autogenerated instead of provided:

You are right, I read the documentation too fast, sorry. Zookeeper allows different password between **keystore** and **truststore** but not between **keystore** and **certificate key**.

> You are right! But then, I believe this is really a Cert-Manager limitation :grin: . One should follow the "hacky" approach because Cert-Manager itself does not allow setting two different encryption passwords.

I agree with this point too, this is a limitation from cert-manager currently...

But I think having the ability to configure the keys for the secret can be useful as you said (and we do not have to maintain a secret with 2 same keys... but yes, we need to point to the same key). I updated the parameters accordingly!

bitnami/zookeeper/templates/scripts-configmap.yaml
yohboy: I checked to define a new helper function for this, but I do not find the point of a helper function here. It will return always the same value? Moreover, I will need 4 helper functions. I think it will just complexify. I would skip this one. What do you think?

yohboy:
Hi @joancafom,

Thank you for your feedback again!
I updated the PR accordingly and I left you one more thought :)

bitnami/zookeeper/templates/scripts-configmap.yaml
joancafom: > but I do not find the point of a helper function here. It will return always the same value?

The thing here is that the keystore file name (that is, the key in the secret) might not be the default one expected by the chart. When an `existingSecret` is not provided, the keystore is named `zookeeper.truststore.jks`. However, if you use Cert-Manager, it will be `truststore.jks`. Maybe I misunderstood or made a wrong statement, but you can see that they might have different values.

By the way (maybe the confusion comes from this), the default value I expect for `existingSecretKeystoreKey` is `""` and NOT `"zookeeper.truststore.jks"` (which is the chart's default value). To me, this parameter conceptually exclusively models the name of the KeystoreKey (a.k.a. the keystore file) in the `existingSecret`.

joancafom: > but I do not find the point of a helper function here. It will return always the same value?

The thing here is that the keystore file name (that is, the key in the secret) might not be the default one expected by the chart. When an `existingSecret` is not provided, the keystore is named `zookeeper.truststore.jks`. However, if you use Cert-Manager, it will be `truststore.jks`. Maybe I misunderstood or made a wrong statement, but you can see that they might have different values.

By the way (maybe the confusion comes from this), the default value I expect for `existingSecretKeystoreKey` is `""` and NOT `"zookeeper.truststore.jks"` (which is the chart's default value). To me, this parameter conceptually exclusively models the name of the KeystoreKey (a.k.a. the keystore file) in the `existingSecret`.

joancafom:
Hi @yohboy 

Replied :)

bitnami/zookeeper/templates/scripts-configmap.yaml
yohboy: You were right about the confusion :D  
I updated both the default values and the functions helper!

yohboy:
Hi @joancafom,

Thanks again for your feedback!
I fixed all threads!
Let me know if you notice anything else :)

bitnami/zookeeper/values.yaml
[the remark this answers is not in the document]
joancafom: Thanks!

bitnami/zookeeper/templates/_helpers.tpl
```diff
@@ -190,6 +190,50 @@ Return the secret containing ZooKeeper client TLS certificates
 {{- end -}}
 {{- end -}}
 
+{{/*
+Get the quorum keystore key to be retrieved from tls.quorum.existingSecret.
+*/}}
+{{- define "zookeeper.quorum.keystoreKey" -}}
```
joancafom, line 196:
```suggestion
{{- define "zookeeper.quorum.tlsKeystoreKey" -}}
```
I think this is more descriptive.

bitnami/zookeeper/templates/_helpers.tpl
```diff
@@ -190,6 +190,50 @@ Return the secret containing ZooKeeper client TLS certificates
...
+    {{- printf "%s" .Values.tls.quorum.existingSecretKeystoreKey -}}
+{{- else -}}
+    {{- printf "zookeeper.keystore.jks" -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
+Get the quorum truststore key to be retrieved from tls.quorum.existingSecret.
+*/}}
+{{- define "zookeeper.quorum.truststoreKey" -}}
```
joancafom, line 207:
```suggestion
{{- define "zookeeper.quorum.tlsTruststoreKey" -}}
```

bitnami/zookeeper/templates/_helpers.tpl
```diff
@@ -190,6 +190,50 @@ Return the secret containing ZooKeeper client TLS certificates
...
+    {{- printf "%s" .Values.tls.quorum.existingSecretTruststoreKey -}}
+{{- else -}}
+    {{- printf "zookeeper.truststore.jks" -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
+Get the client keystore key to be retrieved from tls.client.existingSecret.
+*/}}
+{{- define "zookeeper.client.keystoreKey" -}}
```
joancafom, line 218:
```suggestion
{{- define "zookeeper.client.tlsKeystoreKey" -}}
```

bitnami/zookeeper/templates/_helpers.tpl
```diff
@@ -190,6 +190,50 @@ Return the secret containing ZooKeeper client TLS certificates
...
+    {{- printf "%s" .Values.tls.client.existingSecretKeystoreKey -}}
+{{- else -}}
+    {{- printf "zookeeper.keystore.jks" -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
+Get the client truststore key to be retrieved from tls.client.existingSecret.
+*/}}
+{{- define "zookeeper.client.truststoreKey" -}}
```
joancafom, line 229:
```suggestion
{{- define "zookeeper.client.tlsTruststoreKey" -}}
```

bitnami/zookeeper/templates/_helpers.tpl
```diff
@@ -211,6 +255,50 @@ Return the name of the secret containing the Keystore and Truststore password
 {{- end -}}
 {{- end -}}
 
+{{/*
+Get the quorum keystore password key to be retrieved from tls.quorum.passwordSecretName.
+*/}}
+{{- define "zookeeper.quorum.passwordKeystoreKey" -}}
```
joancafom, line 261:
```suggestion
{{- define "zookeeper.quorum.tlsPasswordKeystoreKey" -}}
```

bitnami/zookeeper/templates/_helpers.tpl
```diff
@@ -211,6 +255,50 @@ Return the name of the secret containing the Keystore and Truststore password
...
+    {{- printf "%s" .Values.tls.quorum.passwordsSecretKeystoreKey -}}
+{{- else -}}
+    {{- printf "keystore-password" -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
+Get the quorum truststore password key to be retrieved from tls.quorum.passwordSecretName.
+*/}}
+{{- define "zookeeper.quorum.passwordTruststoreKey" -}}
```
joancafom, line 272:
```suggestion
{{- define "zookeeper.quorum.tlsPasswordTruststoreKey" -}}
```

bitnami/zookeeper/templates/_helpers.tpl
```diff
@@ -211,6 +255,50 @@ Return the name of the secret containing the Keystore and Truststore password
...
+    {{- printf "%s" .Values.tls.quorum.passwordsSecretTruststoreKey -}}
+{{- else -}}
+    {{- printf "truststore-password" -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
+Get the client keystore password key to be retrieved from tls.client.passwordSecretName.
+*/}}
+{{- define "zookeeper.client.passwordKeystoreKey" -}}
```
joancafom, line 283:
```suggestion
{{- define "zookeeper.client.tlsPasswordKeystoreKey" -}}
```

bitnami/zookeeper/templates/_helpers.tpl
```diff
@@ -211,6 +255,50 @@ Return the name of the secret containing the Keystore and Truststore password
...
+    {{- printf "%s" .Values.tls.client.passwordsSecretKeystoreKey -}}
+{{- else -}}
+    {{- printf "keystore-password" -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
+Get the client truststore password key to be retrieved from tls.client.passwordSecretName.
+*/}}
+{{- define "zookeeper.client.passwordTruststoreKey" -}}
```
joancafom, line 294:
```suggestion
{{- define "zookeeper.client.tlsPasswordTruststoreKey" -}}
```

joancafom (changes requested):
I added some more minor comments about naming. Once addressed, I think that should be it. Thanks again! 😁

yohboy:
Hi @joancafom,
I applied your suggestions and did some updates according to these changes.
I had a hard time with the DCO and the batch suggestions commit from GitHub (no signoff added).
All should be good now :)
Thanks again for your time!

joancafom (approved):
LGTM! Thank you so much for your contribution and efforts on this!! 🚀

yohboy:
Hi @joancafom, great news! Thank you very much for your patience and help :)

PR #9724, opened 2022-04-07
[bitnami/rabbitmq] reuse secret data during upgrade with look up

VannTen:
**Description of the change**

In the `rabbitmq/templates/secrets.yaml` use the lookup function to retrieve the
value of `rabbitmq-password` and `rabbitmq-erlang-cookie`.
If it does not exists, generate a random one.
If the user explicitely speficy those keys, it will override the current value.

**Benefits**

Avoid explicitely specifying the password and cookie when upgrading.

**Possible drawbacks**

- N/A

**Applicable issues**

- N/A

**Additional information**

**Checklist**


bitnami/rabbitmq/templates/secrets.yaml
```diff
@@ -1,4 +1,18 @@
...
+  {{- $rabbitmq_password := index $secret.data "rabbitmq-password" }}
+  {{- $rabbitmq_erlangCookie := index $secret.data "rabbitmq-erlang-cookie" }}
+{{- end -}}
+{{- if .Values.auth.password }}
+   {{ $rabbitmq_password := .Values.auth.password }}
+{{- end -}}
+{{- if .Values.auth.erlangCookie }}
+   {{ $rabbitmq_erlangCookie := .Values.auth.erlangCookie }}
+{{- end -}}
+
```
joancafom, line 15: We actually have a common method implemented for this same purpose:

https://github.com/bitnami/charts/blob/14d91e2bed2e4328434c45f301628987f53e43ec/bitnami/common/templates/_secrets.tpl#L85

You can take a look at this PR, which implements a similar change:
https://github.com/bitnami/charts/pull/9300/files

joancafom (changes requested):
Thanks for much for your PR! Could you please take a look at my comment?

bitnami/rabbitmq/templates/secrets.yaml
VannTen: Thansk for the tip, I'm gonna take a look. That ought to make this PR much simpler :+1:

VannTen: So I rewrited the PR to basically do the same thing that the linked one, opinions ?

joancafom: Great!! It is much simpler now, thanks 😁

bitnami/rabbitmq/templates/secrets.yaml
```diff
@@ -1,29 +1,18 @@
 {{- if or (not .Values.auth.existingErlangSecret) (not .Values.auth.existingPasswordSecret) }}
+{{- $secretName := include "common.secrets.name" (dict "existingSecret" .Values.auth.existingSecret "context" $)}}
 apiVersion: v1
 kind: Secret
 metadata:
-  name: {{ include "rabbitmq.fullname" . }}
+  name: {{ $secretName }}
```
joancafom, line 6: This is tricky here, as the change is not exactly the same as the PR I shared with you. Bear in mind, that this secret will be created **if either** `existingErlangSecret` or `existingPasswordSecret` are empty.

This means, that if you **only provide one** existing secret, say `existingErlangSecret`, the installation will fail as Helm will try to create a new Secret that will be named exactly like the one that already exists, which will lead to failure. 

```bash
$ kubectl create secret generic abc --from-literal rabbitmq-password=ABCDEFG12345 --from-literal rabbitmq-erlang-cookie=ABCDEFG12345
secret/abc created

$ helm install delta bitnami/rabbitmq --set auth.existingSecret=abc
Error: rendered manifests contain a resource that already exists. Unable to continue with install: Secret "abc" in namespace "default" exists and cannot be imported into the current release: invalid ownership metadata; label validation error: missing key "app.kubernetes.io/managed-by": must be set to "Helm"; annotation validation error: missing key "meta.helm.sh/release-name": must be set to "delta"; annotation validation error: missing key "meta.helm.sh/release-namespace": must be set to "default"
```

This case is more similar to Odoo:

https://github.com/bitnami/charts/blob/master/bitnami/odoo/templates/secrets.yaml
```suggestion
  name: {{ include "rabbitmq.fullname" . }}
```

bitnami/rabbitmq/templates/secrets.yaml
```diff
@@ -1,29 +1,18 @@
...
-  {{- end }}
-  {{- end }}
-  {{- if not .Values.auth.existingErlangSecret }}
-  {{- if .Values.auth.erlangCookie }}
-  rabbitmq-erlang-cookie: {{ .Values.auth.erlangCookie | b64enc | quote }}
-  {{- else }}
-  rabbitmq-erlang-cookie: {{ randAlphaNum 32 | b64enc | quote }}
-  {{- end }}
-  {{- end }}
+  rabbitmq-password: {{ include "common.secrets.passwords.manage" (dict "secret" $secretName "key" "rabbitmq-password" "length" 16 "providedValues" (list "auth.password") "context" $) }}
```
joancafom, line 14: Again, you will need to follow the same approach used in Odoo:

https://github.com/bitnami/charts/blob/28661447b627cc8d3d25e7e1ee79fec21d687cff/bitnami/odoo/templates/secrets.yaml#L16-L18

joancafom: Also, note that the old length was `10` and not `16`.

bitnami/rabbitmq/templates/secrets.yaml
```diff
@@ -1,29 +1,18 @@
...
+  rabbitmq-erlang-cookie: {{ include "common.secrets.passwords.manage" (dict "secret" $secretName "key" "rabbitmq-erlang-cookie" "length" 32 "providedValues" (list "auth.password") "context" $) }}
```
joancafom, line 15: The same comment applies here, I think you got the idea :)

joancafom (changes requested):
Good job on integrating changes! I left you some more comments, could you please take a look at them?

bitnami/rabbitmq/templates/secrets.yaml
VannTen: > Also, note that the old length was `10` and not `16`.

That last bit was intentional. A 10-char password struck me as odd.

joancafom (changes requested):
Thanks for implementing the changes. I left you a final minor comment and I think we will be ready to test & merge this!

Also, please remember to resolve the conflicts in the `Chart.yaml` file.

bitnami/rabbitmq/templates/secrets.yaml
```diff
@@ -13,20 +13,12 @@ metadata:
...
-  rabbitmq-erlang-cookie: {{ .Values.auth.erlangCookie | b64enc | quote }}
-  {{- else }}
-  rabbitmq-erlang-cookie: {{ randAlphaNum 32 | b64enc | quote }}
-  {{- end }}
-  {{- end }}
+  {{- if (not .Values.auth.existingPasswordSecret ) }}
+  rabbitmq-password: {{ include "common.secrets.passwords.manage" (dict "secret" (include "common.names.fullname" .) "key" "rabbitmq-password" "length" 16 "providedValues" (list "auth.password") "context" $) }}
+  {{ end }}
+  {{- if (not .Values.auth.existingErlangSecret ) }}
+  rabbitmq-erlang-cookie: {{ include "common.secrets.passwords.manage" (dict "secret" (include "common.names.fullname" .) "key" "rabbitmq-erlang-cookie" "length" 32 "providedValues" (list "auth.password") "context" $) }}
```
joancafom, line 20:
```suggestion
  rabbitmq-erlang-cookie: {{ include "common.secrets.passwords.manage" (dict "secret" (include "common.names.fullname" .) "key" "rabbitmq-erlang-cookie" "length" 32 "providedValues" (list "auth.erlangCookie") "context" $) }}
```

joancafom (dismissed):
LGTM! Thanks for your changes! 🚀

PR #9754, opened 2022-04-11
[bitnami/wordpress] Remove wordpressAutoUpdateLevel parameter in the WordPress chart

jotamartos:
**Description of the change**

This PR removes the wordpressAutoUpdateLevel parameter. If you enable this feature, WordPress will be updated to the latest minor/major version but if you recreate the pod for any reason, the app will come back to the previous version and that may introduce errors. 

**Benefits**

The solution won't be automatically updated so no errors will appear if that happens and the user recreate the pod for any reason

**Possible drawbacks**

This parameter allows the user keep the latest version of the solution automatically

**Checklist**


bitnami/wordpress/README.md
```diff
@@ -499,6 +498,10 @@ Find more information about how to deal with common errors related to Bitnami's
 
 ## Notable changes
 
+### 13.2.0
+
+- Removed support for limiting auto-updates to WordPress core via the `wordpressAutoUpdateLevel` option. To update WordPress core, we recommend to swap the container image version for your deployment instead of using the built-in update functionality.
```
carrodher, line 503: More than swapping the container image, for upgrading WordPress the recommendation should be to upgrade the Helm chart itself because it is possible changes in the application involve changes in the Helm chart.

Apart from that, please note we are not using lists for the notable change, the initial `-` can be safely removed

jotamartos: Thanks for the suggestion. I just applied the changes.

carrodher (dismissed):
LGTM

fmulero (dismissed):

PR #9770, opened 2022-04-12
[bitnami/ghost] Add Ghost Helm Chart tests

joancafom:
**Description of the change**

**Test results**

9 consecutive executions without issues: [Batch #1](https://github.com/joancafom/charts/runs/5980673253?check_suite_focus=true), [Batch #2](https://github.com/joancafom/charts/runs/5985970141?check_suite_focus=true) & [Batch #3](https://github.com/joancafom/charts/runs/5978684886?check_suite_focus=true).

Samples:

- [GKE](https://cp.production.k.bromelia.bitnami.net/v1/execution-graphs/606ebebf-15d3-4007-94aa-68aea2f1efa9/report) - PASS
- [AKS](https://cp.production.k.bromelia.bitnami.net/v1/execution-graphs/11f53449-f1ff-4c33-be6d-07531e97d7cb/report) - PASS
- [TKG](https://cp.production.k.bromelia.bitnami.net/v1/execution-graphs/0d833092-3a2a-4408-a8f6-4bee94bf59d0/report) - PASS

Adding the tests for the Ghost Helm chart. 

**Benefits**

When someone makes a change in the Helm Chart, they will be able to check if their change has broken the Helm chart by running these automated tests.

**Possible drawbacks**

With UI tests, there is always a possibility for flakiness. 

**Checklist**


.github/workflows/vib-ghost.yaml
```diff
@@ -0,0 +1,36 @@
+name: 'vib'
+on: # rebuild any PRs and main branch changes
```
superaleks, line 2: I already have the PR ready for fixing inconsistencies. Ghost will **not** be included, so there is a need to make the inconsistency related changes here.

.github/workflows/vib-ghost.yaml
```diff
@@ -0,0 +1,36 @@
...
+  CSP_API_TOKEN: ${{ secrets.CSP_API_TOKEN }}
+  VIB_PUBLIC_URL: https://cp.bromelia.vmware.com
+jobs:
+  vib-k8s-verify: # verify in multiple target platforms
+    runs-on: ubuntu-latest
+    environment: vmware-image-builder
+    strategy:
+      matrix:
+       include:
+         - target-platform: gke
```
superaleks, line 19:
```suggestion
         - target-platform: Tanzu Kubernetes Grid
```

.github/workflows/vib-ghost.yaml
```diff
@@ -0,0 +1,36 @@
...
+           target-platform-id: 91d398a2-25c4-4cda-8732-75a3cfc179a1
+         - target-platform: aks
+           target-platform-id: 7b13a7bb-011c-474f-ad71-8152fc321b9e
+         - target-platform: tkg
+           target-platform-id: 7ddab896-2e4e-4d58-a501-f79897eba3a0
+      fail-fast: false
+    name: K8s verify ${{ matrix.target-platform}}
```
superaleks, line 26:
```suggestion
    name: Ghost on ${{ matrix.target-platform }}
```

.github/workflows/vib-ghost.yaml
```diff
@@ -0,0 +1,36 @@
...
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          ref: ${{github.event.pull_request.head.ref}}
+          repository: ${{github.event.pull_request.head.repo.full_name}}
```
superaleks, line 31:
```suggestion
          repository: ${{ github.event.pull_request.head.repo.full_name }}
```

.vib/ghost/cypress/cypress/support/commands.js
```diff
@@ -0,0 +1,53 @@
...
+    const origVal = originalFn(...args);
+
+    return new Promise((resolve) => {
+      setTimeout(() => {
+        resolve(origVal);
+      }, COMMAND_DELAY);
+    });
+  });
+}
+
```
superaleks, line 15: I appreciate the necessity to explain various corner case behaviours, but I'd prefer to keep this data documented internally instead of having them in the comments. Let's create a task for this in our Helm chart strategy and brainstorm different ways of documenting this. 
The cognitive complexity of the code itself is already there for the reader, it is quite difficult to cope with if we add a lot of comments instead.

.vib/ghost/cypress/cypress/integration/ghost_spec.js
```diff
@@ -0,0 +1,123 @@
+/// <reference types="cypress" />
+import { random, getPageUrlFromTitle, getUserFromEmail } from './utils';
+
+it('allows to log in and out', () => {
+  cy.login(Cypress.env('email'), Cypress.env('password'), false);
```
superaleks, line 5: Could we always have this set the 'waitForDashboard' to `true`?
We are waiting for it in the login test and in the `commands.js` default settings. 
Better said, In which case will we will **not** wait for the dashboard?

.vib/ghost/cypress/cypress/support/commands.js
```diff
@@ -0,0 +1,53 @@
...
+  (
+    email = Cypress.env('email'),
+    password = Cypress.env('password'),
+    waitForDashboard = true
+  ) => {
+    cy.visit('/ghost/');
+    cy.get('input[type="email"]').should('be.enabled').type(email);
+    cy.get('input[type="password"]').should('be.enabled').type(password);
+    cy.contains('button', 'Sign in').click();
+    if (waitForDashboard) {
```
superaleks, line 36: Same question as in the comment above, can you clarify why we need the `waitForDashboard`?

.vib/ghost/cypress/cypress.json
```diff
@@ -0,0 +1,10 @@
+{
+    "env": {
+      "email": "user@example.com",
+      "password": "Complicated123!4"
+    },
+    "hosts": {
+      "bitnami-ghost.my": "{{ TARGET_IP }}"
```
superaleks, line 7: Very nice!

.vib/ghost/cypress/cypress/integration/ghost_spec.js
```diff
@@ -0,0 +1,123 @@
...
+  cy.contains('Retry').should('not.exist');
+  // Dashboard is hidden in the UI until correctly logged in
+  cy.contains('Dashboard').should('be.visible');
+
+  cy.get("div[class*='avatar']").click();
+  cy.contains('Sign out').click();
+  cy.contains('Sign in');
+});
+
+it('allows to create a new post', () => {
```
superaleks, line 15:
```suggestion
it('allows to create and publish a new post', () => {
```

.vib/ghost/goss/goss.yaml
```diff
@@ -0,0 +1,26 @@
...
+    exec: which ghost
+    exit-status: 0
+    stdout:
+      - "/opt/bitnami/ghost/bin/ghost"
+    stderr: [ ]
+    timeout: 40000
+  ghost-version-ok:
+    exec: ghost version | grep --quiet "$APP_VERSION"
+    exit-status: 0
+    stdout: [ ]
```
superaleks, line 12: Why do we need the `stdout` here, if we are not verifying the output?

.vib/ghost/cypress/cypress/support/index.js
```diff
@@ -0,0 +1,20 @@
...
+// You can change the location of this file or turn off
+// automatically serving support files with the
+// 'supportFile' configuration option.
+//
+// You can read more here:
+// https://on.cypress.io/configuration
+// ***********************************************************
+
+// Import commands.js using ES2015 syntax:
+import './commands'
```
FraPazGal, line 17:
```suggestion
import './commands';
```

.vib/ghost/cypress/cypress/integration/ghost_spec.js
```diff
@@ -0,0 +1,123 @@
...
+  cy.login();
+  cy.visit('/ghost/#/members/import');
+
+  // The element exists in the UI while loading, but we need to
+  // wait until it is visible to proceed
+  cy.contains('h1', 'Import members').should('be.visible');
+  cy.get('input[type="file"]').selectFile('cypress/fixtures/members.csv', {
+    force: true,
+  });
+  cy.contains('button', 'Import').click();
```
FraPazGal, line 89: We are working with non-random data. Let's say the test fails checking the added members because temporarily the page doesn't load. Could we encounter an issue here if we need to retry the test?

FraPazGal (commented):
The PR looks good! Could you please check my comments?

.github/workflows/vib-ghost.yaml
joancafom: This does no longer apply after https://github.com/bitnami/charts/pull/9780. Marking this as resolved.

.vib/ghost/cypress/cypress.json
joancafom: Thanks!

.vib/ghost/cypress/cypress/integration/ghost_spec.js
```diff
@@ -0,0 +1,123 @@
...
+    cy.visit(getPageUrlFromTitle(`${$pages.newPage.title}-${random}`));
+    cy.contains(`${$pages.newPage.title}-${random}`);
+    cy.contains(`${$pages.newPage.content}-${random}`);
+  });
+});
+
+it('allows to import members', () => {
+  cy.login();
+  cy.visit('/ghost/#/members/import');
+
```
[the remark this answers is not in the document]
joancafom: Okay, agreed. If there is an explicit agreement on trying to avoid `be.visible/exists` when possible we should trust the author, and hence not need these kinds of comments.

.vib/ghost/cypress/cypress/integration/ghost_spec.js
joancafom: That's a good point. Another good valid reasoning is that we want to execute tests twice in the future, maybe after scaling up/down replica sets.

In any case, Ghost's import members function does actually overwrite the data if it already existed, so it should not be a problem. In fact, I had manually tested that the tests can be run multiple times on the same deployment 😁

.vib/ghost/vib-platform-verify.json
```diff
@@ -0,0 +1,66 @@
+{
```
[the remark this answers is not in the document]
joancafom: Does no longer apply, I will resolve this thread.

.vib/ghost/cypress/cypress/integration/ghost_spec.js
joancafom: > Better said, In which case will we will not wait for the dashboard?

This is the only case in which we **can't wait for it**. Upon submitting the login form, Ghost will do different redirections and eventually show the Dashboard page. It is not only until those last redirections happen that one can navigate to other pages, else you are presented with an Unauthorized error. In short, this is the reason why I decided to add it to commands.

Why don't we want to wait for it here? As we are validating the log-in feature, I want to explicitly validate that there are no form errors. If I wait for Dashboard to appear, it means that the log-in form is already gone.

I don't know if this is kind of tricky and I should just always validate that in `commands` and that's it. WDYT?

.vib/ghost/cypress/cypress/support/commands.js
joancafom: While I agree that it is good having this documented in our strategy, I still see it valuable to have these corner cases in the specific tests where it applies. If this is a workaround that most assets will use then I don't see the point either, but I believe that we won't have that many using `hosts`. In fact, this is the only one that currently has it out of all the others already implemented.

IMHO, I'd rather be too verbose in these workarounds in the code than not commenting at all. Someone that does not have the context will probably don't understand why I am overwriting the built-in `visit` function here.

.vib/ghost/cypress/cypress/support/commands.js
joancafom: Replied in the thread above. Also, there is a summary in the comments below.

.vib/ghost/goss/goss.yaml
joancafom: I reckon this in fact validates that the output is empty. I mean, if there were an error and it was printed in `stdout`, this will fail as the output contains text.

.vib/ghost/cypress/cypress/integration/ghost_spec.js
```diff
@@ -0,0 +1,116 @@
+/// <reference types="cypress" />
+import { random, getPageUrlFromTitle, getUserFromEmail } from './utils';
+
+it('allows to log in and out', () => {
+  cy.login();
+
```
superaleks, line 6: New line here, is it on purpose?

.vib/ghost/cypress/cypress/integration/ghost_spec.js
```diff
@@ -0,0 +1,116 @@
...
+  cy.get("div[class*='avatar']").click();
+  cy.contains('Sign out').click();
+  cy.contains('Sign in');
+});
+
+it('allows to create and publish a new post', () => {
+  cy.login();
+  cy.visit('/ghost/#/posts');
+
```
superaleks, line 15: Same as above, new line.

.vib/ghost/cypress/cypress/integration/ghost_spec.js
joancafom: I removed new lines 😁  thanks!

superaleks (approved):
LGTM!

FraPazGal (approved):

PR #9776, opened 2022-04-12
[bitnami/minio] Fix minio test

superaleks:
**Description of the change**

I've noticed that the version of the tests for MiniO which is currently on master doesn't have the changes related to 'remove extra asserts' task, which was merged yesterday. 

My theory is that they got overwritten in one of the rebases. Since I have a copy on my fork, I'd like to fix this. 

**Test results**

[Please check here. ](https://github.com/superaleks/charts/runs/5992726291?check_suite_focus=true)

**Checklist**


FraPazGal (approved):
LGTM

.vib/minio/cypress/cypress/integration/minio_spec.js
```diff
@@ -1,67 +1,69 @@
...
-  cy.contains('div','Object uploaded successfully.').should('be.visible');
-})
+  cy.contains('#upload-main', 'Upload').click();
+  cy.contains('span', 'Upload File').should('be.visible').click();
+  cy.get('div#object-list-wrapper > input[type="file"]')
+    .should('not.be.disabled')
+    .selectFile('cypress/fixtures/example.json', { force: true });
+  cy.reload();
+  cy.contains('example.json');
+});
```
joancafom, line 32: Not a big concern, but this might no longer ensure that the test would be valid for multiple executions. If there is an error while uploading the file and then we reload the page, would we still see the `example.json` on the page?

Before we looked for the `Object uploaded successfully` notification.

superaleks: I understand this, however, in the case of MiniO we upload the file to a bucket we create on the fly. Hence, if we run the same test on the same server twice, the files will anyway be uploaded to different buckets so they will never coincide. 
MiniO

joancafom: Fair enough! 😁

joancafom (approved):
LGTM

PR #9778, opened 2022-04-12
[bitnami/nginx bitnami/nginx-intel] Clarify nginx-ldap-auth-daemon deprecation

carrodher:
Update the note for the major version bump adding more info about the deprecation of the container that has coincided in time with a security bug in the same component


agomezmoron (approved):
LGTM

beltran-rubo (approved):

PR #9794, opened 2022-04-13
[bitnami/airflow] feat: Airflow extraEnvVarsSecrets list

addisonjones-gc:
**Description of the change**

Background: The `extraEnvVarsSecret` is a useful feature that allows for a single secret to be loaded into pods as an environment variable using the `envFrom:` capability of the pod spec. 

A use case I encountered was the need to securely add environment variables to our pods using the [External Secrets](https://github.com/external-secrets/kubernetes-external-secrets) custom resource definition, but without loading everything into a single secret. This posed a potential issue with multiple developers modifying the same secret with the chance of overwriting the values. 

The result was the `extraEnvVarsSecrets` list that can be set at the global, scheduler, web, and worker level. Each secret can be loaded by providing the name of the secret just as one would do if defining the pod resource by hand. ex:
```yaml
envFrom:
  - secretRef:
     name: my-secret
```

Format for leveraging the new list:
```yaml
extraEnvVarsSecrets:
  - name: my-secret1
  - name: my-secret2
```

**Benefits**

Benefits of this feature include the ability to push extra environment variables as secrets to all scopes already leveraging the `extraEnvVars...` functionality.

**Possible drawbacks**

The functionality of this addition is somewhat limited in scope: The environment variables will be loaded from the Secret's keys and values without any ability to change the format via the chart. 

**Applicable issues**

  - fixes #

**Additional information**

**Checklist**

### DIFF
--- a/bitnami/airflow/Chart.yaml
+++ b/bitnami/airflow/Chart.yaml
@@ -32,4 +32,4 @@ name: airflow
 sources:
   - https://github.com/bitnami/bitnami-docker-airflow
   - https://airflow.apache.org/
-version: 12.1.9
+version: 12.2.0
--- a/bitnami/airflow/templates/config/configmap.yaml
+++ b/bitnami/airflow/templates/config/configmap.yaml
@@ -131,7 +131,7 @@ data:
             {{- if .Values.worker.extraEnvVars }}
             {{- include "common.tplvalues.render" (dict "value" .Values.worker.extraEnvVars "context" $) | nindent 12 }}
             {{- end }}
-          {{- if or .Values.worker.extraEnvVarsCM .Values.worker.extraEnvVarsSecret .Values.extraEnvVarsCM .Values.extraEnvVarsSecret }}
+          {{- if or .Values.worker.extraEnvVarsCM .Values.worker.extraEnvVarsSecret .Values.worker.extraEnvVarsSecrets .Values.extraEnvVarsCM .Values.extraEnvVarsSecret .Values.extraEnvVarsSecrets }}
           envFrom:
             {{- if .Values.extraEnvVarsCM }}
             - configMapRef:
@@ -148,6 +148,18 @@ data:
             {{- if .Values.worker.extraEnvVarsSecret }}
             - secretRef:
                 name: {{ .Values.worker.extraEnvVarsSecret }}
+            {{- end }}
+            {{- if .Values.extraEnvVarsSecrets }}
+            {{- range .Values.extraEnvVarsSecrets }}
+            - secretRef:
+                name: {{ .name }}
+            {{- end }}
+            {{- end }}
+            {{- if .Values.worker.extraEnvVarsSecrets }}
+            {{- range .Values.worker.extraEnvVarsSecrets }}
+            - secretRef:
+                name: {{ .name }}
+            {{- end }}
             {{- end }}
           {{- end }}
           ports:
--- a/bitnami/airflow/templates/config/configmap.yaml
+++ b/bitnami/airflow/templates/config/configmap.yaml
@@ -152,13 +152,13 @@ data:
             {{- if .Values.extraEnvVarsSecrets }}
             {{- range .Values.extraEnvVarsSecrets }}
             - secretRef:
-                name: {{ .name }}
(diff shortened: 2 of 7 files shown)


bitnami/airflow/templates/config/configmap.yaml
```diff
@@ -149,6 +149,18 @@ data:
             - secretRef:
                 name: {{ .Values.worker.extraEnvVarsSecret }}
             {{- end }}
+            {{- if .Values.extraEnvVarsSecrets }}
+            {{- range .Values.extraEnvVarsSecrets }}
```
addisonjones-gc, line 153: I would love some feedback as to whether there's a more elegant implementation of this. I'm new to Helm, so this was the best I could do with my current knowledge.

addisonjones-gc:
Hi @fmulero, was hoping you might be able to get some eyes on this. This is a feature we're hoping to roll into our environment soon and would like to continue using the source chart.

bitnami/airflow/Chart.yaml
```diff
@@ -32,4 +32,4 @@ name: airflow
 sources:
   - https://github.com/bitnami/bitnami-docker-airflow
   - https://airflow.apache.org/
-version: 12.1.8
+version: 12.1.9
```
javsalgar, line 35: Could you bump the minor version? It is adding a non-breaking feature

bitnami/airflow/templates/config/configmap.yaml
```diff
@@ -149,6 +149,18 @@ data:
...
+            - secretRef:
+                name: {{ .name }}
+            {{- end }}
+            {{- end }}
+            {{- if .Values.worker.extraEnvVarsSecrets }}
+            {{- range .Values.worker.extraEnvVarsSecrets }}
+            - secretRef:
+                name: {{ .name }}
```
javsalgar, line 161: I don't think we need to add a field `name`. I'd rather use a list with the names.

to instead of `{{ .name }}` it would be ` {{. }}`

javsalgar (changes requested):

javsalgar:
Thank you so much! Changes requested

bitnami/airflow/templates/config/configmap.yaml
addisonjones-gc: Ah perfect, this is great learning info for Helm, thanks!

bitnami/airflow/Chart.yaml
addisonjones-gc: @javsalgar I totally can. Just to make sure I'm clear that's 12.X right? So in this case we'd be going 12.1.8 to 12.2?

Thanks!

bitnami/airflow/templates/config/configmap.yaml
addisonjones-gc: So let me make sure I'm formatting this as expected. This worked in testing so I'm just ensuring it's what was intended: 

Current state with `name:`:
`values.yaml`
```yaml
extraEnvVarsSecrets:
  - name: secret-name
```
`configmap.yaml`:
```yaml
{{- range .Values.worker.extraEnvVarsSecrets }}
            - secretRef:
                name: {{ .name }}
```

Desired state: 
`values.yaml`
```yaml
extraEnvVarsSecrets:
  - secret-name
```
`configmap.yaml`:
```yaml
{{- range .Values.worker.extraEnvVarsSecrets }}
            - secretRef:
                name: {{ . }}
```

This produces the following which is what we want: 
```yaml
envFrom:
  - secretRef:
      - name: secret-name
```

bitnami/airflow/Chart.yaml
addisonjones-gc: Hey @javsalgar, I pushed this up just now to 12.2.0. Would you mind confirming for me that's what we're after? 

Thanks!

javsalgar (approved):
Thank you so much! LGTM

addisonjones-gc:
@javsalgar Awesome thank you! It looks like the merge was overridden by the bot. Would you mind re-approving and merging? Thanks!

javsalgar (approved):

Issue #5620, opened 2021-02-25
Running `helm diff` will always yield a change in the `memberlist` secret


phihos:
**Which chart**:
metallb 2.3.2

**Describe the bug**

Running `helm diff` will always yield a change in the `memberlist` secret. The is because the existing `speaker.secretValue` is unquoted and the "new" (but same) `speaker.secretValue` is quoted. 
I already traced the behavior down to `common.secrets.passwords.manage` but I could not find the actual culprit.

**To Reproduce**

1. Create a `values.yaml` with `speaker.secretValue` set to a fixed value.
2. Deploy with helm for the first time: `helm install --values values.yaml metallb bitnami/metallb`
3. Install helm-diff plugin and run something like `helm diff upgrade --values values.yaml metallb bitnami/metallb --show-secrets`

**Expected behavior**

The diff should be empty.

**Version of Helm and Kubernetes**:

- Output of `helm version`:

```
version.BuildInfo{Version:"v3.5.2", GitCommit:"167aac70832d3a384f65f9745335e9fb40169dc2", GitTreeState:"dirty", GoVersion:"go1.15.7"}
```

- Output of `kubectl version`:

```
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.4", GitCommit:"d360454c9bcd1634cf4cc52d1867af5491dc9c5f", GitTreeState:"clean", BuildDate:"2020-11-11T13:17:17Z", GoVersion:"go1.15.2", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.9+k3s1", GitCommit:"630bebf94b9dce6b8cd3d402644ed023b3af8f90", GitTreeState:"clean", BuildDate:"2020-09-17T19:05:07Z", GoVersion:"go1.13.15", Compiler:"gc", Platform:"linux/amd64"}
```

juan131:
Hi @phihos 

It's easy to reproduce by simply running this:

```console
$ helm template metallb bitnami/metallb -s templates/speaker/secret.yaml > secret.yaml
$ yq read secret.yaml
# Source: metallb/templates/speaker/secret.yaml
apiVersion: v1
data:
  secretkey: "c29tZS12YWx1ZQ=="
...
$ kubectl create -f secret.yaml
$ kubectl get secret metallb-memberlist -o yaml
apiVersion: v1
data:
  secretkey: c29tZS12YWx1ZQ==
[...]

PR #9804, opened 2022-04-14
Remove quote in common.secrets.passwords.manage

lindhe:
**Description of the change**

This change removes the `quote` function from the definition of the `password` variable in `common.secrets.passwords.manage`.

**Benefits**

With `quote`, I'm constantly getting a diff (in my deployment of bitnami/harbor version 12.3.1) because the applied secret value never has quotes while the template does have quotes, so `helm diff` reports an diff even though there's nothing to apply really.

**Possible drawbacks**

I think there should be no drawbacks. Since the value is base64, it should be safe to not quote it. But since it's a common library, it is hard for me to verify all places where it's used.

Whether or not this is a breaking change is a matter of preference. I think it's not a breaking change, since I can't think of a reason where the quotes makes a difference for how the value should be parsed. But someone may have a different view on this than I have. [Knowing what's breaking or not is hard.](https://xkcd.com/1172/)

**Applicable issues**

- Previously discussed in #5620 with no clear resolution.

**Additional information**

As far as I can tell, quoting a value for a Kubernetes Secret has no effect:

<table>
<tr>
<th>noquotes.yaml</th>
<th>quotes.yaml</th>
</tr>
<tr>
<td>

```yaml
apiVersion: v1
data:
  foo: YmFy
kind: Secret
metadata:
  name: test
```
</td>
<td>

```yaml
apiVersion: v1
data:
  foo: "YmFy"
kind: Secret
metadata:
  name: test
```
</td>
</tr>
</table>

```console
$ kubectl apply -f noquotes.yaml --dry-run=client -o jsonpath='{ .data.foo }'
YmFy
$ kubectl apply -f quotes.yaml --dry-run=client -o jsonpath='{ .data.foo }'
YmFy
```

**Checklist**

### DIFF
--- a/bitnami/common/Chart.yaml
+++ b/bitnami/common/Chart.yaml
@@ -20,4 +20,4 @@ sources:
   - https://github.com/bitnami/charts
   - https://www.bitnami.com/
 type: library
-version: 1.13.0
+version: 1.13.1
--- a/bitnami/common/templates/_secrets.tpl
+++ b/bitnami/common/templates/_secrets.tpl
@@ -98,7 +98,7 @@ The order in which this function returns a secret password:
     {{- printf "\nPASSWORDS ERROR: The secret \"%s\" does not contain the key \"%s\"\n" .secret .key | fail -}}
   {{- end -}}
 {{- else if $providedPasswordValue }}
-  {{- $password = $providedPasswordValue | toString | b64enc | quote }}
+  {{- $password = $providedPasswordValue | toString | b64enc }}
 {{- else }}
 
   {{- if .context.Values.enabled }}
@@ -114,9 +114,9 @@ The order in which this function returns a secret password:
     {{- $subStr := list (lower (randAlpha 1)) (randNumeric 1) (upper (randAlpha 1)) | join "_" }}
     {{- $password = randAscii $passwordLength }}
     {{- $password = regexReplaceAllLiteral "\\W" $password "@" | substr 5 $passwordLength }}
-    {{- $password = printf "%s%s" $subStr $password | toString | shuffle | b64enc | quote }}
+    {{- $password = printf "%s%s" $subStr $password | toString | shuffle | b64enc }}
   {{- else }}
-    {{- $password = randAlphaNum $passwordLength | b64enc | quote }}
+    {{- $password = randAlphaNum $passwordLength | b64enc }}
   {{- end }}
 {{- end -}}
 {{- printf "%s" $password -}}


migruiz4:
Hi @lindhe,

The `quote` is mainly used to prevent issues with how YAML interprets the base64 encoded field when an integer-only string is provided. 

For example:

<table>
<tr>
<th>noquotes.yaml</th>
<th>quotes.yaml</th>
</tr>
<tr>
<td>

```yaml
apiVersion: v1
data:
  foo: 1234
kind: Secret
metadata:
  name: noquotes
```
</td>
<td>

```yaml
apiVersion: v1
data:
  foo: "1234"
kind: Secret
metadata:
  name: quotes
```
</td>
</tr>
</table>

```console
$ k apply -f noquotes.yaml 
Error from server (BadRequest): error when creating "noquotes.yaml": Secret in version "v1" cannot be handled as a Secret: v1.Secret.Data: base64Codec: invalid input, error found in #10 byte of ...|a":{"foo":1234},"kin|..., bigger context ...|{"apiVersion":"v1","data":{"foo":1234},"kind":"Secret","metadata":{"annotations":{"|...
$ k apply -f quotes.yaml
secret/quotes created
```

This is because, while `YmFy` == `"YmFy"` (both are interpreted as strings), `1234` != `"1234"`, because the first one will be interpreted as an integer while the second one is interpreted as a string.

This not only applies to integer-only strings, it may also happen too when a string could be interpreted as an exponential, for example, `1234e4`.

It is probably very unlikely for `base64` to generate a string that could be interpreted as an integer, but since it is still a possibility, we thought adding that using `quote` would be the safest option since quoting string would not cause any harm.

The optimal solution would be to use `quote` only when the unquoted string could be interpreted as an integer, but I'm not sure Helm provides the tools to do so.

lindhe:
Excellent answer, you made the situation very clear. We should not remove the `quote` function then.

I wonder what we could do about this... I'll discuss it with a colleague and get back here if we find a reasonable approach.

SimonAlling:
> Excellent answer, you made the situation very clear. We should not remove the `quote` function then.
> 
> I wonder what we could do about this... I'll discuss it with a colleague and get back here if we find a reasonable approach.

Colleague here. Excellent answer indeed, @migruiz4.

For the record, here are some examples of strings whose Base64 encodings are numeric literals in YAML:

| Plain | Base64-encoded |
|-------|----------------|
| `ãMw` | `4013`         |
| `Ó~9` | `0345`         |
| `ãmù` | `4235`         |
| `ãn{` | `4257`         |
| `÷}5` | `9301`         |
| `ëÝu` | `6911`         |

> It is probably very unlikely for `base64` to generate a string that could be interpreted as an integer, but since it is still a possibility, we thought adding that using `quote` would be the safest option since quoting string would not cause any harm.

I totally would have agreed with you. Who would have thought that quoting the encoded string _would_ actually be problematic, as described in the OP … I'm not sure what to do tbh, but we'll look into it.

SimonAlling:
[YAML Multiline Strings](https://yaml-multiline.info/) to the rescue!

<table>
<tr>
<th>noquotes.yaml</th>
<th>quotes.yaml</th>
<th>multiline.yaml</th>
</tr>
<tr>
<td>

```yaml
apiVersion: v1
data:
  foo: 1234

kind: Secret
metadata:
  name: noquotes
```

</td>
<td>

```yaml
apiVersion: v1
data:
  foo: "1234"

kind: Secret
metadata:
  name: quotes
```

</td>
<td>

```yaml
apiVersion: v1
data:
  foo: |-
    1234
kind: Secret
metadata:
  name: multiline
```

</td>
</tr>
</table>

```console
$ python3
> import yaml
> yaml.safe_load(open("noquotes.yaml"))
{'apiVersion': 'v1', 'data': {'foo': 1234}, 'kind': 'Secret', 'metadata': {'name': 'noquotes'}}
> yaml.safe_load(open("quotes.yaml"))
{'apiVersion': 'v1', 'data': {'foo': '1234'}, 'kind': 'Secret', 'metadata': {'name': 'quotes'}}
> yaml.safe_load(open("multiline.yaml"))
{'apiVersion': 'v1', 'data': {'foo': '1234'}, 'kind': 'Secret', 'metadata': {'name': 'noquotes'}}

$ kubectl apply --dry-run=server -o jsonpath='{ .data.foo }' -f noquotes.yaml
Error from server (BadRequest): error when creating "noquotes.yaml": Secret in version "v1" cannot be handled as a Secret: v1.Secret.Data: base64Codec: invalid input, error found in #10 byte of ...|a":{"foo":1234},"kin|..., bigger context ...|{"apiVersion":"v1","data":{"foo":1234},"kind":"Secret","metadata":{"annotations":{"|...

$ kubectl apply --dry-run=server -o jsonpath='{ .data.foo }' -f quotes.yaml
1234

$ kubectl apply --dry-run=server -o jsonpath='{ .data.foo }' -f multiline.yaml
1234
```

lindhe:
Good catch. Unfortunately, I think that would require refactoring all places where `common.secrets.passwords.manage` is _used_, since `bitnami/common` is a library chart and the value `common.secrets.passwords.manage` may be used [like this](https://github.com/bitnami/charts/blob/531f7c28e5379fa8370cbc1bf8b23ac2ac486bb0/bitnami/harbor/templates/core/core-secret.yaml#L16) in many different places.

Get what I mean?

migruiz4:
Awesome! Very smart solution and thank you for joining the conversation.

After my first reply, I thought that maybe Helm's `printf` could help us with this 'quote only when strictly needed' issue. But unluckily it won't interpret types quoting them when needed, instead, if using `%s` it will do .

Then I thought maybe Helm had some function equivalent to `int` but for strings that I wasn't aware of, but instead, I found this:
https://helm.sh/docs/chart_best_practices/values/#make-types-clear

Although it states the following:
> The easiest way to avoid type conversion errors is to be explicit about strings, and implicit about everything else. Or, in short, quote all strings.

The latest tip did the work for me. So in addition to your tests:

<table>
<tr>
<th>tagged.yaml</th>
</tr>
<tr>
<td>

```yaml
apiVersion: v1
data:
  foo: !!string 1234
kind: Secret
metadata:
  name: tagged
```

</td>
</tr>
</table>


```console
$ kubectl apply --dry-run=server -o jsonpath='{ .data.foo }' -f tagged.yaml  
1234
```

I don't have a strong opinion on whether to use multiline or `!!string`, but I would like to confirm it fixes the original issue.

Tying to reproduce the OP issue, installed the latest version of `databus23/helm-diff` plugin, created a mock Chart with the following secret:
```mustache
apiVersion: v1
kind: Secret
metadata:
  name: test
type: Opaque
data:
  foo: {{ include "common.secrets.passwords.manage" (dict "secret" "test" "key" "secretKey" "length" 16 "providedValues" (list "foo") "context" $) }}
```
And values.yaml:
```
foo: bar
```

But I can't get `helm diff upgrade mock .` to detect the quote mismatch. @lindhe

lindhe:
Being able to specify the type explicitly sounds very nice! If that works, I'll make sure to suggest that fix to the bitnami/harbor chart.

But we should really try to replicate this issue first. It's very strange, I can reproduce it using my current Harbor installation, but not using the minimal example you suggest. I'll dig around to see if I can figure out the actual MWE.

PR #9835, opened 2022-04-20
[bitnami/spring-cloud-dataflow] Added the Spring Cloud Dataflow Helm chart tests

superaleks:

**Description of the change**
Increasing the test coverage of Spring Cloud Dataflow Helm chart by adding Cypress and Goss tests. 

**Benefits**

Higher quality of the Helm charts. 

**Possible drawbacks**
Test flakiness.

**Test results** - IN PROGRESS

9 consecutive pipelines were executed without failures. 
[Please see them in the history here](https://github.com/superaleks/charts/pull/3)

**Checklist**

### DIFF
--- a/.vib/spring-cloud-dataflow/cypress/cypress/integration/spring_cloud_dataflow_spec.js
+++ b/.vib/spring-cloud-dataflow/cypress/cypress/integration/spring_cloud_dataflow_spec.js
@@ -24,7 +24,13 @@ it('allows a stream to be created and deployed', () => {
     cy.get('.CodeMirror-line').type(stream.newStream.type);
   });
 
-  cy.contains('#v-2', 'mongodb').and('contain', 'cassandra');
+  cy.fixture('applications').then((application) => {
+    cy.contains('#v-2', application.newApplication.streamApplication1).and(
+      'contain',
+      application.newApplication.streamApplication2
+    );
+  });
+
   cy.contains('button', 'Create stream(s)').click();
   cy.get('.modal-content').should('contain', 'Create Stream');
   cy.fixture('streams').then((stream) => {
--- a/.vib/spring-cloud-dataflow/cypress/cypress/fixtures/applications.json
+++ b/.vib/spring-cloud-dataflow/cypress/cypress/fixtures/applications.json
@@ -1,6 +1,8 @@
 {
   "newApplication": {
     "streamApplicationType": "Stream application starters for Kafka/Maven",
-    "taskApplicationType": "Task application starters for Maven"
+    "taskApplicationType": "Task application starters for Maven",
+    "streamApplication1": "mongodb",
+    "streamApplication2": "cassandra"
   }
 }


.vib/spring-cloud-dataflow/cypress/cypress/integration/utils.js
```diff
@@ -0,0 +1,51 @@
+/// <reference types="cypress" />
+export let random = (Math.random() + 1).toString(36).substring(7);
+
```
superaleks, line 3: These are added as separate methods since I am using them to achieve the desired application state without creating dependencies between tests.

.vib/spring-cloud-dataflow/cypress/cypress/integration/utils.js
```diff
@@ -0,0 +1,51 @@
...
+export const importATaskApplication = () => {
+  cy.visit('/dashboard');
+  cy.contains('button', 'Add application(s)').click();
+  cy.contains(
+    '.clr-control-label',
+    'Task application starters for Maven'
+  ).click();
+  cy.contains('.btn-primary', 'Import').click();
+  cy.get('.toast-container').should('contain', 'Application(s) Imported');
```
superaleks, line 27: An alternative way of asserting without using the `cy.contains()`.

.vib/spring-cloud-dataflow/goss/goss.yaml
```diff
@@ -0,0 +1,28 @@
...
+    exec: id -u
+    exit-status: 0
+    stdout:
+      - 1001
+    stderr: []
+  java-test:
+    exec: java -version
+    exit-status: 0
+    stdout:
+    stderr: []
```
superaleks, line 28: There is not a lot to verify using Goss, since the app doesn't come with the CLI installed. 
In order to obtain the dataflow CLI, a separate container should be installed which is out of scope for community at the time being.

.vib/spring-cloud-dataflow/cypress/cypress.json
```diff
@@ -0,0 +1,5 @@
+{
+    "baseUrl": "http://localhost",
+    "defaultCommandTimeout":30000,
+    "pageLoadTimeout": 240000
```
superaleks, line 4: This is one of the applications that load slowly after installation.

.vib/spring-cloud-dataflow/cypress/cypress/support/index.js
```diff
@@ -0,0 +1,20 @@
...
+// You can change the location of this file or turn off
+// automatically serving support files with the
+// 'supportFile' configuration option.
+//
+// You can read more here:
+// https://on.cypress.io/configuration
+// ***********************************************************
+
+// Import commands.js using ES2015 syntax:
+import './commands'
```
FraPazGal, line 17: You may need to pass this file throughout the linter.

.vib/spring-cloud-dataflow/goss/goss.yaml
```diff
@@ -0,0 +1,28 @@
...
+    owner: root 
+command:
+  user-id-test:
+    exec: id -u
+    exit-status: 0
+    stdout:
+      - 1001
+    stderr: []
+  java-test:
+    exec: java -version
```
FraPazGal, line 25: I understand this test just want to check java is installed and part of `$PATH` right? Do we want want to check for a specific java version?

.vib/spring-cloud-dataflow/vib-verify.json
```diff
@@ -1,36 +1,36 @@
 {
-  "phases": {
```
FraPazGal, line 2: AFAIK, we are leaving this blank space in our `.json` files. Could you please recheck your linter config and re-run it here?

.vib/spring-cloud-dataflow/cypress/cypress/integration/utils.js
FraPazGal: I'm not sure these functions can be considered utility ones instead of proper tests. It's true you need them to achieve the desired state for your tests, but I think the issue here is their required starting state. 

For example, the `check created task and schedule one` test could be `create task and schedule one` and we will avoid having the `createATask` function. Moreover, here you are in fact testing that the imports function properly and that the task creation process works.

As I understand it, utility functions should be small functions that help in the creation of the tests such as `random` or `obtainCredentials` functions.

.vib/spring-cloud-dataflow/cypress/cypress/plugins/index.js
```diff
@@ -0,0 +1,22 @@
+/// <reference types="cypress" />
```
FraPazGal, line 1: I think we don't need to add this file to the repo and leave it for local testing only.

FraPazGal (commented):
Hello @superaleks, I added a few comments. Could you please look at them?

.vib/spring-cloud-dataflow/goss/goss.yaml
superaleks: No, we are avoiding checking the version in order to facilitate maintenance. 
I wanted to check that Java is present (If the command executes properly).

.vib/spring-cloud-dataflow/vib-verify.json
superaleks: I passed this through linter and this is the result. Which linter should I use for `JSON` files and what should be the config?

.vib/spring-cloud-dataflow/cypress/cypress/integration/utils.js
superaleks: I also think these methods don't belong to the `utils.js`.
As agreed on our call, I'll create a new .js file with these methods which are used for application state preparation.

.vib/spring-cloud-dataflow/cypress/cypress/fixtures/task-to-import.json
```diff
@@ -0,0 +1 @@
+{"date":1650360393208,"tasks":[{"name":"test-task-imported","dslText":"timestamp","composed":false,"status":"UNKNOWN","composedTaskElement":false,"description":""}]}
```
joancafom, line 1: It is not important, but I'd format this to look similar to the other `json` files

.vib/spring-cloud-dataflow/cypress/cypress/fixtures/tasks.json
```diff
@@ -0,0 +1,5 @@
+{
+    "newTask": {
+        "name": "test-task"
+    }
+}
```
joancafom, line 5: Is this file using 4 spaces to indent instead of 2?

.vib/spring-cloud-dataflow/cypress/cypress/integration/prepare-app-state.js
```diff
@@ -0,0 +1,52 @@
+/// <reference types="cypress" />
+import { random } from './utils';
+
+export const importAStreamApplication = () => {
+  cy.visit('/dashboard');
+  cy.contains('button', 'Add application(s)').click();
+  cy.contains(
+    '.clr-control-label',
```
joancafom, line 8: Instead of a class selector (and as we are already using a very specific string `Stream application starters for Kafka/Maven`), could we use a more generic selector here like `button`, which is less prone to change?

.vib/spring-cloud-dataflow/cypress/cypress/integration/prepare-app-state.js
```diff
@@ -0,0 +1,52 @@
...
+    'Task application starters for Maven'
+  ).click();
+  cy.contains('.btn-primary', 'Import').click();
+  cy.get('.toast-container').should('contain', 'Application(s) Imported');
+  cy.get('[routerlink="tasks-jobs/tasks"]').click();
+  cy.contains('button', 'Create task').click();
+  cy.get('.content-area')
+    .should('contain', 'timestamp')
+    .and('contain', 'timestamp-batch');
+};
```
joancafom, line 33: As you are reusing about half of the lines with `importAStreamApplication`, do you think it would make sense to have a generic method, say `importApplication(String type)`, and act depending on the Application `type`?

.vib/spring-cloud-dataflow/cypress/cypress/integration/spring-cloud-dataflow_spec.js
```diff
@@ -0,0 +1,107 @@
...
+import {
+  importAStreamApplication,
+  importATaskApplication,
+  createATask,
+} from './prepare-app-state.js';
+
+it('allows getting Spring Cloud Dataflow info', () => {
+  cy.visit('/dashboard');
+  cy.get('.signpost-trigger').click();
+  cy.contains('.signpost-content-body', 'Version');
```
joancafom, line 12: It would also be nice if we can check there is an actual value in the Version field.

.vib/spring-cloud-dataflow/cypress/cypress/integration/spring-cloud-dataflow_spec.js
```diff
@@ -0,0 +1,107 @@
...
+  cy.contains('.btn-primary', 'Create stream(s)').click();
+  cy.get('.CodeMirror-line').type(STREAM);
+  cy.contains('#v-2', 'mongodb').and('contain', 'cassandra');
+  cy.contains('button', 'Create stream(s)').click();
+  cy.get('.modal-content').should('contain', 'Create Stream');
+  cy.fixture('streams').then((stream) => {
+    cy.get('input[placeholder="Stream Name"]').type(
+      `${stream.newStream.name}-${random}`
+    );
+    console.log(random);
```
joancafom, line 34: I think you forgot to remove this line.

.vib/spring-cloud-dataflow/cypress/cypress/integration/spring-cloud-dataflow_spec.js
```diff
@@ -0,0 +1,107 @@
...
+    cy.contains('.btn-primary', 'Create the stream').click();
+    cy.contains(
+      '.datagrid-inner-wrapper',
+      `${stream.newStream.name}-${random}`
+    );
+  });
+  cy.contains('.toast-container', 'successfully');
+  cy.contains('clr-dg-cell', 'UNDEPLOYED')
+    .siblings('clr-dg-cell', 'test-stream')
+    .first()
```
joancafom, line 44: Hmm relying on the order of an array wrt to an element in the UI does not sound very resilient. Is there any other possibility to avoid this?

Sometimes, when is difficult to come up with the right global selector, it is useful to first match the div/table containing our desired element and then use the cypress function `within`.

superaleks: It's not really relevant which element we click. 
It is a table and I want to take any element. It can be the first or the last since they are all the same. 
I am taking the last of the elements from the table which has the tag `UNDEPLOYED`.  It's just a convenient use of the `siblings()` method. 
I can show you over a call and we can decide together if it's appropriate or not.

.vib/spring-cloud-dataflow/cypress/cypress/integration/prepare-app-state.js
superaleks: Changed accordingly.

.vib/spring-cloud-dataflow/cypress/cypress/integration/prepare-app-state.js
superaleks: Yes, I like this suggestion a lot! Applied it.

.vib/spring-cloud-dataflow/cypress/cypress/integration/spring-cloud-dataflow_spec.js
superaleks: I've added a new method of verifying this as per your suggestion.

.vib/spring-cloud-dataflow/cypress/cypress/integration/spring-cloud-dataflow_spec.js
```diff
@@ -0,0 +1,108 @@
...
+    .and('contain', 'Shell');
+});
+
+it('allows a stream to be created and deployed', () => {
+  const STREAM = 'mongodb | cassandra';
+  const STREAM_APPLICATION = 'Stream application starters for Kafka/Maven';
+
+  importAnApplication(STREAM_APPLICATION);
+  cy.visit('/dashboard');
+  cy.get('[routerlink="streams/list"').click();
```
FraPazGal, line 23: As a go-to policy, do we want to access the page where the testing will take place by navigating the UI or using `cy.visit`? 

I don't really have any strong thoughts one way or the other, as I think both are error-prone, but having a defined policy here will be helpful to maintain a single style.

.vib/spring-cloud-dataflow/cypress/cypress/integration/spring-cloud-dataflow_spec.js
```diff
@@ -0,0 +1,108 @@
...
+  cy.contains('.btn-danger', 'Destroy the task').click();
+  cy.contains('.toast-container', '1 task definition(s) destroyed.');
+});
+
+it('allows importing a task from a file and destroying it ', () => {
+  cy.visit('/dashboard');
+  cy.get('[routerlink="manage/tools"]').click();
+  cy.contains('a', 'Import tasks from a JSON file').click();
+  cy.get('input[type="file"]').selectFile(
+    'cypress/fixtures/task-to-import.json',
```
FraPazGal, line 85: Sorry to be singing always the same song, but would there be an issue if the corner case of importing the same data twice happens? 

Maybe we need to brainstorm this corner case and its possible solutions, as we will encounter it every time we work with non-randomized data.

.vib/spring-cloud-dataflow/cypress/cypress/integration/prepare-app-state.js
```diff
@@ -0,0 +1,27 @@
...
+
+export const createATask = (task) => {
+  cy.visit('/dashboard');
+  cy.get('[routerlink="tasks-jobs/tasks"]').click();
+  cy.contains('.btn-primary', 'Create task').click();
+  cy.get('.CodeMirror-line').type(task);
+  cy.contains('#v-2', task);
+  cy.contains('button', 'Create task').click();
+  cy.contains('.modal-content', 'Create task');
+  cy.fixture('tasks').then((task) => {
```
joancafom, line 20: Beware that the variable `task` already exists at this level (it is the arg of the `createATask` function!)

.vib/spring-cloud-dataflow/cypress/cypress/integration/spring-cloud-dataflow_spec.js
```diff
@@ -0,0 +1,108 @@
...
+    .siblings('clr-dg-cell', 'test-stream')
+    .first()
+    .click();
+  cy.contains('button#btn-deploy', 'Deploy stream').click();
+  cy.contains('button', 'Deploy stream').click();
+  cy.get('.toast-container').should('contain', 'Deploy success');
+});
+
+it('allows a task to be scheduled and destroyed', () => {
+  const CRON_EXPRESSION = '*/5 * * * *';
```
joancafom, line 50: This is also part of the fixture I think

.vib/spring-cloud-dataflow/cypress/cypress/integration/spring-cloud-dataflow_spec.js
```diff
@@ -0,0 +1,108 @@
...
+  const TASK_APPLICATION = 'Task application starters for Maven';
+  const TASK_TYPE = 'timestamp';
```
joancafom, line 52: So is this!

.vib/spring-cloud-dataflow/cypress/cypress/integration/spring-cloud-dataflow_spec.js
```diff
@@ -0,0 +1,108 @@
...
+
+  importAnApplication(TASK_APPLICATION);
+  createATask(TASK_TYPE);
+  cy.get('[routerlink="tasks-jobs/tasks"]').click();
```
joancafom, line 56: Again, prefer `cy.visit` over UI interaction.

.vib/spring-cloud-dataflow/cypress/cypress/integration/spring-cloud-dataflow_spec.js
```diff
@@ -0,0 +1,108 @@
...
+  cy.get('input[name="example"]').last().type(CRON_EXPRESSION);
+  cy.contains('button', 'Create schedule(s)').click();
+  cy.contains('.toast-container', 'Successfully');
+  cy.contains('a', 'Tasks').click();
+  cy.contains('clr-dg-cell', 'UNKNOWN')
+    .siblings('clr-dg-cell', 'test-task-')
+    .first()
+    .click();
+  cy.contains('button', 'Destroy task').click();
+  cy.contains('.btn-danger', 'Destroy the task').click();
```
joancafom, line 76:
```suggestion
  cy.contains('button', 'Destroy the task').click();
```

.vib/spring-cloud-dataflow/cypress/cypress/integration/spring-cloud-dataflow_spec.js
```diff
@@ -0,0 +1,108 @@
...
+it('allows importing a task from a file and destroying it ', () => {
+  cy.visit('/dashboard');
+  cy.get('[routerlink="manage/tools"]').click();
+  cy.contains('a', 'Import tasks from a JSON file').click();
+  cy.get('input[type="file"]').selectFile(
+    'cypress/fixtures/task-to-import.json',
+    { force: true }
+  );
+  cy.contains('button', 'Import').click();
+  cy.contains('.modal-body', '1 task(s) created');
```
joancafom, line 89:
```suggestion
  cy.contains('1 task(s) created');
```

.vib/spring-cloud-dataflow/cypress/cypress/integration/spring-cloud-dataflow_spec.js
```diff
@@ -0,0 +1,108 @@
...
+  cy.get('.close').click();
+  cy.get('[routerlink="tasks-jobs/tasks"]').click();
```
joancafom, line 91: You can replace this two with a `cy.visit`

.vib/spring-cloud-dataflow/cypress/cypress/integration/spring-cloud-dataflow_spec.js
```diff
@@ -0,0 +1,108 @@
...
+  cy.contains('button', 'Group Actions').click();
+  cy.get('[aria-label="Select All"]').click({ force: true });
+  cy.contains('button', 'Destroy task').click();
+  cy.contains('.modal-content', 'Confirm Destroy Task');
+  cy.contains('button', 'Destroy the task').click();
+  cy.get('.toast-container').should('contain', 'destroyed');
+});
+
+it('allows unregistering of an application', () => {
+  const STREAM_APPLICATION = 'Stream application starters for Kafka/Maven';
```
joancafom, line 101: I am guessing this is kind of the `type` of stream application we want to create. If I am right, then I also think this is a characteristic of the application and should be in fixtures.

.vib/spring-cloud-dataflow/cypress/cypress/integration/prepare-app-state.js
superaleks: Indeed, I changed the argument name for clarity. Now the argument is `taskType`.

.vib/spring-cloud-dataflow/cypress/cypress/integration/spring-cloud-dataflow_spec.js
superaleks: Yes, the same case is happening here. This is why I mentioned this might be a global problem.
In some apps, we have the option to modify the application name which is good (ex. Mediawiki).
However, other apps don't allow for this.
Would it be OK if were leave this for now and brainstorm a global solution we can apply to these types of assets? Spring Cloud Dataflow is not the only one.

.vib/spring-cloud-dataflow/cypress/cypress/integration/spring-cloud-dataflow_spec.js
superaleks: We have the policy to use the URL on the navigation to avoid flakiness. 
I forgot about it for a time, hence the mistake!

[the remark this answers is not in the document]
superaleks: I forgot about this hence the mistakes. Fixed it across the spec.

.vib/spring-cloud-dataflow/cypress/cypress/integration/spring-cloud-dataflow_spec.js
superaleks: Yes, now I created a fixture for this.
I've been indecisive since we are not exactly creating this application but selecting it from a list.

.vib/spring-cloud-dataflow/cypress/cypress/support/commands.js
```diff
@@ -0,0 +1,26 @@
+const COMMAND_DELAY = 800;
+const TYPE_DELAY = 200;
```
FraPazGal, line 2: Could you please give a bit of context regarding the need for a `TYPE_DELAY`? 

Looking now at the const names, shouldn't we be using `CLICK_DELAY` instead of `COMMAND_DELAY`?

superaleks: Yes, you are right. I fixed this mistake.
The context: when tests run in sequence sometimes Cypress types the values too quickly and one of the letters (usually the first one) doesn't get typed in properly. 
This flakiness is solved by adding a small delay in typing.

.vib/spring-cloud-dataflow/cypress/cypress/integration/spring-cloud-dataflow_spec.js
superaleks: I've created a task related to this in our internal EPIC.

.vib/spring-cloud-dataflow/goss/goss.yaml
```diff
@@ -0,0 +1,23 @@
...
+    filetype: directory
+    mode: "0775"
+    owner: root
+command:
+  user-id-test:
+    exec: id -u
+    exit-status: 0
+    stdout:
+      - 1001
+    stderr: []
```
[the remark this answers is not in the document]
superaleks: Changed this test!

.vib/spring-cloud-dataflow/goss/goss.yaml
```diff
@@ -0,0 +1,23 @@
...
+    exists: true
+    filetype: directory
+    mode: "0775"
+    owner: root
+command:
+  user-id-test:
+    exec: id -u
+    exit-status: 0
+    stdout:
+      - "/[^0]+/"
```
FraPazGal, line 17: I checked this regex and it doesn't work for every case. E.g: 1001 (contains 0)

superaleks: OK, let me see how can I improve it.

superaleks: I've used your example to fix this test.

FraPazGal (dismissed):

.vib/spring-cloud-dataflow/cypress/cypress/integration/spring_cloud_dataflow_spec.js
```diff
@@ -0,0 +1,115 @@
...
+  cy.fixture('applications').then((application) => {
+    importAnApplication(application.newApplication.streamApplicationType);
+  });
+  cy.visit('dashboard/#/streams/list');
+  cy.contains('button', 'Create stream(s)').click();
+  cy.fixture('streams').then((stream) => {
+    cy.get('.CodeMirror-line').type(stream.newStream.type);
+  });
+
+  cy.contains('#v-2', 'mongodb').and('contain', 'cassandra');
```
joancafom, line 27: This is directly related to the test data. I think that changing the stream type in the fixture will make the test fail.

superaleks: I moved this to the application data fixture, hope it's better now!

joancafom (approved):

FraPazGal (approved):

PR #9909, opened 2022-04-25
[bitnami/grafana-operator] Add parameters to configure plugins init image

### LINKED ISSUE #8853: [bitnami/grafana-operator]
**Which chart**:

bitnami/grafana-operator release: 2.2.2

**Describe the bug**

The grafana-plugins-init-container-image will always mount the baseImage

https://github.com/bitnami/charts/blob/9fb725ce244079fc8a3cfa1c215d2550b9eef268/bitnami/grafana-operator/templates/deployment.yaml#L84-L88

**Expected behavior**

Seperate baseImage/tag from and plugininitimage/tag in deployment,yaml and value.yaml, so that there is no need to manual adjust the deployment yaml with every bitnami grafana-operator release

**Additional context**

https://github.com/grafana-operator/grafana_plugins_init

ronleb:
Signed-off-by: Ronan LE BRUN ronan.lebrun@gmail.com

**Description of the change**

Add parameters to specify a custom plugins-init-container image instead of the grafana image

**Benefits**

Allows to use a stock image of grafana and a custom image containing the grafana plugins 

**Possible drawbacks**

Requires to specify grafana image twice instead of once if user has a custom grafana image with embedded plugins

**Applicable issues**

  - fixes #8853

**Additional information**

Another solution could be to hide the parameters and default to use the grafana.image parameters

This would have the advantage of requiring only a minor version bump as it would be retro-compatible but would hide the parameters from the documented parameters (or by using nil/empty values) 

This is open to discussion

**Checklist**


fmulero (changes requested):
Hi @ronleb thanks a lot for your contribution!

I think in this case you don't need a major version, you are adding a new feature and there are no breaking changes, so you only need to increase the minor digit ([semver](https://semver.org/))

Great work

ronleb:
I could see one regression case if users of the chart have overridden the default grafana image with a custom one that ship their plugins. In that case the change would break their setup as the plugins init container would not be their custom image anymore until they also override the new parameters.

This could be handled a bit differently by defaulting to the grafana image if the parameters are empty. 
This would make things retro compatible and avoid duplicating references to the base image.

fmulero:
You right! Probably a lot of users have their custom images due to this limitation. Having the base grafana image as default if `pluginsInitContainerImage` is not set would be great

fmulero (approved):
Great! LGTM

PR #9915, opened 2022-04-25
[bitnami/kafka] Add parameters to configure SSL for Zookeeper client connections

yohboy:

**Description of the change**
This PR adds parameters to configure SSL for Zookeeper client connections from Kafka. This feature already exists in the Kafka container. This PR adds the corresponding environment variables if `auth.zookeeper.tls.enabled` is set to true. It also mounts a secret to provide Keystore/Truststore. A second secret is used to provide passwords for Keystore/Truststore. We can customize keys for both of these secrets (useful for cert-manager support).  
This PR also adds a template function to select the kafka<->zookeeper protocol. It was previously unset by default (default to `PLAINTEXT`) and set to `SASL` if `zookeeper.auth.enabled` `auth.sasl.jaas.zookeeperUser` were used.

**Benefits**
Can connect Zookeeper with TLS enabled.

**Possible drawbacks**
None.

**Applicable issues**
None.

**Additional information**
None.

**Checklist**


dgomezleon (dismissed):
LGTM

bitnami/kafka/values.yaml
```diff
@@ -342,6 +342,39 @@ auth:
     ## ref: https://docs.confluent.io/current/kafka/authentication_ssl.html#optional-settings
     ##
     endpointIdentificationAlgorithm: https
+  ## Zookeeper client configuration for kafka brokers
```
joancafom, line 345: From my point of view, I think this configuration does not belong here. This is not really something we want to directly configure on Kafka. Instead, this is more of a "reactive" change: we need to configure Kafka to use TLS because the underlying `zookeeper` is using it. As such, this should belong in the `zookeeper:` entry in the `values.yaml`.

Another good point to it is that this chart is using the zookeeper subchart by default (again, the `zookeeper:` entry in `values.yaml`). Imagine that you wanted to configure TLS for Kafka-Zookeeper communications --> You would need to do it both in `auth.zookeeper.tls` and `zookeeper.tls` with this approach.

This logic already exists in the zookeeper subchart, so we just need to move the entries under `zookeeper:` and make this chart "react" to that configuration. For example, take a look at how the authentication credentials for `redis` are configured in the `bitnami/argo-cd` chart, and how there is logic to detect which is the right `secret` to use:

https://github.com/bitnami/charts/blob/fdd8b691290c8f039d192a2b7c6f87706e7c2390/bitnami/argo-cd/values.yaml#L2325-L2333
https://github.com/bitnami/charts/blob/fdd8b691290c8f039d192a2b7c6f87706e7c2390/bitnami/argo-cd/templates/_helpers.tpl#L143-L153
https://github.com/bitnami/charts/blob/fdd8b691290c8f039d192a2b7c6f87706e7c2390/bitnami/argo-cd/templates/repo-server/deployment.yaml#L114-L121

Does it make sense to you? WDYT?

yohboy: Hi @joancafom,
thank you for the review and sorry for the late reply!  
I think I do not completely agree with this point, let me explain. 

> This is not really something we want to directly configure on Kafka. Instead, this is more of a "reactive" change: we need to configure Kafka to use TLS because the underlying zookeeper is using it. 

In my opinion, this is something we want to configure directly on Kafka. Of course, we need to configure it because zookeeper uses TLS, but we also need to configure it for `mtls` between Kafka and Zookeeper. You can configure zookeeper to accept TLS connection without configuring `mtls` in Kafka. 

> Another good point to it is that this chart is using the zookeeper subchart by default (again, the zookeeper: entry in values.yaml). Imagine that you wanted to configure TLS for Kafka-Zookeeper communications --> You would need to do it both in auth.zookeeper.tls and zookeeper.tls with this approach.

To my mind, even if we configure these parameters in the `zookeeper:` entry, you still need to configure 2 keystore/truststore. One for zookeeper, and one for kafka (mtls). You do not need/want to use the same keystore/truststore for both. So one entry will only impact Zookeeper, while the second will only impact Kafka. Of course there are some links between these two configurations because of the truststore, but it is two different configurations.

> This logic already exists in the zookeeper subchart, so we just need to move the entries under zookeeper: and make this chart "react" to that configuration. For example, take a look at how the authentication credentials for redis are configured in the bitnami/argo-cd chart, and how there is logic to detect which is the right secret to use:

If we do this, you will use the same keystore/truststore for both server and client. I think this is not a good practice. What do you think?

joancafom: Sorry for the delay. Let me discuss it with a teammate and I'll get back to you shortly, I think you're right 😁

yohboy: Thank you! Waiting for your feedback :)

joancafom: After some brief conversation, I think it does make sense to have these two entries as you originally implemented them 😁 Not only will that allow to set different keystore/trustore for Kafka & Zookeeper, but it will also allow the use of TLS with an external zookeeper instance. 

Let me recheck the PR for further comments now :) Thanks!!

bitnami/kafka/templates/_helpers.tpl
```diff
@@ -210,6 +210,21 @@ SASL_PLAINTEXT
 {{- end -}}
 {{- end -}}
 
+{{/*
+Return the protocol used with zookeeper
+*/}}
```
joancafom, line 215: Super nice! 🎉

bitnami/kafka/values.yaml
```diff
@@ -342,6 +342,39 @@ auth:
...
+      enabled: false
+      ## @param auth.zookeeper.tls.type Format to use for TLS certificates. Allowed types: `jks` and `pem`.
+      ##
+      type: jks
+      ## @param auth.zookeeper.tls.verifyHostname Hostname validation.
+      ##
+      verifyHostname: true
+      ## @param auth.zookeeper.tls.existingSecret Name of the existing secret containing the TLS certificates for ZooKeeper client communications.
+      ##
+      existingSecret: ""
```
joancafom, line 362: After taking another look at the changes they LGTM overall, good job! However, I do have a small concern here.

I am not a Kafka expert, so maybe you have more background on this than me. Looking at the chart, it seems that in order to configure TLS for Kafka itself, one has to configure a TLS certificate for EACH broker. I have also found this in the documentation:

https://kafka.apache.org/documentation/#security_ssl_key

Because of that, the Bitnami Kafka chart exposes an **array** of `existingSecrets`. However, they are not explicit when it comes to Zookeeper TLS:

https://kafka.apache.org/documentation/#zk_authz

> When using mTLS alone, every broker and any CLI tools (such as the [ZooKeeper Security Migration Tool](https://kafka.apache.org/documentation/#zk_authz_migration)) should identify itself with the same Distinguished Name (DN) because it is the DN that is ACL'ed

This leaves me in doubt here on whether multiple certs should be used for Zookeeper auth as well or not. Technically, it should be possible to use the same client certs as they mention that is the DN that is ACL, so maybe using a TLS per broker should be regarded as an improvement of this PR.

What are your thoughts on this? Have you tried these changes with more than one broker?

yohboy: Hi @joancafom!
Thank you!

> What are your thoughts on this? Have you tried these changes with more than one broker?
Yes, I tried these changes with 3 Kafka brokers and 3 Zookeeper nodes. If you use the same certificate for Zookeeper auth, you must have the same DN and add all possible SAN to allow hostname validation. Using the same DN is a requirement if you use mtls auth only (unless you write a custom auth provider). If you use SASL+mtls, then you can configure the DN as you want since they will use the ACL'ed SASL. 

To be honest, it can be interesting to add the ability to configure one TLS certificate used for Zookeeper auth per broker. However, I would implement this later as an improvement of this PR :). This PR already provides the ability to configure mtls with Zookeeper et does not restrict the user possibilities.

joancafom: Agreed, thank you so much 🚀

joancafom (approved):
LGTM

PR #9929, opened 2022-04-26
[bitnami/etcd] PodDisruptionBudget should be enabled by default

### LINKED ISSUE #9927: [bitnami/etcd] PodDisruptionBudget should be enabled by default
### Name and Version

bitnami/etcd

Without PDBs, every time a cluster is updated, since Kubernetes is not aware of how to handle the application without PDB, it may cause a majority failure. 

According to [the etcd docs on Majority Failure](https://etcd.io/docs/v3.5/op-guide/failures/#majority-failure):
> When the majority members of the cluster fail, the etcd cluster fails and cannot accept more writes.
>
>The etcd cluster can only recover from a majority failure once the majority of members become available. If a majority of members cannot come back online, then the operator must start disaster recovery to recover the cluster.

This means that manual intervention is needed.

I see that this chart already has PDB configuration set, which is good. However I propose to have PDB be set to default true. Given that "we (chart developers)" are aware of the app we are deploying I don't think we ever want consumers of the chart to enable PDB. The reality is, those who haven't PDB enabled on the chart are most probably unaware how etcd handles a majority failure and if explained to them they will turn it on.

I am pretty new to helm charts so I am not sure if it is possible to automatically set PDB to the minimum of `(N-1)/2` of the replica count

_No response_

cdemi:
**Description of the change**

Without PDBs, every time a cluster is updated, since Kubernetes is not aware of how to handle the application without PDB, it may cause a majority failure. 

According to [the etcd docs on Majority Failure](https://etcd.io/docs/v3.5/op-guide/failures/#majority-failure):
> When the majority members of the cluster fail, the etcd cluster fails and cannot accept more writes.
>
>The etcd cluster can only recover from a majority failure once the majority of members become available. If a majority of members cannot come back online, then the operator must start disaster recovery to recover the cluster.

This means that manual intervention is needed.

**Benefits**

Enabling this by default instructs the Kubernetes cluster and operators that at least 51% must remain available during voluntary disruptions, such as cluster upgrades.

**Possible drawbacks**
None as far as I can see

**Applicable issues**
  - fixes #9927

**Checklist**

### DIFF
--- a/bitnami/etcd/Chart.yaml
+++ b/bitnami/etcd/Chart.yaml
@@ -24,4 +24,4 @@ name: etcd
 sources:
   - https://github.com/bitnami/bitnami-docker-etcd
   - https://coreos.com/etcd/
-version: 8.0.2
+version: 8.0.3
--- a/bitnami/etcd/Chart.yaml
+++ b/bitnami/etcd/Chart.yaml
@@ -24,4 +24,4 @@ name: etcd
 sources:
   - https://github.com/bitnami/bitnami-docker-etcd
   - https://coreos.com/etcd/
-version: 8.0.3
+version: 8.1.0
--- a/bitnami/etcd/README.md
+++ b/bitnami/etcd/README.md
@@ -297,8 +297,8 @@ The command removes all the Kubernetes components associated with the chart and 
 
 | Name                 | Description                                                    | Value   |
 | -------------------- | -------------------------------------------------------------- | ------- |
-| `pdb.create`         | Enable/disable a Pod Disruption Budget creation                | `false` |
-| `pdb.minAvailable`   | Minimum number/percentage of pods that should remain scheduled | `1`     |
+| `pdb.create`         | Enable/disable a Pod Disruption Budget creation                | `true` |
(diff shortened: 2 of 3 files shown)


miguelaeh (commented):
IMHO this change shouldn't be considered a patch, since upgrading a release will set the `pdb.create` to `true` in case you don't have it overridden in your `values.yaml` as well as the `pdb.minAvailable`.
Maybe a major change is too much, but even though it is not a feature addition, I think a minor bump would be better.
WDYT @carrodher ?

cdemi:
I think it makes sense to upgrade the minor yes

carrodher:
I think a minor is better yes, too much for a major

bitnami/etcd/Chart.yaml
```diff
@@ -24,4 +24,4 @@ name: etcd
 sources:
   - https://github.com/bitnami/bitnami-docker-etcd
   - https://coreos.com/etcd/
-version: 8.0.2
+version: 8.0.3
```
carrodher, line 27:
```suggestion
version: 8.1.0
```

miguelaeh (approved):
LGTM!

PR #9939, opened 2022-04-26
[bitnami/*] feat: :sparkles: Add vib-publish pipeline

javsalgar:

### Description of the change

This PR adds the Github pipeline `vib-publish`, which will verify and publish the WordPress chart (first chart) using VIB (VMware Image Builder)

### Applicable issues

  - fixes #

### Checklist


.github/workflows/vib-publish.yaml
```diff
@@ -0,0 +1,113 @@
+name: 'VIB Publish'
```
recena, line 1: In order to follow [the convention](https://github.com/bitnami/charts/blob/master/.github/workflows/vib-verify.yaml#L1) `name: 'VIB'`

.github/workflows/vib-publish.yaml
```diff
@@ -0,0 +1,113 @@
...
+    branches:
+      - master
+    paths:
+      - 'bitnami/wordpress/Chart.yaml'
+env:
+  CSP_API_URL: https://console.cloud.vmware.com
+  CSP_API_TOKEN: ${{ secrets.CSP_API_TOKEN }}
+  VIB_PUBLIC_URL: https://cp.bromelia.vmware.com
+jobs:
+  get-chart:
```
recena, line 13: Do we really need this job for those changes landing in `stable` branch? I think we can guarantee that only one Helm Chart has been modified. Am I wrong?

.github/workflows/vib-publish.yaml
```diff
@@ -0,0 +1,113 @@
...
+    if: ${{ needs.get-chart.outputs.result == 'ok' }}
+    environment: vmware-image-builder
+    name: Publish
+    steps:
+      - uses: actions/checkout@v2
+        name: Checkout Repository
+        with:
+          path: charts
+      - uses: vmware-labs/vmware-image-builder-action@main
+        name: Verify and publish ${{ needs.get-chart.outputs.chart }}
```
recena, line 69: I would say just `Publish`. It is like `mvn package`, it involves prior phases.

.github/workflows/vib-publish.yaml
aruiz14: I noticed that change and hence renamed it to this from `vib-publish`, but then noticed that would create 2 duplicated entries at the [Actions](https://github.com/bitnami/charts/actions) page (note there are already 3: 1 for vib-verify.yaml workflow + the 2 extra workflows).
Do you still think we should name this `VIB` only? Maybe we could name them `VIB - Publish` and `VIB - PR Verify`?

.github/workflows/vib-publish.yaml
aruiz14: I don't think it would be so obvious that a `Publish` pipeline would also verify, unlike `package` phase, which is indeed implicit.
In fact, this is our choice to verify everything again, but we could rely on the success of PR Verify to avoid verifying everything again.

.github/workflows/vib-publish.yaml
recena: @aruiz14 I like your proposal:

- `VIB - PR verify` for verifying the contribution
- `VIB - Publish` for producing final bits and publishing them

.github/workflows/vib-publish.yaml
recena: It was not obvious the first time when I used Maven 😄 but it makes sense.

Up to you, I don't have a strong thought here.

.github/workflows/vib-publish.yaml
```diff
@@ -0,0 +1,113 @@
...
+        with:
+          pipeline: ${{ needs.get-chart.outputs.chart }}/vib-publish.json
+          config: charts/.vib/
+        env:
+          TARGET_PLATFORM: 91d398a2-25c4-4cda-8732-75a3cfc179a1
```
joancafom, line 74:
```suggestion
          VIB_ENV_TARGET_PLATFORM: 91d398a2-25c4-4cda-8732-75a3cfc179a1
```

joancafom: Additionally, I think we should be using the same approach we follow for `vib-verify` now:

1. Do not hardcode the value for `target-platform` here and use a secret instead. That way we prevent adding commits just to simply update the platform.
2. Use TKG by default and add another variable for assets incompatible with this platform

https://github.com/bitnami/charts/blob/5571b65f6b54f74097792c161a47566f755cc01d/.github/workflows/vib-verify.yaml#L93-L96

aruiz14: > That way we prevent adding commits just to simply update the platform.

The downside of using secrets is that it's not easy to retrieve the content of the secret once it's defined. Also, we could argue if that's actually a pro and not a con 😄 (configuration as code)
Having a way of tracing why the target-platform and follow the standard code review process is a feature in my opinion.

aruiz14: > Use TKG by default and add another variable for assets incompatible with this platform
We cannot follow this approach, since we can only launch 1 pipeline (otherwise the publish phase would be executed twice). In addition, we would also verifying twice something that was already checked during PR verify.

.github/workflows/vib-publish.yaml
```diff
@@ -0,0 +1,113 @@
+name: 'VIB - Publish'
```
aruiz14, line 1:
```suggestion
name: 'VIB Publish'
```

.github/workflows/vib-publish.yaml
aruiz14: > Additionally, I think we should be using the same approach we follow for `vib-verify` now:
> 
>     2. Use TKG by default and add another variable for assets incompatible with this platform
> 
> 
> https://github.com/bitnami/charts/blob/5571b65f6b54f74097792c161a47566f755cc01d/.github/workflows/vib-verify.yaml#L93-L96

@jsalmeron I think you didn't cover this.

aruiz14 (approved):

.github/workflows/vib-publish.yaml
```diff
@@ -0,0 +1,114 @@
...
+        with:
+          pipeline: ${{ needs.get-chart.outputs.chart }}/vib-publish.json
+          config: charts/.vib/
+        env:
+          VIB_ENV_TARGET_PLATFORM: 7ddab896-2e4e-4d58-a501-f79897eba3a0
+          VIB_ENV_ALTERNATIVE_TARGET_PLATFORM: 91d398a2-25c4-4cda-8732-75a3cfc179a1
+          VIB_ENV_S3_URL: ${{ secrets.S3_URL }}/bitnami
+          VIB_ENV_S3_USERNAME: ${{ secrets.S3_USERNAME }}
+          VIB_ENV_S3_PASSWORD: ${{ secrets.S3_PASSWORD }}
+          VIB_ENV_S3_ROLE: ${{ secrets.S3_ROLE }}
```
joancafom, line 79: I can't see that these secrets are currently defined. Could you please make sure they are before merging?

aruiz14 (approved):

joancafom (approved):

PR #9961, opened 2022-04-28
[bitnami/kibana] Fix Kibana tests that started failing due to a major release

superaleks:

### Description of the change
There was a major Kibana release which forces the CSP which is a problem for Cypress tests. We need to disable CSP so that the tests will run without issues.
Also, there are small changes in the UI which the tests need to adapt to. 

This is tested on the fork with the [following executions.](https://github.com/superaleks/charts/runs/6206267985?check_suite_focus=true)
### Checklist

### DIFF
--- a/.vib/kibana/cypress/cypress/integration/kibana_spec.js
+++ b/.vib/kibana/cypress/cypress/integration/kibana_spec.js
@@ -1,68 +1,58 @@
 /// <reference types="cypress" />
 
-import {
-  random,
-  closeThePopups,
-  skipTheWelcomeScreen
-} from './utils'
+import { random, closeThePopups, skipTheWelcomeScreen } from './utils';
 
 before(() => {
-  skipTheWelcomeScreen()
-})
+  skipTheWelcomeScreen();
+});
 
 it('allows uploading a file', () => {
-  cy.get('a[data-test-subj="uploadFile"]').click()
-  cy.get(
-    'input#filePicker',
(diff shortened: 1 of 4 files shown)


superaleks:
I'll do this from the fork.

.vib/kibana/cypress/cypress/integration/kibana_spec.js
```diff
@@ -1,68 +1,58 @@
...
-  cy.get('[title*="Index pattern created is complete"]').should(
-    'be.visible',
-  );
+      `${td.testIndexName}.${random}`
+    );
+  });
+  cy.get('[data-test-subj="dataVisualizerFileImportButton"]').click();
+  cy.get('[title*="Index created"]').scrollIntoView().should('be.visible');
+  cy.get('[title*="Data uploaded"').scrollIntoView().should('be.visible');
+  cy.get('[title*="Index created"]').should('be.visible');
```
joancafom, line 26: This assertion is the same one as line `24`. Looking at the old code, maybe you meant `Index pattern created`  instead?

superaleks: Good catch, it should actually say `Data view created`.
I've modified the assertions and removed the extra `should('be.visible')` assertions. Please check now.

migruiz4 (approved):
LGTM!

joancafom (approved):

PR #9985, opened 2022-04-29
[bitnami/wordpress] Update the tests to follow the test strategy

superaleks:
### Description of the change

It has been observed that the WordPress test spec is not completely adherent to the test strategy. It was the first spec to ever be published to GitHub. Since we have learned a lot and defined a test strategy after that, this spec needs to be adjusted and improved. 
Changes

- Removed some test cases 
- Changed the loading of test data
- Tweaked `runtime_parameters` accordingly 

### Test results

[Please check here.](https://github.com/superaleks/charts/pull/10) - PASS

### Benefits

Fewer maintenance issues, and code uniformity across assets. 

### Possible drawbacks

Reducing the number of tests also reduces test coverage. 

### Checklist

 `README.md` using [readme-generator-for-helm](https://github.com/bitnami-labs/readme-generator-for-helm)


.vib/wordpress/cypress/cypress/fixtures/admins.json
```diff
@@ -0,0 +1,4 @@
+{
+  "wordpressFirstName": "TestName",
+  "wordpressLastName": "TestLastName"
```
joancafom, line 3: `admins` should define an array of admin users, so that if the tests are ever expanded, one can add another entry:

```suggestion
  "defaultAdmin": {
    "firstName": "TestName",
    "lastName": "TestLastName"
  }
```

.vib/wordpress/cypress/cypress/fixtures/posts.json
```diff
@@ -3,4 +3,4 @@
   "content": "Welcome to WordPress.",
   "commenter": "A WordPress Commenter",
   "comment": "This is a test comment."
```
joancafom, line 5: Same here, it should all go under the same entry

.vib/wordpress/cypress/cypress/integration/wordpress_spec.js
```diff
@@ -2,88 +2,35 @@
...
-      cy.get('#author').type(user.username).should('have.value', user.username);
-      cy.get('#email').type(user.email).should('have.value', user.email);
-    });
-
-    cy.get('#submit').click();
+  cy.fixture('posts').then((post) => {
+    cy.contains('.wp-block-post-title', post.title);
+    cy.contains('.wp-block-post-content', post.content);
+    cy.get('.wp-block-post-title').click();
+    cy.get('.commentlist').should('have.length', 1);
```
joancafom, line 20: Would this work if the test is executed multiple times?

.vib/wordpress/cypress/cypress/integration/wordpress_spec.js
```diff
@@ -2,88 +2,35 @@
...
-
-    cy.fixture('user').then((user) => {
-      cy.get('#author').type(user.username).should('have.value', user.username);
-      cy.get('#email').type(user.email).should('have.value', user.email);
-    });
-
-    cy.get('#submit').click();
+  cy.fixture('posts').then((post) => {
+    cy.contains('.wp-block-post-title', post.title);
+    cy.contains('.wp-block-post-content', post.content);
```
joancafom, line 18: Hmm it took me some moments to understand this. As this is a test that deals with commenting, does it make sense to assert other entities (_Posts_) here? I'm not against it when there is a need, but I think that the title and content of the post are not relevant to commenting. WDYT?

.vib/wordpress/cypress/cypress/integration/wordpress_spec.js
```diff
@@ -2,88 +2,35 @@
...
-it('checks the blog name and user email configuration', () => {
-  cy.login();
-  cy.visit('/wp-admin/options-general.php');
-  cy.get('#new_admin_email').should(
-    'have.value',
-    Cypress.env('wordpressEmail')
+  cy.get('#submit').click();
+  cy.contains(
+    '.comment-awaiting-moderation',
+    'your comment will be visible after it has been approved'
```
joancafom, line 32: I don't you need a CSS class selector here. If the second string appears on screen, it should be enough to know it worked :)

.vib/wordpress/cypress/cypress/fixtures/admins.json
superaleks: True. I applied this suggestion, and I also simplified the test a bit since I noticed WordPress had another release with different behaviour from the SMTP product.
Since I saw this happens often, I made this test even more minimalistic.

.vib/wordpress/cypress/cypress/integration/wordpress_spec.js
superaleks: Yeah, it does! However I want to rethink this test, you have a good point.
It seems to be working because the comment is always added to a new post.
Since we are creating posts all the time, there is always one comment per post if that makes sense.

.vib/wordpress/cypress/cypress/integration/wordpress_spec.js
superaleks: I modified the test to focus on 'comments' and I just used the entry from 'posts' to select the proper post (`Hello world` one. Hope it's better now!

.vib/wordpress/cypress/cypress/integration/wordpress_spec.js
superaleks: Fixed.

.vib/wordpress/cypress/cypress/integration/wordpress_spec.js
FraPazGal: I'm not sure why we need to check the comments list here. If we are not checking whether its length increases after adding a comment at the end of the test, there is not really a reason to do so at the start right?

FraPazGal (commented):
Thanks for the PR! I left a couple of comments; please check when possible.

.vib/wordpress/cypress/cypress/integration/wordpress_spec.js
superaleks: We are adding a comment each time. 
We want to check if the comment is added properly. 
The length **doesn't** increase because every comment added **needs** to be approved by the moderator. So each time the comment is added it will appear to be the first one.

.vib/wordpress/cypress/cypress/support/commands.js
```diff
@@ -15,6 +14,8 @@ for (const command of ['click']) {
   });
 }
 
+const TYPE_DELAY = 100;
```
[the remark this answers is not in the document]
superaleks: Indeed! It was an error. Now I've fixed it.

FraPazGal (dismissed):
LGTM!

.vib/wordpress/cypress/cypress/fixtures/posts.json
superaleks: Done!

.vib/wordpress/cypress/cypress/fixtures/posts.json
```diff
@@ -0,0 +1,3 @@
+{
+  "title": "Hello world!"
```
joancafom, line 2: This should be an array of `posts`, in which one of them has that property.

.vib/wordpress/cypress/cypress/integration/wordpress_spec.js
```diff
@@ -2,88 +2,33 @@
...
+it('allows adding a comment to the sample blogpost', () => {
   cy.visit('/');
   cy.get('.wp-block-query');
   cy.get('.wp-block-post-title a').click();
-  cy.fixture('helloworld').then((hw) => {
-    cy.contains('.wp-block-post-title', hw.title);
-    cy.contains('.wp-block-post-content', hw.content);
+  cy.fixture('posts').then((post) => {
+    cy.contains('.wp-block-post-title', post.title).click();
   });
```
joancafom, line 18: This is not a hard requirement but just a thought that came to my mind. It doesn't really matter what post you comment in, so you don't really need to load a fixture here. Instead, you could simply get all the posts and click one of them.

.vib/wordpress/cypress/cypress/integration/wordpress_spec.js
superaleks: I think I made a mistake in the previous explanation. Let me try to explain it better
In this test, we first check for the presence of the Hello World comment in the Hello World post. Then we proceed to add our own comment which then goes to approval.
This test verifies if the Hello World comment is there before proceeding to add its own. I think it's a good way to check if the comment section is loaded properly, as well as a good way to check if the comment section works properly. UI for reference:

image

P.S. This test can be executed as many times as possible without issues, which was the original question.

.vib/wordpress/cypress/cypress/fixtures/posts.json
superaleks: Good point, on my way!

.vib/wordpress/cypress/cypress/integration/wordpress_spec.js
joancafom: nice, now I got you! thanks 👍

PR #9441, opened 2022-03-16
[bitnami/kubeapps] Chart standardised 

migruiz4:
**Description of the change**

This PR standardizes the bitnami/kubeapps chart to be in line with the rest of the catalog.

**Checklist**

- [X] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [x] Variables are documented in the values.yaml and added to the README.md using [readme-generator-for-helm](https://github.com/bitnami-labs/readme-generator-for-helm)
- [X] Title of the PR starts with chart name (e.g. [bitnami/<name_of_the_chart>])
- [X] All commits signed off and in agreement of [Developer Certificate of Origin (DCO)](https://github.com/bitnami/charts/blob/master/CONTRIBUTING.md#sign-your-work)

PR #10006, opened 2022-05-02
[bitnami/kubeapps] Values references postgresPassword used by Kubeapps.

absoludity:

### Description of the change
In the change at:
https://github.com/bitnami/charts/pull/9441/files#diff-0f1b69af1b7c9c48d9d4f43c74fd880405aec305533aa650e1c26d7a2d20c300R1976-R1988
the values of the Kubeapps chart switched to reference the non-postgres
user rather than the postgres user which Kubeapps currently uses

I believe Kubeapps could switch to the normal (unprivileged) user (see https://github.com/vmware-tanzu/kubeapps/issues/4643), but currently it is using the postgres user throughout, so this change just restores our values.yaml to reflect that after @jorgemoralespou noticed that our README now mentions the non-priv'd user password rather than setting the postgres password.

### Benefits

Better documentation.

### Applicable issues

  - reference #

### Checklist

### DIFF
--- a/bitnami/kubeapps/Chart.yaml
+++ b/bitnami/kubeapps/Chart.yaml
@@ -30,4 +30,4 @@ maintainers:
 name: kubeapps
 sources:
   - https://github.com/vmware-tanzu/kubeapps
-version: 8.0.15
+version: 8.0.16
--- a/bitnami/kubeapps/README.md
+++ b/bitnami/kubeapps/README.md
@@ -121,7 +121,7 @@ Once you have installed Kubeapps follow the [Getting Started Guide](https://gith
 | ------------------------------------------------ | ----------------------------------------------------------------------------------------- | ---------------------- |
 | `frontend.image.registry`                        | NGINX image registry                                                                      | `docker.io`            |
 | `frontend.image.repository`                      | NGINX image repository                                                                    | `bitnami/nginx`        |
-| `frontend.image.tag`                             | NGINX image tag (immutable tags are recommended)                                          | `1.21.6-debian-10-r62` |
+| `frontend.image.tag`                             | NGINX image tag (immutable tags are recommended)                                          | `1.21.6-debian-10-r93` |
 | `frontend.image.pullPolicy`                      | NGINX image pull policy                                                                   | `IfNotPresent`         |
 | `frontend.image.pullSecrets`                     | NGINX image pull secrets                                                                  | `[]`                   |
 | `frontend.image.debug`                           | Enable image debug mode                                                                   | `false`                |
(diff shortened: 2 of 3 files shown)


absoludity:
@migruiz4 Just adding you as a reviewer in case you had other reasons for changing the password in #9441 .

antgamdia:
From a Kubeapps' point of view, LGTM. Currently, as you mentioned, we are using the "root" user. In fact, the generated secret contains the data `postgres-password`, which represents the password of this root (aka `postgres`) user.

```
kubectl get secrets -n kubeapps kubeapps-postgresql -ojsonpath='{.data}'
{"postgres-password":"<redacted>"}
```

This PR is just to keep it consistent, but mid/longer term we should switch to avoid using this root `postgres` user (as we don't actually need it).

BTW, I've noticed another very minor typo in the doc, if you want to squeeze the change, it would be great. Otherwise, I can just send another PR:
> In https://github.com/bitnami/charts/blob/master/bitnami/kubeapps/README.md#serving-kubeapps-in-a-subpath,   `--set authProxy.additionalFlags=...` should be `  --set authProxy.extraFlags=...`
(

migruiz4 (dismissed):
Nice catch! Looks good to me, after the PostgreSQL `auth` parameters rename, I didn't notice kubeapps was using the `postgres` user and added the wrong value.

absoludity:
> BTW, I've noticed another very minor typo in the doc, if you want to squeeze the change, it would be great. Otherwise, I can just send another PR:
> 
> > In https://github.com/bitnami/charts/blob/master/bitnami/kubeapps/README.md#serving-kubeapps-in-a-subpath,   `--set authProxy.additionalFlags=...` should be `  --set authProxy.extraFlags=...`
> > (

Done.

absoludity:
It needs an extra review since I updated, but also the VIB/Verify seems to fail with a 500?

```
Warning: Environment variable VIB_ENV_TARGET_PLATFORM is set but is not used within pipeline kubeapps/vib-verify.json
Warning: Environment variable VIB_ENV_ALTERNATIVE_TARGET_PLATFORM is set but is not used within pipeline kubeapps/vib-verify.json
Error: Request failed with status code 500
```

migruiz4:
Hi @absoludity,

There seems to be an ongoing issue with the VIB / Verify check that causes the 500 status code. 
The issue can be workaround by setting an empty `verify.context` at `.vib/kubeapps/vib-verify.json`:

```json
{
  "phases": {
    "package": {
      "context": {
        "resources": {
          "url": "{SHA_ARCHIVE}",
          "path": "/bitnami/kubeapps"
        }
      },
      "actions": [
        {
          "action_id": "helm-package"
        },
        {
          "action_id": "helm-lint"
        }
      ]
    },
    "verify": {
      "context": {},
      "actions": [
        {
          "action_id": "trivy",
          "params": {
            "threshold": "CRITICAL",
            "vuln_type": ["OS"]
          }
        }
      ]
    }
  }
}

```

migruiz4:
No need to apply the workaround, the issue is already fixed 👍

absoludity:
> No need to apply the workaround, the issue is already fixed +1

Great, thanks @migruiz4 . So I think this just needs your approval re-applied (as I pushed the small change mentioned by @antgamdia yesterday) and we can land it?

migruiz4 (approved):
LGTM!

PR #10031, opened 2022-05-05
[bitnami/redis] Fix issue with duplicated lifecycles

### LINKED ISSUE #10027: redis lifecycleHooks does not work when sentinel enabled
### Name and Version

bitname/redis 16.8.9

Readme and values imply we can set lifecycle hooks to run but we can't because of duplicate lifecycle keys in the yaml

https://github.com/bitnami/charts/blob/master/bitnami/redis/templates/sentinel/statefulset.yaml#L99

image

is overridden due to a duplicate key later down here

https://github.com/bitnami/charts/blob/master/bitnami/redis/templates/sentinel/statefulset.yaml#L267

image

note the sentinel image has the same problem

image

image

```yaml
  sentinel:
    enabled: true
  replica:
    lifecycleHooks:
      postStart:
        exec:
          command: ["/bin/sh", "-c", "echo Hello from the postStart handler > $HOME/message"]
```

```yaml
...
          lifecycle:
            postStart:
              exec:
                command: ["/bin/sh", "-c", "echo Hello from the postStart handler > $HOME/message"]
            preStop:
              exec:
                command:
                  - /bin/bash
                  - -c
                  - /opt/bitnami/scripts/start-scripts/prestop-redis.sh
...
```

### What do you see instead?

```yaml
...
          lifecycle:
            preStop:
              exec:
                command:
                  - /bin/bash
                  - -c
                  - /opt/bitnami/scripts/start-scripts/prestop-redis.sh
...
```

carrodher:

### Description of the change

As described at https://github.com/bitnami/charts/issues/10027, the `lifecycle` is duplicated in the sentinel template. It works fine if the user doesn't provide a custom `lifecycle` since the one set in the file is going to be taken into account; but if the user sets a custom one both are defined at the same time.

With those changes, by default the one defined in the template is going to be used unless the user sets its own `lifecycle`

### Benefits

Users will be able to set a custom `lifecycle` overriding the default one.

### Applicable issues

  - fixes #10027

### Checklist

### DIFF
--- a/bitnami/redis/Chart.yaml
+++ b/bitnami/redis/Chart.yaml
@@ -24,4 +24,4 @@ maintainers:
 name: redis
 sources:
   - https://github.com/bitnami/bitnami-docker-redis
-version: 16.8.9
+version: 16.8.10
--- a/bitnami/redis/templates/sentinel/statefulset.yaml
+++ b/bitnami/redis/templates/sentinel/statefulset.yaml
@@ -97,6 +97,14 @@ spec:
           {{- if not .Values.diagnosticMode.enabled }}
           {{- if .Values.replica.lifecycleHooks }}
           lifecycle: {{- include "common.tplvalues.render" (dict "value" .Values.replica.lifecycleHooks "context" $) | nindent 12 }}
+          {{- else }}
+          lifecycle:
+            preStop:
+              exec:
+                command:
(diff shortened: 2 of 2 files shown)


migruiz4 (approved):
LGTM!

PR #10039, opened 2022-05-05
[bitnami/redmine] Add Redmine Helm Chart Test

superaleks:

**Description of the change**
Increasing the test coverage of the Redmie Helm Chart by adding Cypress and Goss tests. 

**Benefits**

Ensuring higher quality of the Helm charts

**Possible drawbacks**

**Test results** - IN PROGRESS

9 consecutive pipelines test is in progress
[Please see them in the history here](https://github.com/superaleks/charts/pull/16)

**Checklist**


.vib/redmine/cypress/cypress/plugins/index.js
```diff
@@ -0,0 +1,22 @@
+/// <reference types="cypress" />
```
FraPazGal, line 1: This file can be removed.

.vib/redmine/cypress/cypress/fixtures/issues.json
```diff
@@ -0,0 +1,7 @@
+{
+  "newIssue": {
+    "tracker": "Bug",
```
FraPazGal, line 3: It seems we are not using `newIssue.tracker`.

.vib/redmine/cypress/cypress/integration/redmine_spec.js
```diff
@@ -0,0 +1,81 @@
...
+  cy.contains('Issues').click();
+  cy.get('.icon-add').click();
+  cy.fixture('issues').then((issue) => {
+    cy.get('#issue_subject').type(`${issue.newIssue.subject}.${random}`);
+    cy.get('#issue_description').type(issue.newIssue.description);
+    cy.get('input[type="file"]').selectFile('cypress/fixtures/issues.json');
+  });
+  cy.get('[name="commit"]').click();
+  cy.contains('#flash_notice', 'created');
+  cy.contains('issues.json');
```
FraPazGal, line 48: Would this `contains` succeed if the previous step failed? My concern is that if it stays on the previous page, it will detect the string `issues.json` from the uploaded file to be committed.

.vib/redmine/cypress/cypress/fixtures/roles.json
```diff
@@ -0,0 +1,5 @@
+{
```
FraPazGal, line 1: I may be missing something, are we using this fixture somewhere?

.vib/redmine/cypress/cypress/integration/redmine_spec.js
```diff
@@ -0,0 +1,81 @@
...
+  });
+  cy.contains('[type="submit"]', 'Save').click({ force: true });
+  cy.contains('Successful update.');
+});
+
+it('allows admin to verify and change application configuration', () => {
+  cy.login();
+  cy.visit('/admin');
+  cy.contains('Settings').click();
+  cy.get('#settings_app_title').clear().type(`VMWare.${random}`);
```
FraPazGal, line 74: Should we obtain this value from a `settings.json` fixture?

.vib/redmine/cypress/cypress/integration/redmine_spec.js
```diff
@@ -0,0 +1,81 @@
...
+  cy.contains('input', 'Save').click();
+  cy.contains('#flash_notice', 'Successful update.');
+  cy.visit('/admin/info');
+  cy.contains('Mysql');
+  cy.contains('production');
+  cy.contains('Git');
```
FraPazGal, line 80: I'm not familiar with the asset but shouldn't we use `const` variables for these values?

FraPazGal (changes requested):
Thanks for the PR @superaleks, could you please check my comments?

.vib/redmine/cypress/cypress/fixtures/issues.json
superaleks: True. I added the test data thinking I'll need it. Then I forgot to remove it. Now it's fixed.

.vib/redmine/cypress/cypress/integration/redmine_spec.js
superaleks: I've tried this and this is my conclusion
We have two assertions here:
```
cy.contains('#flash_notice', 'created');
cy.contains('issues.json');
```
If the file is not uploaded properly assertion 2 will fail since it will not find the `issues.json` on the page. 
If the bug is not created, assertion 1 will fail since the bug will not be created. 
So, if the bug is not committed we will not arrive at the file upload test since the first assertion will fail beforehand. This, to me, looks like expected behaviour and doesn't allow for the corner case you worry about.

.vib/redmine/cypress/cypress/integration/redmine_spec.js
superaleks: True. I've moved this to the fixture.

.vib/redmine/cypress/cypress/integration/redmine_spec.js
superaleks: After your comment, I was indecisive about using `const` or fixtures.
Since we already have the `settings.json` file, I think it makes more sense to have this there. I've applied this change as well.

FraPazGal: Maybe a bit unrelated to this, is it necessary to check the page contains these 3 elements? I could be missing something as I'm not familiar with redmine, but we already checked the "Successful update" notice and these elements look like were already as they are before the update.

.vib/redmine/cypress/cypress/fixtures/smtps.json
```diff
@@ -0,0 +1,7 @@
+{
+  "newSMTP": {
+    "smtpEmailAddress": "redmine@vmware.net",
```
joancafom, line 3: I'd rather use a truly fake one, just in case. e.g. `redmine@example.com`

.vib/redmine/cypress/cypress/fixtures/users.json
```diff
@@ -0,0 +1,9 @@
+{
+  "newUser": {
+    "userName": "JaneJessica",
+    "password": "someComplicatedPass12345!",
+    "firstName": "Jane Doe Jess",
+    "lastName": "Doe",
+    "email": "thisisanemail"
```
joancafom, line 7: Is this a valid email?

.vib/redmine/cypress/cypress/integration/redmine_spec.js
```diff
@@ -0,0 +1,86 @@
+/// <reference types="cypress" />
+import { random } from './utils';
+
+it('allows admin to login/logout', () => {
+  cy.login();
+  cy.get('#flash_error').should('not.exist');
```
joancafom, line 6: When possible, rather check the error message than the element's id, e.g.:

```
cy.contains('Invalid user').should('not.exist');
```

.vib/redmine/cypress/cypress/integration/redmine_spec.js
```diff
@@ -0,0 +1,86 @@
...
+  cy.visit('account/register');
+  cy.fixture('users').then((user) => {
+    cy.get('#user_login').type(`${user.newUser.userName}.${random}`);
+    cy.get('#user_password').type(`${user.newUser.password}.${random}`);
+    cy.get('#user_password_confirmation').type(
+      `${user.newUser.password}.${random}`
+    );
+    cy.get('#user_firstname').type(`${user.newUser.firstName}.${random}`);
+    cy.get('#user_lastname').type(`${user.newUser.lastName}.${random}`);
+    cy.get('#user_mail').type(`${user.newUser.email}.${random}@email.com`);
```
joancafom, line 21: Now I understand why the domain was not in the fixtures. Isn't it easier something like `${random}${user.newUser.email}`, though?

.vib/redmine/cypress/cypress/integration/redmine_spec.js
```diff
@@ -0,0 +1,86 @@
...
+
+it('allows admin to create a project and an issue with file uploaded', () => {
+  cy.login();
+  cy.visit('projects');
+  cy.contains('New project').click();
+  cy.fixture('projects').then((project) => {
+    cy.get('#project_name').type(`${project.newProject.name}.${random}`);
+    cy.get('#project_description').type(project.newProject.description);
+    cy.get('[name="commit"]').click();
+    cy.contains('Successful creation.');
```
joancafom, line 35:
```suggestion
    cy.contains('Successful creation');
```

.vib/redmine/cypress/cypress/integration/redmine_spec.js
```diff
@@ -0,0 +1,86 @@
...
+
+it('allows admin to login/logout', () => {
+  cy.login();
+  cy.get('#flash_error').should('not.exist');
+  cy.contains('Sign out').click();
+  cy.contains('Sign in');
+});
+
+it('allows user to register', () => {
+  cy.visit('account/register');
```
joancafom, line 12: You're interchangeably using `cy.visit` with a leading `/` and without it. Use only one approach:

```suggestion
  cy.visit('/account/register');
```

.vib/redmine/cypress/cypress/integration/redmine_spec.js
```diff
@@ -0,0 +1,86 @@
...
+    cy.get('#user_lastname').type(`${user.newUser.lastName}.${random}`);
+    cy.get('#user_mail').type(`${user.newUser.email}.${random}@email.com`);
+  });
+  cy.get('[type="submit"]').click();
+  cy.contains('Your account was created');
+});
+
+it('allows admin to create a project and an issue with file uploaded', () => {
+  cy.login();
+  cy.visit('projects');
```
joancafom, line 29:
```suggestion
  cy.visit('/projects');
```

.vib/redmine/cypress/cypress/integration/redmine_spec.js
```diff
@@ -0,0 +1,86 @@
...
+    cy.get('#issue_subject').type(`${issue.newIssue.subject}.${random}`);
+    cy.get('#issue_description').type(issue.newIssue.description);
+    cy.get('input[type="file"]').selectFile('cypress/fixtures/issues.json');
+  });
+  cy.get('[name="commit"]').click();
+  cy.contains('#flash_notice', 'created');
+  cy.contains('issues.json');
+});
+
+it('allows admin to check smtp configuration', () => {
```
joancafom, line 51: This is not 'checking' but rather 'setting', right?

.vib/redmine/cypress/cypress/integration/redmine_spec.js
```diff
@@ -0,0 +1,86 @@
...
+    cy.get('#settings_emails_header')
+      .scrollIntoView()
+      .clear()
+      .type(`${smtp.newSMTP.header}.${random}`);
+    cy.get('#settings_emails_footer')
+      .clear()
+      .type(`${smtp.newSMTP.footer}.${random}`);
+  });
+  cy.contains('[type="submit"]', 'Save').click({ force: true });
+  cy.contains('Successful update.');
```
joancafom, line 67:
```suggestion
  cy.contains('Successful update');
```

.vib/redmine/cypress/cypress/integration/redmine_spec.js
```diff
@@ -0,0 +1,86 @@
...
+  cy.login();
+  cy.visit('/admin');
+  cy.contains('Settings').click();
+  cy.fixture('settings').then((setting) => {
+    cy.get('#settings_app_title')
+      .clear()
+      .type(`${setting.newSetting.appTitle}.${random}`);
+
+    cy.contains('input', 'Save').click();
+    cy.contains('#flash_notice', 'Successful update.');
```
joancafom, line 80:
```suggestion
    cy.contains('#flash_notice', 'Successful update');
```

.vib/redmine/cypress/cypress/integration/redmine_spec.js
```diff
@@ -0,0 +1,86 @@
...
+    cy.visit('/admin/info');
+    cy.contains(setting.newSetting.database);
+    cy.contains(setting.newSetting.environment);
+    cy.contains(setting.newSetting.versionControl);
```
joancafom, line 84: You are not updating those fields in the previous step, how can the new settings be applied then?

.vib/redmine/cypress/cypress/integration/redmine_spec.js
superaleks: The intention of the test is to **verify**(observe) the application configuration, as well as modify it. 
When we modify the configuration, we go to the info page which is under the admin tab. 
The tests are taken a look at this page:

redmine

Since they are showing important components of this application (database, version control) I thought it is relevant to have them.

.vib/redmine/cypress/cypress/fixtures/smtps.json
superaleks: Why, what is the difference?

.vib/redmine/cypress/cypress/integration/redmine_spec.js
superaleks: Well, it's doing both.
The SMTP input form doesn't appear if the SMTP is not enabled in the configuration (runtime parameters).
The admin can change the email settings, but cannot set them.
I think this will be clearer if I rename the tests to
`allows admins to verify SMTP is enabled and modify it`

.vib/redmine/cypress/cypress/integration/redmine_spec.js
superaleks: In the case of error messages within consistent UI error elements, I disagree. Especially for login.
Imagine that the login test fails for something **other** than username and password being incorrect.
The error message will be different. 
But, the UI element (flash error) showing the error message will still be the same since it's used consistently in the application.
If we limit ourselves to discovering error messages only, we will catch only one type of issue.
WDTY? 

Screenshot 2022-05-11 at 11 57 59

.vib/redmine/cypress/cypress/integration/redmine_spec.js
superaleks: The test is named like this

`it('allows admin to modify and observe application configuration'`

This is because I am not only modifying the application configuration but also verifying the existing one. [Please check a more elaborate explanation I gave to Fran in a separate thread. ](https://github.com/bitnami/charts/pull/10039#discussion_r870100005)

.vib/redmine/cypress/cypress/integration/redmine_spec.js
superaleks: Yeah, fair enough! I fixed it, which means both of the threads related to emails are resolved.

.vib/redmine/cypress/cypress/fixtures/users.json
superaleks: Made the change requested in a separate comment.

.vib/redmine/cypress/cypress/fixtures/smtps.json
joancafom: I am usually not very into using real domains, just in case haha but feel free to leave it as it is. It was just a personal preference

.vib/redmine/cypress/cypress/integration/redmine_spec.js
joancafom: I agree with you here! My concern was that text is less prone to change, but I totally see your point

.vib/redmine/cypress/cypress/integration/redmine_spec.js
joancafom: Ok, now I see it. However, when you first read this it gives the impression that those are `newSetting`s to be updated in the application. I'd move them to another fixture, say `defaultSetting`.

.vib/redmine/goss/goss.yaml
```diff
@@ -0,0 +1,40 @@
...
+    group: root
+    contains:
+      - "postgresql"
+      - "redmine"
+  /bitnami/redmine/files:
+    mode: "0775"
+    filetype: directory
+    exists: true
+    group: root
+  /opt/bitnami/mysql:
```
joancafom, line 31: Not a big deal, but there are 6 tests here

joancafom (approved):
LGTM

PR #10042, opened 2022-05-05
[bitnami/rabbitmq] Use short DNS names for clustering

VannTen:
### Benefits
In a kubernetes pod, /etc/resolv.conf default setup is:

```
search <namespace>.svc.<clusterdomain> svc.<clusterdomain> <clusterdomain> <other searchs domain ...>
options ndots:5
```

using pod.service.namespace.svc.clusterdomain for clustering is less
predictible than using short names:
If clusterdomain = cluster.local, local search will be bypassed, but
if it does not contain '.', it will not, thus performing a DNS query for
each searchdomains.
Regardless of the clusterdomain, the short name (pod.service) will only
need one DNS query, matching with the first search domain.

This also make the manifest simpler.

### Possible drawbacks

N/A

### Checklist


fmulero (dismissed):
Thanks @VannTen.  LGTM, I'll run some extra test, if everything goes well, your changes will be merged soon

VannTen:
Anything missing from my side ?

fmulero:
I think the problem we are facing is related to the fact that we are creating a cluster with the FQDN names (as it was in previous version), running an upgrade with the short names. To be honest I couldn't test it, did you face any issue when you applied these changes in your environment?

VannTen:
> I think the problem we are facing is related to the fact that we are creating a cluster with the FQDN names (as it was in previous version), running an upgrade with the short names. To be honest I couldn't test it, did you face any issue when you applied these changes in your environment?

Sorry for taking a long time to answer.

I just had the occasion to test the upgrade. Initially I thought this
was due to the fact that we're not using persistence for our rabbitmq,
but I was able to perform the upgrade with it enabled without problems
on my environment (k8s 1.19.9).

I tested the following values:
```yaml
replicaCount: 3
clusterDomain: <custom domain>
persistence:
  enabled: false
image:
  debug: true
```
and 
```yaml
replicaCount: 3
clusterDomain: <custom domain>
persistence:
  enabled: false
image:
  debug: true
```
doing a fresh install from b42600d13 and upgrading to 886f258ef

Would you have more details on the problem you're facing on your side ?

fmulero (dismissed):

mjfitzge:
This PR wipes rabbitmq when applied to existing clusters. It changes the volume names within rabbitmq so all existing data is lost. This should be rolled back or come with a massive warning/major version release to prevent people losing data they care about. Luckily for us this was caught in a development upgrade... others won't be so lucky. We tested on another development cluster with the version before this PR and had no data loss.

fmulero:
Hi @mjfitzge. Thanks for raising this issue so quick.

This change causes that an existing cluster  upgraded to this version looks like it's been deleted, but if you come back to previous version, queues and persistent messages are still there, so I' m not sure what you mean with:

> It changes the volume names within rabbitmq so all existing data is lost

I am trying to find a workaround for who is facing this issue, so any help is welcome

dyvol777:
@fmulero we lost all data in this upgrade. But we had `forceboot` option enabled, so this (with changing data folder) could be a reason of this.

arthurk:
This upgrade also broke our rabbitmq and all data was missing (luckily in staging). We reverted back to the previous version which fixed it

PR #10047, opened 2022-05-05
[bitnami/redis] Allow deploying multiple masters

eduardobr:
Signed-off-by: Eduardo Semprebon <eduardobr@gmail.com>

### Description of the change

When configured with Kubernetes features like internalTrafficPolicy (1.22+) and podAntiAffinity, we can afford an interesting setup with multiple masters, for example:
- Set master.service.internalTrafficPolicy=Local
- Create a podAntiAffinity to never schedule more than 1 master per node.

As a result replicas in each node will see only a single master inside the node.

### Benefits

- Cheaper and faster traffic as master <-> replica and writer <-> master communication will happen only within nodes.
- High availability without limitations of Redis Cluster, because with multiple masters we can create a PodDisruptionBudget to ensure at least one master is available all the time.

Cache writers can simply be deployed across all nodes and they will only see the master of own node with master.service.internalTrafficPolicy=Local.

In summary, we can replicate a whole stack of Writer Application + master + replicas per node.
Consumers could be optionally restricted to replicas of the own node or all replicas of the cluster.

### Possible drawbacks

Requires specific configuration in other properties of the values file to work well, as mentioned above.

### Applicable issues

N/A

### Additional information

- Ideally the property should be called `master.replicaCount` instead of `master.count`, but that would confuse the term replica from statefulset with redis replicas of master.
- When deploying a single master, there's no need to go through a non-headless service, that's why the conditional in master statefulset.yaml

### Checklist


alemorcuq:
Is this supported by Redis? How does intra-master communication work?

alemorcuq:
Sorry for the late follow up.

There's something I don't understand: if the masters are independent and unable to see each other, how do we achieve full replication? How do we make sure what's written to one master gets to other masters?

alemorcuq:
Sounds interesting, but it completely changes the way the chart currently works so I'm not sure about changing the behaviour by default. What are your thoughts on this @javsalgar ?

alemorcuq:
Forget about my last message, I didn't notice the replica field is currently hardcoded. I think this would be a nice addition. Could you please add a section to the README explaining how this works when # masters > 1 and that this is currently an experimental feature?

eduardobr:
Thanks for reviewing @alemorcuq 
I've included in README an explanation and recommendation for when master.count > 1

Additionally, fixed a text about `The replicas will be read-replicas of the master`. This is just the default value of redis and replicas can be writable if desired by setting `replica-read-only no` in redis config.

alemorcuq:
The DCO is missing, could you please sign it?

eduardobr:
@alemorcuq, when following the instructions from the DCO Details to signoff all commits using rebase, something went bad.
I'll take a look and revert those hundreds of commits

alemorcuq:
> @alemorcuq, when following the instructions from the DCO Details to signoff all commits using rebase, something went bad. I'll take a look and revert those hundreds of commits

The diff looks fine, but the DCO is still missing.

eduardobr:
@alemorcuq I have rewritten the history with a single commit, signed off. DCO should pass now.

PR #10057, opened 2022-05-06
[bitnami/kibana] Kibana Helm Chart Test improvement

superaleks:
### Description of the change

Move the test data to appropriate fixture files.

### Test results

[Please check here](https://github.com/superaleks/charts/pull/17) - PASS - (Trivy failing, proved to be non relevant)

### Checklist


.vib/kibana/cypress/cypress/fixtures/testdata.json
```diff
@@ -1,9 +1,3 @@
 {
-  "dashboardTitle": "Test_dashboard",
-  "dashboardDescription": "Test_dashboard_description",
-  "visualizationTitle": "Test_visualisation",
-  "testIndexName": "test_index",
-  "remoteClusterName":"remoteCluster",
-  "remoteCluster":"localhost:90",
-  "workpadName":"testWorkpad"
+  "workpadName": "testWorkpad"
```
joancafom, line 2: Are you still using this file? I don't think so

.vib/kibana/cypress/cypress/integration/prepare_app_state.js
```diff
@@ -0,0 +1,18 @@
+/// <reference types="cypress" />
```
joancafom, line 1: Although we are using this `prepare_app_state.js` file in SCDF, the use case there is more complex in my opinion, and does actually create some resources that prepare the app state.

In this case, I regard these two functions simply as two helpers/**utils** that every test uses. They don't really change or prepare the app state, so in my head it makes sense to place them in `utils.js`

superaleks: I discussed this with @FraPazGal  before creating this task when we did SCDF. 
What he was suggesting is that the `utils.js` should be used for methods that serve as tools for our tests. This is clear in the case of random numbers generators etc.
When it comes to simple browser interactions like the ones I have now, the distinction is not so easy.

I agreed it will be cleaner to segregate the application state preparation methods from utils (like random). 
@FraPazGal  WDYT?

FraPazGal: For SCDF, the distinction was indeed clearer. There, we had some methods that were testing some functionalities of the asset itself. In this case, I would say that the `skipTheWelcomeScreen` and `closeThePopups` can be seen as auxiliary tools for our tests. 

Having said that, I do not have any strong thoughts one way or the other as long as we clearly define the use of each file.

superaleks: As agreed, added the mentioned method to the `utils.js`.

.vib/kibana/cypress/cypress/fixtures/testdata.json
superaleks: True! I forgot to delete it.

FraPazGal (dismissed):
LGTM!

joancafom (approved):
LGTM

joancafom (approved):

FraPazGal (approved):

PR #10070, opened 2022-05-08
[bitnami/minio] Allow user provisioning from existing Kubernetes secrets

justusbunsi:
Signed-off-by: Steven Kriegler <61625851+justusbunsi@users.noreply.github.com>

### Description of the change

This PR introduces the ability to take user configurations from Kubernetes secrets into account during MinIO user provisioning.

### Benefits

It is already possible to provision MinIO users through the Chart but sensitive data like passwords can only be defined in plain text inside `values.yaml`. Referring to a Kubernetes secret containing the configuration for a MinIO user allows for storing the user provisioning part of this Helm Chart in Git repositories to go the GitOps way.

This implementation allows using one Kubernetes secret providing all desired users or one Kubernetes secret for each user.

### Additional information

This introduces a new setting `provisioning.usersExistingSecrets` representing an array of names of existing Kubernetes secrets. These secrets are mounted as files into the provisioning job container where they are being parsed and their values processed by MinIO CLI.

There should be no breaking change with this implementation as the new functionality is only active if administrators pro-actively specify `usersExistingSecrets` for user configuration. Otherwise, the Chart behaves the same.
Administrators can use both `provisioning.users` and `provisioning.usersExistingSecrets` together to have as much flexibility as possible.

### Checklist


bitnami/minio/templates/provisioning-job.yaml
```diff
@@ -96,8 +96,13 @@ spec:
               mc admin policy add {{ $minioAlias }} {{ $policy.name }} /etc/ilm/policy-{{ $policy.name }}.json;
               {{- end }}
 
-              {{- range $user := .Values.provisioning.users }}
-              mc admin user add {{ $minioAlias }} {{ $user.username }} {{ $user.password }};
+              {{- range $idx, $user := .Values.provisioning.users }}
+              {{- $userPassword := default "" $user.password }}
```
justusbunsi, line 100: TBH, I am not sure if the empty default value is problematic. It would only be used in case neither inline password nor secret reference is specified. Maybe in that case the release should throw an error? Happy for opinions on that.

justusbunsi: Possible options instead of empty value:

- `hasKey` check to ensure password field is set if new approach is not used
- create random password
- let `helm install` fail with proper error message (as mentioned above)

FraPazGal (commented):
Thanks for the PR @justusbunsi! I agree we should allow for providing the users' passwords with an existing secret but I have some concerns with this approach.

I think we should avoid mixing data from `.Values.provisioning.users` and a secret containing the passwords to create the final `mc admin user add` command. It can be confusing to coordinate the data at `values.yaml` with the one in a secret. 

Given that we are creating the users using a bash script, we could pass all the info of each user as data of the secret and parse it in bash before calling `mc admin`. Following this approach, we would create a key-value pair for each user with a similar structure as the one used in the values:

```
# Having the following values
  users:
    - username: test-username
      password: test-password
      disabled: false
      policies:
        - readwrite
        - consoleAdmin
        - diagnostics
      setPolicies: false
----
# The equivalent in a `secret-users.yaml` would be:
data:
  username1: password;false;readwrite,consoleAdmin,diagnostics;false
```
It is true that we'll need to create the parsing logic in the bash script and that we are losing some flexibility without a `existingSecret` and `existingSecretKey` per user, but this will make for a more transparent solution where a user can replicate the whole `.Values.provisioning.users` in a secret just by knowing the structure for the key/value. As most of the logic will be done on bash, it will also simplify the volumes logic as well as the validation.

justusbunsi:
Thanks for your review. I agree that mixing both configuration sources can become irritating. So either use inline config or a reference to a full user secret config works for me. I'll change that.

But I am a bit concerned regarding the one-line configuration. In your suggestion there is no guarantee for `;` (or any other character) to be safe to use. Imaging having the field delimiter in the password. It would break the parsing.

What do you think of the following secret data structure?

```yaml
data:
  username1: |-
    username=test-username
    password=sup3r=5ecure
    disabled=false
    policies=readwrite,consoleAdmin,diagnostics
    setPolicies=false
```

That way the structure would be clear for all users and flexible enough for another property order. And special characters for password (or username) would work too.

justusbunsi:
And it would allow configuring multiple users in one secret, if the admin like that. I am realizing this now. :D

justusbunsi:
Alright. Will change it in the next few days.

justusbunsi:
@FraPazGal I've changed the implementation and adjusted the PR description accordingly. I had some struggles getting both inline and external reference into the same `provisioning.users` parameter so I decided to introduce a new one for external references. I'm not really happy with the name `provisioning.usersExistingSecrets`. If you have a better name, I'd be happy to change it. 😄

FraPazGal (commented):
I tested the PR and it looks great @justusbunsi! I just added a couple of minor comments. Regarding the param name, I'm drawing a blank there so I think we can leave it as `usersExistingSecrets` 😄

justusbunsi:
> I tested the PR and it looks great @justusbunsi! I just added a couple of minor comments. Regarding the param name, I'm drawing a blank there so I think we can leave it as `usersExistingSecrets` 😄

@FraPazGal Totally agree with your suggestions. I've applied them and updated the branch to match the current target. Ready yet another time.

justusbunsi:
🤔 Anything I can do about the failing build? I'm not sure what the actual error is from reading the logs.

FraPazGal (approved):

PR #10102, opened 2022-05-10
[bitnami/rabbitmq] Add RabbitMQ Helm Chart tests

superaleks:
**Description of the change**
Increasing the test coverage of the RabbitMQ Helm Chart by adding Cypress and Goss tests. 

**Benefits**

Ensuring higher quality of the Helm charts

**Possible drawbacks**

**Test results** - [PASS](https://github.com/superaleks/charts/pull/18)
### Checklist


.vib/rabbitmq/cypress/cypress/integration/rabbitmq_spec.js
```diff
@@ -0,0 +1,100 @@
+/// <reference types="cypress" />
+import { random } from './utils';
+
+it('allows login/logout', () => {
+  const APP_NAME = 'RabbitMQ';
+
+  cy.login();
+  cy.get('#versions').should('contain', APP_NAME);
+  cy.contains('.hider', 'Queued messages');
```
superaleks, line 9: The `logout` test failed intermittently since the button would get clicked before the event listener listened.
I handled this in two ways:

- This assertion, will make sure the page will load before searching and clicking on the Log out button
- Increase click timeout by 200 ms (1100)

.vib/rabbitmq/vib-verify.json
```diff
@@ -22,11 +22,11 @@
           "url": "{SHA_ARCHIVE}",
           "path": "/bitnami/rabbitmq"
         },
-        "runtime_parameters": "ImF1dGgiOgogICJlcmxhbmdDb29raWUiOiAiUmpzcENDM25tNGtMdjRMaldGTmZGSFczeEx4Yjd6VEMiCiAgInBhc3N3b3JkIjogIlBIVkhkcXpyTWI5cyIKICAidXNlcm5hbWUiOiAidXNlciIKInNlcnZpY2UiOgogICJwb3J0cyI6CiAgICAibWFuYWdlciI6IDgwCiAgInR5cGUiOiAiTG9hZEJhbGFuY2VyIgo=",
+        "runtime_parameters": "YXV0aDoKICBlcmxhbmdDb29raWU6IFJqc3BDQzNubTRrTHY0TGpXRk5mRkhXM3hMeGI3elRDCiAgcGFzc3dvcmQ6IENvbXBsaWNhdGVkUGFzc3dvcmQxMjMhNAogIHVzZXJuYW1lOiB1c2VyCnNlcnZpY2U6CiAgcG9ydHM6CiAgICBtYW5hZ2VyOiA4MAogIHR5cGU6IExvYWRCYWxhbmNlcg==",
         "target_platform": {
           "target_platform_id": "{VIB_ENV_TARGET_PLATFORM}",
           "size": {
-            "name": "S4"
+            "name": "L4"
```
superaleks, line 29: Increased the size of the cluster because this application had performance issues and intermittent test failures on S4.

.vib/rabbitmq/cypress/cypress/integration/rabbitmq_spec.js
```diff
@@ -0,0 +1,100 @@
+/// <reference types="cypress" />
+import { random } from './utils';
+
+it('allows login/logout', () => {
+  const APP_NAME = 'RabbitMQ';
+
```
superaleks, line 6: I'm still not sure if this should be a CONST or a fixture.
I think it's a CONST since it's metadata (application name), not the actual test data we are adding.
What do the reviewers think?

.vib/rabbitmq/cypress/cypress/integration/rabbitmq_spec.js
```diff
@@ -0,0 +1,100 @@
...
+      'contain',
+      `${exchange.newExchange.name}${random}`
+    );
+    cy.contains('a', `${exchange.newExchange.name}${random}`).click();
+  });
+  cy.contains('Publish message').click();
+  cy.fixture('messages').then((message) => {
+    cy.get('textarea').type(message.newMessage.payload);
+  });
+  cy.contains('input', 'Publish message').click();
```
superaleks, line 44: I need the CSS here since there are other elements with the same text.

joancafom (changes requested):
Thanks for your PR @superaleks I left you some thoughts 😁

.vib/rabbitmq/cypress/cypress/integration/rabbitmq_spec.js
FraPazGal: I would say that using a `const` for this matches the other instances where we used `const`, though we may be misusing them. Maybe we should be opting to add the const value directly to the tests.

.vib/rabbitmq/cypress/cypress/integration/rabbitmq_spec.js
```diff
@@ -0,0 +1,102 @@
...
+  cy.contains('Add a new queue').click();
+  cy.fixture('queues').then((queue) => {
+    cy.get('[name="name"]').type(`${queue.newQueue.name}${random}`);
+    cy.get('#arguments_1_mfkey').type(queue.newQueue.argument1);
+    cy.get('#arguments_1_mfvalue').type(queue.newQueue.argument2);
+    cy.contains('Add queue').click();
+    cy.get('table[class="list"]').should(
+      'contain',
+      `${queue.newQueue.name}${random}`
+    );
```
FraPazGal, line 62: Could we use `cy.contains` here?
```suggestion
    cy.contains('table', `${queue.newQueue.name}${random}`);
    ---
    cy.contains('[class="list"]', `${queue.newQueue.name}${random}`);
```

This will apply to [L90](https://github.com/bitnami/charts/pull/10102/files#diff-11ca9508e8abe677c1ee5da0453302f35fd31855171af5ba810dd34865b89934R90) as well.

FraPazGal (commented):
Thanks for the PR! Could you please look at my comments?

.vib/rabbitmq/vib-verify.json
superaleks: I will, I didn't know I have the option of `M4`.
I've experienced intermittent failures of multiple tests.

.vib/rabbitmq/cypress/cypress/integration/rabbitmq_spec.js
superaleks: Regarding the value of the test, I wanted to anchor to a page element just to check if the login was successful. 
RabbitMQ

I'm not testing this element as such.  I can apply the Regex improvement, I think it makes sense. 
I will leave the value directly at the test then.
I'm still not really clear on fixture/const/direct usage. We can brainstorm next week!

superaleks: Applied the improvement!

.vib/rabbitmq/cypress/cypress/integration/rabbitmq_spec.js
```diff
@@ -0,0 +1,102 @@
...
+    cy.get('[class="updatable"]').should(
+      'contain',
+      `${admin.newAdmin.userName}${random}`
+    );
+    cy.contains('Log out').click();
+    cy.get('[alt="RabbitMQ logo"]');
+    cy.get('[name="username"]').type(`${admin.newAdmin.userName}${random}`);
+    cy.get('[name="password"]').type(`${admin.newAdmin.password}${random}`);
+    cy.contains('Login').click();
+    cy.get('#versions').should('contain', APP_NAME);
```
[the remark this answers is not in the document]
superaleks: I'm logging in with **different** credentials this time (the newly created admin).
Since the Cypress custom login method doesn't accept arguments as far as I can see, I can't use it. 
Do you think I should modify the login method to accept arguments and then pass admin credentials?

.vib/rabbitmq/cypress/cypress/integration/rabbitmq_spec.js
```diff
@@ -0,0 +1,102 @@
...
+      `${exchange.newExchange.name}${random}`
+    );
+    cy.contains('a', `${exchange.newExchange.name}${random}`).click();
+  });
+  cy.contains('Publish message').click();
+  cy.fixture('messages').then((message) => {
+    cy.get('textarea').type(message.newMessage.payload);
+  });
+  cy.contains('input', 'Publish message').click();
+  cy.contains('Message published, but not routed');
```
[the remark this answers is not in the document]
superaleks: Makes sense!

.vib/rabbitmq/cypress/cypress/integration/rabbitmq_spec.js
superaleks: I've improved the approach here, reducing the click timeout. 
No need to intercept the call, but I'll keep this for future reference.

.vib/rabbitmq/goss/goss.yaml
```diff
@@ -0,0 +1,31 @@
...
+    exec: rabbitmqctl version | grep --quiet $APP_VERSION
+    exit-status: 0
+    stderr: [ ]
+    timeout: 20000
+  rabbitmqctl-add-vhost:
+    exec: rabbitmqctl add_vhost test-host
+    exit-status: 0
+    stderr: [ ]
+    stdout:
+      - 'Adding vhost'
```
FraPazGal, line 12: Does the command above only output this line? I'm not sure if the command could start with `Adding vhost` but failing at the end. If the commands returns something like "vhost added", it would be great.

.vib/rabbitmq/cypress/cypress/integration/rabbitmq_spec.js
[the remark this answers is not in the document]
joancafom: The custom method does accept arguments, but defaults to the `cypress.json` defined username/password:

```javascript
Cypress.Commands.add(
  'login',
  (username = Cypress.env('username'), password = Cypress.env('password')) => {
```

You can provide your own values, though 😁 

```javascript
cy.login('myuser', 'mypass');
```

.vib/rabbitmq/cypress/cypress/integration/rabbitmq_spec.js
```diff
@@ -0,0 +1,109 @@
+/// <reference types="cypress" />
+import { random } from './utils';
+
+it('allows login/logout', () => {
+  cy.login();
+  cy.get('#versions')
+    .invoke('text')
+    .should('match', /RabbitMQ\s\d+\.\d+\.\d+/);
```
joancafom, line 8: I'd say to use a more resilient RegEx. 

```suggestion
    .should('match', /RabbitMQ.*\d+\.\d+\.\d+/);
```

If in the next version, the version is printed as `RabbitMQ: 9.8.2`, this will still match.

.vib/rabbitmq/vib-verify.json
```diff
@@ -22,11 +22,11 @@
           "url": "{SHA_ARCHIVE}",
           "path": "/bitnami/rabbitmq"
         },
-        "runtime_parameters": "ImF1dGgiOgogICJlcmxhbmdDb29raWUiOiAiUmpzcENDM25tNGtMdjRMaldGTmZGSFczeEx4Yjd6VEMiCiAgInBhc3N3b3JkIjogIlBIVkhkcXpyTWI5cyIKICAidXNlcm5hbWUiOiAidXNlciIKInNlcnZpY2UiOgogICJwb3J0cyI6CiAgICAibWFuYWdlciI6IDgwCiAgInR5cGUiOiAiTG9hZEJhbGFuY2VyIgo=",
+        "runtime_parameters": "YXV0aDoKICBlcmxhbmdDb29raWU6IFJqc3BDQzNubTRrTHY0TGpXRk5mRkhXM3hMeGI3elRDCiAgcGFzc3dvcmQ6IENvbXBsaWNhdGVkUGFzc3dvcmQxMjMhNAogIHVzZXJuYW1lOiB1c2VyCnNlcnZpY2U6CiAgcG9ydHM6CiAgICBtYW5hZ2VyOiA4MAogIHR5cGU6IExvYWRCYWxhbmNlcg==",
```
joancafom, line 25: You need to replicate these changes in `vib-publish.json` as well

.vib/rabbitmq/goss/goss.yaml
superaleks: Here is an example of a successful and erroneous command execution.

`I have no name!@rabbitmq-0:/$ rabbitmqctl add_vhost test-host1245
Adding vhost "test-host1245" ... `

`I have no name!@rabbitmq-0:/$ rabbitmqctl add_vhost t!#@$`

Error (argument validation): too many arguments.

Taking this into account, do you think it's acceptable to have it like this?

.vib/rabbitmq/cypress/cypress/integration/rabbitmq_spec.js
superaleks: Improved as per suggestion. Thanks!

.vib/rabbitmq/goss/goss.yaml
FraPazGal: You are right, the command itself only outputs the `Adding...` line and I can't find a more fail-proof option. This is perfectly fine.

FraPazGal (approved):
Thanks for the PR @superaleks, LGTM!

joancafom (approved):
LGTM

FraPazGal (approved):

PR #10169, opened 2022-05-12
[bitnami/{grafana-operator|kube-prometheus}] CRDs should not include unnecessary dashes

juan131:
Signed-off-by: juan131 <juan.ariza.1311993@gmail.com>

### Description of the change

The Grafana Operator and kube-prometheus charts' CRDs start with 3 dashes that are not necessary since these are only required as a separator when N directives are included in a single YAML file (which is not the case). Indeed, if you render the templates including the CRDs (e.g. running  `helm template kube-prometheus bitnami/kube-prometheus --include-crds > foo.yaml` and inspect the output, you'll find blocks such as the one below:

```
---
# Source: crds/crd-alertmanager-config.yaml
# https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/v0.55.0/example/prometheus-operator-crd/monitoring.coreos.com_alertmanagerconfigs.yaml

---
```

This is interpreted as an empty directive which doesn't make sense.

### Benefits

Charts compatible with `helm template --include-crds ...`.

### Possible drawbacks

None

### Applicable issues

N/A

### Additional information

N/A

### Checklist

- [x] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [x] Title of the pull request follows this pattern [bitnami/<name_of_the_chart>] Descriptive title
- [x] All commits signed off and in agreement of [Developer Certificate of Origin (DCO)](https://github.com/bitnami/charts/blob/master/CONTRIBUTING.md#sign-your-work)

PR #10185, opened 2022-05-13
[bitnami/*] Unify k8s directives separators

carrodher:

### Description of the change

The 3 dashes are only required as a separator when N directives are included in a single YAML file, being a delimiter at the **beginning** of a new directive.

Currently, if you render some templates (e.g. running `helm template kube-prometheus bitnami/kube-prometheus --include-crds > foo.yaml` and inspect the output, you'll find blocks such as the one below:

```
---
# Source: crds/crd-alertmanager-config.yaml
# https://raw.githubusercontent.com/prometheus-operator/prometheus-operator/v0.55.0/example/prometheus-operator-crd/monitoring.coreos.com_alertmanagerconfigs.yaml

---
```

This is interpreted as an empty directive that doesn't make sense.

In this diff, those delimiters are unified to

1. Removed when they are not needed (i.e at the end of the last block)
2. Placed right before the k8s directive in case there is a condition, so it is not printed if the condition is not true

Based on #10169

### Benefits

Charts compatible with `helm template ... | kubectl apply`.

### Checklist


juan131:
LGTM

juan131:
By the way, without these changes the charts aren't compatible with [Terraform for Helm](https://registry.terraform.io/providers/hashicorp/helm/latest/docs/data-sources/template). Another good reason to implement this.

migruiz4 (approved):
LGTM!

javsalgar (approved):

PR #10221, opened 2022-05-13
Fix required key name in `auth.existingSecret`

making:
`mongodb-passwords` is correct

https://github.com/bitnami/charts/blob/master/bitnami/mongodb/templates/standalone/dep-sts.yaml#L200


carrodher:

In the same way, please note you need to bump the chart version in the _Chart.yaml_ file

PR #10228, opened 2022-05-16
[bitnami/kubeapps] Restore PG persistence disabling in Kubeapps chart.

absoludity:

### Description of the change

This change restores the Kubeapps configuration for Postgresql to disable persistence. Kubeapps uses postgres as a cache only, so persistence is not necessary. We've recently had a number of users having postgres auth issues due to the persistence (latest in https://github.com/vmware-tanzu/kubeapps/issues/4689), but what we didn't realise is that this is another setting which had accidentally changed when upgrading postgres from 10->11. As you can see in the diff here, we still had intended to disable persistence, but it's still using the location of that knob for the postgres 10 chart (postgresql.persistence.enabled) rather than the 11 location (postgresql.primary.persistence.enabled).

### Benefits

No auth issues when upgrading kubeapps with the default settings.

### Applicable issues

  - fixes https://github.com/vmware-tanzu/kubeapps/issues/4689

### Checklist

### DIFF
--- a/bitnami/kubeapps/Chart.yaml
+++ b/bitnami/kubeapps/Chart.yaml
@@ -30,4 +30,4 @@ maintainers:
 name: kubeapps
 sources:
   - https://github.com/vmware-tanzu/kubeapps
-version: 8.0.19
+version: 8.0.20
--- a/bitnami/kubeapps/values.yaml
+++ b/bitnami/kubeapps/values.yaml
@@ -1974,7 +1974,6 @@ testImage:
 ## PostgreSQL chart configuration
 ## ref: https://github.com/bitnami/charts/blob/master/bitnami/postgresql/values.yaml
 ## @param postgresql.enabled Deploy a PostgreSQL server to satisfy the applications database requirements
-## @param postgresql.architecture PostgreSQL architecture (`standalone` or `replication`)
 ## @param postgresql.auth.postgresPassword Password for 'postgres' user
 ## ref: https://github.com/bitnami/bitnami-docker-postgresql/blob/master/README.md#setting-the-root-password-on-first-run
 ## @param postgresql.auth.database Name for a custom database to create
--- a/bitnami/kubeapps/values.yaml
+++ b/bitnami/kubeapps/values.yaml
@@ -1982,19 +1982,15 @@ testImage:
 ##
 postgresql:
   enabled: true
-  architecture: "replication"
   auth:
     postgresPassword: ""
     database: assets
     existingSecret: ""
-  ## PostgreSQL Persistence parameters
-  ## ref: https://kubernetes.io/docs/user-guide/persistent-volumes/
-  ## @param postgresql.persistence.enabled Enable persistence on PostgreSQL using PVC(s)
-  ## @param postgresql.persistence.size Persistent Volume size
-  ##
-  persistence:
-    enabled: false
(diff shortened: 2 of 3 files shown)


bitnami/kubeapps/values.yaml
```diff
@@ -1974,27 +1974,22 @@ testImage:
...
 ## @param postgresql.enabled Deploy a PostgreSQL server to satisfy the applications database requirements
-## @param postgresql.architecture PostgreSQL architecture (`standalone` or `replication`)
 ## @param postgresql.auth.postgresPassword Password for 'postgres' user
 ## ref: https://github.com/bitnami/bitnami-docker-postgresql/blob/master/README.md#setting-the-root-password-on-first-run
 ## @param postgresql.auth.database Name for a custom database to create
 ## @param postgresql.auth.existingSecret Name of existing secret to use for PostgreSQL credentials
 ##
 postgresql:
   enabled: true
-  architecture: "replication"
```
antgamdia, line 1985: Why remove it instead of using `standalone` ? I mean, just to make it more explicit and let users override the value more easily (because if present, it gets documented in the chart values)

bitnami/kubeapps/Chart.yaml
```diff
@@ -30,4 +30,4 @@ maintainers:
 name: kubeapps
 sources:
   - https://github.com/vmware-tanzu/kubeapps
-version: 8.0.19
+version: 8.0.20
```
antgamdia, line 33: I don't know if this change is seen as a patch version instead of a minor. Let's have someone from the team confirm it first.

bitnami/kubeapps/values.yaml
```diff
@@ -1974,27 +1974,22 @@ testImage:
...
-  ## @param postgresql.persistence.size Persistent Volume size
-  ##
-  persistence:
-    enabled: false
-    size: 8Gi
+  ## PostgreSQL Primary persistence configuration
+  ## @param postgresql.primary.persistence.enabled Enable PostgreSQL Primary data persistence using PVC
+  primary:
+    persistence:
+      enabled: false
```
antgamdia, line 1992: Great! Thanks for the change... it has been causing some headaches recently 😅

antgamdia (approved):
+1 from a Kubeapps' POV

marcosbc (approved):
LGTM!

bitnami/kubeapps/values.yaml
absoludity: Sorry, this PR was landed before I got a chance to revert/reply.

I removed it simply because "standalone" was the default value and i assumed we only had the `architecture` present in our kubeapps values because we were overriding the default value, so now that we weren't overriding it I didn't see an issue removing it. More generally, I think the values of the subchart that we explicitly call out in our values.yaml are a combination of those we override and those that make sense to present for different reasons: requirement (`postgresPassword`), importance (perhaps `architecture`?). Happy to revert and re-include this one if it's an important one.

bitnami/kubeapps/Chart.yaml
absoludity: Same here. Was a worthwhile point to consider, but as the chart had been landed, I didn't come back to it, sorry. Luckily, the minor version was bumped with the next PR that landed here upstream anyway :)

PR #10241, opened 2022-05-16
[bitnami/discourse] Add goss/cypress tests

FraPazGal:

**Description of the change**
Adds Cypress and Goss tests to the `bitnami/discourse` chart.

**Benefits**

Ensuring higher quality of the Helm chart.

**Possible drawbacks**

**Test results** - PASS

[Please see them in the history here](https://github.com/FraPazGal/charts/pull/9)

**Checklist**

### DIFF
--- a/.vib/discourse/cypress/cypress/integration/discourse_spec.js
+++ b/.vib/discourse/cypress/cypress/integration/discourse_spec.js
@@ -1,0 +1,64 @@
+/// <reference types="cypress" />
+import { random } from './utils';
+
+it('allows to log in and out', () => {
+  cy.login();
+  cy.contains('a', 'Unread');
+  cy.get('#current-user').click();
+  cy.get('[title="Preferences"]').click();
+  cy.contains('Log Out').click();
+  cy.contains('button', 'Log In');
+});
+
+it('allows to sign in', () => {
+  cy.contains('button', 'Sign Up').click();
+  cy.fixture('users').then((user) => {
+    cy.get('#new-account-email').type(`${random}.${user.newUser.email}`);
+    cy.get('#new-account-username').type(`${random}.${user.newUser.username}`);
+    cy.get('#new-account-name').type(`${user.newUser.name} ${random}`);
+    cy.get('#new-account-password').type(`${random}.${user.newUser.password}`);
+  });
+  cy.contains('Checking username').should('not.exist');
+  cy.contains('button', 'Create your account').click();
+  cy.contains('button', 'Resend Activation Email');
+});
+
+it('allows to create a topic', () => {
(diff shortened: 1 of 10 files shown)


.vib/discourse/cypress/cypress/integration/discourse_spec.js
```diff
@@ -0,0 +1,64 @@
...
+  cy.login();
+  cy.visit('/u/user/preferences/account');
+  cy.fixture('user-settings').then((userSettings) => {
+    cy.get('[class*="pref-name"]').within(() => {
+      cy.get('input')
+        .clear({ force: true })
+        .type(`${userSettings.user.realName} ${random}`);
+    });
+    cy.contains('button', 'Save Changes').click();
+    cy.contains('Saved').click();
```
joancafom, line 51: This looks like a confirmation msg. Do we really want to click on it?

.vib/discourse/cypress/cypress/integration/discourse_spec.js
```diff
@@ -0,0 +1,64 @@
+/// <reference types="cypress" />
+import { random } from './utils';
+
+it('allows to log in and out', () => {
+  cy.login();
+  cy.contains('a', 'Unread');
```
joancafom, line 6: I guess this is checking for the unread notifs. My concern here is what would happen if the test is executed twice, would the notifs still be unread?

Maybe we can change this check to something that does not depend on a status (`read/unread`) if the goal here is to ensure we are logged in. WDYT?

joancafom (changes requested):
Thanks for your PR @FraPazGal ! I left you some comments, could you please take a look at them?

.vib/discourse/cypress/cypress/integration/discourse_spec.js
FraPazGal: In this case, it is an additional view that only appears when you are logged in and shows you your unread topics as shown below. There are other elements only present when authenticated, but this one shouldn't give us any issue.

Screenshot 2022-05-18 at 13 54 44

.vib/discourse/cypress/cypress/integration/discourse_spec.js
FraPazGal: We don't 😅 Good catch!

.vib/discourse/cypress/cypress/integration/discourse_spec.js
joancafom: Ok, got it!

joancafom (approved):
LGTM

superaleks (approved):
LGTM!

PR #10257, opened 2022-05-16
[bitnami/argo-workflows] Easier configuration for workflowDefaults in Controller configMap

ktaletsk:
Signed-off-by: Konstantin Taletskiy <konstantin.taletskiy@labshare.org>

### Description of the change

Adds a new section in values.yaml to have an easier time overriding the section of the Controller configMap.

```yaml
## Default Workflow Values
## Ref: https://argoproj.github.io/argo-workflows/default-workflow-specs/#setting-default-workflow-values
  workflowDefaults:
    enabled: false
```

Currently, if you would like to override `workflowDefaults`, you are required to copy the default configMap from default values, uncomment `workflowDefaults` and add your desired info there.

### Benefits

With this change, configMap will automatically quote the new section of the values, so you are no longer required to override the whole configMap.

### Checklist


joancafom:
Thanks for the explanation, now I got your point!

As a side note

> In fact, this approach is already used in this config file in case of telemetry parameter

This is also done this way for another reason, and it is that the telemetry config does affect some other resources deployed in the chart, for instance, in a service:

https://github.com/bitnami/charts/blob/ec1526accf12f86b5ae642aa905bf85ee7028f4b/bitnami/argo-workflows/templates/controller/service.yaml#L48

Having said that, let me re-review your changes!

bitnami/argo-workflows/values.yaml
```diff
@@ -672,10 +672,9 @@ controller:
           name: {{ include "argo-workflows.controller.database.password.secret" . }}
           key: {{ include "argo-workflows.controller.database.password.secret.key" . }}
     {{- end }}
-    ## Uncomment to set workflowDefaults
-    ## Ref: https://argoproj.github.io/argo-workflows/default-workflow-specs/#setting-default-workflow-values
-    ## workflowDefaults: {}
-    ##
+    {{- if .Values.controller.workflowDefaults.enabled }}
+    workflowDefaults: {{- include "common.tplvalues.render" (dict "value" .Values.controller.workflowDefaults "context" $) | nindent 2 }}
```
joancafom, line 676: I'm afraid this is not as easy as with `telemetryConfig`. It turns out that `telemetryConfig.enabled` is a valid entry:

https://github.com/argoproj/argo-workflows/blob/1bd841853633ebb71fc569b2975def90afb1a68d/docs/workflow-controller-configmap.yaml#L259-L263

Nevertheless, I don't think `workflowDefaults.enabled` would be valid...

https://github.com/argoproj/argo-workflows/blob/1bd841853633ebb71fc569b2975def90afb1a68d/docs/workflow-controller-configmap.yaml#L320-L329

 You may need to add a subentry to `workflowDefaults`, say `workflowDefaults.spec`, where the config should be specified.

joancafom (changes requested):
Thanks for the PR! I left you a comment, could you please review it?

bitnami/argo-workflows/values.yaml
ktaletsk: I added `workflowDefaults.enabled` as a new variable and in my testing it did not introduce any issues, even though it is not a part of the `workflowDefaults` accepted keys. I understand why it might not be desirable to do it.

My new idea then is to remove `enabled` key from `workflowDefaults`, set the default value for `workflowDefaults` as `{}`, then check that `workflowDefaults` dict is not empty when including the contents of it in the configMap


```diff
-  ## @param controller.workflowDefaults.enabled Enable Default Workflow Values
-  workflowDefaults:
-    enabled: false
+  ## @param controller.workflowDefaults Default Workflow Values
+  workflowDefaults: {}


- {{- if .Values.controller.workflowDefaults.enabled }}
+ {{- if .Values.controller.workflowDefaults }}
```

By default this does not add anything to the configMap (same as before).

If user would like to provide Workflow Defaults, they would do the following:
```
workflowDefaults:
  metadata:
    annotations:
      argo: workflows
    labels:
      foo: bar
  spec:
    ttlStrategy:
      secondsAfterSuccess: 5
    parallelism: 3
```

Importantly, the values of `workflowDefaults` are not using multi-line literal style scalar as they do in [linked docs](https://argoproj.github.io/argo-workflows/default-workflow-specs/#setting-default-workflow-values), but "normal" yaml syntax.

Is this option looking better?

joancafom: > My new idea then is to remove enabled key from workflowDefaults, set the default value for workflowDefaults as {}, then check that workflowDefaults dict is not empty when including the contents of it in the configMap

Super nice! 🚀 

> but "normal" yaml syntax.

I reckon this totally makes sense. I will mark this thread as resolved :)

bitnami/argo-workflows/values.yaml
```diff
@@ -672,10 +672,9 @@ controller:
           name: {{ include "argo-workflows.controller.database.password.secret" . }}
           key: {{ include "argo-workflows.controller.database.password.secret.key" . }}
     {{- end }}
-    ## Uncomment to set workflowDefaults
-    ## Ref: https://argoproj.github.io/argo-workflows/default-workflow-specs/#setting-default-workflow-values
-    ## workflowDefaults: {}
-    ##
+    {{- if .Values.controller.workflowDefaults }}
+    workflowDefaults: {{- include "common.tplvalues.render" (dict "value" .Values.controller.workflowDefaults "context" $) | nindent 2 }}
```
joancafom, line 676: Bear in mind that the entire `config` file is indented using a value of 4, you may need to update the value here too:

```diff
- workflowDefaults: {{- include "common.tplvalues.render" (dict "value" .Values.controller.workflowDefaults "context" $) | nindent 2 }}
+ workflowDefaults: {{- include "common.tplvalues.render" (dict "value" .Values.controller.workflowDefaults "context" $) | nindent 4 }}
```

joancafom (changes requested):
Good job! I added an extra change that you may need to apply. Could you please check it?

bitnami/argo-workflows/Chart.yaml
```diff
@@ -35,4 +35,4 @@ sources:
   - https://github.com/bitnami/bitnami-docker-argo-workflow-controller
   - https://github.com/bitnami/bitnami-docker-argo-workflow-exec
   - https://argoproj.github.io/workflows/
-version: 2.1.0
+version: 2.1.1
```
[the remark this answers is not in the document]
ktaletsk: Sure, I am going to bump MINOR now, but should check for upstream value again before merging (in case it changes in the meantime)

bitnami/argo-workflows/values.yaml
joancafom: >Indent should be counted relative to the config file level and each nested level should add 2 spaces (which is the convention throughout values.yaml).

You are absolutely right! I missed the point that the param itself is also adding indentation. 

> Because of that, I claim that metricsConfig, telemetryConfig, sso and workflowDefaults should all be using indent 2.

Would you mind adding the changes to fix the level of indentation as well in the PR?

ktaletsk: Added the indent changes

joancafom (approved):
LGTM

Issue #4704, opened 2020-12-11
[kafka] When run kafka-ropics.sh it throws port already in use


sergeyshaykhullin:
**Which chart**:
kafka

**Describe the bug**
I started researching to provision kafka topics https://github.com/bitnami/charts/issues/4651 from helm chart and tried to run kafka-topics.sh manually. When i ran it, i got an exception that port is already in use

**To Reproduce**
1. Run kafka cluster
2. `kubectl exec -it kafka-0 -n kafka kafka -- bash`
3. `export KAFKA_OPTS="-Djava.security.auth.login.config=/opt/bitnami/kafka/conf/kafka_jaas.conf"`
4. `/opt/bitnami/kafka/bin/kafka-topics.sh --create --zookeeper SERVER-IP:2181 --replication-factor 1 --partitions 1 --topic test`

**Expected behavior**
I can use kafka-topics.sh without exceptions

**Version of Helm and Kubernetes**:

- Output of `helm version`:

```
version.BuildInfo{Version:"v3.4.1", GitCommit:"c4e74854886b2efe3321e185578e6db9be0a6e29", GitTreeState:"clean", GoVersion:"go1.14.11"}
```

- Output of `kubectl version`:

```
Client Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.4", GitCommit:"d360454c9bcd1634cf4cc52d1867af5491dc9c5f", GitTreeState:"clean", BuildDate:"2020-11-11T13:17:17Z", GoVersion:"go1.15.2", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.4", GitCommit:"d360454c9bcd1634cf4cc52d1867af5491dc9c5f", GitTreeState:"clean", BuildDate:"2020-11-11T13:09:17Z", GoVersion:"go1.15.2", Compiler:"gc", Platform:"linux/amd64"}
```

**Additional context**

Docs https://docs.bitnami.com/aws/infrastructure/kafka/administration/run-producer-consumer/

Exception
```
[...]

PR #10284, opened 2022-05-17
[bitnami/kafka] Added config og JMX port to avoid conflict when using Kafka CLI

3schwartz:
Kafka CLI doesn't work when using it inside the both while the JMX exporter is enabled. This is because the JXM application is running from port 5555, which isn't configurable.

I have added configurations such that the JXM exporter can run on port ex. 5557 which doesn't break the Kafka CLI.

### Description of the change

Even though default values break Kafka CLI there have been keep to avoid side effects when deploying new chart.

### Benefits

Able to use JMX exporter and Kafka CLI.

### Checklist


joancafom:
Hi @3schwartz 

Thanks for your PR! However, I don't think this does resolve the issue. 

> This is because the JXM application is running from port 5555, which isn't configurable.

The problem is that even if you configure the port, running Kafka CLI within the container will try to start another kafka server, which in turn will try to bind to the same JMX port and hence fail. In order to validate that, can you please ensure you are able to make it work with your changes? On my side, I bumped into the same error message.

```bash
$ kubectl exec -it sapphire-kafka-0 -- bash
> kafka-topics.sh
Error: Exception thrown by the agent : java.rmi.server.ExportException: Port already in use: 9999; nested exception is:
	java.net.BindException: Address already in use (Bind failed)
...
```

On the other hand, the chart's installation notes mention the correct way to create a pod and use it as a Kafka client, in which you may be able to execute kafka cli without problems:

```bash
$ helm install sapphire bitnami/kafka -f values.yaml
...
To create a pod that you can use as a Kafka client run the following commands:

    kubectl run sapphire-kafka-client --restart='Never' --image docker.io/bitnami/kafka:3.1.1-debian-10-r4 --namespace default --command -- sleep infinity
    kubectl exec --tty -i sapphire-kafka-client --namespace default -- bash
...
```

joancafom:
Hi @3schwartz 

Thanks for your PR! However, I don't think this does resolve the issue. 

> This is because the JXM application is running from port 5555, which isn't configurable.

The problem is that even if you configure the port, running Kafka CLI within the container will try to start another kafka server, which in turn will try to bind to the same JMX port and hence fail. In order to validate that, can you please ensure you are able to make it work with your changes? On my side, I bumped into the same error message.

```bash
$ kubectl exec -it sapphire-kafka-0 -- bash
> kafka-topics.sh
Error: Exception thrown by the agent : java.rmi.server.ExportException: Port already in use: 9999; nested exception is:
	java.net.BindException: Address already in use (Bind failed)
...
```

On the other hand, the chart's installation notes mention the correct way to create a pod and use it as a Kafka client, in which you may be able to execute kafka cli without problems:

```bash
$ helm install sapphire bitnami/kafka -f values.yaml
...
To create a pod that you can use as a Kafka client run the following commands:

    kubectl run sapphire-kafka-client --restart='Never' --image docker.io/bitnami/kafka:3.1.1-debian-10-r4 --namespace default --command -- sleep infinity
    kubectl exec --tty -i sapphire-kafka-client --namespace default -- bash
...
```

3schwartz:
> Hi @3schwartz
> 
> Thanks for your PR! However, I don't think this does resolve the issue.
> 
> > This is because the JXM application is running from port 5555, which isn't configurable.
> 
> The problem is that even if you configure the port, running Kafka CLI within the container will try to start another kafka server, which in turn will try to bind to the same JMX port and hence fail. In order to validate that, can you please ensure you are able to make it work with your changes? On my side, I bumped into the same error message.
> 
> ```shell
> $ kubectl exec -it sapphire-kafka-0 -- bash
> > kafka-topics.sh
> Error: Exception thrown by the agent : java.rmi.server.ExportException: Port already in use: 9999; nested exception is:
> 	java.net.BindException: Address already in use (Bind failed)
> ...
> ```
> 
> On the other hand, the chart's installation notes mention the correct way to create a pod and use it as a Kafka client, in which you may be able to execute kafka cli without problems:
> 
> ```shell
> $ helm install sapphire bitnami/kafka -f values.yaml
> ...
> To create a pod that you can use as a Kafka client run the following commands:
> 
>     kubectl run sapphire-kafka-client --restart='Never' --image docker.io/bitnami/kafka:3.1.1-debian-10-r4 --namespace default --command -- sleep infinity
>     kubectl exec --tty -i sapphire-kafka-client --namespace default -- bash
> ...
> ```

Strange - I tried default values file and only changes was

```shell
metrics:
  jmx:
    ## @param metrics.jmx.enabled Whether or not to expose JMX metrics to Prometheus
    ##
    enabled: true
    ## @param metrics.jmx.port JMX application exporter port - Default conflicts with Kafka CLI
    ##
    port: 5557
```

Then I get the exceptions. In a older version of the chart manage to have this working by adding this explicitly 
```
metrics:
  jmx:
    ## @param metrics.jmx.enabled Whether or not to expose JMX metrics to Prometheus
    ##
    config: |-
      jmxUrl: service:jmx:rmi:///jndi/rmi://127.0.0.1:557/jmxrmi
      lowercaseOutputName: true
      lowercaseOutputLabelNames: true
      ssl: false
      {{- if .Values.metrics.jmx.whitelistObjectNames }}
      whitelistObjectNames: ["{{ join "\",\"" .Values.metrics.jmx.whitelistObjectNames }}"]
      {{- end }}    
```
and then from a script run before start up of Kafka export JMX_PORT=5557 to override environment variable from the chart. Then both metrics and Kafka CLI from within the pods works.

I will try do some investigation. Do you know if anything has changes related to these configurations back until around when the chart was targeting Kafka 2.8.X?
It is needed to due to the implementation of a readiness check which listen to the same metrics as the one Strimzi is listening to https://github.com/strimzi/strimzi-kafka-operator/blob/main/kafka-agent/src/main/java/io/strimzi/kafka/agent/KafkaAgent.java.

joancafom:
> I will try do some investigation. Do you know if anything has changes related to these configurations back until around when the chart was targeting Kafka 2.8.X?

This dates as back as Oct 2019, so probably this has nothing to do with recent configurations:

https://github.com/bitnami/charts/issues/1522
https://github.com/bitnami/charts/issues/4704

Just to make sure, have you tried using the recommended kafka-client pod mentioned in the helm installation notes? Do you need to execute the CLI from within the pods for a reason? That would be helpful to know 😁

jouve:
have you tried running
``` 
unset JMX_PORT
kafka-topics.sh ...
``` 

 it should solve your problem

ref: https://github.com/apache/kafka/blob/trunk/bin/kafka-run-class.sh#L214

3schwartz:
> > I will try do some investigation. Do you know if anything has changes related to these configurations back until around when the chart was targeting Kafka 2.8.X?
> 
> This dates as back as Oct 2019, so probably this has nothing to do with recent configurations:
> 
> #1522 #4704
> 
> Just to make sure, have you tried using the recommended kafka-client pod mentioned in the helm installation notes? Do you need to execute the CLI from within the pods for a reason? That would be helpful to know 😁

In our production setup I can't deploy new pods on the fly. However I can debug by exec into the container.

joancafom:
> In our production setup I can't deploy new pods on the fly. However I can debug by exec into the container.

In that case, the workaround proposed by @jouve would make it 😁. Could you please give it a try?

3schwartz:
> unset JMX_PORT

It works (sorry for the late response).

@jouve may I ask why this work? We unset the environment variable - does this work because the variable are just needed to start up the JMX exporter? 

The errors messages states port 5555 is in use. Even though we unset the env-variable the JMX process is still running. So why no port conflict just because of the unset?

jouve:
the issue arise because in the container, kafka is running & listening on jmx port 5555.

when you run kafka cli tools in the container, JMX_xxx opts are defined, so kafka-console-consumer.sh also starts listening jmx on port 5555, which conflicts with kafka already listening on 5555.

3schwartz:
> the issue arise because in the container, kafka is running & listening on jmx port 5555.
> 
> when you run kafka cli tools in the container, KAFKA_JMX_OPTS & JMX_PORT=5555 env vars are defined, so kafka-console-consumer.sh also starts listening jmx on port 5555, which conflicts with kafka already listening on 5555.

Thanks - that makes sense

PR #1696, opened 2019-12-03
[bitnami/apache] add path to http probes

whall89:
**Description of the change**

I have added the option to allow the pod health checks(liveness and readiness probes) be pointed at other paths and ports different from the defaults, "/" and "http".

**Benefits**

This allows the pod to point to a specific path and port on apache weather it is an html file, other web app or in our case "/server-status" the built-in apache health check.

**Possible drawbacks**

if the path and port are changed from default and arent configured properly in the apache vhost, config or web-app it could cause an error.

**Applicable issues**

N/A

**Additional information**

**Checklist** 
- [X] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [x] Variables are documented in the README.md
- [X] Title of the PR starts with chart name (e.g. `[bitnami/chart]`)
- [ ] If the chart contains a `values-production.yaml` apart from `values.yaml`, ensure that you implement the changes in both files

:warning: Keep in mind that if you want to make changes to the kubeapps chart, please implement them in the [kubeapps repository](https://github.com/kubeapps/kubeapps/tree/master/chart/kubeapps). This is only a synchronized mirror.

PR #10286, opened 2022-05-18
[bitnami/apache] Fix probes not reading port from .Values

JonasAlfredsson:
### Description of the change

From what I can see pull request #1696 tried to introduce both customizable path and port number for the liveness and readiness probes, however, the part where it actually read the custom value from the .Values file was removed before the merge.

With this pull request, we add the necessary code for the template to read any custom value that the user has added to their .Values file.

Signed-off by: Jonas Alfredsson <jonasal@axis.com>

### Benefits

As it is now the port is defined in the .Values file as it would be a customizable value, but this is completely ignored when you try to change it since this is not propagated to the final deployment.yaml. This might cause confusion when nothing happens even though the user has defined a custom port.

### Possible drawbacks

I don't really see any drawbacks with this. There already are default values for the port, so nothing will change for anyone that doesn't have a custom port defined in their deployment. For those that have changed the port, this will then suddenly start working as the user wants, but it might be so that they have wrongly configured something else which then makes their probes fail. But that is a bug that they probably are interested in finding. 

### Applicable issues

If I want to probe another port than the one serving all the standard HTTP requests I see no reason to stop me from doing that. 

### Checklist


carrodher (approved):
Thank you for this contribution, changes LGTM

JonasAlfredsson:
Nice!
In regards to the failing test I am not sure what goes wrong. Could you please help identifying the issue and will see if I can fix it.

FraPazGal:
Hello @JonasAlfredsson, the issue comes from the trivy scan. We are aware of the related CVE and new versions of our assets are being released, so we can omit the error in this case. We'll merge this PR through our internal CI so there shouldn't be any further issues.

PR #10306, opened 2022-05-18
[bitnami/redis] Support allocateLoadBalancerNodePorts setting

olljanat:
### Description of the change
I noticed that `allocateLoadBalancerNodePorts` looks to be supported by some Bitnami charts but not all of them and we prefer to be able to skip node port allocation on services which use service type `LoadBalancer`


carrodher:
Thanks for your contribution. Could you please solve the issues reported by the GH actions? Mainly you need to bump the chart version in the Chart.yaml and add the metadata for the new parameter in the values.yaml

carrodher:
Thanks for your contribution. Could you please solve the issues reported by the GH actions? Mainly you need to bump the chart version in the Chart.yaml and add the metadata for the new parameter in the values.yaml

recena:
@olljanat Please, could you use the PR template proposed by default?

PR #10372, opened 2022-05-23
[bitnami/grafana] Standardise LDAP configuration

fmulero:
### Description of the change

Add several ldap values to configure authentication against LDAP server.

### Benefits

The goal of this task is to achieve homogeneity in the LDAP parameters exposed by the values.yaml file across the different Charts.

### Checklist


bitnami/grafana/templates/deployment.yaml
```diff
@@ -316,6 +323,11 @@ spec:
             secretName: {{ .Values.config.grafanaIniSecret }}
           {{- end }}
         {{- end }}
+        {{- if .Values.ldap.tls.certificatesSecret }}
+        - name: ldap-tls
+          secret:
+            secretName: {{ .Values.ldap.tls.certificatesSecret }}
+        {{- end }}
```
migruiz4, line 330: The condition for the volumeMount is `.Values.ldap.tls.enabled`, so I think we could use the same condition here.

What do you think about this:

```suggestion
        {{- if .Values.ldap.tls.enabled }}
        - name: ldap-tls
          secret:
            secretName: {{ required "Please provide a secret containing TLS certificates to enable 'ldap.tls.certificatesSecret'" .Values.ldap.tls.certificatesSecret }}
        {{- end }}
```

bitnami/grafana/templates/_helpers.tpl
```diff
@@ -176,3 +177,47 @@ grafana: ldap.enabled ldap.configMapName and ldap.secretName
...
+{{- end }}
+{{- if .Values.ldap.bindpw }}
+# Search user bind password
+# If the password contains # or ; you have to wrap it with triple quotes. Ex """#password;"""
+bind_password = {{ .Values.ldap.bindpw | quote }}
+{{- end }}
+
+# User search filter, for example "(cn=%s)" or "(sAMAccountName=%s)" or "(uid=%s)"
+# Allow login from email or username, example "(|(sAMAccountName=%s)(userPrincipalName=%s))"
+search_filter = "({{ .Values.ldap.uidField }}=%s)"
```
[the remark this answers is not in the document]
fmulero: I completely agree but I am not sure at the moment if the rest of charts could adopt this solution. I'll keep this conversation open to revisit it before removing the draft mode

bitnami/grafana/templates/deployment.yaml
fmulero: Thanks for pointing out, there was a failure there. 

About your suggestion, the certificates are not required when `ldap.tls` is enabled. If the LDAP certificate is signed by a wide trusted authority or the user disabled any validation with `ldap.extraConfiguration` , `tls.certificatesSecret` won't be required

bitnami/grafana/values.yaml
```diff
@@ -149,6 +149,54 @@ ldap:
   ## NOTE: When it's set the ldap.configuration parameter is ignored
```
migruiz4, line 149: Value `ldap.uri` missing here

migruiz4 (changes requested):

bitnami/grafana/values.yaml
fmulero: Argh! you the best!

PR #10387, opened 2022-05-24
[bitnami/keycloak] Add support for distributed caching in kubernetes

abrahamjoc:
### Description of the change

This PR adds support for configuring keycloak distributed caching in kubernetes
Reference: https://www.keycloak.org/server/caching

### Benefits

Users can configure keycloak distributed cache using transport stack as kubernetes and cache as infinity span.

### Applicable issues

relates_to: https://github.com/bitnami/bitnami-docker-keycloak/pull/65

### Checklist

### DIFF
--- a/bitnami/keycloak/Chart.yaml
+++ b/bitnami/keycloak/Chart.yaml
@@ -26,4 +26,4 @@ name: keycloak
 sources:
   - https://github.com/bitnami/bitnami-docker-keycloak
   - https://github.com/keycloak/keycloak
-version: 9.0.5
+version: 9.1.0
--- a/bitnami/keycloak/README.md
+++ b/bitnami/keycloak/README.md
@@ -312,6 +312,13 @@ The command removes all the Kubernetes components associated with the chart and 
 | `externalDatabase.existingSecretPasswordKey` | Name of an existing secret key containing the database credentials      | `""`               |
 
 
+### Keycloak Cache parameters
+
+| Name            | Description                                                               | Value   |
+| --------------- | ------------------------------------------------------------------------- | ------- |
+| `cache.enabled` | Switch to enable or disable the keycloak distributed cache for kubernetes | `false` |
+
+
 Specify each parameter using the `--set key=value[,key=value]` argument to `helm install`. For example,
 
 ```bash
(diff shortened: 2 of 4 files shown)


marcosbc (approved):

ddurham2:
@abrahamjoc, may I ask for clarification on this change (which has been amended a bit since this original PR).

My use-case involves running keycloak in kubernetes.  And I will have control over the replicas count.

I'm trying to ascertain whether deploying keycloak in this simplistic configuration will scale as needed (e.g. potentially a million users, but how active is not yet known).  And, notwithstanding the disappointing fact that all users get logged out any restart/upgrade of keycloak.

But concerning scaling, I've seen horror stories online where much work had to be done to make keycloak scale, and how blowing the cache limit may start causing random users to get logged out, but some of those predate more recent improvements.  

My question here is what exactly this change does.  Does it allow keycloak's internal infinispan instances to discover each other?  If so, how did keycloak behave before this change when there were multiple replicas with synchronized caches?  Or did it not use a cache?

_Or_, does it somehow facility one running their own infinispan instance externally and keycloak discovering that?  I'm doubtful of that given the articles about running hotrod and changing the cache server's address.  All of that seems somewhat unfinished/experimental work at present anyway. 

Does this change improve scalability or fix some problem beyond one relica?

Thanks for any info.

abrahamjoc:
Hi @ddurham2, this PR was made when the option to set the infinispan cache was added to the bitnami keycloak docker image, apparently the track of that change was lost because it had a separate repository before, now the current code declaration can be found here:
- https://github.com/bitnami/containers/blob/main/bitnami/keycloak/21/debian-11/rootfs/opt/bitnami/scripts/keycloak-env.sh#L127-L128
- https://github.com/bitnami/containers/blob/main/bitnami/keycloak/21/debian-11/rootfs/opt/bitnami/scripts/libkeycloak.sh#L161-L165

I don't know the behavior of keycloak with a large number of users.

The change corresponds to the documentation in:
- https://www.keycloak.org/server/caching#_enable_distributed_caching
- https://www.keycloak.org/server/caching#_transport_stacks

which indicate that ispn can be enabled and the transport stack option must be kubernetes to run on a kubernetes cluster.

I did tests with the work team and they worked fine. We saw the ispn discovery service works correctly.

My recommendation is that you try to run one instance and up the keycloak statefulset on 2 replicas, by checking the pod logs you will notice that there is automatic communication with the ispn cache between the pods. 

If you need more information to carry out this test, let me know and I can give you more details.

PD.
The current bitnami keycloak helm chart has the following configuration:
- https://github.com/bitnami/charts/blob/main/bitnami/keycloak/values.yaml#L1068-L1071
- https://github.com/bitnami/charts/blob/main/bitnami/keycloak/templates/configmap-env-vars.yaml#L55-L65

That configuration is not what I originally did, but it seems to be a correct implementation to manage the cache configuration.

PR #10392, opened 2022-05-25
[bitnami/kubeapps] Switch to use the non-admin postgresql database user.

absoludity:

### Description of the change

Fixes https://github.com/vmware-tanzu/kubeapps/issues/4643

### Benefits

Better example to set.

### Possible drawbacks

Breaks the upgrade, since the secret won't exist and the DB is created on psql startup. Included instructions for working around the upgrade, but not sure it's worthwhile. Thoughts @antgamdia , @castelblanque ?

### Applicable issues

  - fixes https://github.com/vmware-tanzu/kubeapps/issues/4643

### Checklist

### DIFF
--- a/bitnami/kubeapps/Chart.yaml
+++ b/bitnami/kubeapps/Chart.yaml
@@ -30,4 +30,4 @@ maintainers:
 name: kubeapps
 sources:
   - https://github.com/vmware-tanzu/kubeapps
-version: 8.1.6
+version: 8.1.7
--- a/bitnami/kubeapps/README.md
+++ b/bitnami/kubeapps/README.md
@@ -7,7 +7,7 @@ Kubeapps is a web-based UI for launching and managing applications on Kubernetes
 [Overview of Kubeapps](https://github.com/vmware-tanzu/kubeapps)
 
 
-                           
+
 ## TL;DR
 
 ```bash
@@ -933,6 +933,31 @@ To reduce this response time, you can increase the number of checks that Kubeapp
 Feel free to [open an issue](https://github.com/vmware-tanzu/kubeapps/issues/new) if you have any questions!
 
 ## Troubleshooting
+
+### Upgrading to chart version 8.2.0
+
+This minor release sees Kubeapps using a non-admin user for the postgresql database connection (used for the Helm chart cache).
+
+To upgrade an existing installation, you will need to first check if your Kubeapps database secret already contains a regular `password` in addition to the `postgres-password`:
+
+```
+kubectl --namespace kubeapps get secret kubeapps-postgresql -o jsonpath='{.data}'
+```
+
+If it does not yet contain a regular `password`, add one with:
+
+```
+DBPASS=$(cat /dev/urandom | tr -dc 'a-zA-Z0-9' | fold -w 32 | head -n 1)
+kubectl --namespace kubeapps patch secret kubeapps-postgresql -p="{\"stringData\":{\"password\": \"${DBPASS}\"}}"
+```
+
+Finally, delete the current Postgres stateful set so that the upgrade will create a new Postgres instance:
+
+```
+kubectl --namespace kubeapps delete statefulset kubeapps-postgresql
+```
+
+After upgrading, when displaying the catalog, Kubeapps will initially display an error because the sync job will not yet have re-created the database tables used for the cache, but this will fix itself in a short time as the sync job runs.
 
 ### Upgrading to chart version 8.0.0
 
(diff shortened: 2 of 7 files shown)


castelblanque:
Given that there is the discussion (and work) about completely dropping Postgres from the Helm plugin (and even the plugin itself as it is today), I'm wondering whether it is worth to make this kind of changes, or just leave it as it is if finally PG is removed? Mainly to avoid existing installations of Kubeapps to go through breaking changes.
Not having a strong opinion on this, though.

absoludity:
Yep, I think I agree - I'm not sure this change is worth the cost of the upgrade hassle, either way. @ppbaena , see what you think - I've outlined what would be required when upgrading Kubeapps with this change. The reality is that yes, we currently use the admin postgres user for historical reasons beyond me, but it isn't really a security issue as the pg instance is not shared, nor does it store sensitive data (operating as a cache). I'm happy to push forward with it if you think it worthwhile, but the only benefit I see is modelling better use of postgres, potentially for a short time (at the cost of non-zero manual upgrade steps).

antgamdia (commented):
I don't know how similar chanegs are being managed in other Bintami charts.
Howver, I don't see that straightforward upgrade path and perhaps it is somehting we should do when releasing a new version.

What I now suggest is:

1. Adding the upgrade path as part of the NOTES.txt, so that users can have handy the commands to execute.
2. Making these changes an opt-in change. Like adding a `postgresql.useNonAdminUser: true` to actually add the changes you proposed herein. This way we can just change the default behavior when releasing a new Kubeapps version (and properly announcing it in the release notes?)  

But otherwise, +1 to this change. I mean, I do think it is worthwhile and brings an actual improvement from our current situation.

absoludity:
> I don't know how similar chanegs are being managed in other Bintami charts. Howver, I don't see that straightforward upgrade path and perhaps it is somehting we should do when releasing a new version.
> 
> What I now suggest is:
> 
>     1. Adding the upgrade path as part of the NOTES.txt, so that users can have handy the commands to execute.
> 
>     2. Making these changes an opt-in change. Like adding a `postgresql.useNonAdminUser: true` to actually add the changes you proposed herein. This way we can just change the default behavior when releasing a new Kubeapps version (and properly announcing it in the release notes?)

I like the idea of 2, yes - means we can encode it now, even switch our CI to use it if we want, and switch the default for a major release later. Might be a good way forward - thanks @antgamdia .

alemorcuq:
For us, a change that breaks the upgrade requires a new chart major.

absoludity:
> > I don't know how similar chanegs are being managed in other Bintami charts. Howver, I don't see that straightforward upgrade path and perhaps it is somehting we should do when releasing a new version.
> > What I now suggest is:
> > ```
> > 1. Adding the upgrade path as part of the NOTES.txt, so that users can have handy the commands to execute.
> > 
> > 2. Making these changes an opt-in change. Like adding a `postgresql.useNonAdminUser: true` to actually add the changes you proposed herein. This way we can just change the default behavior when releasing a new Kubeapps version (and properly announcing it in the release notes?)
> > ```
> 
> I like the idea of 2, yes - means we can encode it now, even switch our CI to use it if we want, and switch the default for a major release later. Might be a good way forward - thanks @antgamdia .

So I spent a bit of time looking at this today and have pushed what I found, but I'm closing this PR as it would still require manual intervention for the upgrade - and I don't see a way around that.

The issue is that, even though I can create the `postgresql.useNonAdminUser` flag to control the kubeapps side of the chart perfectly (so that the correct user/password is passed to our deployments etc. depending on the flag), I cannot control the postgres chart itself. The postgres chart values require setting the `auth.username` field to both generate and look up the secret value. Values in values.yaml are not computed, so the only options are to set the `auth.username` in the chart by default (to "kubeapps" as I've done), or to require the user to set it explicitly if they switch the `useNonAdminUser` (clunky and error prone to maintain). Unfortunately, when the `auth.username` is set, the postgres chart fails if the non-admin password does not yet exist in the secret (rather than adding it), with:

```
helm upgrade --install kubeapps ~/dev/vmware/charts/bitnami/kubeapps --namespace kubeapps --create-namespace

Error: UPGRADE FAILED: execution error at (kubeapps/charts/postgresql/templates/secrets.yaml:20:15): 
PASSWORDS ERROR: The secret "kubeapps-postgresql" does not contain the key "password"

```

I assume this is because the bitnami helpers avoid setting passwords on upgrades etc. , even though in our specific case, it could be done as the password doesn't exist yet and we don't use persistence, so once set, it would have been used the next time the pg pod was cycled.

Anyway, I think this issue has already taken more time than it is worth, given both the non-issue to start with, and the possibility of us removing postgres at some point soon.

PR #10491, opened 2022-05-30
[bitnami/common] use -d flag for base64

### LINKED ISSUE #10486: --decode flag of base64 does not work on alpine linux
### Name and Version

bitnami/common

run any bitnami chart on alpine linux and secret handling will fail due to the lack of a base64 --decode flag

_No response_

base64 decode works

### What do you see instead?

base64 fails with error

### Additional information

_No response_

ak3io:
fixes #10486

change --decode flag to -d to solve a compatibility issue with alpine linux (--decode flag not available on alpine)

another example of this issue on another product: https://github.com/promptworks/aws-secrets/issues/14

### DIFF
--- a/bitnami/common/Chart.yaml
+++ b/bitnami/common/Chart.yaml
@@ -20,4 +20,4 @@ sources:
   - https://github.com/bitnami/charts
   - https://www.bitnami.com/
 type: library
-version: 1.14.1
+version: 1.14.2
--- a/bitnami/common/Chart.yaml
+++ b/bitnami/common/Chart.yaml
@@ -2,7 +2,7 @@ annotations:
   category: Infrastructure
 apiVersion: v2
 # Please make sure that version and appVersion are always the same.
-appVersion: 1.13.1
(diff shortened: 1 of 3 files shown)


bitnami/common/Chart.yaml
```diff
@@ -20,4 +20,4 @@ sources:
   - https://github.com/bitnami/charts
   - https://www.bitnami.com/
 type: library
-version: 1.14.1
+version: 1.14.2
```
carrodher, line 23: Could you please also bump the `appVersion` field to this same value? Both versions should match (although they were not matching at this moment)

carrodher (commented):
Thanks for your contribution, please see my minor comment, apart from that, everything LGTM

bitnami/common/Chart.yaml
carrodher: I did it by myself, so we can merge this PR

carrodher (approved):

PR #10510, opened 2022-05-31
[bitnami/solr] Fix Solr Exporter's container and service default ports

dfquaresma:
### Description of the change

The default port used in the Solr Exporter's container was the 9983, the one allocated to the Sia Exporter, as defined in [Prometheus default port allocations](https://github.com/prometheus/prometheus/wiki/Default-port-allocations).  Additionally, the Solr Exporter's service port was 8983, mismatching the one used in the container. 

In this PR we fix the mismatch of port configuration setting both container and service port to the same port, just like in the [default Zookeeper exporter](https://github.com/bitnami/charts/blob/master/bitnami/zookeeper/values.yaml#L667). Both ports are also set as the one allocated for Solr Exporter in the [Prometheus default port allocations](https://github.com/prometheus/prometheus/wiki/Default-port-allocations), 9231.

### Benefits

Speed up the process of setting up the Solr Exporter since this way we just need to set up the scrape configs in Prometheus to properly scrape the Solr Exporter in the default port.  

### Checklist


marcosbc (approved):

PR #10513, opened 2022-06-01
[bitnami/mongodb] - Fixing mongdb certificate generation issue when external access is enabled - #10262

### LINKED ISSUE #10262: Hostname/IP does not match certificate's altnames: IP: 127.0.0.1 is not in the cert's list - Part II
### Name and Version

bitnami/mongodb 12.1.5

Seems to be a problem since "mongo" was replaced with "mongosh" in #9916 and if you have tls.enabled: true and tls. authGenerated: true in your installation.
I also have set 3 IP's under externalAccess.service.loadBalancerIPs

only part of values.yaml that seems to be interesting in this case

<details>
<summary>values.yaml</summary>

```yaml
architecture: replicaset
## @param useStatefulSet Set to true to use a StatefulSet instead of a Deployment (only when `architecture=standalone`)
##
useStatefulSet: false
## MongoDB(&reg;) Authentication parameters
##
tls:
  ## @param tls.enabled Enable MongoDB(&reg;) TLS support between nodes in the cluster as well as between mongo clients and nodes
  ##
  enabled: true
  ## @param tls.autoGenerated Generate a custom CA and self-signed certificates
  ##
  autoGenerated: false
  ## @param tls.existingSecret Existing secret with TLS certificates (keys: `mongodb-ca-cert`, `mongodb-ca-key`, `client-pem`)
  ## NOTE: When it's set it will disable certificate creation
  ##
  existingSecret: ""
  ## Add Custom CA certificate
  ## @param tls.caCert Custom CA certificated (base64 encoded)
  ## @param tls.caKey CA certificate private key (base64 encoded)
  ##
  caCert: my-base64-encoded-CA-Cert
  caKey: my-base64-encoded-CA-Key
  ## Bitnami Nginx image
  ## @param tls.image.registry Init container TLS certs setup image registry
  ## @param tls.image.repository Init container TLS certs setup image repository
  ## @param tls.image.tag Init container TLS certs setup image tag (immutable tags are recommended)
  ## @param tls.image.pullPolicy Init container TLS certs setup image pull policy
  ## @param tls.image.pullSecrets Init container TLS certs specify docker-registry secret names as an array
  ## @param tls.extraDnsNames Add extra dns names to the CA, can solve x509 auth issue for pod clients
  ##
  image:
    registry: docker.io
    repository: bitnami/nginx
    tag: 1.21.6-debian-10-r103
    pullPolicy: IfNotPresent
    ## e.g:
    ## pullSecrets:
    ##
    pullSecrets: []

  ## e.g:
  ## extraDnsNames
  ##   "DNS.6": "$my_host"
  ##   "DNS.7": "$test"
  ##
  extraDnsNames: []
  
  ## @param tls.mode Allows to set the tls mode which should be used when tls is enabled (options: `allowTLS`, `preferTLS`, `requireTLS`)
  ##
  mode: requireTLS
## @param hostAliases Add deployment host aliases
## https://kubernetes.io/docs/concepts/services-networking/add-entries-to-pod-etc-hosts-with-host-aliases/
##
hostAliases: []
## @param replicaSetName Name of the replica set (only when `architecture=replicaset`)
## Ignored when mongodb.architecture=standalone
##
replicaSetName: replica-name
## @param replicaSetHostnames Enable DNS hostnames in the replicaset config (only when `architecture=replicaset`)
## Ignored when mongodb.architecture=standalone
## Ignored when externalAccess.enabled=true
##
replicaSetHostnames: true
## @param enableIPv6 Switch to enable/disable IPv6 on MongoDB(&reg;)
## ref: https://github.com/bitnami/bitnami-docker-mongodb/blob/master/README.md#enabling/disabling-ipv6
##
## configuration: |-

extraFlags: 
  - "--clusterAuthMode=x509"


externalAccess:
  ## @param externalAccess.enabled Enable Kubernetes external cluster access to MongoDB(&reg;) nodes (only for replicaset architecture)
  ##
  enabled: true
  ## External IPs auto-discovery configuration
  ## An init container is used to auto-detect LB IPs or node ports by querying the K8s API
  ## Note: RBAC might be required
  ##
  ## Parameters to configure K8s service(s) used to externally access MongoDB(&reg;)
  ## A new service per broker will be created
  ##
  service:
    ## @param externalAccess.service.type Kubernetes Service type for external access. Allowed values: NodePort, LoadBalancer or ClusterIP
    ##
    type: LoadBalancer
    ## @param externalAccess.service.portName MongoDB(&reg;) port name used for external access when service type is LoadBalancer
    ##
    portName: "mongodb"
    ## @param externalAccess.service.ports.mongodb MongoDB(&reg;) port used for external access when service type is LoadBalancer
    ##
    ports:
      mongodb: 27017
    ## @param externalAccess.service.loadBalancerIPs Array of load balancer IPs for MongoDB(&reg;) nodes
    ## Example:
    ## loadBalancerIPs:
    ##   - X.X.X.X
    ##   - Y.Y.Y.Y
    ##
    loadBalancerIPs:
    ## @param externalAccess.service.loadBalancerSourceRanges Address(es) that are allowed when service is LoadBalancer
      - 10.0.0.1
      - 10.0.0.2
      - 10.0.0.3
    ## ref: https://kubernetes.io/docs/tasks/access-application-cluster/configure-cloud-provider-firewall/#restrict-access-for-loadbalancer-service
    ## Example:
    ## loadBalancerSourceRanges:
    ## - 10.10.10.0/24
    ##
    loadBalancerSourceRanges: []
    ## @param externalAccess.service.externalTrafficPolicy MongoDB(&reg;) service external traffic policy
    ## ref https://kubernetes.io/docs/tasks/access-application-cluster/create-external-load-balancer/#preserving-the-client-source-ip
    ##
    externalTrafficPolicy: Local
    ## @param externalAccess.service.nodePorts Array of node ports used to configure MongoDB(&reg;) advertised hostname when service type is NodePort
    ## Example:
    ## nodePorts:
    ##   - 30001
    ##   - 30002
    ##
    nodePorts: []
    ## @param externalAccess.service.domain Domain or external IP used to configure MongoDB(&reg;) advertised hostname when service type is NodePort
    ## If not specified, the container will try to get the kubernetes node external IP
    ## e.g:
    ## domain: mydomain.com
    ##
    domain: ""
    ## @param externalAccess.service.extraPorts Extra ports to expose (normally used with the `sidecar` value)
    ##
    extraPorts: []
    ## @param externalAccess.service.annotations Service annotations for external access
    ##
    annotations:
      networking.gke.io/load-balancer-type: "Internal"
    ## @param externalAccess.service.sessionAffinity Control where client requests go, to the same pod or round-robin
    ## Values: ClientIP or None
    ## ref: https://kubernetes.io/docs/user-guide/services/
    ##
    sessionAffinity: None
    ## @param externalAccess.service.sessionAffinityConfig Additional settings for the sessionAffinity
    ## sessionAffinityConfig:
    ##   clientIP:
    ##     timeoutSeconds: 300
    ##
    sessionAffinityConfig: {}
```
</details>

- generated certificates should have added 127.0.0.1 as an ip address along with the ips coming from externalAccess.service.loadBalancerIPs
- "less lazier" matching for grep in all the probes to return real probing result

### What do you see instead?

I'm experiencing at least 2 issues that are somewhat certificate related:

1. like already described in #10104 - failing liveness probes. 
@joancafom already commited a fix but this fixes the issue only, if you dont have any ip's under externalAccess.service.loadBalancerIPs. If you do have any additional loader Balancer IP's there, the resulting certificate will not have added 127.0.0.1 as an ip address because it gets overwritten - here are all the details: https://github.com/bitnami/charts/issues/10104#issuecomment-1127450402

2. "startup-probe.sh" is false positive
The probe used the check the pods readiness is something like (in my case):
```
startup-probe.sh: |
    #!/bin/bash
    TLS_OPTIONS='--tls --tlsCertificateKeyFile=/certs/mongodb.pem --tlsCAFile=/certs/mongodb-ca-cert'
    mongosh  $TLS_OPTIONS --port $MONGODB_PORT_NUMBER --eval 'db.hello().isWritablePrimary || db.hello().secondary' | grep -q 'true'
``` 
due to the fact, that mongosh fails to connect to the mongo instance (see point 1. - missing ips in the certificate), the following happens. 
I've replaced the variables with the actual values and executed the command from inside a running mongo container. 
Everything seems to be ok, right? The commands return code is "0" but wait, what about that error message? 
```
I have no name!@mongodb-b2b-dev-2:/$ mongosh  --tls --tlsCertificateKeyFile=/certs/mongodb.pem --tlsCAFile=/certs/mongodb-ca-cert --port 27017 --eval 'db.hello().isWritablePrimary || db.hello().secondary' | grep -q 'true'
MongoServerSelectionError: Hostname/IP does not match certificate's altnames: IP: 127.0.0.1 is not in the cert's list: 10.43.0.34, 10.43.0.205, 10.43.0.206, 10.43.0.207
I have no name!@mongodb-b2b-dev-2:/$ echo $?
0
```
So, I've ran the command again but this time did not pipe the result to grep:

```
I have no name!@mongodb-b2b-dev-2:/$ mongosh  --tls --tlsCertificateKeyFile=/certs/mongodb.pem --tlsCAFile=/certs/mongodb-ca-cert --port 27017 --eval 'db.hello().isWritablePrimary || db.hello().secondary'
Current Mongosh Log ID:	6283692d22cac549de56d42b
Connecting to:		mongodb://127.0.0.1:27017/?directConnection=true&serverSelectionTimeoutMS=2000&tls=true&tlsCertificateKeyFile=%2Fcerts%2Fmongodb.pem&tlsCAFile=%2Fcerts%2Fmongodb-ca-cert&appName=mongosh+1.4.1
MongoServerSelectionError: Hostname/IP does not match certificate's altnames: IP: 127.0.0.1 is not in the cert's list: 10.43.0.34, 10.43.0.205, 10.43.0.206, 10.43.0.207
I have no name!@mongodb-b2b-dev-2:/$ echo $?
1
```
Now we can see the whole error message and also the reason why the return code was "0" before - although the mongosh connection itself fails - the "grep -q 'true'" just matches on one of the other "true" in the error message - like  "directConnection=true" or "tls=true"
Although the mongosh connection could never be established the probe ist returning a "0" and k8s thinks everything is ok...

3. readiness-probe.sh has the same problem .. mongosh connection is failing but "grep -q 'true'" returns a "0"

### Additional information

this is a follow up to #10104

rmuehlbauer:
### Description of the change

* increases starting value to "index=2" to prevent ip's in generated certs are overwritten if external access is enabled
* bumped chart version

### Applicable issues

  - fixes #10262

### Checklist


bitnami/mongodb/templates/common-scripts-cm.yaml
```diff
@@ -74,7 +74,7 @@ data:
     IP.0 = ${MY_POD_HOST_IP}
     IP.1 = 127.0.0.1
     EOL
-    index=1
+    index=2
```
joancafom, line 77: You need to also decrement the index for the DNS entries in line 84:
 https://github.com/bitnami/charts/blob/3f280328fb4281ca68d42965bcde7795094acd8c/bitnami/mongodb/templates/common-scripts-cm.yaml#L84

joancafom (changes requested):

bitnami/mongodb/templates/common-scripts-cm.yaml
rmuehlbauer: yes, you are right...didn't recognize

alemorcuq:
Could you please increase the chart version?

rmuehlbauer:
> Could you please increase the chart version?

I've increased the chart version

rafariossaa (approved):
LGTM

PR #10524, opened 2022-06-01
[bitnami/grafana-loki] Fix podAntiAffinity label for indexGateway StatefulSet

Dima-Diachenko-work:


Dima-Diachenko-work:
@carrodher 
CI failed due to some reason. 
What we should do in this case? :)

PR #10549, opened 2022-06-02
[bitnami/elasticsearch] Fix ingress service name

### LINKED ISSUE #10460: Elasticsearch chart v18.2.4 "Invalid value: restAPI" when enabling ingress
### Name and Version

bitnami/elasticsearch 18.2.4

update bitnami chart
`helm repo update`

Install elasticsearch with ingress enabled
`helm install test-es -n test bitnami/elasticsearch --set ingress.enabled=true`

I'm using kubernetes v1.21.5-gke.1200 

--set ingress.enabled=true

And ingress to be created

### What do you see instead?

You get this error message: 
`Error: Ingress.extensions "test-es-elasticsearch" is invalid: spec.rules[0].http.paths[0].backend.service.port.name: Invalid value: "restAPI": must contain only alpha-numeric characters (a-z, 0-9), and hyphens (-)
`

### Additional information

The workaround is to create an ingress manually and instead of using the `port.name: restAPI`, use `port.name: tcp-rest-api` as defined in the service.

tonny-m-telus-com:
The service name of the ingress was set to "restApi" which is an invalid value for Kubernetes. It should have been "tcp-rest-api".

Fixes #10460

Signed-off-by: Tonny Miousse <tonny.miousse@telus.com>

### DIFF
--- a/bitnami/elasticsearch/Chart.yaml
+++ b/bitnami/elasticsearch/Chart.yaml
@@ -25,4 +25,4 @@ name: elasticsearch
 sources:
   - https://github.com/bitnami/bitnami-docker-elasticsearch
   - https://www.elastic.co/products/elasticsearch
-version: 18.2.7
+version: 18.2.8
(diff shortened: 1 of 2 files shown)


marcosbc (approved):

PR #10553, opened 2022-06-02
[bitnami/mysql] Make auth.database Optional

yaskinny:
### Description of the change
Make `auth.database` Optional

### Benefits
sometimes people may just want to deploy the database with chart and  do the rest with something else like IAC tools.
I think It should be optional same as original MySQL image.

### Checklist


bitnami/mysql/templates/primary/statefulset.yaml
```diff
@@ -141,8 +141,10 @@ spec:
                   key: mysql-password
             {{- end }}
             {{- end }}
+            {{- if .Values.auth.database }}
             - name: MYSQL_DATABASE
               value: {{ .Values.auth.database | quote }}
+            {{- end }}
```
carrodher, line 147:
```suggestion
            {{- if .Values.auth.createDatabase }}
            - name: MYSQL_DATABASE
              value: {{ .Values.auth.database | quote }}
            {{- end }}
```

bitnami/mysql/values.yaml
```diff
@@ -102,7 +102,7 @@ auth:
   ## @param auth.database Name for a custom database to create
   ## ref: https://github.com/bitnami/bitnami-docker-mysql/blob/master/README.md#creating-a-database-on-first-run
   ##
-  database: my_database
+  database: ""
```
carrodher, line 105:
```suggestion
  database: my_database
  ## @param auth.createDatabase Wheter to create the database or not
  ## ref: https://github.com/bitnami/bitnami-docker-mysql/blob/master/README.md#creating-a-database-on-first-run
  ##
  database: true
```

carrodher (changes requested):
Thanks for your changes. I think this change can break the backward compatibility, what do you think about adding a new parameter to act as a flag for creating or not the database?

bitnami/mysql/values.yaml
yaskinny: Yeah it can handle backward compatibility, but needs more changes
like add schema validation to check if `.createDatabase` is `true` then database shouldn't be empty or something else like:

```
            {{- if .Values.auth.createDatabase }}
            - name: MYSQL_DATABASE
              value: {{ .Values.auth.database | quote }}
            {{- end }}
            ```

Comment (yaskinny) [reply bitnami/mysql/values.yaml:105]: okay but in template file it needs to check if both `.Values.auth.database` and `.Values.auth.createDatabase` is set. something like:
```
            {{- if and .Values.auth.createDatabase .Values.auth.database }}
            - name: MYSQL_DATABASE
              value: {{ .Values.auth.database | quote }}
            {{- end }}
```

I can do this and values file becomes something like:
```yaml
auth:
  ## @param auth.createDatabase Wheter to create the .Values.auth.database or not
  ## ref: https://github.com/bitnami/bitnami-docker-mysql/blob/master/README.md#creating-a-database-on-first-run
  ##
  createDatabase: false
  ## @param auth.database Name for a custom database to create
  ## ref: https://github.com/bitnami/bitnami-docker-mysql/blob/master/README.md#creating-a-database-on-first-run
  ##
  database: ""
```
also, should I add a conditional check in schema validation file to check for something like if `.Values.auth.createDatabase` is set then `.Values.auth.database` is not empty.

carrodher: Yes, there are more changes to be implemented in order to support that, the key point is that the default behavior of the Helm chart shouldn't be modified so the database with that name should be created when installing the chart from scratch using the default parameters.

yaskinny:
I added backward compatibility.

Now default value for `.Values.auth.createDatabase` is `true` and default value for `.Values.auth.database` is `my_database` as it was before.

Also in JSON schema validation I added to check if `.Values.auth.createDatabase` is set to `true`, then `.Values.auth.database` should not be empty.

PR #10560, opened 2022-06-03
Fix remaining Redis trademark references

mdhont:
Signed-off-by: michield <michield@vmware.com>


jotamartos (dismissed):
LGTM but check the image revision changes

fmulero (commented):
Hi @mdhont Thanks a lot! LGTM but why did you bump airflow version but not the rest with exactly the same changes?

PR #10561, opened 2022-06-03
[bitnami/owncloud] Add Helm Chart tests for Owncloud 

superaleks:
**Description of the change**
Increasing the test coverage of the Owncloud Helm Chart by adding Goss & Cypress tests. 

**Benefits**

**Possible drawbacks**

**Test results**

:gear: - [PASS](https://github.com/superaleks/charts/pull/30)

### Checklist


.vib/owncloud/cypress/cypress/support/commands.js
```diff
@@ -0,0 +1,35 @@
...
+}
+
+Cypress.Commands.overwrite('visit', (originalFn, url, options) => {
+  return originalFn(`${BASE_URL}${url}`, options);
+});
+
+Cypress.Commands.add(
+  'login',
+  (username = Cypress.env('username'), password = Cypress.env('password')) => {
+    cy.clearCookies();
```
joancafom, line 23: Unless explicitly needed to, I recall it was safe to delete this line as Cypress would indeed clear the cookies before every test run.

joancafom (commented):
Thanks for your PR! I left you some comments for review 😁

.vib/owncloud/cypress/cypress/integration/owncloud_spec.js
```diff
@@ -0,0 +1,67 @@
...
+it('allows modifying sharing settings', () => {
+  cy.login();
+  cy.visit('index.php/settings/personal');
+  cy.contains('Sharing').click();
+  cy.get('[for="allow_share_dialog_user_enumeration_input"]').click();
+  cy.contains('General').click();
+  cy.contains('Sharing').click();
+  cy.get('#allow_share_dialog_user_enumeration_input').should('not.be.checked');
+  cy.get('[for="allow_share_dialog_user_enumeration_input"]').click();
+});
```
FraPazGal, line 56: We are not really checking whether the new setting has been correctly enabled right?

.vib/owncloud/cypress/cypress/support/commands.js
```diff
@@ -0,0 +1,35 @@
...
+    cy.visit('/');
+    cy.get('.v-align').should('exist').and('be.visible'); //This should is needed to ensure stability of the test
```
FraPazGal, line 25: Is the concatenation needed? Is that different from `cy.get('.v-align').should('be.visible')`

FraPazGal (commented):
Thanks for the PR @superaleks! Could you please check my comments?

.vib/owncloud/cypress/cypress/integration/owncloud_spec.js
```diff
@@ -0,0 +1,67 @@
...
+
+it('allows whitelisting a domain', () => {
+  cy.login();
+  cy.visit('index.php/settings/personal');
+  cy.contains('Security').click();
+  cy.fixture('domains').then((domain) => {
+    cy.get('#domain').type(`${domain.newDomain.domain}`);
```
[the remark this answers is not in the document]
superaleks: I've checked, and there is no need to randomise in this case, hence I'm using a real domain (google.com).
The app doesn't check for duplicates.

.vib/owncloud/vib-publish.json
```diff
@@ -17,14 +17,57 @@
       ]
     },
     "verify": {
+      "context": {
+        "resources": {
+          "url": "{SHA_ARCHIVE}",
+          "path": "/bitnami/owncloud"
+        },
+        "runtime_parameters": "b3duY2xvdWRIb3N0OiB2bXdhcmUtb3duY2xvdWQubXkKb3duY2xvdWRVc2VybmFtZTogdXNlcgpvd25jbG91ZFBhc3N3b3JkOiBDb21wbGljYXRlZFBhc3N3b3JkMTIzITQKb3duY2xvdWRFbWFpbDogdGVzdEBleGFtcGxlLmNvbQpzZXJ2aWNlOgogIHR5cGU6IExvYWRCYWxhbmNlcgogIHBvcnRzOgogICAgaHR0cDogODAKc210cEhvc3Q6IHNtdHB0ZXN0aG9zdApzbXRwUG9ydDogOTgyMQpzbXRwVXNlcjogc210cFVzZXIKc210cFBhc3N3b3JkOiBzbXRwQ29tcGxpY2F0ZWRQYXNzd29yZDEyMyE0",
```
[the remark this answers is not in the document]
superaleks: It made sense that the chart fails with incomplete parameters (SMTP configuration partially set). 
I'll try it again and document it here.

.vib/owncloud/cypress/cypress/integration/owncloud_spec.js
```diff
@@ -0,0 +1,67 @@
...
+    cy.get('[value="New folder"]').type(`${folder.newFolder.name}.${random}`);
+    cy.contains('Create').click();
+    cy.contains('.innernametext', `${folder.newFolder.name}.${random}`).click();
+  });
+  cy.get('.new').click();
+  cy.get('[data-action="upload"]').click();
+  cy.get('input[type="file"]').selectFile(
+    'cypress/fixtures/file_to_upload.json',
+    { force: true }
+  );
```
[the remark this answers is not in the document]
superaleks: Indeed, I must have forgotten. I've now added a check for this.

.vib/owncloud/cypress/cypress/integration/owncloud_spec.js
```diff
@@ -0,0 +1,67 @@
...
+});
+
+it('allows adding a group and a user', () => {
+  cy.login();
+  cy.get('#expand').click();
+  cy.contains('Users').click();
```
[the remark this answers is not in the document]
superaleks: Same as above. Now it's in place!

.vib/owncloud/cypress/cypress/integration/owncloud_spec.js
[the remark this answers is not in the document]
superaleks: I've added a check for this. When the file is uploaded, the text that states `No files in here` should disappear.
Hence the check goes as follows:
`cy.contains('No files in here').should('not.be.visible');`

.vib/owncloud/cypress/cypress/integration/owncloud_spec.js
superaleks: Hey @FraPazGal, I have replaced the previously commented test with a file share test as per @joancafom comment. This means that the comment stops being relevant.

.vib/owncloud/cypress/cypress/integration/owncloud_spec.js
```diff
@@ -0,0 +1,67 @@
...
+  cy.contains('Add Group').click();
+  cy.fixture('groups').then((group) => {
+    cy.get('#newgroupname').type(`${group.newGroup.name}.${random}`);
+    cy.get('.icon-add').click();
```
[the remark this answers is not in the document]
superaleks: Thanks for the suggestion! I've made the change you implied. Please let me know if it's looking good.

.vib/owncloud/cypress/cypress/fixtures/links.json
```diff
@@ -0,0 +1,5 @@
+{
+  "newLinks": {
```
FraPazGal, line 2: Minor style comment, this should be in the singular tense.

.vib/owncloud/goss/goss.yaml
```diff
@@ -0,0 +1,27 @@
...
+    mode: "0770"
+    filetype: directory
+    exists: true
+  /opt/bitnami/owncloud/config/config.php:
+    mode: "0777"
+    filetype: symlink
+    exists: true
+    contains:
+      - dbtype
+      - dbuser
```
FraPazGal, line 26: Does this file contain the values of these parameters? If so, you could do something similar [to this test](https://github.com/FraPazGal/charts/blob/1031ef6e2dfee0d1db7a9c3ddb6339567e44bdc5/.vib/joomla/goss/goss.yaml#L18).

FraPazGal (commented):
I added a couple of new comments, could you please check them @superaleks?

.vib/owncloud/cypress/cypress/fixtures/links.json
superaleks: If I recall correctly, @joancafom and I agreed on a different PR to use the plural for this even though we have only one item due to consistency. Is my memory correct @joancafom?

.vib/owncloud/goss/goss.yaml
superaleks: Yes, but the syntax is a bit strange, look:

I'm not sure if this is possible and how. WDYT?

Screenshot 2022-06-14 at 15 09 41

.vib/owncloud/cypress/cypress/integration/owncloud_spec.js
```diff
@@ -0,0 +1,67 @@
...
+
+it('checks the SMTP configuration', () => {
+  cy.login();
+  cy.visit('index.php/settings/admin?sectionid=general');
+  cy.get('#mail_smtphost').should('have.value', Cypress.env('smtpHost'));
+  cy.get('#mail_smtpname').should('have.value', Cypress.env('smtpUser'));
+  cy.get('#mail_to_address').should('have.value', Cypress.env('owncloudEmail'));
+});
+
+it('allows modifying sharing settings', () => {
```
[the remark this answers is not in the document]
joancafom: Thanks! 🚀

.vib/owncloud/cypress/cypress/fixtures/links.json
joancafom: The plural is for the file name, `links`, as it might potentially have more than one link. If the item inside only contains a single link, I'd go with singular as well.

.vib/owncloud/goss/goss.yaml
joancafom: Yes, it is possible if you use a RegEx instead of regular text.

Check https://github.com/bitnami/charts/blob/df8bd7e304600cd5851516aeb7b9adad381b79f0/.vib/prestashop/goss/goss.yaml#L31

.vib/owncloud/cypress/cypress/integration/owncloud_spec.js
```diff
@@ -0,0 +1,78 @@
...
+  cy.get('.shareDialogLinkShare').then(() => {
+    cy.fixture('links').then((link) => {
+      cy.get('[name="linkName"]')
+        .clear()
+        .type(`${link.newLinks.name}0.${random}`);
+    });
+    cy.contains('.primary', 'Share').click();
+    cy.visit('/');
+    cy.contains('Shared by link').click();
+    cy.contains('No files in here').should('not.be.visible');
```
joancafom, line 65: This will only work if the test is executed once. The following times, you won't be able to tell if a new link was generated.

Why not perform a request to the link and check if it returns 2XX?

.vib/owncloud/goss/goss.yaml
FraPazGal: Mmm I think it should be possible without using a RegEx:

```suggestion
      - {{ printf "'dbuser' => '%s'" .Env.OWNCLOUD_DATABASE_USER }}
```

Nevertheless, you can also use a regex. If I'm not mistaken, it should be something like:

```suggestion
      - {{ printf "/dbuser.*%s/" .Env.OWNCLOUD_DATABASE_USER }}
```

.vib/owncloud/cypress/cypress/integration/owncloud_spec.js
```diff
@@ -0,0 +1,80 @@
...
+  });
+  cy.get('[class="permalink"]').click();
+  cy.get('.detailFileInfoContainer').within(() => {
+    cy.get('input')
+      .invoke('val')
+      .then((text) => {
+        const link = text;
+        cy.request({
+          method: 'GET',
+          url: link,
```
joancafom, line 62:
```suggestion
      .then((link) => {
        cy.request({
          method: 'GET',
          url: link,
```

joancafom: Additionally, it would be nice to have another extra assertion to ensure the link is indeed the one that downloads the file (and not some other `a` element Cypress matched the selector)

If the link contains the name of the file to download, maybe you could check it as an extra.

superaleks: No, the link doesn't contain the name of the file. 
However the link contains the name of the domain, so I added this assertion. Let me know what you think. 

I believe we invented a lot of time and effort into tests since I've been rewriting them a couple of times now. I hope you agree with this last version and like it as it is.

joancafom (dismissed):
LGTM!

.vib/owncloud/cypress/cypress/integration/owncloud_spec.js
joancafom: After checking it in a private call, the current approach does make sense

FraPazGal (dismissed):

joancafom (approved):

FraPazGal (approved):

PR #10580, opened 2022-06-05
Use descriptive name for extra VIB verify workflows

carrodher:

### Description of the change

When using an extra VIB verify pipeline to test some specific features/configuration of a Helm chart, the name used is simply _VIB_, the same name used by the generic VIB pipeline.

Currently, there are 3 VIB actions (the generic one and the other two with the extra tests for Redis and Thanos); but in the future, when this number grows, it won't be easy to navigate over the [actions view](https://github.com/bitnami/charts/actions).

### Benefits

There is a descriptive name to be shown in the [actions view](https://github.com/bitnami/charts/actions) instead of the current
Screenshot 2022-06-05 at 18 37 02


joancafom (approved):
LGTM!

PR #10595, opened 2022-06-06
[bitnami/elasticsearch] fix remaining "restAPI" in Ingress

### LINKED ISSUE #10460: Elasticsearch chart v18.2.4 "Invalid value: restAPI" when enabling ingress

tonny-m-telus-com:
A remaining "restAPI" service reference in ingress yaml file of Elasticsearch chart was missed during the first PR #10549.

Fixes #10460 

Signed-off-by: Tonny Miousse <tonny.miousse@telus.com>

### DIFF
--- a/bitnami/elasticsearch/templates/ingest/ingress.yaml
+++ b/bitnami/elasticsearch/templates/ingest/ingress.yaml
@@ -33,7 +33,7 @@ spec:
             {{- if eq "true" (include "common.ingress.supportsPathType" .) }}
             pathType: {{ .Values.ingest.ingress.pathType }}
             {{- end }}
-            backend: {{- include "common.ingress.backend" (dict "serviceName" (include "common.names.fullname" .) "servicePort" "restAPI" "context" $)  | nindent 14 }}
+            backend: {{- include "common.ingress.backend" (dict "serviceName" (include "common.names.fullname" .) "servicePort" "tcp-rest-api" "context" $)  | nindent 14 }}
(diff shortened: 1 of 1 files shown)


joancafom (changes requested):
Thanks for your PR @tonny-m-telus-com !

I have briefly checked and think there are two remaining reference to `restAPI` in another file:

https://github.com/bitnami/charts/blob/a6747a1a24aa074a417d1bc475d402a51e823f75/bitnami/elasticsearch/templates/ingest/ingress.yaml#L36

https://github.com/bitnami/charts/blob/a6747a1a24aa074a417d1bc475d402a51e823f75/bitnami/elasticsearch/templates/ingest/ingress.yaml#L49

Could you please update those values as well?

PR #10611, opened 2022-06-07
bitnami/keycloak

EinavDanielDX:
### Description of the change
Fixed spelling mistakes in statefulset.yaml and secrets.yaml that prevented the TLS feature to work properly.

Prior to this change - when trying to enable TLS, the user gets TPL error, indicates the function is expecting 2 arguments but receives only one (statefulset.yaml, lines 238 and 250).

In addition, the pod created by the statefullset had wrong set of JKS env params:
Before this change:
KEYCLOAK_TLS_KEYSTORE_PASSWORD:    <set to the key 'tls-keystore-password' in secret 'my-jks-secret'>
KEYCLOAK_TLS_TRUSTSTORE_PASSWORD:  <set to the key '**tls-keystore-password**' in secret ''my-jks-secret'>

After this change:
KEYCLOAK_TLS_KEYSTORE_PASSWORD:    <set to the key 'tls-keystore-password' in secret 'my-jks-secret'>
KEYCLOAK_TLS_TRUSTSTORE_PASSWORD:  <set to the key '**tls-truststore-password**' in secret ''my-jks-secret'>

### Benefits
TLS now works

### Checklist


fmulero (approved):
Brilliant!, thanks a lot @EinavDanielDX. LGTM

fmulero:
Thanks Einav for pointing out. I forced a new release, hopefully it will be available in a few hours

fmulero:
Hi Robert,

The env variables `KEYCLOAK_TLS_TRUSTSTORE_FILE` and `KEYCLOAK_TLS_TRUSTSTORE_PASSWORD` finally set the keycloak properties  `https-trust-store-file` and `https-key-store-file`, more information about those properties [here](https://www.keycloak.org/server/all-config#_httptls). 

For [LDAP over SSL keycloak documentation](https://www.keycloak.org/docs/latest/server_admin/#connecting-to-ldap-over-ssl) talks about setting the global truststore for Keycloak with the Truststore SPI or setting directly the JVM  system property `javax.net.ssl.trustStore`.

Currently we are not supporting the configuration for the Truststore SPI, but you can set `javax.net.ssl.trustStore` using this chart. To do it you can add the values `extraEnvVars` setting the JAVA_OPTS env var:
```
extraEnvVars:
  - name: JAVA_OPTS
    value: "-Djavax.net.ssl.trustStore=$KEYCLOAK_TLS_TRUSTSTORE_FILE -Djavax.net.ssl.trustStorePassword=$KEYCLOAK_TLS_TRUSTSTORE_PASSWORD"
```

bobertrublik:
Yes that's what I did to get it working :) Your link says the following:

> If you do not configure the Truststore SPI, the truststore falls back to the default mechanism provided by Java, which can be the file supplied by the javax.net.ssl.trustStore system property or the cacerts file from the JDK if the system property is unset.

I do have a truststore deployed either through `auth.tls.usePem` containing my CA bundle and client crt.key or `auth.tls.existingSecret` which contains a truststore and keystore. Therefore according to the docs it doesn't require the `javax.net.ssl.trustStore`property set. I don't understand why the error message then says

> No truststore provider found - using default SSLSocketFactory

Next:

> The Use Truststore SPI configuration property, in the LDAP federation provider configuration, controls the truststore SPI. By default, Keycloak sets the property to Only for ldaps, which is adequate for most deployments. Keycloak uses the Truststore SPI if the connection URL to LDAP starts with ldaps only.

My connection URL starts with ldaps and therefore it's supposed to use my truststore.

I don't see why I should set those env vars when all documentation points to my doing everything correctly? Also as I said it used to work using chart version 6.1.5.

fmulero:
In latest versions of  keycloak we can set different truststores files, for different porpouses, for example setting the Truststore SPI or setting the `https-trust-store-file` property. In current `bitnami/keycloak` container version we use the env variable `KEYCLOAK_TLS_TRUSTSTORE_FILE` to set the keycloak property `https-trust-store-file` which affects to the HTTP connections, no LDAP.

In java applications you can set also  the `javax.net.ssl.trustStore` system property as a global/default truststore 

Previous versions of `bitnami/keycloak` container and  keycloak product works in a different way, not sure what versions are you using but that could be the reason for that different behaviour.

PR #10672, opened 2022-06-09
[external-dns] Allow watching release namespace without hardcoding it, when `clusterRole: true`

waldner:
### Description of the change

This change makes it possible to restrict external-dns to watch only the release namespace, without having to hardcode its name in the values.

### Benefits

Currently, the only way to watch a specific namespace when using `clusterRole: true` is to hardcode the namespace in the chart values (eg `namespace: myns`), which is inconvenient when the chart meant to be used as a reusable subchart of another project and the namespace is globally specified in the `helm install` invocation.

### Possible drawbacks

`watchReleaseNamespace` is set to `false` by default, so backward compatibility should be preserved.

### Checklist


marcosbc (approved):

PR #10674, opened 2022-06-09
[bitnami/grafana-loki] Add Grafana Loki Helm Chart Tests

superaleks:
**Description of the change**
Increasing the test coverage of the Grafana Loki Helm Chart by adding Goss tests. 

**Benefits**

**Possible drawbacks**

**Test results**

:gear: - [PASS](https://github.com/superaleks/charts/pull/20)

### Checklist


.vib/grafana-loki/cypress/cypress/support/commands.js
```diff
@@ -0,0 +1,25 @@
+// ***********************************************
```
recena, line 1: Needed?

superaleks: Do you refer to the whole file or just this line? 
I would say we don't need the file at all, since we are just testing the API in this case.

recena: The whole file.

superaleks: I'll delete it, good catch!

.vib/grafana-loki/goss/goss.yaml
```diff
@@ -0,0 +1,25 @@
+file:
```
FraPazGal, line 1: I may be missing something but why are we checking only NGINX and its configuration? Aren't there any other tests related to grafana-loki itself that can be performed in any of the deployed pods? IIRC, there should be both a `loki` and `logcli` binaries in the pods.

joancafom: We are testing the API, which is served through `NGINX`. My point is that if Cypress tests succeed, that would imply that the API was responsive and it is highly likely that NGINX is properly configured then.

I don't have the knowledge background about Grafana-Locki, but maybe there are other pods to test that may be more interesting than the gateway itself (i.e. the suggested `loki` & `logcli` binaries). Else, there is no problem in testing NGINX here.

.vib/grafana-loki/cypress/cypress/integration/grafana_loki_spec.js
```diff
@@ -0,0 +1,71 @@
+/// <reference types="cypress" />
+import body from '../fixtures/logs-body.json';
```
joancafom, line 2: Is this the right way to import a fixture? Have you tried using `cy.fixture` or even `cy.readFile`?

.vib/grafana-loki/cypress/cypress/integration/grafana_loki_spec.js
```diff
@@ -0,0 +1,71 @@
...
+
+it('shows build info endpoint', () => {
+  cy.request({
+    method: 'GET',
+    url: '/loki/api/v1/status/buildinfo',
+    form: true,
+  }).as('buildInfo');
+  cy.get('@buildInfo').should((response) => {
```
joancafom, line 10: There is nothing wrong with this notation (aliasing), but conceptually they were introduced to access properties yielded by an async request at a later moment in time. Check: https://docs.cypress.io/guides/core-concepts/variables-and-aliases#Requests

When we are no performing any other operation in between, I personally consider the use of `then()` way neater than aliasing:

```suggestion
  }).then((response) => {
```

.vib/grafana-loki/cypress/cypress/integration/grafana_loki_spec.js
```diff
@@ -0,0 +1,71 @@
...
+});
+
+it('allows Loki to get Grafana logs', () => {
+  cy.request({
+    method: 'GET',
+    url: 'loki/api/v1/query?query={app="grafana-loki"}',
+    form: true,
+  }).as('grafanaLogs');
+  cy.get('@grafanaLogs').should((response) => {
+    expect(response.status).to.eq(200);
```
joancafom, line 29: Chances are this is returning a JSON file. Then, maybe it is worth to check the content-type is set to JSON as well:

```
expect(response.headers['content-type']).to.eq('application/json');
```

.vib/grafana-loki/cypress/cypress/integration/grafana_loki_spec.js
```diff
@@ -0,0 +1,71 @@
...
+    expect(response.body.data.stats.summary).not.to.be.empty;
+    expect(response.body.data.stats).to.include.all.keys(
+      'summary',
+      'querier',
+      'ingester'
+    );
+  });
+});
+
+it('can publish logs to the endpoint', () => {
```
joancafom, line 50: Maybe there would also be interesting to check if the logs are really published by retrieving them afterward.

joancafom (commented):
Thanks for the PR! I left you some comments, could you please review them?

.vib/grafana-loki/cypress/cypress/integration/grafana_loki_spec.js
superaleks: Yes, I've tried both for this case as far as I remember. 
In this case, I need to use the whole file and insert it into a body and AFAIK this is the only way.

.vib/grafana-loki/cypress/cypress/support/commands.js
superaleks: Actually, I've tried running the tests without this file and they will not compile. I'll put it back. This seems to be required for Cypress to work. 
Screenshot 2022-06-14 at 10 40 28

.vib/grafana-loki/cypress/cypress/support/index.js
```diff
@@ -0,0 +1,20 @@
...
+// You can change the location of this file or turn off
+// automatically serving support files with the
+// 'supportFile' configuration option.
+//
+// You can read more here:
+// https://on.cypress.io/configuration
+// ***********************************************************
+
+// Import commands.js using ES2015 syntax:
+import './commands';
```
dani8art, line 17: In fact, you could remove the file (support/commands.js) but you also need to remove this import statement.

superaleks: Oh OK! Excellent catch, On my way!
I forgot I've had this in the script.

.vib/grafana-loki/cypress/cypress/integration/grafana_loki_spec.js
superaleks: Thanks for the suggestion! I've added the line below.

.vib/grafana-loki/goss/goss.yaml
superaleks: I've added 2 binary tests in a separate file, testing a separate pod. 
Please let me know if it's OK for you.

.vib/grafana-loki/cypress/cypress/integration/grafana_loki_spec.js
superaleks: Applied the said suggestion.

.vib/grafana-loki/cypress/cypress/plugins/index.js
```diff
@@ -0,0 +1,22 @@
+/// <reference types="cypress" />
```
FraPazGal, line 1: We can remove this file, it shouldn't be needed in production.

.vib/grafana-loki/goss/goss.yaml
FraPazGal: I think this is a good solution! No issues from my side 👍

.vib/grafana-loki/cypress/cypress/integration/grafana_loki_spec.js
superaleks: I've investigated a lot of queries as we agreed in a separate call but I haven't found a reliable way of doing this which will not introduce flakiness/unpredictability. 
Therefore I applied the alternative which consists in adding a log and a label and checking for that.
Please take a look and let me know if that works for you.

FraPazGal (approved):
LGTM

.vib/grafana-loki/cypress/cypress/integration/grafana_loki_spec.js
```diff
@@ -0,0 +1,82 @@
...
+    );
+  });
+});
+
+it('can publish and retrieve a label', () => {
+  cy.request({
+    method: 'POST',
+    headers: { 'Content-Type': 'application/json; charset=utf-8' },
+    url: 'loki/api/v1/push',
+    body: body,
```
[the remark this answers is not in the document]
superaleks: Please take a look at the second line of the code.

joancafom (approved):
LGTM! Thanks for the PR 🚀

FraPazGal (approved):
LGTM!!

PR #10679, opened 2022-06-09
[bitnami/external-dns] Modify Cluster wide resources naming and add Helm Chart tests

superaleks:
**Description of the change**
Increasing the test coverage of the External Loki Helm Chart by adding Goss tests. 
P.S. This chart is not really testable with Goss. The POD only contains one binary. I've hence added 2 Goss tests only. 

**Benefits**

Now we can support multiple deployments in the same cluster.

**Possible drawbacks**

**Test results**

:gear: - [IN PROGRESS](https://github.com/superaleks/charts/pull/32)

### Checklist


bitnami/external-dns/Chart.yaml
```diff
@@ -24,4 +24,4 @@ sources:
   - https://github.com/kubernetes-sigs/external-dns
   - https://github.com/bitnami/bitnami-docker-external-dns
   - https://github.com/kubernetes-sigs/external-dns
-version: 6.5.6
```
joancafom, line 27: I think this is a minor, as you will be updating the name of resources.

bitnami/external-dns/templates/clusterrole.yaml
```diff
@@ -2,7 +2,7 @@
 apiVersion: rbac.authorization.k8s.io/{{ .Values.rbac.apiVersion }}
 kind: ClusterRole
 metadata:
-  name: {{ template "external-dns.fullname" . }}
+  name: {{ printf "%s-%s" (include "common.names.fullname.namespace" .) "external-dns" }}
```
joancafom, line 5: I think what you intended to do is:

```suggestion
  name: {{ template "common.names.fullname.namespace" . }}
```

bitnami/external-dns/templates/clusterrolebinding.yaml
```diff
@@ -2,7 +2,7 @@
 apiVersion: rbac.authorization.k8s.io/{{ .Values.rbac.apiVersion }}
 kind: ClusterRoleBinding
 metadata:
-  name: {{ template "external-dns.fullname" . }}
+  name: {{ printf "%s-%s" (include "common.names.fullname.namespace" .) "external-dns" }}
```
joancafom, line 5: Same here (and below in line 17)

.vib/external-dns/goss/goss.yaml
```diff
@@ -0,0 +1,11 @@
+file:
+  /opt/bitnami/external-dns/bin:
+    mode: "0755"
+    filetype: directory
+    exists: true
+command:
+  user-id-test:
```
joancafom, line 7: I'd add another extra test to check that the version matches:

```
$ external-dns --version
v0.12.0
```

joancafom: Also, you may check that port `7979` (metrics) is in use 😁 WDYT?

joancafom (changes requested):
Thanks for the PR @superaleks 

You are missing the naming change for other cluster-wide resources (that are not created when using your `runtime_parameters`, but still would fail to deploy if they were):

- _psp.yaml_
- _psp-clusterrolebinding.yaml_
- _psp-clusterrole.yaml_

Could you please include them as well?

.vib/external-dns/goss/goss.yaml
superaleks: Added both of those tests!

bitnami/external-dns/Chart.yaml
superaleks: @joancafom I've updated this but I'm not sure if I did it correctly now. Can you please check?

.vib/external-dns/goss/goss.yaml
```diff
@@ -0,0 +1,27 @@
+port:
+  tcp6:7979:
```
superaleks, line 2: I would have done this using templating but this port is not stated as an environment variable.

superaleks, line 2: I would have done this using templating but this port is not stated as an environment variable.

bitnami/external-dns/templates/psp-clusterrolebinding.yaml
```diff
@@ -3,7 +3,7 @@
 apiVersion: {{ include "common.capabilities.rbac.apiVersion" . }}
 kind: ClusterRoleBinding
 metadata:
-  name: {{ template "external-dns.fullname" . }}-psp
```
superaleks, line 6: @joancafom  @FraPazGal 
I'm not sure if I should leave the `-psp` or remove it as well?

FraPazGal: In these cases, you need to add `-psp`. You can do so using like:

```
{{ printf "%s-%s" (include "common.names.fullname.namespace" .) "psp" }}
```

.vib/external-dns/goss/goss.yaml
```diff
@@ -0,0 +1,27 @@
...
+    filetype: directory
+    exists: true
+command:
+  user-id-test:
+    exec: if [ "$(id -u)" -eq 0 ]; then exit 1; fi
+    exit-status: 0
+    stdout: []
+    stderr: []
+  external-dns-version:
+    exec: external-dns --version
```
joancafom, line 18: You can take the chance and check that it matches the value in $APP_VERSION:

```suggestion
    exec: external-dns --version 2>&1 | grep --quiet "$APP_VERSION"
```

bitnami/external-dns/Chart.yaml
joancafom: Yes, it is correct now! Thanks 🚀

bitnami/external-dns/templates/psp-clusterrolebinding.yaml
superaleks: I´ve added this line of code to wherever it seems it will fit.

bitnami/external-dns/templates/psp-clusterrole.yaml
```diff
@@ -17,5 +17,5 @@ rules:
   resources: ['podsecuritypolicies']
   verbs:     ['use']
   resourceNames:
-  - {{ template "external-dns.fullname" . }}
+  - {{ printf "%s-%s" (include "common.names.fullname.namespace" .) "psp" }}
```
joancafom, line 20: You should be using the other form

```suggestion
  - {{ template "common.names.fullname.namespace" . }}
```

bitnami/external-dns/templates/psp.yaml
```diff
@@ -3,7 +3,7 @@
 apiVersion: policy/v1beta1
 kind: PodSecurityPolicy
 metadata:
-  name: {{ template "external-dns.fullname" . }}
+  name: {{ printf "%s-%s" (include "common.names.fullname.namespace" .) "psp" }}
```
joancafom, line 6: You should be using the other form

```suggestion
  name: {{ template "common.names.fullname.namespace" . }}
```

.vib/external-dns/vib-publish.json
```diff
@@ -31,13 +31,22 @@
...
+          "action_id": "goss",
+          "params": {
+            "resources": {
+              "path": "/.vib/external-dns/goss"
+            },
+            "remote": {
+              "workload": "deploy-external-dns"
+            }
+          }
+        },
```
joancafom, line 44: I don't really think this makes a difference in functioning, but usually `trivy` is the first action defined in the array of the verification step. 

We append new actions to the end of the existing array (I think this is the convention for the other files) so I'd move it to the end 😁.

bitnami/external-dns/templates/psp-clusterrole.yaml
superaleks: I was using this form initially, but it was suggested by @FraPazGal I should change it to what you see here.
Could you guys agree on what you see fit for the task?
I am totally not an expert so I am following your guidance her.e

bitnami/external-dns/templates/psp.yaml
superaleks: Same comment as above.

bitnami/external-dns/templates/psp-clusterrole.yaml
joancafom: The comment suggested by @FraPazGal is right, but it refers to those lines previously referencing the `-psp` suffix. Lines that did not have the suffix should not add it in this version either.

We are both stating the same thing here.

bitnami/external-dns/templates/psp.yaml
superaleks: Now that I realised the mistake I made, I fixed it.
I added the suffix where it wasn't needed.

bitnami/external-dns/templates/psp-clusterrole.yaml
superaleks: Now that I realised the mistake I made, I fixed it.
I added the suffix where it wasn't needed.

joancafom (approved):
LGTM!

PR #10734, opened 2022-06-13
Automate merge process of CI's PRs

FraPazGal:

### Description of the change

This PR, in conjunction with the necessary internal changes to our CI, allows for the PRs created by `bitnami-bot` in the internal pipeline `Helm Charts (GitHub push)` stage to merge automatically once the `VIB Verify` workflow succeeds. 

For this, the internal pipeline will now add an `auto-merge` label and enable de auto-merge feature in its PRs. Once the `Verify` job is finished successfully, the workflow will approve the PR and it will be merged once any missing actions succeed.

If the `Verify` job fails, the related agents will be added as reviewers and the `auto-merge` label will be removed. Once said label is removed, the workflow won't be able to automatically approve the PR. As such, a reviewer will need to check the failures in the `Verify` workflow before rerunning it and approving the PR.

### Benefits

The PRs created by `bitnami-bot` as part of `Helm Charts (GitHub push)` CI stage no longer would require manual intervention unless an error in the `Verify` job appears.

### Additional information

Tested internally with a `bitnami/charts` fork as the testing environment. A couple of PRs, 1 made to succeed and 1 to fail, can be found there:

- https://github.com/FraPazGal/charts/pull/25
- https://github.com/FraPazGal/charts/pull/30

> NOTE: the failed README action is an expected error related to missing credentials in the fork, though it doesn't impact the final result.

### Checklist

### DIFF
--- a/.github/workflows/vib-verify.yaml
+++ b/.github/workflows/vib-verify.yaml
@@ -135,6 +135,6 @@ jobs:
           --header 'authorization: Bearer ${{ secrets.GITHUB_TOKEN }}' \
           --header 'content-type: application/json' \
           --data '{
-            "reviewers": ["carrodher", "dgomezleon", "Mauraza", "marcosbc", "mdhont", "migruiz4"]
+            "team_reviewers": ["build-maintainers"]
             }' \
           --fail


.github/workflows/vib-verify.yaml
```diff
@@ -93,4 +93,48 @@ jobs:
...
           VIB_ENV_TARGET_PLATFORM: ${{ secrets.VIB_ENV_TARGET_PLATFORM }}
           # Alternative Target-Platform to be used in case of incompatibilities
-          VIB_ENV_ALTERNATIVE_TARGET_PLATFORM: ${{ secrets.VIB_ENV_ALTERNATIVE_TARGET_PLATFORM }}
\ No newline at end of file
+          VIB_ENV_ALTERNATIVE_TARGET_PLATFORM: ${{ secrets.VIB_ENV_ALTERNATIVE_TARGET_PLATFORM }}
+  ci-pr-review:
+    runs-on: ubuntu-latest
+    needs: vib-verify
+    if: |
+      always() &&
```
aruiz14, line 101: nit-pick: Is this equivalent to `if true &&`? Can be dropped?

FraPazGal: Yes, it is equivalent. In this case, we can't drop it as we want our `ci-pr-review` job to run after the `vib-verify` one runs (defined by `needs: vib-verify`) regardless of successful execution of `vib-verify`. Without it, this last job would be skipped if the previous one failed.

More info at: [GitHub Actions doc](https://docs.github.com/en/actions/using-jobs/using-jobs-in-a-workflow#defining-prerequisite-jobs)

aruiz14: Ah! I get it now. So it means "run regardless of previous step failures"
Thanks!

aruiz14 (dismissed):

carrodher (approved):

joancafom (approved):
LGTM!

.github/workflows/vib-verify.yaml
```diff
@@ -93,4 +93,48 @@ jobs:
           # Target-Platform used by default
           VIB_ENV_TARGET_PLATFORM: ${{ secrets.VIB_ENV_TARGET_PLATFORM }}
           # Alternative Target-Platform to be used in case of incompatibilities
-          VIB_ENV_ALTERNATIVE_TARGET_PLATFORM: ${{ secrets.VIB_ENV_ALTERNATIVE_TARGET_PLATFORM }}
\ No newline at end of file
+          VIB_ENV_ALTERNATIVE_TARGET_PLATFORM: ${{ secrets.VIB_ENV_ALTERNATIVE_TARGET_PLATFORM }}
+  ci-pr-review:
```
recena, line 97: We should have added a `name` here to avoid this:

Screenshot 2022-06-14 at 20 20 06

recena: IMO, the jobs **Get modified charts** and **ci-pr-review** don't belong to the VIB service.

FraPazGal: I do not have any strong thoughts against leaving only the **Verify** job in the VIB workflow. As for my thought process here, I ended up leaving it as part of the same VIB workflow due to the data the **ci-pr-review** job needs from the **Verify** one. 

GitHub doesn't provide an easy solution to pass data between workflows; we'll have to use workflow artifacts and the logic gets more complex. The same would happen if we extract the **Get modified charts**.

PR #10762, opened 2022-06-14
[bitnami/kube-state-metrics] Modify naming of cluster-wide resources

joancafom:
### Description of the change

The cluster-wide resources in this chart get their name from the `common.names.fullname` function, which only takes into account the deployment name. As such, when installing the same chart with the same deploy name but different namespaces, you get installation errors of the following type"

```
Error: rendered manifests contain a resource that already exists. Unable to continue with install: PodSecurityPolicy "alpha-kube-state-metrics" in namespace "" exists and cannot be imported into the current release: invalid ownership metadata; annotation validation error: key "meta.helm.sh/release-namespace" must equal "gamma": current value is "default"
```

To solve this, we'll use the new `common.names.fullname.namespace` function added in #10462.

### Benefits

Solves a chart design issue and allows for the actions of our VIB pipelines to run in parallel in the affected charts.

### Possible drawbacks

There shouldn't be any drawbacks. The upgrade process has been tested without issues.

### Additional information

To avoid modifying extensively the current chart, only those resources affected by this issue (cluster-wide ones) will be modified.

### Checklist

### DIFF
--- a/bitnami/kube-state-metrics/Chart.yaml
+++ b/bitnami/kube-state-metrics/Chart.yaml
@@ -23,4 +23,4 @@ name: kube-state-metrics
 sources:
   - https://github.com/bitnami/bitnami-docker-kube-state-metrics
   - https://github.com/kubernetes/kube-state-metrics
-version: 3.0.5
+version: 3.1.0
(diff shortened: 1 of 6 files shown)


FraPazGal (approved):
LGTM!

PR #10781, opened 2022-06-16
[bitnami/index] Archive full index

randradas:
### Description of the change

This change is a new branch that will not be merged with master. The purpose is to hold file called bitnami/index.yaml which will be the home for the archive of Bitnami Charts full index. The index is built once per day.

The workflow downloads the last 6-months index from charts.bitnami.com and merges it with the last full index available, removes the duplicates and checks that there is not dups remaining or any missing release.

### Benefits

We can offer a full index for everybody so the can use it to access every chart version, not only version from the last six months.

### Additional information

This work is part of the efforts to reduce the impact of cutting off the official Bitnami Charts index to 6-months.

### Checklist

- [N/A] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [N/A] Variables are documented in the values.yaml and added to the `README.md` using [readme-generator-for-helm](https://github.com/bitnami-labs/readme-generator-for-helm)


.github/workflows/archive-full-index.yaml
```diff
@@ -0,0 +1,146 @@
...
+    steps:
+      - id: download-artifact
+        name: Download artifacts
+        uses: actions/download-artifact@v3
+        with:
+          name: duplicates-index
+      - id: remove
+        name: Remove duplicates
+        # Removes duplicates per entry using 'digest' as value.
+        run: yq eval '.entries[] |= unique_by(.digest)' duplicates_index.yaml > index.yaml
```
beltran-rubo, line 76: I think there is not value on the "duplicated_index.yaml" intermediate one. Why not running both actions at the same step and get the full result?

.github/workflows/archive-full-index.yaml
```diff
@@ -0,0 +1,146 @@
...
+          EOF
+          yamllint -c config index.yaml
+      - id: check-no-dups
+        name: Checks there are not any duplicates
+        # Try to find duplicate digest attributes which would mean there are duplicates.
+        run: |
+          yq eval '.entries[][].digest' index.yaml | sort | uniq -d | ( ! grep sha256 )
+      - id: check-missing-releases
+        name: Checks there are not missing releases
+        # Available URLs should be fine if everything went well during the merge & deduplication.
```
beltran-rubo, line 120: Good verification

.github/workflows/archive-full-index.yaml
carrodher: +1

.github/workflows/archive-full-index.yaml
```diff
@@ -0,0 +1,146 @@
...
+      - id: download-artifact-indexes
+        name: Download artifacts
+        uses: actions/download-artifact@v3
+        with:
+          name: indexes
+      - id: download-artifact-archive-full-index
+        name: Download artifacts
+        uses: actions/download-artifact@v3
+        with:
+          name: archive-full-index
```
carrodher, line 99: Currently, we are uploading 3 intermediate artifacts: `indexes`, `duplicates-index`, and `archive-full-index`.
If we get rid of the intermediate `duplicates-index`, only two artifacts will be present: `indexes` and `archive-full-index`, which are the ones downloaded in this step. In that case, we can simplify those steps into a single one, since if no `name` is provided, all the present artifacts will be downloaded, see https://github.com/actions/download-artifact#download-all-artifacts

```suggestion
      - id: download-artifacts
        name: Download artifacts
        uses: actions/download-artifact@v3
```

MichelZ:
Will you also have an "index-only" branch? I see it currently has all the charts, too.

.github/workflows/archive-full-index.yaml
```diff
@@ -0,0 +1,146 @@
+name: Generate the full index archive for Bitnami Charts
+on:
+  schedule:
```
MichelZ, line 3: Any reason why you can't just execute it on push of the index branch?

on:
  push:
    branches:
    - index

.github/workflows/archive-full-index.yaml
randradas: I like it! I will get rid of the intermediate `duplicates-index`.

.github/workflows/archive-full-index.yaml
randradas: Hi @MichelZ , the idea is to run it once per day because usually the truncated index that is downloaded is generated several times per day so we need to generated new full index, at least, once per day in order to be on track.

MichelZ: Yes, but then you still "lag" behind the truncated index. Also why would you download the truncated one from CDN and not from the index branch?

agarcia-oss: The point regarding the CDN version of the index is a very good one, @randradas : no real reason to download that file externally when we can just use our repository version of that file, right?

.github/workflows/archive-full-index.yaml
randradas: thank you!

.github/workflows/archive-full-index.yaml
randradas: right, I'm changing this part to get the file internally from the repo.

.github/workflows/archive-full-index.yaml
randradas: Done!

.github/workflows/archive-full-index.yaml
randradas: done!

.github/workflows/archive-full-index.yaml
randradas: done!, very good idea :D

randradas:
The work on this PR has been moved to a new one [PR](https://github.com/bitnami/charts/pull/10827).

PR #10794, opened 2022-06-17
[bitnami/postgresql] fix postgresql.replicationPasswordKey secret key logic

AdrianAxelsson:
Signed-off-by: Adrian Axelsson <adrian.axelsson@hotmail.se>

### Description of the change

When deploying the current version of the postgresql chart with replication, without specifying an existing secret, postgresql.replicationPasswordKey will not default to "replication-password". 
This change fixes the if/else logic

### Benefits

The chart can be installed successfully with architecture=replication without having to specify an existing secret

### Applicable issues

  - fixes #

### Checklist

### DIFF
--- a/bitnami/postgresql/templates/_helpers.tpl
+++ b/bitnami/postgresql/templates/_helpers.tpl
@@ -108,6 +108,8 @@ Get the replication-password key.
         {{- printf "%s" (tpl .Values.global.postgresql.auth.secretKeys.replicationPasswordKey $) -}}
     {{- else if .Values.auth.secretKeys.replicationPasswordKey -}}
         {{- printf "%s" (tpl .Values.auth.secretKeys.replicationPasswordKey $) -}}
+    {{- else -}}
+        {{- "replication-password" -}}
     {{- end -}}
 {{- else -}}
     {{- "replication-password" -}}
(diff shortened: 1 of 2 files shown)


bitnami/postgresql/templates/_helpers.tpl
```diff
@@ -108,9 +108,9 @@ Get the replication-password key.
         {{- printf "%s" (tpl .Values.global.postgresql.auth.secretKeys.replicationPasswordKey $) -}}
     {{- else if .Values.auth.secretKeys.replicationPasswordKey -}}
         {{- printf "%s" (tpl .Values.auth.secretKeys.replicationPasswordKey $) -}}
-    {{- else -}}
-        {{- "replication-password" -}}
     {{- end -}}
+{{- else -}}
+    {{- "replication-password" -}}
```
marcosbc, line 113: Shouldn't we also keep the `else` condition that was removed above, in this PR? If we have both, we can basically leave "replication-password" as a fallback value whenever something is not properly configured.

marcosbc (changes requested):
Thanks for the contribution! Please have a look at my comments.

bitnami/postgresql/templates/_helpers.tpl
AdrianAxelsson: Good thinking, I've added the `else` condition again.

ComputerBaer: But then it should be done the same way for `adminPasswordKey` and `userPasswordKey`, right?

marcosbc (approved):
LGTM!

PR #10835, opened 2022-06-21
[bitnami/joomla] Add Goss and Cypress tests

FraPazGal:

**Description of the change**

Adds Cypress and Goss tests to the `bitnami/joomla` chart.

**Benefits**

Ensuring higher quality of the Helm chart.

**Possible drawbacks**

**Test results** - PASS

[Please see them in the history here](https://github.com/FraPazGal/charts/pull/12)

**Checklist**


.vib/joomla/cypress/cypress.json
```diff
@@ -0,0 +1,8 @@
+{
+  "baseUrl": "https://localhost/",
```
superaleks, line 2: Any reason for having `https` here?

.vib/joomla/cypress/cypress/integration/joomla_spec.js
```diff
@@ -0,0 +1,72 @@
...
+  cy.login();
+  cy.visit('/administrator/index.php?option=com_media');
+  cy.contains('Upload').click();
+  cy.get('input[type="file"]').selectFile(
+    'cypress/fixtures/images/post_image.png',
+    { force: true }
+  );
+  cy.contains('Item uploaded');
+});
+
```
superaleks, line 37: One idea to consider, is maybe we can do this as a part of the article creation test, and instead use this test to test the adding user's functionality.
I think it will be valuable because a CMS is usually used by a small business which will have a couple of users, so this would be the first thing they would need to do.

.vib/joomla/cypress/cypress/integration/joomla_spec.js
```diff
@@ -0,0 +1,72 @@
+/// <reference types="cypress" />
+import { random } from './utils';
+
+it('allows user to log in and log out', () => {
+  cy.login();
+  cy.contains('Username and password do not match').should('not.exist');
```
superaleks, line 6: It could be better to search for the presence of the alert here since the error might happen for other reasons as well. 
In that case, we would also be covering other cases of login failures.

.vib/joomla/cypress/cypress/integration/joomla_spec.js
```diff
@@ -0,0 +1,72 @@
...
+  cy.contains('User Menu').click();
+  cy.contains('Log out').click();
+  cy.get("[name='username']");
```
superaleks, line 9: I would rather search for the whole login form here for additional clarity when reading the test.

.vib/joomla/cypress/cypress/integration/joomla_spec.js
```diff
@@ -0,0 +1,72 @@
...
+    cy.get('#jform_sitename')
+      .scrollIntoView()
+      .clear({ force: true })
+      .type(`${siteSettings.newSiteSetting.sitename} ${random}`);
+  });
+  cy.contains('Save & Close').click();
+  cy.contains('Configuration saved');
+});
+
+it('checks plugin management and API operabilty', () => {
```
superaleks, line 51: This is a really intelligent and valuable test!

.vib/joomla/goss/goss.yaml
```diff
@@ -0,0 +1,26 @@
+file:
+  /opt/bitnami/joomla:
+    filetype: symlink
+    linked-to: /bitnami/joomla
```
superaleks, line 4: Haven't seen this before, but this is useful!

superaleks (changes requested):
Thanks for the PR @FraPazGal! 

I've seen some creativity there.
I've added some ideas and suggestions which you might implement.

.vib/joomla/cypress/cypress/integration/joomla_spec.js
```diff
@@ -0,0 +1,72 @@
...
+  cy.login();
+  cy.contains('User Menu').click();
+  cy.contains('Edit Account').click();
+  cy.contains('API Token').click({ force: true });
+  cy.get('#jform_joomlatoken_token')
+    .invoke('attr', 'value')
+    .as('APIToken')
+    .then((APIToken) => {
```
joancafom, line 59: I think you do not need _aliasing_ here. If I am not wrong, the value returned by `invoke` will be automatically redirected to the following stream:

```suggestion
    .invoke('attr', 'value')
    .then((APIToken) => {
```

.vib/joomla/goss/goss.yaml
```diff
@@ -0,0 +1,26 @@
...
+  /bitnami/joomla/logs:
+    exists: true
+    filetype: directory
+    mode: "0775"
+  /bitnami/joomla/configuration.php:
+    exists: true
+    filetype: file
+    mode: "0664"
+    contains:
+      - {{ printf "$db = '%s'" .Env.JOOMLA_DATABASE_NAME }}
```
joancafom, line 15: This is fine, but I personally prefer to use regex in order to avoid problems with syntax changes in the future. That is, I think that checking that there is a line that contains `db` and the actual name of the Database would suffice:
```suggestion
      - {{ printf "/db.*%s/" .Env.JOOMLA_DATABASE_NAME }}
```
WDYT?

.vib/joomla/goss/goss.yaml
```diff
@@ -0,0 +1,26 @@
...
+command:
+  user-id-test:
+    exec: if [ "$(id -u)" -eq 0 ]; then exit 1; fi
+    exit-status: 0
+    stdout: []
+    stderr: []
+  prestashop-version-ok:
```
joancafom, line 22:
```suggestion
  joomla-version-ok:
```

👀

joancafom (commented):
Thanks for the PR @FraPazGal ! Left you some thoughts 😁

.vib/joomla/cypress/cypress.json
FraPazGal: None, thanks for catching the typo!

.vib/joomla/cypress/cypress/integration/joomla_spec.js
```diff
@@ -0,0 +1,72 @@
...
+  cy.get('.button-new').click();
+  cy.fixture('articles').then((article) => {
+    cy.get('#jform_title')
+      .scrollIntoView()
+      .type(`${article.newArticle.title} ${random}`, {
+        force: true,
+      });
+  });
+  cy.contains('Save & Close').click();
+  cy.contains('Article saved');
```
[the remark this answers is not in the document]
FraPazGal: To avoid having complex URLs y modified the test so the created article is featured, which makes it appear in the home page.

.vib/joomla/cypress/cypress/integration/joomla_spec.js
FraPazGal: Thanks! I needed to obtain a bearer token or manually enable basic auth via a plugin. I think this approach was better.

.vib/joomla/cypress/cypress/integration/joomla_spec.js
FraPazGal: You are right, I actually missed this after trying different approaches. Thanks for catching this!

joancafom (approved):
LGTM! Thanks for the changes 🚀

superaleks (approved):
LGTM!

PR #10854, opened 2022-06-22
[bitnami/nginx] Add nginx Helm Chart tests 

superaleks:
**Description of the change**
Increasing the test coverage of the Nginx Helm Chart by adding Goss tests. 

**Benefits**

**Possible drawbacks**

**Test results**

:gear: - [PASS](https://github.com/superaleks/charts/pull/36)

### Checklist


.vib/nginx/goss/goss.yaml
```diff
@@ -0,0 +1,32 @@
+file:
+  /opt/bitnami/nginx/conf/nginx.conf:
```
joancafom, line 2: It may be interesting to check some of the configuration passed as `runtime_params` is applied here. Something like:

```
contains:
  - /listen\s+8080/
```

.vib/nginx/goss/goss.yaml
```diff
@@ -0,0 +1,32 @@
...
+    exit-status: 0
+    stdout: []
+    stderr: 
+      - "signal process started"
+  nginx-home:
+    exec: cat /opt/bitnami/nginx/html/index.html
+    exit-status: 0
+    stdout: 
+      - "If you see this page, the nginx web server is successfully installed"
+    stderr:
```
joancafom, line 32: I think it might be better to query the server with `curl` and then compare the result with the file in disk. That way, you are double-checking that the server is operational and it returns the expected file.

Another approach would be to move this to the `file` section (I think it makes more sense) & implement a simple `curl` test here, which IMHO is more valuable than checking the static index file.

joancafom (commented):
Thanks for the PR @superaleks ! I left you some thoughts, could you please take a look at them?

.vib/nginx/goss/goss.yaml
FraPazGal: I agree, as it is now this check is a bit redundant given that we also have the `nginx-config-files`. In it, we are implicitly checking the `nginx.conf` file exists.

.vib/nginx/goss/goss.yaml
superaleks: Indeed, this check is now applied as well.

FraPazGal (approved):
LGTM! Thanks for the PR @superaleks!

.vib/nginx/goss/goss.yaml
```diff
@@ -0,0 +1,40 @@
...
+      - "nginx.conf syntax is ok"
+      - "nginx.conf test is successful"  
+  nginx-reload:
+    exec: nginx -s reload
+    exit-status: 0
+    stdout: []
+    stderr: 
+      - "signal process started"
+  nginx-curl:
+    exec: curl localhost:8080
```
joancafom, line 36: If we are checking the `stdout` for a specific text, I'd change the command to something more dynamic:

```
diff <(curl --silent localhost:8080) /opt/bitnami/nginx/html/index.html
```

This will check that the index served by the server is actually the same as the one on disk (and it is fail-proof in case the text changes in the future!)

WDYT?

superaleks: I like the idea but this command returns status code 2. 
I'll try to debug it.
```

        "test_id" : "goss-0-11",
        "name" : "Command nginx-curl exit-status",
        "status" : "FAILED",
        "message" : "ERRORS: [Command: nginx-curl: exit-status: doesn't match, expect: [0] found: [2]] FAILURES: [Command: nginx-curl: exit-status: doesn't match, expect: [0] found: [2]] ",
```

superaleks: Removing this test as agreed due to the fact we shouldn't be using `curl`.

joancafom (approved):
LGTM!

joancafom (approved):

FraPazGal (approved):
LGTM!

PR #10914, opened 2022-06-29
[bitnami/wildfly] Allow appVersion as image tag

talbx:
### Description of the change

This change allows users - that use bitnami/wildfly as a dependency to their main chart - to use its appVersion as docker image tag.
### Benefits

Before, you had to explicitly set the image.tag, especially when you build your own wildfly docker images. With this change you can maintain your appVersion directly in the Chart.yaml. This also makes sense, since the Chart.yaml of bitnami/wildfly always corresponds to a version of the bitnami/wildfly docker image.

To have this flexibility seems to be a "best practice" to some of us.
source: https://jfrog.com/blog/helm-charts-best-practices/
### Possible drawbacks

- none
### Applicable issues

- none

### Additional information

- none
### Checklist

### DIFF
--- a/bitnami/wildfly/Chart.yaml
+++ b/bitnami/wildfly/Chart.yaml
@@ -26,4 +26,4 @@ name: wildfly
 sources:
   - https://github.com/bitnami/bitnami-docker-wildfly
   - http://wildfly.org
-version: 13.3.11
+version: 13.4.0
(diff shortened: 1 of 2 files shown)


pSub (approved):

marcosbc (changes requested):
Hi @talbx, we have standarized our charts to use templates in the following way:

```
{{ template "wildfly.image" . }}
```

If you want to improve the way to specify the tag, we should look into modifying the `_helpers.tpl` file instead.

PR #10917, opened 2022-06-29
Rename Verify/Publish github workflows

FraPazGal:

### Description of the change

This PR aims to use more comprehensible names for the former `vib-verify.yaml` and `vib-publish.yaml` pipelines. Although both pipelines do use the VIB service, that is only a specific job of the pipeline and thus we shouldn't be using `VIB` as the names for the pipelines as a whole. Additionally, we are also looking for a better UX using industry-wide terminology and clearer job names.

Note that the files under `.vib/` such as `vib-verify.json` and `vib-publish.json` remain unchanged. Those are indeed the VIB pipelines DSLs and only contain VIB-related data.

### Checklist

### DIFF
--- a/.github/workflows/cd-pipeline.yaml
+++ b/.github/workflows/cd-pipeline.yaml
@@ -1,4 +1,4 @@
-name: 'VIB Publish'
+name: 'CD Pipeline'
 on: # rebuild any PRs and main branch changes
   push:
     branches:
@@ -90,7 +90,7 @@ jobs:
     runs-on: ubuntu-latest
     needs: get-chart
     if: ${{ needs.get-chart.outputs.result == 'ok' }}
-    name: Publish
+    name: VIB Publish
     steps:
       - uses: actions/checkout@v2
         name: Checkout Repository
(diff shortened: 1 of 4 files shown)


.github/workflows/cd-pipeline.yaml
```diff
@@ -1,4 +1,4 @@
-name: 'VIB Publish'
+name: 'CD Pipeline'
```
recena, line 1: `CI Pipeline` just to avoid confusions. Technically we are releasing.

recena (approved):

.github/workflows/cd-pipeline.yaml
FraPazGal: After talking about this privately, it is true both pipelines have steps related to integration.

I thought over this and having both pipelines named as CI (even using different filenames) could be confusing. Another option would be to use CT (continuous testing) and CI to keep them differentiated but we are also automatically merging PRs in `vib-verify`.

After looking over the options and the pipelines definitions, I think keeping them as CI/CD is appropriate, even if not 100% accurate. We do merge PRs into master in `vib-verify` so it can be considered continuous integration (CI) and we publish the new chart's version to the Helm repository in `vib-publish`, which can be considered continuous delivery (CD).

This can be a bit subjective so I do not have strong thoughts on this and can be left to a majority vote among reviewers 😄

joancafom: In my opinion, we should keep differentiating between the pipeline that publishes and the one that only tests the changes/merge the PR.

joancafom (approved):

javsalgar (approved):

PR #10929, opened 2022-06-29
[bitnami/*] cd: Add Slack notify job to CD pipeline 

FraPazGal:

### Description of the change

As of now, we do not have any way to notify the team when a run of the newly renamed `CD Pipeline` workflow fails. As this workflow runs when a commit is merged into master, failures in it can be easily missed.

This PR adds the feature to notify when the `CD Pipeline` does not succeed (meaning the `VIB Publish` job fails or is skipped) via Slack. These notifications will be sent to a private Slack channel where every interested agent part of the "Fix build process" team is. 

Additionally, the `paths` directive is fixed so the workflow is triggered whenever a change in the whole `bitnami/CHARTNAME/**` folder is modified. This is a bugfix, as looking only for changes in the `Chart.yaml` files restricted when the workflow was triggered and allowed for changes without a version bump to be pushed to master without anyone knowing. This happened recently with https://github.com/bitnami/charts/commit/5e62b53282f56ca55420e96b39f6a7935d449c2a. Moreover, the `Get modified charts` job is the one that checks exactly that condition:

https://github.com/bitnami/charts/blob/0516afa9aa973ce043c2cd9798fcf1f051bf5768/.github/workflows/cd-pipeline.yaml#L68

**Update:** after https://github.com/bitnami/charts/pull/10969, now exists the possibility of merging a PR changing `.md` files without bumping the chart's version. Modified the CD Pipeline to avoid being triggered by commits only containing `.md` changes.

### Benefits

Improves awareness of CD pipeline health and fixes its triggering condition.

### Possible drawbacks

Possible increased noise due to false positives coming from stability issues of tests or testing service.

### Additional information

For now, cases like https://github.com/bitnami/charts/commit/5e62b53282f56ca55420e96b39f6a7935d449c2a where the merged PR didn't include a `Chart.yaml` because of GH not detecting conflicts will make the CD pipeline fail. This is expected and we'll study a way to reduce this occurrence if this is something fairly common. 

### Checklist


.github/workflows/cd-pipeline.yaml
```diff
@@ -141,3 +142,43 @@ jobs:
...
+    runs-on: ubuntu-latest
+    needs: vib-publish
+    if: always()
+    name: Notify unsuccessful CD run
+    steps:
+      - name: Notify in Slack channel
+        if: ${{ needs.vib-publish.result != 'success' }}
+        uses: slackapi/slack-github-action@v1.19.0
+        with:
+          channel-id: ${{ secrets.SLACK_CHANNEL_ID }}
```
joancafom, line 157: I would change the secret name to include something related to the CD process:

```suggestion
          channel-id: ${{ secrets.CD_SLACK_CHANNEL_ID }}
```

.github/workflows/cd-pipeline.yaml
```diff
@@ -141,3 +142,43 @@ jobs:
...
+                        "type": "mrkdwn",
+                        "text": "The CD pipeline for <${{ github.event.head_commit.url }}|bitnami/charts@${{ github.event.head_commit.id }}> did not succeed. Check the related <${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}|action run> for more information."
+                      }
+                    }
+                  ]
+                }
+              ]
+            }
+        env:
+          SLACK_BOT_TOKEN: ${{ secrets.SLACK_BOT_TOKEN }}
```
joancafom, line 184: Same here

```suggestion
          SLACK_BOT_TOKEN: ${{ secrets.CD_SLACK_BOT_TOKEN }}
```

joancafom (commented):
Thanks for the PR, this is great work!

I just added to minor comments and would like to hear your thoughts on them 😁, but the logic does LGTM!

.github/workflows/cd-pipeline.yaml
```diff
@@ -141,3 +142,43 @@ jobs:
           git config user.name "Bitnami Containers"
           git config user.email "bitnami-bot@vmware.com"
           git add bitnami/index.yaml && git commit -m "${chart}-${chart_version}: Update index.yaml" -s && git push
+
+  # If the CD Pipeline does not succeed we should notify the interested agents
+  slack-notif:
+    runs-on: ubuntu-latest
+    needs: vib-publish
+    if: always()
```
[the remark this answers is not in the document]
FraPazGal: Yes, we need for the `slack-notif` job to execute every time, regardless of the execution result of the `vib-publish` one. Without it, it would only run if the previous one succeeded because we are using `needs: vib-publish`. Ref: [GH's doc](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-not-requiring-successful-dependent-jobs)

.github/workflows/cd-pipeline.yaml
FraPazGal: Done! Also modified the ENVs names in the repo settings.

joancafom (approved):
LGTM!

javsalgar (approved):

PR #10933, opened 2022-06-29
Fix CI pipeline to avoid race condition in auto-merge PRs

FraPazGal:

### Description of the change

The latest changes to the auto-merge feature of automated PRs merged at https://github.com/bitnami/charts/pull/10907 also introduced a race condition that can make every action run skip the `VIB Verify` job. 

This has happened in https://github.com/bitnami/charts/pull/10930, where the action run triggered by labeling the PR with the `verify` label didn't start instantly and only started once the `readme-generator-for-helm` commit was pushed to the PR branch. As the pipeline checks in real-time if the latest branch commit is the one created for the readme-generator, the action run triggered by the `labeled` action saw that there was already said commit present and skipped the `VIB Verify` job.

To avoid this race condition we just need to add an additional condition so the job is skipped if the latest commit is the  `readme-generator-for-helm` one created by `bitnami-bot` and also the action that triggered that run was a `synchronize` one. This action type which is defined in the [documentation](https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#webhook-payload-object-35) as:

> synchronize (action): Triggered when a pull request's head branch is updated. For example, when the head branch is updated from the base branch, when new commits are pushed to the head branch, or when the base branch is changed.

### Benefits

Solves existing race condition in automated PRs.

### Checklist

### DIFF
--- a/.github/workflows/ci-pipeline.yaml
+++ b/.github/workflows/ci-pipeline.yaml
@@ -45,7 +45,7 @@ jobs:
           num_charts_changed="$(echo "$charts_dirs_changed" | grep -c "bitnami" || true)"
           num_version_bumps="$(echo "$files_changed_data" | jq -r '[.[] | select(.filename|endswith("Chart.yaml")) | select(.patch|contains("+version")) ] | length' )"
 
-          if [[ ${{ github.actor }} = "bitnami-bot" && "$latest_commit_message" == *"readme-generator-for-helm"*  ]]; then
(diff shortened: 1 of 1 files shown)


recena (approved):

recena:
@FraPazGal In the future, we will see if we can generate the README file in a different way. Maybe only when changes arrive into the stable branch? 🤔

FraPazGal:
I agree, having that particular pipeline run every time the `values.yaml` file is modified seems a bit unnecessary. It is true that having the pipeline run in the PR solves the issue of having to do additional commits into the stable branch once the PR's commit is out, but I think we should investigate possible alternatives.

PR #11049, opened 2022-07-05
[bitnami/rabbitmq] Allow set custom logging

JoniJnm:
### Description of the change

Allow not set RABBITMQ_LOGS env to set custom logging. When RABBITMQ_LOGS is set log parameters in rabbitmq.conf are ignored.

The [official documentation says](https://www.rabbitmq.com/logging.html#log-file-location):

> The environment variable takes precedence over the configuration file. When in doubt, consider overriding log file location via the config file. As a consequence of the environment variable precedence, if the environment variable is set, the configuration key `log.file` will not have any effect.

### Benefits

Allow set `log.*` parameters in `extraConfiguration` value.

### Possible drawbacks

I don't think so.

### Additional information

I have been testing `log.` parameters. When the env `RABBITMQ_LOGS` exists seems like rabbit doesn't read log config. So using the following config:

```yaml
extraConfiguration: |
  log.default.level = warning
  log.file = false
  log.console = true
  log.console.level = warning
  log.console.formatter = json
```

The format output is not JSON. It works fine by removing `RABBITMQ_LOGS` env.

### Checklist

### DIFF
--- a/bitnami/rabbitmq/Chart.yaml
+++ b/bitnami/rabbitmq/Chart.yaml
@@ -23,4 +23,4 @@ name: rabbitmq
 sources:
   - https://github.com/bitnami/bitnami-docker-rabbitmq
   - https://www.rabbitmq.com
-version: 10.1.12
+version: 10.1.13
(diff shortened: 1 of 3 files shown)


fmulero (approved):
Great @JoniJnm. LGTM

PR #11055, opened 2022-07-06
[bitnami/magento] Add Magento Helm Chart tests

superaleks:
TBD


.vib/magento/cypress/cypress/integration/magento_spec.js
```diff
@@ -0,0 +1,103 @@
...
+    cy.get('.field-name-firstname').type(
+      `${customer.newCustomer.firstName}.${random}`
+    );
+    cy.get('#lastname').type(`${customer.newCustomer.lastName}.${random}`);
+    cy.get('#email_address').type(`${random}.${customer.newCustomer.email}`);
+    cy.get('#password').type(customer.newCustomer.password);
+    cy.get('#password-confirmation').type(customer.newCustomer.password);
+  });
+  cy.contains('button', 'Create an Account').click();
+  cy.contains('Thank you for registering with Main Website Store.');
```
joancafom, line 42: I would reduce this to:

```suggestion
  cy.contains('Thank you for registering');
```

which is enough to now registration completed successfully.

.vib/magento/cypress/cypress/integration/magento_spec.js
```diff
@@ -0,0 +1,103 @@
...
+      .select('Main Website');
+
+    cy.get('[class="admin__control-multiselect"]').last().select('General');
+    cy.get('#save').click();
+    cy.get('[name="discount_amount"]').type(discount.newDiscount.amount);
+    cy.get('#save').click();
+  });
+  cy.contains('You saved the rule');
+  cy.get('#apply_rules').click();
+  cy.contains('Updated rules applied.');
```
joancafom, line 69:
```suggestion
  cy.contains('Updated rules applied');
```

.vib/magento/goss/goss.yaml
```diff
@@ -0,0 +1,30 @@
...
+file:    
+  /opt/bitnami/magento:
+    filetype: symlink
+    linked-to: /bitnami/magento
+    exists: true
+  /opt/bitnami/magento/package.json.sample:
+    filetype: file
+    exists: true
+    contains: 
+      - "Magento Commerce Inc."
```
joancafom, line 30: I don't think it is worth checking this file, it is just a sample.

I might be wrong since it's been ages since the last time I checked magento, but I think there was a configuration file somewhere. Maybe it is worth checking it instead.

joancafom (commented):
Thanks for your PR @superaleks ! I left some minor comments for you with some ideas that came to my mind :)

.vib/magento/cypress/cypress/integration/magento_spec.js
```diff
@@ -0,0 +1,103 @@
+/// <reference types="cypress" />
+import { random, allowDataUsage } from './utils';
+
+before(() => {
```
FraPazGal, line 4: I'm not sure how this `before` works as you are always logging in but then in some tests you call `cy.login()` again.  Does this generate any type of conflict?

.vib/magento/cypress/cypress/integration/magento_spec.js
```diff
@@ -0,0 +1,103 @@
...
+  cy.login();
+  allowDataUsage();
+});
+
+it('allows admin to add a product to the store', () => {
+  cy.get('#menu-magento-catalog-catalog').click();
+  cy.get('.item-catalog-products').click();
+  cy.contains('Salable Quantity');
+  cy.get('#add_new_product-button').click();
+  cy.fixture('products.json').then((product) => {
```
FraPazGal, line 14: Minor comment. Has being any changes to the `fixture` function? I think we don't add the file extension in any of our tests.

.vib/magento/cypress/cypress/integration/magento_spec.js
FraPazGal, line 69: Could we check whether the discount appears on a product on the public site?

.vib/magento/cypress/cypress/integration/magento_spec.js
```diff
@@ -0,0 +1,103 @@
...
+  cy.get('.item-promo-catalog').click();
+  cy.get('#add').click({ force: true });
+  cy.fixture('discounts.json').then((discount) => {
+    cy.get('[name="name"]').type(`${discount.newDiscount.ruleName}.${random}`);
+    cy.get('.admin__actions-switch-text').click({ force: true });
+    cy.get('[class="admin__control-multiselect"]')
+      .first()
+      .select('Main Website');
+
+    cy.get('[class="admin__control-multiselect"]').last().select('General');
```
FraPazGal, line 62: Would there be an issue applying more than one discount to the same part of the website? I'm thinking about any potential conflicts during retries.

.vib/magento/cypress/cypress/integration/magento_spec.js
```diff
@@ -0,0 +1,103 @@
...
+  cy.get('#menu-magento-backend-marketing').click();
+  cy.get('.item-newsletter-subscriber').click();
+  cy.fixture('customers.json').then((customer) => {
+    cy.contains(`${random}.${customer.newCustomer.email}`);
+  });
+});
+
+it('allows customizing the home page', () => {
+  cy.login();
+  cy.get('#menu-magento-backend-content').click();
```
FraPazGal, line 89: Just to confirm, is there no option to navigate via `cy.visit()` to the intended page instead of going through UI navigation? Same for the other tests.

.vib/magento/cypress/cypress/support/commands.js
```diff
@@ -0,0 +1,57 @@
...
+});
+
+Cypress.Commands.add(
+  'login',
+  (username = Cypress.env('username'), password = Cypress.env('password')) => {
+    cy.visit('/admin');
+    cy.get('#username').type(username);
+    cy.get('#login').type(password);
+    cy.get('.action-login').click();
+    cy.get('.page-title').should('have.text', 'Dashboard');
```
FraPazGal, line 45: A bit silly, but wouldn't `cy.contains('.page-title', 'Dashboard')` suffice? It's just to maintain consistency, as you are using several of these `contains(selector, text)` in other parts of the tests.

FraPazGal (commented):
Thanks for the PR @superaleks! I added a few comments, please check them when possible. Btw, I loved the test data 😄

.vib/magento/cypress/cypress/integration/magento_spec.js
superaleks: I thought this as well and after trying I've concluded it's not possible. 
This is because the application appends a key during the navigation.
If I try navigating to the URL directly without the key, nothing happens. 

Example

.vib/magento/cypress/cypress/support/commands.js
```diff
@@ -0,0 +1,57 @@
+const CLICK_DELAY = 0;
```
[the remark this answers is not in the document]
superaleks: No, it's actually a test I was running locally but I forgot to put it back. It should be `1000`.

.vib/magento/cypress/cypress/integration/magento_spec.js
superaleks: Indeed!

.vib/magento/cypress/cypress/integration/magento_spec.js
superaleks: I've added a `logout()` to the `before()` just to have a more uniform state.
Other than that, here are some things worth knowing: 

There are 2 types of users for this app

- Customer (no need for login)
- Admin (there is a need for login)

I am not doing the login in the customer tests since it is not needed.

The `before()` hook runs only once before all tests. 
I need a login in the `before()` since I need to allow data usage once before all the tests. In order to do that, the user requires to be logged in.
Hence, before all tests, I log in, allow data usage, and log out.
Then each test individually does or doesn't log in, depending if it's a customer or admin test.

.vib/magento/cypress/cypress/integration/magento_spec.js
```diff
@@ -0,0 +1,103 @@
...
+it('allows admin to add a product to the store', () => {
+  cy.get('#menu-magento-catalog-catalog').click();
+  cy.get('.item-catalog-products').click();
+  cy.contains('Salable Quantity');
+  cy.get('#add_new_product-button').click();
+  cy.fixture('products.json').then((product) => {
+    cy.get('[name="product[name]"]').type(
+      `${product.newProduct.productName}.${random}`
+    );
+    cy.get('[name="product[price]"]').type(product.newProduct.price);
```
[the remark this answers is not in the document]
superaleks: Yes, they do!
I've added a test for that as well, thanks for the suggestion 😄

.vib/magento/cypress/cypress/integration/magento_spec.js
superaleks: I would suggest reading [this comment](https://github.com/bitnami/charts/pull/11055#discussion_r917768681). Here I explained the steps needed to display products on the Magento customer page.
The same goes for discounts.

.vib/magento/goss/goss.yaml
```diff
@@ -0,0 +1,30 @@
+command:
```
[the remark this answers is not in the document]
superaleks: Now that I learned it has a CLI, I've added some appropriate tests. Thanks for the idea!

.vib/magento/goss/goss.yaml
superaleks: I've added a test that checks a real config file instead of checking this one. 
WDYT about the new test?

joancafom: I like this one now 😁 Thanks!

.vib/magento/cypress/cypress/integration/magento_spec.js
```diff
@@ -0,0 +1,103 @@
...
+    cy.get('[class="action-menu _active"]').within(() => {
+      cy.contains('Edit').click({ force: true });
+    });
+  });
+  cy.fixture('pages.json').then((page) => {
+    cy.get('input[name="title"]').type(`${page.newPage.pageTitle}.${random}`, {
+      force: true,
+    });
+  });
+});
```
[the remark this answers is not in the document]
joancafom: Fair enought, we don't want to have/maintain super complicated tests. It just came to my mind because it was a similar pattern we had before, but that's ok.

joancafom (approved):
LGTM!

PR #11080, opened 2022-07-07
[bitnami/kube-prometheus] Fix ingress TLS when secrets are provided

### LINKED ISSUE #11016: [bitnami/kube-prometheus] Ingress does not use provided secrets
### Name and Version

bitnami/kube-prometheus 8.0.9

Deploy the chart to any cluster with an ingress controller.

```yaml
prometheus:
  ingress:
    enabled: true
    hostname: prometheus.k8s-intra-rd.local
    ingressClassName: nginx
    pathType: Prefix
    tls: true
    secrets:
      - name: prometheus.k8s-intra-rd.local-tls
         certificate: "<certificate>"
         key: "<key>"
```

The ingress should use the provided secret to serve TLS.

### What do you see instead?

The ingress uses the default self-signed certificate, because the Ingress secret is created but not *used*. The problem is similar to https://github.com/bitnami/charts/issues/10567 (in fact, the ingress template appears to be almost identical), the TLS key is only enabled if a self-signed certificate is used or CSR annotations are provided.

The fix should also be similar.

### Additional information

 I'm sure there are more charts in this repository with the same issue, as this is the second time I've run into this problem.

Edit: indentation of values.

bh-tt:
### Description of the change
Create the ingress TLS configuration if prometheus.ingress.secrets is defined.

### Applicable issues

  - fixes #11016

### Additional information
Similar to https://github.com/bitnami/charts/pull/10671

### Checklist


bh-tt:
@javsalgar Something seems to fail in the pipeline, but I cannot see the actual failure. Locally the chart seems to render fine, setting only `prometheus.ingress.enabled=true` and a dummy secret for `prometheus.ingress.secrets`.

```
# Source: kube-prometheus/templates/prometheus/ingress.yaml
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: release-name-kube-promethe-prometheus
  namespace: "intranet"
  labels:
    app.kubernetes.io/name: kube-prometheus
    helm.sh/chart: kube-prometheus-8.0.10
    app.kubernetes.io/instance: release-name
    app.kubernetes.io/component: prometheus
  annotations:
spec:
  rules:
    - host: prometheus.local
      http:
        paths:
          - path: /
            backend:
              service:
                name: release-name-kube-promethe-prometheus
                port:
                  name: http
  tls:
    - hosts:
        - prometheus.local
      secretName: prometheus.local-tls
```

javsalgar:
Hi,

It is failing because of the `trivy` scanner. Let me re-trigger it.

javsalgar:
Hi,

Could you rebase from master? we performed changes in the actions

bh-tt:
Done.

migruiz4:
Hi @bh-tt,

First, I would like to clarify something about Ingress TLS in our charts.

We provide three different methods to configure ingressTLS:
- Using `ingress.tls=true` + `certManager` (using `ingress.annotation`).
- Using `ingress.tls=true` + `selfSigned`.
- Using `ingress.extraTls` + `ingress.secrets` (optional, secrets could be provided using other methods).

With your proposal, a new option would be available, which would be configuring TLS like this:
```
ingress:
  tls: true
  secrets:
    - name: prometheus.local-tls
       key: ...
       certificate: ...
```

But I'm worried about the following case:
- An user sets `tls.true`, but the provided secret name does not match `printf "%s-tls" .Values.prometheus.ingress.hostname`, which would cause ingress to fail because the secret does not exist.

Adding the logic to avoid that may be too problematic and add extra complexity, so I would suggest two options:
- Add a `NOTE` under the `ingress.secret` value at `values.yaml`, warning that the secret name must match {{ingress.hostname}}-tls in order to be used along `ingress.tls=true` when no certManager or selfSigned are being used.
- Revert the suggested change and instead add a `NOTE` under `ingress.secrets` that mentions it must be used along `ingress.extraTls`.

Either option is fine for me as far as it is clear for other users, @javsalgar could you please share your thoughts on the above?

bh-tt:
My preference is an extra NOTE about the secret name.

migruiz4 (approved):
The note and the changes look good to me, thank you very much for your contribution @bh-tt !

PR #11090, opened 2022-07-08
[bitnami/solr] enable Solr metrics to work when basic auth is enabled

philmtd:
### Description of the change

This PR enables the Solr metrics exporter when Solr authentication is enabled.

### Benefits

Users of Solr with enabled basic authentication can export Solr metrics.

### Possible drawbacks

* This will work when only Solr has basic authentication enabled. If ZooKeeper requires any authentication, this will not work in the current state.
* This change requires the `JAVA_OPTS` environment variable of the metrics exporter to be set according to the [Solr documentation](https://solr.apache.org/guide/solr/latest/deployment-guide/monitoring-with-prometheus-and-grafana.html). It is set per default but if a user decides to overwrite `JAVA_OPTS` of the metrics exporter it may not work.

### Checklist


philmtd:
Hi @rafariossaa, we see that the VIB Verify check failed but we don't know how to reproduce or fix this. To see more logs it seems like we'd need access to the VMWare Cloud console? Is there a way to run the tests locally? We could't find any info on this in the documentation.
Besides this and looking at the Solr tests and the changes made in this PR we cannot really understand why these tests fail but if there's anything wrong we'd be happy to fix them if we knew how.
We'd be thankful for all kinds of help!

philmtd:
Thank you for your reply! I've checked these logs yesterday and I guess the Cypress tests may be flaky because they failed yesterday for no apparent reason (it's unlikely this change has caused this) and today they're fine. 

We see in the logs that only the Trivy check fails. What can we do about this? We didn't change anything related to the container images and guess that these vulnerabilities also existed before this PR.

Could you help us out with this?

rafariossaa:
Hi,
  First time when it failed there was 2 issue, one in trivy and other. I retriggered the testing because it could be a transient error and it disappeared in this last run. 
  The trivy error can be ignored in this context because, as you said, it is in the image and not in your PR.
  We are working to make this experience better, sorry for the noise.

rafariossaa (approved):
LGTM

philmtd:
Thank you for your help! :)

philmtd:
Sure, done!

rafariossaa:
Hi,
  I am sorry, I gave you wrong instructions. You need to run `helm dependency update` to update properly the Chart.lock file.
  Could you do that ?

philmtd:
Hi, sorry, my bad! I just ran `helm dependency update` in the chart and pushed the changes.

rafariossaa (approved):
Thanks !
LGTM.

PR #11117, opened 2022-07-10
[bitnami/rabbitmq] Fix high CPU usage while idle

### LINKED ISSUE #11116: [bitnami/rabbitmq] RabbitMQ high CPU usage while idle
### Name and Version

bitnami/rabbitmq 10.1.11

Just run it and watch the CPU usage.

_No response_

_No response_

### What do you see instead?

High CPU usage while idle

### Additional information

The liveness and readiness probe execute a separate erlang process, which is very expensive.

This was reported (and fixed) in https://github.com/helm/charts/issues/3855 but not to bitnami.

orgads:
### Description of the change
Use REST APIs for liveness/readiness probes, instead of spawning
expensive erlang processes.

### Benefits
Negligible CPU usage.

### Possible drawbacks
* Might be less accurate

### Applicable issues
- fixes #11116

### Checklist

### DIFF
--- a/bitnami/rabbitmq/templates/statefulset.yaml
+++ b/bitnami/rabbitmq/templates/statefulset.yaml
@@ -266,9 +266,9 @@ spec:
           livenessProbe: {{- include "common.tplvalues.render" (dict "value" (omit .Values.livenessProbe "enabled") "context" $) | nindent 12 }}
             exec:
               command:
-                - /bin/bash
+                - sh
                 - -ec
-                - rabbitmq-diagnostics -q ping
+                - test "$(curl -f --user {{ .Values.auth.username }}:$RABBITMQ_PASSWORD 127.0.0.1:{{ .Values.containerPorts.manager }}/api/healthchecks/node)" = '{"status":"ok"}'
           {{- else if .Values.customLivenessProbe }}
           livenessProbe: {{- include "common.tplvalues.render" (dict "value" .Values.customLivenessProbe "context" $) | nindent 12 }}
           {{- end }}
@@ -276,9 +276,9 @@ spec:
           readinessProbe: {{- include "common.tplvalues.render" (dict "value" (omit .Values.readinessProbe "enabled") "context" $) | nindent 12 }}
             exec:
               command:
-                - /bin/bash
+                - sh
                 - -ec
-                - rabbitmq-diagnostics -q check_running && rabbitmq-diagnostics -q check_local_alarms
+                - curl -f --user {{ .Values.auth.username }}:$RABBITMQ_PASSWORD 127.0.0.1:{{ .Values.containerPorts.manager }}/api/health/checks/local-alarms
           {{- else if .Values.customReadinessProbe }}
           readinessProbe: {{- include "common.tplvalues.render" (dict "value" .Values.customReadinessProbe "context" $) | nindent 12 }}
           {{- end }}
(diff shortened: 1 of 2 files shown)


bitnami/rabbitmq/templates/statefulset.yaml
```diff
@@ -266,19 +266,19 @@ spec:
           livenessProbe: {{- include "common.tplvalues.render" (dict "value" (omit .Values.livenessProbe "enabled") "context" $) | nindent 12 }}
             exec:
               command:
-                - /bin/bash
+                - sh
                 - -ec
-                - rabbitmq-diagnostics -q ping
+                - test "$(curl -f --user {{ .Values.auth.username }}:$RABBITMQ_PASSWORD 127.0.0.1:{{ .Values.service.ports.manager }}/api/healthchecks/node)" = '{"status":"ok"}'
```
fmulero, line 271: This is running in the container, so we have to use the container port
```suggestion
                - test "$(curl -f --user {{ .Values.auth.username }}:$RABBITMQ_PASSWORD 127.0.0.1:{{ .Values.containerPorts.manager }}/api/healthchecks/node)" = '{"status":"ok"}'
```

bitnami/rabbitmq/templates/statefulset.yaml
```diff
@@ -266,19 +266,19 @@ spec:
...
           {{- end }}
           {{- if .Values.readinessProbe.enabled }}
           readinessProbe: {{- include "common.tplvalues.render" (dict "value" (omit .Values.readinessProbe "enabled") "context" $) | nindent 12 }}
             exec:
               command:
-                - /bin/bash
+                - sh
                 - -ec
-                - rabbitmq-diagnostics -q check_running && rabbitmq-diagnostics -q check_local_alarms
+                - curl -f --user {{ .Values.auth.username }}:$RABBITMQ_PASSWORD 127.0.0.1:{{ .Values.service.ports.manager }}/api/health/checks/local-alarms
```
fmulero, line 281: This is running in the container, so we have to use the container port
```suggestion
                - curl -f --user {{ .Values.auth.username }}:$RABBITMQ_PASSWORD 127.0.0.1:{{ .Values.containerPorts.manager }}/api/health/checks/local-alarms
```

fmulero (changes requested):
Hi @orgads, sorry but there is a little failure in your PR. Could you have it a look?

bitnami/rabbitmq/templates/statefulset.yaml
orgads: Done

fmulero (approved):
Great! LGTM. Thanks a lot @orgads

orgads:
The rest checks are not deprecated, only the liveness probe I used is. I preferred to use the deprecated version, because the current version requires vhost, which is %3f (/) by default, but I think can be customized.

Would you prefer that?

orgads:
I think we can use `/api/health/checks/virtual-hosts`. See the documentation [here](https://pulse.mozilla.org/api/index.html).

fmulero:
Thanks @orgads. I agree with you about the CPU usage, but I think is better to keep aligned with RabbitMQ documentation, and report this kind of issues to the upstream project

orgads:
Apparently someone else already asked about this. They **don't** recommend using the CLI for probes. See [here](https://groups.google.com/g/rabbitmq-users/c/sB0IBRdaM0g/m/tBmPQp3EBwAJ).

PR #11139, opened 2022-07-12
[bitnami/php-my-admin] Fix cert-manager feature flag behaviour, clarify tls secret config

### LINKED ISSUE #10867: phpmyadmin: cert-manager flag deprecated but needed
### Name and Version

bitnami/phpmyadmin 10.1.9

Install phpmyadmin in v10.1.9 with tls enabled and cluster-issuer annotation on the ingress ressource. Skip `cert-manager` Flag, as it is marked as deprecated.
See invalid certificate delivered instead of cert-manager generated one.

ingress:
  enabled: true

  hostname: pma.example.com

  annotations:
    cert-manager.io/cluster-issuer: prod-cluster-issuer

  tls: true

Valid, cert-manager generated Certificate will be delivered and managed in a secret called pma.example.com-tls.

### What do you see instead?

Invalid, self signed certifcate is delivered.

### Additional information

https://github.com/bitnami/charts/blob/master/bitnami/phpmyadmin/templates/tls-secrets.yaml#L22 still uses the cert-manager flag to decide, whether the self signed cert should be created and used.
As the flag is deprecated, it should at least exist a new check for the new, alternative way using annotations.

grieshaber:
### Description of the change

This changes make the use of cert-manager for handling of tls-secrets easier and adapts the templating behaviour to the description in the values file.
If the cert-manager annotation is present, no tls cert will be generated and the (deprecated) `certManager` flag can be skipped.

Addtionally, the use of a byo tls cert was made easier. Readme and values were adapted accordingly.
To achieve the above, a new flag `ingress.existingSecretName` was introduced in order to pass an existing tls secret. Before this change, you needed to match the exact name (<hostname>-tls) and somehow skip autogeneration via non-self-explanining flags.

### Benefits

Behaviour of tls secret generation should be much more understandable and easier configurable. Additionally, a mismatch between actual behaviour and description was removed.

### Possible drawbacks

non

### Applicable issues
  - fixes #10867

### Checklist


recena:
@grieshaber Could you resolve the conflicts and update your branch?

grieshaber:
Resolved 👍🏼

rafariossaa (approved):
Thanks for adding this.
LGTM

PR #11145, opened 2022-07-12
[bitnami/odoo] Add Helm Chart tests for Odoo

superaleks:
TBD


.vib/odoo/cypress/cypress/integration/odoo_spec.js
```diff
@@ -0,0 +1,83 @@
...
+  });
+});
+
+it('allows installing/uninstalling an application', () => {
+  cy.login();
+  cy.get('[title="Home Menu"]').click();
+  cy.contains('[data-menu-xmlid="base.menu_management"]', 'Apps').click();
+  cy.get('[title="Sales"]').within(() => {
+    cy.contains('Install').click();
+    cy.get('.o_blockUI').should('not.exist');
```
superaleks, line 26: This is a loader that appears after I click on installing the application. 
I want to make sure it's gone so I check for it in **two** places knowingly. 
The installation is slow so sometimes the loader doesn't appear at this step, but later.

mdhont (dismissed):
Lgtm!

.vib/odoo/cypress/cypress/integration/odoo_spec.js
```diff
@@ -0,0 +1,83 @@
+/// <reference types="cypress" />
+import { random } from '../support/utils';
+import 'cypress-real-events/support';
+
+it('allows creating a company', () => {
+  cy.login();
+  cy.get('[title="Home Menu"]').click();
+  cy.contains('Settings').click();
+  cy.contains('Users & Companies').click();
```
FraPazGal, line 9: Could we avoid some of this steps by using `cy.visit()` or is it needed to navigate through clicks?

.vib/odoo/cypress/cypress/fixtures/companies.json
```diff
@@ -0,0 +1,6 @@
+{
+  "newCompany": {
+    "name": "Thornfield",
+    "phone": "0054326557"
```
FraPazGal, line 4: I think this is not necessary; you are not using `newCompany.phone` in the tests.

.vib/odoo/cypress/cypress/fixtures/newUsers.json
```diff
@@ -0,0 +1,6 @@
+{
+  "newUser": {
+    "name": "Edward Rochester",
```
FraPazGal, line 3: It's a bit confusing having two fixture files so similar, but I think we are not using this `.name` in our tests.

.vib/odoo/cypress/cypress/integration/odoo_spec.js
```diff
@@ -0,0 +1,83 @@
...
+    cy.contains('Manage Users').click();
+    cy.contains(`${user.newUser.name}.${random}`);
+  });
+});
+
+it('allows uploading a photo of the admin', () => {
+  cy.login();
+  cy.contains('Administrator').click();
+  cy.contains('Preferences').click();
+  cy.get('.img').realHover().realClick();
```
FraPazGal, line 78: First time seeing this 😮! Could you please give a bit of context on this? A comment in the code could also be useful future-wise.

FraPazGal (commented):
Hi @superaleks, nice PR! I left some comments, please check them when available.

.vib/odoo/cypress/cypress/integration/odoo_spec.js
```diff
@@ -0,0 +1,83 @@
...
+import { random } from '../support/utils';
+import 'cypress-real-events/support';
+
+it('allows creating a company', () => {
+  cy.login();
+  cy.get('[title="Home Menu"]').click();
+  cy.contains('Settings').click();
+  cy.contains('Users & Companies').click();
+  cy.get('[data-menu-xmlid="base.menu_action_res_company_form"]').click();
+  cy.get('.o_list_button_add').click();
```
joancafom, line 11: The creation buttons are usually easily identifiable by their text (they are not really many of them on screen). Can we replace this with `cy.contains` instead?

.vib/odoo/cypress/cypress/integration/odoo_spec.js
```diff
@@ -0,0 +1,83 @@
...
+  cy.fixture('companies').then((company) => {
+    cy.get('[name="name"]').type(`${company.newCompany.name}.${random}`);
+    cy.contains('Save').click();
+    cy.contains('.oe_title', 'Company Name');
+    cy.contains('.oe_title', `${company.newCompany.name}.${random}`);
```
joancafom, line 16: Isn't this duplicated?

joancafom (commented):
Thanks for your PR @superaleks ! I left you some comments 😁

.vib/odoo/cypress/cypress/integration/odoo_spec.js
superaleks: No, I want to check for both texts in the same element. It is on purpose.

.vib/odoo/cypress/cypress/integration/odoo_spec.js
superaleks: The image upload element is not clickable unless we hover over it. 
So, we need to hover and then click. This was the only way I could hover and click properly, and it was performed using the Cypress plugins. 
[Please check this blog post from Cypress ambassador](https://filiphric.com/hover-in-cypress). This is where I took the reference from.

.vib/odoo/cypress/package.json
```diff
@@ -0,0 +1,5 @@
+{
+  "dependencies": {
+    "cypress-real-events": "1.7.1"
```
superaleks, line 3: @FraPazGal this is where the Cypress plugin you asked about is declared.

.vib/odoo/cypress/cypress/fixtures/newUsers.json
superaleks: Indeed, I don't know what was I thinking when I did this! Fixed!

.vib/odoo/cypress/cypress/integration/odoo_spec.js
superaleks: The URLs are really awkward and long so I would prefer to leave it as it is, please check here: 

Screenshot 2022-07-22 at 12 24 41

.vib/odoo/cypress/cypress/integration/odoo_spec.js
superaleks: I did!

agomezmoron:
Hi @FraPazGal @joancafom in order to keep the idea of reducing the number of tests but ensuring we test the integration (filesystem + database) I would keep only these tests:
* `allows inviting new users`: It will test the database integration and if there is any smtp one, too.
* `allows installing/uninstalling an application`: It will test if the user that runs the app needs any extra permissions in the filesystem to download things.

Having both of them, I strongly think:
* `allows uploading a photo of the admin`: we are testing the same than we covered with `allows installing/uninstalling an application` (at integration level).
*  `allows creating a user` and `allows creating a company`: we are testing again the database and it's tested in the other 2 tests we can keep.

agomezmoron (dismissed):
@FraPazGal now it looks much better. LGTM

.vib/odoo/goss/goss.yaml
```diff
@@ -0,0 +1,17 @@
+file:
+  /bitnami/odoo/data:
```
joancafom, line 2: I think we should check `/bitnami/odoo`, which is the path in which we are mounting the PVC:

https://github.com/bitnami/charts/blob/de983f69068c65b47e51ebdbec874accee0eb924/bitnami/odoo/templates/deployment.yaml#L85-L87

agomezmoron (dismissed):
Now looks much better. LGTM

joancafom (approved):
LGTM! Thanks for the changes 🚀

PR #11146, opened 2022-07-12
[bitnami/solr] Wrong files in integration directory

dani8art:
Signed-off-by: Daniel Arteaga Barba <darteaga@vmware.com>

Cypress the tool itself is considering all of the files in `.vib/solr/cypress/cypress/integration` a test spec and is trying to executed it. In this case `.vib/solr/cypress/cypress/integration/utils.js` is being executed by cypress and resulting in a empty report, look at the screenshot attached.

Screenshot 2022-07-12 at 16 28 35

### Description of the change

I've moved `.vib/solr/cypress/cypress/integration/utils.js` to `.vib/solr/cypress/cypress/support/utils.js` and then it worked successfully.

Screenshot 2022-07-12 at 16 33 02

### Benefits

No empty report is given.

### Possible drawbacks

N/A

### Applicable issues

N/A

### Additional information

A successful execution graph id = 81528ac3-6084-4b2a-9da1-d97d2b8d59bc

### Checklist

- [x] Title of the pull request follows this pattern [bitnami/<name_of_the_chart>] Descriptive title
- [x] All commits signed off and in agreement of [Developer Certificate of Origin (DCO)](https://github.com/bitnami/charts/blob/master/CONTRIBUTING.md#sign-your-work)

PR #11164, opened 2022-07-13
[bitnami/*] Move utilities to support folder

superaleks:
**Description of the change**

Since Cypress will execute all files in the `integration` folder, there can be an empty segment in the reports. 
As an improvement task, we agreed to move all files which are not the test suite itself to the `support` folder. 

Taking a look at the [following example](https://github.com/bitnami/charts/pull/11146) we can apply the same across all covered assets.

### Checklist


dani8art (dismissed):

.vib/spring-cloud-dataflow/cypress/cypress/support/spring_cloud_dataflow_spec.js
```diff
@@ -1,6 +1,9 @@
 /// <reference types="cypress" />
```
joancafom, line 1: As said, I reckon this file belongs to the `integration` folder.

joancafom (changes requested):
In addition to `spring-cloud-dataflow`, I think you are missing changes in the `suitecrm_spec.js` file to indicate the new location of `utils.js`.

.vib/spring-cloud-dataflow/cypress/cypress/support/spring_cloud_dataflow_spec.js
superaleks: Oops, great catch. I fixed this in both instances.

joancafom (approved):
LGTM!

dani8art (approved):

PR #11175, opened 2022-07-13
[bitnami/*] Make cd-pipeline retryable

joancafom:
### Description of the change

This PR makes our CD pipeline more resilient against concurrency failures. It decouples the _VIB Publish_ step from the _Update Index_ one, introducing retry-ability for the latter as well.

### Benefits

* The _Update Index_ step, which can be affected by concurrency errors, is now retried before marking the job as _failed_.
* The _Update Index_ step can be independently retried from the _VIB Publish_ one.
* Less human intervention in failed, automatic releases.

### Possible drawbacks

None

### Applicable issues

NA

### Additional information

NA

### Checklist

### DIFF
--- a/.github/workflows/cd-pipeline.yaml
+++ b/.github/workflows/cd-pipeline.yaml
@@ -129,7 +129,7 @@ jobs:
           path: index
           # The token is persisted in the local git config and enables scripts to run authenticated git commands.
           token: ${{ secrets.BITNAMI_BOT_TOKEN }}
-      - name: Install and configure aws-cli, helm and jq
+      - name: Install and configure aws-cli and helm
         run: |
           # AWS CLI
           sudo DEBIAN_FRONTEND=noninteractive apt-get install -y awscli
@@ -140,8 +140,6 @@ jobs:
           # helm
           HELM_TARBALL="helm-v3.8.1-linux-amd64.tar.gz"
           curl -SsLfO "https://get.helm.sh/${HELM_TARBALL}" && sudo tar xf "$HELM_TARBALL" --strip-components 1 -C /usr/local/bin
-          # jq
-          sudo curl -SsLf https://github.com/stedolan/jq/releases/download/jq-1.5/jq-linux64 -o /usr/local/bin/jq && sudo chmod +x /usr/local/bin/jq
       - id: update-index
         name: Fetch chart and update index
         run: |
--- a/.github/workflows/cd-pipeline.yaml
+++ b/.github/workflows/cd-pipeline.yaml
@@ -116,8 +116,6 @@ jobs:
     needs:
       - vib-publish
     name: Update charts index
-    outputs:
-      result: ${{ steps.update-index.outputs.result }}
     steps:
       - uses: actions/download-artifact@v3
         with:
@@ -171,10 +169,7 @@ jobs:
             git push && is_index_updated=1 || echo "Failed to push during attempt $attempts"
           done
 
-          if [[ $is_index_updated -eq 1 ]]; then
-            echo "::set-output name=result::ok"
-          else
-            echo "::set-output name=result::fail"
+          if [[ $is_index_updated -ne 1 ]]; then
             echo "Could not update the index after $max_attempts attempts"
             exit 1
           fi
@@ -188,7 +183,7 @@ jobs:
     name: Notify unsuccessful CD run
     steps:
       - name: Notify in Slack channel
-        if: ${{ needs.vib-publish.result != 'success' ||  needs.update-index.outputs.result != 'ok' }}
+        if: ${{ needs.vib-publish.result != 'success' ||  needs.update-index.result != 'success' }}
         uses: slackapi/slack-github-action@v1.19.0
         with:
           channel-id: ${{ secrets.CD_SLACK_CHANNEL_ID }}
--- a/.github/workflows/cd-pipeline.yaml
+++ b/.github/workflows/cd-pipeline.yaml
@@ -127,6 +127,8 @@ jobs:
         with:
           ref: 'index'
           path: index
+          # The token is persisted in the local git config and enables scripts to run authenticated git commands.
+          token: ${{ secrets.BITNAMI_BOT_TOKEN }}
       - name: Install and configure aws-cli, helm and jq
         run: |
           # AWS CLI
--- a/.github/workflows/cd-pipeline.yaml
+++ b/.github/workflows/cd-pipeline.yaml
@@ -90,7 +90,6 @@ jobs:
         with:
           script: |
               core.setFailed('${{ steps.get-chart.outputs.error }}')
-
   vib-publish:
     runs-on: ubuntu-latest
     needs: get-chart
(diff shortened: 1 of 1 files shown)


.github/workflows/cd-pipeline.yaml
```diff
@@ -138,24 +152,55 @@ jobs:
           # HACK: Obtain chart tarball out of the latest commit
           chart_version="$(yq ".version" charts/bitnami/${chart}/Chart.yaml)"
```
aruiz14, line 153: I think this can be fixed now, by using `jq`:
```
chart_version=$(jq -re '.actions|map(select(.action_id == "helm-publish"))[0] | .application.version' result.json)
```
Where `result.json` is part of the artifacts from the publish job.

Then, we will also be able to remove the checkout step for the master branch (line 126)

.github/workflows/cd-pipeline.yaml
```diff
@@ -138,24 +152,55 @@ jobs:
...
+            # Push changes
+            git add bitnami/index.yaml && git commit -m "${chart}-${chart_version}: Update index.yaml" -s
+            git push && is_index_updated=1 || true
+
+            if [[ $is_index_updated -eq 1 ]]; then
+              echo "::set-output name=result::ok"
+              break
+            else
+              echo "Failed to push during attempt $attempts"
+            fi
```
aruiz14, line 180: Will break inside the `if` finish the execution? I see you are unconditionally printing the fail message below.
What about this:

```suggestion
            git push && is_index_updated=1 || echo "Failed to push during attempt $attempts"
```

Then add the `$is_index_updated -eq 1` condition to the while?

.github/workflows/cd-pipeline.yaml
```diff
@@ -138,24 +152,55 @@ jobs:
...
-    needs: vib-publish
+    needs:
+      - vib-publish
+      - update-index
     if: always()
     name: Notify unsuccessful CD run
     steps:
       - name: Notify in Slack channel
-        if: ${{ needs.vib-publish.result != 'success' }}
+        if: ${{ needs.vib-publish.result != 'success' ||  needs.vib-publish.result != 'ok' }}
```
aruiz14, line 203: Why this difference? `success` vs. `ok`?

aruiz14 (changes requested):

.github/workflows/cd-pipeline.yaml
fmulero: According to the [documentation](https://docs.github.com/en/actions/learn-github-actions/contexts#needs-context), `ok` is not a valid result

joancafom: My mistake here! This should be `needs.update-index.result`. That way, we will be notified in the Slack channel either if the publishing process **or** the update index one has failed.

Is it clear now? I will update the code accordingly 😁

.github/workflows/cd-pipeline.yaml
joancafom: Yes, you are right. There was some strange logic in here. I have applied your suggestions now, thanks!

.github/workflows/cd-pipeline.yaml
joancafom: I have used a new logic and removed the `checkout` step as well as the dependency with the `get-chart` step 🚀 
Thanks for the feedback

fmulero (dismissed):
LGTM

.github/workflows/cd-pipeline.yaml
```diff
@@ -112,50 +111,91 @@ jobs:
...
+          else
+            echo "::set-output name=result::fail"
+          fi
+      - id: show-error
+        name: 'Show error'
+        if: ${{ steps.update-index.outputs.result == 'fail' }}
+        uses: actions/github-script@v3
+        with:
+          script: |
+              core.setFailed('There was a problem updating the index.')
```
FraPazGal, line 186: Is there a need to use the `output.result` here? Given that it can only succeed or fail, I think we can just make the job finish (when it succeeds) or fail (maybe a `exit 1`). That way we avoid both the output and the `show-error` job.

.github/workflows/cd-pipeline.yaml
```diff
@@ -112,50 +111,91 @@ jobs:
...
           VIB_ENV_S3_USERNAME: ${{ secrets.AWS_ACCESS_KEY_ID }}
           VIB_ENV_S3_PASSWORD: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+  update-index:
+    runs-on: ubuntu-latest
+    needs:
+      - vib-publish
+    name: Update charts index
+    outputs:
+      result: ${{ steps.update-index.outputs.result }}
+    if: ${{ needs.vib-publish.result == 'success' }}
```
FraPazGal, line 121: I think only having this conditional is the same as having the `needs: vib-publish`. With the `need`, this job will only run when `vib-publish` succeeds, so having an `if` to check the same thing is redundant.

.github/workflows/cd-pipeline.yaml
```diff
@@ -112,50 +111,91 @@ jobs:
...
           aws configure set source_profile default
-            # helm
+          # helm
           HELM_TARBALL="helm-v3.8.1-linux-amd64.tar.gz"
           curl -SsLfO "https://get.helm.sh/${HELM_TARBALL}" && sudo tar xf "$HELM_TARBALL" --strip-components 1 -C /usr/local/bin
-          # yq
-          sudo curl -SsLf https://github.com/mikefarah/yq/releases/download/v4.23.1/yq_linux_amd64 -o /usr/local/bin/yq && sudo chmod +x /usr/local/bin/yq
-      - name: Fetch chart and update index
+          # jq
+          sudo curl -SsLf https://github.com/stedolan/jq/releases/download/jq-1.5/jq-linux64 -o /usr/local/bin/jq && sudo chmod +x /usr/local/bin/jq
```
FraPazGal, line 143: AFAIK, you don't need to install `jq` here, as it already comes installed. You can check our [ci-pipeline](https://github.com/bitnami/charts/blob/f870f9cea589adec6387b1faf73066330a50eaf2/.github/workflows/ci-pipeline.yaml#L37) where we are using it with an out-of-the-box ubuntu-latest.

FraPazGal (commented):
This is a nice improvement! I added a few comments, could you take a look at them @joancafom?

.github/workflows/cd-pipeline.yaml
```diff
@@ -112,50 +111,91 @@ jobs:
...
+    steps:
+      - uses: actions/download-artifact@v3
+        with:
+          path: ~/artifacts
       # If we perform a checkout of the master branch, we will find conflicts with the submodules
       - uses: actions/checkout@v2
         with:
           ref: 'index'
           path: index
-          token: ${{ secrets.BITNAMI_BOT_TOKEN }}
```
FraPazGal, line 120: Thinking about this now, I'm not sure we can remove this. As I see it, we need write access to push anything to master in `bitnami/charts` and an unauthenticated user won't be able to do it. It's strange we are using this here for no reason.

joancafom: Hmm I will keep it just in case (and maybe because it is more resilient for the future) but in my own tests, removing the token != creating commits as an unauthenticated user. This can be related to the fact that we are only enforcing the signed-off (DCO) and not a real GPG-sign:

Screenshot 2022-07-25 at 16 39 52

.github/workflows/cd-pipeline.yaml
joancafom: Ohh, okay. Thanks for the heads-up 👍

.github/workflows/cd-pipeline.yaml
joancafom: Removed, thanks!

.github/workflows/cd-pipeline.yaml
joancafom: As we are not printing specific execution errors here, I think you're right to remove the `show-error` step. Updated.

FraPazGal (approved):
LGTM

fmulero (approved):
LGTM

Issue #8746, opened 2022-01-21
Intermittent slow page load times for Discourse


manning-ncsa:
My Discourse site (only a few dozen users and very little activity so far) exhibits intermittent slow page/component load times that can be anywhere from 3-20 seconds. This occurs approximately once every ten page loads (as I repeatedly press keys G, J to cycle through posts, for example). During these events, the spinner spins and the user has to simply wait. 

I have been watching the browser dev console and the server log as I navigate around trying to identify the cause of these slow load times, and I  have identified at least two of the culprits: `/timings` and `123.json?track_visit=true&forceLoad=true` (where `123` is the topic ID).

Changing the "flush timings secs" site setting does not seem to change the behavior.

Does anyone have advice how to troubleshoot this problem? A few seconds here and there would be tolerable, but I fear that these frequent 5-20 second load times will stifle use of the platform.

Our instance is deployed on Kubernetes using the [Bitnami Helm chart](https://bitnami.com/stack/discourse/helm), [Docker image `docker.io/bitnami/discourse:2.7.8-debian-10-r0`](https://github.com/bitnami/bitnami-docker-discourse/tree/2.7.8-debian-10-r0/2/debian-10). I tried [asking for help on the Discourse forum](https://meta.discourse.org/t/intermittent-slow-page-load-times/214187) but this is considered an unsupported deployment method.

carrodher:
Please, [note the Bitnami team is not actively monitoring GitHub Issues for this Helm chart](https://github.com/bitnami/charts/tree/master/bitnami/discourse#community-supported-solution). This issue will stay open for 20 days to allow the community to contribute, after 15 days without activity the issue will be marked as stale being closed after 5 days.

If you would like to contribute to this Helm chart by fixing a bug or adding a new feature, feel free to create a PR, the Bitnami team will be happy to review it and provide feedback.

manning-ncsa:
Thank you @carrodher. It seemed inappropriate to create a GitHub issue like this, but I only did it because the https://community.bitnami.com/ Discourse post editor would not let me create a topic there where I assumed it would receive more feedback.

leventov:
We have exactly the same problem. @manning-ncsa was you able to hunt down this issue?

manning-ncsa:
@leventov Unfortunately, I have had no luck identifying the cause or a solution. The effort it would take to start hacking the source code to debug on my own has been a prohibitive barrier for me so far.

aqua-pro:
[...]

PR #11190, opened 2022-07-14
[bitnami/discourse] Add env for passenger extraFlags and custom Nginx template

deprecastor:
Signed-off-by: Deprecastor Git <deprecastor@users.noreply.github.com>

### Description of the change

With this change, the Helm Chart can provide 'DISCOURSE_PASSENGER_EXTRA_FLAGS' environment variable to the docker image beneath to configure the passenger before start up. This change makes it possible to provide a custom nginx template to passenger.

### Benefits

Can work around performance issues with passenger. For example [this case](https://github.com/bitnami/charts/issues/8746).
The environment variable is already implemented in the docker image in use. Just need to expose it into the values file and deployment.yaml.
At the moment we just have to "exec" into the container and run commands to have changes temporally, but this is not acceptable workaround that has to applied every time the POD moves or is updated.
To have a config change in passenger, the exposed env var in the values is needed.
To have a config change in the nginx template, like we need in the case of the issue 8746 mentioned, we need the configMap to be mounted and then the config option "nginx-config-template" in combination with the mounted file is in use. Please read [this for the option needed](https://github.com/discourse/message_bus#passenger) and [this for the template thingy](https://www.phusionpassenger.com/library/config/standalone/intro.html#nginx-configuration-template).

Also [this case](https://github.com/bitnami/bitnami-docker-discourse/issues/172) is related to my problem and have a good analyses why this is happening.

Please keep in mind that [the provided solution](https://github.com/bitnami/bitnami-docker-discourse/issues/172#issuecomment-922673459) in that old thread of "issue 172" is possible anymore. Like many others, I can see the option in UI is gone, but you can set the same behavior by applying the solution via this exposed env var and the configMap the.
### Possible drawbacks

Static mount path custom template as configMap. Maybe the path is also not as the community would like to have it placed.

Another drawback is the config template itself. The user has to follow the provided documentation link of nginx standalone with passenger, to collect it. Then apply the changes and load it into a configMap and apply the name to the values file.

### Applicable issues

none

### Additional information

The option "DISCOURSE_PASSENGER_EXTRA_FLAGS" wide open, could have tight it down to only use by providing a custom nginx template to passenger. But I guess it is better to have this open for other use cases different from those that led to this PR. But if it is better to have it more closed, I could change that.

### Checklist


fmulero:
Hi @deprecastor, thanks a lot for your contribution!

Before assigning it to one of our engineers, could you review the PR? it seems some Tomcat changes has been included by mistake

deprecastor:
Hi @fmulero, thank you for your assist. I am about to fix that. I guess this is related to Helm version bump, I did this wrong and will fix it after my work shift has ended.

fmulero:
Not sure if this will help (please make a copy of your changes):
```
# d654e308bbfaa98b1b22643bd5c5d80f15b072a5 -> Parent commit before your changes
git reset d654e308bbfaa98b1b22643bd5c5d80f15b072a5 -- bitnami/tomcat
# Undo any change done
git checkout bitnami/tomcat
# Commit
git commit -asm "Undo tomcat changes"
# Merge master in your branch
git merge <upstream>/master
# Check changes
git diff <upstream>/master
```

deprecastor:
@fmulero Thanks a lot for your help. I followed your instructions and it worked!

deprecastor:
Is there anything I could do, to move on with this?

deprecastor:
I solved the merge conflict.

M4NTR4:
Would be great if someone could take a look and merge this :)

alemorcuq (commented):
Thank you for creating this PR! Can't this be done we the already existing `.Values.discourse.extraVolumeMounts` and `.Values.discourse.extraEnvVars` parameters? The main reason those parameters exist is to provide flexibility for extra custom configurations an avoid adding too many parameters to the chart. Is there anything I'm missing?

deprecastor:
Hi @alemorcuq thanks for you response. Well you are right, the extraVolumeMount makes half of the thing obsolete.
I'll try to load the passenger options to the extraEnvVars and test if this fits the solution. Maybe I can do the same there, I have to read into it.

deprecastor:
Well, I decided to do something else and move on.

PR #11191, opened 2022-07-14
[bitnami/spring-cloud-dataflow] use common tpl function instead of toYaml

ooms97:
Signed-off-by: umer <umer.anwar@nuance.com>

### Description of the change

Use common `common.tplvalues.render` function instead of just `toYaml`.

### Benefits

Allows the usage of templates instead of just outputting the yaml. Also aligns the chart where this function is being used [(example)](https://github.com/bitnami/charts/blob/master/bitnami/spring-cloud-dataflow/templates/server/deployment.yaml#L13).

### Additional information

This will allow us to evaluate custom logic and pass it on to the chart as values.

For example:
If we define the following variable:
```yaml
{{- define "common.volumeMount.myVolumeMount" }}
- name: my-volume-mount
  mountPath: {{ .Values.myVolumeMount.containerVolumeMountPath | default "/mnt/foo" }}
  readOnly: true
{{- end }} 
```

Then we want to be able to use it in the spring-cloud-dataflow chart:
```yaml
extraVolumeMounts: '{{ include "common.volumeMount.myVolumeMount" . }}'
```

### Checklist

### DIFF
--- a/bitnami/spring-cloud-dataflow/Chart.yaml
+++ b/bitnami/spring-cloud-dataflow/Chart.yaml
@@ -39,4 +39,4 @@ sources:
   - https://github.com/bitnami/bitnami-docker-spring-cloud-dataflow
   - https://github.com/bitnami/bitnami-docker-spring-cloud-skipper
   - https://dataflow.spring.io/
-version: 12.0.3
+version: 12.0.4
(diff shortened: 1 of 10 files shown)


joancafom (approved):
LGTM! Thanks for your contribution 🚀

PR #11204, opened 2022-07-15
[bitnami/*] Use secrets to define target-platforms

joancafom:
### Description of the change

Use a secret to define the `target-platform` instead of hardcoding the value. This also allows failed jobs to use updated values.

### Benefits

The `target-platform` can be modified dynamically, without having to explicitly alter the workflow file every time.

### Possible drawbacks

Traceability is lost, but failed jobs can be retried using a new `target-platform`.

### Applicable issues

None

### Additional information

NA

### Checklist


FraPazGal (approved):
LGTM

PR #11239, opened 2022-07-19
[bitnami/mongodb] Mongodb custom certs

yohboy:
### Description of the change
This PR adds parameters to provide custom certificates. I tried to add this feature without breaking any existing things. The certificate management of the MongoDB chart is relatively "special". The parameter `tls.autoGenerated` seems useless, but I tried not to touch it. Even if this parameter is set to false, it just cancels the CA auto-generation, but still self-generates certificates for each node. This PR adds 4 parameters to provide custom certificates for each node. Each parameter target a specific group of nodes (hidden, replicaset, arbiter, standalone). To not change the behavior of `tls.autoGenerated`, the chart automatically detects if it needs to auto-generate certificates.  
This PR also removes a useless key (`client-pem`) in the `tls.existingSecret` parameter.

### Benefits
Mongodb can now be configured with custom certificates, bypassing auto-generation of certificates.

### Possible drawbacks
None.

### Applicable issues
None.

### Additional information
None.

### Checklist


bitnami/mongodb/values.yaml
```diff
@@ -175,10 +175,30 @@ tls:
...
+  hidden:
+    ## @param tls.hidden.existingSecrets Array of existing secrets with TLS certificates (`tls.key`, `tls.crt`, `ca.crt`).
+    ## NOTE: When it's set it will disable certificate self-generation from existing CA.
+    ##
+    existingSecrets: []
+  arbiter:
+    ## @param tls.arbiter.existingSecret Existing secret with TLS certificates (`tls.key`, `tls.crt`, `ca.crt`).
+    ## NOTE: When it's set it will disable certificate self-generation from existing CA.
+    ##
+    existingSecret: ""
```
joancafom, line 201: This is one possible approach, but I believe it's kind of complex as it adds a lot of entries under the `tls` param (especially, considering that we already have entries for `hidden` and `arbiter`). 

This is because the MongoDB chart is complex itself, it has a lot of components and different replication configurations. Taking a look into Kafka (which is also a complex chart), we can see that TLS for its components was implemented under their specific entry in the `values.yaml` file (e.g. `provisioning.auth.tls`).

https://github.com/bitnami/charts/blob/1d59ac4394f68177dd349a2008675448bc82f081/bitnami/kafka/values.yaml#L1478-L1487

I think we should follow the same approach here and add `tls` entries for the different components. Moreover, we already follow this schema for other params such as _Service_:

https://github.com/bitnami/charts/blob/1d59ac4394f68177dd349a2008675448bc82f081/bitnami/mongodb/values.yaml#L1795-L1797

----

In that case of the main MongoDB pods, we have to bear in mind that we have two architectures: `standalone` and `replicaset`. 

I would simply use a single array param, `existingSecrets`, and consider `standalone` as a special case of `replicaset` where only a single entry is expected. WDYT?

joancafom (commented):
Big contribution, thank you so much. I have left you some feedback that may affect most of the lines. Could you please review it before I go on with further review? 😁

bitnami/mongodb/values.yaml
yohboy: Hi @joancafom, thank you very much for your feedback! I agree with you, it adds a lot of entry and I already considered a lot of cases when doing these changes. Let me add some more thoughts. At first, I added a single `existingSecrets` containing all secrets for all pods. Fewer parameters, but definitively too complex to manage and use.   
I then considered two options. The one you explained, and the one I implemented. I think each option has some advantages. When all `tls` parameters are in the same entries, it is easier to see what is required to configure tls. The elasticsearch chart is using this method: https://github.com/bitnami/charts/tree/master/bitnami/elasticsearch. All elasticsearch components (master, data, coordinating) required to run an elasticsearch cluster have an entry `security.tls.*component*`. On the other hand, having a specific `*component*.tls` can be interesting as well. In the case of the Kafka chart, it only needs brokers to run a Kafka cluster. The Kafka provisioning is just acting as a temporary client to configure the Kafka brokers. So it can make sense that the Kafka provisioning has its own `tls` entry.  
That says, in the case of MongoDB, to configure autogenerated TLS in all MongoDB components (hidden, replica, arbiter), you have to configure TLS in the main `tls` entry only. Nothing inside `hidden` and `arbiter` entries. So if we put the `tls` entry in `hidden` and `arbiter` entries, there are now two different cases and it may be confusing:
- Autogenerated CA or CA provided+autogenerated certs -> only `tls` entry
- Custom certs -> some entries in `tls` entry, some in `hidden.tls` entry, and some in `arbiter.tls` entry.  

That's why I finally choose to implement the `tls.*component*` entries solution. To my mind, it is not necessarily more complex, but definitively less confusing in this case. However, sharing a single secret array for both `standalone` and `replicaset` may be a good idea to reduce the number of parameters and the complexity :+1: Not sure how to do that to avoid confusion though.  What do you think? :)

joancafom: Okay, I see your points now and I personally like that we already have this implementation in another chart :)

Apologies for the delay, I will add this PR to my reviewing list and provide further feedback 😁

yohboy: Hi @joancafom, thank you! I'm waiting for your feedback then :D

joancafom (commented):
Hi @yohboy 

Thank you so much for your PR, I left you some comments on your changes 😁

bitnami/mongodb/values.yaml
```diff
@@ -175,10 +175,30 @@ tls:
...
+  ## NOTE: When it's set it will disable secret creation.
   ##
   existingSecret: ""
+  standalone:
+    ## @param tls.standalone.existingSecret Existing secret with TLS certificates (`tls.key`, `tls.crt`, `ca.crt`).
+    ## NOTE: When it's set it will disable certificate self-generation from existing CA.
+    ##
+    existingSecret: ""
+  replicaset:
+    ## @param tls.replicaset.existingSecrets Array of existing secrets with TLS certificates (`tls.key`, `tls.crt`, `ca.crt`).
```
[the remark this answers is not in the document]
yohboy: I added this suggestion on both arrays!

bitnami/mongodb/templates/common-scripts-cm.yaml
```diff
@@ -101,4 +103,19 @@ data:
     shopt -s extglob
     rm -rf !(mongodb-ca-cert|mongodb.pem|CAs|openssl.cnf)
     chmod 0600 mongodb-ca-cert mongodb.pem
+    {{- else }}
+    {{- if eq .Values.architecture "standalone" }}
+      ID="0"
+    {{- else }}
+    if [[ "$MY_POD_NAME" =~ "arbiter-"[0-9]{1,} ]]; then
+      ID="${MY_POD_NAME#"{{ printf "%s-arbiter-" $fullname }}"}"
```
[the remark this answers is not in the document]
yohboy: Good catch, changed :)

bitnami/mongodb/values.yaml
```diff
@@ -175,10 +175,30 @@ tls:
   ## @param tls.autoGenerated Generate a custom CA and self-signed certificates
   ##
   autoGenerated: true
-  ## @param tls.existingSecret Existing secret with TLS certificates (keys: `mongodb-ca-cert`, `mongodb-ca-key`, `client-pem`)
-  ## NOTE: When it's set it will disable certificate creation
+  ## @param tls.existingSecret Existing secret with TLS certificates (keys: `mongodb-ca-cert`, `mongodb-ca-key`)
+  ## NOTE: When it's set it will disable secret creation.
   ##
   existingSecret: ""
```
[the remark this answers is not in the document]
yohboy: The `existingSecret` parameter already exists. And yes, you are right, it contains the CA certificate if provided, but this CA is then used to generate a certificate for each MongoDB pod in the init container. So I can not rename it without breaking backward compatibilities. I think we should avoid this modification for now. 
However, you are right with `tls.caCert` and `tls.caKey` ! I changed the `values.yaml` to group with `existingSecret`!

bitnami/mongodb/values.yaml
```diff
@@ -175,10 +175,30 @@ tls:
...
+  standalone:
+    ## @param tls.standalone.existingSecret Existing secret with TLS certificates (`tls.key`, `tls.crt`, `ca.crt`).
```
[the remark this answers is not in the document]
yohboy: To my mind, yes it makes sense! Firstly, it allows us only to share one TLS secret with each component (and it follows the way cert-manager manages certificates). Secondly, the existing parameter for CA is currently used to automatically generate further certificates, not to share CA for each component.

yohboy:
Hi @joancafom, any news about this PR? :)

axelczk:
Hello,

Is there any information about when this PR will be merged ? We need this change on our side too.

bitnami/mongodb/templates/common-scripts-cm.yaml
```diff
@@ -101,4 +103,19 @@ data:
...
+    if [[ "$MY_POD_NAME" =~ "arbiter-0"$ ]]; then
+      ID="0"
+    elif [[ "$MY_POD_NAME" =~ "hidden-"[0-9]{1,}$ ]]; then
+      ID="${MY_POD_NAME#"{{ printf "%s-hidden-" $fullname }}"}"
+    else
+      ID="${MY_POD_NAME#"{{ $fullname }}-"}"
+    fi
+    {{- end }}
+    cat "/certs-${ID}/tls.crt" "/certs-${ID}/tls.key" > "/certs/mongodb.pem"
+    cp "/certs-${ID}/ca.crt" "/certs/mongodb-ca-cert"
```
joancafom, line 119: I think you also need to configure the right permissions for the certificates here, as with the generated ones:

```diff
    cat "/certs-${ID}/tls.crt" "/certs-${ID}/tls.key" > "/certs/mongodb.pem"
    cp "/certs-${ID}/ca.crt" "/certs/mongodb-ca-cert"
+   chmod 0600 /certs/mongodb-ca-cert /certs/mongodb.pem
```

The individual provided files may be mounted with the right permissions, but the generated ones (concatenating using `cat` or copying with `cp`) will more than likely not.

bitnami/mongodb/templates/hidden/statefulset.yaml
```diff
@@ -494,6 +506,14 @@ spec:
             - key: mongodb-ca-key
               path: mongodb-ca-key
               mode: 0600
+        {{- else }}
+        {{- range $index, $secret := .Values.tls.hidden.existingSecrets }}
+        - name: mongodb-certs-{{ $index }}
+          secret:
+            secretName: {{ tpl $secret $ }}
```
joancafom, line 513: Our charts usually make use of our own `common` function for rendering templates, could you please use it here 😁 ?

```suggestion
            secretName: {{ include "common.tplvalues.render" ( dict "value" $secret "context" $) }}
```

bitnami/mongodb/templates/replicaset/statefulset.yaml
```diff
@@ -501,6 +513,14 @@ spec:
             - key: mongodb-ca-key
               path: mongodb-ca-key
               mode: 0600
+        {{- else }}
+        {{- range $index, $secret := .Values.tls.replicaset.existingSecrets }}
+        - name: mongodb-certs-{{ $index }}
+          secret:
+            secretName: {{ tpl $secret $ }}
```
joancafom, line 520: Same comment about the `tpl` function here:

```suggestion
            secretName: {{ include "common.tplvalues.render" ( dict "value" $secret "context" $) }}
```

bitnami/mongodb/templates/arbiter/statefulset.yaml
```diff
@@ -271,6 +281,12 @@ spec:
             - key: mongodb-ca-key
               path: mongodb-ca-key
               mode: 0600
+        {{- else }}
+        - name: mongodb-certs-0
+          secret:
+            secretName: {{ .Values.tls.arbiter.existingSecret }}
```
joancafom, line 287: If we open the possibility to evaluate entries in `.Values.tls.X.existingSecrets` paths as templates, I think we should also do it here. WDYT?

```suggestion
            secretName: {{ include "common.tplvalues.render" ( dict "value" .Values.tls.arbiter.existingSecret "context" $) }}
```

bitnami/mongodb/templates/standalone/dep-sts.yaml
```diff
@@ -434,6 +440,12 @@ spec:
             - key: mongodb-ca-key
               path: mongodb-ca-key
               mode: 0600
+        {{- else }}
+        - name: mongodb-certs-0
+          secret:
+            secretName: {{ .Values.tls.standalone.existingSecret }}
```
joancafom, line 446: Same comment here about templates:

```suggestion
            secretName: {{ include "common.tplvalues.render" ( dict "value" .Values.tls.standalone.existingSecret "context" $) }}
```

joancafom (commented):
Hi again @yohboy and my sincere apologies, the PR got off my radar and I completely forgot to review it. I will keep this pinned for review to prevent it from happening again and try to speed up the remaining review process.

I left some minor comments for you to review.

bitnami/mongodb/templates/_helpers.tpl
```diff
@@ -384,6 +386,32 @@ mongodb: rbac.create
...
+    tls.replicaset.existingSecrets Number of secrets and number of replicaset nodes must be the same.
+{{- end -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
+Validate values of MongoDB&reg; - Number of hidden secrets must be the same than number of hidden nodes.
+*/}}
+{{- define "mongodb.validateValues.hidden.existingSecrets" -}}
+{{- if and .Values.tls.enabled (eq (include "mongodb.hidden.enabled" .) "true") (not (empty .Values.tls.hidden.existingSecrets)) }}
```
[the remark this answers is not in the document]
yohboy: Thanks!

bitnami/mongodb/templates/common-scripts-cm.yaml
yohboy: Agree! Changed

bitnami/mongodb/templates/hidden/statefulset.yaml
yohboy: Sure :)

bitnami/mongodb/templates/arbiter/statefulset.yaml
yohboy: Agree too, changed :)

joancafom (approved):
LGTM! Thanks for your PR!

PR #11270, opened 2022-07-20
Add [CI/CD] prefix to the GH actions name

carrodher:
### Description of the change

Add the `[CI/CD]` prefix to the GH actions name.

### Benefits

In the [Actions view](https://github.com/bitnami/charts/actions) it would be easy to identify what actions are related to a specific purpose.
In the same way, since actions are alphabetically ordered, all the actions with the same prefix will appear together. This is the approach used for the different actions related to support, see
Screenshot 2022-07-20 at 13 18 07

Related to https://github.com/bitnami/charts/pull/11271

### DIFF
--- a/.github/workflows/cd-pipeline.yaml
+++ b/.github/workflows/cd-pipeline.yaml
@@ -1,4 +1,4 @@
-name: 'CD Pipeline'
+name: '[CI/CD] CD Pipeline'
 on: # rebuild any PRs and main branch changes
   push:
     branches:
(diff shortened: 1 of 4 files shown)


.github/workflows/cd-pipeline.yaml
```diff
@@ -1,4 +1,4 @@
-name: 'CD Pipeline'
+name: '[CI/CD] CD Pipeline'
```
recena, line 1: I think the original intention of @FraPazGal was to differentiate CI pipeline vs CD pipelines.

recena (approved):

FraPazGal (approved):
LGTM, but please make sure if there is a rebase needed for the PRs already opened so the newly renamed action runs.

.github/workflows/cd-pipeline.yaml
carrodher: Right now they are appearing together casually because if we add a new workflow starting with `CX` where `X` is any letter between `I` and `D` it is going to be placed in the middle.

I think using this prefix style we can a proper order and the usability is better for Bitnami maintainers and external users.

Being said that, I used `[CI/CD]` but we can use any other string of course

PR #11274, opened 2022-07-20
[bitnami/thanos] Conditionally Set objstore arg and OBJSTORE_CONFIG for Thanos receive

### LINKED ISSUE #11272: [bitnami/thanos]: Receiver still can not start with no object store config
### Name and Version

bitnami/thanos 11.1.2

*    install Thanos helm chart
*    Enable Thanos receiver
*    See the error

```
Error: secret "test-thanos-objstore-secret" not found
```

```
    receive:
      enabled: true
```

thanos receiver can start without the object store config file

### What do you see instead?

Error about missing secret

### Additional information

In #10995 we disabled the validation check for thanos receive to start without object store config, however when we start Thanos receive without `.Values.objstoreConfig` and  `.Values.existingObjstoreSecret`
In the deployment:

* https://github.com/bitnami/charts/blob/master/bitnami/thanos/templates/receive/statefulset.yaml#L155-L159 Needs `thanos.objstoreSecretName`  which is defined here:  https://github.com/bitnami/charts/blob/master/bitnami/thanos/templates/_helpers.tpl#L43-L48

* Now since there is no `Values.existingObjstoreSecret` it will fall to the `else` condition and it will search for `{{- printf "%s-objstore-secret" (include "common.names.fullname" .) -}}` 

* that secret is defined in https://github.com/bitnami/charts/blob/master/bitnami/thanos/templates/objstore-secret.yaml
* The secret will be created only and only if 
```
{{- if (include "thanos.createObjstoreSecret" .) }}
```
* the condition is defined in: 
https://github.com/bitnami/charts/blob/master/bitnami/thanos/templates/_helpers.tpl#L54-L58

* Since there are no  `.Values.objstoreConfig` the condition will not be met and the error will happen.

I will open a PR with a suggested solution

badrabubker:
### Description of the change
This change is to conditionally set objstore arg and OBJSTORE_CONFIG for Thanos receive 
If Thanos receive was deployed without object store config setting the objstore arg and OBJSTORE_CONFIG will lead to an error 

### Benefits

Allow Thanos receive to run without object store config

### Possible drawbacks

I tested this change on Thanos receive only and it's only affecting Thanos receive

### Applicable issues

  - fixes #11272

### Checklist

### DIFF
--- a/bitnami/thanos/Chart.yaml
+++ b/bitnami/thanos/Chart.yaml
@@ -28,4 +28,4 @@ name: thanos
 sources:
   - https://github.com/bitnami/bitnami-docker-thanos
   - https://thanos.io
-version: 11.1.2
+version: 11.1.3
--- a/bitnami/thanos/templates/receive/statefulset.yaml
+++ b/bitnami/thanos/templates/receive/statefulset.yaml
@@ -120,7 +120,9 @@ spec:
             - --grpc-address=0.0.0.0:10901
             - --http-address=0.0.0.0:10902
             - --remote-write.address=0.0.0.0:19291
+            {{- if or .Values.objstoreConfig .Values.existingObjstoreSecret }}
             - --objstore.config=$(OBJSTORE_CONFIG)
+            {{- end }}
             - --tsdb.path=/var/thanos/receive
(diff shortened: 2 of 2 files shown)


alemorcuq (approved):
Thanks

PR #10333, opened 2022-05-20
[bitnami/dataplatform-bp1/2] Deprecate Dataplatform BPs Helm charts

carrodher:
### Description of the change

The Dataplatform BPs Helm charts will be deprecated

### Checklist

- [X] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/).
- [X] Variables are documented in the values.yaml and added to the `README.md` using [readme-generator-for-helm](https://github.com/bitnami-labs/readme-generator-for-helm)
- [X] Title of the pull request follows this pattern [bitnami/<name_of_the_chart>] Descriptive title
- [X] All commits signed off and in agreement of [Developer Certificate of Origin (DCO)](https://github.com/bitnami/charts/blob/master/CONTRIBUTING.md#sign-your-work)

PR #11328, opened 2022-07-25
[bitnami/dataplatform-bp2] Deprecate Helm chart

carrodher:
The Helm chart was supposed to be deprecated in [this PR](https://github.com/bitnami/charts/pull/10333) but it seems the release was not completed since it is not appearing as deprecated in the Helm registry:
```console
$ helm search repo bitnami/dataplatform
NAME                    	CHART VERSION	APP VERSION	DESCRIPTION
bitnami/dataplatform-bp1	12.0.2       	1.0.1      	DEPRECATED This Helm chart can be used for the ...
bitnami/dataplatform-bp2	12.0.3       	1.0.1      	This Helm chart can be used for the automated d...
```

I just created this new PR adding the Helm chart directly with the deprecation changes as well as the VIB action to publish it in the registry.

Once released, I will create a new PR removing the `bitnami/dataplatform-bp2` folder and its associated VIB directory


joancafom (approved):
My main objection to this PR was that you can perform the same VIB call locally to publish the chart, but since you also have to update the index accordingly, maybe this is the right way to do it.

LGTM.

PR #11142, opened 2022-07-12
[bitnami/mariadb] add option for extraVolumeMounts to MariaDB metrics containers

tstollin:
### Description of the change

This change enables adding extra volumeMounts to the metrics containers of the primary and secondary MariaDB StatefulSet. The volumeMounts can be configured via the value.yaml keys `metrics.extraVolumeMounts.primary` and `metrics.extraVolumeMounts.secondary` respectively. 

### Benefits

This change enables mounting volumes with custom password files in the primary and secondary metrics container. 
Without this change the metrics container does not work with custom password files supplied by a volume.

### Possible drawbacks

None.

### Applicable issues

  - none

### Checklist

- [X] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/). This is *not necessary* when the changes only affect README.md files.
- [X] Variables are documented in the values.yaml and added to the `README.md` using [readme-generator-for-helm](https://github.com/bitnami-labs/readme-generator-for-helm)
- [X] Title of the pull request follows this pattern [bitnami/<name_of_the_chart>] Descriptive title
- [X] All commits signed off and in agreement of [Developer Certificate of Origin (DCO)](https://github.com/bitnami/charts/blob/master/CONTRIBUTING.md#sign-your-work)

PR #11348, opened 2022-07-26
[bitnami/mariadb] fix missing volumeMounts key for metrics container

tstollin:
Signed-off-by: Stollin, Thomas <thomas.stollin@interhyp.de>

### Description of the change

This change fixes a templating error caused by adding extraVolumeMounts to the mariadb metrics container when customPasswordFiles are used. This combination of configurations results in a missing ```volumeMounts``` key for the metrics container and dangling list entries. 
The problem was introduced with https://github.com/bitnami/charts/pull/11142. 

### Applicable issues

  - fixes #

### Checklist

### DIFF
--- a/bitnami/mariadb/Chart.yaml
+++ b/bitnami/mariadb/Chart.yaml
@@ -26,4 +26,4 @@ sources:
   - https://github.com/bitnami/containers/tree/main/bitnami/mariadb
   - https://github.com/prometheus/mysqld_exporter
   - https://mariadb.org
-version: 11.1.3
+version: 11.1.4
--- a/bitnami/mariadb/templates/primary/statefulset.yaml
+++ b/bitnami/mariadb/templates/primary/statefulset.yaml
@@ -318,7 +318,6 @@ spec:
           {{- if .Values.metrics.resources }}
           resources: {{- toYaml .Values.metrics.resources | nindent 12 }}
           {{- end }}
-          {{- if or (and .Values.auth.usePasswordFiles (not .Values.auth.customPasswordFiles)) .Values.metrics.extraVolumeMounts.primary }}
           volumeMounts:
           {{- if and .Values.auth.usePasswordFiles (not .Values.auth.customPasswordFiles) }}
             - name: mariadb-credentials
@@ -326,7 +325,6 @@ spec:
           {{- end }}
           {{- if .Values.metrics.extraVolumeMounts.primary }}
           {{- include "common.tplvalues.render" (dict "value" .Values.metrics.extraVolumeMounts.primary "context" $) | nindent 12 }}
-          {{- end }}
           {{- end }}
         {{- end }}
         {{- if .Values.primary.sidecars }}
--- a/bitnami/mariadb/templates/primary/statefulset.yaml
+++ b/bitnami/mariadb/templates/primary/statefulset.yaml
@@ -318,13 +318,15 @@ spec:
           {{- if .Values.metrics.resources }}
           resources: {{- toYaml .Values.metrics.resources | nindent 12 }}
           {{- end }}
-          {{- if and .Values.auth.usePasswordFiles (not .Values.auth.customPasswordFiles) }}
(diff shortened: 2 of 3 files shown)


bitnami/mariadb/templates/primary/statefulset.yaml
```diff
@@ -318,14 +318,16 @@ spec:
...
+          {{- if or (and .Values.auth.usePasswordFiles (not .Values.auth.customPasswordFiles)) .Values.metrics.extraVolumeMounts.primary }}
           volumeMounts:
+          {{- if and .Values.auth.usePasswordFiles (not .Values.auth.customPasswordFiles) }}
             - name: mariadb-credentials
               mountPath: /opt/bitnami/mysqld-exporter/secrets/
           {{- end }}
           {{- if .Values.metrics.extraVolumeMounts.primary }}
           {{- include "common.tplvalues.render" (dict "value" .Values.metrics.extraVolumeMounts.primary "context" $) | nindent 12 }}
           {{- end }}
+          {{- end }}
```
migruiz4, line 330: The extra conditionals are not really needed, could you please refactor this as it follows?

```suggestion
          volumeMounts:
          {{- if and .Values.auth.usePasswordFiles (not .Values.auth.customPasswordFiles) }}
            - name: mariadb-credentials
              mountPath: /opt/bitnami/mysqld-exporter/secrets/
          {{- end }}
          {{- if .Values.metrics.extraVolumeMounts.primary }}
          {{- include "common.tplvalues.render" (dict "value" .Values.metrics.extraVolumeMounts.primary "context" $) | nindent 12 }}
          {{- end }}
```

bitnami/mariadb/Chart.yaml
```diff
@@ -26,4 +26,4 @@ sources:
   - https://github.com/bitnami/bitnami-docker-mariadb
   - https://github.com/prometheus/mysqld_exporter
   - https://mariadb.org
-version: 11.1.1
+version: 11.1.2
```
migruiz4, line 29: Version 11.1.2 has already been released, could you please update this to version `11.1.3`?

migruiz4 (changes requested):
Hi @tstollin,

Thank you very much for your contribution. Could you please take a look at my comments below?

bitnami/mariadb/templates/primary/statefulset.yaml
tstollin: I added the conditionals as I was not sure how Kubernetes would react, if the key `volumeMounts` is rendered, but without any list entries (not even an empty list). This would happen if neither `.Values.auth.usePasswordFiles` nor `.Values.metrics.extraVolumeMounts` are used. But I trust your expertise and I will follow your suggestion.

tstollin:
@migruiz4 I implemented your suggestions. While doing this, I noticed that the same bug existed in the statefulset.yaml of the secondary containers. I fixed this yaml as well. 
Please have a look.

bitnami/mariadb/templates/primary/statefulset.yaml
migruiz4: Hi @tstollin, no worries. 
Kubernetes won't fail if an empty `volumeMounts` is rendered, the same approach is used in other charts, e.g. airflow:
https://github.com/bitnami/charts/blob/21cc5f9ab8127da23f271f085a2e8bd916c2cc0f/bitnami/airflow/templates/worker/statefulset.yaml#L182-L202

tstollin: Perfect. Thanks for the clarification.

tstollin:
The CI/CD Pipeline blocks a merge because of findings by trivy. As far as I understand the logs, there are security findings in the mariadb docker image. As the image is not part of my changes, I'm unsure what I can do to pass the checks. :(
Any help is appreciated!

carrodher:
In order to fix the failing action, could you please rebase from master? Thanks

migruiz4 (approved):
Thank you very much for your contribution @tstollin!

PR #11352, opened 2022-07-26
[bitnami/*] Update URLs to point to the new bitnami/containers monorepo

carrodher:
### Description of the change

Since we are migrating the different `bitnami/bitnami-docker-foo` repositories to a single `bitnami/containers` repository, the aim of this PR is to replace the different URLs reflecting the new source.

### Benefits

Links are updated pointing to the new source.

### Possible drawbacks

Since changes were done in a batch way maybe some links can be broken, in order to avoid this, I used the [`markdown-link-check` tool](https://www.npmjs.com/package/markdown-link-check) to check the markdown ones

⚠️ We are **not** bumping the Chart version on purpose to avoid triggering a new release of every Helm chart in the catalog. On this way changes will be merged at code level but the release will be done when changes from a different nature trigger it

### DIFF
--- a/bitnami/airflow/Chart.yaml
+++ b/bitnami/airflow/Chart.yaml
@@ -30,6 +30,6 @@ maintainers:
     url: https://github.com/bitnami/charts
 name: airflow
 sources:
-  - https://github.com/bitnami/bitnami-docker-airflow
+  - https://github.com/bitnami/containers/tree/main/bitnami/airflow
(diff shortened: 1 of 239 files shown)


fmulero (approved):
Thanks a lot @carrodher and good finding with `markdown-link-check tool`

joancafom (approved):

PR #11371, opened 2022-07-27
[bitnami/Contour] Add default relabel config of servicemonitor

lou-lan:
### Benefits

Because contour's official repo example contains these modifications, it can easily **reuse** contour's **grafana dashboard** without customization.

ref: https://github.com/projectcontour/contour/blob/main/examples/prometheus/02-prometheus-configmap.yaml#L291

### Applicable issues

  - fixes #

### Checklist


carrodher (changes requested):
Thanks for your contribution!

I think the default value should be empty so users can customize it according to their needs instead of hardcoding an example. Although it is good to have specific use cases, other users may be interested in a different example.

In the same way, I have some concerns about future maintenance since this code of code could be updated upstream but not here.

lou-lan:
> Thanks for your contribution!
> 
> I think the default value should be empty so users can customize it according to their needs instead of hardcoding an example. Although it is good to have specific use cases, other users may be interested in a different example.
> 
> In the same way, I have some concerns about future maintenance since this code of code could be updated upstream but not here.


The prometheus configuration created by the `ServiceMonitor` CRD will have additional `relabel_configs`, which may cause some confusion for the SRE. 

Label added by prometheus operator:
```yaml
  - source_labels: [__meta_kubernetes_service_name]
    separator: ;
    regex: (.*)
    target_label: service
    replacement: $1
    action: replace
```

For example the following metrics:
```basic
envoy_cluster_upstream_rq_xx{
  container="envoy", 
  endpoint="metrics", 
  envoy_cluster_name="default_app-golang_80", 
  envoy_response_code_class="2", 
  instance="10.203.0.116:8002", 
  job="contour", 
  namespace="projectcontour",  
  pod="contour-envoy-6c76cc6c94-fzl8r", 
  service="contour-envoy-metrics"
}
```

The actual app service is `app-golang`.

carrodher:
I'm sorry but I don't see the advantage for the whole community of this helm chart about pre-defining a specific configuration for a certain use case when there is a generic way to use the desired configuration without being attached to a specific example

PR #11400, opened 2022-07-28
[bitnami/keycloak] Updating keycloak-config-cli version to match chart requirement

dev-rowbot:
### Description of the change

The `keycloak-config-cli` tag used does not have the 18.0.2 library included, it only has 18.0.1, so the job will fail during deploy because the command is defined as

```
          command:
            - java
            - -jar
            - {{ printf "/opt/bitnami/keycloak-config-cli/keycloak-config-cli-%s.jar" .Chart.AppVersion }}
```

### Benefits

It now works

### Possible drawbacks

It won't work next time the chart version changes. 

I also cannot update the chart version because it will then break the command again

### Checklist


carrodher:
Thanks! The CI job is failing due to the Chart version was not bumped, apart from that, those changes are usually done by our internal test & release pipeline, in fact, I just checked the system and there is an ongoing release updating the container images so an automated commit should appear here soon

carrodher:
There is a new release of the Helm chart bumping the keycloak-config-cli container version, see https://github.com/bitnami/charts/pull/11405

Thanks for your contribution

PR #11410, opened 2022-07-28
[bitnami/elasticsearch] Revisit Goss/Cypress tests

FraPazGal:
### Description of the change

Tested in https://github.com/FraPazGal/charts/pull/68

### Checklist

### DIFF
--- a/.vib/elasticsearch/cypress/cypress/integration/elasticsearch_spec.js
+++ b/.vib/elasticsearch/cypress/cypress/integration/elasticsearch_spec.js
@@ -2,20 +2,6 @@
 import { random } from '../support/utils';
 import body from '../fixtures/documents.json';
 
-const NODE_NUMBER = Cypress.env('nodeNumber');
-
-it('can get cluster info', () => {
-  cy.request({
-    method: 'GET',
-    url: '/',
-    form: true,
-  }).then((response) => {
-    expect(response.status).to.eq(200);
-    expect(response.body.cluster_name).to.contain('elastic');
-    expect(response.body.tagline).to.contain('You Know, for Search');
-  });
-});
-
 it('can check cluster health', () => {
   cy.request({
     method: 'GET',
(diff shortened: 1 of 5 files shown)


.vib/elasticsearch/cypress/cypress/integration/elasticsearch_spec.js
```diff
@@ -24,18 +10,7 @@ it('can check cluster health', () => {
   }).then((response) => {
     expect(response.status).to.eq(200);
     expect(response.body.status).to.contain('green');
```
agomezmoron, line 12: If the response is 200, do we really need to check it's  _green_? I mean, it's possible to have a 200 response and not being _green_? If that is possible, the previous one (eq(200)) is not needed.

agomezmoron (approved):
I left a comment to remove one (if possible) one condition because is redundant and coupled to a color. Apart of that, LGTM.

.vib/elasticsearch/cypress/cypress/integration/elasticsearch_spec.js
FraPazGal: Note that the color indicates the status of the cluster's shards. The color could be [not green](https://www.elastic.co/guide/en/elasticsearch/reference/current/cluster-health.html#cluster-health-api-desc), but I can't say for sure if a `!200` implies a not having the `body.status` param because of insufficient documentation.

By default, we are adding the `200` check to verify the API response looks good overall and then we do specific checks in some of these responses. It is similar to Goss command's `exit-status: 0` and then checking for an specific `stdout` in some cases.

I'm not really against removing the `200` checks but, given that is the equivalent to `exit-status` in APIs, I think it a good way to verify the API request itself is good before checking anything else.

agomezmoron: Then, LGTM

PR #11418, opened 2022-07-29
[bitnami/redis] Fix probe bug, fix config bug, better default timeout, inclusivity

### LINKED ISSUE #11432: [bitnami/redis] Redis config changes are ignored
### Name and Version

bitnami/redis 17.0.6

1. Install the redis chart
2. Set the `replica.configuration` value and update the release
3. New configuration does not take effect

Values for step 1:

```yaml
auth:
  enabled: false
sentinel:
  enabled: true
```

Values for step 2:

```yaml
auth:
  enabled: false
sentinel:
  enabled: true
replica:
  configuration: |
    maxmemory 2gb
    maxmemory-policy noeviction
```

Expected the redis pods to restart and use new configuration.

### What do you see instead?

Redis pods restarted but retained the original configuration, evidenced by both the `config get maxmemory` redis command as well as checking the configuration files in `/opt/bitnami/redis`:

The mounted ConfigMap has the correct config:

```shell
$ cat /opt/bitnami/redis/mounted-etc/replica.conf

dir /data
# User-supplied replica configuration:
maxmemory 2gb
maxmemory-policy noeviction

rename-command FLUSHDB ""
rename-command FLUSHALL ""
# End of replica configuration
```

However the resulting config is not updated:

```shell
$ cat /opt/bitnami/redis/etc/replica.conf

dir /data
# User-supplied replica configuration:
rename-command FLUSHDB ""
rename-command FLUSHALL ""
# End of replica configuration

<snip>
```

### Additional information

The cause appears to be this line of the start-script:

https://github.com/bitnami/charts/blob/5223c9ce5a64ffbc6bffb76e0ca1c1a62d2e5082/bitnami/redis/templates/scripts-configmap.yaml#L175-L177

The previous config is present after pod restart because it is stored in an `emptyDIr` volume:

https://github.com/bitnami/charts/blob/754325f57d391180aa54f224273e264642195e2d/bitnami/redis/templates/sentinel/statefulset.yaml#L601-L608

I assume this is a bug, because care was taken to ensure that the pod restarts when config is updated:

https://github.com/bitnami/charts/blob/5223c9ce5a64ffbc6bffb76e0ca1c1a62d2e5082/bitnami/redis/templates/sentinel/statefulset.yaml#L43

Commit where the bug was introduced: https://git.app.uib.no/caleno/helm-charts/-/commit/34ee421566848cf688dccbc7f9e31af3c19398f1

### LINKED ISSUE #11629: [bitnami/redis] Replica config grows with each restart
### Name and Version

bitnami/redis 17.0.8

1. Install the chart with Sentinel enabled, then trigger a container restart:

   ```shell
   helm install redis bitnami/redis --version 17.0.8 --set auth.enabled=false --set sentinel.enabled=true
   kubectl port-forward redis-node-0 6379
   redis-cli shutdown
   ```

2. Check the `replica.conf` and see that the same data was appended twice:

   ```shell
   kubectl exec redis-node-0 -c redis -- cat /opt/bitnami/redis/etc/replica.conf

   dir /data
   # User-supplied replica configuration:
   rename-command FLUSHDB ""
   rename-command FLUSHALL ""
   # End of replica configuration
   replica-announce-port 6379
   replica-announce-ip redis-node-0.redis-headless.default.svc.cluster.local

   replica-announce-port 6379
   replica-announce-ip redis-node-0.redis-headless.default.svc.cluster.local
   ```

_No response_

_No response_

### What do you see instead?

The `replica.conf` file grows with the same appended data after each container restart.

### Additional information

Issue appears to be caused by these lines:

https://github.com/bitnami/charts/blob/74e05c5c3e31b53508cba7365388b6710ba9c854/bitnami/redis/templates/scripts-configmap.yaml#L183-L185

armsnyder:
### Description of the change

Small bag of changes to the `redis` chart.

Fixes #11432 

Fixes #11629 

### Benefits

* Aligns sentinel defaults with [sentinel configuration documentation](https://redis.io/docs/manual/sentinel/#configuring-sentinel).
* Positions us to having the option to remove "slave" language in the future.
* Fixes a bug where `sentinel.readinessProbe.timeoutSeconds` was not being used.
* Fixes a bug where ConfigMap changes would not result in updated redis config (#11432).

### Possible drawbacks

* Users may be depending on the current default `sentinel.failoverTimeout`.
* In replica mode, will cause a restart of replicas due to changing the `REDIS_REPLICATION_MODE` env value.
* Does anybody know why the ConfigMap updates were being blocked? All I found was [this commit](https://git.app.uib.no/caleno/helm-charts/-/commit/34ee421566848cf688dccbc7f9e31af3c19398f1) which added the `if [[ ! -f ` checks. My best guess is that the goal was to block the ConfigMap live remount from updating a live pod, and that it unintentionally also added this bug.
  * We had some back and forth about this in #11432

### Checklist


bitnami/redis/Chart.yaml
```diff
@@ -24,4 +24,4 @@ maintainers:
 name: redis
 sources:
   - https://github.com/bitnami/containers/tree/main/bitnami/redis
-version: 17.0.8
\ No newline at end of file
+version: 18.0.0
```
migruiz4, line 27: I don't think we should bump a major version because of the change in `sentinel.failoverTimeout`. 
The change is not strictly non-backward compatible and increasing a timeout should not break anything. 

Could you please bump the version `17.0.11`? (latest at this moment is 17.0.10)

migruiz4 (changes requested):
Thank you very much for your contribution! Could you please take a look at my comments?

bitnami/redis/templates/scripts-configmap.yaml
```diff
@@ -172,13 +172,8 @@ data:
...
 
-    if [[ ! -f /opt/bitnami/redis/etc/replica.conf ]];then
-        cp /opt/bitnami/redis/mounted-etc/replica.conf /opt/bitnami/redis/etc/replica.conf
-    fi
-
-    if [[ ! -f /opt/bitnami/redis/etc/redis.conf ]];then
-        cp /opt/bitnami/redis/mounted-etc/redis.conf /opt/bitnami/redis/etc/redis.conf
-    fi
+    cp /opt/bitnami/redis/mounted-etc/replica.conf /opt/bitnami/redis/etc/replica.conf
+    cp /opt/bitnami/redis/mounted-etc/redis.conf /opt/bitnami/redis/etc/redis.conf
```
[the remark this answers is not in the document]
armsnyder: Good catch. I applied this fix in all three places.

bitnami/redis/Chart.yaml
armsnyder: Sounds good. Updated it.

armsnyder:
Hi @migruiz4, can you help review or triage this when you have a chance? Thanks. 🙂

migruiz4 (changes requested):
Hi @armsnyder,

Thank you very much for applying all the suggestions.

I have noticed there is a reference to the `slave` terminology in line 328 of the scripts-configmap.yaml, could you please replace that one too?

armsnyder:
@migruiz4 Sure, I've made the change. As a side note, I do not see where the `REDIS_REPLICATION_MODE` variable is actually referenced in the `start-sentinel.sh` script.

migruiz4 (approved):
Thank you very much for applying the suggested changes, looks good to me!

PR #11425, opened 2022-07-29
[bitnami/dokuwiki] Revisit Goss/Cypress tests

FraPazGal:
### Description of the change

Tested in https://github.com/FraPazGal/charts/pull/69

### Checklist


.vib/dokuwiki/cypress/cypress/integration/dokuwiki_spec.js
```diff
@@ -1,52 +1,18 @@
...
-    );
-    cy.get('#config___dmode').should(
-      'contain.value',
-      setting.newSetting.directoryCreationMode
-    );
-    cy.get('#config___title')
-      .scrollIntoView()
-      .clear()
-      .type(setting.newSetting.newDokuWikiFullName);
+it('allows editing the starting page', () => {
```
agomezmoron, line 4: IMHO, with the login and with this we are testing the same integration: database. I'd maybe complete this test updating an image to test the volume integration too.

agomezmoron (changes requested):

.vib/dokuwiki/cypress/cypress/integration/dokuwiki_spec.js
FraPazGal: As discussed, I have reduced the number of tests to only one (plus login) as this particular chart does not use a DDBB, storing all data in its PVC. That means that as long as we are checking something like login or page edition we are also checking the PVC was mounted without issues.

agomezmoron (approved):
LGTM

PR #11500, opened 2022-08-02
[bitnami/*] Move trivy action to containers and add README note

carrodher:

### Description of the change

In this PR we are moving the Trivy action to containers since the container is the artifact scanned for vulnerabilities and its release process is defined in its own repo where it is being already scanned.

Apart from that, a note in the README was added clarifying how Helm charts (via containers) are analyzed.

It's related to https://github.com/bitnami/containers/pull/1553

### DIFF
--- a/.vib/airflow/vib-publish.json
+++ b/.vib/airflow/vib-publish.json
@@ -32,15 +32,6 @@
       },
       "actions": [
         {
-          "action_id": "trivy",
-          "params": {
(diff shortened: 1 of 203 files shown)


FraPazGal (approved):
LGTM

recena (approved):

pablogalegoc (approved):

PR #11578, opened 2022-08-04
[bitnami/schema-registry] Add Goss/Cypress tests

migruiz4:
### Description of the change

Adds Goss and Cypress tests for the Schema Registry chart.

Tests have been executed in this PR in my fork: https://github.com/migruiz4/charts/pull/1

### Checklist

### DIFF
--- a/.vib/schema-registry/cypress/cypress.json
+++ b/.vib/schema-registry/cypress/cypress.json
@@ -1,0 +1,4 @@
+{
+  "baseUrl": "http://localhost:8081",
+  "responseTimeout": 30000
+}
--- a/.vib/schema-registry/cypress/cypress.json
+++ b/.vib/schema-registry/cypress/cypress.json
@@ -1,4 +1,4 @@
 {
-  "baseUrl": "http://localhost:8081",
+  "baseUrl": "http://localhost",
   "responseTimeout": 30000
 }
--- a/.vib/schema-registry/cypress/cypress.json
+++ b/.vib/schema-registry/cypress/cypress.json
@@ -1,4 +1,7 @@
 {
   "baseUrl": "http://localhost",
-  "responseTimeout": 30000
+  "responseTimeout": 30000,
+  "env": {
+    "compatibilityLevel": "BACKWARD"
+  }
 }
--- a/.vib/schema-registry/cypress/cypress/fixtures/config.json
+++ b/.vib/schema-registry/cypress/cypress/fixtures/config.json
@@ -1,0 +1,3 @@
+{
+  "compatibilityLevel": "BACKWARD"
+}
--- a/.vib/schema-registry/cypress/cypress/fixtures/config.json
+++ b/.vib/schema-registry/cypress/cypress/fixtures/config.json
@@ -1,3 +1,0 @@
-{
-  "compatibilityLevel": "BACKWARD"
-}
--- a/.vib/schema-registry/cypress/cypress/integration/schema_registry_spec.js
+++ b/.vib/schema-registry/cypress/cypress/integration/schema_registry_spec.js
@@ -14,11 +14,12 @@ it('can access the API and obtain global compatibility level', () => {
   });
 });
 
-it('can access the API and create a new schema in the "test" subject', () => {
+it('can create a new schema and subject', () => {
+  const newSubject = 'test';
   cy.request({
     method: 'POST',
     headers: { 'Content-Type': 'application/vnd.schemaregistry.v1+json' },
-    url: 'subjects/test/versions',
+    url: `/subjects/${newSubject}/versions`,
     body: body,
   }).then((response) => {
     expect(response.status).to.eq(200);
@@ -27,11 +28,11 @@ it('can access the API and create a new schema in the "test" subject', () => {
 
   cy.request({
     method: 'GET',
-    url: 'subjects',
+    url: '/subjects',
     form: true,
   }).then((response) => {
     expect(response.status).to.eq(200);
     expect(response.body).to.have.length(1);
-    expect(response.body).to.include('test');
+    expect(response.body).to.include(newSubject);
   });
 });
--- a/.vib/schema-registry/cypress/cypress/integration/schema_registry_spec.js
+++ b/.vib/schema-registry/cypress/cypress/integration/schema_registry_spec.js
@@ -1,5 +1,5 @@
 /// <reference types="cypress" />
-import { body } from '../fixtures/schema.json';
+import body from '../fixtures/schema.json';
 
 it('can access the API and obtain global compatibility level', () => {
   cy.request({
--- a/.vib/schema-registry/cypress/cypress/integration/schema_registry_spec.js
+++ b/.vib/schema-registry/cypress/cypress/integration/schema_registry_spec.js
@@ -1,14 +1,16 @@
 /// <reference types="cypress" />
-import body from '../fixtures/schema.json';
+import { body } from '../fixtures/schema.json';
 
 it('can access the API and obtain global compatibility level', () => {
-  cy.request({
-    method: 'GET',
-    url: '/config',
-    form: true,
-  }).then((response) => {
-    expect(response.status).to.eq(200);
-    expect(response.body.compatibilityLevel).to.contain('BACKWARD');
+  cy.fixture('config').then((config) => {
+    cy.request({
+      method: 'GET',
+      url: '/config',
+      form: true,
+    }).then((response) => {
+      expect(response.status).to.eq(200);
+      expect(response.body.compatibilityLevel).to.contain(config.compatibilityLevel);
+    });
   });
 });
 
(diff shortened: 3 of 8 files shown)


.vib/schema-registry/goss/goss.yaml
```diff
@@ -0,0 +1,8 @@
+file:
+  /opt/bitnami/schema-registry/etc/schema-registry/schema-registry.properties:
+    mode: "0664"
+    filetype: file
+    exists: true
+    contains:
+      - "listeners = http://0.0.0.0:8081"
```
agomezmoron, line 7: This kind of configuration (IP and port) should be in a `vars.yaml` file. Please create it.

agomezmoron (changes requested):
I requested a change for goss parametrization

migruiz4:
@agomezmoron Suggestion applied and execution succeeded at https://github.com/migruiz4/charts/pull/1

.vib/schema-registry/cypress/cypress/integration/schema_registry_spec.js
```diff
@@ -0,0 +1,35 @@
...
+import body from '../fixtures/schema.json';
+
+it('can access the API and obtain global compatibility level', () => {
+  cy.request({
+    method: 'GET',
+    url: '/config',
+    form: true,
+  }).then((response) => {
+    expect(response.status).to.eq(200);
+    expect(response.body.compatibilityLevel).to.contain('BACKWARD');
```
FraPazGal, line 11: I would as `BACKWARD` as a Cypress.env given it is a value configurable in the chart and used in the test. This is a similar scenario to what we do in Goss.

.vib/schema-registry/cypress/cypress/integration/schema_registry_spec.js
```diff
@@ -0,0 +1,35 @@
+/// <reference types="cypress" />
+import body from '../fixtures/schema.json';
```
FraPazGal, line 2: Minor style comment:

```suggestion
import { body } from '../fixtures/schema.json';
```

agomezmoron (approved):
LGTM

.vib/schema-registry/cypress/cypress.json
```diff
@@ -0,0 +1,4 @@
+{
+  "baseUrl": "http://localhost:8081",
```
FraPazGal, line 2: We leave this as `http://localhost` by default, as the value does not really matter here. The VIB Cypress action replaces the value with the URL from the deployment LB. It doesn't really add value to have it with a different port number (even if the chart uses `!80` by default.

.vib/schema-registry/cypress/cypress/integration/schema_registry_spec.js
```diff
@@ -0,0 +1,35 @@
...
+
+it('can access the API and create a new schema in the "test" subject', () => {
+  cy.request({
+    method: 'POST',
+    headers: { 'Content-Type': 'application/vnd.schemaregistry.v1+json' },
+    url: 'subjects/test/versions',
+    body: body,
+  }).then((response) => {
+    expect(response.status).to.eq(200);
+    expect(response.body.id).to.eq(1);
```
FraPazGal, line 23: Is this test retryable? That is, does it succeed two times in a row without issue? I'm concerned about both the API accepting two identical schemas and it returning the same `body.id` for 2 different executions.

FraPazGal (commented):
Thanks for the PR @migruiz4! I left a few comments, could you please check them out?

.vib/schema-registry/cypress/cypress/integration/schema_registry_spec.js
migruiz4: I'm not sure what you mean `is it retryable`.

The test should always return ID 1 unless the schema is changed or the POST request failed, it does not increment with every attempt if that is your concern.

I did test several consecutive executions and, as far as the schema does not change, the call will always return `id: 1`

See:
```
I have no name!@test-schema-registry-0:/$ curl -X POST -H "Content-Type: application/vnd.schemaregistry.v1+json" "http://localhost:8081/subjects/test/versions" -d '{"schema": "{\"type\": \"record\", \"name\": \"User\", \"fields\": [{\"name\": \"name\", \"type\": \"string\"}]}"}'
{"id":1}
I have no name!@test-schema-registry-0:/$ curl -X POST -H "Content-Type: application/vnd.schemaregistry.v1+json" "http://localhost:8081/subjects/test/versions" -d '{"schema": "{\"type\": \"record\", \"name\": \"User\", \"fields\": [{\"name\": \"name\", \"type\": \"string\"}]}"}'
{"id":1}
I have no name!@test-schema-registry-0:/$ curl -X POST -H "Content-Type: application/vnd.schemaregistry.v1+json" "http://localhost:8081/subjects/test/versions" -d '{"schema": "{\"type\": \"record\", \"name\": \"User\", \"fields\": [{\"name\": \"name\", \"type\": \"string\"}]}"}'
{"id":1}
#Changed the schema adding 'refs'
I have no name!@test-schema-registry-0:/$ curl -X POST -H "Content-Type: application/vnd.schemaregistry.v1+json" "http://localhost:8081/subjects/test/versions" -d '{"schema": "{\"type\": \"record\", \"name\": \"User\", \"fields\": [{\"name\": \"name\", \"type\": \"string\", \"refs\": \"this is an example\"}]}"}'
{"id":2}
```

.vib/schema-registry/cypress/cypress/integration/schema_registry_spec.js
migruiz4: Done 👍

.vib/schema-registry/cypress/cypress/integration/schema_registry_spec.js
migruiz4: Done 👍 
At first I thought you were referring to `cy.fixtures`, but then I replaced it with `Cypress.env` as suggested.

.vib/schema-registry/goss/goss.yaml
```diff
@@ -0,0 +1,8 @@
+file:
+  /opt/bitnami/schema-registry/etc/schema-registry/schema-registry.properties:
+    mode: "0664"
+    filetype: file
+    exists: true
+    contains:
+      - "listeners = {{ .Vars.listeners }}"
+      - "avro.compatibility.level = {{ .Vars.avroCompatibilityLevel }}"
```
joancafom, line 8: I have noticed that this chart does also create a serviceAccount and automounts its secret by default. Hence, I think it would be worth adding a test to check this is the case. See:

https://github.com/bitnami/charts/blob/2a88f4661df35140dc39c72e7677357de764bcc3/.vib/kafka/goss/goss.yaml#L20-L23

as an example

joancafom (changes requested):
Thanks for your PR! I left a some comments, could you please review them?

.vib/schema-registry/vib-publish.json
```diff
@@ -16,6 +16,55 @@
...
+          "action_id": "cypress",
+          "params": {
+            "resources": {
+              "path": "/.vib/schema-registry/cypress"
+            },
+            "endpoint": "lb-schema-registry-http",
+            "app_protocol": "HTTP",
+            "env": {
+              "compatibilityLevel": "BACKWARD"
+            }
```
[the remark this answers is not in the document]
migruiz4: In that case, please check the status prior to this commit:
https://github.com/bitnami/charts/pull/11578/commits/c01c002ead84ce8f46121dc7fe30264f4079ed7d

I first I used `cy.fixtures`, but then replaced it with `Cypress.env`. Would you prefer the fixtures approach? What other alternative do you recommend?

.vib/schema-registry/vib-publish.json
```diff
@@ -16,6 +16,55 @@
         }
       ]
     },
+    "verify": {
+      "context": {
+        "resources": {
+          "url": "{SHA_ARCHIVE}",
+          "path": "/bitnami/schema-registry"
+        },
+        "runtime_parameters": "c2VydmljZToKICB0eXBlOiBMb2FkQmFsYW5jZXIKICBwb3J0czoKICAgIGh0dHA6IDgwCg==",
```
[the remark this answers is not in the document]
migruiz4: Understood, thanks!

.vib/schema-registry/cypress/cypress.json
```diff
@@ -0,0 +1,4 @@
+{
+  "baseUrl": "http://localhost",
+  "responseTimeout": 30000
+}
```
joancafom, line 4: The old approach of Cypress env for `compatibilityLevel` was right, but its value should be defined here.

.vib/schema-registry/cypress/cypress/fixtures/config.json
```diff
@@ -0,0 +1,3 @@
+{
+  "compatibilityLevel": "BACKWARD"
+}
```
joancafom, line 3: This is not a fixture

migruiz4:
The suggested change to use `import { body }` instead of `import body` was causing the tests to fail, I think because we are importing a `.json` file and not a `.js` including `export`.

I amended this error and tests are now succeeding in the fork https://github.com/migruiz4/charts/pull/1

joancafom (approved):

PR #11611, opened 2022-08-05
[bitnami/keycloak] Revisit Goss/Cypress tests

FraPazGal:
### Description of the change

Tested in https://github.com/FraPazGal/charts/pull/73

> The VIB Verify executions failed due to the trivy action (now removed from `bitnami/charts` no longer present in `bitnami/charts`

### Checklist

### DIFF
--- a/.vib/keycloak/cypress/cypress/support/commands.js
+++ b/.vib/keycloak/cypress/cypress/support/commands.js
@@ -1,4 +1,4 @@
-const COMMAND_DELAY = 500;
+const COMMAND_DELAY = 2000;
 
 for (const command of ['click']) {
   Cypress.Commands.overwrite(command, (originalFn, ...args) => {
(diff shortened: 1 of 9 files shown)


.vib/keycloak/cypress/cypress/support/commands.js
```diff
@@ -1,4 +1,4 @@
-const COMMAND_DELAY = 1100;
+const COMMAND_DELAY = 500;
```
agomezmoron, line 1: Not blocker for now but we should unify all the COMMAND_DELAY to ensure all tests are aligned. Waiting 500ms or 1300ms (the maximum value I observed in the repo) is nothing after reducing the number of tests. Tuning too much the delays to reduce them at the maximum, sometimes, can give us false positives (due to extra needed delay). IMHO, it's something non blocker now but should be unified in another PR.

agomezmoron (approved):
LGTM but please, review the comment I left for a future PR unifying all (but please, do not block this).

.vib/keycloak/cypress/cypress/support/commands.js
FraPazGal: Good idea, I created https://github.com/bitnami/charts/pull/11617 and updated this PR accordingly.

agomezmoron (approved):
LGTM, thanks for applying the change

joancafom (approved):

PR #11617, opened 2022-08-05
[bitnami/*] Unify Cypress' command delays

FraPazGal:

### Description of the change

This PR unifies the delays used in Cypress for specific types of commands (click/type/etc) to a single duration across the catalog. Besides using the same interval for every asset, it also unifies the different delays (COMMAND_DELAY, TYPE_DELAY, etc.) into a single `COMMAND_DELAY` const. This further reduced the involved logic.

This duration is equal to the largest time interval used by any of our tested assets, which means the majority of our tests now have bigger command delays. Even though this will translate into a (slightly) slower test execution, we are gaining in test stability and readability.

Having said that, given the current effort to reduce the number of strictly necessary tests, the added delayed time will be negligible.

### Benefits

Better readability of tests and reduced possible flakiness due to performance issues. 

### Possible drawbacks

Slower execution of Cypress tests.

### Checklist

### DIFF
--- a/.vib/airflow/cypress/cypress/support/commands.js
+++ b/.vib/airflow/cypress/cypress/support/commands.js
@@ -1,4 +1,4 @@
-const COMMAND_DELAY = 800;
+const COMMAND_DELAY = 2000;
 
 for (const command of ['click']) {
   Cypress.Commands.overwrite(command, (originalFn, ...args) => {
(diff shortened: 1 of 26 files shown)


agomezmoron (approved):
LGTM. Thank you so much for this unification and cleanup.

PR #11632, opened 2022-08-08
[bitnami/elasticsearch] Fix alt names in generated TLS certs

psiroky:
### Description of the change
Minor fix in the generated Elasticsearch TLS certs, specifically one of the alternative DNS names.
Before this change, the name would be generated like this: `DNS:elasticsearch.%!s(MISSING).svc.%!s(MISSING)`.

### Benefits
Those fixed alt names are now usable.

### Possible drawbacks
Can't think of any.

### Applicable issues
I haven't created an issue for this, since it is a very simple fix. If preferred I can also file GitHub issue and link it here.

### Additional information
One additional (super minor) fix in the README.

### Checklist

### DIFF
--- a/bitnami/elasticsearch/README.md
+++ b/bitnami/elasticsearch/README.md
@@ -701,7 +701,7 @@ kibana:
         # Instruct kibana to connect to elastic over https
         enabled: true
         # Bit of a catch 22, as you will need to know the name upfront of your release
-        existingSecret: RELEASENAME-elasticsearch-coordinating-only-crt
+        existingSecret: RELEASENAME-elasticsearch-coordinating-crt # or just 'elasticsearch-coordinating-crt' if the release name happens to be 'elasticsearch'
         # As the certs are auto-generated, they are pemCerts so set to true
         usePemCerts: true
 ```
--- a/bitnami/elasticsearch/Chart.yaml
+++ b/bitnami/elasticsearch/Chart.yaml
@@ -25,4 +25,4 @@ name: elasticsearch
 sources:
   - https://github.com/bitnami/containers/tree/main/bitnami/elasticsearch
   - https://www.elastic.co/products/elasticsearch
-version: 19.1.9
+version: 19.1.10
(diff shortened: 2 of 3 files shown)


bitnami/elasticsearch/README.md
```diff
@@ -700,7 +700,7 @@ kibana:
         # Instruct kibana to connect to elastic over https
         enabled: true
         # Bit of a catch 22, as you will need to know the name upfront of your release
-        existingSecret: RELEASENAME-elasticsearch-coordinating-only-crt
+        existingSecret: RELEASENAME-elasticsearch-coordinating-crt # or just 'elasticsearch-coordinating-crt' if the release name happens to be 'elasticsearch'
```
psiroky, line 703: It seems this depends on which kind of Elastic nodes are actually enabled for specific deployment. If e.g. `coordinating` is disabled, then this secret would not exist and one would need to reference `data` or `master` certs. However, since by default `coordinating` nodes are enabled, I think this is OK.

I could possibly add additional note clarifying the secret naming, but it would clutter the README, so not sure if that's worth it (since this is just an example anyway).

fmulero (approved):
Great Contribution!. Thanks a lot @psiroky.

I am a little bit worried about the certificate change, because it also implies a change in the private key when we are upgrading from the previous version. I've tested it with a simple cluster with data and everything seems OK, I can retrieve the data stored with the previous version. Have you did any tests about this topic?

psiroky:
Yes, this definitely changes the generated private key, but I am not sure I understand what you mean with upgrade.

As far as I was able to observe, the keys/certificates are re-generated on every `helm upgrade` anyway -- even if there are no changes in values.yaml/the chart itself. I suspect this happens, because the certs/keys are always generated differently (which is obviously a good thing), so helm assumes the resource (the secrets) changed and it just replaces them. This feels like a bug, since I would assume the certificates would only be generated once and then re-used if they are already present - however, I am not 100% sure about a proper behavior. Please let me know if I am missing something here, which might be the case.

fmulero:
I was worried about the change in the private key. Some applications use it to encrypt the data, so a change in the private key could cause a data loss, but it seems that is not the case.

Thanks a lot for your contribution, well done

psiroky:
@fmulero. Thanks. I have an slightly unrelated follow-up question -- what is the best place to start a discussion/post a question regarding this chart? I don't want to open Github issue/feature-request since I don't even know if that's correct or if I am just misunderstanding something.

It is basically about the fact that all the certs/keys are re-generated with each `helm upgrade`. I am not sure I would call this a bug or feature request or something else (maybe a known limitation?).

fmulero:
You can open a bug, there is no problem. I think that is a very interesting topic, and likely other charts are affected also

PR #11692, opened 2022-08-09
[bitnami/zookeeper] Fix quorum secret generation

mproch:
There was copy/paste error - the quorum password value was tried to be taken from client-auth secret.

Signed-off-by: Maciej Próchniak <mpr@touk.pl>

### Benefits

The user no longer gets:
` PASSWORDS ERROR: The secret "zookeeper-client-auth" does not contain the key "quorum-learner-password"` 
error after each helm upgrade.

### Checklist


fmulero (approved):
Thanks a lot @mproch. Good catch!

LGTM

PR #11707, opened 2022-08-10
[bitnami/matomo] Revisit tests

joancafom:
### Description of the change

Tested in https://github.com/joancafom/charts/pull/61

### Benefits

### Possible drawbacks

None

### Applicable issues

NA

### Additional information

NA

### Checklist


.vib/matomo/cypress/cypress/integration/matomo_spec.js
```diff
@@ -1,74 +1,62 @@
...
-    cy.get('#user_password').type(`${user.newUser.password}-${random}`);
-    cy.get('#user_email').type(`${random}_${user.newUser.email}`);
-    cy.get('input[value*="Create user"]').click();
-    cy.contains('changes have been saved');
-    cy.visit('/index.php?module=UsersManager');
-    cy.contains('#userLogin', `${user.newUser.username}-${random}`);
-  });
-});
-
 it('allows to create a new website', () => {
```
FraPazGal, line 4: I am missing a lot of context about the asset itself, but does this test verify anything that the following one does not? In the same vein, are any of these tests checking the matomo PVC works as expected?

.vib/matomo/goss/goss.yaml
```diff
@@ -1,36 +1,6 @@
...
-  /opt/bitnami/matomo/js:
-    filetype: symlink
-    linked-to: /bitnami/matomo/js
+  /bitnami/matomo:
     exists: true
-  /opt/bitnami/matomo/plugins:
-    filetype: symlink
-    linked-to: /bitnami/matomo/plugins
-    exists: true
-  /bitnami/matomo/config/config.ini.php:
```
FraPazGal, line 10: Should we maintain this test and modify it to make use of the `vars.yaml` file? 

It is true we would only be checking the host and database parameters (which are implicitly checked by Cypress), but checking them here will verify the values passed to the chart are the ones used. Both the hostname and database can fail due to multiple reasons besides this one. wdyt?

FraPazGal (commented):
Thanks for the PR! I left a couple of comments, please check them when possible.

.vib/matomo/cypress/cypress/integration/matomo_spec.js
joancafom: >  but does this test verify anything that the following one does not?
Yes, the original intention was to test two different components of the app: the website itself and the API.

> are any of these tests checking the matomo PVC works as expected?
About the PVC, they don't perform any changes in the fs (I suppose they are stored in the DB). Nevertheless, based on the folder structure, I don't think there are many operations that modify it.

```
$ ls /bitnami/matomo
config	js  misc  plugins
```

.vib/matomo/goss/goss.yaml
joancafom: My opinion is that we should drop it for this kind of checks (db-related, user accounts, ...). We are either already checking them in Cypress or they do not add any value.

GOSS tests are only executed if and only if the deployment was successful. The majority of the apps do fail when the connection to the DB, so that means that if GOSS tests are executed their DB details must be the correct ones.

FraPazGal: I agree, I don't think this type of test will add enough value to offset the increased test complexity. We may want to document this internally for future reference.

FraPazGal (approved):
Thanks for the changes @joancafom!

PR #11757, opened 2022-08-15
[bitnami/redis] Add seperate serviceAccunts to master and replicas

adrianNEMO:
### Description of the change

This gives the master and replicas separate service accounts.

### Benefits

 This allows running Redis in a service mesh like consul.

### Possible drawbacks

It would be breaking for people who rely on having the same service account for the master and the replicas.

### Checklist


rafariossaa:
Hi,
  Thanks for providing this.
  To not break compatibility, What about using the current `serviceAccountname` for both (master and replicas) when not specific one is set ?.

adrianNEMO:
> Hi, Thanks for providing this. To not break compatibility, What about using the current `serviceAccountname` for both (master and replicas) when not specific one is set ?.

Hi, good Idea I have updated the chart to only use the specific service account when it is manually enabled in the master or replica config.

bitnami/redis/Chart.yaml
```diff
@@ -24,4 +24,4 @@ maintainers:
 name: redis
 sources:
   - https://github.com/bitnami/containers/tree/main/bitnami/redis
-version: 17.0.10
+version: 18.0.0
```
rafariossaa, line 27: In this case, a major version is not longer needed.

rafariossaa (changes requested):

rafariossaa:
Also, you will need to rebase you changes from master.

bitnami/redis/Chart.yaml
adrianNEMO: :heavy_check_mark:

adrianNEMO:
> Also, you will need to rebase you changes from master.

:heavy_check_mark:

rafariossaa (approved):
Thanks for providing this.
LGTM

PR #11802, opened 2022-08-17
[bitnami/kubeapps] Add nodeSelector: kubernetes.io/arch: amd64 in values.yaml

Trackhe:
preventing errors while deploying on mixed clusters

I need to fix the deployments and syncjob by hand. Because i running a mixed cluster amd64 and arm64. The arm64 support from bitnami is planed to 2023/Q3 in the meantime it should have the nodeSelector as default.

### Benefits

Working without hesitate.

### Possible drawbacks

Needs removed or edit, if the arm64 support arrive.

### DIFF
--- a/bitnami/kubeapps/Chart.yaml
+++ b/bitnami/kubeapps/Chart.yaml
@@ -31,4 +31,4 @@ maintainers:
 name: kubeapps
 sources:
   - https://github.com/vmware-tanzu/kubeapps
-version: 10.2.0
+version: 10.2.1
--- a/bitnami/kubeapps/Chart.yaml
+++ b/bitnami/kubeapps/Chart.yaml
@@ -31,4 +31,4 @@ maintainers:
 name: kubeapps
 sources:
   - https://github.com/vmware-tanzu/kubeapps
-version: 10.2.1
+version: 10.2.2
--- a/bitnami/kubeapps/values.yaml
+++ b/bitnami/kubeapps/values.yaml
@@ -402,7 +402,8 @@ frontend:
   ## @param frontend.nodeSelector Node labels for pod assignment
   ## ref: https://kubernetes.io/docs/user-guide/node-selection/
   ##
-  nodeSelector: {}
+  nodeSelector:
+    kubernetes.io/arch: amd64
   ## @param frontend.tolerations Tolerations for pod assignment
   ## ref: https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/
   ##
@@ -750,7 +751,8 @@ dashboard:
   ## @param dashboard.nodeSelector Node labels for pod assignment
   ## ref: https://kubernetes.io/docs/user-guide/node-selection/
   ##
-  nodeSelector: {}
+  nodeSelector:
+    kubernetes.io/arch: amd64
   ## @param dashboard.tolerations Tolerations for pod assignment
   ## ref: https://kubernetes.io/docs/concepts/configuration/taint-and-toleration/
   ##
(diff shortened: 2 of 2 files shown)


antgamdia:
@Trackhe , it seems the PR failed due to a couple of things: 1) the commits need to be signed using DCO (`git commit --signoff`); 2) the metadata annotations seem to be wrong (see the logs: `ERROR: Missing metadata for key: frontend.nodeSelector.kubernetes.io/arch`)

Also @alemorcuq (or any other from the team), I don't know which is the best way to do so: whereas I see the `nodeSelector: {}` in the rest of the bitnami charts, I don't know if defaulting to the `amd64` images could be a good built-in annotation or we better defer to each user the decision on which annotations to use. 
Please, let us know which is the best practice here 🙏 . Thanks!

Trackhe:
> @Trackhe , it seems the PR failed due to a couple of things: 1) the commits need to be signed using DCO (`git commit --signoff`); 2) the metadata annotations seem to be wrong (see the logs: `ERROR: Missing metadata for key: frontend.nodeSelector.kubernetes.io/arch`)
> 
> Also @alemorcuq (or any other from the team), I don't know which is the best way to do so: whereas I see the `nodeSelector: {}` in the rest of the bitnami charts, I don't know if defaulting to the `amd64` images could be a good built-in annotation or we better defer to each user the decision on which annotations to use. Please, let us know which is the best practice here 🙏 . Thanks!

Yea srry ^^' so, i don't have really understood how your system works. So i tried to find the right position for the annotation. I know 2 ways to write them in the deployment.yaml files.

```
nodeSelector:
  kubernetes.io/arch: amd64
```
or
```
nodeAffinity:
    nodeSelectorTerms:
      - matchExpressions:
          - key: kubernetes.io/arch
            operator: In
            values:
              - amd64
```

"I don't know if defaulting to the amd64 images could be a good built-in annotation or we better defer to each user the decision on which annotations to use."

The images running currently only under amd64 on a mixed cluster it creates errors, and i don't find a way to set it localy. As example the Kubeapps dashboard creates the Kubernetes Cron Jobs every time new when i press the registry refresh button. The changes that i made to the CronJob / Sync job are gone after that. It happens automaticly too.

alemorcuq:
Thanks for creating this, @Trackhe. However, I don't like adding this annotation as a default. We try to have as less custom default values as possible to ensure maximum compatibility and ease of maintenance. If your environment requires the annotation, you can set the annotation to your deployment alone.

Trackhe:
@alemorcuq can you tell me how i can do this for the sync jobs because i dont find a way to do this. If it only changing a deployment it would be easy but i don't find a way to edit the deployment of the jobs that are generated when i create a repository in kubeapps. This is the only reason why i created this here.

migruiz4:
Hi @Trackhe, 

I'm sorry I'm not sure I understood your question. 
What @alemorcuq meant is we can not apply this change to the chart default values, but you can deploy the Kubeapps helm chart overriding these values using a file similar to this:

```yaml
# custom-values.yaml
frontend:
  nodeSelector:
    kubernetes.io/arch: amd64

dashboard:
  nodeSelector:
    kubernetes.io/arch: amd64

apprepository:
  nodeSelector:
    kubernetes.io/arch: amd64

kubeops:
  nodeSelector:
    kubernetes.io/arch: amd64

kubeappsapis:
  nodeSelector:
    kubernetes.io/arch: amd64
```

And install the Kubeapps helm chart using those values:
```
helm install <your-release-name> bitnami/kubeapps -f custom-values.yaml
```

PR #11848, opened 2022-08-19
[bitnami/opencart] Revisit tests

joancafom:
### Description of the change

Tested in https://github.com/joancafom/charts/pull/70

### Benefits

### Possible drawbacks

None

### Applicable issues

NA

### Additional information

NA

### Checklist


.vib/opencart/cypress/cypress/integration/opencart_spec.js
```diff
@@ -1,28 +1,11 @@
 /// <reference types="cypress" />
 import { random } from '../support/utils';
 
-it('allows a user to register', () => {
-  cy.visit('/index.php?route=account/login');
-  cy.contains('Continue').click();
-  cy.fixture('users').then((user) => {
```
FraPazGal, line 7: I think you missed deleting the related fixtures' files no longer used. You should be able to remove the `users/downloads.json` and the `images` folder.

.vib/opencart/cypress/cypress/integration/opencart_spec.js
```diff
@@ -66,46 +49,9 @@ it('allows an admin to add a product to the catalog', () => {
...
-    'have.value',
-    Cypress.env('smtpHost')
-  );
-  cy.get('#input-mail-smtp-username').should(
-    'have.value',
-    Cypress.env('smtpUser')
-  );
-});
-
-it('allows the admin to upload a file', () => {
```
FraPazGal, line 92: Would it be possible to have the image addition test included in the one testing a product addition? Usually, the image test is useful to check the asset's PVC is working properly.

FraPazGal (commented):
Hi @joancafom, I left a couple of comments as a review. Thanks for the PR!

.vib/opencart/cypress/cypress/integration/opencart_spec.js
joancafom: You are right, I totally forgot! Removed 🚀

.vib/opencart/cypress/cypress/integration/opencart_spec.js
joancafom: I decided to remove the image upload test because it does not add any value. The app is configured by default to only persist the following files: `config.php` and `admin/config.php`.

https://github.com/bitnami/containers/blob/3694619b3152976363111b80487e93e3fd9214bd/bitnami/opencart/3/debian-11/rootfs/opt/bitnami/scripts/opencart-env.sh#L74

I manually performed the test case and checked that the image is not stored in the PVC, so I don't think this is useful here.

FraPazGal (approved):
LGTM!

PR #11849, opened 2022-08-19
[bitnami/wordpress] Revisit Helm Chart tests

FraPazGal:
### Description of the change

Tested in https://github.com/FraPazGal/charts/pull/80

### Checklist


.vib/wordpress/vib-verify.json
```diff
@@ -22,7 +22,7 @@
...
+        "runtime_parameters": "d29yZHByZXNzVXNlcm5hbWU6IHRlc3RfdXNlcgp3b3JkcHJlc3NQYXNzd29yZDogQ29tcGxpY2F0ZWRQYXNzd29yZDEyMyE0CndvcmRwcmVzc0VtYWlsOiB0ZXN0X3VzZXJfZW1haWxAZW1haWwuY29tCm1hcmlhZGI6CiAgYXV0aDoKICAgIGRhdGFiYXNlOiB0ZXN0X3dvcmRwcmVzc19kYXRhYmFzZQogICAgdXNlcm5hbWU6IHRlc3Rfd29yZHByZXNzX3VzZXJuYW1lCiAgICBwYXNzd29yZDogdGVzdF93b3JkcHJlc3NfcGFzc3dvcmQKcG9kU2VjdXJpdHlDb250ZXh0OgogIGVuYWJsZWQ6IHRydWUKICBmc0dyb3VwOiAxMDAyCmNvbnRhaW5lclNlY3VyaXR5Q29udGV4dDoKICBlbmFibGVkOiB0cnVlCiAgcnVuQXNVc2VyOiAxMDAyCndvcmRwcmVzc1RhYmxlUHJlZml4OiB3b3JkcHJlc3NfCnNlcnZpY2U6CiAgdHlwZTogTG9hZEJhbGFuY2VyCiAgcG9ydHM6CiAgICBodHRwOiA4MAogICAgaHR0cHM6IDQ0Mwpjb250YWluZXJQb3J0czoKICBodHRwOiA4MDgxCiAgaHR0cHM6IDg0NDQKaG9zdEFsaWFzZXM6CiAgLSBpcDogMTI3LjAuMC4xCiAgICBob3N0bmFtZXM6CiAgICAgIC0gc3RhdHVzLmxvY2FsaG9zdA==",
```
agomezmoron, line 25: Why not changing too the LoadBalancer ports?

.vib/wordpress/goss/goss.yaml
```diff
@@ -1,65 +1,32 @@
...
-  wp-post-exists:
-    exec: wp post list
-    exit-status: 0
-    stdout:
-      - "Hello world!"
-    stderr: [ ]
-    timeout: 20000
-  wp-config-list:
-    exec: wp config list
+    exec: wp option list --search="admin_email"
```
joancafom, line 19: You updated the command but forgot to change the title to something that suits.

.vib/wordpress/goss/vars.yaml
```diff
@@ -0,0 +1,12 @@
+wordpressEmail: test_user_email@email.com
+service:
+  ports:
+    https: 443
```
joancafom, line 4: We can use `443 + 1` here as we do with other ports to check customization. I think it'll work 😁

FraPazGal: I left it as `443` because the cypress action is running using `HTTPS` and modifying it breaks the action (as using `http != 80` does if it uses HTTP). I have removed the https test in goss due to redundancy.

joancafom (approved):
LGTM!

Issue #10912, opened 2022-06-28
[bitnami/ghost]  Internal Server Error


jotcodeofficial:
### Name and Version

bitnami/ghost 19.0.6

### What steps will reproduce the bug?

When I visit my site I get: 'Internal Server Error'

The blog pod is alive and fine with no errors. The https cert for the page is working fine too, the issue is getting the service pointing to right thing, which I think is https port issue. When I check the logs of the pod:
```

ghost 05:56:37.35 INFO  ==> ** Ghost setup finished! **

ghost 05:56:37.56 INFO  ==> ** Starting Ghost **

Love open source? We’re hiring Node.js Engineers to work on Ghost full-time.
https://careers.ghost.org/product-engineer-node-js

- Inspecting operating system
```

Here is the relevant part of my values file:

```yaml
service:
  ## @param service.type Ghost service type
  ##
  type: ClusterIP
  ## @param service.ports.http Ghost service HTTP port
  ## @param service.ports.https Ghost service HTTPS port
  ##
  ports:
    http: 80
    https: 443
  ## Node ports to expose
  ## @param service.nodePorts.http Node port for HTTP
  ## @param service.nodePorts.https Node port for HTTPS
  ## NOTE: choose port between <30000-32767>
  ##
  nodePorts:
    http: ""
    https: ""
  ## @param service.clusterIP Ghost service Cluster IP
  ## e.g.:
  ## clusterIP: None
  ##
  clusterIP: ""
```

  here is my ingress route: 
```yaml
  apiVersion: traefik.containo.us/v1alpha1
kind: IngressRoute
metadata:
  namespace: example
  name: example-ingress-route
annotations:
  kubernetes.io/ingress.class: "traefik"
  cert-manager.io/issuer: example-issuer
  traefik.ingress.kubernetes.io/router.entrypoints: web
  traefik.frontend.redirect.entryPoint: https
tls:
  - hosts:
      - example.com
    secretName: example-com-tls
spec:
  entryPoints:
    - web
  routes:
    - match: Host(`example.com`)
      middlewares:
        - name: https-only,remove-www
      kind: Rule
      services:
        - name: example-blog-ghost
          namespace: example
          port: 80
```
If I change the port in here to 80, I just get a 404 page not found when I visit the site
here is my secure ingress route:
```yaml
apiVersion: traefik.containo.us/v1alpha1
kind: IngressRoute
metadata:
  namespace: example
  name: example-secure-ingress-route
annotations:
[...]

Issue #11835, opened 2022-08-18
[GHOST] error on install/upgrade


jalberto:
### Name and Version

bitnami/ghost 19.0.16

### What steps will reproduce the bug?

1. helm upgrade bitnami/ghost
2.  {"errors":[{"message":"Setup has already been completed.","context":null,"type":"NoPermissionError","details":null,"property":null,"help":null,"code":null,"id":"919684b0-1f0d-11ed-b976-fdf77fc63ed5","ghostErrorCode":null}]}

### Are you using any custom parameters or values?

nothing related to PV

### What is the expected behavior?

to work

### What do you see instead?

 {"errors":[{"message":"Setup has already been completed.","context":null,"type":"NoPermissionError","details":null,"property":null,"help":null,"code":null,"id":"919684b0-1f0d-11ed-b976-fdf77fc63ed5","ghostErrorCode":null}]}

jalberto:
I also tried to remove the `depoy` object (but not the mysql stateful one) then `upgrade` again to recreate the `deploy`, and get same error

carrodher:
Please, [note that the Bitnami team is not actively monitoring GitHub Issues for this Helm chart](https://github.com/bitnami/charts/tree/master/bitnami/ghost#community-supported-solution). This issue will stay open for 20 days to allow the community to contribute, after 15 days without activity the issue will be marked as stale, being closed after 5 days.

If you would like to contribute to this Helm chart by fixing a bug or adding a new feature, feel free to create a PR, the Bitnami team will be happy to review it and provide feedback.

jalberto:
Adding this doesn't fix the issue either:

```
volumePermissions:
  enabled: true
```

jalberto:
I fully deleted the volume and did deploy again, everything recreated, same error

jalberto:
when `ghostSkipInstall: true` I get this:

```
ghost 09:59:19.79 
ghost 09:59:19.79 Welcome to the Bitnami ghost container
ghost 09:59:19.80 Subscribe to project updates by watching https://github.com/bitnami/containers
ghost 09:59:19.80 Submit issues and feature requests at https://github.com/bitnami/containers/issues
ghost 09:59:19.80 
ghost 09:59:19.81 INFO  ==> Configuring libnss_wrapper
ghost 09:59:19.87 INFO  ==> Validating settings in MYSQL_CLIENT_* env vars
ghost 09:59:19.97 DEBUG ==> Validating settings in GHOST_* environment variables...
[...]

Issue #10534, opened 2022-06-01
Enabled ghostEnableHttps and traefik config


Mantany:
### Name and Version

bitnami/ghost

### What steps will reproduce the bug?

Steps:

1. using k3s with traefik 
2. set ghostEnableHttps: true
3. visit the domain

Try the same config without https enabled, works but produces "mixed content" errors

- readinessProbe and startupProbe are good
- The logs on the ghost pod looking fine
- when I try to access the site via domain, I get an internal server error (traefik error site)
- when I try to access the site via the IP address of the pod, firefox says "SSL_ERROR_RX_RECORD_TOO_LONG"

I think it has something to do with the missing header information
"x-forwarded-proto: https"

Following the docs for [traefik](https://doc.traefik.io/traefik/v1.7/configuration/backends/kubernetes/) the additional annotations doesn't seem to do the trick.
```
ingress.kubernetes.io/ssl-proxy-headers: "X-Forwarded-Proto: https"
ingress.kubernetes.io/ssl-redirect: "true"
```

Does anybody has the same issues? 

### Are you using any custom parameters or values?
```
ghostHost: "example.com"
ghostPath: /
ghostEnableHttps: true
ingress:
  enabled: true
  pathType: ImplementationSpecific
  apiVersion: ""
  hostname: examle.com
  path: /
  annotations: 
    kubernetes.io/ingress.class: traefik
    cert-manager.io/cluster-issuer: letsencrypt-prod
  tls: true

```
### What is the expected behavior?

I should see the website, when I visit the domain

### What do you see instead?

internal server error

rafariossaa:
Hi,
  Does it work if you access ghost directly ? 
  If so, I think the problem will be in the traefik configuration, in this case I would suggest you ask in traefik forums they are going to provide better support to you.

Mantany:
Hi,

When I try to ping the container directly, it works:
`curl -iv http://<pod-ip-address>:<https-port>/ -H "X-Forwarded-Proto: https"`

I'll ask in traefik forums then, thank you!

rafariossaa:
Hi,
  Don't hesitate to come back and share your findings.

jotcodeofficial:
> Hi,
> 
> When I try to ping the container directly, it works: `curl -iv http://<pod-ip-address>:<https-port>/ -H "X-Forwarded-Proto: https"`
> 
> I'll ask in traefik forums then, thank you!

Did you find a solution for this, I think I have the same issue.

[...]

PR #11852, opened 2022-08-19
[bitnami/ghost]Fix url/https targetPort in traefik

wibed:
Details/Fixes: 
[#10912](https://github.com/bitnami/charts/issues/10912)
https://github.com/bitnami/charts/issues/11835
https://github.com/bitnami/charts/issues/10534

Goal:
targetPort points to https and interprets it as a value instead of a reference. this makes the targetPort unconfigurable besides the default.
additionally, i wont exclude the option for port numbers within the url, but limit it to the service type `LoadBalancer`.

EDIT:

Issues mentioned above are not fixed in regard of https passthrough,

but make a solution using traefik work and build the urls correctly


carrodher:

Apart from that, you also need to bump the chart version

carrodher:

Apart from that, you also need to bump the chart version.

Could you please fill the description about what is the goal of this PR?

wibed:
@carrodher i have admitted requested features.

bitnami/ghost/templates/deployment.yaml
```diff
@@ -154,10 +154,12 @@ spec:
               value: {{ .Values.ghostBlogTitle | quote }}
             - name: GHOST_ENABLE_HTTPS
               value: {{ ternary "yes" "no" .Values.ghostEnableHttps | quote }}
-            - name: GHOST_EXTERNAL_HTTP_PORT_NUMBER
-              value: {{ coalesce .Values.service.ports.http .Values.service.port| quote }}
-            - name: GHOST_EXTERNAL_HTTPS_PORT_NUMBER
-              value: {{ coalesce .Values.service.ports.https .Values.service.httpsPort| quote }}
+            {{- if eq ( toString ( .Values.service.type )) "LoadBalancer" }}
```
fmulero, line 157: Inner parenthesis are not needed
```suggestion
            {{- if eq (toString .Values.service.type) "LoadBalancer" }}
```

bitnami/ghost/templates/deployment.yaml
```diff
@@ -154,10 +154,12 @@ spec:
...
+              - name: GHOST_EXTERNAL_HTTP_PORT_NUMBER
+                value: {{ coalesce .Values.service.ports.http .Values.service.port| quote }}
+              - name: GHOST_EXTERNAL_HTTPS_PORT_NUMBER
+                value: {{ coalesce .Values.service.ports.https .Values.service.httpsPort| quote }}
```
fmulero, line 161: Wrong indentation
```suggestion
            - name: GHOST_EXTERNAL_HTTP_PORT_NUMBER
              value: {{ coalesce .Values.service.ports.http .Values.service.port| quote }}
            - name: GHOST_EXTERNAL_HTTPS_PORT_NUMBER
              value: {{ coalesce .Values.service.ports.https .Values.service.httpsPort| quote }}
```

fmulero (changes requested):
Hi @wibed thanks a lot for your contribution. 

I have some doubts/concerns about this PR and how it tries solve the problem with the redirections. I tried some of the configurations explained in #10912 and I don't see any improvement. Could you share how you test it and how it solves the problem?

Note: I think the problem here is a misconfiguraion in the container or the deployment, because we are enabling HTTPS in the container to do the https redirects but we are not setting any certificate in the container, so we have an inconsistence there

bitnami/ghost/templates/deployment.yaml
wibed: thank you for your concerns!
as far as i can recall, i have not resolved the https passthrough issue. I have indeed had to resort to tls termination, because of this.

this PR adresses the kubernetes  preassumption that https means 443, which would ignore any customization on your end.

fmulero: I see. I think we should try to address the root problem here, the tls termination. We can apply your changes but I don't see any value if the problem persist

wibed: tls termination is not that uncommon for reverse proxies
but it is up to a different issue though, this only solves configuration recognition issues if you try to set a different port from the default.

wibed: @fmulero you can open a new issue were we are able to address the details.
as i am still interested in https passthrough, but i got stuck on the container configuration.

it told the the address it tried to bind to was in use.

wibed:
> Hi @wibed thanks a lot for your contribution.
> 
> I have some doubts/concerns about this PR and how it tries solve the problem with the redirections. I tried some of the configurations explained in #10912 and I don't see any improvement. Could you share how you test it and how it solves the problem?
> 
> Note: I think the problem here is a misconfiguraion in the container or the deployment, because we are enabling HTTPS in the container to do the https redirects but we are not setting any certificate in the container, so we have an inconsistence there

i have openeed a new issue for you
https://github.com/bitnami/charts/issues/12474

fmulero:
I think we can close this PR, there is no real value, the key point here is #12474. Please note this is a community-supported solution.

wibed:
> I think we can close this PR, there is no real value, the key point here is #12474. Please note this is a community-supported solution.

@fmulero 
could you post a working configuration withough those changes?

fmulero:
Can you do it with or without these changes? Theses changes tries to fix: #10912, #11835 and #10534 I tried some of the configurations explained in #10912 and I didn't see any improvement ([see this review](https://github.com/bitnami/charts/pull/11852#pullrequestreview-1105833031), also the code was wrong and I needed to change it, with the suggestions shared, to test it). All my tries ended with TLS failures, and everything points to a problem with TLS termination in the container side as the root cause. Maybe changing the name of the port could help to bypass default behaviour in traefik or istio, but that's only a patch, and probably you can workaround it with specific traefik or istio routing rules.

Please correct me if I am wrong.

wibed:
> Can you do it with or without these changes? Theses changes tries to fix: #10912, #11835 and #10534 I tried some of the configurations explained in #10912 and I didn't see any improvement ([see this review](https://github.com/bitnami/charts/pull/11852#pullrequestreview-1105833031), also the code was wrong and I needed to change it, with the suggestions shared, to test it). All my tries ended with TLS failures, and everything points to a problem with TLS termination in the container side as the root cause. Maybe changing the name of the port could help to bypass default behaviour in traefik or istio, but that's only a patch, and probably you can workaround it with specific traefik or istio routing rules.
> 
> Please correct me if I am wrong.

i couldn't get it to work using tls termination nor using https passthrough,
these changes do not fix the issue with https passthrough, but get it to work with tls termination.

especially keeping the port out of the url. the caption might be misleading ill change that

wibed:
@fmulero 
awaiting review

Mantany:
Thank you for the PR!, the fix works for me in my configuration (k3s using traefik).
Can you merge it please? @fmulero

wibed:
> Hi, I think this PR is not nessesary. I found a solution to tackle the problems with traefik: #10534 #10912
> 
> See: [traefik/traefik#5062](https://github.com/traefik/traefik/pull/5062)
> 
> just add the "serversscheme" annotation to the **service** part in the values.yaml file, like that:
> 
> ```yaml
> service:
>   type: ClusterIP
>   annotations: 
>     traefik.ingress.kubernetes.io/service.serversscheme: http
> ```
> 
> This will override traefiks [automatic https schema detection](https://doc.traefik.io/traefik/routing/providers/kubernetes-ingress/#communication-between-traefik-and-pods).
> 
> It is not very well documented in the traefik docs.
> 
> Also, dont forget to add the right header-middleware (for the x-forwarded-proto header): as described [here](https://ghost.org/docs/faq/proxying-https-infinite-loops/)
> 
> ```yaml
> ingress:
>   enabled: true
>   pathType: ImplementationSpecific
>   apiVersion: ""
>   hostname: example.com
>   path: /
>   annotations: 
>     traefik.ingress.kubernetes.io/router.middlewares: "ghost-blog-add-x-forwarded-proto@kubernetescrd"
> ```
> 
>  ```yaml
> apiVersion: traefik.containo.us/v1alpha1
> kind: Middleware
> metadata:
>   namespace: default
>   name: add-x-forwarded-proto
> spec:
>   headers:
>     customRequestHeaders:
>       X-Forwarded-Proto: https
>       Host: https
>       X-Forwarded-Protocol: https
> ```
> 
> Enjoy!

awesome! closed

PR #12144, opened 2022-08-24
[bitnami/apache] Improve start-stop #11770

### LINKED ISSUE #11770: [bitnami/apache] [Question] Why initialDelaySeconds is needed?
### Name and Version

bitnami/apache

Why the `readinessProbe.initialDelaySeconds` is needed?

https://github.com/bitnami/charts/blob/6cb91fc2558d21754357e5ef2b1dee6fdb93e272/bitnami/apache/values.yaml#L298

The apache chart uses 30 seconds

If the `readinessProbe` is doing a `http` request to `/`, why the `initialDelaySeconds` is needed? If the call is success, why not just to use 0 delay?

Is there a problem in kubernetes services or apache and that's why the chart uses a high initialDelaySeconds? Your [nginx chart](https://github.com/bitnami/charts/blob/b786e6e7e7d2431c61aa6bd2dacef14e29c24768/bitnami/nginx/values.yaml#L317) uses 5 seconds.

set `initialDelaySeconds=0` in `startupProbe`, `livenessProbe` and `readinessProbe`

If the container needs to do a migration or an [initial setup](https://github.com/bitnami/containers/blob/main/bitnami/apache/2.4/debian-11/rootfs/opt/bitnami/scripts/apache/entrypoint.sh#L22) maybe the chart should enable `startupProbe`. So the entry point migrates or do the initial setup and then starts apache? `startupProbe.failureThreshold` could be very high to let the migration/setup ends, but with an initial daily = 0

JoniJnm:
### Description of the change

Set `initialDelaySeconds` to 0 in all probes

Set custom preStop to stop apache gracefully

### Benefits

Pod is ready early

Current processes won't be killed after a scale down (because of the graceful stop)

### Possible drawbacks

Using startupProbe=true and a custom path. This won't work because the startupProbe now runs a custom command (instead a http request)

### Applicable issues

  - fixes #11770

### Additional information

The PR created in bitnami/containers#4234 must be merged

Why the new startupProbe is using a custom command instead of checking if http port is open? Because sometimes the setup needs to restart the service to apply some changes. For example:
* https://github.com/bitnami/containers/blob/main/bitnami/mysql/8.0/debian-11/rootfs/opt/bitnami/scripts/mysql/setup.sh#L38
* https://github.com/bitnami/containers/blob/main/bitnami/rabbitmq/3.10/debian-11/rootfs/opt/bitnami/scripts/rabbitmq/setup.sh#L20

So the probe could be success (port open) but the service is not ready because needs a restart.

I think this change (enable startupProbe and be ensure when setup is done) should be applied in all bitnami charts.

### Checklist

### DIFF
--- a/bitnami/apache/values.yaml
+++ b/bitnami/apache/values.yaml
@@ -52,7 +52,7 @@ extraDeploy: []
 image:
   registry: docker.io
   repository: bitnami/apache
-  tag: 2.4.54-debian-11-r23
+  tag: 2.4.55-debian-11-r23
   digest: ""
   ## Specify a imagePullPolicy
   ## Defaults to 'Always' if image tag is 'latest', else set to 'IfNotPresent'
@@ -224,9 +224,18 @@ command: []
 ## @param args Override default container args (useful when using custom images)
 ##
 args: []
-## @param lifecycleHooks for the Apache server container(s) to automate configuration before or after startup
+## ref: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/
+## ref: https://httpd.apache.org/docs/2.4/en/stopping.html#gracefulstop
+## @param lifecycleHooks.preStop.exec.command Override default apache preStop command
 ##
-lifecycleHooks: {}
+lifecycleHooks:
+  preStop:
+    exec:
+      command:
+        - kill
+        - -s
+        - SIGWINCH
+        - "1"
 ## Apache pods' resource requests and limits
 ## ref: https://kubernetes.io/docs/user-guide/compute-resources/
 ## We usually recommend not to specify default resources and to leave this as a conscious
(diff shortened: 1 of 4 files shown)


bitnami/apache/values.yaml
```diff
@@ -224,9 +224,18 @@ command: []
...
+## @param lifecycleHooks.preStop.exec.command Override default apache preStop command
+##
+lifecycleHooks:
+  preStop:
+    exec:
+      command:
+        - kill
+        - -s
+        - SIGWINCH
+        - "1"
```
JoniJnm, line 238: The process 1 is httpd, but it can me changed by

`$(cat /opt/bitnami/apache/var/run/httpd.pid)`

bitnami/apache/values.yaml
```diff
@@ -52,7 +52,7 @@ extraDeploy: []
 image:
   registry: docker.io
   repository: bitnami/apache
-  tag: 2.4.54-debian-11-r23
+  tag: 2.4.55-debian-11-r23
```
JoniJnm, line 55: bitnami/containers#4234 mudt be merged

fmulero:
Please bear with me. I hope take it a glance soon.

PR #12159, opened 2022-08-25
[bitnami/drupal] Add Cypress and Goss tests to Drupal Chart

mdhont:

### Description of the change

This adds Cypress and Goss tests for the Drupal Chart for the VIB chart testing pipeline

### Additional information

Successful test run: https://github.com/mdhont/charts/runs/8014668068

### Checklist


.vib/drupal/cypress/cypress/integration/drupal_spec.js
```diff
@@ -0,0 +1,18 @@
...
+it('allows creating a new user with a profile picture', () => {
+  cy.login();
+  cy.visit('/admin/people/create');
+  cy.fixture('users').then((user) => {
+    cy.get('#edit-mail').type(`${random}.${user.newUser.email}`);
+    cy.get('#edit-name').type(`${user.newUser.name}${random}`);
+    cy.get('#edit-pass-pass1').type(`${random}${user.newUser.password}`);
+    cy.get('#edit-pass-pass2').type(`${random}${user.newUser.password}`);
+    cy.get('[type=file]').selectFile('cypress/fixtures/images/test_image.jpeg', { force: true });
+    cy.contains('test_image');
```
FraPazGal, line 13: I'll prefer this check to be done after the account is created. Even if the string appears here, there could be issues in the creation process. Is that possible?

.vib/drupal/goss/vars.yaml
```diff
@@ -0,0 +1,16 @@
+drupalUsername: test_user
+drupalPassword: ComplicatedPassword123!4
```
FraPazGal, line 2: You should not just paste the used `runtime_params` here, as there are parameters you won't be checking in goss. Please remove those unrelated to your tests.

Additionally, `bitnami/drupal` automounts a service account, please add the related tests as done for other assets:

https://github.com/bitnami/charts/blob/c6aa735f9d8e6352d951bdfc92befacff69ebe0f/.vib/cassandra/goss/goss.yaml#L14

.vib/drupal/vib-publish.json
```diff
@@ -22,7 +22,7 @@
           "url": "{SHA_ARCHIVE}",
           "path": "/bitnami/drupal"
         },
-        "runtime_parameters": "ImRydXBhbFBhc3N3b3JkIjogIlMzOUJLV2pTa2gi",
+        "runtime_parameters": "ZHJ1cGFsVXNlcm5hbWU6IHRlc3RfdXNlcgpkcnVwYWxQYXNzd29yZDogQ29tcGxpY2F0ZWRQYXNzd29yZDEyMyE0CmNvbnRhaW5lclBvcnRzOgogIGh0dHA6IDgwODEKICBodHRwczogODQ0NApzZXJ2aWNlOgogIHBvcnRzOgogICAgaHR0cHM6IDQ0NApwb2RTZWN1cml0eUNvbnRleHQ6CiAgZnNHcm91cDogMTAwMgpjb250YWluZXJTZWN1cml0eUNvbnRleHQ6CiAgcnVuQXNVc2VyOiAxMDAyCmhvc3RBbGlhc2VzOgogIC0gaXA6IDEyNy4wLjAuMQogICAgaG9zdG5hbWVzOgogICAgICAtIHN0YXR1cy50ZXN0aG9zdA==",
```
FraPazGal, line 25: Similar to what WordPress has, we should modify MariaDB's subchart params under `mariadb.auth` to avoid using default values.

.vib/drupal/cypress/cypress/integration/drupal_spec.js
```diff
@@ -0,0 +1,18 @@
...
+    cy.contains('Create new account').click();
```
FraPazGal, line 14: As a suggestion, try always to use the minimum info necessary to avoid flakiness. For example, we should search for `Create` instead of the whole string if possible.

.vib/drupal/cypress/cypress/support/commands.js
```diff
@@ -0,0 +1,29 @@
...
+    cy.get('#edit-pass').type(password);
+    cy.contains('input', 'Log in').click();
+  }
+);
+
+Cypress.on('uncaught:exception', (err, runnable) => {
+  if (err.message.includes('Unexpected token')) {
+    return false;
+  }
+});
```
FraPazGal, line 29: Is this exception handling necessary? It seems strange to me that drupal outputs an `Unexpected token` error.

FraPazGal (changes requested):
Hi @mdhont, thanks for the PR. I left some comments as a review, could you please check them?

.vib/drupal/cypress/cypress/integration/drupal_spec.js
```diff
@@ -0,0 +1,18 @@
+/// <reference types="cypress" />
+import { random } from '../support/utils';
+
+it('allows creating a new user with a profile picture', () => {
```
joancafom, line 4: The test itself is okay, but when we only implement one test-case in Cypress we try to cover the key feature of the asset. This time, I feel that creating a post (just like WordPress or Discourse) is more the core of Drupal than adding a user.

Having said that I would either: 
- Rewrite this to test a a post publishing with an image (recommended)
- Keep this test, but add another one for the aforementioned (usually, if can't test all the features in only one test or the asset has more than one key functionality)

joancafom (changes requested):
Thanks for your PR! I left you some comments, please review them when you have a chance 😁

mdhont:
Successful test run: https://github.com/mdhont/charts/runs/8072865528

.vib/drupal/cypress/cypress/support/commands.js
mdhont: I've removed this now.

.vib/drupal/cypress/cypress/integration/drupal_spec.js
mdhont: Changed it

.vib/drupal/vib-publish.json
mdhont: Added this.

.vib/drupal/cypress/cypress/support/commands.js
```diff
@@ -0,0 +1,32 @@
...
+
+    return new Promise((resolve) => {
+      setTimeout(() => {
+        resolve(origVal);
+      }, COMMAND_DELAY);
+    });
+  });
+}
+
+Cypress.Commands.add('iframe', { prevSubject: 'element' }, $iframe => {
```
mdhont, line 15: I added this function to handle the iframe used by the application for adding the article text. Cypress is not very straightforward with how it handles Iframes it seems.

mdhont:
Test run with automountServiceAccountToken check:

https://github.com/mdhont/charts/runs/8073520126

.vib/drupal/cypress/cypress/integration/drupal_spec.js
```diff
@@ -0,0 +1,18 @@
+/// <reference types="cypress" />
+import { random } from '../support/utils';
+
+it('allows creating an article with an image', () => {
+  cy.login();
+  cy.visit('/node/add/article');
+  cy.fixture('articles').then((article) => {
+    cy.get('.text-full[name*=title]').type(`${article.newArticle.title} ${random}`);
+    cy.get('iframe.cke_wysiwyg_frame').iframe().find('p').type(`${article.newArticle.text}`);
```
FraPazGal, line 9: I'm concerned about the `get` selectors as they seem too specific, especially the second one. Sometimes there is now good option but we should always try to look for selectors that seem "stable" and that shouldn't be prone to change from version to version. Just as a suggestion, another option for the iframe selector would be:

```suggestion
    cy.get('.text-full[name*=title]').type(`${article.newArticle.title} ${random}`);
    cy.get('[title*="Body"]').iframe().find('p').type(`${article.newArticle.text}`);
```

.vib/drupal/cypress/cypress/integration/drupal_spec.js
```diff
@@ -0,0 +1,18 @@
...
+    cy.get('[type=file]').scrollIntoView();
+    cy.get('[type=file]').selectFile('cypress/fixtures/images/test_image.jpeg', { force: true });
+    cy.get('img').should('have.attr', 'src').and('contains', 'test_image');
```
FraPazGal, line 12:
```suggestion
    cy.get('[src*="test_image"]');
```

.vib/drupal/cypress/cypress/integration/drupal_spec.js
mdhont: I've rewritten the test to publish an article instead.

.vib/drupal/cypress/cypress/integration/drupal_spec.js
```diff
@@ -0,0 +1,18 @@
...
+    cy.get('input[id*="alt"]').type(`${article.newArticle.altText}`);
```
FraPazGal, line 13: Checking the app, this field only appears when the image is fully uploaded so we can avoid checking for `test_image` above as this one does basically the same.

.vib/drupal/cypress/cypress/support/commands.js
FraPazGal: I'm not sure if there is a need to add the whole iframe logic to just add the article's text (body), which is not a mandatory field. Having to fill the title and uploading an image would be testing enough so if typing the body requires extra logic I think it is not worthwhile. WDYT @mdhont @joancafom?

.vib/drupal/vib-publish.json
```diff
@@ -22,7 +22,7 @@
           "url": "{SHA_ARCHIVE}",
           "path": "/bitnami/drupal"
         },
-        "runtime_parameters": "ImRydXBhbFBhc3N3b3JkIjogIlMzOUJLV2pTa2gi",
+        "runtime_parameters": "ZHJ1cGFsVXNlcm5hbWU6IHRlc3RfdXNlcgpkcnVwYWxQYXNzd29yZDogQ29tcGxpY2F0ZWRQYXNzd29yZDEyMyE0CmRydXBhbEVtYWlsOiB1c2VyQGV4YW1wbGUuY29tCm1hcmlhZGI6CiAgYXV0aDoKICAgIGRhdGFiYXNlOiB0ZXN0X2RydXBhbF9kYXRhYmFzZQogICAgdXNlcm5hbWU6IHRlc3RfZHJ1cGFsX3VzZXJuYW1lCiAgICBwYXNzd29yZDogdGVzdF9kcnVwYWxfcGFzc3dvcmQKY29udGFpbmVyUG9ydHM6CiAgaHR0cDogODA4MQogIGh0dHBzOiA4NDQ0CnNlcnZpY2U6CiAgdHlwZTogTG9hZEJhbGFuY2VyCiAgcG9ydHM6CiAgICBodHRwczogNDQ0CnBvZFNlY3VyaXR5Q29udGV4dDoKICBlbmFibGVkOiB0cnVlCiAgZnNHcm91cDogMTAwMgpjb250YWluZXJTZWN1cml0eUNvbnRleHQ6CiAgZW5hYmxlZDogdHJ1ZQogIHJ1bkFzVXNlcjogMTAwMgpob3N0QWxpYXNlczoKICAtIGlwOiAxMjcuMC4wLjEKICAgIGhvc3RuYW1lczoKICAgICAgLSBzdGF0dXMudGVzdGhvc3QKc2VydmljZUFjY291bnQ6CiAgYXV0b21vdW50U2VydmljZUFjY291bnRUb2tlbjogdHJ1ZQ==",
```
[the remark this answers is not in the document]
mdhont: @FraPazGal I was applying @joancafom's comments on this issue: https://github.com/bitnami/charts/pull/12159#discussion_r957215184.

.vib/drupal/cypress/cypress/support/commands.js
mdhont: My idea for adding it was that if we only test the main functionality then let's make sure at least that the common steps are tested.

.vib/drupal/cypress/cypress/integration/drupal_spec.js
mdhont: I think the test fails without this check, but I will double check.

mdhont: You were right, I removed it

.vib/drupal/cypress/cypress/integration/drupal_spec.js
mdhont: This has been changed now

.vib/drupal/cypress/cypress/integration/drupal_spec.js
```diff
@@ -0,0 +1,18 @@
...
+import { random } from '../support/utils';
+
+it('allows creating an article with an image', () => {
+  cy.login();
+  cy.visit('/node/add/article');
+  cy.fixture('articles').then((article) => {
+    cy.get('.text-full[name*=title]').type(`${article.newArticle.title} ${random}`);
+    cy.get('iframe.cke_wysiwyg_frame').iframe().find('p').type(`${article.newArticle.text}`);
+    cy.get('[type=file]').scrollIntoView();
+    cy.get('[type=file]').selectFile('cypress/fixtures/images/test_image.jpeg', { force: true });
```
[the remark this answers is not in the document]
mdhont: I've removed scrollIntoView

.vib/drupal/vib-publish.json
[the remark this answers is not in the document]
mdhont: This has been changed now

agomezmoron (approved):
LGTM with @FraPazGal's suggestion

.vib/drupal/goss/goss.yaml
```diff
@@ -0,0 +1,26 @@
...
+    status: 200
+    allow-insecure: true
+command:
+  check-user-info:
+    exec: id
+    exit-status: 0
+    stdout:
+      - uid={{ .Vars.containerSecurityContext.runAsUser }}
+      - /groups=.*{{ .Vars.podSecurityContext.fsGroup }}/
+file:
```
joancafom, line 12: As this chart mounts a volume, you should also check it:

```
  /bitnami/drupal:
    exists: true
    filetype: directory
    mode: '2775'
    owner: root
```

.vib/drupal/goss/goss.yaml
```diff
@@ -0,0 +1,26 @@
...
+    filetype: directory
+    mode: "3777"
+  /etc/hosts:
+    exists: true
+    filetype: file
+    contains:
+    {{ range .Vars.hostAliases }}
+    {{ $ip := .ip }}
+    {{ range $host := .hostnames }}
+      - /^{{ $ip }}.*{{ $host }}$/
```
joancafom, line 24:
```suggestion
      - /{{ $ip }}.*{{ $host }}/
```

agomezmoron (approved):
LGTM

.vib/drupal/cypress/cypress/integration/drupal_spec.js
```diff
@@ -0,0 +1,15 @@
+/// <reference types="cypress" />
+import { random } from '../support/utils';
+
+it('allows creating an article with an image', () => {
+  cy.login();
+  cy.visit('/node/add/article');
+  cy.fixture('articles').then((article) => {
+    cy.get('.text-full[name*=title]').type(`${article.newArticle.title} ${random}`);
+    cy.get('[type=file]').get('[type=file]').selectFile('cypress/fixtures/images/test_image.jpeg', { force: true });
```
FraPazGal, line 9: This double `get` is really weird. I'm not sure if the intention was to have something similar to [cy.within](https://docs.cypress.io/api/commands/within) but here a single `get` should suffice. Checking how Cypress behaves, it seems it checks 2 times for that same selector, which is redundant. Please try using the following:

```suggestion
    cy.get('[type=file]').selectFile('cypress/fixtures/images/test_image.jpeg', { force: true });
```

FraPazGal (commented):
Hi @mdhont, I added one last comment. Please check it when available.

.vib/drupal/cypress/cypress/integration/drupal_spec.js
mdhont: I didn't notice. Anyway. It's fixed now: 

https://github.com/mdhont/charts/actions/runs/3051810313

FraPazGal (approved):
LGTM!

joancafom (approved):

Issue #11828, opened 2022-08-18
bitnami-common 1.17.0 breaks all image tags


pyromaniac3010:
### Name and Version

bitnami/common 1.17.0

### What steps will reproduce the bug?

all charts which have:
```dependencies:
  - name: common
    repository: https://charts.bitnami.com/bitnami
    tags:
      - bitnami-common
    version: 1.x.x
```
are reporting image tag version "nil" back and break.
this was introduced with https://github.com/bitnami/charts/pull/11797
This also leads to big problems with a huge amount of other bitnami charts depending on common.
bitnami/redis is affected by this, too!

### What do you see instead?

no change for the image tag

### Additional information

https://github.com/bitnami/charts/pull/11797#issuecomment-1219193803
https://github.com/bitnami/charts/pull/11797

pyromaniac3010:
@carrodher please revert the commit and release a 1.17.1 asap as all users with ci/cd pipelines like argo-cd aso have for example stopped redis instances at the moment and cannot really work around that. many charts use the recommended dependency with "version: 1.x.x"!

Panplumousse:
Hello,
same issu here with bitnami/kube-prometheus

pyromaniac3010:
how long does it take to update https://charts.bitnami.com/bitnami/index.yaml ?

carrodher:
Sorry for the inconveneces. Initial changes from #11797 (`1.17.0`) were reverted at #11829 (`1.17.1`). The same changes were implemented again at #11830 but bumping the chart version in a major (`2.0.0`).

The version with the fix is already published, see
```console
$ helm repo update bitnami
Hang tight while we grab the latest from your chart repositories...
...Successfully got an update from the "bitnami" chart repository
Update Complete. ⎈Happy Helming!⎈

$ helm search repo bitnami/common
NAME          	CHART VERSION	APP VERSION	DESCRIPTION
bitnami/common	1.17.1       	1.17.1     	A Library Helm Chart for grouping common logic ...
```

And the new major will be available once reviewed and merged the PR

PR #12175, opened 2022-08-27
[bitnami/fluentd] feat: allow setting `hostNetwork` and `dnsPolicy` for forwarder

### LINKED ISSUE #12156: [bitnami/fluentd] allow option to set `hostNetwork` and `dnsPolicy` for forwarder deployment
### Name and Version

bitnami/fluend 5.3.4

https://aws.amazon.com/blogs/containers/capturing-logs-at-scale-with-fluent-bit-and-amazon-eks/ from this article, I want to allow using the `Use_Kubelet` config for fluent-bit to avoid overloading the kube API server. However, that requires setting the `hostNetwork` and `dnsPolicy` (see https://docs.fluentbit.io/manual/pipeline/filters/kubernetes/).

I am aware that the chart is mainly for fluentd. However, because of the way it was written, it can be easily reused for fluent-bit (was able to successfully deploy a fluent-bit forwarder with fluentd aggregator setup).

I am happy to make and test the changes myself if this is approved.

Allow setting the `hostNetwork` and `dnsPolicy` for custom forwarders

n/a

PJHRobles:
Signed-off-by: Paul Joshua Robles <joshuahrobles@gmail.com>

### Description of the change
This allows setting `hostNetwork` and `dnsPolicy` for forwarder daemonset. The change is backwards-compatible.

### Benefits

This will allow custom forwarders such as fluent-bit to set the mentioned values. The values are necessary in order to query the kubelet which should reduce the load on the kube API server (see [article](https://aws.amazon.com/blogs/containers/capturing-logs-at-scale-with-fluent-bit-and-amazon-eks/) for more info).

### Possible drawbacks

n/a

### Applicable issues

Closes #12156

### Additional information

Manually tested with the following scenarios:
- [X] default values for `hostNetwork` and `dnsPolicy` (backwards compatibility working)
- [X] default values for `hostNetwork` and `dnsPolicy` with `forwarder.rbac.pspEnabled`
- [X] update values for `hostNetwork` (set to true) and `dnsPolicy` (set to `ClusterFirstWithHostNet`)
- [X] update values for `hostNetwork` (set to true) and `dnsPolicy` (set to `ClusterFirstWithHostNet`) with `forwarder.rbac.pspEnabled`

### Checklist


javsalgar (approved):
LGTM! Thanks!

PJHRobles:
#12213 if I understood the problem correct, #12214 should fix the issue

PJHRobles:
From #12214:
> As I already commented in #12213, this does not seem to be related to the tests but to an issue that affected the charts some time ago (see #11828).
> 
> The values used in `vib-verify` change the port the **Service** exposes, but not the Fluentd port itself. Moreover, the `targetPort` defaults to the port named`http`, which is indeed set to `9880`

Any chance we can merge this? Thanks!

javsalgar:
I've click on retry the jobs, because if it does not succeed, the publish process will fail in another action

javsalgar:
The tests are failing because of one goss test

```
command:
  # The Chart is configured to keep track of the logs in the /var/log/containers directory. Creating a log file
  # should trigger the Forwarder to send the contents to the Aggregator.
  #
  # The integration can be verified as the Forwarder pods have access to the logs of every pod in the cluster, so
  # we check that the Aggregator is aware of the newly created file.
  {{ with (printf "test-%s.log" (randAlpha 5)) }}
  forward-aggregator-integration:
    exec: "cd /var/log/containers && echo '{\"time\": \"2022-07-29T11:33:47.408955753Z\", \"data\": \"zwiebel\"}' > {{.}} && sleep 65 && find . -name \"*fluentd-0*\" -exec tail {} \\;"
    exit-status: 0
    stdout:
      - {{.}}
    timeout: 75000
  {{end}}
```

@joancafom  is looking into it

joancafom:
Hi! After some time debugging this I have come to what I think is the root issue.

The tests are designed to verify the integration between _forwarders_ and _aggregators_. In order to do that, it executes the following steps in a _forwarder_ pod:

1.  Create a new log file in a path that *forwarders* are configured to track.
2. The _forwarder_ notices the new file and sends info about it to the _aggregator_.
3. If the _aggregator_ has correctly received the information, a new entry appears in its log. This log is checked for the new entry.

The _forwarder_-_aggregator_ integration can be easily verified as _forwarders_ have access to the logs of the rest of the pods executing *within the same node*.

#### Problem
Nevertheless, this might be problematic for clusters with more than one node, as we should *carefully select the forwarder pod that shares the same aggregator node to have access to its logs*. If the test is executed in a _forwarder_ pod that is not running in the same node, it will report a *failure* as the new entry can not be found.

We have not faced this kind of error in the past with this test, maybe due to a combination of both luck and timing.

#### Solution
I will adapt the verify stage to enforce the provisioning of a cluster with a number of `nodes=1`, ensuring then that _forwarders_ and _aggregators_ pods are scheduled on the same node and then solving the possible issue.

joancafom (approved):

PR #12260, opened 2022-09-02
[bitnami/elasticsearch] Configurable servicenames

### LINKED ISSUE #11505: [bitnami/elasticsearch] configurable service names
bitnami/elasticsearch >= 18

### When upgrading from chart version 17.x to 18.x the upgrade fails with the following errors:

```
Forbidden: updates to statefulset spec for fields other than 'replicas', 'template', and 'updateStrategy' are forbidden ...
```

This is due to the fact that the ```statefulset.spec.serviceName``` changed from: ```{elasticsearch.[...].fullname}``` to ```{elasticsearch.[...].fullname}-hl```

A rolling cluster upgrade is therefore not possible. 
We need to destroy the current release first which is what we don't want.

Before writing code I would like to discuss our idea. Maybe someone has better ideas. 

Introduce a new variable in ```values.yaml``` for each node type section (master, data, coordinating, ingest) 

```
master: 
  # The name of the headless service
  # default to: {elasticsearch.master.fullname}-hl
  serviceName: ""

```

Provide a new ```_helper.tpl``` function for each node type which defines the defaults.
```
{{- define "elasticsearch.master.servicename" -}}
{{- if .Values.master.servicename-}}
{{- .Values.master.servicename | trunc 63 | trimSuffix "-" -}}
{{- else -}}
{{- printf "%s-hl" (include "elasticsearch.master.fullname" .) | trunc 63 | trimSuffix "-" -}}
{{- end -}}
{{- end -}}
```

What do you think about this? I know that no other bitnami chart currently uses configurable service names but in our case it would make sense.

mFranz82:
### Description of the change

Since chart version 18.0.0 the naming of the node services changed **from**: 
```{{ include "elasticsearch.coordinating.fullname" . }}```  
**to:** 
```{{ printf "%s-hl" (include "elasticsearch.coordinating.fullname" .) | trunc 63 | trimSuffix "-" }}```
If one has a cluster running with the latest 17.x.x chart he will not be able to perform a rolling cluster upgrade to chart version >=18.0.0. This is because of kubernetes does not allow to change the servicename in a statefulset once set. Adressed in this issue #11505

This change intruduces the value ```servicenameOverride``` for each node section (**coordinating**, **data**, **master**, **ingest**) which could be used to override the *-hl* suffix.

### Benefits
You can run a rolling cluster upgrade  to elasticserach version >= 8 without any problems now.

### Possible drawbacks
No drawbacks as the defaults are the same

### Applicable issues

  - fixes #11505

### Additional information


javsalgar:
Hi!

Thank you for the PR! Could you fix the merge issue?

mFranz82:
@javsalgar Something went wrong during rebasing and the merge did not had any effects...
Could you please re-review. 
Thanks a lot!

javsalgar:
It seems that there are issues with the DCO, could you check them?

PR #12268, opened 2022-09-03
[bitnami/keycloak] Option to scrape extra monitoring endpoints and configure alerts

### LINKED ISSUE #12250: [bitnami/keycloak] Configure scraping of realm metrics and add rules and a dashboard
### Name and Version

bitnami/keycloak 9.7.2

Keycloak exposes realm metrics on `/realms/master/metrics` (actually they are available at `/realms/<ANY REALM NAME HERE>/metrics`). But the service monitor only configures an endpoint for the server statistics at `/metrics`.

Besides that there are no default Prometheus rules in the chart. And no Grafana dashboard.

Add scraping of the realm metrics, add the [Prometheus rules from the Keycloak Operator](https://github.com/keycloak/keycloak-operator/blob/master/pkg/model/prometheus_rule.go) and add a dashboard based on [this](https://grafana.com/grafana/dashboards/10441) dashboard from Grafana Labs.

I would like to provide the PR for this, but I'm not entirely sure on how to do it. I would like to make it all optional and deactivate it by default, so that updating will not change anything in the release.

My idea is adding keys like this:

```yaml
metrics:
  serviceMonitor:
    enabledRealmMetrics: false

  prometheusRules:
    group1:
      rule1: # add the rules from the operator here
        alert: KeycloakJavaNonHeapThresholdExceeded
        ...

  grafanaDashboards:
    enabled: false
    namespace: monitoring
    labels: {}
    prometheusInstance: default
```

And the respective templates of course.

My open questions are:
* Does it make sense to just add a toggle for the realm metrics? Or should I just enable them when the service monitor is enabled? Or should it get a whole group, so one can configure everything (like relabellings, ...) for it as well?
* Is it ok to add the rules from the operator? Or should it just be empty and the user adds his or her own rules?
* Is it ok to add a dashboard from Grafana Labs? Or should this also be done by the user?

I added the features to my own chart, that depends on this one. But I think it is a valuable addition to the chart. So I'd like to add it here as well.

maxnitze:
### Description of the change

Added the possibility to configure multiple endpoints in the ServiceMonitor and added a template and configuration for PrometheusRules, to make it possible to add alert rules.

### Benefits

Realm metrics are scraped as well and you're able to add alert rules, e.g. the ones from [the Keycloak operator](https://github.com/keycloak/keycloak-operator/blob/master/pkg/model/prometheus_rule.go).

### Possible drawbacks

no limitations known

### Applicable issues

  - fixes #12250

### Checklist


bitnami/keycloak/Chart.yaml
```diff
@@ -26,4 +26,4 @@ name: keycloak
 sources:
   - https://github.com/bitnami/containers/tree/main/bitnami/keycloak
   - https://github.com/keycloak/keycloak
-version: 9.7.2
```
javsalgar, line 29: Could you bump the minor version? It is adding a non-breaking feature

bitnami/keycloak/templates/servicemonitor.yaml
```diff
@@ -1,13 +1,14 @@
-{{- if and .Values.metrics.enabled .Values.metrics.serviceMonitor.enabled }}
+{{- $serviceMonitor := .Values.metrics.serviceMonitor }}
```
javsalgar, line 1: I see that you are using intermediate variables like these. I imagine you are using them for simplicity, right? The issue is that it differs from the other charts and it would make it a bit more difficult to us to apply batch changes in all charts. Could you remove these intermediate vars?

bitnami/keycloak/Chart.yaml
maxnitze: Of course. Fixed that!

bitnami/keycloak/templates/servicemonitor.yaml
maxnitze: Yep, sure. Just some "muscle memory" to do that. But I see your point. Removed them from both the ServiceMonitor and the PrometheusRule templates.

javsalgar (approved):

bitnami/keycloak/templates/servicemonitor.yaml
```diff
@@ -20,25 +20,27 @@ spec:
   jobLabel: {{ .Values.metrics.serviceMonitor.jobLabel }}
   {{- end }}
   endpoints:
-    - port: http
-      {{- if .Values.metrics.serviceMonitor.path }}
-      path: {{ .Values.metrics.serviceMonitor.path }}
+    {{- $endpoints := ternary (.Values.metrics.serviceMonitor.endpoints) (list (dict "path" .Values.metrics.serviceMonitor.path)) (empty .Values.metrics.serviceMonitor.path) }}
+    {{- range $endpoints }}
+    {{- $endpoint := merge . $.Values.metrics.serviceMonitor }}
```
scootermon, line 25: @maxnitze, this is a rather big issue as this merge here creates a reference cycle.
Merge modifies the "dest" argument, which in this case is potentially a dict from the `serviceMonitor.endpoints` list. Since we're merging with the `serviceMonitor` object itself, we end up inserting a reference to `serviceMonitor.endpoints` into every item. In other words, we can do `$endpoint.endpoints[0].endpoints[0].endpoints[0].endpoints[0]` to infinity.

I assume this is also the root cause for #12316.

One way to resolve this is to deep copy the endpoint. Since this is a small dict, we can ignore the performance penalty.

```
{{- $endpoint := merge (deepCopy .) $.Values.metrics.serviceMonitor }}
```

maxnitze: I see the issue. I also thought about that when adding that. But I tested - at least with `helm template` and there was no issue then.

I'll create a PR to fix this. The `deepCopy` fix is even part of the documentation, as it seems. So  I'll go for that.

Just to understand the issue: Does `template` differ from `upgrade --install` in that case? Or is some specific setup necessary?

maxnitze: #12333

scootermon: I haven't confirmed this with tests, so take this with a grain of salt, but this the specific error originates from Helm's storage driver. So when installing the chart, Helm tries to store the final [Release](https://github.com/helm/helm/blob/4b18b19a5e0b11450b9dc92edc75bdd7891c1f4e/pkg/release/release.go#L22) object using the selected storage driver (by default, this is the "secret" storage driver which uses k8s secrets) and at this point, the `Release.Config` (aka the `.Values`) are the ones modified by the rendering process, so they include this reference cycle. I'm not sure whether I consider this a feature or a bug, but I guess that's how it is :D
So when the driver tries to marshal the release object into JSON, the json package throws an error because of this reference cycle:

https://github.com/helm/helm/blob/4b18b19a5e0b11450b9dc92edc75bdd7891c1f4e/pkg/storage/driver/util.go#L36

This error is then caught by the driver which spits out the error message that we see in #12316

https://github.com/helm/helm/blob/a499b4b179307c267bdf3ec49b880e3dbd2a5591/pkg/storage/driver/secrets.go#L154-L157

So yeah, the magic sauce is that Helm needs to try to prepare the stored release, which doesn't happen when using template.

Also, thanks for your swift action :)

PR #12303, opened 2022-09-07
[bitnami/rabbitmq] Use custom probes if given

orgads:
Without this change, in order to use a custom probe, the user has to specify the probe parameters, and also must specify for the original
probe "enabled: false".

### DIFF
--- a/bitnami/rabbitmq/Chart.yaml
+++ b/bitnami/rabbitmq/Chart.yaml
@@ -23,4 +23,4 @@ name: rabbitmq
 sources:
   - https://github.com/bitnami/containers/tree/main/bitnami/rabbitmq
   - https://www.rabbitmq.com
-version: 10.3.2
+version: 10.3.3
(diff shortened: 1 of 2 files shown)


orgads:
Is this a known and documented convention in all the bitnami charts that custom probes require explicitly disabling the default ones?

carrodher (approved):
Changes LGTM, it makes more sense to, by default, use the custom probes if they are specified regardless of if the regular probes are enabled or disabled

orgads:
I see that all the images have the same logic that I replaced here. Not sure what's better - replace everywhere, revert this, or remain inconsistent.

Notice that it has a potential to break compatibility, in case a user did configure custom probe, but prefers not to use it, so they leave the default enabled (I doubt many users are aware of that though).

orgads:
I'll open an issue for that.

orgads:
Done: #12354.

PR #12374, opened 2022-09-12
Use custom probes if given

### LINKED ISSUE #12354: Custom probes only work if the original probe is explicitly disabled
### Name and Version

all

* Configure `customLivenessProbe`, `customReadinessProbe` or `customStartupProbe`
* Do *not* set `enabled: false` for the matching default probe.

_No response_

The custom probe should be activated.

### What do you see instead?

It is not activated, unless I explicitly set `default: false` for the default probe.

### Additional information

Sample template:
```yaml
            exec:
              ...
          {{- else if .Values.customLivenessProbe }}
          livenessProbe: {{- include "common.tplvalues.render" (dict "value" .Values.customLivenessProbe "context" $) | nindent 12 }}
          {{- end }}
```

So if the user configured `customLivenessProbe` but doesn't `livenessProbe.enabled` to false, the custom probe is not effective.

This is unintuitive. If I set a custom probe, I expect it to just work, regardless of other options.

I pushed a fix for rabbitmq: https://github.com/bitnami/charts/pull/12303, but I still have some concerns.

1. This pattern is widely used across all the bitnami images. It should either be fixed in all of them, or none.
2. Was it an intentional design choice, or was this case just overlooked?
3. Changing the order is not completely backwards compatible. If a user intended to configure a custom probe, and still **not** use it, by leaving the default `enabled`, or explicitly setting `enabled: true` for the default, this case will break. I can't find a good reason for a user to do that though.

Thanks!

orgads:
### Description of the change
When custom probes are provided, activate them even if the default probe doesn't have `enabled: false`.

### Benefits
Without this change, in order to use a custom probe, the user has to specify the probe parameters, and also must specify for the original probe `enabled: false`.

### Possible drawbacks
If the user specified a custom probe, but intended to *disable* it by enabling (or not disabling) the default probe, this will no longer work. It is very unlikely that a user would do that though.

Done using this Ruby script:
```ruby
chart_updated = {}
Dir.glob('**/templates/**/*').each do |file|
  next unless File.file?(file)
  s = File.read(file)
  if s.gsub!(/\{\{- if\s+(?:and )?(\S*Probe\.enabled\s+\}\}\n[\s\S]+?)\{\{- else if\s+(\S*\.custom\w+Probe\s+\}\}\n[\s\S]+?)(?=\{\{- end \}\})/, '{{- if \2{{- else if \1')
    File.write(file, s)
    dir = file.split('/')[0..1].join('/')
    next if !dir.start_with?('bitnami/') || chart_updated[dir]
    chart = File.read("#{dir}/Chart.yaml")
    chart.sub!(/(?<=^version: ).*/) { |ver| ver.split('.').tap { |v| v[-1] = v[-1].to_i + 1 }.join('.') }
    File.write("#{dir}/Chart.yaml", chart)
    chart_updated[dir] = true
  end
end
```

### Applicable issues
Fixes #12354.

### Checklist

### DIFF
--- a/bitnami/airflow/Chart.yaml
+++ b/bitnami/airflow/Chart.yaml
@@ -32,4 +32,4 @@ name: airflow
 sources:
   - https://github.com/bitnami/containers/tree/main/bitnami/airflow
   - https://airflow.apache.org/
-version: 13.1.5
+version: 13.1.6
--- a/bitnami/airflow/templates/config/configmap.yaml
+++ b/bitnami/airflow/templates/config/configmap.yaml
@@ -180,26 +180,26 @@ data:
           resources: {{- toYaml .Values.worker.resources | nindent 12 }}
           {{- end }}
           {{- if not .Values.diagnosticMode.enabled }}
-          {{- if .Values.worker.startupProbe.enabled }}
+          {{- if .Values.worker.customStartupProbe }}
+          startupProbe: {{- include "common.tplvalues.render" (dict "value" .Values.worker.customStartupProbe "context" $) | nindent 12 }}
+          {{- else if .Values.worker.startupProbe.enabled }}
           startupProbe: {{- include "common.tplvalues.render" (dict "value" (omit .Values.worker.startupProbe "enabled") "context" $) | nindent 12 }}
             tcpSocket:
               port: worker
-          {{- else if .Values.worker.customStartupProbe }}
-          startupProbe: {{- include "common.tplvalues.render" (dict "value" .Values.worker.customStartupProbe "context" $) | nindent 12 }}
           {{- end }}
-          {{- if .Values.worker.livenessProbe.enabled }}
+          {{- if .Values.worker.customLivenessProbe }}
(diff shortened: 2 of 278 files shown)


orgads:
This is a moving target. Luckily, the script updates all the versions, so I'm resolving the conflicts simply by rerunning it on latest master :)

Please review.

rafariossaa:
Hi,
  First of all, thank you very much for sending this PR.
  However, due to our CI/CD pipeline, we need this PR to be divided into one PR per chart. If not, the verification and publishing won't be performed properly.
  Could you split this PR ?

orgads:
I'll need github API for that.

I'll try.

rafariossaa:
Thanks.

orgads:
Ok, I won't include the script in each commit message. I'll copy it to the issue, and will reference the issue from all the commit messages. So my message template will be:
```
[bitnami/xxx] Use custom probes if given

probe "enabled: false".

Issue: #12354
```

Accepted?

orgads:
@rafariossaa?

rafariossaa:
Hi, Sorry for the delay.
  Those 2 PR looks ok, they need to be reviewed and merged.

orgads:
Pushed as separate PRs.

PR #12458, opened 2022-09-16
[bitnami/spring-cloud-dataflow] Revisit tests

FraPazGal:

### Description of the change

Tested in https://github.com/FraPazGal/charts/pull/107

### Benefits

### Possible drawbacks

None

### Applicable issues

NA

### Additional information

NA

### Checklist


.vib/spring-cloud-dataflow/cypress/cypress/fixtures/imports.json
```diff
@@ -0,0 +1,4 @@
+{
+  "taskStarters": "Task application starters for Docker",
+  "streamStarters": "Stream application starters for RabbitMQ/Docker"
```
agomezmoron, line 3: I don't see where we use here those imports.json. Can we remove RabbitMQ string for another one if it's something to have a string? I don't like mixing product names.

agomezmoron (changes requested):
I requested a minor change. In general, LGTM

.vib/spring-cloud-dataflow/cypress/cypress/fixtures/imports.json
FraPazGal: This use case is a bit special. Spring CD includes some application starters not imported by default. These imported apps are needed to create tasks and streams, so we must do the import before starting our tests. These 2 strings are just used to select the correct option in the imports section:

Screenshot 2022-09-20 at 18 25 27

I could avoid having these strings in a fixture file (maybe as a `const` variable) but I left them there as that's how they were before the revisit.

agomezmoron: No worries @FraPazGal , I wasn't aware of that detail, so LGTM

agomezmoron (approved):

PR #12483, opened 2022-09-19
[bitnami/cassandra] Use custom probes if given

orgads:
probe "enabled: false".

Issue: #12354


migruiz4 (changes requested):
Hi @orgads, could you please update the chart version?
Other that that, the changes look good to me

carrodher (approved):

orgads:
Done

carrodher (approved):

migruiz4 (approved):

PR #12490, opened 2022-09-19
[bitnami/dokuwiki] Use custom probes if given

orgads:
probe "enabled: false".

Issue: #12354

### DIFF
--- a/bitnami/dokuwiki/Chart.yaml
+++ b/bitnami/dokuwiki/Chart.yaml
@@ -26,4 +26,4 @@ name: dokuwiki
 sources:
   - https://github.com/bitnami/containers/tree/main/bitnami/dokuwiki
   - http://www.dokuwiki.org/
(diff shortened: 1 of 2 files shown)


orgads:
ping

corico44 (approved):
LGTM

corico44 (changes requested):
The PR format is not correct.

orgads:
@corico44 I initially opened a PR for all the charts, in which I used the template. See #12374. But then I was asked to split it and create a PR for each chart.

Obviously, this is too much for manual work (96 charts), so I wrote a script that creates these PRs.

All the other charts are already merged, and nobody complained about the template.

Does it really matter? If you insist, I'll change it, but I really don't see the value here.

corico44:

corico44 (approved):

PR #12498, opened 2022-09-19
[bitnami/ghost] Use custom probes if given

orgads:
probe "enabled: false".

Issue: #12354


migruiz4 (commented):
Hi @orgads,

We have released a new version of the bitnami/ghost 19.1.10, could you please rebase this branch and update the version in the Chart.yaml?

PR #12530, opened 2022-09-19
[bitnami/moodle] Use custom probes if given

orgads:
probe "enabled: false".

Issue: #12354


fmulero (approved):
Thanks a lot @orgads

PR #12539, opened 2022-09-19
[bitnami/oauth2-proxy] Use custom probes if given

orgads:
probe "enabled: false".

Issue: #12354


corico44 (changes requested):
The PR format is not correct.

corico44 (approved):

PR #11297, opened 2022-07-21
[bitnami/spark] Init support scripts for spark chart added

corico44:
### Description of the change

Add init scripts support to the Spark chart

### Benefits

Users can to add custom init scripts to Spark so they can, for example, add custom plugins.

### Checklist

- [X] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/). This is *not necessary* when the changes only affect README.md files.
- [X] Variables are documented in the values.yaml and added to the `README.md` using [readme-generator-for-helm](https://github.com/bitnami-labs/readme-generator-for-helm)
- [X] Title of the pull request follows this pattern [bitnami/<name_of_the_chart>] Descriptive title
- [X] All commits signed off and in agreement of [Developer Certificate of Origin (DCO)](https://github.com/bitnami/charts/blob/master/CONTRIBUTING.md#sign-your-work)

PR #12551, opened 2022-09-19
[bitnami/pytorch] Use custom probes if given

orgads:
probe "enabled: false".

Issue: #12354


orgads:
ping

corico44 (approved):
LGTM

corico44 (changes requested):
Please, @orgads , modify the PR format since it is not correct. It should look something like this: https://github.com/bitnami/charts/pull/11297. When you try to create a new PR, GitHub gives you a template. Modify this PR using that format.

corico44 (approved):

PR #12578, opened 2022-09-19
[bitnami/concourse] Add tests and publishing using VIB

joancafom:
### Description of the change

The main objective of this PR is to publish our Bitnami Concourse Helm Chart using VMware Image Builder. In order to do that, several changes are included:

- Adapting the chart to use `common.names.fullname.namespace`, which allows deploying several releases under the same K8s cluster but in different namespaces.

### Benefits

### Possible drawbacks

### Applicable issues

NA

### Additional information

NA

### Checklist


agomezmoron (dismissed):
LGTM

.vib/concourse/goss/web/vars.yaml
```diff
@@ -0,0 +1,34 @@
...
+    fsGroup: 1002
+  containerSecurityContext:
+    runAsUser: 1002
+  serviceAccount:
+    automountServiceAccountToken: true
+service:
+  web:
+    ports:
+      http: 80
+      https: 444
```
FraPazGal, line 15: You are missing the `http` test checking the https port of the K8s svc.

.vib/concourse/goss/web/vars.yaml
```diff
@@ -0,0 +1,34 @@
+secrets:
+  localUser: "bitnamiUser"
```
FraPazGal, line 2: Is there a need to have some of these values between double quotes? Yaml should be treating them as strings in the `goss.yaml` file even without them right?

FraPazGal (commented):
Thanks for your contribution @joancafom! Please check the couple of comments I left as a review when you have some time.

.vib/concourse/goss/web/goss.yaml
```diff
@@ -0,0 +1,26 @@
+file:
```
FraPazGal, line 1: I think you missed checking the `volumeMounts` both for the web and worker pods.

.vib/concourse/goss/web/goss.yaml
```diff
@@ -0,0 +1,26 @@
...
+  {{- $pipeline := printf "pipe_%s" (randAlphaNum 5 | toLower) }}
+  {{- $port := .Vars.service.web.ports.http }}
+  create-pipeline-trigger-job:
+    exec: fly -t {{ $target }} login -c http://concourse-web:{{ $port }} -u {{ $user }} -p '{{ $pwd }}' && (echo '{{ .Vars.examplePipeline | b64enc }}' | base64 -d) > /tmp/pipeline.yaml && fly -t {{ $target }} set-pipeline -p {{ $pipeline }} -c /tmp/pipeline.yaml -n && fly -t {{ $target }} unpause-pipeline -p {{ $pipeline }} && fly -t {{ $target }} trigger-job -j {{ $pipeline }}/job && sleep 35 && fly -t {{ $target }} jobs -p {{ $pipeline }}
+    exit-status: 0
+    stdout:
+      - "succeeded"
+    timeout: 40000
+  check-db-config:
+    exec: $(echo $CONCOURSE_POSTGRES_DATABASE | grep -q {{ .Vars.postgresql.auth.database }}) && $(echo $CONCOURSE_POSTGRES_USER | grep -q {{ .Vars.postgresql.auth.username }}) && $(echo $CONCOURSE_POSTGRES_PASSWORD | grep -q {{ .Vars.postgresql.auth.password }})
```
FraPazGal, line 19: I like a lot this test, I think it tests the db_auth config from the chart viewpoint better than checking the stings in a config file for example.

Actually, could you replicate this test to check the containerPorts? Everyone of them uses a `*_BIND_PORT` env, so I think it is feasible and adds value to the tests.

.vib/concourse/goss/web/vars.yaml
joancafom: The `https` test is skipped on purpose as the chart currently presents some problems with it. We may add it in the future when it is fixed, though.

.vib/concourse/goss/web/vars.yaml
joancafom: Yes, there is no reason to use double quotes for that specific entry, I might have missed it. I generally only use quotes when special characters are used or to ensure the content is not confused with another type.

.vib/concourse/goss/web/goss.yaml
joancafom: As discussed, I will only add an entry for the `/bitnami/concourse/workdir` directory as the other one containing the keys it's being used by the application

.vib/concourse/goss/web/goss.yaml
joancafom: Done!

FraPazGal (approved):
LGTM

PR #12647, opened 2022-09-23
[bitnami/argocd] Additional configuration for wait-for-redis container

### LINKED ISSUE #12645: [bitnami/argo-cd] wait-for-redis init container is hard to configure
### Name and Version

bitnami/argocd

I'm running argocd and want to encrypt the connection to redis. The 3 argocd components themselves offer plenty of support for that using the `extraArgs` parameter, however the init container fails because it does not expect redis to use TLS. I cannot disable the init container or change its arguments. I tried to use istio service mesh for TLS, but that disables networking capabilities of the init container unless it runs as UID 1337. The securityContext for the init container is not configurable either.

Several improvements:
 - allow disabling the init container
 - allow setting the securitycontext for the init container
 - allow adding parameters to the redis-cli call

We could simply not encrypt the connection to redis, but since it handles caching of argocd (which might contain some Secrets) that is not acceptable from a security perspective.

It might be possible to configure the container using an extra initContainer with the same name as the wait-for-redis container, which then contains the extra configuration. This is an alternative solution but is harder to understand (you'll end up with half an init container defined in the values) than a few parameters defined in the chart.

bh-tt:
### Description of the change
Add several `redisWait` values which control the init container waiting for redis. This includes disabling the wait, setting securityContext or adding additional arguments to the redis-cli call.

### Benefits
It makes it possible to run an argoCD deployment with TLS-enabled redis.

### Possible drawbacks
None.

### Applicable issues

  - fixes #12645

### Additional information
The variable in question is named `redisWait` now, (cannot use `redis` prefix as that might cause conflicts with the redis chart dependency). If you have a better name please advise.

### Checklist

### DIFF
--- a/bitnami/argo-cd/templates/application-controller/deployment.yaml
+++ b/bitnami/argo-cd/templates/application-controller/deployment.yaml
@@ -65,9 +65,13 @@ spec:
       securityContext: {{- omit .Values.controller.podSecurityContext "enabled" | toYaml | nindent 8 }}
       {{- end }}
       initContainers:
+        {{- if .Values.redisWait.enabled }}
         - name: wait-for-redis
           image: {{ include "common.images.image" (dict "imageRoot" .Values.redis.image "global" .Values.global) }}
           imagePullPolicy: {{ .Values.redis.image.pullPolicy | quote }}
+          {{- with .Values.redisWait.securityContext }}
+          securityContext: {{ . | toYaml }}
+          {{- end }}
           command:
             - /bin/bash
           args:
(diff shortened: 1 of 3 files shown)


rafariossaa:
Hi, 
  Could you take a look to the error reported by the CI/CD ?
  ```
==> Linting /tmp/puzzle-workplace-11294783707873891899/extracted/bh-tt-charts-d4ed208/bitnami/argo-cd
[ERROR] templates/: parse error at (argo-cd/templates/server/deployment.yaml:111): invalid syntax
```

bh-tt:
CI is fixed, had some confusion between ansible yml with default values and helm.

carrodher (approved):
LGTM

PR #12700, opened 2022-09-27
[bitnami/owncloud] Review Cypress and Goss tests for Owncloud charts

mdhont:

### Description of the change

Review Cypress and Goss tests for Owncloud

### Checklist


agomezmoron (approved):
LGTM

.vib/owncloud/cypress/cypress/integration/owncloud_spec.js
```diff
@@ -4,77 +4,18 @@ import { random } from '../support/utils';
...
+    cy.contains(`${folder.newFolder.name}.${random}`).click();
   });
   cy.get('.new').click();
   cy.get('[data-action="upload"]').click();
   cy.get('input[type="file"]').selectFile(
     'cypress/fixtures/file_to_upload.json',
     { force: true }
   );
-  cy.contains('No files in here').should('not.be.visible');
+  cy.contains('file_to_upload').should('be.visible');
```
FraPazGal, line 19: Would it be possible to expand this test so Cypress downloads the uploaded file? Given owncloud's file-sharing nature, we are only testing "half" of the use case here.

.vib/owncloud/cypress/cypress/integration/owncloud_spec.js
```diff
@@ -4,77 +4,18 @@ import { random } from '../support/utils';
 it('allows creating a folder and uploading a file ', () => {
   cy.login();
   cy.get('.new').click();
-  cy.get('[data-action="folder"]').click();
+  cy.contains('Folder').click();
```
FraPazGal, line 7: Could we use a `visit` here to start the test from the "New Folder" page and avoid these 2 clicks?

.vib/owncloud/goss/vars.yaml
```diff
@@ -0,0 +1,18 @@
+mariadb:
+  auth:
+    database: test_db_name
+    username: test_db_user
+    password: test_db_password
+externalDatabase:
+  port: 3306
```
FraPazGal, line 7: We are using the built-in MariaDB chart so this is not used. Although you are using this param to check for the used MariaDB, we should avoid it to avoid confusion. If there is no param for modifying the MariaDB port in OwnCloud's values we don't need to check for it.

.vib/owncloud/goss/goss.yaml
```diff
@@ -1,18 +1,15 @@
...
 file:
-  /opt/bitnami/apache/conf/extra/httpd-manual.conf:
-    mode: "0664"
-    filetype: file
-    exists: true
-  /opt/bitnami/mysql:
-    mode: "0755"
-    filetype: directory
-    exists: true
   /bitnami/owncloud/data:
```
FraPazGal, line 13: The correct path to be used would be `/bitnami/owncloud` as that is the one used in the volumeMount:

https://github.com/bitnami/charts/blob/8ec7c240ad509a21d938e1627b7281a088c54a3e/bitnami/owncloud/templates/deployment.yaml#L229

.vib/owncloud/vib-publish.json
```diff
@@ -22,7 +22,7 @@
           "url": "{SHA_ARCHIVE}",
           "path": "/bitnami/owncloud"
         },
-        "runtime_parameters": "b3duY2xvdWRIb3N0OiB2bXdhcmUtb3duY2xvdWQubXkKb3duY2xvdWRVc2VybmFtZTogdXNlcgpvd25jbG91ZFBhc3N3b3JkOiBDb21wbGljYXRlZFBhc3N3b3JkMTIzITQKb3duY2xvdWRFbWFpbDogdGVzdEBleGFtcGxlLmNvbQpzZXJ2aWNlOgogIHR5cGU6IExvYWRCYWxhbmNlcgogIHBvcnRzOgogICAgaHR0cDogODAKc210cEhvc3Q6IHNtdHB0ZXN0aG9zdApzbXRwUG9ydDogOTgyMQpzbXRwVXNlcjogc210cFVzZXIKc210cFBhc3N3b3JkOiBzbXRwQ29tcGxpY2F0ZWRQYXNzd29yZDEyMyE0",
+        "runtime_parameters": "b3duY2xvdWRIb3N0OiB2bXdhcmUtb3duY2xvdWQubXkKb3duY2xvdWRVc2VybmFtZTogdGVzdF91c2VyCm93bmNsb3VkUGFzc3dvcmQ6IENvbXBsaWNhdGVkUGFzc3dvcmQxMjMhNApzZXJ2aWNlOgogIHR5cGU6IExvYWRCYWxhbmNlcgogIHBvcnRzOgogICAgaHR0cDogODAKICAgIGh0dHBzOiA4NDQ0Cm1hcmlhZGI6CiAgZW5hYmxlZDogdHJ1ZQogIGF1dGg6CiAgICBkYXRhYmFzZTogdGVzdF9kYl9uYW1lCiAgICB1c2VybmFtZTogdGVzdF9kYl91c2VyCiAgICBwYXNzd29yZDogdGVzdF9kYl9wYXNzd29yZApwb2RTZWN1cml0eUNvbnRleHQ6CiAgZnNHcm91cDogMTAwMgpjb250YWluZXJTZWN1cml0eUNvbnRleHQ6CiAgcnVuQXNVc2VyOiAxMDAyCmNvbnRhaW5lclBvcnRzOgogIGh0dHA6IDgwODEKICBodHRwczogODQ0NApob3N0QWxpYXNlczoKICAtIGlwOiAxMjcuMC4wLjEKICAgIGhvc3RuYW1lczoKICAgICAgLSBzdGF0dXMudGVzdGhvc3Q=",
```
FraPazGal, line 25: Please add the `podSecurityContext.enabled` and `containerSecurityContext.enabled` params.

FraPazGal (commented):
Thanks for the PR @mdhont! I left a few comments, please check them when possible.

.vib/owncloud/cypress/cypress/integration/owncloud_spec.js
mdhont: It's not possible, the app uses javascript to send the form, there is no direct URL.

FraPazGal (commented):
Hi @mdhont, could you please check my newest comments?

.vib/owncloud/cypress/cypress/integration/owncloud_spec.js
```diff
@@ -4,77 +4,31 @@ import { random } from '../support/utils';
...
-    cy.get('input')
-      .invoke('val')
-      .should('contain', 'vmware')
-      .then((link) => {
+    cy.contains(`${folder.newFolder.name}.${random}`).click();
+    cy.get('span').contains(`${folder.newFolder.name}.${random}`).click();
+    cy.contains('No files in here');
+    cy.get('.new').click();
+    cy.get('[data-action="upload"]').click();
+    cy.get('input[type="file"]').selectFile(
```
joancafom, line 16: Don't think you need the element type, `file` usually is specific enough.
```suggestion
    cy.get('[type="file"]').selectFile(
```

.vib/owncloud/cypress/cypress/integration/owncloud_spec.js
```diff
@@ -4,77 +4,31 @@ import { random } from '../support/utils';
...
         cy.request({
+          url: href,
           method: 'GET',
-          url: link,
-          form: true,
         }).then((response) => {
-          expect(response.status).to.eq(200);
-        });
+            expect(response.status).to.eq(200);
+            expect(response.headers['content-type']).to.eq('text/plain;charset=UTF-8');
```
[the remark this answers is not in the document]
mdhont: The shared URL that is created points to the page with the file where the user can click and download. Unfortunately downloading the file through clicking doesn't seem to be working and the file does not have a direct download link. In any case, I will add some text to the file.

.vib/owncloud/cypress/cypress/integration/owncloud_spec.js
```diff
@@ -4,77 +4,31 @@ import { random } from '../support/utils';
...
-    cy.get('[data-action="Share"]').click();
-  });
-  cy.get('[class="permalink"]').click();
-  cy.get('.detailFileInfoContainer').within(() => {
-    cy.get('input')
-      .invoke('val')
-      .should('contain', 'vmware')
-      .then((link) => {
+    cy.contains(`${folder.newFolder.name}.${random}`).click();
+    cy.get('span').contains(`${folder.newFolder.name}.${random}`).click();
```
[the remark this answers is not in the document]
mdhont: This is not duplicate code. The first line creates a folder (using a small form inside a list element). The second line clicks on the created folder, because I want to add the file to the new folder.

.vib/owncloud/cypress/cypress/integration/owncloud_spec.js
```diff
@@ -4,77 +4,31 @@ import { random } from '../support/utils';
...
+    cy.contains('No files in here');
+    cy.get('.new').click();
+    cy.get('[data-action="upload"]').click();
+    cy.get('input[type="file"]').selectFile(
+      'cypress/fixtures/file_to_upload.json',
+      { force: true }
+    );
+    cy.contains('file_to_upload').should('be.visible');
+    cy.get('[href*="file_to_upload"]')
```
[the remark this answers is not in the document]
FraPazGal: Indeed, you need to add extra logic here. The link you are using is a private one but you have the option to generate a public one. That would alow yo to access the link, download the file, and then check if the downloaded file is the same as the one uploaded. You can take as reference MinIO's test:

https://github.com/bitnami/charts/blob/5278e0000ee42dfb5d39df3d385882a15dbffea9/.vib/minio/cypress/cypress/integration/minio_spec.js#L21

.vib/owncloud/cypress/cypress/integration/owncloud_spec.js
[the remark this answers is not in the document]
mdhont: When I remove any of these 2 lines the upload fails for some reason. Instead of spending time trying to find out why that's the case I'd rather leave it like this or remove both two lines. The latter option would add the file outside the created folder but it doesn't really matter in my opinion.

.vib/owncloud/cypress/cypress/integration/owncloud_spec.js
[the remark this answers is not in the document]
mdhont: Is this really necessary? Creating a public link requires a lot of extra clicks and extra fields must be filled, like the password and the expiry date. To download the file the user must log in with the created password and then the download page appears. These steps will add flakiness to the tests in my opinion.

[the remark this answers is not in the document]
FraPazGal: I would say this is indeed necessary, given that it is the core feature of owncloud. I checked the use case and, though you have the option to set a password and expiration date, they are not necessary. It is neither necessary to log in, as I have been able to access and download the file directly from the created public link in an incognito window.

It is true every additional step adds flakiness to the tests but this shouldn't be a concern when creating the necessary ones.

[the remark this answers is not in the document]
mdhont: Ok, got it. I will work on it

.vib/owncloud/cypress/cypress/integration/owncloud_spec.js
```diff
@@ -4,77 +4,36 @@ import { random } from '../support/utils';
 it('allows creating a folder and uploading a file ', () => {
```
joancafom, line 4: If the test does also download the file, please update the description string accordingly

.vib/owncloud/cypress/cypress/integration/owncloud_spec.js
```diff
@@ -4,77 +4,36 @@ import { random } from '../support/utils';
...
-it('allows adding a group and a user', () => {
-  cy.login();
-  cy.visit('/index.php/settings/users');
-  cy.contains('Add Group').click();
-  cy.fixture('groups').then((group) => {
-    cy.get('#newgroupname').type(`${group.newGroup.name}.${random}`);
-    cy.get('#newgroup-form').within(() => {
-      cy.get('[type="submit"]').click();
+    cy.get('span').contains(`${folder.newFolder.name}.${random}`).click();
+    cy.contains('No files in here');
```
joancafom, line 12: I think this line does not add any value (only flakiness). I'd remove it if it is not strictly necessary.

.vib/owncloud/cypress/cypress/integration/owncloud_spec.js
```diff
@@ -4,77 +4,36 @@ import { random } from '../support/utils';
...
       .then((link) => {
         cy.request({
-          method: 'GET',
-          url: link,
-          form: true,
+          url: `${link}/download`,
         }).then((response) => {
-          expect(response.status).to.eq(200);
-        });
+          const mime = response.body
```
joancafom, line 34: You are declaring a constant here, but never using it.

.vib/owncloud/cypress/cypress/integration/owncloud_spec.js
```diff
@@ -4,77 +4,36 @@ import { random } from '../support/utils';
...
+          expect(response.body).to.contain('testJson')
```
joancafom, line 35: This test will fail if the content of the example file is changed. Instead, it is a better practice to read the content of the file and compare it to the downloaded one.

Please, take a look to 

https://github.com/bitnami/charts/blob/6d9b5a9ad6007c2e27a936d3f800b00425aae15d/.vib/minio/cypress/cypress/integration/minio_spec.js#L26 

for a similar test

.vib/owncloud/cypress/cypress/integration/owncloud_spec.js
```diff
@@ -1,80 +1,41 @@
 /// <reference types="cypress" />
 import { random } from '../support/utils';
 
-it('allows creating a folder and uploading a file ', () => {
+it('allows creating a folder, then upload and then download a file ', () => {
```
FraPazGal, line 4:
```suggestion
it('allows creating a folder, uploading a file to it, and downloading said file ', () => {
```

FraPazGal (dismissed):
Added a minor wording suggestion, but otherwise LGTM

agomezmoron (dismissed):

agomezmoron (approved):

FraPazGal (approved):

joancafom (approved):

PR #12797, opened 2022-10-04
[bitnami/redis] Revisit tests and remove extra pipelines

FraPazGal:
### Description of the change

Additionally, this revisit has reduced the number of executed pipelines for Redis to one, and thus its related CI file has been removed.

Tested in https://github.com/FraPazGal/charts/pull/121

### Checklist


.vib/redis/goss/goss.yaml
```diff
@@ -0,0 +1,64 @@
...
+  {{ .Vars.master.persistence.path }}:
+    filetype: directory
+    exists: true
+    mode: "2775"
+    owner: root
+  /opt/bitnami/scripts/start-scripts:
+    filetype: directory
+    exists: true
+    mode: "2777"
+    owner: root
```
joancafom, line 45: I would omit this path. There already are a bunch of other paths in the tests, but in particular, this one is used by the primary container to start the Redis service. An error related to this mounted volume would probably end up as a deployment failure.

.vib/redis/goss/goss.yaml
```diff
@@ -0,0 +1,64 @@
...
+  /opt/bitnami/redis/etc:
+    filetype: directory
+    exists: true
+    mode: "2777"
+    owner: root
+  /health:
+    filetype: directory
+    exists: true
+    mode: "2777"
+    owner: root
```
joancafom, line 60: Same thoughts here.

.vib/redis/goss/goss.yaml
```diff
@@ -0,0 +1,64 @@
...
+  {{- $key := printf "key_%s" (randAlpha 5) }}
+  {{- $value := printf "value_%s" (randAlpha 5) }}
+  redis-set-key-value-pairs:
+    exec: export {{ $auth }} && redis-cli {{ $master_endpoint }} SET {{ $key }} {{ $value }} {{ range $e, $i := until $replicas }} && redis-cli -h redis-replicas-{{ $i }}.redis-headless -p {{ $.Vars.replica.containerPorts.redis }} GET {{ $key }} | grep -q {{ $value }}{{ end }}
+    exit-status: 0
+  redis-disabled-command:
+    exec: {{ $auth }} redis-cli {{ $master_endpoint }} {{ .Vars.master.disableCommands }}
+    exit-status: 0
+    stdout:
+      - ERR unknown command
```
joancafom, line 15: I would probably use a loop here, as `disableCommands` is originally an array.

joancafom (commented):
Thanks for your PR @FraPazGal ! I left you some minor comments for you to review 😁

.vib/redis/goss/goss.yaml
agomezmoron: Good catch!

.vib/redis/goss/goss.yaml
FraPazGal: You are right, removed both of them.

agomezmoron (approved):

joancafom (approved):

PR #12818, opened 2022-10-05
[bitnami/postgresql] Update maintainers

carrodher:
### Description of the change

In this PR the maintainers' metadata is updated to show the team in charge of the maintenance, updates, and support of this Helm chart.

We always appreciate the initial contribution of @desaintmartin when migrating this Helm chart from Helm stable

### Benefits

On websites like [ArtifactHUB](https://artifacthub.io/packages/helm/bitnami/postgresql), the current maintainers are showed

### DIFF
--- a/bitnami/postgresql/Chart.yaml
+++ b/bitnami/postgresql/Chart.yaml
@@ -22,10 +22,8 @@ keywords:
 maintainers:
   - name: Bitnami
     url: https://github.com/bitnami/charts
-  - email: cedric@desaintmartin.fr
-    name: desaintmartin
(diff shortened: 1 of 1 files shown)


desaintmartin:
It's been fun and we're still heavy users of bitnami charts, and we'll happily continue to contribute when needed.

carrodher:
Sure! This is more a cosmetic change since depending on how the metadata is rendered there are some websites where Bitnami is not present as maintainers which can be confusing for users using this Helm chart:
Screenshot 2022-10-05 at 12 44 06

PR #12886, opened 2022-10-10
[bitnami/mariadb-galera] changing podManagementPolicy default value

bouaouda-achraf:
the default value needs to be changed because during the running lifecycle either  POD-1 or POD-2 ... can be the leader and after non-cleaning terminating pod the cluster will crash because it will start with pod-0 and not find the leader 

Signed-off-by: ACHRAF BOUAOUDA <33317417+bouaouda-achraf@users.noreply.github.com>

### Applicable issues

  - fixes #

### Checklist


rafariossaa:
Hi,
 I think this is not needed to be in `Parallel` for the default install, and would produce nodes to restart. However, this is indicated to be set to `Parallel` when you need to [start from other nodes than '0'](https://github.com/bitnami/charts/tree/master/bitnami/mariadb-galera#user-content-bootstraping-a-node-other-than-0) .

PR #12932, opened 2022-10-12
[bitnami/common] Do not explicitly specify namespace in affinity term.

### LINKED ISSUE #12668: [bitnami/common] _affinities namespace
### Name and Version

bitnami/common 1.16.0

related to https://github.com/bitnami/charts/issues/10804

Hello. I struggle with the helmchart inside kustomize. The namespace transformer can't reach this value and change it accordingly.

Here is the question: Why even have the podAffinityTerm populated with namespaces list with this single value? It seems to me this setup has the same effect as not defined at all.

_No response_

The podAffinityTerm applies to namespace of redis deployment as in case where namespace is not specified in the podAffinityTerm docs.

```
   namespaces	<[]string>
     namespaces specifies a static list of namespace names that the term applies
     to. The term is applied to the union of the namespaces listed in this field
     and the ones selected by namespaceSelector. null or empty namespaces list
     and null namespaceSelector means "this pod's namespace"
```

### What do you see instead?

If the resource is created for different kubectl context than the current. The current namespace is injected into `podAffinityTerm.namespaces` which does not correspond with the namespace of pod.

### Additional information

I suggest the template

https://github.com/bitnami/charts/blob/master/bitnami/common/templates/_affinities.tpl#L65

should not contain 
```
    namespaces:
      - {{ include "common.names.namespace" .context | quote }}
```

dosmanak:
the current context namespace defined is identical as not defined at all

> null or empty namespaces list and null namespaceSelector means "this pod's namespace"

docs: https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.23/#podaffinityterm-v1-core
issue: #12668

### Applicable issues

  - fixes #12668

### Checklist


fmulero:
Hi @dosmanak It seems your email doesn't match with your GPG key (take a look at the [unverified message](https://docs.github.com/en/authentication/managing-commit-signature-verification/displaying-verification-statuses-for-all-of-your-commits)) and the DCO wasn't triggered. They could be related, please use the right email in the signature and force a new push to trigger the DCO bot

fmulero (approved):
Hi  Petr, thanks a lot for your PR.

LGTM. I don't see any inconvenience in your changes. As you stated this change should have the same behaviour as it is implemented right now and there is no issues during chart upgrades when the namespace is removed from the `podAffinityTerm`.

PR #12943, opened 2022-10-13
[bitnami/*] Use new default branch name in links

fmulero:
Signed-off-by: Fran Mulero <fmulero@vmware.com>

### Description of the change

Change default branch in links.

### Benefits

Keep documentation up to date and avoid unnecessary redirections.

### Possible drawbacks

None


corico44:
It looks like you should update the version in Chart.yaml.

jotamartos (approved):
Changes LGTM but do not think we need to bump the charts' versions, no? Is there a way to avoid that check?

fmulero:
There is no way at the moment to avoid the validation error

PR #12992, opened 2022-10-17
[bitnami/metallb] Add tests and publishing using VIB

joancafom:
### Description of the change

The main objective of this PR is to publish our Bitnami MetalLB Helm Chart using VMware Image Builder. In order to do that, several changes are included:

- Increasing the existing test coverage of the asset by adding Ginkgo tests.

### Benefits

### Possible drawbacks

### Applicable issues

NA

### Additional information

Tested in https://github.com/joancafom/charts/pull/87

### Checklist


agomezmoron (approved):

.vib/metallb/ginkgo/metallb_test.go
```diff
@@ -0,0 +1,74 @@
...
+			createIpAddressPoolResourceOrDie(ctx, dynamicClient, &ipPoolConfig)
+			createL2AdvertisementResourceOrDie(ctx, dynamicClient, &advertConfig)
+		})
+		AfterEach(func() {
+			// Not need to panic here if failed, the cluster is expected to clean up with the undeployment
+			dynamicClient.Resource(l2AdvertisementType).Namespace(*namespace).Delete(ctx, advertConfig.name, metav1.DeleteOptions{})
+			dynamicClient.Resource(ipAddressPoolType).Namespace(*namespace).Delete(ctx, ipPoolConfig.name, metav1.DeleteOptions{})
+		})
+
+		When("a Service resource is created", func() {
```
FraPazGal, line 48: Could we create the testing service at the chart's installation time instead? Checking Metallb doc, it should remain idle until configured with an IPAdressPool and L2Advertisement. In such a case, you may be able to create the LB using the `extraDeploys` param instead of adding this logic here. The testing logic will be the same except for the service creation.

.vib/metallb/ginkgo/metallb_test.go
```diff
@@ -0,0 +1,74 @@
...
+			var service *v1.Service
+			var serviceName string = "lb-vib-tests"
+
+			BeforeEach(func() {
+				var err error
+				createServiceOrDie(ctx, typedClient, serviceName)
+				service, err = typedClient.Services(*namespace).Get(ctx, serviceName, metav1.GetOptions{})
```
FraPazGal, line 55: This `get()` can be avoided, as the `create()` returns the created resource.

FraPazGal (commented):
Thanks for the PR @joancafom. I added a couple of comments, please review them when possible.

.vib/metallb/ginkgo/metallb_test.go
joancafom: Although your proposed workflow makes sense, we would need to ensure that CRDs are installed before the Service is created. This is usually the case when CRDs are placed within the `crds` directory, but in this case, MetalLB [does not define them there](https://github.com/bitnami/charts/blob/main/bitnami/metallb/templates/controller/crds.yaml) (because they need to be templatized). 

This, combined with the fact that CRDs' registration process usually takes some time, makes using the `extraDeploy` an unsafe option: the K8s cluster cloud provider's native LB controller could interfere with MetalLB here.

Hence, I would keep this as it is to ensure the resource is deployed one MetalLB is marked as `ready` 😁

.vib/metallb/ginkgo/metallb_test.go
joancafom: I have tried to update the logic as you suggested, but the `create()` returns the newly created object, right before MetalLB has updated its configuration (and assigned an IP to the service).

Using the `get()` to query the API returns the service once MetalLB has managed to noticed it, though.

FraPazGal (approved):
LGTM!

Issue #12598, opened 2022-09-20
[bitnami/redis] sentinel failover in prestop hook causes write loss


simon-baatz:
### Name and Version

bitnami/redis 17.1.6

### What steps will reproduce the bug?

1. Start a sentinel deployment with default config:  `helm install redis
   bitnami/redis --set sentinel.enabled=true`. Wait until all nodes are up.
   Usually, `redis-node-0` will be the master. This is assumed in the following.

2. Create writes (once per second is enough), the window during which writes are
   lost is typically a couple of seconds. For example, increment a key every
   second and read it on a replica.

   ```bash
   while /bin/true ; do  echo -n "W: " ; kubectl exec redis-node-0 --container redis -- bash -c 'REDISCLI_AUTH="$REDIS_PASSWORD" redis-cli -h localhost -p 6379 incr key' ; sleep 1 ; echo -n "R: " ; kubectl exec redis-node-1 --container redis -- bash -c 'REDISCLI_AUTH="$REDIS_PASSWORD" redis-cli -h localhost -p 6379 get key' ; done
   ```

   (While this is running, collect the logs of `redis-node-0`: `kubectl logs redis-node-0 --container=redis -f` for later analysis)

3. Shutdown the master by deleting it `kubectl delete pod redis-node-0`. Kubernetes will re-create the pod (as a replica node).

4. Observe that although writes are succeeding during failover, they are lost after the failover.

### Are you using any custom parameters or values?

Enable sentinel deployment as shown above.

### What is the expected behavior?

If no error occurs for a write, the written value should not be lost (there are
cases in which Redis does not guarantee this, but a regular shutdown should not be
such a case as Redis explicitly tries to prevent that with its shutdown procedure).

### What do you see instead?

A typical output of the INCR/GET loop above looks like:

```
W: 3
R: 3
W: 4
R: 4
W: 5
R: 4
W: 6
R: 4
W: 7
R: 4
W: 8
R: 4
W: 9
R: 4
W: 10
R: 4
W: 11
R: 4
W: error: Internal error occurred: error executing command in container: container not running (c1318323ffb243cdc98658b6a9c504121422a3f71bc104932474f57fc380d811)
```

Writes to the old master succeed, but the values are not replicated anymore. The
time window during which this happens is surprisingly large. (Note that
according to the client spec, the client only needs to verify the current master
in case of a reconnection. The original problem occurred with a sentinel
protocol aware Jedis client. In this bug report, `redis-cli` is used to make the
[...]

PR #13021, opened 2022-10-19
[bitnami/redis] Fix data loss when executing failover

jotamartos:
Signed-off-by: Jota Martos <jotamartos@vmware.com>

### Description of the change

The master node stops the client write requests during the prestop hook to avoid data loss

### Benefits

Write requests are stopped to avoid data loss

### Possible drawbacks

None from the tests I performed

### Applicable issues

  - fixes https://github.com/bitnami/charts/issues/12598

### Checklist

### DIFF
--- a/bitnami/redis/templates/scripts-configmap.yaml
+++ b/bitnami/redis/templates/scripts-configmap.yaml
@@ -478,7 +478,7 @@ data:
         if is_boolean_yes "$REDIS_TLS_ENABLED"; then
             redis-cli -h 127.0.0.1 -p "$REDIS_TLS_PORT" --tls --cert "$REDIS_TLS_CERT_FILE" --key "$REDIS_TLS_KEY_FILE" --cacert "$REDIS_TLS_CA_FILE" "$@"
         else
-            redis-cli -h 127.0.0.1 -p ${REDIS_PORT} "$@"
+            redis-cli -h 127.0.0.1 -p "$REDIS_PORT" "$@"
         fi
     }
     failover_finished() {
@@ -491,10 +491,11 @@ data:
     [[ -f "$REDIS_PASSWORD_FILE" ]] && export REDISCLI_AUTH="$(< "${REDIS_PASSWORD_FILE}")"
 
     if ! failover_finished; then
-        echo "I am the master pod and you are stopping me. Stopping client connections"
+        echo "I am the master pod and you are stopping me."
+        echo "Stopping client connections"
         # Stopping client write connections to avoid data loss
         run_redis_command CLIENT PAUSE "{{ mul (sub .Values.sentinel.terminationGracePeriodSeconds 10) 1000 }}" WRITE
-        # if I am the master, issue a command to failover once and then wait for the failover to finish
+        # Waiting for the failover to finish
         echo "Waiting for sentinel to run failover for up to {{ sub .Values.sentinel.terminationGracePeriodSeconds 10 }}s"
         retry_while "failover_finished" "{{ sub .Values.sentinel.terminationGracePeriodSeconds 10 }}" 1
     else
--- a/bitnami/redis/templates/scripts-configmap.yaml
+++ b/bitnami/redis/templates/scripts-configmap.yaml
@@ -495,6 +495,8 @@ data:
         echo "Stopping client connections"
         # Stopping client write connections to avoid data loss
         run_redis_command CLIENT PAUSE "{{ mul (sub .Values.sentinel.terminationGracePeriodSeconds 10) 1000 }}" WRITE
+        # Running failover command
+        run_redis_command FAILOVER TIMEOUT "{{ mul (sub .Values.sentinel.terminationGracePeriodSeconds 10) 1000 }}"
         # Waiting for the failover to finish
         echo "Waiting for sentinel to run failover for up to {{ sub .Values.sentinel.terminationGracePeriodSeconds 10 }}s"
         retry_while "failover_finished" "{{ sub .Values.sentinel.terminationGracePeriodSeconds 10 }}" 1
--- a/bitnami/redis/templates/scripts-configmap.yaml
+++ b/bitnami/redis/templates/scripts-configmap.yaml
@@ -492,8 +492,9 @@ data:
 
     if ! failover_finished; then
         echo "I am the master pod and you are stopping me. Stopping client connections"
-        # if I am the master, issue a command to failover once and then wait for the failover to finish
+        # Stopping client write connections to avoid data loss
         run_redis_command CLIENT PAUSE "{{ mul (sub .Values.sentinel.terminationGracePeriodSeconds 10) 1000 }}" WRITE
+        # if I am the master, issue a command to failover once and then wait for the failover to finish
         echo "Waiting for sentinel to run failover for up to {{ sub .Values.sentinel.terminationGracePeriodSeconds 10 }}s"
         retry_while "failover_finished" "{{ sub .Values.sentinel.terminationGracePeriodSeconds 10 }}" 1
     else
--- a/bitnami/redis/templates/scripts-configmap.yaml
+++ b/bitnami/redis/templates/scripts-configmap.yaml
@@ -491,6 +491,9 @@ data:
     [[ -f "$REDIS_PASSWORD_FILE" ]] && export REDISCLI_AUTH="$(< "${REDIS_PASSWORD_FILE}")"
 
     if ! failover_finished; then
+        echo "I am the master pod and you are stopping me. Stopping client connections"
+        # if I am the master, issue a command to failover once and then wait for the failover to finish
+        run_redis_command CLIENT PAUSE "{{ mul (sub .Values.sentinel.terminationGracePeriodSeconds 10) 1000 }}" WRITE
         echo "Waiting for sentinel to run failover for up to {{ sub .Values.sentinel.terminationGracePeriodSeconds 10 }}s"
         retry_while "failover_finished" "{{ sub .Values.sentinel.terminationGracePeriodSeconds 10 }}" 1
     else
--- a/bitnami/redis/templates/scripts-configmap.yaml
+++ b/bitnami/redis/templates/scripts-configmap.yaml
@@ -492,14 +492,18 @@ data:
 
     if ! failover_finished; then
         echo "I am the master pod and you are stopping me."
-        echo "Stopping client connections"
-        # Stopping client write connections to avoid data loss
-        run_redis_command CLIENT PAUSE "{{ mul (sub .Values.sentinel.terminationGracePeriodSeconds 10) 1000 }}" WRITE
         # Running failover command
+        echo "Running the FAILOVER command"
         run_redis_command FAILOVER TIMEOUT "{{ mul (sub .Values.sentinel.terminationGracePeriodSeconds 10) 1000 }}"
         # Waiting for the failover to finish
-        echo "Waiting for sentinel to run failover for up to {{ sub .Values.sentinel.terminationGracePeriodSeconds 10 }}s"
-        retry_while "failover_finished" "{{ sub .Values.sentinel.terminationGracePeriodSeconds 10 }}" 1
+        echo "Waiting for the failover to finish up to {{ sub .Values.sentinel.terminationGracePeriodSeconds 10 }}s"
+        if retry_while "failover_finished" "{{ sub .Values.sentinel.terminationGracePeriodSeconds 10 }}" 1
+            echo "Master has been successfuly failed over to a different pod."
+            exit 0
+        else
+            echo "Master failover failed"
+            exit 1
+        fi
     else
         exit 0
     fi
--- a/.vib/redis/vib-publish.json
+++ b/.vib/redis/vib-publish.json
@@ -26,7 +26,8 @@
         "target_platform": {
           "target_platform_id": "{VIB_ENV_TARGET_PLATFORM}",
           "size": {
-            "name": "S4"
+            "name": "S4",
+            "worker_nodes_instance_count": "1"
           }
         }
       },
--- a/.vib/redis/vib-verify.json
+++ b/.vib/redis/vib-verify.json
@@ -26,7 +26,8 @@
         "target_platform": {
           "target_platform_id": "{VIB_ENV_TARGET_PLATFORM}",
           "size": {
-            "name": "S4"
+            "name": "S4",
+            "worker_nodes_instance_count": "1"
           }
         }
       },
--- a/bitnami/redis/Chart.yaml
+++ b/bitnami/redis/Chart.yaml
@@ -22,4 +22,4 @@ maintainers:
 name: redis
 sources:
   - https://github.com/bitnami/containers/tree/main/bitnami/redis
-version: 17.3.6
+version: 17.3.7


armsnyder:
Hi @jotamartos 👋 

Isn't there still the possibility of dropping writes here? There looks to be a race, where sentinel may begin failover in sentinel-prestop.sh before we send the `CLIENT WAIT` in redis-prestop.sh.

How about moving this new code from redis-prestop.sh to sentinel-prestop.sh so we can guarantee order:

```yaml
  prestop-sentinel.sh: |
...
if ! failover_finished; then
        echo "I am the master pod and you are stopping me. Stopping client connections"
        # Stopping client write connections to avoid data loss
        run_redis_command CLIENT PAUSE "{{ mul (sub .Values.sentinel.terminationGracePeriodSeconds 10) 1000 }}" WRITE
        # if I am the master, issue a command to failover once and then wait for the failover to finish
        echo "I am the master pod and you are stopping me. Starting sentinel failover"
        # if I am the master, issue a command to failover once and then wait for the failover to finish
        run_sentinel_command failover "{{ .Values.sentinel.masterSet }}"
        if retry_while "failover_finished" "{{ sub .Values.sentinel.terminationGracePeriodSeconds 10 }}" 1; then
            echo "Master has been successfuly failed over to a different pod."
            exit 0
        else
            echo "Master failover failed"
            exit 1
        fi
    else
        exit 0
    fi
```

brettmorien:
Hi folks.

Our team (including @armsnyder) took a fork of this fix and ran it in our project with the same issue at load. For our particular case, this appears to have eliminated our reproduceable data loss issue during redeployments.

Hope to see this released soon!

fmulero (dismissed):
Great! The issue reported in #12598 is not reproducible
LGTM

bitnami/redis/templates/scripts-configmap.yaml
```diff
@@ -491,6 +491,10 @@ data:
     [[ -f "$REDIS_PASSWORD_FILE" ]] && export REDISCLI_AUTH="$(< "${REDIS_PASSWORD_FILE}")"
 
     if ! failover_finished; then
+        echo "I am the master pod and you are stopping me. Stopping client connections"
+        # Stopping client write connections to avoid data loss
+        run_redis_command CLIENT PAUSE "{{ mul (sub .Values.sentinel.terminationGracePeriodSeconds 10) 1000 }}" WRITE
+        # if I am the master, issue a command to failover once and then wait for the failover to finish
         echo "Waiting for sentinel to run failover for up to {{ sub .Values.sentinel.terminationGracePeriodSeconds 10 }}s"
         retry_while "failover_finished" "{{ sub .Values.sentinel.terminationGracePeriodSeconds 10 }}" 1
```
simon-baatz, line 499: While issuing a "client pause" prevents write losses until we notice the role switch here, this can/will take longer than necessary:

The "SENTINEL FAILOVER" command just assumes that the master is dead. This means that it does not actively reconfigure the master as part of the failover. It will promote one of the replicas to become a master and will then reconfigure the remaining replicas of the _former_ master. 

If sentinel is asked for the master address, it will return the new master as soon as it has verified that it switched its role to master. 

However, the former master (which we query in `failover_finished`) will change its role only by the regular checks sentinel makes. Sentinel will notice that this node is supposed to be a replica and reconfigure it (after waiting some additional time).

That's why the write loss window reported in  https://github.com/bitnami/charts/issues/12598 is so big (approx. 10 seconds).

To improve this, we could use the check used by the sentinel script to find out if the failover finished (that one asks the sentinel for the master). When we shut down the former master then, reconnecting clients will connect to the new master, which is what we want.

All in all however, I wonder if it is worth the hassle. Before #6080 we did not do it anything here and, thus, shut down the master as quickly as possible, creating a time window during which clients could not reach the master.  Now we introduce a time window during which only reads are allowed, but which is longer than necessary because of the polling. Which one is better probably depends on the application.

simon-baatz:
@armsnyder 

> Isn't there still the possibility of dropping writes here? There looks to be a race, where sentinel may begin failover in prestop-sentinel.sh before we send the `CLIENT WAIT` in prestop-redis.sh.

Yes and, actually, the race condition is even bigger. If clients are paused, but replication is not fully in sync yet, the sentinel may initiate a role switch before the replica caught up.

To be on the safe side, we would need to stop the clients and wait for the replicas to catch up before initiating the failover (see #12598), but that seems hard to do.

simon-baatz:
I looked "a bit" further into this:

## Long client write phases

I have seen two effects of having client writes paused for a long time:

1. If the client timeout is shorter than the pause time, paused writing clients will timeout, reconnect and,
   if a new master is already elected, connect to that master. Once connected to the new master, they
   can write data again.
    
   However, clients not writing at all will stay connected to the former master and not see updates until Sentinel reconfigures the former master (or Redis shuts down; whatever happens earlier). 
    
   So while writes aren't lost, clients may see stale data for some time (Note: Of course, a replica may
   always see old data because of the replication delay. But here, we don't see new data for the remainder of the client pause time).

2. Clients that are connected to a replica (for read load sharing) will reconnect during the failover and may end up 
   on the former master. The Sentinel client protocol specifies that such clients should check the role after a reconnect, but
   apparently at least the Python Redis client does not do that; it does not seem to care that the "replica" it connects to considers itself to be a "master".

   Such clients will see stale data as well.

## Solution proposed in this MR

Apparently, the "CLIENT PAUSE WRITE" command *inhibits a role change* by the sentinel. So, monitoring for a Redis role change does not work in this case. In all cases I have seen, it took the default 20s timeout to shut down Redis, e.g.

```
1:M 24 Oct 2022 10:26:26.466 # Connection with replica redis-node-2.redis-headless.default.svc.cluster.local:6379 lost.
1:M 24 Oct 2022 10:26:27.396 # Connection with replica redis-node-1.redis-headless.default.svc.cluster.local:6379 lost.
1:M 24 Oct 2022 10:26:46.563 # User requested shutdown...
```
   
Consequently, client writes are paused for 20 seconds. Read-only clients connecting to a master with a timeout of 5 seconds did see stale data for approx. 15 seconds.

## Alternative 1: Delay Redis Shutdown (based on Sentinel GET-MASTER-ADDR-BY-NAME)

To test if my improvement proposed in https://github.com/bitnami/charts/pull/13021#discussion_r1000698612 (monitoring the sentinel) helps, I copied the respective functions from `prestop-sentinel.sh` into `prestop-redis.sh` (renaming `failover_finished` into `sentinel_failover_finished`) and used:

```bash
    if ! failover_finished && ! sentinel_failover_finished; then
        echo "I am the master pod and you are stopping me. Stopping client connections"
        # Stopping client write connections to avoid data loss
        run_redis_command CLIENT PAUSE "{{ mul (add 2 (sub .Values.replica.terminationGracePeriodSeconds 10)) 1000 }}" WRITE
        # if I am the master, issue a command to failover once and then wait for the failover to finish
        echo "Waiting for sentinel to run failover for up to {{ sub .Values.replica.terminationGracePeriodSeconds 10 }}s"
        retry_while "sentinel_failover_finished" "{{ sub .Values.replica.terminationGracePeriodSeconds 10 }}" 1
    else
        exit 0
    fi
```

Now, Redis is shutting down much quicker:

```
1:M 24 Oct 2022 10:54:46.746 * Synchronization with replica redis-node-2.redis-headless.default.svc.cluster.local:6379 succeeded
1:M 24 Oct 2022 10:55:11.137 # Connection with replica redis-node-2.redis-headless.default.svc.cluster.local:6379 lost.
1:M 24 Oct 2022 10:55:12.089 # Connection with replica redis-node-1.redis-headless.default.svc.cluster.local:6379 lost.
1:M 24 Oct 2022 10:55:13.140 # User requested shutdown...
```

Read-only clients connecting to a master using a timeout of 5 seconds did not see stale data. Read-only clients connecting to a master using a timeout of 0.1 seconds did see stale data. Read-only clients connecting to a replica (using redis-py) sometimes did see stale data for 1-2 seconds.

## Alternative 2: Shutdown Redis immediately

Another approach is to shutdown the Redis instance as quickly as possible (i.e. remove `redis-prestop.sh`) but keep waiting in the sentinel pre-stop script for the failover to happen. The latter does not stop the pod unless we know that clients will connect to a different master (and the re-starting pod will start as a replica)

This seems to minimize seeing stale data:

As the duration of the "client pause write" phase (which is part of the Redis shutdown procedure) is usually very short, I could not reproduce the problem with stale data caused by short client timeouts (I used a 0.1s timeout).

Also, the problem in redis-py for replica connections does not appar, as clients can't connect to the "wrong" replica (which is our former master), as it shuts down very quickly. 

In contrast to alternative 1, clients of the master will get "connection refused" messages during the failover as the connection teardown ususally happens before a new master is elected.

## Summary

In summary, I think alternative 1 (pause client writes) is ok if the client timeout is bigger than the duration it takes to shut down Redis (which we can get down to a couple of seconds). If the client timeout is shorter, clients connecting to masters may see stale data. We might want to document that.

Clients connected to replicas might see stale data (however, one could argue that this is a client implementation problem as a client should verify the role when connecting.)

That being said, I prefer alternative 2: It is messier with respect to connection handling (clients will see connection refused errors), but I could not produce issues with stale data which is seconds old. Additionally, it is simpler.

(No solution solves the problem of write losses if the failover happens while replication is lagging. But that is another problem.)

brettmorien:
> ## Summary
> In summary, I think alternative 1 (pause client writes) is ok if the client timeout is bigger than the duration it takes to shut down Redis (which we can get down to a couple of seconds). If the client timeout is shorter, clients connecting to masters may see stale data. We might want to document that.
> 
> Clients connected to replicas might see stale data (however, one could argue that this is a client implementation problem as a client should verify the role when connecting.)
> 
> That being said, I prefer alternative 2: It is messier with respect to connection handling (clients will see connection refused errors), but I could not produce issues with stale data which is seconds old. Additionally, it is simpler.
> 
> (No solution solves the problem of write losses if the failover happens while replication is lagging. But that is another problem.)

For our case, either solution is workable. I tend to agree that the second option is safer in terms of data loss, and code should already be guarded against intermittent connection interruptions to Redis. If it becomes unavailable for a short time, except for the highest availability requirements, this is probably the most predictable source of downtime. And if you need near-100% availability, you're probably already mitigating well beyond stock k8s and this chart.

jotamartos:
Hi all, 

Thanks for all your comments. I'm going to take a look at them and update the PR with the new suggestions.

bitnami/redis/templates/scripts-configmap.yaml
```diff
@@ -491,6 +491,10 @@ data:
     [[ -f "$REDIS_PASSWORD_FILE" ]] && export REDISCLI_AUTH="$(< "${REDIS_PASSWORD_FILE}")"
 
     if ! failover_finished; then
+        echo "I am the master pod and you are stopping me. Stopping client connections"
+        # Stopping client write connections to avoid data loss
+        run_redis_command CLIENT PAUSE "{{ mul (sub .Values.sentinel.terminationGracePeriodSeconds 10) 1000 }}" WRITE
+        # if I am the master, issue a command to failover once and then wait for the failover to finish
```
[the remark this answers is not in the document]
jotamartos: Thanks for the suggestion. I just updated the code

jotamartos:
Hi again, 

After reviewing your comments, we decided that it's better to leave the changes as they are right now. This solution solved the original issue and does not require elaborate changes. Please find more information below:

- The first option requires bigger changes to make sure the client connections are stopped. Please note that we need to take into account if the sentinel server requires authentication or not so exporting the REDISCLI_AUTH env var will require changes in that case. 

- The second solution is easy to implement, and will work out of the box. If there is no master node, clients will not able to write. However, it will bring back [this error](https://github.com/bitnami/charts/pull/6080). We would not like to interrupt the service and waiting for the failover command to finish is a good idea in the master node. Please note that you can specify a different hook by using the replica.lifecycleHooks parameter in your local deployments. 

https://github.com/bitnami/charts/blob/main/bitnami/redis/templates/sentinel/statefulset.yaml#L99

I just made some minor changes in the code would like to merge these changes as soon as possible to be able to release a new version of the chart with these upgrades. I'll wait for your feedback in case there is something we are missing but please note that you can create a separate PR in the future with your suggestions in case you think this solution can be improved (for example if you want to try the first solution). 

Thanks

simon-baatz:
Hi @jotamartos 

> After reviewing your comments, we decided that it's better to leave the changes as they are right now.

Perhaps my comments were not clear enough, but what I was trying to say is that the current approach does not work. 

That is to say, the master never becomes a replica while waiting, the wait loop is always terminated by the timeout (which is 20 seconds with the default grace period of 30 seconds).  Have you seen instances where the wait actually ended due to a master to replica switch?

This means, depending on the timeout setting of clients, any client that does writes will hang for up to 20 seconds (and to me, this counts as a service interruption as well).

To illustrate my point, here is an example with a termination grace period of 180s:

Redis container (node 0) waits almost 3 minutes for the switch to occur (which does not happen), then shuts down:

```
1:M 27 Oct 2022 09:08:45.822 # Connection with replica redis-node-1.redis-headless.default.svc.cluster.local:6379 lost.
1:M 27 Oct 2022 09:08:46.750 # Connection with replica redis-node-2.redis-headless.default.svc.cluster.local:6379 lost.
1:signal-handler (1666861897) Received SIGTERM scheduling shutdown...
1:M 27 Oct 2022 09:11:37.928 # User requested shutdown...
1:M 27 Oct 2022 09:11:37.928 * Calling fsync() on the AOF file.
1:M 27 Oct 2022 09:11:37.928 * Saving the final RDB snapshot before exiting.
1:M 27 Oct 2022 09:11:37.949 * DB saved on disk 1:M 27 Oct 2022 09:11:37.949 # Redis is now ready to exit, bye bye...
```

Sentinel on node 2 (sentinel on node 0 shuts down earlier): Apparently,  the sentinels try to reconfigure the slave again and again, but do not succeed until a new redis container starts:
```
1:X 27 Oct 2022 09:08:46.759 * +slave slave redis-node-0.redis-headless.default.svc.cluster.local:6379 redis-node-0.redis-headless.default.svc.cluster.local 6379 @ mymaster redis-node-1.redis-headless.default.svc.cluster.local 6379
1:X 27 Oct 2022 09:08:46.814 * Sentinel new configuration saved on disk
1:X 27 Oct 2022 09:08:57.751 # +sdown sentinel 2a09ba7abbb41ee71e79087310d75f9809c3c815 redis-node-0.redis-headless.default.svc.cluster.local 26379 @ mymaster redis-node-1.redis-headless.default.svc.cluster.local 6379
1:X 27 Oct 2022 09:08:58.911 # +sdown slave redis-node-0.redis-headless.default.svc.cluster.local:6379 redis-node-0.redis-headless.default.svc.cluster.local 6379 @ mymaster redis-node-1.redis-headless.default.svc.cluster.local 6379
1:X 27 Oct 2022 09:09:02.018 # -sdown slave redis-node-0.redis-headless.default.svc.cluster.local:6379 redis-node-0.redis-headless.default.svc.cluster.local 6379 @ mymaster redis-node-1.redis-headless.default.svc.cluster.local 6379
1:X 27 Oct 2022 09:09:12.989 # +sdown slave redis-node-0.redis-headless.default.svc.cluster.local:6379 redis-node-0.redis-headless.default.svc.cluster.local 6379 @ mymaster redis-node-1.redis-headless.default.svc.cluster.local 6379
1:X 27 Oct 2022 09:09:17.103 # -sdown slave redis-node-0.redis-headless.default.svc.cluster.local:6379 redis-node-0.redis-headless.default.svc.cluster.local 6379 @ mymaster redis-node-1.redis-headless.default.svc.cluster.local 6379
1:X 27 Oct 2022 09:09:28.159 # +sdown slave redis-node-0.redis-headless.default.svc.cluster.local:6379 redis-node-0.redis-headless.default.svc.cluster.local 6379 @ mymaster redis-node-1.redis-headless.default.svc.cluster.local 6379
1:X 27 Oct 2022 09:09:32.275 # -sdown slave redis-node-0.redis-headless.default.svc.cluster.local:6379 redis-node-0.redis-headless.default.svc.cluster.local 6379 @ mymaster redis-node-1.redis-headless.default.svc.cluster.local 6379
1:X 27 Oct 2022 09:09:43.364 # +sdown slave redis-node-0.redis-headless.default.svc.cluster.local:6379 redis-node-0.redis-headless.default.svc.cluster.local 6379 @ mymaster redis-node-1.redis-headless.default.svc.cluster.local 6379
1:X 27 Oct 2022 09:09:47.356 # -sdown slave redis-node-0.redis-headless.default.svc.cluster.local:6379 redis-node-0.redis-headless.default.svc.cluster.local 6379 @ mymaster redis-node-1.redis-headless.default.svc.cluster.local 6379
1:X 27 Oct 2022 09:09:58.337 # +sdown slave redis-node-0.redis-headless.default.svc.cluster.local:6379 redis-node-0.redis-headless.default.svc.cluster.local 6379 @ mymaster redis-node-1.redis-headless.default.svc.cluster.local 6379
1:X 27 Oct 2022 09:10:02.503 # -sdown slave redis-node-0.redis-headless.default.svc.cluster.local:6379 redis-node-0.redis-headless.default.svc.cluster.local 6379 @ mymaster redis-node-1.redis-headless.default.svc.cluster.local 6379
1:X 27 Oct 2022 09:10:13.480 # +sdown slave redis-node-0.redis-headless.default.svc.cluster.local:6379 redis-node-0.redis-headless.default.svc.cluster.local 6379 @ mymaster redis-node-1.redis-headless.default.svc.cluster.local 6379
1:X 27 Oct 2022 09:10:17.597 # -sdown slave redis-node-0.redis-headless.default.svc.cluster.local:6379 redis-node-0.redis-headless.default.svc.cluster.local 6379 @ mymaster redis-node-1.redis-headless.default.svc.cluster.local 6379
1:X 27 Oct 2022 09:10:28.602 # +sdown slave redis-node-0.redis-headless.default.svc.cluster.local:6379 redis-node-0.redis-headless.default.svc.cluster.local 6379 @ mymaster redis-node-1.redis-headless.default.svc.cluster.local 6379
1:X 27 Oct 2022 09:10:32.701 # -sdown slave redis-node-0.redis-headless.default.svc.cluster.local:6379 redis-node-0.redis-headless.default.svc.cluster.local 6379 @ mymaster redis-node-1.redis-headless.default.svc.cluster.local 6379
1:X 27 Oct 2022 09:10:43.737 # +sdown slave redis-node-0.redis-headless.default.svc.cluster.local:6379 redis-node-0.redis-headless.default.svc.cluster.local 6379 @ mymaster redis-node-1.redis-headless.default.svc.cluster.local 6379
1:X 27 Oct 2022 09:10:47.808 # -sdown slave redis-node-0.redis-headless.default.svc.cluster.local:6379 redis-node-0.redis-headless.default.svc.cluster.local 6379 @ mymaster redis-node-1.redis-headless.default.svc.cluster.local 6379
1:X 27 Oct 2022 09:10:58.781 # +sdown slave redis-node-0.redis-headless.default.svc.cluster.local:6379 redis-node-0.redis-headless.default.svc.cluster.local 6379 @ mymaster redis-node-1.redis-headless.default.svc.cluster.local 6379
1:X 27 Oct 2022 09:11:02.952 # -sdown slave redis-node-0.redis-headless.default.svc.cluster.local:6379 redis-node-0.redis-headless.default.svc.cluster.local 6379 @ mymaster redis-node-1.redis-headless.default.svc.cluster.local 6379
1:X 27 Oct 2022 09:11:14.030 # +sdown slave redis-node-0.redis-headless.default.svc.cluster.local:6379 redis-node-0.redis-headless.default.svc.cluster.local 6379 @ mymaster redis-node-1.redis-headless.default.svc.cluster.local 6379
1:X 27 Oct 2022 09:11:18.035 # -sdown slave redis-node-0.redis-headless.default.svc.cluster.local:6379 redis-node-0.redis-headless.default.svc.cluster.local 6379 @ mymaster redis-node-1.redis-headless.default.svc.cluster.local 6379
1:X 27 Oct 2022 09:11:29.032 # +sdown slave redis-node-0.redis-headless.default.svc.cluster.local:6379 redis-node-0.redis-headless.default.svc.cluster.local 6379 @ mymaster redis-node-1.redis-headless.default.svc.cluster.local 6379
1:X 27 Oct 2022 09:11:33.160 # -sdown slave redis-node-0.redis-headless.default.svc.cluster.local:6379 redis-node-0.redis-headless.default.svc.cluster.local 6379 @ mymaster redis-node-1.redis-headless.default.svc.cluster.local 6379
1:X 27 Oct 2022 09:11:47.983 # +sdown slave redis-node-0.redis-headless.default.svc.cluster.local:6379 redis-node-0.redis-headless.default.svc.cluster.local 6379 @ mymaster redis-node-1.redis-headless.default.svc.cluster.local 6379
1:X 27 Oct 2022 09:12:19.525 * +reboot slave redis-node-0.redis-headless.default.svc.cluster.local:6379 redis-node-0.redis-headless.default.svc.cluster.local 6379 @ mymaster redis-node-1.redis-headless.default.svc.cluster.local 6379
1:X 27 Oct 2022 09:12:19.596 # -sdown slave redis-node-0.redis-headless.default.svc.cluster.local:6379 redis-node-0.redis-headless.default.svc.cluster.local 6379 @ mymaster redis-node-1.redis-headless.default.svc.cluster.local 6379
1:X 27 Oct 2022 09:12:21.434 # -sdown sentinel 2a09ba7abbb41ee71e79087310d75f9809c3c815 redis-node-0.redis-headless.default.svc.cluster.local 26379 @ mymaster redis-node-1.redis-headless.default.svc.cluster.local 6379
```

jotamartos:
Hi again @simon-baatz and thank you for your feedback.

I could reproduce the issue you mentioned and I investigated it a bit more. The problem was that the master node continued being the master node till the end and that's why you had a service interruption. With the [new changes](https://github.com/bitnami/charts/pull/13021/commits/b7af6eba84eee4b20d15088dfe691985c3fae5b9), the process is almost instantaneous and you can start using the new master without problems. Could you please give it a try and share your thoughts? 

Thanks

fmulero (dismissed):
LGTM

bitnami/redis/templates/scripts-configmap.yaml
```diff
@@ -491,6 +491,13 @@ data:
     [[ -f "$REDIS_PASSWORD_FILE" ]] && export REDISCLI_AUTH="$(< "${REDIS_PASSWORD_FILE}")"
 
     if ! failover_finished; then
+        echo "I am the master pod and you are stopping me."
+        echo "Stopping client connections"
+        # Stopping client write connections to avoid data loss
+        run_redis_command CLIENT PAUSE "{{ mul (sub .Values.sentinel.terminationGracePeriodSeconds 10) 1000 }}" WRITE
+        # Running failover command
+        run_redis_command FAILOVER TIMEOUT "{{ mul (sub .Values.sentinel.terminationGracePeriodSeconds 10) 1000 }}"
```
armsnyder, line 499: You do not need to run `CLIENT PAUSE WRITE` if you run `FAILOVER`. See docs: https://redis.io/commands/failover/

> The master will internally start a CLIENT PAUSE WRITE, which will pause incoming writes and prevent the accumulation of new data in the replication stream.

armsnyder (changes requested):

bitnami/redis/templates/scripts-configmap.yaml
jotamartos: Good catch! I just updated the code accordingly.

fmulero (approved):

ftagada:
There is a syntax error in bitnami/redis/templates/scripts-configmap.yaml, there is a missing "; then" after the "if retry_while "failover_finished" ...

It was introduced there: https://github.com/bitnami/charts/pull/13021/commits/aac3cd3af56728f4a47f9d9b026d0bdd6a935e14  (line 500)

PR #13054, opened 2022-10-20
[bitnami/common] kubernetes.io/tls-acme Ingress annotation triggers IngressTLS array

ian-llewellyn:
### Description of the change

`kubernetes.io/tls-acme` should be an acceptable Ingress annotation to trigger creation of the IngressTLS array. Many clusters run cert-manager with a default issuer configured for [fully automated TLS](https://cert-manager.io/docs/usage/ingress/#optional-configuration). This PR addresses that in the `_ingress.tpl` template of the bitnami/common Chart.

### Benefits

Clusters admins / operators no longer need to tie their deployments to specific certificate issuers. In many cases this is already defined in the cluster's cert-manager deployment (if present).

### Possible drawbacks

This change does not interfere with previous deployments so there are no known limitations. The only possible drawback would be if `kubernetes.io/tls-acme` is defined without a desire for a IngressTLS array - perhaps when set to `false`, however, this is a misuse of the annotation.

### Applicable issues

No current issues related to this fix - rather than open one, I'm submitting this PR.

### Checklist

Signed-off-by: Ian Cotter-Llewellyn <ian_llewellyn@hotmail.com>

### DIFF
--- a/bitnami/common/Chart.yaml
+++ b/bitnami/common/Chart.yaml
@@ -2,7 +2,7 @@ annotations:
   category: Infrastructure
 apiVersion: v2
 # Please make sure that version and appVersion are always the same.
-appVersion: 2.0.3
+appVersion: 2.0.4
 description: A Library Helm Chart for grouping common logic between bitnami charts. This chart is not deployable by itself.
 home: https://github.com/bitnami/charts/tree/main/bitnami/common
 icon: https://bitnami.com/downloads/logos/bitnami-mark.png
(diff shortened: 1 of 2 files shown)


ian-llewellyn:
Follow up question: Do the other Charts that depend on this `common` Chart have their `Chart.lock` files updated automatically. I'm hoping I'll be able to pin my NGINX deployment if/when this PR makes it through to `main`. Thanks.

marcosbc (approved):
LGTM! Thanks for the improvement!

PR #13114, opened 2022-10-24
[bitnami/kafka] Add support for JMX metrics extra Java args

### LINKED ISSUE #13071: [bitnami/kafka] JMX Exporter gets OOM killed on Cgroups v2 system
### Name and Version

bitnami/kafka 19.0.1

* Deploy chart with jmx exporter enabled and memory limit set.
* See its memory usage grow until dead.

The exporter limits its memory to the provided value.

### What do you see instead?

The exporter fails to read the limit and wants to use the whole system memory.

Visible in logs:
```
OpenJDK 64-Bit Server VM warning: Unable to open cgroup memory limit file /sys/fs/cgroup/memory/memory.limit_in_bytes (No such file or directory)
```

### Additional information

The image uses Java 8 which does support cgroups as a backport from Java 10. But Java 10 only supports cgroups v1. V2 support was added in Java 15 and backported to 11.0.16 (see [this issue](https://bugs.openjdk.org/browse/JDK-8285289) for details).

With Java updated, the container should respect limits on both Cgroups V1 and V2. But it's possible to work around this issue even without updating the image - by passing `-XX:MaxRAM` explicitly to JVM at startup.
The problem with this approach? The chart doesn't provide an option to customize the exporter's command. And it would need to be kept in sync between args and limits manually.

This is probably the source issue of https://github.com/prometheus/jmx_exporter/issues/690 too.

migruiz4:
Signed-off-by: Miguel Ruiz <miruiz@vmware.com>

### Description of the change

Add support for additional arguments when calling the JMX exporter.

Example:
```
#values.yaml
metrics:
  jmx:
    enabled: true
    extraJavaArgs:
      - -XX:MaxRAM=128000000
# Install command
helm install myrelease bitnami/kafka -f values.yaml --dry-run
...
        - name: jmx-exporter
          image: docker.io/bitnami/jmx-exporter:0.17.2-debian-11-r6
          securityContext:
            runAsNonRoot: true
            runAsUser: 1001
          command:
            - java
          args:
            - -XX:MaxRAM=128000000
            - -XX:+UnlockExperimentalVMOptions
            - -XX:+UseCGroupMemoryLimitForHeap
            - -XX:MaxRAMFraction=1
            - -XshowSettings:vm
            - -jar
            - jmx_prometheus_httpserver.jar
            - "5556"
            - /etc/jmx-kafka/jmx-kafka-prometheus.yml
```

### Applicable issues

  - fixes #13071

### Checklist


bitnami/kafka/templates/statefulset.yaml
```diff
@@ -472,6 +472,9 @@ spec:
           command:
             - java
           args:
+            {{- if .Values.metrics.jmx.extraJavaArgs }}
```
fmulero, line 475: LGTM but did you consider the same approach we apply for main containers where users are able to change whole `args` or `command`  values? Some lines above we are taking into account the diagnosticMode also:
```yaml
          ....
          {{- if .Values.%%MAIN_OBJECT_BLOCK%%.args }}
          args: {{- include "common.tplvalues.render" (dict "value" .Values.%%MAIN_OBJECT_BLOCK%%.args "context" $) | nindent 12 }}
           {{- else }}
          ....
```

fmulero (approved):
Thanks a lot @migruiz4! LGTM but could you have a look to my comment?

PR #13116, opened 2022-10-24
[bitnami/keycloak] Fix prometheus rule

cleverhu:
Signed-off-by: cleverhu <shouping.hu@daocloud.io>

### Description of the change
 Fix prometheus rule bugger. 

### Applicable issues

  - fixes #

### Checklist


migruiz4 (approved):

maxnitze:
Hey @cleverhu ,

could you elaborate why this solution is better than the old one? For me, this breaks my current setup where I define a chart with some basic rules and add additional ones in another chart, that "extends" this.

With this solution I have to overwrite and duplicate all my rules through all my Keycloak instances.

Unfortunately there is no description in this PR and no Issue linked. Was this discussed somewhere?

migruiz4:
I'm very sorry for the inconvenience @maxnitze.

This PR was accepted because in other charts we also use `{{ include "common.tplvalues.render" }}` instead of `{{ range }}` as it allows an extended usage of templating inside the values, my mistake was that I didn't notice this change wasn't backward-compatible with the previous syntax.

maxnitze:
Hey @migruiz4 ,

does it make sense to change it back again? Of course with the `{{ include "common.tplvalues.render" }}` option?

Maybe even supporting both a list and a dict?

migruiz4:
I'm sorry for the inconvenience, we should not have released this change in a minor version for the first instance.

Nonetheless, I don't think it makes sense to revert or support both list and dict formats, as the new method suits better the coding style used in the catalog.

Adapting the value `metrics.prometheusRule.groups` should be as simple as:
```diff
groups:
- rule_group_1:
+ - name: rule_group_1
    rules:
      ...
- rule_group_2:
+ - name: rule_group_2
    rules:
      ...
...
```

maxnitze:
I know how to adapt this for a single chart. This is not the issue.

My issue is, that I have a "base chart" based on this one from the bitnami repo. That base chart adds a lot of configuration that is then shared by about 10 other charts for the specific instances I'm currently running.

The configuration in my "base chart" contains a default set of prometheus rules, that are shared by all the instances (12 currently). Some instances have additional instance-specific alerts though.

In the former setup they were added as a new group and I was done. In the new setup I have to copy all rules from the base chart and then add my group with the additional one or two alerts.

For visualization here's what my setup basically looks like:

```
~ $ ls -alh ./keycloak
total 20K
drwxr-xr-x ... .
drwxr-xr-x ... ..
drwxr-xr-x ... base
drwxr-xr-x ... instance1
drwxr-xr-x ... instance2
...
~ $ cat ./keycloak/base/Chart.yaml
...
dependencies:
  - name: keycloak
    version: 10.1.6
~ $ cat ./keycloak/base/values.yaml
  metrics:
    ...
    prometheusRule:
      enabled: true
      groups:
        default:
          ...
        security:
          ...
~ $ cat ./keycloak/instance1/Chart.yaml
...
dependencies:
  - name: base
    version: ...
    repository: file://../base
~ $ cat ./keycloak/instance1/values.yaml
  metrics:
    ...
    prometheusRule:
      enabled: true
      groups:
        instance1:
          ...
```

PR #13117, opened 2022-10-24
[bitnami/kong] Removes check on service.type when custom service.clusterIP is provided 

davidchriqui:
**Description of the change**

This change removes the check on service.type in order to set provided service.clusterIP for any service type.

This is required in the case you've set service.type to "LoadBalancer" but you need to set a pre-defined clusterIP to the service (required if used in conjunction with hostAliases for instance)


jotamartos (dismissed):
Changes LGTM!  We already have the same configuration in other solutions like RabbitMQ

PR #13056, opened 2022-10-20
[bitnami/rabbitmq]fix secure password env hardcode

cleverhu:
fix secure password env hardcode.
Fixes: https://github.com/bitnami/charts/issues/12924
### Applicable issues

  - fixes https://github.com/bitnami/charts/issues/12924

### Checklist

- [x] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/). This is *not necessary* when the changes only affect README.md files.
- [x] Variables are documented in the values.yaml and added to the `README.md` using [readme-generator-for-helm](https://github.com/bitnami-labs/readme-generator-for-helm)
- [x] Title of the pull request follows this pattern [bitnami/<name_of_the_chart>] Descriptive title
- [x] All commits signed off and in agreement of [Developer Certificate of Origin (DCO)](https://github.com/bitnami/charts/blob/main/CONTRIBUTING.md#sign-your-work)

PR #13141, opened 2022-10-25
[bitnami/rabbitmq] Configure password correctly when RABBITMQ_SECURE_PASSWORD is not enabled

jotamartos:
### Description of the change

If you disable the securePassword parameter (RABBITMQ_SECURE_PASSWORD env var) the password is set to CHANGEME due to it's hardcoded in the configuration file.  

### Benefits

The password generated in the secret will be always used to configure the application and not the CHANGEME one

### Possible drawbacks

This behavior differs from the one in the containers that configures the password during the initialization 

https://github.com/bitnami/containers/blob/main/bitnami/rabbitmq/3.11/debian-11/rootfs/opt/bitnami/scripts/librabbitmq.sh#L375
https://github.com/bitnami/containers/blob/main/bitnami/rabbitmq/3.11/debian-11/rootfs/opt/bitnami/scripts/librabbitmq.sh#L822

### Applicable issues

  - Related to this other PR: https://github.com/bitnami/charts/pull/13056

### Checklist


corico44 (approved):
LGTM!

joancafom (approved):

siryur:
Hi 👋 
I'm afraid I have to disagree with this approach. If one has the "defintions_load" in place, that might mean that they don't want to use the plain password and this logic is currently saying:

```
If .Values.secureboot == true OR if .Values.auth.password == EMPTY:
    SECURE_BOOT = "yes"
else
    SECURE_BOOT = "no"
```

Why should I have a secure boot if I have my password EMPTY (like the default behavior)? I can overcome this by passing a random value to the `auth.password` - which makes no sense.

PR #13149, opened 2022-10-25
add namespaces to clusterrole for grafana-operator

cheddarwhizzy:
Fixes `scanAllNamespaces: true`
```
failed to list *v1.Namespace: namespaces is forbidden: User "system:serviceaccount:grafana:grafana-operator" cannot list resource "namespaces" in API group "" at the cluster scope
```

### DIFF
--- a/bitnami/grafana-operator/templates/rbac.yaml
+++ b/bitnami/grafana-operator/templates/rbac.yaml
@@ -112,6 +112,7 @@ rules:
       - ""
     resources:
       - pods
+      - namespaces
     verbs:
       - get
       - list
--- a/bitnami/grafana-operator/Chart.yaml
+++ b/bitnami/grafana-operator/Chart.yaml
@@ -22,4 +22,4 @@ name: grafana-operator
 sources:
   - https://github.com/grafana-operator/grafana-operator
   - https://github.com/bitnami/containers/tree/main/bitnami/grafana-operator
-version: 2.7.8
+version: 2.7.9


bitnami/grafana-operator/templates/rbac.yaml
```diff
@@ -112,6 +112,7 @@ rules:
       - ""
     resources:
       - pods
+      - namespaces
```
jotamartos, line 115: Taking a look at [the grafana-operator's configuration](https://github.com/grafana-operator/grafana-operator/blob/955daa6abb9b6efe85d921a783eeffd51853fae4/deploy/manifests/v4.7.1/rbac.yaml), I can see this is not necessary. 

Please take a look at [this README file](https://github.com/grafana-operator/grafana-operator/tree/955daa6abb9b6efe85d921a783eeffd51853fae4/deploy/cluster_roles) for more information about how to solve this.

jotamartos (changes requested):

bitnami/grafana-operator/templates/rbac.yaml
Yaytay: I think that the Grafana-operator documentation is insufficient.
If the operator is configured with a dashboardNamespaceSelector then it needs to be able to read and monitor namespace objects, which the existing RBAC does not provide.
I think this is _only_ required with the dashboardNamespaceSelector, so this PR isn't correct either, the inclusion of the namespaces resources should be conditional on the dashboardNamespaceSelector being specified.

Would you like a new PR?

Issue #11847, opened 2022-08-19
[bitnami/elasticserach] TLS keys/certificates are re-generated even on empty helm upgrade


psiroky:
### Name and Version

bitnami/elasticserach 19.1.11

### What steps will reproduce the bug?

1. I am running a local `kind` k8s "cluster" (v1.24.0)
2. I installed elasticsearch with
```
helm install -n elastic elasticsearch bitnami/elasticsearch --version="19.1.11" \
    --set security.enabled=true \
    --set security.tls.autoGenerated=true
```
Everything up to this point is OK.

The generated secret for the coordinating nodes looks like this:
```yaml
apiVersion: v1
data:
[...]

PR #13150, opened 2022-10-26
[bitnami/common] Add new function 'common.secrets.lookup'

migruiz4:
Signed-off-by: Miguel Ruiz <miruiz@vmware.com>

### Description of the change

This PR adds a new common function, that depends on the Helm `lookup` function.

This function searches for a key in an existing secret, otherwise sets a default value.

### Benefits

This function can be used to prevent regenerating secrets when upgrading, for example, for autogenerated certs.

Currently, we do not have a method to lookup for a value in an existing secret, and it could be used, for example, in cases like this:
```diff
-  ca.crt: {{ $ca.Cert | b64enc | quote }}
-  tls.crt: {{ $crt.Cert | b64enc | quote }}
-  tls.key: {{ $crt.Key | b64enc | quote }}
+  ca.crt: {{ include "common.secrets.lookup" (dict "secret" $secretName "key" "ca.crt" "defaultValue" $ca.Cert "context" $) }}
+  tls.crt: {{ include "common.secrets.lookup" (dict "secret" $secretName "key" "tls.crt" "defaultValue" $crt.Cert "context" $) }}
+  tls.key: {{ include "common.secrets.lookup" (dict "secret" $secretName "key" "tls.key" "defaultValue" $crt.Key "context" $) }}
```

### Possible drawbacks

This function is similar to "common.secrets.passwords.manage", but much more simple. 
If configured with some specific parameters, its functionality could overlap and we could end up with both methods used for the same purpose in some cases.

### Applicable issues

  Would help fixing #11847

### Checklist

### DIFF
--- a/bitnami/common/templates/_secrets.tpl
+++ b/bitnami/common/templates/_secrets.tpl
@@ -123,6 +123,32 @@ The order in which this function returns a secret password:
 {{- end -}}
 
 {{/*
+Reuses the value from an existing secret, otherwise sets its value to a default value.
+
+Usage:
+{{ include "common.secrets.lookup" (dict "secret" "secret-name" "key" "keyName" "defaultValue" .Values.myValue "context" $) }}
+
+Params:
+  - secret - String - Required - Name of the 'Secret' resource where the password is stored.
+  - key - String - Required - Name of the key in the secret.
+  - defaultValue - String - Required - The path to the validating value in the values.yaml, e.g: "mysql.password". Will pick first parameter with a defined value.
+  - context - Context - Required - Parent context.
+
+*/}}
+{{- define "common.secrets.lookup" -}}
+
+{{- $value := "" -}}
+{{- $defaultValue := required "\n'common.secrets.lookup': Argument 'defaultValue' missing or empty" .defaultValue -}}
+{{- $secretData := (lookup "v1" "Secret" $.context.Release.Namespace .secret).data -}}
+{{- if and $secretData (hasKey $secretData .key) -}}
+  {{- $value = index $secretData .key | quote -}}
+{{- else -}}
+  {{- $value = $defaultValue | toString | b64enc | quote -}}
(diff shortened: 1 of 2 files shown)


aoterolorenzo (approved):

bitnami/common/templates/_secrets.tpl
```diff
@@ -122,6 +122,32 @@ The order in which this function returns a secret password:
...
+*/}}
+{{- define "common.secrets.lookup" -}}
+
+{{- $value := "" -}}
+{{- $defaultValue := required "\n'common.secrets.lookup': Argument 'defaultValue' missing or empty" .defaultValue -}}
+{{- $secretData := (lookup "v1" "Secret" $.context.Release.Namespace .secret).data -}}
+{{- if and $secretData (hasKey $secretData .key) -}}
+  {{- $value = index $secretData .key | quote -}}
+{{- else -}}
+  {{- $value = $defaultValue | toString | b64enc | quote -}}
```
javsalgar, line 146: do we normally quote values in the helpers? Not sure if this could be problematic for certain cases where you need to add the password like, for example a connection string

migruiz4: I used "common.secrets.passwords.manage" as reference.

I'm sorry I don't understand the connection string example, as the output of this function will always be a base64 encoded.

Although, the suggestion made me think of a different use case.

If we add `| quote` inside the helper, it wouldn't be possible to run `| b64dec` if needed. For example:

`
include "common.secrets.lookup" (dict "secret" $secretName "key" "ca.crt" "defaultValue" $ca.Cert "context" $) | b64dec
`
Can't think of when would that be needed, but I would prefer not to limit the helper usage by forcing the `| quote`, so I'm removing it.

aoterolorenzo (approved):

aoterolorenzo (approved):

PR #13228, opened 2022-10-28
[bitnami/cert-manager] Add tests and VIB publishing

joancafom:
### Description of the change

The main objective of this PR is to publish our Bitnami CertManager Helm Chart using VMware Image Builder. In order to do that, several changes are included:

- Increasing the existing test coverage of the asset by adding Ginkgo tests.

### Benefits

### Possible drawbacks

### Applicable issues

NA

### Additional information

Tested in https://github.com/joancafom/charts/pull/93

### Checklist


.vib/cert-manager/goss/goss.yaml
```diff
@@ -0,0 +1,12 @@
+file:
```
FraPazGal, line 1: I think you should be able to check the `webhook.httpsPort` param with a `http` test:

```
$ curl -LIk https://cert-manager-webhook
HTTP/1.1 404 Not Found
Content-Type: text/plain; charset=utf-8
```

Additionally, is there any way to check the cert-manager/cert-manager-webhook CLIs?

FraPazGal (commented):
Just a comment before approving the PR @joancafom, please take a look when possible.

.vib/cert-manager/goss/goss.yaml
joancafom: > I think you should be able to check the webhook.httpsPort param with a http test:
The port is tested by the controller when a new Issuer CRD is created. In fact, changing the default port results in Ginkgo tests failures:

```
...
  Test Panicked
  In [BeforeEach] at: /.vib/cert-manager/ginkgo/integration_suite_test.go:71

  There was an error creating the Issuer resource: Internal error occurred: failed calling webhook "webhook.cert-manager.io": failed to call webhook: Post "https://cert-manager-webhook.default.svc:443/mutate?timeout=10s": no service port 443 found for service "cert-manager-webhook"
```

So there is definitely a test that checks this connection.

> Additionally, is there any way to check the cert-manager/cert-manager-webhook CLIs?

I think the current approach is enough, I am following the [official notes from the developer](https://cert-manager.io/docs/installation/verify/#manual-verification) on how to assess whether the app was correctly installed.

FraPazGal: In that case, there is indeed no need for an additional test. As you are already testing it, please add `webhook.httpsPort=443` to the runtime params.

Regarding the need for test using the CLI, given that the chart uses them directly to configure the pods I agree we can avoid testing them using goss.

FraPazGal (approved):
LGTM!

PR #13271, opened 2022-10-31
[bitnami/kafka] removing default value updateStrategy.rollingUpdate

davidballano:

### Description of the change

When selecting `updateStrategy.type: OnDelete `  and applying the manifests kubernetes will give this error:
`The StatefulSet "kafka-cluster-b" is invalid: spec.updateStrategy.rollingUpdate: Invalid value: apps.RollingUpdateStatefulSetStrategy{Partition:0}: only allowed for updateStrategy 'RollingUpdate'`

`updateStrategy.rollingUpdate` is default to `{}` in values.yaml and there is no way to remove it or set it to something that works with OnDelete. Kubernetes API doesnt like to have  `updateStrategy.rollingUpdate: {}` since it seems to be a setting for type RollingUpgrade only.

### Benefits

Being able to set statefulset type OnDelete,

### Possible drawbacks

The setting on values.yaml is an empty dict `updateStrategy.rollingUpdate: {}`  so it is already set to what kubernetes default is. If there is anybody using this setting they probably overrided it in their values.yaml. so removing this default value shouldn't affect anything as far as I know.

### Applicable issues

I recently opened a question about rolling restart of kafka when using a Statefulset,  Brokers get restarted as soon as they become ready. I would like to be able to control the roll and restart each node once the cluster has passed some checks before it continues with the next one. The only way to be able to do this is to set the statefulset type OnDelete and orchestrate the restarts from outside kubernetes.

### Additional information
n/a
### Checklist


corico44 (approved):
LGTM!

PR #13319, opened 2022-11-02
[bitnami/zookeeper] removing updateStrategy.rollingUpdate: {} so statefultset can be set to

davidballano:
OnDelete

### Description of the change
When selecting updateStrategy.type: OnDelete and applying the manifests kubernetes will give this error:
`Error from server (Invalid): error when creating "manifests/x.kube.demonware.net/kafka-cluster-b/dw-kafka/charts/zookeeper/templates/statefulset.yaml": StatefulSet.apps "kafka-cluster-b-zookeeper" is invalid: spec.updateStrategy.rollingUpdate: Invalid value: apps.RollingUpdateStatefulSetStrategy{Partition:0}: only allowed for updateStrategy 'RollingUpdate'`

updateStrategy.rollingUpdate is default to {} in values.yaml and there is no way to remove it or set it to something that works with OnDelete. Kubernetes API doesnt like to have `updateStrategy.rollingUpdate: {}` since it seems to be a setting for type RollingUpgrade only.
### Benefits

Being able to set statefulset type OnDelete,

### Possible drawbacks

The setting on values.yaml is an empty dict `updateStrategy.rollingUpdate: {}`  so it is already set to what kubernetes default is. If there is anybody using this setting they probably overrided it in their values.yaml. so removing this default value shouldn't affect anything as far as I know.

### Applicable issues
I recently opened a question about rolling restart of kafka when using a Statefulset,  Brokers get restarted as soon as they become ready. Same happens for Zookeeper I would like to be able to control the roll and restart each zookeeper node once the cluster has passed some checks before it continues with the next one. The only way to be able to do this is to set the statefulset type OnDelete and orchestrate the restarts from outside kubernetes.

### Additional information
n/a

### Checklist


marcosbc:
Hi @davidballano, have you tried setting `updateStrategy.rollingUpdate=null`? In principle it should allow you to remove the key, as documented here: https://helm.sh/docs/chart_template_guide/values_files/#deleting-a-default-key

davidballano:
Hi @marcosbc, yes I did try. bellow is my values.yaml
```
zookeeper:
  replicaCount: 5
  updateStrategy:
    type: OnDelete
    rollingUpdate: null
```
but once I render the manifests I get this on the statefulset.yaml
```
  updateStrategy:
    rollingUpdate: {}
    type: OnDelete
```
when I try to apply the manifests this is what i get:
`The StatefulSet "kafka-cluster-b-zookeeper" is invalid: spec.updateStrategy.rollingUpdate: Invalid value: apps.RollingUpdateStatefulSetStrategy{Partition:0}: only allowed for updateStrategy 'RollingUpdate'`

davidballano:
Hi @marcosbc 

Could you try the same command but this time also changing zookeeper.updateStrategy.type=OnDelete, the problem only  is when the unpdate startategy is set Ondelete and updateStrategy.rollingUpdate is set.

davidballano:
Actually you are right, I was able to test this, ignore my last comment. I will check why I cant get it working on my setup.
thanks!

PR #10445, opened 2022-05-26
[bitnami/mongodb] enabled mdb all collectors

shnurok672:
### Description of the change

Fixes mdb exporter not showing other metrics except mongodb_up
As per https://github.com/percona/mongodb_exporter/blob/main/REFERENCE.md , we need to pass --collect-all argument to enable all available collectors. Also it closes(resolves) https://github.com/bitnami/charts/issues/10440 https://github.com/bitnami/charts/issues/10379 https://github.com/bitnami/charts/issues/10294

Additionally enabled compatible mode as suggested here https://github.com/percona/mongodb_exporter/issues/472

PR #13385, opened 2022-11-07
[bitnami/mongodb-sharded] Enable all collectors in mongodb_exporter

henrifroese:
Same as #10445, but for mongodb-sharded. It's currently only exporting few metrics, this will make it export all.


CeliaGMqrz (approved):

Issue #13378, opened 2022-11-07
[bitnami/thanos]: no template "thanos.serviceAccountName" associated with template "gotpl"


willhughes-au:
### Name and Version

bitnami/thanos 11.5.7

### What steps will reproduce the bug?

Create a thanos deployment that works with 11.5.5. 

Upgrade to 11.5.7 and we get
```Error: template: thanos/templates/serviceaccount.yaml:5:11: executing "thanos/templates/serviceaccount.yaml" at <include "thanos.serviceAccountName" .>: error calling include: template: no template "thanos.serviceAccountName" associated with template "gotpl"```

### What do you see instead?

```Error: template: thanos/templates/serviceaccount.yaml:5:11: executing "thanos/templates/serviceaccount.yaml" at <include "thanos.serviceAccountName" .>: error calling include: template: no template "thanos.serviceAccountName" associated with template "gotpl"```

### Additional information

This appears to be because of an error in PR https://github.com/bitnami/charts/pull/13312 which changed  [serviceaccount.yaml at Line 5](https://github.com/bitnami/charts/pull/13312/files#diff-0f890d772223c137385fe2aab86b7d5950dd0075e98f62e833b411d3733b2ab4L5): 

```yaml 
   name: {{ include "thanos.serviceAccountName" . }}
```

This appears to be a typo, and should be referencing `thanos.serviceAccount.name` as defined in [_helpers.tpl at Line 331](https://github.com/bitnami/charts/blob/main/bitnami/thanos/templates/_helpers.tpl#L331)

calvinbui:
@corico44 i'm facing the same problem from your change in https://github.com/bitnami/charts/pull/13312

javsalgar:
Hi! 

Thank you so much for reporting! I will notify it so we can revert it.

calvinbui:
> Hello @calvinbui,
> 
> Using your values.yaml that you mentioned in the previous issue:

That was not me.

> You have modified something different from what was specified in the previous issue (https://github.com/bitnami/charts/issues/12954)?

As @willhughes-au mentioned:

> This appears to be a typo, and should be referencing thanos.serviceAccount.name as defined in [_helpers.tpl at Line 331](https://github.com/bitnami/charts/blob/main/bitnami/thanos/templates/_helpers.tpl#L331)

corico44:
Sorry for the confusion @calvinbui. The changes have been applied and it should now work correctly. 
Thank you very much for the contribution.

PR #13406, opened 2022-11-08
[bitnami/thanos] Fix serviceAccount name logic for Thanos

corico44:
### Description of the change

The template/serviceaccount file sets its name value using the _helpers.tlp file just like the other assets and not the common fullname value.

### Benefits

The format is the same as the other assets and the value is really being set with the value indicated.

### Additional information

- Fixes https://github.com/bitnami/charts/issues/13378

### Checklist

### DIFF
--- a/bitnami/thanos/templates/serviceaccount.yaml
+++ b/bitnami/thanos/templates/serviceaccount.yaml
@@ -2,7 +2,7 @@
 apiVersion: v1
 kind: ServiceAccount
 metadata:
-  name: {{ include "thanos.serviceAccount.name" . }}
+  name: {{ include "thanos.serviceAccountName" . }}
   namespace: {{ .Release.Namespace | quote }}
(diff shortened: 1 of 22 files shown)


bitnami/thanos/templates/serviceaccount.yaml
```diff
@@ -2,7 +2,7 @@
 apiVersion: v1
 kind: ServiceAccount
 metadata:
-  name: {{ include "thanos.serviceAccountName" . }}
+  name: {{ include "thanos.serviceAccount.name" . }}
```
jotamartos, line 5: Why don't we follow the same approach we have in others here?

https://github.com/bitnami/charts/blob/main/bitnami/wordpress/templates/_helpers.tpl#L50

corico44: Done!

jotamartos (approved):

PR #13414, opened 2022-11-08
[bitnami/fluentd] Make servicemonitor URL/Path configurable

### LINKED ISSUE #13388: [bitnami/fluentd] Make servicemonitor URL/Path configurable
### Name and Version

bitnami/fluentd

GIves the user the ability to to scrape aggregated_metrics via Prometheus

Currently, the metrics path is hard coded and would like to make it configurable in servicemonitor.yaml

Hardcoded
`spec:endpoints:path: /metrics`

Please make path configurable

Overriding the fluentd.conf files /aggregated_metrics path with /metrics

cmbaatz:
### Description of the change
Adds logic that allows a user to override both the port and path used in the prometheus ServiceMonitor. <br/>
This change allows for the port & path configuration for the ServiceMonitor to be configured in two different ways.
1. Via dedicated values<br/>
    - `.Values.metrics.service.port`: Already defines the port for the service.  Reusing this value prevents a chart user for accidentally using two separate values.
    -  `.Values.metrics.serviceMonitor.path`: Defines the path that prometheus should scrape.  
2. Via legacy annotations<br/>
    Supports the legacy non-operator prometheus deployment in which these values would typically be set via annotations.  This will aid legacy users if they transition over to the new operator prometheus.<br/>
    - .Values.metrics.service.annotations.prometheus.io/port 
    - .Values.metrics.service.annotations.prometheus.io/path

### Benefits
Allows a user to change both the port and path used by prometheus.  

### Possible drawbacks
N/A

### Applicable issues
fixes #13388 

### Additional information
N/A

### Checklist


bitnami/fluentd/templates/servicemonitor.yaml
```diff
@@ -26,8 +26,8 @@ spec:
       {{- include "common.tplvalues.render" (dict "value" .Values.metrics.serviceMonitor.selector "context" $) | nindent 6 }}
       {{- end }}
   endpoints:
-    - port: tcp-metrics
-      path: "/metrics"
+    - port: {{ default "tcp-metrics" (default (index .Values.metrics.service.annotations "prometheus.io/port") .Values.metrics.service.port) | quote}}
```
javsalgar, line 29: is it necessary to override the tcp-metrics string by `metrics.service.port`? In principle this should be the same, right?

cmbaatz: Fixed . thanks for the suggestion

bitnami/fluentd/templates/servicemonitor.yaml
```diff
@@ -26,8 +26,8 @@ spec:
       {{- include "common.tplvalues.render" (dict "value" .Values.metrics.serviceMonitor.selector "context" $) | nindent 6 }}
       {{- end }}
   endpoints:
-    - port: tcp-metrics
-      path: "/metrics"
+    - port: "tcp-metrics"
+      path: {{ default "/metrics" (default (index .Values.metrics.service.annotations "prometheus.io/path") .Values.metrics.serviceMonitor.path) | quote}}
```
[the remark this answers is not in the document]
cmbaatz: @javsalgar  actually after switching up to coalesce I realized that this wont work due to the fact I'm setting /metrics as the default value for .Values.metrics.serviceMonitor.path in values.yaml.  Therefore I switched by to using default, but now it only one statement first checking if the annotation is there and then defaulting to what's defined specified value.

cmbaatz:
@javsalgar or @CeliaGMqrz is it possible to move this pull request forward?

CeliaGMqrz (dismissed):
Sorry for the delay, LGTM! Thanks

CeliaGMqrz (approved):

PR #13465, opened 2022-11-10
[bitnami/ejbca] Fix EJBCA Cypress Test

mdhont:

### Description of the change

Currently the EJBCA Cypress test is unstable. The p12 is not always downloaded due to a bug in [Cypress](https://github.com/cypress-io/cypress/issues/14857). A workaround for this bug is to avoid checking for the file, and continue with the test after a few seconds. As the test later checks if the certificate exists, we can safely assume it has been possible to create a certificate.

### Applicable issues

  - fixes #


aoterolorenzo (approved):
Seems pretty straight forward if there's a subsequent test point already checking the cert directly or indirectly. LGTM

PR #13489, opened 2022-11-11
[bitnami/external-dns] Add support for google provider batch change size

nisc-acooper:
### Description of the change

Adding support to set the batch change size for the google provider.  This is undocumented [1] but has been confirmed to work with the latest bitnami docker image of external-dns.  

[1] https://github.com/kubernetes-sigs/external-dns/blob/master/provider/google/google.go#L108

### Benefits

This parameter was needed when we upgraded to the new txt registry.  Our existing dns record count was so high that external-dns would never succeed with adding all new registry entries without over-ridding this setting to a lower value.  This parameter is already defined in the helm chart for the aws provider.

### Possible drawbacks

I used the same default of 1000 to match what is set in the helm chart for the aws provider.  If there is a use case (I can not think of one) for this to be higher than 1000, the default of the values.yaml could be changed.

### Checklist


bitnami/external-dns/Chart.yaml
```diff
@@ -24,4 +24,4 @@ sources:
   - https://github.com/kubernetes-sigs/external-dns
   - https://github.com/bitnami/containers/tree/main/bitnami/external-dns
   - https://github.com/kubernetes-sigs/external-dns
-version: 6.11.3
+version: 6.11.4
```
marcosbc, line 27: Since this is adding a new feature, we should increase the minor instead (e.g. to `6.12.0`). Could you change that?

marcosbc (changes requested):

marcosbc:
Once fixed, could you add a separate comment? That way our support system will be properly able to track this.

bitnami/external-dns/Chart.yaml
carrodher, line 27:
```suggestion
version: 6.12.0
```

nisc-acooper:
I have approved and signed off on the change

marcosbc (approved):
LGTM, thanks!

PR #13504, opened 2022-11-14
[bitnami/redis] Allow auth.existingSecret to be templated

### LINKED ISSUE #13503: [bitnami/redis] auth.existingSecret cannot be templated
### Name and Version

bitnami/redis 17.3.9

I would like to create a secret with a password for Redis that is named after my Helm release like this:

```yaml
apiVersion: v1
kind: Secret
metadata:
  name: {{ include "my-chart.fullname" . }}-redis-auth
```

So, I would like to reference the secret like this in `values.yaml`:

```yaml
redis:
  auth:
    existingSecret: '{{ include "my-chart.fullname" . }}-auth'
```

However, the chart currently doesn't allow this value to be templated, so it'll raise an error about the secret name being invalid.

Allow `existingSecret` to be templated using `tpl`.

I currently work around this by specifying the `existingSecret` value on the Helm commandline using `--set`, but I really want to be able to do it in `values.yaml` instead.

OvervCW:
### Description of the change

The `auth.existingSecret` parameter can now include Helm templating.

### Benefits

This allows users of the chart to derive the name of the secret from other variables, like the release name.

### Possible drawbacks

It could cause backwards incompatibility if existing users used characters in their secret name that could be misinterpreted as Helm templating syntax, but I don't think there are any such characters that would have previously referred to a valid secret name.

### Applicable issues

- Fixes #13503

### Additional information

N/A

### Checklist

### DIFF
--- a/bitnami/redis/Chart.yaml
+++ b/bitnami/redis/Chart.yaml
@@ -22,4 +22,4 @@ maintainers:
 name: redis
 sources:
   - https://github.com/bitnami/containers/tree/main/bitnami/redis
-version: 17.3.10
+version: 17.3.11
--- a/bitnami/redis/templates/_helpers.tpl
+++ b/bitnami/redis/templates/_helpers.tpl
@@ -193,7 +193,7 @@ Get the password secret.
 */}}
 {{- define "redis.secretName" -}}
 {{- if .Values.auth.existingSecret -}}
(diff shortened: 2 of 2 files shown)


rafariossaa (approved):
Thanks for changing this.


LGTM

PR #13520, opened 2022-11-15
[bitnami/consul] fix apiversion hardcode for statefulset

my-git9:
[bitnami/consul] fix apiversion hardcode for statefulset


jotamartos (approved):
Thanks for your contribution!

PR #13527, opened 2022-11-15
[bitnami/kube-prometheus] Support prometheus-operator namespace allowlist. (#13525)

### LINKED ISSUE #13525: Support kube-prometheus co-existence with the prometheus-operator in 'openshift-monitoring'
### Name and Version

bitnami/kube-prometheus

If you deploy the kube-prometheus chart to an Openshift cluster already running RH's cluster-monitoring functionality - you end up with 2 prometheus-operators fighting of the same kind:Prometheus and you break the existing openshift-controlled prometheus-etc. instances.

Expose the existing operator's `-namespace` argument in the helm chart. Not specifying a value causes the existing behaviour (no argument is added, all namespaces are watched.), specifying a value injects the `-namespaces` argument to the operator, and causes it to watch the specific NS only (allowlist)

operator.args _could_ be used here, but then I have to keep all the the other currently-templated-args under my control - it is a hammer to crack a nut.
Better, have the chart support the upstream operator configurable explicitly.

mlashley:
### Description of the change

Support specifying the namespaces that the prometheus-operator watches.

### Benefits

Chart can co-exists with 'Openshift Monitoring' (or presumably multiple operator-versions in a single cluster)

### Possible drawbacks

None known 

### Applicable issues

  - fixes #13525

### Additional information

First time PR here - hope I got this right. I will be AFK for a couple of weeks from the end of the week - so I haven't died if we don't get this in before then.

### Checklist

### DIFF
--- a/bitnami/kube-prometheus/Chart.yaml
+++ b/bitnami/kube-prometheus/Chart.yaml
@@ -35,4 +35,4 @@ sources:
   - https://github.com/bitnami/containers/tree/main/bitnami/prometheus
   - https://github.com/bitnami/containers/tree/main/bitnami/alertmanager
   - https://github.com/prometheus-operator/kube-prometheus
-version: 8.1.13
+version: 8.2.0
--- a/bitnami/kube-prometheus/Chart.yaml
+++ b/bitnami/kube-prometheus/Chart.yaml
@@ -35,4 +35,4 @@ sources:
   - https://github.com/bitnami/containers/tree/main/bitnami/prometheus
   - https://github.com/bitnami/containers/tree/main/bitnami/alertmanager
   - https://github.com/prometheus-operator/kube-prometheus
-version: 8.1.12
+version: 8.1.13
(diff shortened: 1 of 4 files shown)


bitnami/kube-prometheus/Chart.yaml
```diff
@@ -35,4 +35,4 @@ sources:
   - https://github.com/bitnami/containers/tree/main/bitnami/prometheus
   - https://github.com/bitnami/containers/tree/main/bitnami/alertmanager
   - https://github.com/prometheus-operator/kube-prometheus
-version: 8.1.12
+version: 8.1.13
```
marcosbc, line 38: Since this is adding a new feature, could you increase the minor version instead? (e.g. to `8.2.0`)

marcosbc (changes requested):
Thanks for the contribution! Please have a look at my comments whenever you have a chance.

Once you have checked them (and fixed any issue if applicable), please reply back with an individual comment, for the response to be properly tracked by our support system.

bitnami/kube-prometheus/Chart.yaml
mlashley: Sure thing - I figured this was 50/50 bugfix/feature.

mlashley:
Thanks for the review @marcosbc  - all comments attended to.

marcosbc (approved):
LGTM! Thanks!

PR #13528, opened 2022-11-15
[bitnami/fluentd] fix apiversion hardcode for statefulset

my-git9:
[bitnami/fluentd] fix apiversion hardcode for statefulset


corico44 (approved):
LGTM

PR #13538, opened 2022-11-16
[bitnami/harbor]fix apiversion hardcode for statefulset

my-git9:
[bitnami/harbor]fix apiversion hardcode for statefulset


migruiz4 (approved):
Thank you very much for your contribution! Looks good to me!

PR #13545, opened 2022-11-16
[bitnami/redis-cluster] add support to disable persistence and add custom config

### LINKED ISSUE #4484: charts/redis-cluster disable persistence
Hi,

Is it possible to disable RDB in the redis-cluster chart? I tried to disable persistence but when I run `config get save` it still shows that it's enabled. Is it possible to disable it? I want to run redis cluster without any storage, only in memory since I'm using only for cache

Thanks

### LINKED ISSUE #13409: Make redis-cluster chart backwards compatible for Redis 6 (not only Redis 7)
### Name and Version

bitnami/redis-cluster 8.2.7

Currently the bitnami/redis-cluster chart version 8.2.7 only supports Redis 7 but I still need to support Redis 6. I can't just use the old chart version that was still supporting Redis 6 because that version (7.6.4) does not have additional features implemented in the meantime (e.g. `securityContext` on the `metrics` image).

Looking at the changes that have been implemented between the last version that supported Redis 6 (chart version 7.6.4, https://github.com/bitnami/charts/commit/d1b297eb06a01c57194e9434cd5b3d026038f7cf) and the current version, there seems to be only one incompatible change which is the default configuration used in https://github.com/anessi/bitnami-charts/blob/main/bitnami/redis-cluster/templates/configmap.yaml
This configuration is specific to Redis 7 and does not work for Redis 6. There is no way to override this configuration in the chart values file.

A possible approach could be to provide both default configurations and use the correct one depending on an optional property in the values file (e.g. `redis.majorVersion`). If this property is not provided, one could revert back to the `appVersion` in the `Chart.yaml` file.

Please let me know if you're willing to receive a PR for this or if this should be addressed differently.

It would be possible to back-port the changes done on chart version 8.2.7 to the 7.6.4 version, but I did not see any way how to do this (e.g. how to fix issues on older chart versions). Is this supported or not?

Another approach could be to allow to set the default config in the values file and not have it hardcoded in the chart.

anessi:
### Description of the change

* Adds support to disable persistence (uses emptyDir instead of a PVC). This is required if one wants to use Redis with in-memory data only. It makes sense to disable AOF and RDB persistence in the config in that case.
```
    # Disable AOF and RDB persistence as we keep everything in memory only, see https://redis.io/topics/persistence
    appendonly no
    # Disable RDB persistence, AOF persistence already disabled above.
    save ""
```

* Adds support to supply a custom Redis configuration, e.g. to support an older version of Redis. With the current setup the config is hard-coded and not backwards compatible, so the current chart version can't be used with an older version or Redis.

### Benefits

Adds additional features already covered, e.g. in the bitnami/redis chart. Does not change the default behavior and the changes are fully backwards compatible.

### Possible drawbacks

Additional complexity

### Applicable issues

  - fixes #4484
  - fixes #13409

### Checklist

### DIFF
--- a/bitnami/redis-cluster/Chart.yaml
+++ b/bitnami/redis-cluster/Chart.yaml
@@ -23,4 +23,4 @@ name: redis-cluster
 sources:
   - https://github.com/bitnami/containers/tree/main/bitnami/redis
   - http://redis.io/
-version: 8.2.7
+version: 8.3.0
--- a/bitnami/redis-cluster/README.md
+++ b/bitnami/redis-cluster/README.md
@@ -145,6 +145,7 @@ The command removes all the Kubernetes components associated with the chart and 
 | `service.externalTrafficPolicy`                         | Service external traffic policy                                                                                                                     | `Cluster`               |
 | `service.sessionAffinity`                               | Session Affinity for Kubernetes service, can be "None" or "ClientIP"                                                                                | `None`                  |
 | `service.sessionAffinityConfig`                         | Additional settings for the sessionAffinity                                                                                                         | `{}`                    |
+| `persistence.enabled`                                   | Enable persistence on Redis&reg;                                                                                                                    | `true`                  |
 | `persistence.path`                                      | Path to mount the volume at, to use other images Redis&reg; images.                                                                                 | `/bitnami/redis/data`   |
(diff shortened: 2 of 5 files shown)


marcosbc (approved):

PR #13565, opened 2022-11-17
[bitnami/nats] fix apiversion hardcode for statefulset

my-git9:
[bitnami/nats] fix apiversion hardcode for statefulset


CeliaGMqrz (approved):
LGTM! Thanks

PR #13574, opened 2022-11-17
zookeeper: Don't re-initialize keystores after success

kellyma2:
Init containers must be idempotent because they may get re-run in some cases even if the pod itself is not being restarted. For example, if the underlying node garbage collects the associated docker container.

In the current incarnation of the init script we will fail if the init container is re-run resulting in a state where the init container is continually flapping.  To avoid this, we can add a initialization check file and short circuit the initialization process if it is present.


kellyma2:
Thanks. I'll update that as well.

bitnami/zookeeper/templates/scripts-configmap.yaml
```diff
@@ -31,6 +34,7 @@ data:
               -keystore "/opt/bitnami/zookeeper/config/certs/client/zookeeper.truststore.jks" \
               -storepass "${ZOO_TLS_CLIENT_TRUSTSTORE_PASSWORD}" \
               -noprompt
+        echo "1" > /opt/bitnami/zookeeper/config/certs/client/initialized
```
jotamartos, line 37: I'd create a hidden file without any content, this way the file won't be listed by default so it'll create less noise. What do you think?

```
touch /opt/bitnami/zookeeper/config/certs/client/.initialized
```

kellyma2: Done

bitnami/zookeeper/Chart.yaml
```diff
@@ -21,4 +21,4 @@ name: zookeeper
 sources:
   - https://github.com/bitnami/containers/tree/main/bitnami/zookeeper
   - https://zookeeper.apache.org/
-version: 11.0.0
+version: 11.0.1
```
jotamartos, line 24: Version 11.0.1 is already published, please bump the version to 11.0.2

jotamartos (commented):
Thanks for applying the suggestions. I didn't receive any notification and that's why I didn't review this before. Please apply this final change before the PR can be closed.

bitnami/zookeeper/Chart.yaml
jotamartos: I already pushed this change to  your repo.

jotamartos (approved):

PR #13576, opened 2022-11-18
[bitnami/jupyterhub] fix apiversion hardcode for NetworkPolicy

my-git9:
[bitnami/jupyterhub] fix apiversion hardcode for NetworkPolicy


corico44 (approved):
LGTM

PR #13584, opened 2022-11-18
[bitnami/mongodb-sharded] Evaluate *.persistence.subPath as a template

cferns08:
### Description of the change

Evaluate *.persistence.subPath value as a template.

### Benefits

Ability to use dynamic subPath for volume mounts through template evaluation.

### Checklist


carrodher:

CeliaGMqrz:
Hi @cferns08 

Thanks for your contribution, but I think this could be a breaking change if the value is already set in the values. Could you please give me a usage example of this change? Also, I think this could be a new minor instead of a patch.

cferns08:
Hi @CeliaGMqrz 

I fixed some things my changes had wrong.

In regards to your comment:

> I think this could be a breaking change if the value is already set in the values

Does _common.tplvalues.render_ not work with literals not containing a template?

> Could you please give me a usage example of this change?

The use case is similar to subPathExpr but I thought this would be more flexible (also the commands run from the init containers make it more difficult). For example:

`shardsvr.persistence.subPath: {{ printf "data-%d" $i }}`

> I think this could be a new minor instead of a patch.

OK, do I increment the minor instead then?

CeliaGMqrz:
Hi @cferns08,

Yes, it's clearer now. Great.

> OK, do I increment the minor instead then?

Yes, please. I think this is a new feature so it should be minor.

Thanks.

CeliaGMqrz (approved):
LGTM! Thanks for your contribution!

PR #13592, opened 2022-11-18
[bitnami/kafka] adding back rollingUpdate default setting

davidballano:
helm bug and there is a workaround.

### Description of the change

Fews week ago created a [Kafka PR](https://github.com/bitnami/charts/pull/13271) that got merged regarding rollingUpdate default setting on Kafka chart.
after digging into the same behavior trying to set the Statefulset type OnDelete on Zookeeper  chart  [zookeper PR](https://github.com/bitnami/charts/pull/13319) found this was a bug in Helm chart. not a limitation of the chart/s. so i am rolling back the change. 

### Benefits

leave everything as it was.

### Possible drawbacks

 there are no drawbacks. the setting is added back as it was before.

### Applicable issues

n/a

### Additional information

If you want to set Kafka statefulset `Ondelete` you will need to delete `rollingUpdate:{}`  key. if you are using this chart as a dependency of a parent chart you will need to follow a workaround for now:
[bug/workaround](https://github.com/helm/helm/issues/9136#issuecomment-1205728157)

### Checklist


aoterolorenzo:
Hi @davidballano,

I'm afraid I may be missing something: I understand that the error mentioned in [zookeper PR](https://github.com/bitnami/charts/pull/13319) is provoked for having a default setting `rollingUpdate: {}`, which is what you are tying to add in this PR. Shouldn't be left as it is right now to avoid issues when changing the `updateStrategy.type`?

davidballano:
sorry, let me explain. At first when I tried to change `updateStrategy.type` to Zookeeper and Kafka charts it didn't work becasue like you mentioned the default `rollingUpdate: {} ` was in both charts.  I did try setting the key to `null` in my values.yaml but it didnt work for me at first and I though it was a limitation of the chart. So I made both PR one for kafka and another for Zookeeper removing the `rollingUpdate: {}` key. Kafka PR got merged Zookeeper didn't.

zookepeper didn't get merged becasue after debugging this further we found there is a bug in helm [bug](https://github.com/helm/helm/issues/9136#issuecomment-1205728157) and using a workaround I could finally get rid of that key from the manifest without modifying the charts.

Since there is a workaraound I felt the change I made and got merged in Kafka chart was unessesary so I created **this** PR  to rollback the changes in Kafka chart and put back the key. Since zookeeper chat didnt get merged I thought it made sense to roll back Kafka chart.

Now, there is no harm in leaving the chart like this as far as I know.

Hope is clear now.
Thanks!

aoterolorenzo (approved):
> 
> Hope is clear now. Thanks!

Fair enough, thank you for taking the time!

PR #13596, opened 2022-11-18
[bitnami/redis] Add minReadySeconds configuration

gmbnomis:
### Description of the change

Allow to set `minReadySeconds` in the `StatefulSet` or `Deployment` resource of a `master`, `replica`, or `sentinel` deployment.

### Benefits

For larger setups with lots of data to sync after startup, an additional delay before upgrading the next
node may be helpful.

### Possible drawbacks

As minReadySeconds is a recent addition to Statefulset (graduated to beta in Kubernetes 1.23), the field is added conditionally, i.e. only if the value is not zero.

The configuration will not work when set on older Kubernetes versions.

### Checklist


rafariossaa:
Hi,
  Just asking, for that use case, instead of a delay that you need to guess for the proper vaule, wouldn't be more useful to set your [own probes](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#types-of-probe). Maybe a startup probe would do. 
  What do you think ? I am not very into adding features that are in alpha state.

gmbnomis:
@rafariossaa I am sorry for answering so late, I had an "interrupt" that kept me busy.

Perhaps I should have been more specific in the "benefits" section. We intend to use this during updates of Redis to slow down the rate of upgrades.

Upgrades may cause multiple failovers to happen (for a three node sentinel setup, up to three failovers: node 2 -> node 1 -> node 0 -> node 1 or 2). Setting  `minReadySeconds`  will reduce the rate of failovers in quick succession and reduce the strain on the network if there is much data to move for initial sync (because there are pauses in-between).

One could even slow it down to such a rate to allow manual inspection, but in this case, a [partitioned rollout](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#partitions) would probably be more appropriate.

There are two alternatives to this:
1. Partitioned rollout: One could change the partitions multiple times. This is more work (especially as we use GitOps and don't access k8s resources directly), but is doable.
2. As you say, we could e.g. increase `initialDelaySeconds` of the startup or readiness probe, but this also delays the readiness of the container.

In conclusion, I think `minReadySeconds` will still be helpful in some situations. Btw. the feature is not in alpha. Since 1.25 it is stable, see https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#minimum-ready-seconds.

rafariossaa:
Hi,
  My main concern was about this "being" an alfa feature. Thanks for pointing me to the right documentation.

rafariossaa (approved):
LGTM

rafariossaa:
I am adding the version checking to be able to support older version of k8s (1.19+) without errors.

PR #13598, opened 2022-11-19
[bitnami/keycloak] fix apiversion hardcode for NetworkPolicy

my-git9:
[bitnami/keycloak] fix apiversion hardcode for NetworkPolicy


jotamartos (approved):
Thanks for your contribution

PR #13600, opened 2022-11-19
[bitnami/rabbitmq] fix apiversion hardcode for NetworkPolicy

my-git9:
[bitnami/rabbitmq] fix apiversion hardcode for NetworkPolicy


CeliaGMqrz (approved):
Thanks! LGTM

PR #13610, opened 2022-11-20
[bitnami/thanos] update PrometheusRule names to be unique

zzorica:
### Description of the change

When setting `prometheusRule.enabled: true` in chart verstion 11.6.1 those CRDs resources fail with: 

> Helm upgrade failed: error while running post render on files: may not add resource with an already registered id

Issue is that it wants to create all of `PrometheusRule` resources with the same name in the same namespace.

### Benefits

This will add suffix referring to components that rules are for which will set different names to all those resources.

### Possible drawbacks

Maybe some better logic instead of hardcoding these. Though it should be ok since its related to this specific chart and should not clash with other resources of the same kind.

### Applicable issues

None that I know off.

### Checklist

### DIFF
--- a/bitnami/thanos/templates/alert-rule/ruler.yml
+++ b/bitnami/thanos/templates/alert-rule/ruler.yml
@@ -5,7 +5,7 @@ Generated from https://github.com/thanos-io/thanos/blob/main/examples/alerts/ale
 apiVersion: monitoring.coreos.com/v1
 kind: PrometheusRule
 metadata:
-  name: {{ template "common.names.fullname" . }}-rule
+  name: {{ template "common.names.fullname" . }}-ruler
   namespace: {{ default .Release.Namespace .Values.metrics.prometheusRule.namespace | quote }}
   labels: {{- include "common.labels.standard" . | nindent 4 }}
     {{- if .Values.metrics.prometheusRule.additionalLabels }}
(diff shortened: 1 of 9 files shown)


bitnami/thanos/templates/alert-rule/ruler.yml
```diff
@@ -5,7 +5,7 @@ Generated from https://github.com/thanos-io/thanos/blob/main/examples/alerts/ale
 apiVersion: monitoring.coreos.com/v1
 kind: PrometheusRule
 metadata:
-  name: {{ template "common.names.fullname" . }}
+  name: {{ template "common.names.fullname" . }}-rule
```
corico44, line 8: I think it would be better to use the full name as the identifier. In this case, you would use ruler instead of rule.

bitnami/thanos/templates/alert-rule/store_gateway.yml
```diff
@@ -5,7 +5,7 @@ Generated from https://github.com/thanos-io/thanos/blob/main/examples/alerts/ale
 apiVersion: monitoring.coreos.com/v1
 kind: PrometheusRule
 metadata:
-  name: {{ template "common.names.fullname" . }}
+  name: {{ template "common.names.fullname" . }}-store
```
corico44, line 8: I think it would be better to use the full name as the identifier. In this case, you would use store-gateway instead of store.

corico44:
Hello @zzorica, can you resolve the conflict in the Chart.yaml file, please?

bitnami/thanos/templates/alert-rule/ruler.yml
zzorica: done

zzorica:
@corico44 sry for delay on this. I adjusted the naming and bumped up the chart versionm

ajoux-accor:
Hi,

We are impacted by this issue, any news for the release ?

corico44 (approved):
LGTM

PR #13628, opened 2022-11-22
[bitnami/pinniped] fix apiversion hardcode ClusterRoleBinding

my-git9:
[bitnami/pinniped] fix apiversion hardcode ClusterRoleBinding


migruiz4 (approved):

PR #13639, opened 2022-11-22
[bitnami/kubernetes-event-exporter]fix apiversion hardcode for Role

my-git9:
[bitnami/kubernetes-event-exporter]fix apiversion hardcode for Role


migruiz4 (approved):

PR #13655, opened 2022-11-23
[bitnami/phpmyadmin]fix apiversion hardcode for deployment

my-git9:
[bitnami/phpmyadmin]fix apiversion hardcode for deployment


corico44 (approved):
LGTM

PR #13662, opened 2022-11-23
[bitnami/zookeeper] replace commonAnnotations & commonLabels with persistence.annotations & persistence.labels in volumeClaimTemplates

itscaro:
New PR following https://github.com/bitnami/charts/pull/8272
Cf. https://github.com/bitnami/charts/pull/8272#issuecomment-1324830168

### DIFF
--- a/bitnami/zookeeper/templates/statefulset.yaml
+++ b/bitnami/zookeeper/templates/statefulset.yaml
@@ -492,9 +492,6 @@ spec:
         {{- range $key, $value := .Values.persistence.annotations }}
           {{ $key }}: {{ $value }}
         {{- end }}
-        {{- if .Values.commonAnnotations }}
-          {{- include "common.tplvalues.render" ( dict "value" .Values.commonAnnotations "context" $ ) | nindent 10 }}
-        {{- end }}
         {{- if .Values.commonLabels }}
         labels:
           {{- include "common.tplvalues.render" ( dict "value" .Values.commonLabels "context" $ ) | nindent 10 }}
@@ -518,9 +515,6 @@ spec:
         annotations:
         {{- range $key, $value := .Values.persistence.annotations }}
           {{ $key }}: {{ $value }}
-        {{- end }}
-        {{- if .Values.commonAnnotations }}
-          {{- include "common.tplvalues.render" ( dict "value" .Values.commonAnnotations "context" $ ) | nindent 10 }}
         {{- end }}
         {{- if .Values.commonLabels }}
         labels:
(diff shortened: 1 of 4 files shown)


bitnami/zookeeper/templates/statefulset.yaml
```diff
@@ -515,16 +510,11 @@ spec:
...
     - metadata:
         name: data-log
-        annotations:
-        {{- range $key, $value := .Values.persistence.annotations }}
-          {{ $key }}: {{ $value }}
-        {{- end }}
-        {{- if .Values.commonAnnotations }}
-          {{- include "common.tplvalues.render" ( dict "value" .Values.commonAnnotations "context" $ ) | nindent 10 }}
+        {{- if .Values.persistence.annotations }}
+        annotations: {{- include "common.tplvalues.render" (dict "value" .Values.persistence.annotations "context" $) | nindent 10 }}
```
aoterolorenzo, line 514: Thanks for your contribution @itscaro!

I think that both `commonLabels` and `commonAnnotations` could still coexist with the `persistence.annotations` and `persistence.labels`.

aoterolorenzo (changes requested):

bitnami/zookeeper/templates/statefulset.yaml
itscaro: But it's not the case with other charts kafka, postgresql.

jouve: I agree with @itscaro arguments (commonAnnotation may be mutated & consistency with other charts)

itscaro: https://github.com/bitnami/charts/blob/2814eed/bitnami/kafka/templates/statefulset.yaml#L571-L578
https://github.com/bitnami/charts/blob/2814eed/bitnami/postgresql/templates/primary/statefulset.yaml#L611-L618
https://github.com/bitnami/charts/blob/2814eed/bitnami/thanos/templates/storegateway/statefulset-sharded.yaml#L312-L319

etc...

aoterolorenzo: I understand the situation. The thing is, at the end of the day neither of both `persistence.annotations` and `commonAnotations` "match the "immutability" of persistence claim" (as helm values are not persistent by nature), but when it gets to its usage, makes sense that in case you need, you use the commonAnnotations dinamically rather than `persistence.annotations`. But even like this, we may have to rely to the nature of how persistence objects are interpreted by kubernetes and if shuold be a way to provide some not inmutable references to a pv/pvc.

Anyway, completely understand now. Approved! :)

aoterolorenzo (approved):

PR #13666, opened 2022-11-23
[bitnami/redis-cluster] Revisit tests

FraPazGal:
### Description of the change

Tested in https://github.com/FraPazGal/charts/pull/135

### Checklist

### DIFF
--- a/.vib/redis-cluster/goss/goss.yaml
+++ b/.vib/redis-cluster/goss/goss.yaml
@@ -1,39 +1,50 @@
+{{- $auth := printf "REDISCLI_AUTH=%s" .Vars.password }}
+{{- $headless_port := .Vars.redis.containerPorts.redis }}
+{{- $lb_endpoint := printf "-h redis-cluster -p %d" .Vars.service.ports.redis }}
+{{- $nodes := .Vars.cluster.nodes }}
 command:
-  redis-version:
-    exec: redis-server --version | grep --quiet "$APP_VERSION"
+  {{- $key := printf "key_%s" (randAlpha 5) }}
+  {{- $value := printf "value_%s" (randAlpha 5) }}
+  redis-set-key-value-pairs:
+    exec: |
+      export {{ $auth }} && \
+      redis-cli {{ $lb_endpoint }} CLUSTER SLOTS | sed -n '3p' | \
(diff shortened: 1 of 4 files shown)


agomezmoron (dismissed):
LGTM

.vib/redis-cluster/goss/goss.yaml
```diff
@@ -1,39 +1,50 @@
...
+{{- $nodes := .Vars.cluster.nodes }}
 command:
-  redis-version:
-    exec: redis-server --version | grep --quiet "$APP_VERSION"
+  {{- $key := printf "key_%s" (randAlpha 5) }}
+  {{- $value := printf "value_%s" (randAlpha 5) }}
+  redis-set-key-value-pairs:
+    exec: |
+      export {{ $auth }} && \
+      redis-cli {{ $lb_endpoint }} CLUSTER SLOTS | sed -n '3p' | \
```
joancafom, line 11: As per [documentation](https://redis.io/commands/cluster-slots/), this command is now marked as deprecated. If you are trying to obtain the IP of the master node, you could try using `CLUSTER NODES` instead and then filter by role type.

.vib/redis-cluster/goss/goss.yaml
```diff
@@ -1,39 +1,50 @@
+{{- $auth := printf "REDISCLI_AUTH=%s" .Vars.password }}
+{{- $headless_port := .Vars.redis.containerPorts.redis }}
```
joancafom, line 2: It took me some time to understand the tests where this variable was involved because of its naming. It is used both in connections that use the headless svc and plain IPs:

- `-h redis-cluster-{{ $i }}.redis-cluster-headless -p {{ $headless_port }}`: OK, it is easy to understand as the hostname is a `headless` svc.
- `xargs -I '{}' redis-cli -c -h '{}' -p {{ $headless_port }}`: Not that straightforward to notice :(, the host is an IP

I would simplify to something like `redis_port`, which IMHO makes it easier to understand.

joancafom (commented):
Thanks for your contribution! I left you some comments as a review 😁

.vib/redis-cluster/goss/goss.yaml
FraPazGal: Applied both suggestions

joancafom (approved):

PR #13689, opened 2022-11-24
[bitnami/harbor] fix apiversion hardcode for deployment

my-git9:
[bitnami/harbor] fix apiversion hardcode for deployment


jotamartos (approved):
Thanks for your contribution

PR #13778, opened 2022-12-01
[bitnami/prestashop] Revisit tests

joancafom:
### Description of the change

Tested in https://github.com/joancafom/charts/pull/102

### Checklist


.vib/prestashop/cypress/cypress/integration/prestashop_spec.js
```diff
@@ -12,119 +12,32 @@ it('allows a user to place an order and an admin to list it', () => {
...
-  const DEFAULT_ACTIVE_PAY_METHODS = ['Bank transfer', 'Payments by check'];
-
-  cy.contains('div.card', 'Active payment').within(() => {
-    DEFAULT_ACTIVE_PAY_METHODS.forEach((method) => {
-      cy.contains(method);
-    });
-  });
-});
-
-it('allows performing backups', () => {
```
FraPazGal, line 106: I suppose this test was added to check prestashop's PV. Are we using the PV in any way with the remaining test or chart installation? If not, I would include again this backup test.

FraPazGal (commented):
Thanks for the PR @joancafom, please check my comments when possible!

.vib/prestashop/cypress/cypress/integration/prestashop_spec.js
joancafom: Agreed, I added it back. Thanks!

FraPazGal (approved):
LGTM, thanks!

PR #13782, opened 2022-12-01
Revert "[bitnami/redis] Add minReadySeconds configuration"

rafariossaa:
Reverts bitnami/charts#13596

I revert this because we need to support k8s 1.19+ and this feature was not in stable until 1.25. So, k8s version checking is needed and I missed it.


rafariossaa:
updated in https://github.com/bitnami/charts/pull/13783

Issue #13800, opened 2022-12-02
[bitnami/postgresql-ha] Change of postgres password key avoids startup


sjoukedv:
### Name and Version

bitnami/postgresql-ha 10.0.6

### What steps will reproduce the bug?

This PR #13733 renamed `postgres-password` to `password` which is fine in essence, but the statefulset use envFrom with the new key. I am using a custom secret for these values (so that it is managed outside of the chart) and therefore the key does not exist. 

### Are you using any custom parameters or values?

n.a.

### What do you see instead?

I was unable to upgrade because of `CreateConfigError` on the pod as the key did not exist. We should have an if statement to check if the new value exists and otherwise default to `postgres-password`

CeliaGMqrz:
Hi @sjoukedv,

Thanks for reporting this issue. Would you like to contribute by creating a PR to solve the issue? The Bitnami team will be happy to review it and provide feedback. [Here](https://github.com/bitnami/charts/blob/master/CONTRIBUTING.md) you can find the contributing guidelines.

sjoukedv:
> Hi @sjoukedv,
> 
> Thanks for reporting this issue. Would you like to contribute by creating a PR to solve the issue? The Bitnami team will be happy to review it and provide feedback. [Here](https://github.com/bitnami/charts/blob/master/CONTRIBUTING.md) you can find the contributing guidelines.

Yes, I have created a PR which looks up the `postgres-password` in the secret and defaults to the new `password` key

```
- name: POSTGRES_PASSWORD
              valueFrom:
                secretKeyRef:
                  name: test-postgresql-ha-postgresql
                  key: password
```

I will link the PR here in a sec

sjoukedv:
Sorry for the duplicate PR I was using commitizen which did not sign my commits by default anymore. The open PR is ready for review.

CeliaGMqrz:
Thanks for the contribution! Our team will review your PR.

PR #13808, opened 2022-12-02
[bitnami/postgresql-ha] postgresql-ha POSTGRES_PASSWORD envFrom backward compatibility postgres-password key

sjoukedv:
### Description of the change

First determine if this charts will create the secret (a.k.a no `postgresql.existingSecret` is defined, using [ `postgresql-ha.postgresqlCreateSecret`](https://github.com/bitnami/charts/blob/main/bitnami/postgresql-ha/templates/_helpers.tpl#L304)), if so it will use the new `password` envFrom key. In case there is a custom secret provided it will lookup the secret and ONLY if it has a `postgres-password` key this will be used. Otherwise it will default to the new `password` key.

### Benefits

Backward compatibility with users that provide an existing secret with the old `postgres-password` key. Otherwise this lead to create container errors.

### Possible drawbacks

None, default to the new `password` key.

### Applicable issues

#13800

### Checklist

### DIFF
--- a/bitnami/postgresql-ha/Chart.yaml
+++ b/bitnami/postgresql-ha/Chart.yaml
@@ -29,4 +29,4 @@ name: postgresql-ha
 sources:
   - https://github.com/bitnami/containers/tree/main/bitnami/postgresql
   - https://www.postgresql.org/
-version: 10.0.6
+version: 10.0.7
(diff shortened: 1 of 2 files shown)


sjoukedv:
Close in favor of #13809

PR #13809, opened 2022-12-02
[bitnami/postgresql-ha] backwards compat postgres-password

### LINKED ISSUE #13800: [bitnami/postgresql-ha] Change of postgres password key avoids startup

sjoukedv:
### Description of the change

First determine if this charts will create the secret (a.k.a no `postgresql.existingSecret` is defined, using [ `postgresql-ha.postgresqlCreateSecret`](https://github.com/bitnami/charts/blob/main/bitnami/postgresql-ha/templates/_helpers.tpl#L304)), if so it will use the new `password` envFrom key. In case there is a custom secret provided it will lookup the secret and ONLY if it has a `postgres-password` key this will be used. Otherwise it will default to the new `password` key.

### Benefits

Backward compatibility with users that provide an existing secret with the old `postgres-password` key. Otherwise this lead to create container errors.

### Possible drawbacks

None, default to the new `password` key.

### Applicable issues

- Fixes #13800

### Checklist

### DIFF
--- a/bitnami/postgresql-ha/Chart.yaml
+++ b/bitnami/postgresql-ha/Chart.yaml
@@ -29,4 +29,4 @@ name: postgresql-ha
 sources:
   - https://github.com/bitnami/containers/tree/main/bitnami/postgresql
   - https://www.postgresql.org/
-version: 10.0.6
+version: 10.0.7
--- a/bitnami/postgresql-ha/templates/postgresql/statefulset.yaml
+++ b/bitnami/postgresql-ha/templates/postgresql/statefulset.yaml
@@ -191,7 +191,15 @@ spec:
               valueFrom:
                 secretKeyRef:
                   name: {{ include "postgresql-ha.postgresqlSecretName" . }}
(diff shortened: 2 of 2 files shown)


marcosbc (approved):
LGTM! Thanks for the fix!

rivoalrivoal:
For now, this not working :
In [helm doc](https://helm.sh/docs/chart_template_guide/function_list/#index), if you need to access in subvalue, need to do this : 
`{{- if ( index ( lookup "v1" "Secret" (include "common.names.namespace" .) (include "postgresql-ha.postgresqlSecretName" .) ) "data" "postgres-password" ) }}`

PR #13876, opened 2022-12-08
[bitnami/pinniped] Separate service configuration for supervisor api and aggregator server

### LINKED ISSUE #13875: [bitnami/pinniped] Reused nodePorts for two services
### Name and Version

bitnami/pinniped 0.4.6

1. In a kubeadm setup kubernetes cluster
2. I have the nodePort range set to 49000-49100
3. With the values yaml provided below
4. I run `helm upgrade --install --values helm-values.yml --version=0.4.6 pinniped-authenticator bitnami/pinniped`

```yaml
concierge:
  enabled: true
  replicaCount: 2
  # TODO
  resources:
    limits: {}
    requests: {}

supervisor:
  enabled: true
  replicaCount: 2

  # TODO
  resources:
    limits: {}
    requests: {}

  service:
    type: NodePort
    nodePorts:
      https: "49000"
```

I expect the installation to succeed with a defined nodePort for the supervisor service.

### What do you see instead?

The installation fails because nodePort 49000 is already in use.
```console
$ helm upgrade --install --values helm-values.yml --version=0.4.6 pinniped-authenticator bitnami/pinniped 
Release "pinniped-authenticator" does not exist. Installing it now.
Error: Service "pinniped-authenticator-supervisor-api" is invalid: spec.ports[0].nodePort: Invalid value: 49000: provided port is already allocated
```

```console
$ kubectl get svc
NAME                                                          TYPE        CLUSTER-IP       EXTERNAL-IP   PORT(S)         AGE
pinniped-authenticator-concierge                              ClusterIP   10.103.36.71     <none>        443/TCP         70s
pinniped-authenticator-concierge-pinniped-authenticator-api   ClusterIP   10.103.213.9     <none>        443/TCP         70s
pinniped-authenticator-supervisor                             NodePort    10.104.181.158   <none>        443:49000/TCP   70s
```

### Additional information

I have investigated the problem, and I think it is that two service definitions use the same configuration from the values file.
* https://github.com/bitnami/charts/blob/main/bitnami/pinniped/templates/supervisor/service-api.yaml#L51
* https://github.com/bitnami/charts/blob/main/bitnami/pinniped/templates/supervisor/service.yaml#L51

In fact, based on the services included in the deployment based on pinnipads own documentation (see below), I think the `.Values.supervisor.service` should be broken up into two, because separate parameters of it correspond to two different services.

---

The original service definition looks like this:
```yaml
# Included in the manifest at https://get.pinniped.dev/v0.20.0/install-pinniped-supervisor.yaml
---
apiVersion: v1
kind: Service
metadata:
  name: pinniped-supervisor-api
  namespace: pinniped-supervisor
  labels:
    app: pinniped-supervisor
  annotations:
    kapp.k14s.io/disable-default-label-scoping-rules: ""
spec:
  type: ClusterIP
  selector:
    deployment.pinniped.dev: supervisor
  ports:
  - protocol: TCP
    port: 443
    targetPort: 10250

# Suggested by the documentation, done by me
---
apiVersion: v1
kind: Service
metadata:
  name: pinniped-supervisor-loadbalancer
  # Assuming that this is the namespace where the Supervisor was installed.
  # This is the default.
  namespace: pinniped-supervisor
spec:
  type: NodePort
  selector:
    # Assuming that this is how the Supervisor Pods are labeled.
    # This is the default.
    app: pinniped-supervisor
  ports:
  - protocol: TCP
    port: 443
    targetPort: 8443  # 8443 is the TLS port.
    nodePort: 49000
```

ecklm:
One single configuration is in parallel for two services. This PR separates them so that the can be configured separately. 

Fixes #13875

### DIFF
--- a/bitnami/pinniped/Chart.yaml
+++ b/bitnami/pinniped/Chart.yaml
@@ -22,4 +22,4 @@ name: pinniped
 sources:
   - https://github.com/bitnami/containers/tree/main/bitnami/pinniped
   - https://github.com/vmware-tanzu/pinniped/
-version: 0.4.7
+version: 1.0.0
--- a/bitnami/pinniped/README.md
+++ b/bitnami/pinniped/README.md
@@ -19,7 +19,7 @@ $ helm install my-release my-repo/pinniped
 
 Bitnami charts for Helm are carefully engineered, actively maintained and are the quickest and easiest way to deploy containers on a Kubernetes cluster that are ready to handle production workloads.
 
-This chart bootstraps a [Grafana Loki](https://github.com/grafana/loki) Deployment in a [Kubernetes](https://kubernetes.io) cluster using the [Helm](https://helm.sh) package manager.
+This chart bootstraps a [Pinniped](https://pinniped.dev/) Deployment in a [Kubernetes](https://kubernetes.io) cluster using the [Helm](https://helm.sh) package manager.
 
 Bitnami charts can be used with [Kubeapps](https://kubeapps.com/) for deployment and management of Helm Charts in clusters. This Helm chart has been tested on top of [Bitnami Kubernetes Production Runtime](https://kubeprod.io/) (BKPR). Deploy BKPR to get automated TLS certificates, logging and monitoring for your applications.
 
--- a/bitnami/pinniped/README.md
+++ b/bitnami/pinniped/README.md
@@ -367,6 +367,12 @@ Find more information about how to deal with common errors related to Bitnami's 
 
 ## Upgrading
 
+### To 1.0.0
+
+This version brings a breaking change into the configuration, eliminating abused reuse of Pinniped service parameters.
+The `supervisor.service` object is now separated into `supervisor.service.api` which configures the service used by Pinniped internally, and `supervisor.service.public` which configures the service the users interact with.
+In case configuration was specified in the `supervisor.service` object, now it needs to be redistributed into the two new objects. Keep in mind that the API service default service type was also changed to `ClusterIP` to reflect more on how the API service is used by default. Also the formerly `supervisor.service.ports.aggregatedAPIService` parameter is now only available under the API service configuration, because it is not a relevant parameter for the user-facing service.
+
 ### To 0.4.0
 
 This version updates Pinniped to its newest version, 0.20.x. For more information, please refer to [the release notes](https://github.com/vmware-tanzu/pinniped/releases/tag/v0.20.0).
--- a/bitnami/pinniped/README.md
+++ b/bitnami/pinniped/README.md
@@ -273,35 +273,47 @@ The command removes all the Kubernetes components associated with the chart and 
 
 ### Supervisor Traffic Exposure Parameters
 
-| Name                                           | Description                                                                                                                      | Value                       |
-| ---------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------- | --------------------------- |
-| `supervisor.service.type`                      | Supervisor service type                                                                                                          | `LoadBalancer`              |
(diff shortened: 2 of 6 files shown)


bitnami/pinniped/Chart.yaml
```diff
@@ -22,4 +22,4 @@ name: pinniped
 sources:
   - https://github.com/bitnami/containers/tree/main/bitnami/pinniped
   - https://github.com/vmware-tanzu/pinniped/
-version: 0.4.6
\ No newline at end of file
+version: 0.4.7
```
marcosbc, line 25: Since this is adding a feature, please increase the minor.

marcosbc (commented):
Thanks for the contribution! I believe this contribution makes sense, however, it seems like this is introducing a lot of breaking changes due to the renaming of options.

Would it be possible to adapt it to support both the "original" parameters and the new ones? If not, the only option would be to increase the major version of the chart, and document this breaking change ([example](https://github.com/bitnami/charts/tree/main/bitnami/etcd#upgrading)). With this change, I believe that could make sense though.

ecklm:
> Thanks for the contribution! I believe this contribution makes sense, however, it seems like this is introducing a lot of breaking changes due to the renaming of options.
> 
> Would it be possible to adapt it to support both the "original" parameters and the new ones? If not, the only option would be to increase the major version of the chart, and document this breaking change ([example](https://github.com/bitnami/charts/tree/main/bitnami/etcd#upgrading)). With this change, I believe that could make sense though.

@marcosbc TLDR: By keeping it compatible, it'd keep being ambiguous, incuring unwanted costs (by duplicating load balancers) or would loose configuration possibilities.

First I thought, I'll just fix which service is using which variable. Then the aggregator service would use the `nodePort` and the api service wouldn't. And vice versa for `aggregatedAPIServer`. But then their types would still be forced to go along, and based on the original deployment, the api service is not necessary to be of type NodePort when the aggregator might be. It also doesn't need a load balancer, which potentially adds costs in a public cloud environment unnecessarily.

So we either
1. drop all configurability from the api service. It'd still work as a `ClusterIP` service with the default `aggregatedAPIServer` (which by software restriction, doesn't seem configurable anyway), but we'd loose configurability of the metadata. I personally think this is feasible, but I guess exposing every possible configuration parameter is a policy here so that every user can decide what they want to modify.
2. Create separate configuration for both of them. -> Depending on the naming, this is likely a breaking change.

@marcosbc if you agree this should keep being a breaking change, I'll add the necessary parts to the documentation. In this case, please also comment on the naming I pointed out earlier myself.

ecklm:
Can I get some feedback on which direction I should take?

CeliaGMqrz:
Hi @ecklm, 

We agree with you on creating a separate configuration for both services, which should be a major. Also, please remember to document the changes in the README.md as my colleague explained.

bitnami/pinniped/Chart.yaml
ecklm: I've updated 1.0.0 since it's definitely a breaking change and it fixes a bux. See 54eee0ea2acb33dfffed531e3d7fb05b92d704b6.

ecklm:
@CeliaGMqrz I added some description in d6dc9d6f3f2eeadf379b59b8388eae2fa33a96d5, I'm not sure how deep I should go.

ecklm:
Can I have my new changes reviewed, please?

marcosbc (approved):

Also, just wanted to apologize for the delay. Due to christmas holidays, many of our team members have been off and we have had a very limited capacity for checking PRs and issues.

ecklm:
No problem at all. I thought it was the case. I just wanted to comment something to prevent the bot closing the PR.

PR #13971, opened 2022-12-15
[bitnami/grafana] Catch typeError in Cypress test

mdhont:

### Description of the change

It's a workaround for a typeError thrown in the console, which makes the Cypress test fail. An issue has been created:
https://github.com/grafana/grafana/issues/60377

### Checklist

(https://github.com/bitnami-labs/readme-generator-for-helm)

### DIFF
--- a/.vib/grafana/cypress/cypress/support/commands.js
+++ b/.vib/grafana/cypress/cypress/support/commands.js
@@ -12,6 +12,16 @@ for (const command of ['click']) {
   });
 }
 
+Cypress.on('uncaught:exception', (err, runnable) => {
+  // In Grafana 9.3.0 the console throws a typeError with message
+  //'Cannot read properties of undefined (reading 'eventTrackingNamespace')'.
+  // An issue has been created:
+  // https://github.com/grafana/grafana/issues/60377
+  if (err.message.includes("Cannot read properties of undefined (reading 'eventTrackingNamespace')")) {
+    return false
+  }
+})
+
 Cypress.Commands.add(
   'login',
   (username = Cypress.env('username'), password = Cypress.env('password')) => {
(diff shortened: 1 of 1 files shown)


.vib/grafana/cypress/cypress/support/commands.js
```diff
@@ -12,6 +12,16 @@ for (const command of ['click']) {
   });
 }
 
+Cypress.on('uncaught:exception', (err, runnable) => {
+  // In Grafana 9.3.0 the console throws a typeError with message
+  //'Cannot read properties of undefined (reading 'eventTrackingNamespace')'.
```
corico44, line 17: Could you add a space just after the slashes to follow the same structure as the other lines, please?

.vib/grafana/cypress/cypress/support/commands.js
```diff
@@ -12,6 +12,16 @@ for (const command of ['click']) {
...
+  // An issue has been created:
+  // https://github.com/grafana/grafana/issues/60377
+  if (err.message.includes("Cannot read properties of undefined (reading 'eventTrackingNamespace')")) {
```
corico44, line 20: Should the error message be so specific or would a simple **"Cannot read properties of undefined"** be enough?

mdhont: I'd rather use a specific message in this case, to not cover up any other errors.

mdhont:
It seems the issue is being fixed: https://github.com/grafana/grafana/pull/60382, so I think we should just wait a bit. Wdyt @FraPazGal?

FraPazGal:
I have some doubts about waiting, mainly because we don't know when they will release a new version including the fix. If it is a minor bug and upstream doesn't release a hotfix, we may have to wait until the next scheduled release. Does the bug break anything major in the asset itself? If it doesn't, we can add the `Cypress.on` directive to catch it and release this version.

Regardless, I would update the comment to explain that a fix has been merged and to review this `Cypress.on` in the future (when the asset version `>9.3.1`).

PR #14018, opened 2022-12-19
[bitnami/etcd] Add support for shareProcessNamespace

### LINKED ISSUE #13989: [bitnami/etcd] healthcheck.sh leaving behind etcdctl zombies on timeout
### Name and Version

bitnami/etcd 8.5.8

We run a kubeadm cluster using calico as the networking layer on physical machines. We've recently had some calico issues which caused networking interruptions between pods. When something like that happens we end up with a bunch of etcdctl zombies.

My _guess_ is that due to the networking issues the etcdctl command takes a long time, which causes healthcheck.sh to get killed. For reasons unclear to me that doesn't properly kill the etcdctl command at that point, which, because the parent process is gone then gets attached to the etcd command running as PID 1, which obviously doesn't handle cleaning up zombies properly. I suspect it is very similar to this issue seen with redis: https://github.com/bitnami/charts/issues/5328

Due to the redis issue I think that enabling shareProcessNamespace on the etcd pod should already fix this issue, but the chart currently doesn't allow doing that.

We use a replicacount of 3

No zombies

### What do you see instead?

An etcdctl zombie for every timed out healthcheck

RobinGeuze:
Signed-off-by: Robin Geuze <robing@transip.nl>

### Description of the change
Add the option to set shareProcessNamespace on the etcd statefulset. This will cause the /pause process to get PID 1 and enable it to reap any zombies caused by the healthchecks

### Benefits
People that have issues with zombie etcdctl processes can solve it by enabling the setting

### Possible drawbacks
None, its disabled by default

### Applicable issues
  - fixes #13989

### Additional information
Based on a similar solution in redis-cluster

### Checklist


bitnami/etcd/values.yaml
```diff
@@ -464,6 +464,13 @@ priorityClassName: ""
 ## ref: https://kubernetes.io/docs/concepts/containers/runtime-class/
 ##
 runtimeClassName: ""
+## @param shareProcessNamespace Enable shared process namespace in a pod.
+## If set to false (default), each container will run in separate namespace, etcd will have PID=1.
+## If set to true, the /pause will run as init process and will reap any zombie PIDs,
+## for example, generated by a custom exec probe running longer than a probe timeoutSeconds.
+## Enable this only if customLivenessProbe or customReadinessProbe is used and zombie PIDs are accumulating.
+## Ref: https://kubernetes.io/docs/tasks/configure-pod-container/share-process-namespace/
```
CeliaGMqrz, line 472:
```suggestion
## Ref: https://kubernetes.io/docs/tasks/configure-pod-container/share-process-namespace/
##
```

CeliaGMqrz (commented):
Thanks for your contribution. Can you update the minor version of the chart instead? Since you added a new feature, it is considered a new minor version instead of a patch version. In order to follow a similar approach to other READMEs in the repository, I proposed some minor changes.

RobinGeuze:
Hey @CeliaGMqrz, thanks for the comments, I've applied both changes.

CeliaGMqrz (approved):
Thanks for your patience, LGTM!

PR #14071, opened 2022-12-22
[bitnami/rabbitmq] Set advanced configuration from an existing secret

### LINKED ISSUE #14017: [rabbitmq chart] Add an option to override advancedConfiguration with a pre existing secret
### Name and Version

bitnami/rabbitmq

The "advancedConfiguration" field could have sensitive data (Azure active directory configuration in my case) or be different between environments.

My proposal is that it would be nice to have a way to load that configuration from a secret. Then we can address both issues, not expose sensitive data and have configurations per environment easily.

If needed I am willing to help, but I don't know where to start so I appreciate any help.

I think the helm post render option could solve this, but doesn't fit my needs because argocd isn't compatible with that helm's feature. And I think a native solution would be better to not add that complexity to the deploy

FranAguiar:
### Description of the change
Set advanced configuration from an existing secret instead of plain text in values yaml. Plain text still available

### Benefits
Loading advanced configuration from a secret allow to not expose sensitive data or have different configurations between different instances with the same chart.

### Possible drawbacks
If both option are set, the secret will be ignored.

### Applicable issues
  - fixes #14017 

### Checklist


bitnami/rabbitmq/templates/statefulset.yaml
```diff
@@ -360,15 +360,14 @@ spec:
...
-                path: advanced.config
-              {{- end }}
+          projected:
+            sources:
+              - secret:
+                  name: {{ printf "%s-config" (include "common.names.fullname" .) }}
+            {{- if and .Values.advancedConfigurationExistingSecret (not .Values.advancedConfiguration) }}
+            - secret:
+                name: {{ tpl .Values.advancedConfigurationExistingSecret . | quote }}
+            {{- end }}
```
fmulero, line 370: Wrong indentation
```suggestion
              {{- if and .Values.advancedConfigurationExistingSecret (not .Values.advancedConfiguration) }}
              - secret:
                  name: {{ tpl .Values.advancedConfigurationExistingSecret . | quote }}
              {{- end }}
```

fmulero (changes requested):
Great!! thanks a lot Francisco for your contribution, it simplifies the implementation, I like it.

There is a little failure, Could you review it?

FranAguiar:
Reviewed the indentation error and also I added the mandatory key in the secret to the values yaml and the readme

fmulero (approved):
Great!!! LGTM

PR #14085, opened 2022-12-23
 [bitnami/wordpress] Invert memory/cpu in hpa for Argo CD users

Callixte-de-Vulpillieres:
### Description of the change

Invert CPU and Memory in the hpa definition, because it causes some issues when using Argo CD : Kubernetes always puts Memory first even if it not the case in the yaml file, and so when Argo CD compares the "live manifest" with the one of the repo, it detects a diff and so the app is falsely described at "out-of-sync". See attached screenshot if my explanation is unclear.

Capture d’écran du 2022-12-23 12-59-34

### Benefits

No more constant "out-of-sync" status for Argo CD users. 

### Checklist


CeliaGMqrz (approved):
Sorry for the delay, Looks good to me!
Thanks for your contribution!

Issue #13541, opened 2022-11-16
ClickHouse does not have cluster authentication set up by default.


RossRothenstine:
### Name and Version

bitnami/clickhouse 1.0.3

### What steps will reproduce the bug?

1. Install the chart via helm with default values. 
```
helm3 install clickhouse bitnami/clickhouse --namespace clickhouse --create-namespace
```

2. Exec into a ClickHouse shard.
```
k exec -n clickhouse clickhouse-shard0-0 -it -- clickhouse-client --user default --password $(kubectl get secret --namespace clickhouse clickhouse -o jsonpath="{.data.admin-password}" | base64 -d)
```

3. Create a database. Note the on cluster and engine are important.
```
 CREATE DATABASE my_database ON CLUSTER default ENGINE=Atomic;
USE my_database;
```

4. Create a table. This will be the local part of the distributed table.
```
 CREATE TABLE logs_local ON CLUSTER default(ts DateTime, text String) ENGINE=MergeTree ORDER BY ts PARTITION BY toStartOfHour(ts) TTL ts + toIntervalWeek(1);
```

5. Create the distributed table.

```
CREATE TABLE logs ON CLUSTER default AS logs_local ENGINE = Distributed(default, my_database, logs_local, rand());
```

6. Attempt to select.

```
 SELECT * FROM logs;
```
 
 7. Observe the following output:
 
 ```
clickhouse-shard0-0.clickhouse-headless.clickhouse.svc.cluster.local :) SELECT * FROM logs;

SELECT *
FROM logs

Query id: 34e3d725-8ce4-433c-a778-779d7b29f5b7


0 rows in set. Elapsed: 0.060 sec.

Received exception from server (version 22.9.4):
Code: 516. DB::Exception: Received from localhost:9000. DB::Exception: Received from clickhouse-shard1-1.clickhouse-headless.clickhouse.svc.cluster.local:9000. DB::Exception: default: Authentication failed: password is incorrect or there is no user with such name. (AUTHENTICATION_FAILED)
```

### What is the expected behavior?

Shards should be able to communicate with each other.

### What do you see instead?

There are permission problems talking to other shards.

### Additional information

The default user should have access to all other shards. Its possible to achieve this by overriding the config and the user. However this is unintuitive to new users who would use ClickHouse.

RossRothenstine:
[...]

PR #14106, opened 2022-12-26
[bitnami/clickhouse] Cluster authentication fix

clintz:
Signed-off-by: Christopher Lintz <chlintz@twilio.com>

### Description of the change
Added a non-empty remote_servers/default/secret in values.yaml.

This utilizes a Clickhouse feature that uses the initial
user as the current user for the inter-cluster queries. All nodes with the same non-empty secret string will trust each other.  The value can be any non-empty string.

For more info, see https://github.com/ClickHouse/ClickHouse/pull/13156

### Benefits

Resolves the cluster authentication issue securely without user/pass exposed in remote_server shards.

### Possible drawbacks

N/A

### Applicable issues

https://github.com/bitnami/charts/issues/13541

### Checklist


clintz:
Sorry missed moving to draft earlier. This is good now - no need for draft.

fmulero (commented):
Thank you so much @clintz for taking care of this topic.

I've been taken a look and I have some concerns about the implementation. Having a hard coded default password is not a good practice and I am not sure about the impact of this change in existing installations, have you tested an upgrade from previous version?

trennepohl:
Hey @clintz  thanks for this, I've spent a good chunk of time trying to figure out why I was having issues with distributed queries even though I had users with enough permissions for that.

Regarding @fmulero  comment you can use the `from_env` flag available in the config and set the password as a Kubernetes secret and mount it as an environment variable.

fmulero:
> Regarding @fmulero comment you can use the from_env flag available in the config and set the password as a Kubernetes secret and mount it as an environment variable.

Good point! that's very useful. In addition, we should avoid the use of hardcoded passwords, we should autogenerate it during the installation. Here is my proposal:
* Add a new key to the [secret](https://github.com/bitnami/charts/blob/main/bitnami/clickhouse/templates/secret.yaml) using the [common.secrets.passwords.manage](https://github.com/bitnami/charts/blob/610f860fd8197f6f1dc1906511b56a96df42dac9/bitnami/common/templates/_secrets.tpl#L61-L85) template to create a new secret if the value is not provided by the user.
* Load the secret in a new env var like is already done for the [admin password](https://github.com/bitnami/charts/blob/610f860fd8197f6f1dc1906511b56a96df42dac9/bitnami/clickhouse/templates/statefulset.yaml#L202-L206)
* Use `env_from` in the configuration file to load the secret from the env var created in the previous step.

fermuch:
hey @clintz ! Thanks for the PR, I didn't know what was wrong with the cluster settings.

Are you still able to update the PR to accomodate what @fmulero proposed? If not, is it ok if I continue from where you've left it?

clintz:
Hi @fermuch - sorry for the long delay. Please feel free to pick up where I left off.

fermuch:
hey @clintz ! Thanks! Sadly, at $WORK we're already setting this configuration manually, so I won't have that much time anymore to implement this.

@fmulero is there someone who can implement this available? If not, I'll try to make some time next month.

fmulero:
We're a small team and our capacity is not too high. This is currently in our radar but the priority is not too high. Said that, we can not guarantee any ETA.

yasha-dev1:
what's the status of this PR? as a hotfix we are creating a `default` user and granting all privilleges. I can contribute with this PR if any help is needed.

fmulero:
Thanks @yasha-dev1  for your interest in this PR.

The status is what you can see here. @clintz couldn't continue with my suggestions and @fermuch expressed his interest follow up but it seems doesn't have enough time also.

From my side there is no problem if you want to continue working on this.

PR #14127, opened 2022-12-29
bitnami/mariadb-galera: Add optional Annotation checksums of configmaps and secrets to Statefulset

mcc138:
### Description of the change

If the parameter checksumAnnotations is true, adds Annotations to the statefulset containing the checksums
of configmap, initialization-configmap, secrets if installed

### Benefits

When using ArgoCD or similar, the pods will automatically get restarted if the secret or configmaps get updated

### Possible drawbacks

None that I know of, as this is optional.

### Applicable issues

n/a

### Checklist

### DIFF
--- a/bitnami/mariadb-galera/templates/statefulset.yaml
+++ b/bitnami/mariadb-galera/templates/statefulset.yaml
@@ -7,8 +7,20 @@ metadata:
     {{- if .Values.commonLabels }}
     {{- include "common.tplvalues.render" (dict "value" .Values.commonLabels "context" $) | nindent 4 }}
     {{- end }}
-  {{- if .Values.commonAnnotations }}
-  annotations: {{- include "common.tplvalues.render" (dict "value" .Values.commonAnnotations "context" $) | nindent 4 }}
+  {{- if or .Values.commonAnnotations .Values.checksumAnnotations }}
+  annotations:
+    {{- if .Values.commonAnnotations }}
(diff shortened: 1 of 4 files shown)


bitnami/mariadb-galera/templates/statefulset.yaml
```diff
@@ -7,8 +7,20 @@ metadata:
     {{- if .Values.commonLabels }}
     {{- include "common.tplvalues.render" (dict "value" .Values.commonLabels "context" $) | nindent 4 }}
     {{- end }}
-  {{- if .Values.commonAnnotations }}
-  annotations: {{- include "common.tplvalues.render" (dict "value" .Values.commonAnnotations "context" $) | nindent 4 }}
+  {{- if or .Values.commonAnnotations .Values.checksumAnnotations }}
```
jotamartos, line 10: I'd follow the same approach we have in other solutions (for example RabbitMQ) and would not create a new `.Values.checksumAnnotations` parameter. I'd simply add the code and remove this "if" line. What do you think?

jotamartos (changes requested):

bitnami/mariadb-galera/templates/statefulset.yaml
mcc138: OK, I'll take a look at RabbitMQ and rework it :)

PR #14133, opened 2022-12-29
[bitnami/redis] Sentinel deployment: Fix Failover at graceful shutdown

### LINKED ISSUE #13956: syntax error in bitnami/redis/templates/scripts-configmap.yaml
Hi,

while debugging an issue with the Redis helm package where the pods keep crashing, I found a syntax error in bitnami/redis/templates/scripts-configmap.yaml

there is a missing "; then" after the "if retry_while "failover_finished" ...

It was introduced there: https://github.com/bitnami/charts/pull/13021/commits/aac3cd3af56728f4a47f9d9b026d0bdd6a935e14  (line 500)

_Originally posted by @ftagada in https://github.com/bitnami/charts/issues/13021#issuecomment-1346264529_

gmbnomis:
### Description of the change

Achieving a proper failover at graceful shutdown of the redis master pod is tricky (see #13021 and #12598 for discussions of different options).

The currently implemented solution does not work as intended because of a syntax error (see #13956). After fixing the syntax error, it turns out that the intended solution does not work, as it issues a "SENTINEL FAILOVER" as well as a Redis "FAILOVER". These failover commands are unrelated and were not designed to work together. Problems are for example:

1. Redis "FAILOVER" and "SENTINEL FAILOVER" may select different new master nodes.

2. As part of Redis "FAILOVER", the master will wait for lagging replicas. But while waiting, the replicas may "disappear" because the sentinel failover procedure is reconfiguring them. This causes the redis failover procedure to hang until it times out.

In summary, the Redis "FAILOVER" command can't be used in this context.

Instead, implement the following when the redis master pod shuts down:

1. Issue a "CLIENT PAUSE" command on the Redis master. This will block all writes. Reads can still happen.

2. Trigger a failover to a different master by issuing a "SENTINEL FAILOVER" command.

3. The shutdown of the sentinel container is delayed until the sentinel returns a different master (i.e. the failover has succeeded from a client's perspective)

   Optionally (the default), also delay the shutdown of the redis container until that point.

The delayed shutdown of the Redis container maximizes the availability of Redis during the failover (at least for reads). However, as discussed in
https://github.com/bitnami/charts/pull/13021#issuecomment-1289177976, it has drawbacks when short timeouts are used by clients and can therefore be switched off by setting
`sentinel.redisShutdownWaitFailover=false`.

### Benefits

Implement proper failover when master pod is shut down. Depending on the need of the application, the failover may be tuned to favor read availability or favor read consistency.

### Possible drawbacks

Details of this tuning and its consequences may not be easy to understand. But there seems to be no "one size fits all" solution.

### Applicable issues

  - fixes #13956

### Additional information

Note: Because of the syntax error in the previous code, its behavior actually corresponds to the behavior implemented by this commit when `sentinel.redisShutdownWaitFailover=False`.

### Checklist


fmulero (dismissed):
Brilliant!!! Thanks a lot @gmbnomis!. As you said probably doesn't fit all the scenarios but  I've running some tests and everything looks good (and better than it is right now).

PR #14134, opened 2022-12-29
bitnami/keycloak: Document possible use of MSSQL as an external DB

samm-git:
I was able to run KC with a MSSQL 2019 and the bitnami Keycloak chart, so decided to add this to the documentation. 

Proposed works because KeyCloak gives preferences to the environment compared to the statements in the configuration file. I think it could be useful for other chart users.

Signed-off-by: Alex Samorukov <samm@net-art.cz>

### Description of the change

Update Keycloak chart readme to document possible use of MSSQL as an external DB

### Checklist

### DIFF
--- a/bitnami/keycloak/README.md
+++ b/bitnami/keycloak/README.md
@@ -387,6 +387,12 @@ Refer to the [chart documentation on using an external database](https://docs.bi
 It is not supported but possible to run Keycloak with an external MSSQL database with the following settings:
 
 ```yaml
+externalDatabase:
+  host: "mssql.example.com"
+  port: 1433
+  user: keycloak
+  database: keycloak
+  existingSecret: passwords
 extraEnvVars:
   - name: KC_DB # override values from the conf file
     value: 'mssql'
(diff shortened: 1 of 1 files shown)


rafariossaa:
Hi,
  Thanks forehand, but I think we should not add this to the README, because as you said, this is not official. They could remove the mysql connector in any moment and leave this broken.
  However, other users could find this PR thread and use it if they want to take the risk.

samm-git:
Its MS-SQL here, not MySQL. I do not think that there are any plans to remove MSSQL support, it is not marked as depricated and bugs are being addressed. It is only limitation of the helm, not of the KC itself.

samm-git:
(so may be i was not clear enough. it is not official in the helm chart, not in the KC upstream)

rafariossaa (approved):
LGTM

rafariossaa:
> Its MS-SQL here, not MySQL. I do not think that there are any plans to remove MSSQL support, it is not marked as depricated and bugs are being addressed. It is only limitation of the helm, not of the KC itself.

Sorry, My mistake .

PR #14144, opened 2022-12-30
[bitnami/zookeeper] Configure headless service name

fabien-marchand:
A new configuration parameter `service.headless.name` now allows to override the default name given to the headless service.

Signed-off-by: Fabien Marchand <fabien.marchand@intersec.com>

### Applicable issues

  - fixes #

### Checklist

### DIFF
--- a/bitnami/zookeeper/Chart.yaml
+++ b/bitnami/zookeeper/Chart.yaml
@@ -22,4 +22,4 @@ name: zookeeper
 sources:
   - https://github.com/bitnami/containers/tree/main/bitnami/zookeeper
   - https://zookeeper.apache.org/
-version: 11.0.3
+version: 11.1.0
--- a/bitnami/zookeeper/README.md
+++ b/bitnami/zookeeper/README.md
@@ -216,7 +216,7 @@ The command removes all the Kubernetes components associated with the chart and 
 | `service.extraPorts`                        | Extra ports to expose in the ZooKeeper service (normally used with the `sidecar` value) | `[]`        |
 | `service.headless.annotations`              | Annotations for the Headless Service                                                    | `{}`        |
 | `service.headless.publishNotReadyAddresses` | If the ZooKeeper headless service should publish DNS records for not ready pods         | `true`      |
-| `service.headless.name`                     | Overrides the name given to the headless service, if set                                | `""`        |
+| `service.headless.servicenameOverride`      | String to partially override headless service name                                      | `""`        |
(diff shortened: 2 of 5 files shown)


bitnami/zookeeper/Chart.yaml
```diff
@@ -21,4 +21,4 @@ name: zookeeper
 sources:
   - https://github.com/bitnami/containers/tree/main/bitnami/zookeeper
   - https://zookeeper.apache.org/
-version: 11.0.2
+version: 11.0.3
```
jotamartos, line 24: As you are adding a new parameter, please bump the version to a new minor (`11.1.0`)

jotamartos (changes requested):
Thank you for your  contribution. However, I'd like you to take a look at how other charts were configured to apply the same changes here. You can take a look at RabbitMQ as an example 

https://github.com/bitnami/charts/blob/main/bitnami/rabbitmq/values.yaml#L80

fabien-marchand:
> Thank you for your contribution. However, I'd like you to take a look at how other charts were configured to apply the same changes here. You can take a look at RabbitMQ as an example
> 
> https://github.com/bitnami/charts/blob/main/bitnami/rabbitmq/values.yaml#L80

Thanks for the review. I re-pushed but did not exactly as it is done in "rabbitmq" (the service name is not a full override as you have a `%s-%s` constraint, which I d

jotamartos:
Please resolve the conflict so all tests pass. Thanks!

jotamartos:
Hi @fabien-marchand,

Did you take a look at the comment I posted above? 

https://github.com/bitnami/charts/pull/14144#pullrequestreview-1242134542

I suggested you use the "servicenameOverride" variable as others solution use.

fabien-marchand:
> servicenameOverride

Hi @jotamartos ,
Yes, sorry I did not undertstand you were also talking about setting the exact same name for the option. I just repushed that way, though the name `service.headless.servicenameOverride` seems a bit redundant to me and not really exact as the name is not fully overriden.

jotamartos:
Hi again @fabien-marchand,

Sorry if the info was not clear enough. I asked you to implement these changes in the bitnami/zookeeper chart. Can you update the PR to be like this other one?

https://github.com/bitnami/charts/commit/e963f17318085b18eb87f88d44415da7ecd4f3d0

Thanks

jotamartos:
>though the name service.headless.servicenameOverride seems a bit redundant to me and not really exact as the name is not fully overriden.

I see your point of view but I'd like to keep it in sync with other charts we have in which we also use that variable to partially change the name of the service.

jotamartos (approved):

PR #14163, opened 2023-01-03
[bitnami/redis] Added commonLabels to volumeClaimTemplates

likeon:
Signed-off-by: Aleksandr Sterkhov <i@likeon.dev>

### Description of the change

`commonLabels` parameter supposed to add labels to all deployed objects, but PVCs created via volumeClaimTemplates did not have `commonLabels`. This PR fixes it.

### Benefits

 `commonLabels` added to volumeClaimTemplates.

### Checklist

### DIFF
--- a/bitnami/redis/Chart.yaml
+++ b/bitnami/redis/Chart.yaml
@@ -22,4 +22,4 @@ maintainers:
 name: redis
 sources:
   - https://github.com/bitnami/containers/tree/main/bitnami/redis
-version: 17.4.0
+version: 17.5.0
--- a/bitnami/redis/templates/master/application.yaml
+++ b/bitnami/redis/templates/master/application.yaml
@@ -462,6 +462,9 @@ spec:
         name: redis-data
         labels: {{- include "common.labels.matchLabels" . | nindent 10 }}
           app.kubernetes.io/component: master
+          {{- if .Values.commonLabels }}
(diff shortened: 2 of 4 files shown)


fmulero (changes requested):
Hi @likeon Thanks a lot for your contribution!!

I am not sure if we can accept this change because some users put dynamic content in `commonLabels` and/or `commonAnnotations`. In that case it would brake the "immutability" of the persistence claim and other users wouldn't be able to upgrade their deployments. I am going to open an internal discussion about this topic because I recognize that we have a bit of mess about it

likeon:
@fmulero I understand your concern here, but please bare in mind that currently there is no way to apply user-defined labels to PVCs. `commonLabels` seems like a logical place to provide it, but a new variable could work as well.

fmulero:
> @fmulero I understand your concern here, but please bare in mind that currently there is no way to apply user-defined labels to PVCs. `commonLabels` seems like a logical place to provide it, but a new variable could work as well.

In [kafka for example](https://github.com/bitnami/charts/blob/2814eed/bitnami/kafka/templates/statefulset.yaml#L571-L578) (and you can find it also in other charts: thanos, postgresql, zookeeper, etc), there is a `persistence.labels` value to allow users set particular labels in the PVCs. I think this solution will fit your needs.

Issue #14208, opened 2023-01-06
[bitnami/oauth2-proxy] Google Provider issue


vvdaal:
### Name and Version

bitnami/oauth2-proxy 3.4.2

### What steps will reproduce the bug?

* Deploy chart with Google provider and a JSON
* Pod will never start due to the JSON file not being mounted as the secret is referenced incorrectly

### What is the expected behavior?

Secret name should be appended with -google as that is how the secret is created

### What do you see instead?

Secret name does not have -google appended and thus can't be referenced

vvdaal:
Proposed fix https://github.com/bitnami/charts/pull/14210

carrodher:
Thanks for creating this issue and the associated PR, the team will review it and provide feedback

PR #14210, opened 2023-01-06
[bitnami/oauth2-proxy] Fix Google secret reference when using Google provider

vvdaal:
### Description of the change

Fixes the reference to the google secret. By default now the name of the actual secret created has `-google` appended. However, in the chart it doesn't reference it like this, causing the service account JSON which is stored in this secret to never work.

### Benefits

Fixes a bug https://github.com/bitnami/charts/issues/14208

### Possible drawbacks

Can't think of any but not an expert on this chart

### Applicable issues

  - fixes https://github.com/bitnami/charts/issues/14208

### Checklist


corico44 (approved):
LGTM

PR #14265, opened 2023-01-10
Bug/discourse remove redis username

### LINKED ISSUE #14264: [bitnami/discourse] Using external redis authentication with username is not implemented
### Name and Version

bitnami/discourse 9.0.6

1. Configure an external redis service 
2. Create a credentials providing username and password. Like:
```
AUTH user1 pass1
```
3. Configure the Helm chart accordingly to enable the external redis
4. The sidekiq container starts restarting again and again showing this stacktrace
```
sidekiq 2023-01-10T21:06:04.064Z pid=1 tid=chh WARN: WRONGPASS invalid username-password pair or user is disabled. subscribe failed, reconnecting in 1 second. Call stack /opt/bitn │
│ ami/discourse/vendor/bundle/ruby/2.7.0/gems/redis-4.5.1/lib/redis/client.rb:163:in `call'         
```

I've been debugging 

```
    redis:
      enabled: false
    externalRedis:
      host: redis-ha.redis.svc.cluster.local
      port: 6379
      username: user1
      password: pass1
``` 

I would expect from discourse to be able to configure as well a username. But :
1. The Helm Chart parameter `externalRedis.username` is only present in the README.md and values file, but not used anywehere, so it's useless. 
2. The [docker image](https://github.com/bitnami/containers/blob/main/bitnami/discourse/2/debian-11/rootfs/opt/bitnami/scripts/libdiscourse.sh#L255) does not implement either this parameter
3. The [discourse code](https://github.com/discourse/discourse/blob/main/app/models/global_setting.rb#L200) does not implement either to configure a username for Redis, just a password

### What do you see instead?

An error log that provokes the container to not start successfully, and then it restarts again and again.

### Additional information

I would suggest to remove it from the Helm Chart as it's none sense.

ausias-armesto:
Fixes #14264

### DIFF
--- a/bitnami/discourse/Chart.yaml
+++ b/bitnami/discourse/Chart.yaml
@@ -1,7 +1,7 @@
 annotations:
   category: Forum
 apiVersion: v2
-appVersion: 2.8.13
+appVersion: 2.8.14
 dependencies:
   - condition: redis.enabled
     name: redis
--- a/bitnami/discourse/Chart.yaml
+++ b/bitnami/discourse/Chart.yaml
@@ -1,7 +1,7 @@
 annotations:
   category: Forum
 apiVersion: v2
-appVersion: 2.8.14
+appVersion: 2.8.13
 dependencies:
   - condition: redis.enabled
     name: redis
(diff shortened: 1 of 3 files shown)


migruiz4 (approved):
Thank you for your contribution @ausias-armesto.

The value `externalRedis.username` is not being referenced in the chart, so I agree with removing it.

At the moment, Discourse does not yet support setting a custom Redis username: https://github.com/discourse/discourse/blob/v2.8.13/config/discourse_defaults.conf

PR #14293, opened 2023-01-11
[bitnami/*] Add licenses annotation and remove obsolete engine parameter

carrodher:
This PR adds the `licenses` annotations as a list of [SPDX licenses](https://spdx.org/licenses/):
```yaml
annotations: 
  licenses: |
    - Apache-2.0
```

In the same way, it removes the `engine` parameter that was optional in Helm 2 (https://v2.helm.sh/docs/charts/#the-chart-yaml-file) but removed in Helm 3 (https://helm.sh/docs/topics/charts/#the-chartyaml-file), see https://github.com/helm/helm/pull/5260

-----------------

PS: CI/CD failure is expected since there is more than one chart included in the same PR

### DIFF
--- a/bitnami/airflow/Chart.yaml
+++ b/bitnami/airflow/Chart.yaml
@@ -1,5 +1,7 @@
 annotations:
   category: WorkFlow
+  licenses: |
+    - Apache-2.0
 apiVersion: v2
 appVersion: 2.5.0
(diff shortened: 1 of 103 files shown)


FraPazGal (approved):
LGTM

PR #14296, opened 2023-01-11
[bitnami/*] Update serviceAccount Goss' test

FraPazGal:

### Description of the change

Until now, our Goss test that checks the mounted serviceAccounts only makes sure that the related directory is mounted. The issue with this is that Kubernetes will always mount the SA metadata into every pod, be it the one created by default by the cluster or the one created by the chart itself.

That makes this test useless, as there will always exist the `/var/run/secrets/kubernetes.io/serviceaccount` directory. The revamp of the test now checks for the mounted SA name and compares it to the app's name via its env.

Reviewing the affected tests, it has been found 2 instances where the test wasn't necessary:

- The `bitnami/wavefront` deployment of the tested pod (wavefront-proxy) does not use the custom SA
- The SAs of the `bitnami/jaeger` chart are not used. An internal task has been created to address this.
- Both `bitnami/keycloak` and `bitnami/schema-registry` runtime parameters and vars were updated to use SA-related params.

### Benefits

- The test now works as expected.

### Possible drawbacks

- There may be charts where the SA name is not the same as the `BITNAMI_APP_NAME`. See: `bitnami/parse` changes.
- Added complexity to the test logic.

### Additional information

Change tested at https://github.com/FraPazGal/charts/pull/140

### Checklist

### DIFF
--- a/.vib/appsmith/goss/goss.yaml
+++ b/.vib/appsmith/goss/goss.yaml
@@ -1,16 +1,3 @@
-command:
-  {{- $uid := .Vars.client.containerSecurityContext.runAsUser }}
-  {{- $gid := .Vars.client.podSecurityContext.fsGroup }}
-  check-user-info:
-    # The UID and GID should always be either the one specified as vars (always a bigger number that the default)
-    # or the one randomly defined by openshift (larger values). Otherwise, the chart is still using the default value.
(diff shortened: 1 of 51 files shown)


joancafom (approved):
LGTM!

agomezmoron (approved):

joancafom (approved):

PR #14396, opened 2023-01-17
[bitnami/thanos] Improvements in param description

### LINKED ISSUE #14262: [thanos] Client/server TLS certificate keys not compatible with certs issued by cert-manager
### Name and Version

bitnami/thanos 11.6.4

Ref: https://github.com/bitnami/charts/blob/d5577f41e99f7e7ec07d3131a9f9d829c53a0a92/bitnami/thanos/templates/query/deployment.yaml#L141-L151

The way the chart is structured, we can't use Secrets created by `cert-manager` in the `(query|storegateway).grpc.(server|client).tls.existingSecret` parameters. The current chart version assume the CA, cert and key files are exposed through the `ca-cert`, `tls-cert` and `tls-key` attributes, instead of the standard `ca.crt`, `tls.crt` and `tls.key` attributes for Secrets of type `kubernetes.io/tls`.

This chart should use the same convention for compatibility with other projects, such as `cert-manager`.

We need a way to specify the correct file path for the client/server secrets, in a way that we can reference the Secret fields `ca.crt`, `tls.crt` and `tls.key`, which are the keys populated by `cert-manager`.

I managed to make this work by disabling the client/server TLS configuration in the chart, mounting the certificate secrets via the `volumes` and `volumeMounts` settings, and providing the TLS flags manually.

rafariossaa:
### Description of the change

State explicitly that you may need to use the `keyMapping` when using an existing secret to map the files.

### Applicable issues

  - fixes #14262

### Checklist

- [] Variables are documented in the values.yaml and added to the `README.md` using [readme-generator-for-helm](https://github.com/bitnami-labs/readme-generator-for-helm)


bitnami/thanos/values.yaml
```diff
@@ -408,7 +408,7 @@ query:
       ## @param query.grpc.server.tls.cert TLS Certificate for GRPC server - ignored if existingSecret is provided
       ## @param query.grpc.server.tls.key TLS Key for GRPC server - ignored if existingSecret is provided
       ## @param query.grpc.server.tls.ca TLS CA to verify clients against - ignored if existingSecret is provided
-      ## @param query.grpc.server.tls.existingSecret Existing secret containing your own TLS certificates
+      ## @param query.grpc.server.tls.existingSecret Existing secret containing your own TLS certificates. In the case that they are not named ca-cert, tls-cert and tls-key you will need to use the keyMapping accordingly.
```
danielfm, line 411: Maybe also include a note to tell the user to check for an example in the `values.yaml` file?

For instance, if the user is reading the chart documentation in ArtifactHub, these example comments are not exposed via the main documentation, so being explicit there's an example available in `values.yaml` might help.

PR #14404, opened 2023-01-17
[bitnami/mariadb] Allow templates for initdbScriptsConfigMap and auth.existingSecret

FabianScheidt:
### Description of the change

I noticed that `initdbScriptsConfigMap` and `auth.existingSecret` don't allow us to use template strings. I don't see a reason why this shouldn't be allowed, so this change will apply Helm's `tpl` function to both of them.

The same setup is already used for `primary.existingConfigmap` and `secondary.existingConfigmap`.

### Benefits

This will allow us to configure the following, which currently causes an error:

```yaml
auth:
    database: my-database
    username: my-user
    existingSecret: "{{ .Release.Name }}-mariadb-secret"
initdbScriptsConfigMap: "{{ .Release.Name }}-mariadb-init-config"
```

Without this change we would be unable to e.g. include the release name in the secret or config map.

### Possible drawbacks

No known drawbacks.

### Applicable issues

Not reported yet.

### Additional information

\-

### Checklist

### DIFF
--- a/bitnami/mariadb/Chart.yaml
+++ b/bitnami/mariadb/Chart.yaml
@@ -26,4 +26,4 @@ sources:
   - https://github.com/bitnami/containers/tree/main/bitnami/mariadb
   - https://github.com/prometheus/mysqld_exporter
   - https://mariadb.org
-version: 11.4.3
+version: 11.4.4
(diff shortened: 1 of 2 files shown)


CeliaGMqrz (approved):
Thanks for your contribution! LGTM!

PR #14399, opened 2023-01-17
[bitnami/redis] make emptyDir sizeLimit work everywhere

monotek:
Signed-off-by: André Bauer <andre.bauer@staffbase.com>

### Description of the change

* make emptyDir sizeLimit work everywhere

### Benefits

* before the change sizeLimit could only be used if `persistence.medium` was set too (now both options can be set independently)
* sentinel config was missing sizeLimit config option completly

### Applicable issues

  - fixes #

### Checklist

- [x] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/). This is *not necessary* when the changes only affect README.md files.
- [x] Variables are documented in the values.yaml and added to the `README.md` using [readme-generator-for-helm](https://github.com/bitnami-labs/readme-generator-for-helm)
- [x] Title of the pull request follows this pattern [bitnami/<name_of_the_chart>] Descriptive title
- [x] All commits signed off and in agreement of [Developer Certificate of Origin (DCO)](https://github.com/bitnami/charts/blob/main/CONTRIBUTING.md#sign-your-work)

PR #14420, opened 2023-01-18
[bitnami/redis] add probes for metrics container

monotek:
Signed-off-by: André Bauer <andre.bauer@staffbase.com>

### Description of the change

* add probes for metrics container

### Benefits

* startupProbe, livenessProbe & readinessProbe can be defined for metrics container
* default is disabled so current behaviour is not changed
* bumped version to 17.6.0 because there is a pr for 17.5.0 already: https://github.com/bitnami/charts/pull/14399

### Applicable issues

  - fixes #

### Checklist

### DIFF
--- a/bitnami/redis/templates/master/application.yaml
+++ b/bitnami/redis/templates/master/application.yaml
@@ -311,11 +311,6 @@ spec:
           startupProbe: {{- include "common.tplvalues.render" (dict "value" .Values.metrics.customStartupProbe "context" $) | nindent 12 }}
           {{- else if .Values.metrics.startupProbe.enabled }}
           startupProbe: {{- include "common.tplvalues.render" (dict "value" (omit .Values.metrics.startupProbe "enabled") "context" $) | nindent 12 }}
-            initialDelaySeconds: {{ .Values.metrics.startupProbe.initialDelaySeconds }}
-            periodSeconds: {{ .Values.metrics.startupProbe.periodSeconds }}
-            timeoutSeconds: {{ .Values.metrics.startupProbe.timeoutSeconds }}
-            successThreshold: {{ .Values.metrics.startupProbe.successThreshold }}
-            failureThreshold: {{ .Values.metrics.startupProbe.failureThreshold }}
             tcpSocket:
               port: metrics
           {{- end }}
--- a/bitnami/redis/templates/master/application.yaml
+++ b/bitnami/redis/templates/master/application.yaml
@@ -303,11 +303,23 @@ spec:
             {{- if .Values.metrics.extraEnvVars }}
             {{- include "common.tplvalues.render" (dict "value" .Values.metrics.extraEnvVars "context" $) | nindent 12 }}
             {{- end }}
+          {{- with .Values.metrics.livenessProbe }}
+          livenessProbe:
+            {{- toYaml . | nindent 12 }}
+          {{- end }}
           ports:
             - name: metrics
               containerPort: 9121
+          {{- with .Values.metrics.readinessProbe }}
+          readinessProbe:
+            {{- toYaml . | nindent 12 }}
+          {{- end }}
           {{- if .Values.metrics.resources }}
           resources: {{- toYaml .Values.metrics.resources | nindent 12 }}
+          {{- end }}
+          {{- with .Values.metrics.startupProbe }}
+          startupProbe:
+            {{- toYaml . | nindent 12 }}
           {{- end }}
           volumeMounts:
             {{- if .Values.auth.usePasswordFiles }}
--- a/bitnami/redis/templates/master/application.yaml
+++ b/bitnami/redis/templates/master/application.yaml
@@ -318,11 +318,6 @@ spec:
           livenessProbe: {{- include "common.tplvalues.render" (dict "value" .Values.metrics.customLivenessProbe "context" $) | nindent 12 }}
           {{- else if .Values.metrics.livenessProbe.enabled }}
           livenessProbe:
-            initialDelaySeconds: {{ .Values.metrics.livenessProbe.initialDelaySeconds }}
-            periodSeconds: {{ .Values.metrics.livenessProbe.periodSeconds }}
-            timeoutSeconds: {{ .Values.metrics.livenessProbe.timeoutSeconds }}
-            successThreshold: {{ .Values.metrics.livenessProbe.successThreshold }}
-            failureThreshold: {{ .Values.metrics.livenessProbe.failureThreshold }}
             tcpSocket:
               port: metrics
           {{- end }}
@@ -330,11 +325,6 @@ spec:
           readinessProbe: {{- include "common.tplvalues.render" (dict "value" .Values.metrics.customReadinessProbe "context" $) | nindent 12 }}
           {{- else if .Values.metrics.readinessProbe.enabled }}
           readinessProbe:
-            initialDelaySeconds: {{ .Values.metrics.readinessProbe.initialDelaySeconds }}
-            periodSeconds: {{ .Values.metrics.readinessProbe.periodSeconds }}
-            timeoutSeconds: {{ .Values.metrics.readinessProbe.timeoutSeconds }}
-            successThreshold: {{ .Values.metrics.readinessProbe.successThreshold }}
-            failureThreshold: {{ .Values.metrics.readinessProbe.failureThreshold }}
             httpGet:
               path: /
               port: metrics
--- a/bitnami/redis/templates/master/application.yaml
+++ b/bitnami/redis/templates/master/application.yaml
@@ -303,23 +303,50 @@ spec:
             {{- if .Values.metrics.extraEnvVars }}
             {{- include "common.tplvalues.render" (dict "value" .Values.metrics.extraEnvVars "context" $) | nindent 12 }}
             {{- end }}
-          {{- with .Values.metrics.livenessProbe }}
-          livenessProbe:
-            {{- toYaml . | nindent 12 }}
-          {{- end }}
           ports:
             - name: metrics
               containerPort: 9121
-          {{- with .Values.metrics.readinessProbe }}
+          {{- if not .Values.diagnosticMode.enabled }}
+          {{- if .Values.metrics.customStartupProbe }}
+          startupProbe: {{- include "common.tplvalues.render" (dict "value" .Values.metrics.customStartupProbe "context" $) | nindent 12 }}
+          {{- else if .Values.metrics.startupProbe.enabled }}
+          startupProbe: {{- include "common.tplvalues.render" (dict "value" (omit .Values.metrics.startupProbe "enabled") "context" $) | nindent 12 }}
+            initialDelaySeconds: {{ .Values.metrics.startupProbe.initialDelaySeconds }}
+            periodSeconds: {{ .Values.metrics.startupProbe.periodSeconds }}
+            timeoutSeconds: {{ .Values.metrics.startupProbe.timeoutSeconds }}
+            successThreshold: {{ .Values.metrics.startupProbe.successThreshold }}
+            failureThreshold: {{ .Values.metrics.startupProbe.failureThreshold }}
+            tcpSocket:
+              port: metrics
+          {{- end }}
+          {{- if .Values.metrics.customLivenessProbe }}
+          livenessProbe: {{- include "common.tplvalues.render" (dict "value" .Values.metrics.customLivenessProbe "context" $) | nindent 12 }}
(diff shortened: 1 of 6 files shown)


migruiz4 (changes requested):
Hi @monotek,

Could you please adapt this PR to use the same format used for probes in the catalog?

In the Statefulsets:
```
          {{- if .Values.%%MAIN_OBJECT_BLOCK%%.customLivenessProbe }}
          livenessProbe: {{- include "common.tplvalues.render" (dict "value" .Values.%%MAIN_OBJECT_BLOCK%%.customLivenessProbe "context" $) | nindent 12 }}
          {{- else if .Values.%%MAIN_OBJECT_BLOCK%%.livenessProbe.enabled }}
          livenessProbe: {{- include "common.tplvalues.render" (dict "value" (omit .Values.%%MAIN_OBJECT_BLOCK%%.livenessProbe "enabled") "context" $) | nindent 12 }}
            %%httpGet || command || etc%%
          {{- end }}
          {{- if .Values.%%MAIN_OBJECT_BLOCK%%.customReadinessProbe }}
          readinessProbe: {{- include "common.tplvalues.render" (dict "value" .Values.%%MAIN_OBJECT_BLOCK%%.customReadinessProbe "context" $) | nindent 12 }}
          {{- else if .Values.%%MAIN_OBJECT_BLOCK%%.readinessProbe.enabled }}
          readinessProbe: {{- include "common.tplvalues.render" (dict "value" (omit .Values.%%MAIN_OBJECT_BLOCK%%.readinessProbe "enabled") "context" $) | nindent 12 }}
            %%httpGet || command || etc%%
          {{- end }}
          {{- if .Values.%%MAIN_OBJECT_BLOCK%%.customStartupProbe }}
          startupProbe: {{- include "common.tplvalues.render" (dict "value" .Values.%%MAIN_OBJECT_BLOCK%%.customStartupProbe "context" $) | nindent 12 }}
          {{- else if .Values.%%MAIN_OBJECT_BLOCK%%.startupProbe.enabled }}
          startupProbe: {{- include "common.tplvalues.render" (dict "value" (omit .Values.%%MAIN_OBJECT_BLOCK%%.startupProbe "enabled") "context" $) | nindent 12 }}
            %%httpGet || command || etc%%
          {{- end }}
          {{- end }}
```
In the values.yaml:
```yaml
  ## Configure extra options for %%MAIN_CONTAINER_NAME%% containers' liveness and readiness probes
  ## ref: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/#configure-probes
  ## @param %%MAIN_OBJECT_BLOCK%%.livenessProbe.enabled Enable livenessProbe on %%MAIN_CONTAINER_NAME%% containers
  ## @param %%MAIN_OBJECT_BLOCK%%.livenessProbe.initialDelaySeconds Initial delay seconds for livenessProbe
  ## @param %%MAIN_OBJECT_BLOCK%%.livenessProbe.periodSeconds Period seconds for livenessProbe
  ## @param %%MAIN_OBJECT_BLOCK%%.livenessProbe.timeoutSeconds Timeout seconds for livenessProbe
  ## @param %%MAIN_OBJECT_BLOCK%%.livenessProbe.failureThreshold Failure threshold for livenessProbe
  ## @param %%MAIN_OBJECT_BLOCK%%.livenessProbe.successThreshold Success threshold for livenessProbe
  ##
  livenessProbe:
    enabled: true
    initialDelaySeconds: foo
    periodSeconds: bar
    timeoutSeconds: foo
    failureThreshold: bar
    successThreshold: foo
  ## @param %%MAIN_OBJECT_BLOCK%%.readinessProbe.enabled Enable readinessProbe on %%MAIN_CONTAINER_NAME%% containers
  ## @param %%MAIN_OBJECT_BLOCK%%.readinessProbe.initialDelaySeconds Initial delay seconds for readinessProbe
  ## @param %%MAIN_OBJECT_BLOCK%%.readinessProbe.periodSeconds Period seconds for readinessProbe
  ## @param %%MAIN_OBJECT_BLOCK%%.readinessProbe.timeoutSeconds Timeout seconds for readinessProbe
  ## @param %%MAIN_OBJECT_BLOCK%%.readinessProbe.failureThreshold Failure threshold for readinessProbe
  ## @param %%MAIN_OBJECT_BLOCK%%.readinessProbe.successThreshold Success threshold for readinessProbe
  ##
  readinessProbe:
    enabled: true
    initialDelaySeconds: foo
    periodSeconds: bar
    timeoutSeconds: foo
    failureThreshold: bar
    successThreshold: foo
  ## @param %%MAIN_OBJECT_BLOCK%%.startupProbe.enabled Enable startupProbe on %%MAIN_CONTAINER_NAME%% containers
  ## @param %%MAIN_OBJECT_BLOCK%%.startupProbe.initialDelaySeconds Initial delay seconds for startupProbe
  ## @param %%MAIN_OBJECT_BLOCK%%.startupProbe.periodSeconds Period seconds for startupProbe
  ## @param %%MAIN_OBJECT_BLOCK%%.startupProbe.timeoutSeconds Timeout seconds for startupProbe
  ## @param %%MAIN_OBJECT_BLOCK%%.startupProbe.failureThreshold Failure threshold for startupProbe
  ## @param %%MAIN_OBJECT_BLOCK%%.startupProbe.successThreshold Success threshold for startupProbe
  ##
  startupProbe:
    enabled: false
    initialDelaySeconds: foo
    periodSeconds: bar
    timeoutSeconds: foo
    failureThreshold: bar
    successThreshold: foo
  ## @param %%MAIN_OBJECT_BLOCK%%.customLivenessProbe Custom livenessProbe that overrides the default one
  ##
  customLivenessProbe: {}
  ## @param %%MAIN_OBJECT_BLOCK%%.customReadinessProbe Custom readinessProbe that overrides the default one
  ##
  customReadinessProbe: {}
  ## @param %%MAIN_OBJECT_BLOCK%%.customStartupProbe Custom startupProbe that overrides the default one
  ##
  customStartupProbe: {}
```

See https://github.com/bitnami/charts/tree/main/template/CHART_NAME/templates for more examples.

Additionally, you skipped version `17.5.0` in the Chart.yaml

monotek:
I thought about that too but guessed that this implementation just came from the fact that the probes were not configurable in the early days of the chart and that is just an ugly hack to fix that?

If it's still wanted that way i'll do it but imho the current way is cleaner.

The thing about the version was mentioned in the description already. Did it as i have another pr for 17.5.0 open. See:  https://github.com/bitnami/charts/pull/14399

migruiz4:
> I thought about that too but guessed that this implementation just came from the fact that the probes were not configurable in the early days of the chart and that this implementation is just a hack to fix that?

We do it this way for several reasons:

The `readinessProbe` is enabled by default and its probe command is hard-coded on purpose. This way, we provide an easier way to configure Probes, and we can ensure the probe is widely tested by both our CI/CD pipeline and other users.
The most difficult part is setting the Probe default timing values to fit every use case, that is why we always try to provide default values that could fit most of the cases, but also give users the chance to fine-tune the Probe settings.

For users who'd like to use a different command or omit our default values, we provide `customReadinessProbe` with the same functionality you proposed here. 
It also helps workaround an issue where Helm will not allow removing default parameters from object type values.

In case you only want to contribute to that functionality, I will need to ask you to rename `(.*)Probe` to `custom(.*)Probe`, so it doesn't interfere in case we would like to implement default probes in the future.

bitnami/redis/templates/master/application.yaml
```diff
@@ -306,6 +306,45 @@ spec:
...
+          {{- if not .Values.diagnosticMode.enabled }}
+          {{- if .Values.metrics.customStartupProbe }}
+          startupProbe: {{- include "common.tplvalues.render" (dict "value" .Values.metrics.customStartupProbe "context" $) | nindent 12 }}
+          {{- else if .Values.metrics.startupProbe.enabled }}
+          startupProbe: {{- include "common.tplvalues.render" (dict "value" (omit .Values.metrics.startupProbe "enabled") "context" $) | nindent 12 }}
+            initialDelaySeconds: {{ .Values.metrics.startupProbe.initialDelaySeconds }}
+            periodSeconds: {{ .Values.metrics.startupProbe.periodSeconds }}
+            timeoutSeconds: {{ .Values.metrics.startupProbe.timeoutSeconds }}
+            successThreshold: {{ .Values.metrics.startupProbe.successThreshold }}
+            failureThreshold: {{ .Values.metrics.startupProbe.failureThreshold }}
```
migruiz4, line 318: Thank you for applying the template changes.

These lines can be omitted, as they will be added by the render command in line 313:
```
{{- include "common.tplvalues.render" (dict "value" (omit .Values.metrics.startupProbe "enabled") "context" $) | nindent 12 }}
```

migruiz4: I'm sorry I forgot to mention this applies to all `startup`, `readiness` and `liveness` probes

bitnami/redis/values.yaml
```diff
@@ -1412,6 +1412,63 @@ metrics:
     ##   - myRegistryKeySecretName
     ##
     pullSecrets: []
+  ## @param metrics.livenessProbe define livenessProbe for metrics container
+  ##
+  ## @param metrics.livenessProbe define livenessProbe for metrics container
+  ##
```
migruiz4, line 1418: These lines are causing the CI/CD to fail when attempting to autogenerate the README.md tables, could you please remove them?

bitnami/redis/values.yaml
```diff
@@ -1412,6 +1412,59 @@ metrics:
...
+  ## ref: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/
+  ## @param metrics.startupProbe.enabled Enable startupProbe on Redis&reg; replicas nodes
+  ## @param metrics.startupProbe.initialDelaySeconds Initial delay seconds for startupProbe
+  ## @param metrics.startupProbe.periodSeconds Period seconds for startupProbe
+  ## @param metrics.startupProbe.timeoutSeconds Timeout seconds for startupProbe
+  ## @param metrics.startupProbe.failureThreshold Failure threshold for startupProbe
+  ## @param metrics.startupProbe.successThreshold Success threshold for startupProbe
+  ##
+  startupProbe:
+    enabled: true
```
migruiz4, line 1425: Please set `startupProbe.enabled=false`.
For backward-compatibility reasons with older Kubernetes versions, we don't enable startupProbes by default.

```suggestion
    enabled: false
```

bitnami/redis/values.yaml
```diff
@@ -1412,6 +1412,59 @@ metrics:
...
+  ## @param metrics.readinessProbe.failureThreshold Failure threshold for readinessProbe
+  ## @param metrics.readinessProbe.successThreshold Success threshold for readinessProbe
+  ##
+  readinessProbe:
+    enabled: true
+    initialDelaySeconds: 5
+    periodSeconds: 10
+    timeoutSeconds: 1
+    successThreshold: 1
+    failureThreshold: 3
```
migruiz4, line 1458: Default values for the readiness probe may be too strict.

Its `initialDelaySeconds` is shorter than the livenessProbe's initial delay, and the 1 second timeout may be too short too.

I would suggest, at least, 15 seconds for initialDelaySeconds and 5 seconds timeout.

migruiz4 (changes requested):
Hi @monotek,

Please take a look at my suggestion. Apart from that, your contribution looks great to me.

bitnami/redis/values.yaml
monotek: Can you please elaborate why?

The kubernetes default for initialDelaySeconds is "0". So "5s" should already giving the exporter enough time to start and be able to receive traffic.

Imho initialDelaySeconds of the readinessprobe should be independent from the initialDelaySeconds setting of the livenessprobe. 

Livenessprobe would fail after "initialDelaySeconds + (periodSeconds * failureThreshold)" = 60s.
Readinessprobe would fail after: "initialDelaySeconds + (periodSeconds * failureThreshold)" = 35s.

So having the readinessprobe failing before the livenessprobe sounds reasonable to me.

The 1 second timeout is also the k8s default and it should also be enough to render a mostly empty html page, as metrics are not part of it. If thats not possible there is likely already a problem and guess i don't want to send more traffic there.

migruiz4: I'm sorry, you are absolutely right. It is the liveness that should be performed after the readiness.

With that, the PR looks good to me, thank you very much for your contribution!

migruiz4 (approved):
Looks good to me, thank you very much for your contribution and for taking the time to apply the suggestions!

PR #14433, opened 2023-01-18
[bitnami/template] PodDisruptionBudget template now works alongside autoscaling

### LINKED ISSUE #14407: [bitnami/template] Enabling PodDisruptionBudget requires setting .Values.replicaCount > 1, even if using HorizontalPodAutoscaler
### Name and Version

bitnami/template

1. Clone repository
2. Under `template/CHART_NAME`, replace all placeholder markers (i.e. `%%COMPONENT_NAME%%`) with valid variables to enable Helm templating. You can remove things such as the subchart template in `Chart.yaml` .
3. Run `helm dependency update template/CHART_NAME`
4. Run `helm template template/CHART_NAME --set pdb.create=true --set autoscaling.enabled=true`, you can see that the PodDisruptionBudget will not be applied
5. Run `helm template template/CHART_NAME --set pdb.create=true --set autoscaling.enabled=true --set replicaCount=2` (replicaCount can have any value greater than 1), the PodDisruptionBudget template will be created.

pdb.create = true
autoscaling.enabled = true
replicaCount = 2 (this should _not_ need to be set)

If I set `.Values.autoscaling.enabled = true` and `.Values.pdb.create = true`, I should be able to template a PodDisruptionBudget YAML generated, without touching `.Values.replicaCount` . HorizontalPodAutoscaler users should not need to care about this value.

If `.Values.autoscaling.enabled = false`, then in order to generate the PDB YAML I should have to set _both_ `.Values.pdb.create = true` and `.Values.replicaCount = {{ some integer > 1 }}`

### What do you see instead?

No PodDisruptionBudget generated in Helm templating.

### Additional information

There are two areas `.Values.replicaCount` will be used:

1) Using logic from `template/CHART_NAME/templates/deployment.yaml` and `template/CHART_NAME/templates/statefulset.yaml`, replicas are only set if autoscaling is not enabled. This is expected behavior.

2) In `template/CHART_NAME/templates/pdb.yaml`, the logic is checking for both `.Values.pdb.create = true` and `gt .Values.replicaCount 1` . The replicaCount size check should be skipped if `.Values.autoscaling.enabled = true`

Possible fix to be applied in `pdb.yaml`:

```
{{- if and .Values.pdb.create (or (gt $replicaCount 1) .Values.autoscaling.enabled) }}
```

Lance-Drane:
### Description of the change

The PodDisruptionBudget (PDB) template can now be generated alongside the HorizontalPodAutoscaling (HPA) template without illogically setting ".Values.replicaCount" to a value greater than 1.

PDB and HPA are both fairly generic API resources which can both be applied to a large number of applications, so it makes sense for the templates to assume that a chart may want to deploy both/one-of/neither of these resources.

### Benefits

Templates are more flexible

### Possible drawbacks

If a user only wants a PDB in their chart, and not a HPA, the PDB template requires additional modification.

### Applicable issues

- fixes #14407

### Additional information

I looked through the current Bitnami charts and didn't see any charts this would fix, but if the Kafka chart and the Zookeeper chart want to add a HorizontalPodAutoscaler template in the future then they should be sure to update their `poddisruptionbudget.yaml` or `pdb.yaml` files.

### Checklist

- [ X ] Title of the pull request follows this pattern [bitnami/<name_of_the_chart>] Descriptive title
- [ X ] All commits signed off and in agreement of [Developer Certificate of Origin (DCO)](https://github.com/bitnami/charts/blob/main/CONTRIBUTING.md#sign-your-work)

### DIFF
--- a/template/CHART_NAME/templates/pdb.yaml
+++ b/template/CHART_NAME/templates/pdb.yaml
@@ -1,5 +1,5 @@
 {{- $replicaCount := int .Values.%%MAIN_OBJECT_BLOCK%%.replicaCount }}
-{{- if and .Values.%%MAIN_OBJECT_BLOCK%%.pdb.create (gt $replicaCount 1) }}
+{{- if and .Values.%%MAIN_OBJECT_BLOCK%%.pdb.create (or (gt $replicaCount 1) .Values.%%MAIN_OBJECT_BLOCK%%.autoscaling.enabled) }}
 apiVersion: {{ include "common.capabilities.policy.apiVersion" . }}
 kind: PodDisruptionBudget
 metadata:


dgomezleon (approved):
LGTM

PR #14434, opened 2023-01-18
[bitnami/external-dns] Add container securityContext to comply with restricted PSS

selyx:
Signed-off-by: Andreas Grünenfelder <selyx@users.noreply.github.com>

### Description of the change

This PR adds the possibility to deploy the chart to a namespace with PodSecurityStandard _restricted_ enabled.

### Benefits

Before this PR chart couldn't be deployed to a namespace with the restricted PodSecurityStandard enabled even though no other permission was required.

### Checklist


fmulero (commented):
Thanks Andreas for your contribution.

I think this change is very focussed in your environment and I am not very sure about the real benefit for the rest of the users. Most of your changes can be done in your installation, setting your own `values.yml` file.

selyx:
Yes indeed, that's what we did already. I just thought it might be nice to provide this out of the box for every user since PSA is a kubernetes native feature and being compliant with a security profile doesn't hurt anybody.

fmulero:
Sorry @sylex for my late response.

I agree that security should be a must, but we have to prepare our charts for several use cases and allow the users customize them according to their needs. If we accept your changes, we will be enforcing seccomp by default (please keep in mind that we are supporting kubernetes >= 1.19 ) and it could break a lot of running deployments. For that reason we can not accept this change in a bugfix release.

PR #14451, opened 2023-01-19
[bitnami/mariadb] Forbid privileged containers by default

### LINKED ISSUE #13942: [bitnami/mariadb] Can't create/write to /opt/bitnami/mariadb/tmp whithout privilege escalation
### Name and Version

bitnami/mariadb 11.4.1

Launch a mariadb instance without privilege escalation allowed (I suppose that is the reason for the crashing, since that is what I've changed from the defaults)

```yaml
auth:
  ...

primary:
  resources:
    ...
  persistence:
    ...

  containerSecurityContext:
    enabled: true
    runAsNonRoot: true
    runAsUser: 1001
    runAsGroup: 1001
    privileged: false
```

Mariadb should be running fine without privilege escalation. It would be nice to have this set by default in all of the `containerSecurityContext`s

### What do you see instead?

The instance starts successfully in the first place (iirc), but fails as soon as it attempts to write to /opt/bitnami/mariadb/tmp and continues crash-looping from there

### Additional information

Since `primary.podSecurityContext.fsGroup` already defaults to `1001`, a possible solution is adding an `emtyDir` volume for this tmp directory to make it user-writable.

My current workaround:
```yaml
primary:
  # this tmp directory must be user-writable
  extraVolumes:
    - name: tmp
      emptyDir: {}
  extraVolumeMounts:
    - name: tmp
      mountPath: /opt/bitnami/mariadb/tmp
```

bauerjs1:
### Description of the change

ContainerSecurityContexts set the parameters `privileged` and `allowPrivilegeEscalation` to `false` by default.

### Benefits

Kubernetes Policy Engines typically enforce these parameters as a baseline standard and might reject Pods that do not have them set. This change implements those security standards.

### Possible drawbacks

Mariadb should not require privilege escalation, so there should be no drawbacks.

### Applicable issues

  - fixes #13942 

### Additional information

–

### Checklist


Mauraza (approved):
Hi @bauerjs1!!

Thanks for the contribution! Could you check the comment?

bitnami/mariadb/values.yaml
```diff
@@ -1089,7 +1099,9 @@ metrics:
   ##     readOnlyRootFilesystem: true
   ##
   containerSecurityContext:
-    enabled: false
+    enabled: true
```
Mauraza, line 1102: Could you explain why you change this to true?

Mauraza (changes requested):
Hi @bauerjs1!!

Thanks for the contribution! Could you check the comment?

bitnami/mariadb/values.yaml
bauerjs1: Hi @Mauraza!

I guess this was set to `false`, because the `containerSecurityContext` did not include any default parameters so far. Since this PR adds defaults, it should also apply them by enabling the `containerSecurityContext`. That's why I changed to `enabled=true`.

bauerjs1: Same as with the other containers, defaulting to `privileged=false` and `allowPrivilegeEscalation=false` should not impact the container in any way (I would be very suspicious if a metrics container needed privileged execution 😉) and just implements common security best-practices.

Mauraza: But you can keep enabled to false because if someone updates the chart to this version it will not change. 
Could you change it?

Mauraza (commented):
Hi @bauerjs1, 

could you change `containerSecurityContext.enable` to false?

bitnami/mariadb/values.yaml
bauerjs1: @Mauraza when the chart is updated, this will only change if people stick to the defaults and that's what this PR is actually all about. It is not a breaking change. But sure, I can revert this to disabled by default

Mauraza (approved):
Hi @bauerjs1,

Thanks again for the contribution! LGTM!

PR #14471, opened 2023-01-20
[bitnami/minio] Deleted all gateway features

corico44:
### Description of the change

MinIO has made a decision to deprecate its MinIO gateway feature, as can be seen in the following blogpost: https://blog.min.io/deprecation-of-the-minio-gateway/
They already deprecated Azure, GCS and HDFS in February: [minio/minio#14418](https://github.com/minio/minio/pull/14418)
And as of today, the only remaining compatible gateways are S3 and NAS: https://docs.min.io/docs/minio-gateway-for-s3.html
We are already receiving support cases of the chart gateway parameters not working, and we can expect more to come, so we are removing all gateway features.

### Checklist


bitnami/minio/values.yaml
```diff
@@ -2,6 +2,7 @@
 ## Global Docker image parameters
 ## Please, note that this will override the image parameters, including dependencies, configured to use the global value
 ## Current available global Docker image parameters: imageRegistry, imagePullSecrets and storageClass
+## Warning: All gateway features have been removed from Minio since upstream completely dropped this feature
```
dgomezleon, line 5: I think adding it to README.md is enough.

bitnami/minio/README.md
```diff
@@ -589,4 +528,8 @@ Unless required by applicable law or agreed to in writing, software
 distributed under the License is distributed on an "AS IS" BASIS,
 WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
-limitations under the License.
\ No newline at end of file
+limitations under the License.
+
+## Warning
+
+All gateway features have been removed from Minio since upstream completely dropped this feature.
```
dgomezleon, line 535: I think we can add it in `Upgrade` section. I cannot see another `Warning` section used for this kind of notification.

dgomezleon (changes requested):
The rest LGTM

bitnami/minio/README.md
carrodher: ➕1️⃣

bitnami/minio/values.yaml
carrodher: ➕1️⃣

carrodher (approved):
LGTM

bitnami/minio/README.md
```diff
@@ -516,6 +455,10 @@ Find more information about how to deal with common errors related to Bitnami's
 
 ## Upgrading
 
+### To 12.0.0
```
dgomezleon, line 458: The version is 12.1.0.

dgomezleon (changes requested):

dgomezleon (approved):

bitnami/minio/Chart.yaml
```diff
@@ -25,4 +25,4 @@ name: minio
 sources:
   - https://github.com/bitnami/containers/tree/main/bitnami/minio
   - https://min.io
-version: 12.0.3
+version: 12.1.0
```
migruiz4, line 28: I would like to open a discussion on the Chart version for this PR.

On one side, MinIO deprecated the Gateway feature some versions ago, so this probably won't affect users using the latest version of the `bitnami/minio` image.

Although, for those users using an older version of the image, this change could be considered non-backward reversible, which should be included in a major version.

@corico44 @carrodher @dgomezleon What do you think? Is it fine to release this in version `12.1.0` or should be instead released as `13.0.0`?

migruiz4 (approved):
Changes look good to me! I left a comment on the Chart version to discuss if the chart version should remain `12.1.0` or instead should be pushed to a major `13.0.0`.

bitnami/minio/Chart.yaml
carrodher: Sure, if backward compatibility is not guaranteed the version should be bumped in a major. I'm not familiar with this feature but if you think it could be non-backward compatible, feel free to bump the chart version in a major.

Being said that, please note bumping a major version implies bumping the major version as well in those Helm charts using MinIO as a subchart, in this case:
```console
$ ag '    name: minio' bitnami/*/Chart.yaml -l
bitnami/mastodon/Chart.yaml
bitnami/thanos/Chart.yaml
```

dgomezleon: But the feature was removed from 11.x.x to 12.x.x, we are only removing related variables. Am I wrong?

migruiz4: Not exactly. MinIO deprecated Gateway in version `2022-10-24`, meaning that the latest version where MinIO gateway works without changing the default image version would be `11.10.12` (https://github.com/bitnami/charts/pull/13132), and that only applies to 'nas' gateway, as support for other Gateway platforms have been progressively removed in even earlier versions of MinIO.

Although, users could still run the chart up to the latest version by providing an older MinIO tag to continue using gateway feature, although that may not ensure the chart would work properly.

dgomezleon: After the discussion, please we decided to go ahead with version 12.1.0, with a few clarifications in the README.md

bitnami/minio/README.md
```diff
@@ -516,6 +455,10 @@ Find more information about how to deal with common errors related to Bitnami's
 
 ## Upgrading
 
+### To 12.1.0
+
+All gateway features have been removed from Minio since upstream completely dropped this feature.
+
```
dgomezleon, line 461:
```suggestion
### To 12.0.0

This version updates MinIO&reg; to major version 2023.  All gateway features have been removed from Minio since upstream completely dropped this feature. The related options have been removed in version 12.1.0.

```

dgomezleon (changes requested):

dgomezleon (approved):

PR #14512, opened 2023-01-24
[bitnami/contour] Added option to override envoy-service-name and envoy-service-namespace.

kahirokunn:
Signed-off-by: kahirokunn <okinakahiro@gmail.com>

### Description of the change

I am using a multistage contour configuration because I am using a cilium cluster mesh.
I have three contours running in a cluster.
In such a configuration, you may want to change envoy-service-name and envoy-service-namespace freely.

### Benefits

envoy-service-name and envoy-service-namespace can be changed

### Possible drawbacks

None

### Applicable issues

  - fixes #

### Checklist


kahirokunn:
I'm ready to review.

kahirokunn:
The version of Chart in the main branch has been raised, so the version of Chart has been raised again.

bitnami/contour/Chart.yaml
```diff
@@ -25,4 +25,4 @@ sources:
   - https://github.com/envoyproxy/envoy
   - https://github.com/bitnami/containers/tree/main/bitnami/contour
   - https://projectcontour.io
-version: 10.1.5
+version: 10.1.6
```
jotamartos, line 28: As you are adding new parameters, please update the version to `10.2.0`.

jotamartos (changes requested):
Thanks for your contribution! Could you please take a look at the comment I posted and ping me again when applied? Thanks!

bitnami/contour/Chart.yaml
kahirokunn: Done.

kahirokunn:
@jotamartos Hello! I did it.

bitnami/contour/templates/contour/deployment.yaml
```diff
@@ -103,7 +103,8 @@ spec:
             - --contour-cert-file=/certs/tls.crt
             - --contour-key-file=/certs/tls.key
             - --config-path=/config/contour.yaml
-            - --envoy-service-name={{ printf "%s-envoy" (include "common.names.fullname" .) | trunc 63 | trimSuffix "-" }}
+            - --envoy-service-namespace={{ default .Release.Namespace .Values.contour.envoyServiceNamespace }}
+            - --envoy-service-name={{ default (printf "%s-envoy" (include "common.names.fullname" .) | trunc 63 | trimSuffix "-") .Values.contour.envoyServiceName }}
```
jotamartos, line 107: Taking a deeper look at these new parameters you added, I noticed that the chart is using different values. You need to use the same ones we are already configuring in the envoy service.yaml file 

https://github.com/bitnami/charts/blob/main/bitnami/contour/templates/envoy/service.yaml#L5

Those new parameters you added are not needed and the version should be moved back to 10.1.6

jotamartos (changes requested):
Sorry for not noticing this before. Those new parameters are not needed as the configuration  may not match with the created service.

bitnami/contour/templates/contour/deployment.yaml
kahirokunn: It is not necessary to specify a Service created in the envoy service.yaml file. In practice, users may create their own LoadBalancer Service or create multiple LoadBalancer Services. If only Services created in the envoy service.yaml file can be specified, users would have to make significant changes to their configuration. This PR is to avoid this issue and allow users to customize as they wish, as this feature is truly necessary for them.

PR #14671, opened 2023-01-30
[bitnami/postgresql-ha] Service annotations (#14523)

### LINKED ISSUE #14523: service.type=LoadBalancer can't be securely used with aws-load-balancer-controller
### Name and Version

bitnami/postgresql-ha:11.0.0

1. **Environment:** AWS EKS with aws-load-balancer-controller-1.4.6 installed. Installing `bitnami/postgresql-ha:11.0.0` chart.
2. **Config:** see below, file `/tmp/example.yaml`
3. **Run:** `helm install postgresql-ha bitnami/postgresql-ha -f /tmp/example.yaml`
4. **Error**: No error, but dangerous Service configurations are generated

```
service:
  type: LoadBalancer
  loadBalancerSourceRanges:
    - 1.2.3.4/32
    - 5.6.7.8/32
  externalTrafficPolicy: Local
  annotations:
    service.beta.kubernetes.io/aws-load-balancer-backend-protocol: tcp
    service.beta.kubernetes.io/aws-load-balancer-healthcheck-protocol: tcp
    service.beta.kubernetes.io/aws-load-balancer-name: pgpool
    service.beta.kubernetes.io/aws-load-balancer-nlb-target-type: ip
    service.beta.kubernetes.io/aws-load-balancer-scheme: internet-facing
    service.beta.kubernetes.io/aws-load-balancer-target-group-attributes: preserve_client_ip.enabled=true
    service.beta.kubernetes.io/aws-load-balancer-type: external
```

I expect the possibility to control annotations per service. The reason is that only LoadBalancer service must have annotations that control the creation of cloud resources like Target Groups or updating Security Groups.

### What do you see instead?

There's just one common variable `service.annotations` that is shared between pgpool and postgresql services, so both services will have the same set of annotations.

# Why it is wrong/dangerous:
1. pgpool creates target group and updates security group with a list of IPs, allowing access to the **pgpool port 5432**. A new Load Balancer is created and target group is attached to its listener. So far so good, this is desired configuration. 
2. postgresql service has hard-coded ClusterIP type, but `service.beta.kubernetes.io/aws-load-balancer-*` annotations cause ingress controller to create another target group for this service. This service doesn't have `loadBalancerSourceRanges` specified, so default value 0.0.0.0/0 will be used in security group rule allowing access from anywhere to **postgresql port 5432**. No LoadBalacer will be created.

The outcome of this configuration is that security group will allow access to port 5432 from anywhere regardless of the contents of loadBalancerSourceRanges. This is because port 5432 is the same for pgpool and postgresql.

### Additional information

Remedy?

Allow configuring pgpool service annotations separately from potgresql service annotations. The same pattern is already used for metrics. It might be problematic to make it backward compatible.

Or at least describe this behavior in documentation to let people know that they can't rely on `loadBalancerSourceRanges` when using aws-load-balancer-controller. I don't know how much are other controllers affected.

mouchar:
### Description of the change
Add `pgpool.annotations` and `postgresql.annotations` to the chart values.

### Benefits
It allows Pgpool and PostgreSQL service annotations to be separately configurable. As the pgpool service can be set to `LoadBalancer`, it may need a different set of annotations than postgresql service that is always `ClusterIP`.

### Possible drawbacks
None, the change is backward compatible. The original `service.annotations` still behaves as before and affects both postgresql and pgpool services, although it should be discouraged to use, especially on AWS with ALB ingress controller, because it may lead to unintentional public exposure of the 5432 port.

### Applicable issues
  - fixes #14523

### Checklist

### DIFF
--- a/bitnami/postgresql-ha/Chart.yaml
+++ b/bitnami/postgresql-ha/Chart.yaml
@@ -29,4 +29,4 @@ name: postgresql-ha
 sources:
   - https://github.com/bitnami/containers/tree/main/bitnami/postgresql
   - https://www.postgresql.org/
-version: 11.0.1
+version: 11.1.0
--- a/bitnami/postgresql-ha/README.md
+++ b/bitnami/postgresql-ha/README.md
@@ -7,7 +7,7 @@ This PostgreSQL cluster solution includes the PostgreSQL replication manager, an
 [Overview of PostgreSQL HA](https://www.postgresql.org/)
 
 Trademarks: This software listing is packaged by Bitnami. The respective trademarks mentioned in the offering are owned by the respective companies, and use of them does not imply any affiliation or endorsement.
-                           
+
 ## TL;DR
 
 ```console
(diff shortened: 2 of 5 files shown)


fmulero (commented):
Great Robert! Thanks a lot for your contribution.

I am happy with your change (and the way you bypassed possible backward issues) but could you call the new values something like `pgpool.serviceAnnotations` and `postgresql.serviceAnnotations`? In the same way is already done for `serviceLabels`

mouchar:
@fmulero 
I tried to keep naming consistent with the rest of the chart and also with other charts. `serviceAnnotations` is not used by any chart in bitnami/charts repo, while `annotations` is pretty common. Do you insist on your proposal?

fmulero:
Thanks @mouchar, I know it. To be perfectly blunt my initial thought was Why all these services are using the same config, in other charts we [split the config service values according to the component](https://github.com/bitnami/charts/blob/main/bitnami/grafana-loki/values.yaml#L831-L881) we are setting? Because I can't ask you to do the same (this is not your fault), I thought that using `serviceAnnotations` value could help users understand it a little bit better and could help us to identify it in a future standardization/housekeeping task. 

In summary, yes I would be a little bit more happier if you could do it ;)

fmulero (approved):
Thanks a lot @mouchar . LGTM

PR #14694, opened 2023-02-01
[bitnami/mastodon] Add required LOCAL_DOMAIN env variable

joshleecreates:
Signed-off-by: Josh Lee <joshleecreates@users.noreply.github.com>

### Description of the change

The mastodon application requires that this environment variable be set at startup. With this variable set, `WEB_DOMAIN` is optional but I left both for thoroughness. Ideally this should probably be two separate values options to support Mastodon's built-in option to have the WEB UI hosted at a different URL than the backend API.

### Benefits

This fixes the mastodon chart so that the application actually works after log in.

### Checklist

### DIFF
--- a/bitnami/mastodon/Chart.yaml
+++ b/bitnami/mastodon/Chart.yaml
@@ -42,4 +42,4 @@ name: mastodon
 sources:
   - https://github.com/bitnami/containers/tree/main/bitnami/mastodon
   - https://github.com/mastodon/mastodon/
-version: 1.0.1
+version: 1.0.2
--- a/bitnami/mastodon/values.yaml
+++ b/bitnami/mastodon/values.yaml
@@ -118,6 +118,7 @@ defaultConfig: |
   ES_HOST: {{ include "mastodon.elasticsearch.host" . | quote }}
   ES_PORT: {{ include "mastodon.elasticsearch.port" . | quote }}
   WEB_DOMAIN: {{ include "mastodon.web.domain" . | quote }}
+  LOCAL_DOMAIN: {{ include "mastodon.web.domain" . | quote }}
   STREAMING_API_BASE_URL: {{ include "mastodon.streaming.url" . | quote }}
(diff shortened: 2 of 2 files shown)


carrodher:
Thanks for this contribution, feel free to move it from _Draft_ to _Ready for review_ once you consider it is ready for taking a look so the team can provide feedback

migruiz4 (changes requested):
Hi @joshleecreates,

I'm sorry for the late response.

Could you please provide more details about your issues with the Mastodon chart?

While it is true that the LOCAL_DOMAIN variable is required to federate the mastodon instances, we can not accept this PR as it is, because setting it to a default value could be dangerous.

According to Mastodon documentation, the value LOCAL_DOMAIN can't be changed once it is set for the first time, so users could accidentally deploy Mastodon with the wrong LOCAL_DOMAIN value, or even using the default value.

Instead, the chart should include a new env variables `localDomain` set to an empty string, and the LOCAL_DOMAIN should only be set when a value is provided.

joshleecreates:
> Hi @joshleecreates,
> 
> I'm sorry for the late response.
> 
> Could you please provide more details about your issues with the Mastodon chart?
> 
> While it is true that the LOCAL_DOMAIN variable is required to federate the mastodon instances, we can not accept this PR as it is, because setting it to a default value could be dangerous.
> 
> According to Mastodon documentation, the value LOCAL_DOMAIN can't be changed once it is set for the first time, so users could accidentally deploy Mastodon with the wrong LOCAL_DOMAIN value, or even using the default value.
> 
> Instead, the chart should include a new env variables `localDomain` set to an empty string, and the LOCAL_DOMAIN should only be set when a value is provided.

Your proposed solution sounds good to me in the case where users are not ready to commit to their local domain on first launch. I will update the PR when I have time. Thanks!

migruiz4:
Thanks @joshleecreates!

Our goal is to continue providing a non-federated deployment by default but make it as easy as possible to deploy a federated instance. Your contribution is appreciated.

Issue #14765, opened 2023-02-06
[bitnami/clickhouse] Pods are not updated when more replicas are added


mohamedazouz:
### Name and Version

bitnami/clickhouse 2.3.1

### What steps will reproduce the bug?

1. Install chart with 1 replicas `replicaCount: 1` and 2 shards
2. 2 pods will be created `my-clickhouse-shard0-0 , my-clickhouse-shard1-0`
3. check pod default config `/etc/clickhouse-server/conf.d/00_default_overrides.xml`, there will be correct `remote_servers` setup as following
```xml
<remote_servers>
    <default>
      <shard>
          <replica>
              <host>my-clickhouse-shard0-0.my-clickhouse-headless.analytics.svc.cluster.local</host>
              <port>9000</port>
          </replica>
      </shard>
      <shard>
          <replica>
              <host>my-clickhouse-shard1-0.my-clickhouse-headless.analytics.svc.cluster.local</host>
              <port>9000</port>
          </replica>
      </shard>
    </default>
  </remote_servers>
```
4. upgrade replicas to 2 `replicaCount: 2`
5. 2 additional pods will be added `my-clickhouse-shard0-1, my-clickhouse-shard0-1`

Config On previous pods are still holding the old config But the config for the new pods is having new correct ones 
```xml
<remote_servers>
    <default>
      <shard>
          <replica>
              <host>my-clickhouse-shard0-0.my-clickhouse-headless.analytics.svc.cluster.local</host>
              <port>9000</port>
          </replica>
          <replica>
              <host>my-clickhouse-shard0-1.my-clickhouse-headless.analytics.svc.cluster.local</host>
              <port>9000</port>
          </replica>
      </shard>
      <shard>
          <replica>
              <host>my-clickhouse-shard1-0.my-clickhouse-headless.analytics.svc.cluster.local</host>
              <port>9000</port>
          </replica>
          <replica>
              <host>my-clickhouse-shard1-1.my-clickhouse-headless.analytics.svc.cluster.local</host>
              <port>9000</port>
          </replica>
      </shard>
    </default>
  </remote_servers>
```

* I had to delete the pods manually to grab the latest configMap changes.

### What is the expected behavior?

I was expected that all pods to be consistent having the same config specially for replicas/shards similar to the latest new added replicas

```xml
<remote_servers>
    <default>
      <shard>
          <replica>
              <host>my-clickhouse-shard0-0.my-clickhouse-headless.analytics.svc.cluster.local</host>
[...]

PR #14779, opened 2023-02-07
[bitnami/clickhouse] Add checksum annotation to restart pods when config updated

mohamedazouz:
### Description of the change

When Config changes, Pods needs to be restarted therefore, adding checksum annotation to make this happend
link to issue: https://github.com/bitnami/charts/issues/14765

### Checklist


bitnami/clickhouse/templates/statefulset.yaml
```diff
@@ -25,8 +25,10 @@ spec:
   {{- end }}
   template:
     metadata:
-      {{- if or $.Values.podAnnotations (and $.Values.metrics.enabled $.Values.metrics.podAnnotations) }}
       annotations:
+        checksum/config: {{ include (print $.Template.BasePath "/configmap.yaml") $ | sha256sum }}
+        checksum/config-extra: {{ include (print $.Template.BasePath "/configmap-extra.yaml") $ | sha256sum }}
+        {{- if or $.Values.podAnnotations (and $.Values.metrics.enabled $.Values.metrics.podAnnotations) }}
```
fmulero, line 31: You don't need this:
```
        {{- if or $.Values.podAnnotations (and $.Values.metrics.enabled $.Values.metrics.podAnnotations) }}
```
That condition was included at the beginning of the block to avoid an empty `annotations:` in the yaml. In this case we will have always at least the checksum, so you don't need the condition. Please remove also the final `{{- end }}` for that `if`

fmulero (commented):
Great!!!. Thanks a lot @mohamedazouz for your contribution.

I left a little comment, could you review it?

bitnami/clickhouse/templates/statefulset.yaml
mohamedazouz: That's done, please have a look

fmulero (approved):
Great! Thanks a lot @mohamedazouz. LGTM

PR #14780, opened 2023-02-07
[bitnami/thanos] Add Support for timePartitioning resource management

yuval-almog:
Signed-off-by: Yuval Almog <yuval.almog@riskified.com>

### Description of the change

It is now possible to specify different requests and resources for each StatefulSet storegateway container, with the default values being storegateway.resources.requests and storegateway.resources.limits.

Example for specifying resources per partition: 
```
timePartitioning:
   # One store for data older than 6 weeks
  - min: ""
    max: -6w
    resources: #this will change the StatefulSet storegateway container resources just for this partition 
      requests:
        cpu: 10m
        memory: 100Mi
      limits:
        cpu: 20m
        memory: 100Mi
  # One store for data newer than 6 weeks and older than 2 weeks
  - min: -6w
    max: -2w
  # One store for data newer than 2 weeks
  - min: -2w
    max: ""
```

### Benefits

The current implementation of timeaPrtitioning creates all Statefulsets with the same requests and limits, thus forcing the user to set the resources to be as high as the most resource intensive partition. 

As a result of this change, the user will have the ability to specify exactly the amount of memory and CPU each partition needs, thereby minimizing resource waste  

### Possible drawbacks

None

### Checklist

### DIFF
--- a/bitnami/thanos/README.md
+++ b/bitnami/thanos/README.md
@@ -1303,7 +1303,7 @@ timePartitioning:
     max: ""
 ```
 
-You can also specify different resources and limits configurations for each storegateway statefulset. This is done by adding a `resources.requests` and `resources.limits` to each item you wish to change , as shown below:
+You can also specify different resources and limits configurations for each storegateway statefulset. This is done by adding a `resources.requests` and `resources.limits` to each item you wish to change, as shown below:
 
 ```yaml
 timePartitioning:
--- a/bitnami/thanos/README.md
+++ b/bitnami/thanos/README.md
@@ -1286,6 +1286,8 @@ Thanos store supports partion based on time.
 
 Setting time partitions will create N number of store statefulsets based on the number of items in the `timePartitioning` list. Each item must contain the min and max time for querying in the supported format (find more details at [Thanos documentation](https://thanos.io/tip/components/store.md/#time-based-partitioning)).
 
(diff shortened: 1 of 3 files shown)


bitnami/thanos/README.md
```diff
@@ -1286,6 +1286,8 @@ Thanos store supports partion based on time.
 
 Setting time partitions will create N number of store statefulsets based on the number of items in the `timePartitioning` list. Each item must contain the min and max time for querying in the supported format (find more details at [Thanos documentation](https://thanos.io/tip/components/store.md/#time-based-partitioning)).
 
+You can speficiy different resources and limits configurations to each storegateway statefulset. This is done by adding a `resources.requests` and `resources.limits` to each item. 
+
```
rafariossaa, line 1290: I think adding an example of this in the example shown below would be nice.
(Like the example you posted in the PR description).

rafariossaa (changes requested):
Thanks for adding this.
Could you take a look to my suggestion ?

bitnami/thanos/README.md
yuval-almog: Hi @rafariossaa thanks for you comment. I agree. Added the example

rafariossaa (approved):
Thanks !
LGTM.

PR #14785, opened 2023-02-07
Fonction name was wrong

SoulKyu:
correct
```
executing "argo-cd/templates/applicationset/servicemonitor.yaml" at <include "argo-cd.applicationSet" .>: error calling include: template: no template "argo-cd.applicationSet" associated with template "gotpl"
```
Signed-off-by: SoulKyu <30684712+SoulKyu@users.noreply.github.com>


carrodher:
Thanks for your fix, please note it's needed to bump the chart version in the _Chart.yaml_ file

PR #14850, opened 2023-02-12
[bitnami/kafka] Kafka Kraft implementation

koraven:
Signed-off-by: Danil Kuchukov <danil.kuchukov@outlook.com>

### Kraft mode implementation

This PR implements Kraft mode of Kafka (without Zookeeper). 

### Benefits

It is a pretty important feature of Kafka and the current chart does not support it.

### Possible drawbacks

I tested it with Zookeeper and without it, but it would be great if other people help with testing for backward compatibility.

### Applicable issues

- Currently, bitnami docker image of Kafka does not support sasl_tls and sasl modes with Kraft mode. It hangs on "Creating zookeeper users" even when Kraft mode is enabled. I know what must be fixed and plan to create a PR next week.
- Scaling is not supported. It is known issue when using Kraft. To fix that we need to delete __cluster_metadata/quorum-state file before scaling. I'm not sure how to do that better. We can do it either with helm hooks or init-container.

### Checklist

### DIFF
--- a/bitnami/kafka/Chart.yaml
+++ b/bitnami/kafka/Chart.yaml
@@ -29,4 +29,4 @@ name: kafka
 sources:
   - https://github.com/bitnami/containers/tree/main/bitnami/kafka
   - https://kafka.apache.org/
-version: 21.2.0
+version: 21.3.0
--- a/bitnami/kafka/templates/scripts-configmap.yaml
+++ b/bitnami/kafka/templates/scripts-configmap.yaml
@@ -76,12 +76,35 @@ data:
     #!/bin/bash
 
     ID="${MY_POD_NAME#"{{ $fullname }}-"}"
+    # If process.roles is not set at all, it is assumed to be in ZooKeeper mode.
+    # https://kafka.apache.org/documentation/#kraft_role
+    
     if [[ -f "{{ .Values.logsDirs | splitList "," | first }}/meta.properties" ]]; then
-        export KAFKA_CFG_BROKER_ID="$(grep "broker.id" "{{ .Values.logsDirs | splitList "," | first }}/meta.properties" | awk -F '=' '{print $2}')"
+        if [[ $KAFKA_CFG_PROCESS_ROLES == "" ]]; then
+            export KAFKA_CFG_BROKER_ID="$(grep "broker.id" "{{ .Values.logsDirs | splitList "," | first }}/meta.properties" | awk -F '=' '{print $2}')"
+        else
+            export KAFKA_CFG_BROKER_ID="$(grep "node.id" "{{ .Values.logsDirs | splitList "," | first }}/meta.properties" | awk -F '=' '{print $2}')"
+        fi
     else
         export KAFKA_CFG_BROKER_ID="$((ID + {{ .Values.minBrokerId }}))"
     fi
 
+    if [[ $KAFKA_CFG_PROCESS_ROLES == *"controller"* ]]; then
+        node_id={{ .Values.minBrokerId }}
+        pod_id=0
+        while :
+        do 
+            VOTERS="${VOTERS}$node_id@{{  include "common.names.fullname" . }}-$pod_id.{{ include "common.names.fullname" . }}-headless.{{ $releaseNamespace }}.svc.cluster.local:{{ .Values.kraft.controllerPort }}"
+            node_id=$(( $node_id + 1 ))
+            pod_id=$(( $pod_id + 1 ))
+            if [[ $pod_id -ge {{ .Values.replicaCount }} ]]; then
+                break
+            else
+                VOTERS="$VOTERS,"
+            fi
+        done
+        export KAFKA_CFG_CONTROLLER_QUORUM_VOTERS=$VOTERS
+    fi
     {{- if eq .Values.brokerRackAssignment "aws-az" }}
     export KAFKA_CFG_BROKER_RACK=$(curl "http://169.254.169.254/latest/meta-data/placement/availability-zone-id")
     {{- end }}
--- a/bitnami/kafka/templates/scripts-configmap.yaml
+++ b/bitnami/kafka/templates/scripts-configmap.yaml
@@ -94,7 +94,7 @@ data:
         pod_id=0
         while :
         do 
-            VOTERS="${VOTERS}$node_id@{{  include "common.names.fullname" . }}-$pod_id.{{ include "common.names.fullname" . }}-headless.{{ $releaseNamespace }}.svc.cluster.local:{{ .Values.kraft.controllerPort }}"
+            VOTERS="${VOTERS}$node_id@{{  include "common.names.fullname" . }}-$pod_id.{{ include "common.names.fullname" . }}-headless.{{ $releaseNamespace }}.svc.cluster.local:{{ .Values.containerPorts.controller }}"
             node_id=$(( $node_id + 1 ))
             pod_id=$(( $pod_id + 1 ))
             if [[ $pod_id -ge {{ .Values.replicaCount }} ]]; then
--- a/bitnami/kafka/templates/scripts-configmap.yaml
+++ b/bitnami/kafka/templates/scripts-configmap.yaml
@@ -94,7 +94,7 @@ data:
         pod_id=0
         while :
         do 
-            VOTERS="${VOTERS}$node_id@{{  include "common.names.fullname" . }}-$pod_id.{{ include "common.names.fullname" . }}-headless.{{ $releaseNamespace }}.svc.cluster.local:{{ .Values.containerPorts.controller }}"
+            VOTERS="${VOTERS}$node_id@{{  include "common.names.fullname" . }}-$pod_id.{{ include "common.names.fullname" . }}-headless.{{ $releaseNamespace }}.svc.{{ $clusterDomain }}:{{ .Values.containerPorts.controller }}"
             node_id=$(( $node_id + 1 ))
             pod_id=$(( $pod_id + 1 ))
             if [[ $pod_id -ge {{ .Values.replicaCount }} ]]; then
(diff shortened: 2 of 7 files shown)


bitnami/kafka/values.yaml
```diff
@@ -1748,5 +1782,5 @@ zookeeper:
 ##
 externalZookeeper:
   ## @param externalZookeeper.servers List of external zookeeper servers to use. Typically used in combination with 'zookeeperChrootPath'.
-  ##
+  ## Must be empty if you use Kraft mode.
```
fmulero, line 1785: Same as above

bitnami/kafka/values.yaml
```diff
@@ -1698,13 +1701,44 @@ provisioning:
...
+
+
 ## @section ZooKeeper chart parameters
 
 ## ZooKeeper chart configuration
 ## https://github.com/bitnami/charts/blob/main/bitnami/zookeeper/values.yaml
 ##
 zookeeper:
   ## @param zookeeper.enabled Switch to enable or disable the ZooKeeper helm chart
+  ## Must be false if you use Kraft mode.
```
fmulero, line 1741: To include that information in the README file please move this sentence to the end of the upper line.

bitnami/kafka/templates/statefulset.yaml
```diff
@@ -202,9 +209,15 @@ spec:
               value: {{ join "," .Values.listeners }}
               {{- else if .Values.externalAccess.enabled }}
               value: "INTERNAL://:{{ .Values.containerPorts.internal }},CLIENT://:{{ .Values.containerPorts.client }},EXTERNAL://:{{ .Values.containerPorts.external }}"
+              {{- else if .Values.kraft.enabled }}
+                {{- if .Values.externalAccess.enabled }}
```
fmulero, line 213: Same as above

bitnami/kafka/templates/statefulset.yaml
```diff
@@ -188,6 +189,12 @@ spec:
               value: {{ .Values.listenerSecurityProtocolMap | quote }}
               {{- else if .Values.externalAccess.enabled }}
               value: "INTERNAL:{{ $interBrokerProtocol }},CLIENT:{{ $clientProtocol }},EXTERNAL:{{ $externalClientProtocol }}"
+              {{- else if .Values.kraft.enabled }}
+                {{- if .Values.externalAccess.enabled }}
```
fmulero, line 193: Something is wrong here. A previous branch in the if also checks if .Values.externalAccess.enabled is true doesn't make sense to test it here again

bitnami/kafka/templates/statefulset.yaml
```diff
@@ -202,9 +209,15 @@ spec:
...
+              value: "INTERNAL:{{ .Values.containerPorts.internal }},CLIENT://:{{ .Values.containerPorts.client }},CONTROLLER://:{{ .Values.kraft.controllerPort }},EXTERNAL://:{{ .Values.containerPorts.external }}"
+                {{- else }}
+              value: "INTERNAL://:{{ .Values.containerPorts.internal }},CLIENT://:{{ .Values.containerPorts.client }},CONTROLLER://:{{ .Values.kraft.controllerPort }}"
+                {{- end }}
               {{- else }}
               value: "INTERNAL://:{{ .Values.containerPorts.internal }},CLIENT://:{{ .Values.containerPorts.client }}"
-              {{- end }}
+              {{- end }}
```
fmulero, line 220: Please remove whitespaces
```suggestion
              {{- end }}
```

bitnami/kafka/Chart.yaml
```diff
@@ -29,4 +29,4 @@ name: kafka
 sources:
   - https://github.com/bitnami/containers/tree/main/bitnami/kafka
   - https://kafka.apache.org/
-version: 20.0.6
+version: 20.0.7
```
fmulero, line 32: As you are adding new features, please increase the minor version. https://semver.org/
```suggestion
version: 20.1.0
```

fmulero (changes requested):
Thanks a lot @koraven for your contribution.

I have to run more tests but I left some comments, Could you review them in the mean time?

bitnami/kafka/templates/statefulset.yaml
koraven: fixed it

fzyzcjy:
Hi, is there any updates, or estimated time to merge the PR? Thanks for the PR and looking forward to it!

AntonOfTheWoods:
@fmulero , are we just waiting now for the tests you mention above?

Kasape:
Hi @koraven, could you please explain how `kraft.controllerPort` value works? I'm trying your PR and leaving `kraft.controllerPort` with the default value `9095`. As a result, `KAFKA_CFG_LISTENERS` variable contains `CONTROLLER://:9095`, but that is the only place that was changed by `kraft.controllerPort`. With three replicas, everything except `kafka-0` pod fails with the exception 
```
[2023-02-21 15:14:28,891] WARN [RaftManager nodeId=2] Error connecting to node sentry-kafka-0.sentry-kafka-headless.debug.svc.cluster.local:9095 (id: 0 rack: null) (org.apache.kafka.clients.NetworkClient)
```
and truly a list of ports for the StatefulSet is only `9092/TCP, 9093/TCP` and headless service is also not open on port `9095`.  Maybe the headless service as well as pods should be open for port ` kraft.controllerPort`?

Kasape:
Also, it is unclear what _Must be false if you use Kraft mode._ means in the case of `externalZookeeper.servers` because it is a list. I understand the list must be empty, but I had to find out by looking into the code. Maybe it would be better to replace it with _Must be a empty list if you use Kraft mode_ or something similar. Someone can specify `externalZookeeper.servers: false` and it leads to
```
error calling include: template: kafka/templates/_helpers.tpl:514:22: executing "kafka.validateValues.kraftMode" at <len .Values.externalZookeeper.servers>: error calling len: len of type bool
```

Kasape:
There is another problem leading to `Helm install failed: error while running post render on files: map[string]interface {}(nil): yaml: unmarshal errors` exception. Helm itself can handle situation when there are multiple `value` keys so `helm install` command works, but Flux's helm-controller is unable to handle it and throws an exception. The cause of the problem is that statefulset contains the following:
```
- name: KAFKA_CFG_LISTENER_SECURITY_PROTOCOL_MAP
  value: "INTERNAL:PLAINTEXT,CLIENT:PLAINTEXT"
  value: "INTERNAL:PLAINTEXT,CLIENT:PLAINTEXT,CONTROLLER:PLAINTEXT"
- name: KAFKA_CFG_LISTENERS
  value: "INTERNAL://:9093,CLIENT://:9092"
  value: "INTERNAL://:9093,CLIENT://:9092,CONTROLLER://:9095"
```
for values.yaml
```
zookeeper:
  enabled: false
kraft:
  enabled: true
  clusterId: "Sy957XgrSh-V5yTzfwfwoCMg" 
```

koraven:
@Kasape I'll check what's wrong with controller port.
For the second issue, it is a typo. I was sure I set "empty" there :)
For the third one, I'll fix it. I know why this happens

fmulero (changes requested):
Thanks a lot @Kasape for your review!

I've added some additional comments, @koraven Could you take them a look?

bitnami/kafka/values.yaml
```diff
@@ -1698,13 +1701,43 @@ provisioning:
...
+  ##
+  processRoles: broker,controller
+  ## @param kraft.controllerListenerNames Controller listener names
+  ##
+  controllerListenerNames: CONTROLLER
+  ## @param kraft.clusterId Kafka ClusterID. You must set it if your cluster contains more than one node.
+  ## Generate with `cat /proc/sys/kernel/random/uuid | tr -d '-' | base64 | cut -b 1-22`. Run `export LC_ALL=C` before if you generate it on MacOS.
+  ## Example: k2yipv1sRue7z2_Y3o976A
+  ##
+  clusterId: ""
```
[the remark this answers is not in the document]
koraven: I decided to add a validator and put a default value for clusterId. We cannot use auto-generated Cluster ID because it is saved on a disk by Kafka. If you change the cluster ID variable, it will not be changed on the disk.
This option also allows to avoid cases when somebody deploys 1 replica, then tries to scale the cluster and gets issues because of it.

bitnami/kafka/values.yaml
```diff
@@ -1698,13 +1701,43 @@ provisioning:
...
+## Kraft configuration
+## Kafka mode without Zookeeper. Kafka nodes can work as controllers in this mode.
+##
+kraft:
+  ## @param kraft.enabled Switch to enable or disable the Kraft mode for Kafka
+  ##
+  enabled: false
+  ## @param kraft.controllerPort Kafka Controller listener port
+  ##
+  controllerPort: 9095
```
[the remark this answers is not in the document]
koraven: Added the ports for headless-svc and for the kafka-container. I intentionally did not define .Values.service.ports.controller because we need only headless service for the controller. And for the headless service controller port will be always equal to service port

bitnami/kafka/templates/scripts-configmap.yaml
```diff
@@ -76,12 +76,35 @@ data:
...
     else
         export KAFKA_CFG_BROKER_ID="$((ID + {{ .Values.minBrokerId }}))"
     fi
 
+    if [[ $KAFKA_CFG_PROCESS_ROLES == *"controller"* ]]; then
+        node_id={{ .Values.minBrokerId }}
+        pod_id=0
+        while :
+        do 
+            VOTERS="${VOTERS}$node_id@{{  include "common.names.fullname" . }}-$pod_id.{{ include "common.names.fullname" . }}-headless.{{ $releaseNamespace }}.svc.cluster.local:{{ .Values.containerPorts.controller }}"
```
AntonOfTheWoods, line 97: @koraven , I'm pretty sure this should not be using `cluster.local`, and should instead be using `{{ $clusterDomain }}`. I am not using the default and it doesn't work as is (and does with that change).

AntonOfTheWoods (changes requested):
I have things working with this change

fmulero (commented):
Thanks a lot @koraven! Could you review the @AntonOfTheWoods comment and update your branch from main?

fmulero (commented):
Please bump the chart version, it was missed during merge action.

fmulero (approved):
LGTM

fzyzcjy:
🎉  Congratulations!

Issue #14847, opened 2023-02-12
Support Labels on ServiceAccount


UXabre:
### Name and Version

bitnami/keycloak:14.0.0

### What is the problem this feature will solve?

Currently, we're using Workload Identity to federate our Service Account with Azure AD. Via a label `"azure.workload.identity/use: "true""` is communicated to the AKS binding that this serviceaccount should use `workload.identity`. Currently, only annotations are supported (which is also needed for workload identity to work, as I needed to set the ClientID)

### What is the feature you are proposing to solve the problem?

Next to annotations, support labels for the service account, I see many other projects that use serviceaccounts to also expose this.

### What alternatives have you considered?

The alternative, like I'm doing right now, is not letting the helm chart create the service account but creating it as part of a parent chart which creates this for me.

carrodher:
Thanks for reporting this issue. Would you like to contribute by creating a PR to solve the issue? The Bitnami team will be happy to review it and provide feedback. [Here](https://github.com/bitnami/charts/blob/main/CONTRIBUTING.md) you can find the contributing guidelines.

UXabre:
Yes, i'm already working on it as it's trivial to add it i think. Will create the pr tomorrow as I want to go over the contribution doc thoroughly:-)

PR #14852, opened 2023-02-13
[bitnami/keycloak] Support extraLabels for serviceAccount

UXabre:
Signed-off-by: Arend Lapere <arend.lapere@gmail.com>

The reason for this PR is that, as I'm injecting secrets via a CSI provider, I need to federate my serviceaccount using Workload Identity. In AKS this is done by passing an annotation with my clientId and a label stating it to use aks workload identity. The latter is not yet supported and I would have to create a service account outside this chart to support 

### Description of the change

This PR adds the configuration of extraLabels to serviceAccount configuration block. This allows to add extra labels on top of the ones already there (hence the reason I named it extraLabels and not just labels)

### Benefits

We don't need to create a service account outside of our helm chart. It's more integrated this way.

### Possible drawbacks

None that I can see as it is optionally to add labels but they don't really affect the current labels

### Applicable issues

#14847 

### Checklist

### DIFF
--- a/bitnami/keycloak/Chart.yaml
+++ b/bitnami/keycloak/Chart.yaml
@@ -26,4 +26,4 @@ name: keycloak
 sources:
   - https://github.com/bitnami/containers/tree/main/bitnami/keycloak
   - https://github.com/keycloak/keycloak
-version: 13.0.4
+version: 13.0.5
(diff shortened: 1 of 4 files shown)


corico44 (approved):
LGTM

Issue #5509, opened 2021-02-16
[bitnami/thanos] Option to run Thanos Compactor as CronJob


lorenzo-biava:
**Which chart**:
Thanos (latest) v3.9.0 +

**Is your feature request related to a problem? Please describe.**
The Compactor is run as a deployment. While this is a convenient way to run it, it's not necessarily efficient from a resource usage point of view. For instance, with our dataset, compactions for longer intervals (e.g. 7/14 days) require a significant amount of memory (20-40GB) and only happen once a week or so.

Having the Compactor running constantly, using minimal resources most of the time, tends to be kind of wasteful in this situation. On the other hand, granting it a smaller amount of memory requires us to react to incidents when big compactions happen and it gets stuck due to OOM.

**Describe the solution you'd like**
The ability to choose how to run the Compactor: Deployment or CronJob.

**Describe alternatives you've considered**
N/A

**Additional context**
The Compactor command does indeed run to completion by default, so the CronJob support has been envisioned (https://github.com/thanos-io/thanos/blob/master/docs/components/compact.md#compactor).

I might be able to work on a PR if you believe this feature is viable.

andresbono:
Hi @lorenzo-biava, thank you very much for suggesting this feature! I think having the option to run Compactor as Deployment or Cronjob makes sense, and as you point, Compactor supports both modes.

> I might be able to work on a PR if you believe this feature is viable.

That would be great! Contributions are more than welcome. Please link this issue into the PR for reference. We will be happy to review it :+1:

djsly:
reopen

migruiz4:
Reopening because it shouldn't have been closed by the stale-bot

carrodher:
Unfortunately, this issue was created a long time ago and although there is an internal task to fix it, it was not prioritized as something to address in the short/mid term. It's not a technical reason but something related to the capacity since we're a small team.

Being said that, contributions via PRs are more than welcome in both repositories (containers and charts). Just in case you would like to contribute.

During this time, there are several releases of this asset and it's possible the issue has gone as part of other changes. If that's not the case and you are still experiencing this issue, please feel free to reopen it and we will re-evaluate it.

PR #14865, opened 2023-02-13
[bitnami/thanos] Compactor cronjob

gjacquet:
### Description of the change

Support running the compactor as a cronjob rather than as a deployment. Compactor supports running as a cronjob. 

### Benefits

Running compactor as a job instead of deployment would allow for more efficient use of resources in Kubernetes as it would be possible to request resources only when the compactor is actively running.

### Possible drawbacks

None as it is still possible to run as a deployment. This new configuration is purely optional.

### Applicable issues

- Fixes https://github.com/bitnami/charts/issues/5509.

### Checklist


corico44:
Hello @gjacquet, could you provide me how you have tested these changes to reproduce it, please?

gjacquet:
@corico44 for this change, I picked 2 of our environments running Thanos and set them up to use the chart with this change.

I configured the first one to use compactor as a cronjob and the second one as a deployment.

Then I ran helm template and compared the new output (with a `git diff`) with what we had previously. Deployment showed no significant difference.

Cronjob change looked fine and I applied it to one of our environment. I checked job log output and it seemed fine, it has been running for nearly a day.

As a reviewer, you might want to compare the new `_pod_template.tpl` with `deployment.yaml` from the main branch. `_pod_template.tpl` is pretty much the pod template from `deployment.yaml` with indentation changes. The idea was that since the pod template is nearly the same for a deployment or a cronjob, it could make sense to move it to a shared template.

gjacquet:
I guess the compactor only makes sense if you are using an object store, in s3 for instance. I have something like this:

```yaml
objstoreConfig: |
        type: S3
        config:
          bucket: "<my s3 bucket>"
          endpoint: "s3.us-west-2.amazonaws.com"
          region: "us-west-3"
          insecure: false
          signature_version2: false
          put_user_metadata: {}
          trace:
            enable: true
```

corico44:
Thank you @gjacquet. Now, I get the following error:
`Error: INSTALLATION FAILED: CronJob.batch "sp-th-thanos-compactor" is invalid: spec.jobTemplate.spec.template.spec.restartPolicy: Required value: valid values: "OnFailure", "Never"`

So, restartPolicy should either have a default value or be no longer required.

gjacquet:
@corico44 thanks for the feedback.
I pushed a change to set a default restart policy if unset.

corico44 (approved):
LGTM

onedr0p:
@gjacquet if using a cronjob for the compactor is setting any sort of persistence still required?

PR #14893, opened 2023-02-15
[bitnami/postgresql] feat: :sparkles: Add ServiceBinding-compatible secrets

javsalgar:
### Description of the change

THIS FEATURE IS EXPERIMENTAL

This PR adds a `serviceBindings.enabled` flag that, apart from the regular secret, it will create extra ones to be integrated with the [ServiceBindings](https://servicebinding.io/). In order to not make it very breaking, the flag is disabled by default and also it will not replace the existing secrets. 

It will create two extra secrets: one for the Postgres user and one for the custom user. It can be simply integrated with a ServiceBinding object out of the box. 

### Applicable issues

- fixes #

### Checklist

### DIFF
--- a/bitnami/postgresql/templates/secrets.yaml
+++ b/bitnami/postgresql/templates/secrets.yaml
@@ -1,12 +1,24 @@
-{{- if (include "postgresql.createSecret" .) }}
 {{- $host := include "postgresql.primary.fullname" . }}
 {{- $port := include "postgresql.service.port" . }}
-{{- $postgresPassword := include "common.secrets.passwords.manage" (dict "secret" (include "common.names.fullname" .) "key" "postgres-password" "providedValues" (list "global.postgresql.auth.postgresPassword" "auth.postgresPassword") "context" $) | trimAll "\"" | b64dec }}
-{{- $replicationPassword := include "common.secrets.passwords.manage" (dict "secret" (include "common.names.fullname" .) "key" "replication-password" "providedValues" (list "auth.replicationPassword") "context" $) | trimAll "\"" | b64dec }}
-{{- $ldapPassword := coalesce .Values.ldap.bind_password .Values.ldap.bindpw }}
+{{- $postgresPassword := "" }}
+{{- if .Values.auth.enablePostgresUser }}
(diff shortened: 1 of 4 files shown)


bitnami/postgresql/values.yaml
```diff
@@ -1158,6 +1164,12 @@ volumePermissions:
     runAsUser: 0
 
 ## @section Other Parameters
+##
+
+## @param serviceBindings.enabled Create secret for service binding (Experimental)
```
mpermar, line 1169:
```suggestion
## @param serviceBindings.enabled Create secret for service binding (Experimental)
## Ref: https://servicebinding.io/service-provider/
```

Worth adding the ref?

bitnami/postgresql/templates/secrets.yaml
```diff
@@ -14,16 +22,62 @@ metadata:
...
-  replication-password: {{ include "common.secrets.passwords.manage" (dict "secret" (include "common.names.fullname" .) "key" "replication-password" "providedValues" (list "auth.replicationPassword") "context" $) }}
+  replication-password: {{ $replicationPassword | b64enc | quote }}
   {{- end }}
   # We don't auto-generate LDAP password when it's not provided as we do for other passwords
   {{- if and .Values.ldap.enabled (or .Values.ldap.bind_password .Values.ldap.bindpw) }}
-  ldap-password: {{ coalesce .Values.ldap.bind_password .Values.ldap.bindpw  | b64enc | quote }}
+  ldap-password: {{ $ldapPassword  | b64enc | quote }}
   {{- end }}
-{{- end -}}
+{{- if .Values.serviceBindings.enabled }}
```
mpermar, line 37: Does it make sense to have this`if` condition outside of the global createSecret if. That way at least in some scenarios there would only be one secret.

javsalgar: Not sure if I follow. There will be only one secret when `serviceBindings.enabled=false`, which will be the default secret. When this flag is set to true, we will have the default secret plus the serviceBindings one.

mpermar (approved):
Looks great.

PR #14934, opened 2023-02-17
[bitnami/redis] Update health-configmap.yaml

### LINKED ISSUE #14738: [bitnami/redis] Non persistent Redis sentinel
### Name and Version

bitnami/redis 17.0.3

1. Using k8s with version 1.21
2. We have disabled persistance and using redis with sentinel
3. Deploy redis sentinel
4. Try to update it

no

Data not lost

### What do you see instead?

Data lost

### Additional information

I suppose, the readiness probe isn't correct. The pod is ready when replication to this pod isn't ended. We need to check is pod ready somehow else. It happens not always but it's dangerous. Can you help with it?

DevKhaverko:
### Description of the change

This change is needed because when k8s-cluster doesn't allow saving on disk, we need to disable persistance. When we start deploy changes in config of redis, first pod is going down, and replication to the new pod starts. In period of replication master can go down and new pod, which was replicated to this master can be promoted to be a master. But this pod doesn't has all data, actually when we have a lot of data, on new pod can small piece of this data, so we can lose big part of our data. My change is about watching status of replication and whet it ends we can mark new pod as ready and after that terminate another old pod

### Benefits

Data in redis without persistance isn't lost

### Applicable issues

- fixes #14738

### Checklist

### DIFF
--- a/bitnami/redis/templates/health-configmap.yaml
+++ b/bitnami/redis/templates/health-configmap.yaml
@@ -12,35 +12,49 @@ metadata:
   {{- end }}
 data:
   ping_readiness_local.sh: |-
-    #!/bin/bash
+    #!/usr/bin/env bash
+    REDIS_HOST="localhost"
+    REDIS_PORT="6379"
+    redis_cli=(timeout -s 3 1 redis-cli -h "$REDIS_HOST" -p "$REDIS_PORT")
 
     [[ -f $REDIS_PASSWORD_FILE ]] && export REDIS_PASSWORD="$(< "${REDIS_PASSWORD_FILE}")"
     [[ -n "$REDIS_PASSWORD" ]] && export REDISCLI_AUTH="$REDIS_PASSWORD"
-    response=$(
-      timeout -s 3 $1 \
-      redis-cli \
-        -h localhost \
-{{- if .Values.tls.enabled }}
-        -p $REDIS_TLS_PORT \
-        --tls \
-        --cacert {{ template "redis.tlsCACert" . }} \
-        {{- if .Values.tls.authClients }}
-          --cert {{ template "redis.tlsCert" . }} \
-          --key {{ template "redis.tlsCertKey" . }} \
-        {{- end }}
-{{- else }}
-        -p $REDIS_PORT \
-{{- end }}
-        ping
-    )
-    if [ "$?" -eq "124" ]; then
-      echo "Timed out"
-      exit 1
-    fi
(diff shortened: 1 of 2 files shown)


marcosbc:
@DevKhaverko Please add some more information to the PR, especially the context of the change and some explanations to understand why this is needed.

DevKhaverko:
@marcosbc added description

bitnami/redis/templates/health-configmap.yaml
```diff
@@ -12,35 +12,49 @@ metadata:
...
-        -p $REDIS_TLS_PORT \
-        --tls \
-        --cacert {{ template "redis.tlsCACert" . }} \
-        {{- if .Values.tls.authClients }}
-          --cert {{ template "redis.tlsCert" . }} \
-          --key {{ template "redis.tlsCertKey" . }} \
-        {{- end }}
-{{- else }}
-        -p $REDIS_PORT \
-{{- end }}
```
marcosbc, line 33: It seems like these options are removed, which allows for stuff like specifying a custom Redis port or SSL/TLS support. Could you re-add them?

bitnami/redis/templates/health-configmap.yaml
```diff
@@ -12,35 +12,49 @@ metadata:
   {{- end }}
 data:
   ping_readiness_local.sh: |-
-    #!/bin/bash
+    #!/usr/bin/env bash
```
marcosbc, line 15: In this base image, Bash is guaranteed to be located at:
```suggestion
    #!/bin/bash
```

bitnami/redis/templates/health-configmap.yaml
```diff
@@ -12,35 +12,49 @@ metadata:
...
-      echo "Timed out"
-      exit 1
-    fi
-    if [ "$response" != "PONG" ]; then
-      echo "$response"
-      exit 1
-    fi
+
+    function getRedisRole() {
+        "${redis_cli[@]}" info replication 2>/dev/null | grep "role" | tr -d "\r" | tr -d "\n"
```
marcosbc, line 24: Isn't `redis info replication` specific to `--set architecture=replication` option for the chart (and not `standalone`)?

In addition, a small improvement that could be made:

```suggestion
        "${redis_cli[@]}" info replication 2>/dev/null | grep -Eo "^role:.*"
```

(please also apply this below)

marcosbc (changes requested):

PR #15057, opened 2023-02-18
[bitnami/redis] Use SIGTERM on timeout for probes

### LINKED ISSUE #15055: [bitnami/redis] - Probes uses SIGQUIT and generates coredumps
### Name and Version

[redis](https://github.com/bitnami/charts/tree/main/bitnami/redis) 17.7.4 and [redis-cluster](https://github.com/bitnami/charts/tree/main/bitnami/redis-cluster) have similar code for probes.

1. Deploy redis in standalone mode
2. Exec into the pod
3. Connect to the local server using `redis-cli`
4. [Send `SIGQUIT` signal to the redis-cli](https://github.com/bitnami/charts/blob/main/bitnami/redis/templates/health-configmap.yaml#L20-L22)
   ```bash
   pkill -3 redis-cli
   ```
5. See the information about core-dump
   ```
   Quit (core dumped)
   ```
6. List coredump on the cluster node
   ```bash
   coredumpctl
   ```

We use default probes values

In case when probe is failing it should not generate a coredump - [Termination Signals](https://www.gnu.org/software/libc/manual/html_node/Termination-Signals.html)

**SIGTERM**
> The SIGTERM signal is a generic signal used to cause program termination. Unlike SIGKILL, this signal can be blocked, handled, and ignored. It is the normal way to politely ask a program to terminate.
>
> The shell command kill generates SIGTERM by default.

**SIGQUIT**
> The SIGQUIT signal is similar to SIGINT, except that it’s controlled by a different key—the QUIT character, usually C-\—and produces a core dump when it terminates the process, just like a program error signal. You can think of this as a program error condition “detected” by the user.

### What do you see instead?

On the current node we have a lot of coredumps
```bash
coredumpctl list | wc -l
218
```
Screenshot 2023-02-18 at 21 42 19

### Additional information

https://github.com/bitnami/charts/blob/08e0f4207adbe191e16e26a7cb43cfb609443e54/bitnami/redis/templates/health-configmap.yaml#L20-L22

air3ijai:
### Description of the change

 Change signal to [SIGTERM](https://www.gnu.org/software/libc/manual/html_node/Termination-Signals.html) on probes
> The SIGTERM signal is a generic signal used to cause program termination. Unlike SIGKILL, this signal can be blocked, handled, and ignored. It is the normal way to politely ask a program to terminate.
>
> The shell command kill generates SIGTERM by default. 

### Benefits

 Using `SIGTERM` instead of the `SIGQUIT` on probes will remove unnecessary coredumps.

### Possible drawbacks

 Probably nothing and in the case of the issue, `redis-cli` will generate a proper coredump.

### Applicable issues

- fixes #15055

### Additional information

We tested it on our case when redis does not respond in time and `redis-cli` is killed
```bash
# SIGQUIT - 3
timeout -s 3 2 redis-cli -h localhost -p 6379 ping
timeout: the monitored command dumped core
echo $?
124

# SIGTERM - 15
timeout -s 15 2 redis-cli -h localhost -p 6379 ping
echo $?
124
```

### Checklist


agrevtsev (dismissed):

marcosbc (approved):
LGTM! Thanks for the improvement!

PR #15077, opened 2023-02-21
[bitnami/wordpress] Template hostname usages to allow using variables

andreas-eberle:
### Description of the change

This change introduced the tpl function into the ingress template and to template the hostname variable. This allows to use variables in the hostname that are then resolved by the tpl function. 

This change is important for people using the wordpress helm chart as a subchart of a bigger installation together with other services. In such cases, you do not want to specify the full hostname for every subchart but you want to be able to reference a base url variable and only extend the hostname from that. This PR allows to do that. E.g. the following would now be possible:

```
wordpress:
  ingress:
    enabled: true
    hostname: shop.{{ .Values.global.baseUrl }}
```

### Benefits

The chart is easier to use in multichart szenarios without duplicating variables. 

### Possible drawbacks

I don't think there exist drawbacks. Since we are using the tpl function for the host, it can never happend that a valid hostname would be interpreted as a variable. Therefore this change is backwards compatible as it will break a valid installation. 

### Checklist


bitnami/wordpress/templates/ingress.yaml
```diff
@@ -21,7 +23,7 @@ spec:
   {{- end }}
   rules:
     {{- if .Values.ingress.hostname }}
-    - host: {{ .Values.ingress.hostname | quote }}
+    - host: {{ tpl .Values.ingress.hostname $baseContext | quote }}
```
jouve, line 26:
```suggestion
    - host: {{ tpl .Values.ingress.hostname $ | quote }}
```

[When execution begins, $ is set to the data argument passed to Execute, that is, *to the starting value of dot.](https://pkg.go.dev/text/template)

aoterolorenzo: Shouldn't this apply to all other $baseContext entries for this file? @andreas-eberle

jouve: indeed :)

andreas-eberle: True, sorry guys. I had another look at it and in all but one case, I can even use the normal .
I updated the branch. Hopefully its good now :)

aoterolorenzo: I think there could be a misunderstanding here: 

Using $ instead of baseContext in all the occurrences in this diff would be the best solution, since `$` saves the context of `.` at the start of the templating, so you don't need actually to use a auxiliar variable `$baseContext`to save the `.` context: it is the same as `$`.

`- host: {{ tpl .Values.ingress.hostname $ | quote }}`

would work here.

bitnami/wordpress/templates/ingress.yaml
```diff
@@ -1,4 +1,6 @@
 {{- if .Values.ingress.enabled }}
+{{- $baseContext := . -}}
+
```
aoterolorenzo, line 3:
```suggestion
```

bitnami/wordpress/templates/ingress.yaml
```diff
@@ -21,7 +23,7 @@ spec:
   {{- end }}
   rules:
     {{- if .Values.ingress.hostname }}
-    - host: {{ .Values.ingress.hostname | quote }}
+    - host: {{ tpl .Values.ingress.hostname . | quote }}
```
aoterolorenzo, line 26:
```suggestion
    - host: {{ tpl .Values.ingress.hostname $ | quote }}
```

bitnami/wordpress/templates/ingress.yaml
```diff
@@ -34,7 +36,7 @@ spec:
             backend: {{- include "common.ingress.backend" (dict "serviceName" (include "common.names.fullname" .) "servicePort" "http" "context" $)  | nindent 14 }}
     {{- end }}
     {{- range .Values.ingress.extraHosts }}
-    - host: {{ .name | quote }}
+    - host: {{ tpl .name $baseContext | quote }}
```
aoterolorenzo, line 39:
```suggestion
    - host: {{ tpl .name $ | quote }}
```

bitnami/wordpress/templates/ingress.yaml
```diff
@@ -50,11 +52,11 @@ spec:
   tls:
     {{- if and .Values.ingress.tls (or (include "common.ingress.certManagerRequest" ( dict "annotations" .Values.ingress.annotations )) .Values.ingress.selfSigned) }}
     - hosts:
-        - {{ .Values.ingress.hostname | quote }}
+        - {{ tpl .Values.ingress.hostname $baseContext | quote }}
```
aoterolorenzo, line 55:
```suggestion
        - {{ tpl .Values.ingress.hostname $ | quote }}
```

bitnami/wordpress/templates/ingress.yaml
```diff
@@ -50,11 +52,11 @@ spec:
...
         {{- if or (.Values.ingress.tlsWwwPrefix) (eq (index .Values.ingress.annotations "nginx.ingress.kubernetes.io/from-to-www-redirect") "true" ) }}
-        - {{ printf "www.%s" .Values.ingress.hostname | quote }}
+        - {{ printf "www.%s" (tpl .Values.ingress.hostname $baseContext) | quote }}
```
aoterolorenzo, line 57:
```suggestion
        - {{ printf "www.%s" (tpl .Values.ingress.hostname $) | quote }}
```

bitnami/wordpress/templates/ingress.yaml
```diff
@@ -50,11 +52,11 @@ spec:
...
         {{- end }}
-      secretName: {{ printf "%s-tls" .Values.ingress.hostname }}
+      secretName: {{ printf "%s-tls" (tpl .Values.ingress.hostname $baseContext) }}
```
aoterolorenzo, line 59:
```suggestion
      secretName: {{ printf "%s-tls" (tpl .Values.ingress.hostname $) }}
```

aoterolorenzo (changes requested):

bitnami/wordpress/templates/ingress.yaml
andreas-eberle: Why would you use $ here? . does the same, right? It is not in a loop, so it should do the same?

andreas-eberle:
@aoterolorenzo: Can you have another look? I applied your suggestions. Sorry, I thought I already did this last time, but I must have messed up the commit/push. Sorry for the extra work :(

aoterolorenzo:
@andreas-eberle let me check with the team. We are having some issues ongoing with the linters so probably it's on our side

aoterolorenzo:
@andreas-eberle I'm aware of the team fixing some linter issues lately, so probably the markdown have been updated on main, meanwhile your branch still have the older changes, so the PR is trying to re-push them all. Could you try to rebase/merge the updated main branch?

andreas-eberle:
@aoterolorenzo : I merged main again into my PR. Unfortunately the error persists :( 
I also tried regenerating the readme, but it had no effect. Do you have an idea what I can do?

andreas-eberle:
@aoterolorenzo : I found the problematic newlines and removed them manually.

aoterolorenzo (dismissed):

aoterolorenzo (changes requested):
Hey @andreas-eberle,

Seems fine now. I just noticed a couple of details above, please check to finally merge.

bitnami/wordpress/templates/ingress.yaml
```diff
@@ -1,4 +1,5 @@
 {{- if .Values.ingress.enabled }}
+
```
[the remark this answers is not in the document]
andreas-eberle: done

andreas-eberle:
@aoterolorenzo: Thanks. I applied the requested changes. Can you check again?

andreas-eberle:
@aoterolorenzo : Do you know why the pipeline failed? I don't get it from the logs...

andreas-eberle:
@aoterolorenzo : I merged in main again but without any success... How can I find out what's wrong with this test?

andreas-eberle:
@aoterolorenzo, I merged again with main and it seems to have resolved the issue. Can you check the PR again so we finally get it merged?

aoterolorenzo (approved):
LGTM!

@andreas-eberle thank you so much for contributing, patiently collaborating in the review and for your time. Sorry this having got so delayed.

aoterolorenzo (approved):

PR #15091, opened 2023-02-22
[bitnami/redis] add option to allow sentinel to announce podIp

beatkind:
### Description of the change

This change allows the sentinel and master instances to announce their IP based on the podIP which is set as env to the container.

### Benefits

This allows the correct announcement of the sentinel & master IP even if externalDNS is enabled and a discovery via the get_full_hostname would reveal old IPs.

### Possible drawbacks

`¯\_(ツ)_/¯`

### Applicable issues

- This fixes the issue that the container might be up and looking up the sentinels IP before the externalDNS / TTL has published new DNS entries

### Additional information

`¯\_(ツ)_/¯`

### Checklist


ChrisNoSim (approved):
Very useful, because we are currently having some problems with the chart. If you are using `externalDNS` and the `useHostnames` parameter our sentinels got no quorum, because old IPs were fetched from the external DNS. With this PR the pods will know their IPs without asking any DNS server. Would be very happy if you merge that. :+1:

bitnami/redis/templates/scripts-configmap.yaml
```diff
@@ -671,7 +680,11 @@ data:
 
     echo "" >> /opt/bitnami/redis/etc/replica.conf
     echo "replica-announce-port $REDISPORT" >> /opt/bitnami/redis/etc/replica.conf
+    {{- if .Values.replica.announceContainerIP }}
```
fmulero, line 683: `replica.announceContainerIP` is not defined in values.yaml file, but is it really needed? I think we should have a global variable for this porpoise to avoid mixing IP addresses and names

fmulero (changes requested):
Great contribution, Thanks a lot @beatkind!

I left some comments and I have a general doubt, Shouldn't we modify also `resolve-hostnames` and `resolve-hostnames` sentinel parameters if announceIP is enabled?

fmulero:
Thanks Jan

> I am a little bit unsure what you mean with the resolve-hostnames. Furthermore, I think it is still OK for redis to resolve the hostnames when publishing IPs instead of FQDNs.

Sorry, should have shared this before. I was reviewing [redis documentation about this topic](https://redis.io/docs/management/sentinel/#ip-addresses-and-dns-names) and I am not sure if we should disable [announce-hostnames and/or resolve-hostnames](https://github.com/bitnami/charts/blob/624964e56085213cff89a94238b13710e457310d/bitnami/redis/templates/scripts-configmap.yaml#L402-L407) also.

beatkind:
Thanks for clarification. You are totally right. To honour this primitive:

> You should use hostnames everywhere and avoid mixing hostnames and IP addresses. To do that, use replica-announce-ip <hostname> and sentinel announce-ip <hostname> for all Redis and Sentinel instances, respectively.

I will disable hostnames when announceContainerIP is enabled.

fmulero (commented):
Hi Jan,

I was testing your changes and I've got this error:
```console
```

beatkind:
Thank you very much, good catch! If only I had been a little more careful.

fmulero:
Hi @beatkind

Thanks for fixing it, but I'm sorry to say you that now is failing when sentinel is enabled:
```console
$ helm install redis bitnami/redis --set global.announceContainerIP=true --set sentinel.enabled=true
...
$ kubectl logs redis-node-0 -c sentinel
 09:26:53.40 INFO  ==> about to run the command: REDISCLI_AUTH=$REDIS_PASSWORD timeout 220 redis-cli -h redis.default.svc.cluster.local -p 26379 sentinel get-master-addr-by-name mymaster
Could not connect to Redis at redis.default.svc.cluster.local:26379: Connection refused

*** FATAL CONFIG FILE ERROR (Redis 7.0.10) ***
Reading the configuration file, at line 3
>>> 'sentinel monitor mymaster redis-node-0.redis-headless.default.svc.cluster.local 6379 2'
Can't resolve instance hostname.
1:X 27 Mar 2023 09:27:07.999 # Failed to resolve hostname 'redis-node-0.redis-headless.default.svc.cluster.local'
```

It seems that `REDIS_MASTER_HOST` should use IPs also.

beatkind:
@fmulero interesting, we have this setup running in our environment. Let me check what the case here is.

PR #15138, opened 2023-02-24
[bitnami/contour-operator] Adapt ginkgo tests for non-GKE

FraPazGal:
### Description of the change

There were several problems with `bitnami/contour-operator` Ginkgo tests:

- Sometimes GKE was affected by a race condition. By Ginkgo design, the testing environment is set up and removed for each individual test. The issue here came from the created testing service taking too long to be deleted sometimes. That created conflicts with the assigned IPs when 2 services were up simultaneously.
- TKG/EKS/etc: the test suite expected the service's External-IP always to be an IP, but AWS-based clusters use a host instead.
- After fixing the previous error, an additional one appeared in AWS-based clusters by which the provided host didn't resolve the testing deployment. After investigation, the cause of this related to the Envoy LB configuration in AWS clusters, where the proxy protocol was needed for ELBs.

The first two issues have now been fixed by moving the creation of the Contour and Ingress objects to Ginkgo (which also solves some pullSecrets and endpoint limitations from VIB) as well as implementing some retriable checks.

The final error required specific changes to the contour CRD objects for AWS clusters at pre-deployment time or at ginkgo runtime. This would have added its own drawbacks and complexity to the logic. After reviewing the needed changes, it has been decided to remove the test for which the changes were needed.

This removal decision comes from evaluating that the withdrawn test does not add that much value to the operator's test suite. Said test was thought to verify a functionality of the proper contour app, but not the operator's. Given that a similar test is already used for the bitnami/contour chart and that our operator deploys a contour instance that uses our images, this test can be safely removed.

### Benefits

Increased test stability and proper execution in non-GKE clusters. Decreased testing time.

### Possible drawbacks

There should be none.

### Checklist


.vib/contour-operator/ginkgo/contour_operator.go
```diff
@@ -19,44 +21,85 @@ import (
...
+					return isContourReady(ctx, dynamicClient, contourName)
+				})
+				Expect(isReady).To(BeTrue())
+			}
+			hasIP, err = retry("isServiceLBReady", 10, 30*time.Second, func() (bool, error) {
+				return isServiceLBReady(ctx, coreclient, "envoy")
+			})
 			if err != nil {
-				panic(fmt.Sprintf("There was an error retrieving the %q Ingress resource: %q", *ingressName, err))
+				panic(fmt.Sprintf("There was an error checking whether the testing service had an IP assigned: %q", err))
```
joancafom, line 62:
```suggestion
				panic(fmt.Sprintf("There was an error checking whether the testing Envoy service had an IP assigned: %q", err))
```

.vib/contour-operator/ginkgo/integration_suite_test.go
```diff
@@ -46,6 +53,115 @@ func clusterConfigOrDie() *rest.Config {
...
+	conditions, found, err := unstructured.NestedSlice(contour.Object, "status", "conditions")
+	if !found {
+		fmt.Printf("Status conditions for Contour %q were not found", name)
+		return false, err
+	}
+
+	mostRecentConditionUpdate := conditions[len(conditions)-1]
+	conditionToMap, conversionOk := mostRecentConditionUpdate.(map[string]interface{})
+	if !conversionOk {
+		return false, fmt.Errorf("Error converting last Certificate condition to type map[string]interface{};  got %T", conditionToMap)
```
joancafom, line 159:
```suggestion
		return false, fmt.Errorf("Error converting last Contour condition to type map[string]interface{};  got %T", conditionToMap)
```

joancafom (approved):
LGTM!

PR #15184, opened 2023-02-28
[bitnami/keycloak] add support to disable service links

nkreiger:
### Description of the change

Adds configuration option to disable kubernetes service links. 

This is useful for naming conflicts and to resolve `too many arg` issues if you are overloading with too much in one namespace. Which you can be forced to do sometimes...First contrib, hopefully I followed standards correctly, sorry if I didn't.

### Checklist

- [x ] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/). This is *not necessary* when the changes only affect README.md files.
- [x ] All commits signed off and in agreement of [Developer Certificate of Origin (DCO)](https://github.com/bitnami/charts/blob/main/CONTRIBUTING.md#sign-your-work)


bitnami/keycloak/values.yaml
```diff
@@ -35,6 +35,12 @@ namespaceOverride: ""
 ## @param commonLabels Labels to add to all deployed objects
 ##
 commonLabels: {}
+## @param serviceLinks Boolean to disable kubernetes service links
+##
+serviceLinks:
+  ## @param serviceLinks.disabled Disable kubernetes service links (enabled by default in kubernetes)
+  ##
+  disabled: false
```
carrodher, line 43: Thanks for your contribution, there is an integration that is failing due to an issue in the metadata used to generate the README. Could you apply something like this?
```suggestion
## @param serviceLinks.disabled Disable kubernetes service links (enabled by default in kubernetes)
##
serviceLinks:
  disabled: false
```

bitnami/keycloak/templates/statefulset.yaml
```diff
@@ -72,6 +72,9 @@ spec:
       {{- if .Values.podSecurityContext.enabled }}
       securityContext: {{- omit .Values.podSecurityContext "enabled" | toYaml | nindent 8 }}
       {{- end }}
+      {{- if .Values.serviceLinks.disabled }}
+      enableServiceLinks: false
+      {{- end }}
```
migruiz4, line 77: `enableServiceLinks` is only available in Kubernetes 1.13 and later, to keep backward compatibility, I suggest the following change:
```suggestion
      {{- if semverCompare ">= 1.13" (include "common.capabilities.kubeVersion" .) }}
      enableServiceLinks: {{ .Values.enableServiceLinks }}
      {{- end }}
```

migruiz4 (changes requested):
Hi @nkreiger,

Thank you very much for your contribution. Could you please take a look at my suggestion?

I think the value `serviceLinks.disabled` to configure `enableServiceLinks` could be a bit confusing. 

My suggestion is to use the same name as the Kubernetes attribute, to make it intuitive and easier for users to find.

bitnami/keycloak/Chart.yaml
```diff
@@ -26,4 +26,4 @@ name: keycloak
 sources:
   - https://github.com/bitnami/containers/tree/main/bitnami/keycloak
   - https://github.com/keycloak/keycloak
-version: 13.1.3
+version: 13.1.4
```
migruiz4, line 29: Hi @nkreiger,

As this is PR introduces a new feature, could you please bump a minor version? (13.2.0)

migruiz4 (changes requested):
Thank you very much for applying the suggested changes, looks good to me!

Just one last change before we can merge this PR. Could you please bump a minor version?

bitnami/keycloak/Chart.yaml
nkreiger: sure!

migruiz4 (approved):

PR #15296, opened 2023-03-02
[bitnami/common] Allow empty registry name

RafaelOm:
This function is modified in order to allow empty registry name.
This allows companies which have a a default proxy to the registries, not to set the registry in every field.

This changes have been tested in rabbitmq operator bitnami chart.

### Description of the change

This function is modified in order to allow empty registry name.
Currently, this is not possible because it generates an image name like this: /reponame:tag

### Benefits

This allows companies which have a default proxy to the registries, not to set the registry in every field.

### Checklist


corico44 (approved):
LGTM

PR #15315, opened 2023-03-03
[bitnami/cassandra] set sizeLimit only for certs-shared emptyDir

jmederer:
### Description of the change

In our Production Environment we had to limit every emptyDir Volume. And the certs-shared emptyDir is the only one without sizeLimit. In this PR i hardcoded the Limit to 1Gi I think this should be enough for two certificate files ;-) 

### Applicable issues

- fixes #

### Checklist

### DIFF
--- a/bitnami/cassandra/Chart.yaml
+++ b/bitnami/cassandra/Chart.yaml
@@ -23,4 +23,4 @@ name: cassandra
 sources:
   - https://github.com/bitnami/containers/tree/main/bitnami/cassandra
   - http://cassandra.apache.org
-version: 10.0.4
+version: 10.0.5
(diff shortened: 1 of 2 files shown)


corico44:
Hello @jmederer,

Thank you for opening this PR. Did the value given to emptyDir come from a consulted source or how did you set that value?

jmederer:
Hi @corico44 

Did you mean the value sizeLimit? This is documented by https://kubernetes.io/docs/concepts/storage/volumes/#emptydir

Or did you mean the 1Gi? The 1Gi is a random value from my side. I don't think that so mach space will be ever used by certificates and this value with 1Gi should never cause problems. 

Do you have any concerns with this change?

corico44:
Hi @jmederer,

In the [example provided in the documentation](https://kubernetes.io/docs/concepts/storage/volumes/#emptydir-configuration-example) they use the value **500Mi**. I think that as a default value it would be advisable to use the one indicated in the documentation.

corico44 (approved):
LGTM

PR #15329, opened 2023-03-04
mysql: Adding the port mapping for mysqlx protocol

### LINKED ISSUE #15328: Expose the mysqlx protocol
### Name and Version

bitnami/mysql 9.5.1

None

mysql chart does not yet expose the mysql X protocol's port even if mysql is built using the mysqlx plugin. This requires a manual change of the deployment and service after using the helm chart.

```mysqlx.port.enabled: true```

It should expose the mysqlx port

### What do you see instead?

It does not expose the protocol

### Additional information

_No response_

### PR DESCRIPTION
**Pull request by vkhater**
* Added the config for allowing the mysqlx protocol

### Description of the change

Enables the port for mysql X plugin so clients can connect with it

### Benefits

Clients would be able to connect on this protocol with the mysql version

### Possible drawbacks

NONE

### Applicable issues

- fixes #15328

### Additional information
Have added the following

Exposed 33060 on deployment
Exposed 33060 on service

### Checklist

### DIFF
--- a/bitnami/mysql/Chart.yaml
+++ b/bitnami/mysql/Chart.yaml
@@ -25,4 +25,4 @@ name: mysql
 sources:
   - https://github.com/bitnami/containers/tree/main/bitnami/mysql
   - https://mysql.com
-version: 9.5.1
+version: 9.5.2
--- a/bitnami/mysql/templates/primary/statefulset.yaml
+++ b/bitnami/mysql/templates/primary/statefulset.yaml
@@ -186,6 +186,10 @@ spec:
           ports:
             - name: mysql
               containerPort: 3306
+            {{- if .Values.mysqlx.port.enabled }}
+            - name: mysqlx
+              port: 33060
+            {{- end }}
           {{- if not .Values.diagnosticMode.enabled }}
           {{- if .Values.primary.customLivenessProbe }}
           livenessProbe: {{- include "common.tplvalues.render" (dict "value" .Values.primary.customLivenessProbe "context" $) | nindent 12 }}
--- a/bitnami/mysql/templates/primary/svc.yaml
+++ b/bitnami/mysql/templates/primary/svc.yaml
@@ -40,6 +40,12 @@ spec:
       port: {{ .Values.primary.service.ports.mysql }}
       protocol: TCP
       targetPort: mysql
+    {{- if .Values.mysqlx.port.enabled }}
+    - name: mysqlx
+      port: 33060
+      targetPort: mysqlx
+      protocol: TCP
+    {{- end }}
       {{- if (and (or (eq .Values.primary.service.type "NodePort") (eq .Values.primary.service.type "LoadBalancer")) .Values.primary.service.nodePorts.mysql) }}
       nodePort: {{ .Values.primary.service.nodePorts.mysql }}
       {{- else if eq .Values.primary.service.type "ClusterIP" }}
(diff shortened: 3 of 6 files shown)

### CODE REVIEW

**Comment by vkhater** [comment]
@fmulero can you please review this?

File: bitnami/mysql/Chart.yaml [hunk 1]
```diff
@@ -25,4 +25,4 @@ name: mysql
 sources:
   - https://github.com/bitnami/containers/tree/main/bitnami/mysql
   - https://mysql.com
-version: 9.5.1
+version: 9.5.2
```
fmulero, line 28: You are adding a new feature. According to [semver](https://semver.org/) you should bump the minor version
```suggestion
version: 9.6.0
```

bitnami/mysql/templates/primary/svc.yaml
```diff
@@ -40,6 +40,12 @@ spec:
       port: {{ .Values.primary.service.ports.mysql }}
       protocol: TCP
       targetPort: mysql
+    {{- if .Values.mysqlx.port.enabled }}
+    - name: mysqlx
+      port: 33060
```
fmulero, line 45: I'd parameterize the port as it is done for the `mysql` one
```suggestion
      port: {{ .Values.primary.service.ports.mysqlx }}
```

bitnami/mysql/templates/primary/statefulset.yaml
```diff
@@ -186,6 +186,10 @@ spec:
           ports:
             - name: mysql
               containerPort: 3306
+            {{- if .Values.mysqlx.port.enabled }}
```
fmulero, line 189: Does it make sense for replicas also?

fmulero (commented):
Hi @vkhater. Thanks a lot for your contribution.

This contribution seems very specific and requires the installation of an additional plugin. Have you considered somehing more generic like `extraPorts`

https://github.com/bitnami/charts/blob/102ad241d4b84ba50ec4c50bedce290adb4eebb1/template/CHART_NAME/templates/service.yaml#L49-L51

bitnami/mysql/templates/primary/svc.yaml
vkhater: Updated to use the parameter. Thanks for noticing it.

bitnami/mysql/templates/primary/statefulset.yaml
vkhater: I am not sure if replicas can also listen on mysqlx protocol.

vkhater:
@fmulero: Thanks for the quick feedback. I was thinking of using the `service.extraPorts` but I could not find the same port exposed in pod template. So I had to go ahead with this specific use case. Do let me know if I should use that in statefuleset template too.

vkhater:
> Hi @vkhater. Thanks a lot for your contribution.
> 
> This contribution seems very specific and requires the installation of an additional plugin. Have you considered somehing more generic like `extraPorts`
> 
> https://github.com/bitnami/charts/blob/102ad241d4b84ba50ec4c50bedce290adb4eebb1/template/CHART_NAME/templates/service.yaml#L49-L51

I have now used the `extraPorts` in service

fmulero (commented):
Keeping in mind these two comments:
* https://github.com/bitnami/charts/pull/15329/files#r1127758629
* https://github.com/bitnami/charts/pull/15329#issuecomment-1459638341

I think will be better if we create 2 new values:
* `primary.extraPorts`
* `secondary.extraPorts`

That would allow users set additional ports in the statefulsets according to their installed plugins or needs and it is not coupled to a particular feature. WDYT?

vkhater:
> * extraPorts

Thanks for the comment @fmulero and I like the idea of exposing two `extraPorts`. I have made the required changes.

bitnami/mysql/templates/primary/statefulset.yaml
vkhater: have made the necessary changes.

vkhater:
@fmulero can you please review this?

fmulero (changes requested):
Great! Please review the indentation. Simple installation with those files is failing because of that.

PR #15411, opened 2023-03-09
[bitnami/redis-cluster] Use SIGTERM on timeout for probes

air3ijai:
### Description of the change

 Change signal to [SIGTERM](https://www.gnu.org/software/libc/manual/html_node/Termination-Signals.html) on probes
> The SIGTERM signal is a generic signal used to cause program termination. Unlike SIGKILL, this signal can be blocked, handled, and ignored. It is the normal way to politely ask a program to terminate.
>
> The shell command kill generates SIGTERM by default. 

It is the same changes like we recently did for the redis #15057

### Benefits

 Using `SIGTERM` instead of the `SIGQUIT` on probes will remove unnecessary coredumps.

### Possible drawbacks

 Probably nothing and in the case of the issue, `redis-cli` will generate a proper coredump.

### Applicable issues

Similar to the #15055

### Checklist

### DIFF
--- a/bitnami/redis-cluster/Chart.yaml
+++ b/bitnami/redis-cluster/Chart.yaml
@@ -23,4 +23,4 @@ name: redis-cluster
 sources:
   - https://github.com/bitnami/containers/tree/main/bitnami/redis
   - http://redis.io/
-version: 8.3.10
+version: 8.3.11
--- a/bitnami/redis-cluster/templates/scripts-configmap.yaml
+++ b/bitnami/redis-cluster/templates/scripts-configmap.yaml
@@ -24,7 +24,7 @@ data:
     if [ ! -z "$REDIS_PASSWORD" ]; then export REDISCLI_AUTH=$REDIS_PASSWORD; fi;
     {{- end }}
     response=$(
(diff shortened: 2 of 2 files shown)


aoterolorenzo:
Hi @air3ijai, thanks for the contribution.

The changes seems fine but I cannot see changes in any probes, only a timeout modification. Could you explain it?

air3ijai:
@aoterolorenzo, hello

We probably can refer to the [[bitnami/redis] - Probes uses SIGQUIT and generates coredumps](15055) as same changes were [already applied to the redis chart](15057).

And in shorts, `-s` parameter for `timeout` utility is the signal we send on timeout, not timeout value. Timeout value is another option, we send as a first argument (`$1`) and there no changes.
```
> timeout -s 3 $1
< timeout -s 15 $1
```

For some (unknown?) reason we send `SIGQUIT (3)` signal on timeout and it by design produces coredump which is not an expected behaviour. The correct signal should be `SIGTERM (15)`.

aoterolorenzo (approved):
LGTM! Sorry for the previous confusion, thought the -s was actually for seconds.

PR #15448, opened 2023-03-10
[bitnami/contour] Support all securitycontext fields

### LINKED ISSUE #15392: [Contour] Envoy not exposing all SecurityContext fields breaks security policies
### Name and Version

bitnami/contour:11.0.1

amd64

1. Using the values:
```yaml
contour:
    containerSecurityContext:
      runAsGroup: 1000
      capabilities:
        drop: ["ALL"]
      seccompProfile:
        type: "RuntimeDefault"
  envoy:
    containerSecurityContext:
      runAsGroup: 1000
      capabilities:
        drop: ["ALL"]
      seccompProfile:
        type: "RuntimeDefault"
```
It seems not all securitycontext fields are exposed (some make more sense for a pod than others, depending on what the use case is). I've set a quite hard policy on my AKS cluster which needs a seccompProfile to be set.
2. In envoy at least, I saw that not all fields are integrally copied. For contour, it seems that a nice trace-off was found, because I can either specify enabled and some fields or directly set my securityContext.
3. Not setting it prevents it from being deployed

See my values above

That I can set all securitycontext fields

### What do you see instead?

Not all securitycontext fields are copied verbatim

UXabre:
### Description of the change

This change removes the rigor from the securityContext settings and just allows to set everything you need (or what company policies dictate is needed).

This is a non-breaking change as the values already provide with safe defaults + default fields to use. It is just adding the ability to set the remainder of fields.

### Benefits

The runtime chart can be compliant with whatever policy is set or needed.

### Possible drawbacks

The limitation is that, before the fields that make sense are only used, in the new case, one would need to prevent entering garbage fields. This is only true to some extent, as some of the securitycontext configs already copied the entire value over verbatim.

### Applicable issues

- fixes #15392

### Checklist

### DIFF
--- a/bitnami/contour/Chart.yaml
+++ b/bitnami/contour/Chart.yaml
@@ -25,4 +25,4 @@ sources:
   - https://github.com/envoyproxy/envoy
   - https://github.com/bitnami/containers/tree/main/bitnami/contour
   - https://projectcontour.io
-version: 11.0.2
+version: 11.0.3
--- a/bitnami/contour/templates/contour/deployment.yaml
+++ b/bitnami/contour/templates/contour/deployment.yaml
@@ -225,8 +225,7 @@ spec:
       dnsPolicy: ClusterFirst
       serviceAccountName: {{ include "contour.contourServiceAccountName" . }}
       {{- if .Values.contour.podSecurityContext.enabled }}
-      securityContext:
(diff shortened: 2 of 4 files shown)


Mauraza (approved):
Hi @UXabre,

Thanks for the contribution

PR #15461, opened 2023-03-11
[bitnami/kube-prometheus] Use actual nulls in alertmanager config

dolohow:
On start, alertmanager spits out error
```
undefined receiver "null" used in route
```
the solution is use actual null's instead of null strings

https://stackoverflow.com/questions/65387198/prometheus-undefined-receiver-null-used-in-route

### Description of the change
Use actual nulls instead of null strings

### Benefits
There is no error on startup on alertmanager and you can reload config without problems.

### Possible drawbacks
None to my knowledge

### Checklist

### DIFF
--- a/bitnami/kube-prometheus/values.yaml
+++ b/bitnami/kube-prometheus/values.yaml
@@ -1751,13 +1751,13 @@ alertmanager:
       group_wait: 30s
       group_interval: 5m
       repeat_interval: 12h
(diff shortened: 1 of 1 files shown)


javsalgar:
Hi,

Thank you so much for the PR! You may need to add @skip (check the comments above the section) so the metadata action passes. Could you also bump the chart version?

PR #15464, opened 2023-03-11
[bitnami/redis] Update values.yaml to provide out-of-the-box compliance with "Restricted" level PSS

udf2457:
Suggested out-of-the-box compliance with "Restricted" level PSS(https://kubernetes.io/docs/concepts/security/pod-security-standards/).

Also added podSecurityContext vars at `sentinel:` level which were previously missing (only `containerSecurityContext` was there before)

### DIFF
--- a/bitnami/redis/values.yaml
+++ b/bitnami/redis/values.yaml
@@ -266,6 +266,9 @@ master:
   podSecurityContext:
     enabled: true
     fsGroup: 1001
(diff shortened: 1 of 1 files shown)


carrodher:
Thanks for your contribution, please note it is needed to bump the chart version in the Chart.yaml as well as add the required metadata in the values.yaml for any new parameter. That's the reason why there are some verification failing

PR #15523, opened 2023-03-16
[bitnami/contour] Adds support for Overload Manager (#15513)

### LINKED ISSUE #15513: Add support for Overload Manager
### Name and Version

bitnami/contour:11.0.3

It is not yet possible to enable the Envoy Overload Manager via Contour ([documentation](https://projectcontour.io/docs/main/config/overload-manager/)) through this chart.
Have searched the issues but found no mention of this being discussed already.

The functionality is enabled by adding the parameter `--overload-max-heap=$bytes` to the `contour bootstrap` command.
It would e.g. be sufficient to introduce the `extraArgs` value for this call as well - i.e. adding code similar to [how it is done for the Envoy call here] (https://github.com/bitnami/charts/blob/main/bitnami/contour/templates/envoy/deployment.yaml#L133) at the location where `contour bootstrap` [is being called.](https://github.com/bitnami/charts/blob/main/bitnami/contour/templates/envoy/deployment.yaml#L257)

Alternatively a distinct value for the amount of bytes e.g. `.values.contour.overload-max-heap` could be introduced. We would only render the parameter & value whenever it is filled with numeric content.

Would be willing to work on this.

_No response_

BerGer23:
### Description of the change

Adds support for enabling Envoy's Overload Manager through Contour (documentation can be found [here](https://projectcontour.io/docs/1.23/config/overload-manager/)).

### Benefits

Exposing this configuration will allow users better control over Envoy's heap behaviour in high footprint situations.

### Possible drawbacks

To me knowledge none. Default behaviour will stay the same.

### Applicable issues

- fixes #15513

### Additional information

Ran into the issue of Helm rendering high integer values scientifically/as floats, worked around it by making the value an explicit String and casting to `int` in the templates. 
Not sure whether this is _the bitnami way_ - if you have another suggestion, I am happy to adjust it.

### Checklist


bitnami/contour/Chart.yaml
```diff
@@ -25,4 +25,4 @@ sources:
   - https://github.com/envoyproxy/envoy
   - https://github.com/bitnami/containers/tree/main/bitnami/contour
   - https://projectcontour.io
-version: 11.0.4
+version: 11.0.5
```
marcosbc, line 28: Since this is adding a new feature, we should at least increase the minor

marcosbc (changes requested):
Thanks for the contribution! I have added a comment about the versioning change, but the rest looks fine.

marcosbc (approved):
LGTM, thanks!

PR #15530, opened 2023-03-16
[bitnami/mastodon] Better minio s3 TLS and Secure WebSocket support when using cert-manager to terminate TLS

joelparker:
### Description of the change

This change introduces 2 settings in values.yaml to allow mastodon-web to work when used with cert-manager or another ingress that terminates TLS.  In this case mastodon-web can incorrectly create broken links to S3 when using minio instead of an external S3 Object Store and also break web socket connections by defaulting to the insecure ws:// protocol.

useSecureWebSocket - defaults to false.  When set to true it will set STREAMING_API_BASE_URL using the secure web socket protocol (was://) instead of ws://.  This will allow the Streaming URL to work when the page is served as TLS.

forceS3Https - defaults to false.  When set to true Mastodon's S3_PROTOCOL will always use https as a protocol.  When using an external S3 provider this usually isn't a problem setting S3_PROTOCOL but when minio is used and is TLS terminated prior to the deployment S3_PROTOCOL will be set to http which causes profile pictures img URLS served by mastodon-web to use use the wrong protocol.
### Benefits

When using cert-manager to manage and provision TLS where an ingress terminates TLS prior to the container this allows mastodon-web to correctly generate links that won't be blocked by the browser's Content Security Policy.

### Possible drawbacks

Ideally I would have used minio.tls.enabled but that then causes the the chart to either have auto generated self signed certificates which are rejected by the browser or minio.tls.existingSecret which isn't used.  

### Checklist

### DIFF
--- a/bitnami/mastodon/Chart.yaml
+++ b/bitnami/mastodon/Chart.yaml
@@ -42,4 +42,4 @@ name: mastodon
 sources:
   - https://github.com/bitnami/containers/tree/main/bitnami/mastodon
   - https://github.com/mastodon/mastodon/
-version: 1.1.4
+version: 1.2.0
--- a/bitnami/mastodon/Chart.yaml
+++ b/bitnami/mastodon/Chart.yaml
@@ -42,4 +42,4 @@ name: mastodon
 sources:
   - https://github.com/bitnami/containers/tree/main/bitnami/mastodon
   - https://github.com/mastodon/mastodon/
-version: 1.1.3
+version: 1.1.4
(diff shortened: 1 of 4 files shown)


corico44:
Hello @joelparker,

Could you tell me how you tested this change? Also, I think that the version of the chart should not only be a _patch_ but rather modify it to be a _minor_. The value should be:
`1.2.0 instead of 1.1.4`

joelparker:
I'm going to close this pull request and resubmit just the secure web socket work.  forceS3Https requires more work from additional testing I've done.

PR #15543, opened 2023-03-17
[bitnami/prometheus] Add chart

fmulero:
### Description of the change

This PR adds a helm chart for Prometheus.

### Benefits

New product in our catalog

### Possible drawbacks

N/A

### Checklist


bitnami/prometheus/templates/server/extra-list.yaml
```diff
@@ -0,0 +1,4 @@
+{{- range .Values.extraDeploy }}
+---
+{{ include "common.tplvalues.render" (dict "value" . "context" $) }}
+{{- end }}
```
javsalgar, line 4: This file should be at templates/, not templates/server

.vib/prometheus/cypress/cypress.env.json
```diff
@@ -0,0 +1,18 @@
...
+  },
+  "targets": {
+    "alertmanager": {
+      "replicaCount": 2
+    },
+    "prometheus": {
+      "replicaCount": 1
+    }
+  }
+}
```
javsalgar, line 18: Don't we normally use fixtures for this?

.vib/prometheus/cypress/cypress/integration/prometheus_spec.js
```diff
@@ -0,0 +1,25 @@
+/// <reference types="cypress" />
+
+it('allows executing a query and displaying response data for each deployment', () => {
+  const deployments = Cypress.env('deployments');
```
javsalgar, line 4: Check other charts, we normally use fixtures here

.vib/prometheus/goss/goss.yaml
```diff
@@ -0,0 +1,60 @@
...
+    - "--enable-feature={{ .Vars.server.enableFeatures | first }}"
+http:
+  http://localhost:{{.Vars.server.containerPorts.http}}/-/ready:
+    status: 200
+    body:
+    - "Prometheus Server is Ready."
+  http://localhost:{{.Vars.server.containerPorts.http}}/-/healthy:
+    status: 200
+    body:
+    - "Prometheus Server is Healthy."
```
javsalgar, line 48: Is it possible to check the status of the other servers like alertmanager? Using the service name it should be possible to acess. I would also try accessing via the prometheus server service endpoint (if it exists)

.vib/prometheus/vib-action.config
```diff
@@ -0,0 +1 @@
+verification-mode=SERIAL
```
javsalgar, line 1: Could you add a comment here in the PR explaining the reason of using serial verification? Especially taking into account that it is a S4 instance

bitnami/prometheus/README.md
```diff
@@ -0,0 +1,678 @@
...
+server:
+  remoteWrite:
+    - url: http://grafana-mimir-gateway.svc.cluster.local/api/v1/push
+      headers:
+        X-Scope-OrgID: demo
+```
+
+- Install Prometheus and Grafana Mimir charts:
+
+For Helm 3:
```
javsalgar, line 523: We do not currently support any other Helm version different from Helm 3

bitnami/prometheus/templates/NOTES.txt
```diff
@@ -0,0 +1,34 @@
...
+
+  kubectl exec --namespace {{ include "common.names.namespace" . | quote }} -ti <NAME OF THE POD> -- bash
+
+In order to replicate the container startup scripts execute this command:
+
+    /opt/bitnami/prometheus/bin/prometheus --config.file=/opt/bitnami/prometheus/conf/prometheus.yml --storage.tsdb.path=/opt/bitnami/prometheus/data --web.console.libraries=/opt/bitnami/prometheus/conf/console_libraries --web.console.templates=/opt/bitnami/prometheus/conf/consoles
+
+{{- else }}
+
+%%Instructions to access the application depending on the serviceType and other considerations%%
```
javsalgar, line 27: Missing the instructions here

bitnami/prometheus/templates/_helpers.tpl
```diff
@@ -0,0 +1,147 @@
...
+{{- end }}
+
+{{/*
+Get the Prometheus configuration configmap.
+*/}}
+{{- define "prometheus.server.configmapName" -}}
+{{- if .Values.server.existingConfigmap -}}
+    {{- include "common.tplvalues.render" (dict "value" .Values.server.existingConfigmap "context" .) -}}
+{{- else }}
+    {{- printf "%s" (include "common.names.fullname" . ) -}}
```
javsalgar, line 87: No need to use printf if you are using a simple template like %s. With print should be enough
{{- print (include "common.names.fullname" .) -}}
Or even:
{{- include "common.names.fullname" . -}}

bitnami/prometheus/templates/_helpers.tpl
```diff
@@ -0,0 +1,147 @@
...
+{{- end -}}
+
+{{/*
+Get the Prometheus configuration configmap key.
+*/}}
+{{- define "prometheus.server.configmapKey" -}}
+{{- if .Values.server.existingConfigmapKey -}}
+    {{- include "common.tplvalues.render" (dict "value" .Values.server.existingConfigmapKey "context" .) -}}
+{{- else }}
+    {{- printf "prometheus.server.yaml" -}}
```
javsalgar, line 98: Is this filename expected? It seems to me that it should be prometheus.yaml

bitnami/prometheus/templates/_helpers.tpl
```diff
@@ -0,0 +1,147 @@
...
+{{- end -}}
+
+{{/*
+Get the Alertmanager configuration configmap.
+*/}}
+{{- define "prometheus.alertmanager.configmapName" -}}
+{{- if .Values.alertmanager.existingConfigmap -}}
+    {{- include "common.tplvalues.render" (dict "value" .Values.alertmanager.existingConfigmap "context" .) -}}
+{{- else }}
+    {{- printf "%s" (include "prometheus.alertmanager.fullname" . ) -}}
```
javsalgar, line 145: Same as before, no need to use printf

bitnami/prometheus/templates/server/clusterrole.yaml
```diff
@@ -0,0 +1,49 @@
+{{- if .Values.rbac.create }}
+kind: ClusterRole
+apiVersion: {{ include "common.capabilities.rbac.apiVersion" . }}
+metadata:
+  name: {{ include "common.names.fullname" . }}
```
javsalgar, line 5: If this is inside the "server" section, shouldn't we use the naming of the server instead of the chart?
Also, if this is a cluster-wide resource, then you should use *.fullname.namespace to avoid issues with the parallel
execution of VIB. This also ensures that you can use the same release name in different namespaces

bitnami/prometheus/values.yaml
```diff
@@ -0,0 +1,1476 @@
...
+    ##
+    ports:
+      http: 80
+      cluster: 9094
+    ## Node ports to expose
+    ## @param alertmanager.service.nodePorts.http Node port for HTTP
+    ## NOTE: choose port between <30000-32767>
+    ##
+    nodePorts:
+      http: ""
```
javsalgar, line 491: Missing reference to the cluster nodePort

bitnami/prometheus/values.yaml
```diff
@@ -0,0 +1,1476 @@
...
+      ##     key: |-
+      ##       -----BEGIN RSA PRIVATE KEY-----
+      ##       ...
+      ##       -----END RSA PRIVATE KEY-----
+      ##     certificate: |-
+      ##       -----BEGIN CERTIFICATE-----
+      ##       ...
+      ##       -----END CERTIFICATE-----
+      ##
+      secrets: []
```
javsalgar, line 1168: I think this is not being used in the template

bitnami/prometheus/values.yaml
```diff
@@ -0,0 +1,1476 @@
...
+## @section Persistence Parameters
+##
+
+## Enable persistence using Persistent Volume Claims
+## ref: https://kubernetes.io/docs/user-guide/persistent-volumes/
+##
+persistence:
+  ## @param persistence.enabled Enable persistence using Persistent Volume Claims. If you have multiple instances (server.repicacount > 1), please considere using an external storage service like Thanos or Grafana Mimir
+  ##
+  enabled: false
```
javsalgar, line 1360: Does it make sense to have a general persistence section instead of being inside the `server` section?

bitnami/prometheus/values.yaml
```diff
@@ -0,0 +1,1476 @@
...
+  ## @param persistence.mountPath Path to mount the volume at.
+  ##
+  mountPath: /opt/bitnami/prometheus/data
```
javsalgar, line 1363: Any reason for not using /bitnami/prometheus/data as we normally do with volumes?

bitnami/prometheus/values.yaml
```diff
@@ -0,0 +1,1476 @@
...
+  ##       - pods
+  ##     verbs:
+  ##       - get
+  ##       - list
+  ##
+  rules: []
+
+## Prometheus metrics
+##
+metrics:
```
javsalgar, line 1473: This is not being used

javsalgar (changes requested):

bitnami/prometheus/values.yaml
fmulero: Removed

bitnami/prometheus/values.yaml
fmulero: 👍

bitnami/prometheus/values.yaml
fmulero: There is no specific reason. Changed

.vib/prometheus/vib-action.config
fmulero: Changed. We are setting now the `ClusteRrole` name appending the namespace, we don't need serial verification anymore.

bitnami/prometheus/README.md
```diff
@@ -0,0 +1,674 @@
...
+> Note: in this example we will use MinIO&reg; (subchart) as the Objstore. Every component will be deployed in the "monitoring" namespace.
+
+- Create a **values.yaml** like the one below for Thanos:
+
+```yaml
+objstoreConfig: |-
+  type: s3
+  config:
+    bucket: thanos
+    endpoint: {{ include "thanos.minio.fullname" . }}.{{ .Release.Namespace }}.svc.cluster.local:9000
```
javsalgar, line 445: Note that thanos.minio.fullname is not defined in the parent chart and could lead to problems. Did you check if this configuration works?

bitnami/prometheus/templates/alertmanager/pdb.yaml
```diff
@@ -0,0 +1,28 @@
+{{- $replicaCount := int .Values.alertmanager.replicaCount }}
+{{- if and .Values.alertmanager.enabled .Values.alertmanager.pdb.create (or (gt $replicaCount 1) .Values.alertmanager.autoscaling.enabled) }}
```
javsalgar, line 2: There is no HPA object for Alertmanager, so no need to check if alertmanager.autoscaling.enabled

bitnami/prometheus/templates/server/deployment.yaml
```diff
@@ -0,0 +1,277 @@
...
+            - --grpc-address=0.0.0.0:10901
+            - --http-address=0.0.0.0:10902
+            - --tsdb.path=/prometheus/
+            {{- if .Values.server.thanos.objectStorageConfig }}
+            - --objstore.config=$(OBJSTORE_CONFIG)
+            {{- end }}
+            {{- if .Values.server.thanos.extraArgs }}
+            {{ toYaml .Values.server.thanos.extraArgs | indent 12  | trim }}
+            {{- end }}
+          {{- if .Values.server.thanos.objectStorageConfig }}
```
javsalgar, line 215: I'd rather check tha objectStorageConfig.secretName is set and define both secretName and secretKey in values.yaml, so they get an entry in the README.md file

bitnami/prometheus/templates/alertmanager/pdb.yaml
fmulero: Good catch!

javsalgar (approved):

PR #15737, opened 2023-03-26
[bitnami/redis] Add notes about RDB compatibility in upgrades

eduardobr:
### Description of the change

Add notes about RDB compatibility to README explaining how to mitigate disruptions in rolling updates caused by upgrades to Redis with forward incompatible RDBs.

In upcoming Redis 7.2 we have a new RDB format.

### Benefits

Smoother rolling update when RDB format changed.

### Possible drawbacks

None afaik

### Checklist

### DIFF
--- a/bitnami/redis/README.md
+++ b/bitnami/redis/README.md
@@ -750,6 +750,18 @@ Find more information about how to deal with common errors related to Bitnami's 
 
 A major chart version change (like v1.2.3 -> v2.0.0) indicates that there is an incompatible breaking change needing manual actions.
 
+### RDB compatibility
+
+It's common to have RDB format changes across Redis&reg; releases where we see backward compatibility but no forward compatibility. For example: version 7.0 can load an RDB created by version 6.2, but version 6.2 can't load an RDB created by 7.0.  
+When that's the case the rolling update can cause replicas to temporarily stop synchronizing while they are running a lower version than master.  
+For example: on a rolling update `master-0` and `replica-2` are updated first from version 6.2 to 7.0, `replica-0` and `replica-1` won't be able to start a full sync with `master-0` because they still run 6.2 and can't understand the RDB format from version 7.0 that master is using.  
+This issue can be mitigated by splitting upgrade in two stages: one for all replicas and another for any master.  
+
+- Stage 1 (replicas only, as there's no master with an ordinal higher than 99):  
+`helm upgrade my-repo/redis --set master.updateStrategy.rollingUpdate.partition=99`  
+- Stage 2 (anything else that is not up to date, in this case only master):  
+`helm upgrade my-repo/redis`  
+
 ### To 17.0.0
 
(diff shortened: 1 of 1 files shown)


bitnami/redis/README.md
```diff
@@ -750,6 +750,18 @@ Find more information about how to deal with common errors related to Bitnami's
...
+It's common to have RDB format changes across Redis&reg; releases where we see backward compatibility but no forward compatibility. For example: version 7.0 can load an RDB created by version 6.2, but version 6.2 can't load an RDB created by 7.0.  
+When that's the case the rolling update can cause replicas to temporarily stop synchronizing while they are running a lower version than master.  
+For example: on a rolling update `master-0` and `replica-2` are updated first from version 6.2 to 7.0, `replica-0` and `replica-1` won't be able to start a full sync with `master-0` because they still run 6.2 and can't understand the RDB format from version 7.0 that master is using.  
+This issue can be mitigated by splitting upgrade in two stages: one for all replicas and another for any master.  
+
+- Stage 1 (replicas only, as there's no master with an ordinal higher than 99):  
+`helm upgrade my-repo/redis --set master.updateStrategy.rollingUpdate.partition=99`  
+- Stage 2 (anything else that is not up to date, in this case only master):  
+`helm upgrade my-repo/redis`  
+
```
dgomezleon, line 764:
```suggestion
### RDB compatibility

It's common to have RDB format changes across Redis&reg; releases where we see backward compatibility but no forward compatibility. A new major version can load an RDB created by version the previous major version, but the opposite is not true.
When that's the case, the rolling update can cause replicas to temporarily stop synchronizing while they are running a lower version than master. For example, on a rolling update `master-0` and `replica-2` are updated first from version v6.2 to v7.0; `replica-0` and `replica-1` won't be able to start a full sync with `master-0` because they are still running v6.2 and can't support the RDB format from version 7.0 that master is now using.  
This issue can be mitigated by splitting the upgrade into two stages: one for all replicas and another for any master.  

- Stage 1 (replicas only, as there's no master with an ordinal higher than 99):  
`helm upgrade my-repo/redis --set master.updateStrategy.rollingUpdate.partition=99`  
- Stage 2 (anything else that is not up to date, in this case only master):  
`helm upgrade my-repo/redis`  

```

dgomezleon (commented):
Thanks a lot for this contribution.

Could you please take a look at my suggestions?

bitnami/redis/README.md
eduardobr: Thanks for review
> "A new major version can load an RDB created by the previous major version, but the opposite is not true"

We have a different RDB format between 7.0 and 7.2, so this is not really about major version.
I will amend text soon

eduardobr: @dgomezleon Changes done according to your suggestions

dgomezleon (approved):
LGTM

Issue #15299, opened 2023-03-02
Quorum queues created during deploy via loadDefinition do not adopt any followers


Millz12:
### Name and Version

bitnami/rabbitmq

### What architecture are you using?

amd64

### What steps will reproduce the bug?

Hello. We're deploying RabbitMQ with 3 replicas in a cluster, and loading in a default configuration via the loadDefinitions parameter. We're creating 2 quorum queues as below:

```        "queues": [
            {
                "name": "user",
                "vhost": "/",
                "durable": true,
                "auto_delete": false,
                "type": "quorum",
                "arguments": {
                    "x-queue-type": "quorum"
                }
            },
            {
                "name": "admin",
                "vhost": "/",
                "durable": true,
                "auto_delete": false,
                "type": "quorum",
                "arguments": {
                    "x-queue-type": "quorum"
                }
            }
        ]
```

The queues are created (along with bindings, user accounts etc) without any errors, however the queues created via this method only allocate the first node without any followers. When creating additional queues manually later via the GUI then the primary node is assigned with 2 followers (which is the expected behaviour).

![image](https://user-images.githubusercontent.com/8068162/222433632-59df8b93-3d90-4fa7-8f49-60820e7cfa2d.png)

We suspect this is because the configuration is applied prior to the cluster being fully created/initialised. 

Is there a way around this without having to apply a workaround of applying the config after deployment?

Thank you

corico44:
Hello @Millz12,

Thank you for opening this issue. It is possible that the logic that is carried out when adding the configuration by file vs manually is different. I'm going to dig deeper to see what the actual difference between the two methods is. I will notify you about this issue when I have more information.

corico44:
Digging deeper, there doesn't seem to be such a distinct logic justifying the behavior that the queues created using loadDefinitions only allocate the first node without any followers. But when creating additional queues manually later via the GUI then the primary node is assigned with 2 followers. Therefore, I am going to create a task internally to better study this behavior.
We will notify you about this issue when we have more information. 
Thanks for reporting it.

Millz12:
Thanks for the update @corico44 

I've found that if you wait until the cluster is fully initialised then load the same definition file from the CLI within a pod then it sets up the queues correctly.

```
$ rabbitmqctl import_definitions load.json
Importing definitions in JSON from a file at "load.json" ...
Successfully started definition import. This process is asynchronous and can take some time.
$ rabbitmq-queues quorum_status admin-events
[...]

PR #15934, opened 2023-04-03
[bitnami/rabbitmq] Updated values.yaml

Millz12:
### Description of the change

Updated the values.yaml file to prevent loading definition file prior to the cluster being fully initialised. This allows the quorum queues defined within the definition file to setup correctly with followers instead of only a master with 0 followers.

### Benefits

The current configuration means if you deploy a cluster of 3 nodes or more, and enable load_definitions with a quorum type queue, the definitions are loaded on the first node to initialise so queues are only built on the first node.

This change waits until the cluster reaches the target size until it loads the definition file which then allows the quorum queue to be setup with a master plus 2 followers.

### Applicable issues

Resolves issue: https://github.com/bitnami/charts/issues/15299

### Checklist


bitnami/rabbitmq/Chart.yaml
```diff
@@ -23,4 +23,4 @@ name: rabbitmq
 sources:
   - https://github.com/bitnami/containers/tree/main/bitnami/rabbitmq
   - https://www.rabbitmq.com
-version: 11.12.2
+version: 11.12.3
```
Mauraza, line 26:
```suggestion
version: 11.13.0
```

Mauraza (commented):
Hi @Millz12,

Thanks for the contribution! As this change the configuration, could you change the version to a minor?

Mauraza:
Hi @Millz12,

Could you be so kind as resolve the conflicts?

Issue #14600, opened 2023-01-30
[bitnami/kafka] TLS enabled brokers do not return certificate chain (cert-manager + pemChainIncluded=true )


kxs-jnadeau:
### Name and Version

bitnami/kafka 20.0.5

### What steps will reproduce the bug?

1.  Deploying a 3 replicas cluster on AKS v1.23,  with cert-manager and external DNS
2. Chart deployed from helm command below.  The existing certs are generated by cert-manager with Sectigo's ACME endpoint.   Like Let'sEncrypt,  the generated secrets only contains a tls.crt (with chain) and a tls.key.    

3.  Looking at what the chart does with the pemChainIncluded flag,   I can see the certs split in the broker pods under /opt/bitnami/kafka/config/certs:

```
 -rw-r--r-- 1 1001 root 1708 Jan 27 20:22 kafka.keystore.key
-rw-r--r-- 1 1001 root 2472 Jan 27 20:22 kafka.keystore.pem
-rw-r--r-- 1 1001 root 4143 Jan 27 20:22 kafka.truststore.pem
-rw-r--r-- 1 1001 root 2472 Jan 27 20:22 xx00
-rw-r--r-- 1 1001 root 2175 Jan 27 20:22 xx01
-rw-r--r-- 1 1001 root 1968 Jan 27 20:22 xx02
```

4.  TLS validation is failing from clients (Java/Python) as shown below with openssl s_connect..   Validation goes ok if I pass the full chain to openssl s_client in a PEM file.

### Are you using any custom parameters or values?

```
helm template demo \
  --set replicaCount=3 \
  --set resources.requests.cpu=0.5 \
  --set resources.requests.memory=1024Mi \
  --set resources.limits.cpu=1 \
  --set resources.limits.memory=2024Mi \
  --set persistence.enabled=true \
  --set persistence.storageClass=managed-csi \
  --set zookeeper.persistence.storageClass=managed-csi \
  --set zookeeper.resources.requests.cpu=0.5 \
  --set zookeeper.resources.requests.memory=512Mi \
  --set zookeeper.resources.limits.cpu=1 \
  --set zookeeper.resources.limits.memory=1024Mi \
  --set auth.clientProtocol=sasl_tls \
  --set auth.externalClientProtocol=sasl_tls \
  --set auth.sasl.mechanisms=scram-sha-512 \
  --set auth.sasl.jaas.clientUsers[0]=test \
[...]

PR #15941, opened 2023-04-03
[bitnami/kafka] Keep intermediate certs in PEM server certificate chain

sophokles73:
### Description of the change

The *setup.sh* script from the *scripts-configmap* currently removes all intermediary certificates from the configured server certificate chain provided in a tls.crt PEM file as created by k8s cert-manager.

This effectively prevents the kafka broker to include any of the intermediary certificates in a TLS handshake with a peer/client.

### Benefits

The PEM file is now being copied as-is to the kafka container.

However, this change alone does not fix the issue yet because the kafka_configure_ssl() function from the "libkafka.sh" bitnami kafka container helper library transforms the keystore PEM file to a multi-line string to be include in the server.properties in a way that is not compatible with the regular expression used by kafka's DefaultSslEngineFactory$PemParser.pemEntries() method when starting up the broker. The regular expression only matches the last certificate in the chain which leads to the broker only serving this certificate in TLS handshakes.

### Applicable issues

This is a pre-requisite for fixing #14600

### Checklist

### DIFF
--- a/bitnami/kafka/Chart.yaml
+++ b/bitnami/kafka/Chart.yaml
@@ -29,4 +29,4 @@ name: kafka
 sources:
   - https://github.com/bitnami/containers/tree/main/bitnami/kafka
   - https://kafka.apache.org/
-version: 21.4.4
+version: 21.4.5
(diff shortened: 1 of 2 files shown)


aoterolorenzo:
Hi @sophokles73, 

Sorry for the delay, since another PR related to this needed to be reviewed in order to decide the best approach. 

Could you provide a brief set of values and commands to test this out?

aoterolorenzo:
Yes, I meant a 'test script'. This kind of cases needs several time to be replicated and tested so would be very helpful if you could provide a set of values and commands to just test the feature. I'm aware that we will need to use the latest image in the chart for this to work, after merged https://github.com/bitnami/containers/pull/29026

PR #15950, opened 2023-04-04
[bitnami/etcd] Use common.labels.matchLabels instead of common.labels.standard in ne…

### LINKED ISSUE #15936: [bitnami/etcd] Networkpolicy uses chart version as podSelector
### Name and Version

bitnami/etcd:8.8.0

None

1. Deploy a deny-all networkpolicy matching all pods in a namespace
2. Deploy etcd with `networkPolicy.enabled=true`
3. do a helm upgrade to a different chart version
4. the new pod fails to become ready as the networkpolicy is now set to the new chart version. This means it cannot reach the old nodes. Simultaneously the 2 running nodes are unavailable since only our default deny-all networkpolicy applies to them.

```
networkPolicy:
  enabled: true
```

The networkpolicy should not select pods based on the chart version, it should instead use only annotations like app.kubernetes.io/instance, app.kubernetes.io/name which do not change between upgrades.

### What do you see instead?

The networkpolicy selects only pods on the current version of the chart, which breaks rolling upgrades.

### Additional information

It's good I tested this on a test cluster before upgrading the production etcd because this is a severe bug on an application that is normally safe to do a rolling upgrade with.

bh-tt:
…tworkpolicy matchLabels

### Description of the change
Use common.labels.matchLabels to select pods belonging to the etcd networkpolicy, as common.labels.standard includes the chart version which breaks rolling upgrades.

### Benefits
Rolling upgrades will work for etcd with networkpolicies enabled.

### Applicable issues

- fixes #15936

### Checklist

### DIFF
--- a/bitnami/etcd/Chart.yaml
+++ b/bitnami/etcd/Chart.yaml
@@ -25,4 +25,4 @@ name: etcd
 sources:
   - https://github.com/bitnami/containers/tree/main/bitnami/etcd
   - https://coreos.com/etcd/
-version: 8.8.1
+version: 8.8.2
--- a/bitnami/etcd/templates/networkpolicy.yaml
+++ b/bitnami/etcd/templates/networkpolicy.yaml
@@ -13,7 +13,7 @@ metadata:
   {{- end }}
 spec:
   podSelector:
-    matchLabels: {{- include "common.labels.standard" . | nindent 6 }}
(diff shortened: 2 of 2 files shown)


bh-tt:
I'm getting an hourly failure to run 'support closing issues' in my fork since I do not have a token, is it possible to disable that job when running outside of the bitnami/charts repository?

rafariossaa (approved):
Thanks for updating this.
LGTM.

PR #15998, opened 2023-04-10
[bitnami/*] Make Helm charts 100% OCI

carrodher:
### Description of the change

Changes in the Helm chart templates (_Chart.yaml_ and _Chart.lock_) were implemented by using the following script:
```bash
for chart in *; do  
  cd $chart || exit
  sed -i '' -e 's@https://charts.bitnami.com/bitnami@oci://registry-1.docker.io/bitnamicharts@g' Chart.yaml
  rm -rf charts/ Chart.lock
  helm dep update
  cd .. || exit
done
```
In this way the _Chart.lock_ and the _charts/_ subfolder were rebuilt by using the OCI Helm charts present at DockerHub instead of the ones from the _index.yaml_.

Apart from that, the documentation and installation instructions were updated to reflect the new preferred way of installing Helm charts via OCI.

You can find more info about the reason behind those changes in [this blog post](https://blog.bitnami.com/2023/04/httpsblog.bitnami.com202304bitnami-helm-charts-now-oci.html).

------

⚠️ GH action is failing because the bitnami/common chart version was not bumped. This is expected since only the _README.md_ was modified


README.md
```diff
@@ -72,9 +54,8 @@ Please refer to the [Quick Start guide](https://helm.sh/docs/intro/quickstart/)
 
 Useful Helm Client Commands:
 
-- View available charts: `helm search repo`
-- Install a chart: `helm install my-release bitnami/<package-name>`
-- Upgrade your application: `helm upgrade`
+- Install a chart: `helm install my-release oci://registry-1.docker.io/bitnamicharts/<chart>`
+- Upgrade your application: `helm upgrade ...`
```
agomezmoron, line 58: I'd add an example if possible

agomezmoron (approved):
LGTM!

fmulero (approved):
LGTM

agomezmoron (approved):

morsik:
Hi,

this excerpt from that blog post is kinda funny:

> We will be moving index.yaml to a reliable CDN but we would like you to note that any possible rate limits that the CDN provider might choose to apply to individual IP addresses consuming Helm Charts at a large scale are out of our control. In addition to this, keeping all the limitations and issues faced when using large Helm chart catalogs in a single index, we strongly recommend users start consuming Helm Charts as OCI packages going forward (although the index will remain available).

I've never had issues with your previous CDN (except when I need to use older software and need to find old index.yaml for that), and now when trying to use OCI on Docker Hub, I hit first IP limitation and I can't use Bitnami Helm Chart from `registry-1.docker.io` :/

So... for some people it is moved to even less reliable place than before :P

carrodher:
I understand that you're facing some issues with the OCI registry, and I'm here to help.

It's important to note that the OCI registry isn't hosted or maintained by our team. If you're encountering problems, I recommend reaching out to the DockerHub community forums. Helm charts, just like container images for several years, are hosted on an external trusted registry like DockerHub, although it's true that during the latest days, they faced some issues. That's why my recommendation is to report there any issues.

Regarding `https://charts.bitnami.com/bitnami`, I want to clarify that although it's currently available, it's not the recommended way to consume Bitnami Helm charts. While we have no immediate plans to decommission it, we'll provide ample notice to users if any changes are on the horizon. If such a decision is ever made, we'll ensure a proper communication strategy is in place to give users sufficient time to adapt.

If you have any more questions or need additional information, please don't hesitate to ask.

PR #16009, opened 2023-04-11
[bitnami/discourse] Updated Discourse README with instructions for plugins

RickFoland:
Directions were added to the Discourse README for installing Discourse plugins. It wasn't obvious how to add the original startup command in order to get the plugin install commands to run.


migruiz4:
Hi @RickFoland,

Would you like to contribute by implementing this feature in the bitnami/discourse chart? Maybe under the value `discourse.plugins`.

This way it would be easier to support and maintain this feature as official, rather than documenting a custom use case that is more likely to be outdated over time.

Additionally, please note that the logic described in your PR is not correct, as commands after `run.sh` will not be executed as it would start the discourse process. 
```yaml
command:
  - /bin/bash
args:
  - -c
  - |
    /opt/bitnami/scripts/discourse/entrypoint.sh /opt/bitnami/scripts/discourse/run.sh &&
    cd /opt/bitnami/discourse &&
    RAILS_ENV=production bundle exec rake plugin:install repo=https://github.com/discourse/discourse-oauth2-basic &&
    RAILS_ENV=production bundle exec rake assets:precompile
```

My suggestion would be the following:
```yaml
  command:
    - /bin/bash
  args:
    - -c
    - |
      pushd "/opt/bitnami/discourse" >/dev/null || exit
      RAILS_ENV=production bundle exec rake plugin:install repo=https://github.com/discourse/discourse-oauth2-basic
      popd >/dev/null || exit
      /opt/bitnami/scripts/discourse/entrypoint.sh /opt/bitnami/scripts/discourse/run.sh
```

Pushd / popd are used to restore the original working directory.
The precompile command is not needed, as it would be executed during the container initialization.

This could be implemented in the chart as:
```
          {{- if .Values.diagnosticMode.enabled }}
          command: {{- include "common.tplvalues.render" (dict "value" .Values.diagnosticMode.command "context" $) | nindent 12 }}
          {{- else if .Values.discourse.command }}
          command: {{- include "common.tplvalues.render" (dict "value" .Values.discourse.command "context" $) | nindent 12 }}
          {{- else }}
          command:
            - /bin/bash
          {{- end }}
          {{- if .Values.diagnosticMode.enabled }}
          args: {{- include "common.tplvalues.render" (dict "value" .Values.diagnosticMode.args "context" $) | nindent 12 }}
          {{- else if .Values.discourse.args }}
          args: {{- include "common.tplvalues.render" (dict "value" .Values.discourse.args "context" $) | nindent 12 }}
          {{- else }}
          args:
            - -c
            - |
              {{- if .Values.discourse.plugins }}
              pushd "/opt/bitnami/discourse" >/dev/null || exit
              {{- range $plugin := .Values.discourse.plugins }}
              RAILS_ENV=production bundle exec rake plugin:install repo=$plugin
              {{- end }}
              popd >/dev/null || exit
              {{- end }}
              /opt/bitnami/scripts/discourse/entrypoint.sh /opt/bitnami/scripts/discourse/run.sh
          {{- end }}
```

obeyler:
Note that in case of adding plugins, the statupProbe must be set as the start process can take a long long time... for our case with only 4 plugins it takes 10min !

migruiz4:
Thanks for the suggestion @obeyler! 
We should definitely add that as a side note for this feature.

@RickFoland did you have time to look into my previous comment? Please let us know If you find any issues implementing this as a feature into the chart, or if you would prefer if we went ahead and implemented it ourselves.

RickFoland:
I haven't had time to go back and test it yet. One thing that stood out was about the original logic not working. I was able to implement it that way and it was the only way I could get the plugins to install (I can verify that they install from the console). For some reason the install would hang when I tried to run it before the start commands (but maybe I tried to do it incorrectly)

RickFoland:
I've had a bit more time to test and it turns out that neither of the commands work on their own for initially installing the plugin. I pushed the chart to a new env and the Oauth plugin didn't work because it wasn't installed. I tired using the command you gave and had the same issue. 

Then I tried manually installing it and it worked. My guess is that the manual install is saved on the PVC for the discourse pod and the plugin doesn't actually need to be installed every time.

migruiz4:
Hi @RickFoland,

As I see you had some difficulties in the process, I have created this PR to submit my proposal: https://github.com/bitnami/charts/pull/16196

I would appreciate it if you took a look at the changes and even checkout the branch and try it, so we can ensure the changes meet your needs.

In my local tests I didn't find any issues installing the Oauth2 plugin:
image

migruiz4:
Hi @RickFoland,

Version 10.2.0 of the `bitnami/discourse` is now live, including support for the value `discourse.plugins`.

I will then close this issue, as the `discourse.plugins` value is self-descriptive and I don't think a new section in the README would be required.

Please give try out this new version and let me know if you find any issues.

PR #16021, opened 2023-04-11
Added possibility to have a Custom Secretname for the certificate

VF-mbrauer:
### Description of the change

The change will make it possible to define your own secretName which is part of the ingress rule configuration.
Right now, it is concatenated with some static names and the hostname. Due to existing certification automation, it
might be for several people to specify their own pattern.
If this value will not be set, then the original one before this PR will be taken.

### Benefits

Flexibility to configure the secretName.

### Possible drawbacks

No drawback in case the value file will be empty or not set.

### Applicable issues

n.a.

### Additional information

n.a.

### Checklist


bitnami/oauth2-proxy/Chart.yaml
```diff
@@ -30,4 +30,4 @@ name: oauth2-proxy
 sources:
   - https://github.com/bitnami/containers/tree/main/bitnami/oauth2-proxy
   - https://github.com/oauth2-proxy/oauth2-proxy
-version: 3.4.9
+version: 3.4.10
```
Mauraza, line 33: I think is a better upgrade to a minor since you are adding a new possibility.
```suggestion
version: 3.5.0
```

Mauraza (commented):
Hi @VF-mbrauer,

Thanks for the contribution (again 😃)!! Could you check the small comments?

bitnami/oauth2-proxy/templates/tls-secrets.yaml
```diff
@@ -1,5 +1,4 @@
-{{- if .Values.ingress.enabled }}
-{{- if .Values.ingress.secrets }}
+{{- if and .Values.ingress.enabled (not .Values.ingress.existingSecretName) }}
```
Mauraza, line 1:
```suggestion
{{- if and .Values.ingress.enabled (not .Values.ingress.existingSecretName) }}
{{- if .Values.ingress.secrets }}
```

Mauraza (commented):
Sorry, I missed a line in the tls-secrets.yaml, could you add it, please?

PR #16112, opened 2023-04-18
[bitnami/thanos] Added support for query-url for Thanos ruler 

k4kratik:
### Description of the change
- Added argument in Thanos Ruler Staefulset for `--query-url`
- Added param in `query_url` in values.yaml so this value can be supplied back to Thanos Ruler Statefulset.
- Updated Docs

### Benefits
This will add enable users to check their Thanos Ruler rules in Thanos Query (or Query Frontend) by clicking on `alert` or `expr` .

### Possible drawbacks
You should provide correct Thanos Query URL otherwise you won't be able to see your thanos rules/expr from Ruler UI.
**param name** : `ruler.query_url`

### Checklist


bitnami/thanos/templates/ruler/statefulset.yaml
```diff
@@ -135,6 +135,7 @@ spec:
             {{- if and $query.enabled .Values.ruler.dnsDiscovery.enabled }}
             - --query=dnssrv+_http._tcp.{{ include "common.names.fullname" . }}-query.{{ .Release.Namespace }}.svc.{{ .Values.clusterDomain }}
             {{- end }}
+            - --alert.query-url={{ .Values.ruler.query_url }}
```
fmulero, line 138: As thanos-query service is part of the chart, I'd calculate the url asw we do for example in: https://github.com/bitnami/charts/blob/0594dcb3c0161515aa8ecf75cfba0a1f5ab468c0/bitnami/thanos/templates/NOTES.txt#L30

Maybe is worth to create a helper to avoid that ugly piece of code twice

bitnami/thanos/values.yaml
```diff
@@ -4205,7 +4208,7 @@ minio:
     rootPassword: ""
   ## @param minio.defaultBuckets Comma, semi-colon or space separated list of MinIO&reg; buckets to create
   ##
-  defaultBuckets: 'thanos'
+  defaultBuckets: "thanos"
```
fmulero, line 4211: Good catch!

bitnami/thanos/values.yaml
```diff
@@ -2762,6 +2762,9 @@ ruler:
   ##
   dnsDiscovery:
     enabled: true
+  ## @param ruler.query_url Thanos query/query-frontend URL to link in Ruler UI.
+  ##
+  query_url: http://thanos-query:9090
```
fmulero, line 2767: I think is not needed. Also it could infer to error if we deploy the chart with a name different than `thanos`

fmulero (commented):
Good catch @k4kratik! Thanks a lot for reporting the issue and creating the PR to solve it. You the best!

I have some concerns, could you take them a glance?

bitnami/thanos/values.yaml
k4kratik: I thought we need to supply a default value, then should I keep it empty?

bitnami/thanos/templates/ruler/statefulset.yaml
k4kratik: Hi @fmulero!
Thank you for reviewing and kind comments 🤗 

**Re: calculating the URL**
Suppose someone has a public routable domain(for example) for their Thanos query service, Now, if we calculate and provide them our k8s svc FQDN of Thanos Query, it won't work unless they have set up some kind of VPN or something to access k8s service in browser.
So the point is, if we calculate and use the value for Thanos query for  `--alert.query-url` param, we might be restricting the users for values of query-url which they can use with this setup. (because now no one will be able to set any other value than the calculated one)

thoughts?

k4kratik: @fmulero Can you please check this ?

fmulero: We can keep both approaches, if the user provides the `queryURL` that value should be used. In other case, if `queryURL` is empty we should calculate it.

bitnami/thanos/values.yaml
fmulero: We can keep it empty as I mentioned before. Please follow camel case (`queryURL` instead of `query_url`) for this name according to the name convention in this file

k4kratik: Sure!

fmulero (changes requested):
Sorry @k4kratik I think I didn't explain very well. I left a suggestion

bitnami/thanos/templates/ruler/statefulset.yaml
k4kratik: Totally agree. Will do the same!

bitnami/thanos/templates/ruler/statefulset.yaml
```diff
@@ -135,6 +135,7 @@ spec:
             {{- if and $query.enabled .Values.ruler.dnsDiscovery.enabled }}
             - --query=dnssrv+_http._tcp.{{ include "common.names.fullname" . }}-query.{{ .Release.Namespace }}.svc.{{ .Values.clusterDomain }}
             {{- end }}
+            - --alert.query-url={{ .Values.ruler.queryUrl }}
```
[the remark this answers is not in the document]
k4kratik: You explained it very clearly, actually, I was trying to figure it out @fmulero :)
Let me check. Thanks!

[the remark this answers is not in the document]
k4kratik: I'd really love to learn and contribute using Helper functions, possibly can you point me to relevant documentation(s)? @fmulero.

k4kratik:
Hi @fmulero 
Can you please check and test now?
I am pretty confident about this one, As I tested on local and It was working as expected. 
Thanks and Apologies for the delay :)

PR #16158, opened 2023-04-21
[bitnami/keycloak]: adjust default service to ClusterAPI

mweibel:
### Description of the change

Exposing keycloak by default as a LoadBalancer might not be the best option since people unaware expose keycloak to the internet without an ingress in front which is what people would want usually.

### Benefits

Slightly better defaults. Most people running keycloak will run it behind an ingress anyway. 

### Possible drawbacks

It's a breaking change of the chart.

### Checklist

Please let me know if this is something you'd accept - if yes I'll update the Chart.yaml and README files accordingly.


javsalgar:
Hi,

Thank you for creating the ticket. Could you bump the version of the chart? Otherwise the action will continue failing.

mweibel:
Thanks @javsalgar for your reply. Yeah I don't mind doing that - I didn't do it because I wasn't sure if you'd accept this PR. 

Given that this is a potential breaking change to change the default - would you agree I'd need to bump the major version?

rafariossaa:
In this case I would bump the major version and add a note in the README so users would know about this change.

mweibel:
@rafariossaa @javsalgar thanks for the feedback - Chart version and README is updated.

rafariossaa:
Hi, 
  I seems you need to rebase your changes.

rafariossaa (approved):
LGTM

PR #16234, opened 2023-04-26
[bitnami/redis] Fix replica-announce-ip wrong number of arguments whe…

### LINKED ISSUE #16177: [bitnami/redis] FATAL CONFIG FILE ERROR when (useHostnames: false and useExternalDNS enabled)
### Name and Version

bitnami/redis 17.9.4

amd64

install helm chart with custom parameters

```yaml
architecture: replication
useHostnames: false
sentinel:
  enabled: true
  quorum: 2
  service:
    type: NodePort
    nodePorts:
      redis: 32010
      sentinel: 32020
useExternalDNS:
  enabled: true
  suffix: example.com
  additionalAnnotations:
    endpoints-type: HostIP
```

```
 12:46:52.19 INFO  ==> about to run the command: timeout 220 redis-cli -h redis.test-redis.svc.cluster.local -p 32020 sentinel get-master-addr-by-name master
Could not connect to Redis at redis.test-redis.svc.cluster.local:32020: Connection refused
 12:46:52.22 ERROR ==> Redis master host is configured incorrectly **(host: , port: 32011)**
 12:46:52.03 INFO  ==> about to run the command: timeout 220 redis-cli -h redis.test-redis.svc.cluster.local -p 32020 sentinel get-master-addr-by-name master
Could not connect to Redis at redis.test-redis.svc.cluster.local:32020: Connection refused
 12:46:52.07 INFO  ==> Configuring the node as master

*** FATAL CONFIG FILE ERROR (Redis 7.0.10) ***
Reading the configuration file, at line 7
>>> 'replica-announce-ip'
wrong number of arguments
```
 

### What do you see instead?

```
{{- if .Values.useHostnames }}
echo "${full_hostname}"
{{- else }}
until getent hosts "${full_hostname}"
do 
echo waiting for redis
sleep 10
done;
getent hosts "${full_hostname}" | awk '{ print $1 ; exit }'
{{- end }}
```

### Additional information

`getent hosts "${full_hostname}" | awk '{ print $1; exit }'` 
passed an empty field to `replica-announce-ip` because the server can't find NXDOMAIN
need a little more time to create a record

AlehYarmalovich:
Signed-off-by: Aleh Yarmalovich <yarmolmi@gmail.com>

### Description of the change

This change fix replica-announce-ip wrong number of arguments when useExternalDNS enabled.
When the useExternalDNS option is enabled and useHostnames is disabled, Redis containers fail because the external DNS doesn't have time to create an record, and the getent command gets empty output.

### Benefits

This solves the issue reported [here](https://github.com/bitnami/charts/issues/16177)

### Possible drawbacks

None from the tests I performed

- Launched a new deployment with the useHostnames parameter set to false and useExternalDNS set to true.
- Launched a new deployment with the useHostnames parameter set to true and useExternalDNS set to true.
- Launched a new deployment with the useHostnames parameter set to true and useExternalDNS set to false.
- Removed a secondary/master replicas and confirmed everything worked as expected.

### Applicable issues

- fixes #16177

### Checklist

### DIFF
--- a/bitnami/redis/Chart.yaml
+++ b/bitnami/redis/Chart.yaml
@@ -22,4 +22,4 @@ maintainers:
 name: redis
 sources:
   - https://github.com/bitnami/containers/tree/main/bitnami/redis
-version: 17.11.1
+version: 17.10.2
--- a/bitnami/redis/Chart.yaml
+++ b/bitnami/redis/Chart.yaml
@@ -22,4 +22,4 @@ maintainers:
 name: redis
 sources:
   - https://github.com/bitnami/containers/tree/main/bitnami/redis
-version: 17.10.1
+version: 17.11.1
--- a/bitnami/redis/templates/scripts-configmap.yaml
+++ b/bitnami/redis/templates/scripts-configmap.yaml
@@ -54,7 +54,16 @@ data:
         {{- if .Values.useHostnames }}
         echo "${full_hostname}"
         {{- else }}
-        until getent hosts "${full_hostname}"; do sleep 5; done | awk '{ print $1 ; exit }';
+        retry_count=0
+        until getent hosts "${full_hostname}" | awk '{ print $1; exit }' | grep .; do 
+        if [[ $retry_count -lt {{ .Values.useHostnamesThreshold }} ]]; then
+            sleep {{ .Values.useHostnamesTimeout }}
+        else
+            error "IP address for ${full_hostname} not found"
+            exit 1
+        fi
+        ((retry_count++))
+        done
         {{- end }}
     }
 
@@ -278,7 +287,16 @@ data:
         {{- if .Values.useHostnames }}
         echo "${full_hostname}"
         {{- else }}
-        until getent hosts "${full_hostname}"; do sleep 5; done | awk '{ print $1 ; exit }';
+        retry_count=0
+        until getent hosts "${full_hostname}" | awk '{ print $1; exit }' | grep .; do 
+        if [[ $retry_count -lt {{ .Values.useHostnamesThreshold }} ]]; then
+            sleep {{ .Values.useHostnamesTimeout }}
+        else
+            error "IP address for ${full_hostname} not found"
+            exit 1
+        fi
+        ((retry_count++))
+        done
         {{- end }}
     }
 
@@ -448,7 +466,16 @@ data:
         {{- if .Values.useHostnames }}
         echo "${full_hostname}"
         {{- else }}
-        until getent hosts "${full_hostname}"; do sleep 5; done | awk '{ print $1 ; exit }';
+        retry_count=0
+        until getent hosts "${full_hostname}" | awk '{ print $1; exit }' | grep .; do 
+        if [[ $retry_count -lt {{ .Values.useHostnamesThreshold }} ]]; then
+            sleep {{ .Values.useHostnamesTimeout }}
+        else
+            error "IP address for ${full_hostname} not found"
+            exit 1
+        fi
+        ((retry_count++))
+        done
         {{- end }}
     }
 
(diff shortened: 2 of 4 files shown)


bitnami/redis/templates/scripts-configmap.yaml
```diff
@@ -54,7 +54,7 @@ data:
         {{- if .Values.useHostnames }}
         echo "${full_hostname}"
         {{- else }}
-        getent hosts "${full_hostname}" | awk '{ print $1 ; exit }'
+        until getent hosts "${full_hostname}"; do sleep 5; done | awk '{ print $1 ; exit }';
```
fmulero, line 57: It could become an infinite loop if the name cannot be registered. I think a retry limit should be included and fail if the limit is reached. I mean something similar to this piece of code (not tested):
```suggestion
        let retry=0
        until [[ retry -lt 5 ]] && getent hosts "${full_hostname}"; do ((retry+=1)); sleep 5; done | grep . | awk '{ print $1 ; exit }';
```

NOTE: `grep` should fail if getent doesn't print anything.

Please keep in mind this delay in the readiness and liveness checks.

bitnami/redis/templates/scripts-configmap.yaml
```diff
@@ -278,7 +278,7 @@ data:
...
```
fmulero, line 281: Same as above

bitnami/redis/templates/scripts-configmap.yaml
```diff
@@ -448,7 +448,7 @@ data:
...
```
fmulero, line 451: Same as above

bitnami/redis/templates/scripts-configmap.yaml
```diff
@@ -521,7 +521,7 @@ data:
...
```
fmulero, line 524: same as above

bitnami/redis/templates/scripts-configmap.yaml
```diff
@@ -650,7 +650,7 @@ data:
...
```
fmulero, line 653: Same as above

fmulero (changes requested):
Good catch @AlehYarmalovich! Thanks a lot for your contribution.

I left some comments, Could you review them?

bitnami/redis/Chart.yaml
```diff
@@ -22,4 +22,4 @@ maintainers:
 name: redis
 sources:
   - https://github.com/bitnami/containers/tree/main/bitnami/redis
-version: 17.10.1
+version: 17.11.1
```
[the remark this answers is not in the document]
AlehYarmalovich: Agree, my mistake

bitnami/redis/templates/scripts-configmap.yaml
AlehYarmalovich: Yes, I thought about it, and agree that the retry limit is still needed, but let's make it a parameter because the amount of retries depends on different DNS/K8s configuration options.
For example, without adding Route53 zone to CoreDNS in a test environment, getent gives me an external IP in 5 minutes, and with CoreDNS set up in less than 1 minute

AlehYarmalovich (commented):
Added and parameterized retry limit

bitnami/redis/values.yaml
```diff
@@ -50,6 +50,12 @@ extraDeploy: []
 ## @param useHostnames Use hostnames internally when announcing replication
 ###
 useHostnames: true
+## @param useHostnamesThreshold Failure threshold for internal hostnames collection
+###
+useHostnamesThreshold: 5
+## @param useHostnamesTimeout Timeout seconds between probes for internal hostnames collection
+###
+useHostnamesTimeout: 5
```
fmulero, line 58: As an user I wouldn't understand very well the use of these variables. If `useHostnames` is false then we will retrieve the IP addresses for the mongodb instances and only in that case `useHostnamesThreshold` and `useHostnamesTimeout` makes sense but `useHostnames` has to be false... I have to say that I am very bad naming variables and functions but I think something like this will be clearer for the final user:
```suggestion
## @param useHostnames Use hostnames internally when announcing replication. If false, the hostname will be resolved to an IP address.
###
useHostnames: true
## @param nameResolutionThreshold Failure threshold for internal hostnames resolution.
###
nameResolutionThreshold: 5
## @param nameResolutionTimeout Timeout seconds between probes for internal hostnames collection.
###
nameResolutionTimeout: 5
```
What do you think?

fmulero (commented):
Great @AlehYarmalovich! I left a little concern about the name of the variables. Could you take it a glance?

bitnami/redis/values.yaml
AlehYarmalovich: Yes, I'm bad at naming variables too :)
Your suggestion sounds much better

AlehYarmalovich:
Hey @fmulero 
Could you review?
Hope everything is as expected :)

fmulero (approved):
Great Aleh!! LGTM

PR #16245, opened 2023-04-26
[bitnami/clickhouse] Fix volumePermissions's command syntax and security context

spijet:
### Description of the change

The command used in `volumePermissions` init container was specified as string instead of a list of strings. This may have resulted in errors reported by Helm:
```
Error: UPGRADE FAILED: error validating "": error validating data: ValidationError(StatefulSet.spec.template.spec.initContainers[0].command): invalid type for io.k8s.api.core.v1.Container.command: got "string", expected "array"
```

Also, this chart appears to have an unused `.Values.volumePermissions.containerSecurityContext` knob and unconditionally uses the main container's securityContext for `volumePermissions` init container instead. This breaks the startup in some cases, e.g. when we require root privileges in order to set up file permissions properly. This PR makes `volumePermissions` honor the securityContext specified for it in `values.yaml`.

### Benefits

With this PR merged, both initContainers will use the same array-based command syntax.

### Possible drawbacks

(I'm not sure, please let me know if you know of any possible problems this may introduce 😅)

### Checklist

### DIFF
--- a/bitnami/clickhouse/templates/statefulset.yaml
+++ b/bitnami/clickhouse/templates/statefulset.yaml
@@ -123,7 +123,7 @@ spec:
               {{- end }}
               chmod 600 {{ include "clickhouse.tlsCertKey" $ }}
               {{- end }}
-          securityContext: {{ $.Values.volumePermissions.containerSecurityContext | toYaml | nindent 12 }}
+          securityContext: {{- include "common.tplvalues.render" (dict "value" $.Values.volumePermissions.containerSecurityContext "context" $) | nindent 12 }}
           {{- if $.Values.volumePermissions.resources }}
           resources: {{- toYaml $.Values.volumePermissions.resources | nindent 12 }}
           {{- end }}
--- a/bitnami/clickhouse/templates/statefulset.yaml
+++ b/bitnami/clickhouse/templates/statefulset.yaml
@@ -130,8 +130,6 @@ spec:
           volumeMounts:
             - name: data
               mountPath: /bitnami/clickhouse
-            - name: config
-              mountPath: /bitnami/clickhouse/conf/default
             {{- if $.Values.tls.enabled }}
             - name: raw-certificates
               mountPath: /tmp/certs
--- a/bitnami/clickhouse/templates/statefulset.yaml
+++ b/bitnami/clickhouse/templates/statefulset.yaml
@@ -82,8 +82,8 @@ spec:
           {{- if $.Values.resources }}
           resources: {{- toYaml $.Values.resources | nindent 12 }}
           {{- end }}
-          # We don't require a privileged container in this case
           {{- if $.Values.containerSecurityContext.enabled }}
+          # We don't require a privileged container in this case
           securityContext: {{- omit $.Values.containerSecurityContext "enabled" | toYaml | nindent 12 }}
           {{- end }}
           command:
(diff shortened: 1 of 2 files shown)


spijet:
I also found a problem where `volumePermissions` init container didn't use its' own security context and re-used the main container's one instead. [be8734f](https://github.com/bitnami/charts/pull/16245/commits/be8734f69babdaf2ac061e9d4c0e27771178a052) fixes it.

spijet:
Also, do we really need to mount the `config` ConfigMap to `/bitnami/clickhouse/conf/default` inside the `volumePermissions` container? It doesn't look like it's used there, and the `chown` command cannot do anything to the `conf` directory anyway, as the ConfigMap is mounted read-only.

bitnami/clickhouse/templates/statefulset.yaml
```diff
@@ -101,28 +101,30 @@ spec:
...
+              cp /tmp/certs/* /opt/bitnami/clickhouse/certs/
+              {{- if eq ( toString ( $.Values.volumePermissions.containerSecurityContext.runAsUser )) "auto" }}
+              chown -R `id -u`:`id -G | cut -d " " -f2` /opt/bitnami/clickhouse/certs/
+              {{- else }}
+              chown -R {{ $.Values.containerSecurityContext.runAsUser }}:{{ $.Values.podSecurityContext.fsGroup }} /opt/bitnami/clickhouse/certs/
+              {{- end }}
+              chmod 600 {{ include "clickhouse.tlsCertKey" $ }}
+              {{- end }}
+          securityContext:
+            runAsUser: {{ $.Values.volumePermissions.containerSecurityContext.runAsUser }}
```
migruiz4, line 127: Could you please render the entire `volumePermissions.containerSecurityContext` object? This way users can add extra settings if needed:
```suggestion
          securityContext: {{- include "common.tplvalues.render" (dict "value" $.Values.volumePermissions.containerSecurityContext "context" $) | nindent 12 }}
```

migruiz4 (changes requested):
Hi @spijet ,

Thank you for your contribution! Could you please take a look at my suggestion?

bitnami/clickhouse/templates/statefulset.yaml
spijet: @migruiz4, thank you for your suggestion!

Implemented in 527a0f84. :)

spijet:
Hello @migruiz4!

All done! Thank you for your suggestion! :)

migruiz4 (changes requested):
Hi @spijet,

While testing this feature, the changes look great, but I found this other issue.
The clickhouse deployment will fail with volume-permissions init-containers running into crashloopback with the following error:
```
$ k logs test-clickhouse-shard1-0 -c volume-permissions
chown: changing ownership of '/bitnami/clickhouse/conf/default/00_default_overrides.xml': Read-only file system
chown: changing ownership of '/bitnami/clickhouse/conf/default/..2023_04_28_10_31_51.2718970610/00_default_overrides.xml': Read-only file system
chown: changing ownership of '/bitnami/clickhouse/conf/default/..2023_04_28_10_31_51.2718970610': Read-only file system
chown: changing ownership of '/bitnami/clickhouse/conf/default/..data': Read-only file system
chown: changing ownership of '/bitnami/clickhouse/conf/default': Read-only file system
```

This is not related to your changes, but another issue with the original logic.

Could you please confirm if you have the same issue? If so, please apply the following change in your PR by removing these lines from the volume-permission container's volumeMounts:

```
- name: config
  mountPath: /bitnami/clickhouse/conf/default
```

spijet:
I was unsure if the `config` volume is needed in some cases, but from the looks of it there's no sense in trying to `chown`/`chmod` it, since it's a ConfigMap (i.e. always mounted as read-only). Will remove it now.

migruiz4 (approved):
Thank you very much for applying the suggested changes! Looks good to me!

spijet:
Thank you very much!

PR #16364, opened 2023-05-03
[bitnami/mastodon] Tootctl media management

joelparker:
### Description of the change

This PR creates a new CronJob and ConfigMap to manage the cache of media files created by running a Mastodon instance.
Specifically mastodon will cache attachments to posts, avatars, profile header images, and custom emoji from other mastodon servers indefinitely.  

This PR creates a script stored in a configmap that uses tootctl media remove commands documented at:
https://docs.joinmastodon.org/admin/tootctl/#media

The script is mounted in a volume to the mastodon image used for the web component of this chart as a cronjob.

The configmap script contents and cronjob are configured and controlled by the new tootctlMediaManagement section in values.yaml.  This cronjob is disabled by default.

### Benefits

The main benefit is managing costs.  The cache files are stored on S3 storage.  It might be the minio component of this chart or an external S3 bucket but either will incur monthly hosting costs.  In the case of the default operation of this chart those cache files build up indefinitely meaning each month costs go up for the administer of the instance.  This allows the administer to pick which components to remove after a configurable amount of time measured in days for each type of media cache.

* Attachements - cached images and videos posted from remote mastodon instances that local users follow
* Preview cards generated to display the images and videos as well as URL previews
* Avatars and Profile headers of remote users.  (By default avatars and headers are removed after a number of days if no local account is following the user.  This can be configured to also include remote followers that local users follow.)
* Orphaned media (a user may have deleted posts, unfollowed a user, etc.)

### Possible drawbacks

* Currently there is a good amount of code from the web/deployment.yaml duplicated in the tootctlMediaManagement/configmap.yaml and tootctlMediaManagement/cronjob.yaml for setting up environment variables as well as setting up the container to run the tootctl command.  This should be moved to a new configmap that web/deployment.yaml and tootctlMediaManagement/*.yaml share.

### Applicable issues

- fixes #

### Checklist


bitnami/mastodon/templates/tootctlMediaManagement/cronjob.yaml
```diff
@@ -0,0 +1,120 @@
...
+  name: {{ template "mastodon.tootctlMediaManagement.fullname" . }}
+  namespace: {{ include "common.names.namespace" . | quote }}
+  labels: {{- include "common.labels.standard" . | nindent 4 }}
+    app.kubernetes.io/part-of: mastodon
+    app.kubernetes.io/component: tootctlMediaManagement
+    {{- if .Values.commonLabels }}
+    {{- include "common.tplvalues.render" ( dict "value" .Values.commonLabels "context" $ ) | nindent 4 }}
+    {{- end }}
+spec:
+  concurrencyPolicy: Allow
```
rafariossaa, line 15: Hi,
  I am not into the inners of mastodon, so I am not sure if this job could be run in parallel. 
  Could you confirm it ? My concerns are mostly regarding the deleting of the same object by different job instances and the error handling of that.

joelparker: Multiple instances of the same job can be run in parallel.  Ideally this CronJob is scheduled daily.  In my testing I’ve seen large cleanups (Gigabytes removed on the first run using External S3) take on the order of minutes.  After the initial run the job usually takes seconds to run when all cacheable objects are configured to remove.  .spec.concurrencyPolicy.Allow is the default value if not specified for a CronJob but it could be made configurable.

joelparker:
tootctlMediaManagement.concurrencyPolicy added to allow a user to configure the concurrency policy of the CronJob created.

rafariossaa (approved):
LGTM

PR #16366, opened 2023-05-03
Change the docs of ConfigCli.configuration to array

MohammedNoureldin:
Usage example:
```yaml
      keycloakConfigCli:
        enabled: true
        configuration:
          - '{"enabled": true, "realm": "DOMAIN"}'
```
Which is obviously an array


MohammedNoureldin (commented):
I am not sure if the variables are added manually or generated somehow. Please let me know if I need to edit something else to get the list of variables auto-generated.

corico44:
Hello @MohammedNoureldin,

The following example is found in the values.yaml file:
  ## @param keycloakConfigCli.configuration keycloak-config-cli realms configuration
```
  ## NOTE: nil keys will be considered files to import locally
  ## Example:
  ## configuration:
  ##   realm1.json: |
  ##     {
  ##       "realm": "realm1",
  ##       "clients": []
  ##     }
  ##   files/realm2.yaml:
  ##   realm3.yaml: |
  ##     realm: realm3
  ##     clients: []
  ##
  configuration: {}

```
Why do you think that this format is wrong?

MohammedNoureldin:
Just to let you know, latest versions does not support the array anyway. The example shown by cori is the only format that works.

PR #16398, opened 2023-05-05
[bitnami/mongodb] Handle mongodb-exporter arguments with chart values

### LINKED ISSUE #15982: [bitnami/mongodb] proposal to expose less metrics by default / drastically reduce CPU usage
### Name and Version

bitnami/mongodb-13.9.4

amd64

1. have an idle mongoDB replicaset (2 mongodb, 1 arbiter, but single standalone sees the same behavioud) with a medium amount of data in it (few GBs in several collections). Enable metrics.
2. observe the primary node consume about 1700millis of CPU time in idle
3. Override args and replace `collect-all` with `--collector.diagnosticdata --collector.replicasetstatus`
4. observe mongodb CPU usage drop to about 300-400 millis of CPU time.

```yaml
architecture: replicaSet
metrics:
  enabled: true
  serviceMonitor:
    enabled: true
```

mongodb_exporter collects a basic amount of metrics per default and can be "extended" to collect more.

### What do you see instead?

IMHO the current default exposes an excessive amount of metrics (in my case almost 1.5MB per scrape in a modestly sized DB).

### Additional information

I'd like to know if the bitnami community is open to toning down the amount of metrics collected by default. AFAICS `diagnosticdata` + `replicasetstatus` is roughly the same amount of data as what the older mongo-exporter collected (before they switched to the new system that exposes _all_ of the backend objects). And I didn't see any difference in the grafana dashboards we're using day to day (though I'm aware that's only _my_ point of view).

The problem currently is that you're not able to tone down the default `--collect-all` without completely replacing the full command (via `metrics.args`). I'd prefer it the other way - collect only a basic amount of data and if someone likes to have "all", they can add `--collect-all` to extraArgs easily. I'm aware that this could be considered a breaking change though.

Alternatilely I can imagine having a switch for "--collect-all" that makes it easy to flick to "basic amount".

If you think that's worth considering, I could start a PR.

horihel:
### Description of the change

This PR refers to https://github.com/bitnami/charts/issues/15982
It replaces the hardcoded "--collect-all" with configurable values. All collector options available from mongodb-exporter are now represented by values.

This change also affects the default behaviour of metrics. It no longer defaults to "--collect-all" - instead, by default only diagnosticData and replicasetStatus are enabled (and compatible mode). In my tests, this should be enough to allow the published grafana dashboards (on grafana.com) to work the same as they do with --collect all.

Reason for the change is outlined in the above issue: "--collect-all" is very CPU intensive and collects megabytes of metric data (bloating prometheus in the process) that go mostly unused.

### Benefits

- fine grained and easily accessible control over what metrics are collected
- defaults use less CPU (80% less in my specific use case)
- adding mongodb.direct-connect and global-conn-pool should decrease logspam in mongod as the exporter now doesn't reconnect for every scrape

### Possible drawbacks

- possibly higher maintenance, depending on how the arguments of mongodb-exporter evolve
- this could be considered a breaking change by anyone depending on "--collect-all" being the default (but it actually became a default only a short while ago as a workaround for the exporter's new behavious (not exporting anything be default))

### Applicable issues

- fixes #15982

### Checklist

### DIFF
--- a/bitnami/mongodb/Chart.yaml
+++ b/bitnami/mongodb/Chart.yaml
@@ -25,4 +25,4 @@ maintainers:
 name: mongodb
 sources:
   - https://github.com/bitnami/charts/tree/main/bitnami/mongodb
-version: 13.11.0
+version: 13.12.0
--- a/bitnami/mongodb/README.md
+++ b/bitnami/mongodb/README.md
@@ -498,73 +498,73 @@ Refer to the [chart documentation for more information on each of these architec
 
 ### Metrics parameters
 
-| Name                                         | Description                                                                                                           | Value                      |
-| -------------------------------------------- | --------------------------------------------------------------------------------------------------------------------- | -------------------------- |
-| `metrics.enabled`                            | Enable using a sidecar Prometheus exporter                                                                            | `false`                    |
-| `metrics.image.registry`                     | MongoDB(&reg;) Prometheus exporter image registry                                                                     | `docker.io`                |
-| `metrics.image.repository`                   | MongoDB(&reg;) Prometheus exporter image repository                                                                   | `bitnami/mongodb-exporter` |
-| `metrics.image.tag`                          | MongoDB(&reg;) Prometheus exporter image tag (immutable tags are recommended)                                         | `0.37.0-debian-11-r26`     |
-| `metrics.image.digest`                       | MongoDB(&reg;) image digest in the way sha256:aa.... Please note this parameter, if set, will override the tag        | `""`                       |
-| `metrics.image.pullPolicy`                   | MongoDB(&reg;) Prometheus exporter image pull policy                                                                  | `IfNotPresent`             |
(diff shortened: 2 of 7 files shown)


aoterolorenzo:
@horihel thanks for the contribution!

Is the PR ready to test and review?

horihel:
Hi @aoterolorenzo .
Yes it is - i'm trying to keep it updated/conflict-free only.

aoterolorenzo:
@horihel let me then give a try to the implementation before approving.

aoterolorenzo (approved):
Seems to be working fine!

Generated args for replicaset's mode statefulset:
```
args:
            - |
              /bin/mongodb_exporter  --collector.diagnosticdata --collector.replicasetstatus --compatible-mode --mongodb.direct-connect --mongodb.global-conn-pool --web.listen-address ":9216" --mongodb.uri "mongodb://$MONGODB_ROOT_USER:$(echo $MONGODB_ROOT_PASSWORD | sed -r "s/@/%40/g;s/:/%3A/g")@localhost:27017/admin?"
              ```

PR #16407, opened 2023-05-05
[bitnami/minio] Use `mc quota` instead of old deprecated (removed) sy…

### LINKED ISSUE #16111: [minio] Provisionning hook failed if quota is defined
### Name and Version

bitnami/minio 12.2.6

amd64

Deploy minio with provided values:

```shell
helm upgrade --install -n temp -f local-values.yaml --version 12.2.6 bitnami/minio minio
```

```yaml
auth:
  rootPassword: mysupersecret
  rootUser: bar
mode: distributed
pdb:
  create: true
provisioning:
  buckets:
  - lifecycle:
    - disabled: false
      expiry:
        days: 14
        nonconcurrentDays: 3
      id: retention
    name: baz
    quota:
      size: 24GiB
      type: hard
  enabled: true
```

_No response_

### What do you see instead?

The deployment fail with this error:

```shell
$ kubectl -n temp logs -f jobs/minio-provisioning
Defaulted container "minio" out of: minio, wait-for-available-minio (init)
Start Minio provisioning
Added `provisioning` successfully.
Restart command successfully sent to `provisioning`. Type Ctrl-C to quit or wait to follow the status of the restart process.

Restarted `provisioning` successfully in 1 seconds
Bucket created successfully `provisioning/baz`.
Lifecycle configuration imported successfully to `provisioning/baz`.
mc: <ERROR> Deprecated command. Please use 'mc quota' instead.
```

### Additional information

The cause is that `mc` changed, [new way to set quota](https://min.io/docs/minio/linux/reference/minio-mc/mc-quota-set.html#command-mc.quota.set)

I tested a patch that works:

```patch
diff --git a/bitnami/minio/templates/provisioning-job.yaml b/bitnami/minio/templates/provisioning-job.yaml
index 60d7baee2..8d9d3c316 100755
--- a/bitnami/minio/templates/provisioning-job.yaml
+++ b/bitnami/minio/templates/provisioning-job.yaml
@@ -177,7 +177,11 @@ spec:
               {{- end }}

               {{- with $bucket.quota }}
-              mc admin bucket quota {{ $minioAlias }}/{{ $bucket.name }} --{{ .type }} {{ .size }};
+              {{- if eq .type "hard" }}
+              mc quota set {{ $minioAlias }}/{{ $bucket.name }} {{ if .size }}--size {{ .size }}{{ end }};
+              {{- else }}
+              mc quota {{ .type }} {{ $minioAlias }}/{{ $bucket.name }} {{ if .size }}--size {{ .size }}{{ end }};
+              {{- end }}
               {{- end }}

               {{- if $isDistributedMode }}
```

Which allow to still use types `hard`/`clear` but it may be useful to switch to `set`/`clear` or whatever. Will probably open a PR later.

sereinity:
### Description of the change

Correctly (un)define quota limits with `mc` tools, the current code doesn't work with the current default version of minio loaded in the chart.

### Benefits

We can again set the quota, upgrades doesn't fail.

### Possible drawbacks

This code works with recent version of minio, but will not with older versions.

### Applicable issues

- fixes #16111 

### Additional information

 I defined a new keyword to set quota: `set` that does the same has `hard` but have a better matching with the `mc` arguments which will help users understand what it does (and need less cases in the template).

I added a special check on `hard` to keep the compatibility with older configuration (added on not about it being deprecated).

### Checklist


bitnami/minio/Chart.yaml
```diff
@@ -25,4 +25,4 @@ name: minio
 sources:
   - https://github.com/bitnami/containers/tree/main/bitnami/minio
   - https://min.io
-version: 12.4.3
+version: 12.4.4
```
Mauraza, line 28: As you said is a breaking change so I think we should update a mayor version
```suggestion
version: 13.0.0
```

Mauraza (changes requested):
Hi @sereinity,

Thanks for the contribution!! Could you check my comments?

Mauraza:
Hi @sereinity,

Could you check the README.md? It seems there is an error with this part. 
Maybe [this](https://github.com/bitnami/charts/blob/main/CONTRIBUTING.md#documentation-requirements) help.

sereinity:
> Hi @sereinity,
> 
> Could you check the README.md? It seems there is an error with this part. Maybe [this](https://github.com/bitnami/charts/blob/main/CONTRIBUTING.md#documentation-requirements) help.

done

bitnami/minio/Chart.yaml
```diff
@@ -24,4 +24,4 @@ maintainers:
 name: minio
 sources:
   - https://github.com/bitnami/charts/tree/main/bitnami/minio
-version: 12.5.0
+version: 12.5.1
```
Mauraza, line 27:
```suggestion
version: 12.6.0
```

Mauraza (commented):
On second thought, if it is not a breaking change, I think it makes more sense to have a minor because the same command is no longer executed.

With that change, we are ready to merge!

bitnami/minio/Chart.yaml
sereinity: done

Mauraza (approved):
Hi @sereinity,

Thanks for the contribution!!! LGTM!

Issue #16559, opened 2023-05-10
Attach existing volume as persistence storage


yywing:
### Name and Version

all chart

### What is the problem this feature will solve?

Can I disable pvc and use exist volume in all bitnami chart?

postgrest-ha as example:

value.yaml
```yaml
    extraVolumes:
      - name: data
        nfs:
          server: "1"
  persistence:
    enabled: false
```

gen k8s config:
```yaml
        - name: data
          nfs:
            server: "1"
         ...
        - name: data
          emptyDir: {}
```

### What is the feature you are proposing to solve the problem?

some thing like:
```
        - name: data
          nfs:
            server: "1"
```

yywing:
but almost chart write
```
        {{- if not $.Values.persistence.enabled }}
        - name: data
          emptyDir: {}
        {{- end }}
```

how to replace data volume?

PR #16667, opened 2023-05-16
[bitnami/cassandra] Add Existing Claim for PVC

rdeveen:
### Description of the change

The Cassandra chart creates a new Persistence Volume and Persistence Volume Claim if you enable persistence but doesn't allow to use an existing PVC. This change allows to specify an existing PVC to use and not create a new one.

### Benefits

Use pre-created/existing PVC.

### Additional information

See: https://github.com/bitnami/charts/issues/16559

### Checklist


bitnami/cassandra/Chart.yaml
```diff
@@ -22,4 +22,4 @@ maintainers:
 name: cassandra
 sources:
   - https://github.com/bitnami/charts/tree/main/bitnami/cassandra
-version: 10.3.0
+version: 10.3.1
```
corico44, line 25: As we are adding a new parameter to the chart, I think it would be more appropriate if it were a minor instead of a patch. So the version should be 10.4.0 instead of 10.3.1.

rdeveen: Updated the version number.

corico44 (approved):
LGTM

Issue #16341, opened 2023-05-02
[bitnami/mongodb] Mongodb Replicaset existingSecrets doesn't work with cert-manager


rrileyca:
### Name and Version

bitnami/mongodb 13.9.4

### What architecture are you using?

amd64

### What steps will reproduce the bug?

When setting values.yaml with the following values:
```yaml
tls:
  enabled: true
  autoGenerated: false
  replicaset:
    existingSecrets:
      - mongodb-tls-0
      - mongodb-tls-1
      - mongodb-tls-2
```

Each secret must have a field `ca.crt`, as documented in the `values.yaml` comment. However, the Jetstack `cert-manager` (arguably a very common if not "standard" kubernetes application) does not produce such a field. Without the ability to inject the field manually in the helm chart, there is no integration capability out of the box.

### Are you using any custom parameters or values?

values:
```yaml
tls:
  enabled: true
  autoGenerated: false
  replicaset:
    existingSecrets:
      - mongodb-tls-0
      - mongodb-tls-1
      - mongodb-tls-2
```

### What is the expected behavior?

Either the helm chart can deploy without a `ca.crt`, or it would provide a way to inject one.

### What do you see instead?

The Pods fail to `Init`.

Running the command:
```shell
kubectl logs mongodb-0 -n mongodb -c generate-tls-certs 
```

I see the output:
```
cp: cannot stat '/certs-0/ca.crt': No such file or directory
chmod: cannot access '/certs/mongodb-ca-cert': No such file or directory
```

jotamartos:
Hi @rrileyca,

Thank you for taking the time to report this issue. Would you like to contribute again by creating a PR to solve the issue? As you are familiar with the solution and the repo, you will be able to make the necessary changes to make this work. The Bitnami team will be happy to review it and provide feedback. As always, you can find [here](https://github.com/bitnami/charts/blob/main/CONTRIBUTING.md) the contributing guidelines.

Thanks

rrileyca:
Hi @jotamartos,

I'm actually kind of confused as to how this chart would support any PKI certificates in ReplicaSet architecture. There are several problems:
[...]

PR #16731, opened 2023-05-18
[bitnami/mongodb] Adding tls.pemChainIncluded value to support wider range of TLS certificates

dtrts:
### Description of the change

Adding the option tls.pemChainIncluded to values. 

When enabled this treats the `tls.crt` attribute as a certificate chain, with the assumption that the first certificate is the leaf node, and the remaining certificates are the intermediaries. 

The generate-tls-certs init container will then split this file, using the leaf node and tls.key for the /certs/mongodb.pem file and the remaining certificates for the /certs/mongodb-ca-cert file. 

### Benefits

This will enable users to use a wider range of certificates. Specifically ACME validated certificates from cert-manager which do not return the `ca.crt` attribute in the secret. 

### Possible drawbacks

Depending on how mutual TLS is configured, and how the cert manager is configured, the pem chain may not be suitable / or easily constructed in such an order. 

It will also perform this file operation on all certificates when supplying multiple for replicasets or hidden nodes. 
I feel that is a user is using certificates from different sources for each node then they have bigger problems to worry about! 

### Applicable issues

- https://github.com/bitnami/charts/issues/16341

### Additional information

This is mimicking a change to the Kafa chart https://github.com/bitnami/charts/pull/9422

### Checklist


bitnami/mongodb/values.yaml
```diff
@@ -1845,7 +1849,7 @@ hidden:
     ## @param hidden.persistence.accessModes PV Access Mode
     ##
     accessModes:
-      - ReadWriteOnce
+    - ReadWriteOnce
     ## @param hidden.persistence.size PVC Storage Request for hidden node data volume
     ##
     size: 8Gi
```
dgomezleon, line 1855: Please keep the indentation of the document.

dtrts:
@dgomezleon - The amendments have been made. Thank you for the review.

dgomezleon (approved):
Thank you @dtrts.

Issue #16149, opened 2023-04-20
can't Access supabase studio/Kong when setting Kong to ClusterIP


markbonfert:
### Name and Version

bitnami/supabase 0.1.5

### What architecture are you using?

amd64

### What steps will reproduce the bug?

After installing the chart pods and services came up fine. But I can't access the application trough port forwarding...

### Are you using any custom parameters or values?

`helm install bitnami-supabase-hr --set kong.service.type=ClusterIP oci://registry-1.docker.io/bitnamicharts/supabase`

also tried

`helm install bitnami-supabase-hr --set kong.service.type=ClusterIP --set publicURL=127.0.0.1:8080 oci://registry-1.docker.io/bitnamicharts/supabase`

### What is the expected behavior?
```console
$ kubectl get svc
NAME                                TYPE        CLUSTER-IP     EXTERNAL-IP   PORT(S)          AGE
bitnami-supabase-hr-auth            ClusterIP   10.3.1.161     <none>        80/TCP           16m
bitnami-supabase-hr-kong            ClusterIP   10.3.139.176   <none>        80/TCP,443/TCP   16m
bitnami-supabase-hr-meta            ClusterIP   10.3.53.11     <none>        80/TCP           16m
bitnami-supabase-hr-postgresql      ClusterIP   10.3.9.129     <none>        5432/TCP         16m
bitnami-supabase-hr-postgresql-hl   ClusterIP   None           <none>        5432/TCP         16m
bitnami-supabase-hr-realtime        ClusterIP   10.3.10.13     <none>        80/TCP           16m
bitnami-supabase-hr-rest            ClusterIP   10.3.244.54    <none>        80/TCP           16m
bitnami-supabase-hr-storage         ClusterIP   10.3.124.206   <none>        80/TCP           16m
bitnami-supabase-hr-studio          ClusterIP   10.3.188.80    <none>        80/TCP           16m

$ kubectl port-forward svc/bitnami-supabase-hr-kong 8080:80
Forwarding from 127.0.0.1:8080 -> 8000
Forwarding from [::1]:8080 -> 8000
Handling connection for 8080
Handling connection for 8080

$ kubectl port-forward svc/bitnami-supabase-hr-studio 8080:80
Forwarding from 127.0.0.1:8080 -> 3000
Forwarding from [::1]:8080 -> 3000
Handling connection for 8080
Handling connection for 8080
Handling connection for 8080`
```
### What do you see instead?

[...]

PR #16744, opened 2023-05-19
[bitnami/moodle] Bugfix: container port different from http

rafacouto:
### Description of the change

This change fixes the problem related to issue #16149 and any other when the service is ClusterIP type and the container listening port is not the standard HTTP port (e.g. moodle server running as a non-privileged user on 8080).

### Benefits

With this fix, the service target port gets configured with the value specified on `containerPorts.http` and the container service can be reached by the ingress.

### Checklist

### DIFF
--- a/bitnami/moodle/Chart.yaml
+++ b/bitnami/moodle/Chart.yaml
@@ -26,4 +26,4 @@ maintainers:
 name: moodle
 sources:
   - https://github.com/bitnami/charts/tree/main/bitnami/moodle
-version: 16.1.1
+version: 16.1.2
--- a/bitnami/moodle/templates/svc.yaml
+++ b/bitnami/moodle/templates/svc.yaml
@@ -42,6 +42,7 @@ spec:
       nodePort: {{ .Values.service.nodePorts.http }}
       {{- else if eq .Values.service.type "ClusterIP" }}
       nodePort: null
+      targetPort: {{ .Values.containerPorts.http }}
       {{- end }}
     - name: https
       port: {{ coalesce .Values.service.ports.https .Values.service.httpsPort }}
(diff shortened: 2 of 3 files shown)


rafacouto:
> In this case there is still an issue, see https://github.com/bitnami/charts/pull/16744/checks?check_run_id=13607956236 for more info about the issue and how to fix it.

Fixed! Thank you :+1:

carrodher:
Please note it's needed to bump the chart version in the _Chart.yaml_ file

rafacouto:
Bumped 16.1.2 chart version :)

rafacouto:
Hi @aoterolorenzo ,

You are right, the fix is transversal and it should be modified on every Chart using a custom port for http/https. More over, to include standard http/https ports in the base service template should be reviewed (an application could not use web port services, the postgresql chart for instance).

However, if it implies to review every chart, I suggest to open another issue to do so. Meanwhile, IMO, we can patch the Moodle chart and other known affected charts.

Please, if it looks good to you, discuss this issue with bitnami leaders and proceed with merge or close this PR :smiley_cat:

aoterolorenzo:
Hi [rafacouto](https://github.com/rafacouto),

Would you prepare this PR to fix only Moodle? (changes are still affecting the template, not moodle)

Also, we also should need to file a new issue to properly escalate this to other charts, in case you want to keep contributing. From that point, we should create a PR for each asset.

rafacouto:
Bo día @aoterolorenzo,

- Moodle chart fixed. Please, merge this PR.
- I'll also fix the well-known #16149 with the same problem.
- Please, if it LGTY, open another issue to implement the [solution you've proposed here](https://github.com/bitnami/charts/pull/16744#issuecomment-1556877168).

Thanks for your suggestions :+1:

aoterolorenzo (approved):
LGTM @rafacouto !

aoterolorenzo:
PS: I just raised an internal task for the team to address this. Your proposal will be referenced there!

PR #16881, opened 2023-05-23
[bitnami/harbor] Add existingSecret to exporter

Harrys-git:
### Description of the change

Added existingSecret to exporter section.

### Benefits

Currently the env variable is pulled from Harbor-core-envvars but for external databases this does not work without adding postgres password in plain text.

### Possible drawbacks

N/A

### Applicable issues

N/A

### Additional information

N/A

### Checklist


bitnami/harbor/Chart.yaml
```diff
@@ -31,4 +31,4 @@ maintainers:
 name: harbor
 sources:
   - https://github.com/bitnami/charts/tree/main/bitnami/harbor
-version: 16.6.5
+version: 16.6.6
```
CeliaGMqrz, line 34: I think this change is a minor version. Could you please update the chart version?

CeliaGMqrz:
Hi @Harrys-git, 

Sorry for the delay. Thanks for your contribution, I have added comments, can you check it?

Harrys-git:
> Hi @Harrys-git,
> 
> Sorry for the delay. Thanks for your contribution, I have added comments, can you check it?

Thanks for the message @CeliaGMqrz I have changed it to a minor change now. Let me know if there's anything else I can do.

bitnami/harbor/templates/exporter/exporter-dpl.yaml
```diff
@@ -99,14 +99,27 @@ spec:
               value: {{ ternary "true" "false" .Values.exporter.image.debug | quote }}
             - name: HARBOR_DATABASE_PASSWORD
               valueFrom:
+              valueFrom:
+                {{- if .Values.exporter.existingSecret }}
+                secretKeyRef:
+                  name: {{ .Values.exporter.existingSecret }}
```
fmulero, line 105: I don't see clearly the use case but I think we should use here something like [this](https://github.com/bitnami/charts/blob/443d63e96275bea2525ed0760917f35b443d6154/bitnami/harbor/templates/core/core-dpl.yaml#L145-L149) because the passwords you are looking for should be in `-envvars` secret

fmulero (changes requested):
Hi @harry-cai, 

First of all, thanks a lot for your contribution. Good work!!

I left some comments, mainly because I am not sure about the scenario you are trying to solve. I think the code in versions 16.6.X are wrong about how we are managing these secrets, but I don't fully understand why you need to create a new variable instead of using `core.existingEnvVarsSecret`. Could you share more details about this?

bitnami/harbor/templates/exporter/exporter-dpl.yaml
Harrys-git: The exporter pulls envvars from Harbor-core-envvars for the admin password and postgres password. For my use case i have set the postgres password in the chart as $POSTGRES_PASSWORD to avoid plain text password which then causes the exporter to think the password is literally "$POSTGRES_PASSWORD". I set extraenvvars secret but it does not over write the harbor-core-envvar. I hope this made sense. Thats why I was hoping the chart could do either default harbor-core-envvar or an existing secret option. Otherwise we could have externaldatabase.existingSecret for password and same for admin password. Please let me know if im missing something

Harrys-git: I just seen you had a link to core-envvars. Yeah that could work aswell. I think the exporter only pulls from the harbor-core-envvars instead of having an else statement which is causing it to not be overwritten by a different secret.

fmulero: Please see my comment below. 

About _setting the postgres password in the chart as `$POSTGRES_PASSWORD`_, I am not sure about how you are doing it, but if you want to use an environment variable already set in your PC or in the machine where you are running the deployment command you can set the helm chart value with the `--set` option in helm:
```console
helm install harbor oci://registry-1.docker.io/bitnamicharts/harbor --set externalDatabase.password=$POSTGRES_PASSWORD
```

bitnami/harbor/values.yaml
```diff
@@ -3199,6 +3199,12 @@ exporter:
   ## @param exporter.extraEnvVarsCM ConfigMap containing extra env vars
   ##
   extraEnvVarsCM: ""
+  ## @param exporter.existingSecret Existing secret for exporter
+  ## The secret must contain the keys:
+  ## `POSTGRESQL_PASSWORD` (required),
+  ## `HARBOR_ADMIN_PASSWORD` (required),
+  ##
+  existingSecret: ""
```
fmulero, line 3207: If you can adopt the solution above, I think this value is not needed.

bitnami/harbor/templates/exporter/exporter-dpl.yaml
Harrys-git: I am using ArgoCD to deploy harbor as an applicationset. I use ExternalSecrets or Vault for all sensitive data including PSQL passwords.

bitnami/harbor/templates/exporter/exporter-dpl.yaml
```diff
@@ -99,14 +99,26 @@ spec:
...
               valueFrom:
+                {{- if .Values.exporter.existingSecret }}
+                secretKeyRef:
+                  name: {{ .Values.exporter.existingSecret }}
+                  key: HARBOR_ADMIN_PASSWORD
+                {{- else }}
                 secretKeyRef:
                   name: {{ include "harbor.core" . }}-envvars
                   key: HARBOR_ADMIN_PASSWORD
+                {{- end }}
```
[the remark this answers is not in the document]
Harrys-git: Added that block of code. It will do the same thing as what I proposed but its a lot better than using existing secret. Thanks

Harrys-git:
Not sure what's happening with the VIB verify but I tested the change in my environment and it works as expected. I will look at add opening a new PR in future for external.Database.existingSecret to be added.

bitnami/harbor/templates/exporter/exporter-dpl.yaml
```diff
@@ -100,13 +100,22 @@ spec:
...
               valueFrom:
                 secretKeyRef:
-                  name: {{ include "harbor.core" . }}-envvars
+                  {{- if .Values.exporter.existingEnvVarsSecret }}
+                  name: {{ .Values.exporter.existingEnvVarsSecret }}
+                  {{- else }}
+                  name: {{ printf "%s-envvars" (include "harbor.core" .) }}
+                  {{- end }}
                   key: HARBOR_ADMIN_PASSWORD
+                {{- end }}
```
fmulero, line 118: This line is causing the error. The `if` conditional is closed above, this `{{- end }}` line is unexpected.

Harrys-git: cheers

Harrys-git: Is there anything else I need to do or is this good to go ?

fmulero (approved):
LGTM

Issue #14069, opened 2022-12-22
[bitnami/minio]: helm upgrade fails and reports rootPassword is not set


kapilgarg1996:
### Name and Version

bitnami/minio: 11.9.1

### What steps will reproduce the bug?

1. Install bitnami/minio chart
2. Upgrade bitnami/minio chart

### Are you using any custom parameters or values?

values.yaml
```yaml
fullnameOverride: "minio"
mode: standalone
defaultBuckets: "logs"  
```

### What is the expected behavior?

Upgrades without any problem

### What do you see instead?
```
Error: UPGRADE FAILED: template: test/charts/minio/templates/NOTES.txt:91:4: executing "test/charts/minio/templates/NOTES.txt" at <include "common.errors.upgrade.passwords.empty" (dict "validationErrors" (list $requiredMinioPasswordErrors) "context" $)>: error calling include: template: test/charts/common/templates/_errors.tpl:21:48: executing "common.errors.upgrade.passwords.empty" at <fail>: error calling fail:
PASSWORDS ERROR: You must provide your current passwords when upgrading the release.
Note that even after reinstallation, old credentials may be needed as they may be kept in persistent volume claims.
Further information can be obtained at https://docs.bitnami.com/general/how-to/troubleshoot-helm-chart-issues/#credential-errors-while-upgrading-chart-releases
'auth.rootPassword' must not be empty, please add '--set auth.rootPassword=$ROOT_PASSWORD' to the command. To get the current value:

export ROOT_PASSWORD=$(kubectl get secret --namespace "test" minio -o jsonpath="{.data.root-password}" | base64 -d)
```
### Additional information

This is not an issue in the templates itself, 
standalone deployment is configured to use the secret value from auto-generated secret
```yaml
- name: MINIO_ROOT_PASSWORD
              valueFrom:
                secretKeyRef:
                  name: {{ include "minio.secretName" . }}
                  key: root-password
```
This is an issue in the NOTES.txt file. The following checks fail the upgrade.
```
{{- if and (not .Values.auth.existingSecret) (not .Values.auth.forceNewKeys) (not .Values.gateway.enabled) -}}
  {{- $requiredRootUser := dict "valueKey" "auth.rootUser" "secret" $secretName "field" "root-user" -}}
  {{- $requiredRootPassword := dict "valueKey" "auth.rootPassword" "secret" $secretName "field" "root-password" -}}
  {{- $requiredPasswords = append $requiredPasswords $requiredRootUser -}}
[...]

PR #16962, opened 2023-05-30
[bitnami/minio] remove unnecessary password conditions in NOTES.txt

### LINKED ISSUE #16570: [influxdb] enabling tls for LoadBalancer services
### Name and Version

bitnami/influxdb 5.5.2

The tls support in case of LoadBalancer service being selected and without the usage of ingress.

1- Add new tls enable value and secret option 
2- The secret should contain tls.crt and tls.key
3- The secret should be mounted to be used as files in order to add them for influnxdb config to support tls
4- The configmap for influxdb.conf should be updated to update https-enabled value, tls-key and tls-cert paths

_No response_

### LINKED ISSUE #16571: [minio] Passwords error on dry run with existingSecret
### Name and Version

bitnami/minio 12.4.4

amd64

0. `$ cat pw.yaml`
```yaml
apiVersion: v1
kind: Secret
metadata:
  name: minio
type: Opaque
stringData:
  root-user: admin
  root-password: test-password
```
1. `$ kubectl apply -f pw.yaml`
2. `$ helm install --wait -f values.yaml minio oci://registry-1.docker.io/bitnamicharts/minio`
3. `$ helm upgrade --dry-run --wait -f values.yaml minio oci://registry-1.docker.io/bitnamicharts/minio`

```yaml
auth:
  existingSecret: minio
```

The command should exit with code 0 and show no diff.

### What do you see instead?

```
Pulled: registry-1.docker.io/bitnamicharts/minio:12.4.4
Digest: sha256:c850f774f276f7b1213308c86a702a6709c09ed423bf14143120798b792df75c
Error: UPGRADE FAILED: execution error at (minio/templates/NOTES.txt:88:4):

    'auth.rootPassword' must not be empty, please add '--set auth.rootPassword=$ROOT_PASSWORD' to the command. To get the current value:

        export ROOT_PASSWORD=$(kubectl get secret --namespace "default" minio -o jsonpath="{.data.root-password}" | base64 -d)
```

### Additional information

I'm not sure wether this was introduced in https://github.com/bitnami/charts/pull/14356 . The changes make sense to me, but someting seems to be off.

### Impact

This makes the chart unusable with `helmfile`, which uses `helm diff`, which in turn uses `helm upgrade --dry-run`.

c4tz:
### Description of the change

**I removed the check for the root-password `NOTES.txt`.**

The problem with the current implementation is a [known](https://github.com/helm/helm-www/issues/635) and [documented](https://github.com/helm/helm-www/blob/main/content/en/docs/chart_template_guide/functions_and_pipelines.md#using-the-lookup-function) behaviour:

> Keep in mind that Helm is not supposed to contact the Kubernetes API Server during a `helm template` or a `helm install|upgrade|delete|rollback --dry-run`, so the lookup function will return an empty list (i.e. dict) in such a case.

This (currently) makes the condition in `NOTES.txt` fail in all cases when using one of the aforementioned `helm` commands. I took a step back and thought about a case were the password could be not set/empty, but I couldn't really find one.  

I considered the following situations:

- `auth.rootPassword=""`, `auth.existingSecret=""`: The root password will be generated automatically once and kept in the default secret. On upgrades, it is looked up by Helm.
- `auth.rootPassword="xyz"`, `auth.existingSecret=""`: Same as above, without the auto-generation.
- `auth.rootPassword=""`, `auth.existingSecret="xyz"`: The password (and username) for root will be looked up from the given secret, on both installing and upgrading the chart. 

As you can see, the password will be set in all constellations.

### Benefits

Chart would be usable with `helmfile`, `helm diff` and `helm upgrade --dry-run`. Also, the problem from #14069 would still be answered.

### Possible drawbacks

The password in the existing Secret (be it auto-generated or created manually) will not be looked up anymore, so there could be a empty string inside. But if you'd consider this a problem, you'd probably have a `lookup` in the other charts as well :wink:  

### Applicable issues

- fixes #16571 

### Checklist

### DIFF
--- a/bitnami/minio/Chart.yaml
+++ b/bitnami/minio/Chart.yaml
@@ -24,4 +24,4 @@ maintainers:
 name: minio
 sources:
   - https://github.com/bitnami/charts/tree/main/bitnami/minio
-version: 12.6.3
+version: 12.6.4
--- a/bitnami/minio/templates/NOTES.txt
+++ b/bitnami/minio/templates/NOTES.txt
@@ -74,15 +74,3 @@ To access the MinIO&reg; web UI:
 {{- include "common.warnings.rollingTag" .Values.clientImage }}
 {{- include "common.warnings.rollingTag" .Values.volumePermissions.image }}
 {{- include "minio.validateValues" . }}
-
-{{- $requiredPasswords := list -}}
-{{- $secretName := include "minio.secretName" . -}}
-{{- $rootPassword := include "minio.secret.existingPassword" . -}}
-{{- if not $rootPassword }}
-  {{- $requiredRootUser := dict "valueKey" "auth.rootUser" "secret" $secretName "field" "root-user" -}}
-  {{- $requiredRootPassword := dict "valueKey" "auth.rootPassword" "secret" $secretName "field" "root-password" -}}
-  {{- $requiredPasswords = append $requiredPasswords $requiredRootUser -}}
-  {{- $requiredPasswords = append $requiredPasswords $requiredRootPassword -}}
-{{- end -}}
-{{- $requiredMinioPasswordErrors := include "common.validations.values.multiple.empty" (dict "required" $requiredPasswords "context" $) -}}
-{{- include "common.errors.upgrade.passwords.empty" (dict "validationErrors" (list $requiredMinioPasswordErrors) "context" $) -}}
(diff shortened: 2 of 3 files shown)


jotamartos (approved):
Thanks for taking the time to create this PR. I can see we made similar changes in the past in other charts to solve similar issues so these changes are completely valid.

PR #16992, opened 2023-06-01
[bitnami/metallb] Add missing rule to controller role

brenix:
### Description of the change

The clusterrole in the bitnami chart does not match some of the rbac rules from the upstream sources. This change adds a missing rule for listing nodes in the cluster as per the
[upstream](https://github.com/metallb/metallb/blob/e708fc9010afb80c379bdfadd5d543931164ca5e/config/rbac/role.yaml#L17-L22) configuration

This should fix errors like this:

```
BGPAdvertisement/kube-system/default dry-run failed, reason: Forbidden: admission webhook "bgpadvertisementvalidationwebhook.metallb.io" denied the request: nodes is forbidden: User "system:serviceaccount:kube-system:metallb-controller" cannot list resource "nodes" in API group "" at the cluster scope
```

### Benefits

Allows helm chart upgrades to succeed

### Possible drawbacks

Helm chart fails to upgrade

### Applicable issues

- fixes #

### Checklist

### DIFF
--- a/bitnami/metallb/Chart.yaml
+++ b/bitnami/metallb/Chart.yaml
@@ -27,4 +27,4 @@ maintainers:
 name: metallb
 sources:
   - https://github.com/bitnami/charts/tree/main/bitnami/metallb
-version: 4.5.2
+version: 4.5.3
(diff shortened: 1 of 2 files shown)


fmulero (approved):
Good catch!! Thank you so much @brenix. LGTM

PR #17001, opened 2023-06-02
[bitnami/etcd] Fix label selector for pdb

fhopfensperger:
Hi,

We found an issue, that the `PodDisruptionBudget` is not working as desired, as it currently takes  `Jobs` created by the `Cronjob` (if `disasterRecovery.enabled=true`) into consideration:
```
$ k describe poddisruptionbudgets.policy -n sandbox-profile-0-n5c8h
Name:           sandbox-profile-0-n5c8h-etcd
Namespace:      sandbox-profile-0-n5c8h
Min available:  51%
Selector:       app.kubernetes.io/instance=sandbox-profile-0-n5c8h-etcd,app.kubernetes.io/name=etcd
Status:
    Allowed disruptions:  0
    Current:              3
    Desired:              2
    Total:                3
Events:
  Type     Reason                           Age              From               Message
  ----     ------                           ----             ----               -------
  Normal   NoPods                           18m              controllermanager  No matching pods found
  Warning  CalculateExpectedPodCountFailed  2s (x2 over 2s)  controllermanager  Failed to calculate the number of expected pods: jobs.batch does not implement the scale subresource
```

### Description of the change

Adding a label to `Statefulset` and `PodDisruptionBudget` to select the correct resources, similar what is done here: https://github.com/bitnami/charts/blob/487c893862a87992f2f75ff8192c84eb0cd17ad5/bitnami/redmine/templates/pdb.yaml

**Note:** If I would change `spec.selector.matchLabels` in the `StatefulSet` it would be a breaking change, as it's forbidden to update this field for already existing instances.

**Another option**: Define the selector of `PodDisruptionBudget` like this, by adding the `matchExpressions`
```
  selector:
    matchLabels: {{- include "common.labels.matchLabels" . | nindent 6 }}
    {{- if .Values.disasterRecovery.enabled }}
    matchExpressions:
      - key: app.kubernetes.io/component
        operator: NotIn
        values:
          - snapshotter
    {{- end }}
```

### Benefits

We can drain nodes and `PodDisruptionBudget` is working if disaster recovery is enabled.

### Possible drawbacks

- Breaking change if we decide to update `spec.selector.matchLabels` as well

### Checklist

### DIFF
--- a/bitnami/etcd/Chart.yaml
+++ b/bitnami/etcd/Chart.yaml
@@ -24,4 +24,4 @@ maintainers:
 name: etcd
 sources:
   - https://github.com/bitnami/charts/tree/main/bitnami/etcd
-version: 8.12.0
+version: 9.0.0
--- a/bitnami/etcd/README.md
+++ b/bitnami/etcd/README.md
@@ -440,6 +440,29 @@ Find more information about how to deal with common errors related to Bitnami's 
 
 ## Upgrading
 
+### To 9.0.0
+
+This version adds a new label `app.kubernetes.io/component=etcd` to the StatefulSet and pods. Due to this change, the StatefulSet will be replaced (as it's not possible to add additional `spec.selector.matchLabels` to an existing StatefulSet) and the pods will be recreated. To upgrade to this version from a previous version, you need to run the following steps:
+
+1. Add new label to your pods
+
+    ```console
+    kubectl label pod my-release-0 app.kubernetes.io/component=etcd
+    # Repeat for all etcd pods, based on configured .replicaCount (excluding the etcd snappshoter pod, if .disasterRecovery.enabled is set to true)
+    ````
+
+2. Remove the StatefulSet keeping the pods:
+
+    ```console
+    kubectl delete statefulset my-release --cascade=orphan
+    ```
+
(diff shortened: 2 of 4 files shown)


fmulero (changes requested):
Thanks a lot @fhopfensperger to fix this issue and such detailed explanation.

To be honest I'd be happier if we change the selectors also in the statatefulset but as you pointed out, that change will imply a major change. Could you change it and add also some instructions to do the upgrade? for example:
1. Add new label to your pods `kubectl label pod my-release-0 app.kubernetes.io/component=etcd`
2. Remove the StatefulSet object keeping the pods: `kubectl delete sts my-release --cascade=orphan`
3. Upgrade your cluster: `helm upgrade my-release oci://registry-1.docker.io/bitnamicharts/etcd  --set auth.rbac.rootPassword=$ETCD_ROOT_PASSWORD`

fhopfensperger:
Hi @fmulero,
I've just addressed your proposed changes and added a new "chapter" inside the Upgrading section. Is that about how you imagined it?

fmulero (approved):
BRILLIANT!! Thanks a lot @fhopfensperger. Big kudos for you

PR #17029, opened 2023-06-05
[bitnami/schema-registry] remove storageClass references

jeremysprofile:
### Description of the change

The schema-registry does not actually use ReadWriteMany volumes in any capacity. The only place where a `storageClass` can be specified is under `global`, which in turn is only used in `templates/_helpers.tpl` to define `schema-registry.image` and `schema-registry.imagePullSecrets`. The actual storage mechanism for the registry is via Kafka topic, so remove these references for clarity.

### Benefits

Documentation clarity - remove unused code.

### Possible drawbacks

If I am wrong and Volumes are used in some way by this chart, this would break that.

### Applicable issues

N/A

### Checklist


corico44:
Thank you for opening this PR @jeremysprofile,
This change that you mention caused by "The schema-registry does not actually use ReadWriteMany volumes in any capacity", has it been obtained from some kind of documentation source?

corico44:
> > Thank you for opening this PR @jeremysprofile, This change that you mention caused by "The schema-registry does not actually use ReadWriteMany volumes in any capacity", has it been obtained from some kind of documentation source?
> 
> @corico44 , does https://docs.confluent.io/platform/current/schema-registry/installation/deployment.html#memory suffice?
> 
> > Schema Registry uses Kafka as a commit log to store all registered schemas durably, and maintains a few in-memory indices to make schema lookups faster.
> > ...
> > Schema Registry does not have any disk resident data. It currently uses Kafka as a commit log to store all schemas durably and holds in-memory indices of all schemas. Therefore, the only disk usage comes from storing the log4j logs.

From the documentation, it appears that some information is stored in memory: "Schema Registry uses Kafka as a commit log to store all registered schemas durably, and maintains a few in-memory indices to make schema lookups faster."

jeremysprofile:
@corico44 , yes, it caches things **in-memory**, not on disk. If the pod restarted, that cache would be lost whether or not you were using a persistent volume.

This might be a better way to convince you:

Look at https://github.com/bitnami/charts/blob/main/bitnami/schema-registry/templates/statefulset.yaml#L234 .  The only volumes it mounts are ConfigMaps and Secrets. It allows users to mount their own PVs, but users have to specify the mount points themselves - nothing in this chart uses those by default. In fact, the description of `extraVolumes` and `extraVolumeMounts` in https://github.com/bitnami/charts/tree/main/bitnami/schema-registry is copy/pasted straight from https://github.com/bitnami/charts/tree/main/bitnami/minio. 

That is a relevant concern, though. What should the descriptions on these `extra[Volume]Mounts` values be? I cannot think of a use, but I am not especially knowledgable of `schema-registry` and would rather have an expert chime in.

corico44:
@jeremysprofile OK. Thank you very much for the clarification. Lastly, I think it would be more correct for the version to increase in a minor and not in a patch. Therefore, the version should be _11.1.0_

jeremysprofile:
@corico44 , updated

corico44 (approved):
LGTM

PR #17066, opened 2023-06-07
[bitnami/postgresql-ha] Fix password reference for created secret

tgh19:
### Description of the change

when I set both of:
```
postgresql.usePasswordFile	
pgpool.usePasswordFile	
````

I’m getting this error
```
MountVolume.SetUp failed for volume "postgresql-password" : references non-existent secret key: postgresql-password
```

This is my values.yaml

```yaml
postgresql:
  usePasswordFile: "true"
  replicaCount: 1

pgpool:
  usePasswordFile: "true"
``` 

`password` is always included in the pgsql secret. However, `postgres-password` may or may not be included.

### Benefits

Deploying with usePasswordFile for both pgpool and postgresql now works.

### Possible drawbacks

None I can think of

### Applicable issues

NA

### Additional information

The key here: https://github.com/bitnami/charts/blob/main/bitnami/postgresql-ha/templates/pgpool/deployment.yaml#L391 would still not be correct if the secret https://github.com/bitnami/charts/blob/main/bitnami/postgresql-ha/templates/postgresql/secrets.yaml#LL1C1-L22C12 included the key `postgres-password` since the volume references the key `postgresql-password`. 

### Checklist

### DIFF
--- a/bitnami/postgresql-ha/templates/pgpool/deployment.yaml
+++ b/bitnami/postgresql-ha/templates/pgpool/deployment.yaml
@@ -388,7 +388,7 @@ spec:
           secret:
             secretName: {{ include "postgresql-ha.postgresqlSecretName" . }}
             items:
-              - key: postgresql-password
+              - key: password
                 path: pgpool-password
               - key: repmgr-password
                 path: pgpool-sr-check-password
--- a/bitnami/postgresql-ha/templates/pgpool/deployment.yaml
+++ b/bitnami/postgresql-ha/templates/pgpool/deployment.yaml
@@ -388,8 +388,13 @@ spec:
           secret:
             secretName: {{ include "postgresql-ha.postgresqlSecretName" . }}
             items:
+              {{- if eq (include "postgresql-ha.postgresqlUsername" .) "postgres" }}
               - key: password
                 path: pgpool-password
+              {{- else }}
+              - key: postgresql-password
+                path: pgpool-password
+              {{- end }}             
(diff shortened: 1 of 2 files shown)


bitnami/postgresql-ha/templates/pgpool/deployment.yaml
```diff
@@ -388,7 +388,7 @@ spec:
           secret:
             secretName: {{ include "postgresql-ha.postgresqlSecretName" . }}
             items:
-              - key: postgresql-password
+              - key: password
                 path: pgpool-password
```
jotamartos, line 392: There is something to take into account here, if `postgresql-ha.postgresqlUsername` is not `postgres`, we need to use `postgresql-password` instead of `password` as you can see in the secret

https://github.com/bitnami/charts/blob/main/bitnami/postgresql-ha/templates/postgresql/secrets.yaml#L17

In order to fix this issue, I'd use the following code

```suggestion
              {{- if eq (include "postgresql-ha.postgresqlUsername" .) "postgres" }}
              - key: password
                path: pgpool-password
              {{- else }}
              - key: postgresql-password
                path: pgpool-password
              {{- end }}             
```

Can you try these changes and confirm they work as expected? 

Thanks

tgh19:
I'll try and take a look at this tomorrow!

jotamartos:
Our tests look good, please confirm when you test it on your side and I'll take care of merging the changes and releasing a new version of the solution.

PR #17124, opened 2023-06-14
[bitnami/apache] Rewrite Ingress TLS logic

### LINKED ISSUE #17112: [bitnami/apache] The ingress.certManager value is not deprecated
### Name and Version

bitnami/apache 9.6.3

amd64

I'm trying to configure the chart so that the Ingress resource works with cert-manager (configured with a Lets Encrypt `ClusterIssuer` in my case). In order to do that, I need to populate the `tls` field in the Ingress object. To do that, I need to not only populate the corresponding value in the chart Values (i.e. `ingress.tls`), I also need to set the `ingress.certManager` field to `true` in order for the `tls` field to be rendered in the Ingress object.

That in itself is not very strange. What is strange is that the `ingress.certManager` field is commented out (and thus undocumented in the README) with a comment explaining that it's deprecated in favor of `ingress.annotations`. This is clearly nonsense, since `ingress.certManager` is required to render the `tls` field in the Ingress object at all.

To be precise, there are two toggles for the `tls` object: either [`ingress.certManager` _or_ `ingress.selfSigned`](https://github.com/bitnami/charts/blob/ef3e8fb307c06342abbeb1f9c986e2e59e86f7c4/bitnami/apache/templates/ingress.yaml#L54). It is, however, not possible to set `selfSigned` instead, since that will overwrite my cert-manager's certificate secret with a self-signed secret ([here](https://github.com/bitnami/charts/blob/ef3e8fb307c06342abbeb1f9c986e2e59e86f7c4/bitnami/apache/templates/tls-secrets.yaml#L24)).

I think that this deprecation has been made by mistake. But there are other solutions to the situations too, which would involve refactoring the template code quite a lot.

_No response_

_No response_

### What do you see instead?

It says here that the `ingress.certManager` field is deprecated: https://github.com/bitnami/charts/blob/ef3e8fb307c06342abbeb1f9c986e2e59e86f7c4/bitnami/apache/values.yaml#L513

Yet the value is used here: https://github.com/bitnami/charts/blob/ef3e8fb307c06342abbeb1f9c986e2e59e86f7c4/bitnami/apache/templates/ingress.yaml#L54

And `ingress.selfSigned` cannot be used instead since it generates a new TLS secret here: https://github.com/bitnami/charts/blob/ef3e8fb307c06342abbeb1f9c986e2e59e86f7c4/bitnami/apache/templates/tls-secrets.yaml#L24

### Additional information

_No response_

lindhe:
### Description of the change

This change rewrites the logic pertaining to when and how to render the `tls` field in the Ingress object.

### Benefits

Because of earlier rewrites of the code, there were a bunch of unused, incorrectly documented and overall strange values with regards to setting the `tls` field in the Ingress object. This PR fixes that.

### Possible drawbacks

As noted in the commit messages, the commit `62a8c633f71f21631c4f5b83c66efa244cf9f623` introduces a (fairly small) breaking change. If we for some reason want to hold off with breaking changes, I can simply drop that commit and its version bump. Tell me if you want that.

### Applicable issues

- closes #17112

### Checklist


lindhe:
Still waiting for input from the maintainers.

bitnami/apache/values.yaml
```diff
@@ -503,25 +503,11 @@ ingress:
...
   ## ref: https://kubernetes.io/docs/concepts/services-networking/ingress/#tls
   ##
-  tls:
-    - hosts:
-        - example.local
-      secretName: example.local-tls
-  ## DEPRECATED: Use ingress.annotations instead of ingress.certManager
-  ## certManager: false
-  ##
+  tls: false
```
andresbono, line 509: This change is introducing a breaking change for someone that for instance installs the chart with the following values, right?

```yaml
ingress:
  enabled: true
  selfSigned: true
```

```bash
$ helm template ./bitnami/apache --set ingress.enabled=true,ingress.selfSigned=true
...
  tls:
    - hosts:
        - "example.local"
      secretName: example.local-tls
```

With your changes, the `tls` part will not be rendered unless you set `ingress.tls=true`, is that right?

Maybe a note should be added in the README: https://github.com/bitnami/charts/tree/main/bitnami/apache#upgrading

andresbono (commented):
Thank you for your contribution! Sorry for the late reply on this PR. I left a comment with some feedback.

bitnami/apache/values.yaml
lindhe: You are right! I'll look into it. 👍

lindhe: I've pushed a change now that I believe fixes this issue. The behavior when using `--set 'ingress.enabled=true,ingress.selfSigned=true'` should now be unchanged, if I didn't mess something up. Please help me double check the logic again. :)

andresbono:
While your proposed changes certainly improves the ingress object logic, taking advantage of the new major version bump, I think we can go further and do a full standardization of the ingress, based on the best practices defined in the `./template` and in the latest created charts.

I have just pushed a commit introducing some more changes, all aligned with your initial proposal.

If you are ok, we can merge the PR now.

bitnami/apache/values.yaml
andresbono: I was trying to keep backwards compatibility but then I thought it would be better to do a full standardization of the ingress, as we are already bumping the major version.

andresbono (approved):

lindhe:
Your changes looks good. Just one more thing: is it a bit weird to have `ingress.tls` as a  bool? The name doesn't "sound" like a bool...

One common pattern is to have `ingress.tls.enabled=true`, but that results in a slightly wordy YAML:

```yaml
ingress:
  tls:
    enabled: true
```

Another option might be `ingress.enableTLS=true`.

WDYT?

andresbono:
You are right, the name doesn't sound like a boolean. But for the sake of homogeneity with the rest of charts, I would leave it like it is. Also, the value is properly documented.

lindhe:
Got it. I'll leave it for you to decide. Feel free to merge! :)

> But for the sake of homogeneity with the rest of charts, I would leave it like it is. 

I don't get this though. There's plenty of `foo.enabled=true` in the charts... In the apache chart, even.

andresbono:
Merged. Thank you for your contribution!

PR #17142, opened 2023-06-15
[bitnami/kube-prometheus] fix: Fix servicemonitor annotations and ArgoCD sync

Balsir:
### Description of the change

1. Fixes mismatched values references in `ServiceMonitor` (e.g. coreDns referencing kubeProxy annotations...)
2. Makes annotation field in `ServiceMonitor` optional. Fixes issue due to which ArgoCD with server-side apply keeps `ServiceMonitor` resources in out-of-sync state. Only `ServiceMonitor`s with non-conditional annotations field were affected.

image

### Benefits

1. Correct labels and annotations for correct components
2. Synced status in ArgoCD

### Possible drawbacks

If people used the mismatched annotations, this may reannotate their resources

### Checklist

### DIFF
--- a/bitnami/kube-prometheus/Chart.yaml
+++ b/bitnami/kube-prometheus/Chart.yaml
@@ -32,4 +32,4 @@ maintainers:
 name: kube-prometheus
 sources:
   - https://github.com/bitnami/charts/tree/main/bitnami/kube-prometheus
-version: 8.13.2
+version: 8.14.0


bitnami/kube-prometheus/Chart.yaml
```diff
@@ -32,4 +32,4 @@ maintainers:
 name: kube-prometheus
 sources:
   - https://github.com/bitnami/charts/tree/main/bitnami/kube-prometheus
-version: 8.13.1
+version: 8.13.2
```
rafariossaa, line 35: As you are changing the annotations and this could cause some noise to the users, I think it would be better to move to `8.14.0`.

Balsir: Done :)

rafariossaa (approved):
Thanks !
LGTM

PR #17144, opened 2023-06-16
[bitnami/postgresql-ha] pgPool Load Balanced Write Bugfix

JohnLahr:
### Description of the change

Added missing quotes around PGPOOL_DISABLE_LOAD_BALANCE_ON_WRITE value. The missing `| quote` filter would render an invalid deployment manifest when `pgpool.loadBalancingOnWrite` was set to `off` because YAML interprets `off` as `false` when unquoted.

### Benefits

Allows users to correctly configure load balancing in pgPool for reads from the PostgreSQL servers without also load balancing writes to the pool, should they choose to do so.

### Possible drawbacks

N/A

### Applicable issues

I personally encountered this bug just recently, but I haven't filed an issue for it. Quite frankly, the fix took less time than creating the issue would, so here we are.

### Additional information

N/A

### Checklist

### DIFF
--- a/bitnami/postgresql-ha/Chart.yaml
+++ b/bitnami/postgresql-ha/Chart.yaml
@@ -28,4 +28,4 @@ maintainers:
 name: postgresql-ha
 sources:
   - https://github.com/bitnami/charts/tree/main/bitnami/postgresql-ha
-version: 11.7.5
+version: 11.7.6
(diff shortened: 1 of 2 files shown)


dgomezleon (approved):
LGTM. Thanks for contributing

Issue #16963, opened 2023-05-30
schema-registry helm chart requires .Values.kafka.auth.clientProtocol when kakfa.enabled: false


ssontaclaus:
### Name and Version

bitnamicharts/schema-registry

### What architecture are you using?

None

### What steps will reproduce the bug?

1. `helm upgrade --install schema-registr-test oci://registry-1.docker.io/bitnamicharts/schema-registry -f values.yaml -n test`

### Are you using any custom parameters or values?

```yaml
# values.yaml
kafka:
  enabled: false
externalKafka:
  brokers:
    - PLAINTEXT://my.kafka.domain:9092
    auth:
      protocol: plaintext
```

### What is the expected behavior?

I think it should require a `.Values.externalKafka.auth.protocol` value.
https://github.com/bitnami/charts/blob/main/bitnami/schema-registry/templates/statefulset.yaml#L4

### What do you see instead?
```
Error: UPGRADE FAILED: template: schema-registry/templates/statefulset.yaml:4:95: executing "schema-registry/templates/statefulset.yaml" at <.Values.kafka.auth.clientProtocol>: nil pointer evaluating interface {}.clientProtocol
```

dgomezleon:
Hello @ssontaclaus ,

Thanks for reporting this issue. It seems there are several missing values for Kafka. Would you like to contribute by creating a PR to solve the issue? The Bitnami team will be happy to review it and provide feedback. [Here](https://github.com/bitnami/charts/blob/main/CONTRIBUTING.md) you can find the contributing guidelines.

PR #17166, opened 2023-06-18
[bitnami/schema-registry] Add value to use existing secret for external Kafka SASL #17084

### LINKED ISSUE #17084: [bitnami/schema-registry] Add value to use existing secret for external Kafka SASL
### Name and Version

bitnami/schema-registry 11.0.0

The password for external Kafka SASL authentication is set as plaintext value in the Helm chart:
https://github.com/bitnami/charts/blob/main/bitnami/schema-registry/values.yaml#L673

While there is a mechanism to use an existing secret, which is implemented here:
https://github.com/bitnami/charts/blob/main/bitnami/schema-registry/templates/external-kafka-secrets.yaml#L20

However, this mechanism does not work, if a Helm chart is installed from templates, as e.g. ArgoCD does, because the `lookup` function from `common.secrets.passwords.manage` does not work there, and so the existing secret is never used.

Thus, the only way to set a correct password is to have it plaintext in `values.yaml` file in a GitOps repo, which is not a great choice from security perspective.

Please add a value `existingSecret`, analogous to this:
https://github.com/bitnami/charts/blob/main/bitnami/kafka/values.yaml#L293

and add a condition for `external-kafka-secrets.yaml` to be created only if this value is not set.

P.S.: I can contribute the PR if needed, the change itself is small (~ 3 LOC)...

_No response_

vikiscience:
### Description of the change

Add a value "externalKafka.auth.jaas.existingSecret", analogous to the one used in Kafka Helm chart.

### Benefits

While there is a mechanism to use an existing secret, which is implemented here:
https://github.com/bitnami/charts/blob/main/bitnami/schema-registry/templates/external-kafka-secrets.yaml#L20

However, this mechanism does not work, if a Helm chart is installed from templates, as e.g. ArgoCD does, because the lookup function from common.secrets.passwords.manage does not work there, and so the existing secret is never used.

Thus, the only way to set a correct password is to have it plaintext in values.yaml file in a GitOps repo, which is not a great choice from security perspective.

This PR makes it possible to avoid that.

### Applicable issues

- fixes #17084

### Checklist


javsalgar:
Hi!

Thank you so much for the PR! Could you fix the merge issues?

vikiscience:
> Hi!
> 
> Thank you so much for the PR! Could you fix the merge issues?

Done.

bitnami/schema-registry/Chart.yaml
```diff
@@ -28,4 +28,4 @@ maintainers:
 name: schema-registry
 sources:
   - https://github.com/bitnami/charts/tree/main/bitnami/schema-registry
-version: 11.0.1
+version: 11.0.2
```
FraPazGal, line 31: As you are adding a new feature, and per [semver conventions](https://semver.org/#semantic-versioning-specification-semver), we should bump the chart's version to a new minor.

bitnami/schema-registry/values.yaml
```diff
@@ -639,6 +639,15 @@ kafka:
...
   replicaCount: 1
+  ## @param kafka.auth.clientProtocol Authentication protocol for communications with clients. Allowed protocols: `plaintext`, `tls`, `mtls`, `sasl` and `sasl_tls`
+  ##
+  auth:
+    clientProtocol: "plaintext"
+  ## @param kafka.service.ports.client Kafka svc port for client connections
+  ##
+  service:
+    ports:
+      client: 9092
```
FraPazGal, line 650: Are these parameters added as a typo or is a conscious decision? I don't have any strong thoughts about this but could you please explain why you included them if done on purpose?

FraPazGal (commented):
Really nice PR @vikiscience, you are right the current logic does not cover your scenario. I have added a few minor comments, please review them when you can!

bitnami/schema-registry/values.yaml
vikiscience: Without these default values, the chart won't install, if used with `kafka.enabled: false` and external Kafka.
This was already discovered in the issue #16963 (which I added to this PR's description, for documentation purposes).
The problem is, these values are referenced e.g. in https://github.com/bitnami/charts/blob/main/bitnami/schema-registry/templates/statefulset.yaml#L5, but there is no check if kafka is enabled as a subchart, or external Kafka is used. 
So I had to fix it too, in order to test my changes. Do you have another suggestion how to fix it?

FraPazGal: You are right, I didn't notice the issue was on the upgrade phase and couldn't reproduce it the other day. In this case, I think this is the cleanest solution without having to rework any logic.

mingmingshiliyu:
> Thanks for the quick review @FraPazGal ! I've addressed your comments below.

met up with some problems, what is the difference between 
``` yaml
## @param externalKafka.sasl.password Password for SASL authentication
  ## @param externalKafka.sasl.existingSecret Name of the existing secret containing a password for SASL authentication (under the key named "client-passwords")
```

PR #17195, opened 2023-06-20
[bitnami/postgresql] Set auto-mount service account token in PodSpec

### LINKED ISSUE #17194: Allow setting automountServiceAccountToken for PodSpecs
### Name and Version

bitnami/postgresql 12.5.8

Currently, the setting `automountServiceAccountToken` is present to be set in `ServiceAccount`s and can be set through the `values.yaml` (e.g. `.Values.serviceAccount.automountServiceAccountToken`). Kubernetes resources split this automatism into two settings, one in the `ServiceAccount`, one in the `PodSpec` [see reference](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#opt-out-of-api-credential-automounting).

With the current settings, it is not possible to influence the `PodSpec` setting via `values.yaml`. This means a service account token created by a third party service may be inadvertently mounted when matching the `serviceAccountName` (defaulting to `default` when `.Values.serviceAccount.create`).

Re-use the existing settings `.Values.serviceAccount.automountServiceAccountToken` to be also set in the `PodSpec`.

The default for those values is currently `true`, which matches the default when not set in `PodSpec`. Someone who has those explicitly set to `false` will now get a pod without a service account token, which I IMHO is the meaning of this setting. Someone who previously had this setting on `false` but wanted to mount it's own token will need to change the setting to `true`, keeping `.Values.serviceAccount.create: false`, which will still produce a pod which auto-mounts the token.

Not that this is the same issue and proposed solution for #17174 

1. Use a separate setting for `automountServiceAccountToken` for the `PodSpec`. But, this would probably double up the setting in the `values.yaml` without matching the current structure of those values (everything is grouped into logical groups, not into K8S resource kinds).
2. Not doing it. This is IMHO a relatively small change and may not be necessary. On the other hand, it's something that completes the settings possible, groundwork is already there and existing settings can be re-used.

girstenbrei:
### Description of the change

Add the `automountServiceAccountToken` to postgresql `primary` and `read`.

### Benefits

Allow a user to influence the `automountServiceAccountToken` setting for `PodSpec`s.

### Possible drawbacks

A user having a custom `ServiceAccount` configured but explicitly having `automountServiceAccountToken` set to `false` might inadvertently rely on the current behavior of auto-mounting. I would argue, that the intent of `automountServiceAccountToken: false` is to not mount the token, making this a bug.

### Applicable issues

- fixes #17194 

### Additional information

This is the same issue / solution as in PR #17175 

### Checklist

### DIFF
--- a/bitnami/postgresql/Chart.yaml
+++ b/bitnami/postgresql/Chart.yaml
@@ -25,4 +25,4 @@ maintainers:
 name: postgresql
 sources:
   - https://github.com/bitnami/charts/tree/main/bitnami/postgresql
-version: 12.5.8
+version: 12.5.9
--- a/bitnami/postgresql/templates/primary/statefulset.yaml
+++ b/bitnami/postgresql/templates/primary/statefulset.yaml
@@ -58,6 +58,7 @@ spec:
       {{- include "common.tplvalues.render" (dict "value" .Values.primary.extraPodSpec "context" $) | nindent 6 }}
       {{- end }}
       serviceAccountName: {{ include "postgresql.serviceAccountName" . }}
+      automountServiceAccountToken: {{ .Values.serviceAccount.automountServiceAccountToken }}
       {{- include "postgresql.imagePullSecrets" . | nindent 6 }}
       {{- if .Values.primary.hostAliases }}
       hostAliases: {{- include "common.tplvalues.render" (dict "value" .Values.primary.hostAliases "context" $) | nindent 8 }}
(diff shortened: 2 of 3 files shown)


girstenbrei:
So some action failed with ` exceeded the maximum execution time of 360 minutes`, but there are no logs for it or any other reason why it failed that I can see. If I can support here, let me know.

jcpunk:
Can someone try to re-run the failed workflows?

bitnami/postgresql/Chart.yaml
```diff
@@ -25,4 +25,4 @@ maintainers:
 name: postgresql
 sources:
   - https://github.com/bitnami/charts/tree/main/bitnami/postgresql
-version: 12.5.8
+version: 12.5.9
```
fevisera, line 28: Please, bump the chart version to resolve the conflict.

fevisera:
Hi @girstenbrei, sorry for the delay.
Please, resolve the conflict by bumping the version.

Thank you.

PR #17201, opened 2023-06-20
Common - Adding app.kubernetes.io/version to common labels

psimms-r7:
### Description of the change

Adding app.kubernetes.io/version to common labels

https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/

### Benefits

Adding version label to all charts

### Applicable issues

- fixes #

### Checklist


joancafom:
Hi @psimms-r7 

Thanks for the PR! Could you please rebase and solve the conflicts?

As this would ultimately affect the entire catalog (the `common.labels.standard` is virtually used everywhere in the repo) I would like to be extra-cautious and perform some upgrade test in order to ensure nothing breaks (some fields might be immutable).

juan131:
I agree with @joancafom to be very cautious wit this change.

The `app.kubernetes.io/version` label must be treated with caution given its non-inmutable values on each chart release (something similar happens with `helm.sh/chart`).

There are fields on K8s objects that are inmutable. For instance, `spec.selector.matchLabels` are inmutable in deployments, daemonsets, statefulsets, etc. Therefore, we might be very cautious to avoid including these kind of labels on these fields and avoid introducing backwards incompatibilities.

juan131:
We might be ready to accept this contribution once we finish the work at https://github.com/bitnami/charts/pull/18154 + all related follow-up PRs

juan131:
@psimms-r7 I just merged the latest changes from `main` in your branch so we can continue with the review process of this PR

bitnami/common/templates/_labels.tpl
```diff
@@ -26,6 +27,7 @@ app.kubernetes.io/name: {{ include "common.names.name" . }}
 helm.sh/chart: {{ include "common.names.chart" . }}
 app.kubernetes.io/instance: {{ .Release.Name }}
 app.kubernetes.io/managed-by: {{ .Release.Service }}
+app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
```
juan131, line 30: There's a main concern regarding this label to be included in the `common.labels.standard`: multi-application charts.

I mean, this works well in charts such as NGINX where there's a single component (`nginx`), see:

- https://github.com/bitnami/charts/tree/main/bitnami/nginx/templates

In these cases, it makes sense to set this label using `.Chart.AppVersion` since all the manifests refer to NGINX.

However, in charts such as Airflow where there are several components (airflow, airflow-scheduler, airflow-worker, etc) each of them having a different version, see:

- https://github.com/bitnami/charts/tree/main/bitnami/airflow/templates
- https://github.com/bitnami/charts/blob/main/bitnami/airflow/values.yaml

It makes less sense to use this approach, since we'll have components manifests (e.g. airflow-scheduler) with `app.kubernetes.io/version` label set with the value of the main component, which might be different.

What do you think @javsalgar @carrodher ?

juan131: A possible approach would be to use "customLabels" argument received by `common.labels.standard` to overwrite the value of `app.kubernetes.io/version` on sub-components labels.

juan131: Good idea! Please check this PR @javsalgar : https://github.com/bitnami/charts/pull/19223

juan131 (approved):
LGTM

psimms-r7:
thank you all 👍

jouve:
@juan131 @joancafom semver  is expected for [version](https://helm.sh/docs/topics/charts/#the-chartyaml-file):

Given a version number MAJOR.MINOR.PATCH, increment the:

    MAJOR version when you make incompatible API changes
    MINOR version when you add functionality in a backward compatible manner
    PATCH version when you make backward compatible bug fixes


so this should have been at least a minor increment.

PR #17237, opened 2023-06-20
[bitnami/redis] Try to seed redis with pss-restricted

jcpunk:
### Description of the change

Enhance security controls for redis pods

### Benefits

Greater compliance with kubernetes pod security best practices

### Possible drawbacks

Users doing custom things may need to adjust their settings.

### Applicable issues

- fixes #

### Additional information

https://kubernetes.io/docs/concepts/security/pod-security-standards/

### Checklist

### DIFF
--- a/bitnami/redis/values.yaml
+++ b/bitnami/redis/values.yaml
@@ -279,10 +279,23 @@ master:
   ## ref: https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-pod
   ## @param master.containerSecurityContext.enabled Enabled Redis&reg; master containers' Security Context
   ## @param master.containerSecurityContext.runAsUser Set Redis&reg; master containers' Security Context runAsUser
+  ## @param master.containerSecurityContext.runAsGroup Set Redis&reg; master containers' Security Context runAsGroup
+  ## @param master.containerSecurityContext.runAsNonRoot Set Redis&reg; master containers' Security Context runAsNonRoot
+  ## @param master.containerSecurityContext.allowPrivilegeEscalation Is it possible to escalate Redis&reg; pod(s) privileges
+  ## @param master.containerSecurityContext.seccompProfile.type Set Redis&reg; master containers' Security Context seccompProfile
+  ## @param master.containerSecurityContext.capabilities.drop Set Redis&reg; master containers' Security Context capabilities to drop
   ##
   containerSecurityContext:
     enabled: true
     runAsUser: 1001
+    runAsGroup: 0
+    runAsNonRoot: true
+    allowPrivilegeEscalation: false
+    seccompProfile:
+      type: RuntimeDefault
+    capabilities:
+      drop:
+        - ALL
   ## @param master.kind Use either Deployment or StatefulSet (default)
   ## ref: https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/
   ##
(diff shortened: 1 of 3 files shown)


carrodher:
Thanks for your contribution! It seems the metadata is missing for the new values, see https://github.com/bitnami/charts/actions/runs/5326572489/jobs/9648784747?pr=17237

jcpunk:
Does this look better?

andresbono (commented):
Thank you for your contribution! While it is true that restricting the container security context as much as possible is a best practice in terms of security, I'm hesitant to apply these changes if we take into account existing deployments. Can you think of any breaking changes or drawbacks for existing users that may see the security context more restricted if they attempt to upgrade?

jcpunk:
It don't think this should break anything.  The GID change is the only one that I think may carry some risk.  But I'd not expect too many folks are using the GID based assignment when the UID is already set.

andresbono:
Just a quick update. We would like to unify the rules for all your PRs as much as possible to have some homogeneity. We are still pending to evaluate the regressions and risks this change may carry.

bitnami/redis/values.yaml
```diff
@@ -1194,10 +1220,21 @@ sentinel:
...
   ## @param sentinel.containerSecurityContext.runAsUser Set Redis&reg; Sentinel containers' Security Context runAsUser
+  ## @param sentinel.containerSecurityContext.runAsGroup Set Redis&reg; Sentinel containers' Security Context runAsGroup
+  ## @param sentinel.containerSecurityContext.runAsNonRoot Set Redis&reg; Sentinel containers' Security Context runAsNonRoot
+  ## @param sentinel.containerSecurityContext.seccompProfile.type Set Redis&reg; Sentinel containers' Security Context seccompProfile
+  ## @param sentinel.containerSecurityContext.capabilities.drop Set Redis&reg; Sentinel containers' Security Context capabilities to drop
   ##
   containerSecurityContext:
     enabled: true
     runAsUser: 1001
+    runAsGroup: 0
```
andresbono, line 1234: Is there any reason to not include `allowPrivilegeEscalation: false` for this container?

bitnami/redis/values.yaml
```diff
@@ -1531,10 +1568,21 @@ metrics:
...
   ## @param metrics.containerSecurityContext.runAsUser Set Redis&reg; exporter containers' Security Context runAsUser
+  ## @param metrics.containerSecurityContext.runAsGroup Set Redis&reg; exporter containers' Security Context runAsGroup
+  ## @param metrics.containerSecurityContext.runAsNonRoot Set Redis&reg; exporter containers' Security Context runAsNonRoot
+  ## @param metrics.containerSecurityContext.seccompProfile.type Set Redis&reg; exporter containers' Security Context seccompProfile
+  ## @param metrics.containerSecurityContext.capabilities.drop Set Redis&reg; exporter containers' Security Context capabilities to drop
   ##
   containerSecurityContext:
     enabled: true
     runAsUser: 1001
+    runAsGroup: 0
```
andresbono, line 1582: Is there any reason to not include `allowPrivilegeEscalation: false` for this container?

bitnami/redis/values.yaml
jcpunk: Those are set at the pod level rather than the individual container. I'm not sure I can set them on the individual container.

jcpunk: That was an oversight, thanks for pointing it out!

bitnami/redis/values.yaml
jcpunk: That was an oversight, thanks for pointing it out!

andresbono (approved):

andresbono:

PR #17279, opened 2023-06-21
[bitnami/jupyterhub] Add jupyterhub proxy serviceAccount settings

### LINKED ISSUE #17255: [bitnami/jupyterhub] proxy misses serviceAccount settings
### Name and Version

bitnami/jupyterhub 4.1.4

Currently, the setting automountServiceAccountToken is present to be set in ServiceAccounts and can be set through the values.yaml (e.g. .Values.serviceAccount.automountServiceAccountToken). Kubernetes resources split this automatism into two settings, one in the ServiceAccount, one in the PodSpec [see reference](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#opt-out-of-api-credential-automounting).

With the current settings, it is not possible to influence the PodSpec setting via values.yaml. This means a service account token created by a third party service may be inadvertently mounted when matching the serviceAccountName (defaulting to default when .Values.serviceAccount.create).

Adding a `.Values.proxy.serviceAccount` setting similar to the `.Values.hub.serviceAccount`. 

Something I have noticed is that there are many of these small features which are not standardized between charts in this repository. Other things are: inability to disable init containers which simply wait for some networking, missing default resource limits. Could there be a way to standardize (some of) those into the standard library bitnami uses?

I'm mainly concerned about using the default serviceAccount, and since this chart has no settings for it the only alternative is keeping my own copy of it. At that point a PR is less work.

bh-tt:
### Description of the change
Add serviceaccount settings to proxy deployment

### Benefits
Allow disabling or use of custom serviceaccount

### Applicable issues

- fixes #17255

### Checklist

### DIFF
--- a/bitnami/jupyterhub/Chart.yaml
+++ b/bitnami/jupyterhub/Chart.yaml
@@ -25,4 +25,4 @@ maintainers:
 name: jupyterhub
 sources:
   - https://github.com/bitnami/charts/tree/main/bitnami/jupyterhub
-version: 4.1.4
+version: 4.1.5
--- a/bitnami/jupyterhub/README.md
+++ b/bitnami/jupyterhub/README.md
@@ -275,6 +275,15 @@ The command removes all the Kubernetes components associated with the chart and 
 | `proxy.pdb.minAvailable`                      | Set minimum available proxy instances                                                                                    | `""`                              |
 | `proxy.pdb.maxUnavailable`                    | Set maximum available proxy instances                                                                                    | `""`                              |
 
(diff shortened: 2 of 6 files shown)


rafariossaa (changes requested):
I think you forgot the corresponding `service-account.yaml`.
Am I right ?

bh-tt:
Yes, did a `git commit -a` which of course does not add new files to the commit.

rafariossaa (approved):
LGTM
Thanks for adding this.

PR #17302, opened 2023-06-22
[bitnami/postgresql] checksum only data part of the ConfigMap

jouve:
### Description of the change

Adapt the rollout of pods on configmap change pattern : https://helm.sh/docs/howto/charts_tips_and_tricks/#automatically-roll-deployments) to trigger the rollout only when the "data" of the configmap is change, instead of the whole configmap.

### Benefits

in the current situation, deploying an upgrade of the chart (version change in Chart.yaml) will trigger a reload of the pods, even it the pod template and configmap config data are not changed.

This happens because the version of the chart is included in "helm.sh/chart" in metadata.labels (`printf "%s-%s" .Chart.Name .Chart.Version`) of the configmap,

so a change of chart version changes the configmap metadata.labels which changes the checksum.

With this change, only changes to data will be considered, so reload of pod will happen only if the pod template change or the configuration changes.

### Applicable issues

- fixes #

### Checklist


joancafom (approved):
LGTM! Thanks for your PR 🚀

PR #17325, opened 2023-06-23
[bitnami/influxdb] Add possibility to specify labels for ServiceMonitor

stiantoften:
### Description of the change

This change adds an extra parameter for labels for the created servicemonitor. This is required if we want to use the prometheus-operator serviceMonitorSelector with the matchLabels parameter. Additional labels for the servicemonitor could be specified via the current metrics.serviceMonitor.selector, but this also adds it to the selector inside the servicemonitor. I believe it to be an error that serviceMonitor.selector is added to the labels field, but it is probably a good idea to keep it in case someone relies on this functionality. This also makes the configuration more inline with servicemonitor configuration from other charts.

### Benefits

- Ability to specify labels on the servicemonitor without also adding labels in other places.
- Better consistency with other similar charts.

### Possible drawbacks

N/A

### Applicable issues

- fixes #

### Checklist


rafariossaa (approved):
Thanks for adding this.
LGTM.

PR #17333, opened 2023-06-23
[bitnami/clickhouse] Add clickhouse shard labels

amalgamm:
### Description of the change

Adds a customShardLabel to clickhouse statefullset podTemplate

### Benefits

Allows to create more accurate affinity rules. For example force replica pods of same shard to schedule in different AZ.

### Possible drawbacks

--

### Applicable issues

--

### Additional information

--

### Checklist

### DIFF
--- a/bitnami/clickhouse/templates/statefulset.yaml
+++ b/bitnami/clickhouse/templates/statefulset.yaml
@@ -39,6 +39,9 @@ spec:
         {{- if $.Values.podLabels }}
         {{- include "common.tplvalues.render" (dict "value" $.Values.podLabels "context" $) | nindent 8 }}
         {{- end }}
+        {{- if $.Values.customShardLabel }}
+        {{- printf "%s: %d" $.Values.customShardLabel.name $i | nindent 8 }}
+        {{- end }}
     spec:
       serviceAccountName: {{ template "clickhouse.serviceAccountName" $ }}
       {{- include "clickhouse.imagePullSecrets" $ | nindent 6 }}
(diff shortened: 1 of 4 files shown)


andresbono:
Hi, sorry for the late reply. We will get back to you soon with some review comments. Thank you for your contribution!

bitnami/clickhouse/templates/statefulset.yaml
```diff
@@ -39,6 +39,9 @@ spec:
         {{- if $.Values.podLabels }}
         {{- include "common.tplvalues.render" (dict "value" $.Values.podLabels "context" $) | nindent 8 }}
         {{- end }}
+        {{- if $.Values.customShardLabel }}
+        {{- printf "%s: %d" $.Values.customShardLabel.name $i | nindent 8 }}
+        {{- end }}
```
andresbono, line 44: I think it is a great idea to set a label that allows identifying the shard the pod belongs to. But what is the value of receiving the label name from inputs? I would suggest to just add the label in all the cases, as we are already doing for [bitnami/mongodb-sharded](https://github.com/bitnami/charts/blob/a8eb95ec6a71e695157a56c3d60122c8e3844b5e/bitnami/mongodb-sharded/templates/shard/shard-data-statefulset.yaml#L40). It will be good to follow the same approach for consistency between solutions:

```suggestion
        shard: {{ $i | quote }}
```

If you still see some value on receiving the label name from inputs, we can discuss it. But I would use a string `customShardLabelName` directly instead of a hash/dictionary if, in the end, only `customShardLabel.name` is taken into account.

andresbono (commented):
Thank you for your contribution! Could you please take a look at my review?

PR #17395, opened 2023-06-29
[bitnami/cert-manager] Remove conflicting labels from CRDs

gw0:
### Description of the change

Removes a conflict when deployed through ArgoCD that sets up its own `app.kubernetes.io/instance` label (https://github.com/argoproj/argo-cd/issues/6728). This is also in-line with other Bitnami-packaged CRDs (e.g. https://github.com/bitnami/charts/blob/main/bitnami/kube-prometheus/crds/crd-servicemonitor.yaml).

### Benefits

No need for workarounds like https://github.com/HariSekhon/Kubernetes-configs/blob/master/cert-manager/base/remove-instance-label.jsonpatch.yaml.

If the `app.kubernetes.io/instance` label is set then ArgoCD reports warnings like:
```
SharedResourceWarning 	CustomResourceDefinition/certificaterequests.cert-manager.io is part of applications cert-manager-production and cert-manager
```

### Checklist

### DIFF
--- a/bitnami/cert-manager/Chart.yaml
+++ b/bitnami/cert-manager/Chart.yaml
@@ -26,4 +26,4 @@ maintainers:
 name: cert-manager
 sources:
   - https://github.com/bitnami/charts/tree/main/bitnami/cert-manager
-version: 0.11.5
+version: 0.12.0
--- a/bitnami/cert-manager/templates/crds/crd-certificaterequests.yaml
+++ b/bitnami/cert-manager/templates/crds/crd-certificaterequests.yaml
@@ -9,8 +9,6 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   name: certificaterequests.cert-manager.io
-  labels: {{- include "common.labels.standard" . | nindent 4 }}
-    app: {{ template "common.names.name" . }}
 spec:
   group: cert-manager.io
   names:
--- a/bitnami/cert-manager/templates/crds/crd-certificates.yaml
+++ b/bitnami/cert-manager/templates/crds/crd-certificates.yaml
@@ -9,8 +9,6 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   name: certificates.cert-manager.io
-  labels: {{- include "common.labels.standard" . | nindent 4 }}
-    app: {{ template "common.names.name" . }}
 spec:
   group: cert-manager.io
   names:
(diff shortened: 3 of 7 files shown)


aoterolorenzo:
Hi @gw0,

Could you explain a little more about the issue you mention and how this helps? I understand that some charts have already removed the labels but my concern is that this could be a breaking change for existent deployments when upgrade.

gw0:
> Could you explain a little more about the issue you mention and how this helps? I understand that some charts have already removed the labels but my concern is that this could be a breaking change for existent deployments when upgrade.

The purpose of ArgoCD (https://argo-cd.readthedocs.io/) is to deploy K8s resources using GitOps principles. On one hand it renders Helm charts that should be deployed and on the other it tracks which resources it is managing. If auto-sync and self-healing are turned on it, it even automatically recreates or deletes resources so that the live state matches the Git state. To accomplish tracking of resources it by default uses the `app.kubernetes.io/instance` label and sets it to the ArgoCD Application name which can be different from the Helm Release name (https://argo-cd.readthedocs.io/en/stable/user-guide/resource_tracking/).

From my understanding the issue comes from the `{{- include "common.labels.standard" . | nindent 4 }}` that sets `app.kubernetes.io/instance: cert-manager` (https://github.com/bitnami/charts/blob/main/bitnami/common/templates/_labels.tpl#L13). In the rare case where ArgoCD Application name matches the Helm Release name, i.e. both match `cert-manager`, everything works. But ArgoCD Application names need to be globally unique across all K8s clusters it is managing, therefore ArgoCD Application names are typically prefixed/postfixed with the environment or cluster name (e.g. `cert-manager-production`). So, ArgoCD expects `app.kubernetes.io/instance: cert-manager-production`, but the rendered Helm chart sets it to `app.kubernetes.io/instance: cert-manager`. For non-CRD resources ArgoCD seems to have a workaround to deal with this, but it seems it deploys CRDs as-is (maybe other cluster-wide resources?).

A similar discrepancy would occur if one would deploy multiple `cert-manager` (e.g. different versions or certificate issuers) to enforce isolation. Although the CRDs would likely be the same, the label `app.kubernetes.io/instance` would change due to a different Helm Release name.

The only breaking change I could imagine is if somebody is using `app.kubernetes.io/instance` to delete all resources deployed together.

aoterolorenzo:
@gw0 I understand the problematic, but apart of my concerns about if it is or not a breaking change, I see that this same proposal is opened upstream at cert-manager/cert-manager#6192, sill waiting for approval. Let's see their response to keep on sync with them in this topic and if there is any reason preventing this going ahead.

PR #17421, opened 2023-06-30
[bitnami/charts] Automate license checks

rloporp:
### Description of the change

Licenses and copyright notices are included as headers in all our public repositories.
We'll automatically check license headers with the [SkyWalking Eyes](https://github.com/apache/skywalking-eyes) GitHub action.

### Benefits

Automatic license compliance.

### Possible drawbacks

No known drawbacks.

### Applicable issues

None.

### Checklist

### DIFF
--- a/.github/workflows/license-headers.yml
+++ b/.github/workflows/license-headers.yml
@@ -1,0 +1,16 @@
+# Copyright VMware, Inc.
+# SPDX-License-Identifier: APACHE-2.0
+
+---
+name: license-headers
+
+on:
+  pull_request:
+
+jobs:
+  license-headers-linter:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - name: Check license Headers
+        uses: apache/skywalking-eyes/header@v0.4.0
--- a/.github/workflows/license-headers.yml
+++ b/.github/workflows/license-headers.yml
@@ -2,7 +2,7 @@
 # SPDX-License-Identifier: APACHE-2.0
 
 ---
-name: [License] Check license headers
+name: '[License] Check license headers'
 
 on:
   pull_request:
--- a/.github/workflows/license-headers.yml
+++ b/.github/workflows/license-headers.yml
@@ -2,7 +2,7 @@
 # SPDX-License-Identifier: APACHE-2.0
 
 ---
-name: license-headers
+name: [License] Check license headers
 
 on:
   pull_request:
--- a/.licenserc.yaml
+++ b/.licenserc.yaml
@@ -1,0 +1,19 @@
+# Copyright VMware, Inc.
+# SPDX-License-Identifier: APACHE-2.0
+
+---
+header:
+  comment: on-failure
+  license:
+    spdx-id: Apache-2.0
+    copyright-owner: VMware, Inc.
+
+    pattern: |
+      {{- /*
+      Copyright VMware, Inc.
+      SPDX-License-Identifier: APACHE-2.0
+      */}}
+
+  paths:
+    - "bitnami/**/*.yml"
+    - "bitnami/**/*.yaml"
--- a/.licenserc.yaml
+++ b/.licenserc.yaml
@@ -3,22 +3,6 @@
 
 ---
 header:
-  # License format for Chart.yaml and values files
-  - comment: on-failure
-    license:
-      spdx-id: Apache-2.0
-      copyright-owner: VMware, Inc.
-
-      content: |
-        # Copyright VMware, Inc.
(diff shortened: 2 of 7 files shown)


agarcia-oss:
I think this PR should also update the [CONTRIBUTING](https://github.com/bitnami/charts/blob/main/CONTRIBUTING.md) guidelines since it includes the restriction of license headers. WDTY?

bitnami/clickhouse/Chart.yaml
```diff
@@ -28,4 +28,4 @@ maintainers:
 name: clickhouse
 sources:
   - https://github.com/bitnami/charts/tree/main/bitnami/clickhouse
-version: 3.5.1
+version: 3.5.2
```
carrodher, line 31:
```suggestion
version: 3.5.1
```

bitnami/jupyterhub/Chart.yaml
```diff
@@ -28,4 +28,4 @@ maintainers:
 name: jupyterhub
 sources:
   - https://github.com/bitnami/charts/tree/main/bitnami/jupyterhub
-version: 4.1.6
+version: 4.1.7
```
carrodher, line 31:
```suggestion
version: 4.1.6
```

carrodher (commented):
Thanks, Raul! Given that this contribution is a little bit special let's try to simplify the process.

Usually, it is required to create a single PR per chart but in this case, we are modifying a couple of Helm charts in a single PR, this will cause issues when testing/releasing the Helm charts since the test & release pipelines only expect a single chart.
Given said that, we can avoid the version bump in the _Chart.yaml_ to avoid triggering a new release. This is possible because the nature of the change is not modifying the templates themselves, just adding a comment. If we don't bump the chart version, the changes will be present in the GitHub code but we won't publish a new Helm chart.

TL;DR
By reverting the version bump there will be a failure in one of the GH actions but the whole test & release process will be smooth, mainly because a release won't be triggered.

.github/workflows/license-headers.yml
```diff
@@ -0,0 +1,16 @@
+# Copyright VMware, Inc.
+# SPDX-License-Identifier: APACHE-2.0
+
+---
+name: license-headers
```
carrodher, line 5: We usually add a prefix to order the actions (i.e `[CI/CD] ...`, `[Support] ...`, etc) and a descriptive title

What do you think about using something like this?
```suggestion
name: [License] Check license headers
```

.licenserc.yaml
```diff
@@ -0,0 +1,37 @@
...
+      spdx-id: Apache-2.0
+      copyright-owner: VMware, Inc.
+
+      content: |
+        {{- /*
+        Copyright VMware, Inc.
+        SPDX-License-Identifier: APACHE-2.0
+        */}}
+
+    paths:
```
carrodher, line 18: There are also a `_helpers.tpl` file present in most Helm charts that would be nice to check as well, see https://github.com/bitnami/charts/blob/main/bitnami/postgresql/templates/_helpers.tpl

.licenserc.yaml
```diff
@@ -0,0 +1,37 @@
...
+      - "bitnami/**/templates/**/*.yml"
+      - "bitnami/**/templates/**/*.yaml"
```
carrodher, line 20: Are we contemplating files directly in `/templates` with that approach or just files under another subdirectory `templates/foo`?
I.e `bitnami/postgresql/templates/read/svc.yaml` and `bitnami/postgresql/templates/svc.yaml`

carrodher:
PS: Failure in the _"Get modified charts"_ GH action is expected, we can merge the PR with this error

.licenserc.yaml
```diff
@@ -0,0 +1,37 @@
+# Copyright VMware, Inc.
+# SPDX-License-Identifier: APACHE-2.0
+
+---
+header:
```
andresbono, line 5: Should we cover [.vib](https://github.com/bitnami/charts/tree/main/.vib) and [template](https://github.com/bitnami/charts/tree/main/template) in this PR or in a follow-up?

rloporp: I agree we should double-check several other files, but we're limiting the scope to YAML files (`Chart.yaml`, `values.yaml` and Helm templates) in this first iteration.

.licenserc.yaml
carrodher: 👍 Thanks

.licenserc.yaml
carrodher: Perfect, totally fine

carrodher (approved):
LGTM

Issue #17479, opened 2023-07-04
 Cannot invoke "kafka.raft.KafkaRaftManager.apiVersions()" because the return value of "kafka.server.SharedServer.raftManager()" is null


shameoff:
### Name and Version

bitnami/kafka 23.0.1

### What architecture are you using?

amd64

### What steps will reproduce the bug?

Using this docker-compose.yml(delete .txt) start first container.
[docker-compose.yml.txt](https://github.com/bitnami/charts/files/11952647/docker-compose.yml.txt)
Or with same credentials run helm chart

### What is the expected behavior?

Sucessful start of Kafka container

### What do you see instead?

`"Cannot invoke "kafka.raft.KafkaRaftManager.apiVersions()" because the return value of "kafka.server.SharedServer.raftManager()" is null"`

### Additional information
```
 [2023-07-04 23:06:35,575] INFO Initialized snapshots with IDs Set() from /bitnami/kafka/data/__cluster_metadata-0 (kafka.raft.KafkaMetadataLog$)
ejudge-infrastructure-kafka-1  | [2023-07-04 23:06:35,579] ERROR [SharedServer id=1] Got exception while starting SharedServer (kafka.server.SharedServer)
ejudge-infrastructure-kafka-1  | org.apache.kafka.common.KafkaException: java.lang.IllegalArgumentException: No serviceName defined in either JAAS or Kafka config
ejudge-infrastructure-kafka-1  |        at org.apache.kafka.common.network.SaslChannelBuilder.configure(SaslChannelBuilder.java:184)
ejudge-infrastructure-kafka-1  |        at org.apache.kafka.common.network.ChannelBuilders.create(ChannelBuilders.java:192)
ejudge-infrastructure-kafka-1  |        at org.apache.kafka.common.network.ChannelBuilders.clientChannelBuilder(ChannelBuilders.java:81)
ejudge-infrastructure-kafka-1  |        at kafka.raft.KafkaRaftManager.buildNetworkClient(RaftManager.scala:285)
ejudge-infrastructure-kafka-1  |        at kafka.raft.KafkaRaftManager.buildNetworkChannel(RaftManager.scala:253)
ejudge-infrastructure-kafka-1  |        at kafka.raft.KafkaRaftManager.<init>(RaftManager.scala:171)
ejudge-infrastructure-kafka-1  |        at kafka.server.SharedServer.start(SharedServer.scala:247)
[...]

PR #13308, opened 2022-11-02
[bitnami/kafka] Add support set config broker.rack with AWS availability zone info

DingGGu:
### Benefits
- `broker.rack` will helpful to failures across AZ by setting a partition assignment in the cloud environment or physical racks.
- Client can get hints from brokers for consuming from same rack.

### Applicable issues

  - fixes https://github.com/bitnami/charts/issues/11226

### Additional information
https://cwiki.apache.org/confluence/display/KAFKA/KIP-392%3A+Allow+consumers+to+fetch+from+closest+replica
https://aws.amazon.com/blogs/big-data/reduce-network-traffic-costs-of-your-amazon-msk-consumers-with-rack-awareness/

### Checklist

- [x] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/). This is *not necessary* when the changes only affect README.md files.
- [x] Variables are documented in the values.yaml and added to the `README.md` using [readme-generator-for-helm](https://github.com/bitnami-labs/readme-generator-for-helm)
- [x] Title of the pull request follows this pattern [bitnami/<name_of_the_chart>] Descriptive title
- [x] All commits signed off and in agreement of [Developer Certificate of Origin (DCO)](https://github.com/bitnami/charts/blob/main/CONTRIBUTING.md#sign-your-work)

PR #17507, opened 2023-07-06
[bitnami/kafka] Chart refactor

migruiz4:
### Description of the change

This PR is a refactor for the `bitnami/kafka` chart which includes the following changes:

- Architecture refactor. This chart now includes two statefulsets: 1 for either `controller`/`controller+broker` nodes and another for `broker` only nodes.
- Listeners configuration refactored to map with Kafka settings
- SASL settings have been enhanced.
- SSL settings have been slightly modified.
- Support for Zookeeper mode migration.
- Minor bug fixes, especially related to Kraft mode.
- Many values have been removed in favor of `extraConfig` settings. Instead of having 1 chart value per Kafka setting, users can now extend their Kafka configuration as they please by using the `extraConfig` value. 
- Adds support for SCRAM authentication in KRaft mode.
- ACL authorization settings refactored and updated notes on KRaft configuration.

More in-depth details about values that have been removed or renamed can be found in the upgrading notes.
Please carefully look at the notes when upgrading your Kafka chart.

### Benefits

- The new architecture allows scaling Kafka cluster per role individually.
- Enhanced security by enabling `readOnlyRootFs=true`.
- Improve user experience with clearer `listeners`, `sasl` and `ssl` values, as well as the direct relation between chart settings and Kafka settings.
- Avoid issues caused by outdated default values.  By removing those settings from the chart, their values directly rely on upstream Kafka defaults.

### Possible drawbacks

Due to the new architecture design, this major upgrade requires users to manually attach existing PVC in the new pods.

If migrating from Zookeeper mode, we recommend using controller-eligible pods as a replacement for previous Kafka brokers.

If migrating from Kraft mode, use controller-eligible for `controler-only` or `controller+broker` nodes and `broker-only` nodes for dedicated brokers.

### Additional information

CURRENTLY WIP:

- Add detailed upgrading notes.
- Add a migration guide from Zookeeper mode into Kraft mode.
- Fix issue with affinities, where a `controller-eligible` and a `broker-only` could be deployed in the same node (at the moment, this also affects other charts with multiple statefulsets).

### Checklist


javsalgar:
I tried running `helm template` but it failed

```
» helm template .
Error: parse error at (kafka/templates/_helpers.tpl:652): function "joing" not defined

```

javsalgar:
I tried generating the template but it is not working

```
» helm template kafka .                                                                                                              [17/07/23 | 15:20:49]
Error: parse error at (kafka/templates/secrets.yaml:50): undefined variable "$secretValue"

```

.vib/kafka/goss/goss.yaml
```diff
@@ -2,32 +2,30 @@
...
     owner: root
-  /opt/bitnami/kafka/config/kraft/server.properties:
+  /opt/bitnami/kafka/config/server.properties:
     exists: true
     filetype: file
-    mode: "0664"
-    owner: root
-    group: root
+    mode: "0644"
     contains:
```
javsalgar, line 19: I'm missing connectivity and capability testing. See the example of Milvus goss.yaml. If there are metrics it would be nice to test them too.

bitnami/kafka/templates/_helpers.tpl
```diff
@@ -346,19 +353,457 @@
...
+ssl.truststore.location=/opt/bitnami/kafka/config/certs/kafka.truststore.jks
+#ssl.keystore.password=
+#ssl.truststore.password=
+#ssl.key.password=
+ssl.client.auth={{ .Values.tls.sslClientAuth }}
+ssl.endpoint.identification.algorithm={{ .Values.tls.endpointIdentificationAlgorithm }}
+{{- end }}
+{{- if .Values.acl.authorizerClassName }}
+authorizer.class.name={{ .Values.acl.authorizerClassName }}
+{{- end }}
```
javsalgar, line 509: I see values like `acl.allowEveryineIfNoAclFound` that are directly translated into a `allow.everyone.if.no.acl.found` entry. In other charts like Milvus or APISIX, we have something like this

```
  defaultConfig: |
    {{- if .Values.dataPlane.metrics.enabled }}
    plugin_attr:
      prometheus:
        export_uri: /apisix/prometheus/metrics
        metric_prefix: apisix_
        enable_export_server: true
        export_addr:
          ip: 0.0.0.0
          port: {{ .Values.dataPlane.containerPorts.metrics }}
    {{- end }}
    apisix:
      node_listen: {{ .Values.dataPlane.containerPorts.http }}

  extraConfig: {}
```
(the reason for having the default configuration as a value is for giving more flexibility to advanced users to modify all configurations at their will, even though that may imply other parts of the deployment not working as expected. This was taken from other upstream charts examples like grafana-tempo)

Suppose that I want to change `plugin_attr.prometheus.metric_prefix`, I could do something like this in my custom values.yaml file:

```
extraConfig:
  plugin_attr:
    prometheus:
      metric_prefix: test_
```

or to do `helm install ... --set extraConfig.plugin_attr.prometheus.metric_prefix=test_`. This is afterward merged and overridden with the proper priority order in an init container. Would it be possible to do something similar here? It would make the chart simpler and more flexible

bitnami/kafka/templates/broker/statefulset.yaml
```diff
@@ -0,0 +1,441 @@
...
+      {{- if .Values.broker.priorityClassName }}
+      priorityClassName: {{ .Values.broker.priorityClassName }}
+      {{- end }}
+      {{- if .Values.broker.podSecurityContext.enabled }}
+      securityContext: {{- omit .Values.broker.podSecurityContext "enabled" | toYaml | nindent 8 }}
+      {{- end }}
+      serviceAccountName: {{ include "kafka.serviceAccountName" . }}
+      initContainers:
+        {{- if and .Values.volumePermissions.enabled .Values.broker.persistence.enabled }}
+        - name: volume-permissions
```
javsalgar, line 94: Actually, it is not clear why this one is not using a helper like the rest of the init containers.

bitnami/kafka/templates/controller-eligible/statefulset.yaml
```diff
@@ -0,0 +1,438 @@
...
+            - name: KAFKA_CONTROLLER_PASSWORD
+              valueFrom:
+                secretKeyRef:
+                  name: {{ include "kafka.saslSecretName" . }}
+                  key: controller-password
+            {{- end }}
+            {{- end }}
+            {{- if .Values.metrics.jmx.enabled }}
+            - name: JMX_PORT
+              value: "5555"
```
javsalgar, line 192: Any reason for hardcoding the metrics port?

bitnami/kafka/templates/broker/statefulset.yaml
```diff
@@ -0,0 +1,441 @@
...
+                name: {{ include "common.tplvalues.render" (dict "value" .Values.broker.extraEnvVarsSecret "context" $) }}
+            {{- end }}
+            {{- if .Values.extraEnvVarsSecret }}
+            - secretRef:
+                name: {{ include "common.tplvalues.render" (dict "value" .Values.extraEnvVarsSecret "context" $) }}
+            {{- end }}
+          {{- end }}
+          ports:
+            - name: client
+              containerPort: {{ .Values.listeners.client.port }}
```
javsalgar, line 230: Could you explain why this one is not following the `containerPorts` model?

bitnami/kafka/templates/_helpers.tpl
```diff
@@ -346,19 +353,457 @@ Return the Kafka configuration configmap
...
+{{/*
+Zookeeper connection section of the server.properties
+*/}}
+{{- define "kafka.zookeeperConfig" -}}
+zookeeper.connect={{ include "kafka.zookeeperConnect" . }}
+#broker.id=
+{{- if .Values.sasl.zookeeper.user }}
+sasl.jaas.config=org.apache.kafka.common.security.plain.PlainLoginModule required \
+    username="{{ .Values.sasl.zookeeper.user }}" \
+    password="zookeeper-password-placeholder";
```
javsalgar, line 559: FWIW, In other solutions we use `render-template`, mapping it to an env var and thus not having to perform logic like `sed` in init containers. Not sure if kafka has `render-template`

bitnami/kafka/templates/controller-eligible/statefulset.yaml
migruiz4: That part is imported from current logic: https://github.com/bitnami/charts/blob/e274a79d85ef6abe97bd3ebb9182fe549692829b/bitnami/kafka/templates/statefulset.yaml#L315-L318

bitnami/kafka/templates/broker/statefulset.yaml
migruiz4: Because I wanted to model all Kafka settings under `listeners`, in this case, `listeners.<listener_name>.port`.
Adding `containerPorts.*`, would mean removing `listeners.<listener_name>.port` as it would not make sense having to configure both.

bitnami/kafka/templates/broker/statefulset.yaml
migruiz4: This section is copied as it is from current logic, I did not modify it to use a helper

.vib/kafka/goss/goss.yaml
migruiz4: Isn't `create-kafka-topic` connectivity and capability testing?

bitnami/kafka/templates/_helpers.tpl
migruiz4: In this case, init-container does not use `bitnami/kafka` container but `bitnami/bitnami-shell` container which does not include `render-template`.

bitnami/kafka/templates/_helpers.tpl
migruiz4: My issue with ACL is that we are setting default `allowEveryineIfNoAclFound=true`:
https://github.com/bitnami/charts/blob/3bfcc0812171abc52f11edc32c2d31650a1bbc8c/bitnami/kafka/values.yaml#L223-L225

While Kafka sets it default to `false` (safest) and discourages users to set it to `true` in production environments:
https://docs.confluent.io/platform/current/kafka/authorization.html#super-users

For that reason, I made it so it relies on default settings unless explicitly set.

javsalgar: The thing is, would it be possible to extend the extra config section to allow overriding values?

bitnami/kafka/templates/controller-eligible/statefulset.yaml
javsalgar: Yes, but now that we're performing a major bump, would it be possible to parameterize it?

bitnami/kafka/templates/broker/statefulset.yaml
javsalgar: The thing is, this would force to have the same port in the service and the container, is that correct?

.vib/kafka/goss/goss.yaml
javsalgar: I see these ports:

```
service:
  ports:
    client: 9092
    controller: 9093
    interbroker: 9094
    external: 9095
```

If these are changed, does it work?

bitnami/kafka/templates/broker/statefulset.yaml
migruiz4: No no, `service.ports` still applies, but `containerPorts.client` would be replaced by `listeners.client.port`.

I thought it may be more intuitive for users to have all listener-related settings under a common object instead of spread among different places `containerPorts`, `auth.client.protocol`, `tls.client.sslClientAuth`, `clientListenerName`, etc.:
```
listeners:
  client:
    port: 9092
    protocol: SASL_PLAINTEXT
    name: CLIENT
    sslClientAuth: ""
```
This common configuration interface is also used to configure the `listeners.extraListeners`, see for example:
```
{{- $listeners := list .Values.listeners.client .Values.listeners.interbroker -}}
{{- range $i := .Values.listeners.extraListeners -}}
{{- $listeners = append $listeners $i -}}
{{- end -}}
...
{{- range $listener := $listeners -}}
{{- if regexFind "SSL" (upper $listener.protocol) -}}
...
{{- end -}}
{{- end -}}
```

bitnami/kafka/templates/_helpers.tpl
migruiz4: I'm not sure what you mean by 'overriding values'.
Configuration can be extended using the new `extraConfig` value or entirely override using `config`/`existingConfigmap`. 
Overriding may be complicated as I don't know how Kafka handles duplicated settings.

That is why, by default, the configuration would be minimal:
- Kraft settings (if kraft enabled): `node.id, process.roles, controller.quorum.voters , controller.listener`
- Zookeeper settings (if zookeeper enabled): `broker.id, zookeeper.connect`
- Listeners settings: `listeners, listeners.security.protocol.map, advertised.listeners`
- Sasl settings: `listener.name.*.jaas.config`
- TLS settings: `(certs location, passwords, etc.) + listener.name.*.ssl.client.auth`

Using a default value for `acl.*` settings would mean including them in the list of 'default' settings

bitnami/kafka/templates/_helpers.tpl
```diff
@@ -98,181 +79,181 @@ Return the proper Docker Image Registry Secret Names
...
+{{- true -}}
 {{- end -}}
 {{- end -}}
 
 {{/*
-Return true if encryption via TLS for inter broker communication connections should be configured
+Return the Kafka SASL credentials secret
 */}}
-{{- define "kafka.interBroker.tlsEncryption" -}}
-{{- $tlsProtocols := list "tls" "mtls" "sasl_tls" -}}
```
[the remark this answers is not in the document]
migruiz4: Do you mean instead of `{{- printf "%s" (tpl $secretName $) -}}`?

.vib/kafka/goss/goss.yaml
migruiz4: Fixed an issue with the tests when changing ports and added new parameters to modify the service ports.

Aaron-23:
https://github.com/bitnami/charts/issues/17479
Will the problem be resolved in the next version？

bitnami/kafka/templates/broker/statefulset.yaml
javsalgar: I see the point, but it's true that it is not directly intuitive that listener.client.port would be mapped to the container port and not the service port. Would it make sense to at least call it `listener.client.containerPort` so it is more closely related to the standards we have in the other charts? I'm concerned of the deviation this may cause.

migruiz4: Sure, `listener.client.containerPort` sounds good to me

bitnami/kafka/templates/controller-eligible/statefulset.yaml
migruiz4: Done 👍

bitnami/kafka/values.yaml
```diff
@@ -146,541 +124,1017 @@ existingLog4jConfigMap: ""
...
-##
-transactionStateLogReplicationFactor: 1
-## @param transactionStateLogMinIsr Overridden min.insync.replicas config for the transaction topic
-##
-transactionStateLogMinIsr: 1
-## @param numIoThreads The number of threads doing disk I/O
-##
-numIoThreads: 8
-## @param numNetworkThreads The number of threads handling network requests
+## @param interBrokerProtocolVersion Override the setting 'inter.broker.protocol.version' during the ZK migration.
```
javsalgar, line 127: Putting it here but this is actually something for the statefulsets. Could you add the `runtimeClassName` value (check other charts)? This is a requirement of the NSA Kubernetes checklist

bitnami/kafka/templates/_helpers.tpl
```diff
@@ -98,181 +79,181 @@ Return the proper Docker Image Registry Secret Names
...
 {{- end -}}
 {{- end -}}
 
 {{/*
-Return the Kafka JAAS credentials secret
+Return the Kafka TLS credentials secret
 */}}
-{{- define "kafka.jaasSecretName" -}}
-{{- $secretName := .Values.auth.sasl.jaas.existingSecret -}}
+{{- define "kafka.tlsPasswordsSecretName" -}}
```
javsalgar, line 216: There are still references to the to the old `kafka.jaasSecretName` helper

bitnami/kafka/values.yaml
```diff
@@ -146,541 +124,1017 @@
...
+    requests: {}
+  ## Kafka pods' Security Context
+  ## ref: https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-pod
+  ## @param controller.podSecurityContext.enabled Enable security context for the pods
+  ## @param controller.podSecurityContext.fsGroup Set Kafka pod's Security Context fsGroup
+  ##
+  podSecurityContext:
+    enabled: true
+    fsGroup: 1001
+  ## Kafka containers' Security Context
```
javsalgar, line 554: Missing seccompPolicy here

bitnami/kafka/values.yaml
```diff
@@ -146,541 +124,1017 @@
...
+  ## @param controller.containerSecurityContext.allowPrivilegeEscalation Force the child process to be run as non-privileged
+  ## @param controller.containerSecurityContext.readOnlyRootFilesystem Allows the pod to mount the RootFS as ReadOnly only
+  ## e.g:
+  ##   containerSecurityContext:
+  ##     enabled: true
+  ##     capabilities:
+  ##       drop: ["NET_RAW"]
+  ##     readOnlyRootFilesystem: true
+  ##
+  containerSecurityContext:
```
javsalgar, line 568: Missing the capabalities.drop=["ALL"] value, which is a requirement for running kafka in more restricted envs

bitnami/kafka/templates/_helpers.tpl
javsalgar: What about the `mergeOverwrite` function that we use in other charts?

.vib/kafka/runtime-parameters.yaml
```diff
@@ -2,31 +2,44 @@ externalAccess:
...
-      client: 9092
-      controller: 9093
-      internal: 9094
-      external: 80
-    type: LoadBalancer
+  controller:
+    service:
+      ports:
+        external: 80
+      type: LoadBalancer
```
OneCricketeer, line 9: Is this strictly necessary since Kafka is not an HTTP service? What actually happens when you try to connect to this since clients must talk to individual brokers, not through a load balancer?

bitnami/kafka/templates/secrets.yaml
```diff
@@ -0,0 +1,136 @@
...
+  provider: {{ print "bitnami" | b64enc | quote }}
+  type: {{ print "kafka" | b64enc | quote }}
+  host: {{ join "," $host | b64enc | quote }}
+  port: {{ print $port | b64enc | quote }}
+  bootstrap-servers: {{ join "," $bootstrapServers | b64enc | quote }}
+{{- end }}
+{{- end }}
+{{- end }}
+{{- if .Values.kraft.enabled }}
+apiVersion: v1
```
OneCricketeer, line 120: There seems to be a `---` separation above here

.vib/kafka/runtime-parameters.yaml
migruiz4: Hi @OneCricketeer,

Port 80 for external access service is not mandatory nor recommended for users.
The file `runtime-parameters.yaml` is the file used to deploy Kafka in our CI/CD while testing Kafka, and we use port 80 due to internal restrictions.

bitnami/kafka/templates/secrets.yaml
migruiz4: Nice catch! Indeed I missed a `---` in here, thank you for your feedback

bitnami/kafka/values.yaml
migruiz4: Added 👍

bitnami/kafka/values.yaml
migruiz4: Added 👍

bitnami/kafka/templates/_helpers.tpl
migruiz4: Nice catch, thanks!

bitnami/kafka/values.yaml
migruiz4: Added 👍

bitnami/kafka/templates/_helpers.tpl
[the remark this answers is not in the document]
javsalgar: Yes

bitnami/kafka/templates/_helpers.tpl
migruiz4: I removed `acl.*` parameters in favor of `extraConfig`

javsalgar (approved):
LGTM

gheorghestrimtu:
Hey, I am seeing this error with this version of the chart:
```
: Error: template: kafka/templates/provisioning/job.yaml:156:87: executing "kafka/templates/provisioning/job.yaml" at <.context.Values.provisioning.replicationFactor>: nil pointer evaluating interface {}.Values
```

I think it is due to these changes
https://github.com/bitnami/charts/pull/17507/files#diff-9e0087488ab0aed49e37d7ceb4eac746d2566c8abbf5e0191f48092a94c38c23L157-R157

How should I set the values correctly to avoid this error?

SergkeiM:
Hi @gheorghestrimtu there is an open issue about this: https://github.com/bitnami/charts/issues/18191

Neustradamus:
@migruiz4: Good job!

Linked to:
- https://github.com/scram-sasl/info/issues/1

bitnami/kafka/templates/scripts-configmap.yaml
```diff
@@ -62,177 +55,300 @@ data:
...
-            if [[ $pod_id -ge {{ .Values.replicaCount }} ]]; then
-                break
-            else
-                VOTERS="$VOTERS,"
-            fi
-        done
-        export KAFKA_CFG_CONTROLLER_QUORUM_VOTERS=$VOTERS
-    fi
-    {{- if eq .Values.brokerRackAssignment "aws-az" }}
-    export KAFKA_CFG_BROKER_RACK=$(curl "http://169.254.169.254/latest/meta-data/placement/availability-zone-id")
```
janvanbesien-ngdata, line 118: Was this intentionally dropped? This is a feature that was introduced in PR #13308 that we were relying on.

PR #17524, opened 2023-07-07
[bitnami/oauth2-proxy] make google-group as list instead of string

omer2500:
Signed-off-by: omer nitzan <omer2500@gmail.com>

### Description of the change

The --google-group flag on oauth2proxy may be given multiple times depends of how many groups you have.
so i converted the variable name to groups(because its already nested inside google property) and made it of type list and not string.

### Benefits

Will give the option to pass multiple google groups without the need to add it as extra args

### Possible drawbacks

None its how oauth2-proxy working

### Applicable issues

- fixes #

### Additional information

image

### Checklist

### DIFF
--- a/bitnami/oauth2-proxy/Chart.yaml
+++ b/bitnami/oauth2-proxy/Chart.yaml
@@ -32,4 +32,4 @@ maintainers:
 name: oauth2-proxy
 sources:
   - https://github.com/bitnami/charts/tree/main/bitnami/oauth2-proxy
-version: 3.7.4
+version: 3.7.5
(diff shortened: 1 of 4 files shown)


omer2500:
@carrodher 

Hi,

Im not sure hot to handle the failure on the license headers check
it seems that its not related to my changes 
what should i do?

Thanks

carrodher:
Thanks for the PR, the issue with that action is something unrelated, we are taking a look at it but it shouldn't block this contribution

carrodher:
The failing action should be solved by rebasing from _main_, could you please do that? Thanks!

omer2500:
@corico44 

first, I'm already using this flag with the bitnami chart, but passing multiple --google-group fields to the extraArgs property on the value file.

I tested my change with helm templates command and saw the output
with empty list
and a list with at least 2 items 

let me know if you want me to test this in more ways

corico44 (approved):
LGTM

PR #17551, opened 2023-07-11
[bitnami/rabbitmq-cluster-operator] update CRDs

ihor-hrytskiv:
### Description of the change

Upgrade CRD definition up to v0.12.0

### Benefits

Keep CRDs up to date.

### DIFF
--- a/bitnami/rabbitmq-cluster-operator/Chart.yaml
+++ b/bitnami/rabbitmq-cluster-operator/Chart.yaml
@@ -28,4 +28,4 @@ maintainers:
 name: rabbitmq-cluster-operator
 sources:
   - https://github.com/bitnami/charts/tree/main/bitnami/rabbitmq-cluster-operator
-version: 3.4.4
+version: 3.5.0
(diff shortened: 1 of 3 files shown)


bitnami/rabbitmq-cluster-operator/Chart.yaml
```diff
@@ -28,4 +28,4 @@ maintainers:
 name: rabbitmq-cluster-operator
 sources:
   - https://github.com/bitnami/charts/tree/main/bitnami/rabbitmq-cluster-operator
-version: 3.4.3
+version: 3.4.4
```
Mauraza, line 31: I think the chart version should be minor because the change in the crds is minor too.

```suggestion
version: 3.5.0
```

Mauraza (commented):
Hi @ihor-hrytskiv,

Thanks for the contribution!! Could you check my comment?

ihor-hrytskiv:
> Hi @ihor-hrytskiv,
> 
> Thanks for the contribution!! Could you check my comment?

I've committed a suggestion, but in [that](https://github.com/bitnami/charts/pull/15744/files) PR was changed only patch version for example

Mauraza:
> > Hi @ihor-hrytskiv,
> > Thanks for the contribution!! Could you check my comment?
> 
> Hi @Mauraza I've committed a suggestion, but in [that](https://github.com/bitnami/charts/pull/15744/files) PR was changed only patch version for example

You are right!

Mauraza (approved):
Hi @ihor-hrytskiv,

PR #17456, opened 2023-07-03
[bitnami/contour] allow to manipulate container securityContext fields for each container

ugur99:
### Description of the change
`envoy.containerSecurityContextShutdownManager` and `envoy.containerSecurityContextInitConfig` have been added and `envoy.containerSecurityContext` has been renamed to `envoy.containerSecurityContextEnvoy`.

### Benefits

This pull request aims to be able to configure the securityContext fields of the Envoy pods for each container. This would be useful for easier hardening of contour pods. Especially when using pod security admission/standards, as a k8s admin, every workload should be hardened as much as possible.

### Possible drawbacks
None

### Checklist

- [X] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/). This is *not necessary* when the changes only affect README.md files.
- [X] Variables are documented in the values.yaml and added to the `README.md` using [readme-generator-for-helm](https://github.com/bitnami-labs/readme-generator-for-helm)
- [X] Title of the pull request follows this pattern [bitnami/<name_of_the_chart>] Descriptive title
- [X] All commits signed off and in agreement of [Developer Certificate of Origin (DCO)](https://github.com/bitnami/charts/blob/main/CONTRIBUTING.md#sign-your-work)

PR #17764, opened 2023-07-18
[bitnami/contour] Support security context for shutdown manager (#17757)

### LINKED ISSUE #17757: [bitnami/contour] shutdownManager should also have a containerSecurityContext applied
### Name and Version

bitnami/contour:12.1.2

In our AKS cluster, we have policies applied, one of these is that a securityContext is required and that certain fields are provided.

I'd like to implement it so that this defaults to being the same as the envoy containerSecurityContext but that it can be overriden to a specific shutdownManager.containerSecurityContext as well.

Disabling my policies that will not make me happy nor our IT team

UXabre:
### Description of the change

This change add supports for security context on the shutdown manager sidecar

### Benefits

With this change, contour can be deployed in a compliant manner if policies (e.g. OPA policies) are applied on a kubernetes cluster.

### Possible drawbacks

None, it makes it more secure. By default however, it takes over the envoy settings, so if envoy has it configured, at least shutdown manager will have it setup as well. It can be overridden to be more (or less) strict.

### Applicable issues

- fixes #17757

### Checklist


migruiz4:
Hi @UXabre,

Thank you for your contribution, but this PR conflicts with https://github.com/bitnami/charts/pull/17456 which adds a similar feature

Issue #17727, opened 2023-07-17
[bitnami/postgresql] Identical passwords for custom user and postgres user in secret


ShiroDN:
### Name and Version

bitnami/postgresql >12.6.0

### What architecture are you using?

amd64

### What steps will reproduce the bug?

1. run `helm upgrade --install --create-namespace -n pgsql pgsql bitnami/postgresql --set auth.username=user01 --set auth.database=db01 --set auth.postgresPassword=xxxxx --set auth.password=yyyyyy --version 12.6.1`
2. secret pgsql-postgresql will have same password (yyyyyy) for both password (custom user) and postgres-password (postgres user)

### Are you using any custom parameters or values?

 --set auth.username=user01 --set auth.database=db01 --set auth.postgresPassword=xxxxx --set auth.password=yyyyyy --version 12.6.1

### What is the expected behavior?

The passwords in the pgsql-postgresql secret should correspond to their set values and should not be the same.

### What do you see instead?

Both passwords in the pgsql-postgresql secret are the same, even though different passwords were set with values for both users.

### Additional information

This bug seems to be caused by the commit https://github.com/bitnami/charts/commit/262f4f1d2906226cee8cabb49c8fedf97ecc30e4

ShiroDN:
Fixed by https://github.com/bitnami/charts/pull/17780

PR #17780, opened 2023-07-19
[bitnami/postgresl] Remove the password in text of the statefulset

Mauraza:
### Description of the change

The password of postgres appear in file o in a secret

### Applicable issues

- fixes #

### Checklist

### DIFF
--- a/bitnami/postgresql/templates/_helpers.tpl
+++ b/bitnami/postgresql/templates/_helpers.tpl
@@ -160,7 +160,7 @@ Return true if a secret object should be created
 */}}
 {{- define "postgresql.createSecret" -}}
 {{- $customUser := include "postgresql.username" . -}}
-{{- $postgresPassword := include "common.secrets.lookup" (dict "secret" (include "common.names.fullname" .) "key" .Values.auth.secretKeys.adminPasswordKey "defaultValue" (ternary (coalesce .Values.global.postgresql.auth.password .Values.auth.password .Values.global.postgresql.auth.postgresPassword .Values.auth.postgresPassword) (coalesce .Values.global.postgresql.auth.postgresPassword .Values.auth.postgresPassword) (or (empty $customUser) (eq $customUser "postgres"))) "context" $) -}}
+{{- $postgresPassword := include "common.secrets.lookup" (dict "secret" (include "common.names.fullname" .) "key" .Values.auth.secretKeys.adminPasswordKey "defaultValue" (ternary (coalesce .Values.global.postgresql.auth.postgresPassword .Values.auth.postgresPassword .Values.global.postgresql.auth.password .Values.auth.password) (coalesce .Values.global.postgresql.auth.postgresPassword .Values.auth.postgresPassword) (or (empty $customUser) (eq $customUser "postgres"))) "context" $) -}}
 {{- if and (not (or .Values.global.postgresql.auth.existingSecret .Values.auth.existingSecret))
     (or $postgresPassword .Values.auth.enablePostgresUser (and (not (empty $customUser)) (ne $customUser "postgres")) (eq .Values.architecture "replication") (and .Values.ldap.enabled (or .Values.ldap.bind_password .Values.ldap.bindpw))) -}}
     {{- true -}}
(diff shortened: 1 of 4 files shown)


migruiz4:
The password is also plaintext in lines 256-258, could you please remove that too?

We should definitely not be setting passwords using plaintext env vars.

migruiz4 (commented):
The password is also plaintext in lines 256-258, could you please remove that too?

We should definitely not be setting passwords using plaintext env vars.

migruiz4:
I added a commit fixing an issue with the `postgres-password` value.

bitnami/postgresql/templates/_helpers.tpl
```diff
@@ -160,7 +160,7 @@ Return true if a secret object should be created
 */}}
 {{- define "postgresql.createSecret" -}}
 {{- $customUser := include "postgresql.username" . -}}
-{{- $postgresPassword := include "common.secrets.lookup" (dict "secret" (include "common.names.fullname" .) "key" .Values.auth.secretKeys.adminPasswordKey "defaultValue" (ternary (coalesce .Values.global.postgresql.auth.postgresPassword .Values.auth.postgresPassword) (coalesce .Values.global.postgresql.auth.password .Values.auth.password) (or (empty $customUser) (eq $customUser "postgres"))) "context" $) -}}
+{{- $postgresPassword := include "common.secrets.lookup" (dict "secret" (include "common.names.fullname" .) "key" .Values.auth.secretKeys.adminPasswordKey "defaultValue" (ternary (coalesce .Values.global.postgresql.auth.password .Values.auth.password .Values.global.postgresql.auth.postgresPassword .Values.auth.postgresPassword) (coalesce .Values.global.postgresql.auth.postgresPassword .Values.auth.postgresPassword) (or (empty $customUser) (eq $customUser "postgres"))) "context" $) -}}
```
jotamartos, line 163: I do not understand why we are changing this. 

- If customUser is empty or if it's postgres, we should use .Values.global.postgresql.auth.postgresPassword or .Values.auth.postgresPassword that are the parameters for that
- If the user is using a custom user, he should be setting a password in the correct parameter: .Values.global.postgresql.auth.password .Values.auth.password

jotamartos (changes requested):

bitnami/postgresql/templates/_helpers.tpl
migruiz4: Let me explain in detail:

`$postgresPassword` stores the password of the 'postgres' user.

We want this value to be set to:
- If 'username' is empty or equals 'postgres', the password for Postgres user can be set using any of the following, in this order: `.Values.global.postgresql.auth.password`, `.Values.auth.password`, `.Values.global.postgresql.auth.postgresPassword`, `.Values.auth.postgresPassword`.
- If 'username' is not empty and not equals 'postgres' (in other words, it is set), the password for Postgres user can be set using any of the following, in this order: `.Values.global.postgresql.auth.postgresPassword`, `.Values.auth.postgresPassword`.

The currently logic prior to this PR is completely wrong, were if `auth.username` is set and password `auth.password` is provided, the Postgres user will also use that password.

jotamartos: In that case, let's update the values/README to because we say that the parameter is for the custom user to create 

https://github.com/bitnami/charts/blob/main/bitnami/postgresql/values.yaml#L23

migruiz4: I don't think it is needed, as the current description still suits the behavior, this is simply to allow the following use cases:
- ```yaml
  auth:
    enablePostgresUser: true
    postgresPassword: "my_pass"
    username: ""
    password: ""
  ```
- ```yaml
  auth:
    enablePostgresUser: true
    postgresPassword: ""
    username: "postgres"
    password: "my_pass"
  ```

It simply, to avoid an even more complex logic, also allows the following settings which may feel a bit weird:
- ```yaml
  auth:
    enablePostgresUser: true
    postgresPassword: ""
    username: ""
    password: "my_pass"
  ```
- ```yaml
  auth:
    enablePostgresUser: true
    postgresPassword: "my_pass"
    username: "postgres"
    password: ""
   ```

Please notice this logic is inherited from the Postgresql container logic which also allows this behavior: https://github.com/bitnami/containers/blob/8384822f1fcaea2991ad5e64849a046c7d9671ae/bitnami/postgresql/15/debian-11/rootfs/opt/bitnami/scripts/libpostgresql.sh#L641-L648

jotamartos (approved):

ShiroDN:
This fixes an issue https://github.com/bitnami/charts/issues/17727

PR #17300, opened 2023-06-22
Add copyright header

fevisera:
### Description of the change

Add copyright header to files in the project.

### Applicable issues

- fixes #

### Checklist

- [ ] ~Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/). This is *not necessary* when the changes only affect README.md files.~
- [ ] ~Variables are documented in the values.yaml and added to the `README.md` using [readme-generator-for-helm](https://github.com/bitnami-labs/readme-generator-for-helm)~
- [ ] ~Title of the pull request follows this pattern [bitnami/<name_of_the_chart>] Descriptive title~
- [x] All commits signed off and in agreement of [Developer Certificate of Origin (DCO)](https://github.com/bitnami/charts/blob/main/CONTRIBUTING.md#sign-your-work)

PR #17816, opened 2023-07-21
[bitnami/mongodb] Add trivial backup option

### LINKED ISSUE #17394: [bitnami/mongodb] backup database
### Name and Version

bitnami/mongodb 13.15.2

Disaster recovery

Some sort of backup script/tooling integrated into the helm chart

Some sort of roll my own

jcpunk:
### Description of the change

Sets up an optional cronjob to run `mongodump` for trivial backups

### Benefits

Users will have a quick way to enable database backups

### Possible drawbacks

Mongodump has a few limitations that make it not ideal in all scenarios.

### Applicable issues

- fixes #17394

### Additional information

This is in theory a workable middle ground between no backups and a complex snapshot based system.

### Checklist


carrodher:
Thanks for your contribution. There are a couple of checks/verifications failing, could you please take a look?

- In the first case it's needed to add the metadata for the newly added parameters, see https://github.com/bitnami/charts/actions/runs/5624341162/job/15240965472?pr=17816
- In the second case, it's needed to add the license header to the new files, see https://github.com/bitnami/charts/actions/runs/5624341156/job/15240965428?pr=17816. You can use any [existing template](https://github.com/bitnami/charts/blob/main/bitnami/mongodb/templates/common-scripts-cm.yaml) as an example.

jcpunk:
In theory I've addressed the errors.

javsalgar:
Hi,

These values are not allowed

```
    ## @param backup.cronjob.startingDeadlineSeconds Set the cronjob parameter startingDeadlineSeconds
    startingDeadlineSeconds: NULL
    ## @param backup.cronjob.ttlSecondsAfterFinished Set the cronjob parameter ttlSecondsAfterFinished
    ttlSecondsAfterFinished: NULL
```

You would need to set up a value

jcpunk:
In theory I've got this sorted out.

bitnami/mongodb/values.yaml
```diff
@@ -1064,6 +1064,100 @@ persistence:
...
+  enabled: false
+  cronjob:
+    ## @param backup.cronjob.schedule Set the cronjob parameter schedule
+    schedule: "@daily"
+    ## @param backup.cronjob.concurrencyPolicy Set the cronjob parameter concurrencyPolicy
+    concurrencyPolicy: Allow
+    ## @param backup.cronjob.failedJobsHistoryLimit Set the cronjob parameter failedJobsHistoryLimit
+    failedJobsHistoryLimit: 1
+    ## @param backup.cronjob.successfulJobsHistoryLimit Set the cronjob parameter successfulJobsHistoryLimit
+    SuccessfulJobsHistoryLimit: 3
```
FraPazGal, line 1083: The [CI/CD] Update README metadata workflow failed here, this should be formatted using lowerCamelCase.

bitnami/mongodb/templates/backup/cronjob.yaml
FraPazGal, line 1: Could you please add the related Copyright header to the new templates as done in https://github.com/bitnami/charts/pull/17300?

```
{{- /*
Copyright VMware, Inc.
*/}}
```

bitnami/mongodb/values.yaml
jcpunk: In theory this is sorted now.

bitnami/mongodb/templates/backup/cronjob.yaml
jcpunk: In theory this is sorted now.

bitnami/mongodb/templates/backup/pvc.yaml
```diff
@@ -0,0 +1,41 @@
...
+  annotations:
+    {{- if .Values.backup.cronjob.annotations }}
+    {{- include "common.tplvalues.render" ( dict "value" .Values.backup.cronjob.annotations "context" $) | nindent 4 }}
+    {{- end }}
+    {{- if .Values.commonAnnotations }}
+    {{- include "common.tplvalues.render" ( dict "value" .Values.commonAnnotations "context" $ ) | nindent 4 }}
+    {{- end }}
+    {{- end }}
+    {{- if .Values.backup.cronjob.storage.resourcePolicy }}
+    helm.sh/resource-policy: {{ .Values.backup.cronjob.storage.resourcePolicy | quote }}
```
FraPazGal, line 30: I think there is a typo here and the `{{- end }}` should be after this `{{ if }}`:

```suggestion
    {{- if .Values.backup.cronjob.storage.resourcePolicy }}
    helm.sh/resource-policy: {{ .Values.backup.cronjob.storage.resourcePolicy | quote }}
    {{- end }}
```

bitnami/mongodb/templates/backup/pvc.yaml
```diff
@@ -0,0 +1,41 @@
...
+    app.kubernetes.io/component: mongodump
+    {{- if .Values.backup.cronjob.labels }}
+    {{- include "common.tplvalues.render" (dict "value" .Values.backup.cronjob.labels "context" $) | nindent 4 }}
+    {{- end }}
+    {{- if .Values.commonLabels }}
+    {{- include "common.tplvalues.render" ( dict "value" .Values.commonLabels "context" $ ) | nindent 4 }}
+    {{- end }}
+  {{- if or .Values.backup.cronjob.annotations .Values.commonAnnotations .Values.backup.cronjob.storage.resourcePolicy}}
+  annotations:
+    {{- if .Values.backup.cronjob.annotations }}
```
FraPazGal, line 22: This should be `.Values.backup.cronjob.storage.annotations` right?

bitnami/mongodb/templates/backup/pvc.yaml
```diff
@@ -0,0 +1,41 @@
...
+  {{- end }}
+spec:
+  accessModes:
+  {{- range .Values.backup.cronjob.storage.accessModes }}
+    - {{ . | quote }}
+  {{- end }}
+  resources:
+    requests:
+      storage: {{ .Values.backup.cronjob.storage.size | quote }}
+  {{ include "common.storage.class" (dict "persistence" .Values.backup.cronjob.storage "global" .Values.global) }}
```
FraPazGal, line 40:
```suggestion
  {{ include "common.storage.class" (dict "persistence" .Values.backup.cronjob.storage "global" .Values.global) | nindent 2 }}
```

bitnami/mongodb/values.yaml
```diff
@@ -1064,6 +1064,100 @@ persistence:
...
+      seccompProfile:
+        type: RuntimeDefault
+      capabilities:
+        drop:
+          - ALL
+    ## @param backup.cronjob.command Set backup container's command to run
+    command: 
+      - /bin/sh
+      - -c
+      - "mongodump --username=${MONGODB_ROOT_USER} --password=${MONGODB_ROOT_PASSWORD} --host=${MONGODB_SERVICE_NAME} --port=${MONGODB_PORT_NUMBER} ${MONGODB_CLIENT_EXTRA_FLAGS} --oplog --gzip --archive=${MONGODUMP_DIR}/mongodump-$(date '+%Y-%m-%d-%H-%M').gz"
```
FraPazGal, line 1114: I would move this to the job's template. You can leave this param as an empty array and add the related logic to use  the default command only when `.Values.backup.cronjob.command` is not set. You can check a similar template in our `bitnami/keycloak` chart:

https://github.com/bitnami/charts/blob/494f4e88ed2e6098aa15a07b24e40d7882495e0f/bitnami/keycloak/templates/keycloak-config-cli-job.yaml#L65-L75

bitnami/mongodb/values.yaml
```diff
@@ -1064,6 +1064,100 @@ persistence:
     ##
     dataSource: {}
 
+## @section Backup parameters
+## This section implements a trivial logical dump cronjob of the database.
+## This only comes with the consistency guarantees of the dump program.
+## This is not a snapshot based roll forward/backward recovery backup.
+## ref: https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/
+backup:
```
FraPazGal, line 1072: This is a style comment, but could you add padding to the new parameters as done for the rest of the file? This helps increase readability. E.g.:

```suggestion
## ref: https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/
##
backup:
```

FraPazGal (commented):
Really nice PR @jcpunk! I have added a few comments as a review, please check them whenever you can.

bitnami/mongodb/templates/backup/pvc.yaml
jcpunk: I've got it at the same level as `schedule`.  I could move it under `storage` if you think that would make more sense.

bitnami/mongodb/values.yaml
jcpunk: I was thinking having it in the `values.yaml` would help by having it exposed in the README so folks could easily see what it is by default, but I can move it easily enough.

bitnami/mongodb/templates/backup/pvc.yaml
FraPazGal: I'm not sure I understood you. In my previous comment I was asking because you added `.Values.backup.cronjob.annotations` to the cronjob template (that's fine) but also here, in the pvc template. Unless I'm mistaken, you are not using `.Values.backup.cronjob.storage.annotations` anywhere in the templates so this code should be:

```suggestion
    {{- if .Values.backup.cronjob.storage.annotations }}
```

bitnami/mongodb/values.yaml
FraPazGal: It's true that is leaving it in the values helps readability but it also clusters the `values.yaml` file. At the same time, this is the go to schema we usually use so users are already used to it.

Adding the option to override the default command as I explained in my previous comment should also help to make users aware of this feature.

jcpunk: Works for me, in theory I've sorted out your requested changes.  In practice...

jcpunk:
Just checking back in on this.

FraPazGal:
Sorry about that @jcpunk, I was checking on my side everything looked good.

Could you check the PR? It seems your last forced push removed the cronjob template from it. Besides, the PRs seems about ready once you recover the template file.

jcpunk:
Crap, I don't seem to have a copy of the templates.  I don't suppose you've got a copy?

jcpunk:
In theory I'm rebased and have the templates all loaded.

FraPazGal (approved):
LGTM!

PR #17877, opened 2023-07-25
[bitnami/dokuwiki] Testing changes in the Dokuwiki README and the validations in the pipeline

jotamartos:
Testing purposes


jotamartos:
Just confirmed that the commit is verified with the new configuration of the laptop.

PR #18010, opened 2023-07-28
[bitnami/kong] Adapt Kong tests to Air-gapped environments

jotamartos:
### Description of the change

Some tests do not work in air-gapped environments.

### Benefits

We are modifying the tests to make them work in all environments. This way we can verify that the solution works as expected. If the solution doesn't work in that environment, the tests will not be changed and support for that env will be removed.

### Applicable issues

- fixes #

### Checklist

### DIFF
--- a/.vib/kong/cypress/cypress/integration/kong_spec.js
+++ b/.vib/kong/cypress/cypress/integration/kong_spec.js
@@ -51,6 +51,14 @@ it('allows adding a new service and route through admin API', () => {
       expect(response.status).to.eq(201);
     });
 
+    cy.request({
+      method: 'GET',
+      url: `${BASE_URL}/services`,
+    }).then((response) => {
+      expect(response.status).to.eq(200);
+      expect(JSON.stringify(response.body)).to.contain(services.newService.upstreamURL.replace(/^https?:\/\//, ''));
+    });
+
     cy.fixture('routes').then((routes) => {
       cy.request({
         method: 'POST',
@@ -68,10 +76,10 @@ it('allows adding a new service and route through admin API', () => {
       cy.wait(7000);
       cy.request({
         method: 'GET',
-        url: `${routes.newRoute.path}${random}`,
+        url: `${BASE_URL}/services/${services.newService.name}-${random}/routes`,
       }).then((response) => {
         expect(response.status).to.eq(200);
-        expect(response.body).to.contain(services.newService.upstreamContent);
+        expect(JSON.stringify(response.body)).to.contain(`${routes.newRoute.path}${random}`);
       });
     });
   });
--- a/.vib/kong/cypress/cypress/fixtures/services.json
+++ b/.vib/kong/cypress/cypress/fixtures/services.json
@@ -1,7 +1,6 @@
 {
   "newService": {
     "name": "cypress-service",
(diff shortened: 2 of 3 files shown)


.vib/kong/cypress/cypress/integration/kong_spec.js
```diff
@@ -51,6 +51,14 @@ it('allows adding a new service and route through admin API', () => {
       expect(response.status).to.eq(201);
     });
 
+    cy.request({
+      method: 'GET',
+      url: `${BASE_URL}/services`,
```
FraPazGal, line 56: IIRC, we shouldn't need to add `${BASE_URL}` here, as Cypress already does so by default.

.vib/kong/cypress/cypress/integration/kong_spec.js
```diff
@@ -51,6 +51,14 @@ it('allows adding a new service and route through admin API', () => {
...
+    }).then((response) => {
+      expect(response.status).to.eq(200);
+      expect(JSON.stringify(response.body)).to.contain(services.newService.upstreamURL.replace(/^https?:\/\//, ''));
```
FraPazGal, line 59: Not really an issue but just a FYI, I'll say there is no need to use `JSON.stringify()` for the response's body.

.vib/kong/cypress/cypress/integration/kong_spec.js
```diff
@@ -68,10 +76,10 @@ it('allows adding a new service and route through admin API', () => {
       cy.wait(7000);
       cy.request({
         method: 'GET',
-        url: `${routes.newRoute.path}${random}`,
+        url: `${BASE_URL}/services/${services.newService.name}-${random}/routes`,
       }).then((response) => {
         expect(response.status).to.eq(200);
-        expect(response.body).to.contain(services.newService.upstreamContent);
+        expect(JSON.stringify(response.body)).to.contain(`${routes.newRoute.path}${random}`);
```
FraPazGal, line 82: We should remove `services.newService.upstreamContent` from the `services.json` file if we no longer use it.

FraPazGal (commented):
Thanks for the PR @jotamartos, I have added a few comments as a review. Please check them whenever you can

.vib/kong/cypress/cypress/integration/kong_spec.js
jotamartos: The response is an object and we need to make it a string for this to work.

.vib/kong/cypress/cypress/integration/kong_spec.js
jotamartos: You are right, will do it!

.vib/kong/cypress/cypress/integration/kong_spec.js
jotamartos: There are 2 endpoints here, the proxy and the API. Will investigate why it's needed but if I'm not mistaken, it's necessary for the test to work.

jotamartos: BASE_URL is set some lines above to include the port of the API endpoint.

jotamartos:
Tests succeeded here: https://github.com/bitnami/charts/actions/runs/5691123284

PR #17833, opened 2023-07-23
[bitnami/external-dns] Fix missing rfc2136.secretName secret key documentation

karlivory:
### Description of the change

Documentation of key name was missing for rfc2136.secretName. The key in question:

https://github.com/bitnami/charts/blob/567f1fc92e6031b12f188550a1ab2d8bae4edc56/bitnami/external-dns/templates/dep-ds.yaml#L587C24-L587C43

### Benefits

Improves documentation.

### Checklist

- [x] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/). This is *not necessary* when the changes only affect README.md files.
- [x] Variables are documented in the values.yaml and added to the `README.md` using [readme-generator-for-helm](https://github.com/bitnami-labs/readme-generator-for-helm)
- [x] Title of the pull request follows this pattern [bitnami/<name_of_the_chart>] Descriptive title
- [x] All commits signed off and in agreement of [Developer Certificate of Origin (DCO)](https://github.com/bitnami/charts/blob/main/CONTRIBUTING.md#sign-your-work)

PR #18030, opened 2023-07-30
[bitnami/external-dns] tsig keyname as a variable

kirillvelikov:
### Description of the change

After defining a custom tsigKeyname value and manually generating the secret (with a different key) the pod fails to start because of incorrectly generated deployment resource. To be precise the variable `EXTERNAL_DNS_RFC2136_TSIG_SECRET` will have incorrect value.

Currently the key which is storing the value of `EXTERNAL_DNS_RFC2136_TSIG_SECRET` is hard-coded in the deployment template while the default values.yaml file already presents the option to change its value. And this value is used for generation of the arguments list, but not for the environment variable.
The result is that the deployment is looking for non-existing key in the specified secret and cannot start.

This commits is replacing the default value currently present in values.yaml (in order not to change the secret template) and is sourcing it in the in deployment template.
Now users can create manually the secret, change the keyname for it and the pod will be able to start.

### Benefits

With the proposed change it is possible for the pod to start after using a custom tsigKeyname value. 

### Applicable issues

- fixes #

### Additional information

Here is an example of values used for testing:
```yaml
provider: rfc2136
domainFilters:
  - <my.domain>
rfc2136:
  host: <ca.my.domain>
  zone: <my.domain>
  tsigKeyname: <custom-value>
  secretName: <secret-name-created-manually>
```

### Checklist


bitnami/external-dns/templates/dep-ds.yaml
```diff
@@ -587,7 +587,7 @@ spec:
               valueFrom:
                 secretKeyRef:
                   name: {{ template "external-dns.secretName" . }}
-                  key: rfc2136_tsig_secret
+                  key: {{ .Values.rfc2136.tsigKeyname }}
```
trunet, line 590: why is this here? the secret.yaml template never set a key named the value of tsigKeyname.

NEOhidra: If the secret is created externally, not by the chart, and the key is named differently.

trunet: this is what I meant: https://github.com/bitnami/charts/pull/17833

trunet: this is what I meant: https://github.com/bitnami/charts/pull/18768/files#diff-ee55b5af91bd556b59c1dc234a5d17aae2f02e63287971530dc4b752a7b48ad0R593

kirillvelikov: According to the description
>   ## rfc2136.secretName When using the rfc2136 provider, specify the existing secret which contains your tsig secret in the key "rfc2136_tsig_secret". Disables the usage of `rfc2136.tsigSecret` (optional)

"secretName" should be specified if the secret exist. This commit allows to mount such secret with a custom name of the key which holds the tsig secret.

Here is an example "values.yaml" for a cluster in which there is already a secret present. And this secret contains a the  tsig secret in a key not named "rfc2136_tsig_secret".

```yaml
# values.yaml
rfc2136:
  tsigKeyname: <custom-key-name-not-rfc2136_tsig_secret>
  secretName: <secret-created-earlier>
```
With such values.yaml the chart will not generate a new secret - there is no need.

Reverting this change will produce a deployment like this:
```yaml
# Source: external-dns/templates/dep-ds.yaml
apiVersion: apps/v1
kind: Deployment
metadata:
spec:
  template:
    spec:
      containers:
        - name: external-dns
          args:
            - --rfc2136-tsig-keyname=<custom-key-name-not-rfc2136_tsig_secret>
          env:
            # RFC 2136 environment variables
            - name: EXTERNAL_DNS_RFC2136_TSIG_SECRET
              valueFrom:
                secretKeyRef:
                  name: <secret-created-earlier>
                  key: rfc2136_tsig_secret
```

Such deployment will not start because there is no key named "rfc2136_tsig_secret" inside secret "\<secret-created-earlier>".

I see no issue in this commit.

If you decide to specify a "rfc2136.tsigKeyname" in a combination with the tsig secret ("rfc2136.tsigSecret") - then you are creating an issue!
> ## rfc2136.tsigSecret When using the rfc2136 provider, specify the tsig secret to enable security. (do not specify if rfc2136.secretName is provided.) (optional)

Because the helm chart will generate a secret, when "rfc2136.tsigSecret" is specified, but will not use the name of the key in the deployment:

```yaml
# values.yaml
rfc2136:
  host: ipa-ca.iam.geek # replace with your host
  zone: noi.geek # replace with your zone
  tsigSecret: "<the-tsig-secret>"
  tsigKeyname: "<custom-key-name-not-rfc2136_tsig_secret>"
---
# Source: external-dns/templates/secret.yaml
apiVersion: v1
kind: Secret
metadata:
  name: testing-external-dns
type: Opaque
data:
  rfc2136_tsig_secret: "aXRJU0hFUkVFRUVF"
---
# Source: external-dns/templates/dep-ds.yaml
apiVersion: apps/v1
kind: Deployment
metadata:
spec:
  template:
    spec:
      containers:
        - name: external-dns
          args: 
            - --rfc2136-tsig-keyname=<custom-key-name-not-rfc2136_tsig_secret>
          env:
            # RFC 2136 environment variables
            - name: EXTERNAL_DNS_RFC2136_TSIG_SECRET
              valueFrom:
                secretKeyRef:
                  name: testing-external-dns
                  key: rfc2136_tsig_secret
```

Reverting this commit will break "rfc2136.secretName". 

Is there any other case in which you will get "CreateContainerConfigError" because of such mismatch?

PR #18059, opened 2023-07-31
[bitnami/external-dns] Support pod restarts when using gitops

### LINKED ISSUE #17562: [bitnami/external-dns] do not set pod annotation to null
### Name and Version

bitnami/external-dns

amd64

We deploy External-dns with ArgoCD and have autosync enabled.
When doing a restart of the external-dns deployment kubectl add the `kubectl.kubernetes.io/restartedAt` annotation to the pod spec. When the pod has pod spec has no other annotations this field should not be included.

The chart currently sets the pod spec annotation to null when there is no addition annotation given while it should leave the field empty.

1. Deploy external-dns via Argocd with autosync enabled
2. Kubectl rolllout restart deployement external-dns
3. New replicaset and pods will be remove instantly

_No response_

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: externaldns-external-dns
  namespace: externaldns
spec:
  template:
    metadata:
      annotations:
        kubectl.kubernetes.io/restartedAt: ""
```
### What do you see instead?
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: externaldns-external-dns
  namespace: externaldns
spec:
  template:
    metadata:
      annotations: null
```

rouke-broersma:
### Description of the change

Add an if around the pod spec annotations in dep-ds so when there are no annotations the key is not placed.

### Benefits

When doing a restart of the external-dns deployment kubectl add the kubectl.kubernetes.io/restartedAt annotation to the pod spec. The chart sets the annotations to null when there are no other annotations. When using gitops the gitops tooling will enforce the chart configuration so it will reapply the null, cancelling the restart.

### Applicable issues

- fixes #17562

### Checklist


bitnami/external-dns/templates/dep-ds.yaml
```diff
@@ -31,6 +31,7 @@ spec:
   template:
     metadata:
       labels: {{ include "external-dns.labels" . | nindent 8 }}
+      {{- if or .Values.podAnnotations (and .Values.metrics.enabled .Values.metrics.podAnnotations) (include "external-dns.createSecret" .) (and (eq .Values.provider "designate") .Values.designate.customCA.enabled) }}
       annotations:
         {{- if or .Values.podAnnotations .Values.metrics.enabled }}
```
migruiz4, line 36: Thank you for your contribution! Could you please update this line to match the conditional you added?
```suggestion
        {{- if or .Values.podAnnotations (and .Values.metrics.enabled .Values.metrics.podAnnotations) }}
```

rouke-broersma: I did not add it because effectively the result will be the same due to the condition already being in the template helper, but it also doesn't hurt putting it here.

rouke-broersma:
@migruiz4 @carrodher let me know if you need any other changes

migruiz4 (approved):
Looks good to me!

PR #18060, opened 2023-07-31
[bitnami/solr] [bitnami/spark] Add noprompt into keytool commands to avoid init crash on pod resume

rajan123456:
### Description of the change

This PR fixes issues in the initContainer script for the solr and spark helm charts. By adding the noprompt option, the keytool command is able to successfully complete even if the container is being resumed and the certificate is already added to the keystore, instead of erroring out and expecting interactive user input.

### Benefits

With this change, the spark-master, spark-worker and solr statefulset pods will be able to successfully startup even if the init script has been executed on the filesystem before.

### Possible drawbacks

NA

### Applicable issues

NA

### Additional information

NA

### Checklist


carrodher:
Thanks for your contribution, please note in order to properly test the Helm charts it's needed to send a PR per chart. Could you please split this PR into two different ones?

Issue #18061, opened 2023-07-31
[bitnami/rabbitmq] Secret name for external load-definition did not consider extraSecretsPrependReleaseName value


DanielFran:
### Name and Version

bitnami/rabbitmq

### What architecture are you using?

None

### What steps will reproduce the bug?

1. Define use of external load-definition
2. Define name of existing secret
3. Define extraSecretsPrependReleaseName = true
4. Confirm that secret name is different than the one defined in statefulset volumes

### What is the expected behavior?

The secret name in statefullset volume should be the same as the secret name created, i.e. the secret name prepend the Release name

### What do you see instead?

The secret name in statefullset volume did not prepend the Release name

carrodher:
Thanks for creating this issue and the associated PR. The team will review it and provide feedback. Once merged the PR, this issue will be automatically closed.

DanielFran:
See comments on https://github.com/bitnami/charts/pull/18062

PR #18062, opened 2023-07-31
[bitnami/rabbitmq] Load definition volume secret name should consider prepend release name

DanielFran:
### Applicable issues

- fixes https://github.com/bitnami/charts/issues/18061

### Checklist

### DIFF
--- a/bitnami/rabbitmq/Chart.yaml
+++ b/bitnami/rabbitmq/Chart.yaml
@@ -25,4 +25,4 @@ maintainers:
 name: rabbitmq
 sources:
   - https://github.com/bitnami/charts/tree/main/bitnami/rabbitmq
-version: 12.0.9
+version: 12.0.10
--- a/bitnami/rabbitmq/templates/statefulset.yaml
+++ b/bitnami/rabbitmq/templates/statefulset.yaml
@@ -414,7 +414,7 @@ spec:
         {{- if .Values.loadDefinition.enabled }}
         - name: load-definition-volume
           secret:
-            secretName: {{ ternary (printf "%s-%s" $.Release.Name $key) $key $.Values.extraSecretsPrependReleaseName }}
+            secretName: {{ ternary (printf "%s-%s" .Release.Name .Values.loadDefinition.existingSecret) .Values.loadDefinition.existingSecret .Values.extraSecretsPrependReleaseName }}
         {{- end }}
         {{- if .Values.initScripts }}
         - name: custom-init-scripts
--- a/bitnami/rabbitmq/templates/statefulset.yaml
+++ b/bitnami/rabbitmq/templates/statefulset.yaml
@@ -414,7 +414,7 @@ spec:
         {{- if .Values.loadDefinition.enabled }}
         - name: load-definition-volume
           secret:
-            secretName: {{ tpl .Values.loadDefinition.existingSecret . | quote }}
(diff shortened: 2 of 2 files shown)


DanielFran:
@fmulero The seccret name created [here](https://github.com/bitnami/charts/blob/main/bitnami/rabbitmq/templates/secrets.yaml#L38) is not matching the name in the volume.

This PR fix it.

fmulero:
Hi Daniel,

In the pictures you shared, you can see how the value `extraSecretsPrependReleaseName` works with `extraSecrets` not with `loadDefinition.existingSecret` value. Don't get me wrong but I think you are mixing both values. Please correct if I am wrong sharing an scenario to test it.

DanielFran:
> Hi Daniel,
> 
> In the pictures you shared, you can see how the value `extraSecretsPrependReleaseName` works with `extraSecrets` not with `loadDefinition.existingSecret` value. Don't get me wrong but I think you are mixing both values. Please correct if I am wrong sharing an scenario to test it.

@fmulero No problem.
To instantiate a rabbitmq with a load definition, we should use load-definition properties, and in my case I want that the secret prepend with the instance name (example: rabbitmq-example).

So the config to use is:
- @param loadDefinition.enabled Enable loading a RabbitMQ definitions file to configure RabbitMQ
loadDefinition.enabled: true
- @param loadDefinition.existingSecret Existing secret with the load definitions file
loadDefinition.existingSecret: "load-definition"
- @param extraSecrets Optionally specify extra secrets to be created by the chart. This can be useful when combined with load_definitions to automatically create the secret containing the definitions to be loaded.
````yaml
extraSecrets:
  load-definition:
    load_definition.json: |
      {
        ...
      }
````
- @param extraSecretsPrependReleaseName Set this flag to true if extraSecrets should be created with <release-name> prepended.
extraSecretsPrependReleaseName: true

[secrets.yaml](https://github.com/bitnami/charts/blob/main/bitnami/rabbitmq/templates/secrets.yaml#L33-L49) will create an a secret example like:

````yaml
# Source: rabbitmq-example/templates/secrets.yaml
apiVersion: v1
kind: Secret
metadata:
  name: rabbitmq-example-load-definition
  namespace: "dev"
  labels:
    app.kubernetes.io/name: rabbitmq-example
    helm.sh/chart: rabbitmq-example-12.0.7
    app.kubernetes.io/instance: rabbitmq-example
type: Opaque
stringData:
  load-definition.json: |-
      {
        ...
      }
````

[statefulset.yaml](https://github.com/bitnami/charts/blob/main/bitnami/rabbitmq/templates/statefulset.yaml#L414-L417) will create an a volume with wrong secret name:

````yaml
# Source: rabbitmq-example/templates/statefulset.yaml
apiVersion: apps/v1
kind: StatefulSet
metadata:
  name: rabbitmq-example
  namespace: "dev"
  labels:
    app.kubernetes.io/name: rabbitmq-example
    helm.sh/chart: rabbitmq-example-12.0.7
    app.kubernetes.io/instance: rabbitmq-example
spec:
  ...
      volumes:
        - name: load-definition-volume
          secret:
            secretName: load-definition
````

DanielFran:
@fmulero I realize that extraSecrets can be used to create the load-definition secret as per the readme information, but it is not mandatory, so the volume secret name can be handled using the templating.
Closing the PR

PR #18074, opened 2023-08-01
[bitnami/solr] Add noprompt to init certs command to avoid crash on resume

rajan123456:
### Description of the change

This PR fixes issues in the initContainer script for the solr helm chart. By adding the noprompt option, the keytool command is able to successfully complete even if the container is being resumed and the certificate is already added to the keystore, instead of erroring out and expecting interactive user input.

### Benefits

With this change, solr statefulset pods will be able to successfully startup even if the init script has been executed on the filesystem before.

### Checklist


juan131 (approved):
Thanks for the fix!! LGTM

PR #18075, opened 2023-08-01
[bitnami/spark] Add noprompt to init certs command to avoid crash on resume

rajan123456:
### Description of the change

This PR fixes issues in the initContainer script for the spark helm chart. By adding the noprompt option, the keytool command is able to successfully complete even if the container is being resumed and the certificate is already added to the keystore, instead of erroring out and expecting interactive user input.

### Benefits

With this change, spark statefulset pods will be able to successfully startup even if the init script has been executed on the filesystem before.

### Possible drawbacks

NA

### Applicable issues

NA

### Checklist

### DIFF
--- a/bitnami/spark/Chart.yaml
+++ b/bitnami/spark/Chart.yaml
@@ -24,4 +24,4 @@ maintainers:
 name: spark
 sources:
   - https://github.com/bitnami/charts/tree/main/bitnami/spark
-version: 7.1.2
+version: 7.1.3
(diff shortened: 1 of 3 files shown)


fmulero:
Hi @rajan123456, Thanks a lot for your contribution.

Everything looks good, but could you provide an example/scenario where we can see the problem you mention and how your PR can fix it? We have several charts with similar code and could be great if we can add a reference to the concrete issue.

rajan123456:
Hi @fmulero -

I can describe the scenario for which we face issues:

- Deployed the chart as a fresh installation, and everything works fine.
- Power outage causing the entire kubernetes cluster to be restarted.
- Spark (master/worker) pods are stuck in Init:CrashLoopBackOff with the error: the alias in keystore already exists, do you want to overwrite: [y/N] . . .

With this change, the pods can resume seamlessly. I am sure there may be a simpler method to reproduce this. Should you require the crash logs, I will have to rollback and reproduce the error again, please let me know should that be required.

Thanks!

fmulero (approved):
Great! LGTM

Issue #17863, opened 2023-07-25
 [bitnami/grafana-operator] New Grafana-Operator versions produces invalid `grafana` specs


luringens:
### Name and Version

bitnami/grafana-operator 5.2.0

### What architecture are you using?

None

### What steps will reproduce the bug?

```sh
helm template my-release oci://registry-1.docker.io/bitnamicharts/grafana-operator --version=3.1.1 -f values.yaml > test.yaml
kubectl create -f test.yaml 
```

### Are you using any custom parameters or values?

```yaml
grafana:
  labels:
    foo: bar
  updateStrategy:
    type: Recreate
  envFrom:
  - configMapRef:
      name: grafana-env
  persistence:
    enabled: true
    storageClass: fast
    size: 4Gi
  ingress:
    enabled: true
    hostname: "example.com"
    path: /
    tls: false
    annotations:
      kubernetes.io/ingress.class: kong
      konghq.com/protocols: https
```

### What is the expected behavior?

Templating to include:
- `Grafana.spec.persistentVolumeClaim`
- `Grafana.spec.deployment.spec.template.spec.containers[].envFrom`
- `Grafana.spec.ingress.metadata.annotations`
- `Grafana.spec.ingress.metadata.labels`
- `Grafana.spec.ingress.spec.tls[].secretName`

Expected behaviour is that the Grafana yaml should be a valid `grafanas` as [specified in the CRD](https://grafana-operator.github.io/grafana-operator/docs/api/).

### What do you see instead?

Templating instead produces:
- `Grafana.spec.dataStorage`
- `Grafana.spec.deployment.envFrom`
- `Grafana.spec.ingress.annotations`
- `Grafana.spec.ingress.labels`
- `Grafana.spec.ingress.spec.tlsEnabled`
- `Grafana.spec.ingress.spec.tlsSecretName`

Which are all invalid `org.integreatly.grafana.v1beta1.Grafana` specs, and produce the following error when applied to Kubernetes (newlines added for readability):
```
error: error validating "testgrafana.yaml": error validating data: [
ValidationError(Grafana.spec): unknown field "dataStorage" in org.integreatly.grafana.v1beta1.Grafana.spec, 
ValidationError(Grafana.spec.deployment): unknown field "envFrom" in org.integreatly.grafana.v1beta1.Grafana.spec.deployment, 
ValidationError(Grafana.spec.deployment): unknown field "strategy" in org.integreatly.grafana.v1beta1.Grafana.spec.deployment,
[...]

PR #18122, opened 2023-08-01
[grafana-operator] fix: move sidecar definition to deployment section

arnecls:
### Description of the change

With the latest version of the operator, the sidecar container section has been removed.
As of this, all additional containers have to be placed into the `deployment.spec.template.spec.containers` section.

The readiness and liveness probe definition also had a couple of issue:

- It was put at the end of a section that actually is an array (rendering error)
- It was not indented correctly, and even if it was, it would only affect the last sidecar
- The main container was not affected by the probes, but the location in the value file somewhat implied that

This PR moves the probe to the main grafana container, as the sidecars can (and should) define this section inline.

### Benefits

- The resource is correctly rendered when using sidecars
- Readiness and Liveness probe can be configured for the grafana instance

### Possible drawbacks

There should be none

### Applicable issues

- relates to #17863

### Checklist


arnecls:
I added another fix I found while rolling out these changes (note: this fix alone does not make the chart work).
The liveliness and readiness probe need to be "complete" now, as the CRD now expects a proper deployment definition.
There's other places in the chart where this is also an issue, but this is out of scope for this specific fix.

juan131 (dismissed):
Thanks for the fix! LGTM

juan131:
@arnecls we merged another PR affecting the Grafana Operator, could you please merge `main` in your branch, resolve the conflicts and update the PR? Thanks in advance.

arnecls:
@juan131 merged-in the latest changes

PR #18137, opened 2023-08-02
[bitnami/kubernetes-event-exporter] Allow changing the default revisionHistoryLimit

barney71:
### Description of the change

I want to clean up old replica sets so that these don't appear in metrics, OPA vioalations, ArgoCD UI and so on.
Therefore I want to overwrite the Kubernetes default revision history limit.

See [Kubernetes Clean up Policy](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#clean-up-policy)

### Benefits

Old replica sets are deleted if not needed.

### Possible drawbacks

-

### Applicable issues

- 

### Additional information

-

### Checklist


bitnami/kubernetes-event-exporter/templates/deployment.yaml
```diff
@@ -18,6 +18,9 @@ metadata:
     {{- end }}
 spec:
   replicas: {{ .Values.replicaCount }}
+  {{- if .Values.revisionHistoryLimit }}
+  revisionHistoryLimit: {{ .Values.revisionHistoryLimit }}
+  {{- end }}
```
fevisera, line 23: Is that if-block really needed?

barney71: Hi @fevisera, 
thanks for your review. :-)

You are right. It isn't really needed for if the value is not present it would fallback to the default of 10. Only if one wants to overwrite this default behavior, the value must be set.

So i will remove the if-block.

fevisera (approved):
Thank you for your contribution and for applying the changes!
LGTM!

PR #18194, opened 2023-08-03
[bitnami/common] Add app.kubernetes.io/version based on AppVersion

jw-maynard:
### Description of the change

Adds the recommended label app.kubernetes.io/version to the Common chart

### Benefits

app.kubernetes.io/version is a recommended label and allows you to query objects in k8s based on the app version. Also good for telemetry data also.

### Possible drawbacks

I suppose if people wish to hid the app version currently deployed this will not allow for that but I suppose a bigger PR could be made to allow this to be turned on and off.

### Applicable issues

- fixes NONE

### Checklist

### DIFF
--- a/bitnami/common/Chart.yaml
+++ b/bitnami/common/Chart.yaml
@@ -6,7 +6,7 @@ annotations:
   licenses: Apache-2.0
 apiVersion: v2
 # Please make sure that version and appVersion are always the same.
-appVersion: 2.6.0
+appVersion: 2.7.0
 description: A Library Helm Chart for grouping common logic between bitnami charts. This chart is not deployable by itself.
 home: https://bitnami.com
 icon: https://bitnami.com/downloads/logos/bitnami-mark.png
@@ -23,4 +23,4 @@ name: common
 sources:
   - https://github.com/bitnami/charts
 type: library
-version: 2.6.0
+version: 2.7.0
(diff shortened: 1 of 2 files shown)


corico44 (approved):
LGTM

corico44:
@jw-maynard After analyzing this change more in depth we have decided to revert it, the explanation of why is the following:

> The `app.kubernetes.io/version` label must be treated with caution given its non-inmutable values (Helm will set a different value to the `appVersion` changes on `Chart.yaml`).
> There are fields on K8s objects that are inmutable (e.g. `deployment.spec.selector` or `statefulset.spec.volumeClaimTemplates.metadata.labels`). Therefore, we might be very cautious to avoid including these kind of labels to avoid introducing backwards incompatibilities.
> Before applying a changes such as the one introduced at https://github.com/bitnami/charts/pull/18194, we should previously ensure we're not making a bad use of `common.labels.matchLabels` helper **on every single chart**. Without doing this study, we shouldn't include it.

jw-maynard:
@corico44 We use a few of the charts in this repo and I'd like to get version added to them. I went for the common chart because it seemed like the place where you guys would prefer some sort of "common" tag that would reply to all charts should be. What's the path forward from here? Can I submit some PRs to add this functionality to to the specific charts but that seems something that would just need to be removed down the line and kicks a can down the road. Does anyone have thoughts on this on the Bitnami side?

Issue #18172, opened 2023-08-03
Conflict between HPA and replicaCount setting in HAProxy chart


MuhammadTalhaPasha:
### Name and Version

bitnami/haproxy 0.8.5

### What architecture are you using?

None

### What steps will reproduce the bug?

1. Deployed bitnami/haproxy chart with the ArgoCD application, using the provided yaml configuration.
2. The HPA is not scaling the pods as expected.

### Are you using any custom parameters or values?

apiVersion: argoproj.io/v1alpha1
kind: Application
metadata:
  name: db-haproxy
  namespace: argocd
spec:
  project: default
  source:
    repoURL: https://charts.bitnami.com/bitnami
    chart: haproxy
    targetRevision: 0.8.5
    helm:
      values: |
         replicaCount: 1
         service:
          type: LoadBalancer
          annotations:
            cloud.google.com/load-balancer-type: "Internal"
          ports:
              - name: mysql
                protocol: TCP
                port: 3306
                targetPort: 3306
              - name: http
                protocol: TCP
                port: 80
                targetPort: 8080
         resources:
          requests:
            cpu: 1000m
            memory: 300Mi
          limits:
            cpu: 1000m
            memory: 300Mi
         autoscaling:
          enabled: false
          minReplicas: 1
          maxReplicas: 10
          targetCPU: 60
          targetMemory: null
         configuration: |
           frontend stats
              mode http
              bind *:8080
              stats enable
              stats uri /
              stats refresh 10s
              stats admin if TRUE
            frontend db_frontend
              bind :3306
              default_backend db_backend
              timeout client 1m
              log global
              option tcplog
            backend db_backend
              mode tcp
              balance leastconn
              timeout server 1m
              log global
              option tcplog
              timeout connect 5s
              option tcp-check
              server db_read_replica_1 10.200.0.77:3306 check
  destination:
    server: https://kubernetes.default.svc
    namespace: ecs
  syncPolicy:
    automated: 
      prune: true
      selfHeal: true

### What is the expected behavior?

The HPA should scale the number of pods between 1 and 10 based on the CPU utilization.

### What do you see instead?

The HPA does not seem to be scaling the pods as expected.

### Additional information

This issue might be related to the replica count being hardcoded in the deployment template in this line: replicas: {{ .Values.replicaCount }}. Usually, in our charts, we have a condition to set that property only if autoscaling is disabled.

[...]

PR #18207, opened 2023-08-04
fix: allow HPA to manage replicas when autoscaling is enabled

MuhammadTalhaPasha:
### Description of the change
This change modifies the HAProxy chart's Deployment.yaml to conditionally set replicaCount based on whether autoscaling is enabled. This change allows the Horizontal Pod Autoscaler (HPA) to properly control the number of replicas when autoscaling is enabled and falls back to use the value from replicaCount when autoscaling is not enabled.

### Benefits
This fix allows users of the HAProxy chart to properly use Kubernetes HPA with the chart. This resolves an issue where the `replicaCount` setting in values.yaml would override the HPA's replica count, preventing proper autoscaling.

### Possible drawbacks
As this change is adding a condition to the setting of replicaCount, it should not introduce any drawbacks unless there is a dependency on the replicaCount field being explicitly set in the Deployment even when autoscaling is enabled.

### Applicable issues
- Fixes https://github.com/bitnami/charts/issues/18172

### Additional information
The conditional check for autoscaling has been added in the Deployment.yaml as follows:
```yaml
{{- if not .Values.autoscaling.enabled }}
{{- end }}
```

### Checklist:

-  Chart version bumped in Chart.yaml according to [semver](http://semver.org/).
-  Variables are documented in the values.yaml and added to the README.md using [readme-generator-for-helm](https://github.com/bitnami-labs/readme-generator-for-helm)
-  Title of the pull request follows this pattern [bitnami/haproxy] Conditionally set replicaCount
-  All commits signed off and in agreement of [Developer Certificate of Origin (DCO)]

### DIFF
--- a/bitnami/haproxy/Chart.yaml
+++ b/bitnami/haproxy/Chart.yaml
@@ -25,4 +25,4 @@ maintainers:
 name: haproxy
 sources:
   - https://github.com/bitnami/charts/tree/main/bitnami/haproxy
-version: 0.9.0
+version: 0.8.8
--- a/bitnami/haproxy/Chart.yaml
+++ b/bitnami/haproxy/Chart.yaml
@@ -25,4 +25,4 @@ maintainers:
 name: haproxy
 sources:
   - https://github.com/bitnami/charts/tree/main/bitnami/haproxy
-version: 0.8.7
+version: 0.9.0
--- a/bitnami/haproxy/templates/deployment.yaml
+++ b/bitnami/haproxy/templates/deployment.yaml
@@ -17,7 +17,9 @@ metadata:
   annotations: {{- include "common.tplvalues.render" ( dict "value" .Values.commonAnnotations "context" $ ) | nindent 4 }}
   {{- end }}
 spec:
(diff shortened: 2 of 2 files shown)


bitnami/haproxy/Chart.yaml
```diff
@@ -25,4 +25,4 @@ maintainers:
 name: haproxy
 sources:
   - https://github.com/bitnami/charts/tree/main/bitnami/haproxy
-version: 0.8.7
+version: 0.9.0
```
fmulero, line 28: According to [semver](https://semver.org) this new version should be a patch version because there is no new features.

```suggestion
version: 0.8.8
```

fmulero (commented):
Good catch @MuhammadTalhaPasha! Thanks a lot for your contribution.

I left only a little comment about the versioning. Could you give it a glance?

fmulero (approved):
Great! Thanks a lot! LGTM

MuhammadTalhaPasha:
Thank you.

PR #18208, opened 2023-08-04
[bitnami/wildfly] Adapt WildFly tests to Air-gapped environments 

jotamartos:
### Description of the change

Some tests do not work in air-gapped environments. We will store the helloworld.war file in our repo to run it when needed.

### Benefits

We are modifying the tests to make them work in all environments. This way we can verify that the solution works as expected. If the solution doesn't work in that environment, tests will not be changed and support for that env will be removed.

### Checklist


FraPazGal (approved):
LGTM!

Issue #18209, opened 2023-08-04
[bitnami/etcd] Make cluster service optional in ETCD_ADVERTISE_CLIENT_URLS


svenwiltink:
### Name and Version

bitnami/etcd latest

### What is the problem this feature will solve?

When trying to do cluster wide defragmentation using a tool such as https://github.com/ahrtr/etcd-defrag autodiscovery of members and their clientURLs is performed. For each clientURL a defragmentation is perfomed. The problem however is that each pod currently advertises the cluster service. This causes a defrag to be called on the service url itself which could make it hit the current leader causing a brief interruption in etcd. 

### What is the feature you are proposing to solve the problem?

In deployments where the cluster URL is unused it could be allowed to remove the cluster service from the advertised urls. There is already a feature flag `service.enabled` that controls the creation of the extra service. When this service is disabled we can remove the service URL from the advertised urls here: https://github.com/bitnami/charts/blob/main/bitnami/etcd/templates/statefulset.yaml#L182

PR #18211, opened 2023-08-04
[bitnami/etcd] only add the service to advertised addresses when it is enabled

svenwiltink:
### Description of the change

Title is self explanatory

### Benefits

Removes a non existing service url from the advertised address list

### Applicable issues

fixes https://github.com/bitnami/charts/issues/18209

### Checklist

### DIFF
--- a/bitnami/etcd/templates/statefulset.yaml
+++ b/bitnami/etcd/templates/statefulset.yaml
@@ -179,7 +179,7 @@ spec:
               {{- end }}
             {{- end }}
             - name: ETCD_ADVERTISE_CLIENT_URLS
-              value: "{{ $etcdClientProtocol }}://$(MY_POD_NAME).{{ $etcdHeadlessServiceName }}.{{ .Release.Namespace }}.svc.{{ $clusterDomain }}:{{ .Values.containerPorts.client }}{% if .Values.service.enabled %},{{ $etcdClientProtocol }}://{{ $etcdFullname }}.{{ .Release.Namespace }}.svc.{{ $clusterDomain }}:{{ coalesce .Values.service.ports.client .Values.service.port }}{% endif %}"
+              value: "{{ $etcdClientProtocol }}://$(MY_POD_NAME).{{ $etcdHeadlessServiceName }}.{{ .Release.Namespace }}.svc.{{ $clusterDomain }}:{{ .Values.containerPorts.client }}{{- if .Values.service.enabled }},{{ $etcdClientProtocol }}://{{ $etcdFullname }}.{{ .Release.Namespace }}.svc.{{ $clusterDomain }}:{{ coalesce .Values.service.ports.client .Values.service.port }}{{- end }}"
             - name: ETCD_LISTEN_CLIENT_URLS
               value: "{{ $etcdClientProtocol }}://0.0.0.0:{{ .Values.containerPorts.client }}"
             - name: ETCD_INITIAL_ADVERTISE_PEER_URLS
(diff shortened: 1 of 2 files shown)


carrodher:
Thanks for your contribution, please note it is needed to bump the chart version in the _Chart.yaml_ file

svenwiltink:
@carrodher thanks for the reminder, forgot about that. Besides bumping the version, is the fix in the chart itself ok regarding the issue I opened or are there more changes required?

bitnami/etcd/templates/statefulset.yaml
```diff
@@ -179,7 +179,7 @@ spec:
               {{- end }}
             {{- end }}
             - name: ETCD_ADVERTISE_CLIENT_URLS
-              value: "{{ $etcdClientProtocol }}://$(MY_POD_NAME).{{ $etcdHeadlessServiceName }}.{{ .Release.Namespace }}.svc.{{ $clusterDomain }}:{{ .Values.containerPorts.client }},{{ $etcdClientProtocol }}://{{ $etcdFullname }}.{{ .Release.Namespace }}.svc.{{ $clusterDomain }}:{{ coalesce .Values.service.ports.client .Values.service.port }}"
+              value: "{{ $etcdClientProtocol }}://$(MY_POD_NAME).{{ $etcdHeadlessServiceName }}.{{ .Release.Namespace }}.svc.{{ $clusterDomain }}:{{ .Values.containerPorts.client }}{% if .Values.service.enabled %},{{ $etcdClientProtocol }}://{{ $etcdFullname }}.{{ .Release.Namespace }}.svc.{{ $clusterDomain }}:{{ coalesce .Values.service.ports.client .Values.service.port }}{% endif %}"
```
FraPazGal, line 182: I think you made a format typo here:

```suggestion
              value: "{{ $etcdClientProtocol }}://$(MY_POD_NAME).{{ $etcdHeadlessServiceName }}.{{ .Release.Namespace }}.svc.{{ $clusterDomain }}:{{ .Values.containerPorts.client }}{{- if .Values.service.enabled }},{{ $etcdClientProtocol }}://{{ $etcdFullname }}.{{ .Release.Namespace }}.svc.{{ $clusterDomain }}:{{ coalesce .Values.service.ports.client .Values.service.port }}{{- end }}"
```

FraPazGal (commented):
Hi @svenwiltink, I understand the logic behind removing the svc URL from `ETCD_ADVERTISE_CLIENT_URLS` when `service.enabled=false` but reading your use case in https://github.com/bitnami/charts/issues/18209 wouldn't you still have the same problem? 

If you are using the envar as a list of endpoints, the difference between having or not the svc URL when the service is disabled is the fact that you will have an endpoint that won't resolve or not have it altogether right? What I mean is that, having the service disabled, its URL advertised shouldn't affect you right?

svenwiltink:
> Hi @svenwiltink, I understand the logic behind removing the svc URL from `ETCD_ADVERTISE_CLIENT_URLS` when `service.enabled=false` but reading your use case in #18209 wouldn't you still have the same problem?
> 
> If you are using the envar as a list of endpoints, the difference between having or not the svc URL when the service is disabled is the fact that you will have an endpoint that won't resolve or not have it altogether right? What I mean is that, having the service disabled, its URL advertised shouldn't affect you right?

The problem I'm describing in the linked issue is etcd-defrag using all the advertised addresses to perform a healthcheck. Even when the dns doesn't resolve the defrag tool will still try to perform a healthcheck, which in turn fails. Removing the non existing endpoint entirely fixes that particular problem.

FraPazGal (approved):
Thanks for the explanation @svenwiltink, LGTM!

PR #18213, opened 2023-08-04
[bitnami/external-dns] Add support for externalName service

fevisera:
### Description of the change

Add support for using `externalName` service.

### Benefits

Users can setup ExternalDNS with an ExternalName service.

### Possible drawbacks

N/A

### Applicable issues

- fixes #

### Additional information

[Example](https://github.com/kubernetes-sigs/external-dns/blob/master/docs/tutorials/externalname.md) of configuring externalName.

### Checklist


bitnami/external-dns/templates/service.yaml
```diff
@@ -30,6 +30,9 @@ spec:
   {{- if .Values.service.externalIPs }}
   externalIPs: {{ toYaml .Values.service.externalIPs | nindent 4 }}
   {{- end }}
+  {{- if .Values.service.externalName }}
```
fmulero, line 33: This new value only makes sense when [`service.type=ExternalName`](https://kubernetes.io/docs/concepts/services-networking/service/#externalname), I think we should add a condition like this:
```suggestion
  {{- if and (eq .Values.service.type "ExternalName") .Values.service.externalName }}
```

fmulero (commented):
Thanks @fevisera for your contibution.

I left only a little suggestion. Could you give it a glance?

bitnami/external-dns/templates/service.yaml
fevisera: Thank you for your review! You are right, I have applied your suggestion. Thanks.

fmulero (commented):
Great! LGTM

fmulero (approved):

PR #18236, opened 2023-08-07
[bitnami/nats] Adapt NATS tests to Air-gapped environments 

jotamartos:
### Description of the change

Some tests do not work in air-gapped environments. We will store the nats binary file in our repo to run it when needed.

The client is not updated as often as the solution so we can store the binary here and shouldn't introduce issues in our tests.

### Benefits

We are modifying the tests to make them work in all environments. This way we can verify that the solution works as expected. If the solution doesn't work in that environment, tests will not be changed and support for that env will be removed.

### Checklist

### DIFF
--- a/bitnami/nats/Chart.yaml
+++ b/bitnami/nats/Chart.yaml
@@ -31,4 +31,4 @@ maintainers:
 name: nats
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/nats
-version: 7.8.9
+version: 7.8.8


jotamartos:
Verification job succeeded [here](https://github.com/bitnami/charts/actions/runs/5786166158/job/15680787179)

bitnami/nats/Chart.yaml
```diff
@@ -31,4 +31,4 @@ maintainers:
 name: nats
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/nats
-version: 7.8.8
+version: 7.8.9
```
joancafom, line 34: Please, remove the trigger

joancafom (approved):

PR #18243, opened 2023-08-07
[bitnami/Thanos] Enable alerting rules for Thanos components only if enabled

davisgomes:
### Description of the change

Only enable Alerting rules for Thanos components if they are enabled. This change applies to the compactor, query, receive, ruler, and storegateway components. 

There are two locations where rules may be disabled based on its corresponding component
1. In the `absent_rules` rule file, the individual "IsDown" rules are removed from the `PrometheusRule` if the component is disabled
2. For each component Rule file (i.e. the `compaciton.yml` rule file), the entire `PrometheusRule` object is disabled if the component is disabled.

I have updated the README to reflect these chart changes.

### Benefits

If users disable components or components are disabled by default, their corresponding alerts will not fire if the component does not exist. Before this change, component alerts would fire if the component did not exist on the cluster, or if it was disabled at any point in the lifecycle of metrics collection.

### Possible drawbacks

It is possible that users would want to enable the component rules if the component is not enabled. This change precludes this from being possible without chart modifications.

### Applicable issues

- fixes #

### Checklist


FraPazGal (commented):
Hi @davisgomes,

I have some doubts regarding this PR. Checking the code, shouldn't the `metrics.prometheusRule.default.xxxxx` parameters be enough for your use case? Additionally, if having the component disabled but its rule active could be useful in some scenarios I don't think we should remove the feature.

PR #18324, opened 2023-08-09
[bitnami/kafka] Add enableServiceLinks

### LINKED ISSUE #18248: [bitnami/kafka] Support enableServiceLinks
### Name and Version

bitnami/kafka

When deploy kafka in a namespace with lots of services, then kafka is very slow to be launched because of injecting too many environment variables .

In the kafka container, simply tried `./kafka-topics.sh --bootstrap-server localhost:9092 --list`, it took 35 seconds to list the topics. This also makes the provisioning job running very very slow.

In my case, helm install kafka chart, it took almost 20 minutes to launch kafka and then provision 17 topics.

The configuration `enableServiceLinks` for pod is set to true by default, add this to kafka helm chart(kafka pod and provisioning job pod) and allow set this to false can stop injecting environment variables.

_No response_

VanessaLi-oqton:
### Description of the change

Add enableServiceLinks to the pod's spec including kafka broker and controller statefulset, kafka exporter deployment, provisioning job.

### Benefits

EnableServiceLinks indicates whether information about services should be injected into pod's environment variables, the default value from Kubernetes is true. 

If there are many services in a namespace,`enableServiceLinks: true` will result in inject too many environment variables to a pod, then the kafka startup as well as running scripts in a pod becomes very slow. Make enableServiceLinks configurable and set to false in this case can stop injecting services related environment variables to the pod.

### Applicable issues

- fixes #18248

### Checklist


dgomezleon (approved):

PR #18393, opened 2023-08-14
[bitnami/opensearch] Fix guards for OpenSearch Dashboards

f4z4on:
### Description of the change

We already declare Helm chart value to enable or disable OpenSearch Dashboards. We also use guards checking similar and other values for other types of workloads. We use these guards to ultimately decide if particular workloads and Kubernetes resources associated with them should be configured. What we have been missing is a similar guard which checks if OpenSearch Dashboards should be configured.

We fix this by introducing opensearch.dashboards.enabled helper template definition. It is similar to opensearch.ingest.enabled because Helm chart values configuring ingest contain `*.enabled` field as well as fields all the guards check: `*.autoscaling.hpa.enabled` and `*.replicaCount`. We implement equivalent condition for the new helper template definition.

We also apply this new guard consistently with other types of workloads…

When implementing this fix, we have noticed another issue with guard already in place. We have implemented them, so they check at least if autoscaling is enabled or if replica count is greater than zero. However, we support both horizontal pod autoscaler (HPA) and vertical pod autoscaler (VPA). Our Helm chart values have no unspecified autoscaling.

We change guards to check if HPA is enabled. If it is (and other conditions are true), we should configure Kubernetes resource associated with particular role. This is because pods for this particular role will be created as needed. We do not care about VPA, because VPA requires some pods to exists in the first place and we already cover that (or fix in this commit).

### Benefits

If we disable OpenSearch Dashboards, no workloads are scheduled in Kubernetes anymore.

We fix and consistently apply guards for different types of workloads checking replica count or HPA. For some types of workloads, we also check their `enabled` flag if it is a declared Helm chart value.

### Possible drawbacks

Change of undocumented, buggy behavior someone may rely upon.

### Applicable issues

- fixes scheduling of workload with OpenSearch Dashboards
- fixes checking if HPA is enabled instead undocumented Helm chart values not used anywhere else

### Additional information

None.

### Checklist


Mauraza:
Hi @f4z4on,

Let me check why the tests are falling with this change.

Mauraza (approved):
Hi @f4z4on,

Thanks for the contribution!!! LGTM!!

f4z4on:
Thank you too, @Mauraza!

Have you found out what the issue with the VIB has been about? I guess you must have noticed, there is [another issue on VIB](https://github.com/bitnami/charts/actions/runs/5938646573/job/16103520894) caused by merging this PR.

I wanted to check myself over the weekend separately using my own VMware account, but ended spending my computer time differently, so I'm still a bit blind about what’s going on in the VIB. My help is not entirely altruistic as I and other rely on this Helm chart on one project. The PR works for us, but we’re not running your pipelines. I’m still okay spending some time making sure I'm not breaking any tests if it helps… But I wanted check if you know more since you merged it…

Issue #17482, opened 2023-07-05
[bitnami/logstash] Can't create NodePort through logstash helm chart 


copy1dollar:
### Name and Version

bitnami/logstash 5.4.4

### What architecture are you using?

amd64

### What steps will reproduce the bug?

Installing Logstash through Helm-Chart with `service.type=NodePort` 

### What do you see instead?

```
Error: UPGRADE FAILED: failed to create resource: Service "logstash-headless" is invalid: spec.ports[0].nodePort: Forbidden: may not be used when type is 'ClusterIP'
```

### Additional information

duplicated issue : https://github.com/bitnami/charts/issues/15393 
but I don't know how reopen the issue.

and related : https://github.com/elastic/helm-charts/pull/839

rafariossaa:
Hi,
  Thanks, #15393 was reopened yesterday after your comment.
  I am closing this issue, let's use the initial github issue to track this.

Issue #15393, opened 2023-03-08
[bitnami/logstash] Can't create NodePort through logstash helm chart


joshua-heidrich:
### Name and Version

bitnami/logstash 5.1.15

### What architecture are you using?

amd64

### What steps will reproduce the bug?

1. Installing Logstash through Helm-Chart with service.type=NodePort

### Are you using any custom parameters or values?

```
service:
  type: NodePort
  ports:
    - name: gelf
      port: 1234
      targetPort: 1234
      nodePort: 30345
      protocol: UDP
```

### What is the expected behavior?

The install works.

### What do you see instead?

`Error: UPGRADE FAILED: failed to create resource: Service "logstash-headless" is invalid: spec.ports[0].nodePort: Forbidden: may not be used when `type` is 'ClusterIP'`

### Additional information

I think I already have the source of the problem. There is this section in headless-svc.yaml:
```
spec:
  type: ClusterIP
  clusterIP: None
  ports:
    {{- include "common.tplvalues.render" (dict "value" .Values.service.ports "context" $) | nindent 4 }}
  selector: {{ include "common.labels.matchLabels" . | nindent 4 }}
```
It uses  `.Values.service.ports` which I have defined as follows:
```
    - name: gelf
      port: 1234
      targetPort: 1234
      nodePort: 30345
      protocol: UDP
```
I need the "nodePort"-part because I want to create a Nodeport. But this uses the NodePort port to create a ClusterIP which obviously does not work.

javsalgar:
Hi!

Thank you so much for reporting! I believe we could use the `omit` go template function to not add the nodePort section inside the headless service. I will forward this to the team. However, if you want to speed up the process, would you like to submit a PR?

copy1dollar:
This issue is not yet resolved.
The same problem is happening with 5.4.4.

Issue #6278, opened 2021-05-03
[bitnami/logstash] Can't manually attribute port number for NodePort


pbrissaud:
**Which chart**:
Logstash 3.3.2

**Is your feature request related to a problem? Please describe.**
I can't manually define node port number in the chart. If I modify `service.type` to **NodePort** and I add a **nodePort** attribute to each port in `service.ports`, I have the following error :
```bash
Error: UPGRADE FAILED: cannot patch "logstash-headless" with kind Service: Service "logstash-headless" is invalid: spec.ports[1].nodePort: Forbidden: may not be used when `type` is 'ClusterIP'
```

**Describe the solution you'd like**
I will like to have the possibility to manually attribute the nodePort like the official chart does.  

**Describe alternatives you've considered**
If I delete the `templates/headless-svc` file, I can deploy the chart and the node port is well defined

**Additional context**

working values.yaml file if I delete the headless-svc template :
```yaml
service:
  type: NodePort
  ports:
    syslog-tcp:
      port: 1514
      targetPort: syslog-tcp
      protocol: TCP
      nodePort: 31044 
```

marcosbc:
Hi @pbrissaud, thanks for reporting this! It indeeds looks like a non-trivial actual bug. I've created an internal task to fix this issue, but unfortunately I cannot give an ETA for when it will be solved.

Note that we are open for contributions, so if you happen to identify a solution, feel free to send a PR! We'd be glad to help with the merge and release processes.

carrodher:
Unfortunately, this issue was created a long time ago and although there is an internal task to fix it, it was not prioritized as something to address in the short/mid term. It's not a technical reason but something related to the capacity since we're a small team.

Being said that, contributions via PRs are more than welcome in both repositories (containers and charts). Just in case you would like to contribute.

During this time, there are several releases of this asset and it's possible the issue has gone as part of other changes. If that's not the case and you are still experiencing this issue, please feel free to reopen it and we will re-evaluate it.

dansou901:
Hi, I'm sorry but the issue is still there, just came across it. I don't have a solution for now, but if I have time to search for one, I'll let you know.

dansou901:
Ok, easiest solution would be to add a special param "nodePort" under the ports section like you did with loadBalancerIP for example. Then you add another if statement in the service template and evaluate that nodePort param there if service.type is NodePort. Should be easy enough to implement

PR #18476, opened 2023-08-16
[bitnami/logstash] fix: support NodePorts in bitnami logstash chart

cbrwflo:
### Description of the change

Omits the `nodePort` field on the logstash headless service.

This fixes the error `spec.ports[1].nodePort: Forbidden: may not be used when type is 'ClusterIP'` .

### Applicable issues

Fixes:

- #17482
- #15393
- #6278

### Additional information

I didn't see a consistent way of expanding the `service.spec.ports` for headless services across bitnami charts, so I modelled this after the [fluentd headless service](https://github.com/bitnami/charts/blob/main/bitnami/fluentd/templates/aggregator-svc-headless.yaml#L36). 

Instead of using `service.spec.ports.name` (since it's an optional field when only 1 port is used), I opted to use `service.spec.port.port` instead (a required field). I don't love the nested `omit`, and happy to change the PR to "expand" the fields instead if you prefer.

### Checklist


bitnami/logstash/templates/headless-svc.yaml
```diff
@@ -25,5 +25,8 @@ spec:
   type: ClusterIP
   clusterIP: None
   ports:
-    {{- include "common.tplvalues.render" (dict "value" .Values.service.ports "context" $) | nindent 4 }}
+  {{- range $service := .Values.service.ports }}
+    - port: {{ $service.port }}
+      {{- (omit (omit $service "nodePort") "port") | toYaml | nindent 6 }}
+  {{- end }}
```
FraPazGal, line 31:
```suggestion
  {{- range $service := .Values.service.ports }}
    - name: {{ $service.name }}
      port: {{ $service.port }}
      targetPort: {{ $service.targetPort }}
      protocol: {{ $service.protocol }}
  {{- end }}
```

A bit more verbose but IMO it's also more intuitive and avoids using the nested `omit`.

FraPazGal (commented):
Hello @cbrwflo, thanks for catching this error and creating a PR!

Checking the existing logic, you couldn't find any similar approach because we don't normally use an array for `service.ports`. To fix this, we could also refactor the whole ports logic and use the `service.extraPorts` parameter (see `bitnami/schema-registry` for an example). The problem with that solution is that it would require a new major version for the chart and I don't think it is worthwhile.

For the reasons explained above, I think we can continue with you approach. I've added a suggestion to avoid the `omit` usage, please check it and tell me what you think.

bitnami/logstash/templates/headless-svc.yaml
```diff
@@ -25,5 +25,10 @@ spec:
   type: ClusterIP
   clusterIP: None
   ports:
-    {{- include "common.tplvalues.render" (dict "value" .Values.service.ports "context" $) | nindent 4 }}
+  {{- range $service := .Values.service.ports }}
+    - name: {{ $service.name }}
+      port: {{ $service.port }}
+      targetPort: {{ $service.targetPort }}
+      protocol: {{ $service.protocol }}
+  {{- end }}
```
juan131, line 33: There's a problem with this implementation: it's not compatible with using templates at `.Values.service.ports` while the previous solution was.

FraPazGal: True, we'll be losing that option in this case... I guess we can con with the `omit` solution and work on reworking this logic in a future refactor for this chart.

juan131, line 33: This should be compatible with using templates.
```suggestion
    {{- range $port := .Values.service.ports }}
    {{- omit (include "common.tplvalues.render" ( dict "value" $port "context" $ ) | fromYaml) "nodePort" | toYaml | nindent 4 }}
    {{- end }}
```

carrodher:
We just reopened this PR, @cbrwflo could you please take a look at the review comments from my teammates? Apart from that, it seems there was a new version during this time, could you please fix the conflicts? Thanks

cbrwflo:
Make fresh! Still interested in this (need to rebase), back from vacay 10/25

cbrwflo:
I intend to work on this

bitnami/logstash/templates/headless-svc.yaml
```diff
@@ -17,6 +17,7 @@ spec:
   type: ClusterIP
   clusterIP: None
   ports:
-    {{- include "common.tplvalues.render" (dict "value" .Values.service.ports "context" $) | nindent 4 }}
-  {{- $podLabels := include "common.tplvalues.merge" ( dict "values" ( list .Values.podLabels .Values.commonLabels ) "context" . ) }}
-  selector: {{- include "common.labels.matchLabels" ( dict "customLabels" $podLabels "context" $ ) | nindent 4 }}
+    {{- range $port := .Values.service.ports }}
+    {{- omit (include "common.tplvalues.render" ( dict "value" $port "context" $ ) | fromYaml) "nodePort" | toYaml | nindent 4 }}
+    {{- end }}
+  selector: {{ include "common.labels.matchLabels" . | nindent 4 }}
```
juan131, line 23: Please maintain the original way to setup the selector:
```suggestion
  {{- $podLabels := include "common.tplvalues.merge" ( dict "values" ( list .Values.podLabels .Values.commonLabels ) "context" . ) }}
  selector: {{- include "common.labels.matchLabels" ( dict "customLabels" $podLabels "context" $ ) | nindent 4 }}
```

cbrwflo:
Commit 25bb38d fixes the issue VIB caught (port items rendered without the `-`)

Without:
```yaml
# helm template --values .vib/logstash/runtime-parameters.yaml logstash ./bitnami/logstash > app.yaml
spec:
  type: ClusterIP
  clusterIP: None
  ports:
    name: http
    port: 8081
    protocol: TCP
    targetPort: http
    name: monitoring
    port: 80
    protocol: TCP
    targetPort: monitoring
```

With:
```yaml
spec:
  type: ClusterIP
  clusterIP: None
  ports:
    -
      name: http
      port: 8081
      protocol: TCP
      targetPort: http
    -
      name: monitoring
      port: 80
      protocol: TCP
      targetPort: monitoring
```

cbrwflo:
I guess there's also `- {{ ... | toYaml | indent 6 | trim }}`, which looks like:

```yaml
spec:
  type: ClusterIP
  clusterIP: None
  ports:
    - name: http
      port: 8081
      protocol: TCP
      targetPort: http
    - name: monitoring
      port: 80
      protocol: TCP
      targetPort: monitoring
```

juan131 (approved):
LGTM

PR #13723, opened 2022-11-28
[bitnami/argo-cd] Add applicationSet deployment to the chart

jotamartos:
Signed-off-by: jotamartos <jotamartos@vmware.com>

### Description of the change

This PR adds support to the applicationSet deployment in the Bitnami's chart. The upstream project added support to its chart [here](https://github.com/argoproj/argo-helm/tree/main/charts/argo-cd/templates/argocd-applicationset)

### Benefits

The ApplicationSet controller is a [Kubernetes controller](https://kubernetes.io/docs/concepts/architecture/controller/) that adds support for an ApplicationSet [CustomResourceDefinition](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/) (CRD).

More info in [the official documentation](https://argocd-applicationset.readthedocs.io/en/stable/)

### Possible drawbacks

None

### Applicable issues

  - fixes https://github.com/bitnami/charts/issues/9490

### Checklist

- [X] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/). This is *not necessary* when the changes only affect README.md files.
- X] Variables are documented in the values.yaml and added to the `README.md` using [readme-generator-for-helm](https://github.com/bitnami-labs/readme-generator-for-helm)
- [X] Title of the pull request follows this pattern [bitnami/<name_of_the_chart>] Descriptive title
- [X] All commits signed off and in agreement of [Developer Certificate of Origin (DCO)](https://github.com/bitnami/charts/blob/main/CONTRIBUTING.md#sign-your-work)

PR #18624, opened 2023-08-18
[bitnami/argo-cd] Provide configmap for rbac configuration

maxnitze:
### Description of the change

Besides the `gpg-keys` and `tls-certs` configmaps, there is also a `rbac` configmap for role-based authentication configuration.

### Benefits

Configure rbac as part of the chart.

### Possible drawbacks

None

### Additional information

See also:
* https://argo-cd.readthedocs.io/en/stable/operator-manual/rbac/
* https://github.com/argoproj/argo-cd/blob/master/docs/operator-manual/argocd-rbac-cm.yaml

### Checklist

### DIFF
--- a/bitnami/argo-cd/Chart.yaml
+++ b/bitnami/argo-cd/Chart.yaml
@@ -39,4 +39,4 @@ maintainers:
 name: argo-cd
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/argo-cd
-version: 4.7.25
+version: 4.8.0
--- a/bitnami/argo-cd/README.md
+++ b/bitnami/argo-cd/README.md
@@ -810,6 +810,7 @@ The command removes all the Kubernetes components associated with the chart and 
 | `config.existingStylesConfigmap`               | Use an existing styles configmap                                                                      | `""`   |
 | `config.tlsCerts`                              | TLS certificates used to verify the authenticity of the repository servers                            | `{}`   |
 | `config.gpgKeys`                               | GnuPG public keys to add to the keyring                                                               | `{}`   |
+| `config.rbac`                                  | Role-based authentication configuration                                                               | `{}`   |
 | `config.secret.create`                         | Whether to create or not the secret                                                                   | `true` |
 | `config.secret.annotations`                    | General secret extra annotations                                                                      | `{}`   |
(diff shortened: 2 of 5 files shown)


Mauraza:
Hi @maxnitze,

Thanks for the contribution!! I was checking the PR and missed adding the 'config.rbac' in the [deployment.yaml ](https://github.com/bitnami/charts/blob/main/bitnami/argo-cd/templates/applicationset/deployment.yaml)as 'config.gpgKeys'. Could you add it to 'deployment.yaml'?

maxnitze:
Oh. Good point!

I never actually worked with `ApplicationSet`. And in the _regular_ controller the ConfgMap just needs to be there and the controller knows what to do with it (no explicit mounting necessary).

I'll try to figure out where I need to mount it in the ApplicationSet deployment.

maxnitze:
@jotamartos , maybe you can help me out here? You initially added the support for applicationsets in #13723.

Just for consistency I would now add the following:

```yaml
          volumeMounts:
            # ...
            {{- if .Values.config.rbac }}
            - mountPath: /app/config/rbac
              name: rbac
            {{- end }}
      volumes:
        # ...
        {{- if .Values.config.rbac }}
        - name: rbac
          configMap:
            name: argocd-rbac-cm
        {{- end }}
```

But I have no idea, if this is the correct place for the config? Can you point me to some documentation about that? I only found the general docs about ApplicationSet (https://argo-cd.readthedocs.io/en/stable/operator-manual/applicationset/), but I cannot find any info about the config in there.

Mauraza:
Hi @maxnitze,

I was investigating, and it seems that the name of the config map is enough (`argocd-rbac-cm`), so I have a question, did you test this change?

maxnitze:
Yes, I have this in production already with a manually created configmap with that name. I did not test it with an applicationset controller. Though I would expect it to work without mounting it as well, since there is no other documentation on it. Not aure if mounting the other configmaps there is even necessary.

Do you think I should mount it there anyways?

maxnitze:
I added the config to mount the rbac-cm content to `/app/config/rbac` in the ApplicationSet container. Just for consistency with the other config files.

Could you have another look? (I will take care of the conflict in a sec)

Mauraza (approved):
Hi @maxnitze,

PR #18154, opened 2023-08-03
[bitnami/common] Add support for customizing standard labels

juan131:
### Description of the change

This PR makes it possible to overwrite the standard labels below that are currently set with predefined values based on chart name, chart release name, etc.

- `app.kubernetes.io/name`
- `helm.sh/chart`
- `app.kubernetes.io/managed-by`
- `app.kubernetes.io/instance`

### Benefits

Users can set any desired custom value to the labels mentioned above.

### Possible drawbacks

We'll have to update every single chart right away to make them compatible with new `common.labels.standard` and `common.labels.matchLabels` helpers' format.

### Applicable issues

- fixes #18025

### Checklist

- [x] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/). This is *not necessary* when the changes only affect README.md files.
- [x] Title of the pull request follows this pattern [bitnami/<name_of_the_chart>] Descriptive title
- [x] All commits signed off and in agreement of [Developer Certificate of Origin (DCO)](https://github.com/bitnami/charts/blob/main/CONTRIBUTING.md#sign-your-work)

### Additional information

Find below an example for adopting the new helper format on Apache:

```diff
diff --git a/bitnami/apache/templates/deployment.yaml b/bitnami/apache/templates/deployment.yaml
index 7fc270be8..da9a75a89 100644
--- a/bitnami/apache/templates/deployment.yaml
+++ b/bitnami/apache/templates/deployment.yaml
@@ -8,16 +8,13 @@ kind: Deployment
 metadata:
   name: {{ include "common.names.fullname" . }}
   namespace: {{ .Release.Namespace | quote }}
-  labels: {{- include "common.labels.standard" . | nindent 4 }}
-    {{- if .Values.commonLabels }}
-    {{- include "common.tplvalues.render" ( dict "value" .Values.commonLabels "context" $ ) | nindent 4 }}
-    {{- end }}
+  labels: {{- include "common.labels.standard" ( dict "customLabels" .Values.commonLabels "context" $ ) | nindent 4 }}
   {{- if .Values.commonAnnotations }}
   annotations: {{- include "common.tplvalues.render" ( dict "value" .Values.commonAnnotations "context" $ ) | nindent 4 }}
   {{- end }}
 spec:
   selector:
-    matchLabels: {{- include "common.labels.matchLabels" . | nindent 6 }}
+    matchLabels: {{- include "common.labels.matchLabels" ( dict "customLabels" .Values.podLabels "context" $ ) | nindent 6 }}
[...]

PR #18742, opened 2023-08-21
[bitnami/phpmyadmin] Support for customizing standard labels

juan131:
### Description of the change

This PR follows up #18154 ensuring this chart is adapted to use new helpers' format, hence adding support for customizing standard labels.

### Benefits

User can customize standard labels.

### Possible drawbacks

None

### Applicable issues

N/A

### Additional information

#18154 must be merged **before** this PR

### Checklist


bitnami/phpmyadmin/templates/deployment.yaml
```diff
@@ -8,29 +8,24 @@ kind: Deployment
...
   strategy: {{- include "common.tplvalues.render" (dict "value" .Values.updateStrategy "context" $ ) | nindent 4 }}
   template:
     metadata:
-      labels: {{- include "common.labels.standard" . | nindent 8 }}
-        {{- if .Values.podLabels }}
-        {{- include "common.tplvalues.render" ( dict "value" .Values.podLabels "context" $) | nindent 8 }}
-        {{- end }}
-      {{- if or .Values.podAnnotations .Values.metrics.enabled }}
+      labels: {{- include "common.labels.standard" ( dict "customLabels" $podLabels "context" $ ) | nindent 8 }}
+      {{- if or .Values.podAnnotations (and .Values.metrics.enabled .Values.metrics.podAnnotations) }}
```
fevisera, line 23: Is this `if` condition needed? I think there are already existing `if` blocks that evaluate the same condition.

bitnami/phpmyadmin/templates/ingress.yaml
```diff
@@ -9,20 +9,15 @@ kind: Ingress
...
   annotations:
     {{- if .Values.ingress.certManager }}
     kubernetes.io/tls-acme: "true"
     {{- end }}
-    {{- if .Values.ingress.annotations }}
-    {{- include "common.tplvalues.render" ( dict "value" .Values.ingress.annotations "context" $ ) | nindent 4 }}
-    {{- end }}
-    {{- if .Values.commonAnnotations }}
-    {{- include "common.tplvalues.render" ( dict "value" .Values.commonAnnotations "context" $ ) | nindent 4 }}
+    {{- if or .Values.ingress.annotations .Values.commonAnnotations }}
```
fevisera, line 18: I believe you can eliminate this `if`. It appears that by using `merge .Values.ingress.annotations .Values.commonAnnotations` there is no need to verify `or .Values.ingress.annotations .Values.commonAnnotations`.

juan131: It is required. Please note that both might be empty if we're evaluating this block because of the third condition> `if .Values.ingress.certManager`

bitnami/phpmyadmin/templates/deployment.yaml
juan131: It's simply to avoid having an empty annotations object when none of the conditions are met. Avoiding this:
```yaml
spec:
  template:
    metadata:
      labels:
        app.kubernetes.io/name: phpmyadmin
        (...)
      annotations:
    spec:
```

bitnami/phpmyadmin/templates/ingress.yaml
fevisera: You're right, we should keep it. Thanks for the explanation.

bitnami/phpmyadmin/templates/deployment.yaml
fevisera: You're right, we should keep it. Thanks for the explanation :)

fevisera (approved):
LGMT!

PR #18922, opened 2023-08-29
[bitnami/redis] expose service binding with correct uri

davy-ikv:
### Description of the change

Redis URI format should be like "redis://user:secret@localhost:6379/0?foo=bar&qux=baz"

The "userinfo" part should be "user:secret".

Since user part is empty, the uri will have to leave a colon then concatenate with the secret.

ref: https://www.iana.org/assignments/uri-schemes/prov/redis

### Benefits

Change uri format to match the RFC so that other applications follows the RFC can connect to redis service by the uri directly.

### Checklist

### DIFF
--- a/bitnami/redis/Chart.yaml
+++ b/bitnami/redis/Chart.yaml
@@ -34,4 +34,4 @@ maintainers:
 name: redis
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/redis
-version: 18.0.1
+version: 18.0.2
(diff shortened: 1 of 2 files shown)


jotamartos (changes requested):
Thanks for your contribution. However, as you can see, verification is failing because you didn't bump the chart's version. Could you please do so?

davy-ikv:
@jotamartos I bumped the chart version, but the CI failed, how can I fix that? thanks

PR #18954, opened 2023-08-31
[bitnami/minio] Add `disableVersioning` flag for provisioning unversioned buckets

pvallone:
### Description of the change

It isn't currently possible to provision unversioned buckets from the Helm chart. In MinIO, bucket versioning can be in three states: `versioned`, `suspended`, and `unversioned`. The Helm chart's existing `bucket.versioning` field is a boolean, and so it only allows buckets to be provisioned with `suspended` or `versioned` versioning statuses. It's desirable to be able to provision unversioned buckets, as otherwise, folders/prefixes are not automatically deleted when their contents are deleted.

This PR adds a `disableVersioning` flag to the MinIO bucket provisioning logic. If set to `true`, provisioned buckets will be left unversioned.

I verified the following by deploying to Minikube:
- With `disableVersioning: false` and `versioning: false`, provisioned buckets have a "suspended" versioning status. This is the same behavior as before.
- With `disableVersioning` not specified at all in the bucket configuration and `versioning: false`, provisioned buckets have a "suspended" versioning status. This is the same behavior as before, and verifies that this won't be a breaking change.
- With `disableVersioning: true`, `mc version info myminio/bucket-name` reports an "unversioned" versioning status, as desired.

### Benefits

Clients will be able to provision unversioned buckets.

### Possible drawbacks

It's confusing to have two similarly named parameters, `disableVersioning` and `versioning`, which at face value should be doing the same thing. In a perfect world, `versioning` would be an enum with three possible values (`unversioned`, `suspended`, and `versioned`), but changing it now would constitute a breaking change. I tried to add comments to help clear this up. I'm open to suggestions!

### Checklist

### DIFF
--- a/bitnami/minio/Chart.yaml
+++ b/bitnami/minio/Chart.yaml
@@ -27,4 +27,4 @@ maintainers:
 name: minio
 sources:
   - https://github.com/bitnami/charts/tree/main/bitnami/minio
-version: 12.8.1
+version: 12.8.2
--- a/bitnami/minio/templates/provisioning-job.yaml
+++ b/bitnami/minio/templates/provisioning-job.yaml
@@ -194,9 +194,11 @@ spec:
               {{- end }}
 
               {{- if $isDistributedMode }}
+              {{- if not $bucket.disableVersioning }}
               {{- if (or ((empty $bucket.withLock)) (not $bucket.withLock)) }}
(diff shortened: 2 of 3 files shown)


pvallone:
Actually, it might be possible to turn the existing `versions` field into an enum without making a breaking change. I'll abandon this while I experiment with that.

PR #18978, opened 2023-09-01
[bitnami/kafka] Added values for supplying additional configuration as a secret object

### LINKED ISSUE #18420: [bitnami/kafka] Add a secret extraConfig in addition to the configMap one.
### Name and Version

bitnami/kafka 24.0.10

Currently, the only way too add custom settings to the kafka config, is by using the extraConfig value. This is problematic as i might want to add custom settings that are secret. Since the new upgrade, the environment variables with prefix "KAFKA_CFG" are no longer working, and therefore the extraEnvVarsSecret value does not provide this functionality anymore.

I am using the chart with the sasl mechanism oauthbearer, which is relatively easy to configure, even though the chart doesn't officially support it. I have to overwrite the JAAS config, and the config will contain secrets, that i don't want to be stored in a configMap.

A value called extraConfigSecret in which i can add extra configs as secrets. These would also need to not be included in the normal kafka config configMap of the template, and instead be appended to the server.properties separately from the configMap.

None. I don't see an alternative other than letting the secrets be stored in configMaps.

LuxTheDude:
### Description of the change

Added the following values for supplying additional configuration as a secret object:

_secretConfig_ --> A string that will be stored in a secret, and added to the server.properties during the init-container run.
_controller.secretConfig_ --> Same as _secretConfig_ but will only be applied to controller-eligible pods.
_broker.secretConfig_ --> Same as _secretConfig_ but will only be applied to broker-only nodes.

If both _secretConfig_ and one of the others are supplied, both will be appended, with _secretConfig_ being appeneded first.

_existingSecretConfig_ --> Name of an existing secret, containing additonal configuration that will be appended to server.properties during the init-container run. If supplied, none of the previous values above will be used.
_controller.existingSecretConfig_ --> Same as _existingSecretConfig_ but will only be applied to controller-eligible pods.
_broker.existingSecretConfig_ --> Same as _existingSecretConfig_ but will only be applied to broker-only pods.

If either of the controller/broker specific existingSecretConfigs are supplied, the non-specific _existingSecretConfig_ will be ignoed.

### Benefits

Currently the only way of providing additional config is via the _extraConfig_ value, which is put into a configMap.
I was missing the possibility to add additional config that might contain secret values. This PR adds this feature.

### Possible drawbacks

I do not see any drawbacks.

### Applicable issues

- fixes #18420 

### Additional information

-

### Checklist


rafariossaa (approved):
Thanks for adding this.
LGTM

PR #19177, opened 2023-09-07
[bitnami/common]: Compatiblity with Helm 3.2.0+

### LINKED ISSUE #19011: binami/common breaks with helm 3.5.4
### Name and Version

binami/common 2.10.0

amd64

The following script creates a plain helm chart (with `helm create`) and _only_ adds `bitnami/common` as dependency.
Just the import of the dependency causes the rendering of the template to fail.
```bash
#!/bin/bash
# create a tmp folder
mkdir /tmp/test
cd /tmp/test

# download helm 3.5.4
wget https://get.helm.sh/helm-v3.5.4-linux-amd64.tar.gz
tar -xf helm-v3.5.4-linux-amd64.tar.gz
cd linux-amd64/

# create a new test chart
./helm create incompatible-common

# add the dependency
cat <<EOT >> ./incompatible-common/Chart.yaml
dependencies:
  - name: common
    version: ^2.10.0
EOT

# update the dependencies
./helm repo add bitnami https://charts.bitnami.com/bitnami
./helm dependency update incompatible-common

# render the plain sample template (without even using bitnami/common)
./helm template incompatible-common
# -> this command fails with `Error: parse error at (incompatible-common/charts/common/templates/_labels.tpl:14): unclosed action`
```

- No

I would expect `bitnami/common` to be compatible with all helm versions stated in the README (i.e. the script above should terminate without an error).

### What do you see instead?

The default chart (created with `helm create`) with only the `common` chart added cannot be rendered. 
It fails with the following error:
```
Error: parse error at (incompatible-common/charts/common/templates/_labels.tpl:14): unclosed action

```

### Additional information

It seems like this issue has been introduced with #18154.
It was even discussed in the PR in this comment: https://github.com/bitnami/charts/pull/18154#issuecomment-1691404353
However, neither the prerequisites of the chart in the README nor the template have been changed to fix the issue.

This doesn't seem to be the biggest issue, and it seems to be fixed with Helm 3.6.0+. But we basically use the same prerequisites for our helm chart where we use `bitnami/common` as a dependency. So we would either have to bump our supported versions or pin the `bitnami/common` (at least for now) to the last release.

juan131:
### Description of the change

This PR attempts to fix an incompatibility introduced at #18154 due to the use of "new line" characters on some `dict` instructions. [This limitation](https://github.com/golang/go/issues/29770) that was fixed in `go1.16` affects Helm < 3.6 versions.

### Benefits

Users can use any Helm 3.2.0+ version with `bitnami/common`.

### Possible drawbacks

None

### Applicable issues

- fixes #19011

### Additional information

N/A

### Checklist

### DIFF
--- a/bitnami/common/Chart.yaml
+++ b/bitnami/common/Chart.yaml
@@ -6,7 +6,7 @@ annotations:
   licenses: Apache-2.0
 apiVersion: v2
 # Please make sure that version and appVersion are always the same.
-appVersion: 2.9.2
+appVersion: 2.10.1
 description: A Library Helm Chart for grouping common logic between bitnami charts. This chart is not deployable by itself.
 home: https://bitnami.com
 icon: https://bitnami.com/downloads/logos/bitnami-mark.png
@@ -23,4 +23,4 @@ name: common
 sources:
   - https://github.com/bitnami/charts
 type: library
-version: 2.10.0
+version: 2.10.1
--- a/bitnami/common/templates/_labels.tpl
+++ b/bitnami/common/templates/_labels.tpl
@@ -11,16 +11,7 @@ Kubernetes standard labels
 */}}
 {{- define "common.labels.standard" -}}
 {{- if and (hasKey . "customLabels") (hasKey . "context") -}}
-{{ merge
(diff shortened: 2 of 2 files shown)


corico44 (approved):
LGTM

Issue #18924, opened 2023-08-29
[bitnami/redis-cluster] Cluster resize with ArgoCD fails


ritmas:
### Name and Version

bitnami/redis-cluster 9.0.0

### What architecture are you using?

amd64

### What steps will reproduce the bug?

1. Create ArgoCD application with bitnami/redis-cluster chart: 3 nodes without replicas, with persistence on (for cluster state)
2. Adjust deployment by adding extra node, 4 nodes in total
3. The 4th node runs as standalone server

### Setup cluster
```yaml
apiVersion: "argoproj.io/v1alpha1"
kind: "Application"
metadata:
  name: "redis-cluster"
spec:
  source:
    chart: "redis-cluster"
    repoURL: "https://charts.bitnami.com/bitnami"
    targetRevision: "9.0.0"
    helm:
      parameters:
        - name: usePassword
          value: "false"
        - name: persistence.size
          value: 1Gi
        - name: redis.configmap
          value: |
            maxmemory 128mb
            maxmemory-policy allkeys-lfu
            appendonly no
            save ""
        - name: redis.resources.requests.cpu
          value: 500m
        - name: redis.resources.requests.memory
          value: 150Mi
        - name: redis.resources.limits.memory
          value: 150Mi
        - name: metrics.enabled
          value: "true"
        - name: cluster.nodes
          value: "3" # ==> 3 master nodes + 0 replicas
        - name: cluster.replicas
          value: "0"
      values: |
        redis:
          useAOFPersistence: "no"
  project: development
  syncPolicy:
    automated:
      prune: true
      selfHeal: true
```

### Resize cluster
```yaml
apiVersion: "argoproj.io/v1alpha1"
kind: "Application"
metadata:
  name: "redis-cluster"
spec:
  source:
    chart: "redis-cluster"
    repoURL: "https://charts.bitnami.com/bitnami"
    targetRevision: "9.0.0"
    helm:
      parameters:
        - name: usePassword
          value: "false"
        - name: persistence.size
          value: 1Gi
        - name: redis.configmap
          value: |
            maxmemory 128mb
            maxmemory-policy allkeys-lfu
            appendonly no
            save ""
        - name: redis.resources.requests.cpu
          value: 500m
        - name: redis.resources.requests.memory
          value: 150Mi
        - name: redis.resources.limits.memory
          value: 150Mi
        - name: metrics.enabled
          value: "true"
        - name: cluster.nodes
[...]

PR #19221, opened 2023-09-11
[bitnami/redis-cluster] Allow set Helm/ArgoCD hook for updateJob

ritmas:
### Description of the change

Ability to setup other Helm hook rather than hard coded one.

### Benefits

On ArgoCD side, a hard coded `"helm.sh/hook": post-upgrade` hook is equivalent to `argocd.argoproj.io/hook: PostSync`, which means, hook will be launched only when ArgoCD application is fully synced.

But it never does as new added node runs in standalone mode and as the default `readinessProbe` is [checking for cluster info](https://github.com/bitnami/charts/blob/main/bitnami/redis-cluster/templates/scripts-configmap.yaml#L52-L77) it will always fail.

Ability to setup other Helm hook would help to launch the script on ArgoCD side.

### Possible drawbacks

N/A

### Applicable issues

- fixes https://github.com/bitnami/charts/issues/18924

### Additional information

N/A

### Checklist

### DIFF
--- a/bitnami/redis-cluster/Chart.yaml
+++ b/bitnami/redis-cluster/Chart.yaml
@@ -32,4 +32,4 @@ maintainers:
 name: redis-cluster
 sources:
   - https://github.com/bitnami/charts/tree/main/bitnami/redis-cluster
-version: 9.0.4
+version: 9.0.5
--- a/bitnami/redis-cluster/README.md
+++ b/bitnami/redis-cluster/README.md
@@ -237,33 +237,33 @@ The command removes all the Kubernetes components associated with the chart and 
 
 ### Cluster update job parameters
 
-| Name                                  | Description                                                                                                     | Value          |
-|---------------------------------------|-----------------------------------------------------------------------------------------------------------------|----------------|
-| `updateJob.activeDeadlineSeconds`     | Number of seconds the Job to create the cluster will be waiting for the Nodes to be ready.                      | `600`          |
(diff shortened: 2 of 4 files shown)


rafariossaa (approved):
LGTM

nan-coupa:
@ritmas what do you set the configuration to for ArgoCD? Looks like Argo only maps to pre-sync/post-sync, so unclear how you resolved it.
https://argo-cd.readthedocs.io/en/stable/user-guide/helm/#helm-hooks

alcidesmig:
> @ritmas what do you set the configuration to for ArgoCD deployments? Looks like Argo only maps to pre-sync/post-sync, so unclear how you resolved it. https://argo-cd.readthedocs.io/en/stable/user-guide/helm/#helm-hooks

I couldn't solve it either. When mapping it to pre-sync, it doesn't solve the problem because it runs before the pods change; in post-sync, it never gets executed because the sync action never ends - because it isn't healthy.

ritmas:
@nan-coupa @alcidesmig This PR didn't fixed the issue with ArgoCD. As you've said, ArgoCD has less hooks than Helm and those cannot cope with [bitnami/redis-cluster](https://github.com/bitnami/charts/tree/main/bitnami/redis-cluster) management fluently.

In my experience, cluster looses data, is not accessible for a long time or just corrupts itself. Anyhow, it's not production ready, so I had to ditch it eventually and use manual way.

PR #19238, opened 2023-09-11
[bitnami/postgresql] Add SecurityContext to CronJob of backup

robson90:
Description of the change

Added the possibility to change/add the attributes to securityContext of CronJob.

Benefits

The CronJob in version 12.11.1 didn't have enough permissions. So it was not possible to write to the Volume. Now it is.

Possible drawbacks

More configuration is present and naming is a little overused.

Applicable issues

none found (I didn't open one.)
Additional information

I searched the issues for someone, who has the same issue. Didn't find anything. I searched inside the tests for some tests regarding the CronJob, nothing. So I did fix it myself and now want to give it to the community

### Checklist


andresbono (commented):
Thank you so much for your contribution! While I think your change totally makes sense, what if we try to homogenize the securityContext for pods with the existing ones for bitnami/postgresql? Basically:

- Call the value `podSecurityContext.
- Add a `enabled` parameter.

You can search for `podSecurityContext` to follow the pattern.

bitnami/postgresql/values.yaml
```diff
@@ -1099,6 +1099,11 @@ backup:
     ttlSecondsAfterFinished: ""
     ## @param backup.cronjob.restartPolicy Set the cronjob parameter restartPolicy
     restartPolicy: OnFailure
+    ## @param backup.cronjob.podSecurityContext.enabled Enable PodSecurityContext for CronJob/Backup
+    ## @param backup.cronjob.podSecurityContext.fsGroup Group ID for the CronJob
+    podSecurityContext:
+      enabled: false
```
andresbono, line 1105: Why are you setting this to `false` by default? Is it for backwards compatibility? Do you have any use case where setting it to `true` could be problematic?

bitnami/postgresql/values.schema.json
andresbono, line 1: These changes were reverted: #19335. Could you please rebase origin/main again?

bitnami/postgresql/Chart.yaml
```diff
@@ -35,4 +35,4 @@ maintainers:
 name: postgresql
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/postgresql
-version: 12.11.3
+version: 12.11.4
```
andresbono, line 38: After rebasing origin/main, can you please bump the version again?

If the conflict happens again before we finally merge the PR, we will do the bump ourselves.

bitnami/postgresql/values.yaml
robson90: I don't think it's a problem to have it enabled by default. In fact the CronJob didn't even work for me, when this setting weren't set. 

I will enable it by default.

andresbono (approved):
LGTM. I have just rebased from `origin/main` and I have fixed a conflict. The PR is ready to go. Thank you!

robson90:
@andresbono is it possible that you trigger an automatic dependency upgrade on the Keycloak Chart? To use the newest Postgresql Chart.

PS: It was a pleasure to contribute 🥇

carrodher:
> @andresbono is it possible that you trigger an automatic dependency upgrade on the Keycloak Chart? To use the newest Postgresql Chart.
> 
> PS: It was a pleasure to contribute 🥇

We just forced a new bitnami/keycloak release, it should be released in a while depending on our testing pipeline load.

PR #19261, opened 2023-09-13
[bitnami/common] new macro to checksum config resources

### LINKED ISSUE #18376: [bitnami/common] new macro to evaluate a checksum of configmap/secrets
### Name and Version

bitnami/common

in https://github.com/bitnami/charts/pull/17302 https://github.com/bitnami/charts/pull/17579, I patched the checksum calculation of pod annotations that are used to trigger a pod rollout.

the pattern described in [helm docs](https://helm.sh/docs/howto/charts_tips_and_tricks/#automatically-roll-deployments) has the issue that it will trigger a reload of the pods when the chart version change, even it the pod template and configmap data are not changed. 

This happens because the version of the chart is included in "helm.sh/chart" in metadata.labels (printf "%s-%s" .Chart.Name .Chart.Version) of the configmap and the checksum is done on the whole file.

instead of copy/pasting the snippets from my PRs in other charts, I suggest a new macro to checksum only the data part of the ConfigMap/Secret, something like:

```
{{ define "common.utils.checksumData" }}
{{ omit ( include (print $.Template.BasePath .file ) . | fromYaml ) "apiVersion" "kind" "metadata" | toYaml | sha256sum }}
{{ end }}

{{ include "common.utils.checksumData" (dict "file" "/health-configmap.yaml") }}
```

this works only if there is one resource/document in the file (fromYaml does not support multiple documents).

I'll do the PR if the idea is found useful .

_No response_

jouve:
### Description of the change

The pattern to rollout workloads on config change, described in [helm docs](https://helm.sh/docs/howto/charts_tips_and_tricks/#automatically-roll-deployments), has the issue that it will trigger a reload of the pods when the chart version change, even it the pod template and configmap data are not changed.

This happens because the version of the chart is included in "helm.sh/chart" in metadata.labels (printf "%s-%s" .Chart.Name .Chart.Version) of the configmap and the checksum is done on the whole file.

This new macro calculates the checksum by omitting metadata.

### Applicable issues

- fixes #18376

### Checklist

### DIFF
--- a/bitnami/common/templates/_utils.tpl
+++ b/bitnami/common/templates/_utils.tpl
@@ -67,11 +67,11 @@ Usage:
 {{- end -}}
 
 {{/*
-Checksum a template containing a *single* resource (ConfigMap,Secret) for use in pod annotations, excluding the metadata (see #18376).
+Checksum a template at "path" containing a *single* resource (ConfigMap,Secret) for use in pod annotations, excluding the metadata (see #18376).
 Usage:
-{{ include "common.utils.checksum" (dict "template" "/configmap.yaml" "context" $) }}
+{{ include "common.utils.checksumTemplate" (dict "path" "/configmap.yaml" "context" $) }}
 */}}
-{{- define "common.utils.checksum" -}}
-{{- $obj := include (print .context.Template.BasePath .template) .context | fromYaml -}}
+{{- define "common.utils.checksumTemplate" -}}
+{{- $obj := include (print .context.Template.BasePath .path) .context | fromYaml -}}
 {{ omit $obj "apiVersion" "kind" "metadata" | toYaml | sha256sum }}
 {{- end -}}
--- a/bitnami/common/templates/_utils.tpl
+++ b/bitnami/common/templates/_utils.tpl
@@ -65,3 +65,13 @@ Usage:
 {{- end -}}
 {{- printf "%s" $key -}} 
 {{- end -}}
+
+{{/*
+Checksum a template containing a *single* resource (ConfigMap,Secret) for use in pod annotations, excluding the metadata (see #18376).
+Usage:
+{{ include "common.utils.checksum" (dict "template" "/configmap.yaml" "context" $) }}
+*/}}
+{{- define "common.utils.checksum" -}}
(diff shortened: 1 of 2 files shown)


bitnami/common/templates/_utils.tpl
```diff
@@ -65,3 +65,13 @@ Usage:
 {{- end -}}
 {{- printf "%s" $key -}} 
 {{- end -}}
+
+{{/*
+Checksum a template containing a *single* resource (ConfigMap,Secret) for use in pod annotations, excluding the metadata (see #18376).
+Usage:
+{{ include "common.utils.checksum" (dict "template" "/configmap.yaml" "context" $) }}
+*/}}
+{{- define "common.utils.checksum" -}}
```
jotamartos, line 74: To be in sync with other functions defined here, I'd use a different name for the function (getSHA256sum, getTemplateSHA256sum, ... ) to describe better what the function does. WDTY?

jotamartos (commented):
Thank you for your contribution, please take a look at my comment.

bitnami/common/templates/_utils.tpl
jouve: let's try [checksum](https://en.wiktionary.org/wiki/checksum#Verb)Template (sha256 is an implementation detail and does not need to be exposed).
The symmetry in *checksum*/xxx et common.utils.*checksumTemplate*  looks neat and it makes the lines around 120 chars.
```yaml
      annotations:
        checksum/health: {{ template "common.utils.checksumTemplate" (dict "path" "/health-configmap.yaml" "context" .) }}
        checksum/scripts: {{ template "common.utils.checksumTemplate" (dict "path" "/scripts-configmap.yaml" "context" .) }}
```

jotamartos: Great! That's a good name for the function too! 👍

jotamartos (approved):

jotamartos:
@jouve, once the new version of the common chart is available, could you please take care of updating the required charts to use this new function? 

Thanks

LeoColomb:
@jouve (@jotamartos) Is there a specific reason for the helper not being used, or “simply” lack of resource/interest?

jouve:
I use it a lot for private charts, as of this repo, I hope to find some time soon :)

PR #19263, opened 2023-09-13
[bitnami/magento] Fix magentoHost when using Ingress #19262

### LINKED ISSUE #19231: [bitnami/magento] Magento hostname when using ingress is wrong.
### Name and Version

bitnami/magento 22.1.6

amd64

1. Install the Magento chart.
2. Open the hostname defined in the ingress.
3. The application loads but if we check the browser console we see a lot of errors because the hostname configured in Magento includes the port defined in the service.

```yaml
ingress:
  enabled: true
  hostname: magento.example.com
```

The base_url used by Magento should be configured properly to "http://magento.example.com" or "https://magentoexample.com".

### What do you see instead?

I see many errors in the browser console and the application is not loading correctly.

The browser is trying to retrieve data from "magento.example.com:8080" (Trying to use the service HTTP port).

### Additional information

The only way I was able to configure it is by setting the service port to 80.

```yaml
service:
  ports:
    http: 80
```
This works for HTTP.

The problem is that there is no way to use HTTPS in the ingress since the application will try to connect from the browser to `http://magento.example.com` (port 80). 

From the docker image there is no way to have all the traffic go to HTTPS (So this is also an issue with the image).

https://github.com/bitnami/containers/blob/99465ed8b18a00231cce2698101000ae055b844c/bitnami/magento/2/debian-11/rootfs/opt/bitnami/scripts/libmagento.sh#L515

dgomezleon:
### Description of the change

It uses ports 80 and 443 as default ports for the service. It also adds the use of `magentoHost`.

### Benefits

Fix usage with ingress.

### Possible drawbacks

N/A

### Applicable issues

- fixes #19231

### Additional information

N/A

### Checklist

### DIFF
--- a/bitnami/magento/Chart.yaml
+++ b/bitnami/magento/Chart.yaml
@@ -48,4 +48,4 @@ maintainers:
 name: magento
 sources:
   - https://github.com/bitnami/charts/tree/main/bitnami/magento
-version: 23.1.1
+version: 23.1.2
--- a/bitnami/magento/README.md
+++ b/bitnami/magento/README.md
@@ -261,8 +261,8 @@ The command removes all the Kubernetes components associated with the chart and 
 | Name                               | Description                                                                                                                      | Value                    |
 | ---------------------------------- | -------------------------------------------------------------------------------------------------------------------------------- | ------------------------ |
 | `service.type`                     | Kubernetes Service type                                                                                                          | `LoadBalancer`           |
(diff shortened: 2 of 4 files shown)


FraPazGal (approved):
LGTM!

rafariossaa (dismissed):
LGTM

FraPazGal (approved):

PR #19310, opened 2023-09-17
[bitnami/nginx] add `cloneStaticSiteFromGit.extraEnvVarsSecret`

smehboub:
### Description of the change

 Add `cloneStaticSiteFromGit.extraEnvVarsSecret` value

### Benefits

Allow to use credentials as secret mounted as variables to clone the content of a private repository by using a token instead of a ssh key.
With this, we can use cloneStaticSiteFromGit.repository: "https://oauth2:${token}@github.com/example/test.git" by pickup the token from a secret.

### Possible drawbacks

None

### Checklist

### DIFF
--- a/bitnami/nginx/values.yaml
+++ b/bitnami/nginx/values.yaml
@@ -452,6 +452,9 @@ cloneStaticSiteFromGit:
   ## @param cloneStaticSiteFromGit.interval Interval for sidecar container pull from the Git repository
   ##
   interval: 60
+  ## @param cloneStaticSiteFromGit.extraEnvVarsSecret Secret with extra environment variables
+  ##
+  extraEnvVarsSecret: ""
   ## Additional configuration for git-clone-repository initContainer
   ##
   gitClone:
@@ -461,9 +464,6 @@ cloneStaticSiteFromGit:
     ## @param cloneStaticSiteFromGit.gitClone.args Override default container args for git-clone-repository
     ##
     args: []
-    ## @param cloneStaticSiteFromGit.gitClone.extraEnvVarsSecret Secret with extra environment variables
-    ##
-    extraEnvVarsSecret: ""
   ## Additional configuration for the git-repo-syncer container
   ##
   gitSync:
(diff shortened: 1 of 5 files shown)


smehboub:
Hello @andresbono,

Kindly please, is it possible to have any news about this MR ?

Thanks in advance.
Rgs,

bitnami/nginx/values.yaml
```diff
@@ -461,6 +461,9 @@ cloneStaticSiteFromGit:
     ## @param cloneStaticSiteFromGit.gitClone.args Override default container args for git-clone-repository
     ##
     args: []
+    ## @param cloneStaticSiteFromGit.gitClone.extraEnvVarsSecret Secret with extra environment variables
+    ##
+    extraEnvVarsSecret: ""
```
andresbono, line 466: For the use case that you described, it seems the token would be only necessary in the `git-clone-repository` initContainer, right? I guess that is why you included the new value under `gitClone`.

But I wonder if it wouldn't be more convenient to generalize `extraEnvVarsSecret` and set it under `cloneStaticSiteFromGit.extraEnvVarsSecret`, adding it for the 2 containers: `git-clone-repository` and `git-repo-syncer`. That would give the value more flexibility for different use cases that other users may come up with.

What do you think?

andresbono (commented):
Thank you so much for your contribution! Could you please check my comments?

smehboub:
> Thank you so much for your contribution! Could you please check my comments?

Thank you for your return!
I commited your proposition :-)

bitnami/nginx/values.yaml
smehboub: Yes it's a good idea, for the value flexibility for different use case.
I will make the change...

smehboub: The change is done ;-)

andresbono (approved):
LGTM thank you very much!

I made a minor change (moving the value below `extraEnvVars`).

PR #19598, opened 2023-09-28
[bitnami/appsmith] Create options for local git repo data on persistent storage

### LINKED ISSUE #19258: Appsmith - Problem with connecting the GIT repository
### Name and Version

bitnami/appsmith

amd64

1 Run Appsmith using bitnami charts
2 Create a sample application in Appsmith
3 Try connecting to a GIT repository using Appsmith's built-in interface (button on the bottom left bar)

After this, Appsmith attempts to create a local GIT repository in the /data resource. Unfortunately, the bitnami deployment does not provide such a resource, and Appsmith does not have the ability to create the directory.

In the official bitnami repository, this directory is mounted on a resource inside the appsmith-stacks directory.

I used the parameters, but also tested on a pure chart straight from the bitnami repository.
I ran on localhost, on kubernetes and on openshift - the same in all cases.

Correct connection to the GIT repository.

### What do you see instead?

Error about no access to /data directory.

### Additional information

I managed to get the mechanism working by mounting an additional resource as extraVolumes in the form of emptyDir, but I don't know if this is the right solution.
Maybe a PVC should be considered, or analysed to use one of the PVCs already created.

jliedy:
### Description of the change

Currently on an install with default values, when storage persistence is set for the backend deployment apps try to save the local git repo under /data which does not exist.  This change will add functionality to set the APPSMITH_GIT_ROOT environment variable for the backend deployment in a way that does not disrupt current deployments.  This environment variable sets the destination of the local git repository data when the app is connected to a git repository.

### Benefits

This makes it a bit more obvious that this "gotcha" exists and offers a way to get this set "automagically" without disrupting current installs.

### Possible drawbacks

Since the default value for this is empty, folks will probably not set this until they run into the snag in their current deployment.  I just can't think of a good way to easily do this that doesn't muck up current installs.

### Applicable issues

- fixes #19258 

### Checklist

### DIFF
--- a/bitnami/appsmith/Chart.yaml
+++ b/bitnami/appsmith/Chart.yaml
@@ -37,4 +37,4 @@ maintainers:
 name: appsmith
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/appsmith
-version: 2.0.2
+version: 2.0.3
--- a/bitnami/appsmith/README.md
+++ b/bitnami/appsmith/README.md
@@ -269,18 +269,19 @@ The command removes all the Kubernetes components associated with the chart and 
 
 ### Backend Persistence Parameters
 
-| Name                                | Description                                                                                             | Value               |
-| ----------------------------------- | ------------------------------------------------------------------------------------------------------- | ------------------- |
-| `backend.persistence.enabled`       | Enable persistence using Persistent Volume Claims                                                       | `true`              |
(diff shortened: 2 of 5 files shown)


jliedy:
Reopening after trying to figure out how to pull in changes due to a conflict that came from bumping the chart version, but ended up making a mess trying to clean up after myself.  Git is weird.

aoterolorenzo (approved):
LGTM

PR #19702, opened 2023-10-03
[bitnami/common] update imagePullSecrets to handle map and list format

greenday9:
[bitnami/common] update imagePullSecrets to handle map and list format

### Description of the change

This change allows imagePullSecrets to be either in list or map format.

```yaml
global:
  imagePullSecrets:
    - myRegistryKeySecretName
```

or

```yaml
global:
  imagePullSecrets:
    - name: myRegistryKeySecretName
```

### Benefits

Other charts use the map format and even helm create uses this format when generating a skeleton.  Having the flexibility to support both formats allows users to create umbrellas combining other charts with bitnami charts even if the other charts support map rendering of imagePullSecrets.

### Checklist


bitnami/common/Chart.yaml
```diff
@@ -23,4 +23,4 @@ name: common
 sources:
   - https://github.com/bitnami/charts
 type: library
-version: 2.13.0
+version: 2.13.1
```
migruiz4, line 26: Could you please bump a new version? 

Version 2.13.1 has already been released: https://github.com/bitnami/charts/blob/main/bitnami/common/Chart.yaml

migruiz4:
Thank you very much for your contribution! Changes look good to me, but a new version bump is required before I can approve it

bitnami/common/Chart.yaml
jouve:
```suggestion
version: 2.14.0
```

> Given a version number MAJOR.MINOR.PATCH, increment the:
>
>    MAJOR version when you make incompatible API changes
>    __MINOR version when you add functionality in a backward compatible manner__
>    PATCH version when you make backward compatible bug fixes

migruiz4: @jouve This is not really a new feature such as adding a new helper, but a patch for an existing one

greenday9: Thanks @migruiz4 , @jouve .  I've updated to 2.13.2.

migruiz4 (approved):

PR #20001, opened 2023-10-10
[bitnami/external-dns] ability to disable validation

### LINKED ISSUE #19942: [bitnami/external-dns] unable to disable validation on external-dns
### Name and Version

bitnami/external-dns:6.23.3

amd64

Using ArgoCD with the ArgoCD Vault Plugin, i'm unable to bypass the validation of the external-dns chart.

My use case is that we pass in the path to the secret location inside of Hashicorp Vault and then argocd-vault-plugin does a `helm template --values values.yaml | argocd-vault-plugin generate -` to substitute the path out of the yaml files into the correct values.

Which then ARGO will apply to the cluster.

https://github.com/argoproj-labs/argocd-vault-plugin

```yaml
aws:
  region: <path:omnibus/data/cluster123/external-dns#awsRegion>
  assumeRoleArn: <path:omnibus/data/cluster123/external-dns#assumeRoleArn>
serviceAccount:
  name: <path:omnibus/data/cluster123/external-dns#serviceAccountName>
  annotations:
    eks.amazonaws.com/role-arn:  <path:omnibus/data/cluster123/external-dns#serviceAccountArn>
txtOwnerId: <path:omnibus/data/cluster123/external-dns#txtOwnerId>
domainFilters: [<path:omnibus/data/cluster123/external-dns#domainFilters>]
sources: [<path:omnibus/data/cluster123/external-dns#sources>]
```

Outputs the yaml with what was passed in, that way the argocd-vault-plugin can substitute out the values correctly. 

### What do you see instead?

Error: execution error at (external-dns/charts/external-dns/templates/NOTES.txt:11:3): 
VALUES VALIDATION:\nexternal-dns: aws.assumeRoleArn
    The AWS Role to assume must follow ARN format: `arn:aws:iam::123455567:role/external-dns`
    Ref: https://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html
    Please set a valid ARN (--set aws.assumeRoleARN="xxxx")

### Additional information

It would be nice to have a flag that could be set to disable the validation, so more advance tools like this could be used.

Thanks,

jwhitcraft:
### Description of the change

There are some uses cases where the chart value validation should not be ran, this provides a way to disable it.

### Benefits

When running inside of ArgoCD with the Vault plugin you need to pass in the path to the param inside of vault, which causes the validation to fail.

### Possible drawbacks

Someone could disable the validation and not know why things are not working.

### Applicable issues

- fixes #19942

### Checklist


jwhitcraft:
@FraPazGal any movement this?

bitnami/external-dns/templates/_helpers.tpl
```diff
@@ -308,10 +308,14 @@ Compile all warnings into a single message, and call fail.
...
 {{- $message := join "\n" $messages -}}
 
+{{- if .Values.validation.enabled -}}
 {{- if $message -}}
 {{-   printf "\nVALUES VALIDATION:\n%s" $message | fail -}}
 {{- end -}}
 {{- end -}}
+{{- end -}}
+
+{{/*
```
FraPazGal, line 318: Is this a typo? These lines shouldn't be needed right?

bitnami/external-dns/Chart.yaml
```diff
@@ -28,4 +28,4 @@ maintainers:
 name: external-dns
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/external-dns
-version: 6.26.4
+version: 6.26.5
```
FraPazGal, line 31: Given this is a new feature, we should bump the chart's version to a new minor

bitnami/external-dns/values.yaml
```diff
@@ -1110,3 +1110,11 @@ metrics:
     ## @param  metrics.googlePodMonitor.endpoint The endpoint for Google Managed Prometheus scraping the metrics
     ##
     endpoint: /metrics
+
+
+## Chart Validation
```
FraPazGal, line 1115: This is just a nitpick on my part but I would put this param before the providers ones so it gets more visibility.

FraPazGal (commented):
Thanks for the PR @jwhitcraft! Please check my comments whenever you have a chance

bitnami/external-dns/values.yaml
jwhitcraft: Moved this up, thanks for the suggestion!

bitnami/external-dns/templates/_helpers.tpl
jwhitcraft: Removed!

FraPazGal (approved):
LGTM!

PR #20003, opened 2023-10-10
[bitnami/redis] Implementing an option to define replicas as Daemonsets

m1k3:
### Description of the change

With this change it is possible to define replicas as the `DaemonSet` kind (defaults to `StatefulSet`) similar to how the master can be defined as a `StatefulSet` or a `Deployment`. 

```yaml
replica:
  kind: DaemonSet
```

### Benefits

Defining replicas as a `DaemonSet` has the advantage of being able to run replicas directly on nodes so that applications can connect directly to a replica running on the same node using file sockets. The replica can be configured to expose the file socket

```yaml
commonConfiguration: |
  # create a unix domain socket to listen on
  unixsocket /host-tmp/redis.socket
  # set permissions for the socket
  unixsocketperm 775

replica:
  kind: DaemonSet
  extraVolumes:
    - name: host-tmp
      hostPath:
        path: /tmp
        type: Directory
  extraVolumeMounts:
    - mountPath: /host-tmp
      name: host-tmp
```

the client app can then mount the socket and connect directly to the replica instead of going through TCP connections. In low-latency scenarios, this saves on the overhead of going through the network.

### Possible drawbacks

This feature does not work when the `sentinel` option is enabled. The reason is that these replicas should always stay replicas. They are not intended to be promoted to a master so it doesn't make sense to run them in sentinel mode.

### Additional information

There are some other minor changes to the templates that include the following:

* ability to avoid deploying `master` resources when the master count is 0
* when the `replica.externalMaster.enabled` is true, the `REDIS_MASTER_HOST` and `REDIS_MASTER_PORT` environment variables are set to the `replica.externalMaster.(host|port)` values.

### Checklist


CeliaGMqrz:
Hi @m1k3 

Thanks for your contribution! 

Sorry for the delay. It seems that there are some conflicts. Can you please resolve them?

m1k3:
Hey @CeliaGMqrz,

looks like there were changes merged into main after I submitted the PR which caused the conflict 🙈 . I have pulled in the most up-to-date version of main resolving the conflict. Please have another look.

CeliaGMqrz:
Hi @m1k3 ,

Thanks for your contribution! This is a good feature, however, It seems that when you implement this feature in multinode, for example, using 3 nodes there are collisions due to the persistent volumes. 

```
Type     Reason            Age               From               Message
  ----     ------            ----              ----               -------
  Warning  FailedScheduling  59s               default-scheduler  0/3 nodes are available: 3 pod has unbound immediate PersistentVolumeClaims.
```

Maybe we should follow another approach to solve this issue.

m1k3:
Good catch. I simply replicated the PVC setup that was already present in to support the `master.kind: Deployment` without thinking too much about it. I also wasn't focusing that much on the PVC aspect since these (being replicas) don't really need persistence.

I did a little digging and discovered that this is actually not a completely new problem with the chart. The same problem is present with the current version `main` version of the chart if you try to use more than a single **master** when the `kind` is set to `Deployment`. This is not a problem when used with a StatefulSet because with a StatefulSet the PVC are managed by the StatefulSet instead of a separate PersistenceVolumeClaim. The problem is also present if one tries to use the `(replica|master).persistence.existingClaim` feature with multiple replicas even when used with a StatefulSet. There (again) K8s will refuse to bind the claim with multiple pods. Not sure if we should do anything about the existing volume claim problems 🤷 

As a a solution that works for this feature, it would make sense to avoid the claims altogether when used in DaemonSet mode. We would use an `emptyDir` the same as if persistence was "disabled". It's not completely disabled, because the data [survives](https://kubernetes.io/docs/concepts/storage/volumes/#emptydir) a container crash. It will only get wiped if the pod is removed from a node which only happens if the node is removed. Given these are supposed to only ever be replicas, this makes sense. I made the changes avoiding the use of the PVC when used in DaemonSet mode to this PR. What do you think?

CeliaGMqrz (approved):
Hi @m1k3,

Thanks for your detailed information. It appears to be a widespread issue in the chart and requires a review.  However, I agree with your approach for this feature as it makes sense. So, LGTM.

Thanks for your contribution!

PR #20074, opened 2023-10-11
[bitnami/schema-registry] Added customPodMetrics option to trigge…

ETisREAL:
I added the possibility to specify a list of custom pod metrics to trigger autoscaling.

This will allow users to have a more fine grained control on how to manage the autoscaling feature

As far as I can tell, I don't see any possible drawbacks from this.

The deployment was tested on a GKE cluster 1.27.3-gke.100 version


carrodher:
Thanks for your contribution! Is it necessary to bump the major version because backward compatibility is not guaranteed? Maybe it is enough to bump just the minor version as it is a new feature addition.

In the same way, please check the [failing action](https://github.com/bitnami/charts/actions/runs/6485594749/job/17612008940?pr=20074). It seems the metadata of the new parameters added in the _values.yaml_ is incorrect.

ETisREAL:
> Thanks for your contribution! Is it necessary to bump the major version because backward compatibility is not guaranteed? Maybe it is enough to bump just the minor version as it is a new feature addition.
> 
> In the same way, please check the [failing action](https://github.com/bitnami/charts/actions/runs/6485594749/job/17612008940?pr=20074). It seems the metadata of the new parameters added in the _values.yaml_ is incorrect.

Hi @carrodher I corrected the metadata issue and bumped down the chart version (since it doesn't affect compatibility).

rafariossaa (approved):
Thanks for adding this.
LGTM.

ETisREAL:
@rafariossaa Thank you guys for letting me add this to the default chart :) It comes really handy

PR #20493, opened 2023-10-29
[bitnami/rabbitmq-cluster-operator] - Make zap-logging flags easier to configure

shlomitubul:
**Description of the Change**
Currently, to customize the zap-logger, one has to override all the arguments. This PR aims to simplify the logging configuration by adding a 'logging' section for both the message and cluster operators. Additionally, the default log level is set to 'error,' which is more suitable for most environments. 
NOTE: This PR aims to be fully backward compatible.

**Benefits**

Possible drawbacks

Applicable issues

Additional information


shlomitubul:
Is there anything from my side that I need to do?

bitnami/rabbitmq-cluster-operator/templates/messaging-topology-operator/deployment.yaml
```diff
@@ -97,6 +97,10 @@ spec:
           {{- else }}
           args:
             - --metrics-bind-address=:{{ .Values.msgTopologyOperator.containerPorts.metrics }}
+            - --zap-devel={{ with .Values.msgTopologyOperator.logging }}{{ default "false" .development }}{{ else }}"false"{{ end }}
```
fmulero, line 100: Same here

bitnami/rabbitmq-cluster-operator/templates/cluster-operator/deployment.yaml
```diff
@@ -88,6 +88,10 @@ spec:
           {{- else }}
           args:
             - --metrics-bind-address=:{{ .Values.clusterOperator.containerPorts.metrics }}
+            - --zap-devel={{ with .Values.clusterOperator.logging }}{{ default "false" .development }}{{ else }}"false"{{ end }}
+            - --zap-encoder={{ with .Values.clusterOperator.logging }}{{ default "json" .encoder }}{{ else }}"json"{{ end }}
+            - --zap-log-level={{ with .Values.clusterOperator.logging }}{{ default "error" .logLevel }}{{ else }}"error"{{ end }}
+            - --zap-stacktrace-level={{ with .Values.clusterOperator.logging }}{{ default "error" .stacktraceLevel }}{{ else }}"error"{{ end }}
```
fmulero, line 94: I think this logic can be simplified, I think all these condition only makes sense when an user remove the logging configuration with `--set clusterOperator.logging=null`. In that case I think the following could be applied (because the user wants to remove the logging section):
```suggestion
            {{- if .Values.clusterOperator.logging }}
            - --zap-devel={{ .Values.clusterOperator.logging.development }}
            - --zap-encoder={{ .Values.clusterOperator.logging.encoder }}
            - --zap-log-level={{ .Values.clusterOperator.logging.logLevel }}
            - --zap-stacktrace-level={{ .Values.clusterOperator.logging.stacktraceLevel }}
            {{- end }}
```
I think this is little bit clearer than the initial approach. WDYT?

fmulero (changes requested):
Hi @shlomitubul thanks for your contribution and sorry about our late response.

I left a little suggestion, could you give it a glance? Don't hesitate to share any doubt or thought.

bitnami/rabbitmq-cluster-operator/Chart.yaml
```diff
@@ -37,4 +37,4 @@ maintainers:
 name: rabbitmq-cluster-operator
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/rabbitmq-cluster-operator
-version: 3.10.6
+version: 3.11.1
```
fmulero, line 40: There is also conflicts in the version. Please take it a glance.

fmulero (commented):
There some conflicts with the main branch

PR #20503, opened 2023-10-30
[bitami/rabbitmq] Added ability to enable/disable allocateLoadBalancerNodePorts

### LINKED ISSUE #20495: Add option to disable Load Balancer Node Port Allocation
### Name and Version

bitnami/rabbitmq 12.3.0

When using the LoadBalancer service type with the RabbitMQ helm chart, NodePorts are assigned for each service port. 

Being able to set `spec.allocateLoadBalancerNodePorts` to `false `would address this.

`service.allocateLoadBalancerNodePorts: false`

Creating a loadbalancer manually

belps96:
### Description of the change

Now allows the allocateLoadBalancerNodePorts attribute of the RabbitMQ service to be disabled (normally always enabled)

### Benefits

This will allow only the desired/specified ports to be allocated a NodePort 

### Possible drawbacks

None

### Applicable issues

- fixes #20495

### Additional information
 None

### Checklist


bitnami/rabbitmq/Chart.yaml
```diff
@@ -30,4 +30,4 @@ maintainers:
 name: rabbitmq
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/rabbitmq
-version: 12.3.0
+version: 12.3.1
```
joancafom, line 33: As this is a new feature, it should be a minor bump :) 
```suggestion
version: 12.4.0
```

bitnami/rabbitmq/templates/svc.yaml
```diff
@@ -26,6 +26,9 @@ spec:
   {{- if .Values.service.loadBalancerSourceRanges }}
   loadBalancerSourceRanges: {{- toYaml .Values.service.loadBalancerSourceRanges | nindent 4 }}
   {{- end }}
+  {{- if .Values.service.allocateLoadBalancerNodePorts }}
+  allocateLoadBalancerNodePorts: {{ .Values.service.allocateLoadBalancerNodePorts }}
+  {{- end }}
```
joancafom, line 31: This only happens when service is `LoadBalancer`, you will need to add a check here as done in L22.

joancafom (changes requested):
Thanks for your PR! I left you some comments as a review

bitnami/rabbitmq/templates/svc.yaml
belps96: Is this not nested within L22's if condition? Notice the {{- end }} at line 32

joancafom: You are right! I missed the `{{-end} }` at line 32 :) Just a follow-up suggestion, I don't think we need the conditional here, .if `Values.service.allocateLoadBalancerNodePorts=false` then it's okay to have `allocateLoadBalancerNodePorts=false`  (once we have ensured the service type is LoadBalancer)

```suggestion
  allocateLoadBalancerNodePorts: {{ .Values.service.allocateLoadBalancerNodePorts }}
```

joancafom:
Just a small follow-up comment to go 😁

joancafom (approved):

PR #20409, opened 2023-10-25
[bitnami/redis] - Add support for PodMonitor

peplc:
**Description of the change**
Adding PodMonitor support for redis chart

**Benefits**
Removing the need for ClusterIP service that solely exist for gathering metrics

**Possible drawbacks**

**Applicable issues**

**Additional information**

PR #20517, opened 2023-10-30
[bitnami/redis] Disable metrics-svc with podMonitor

peplc:
**Description of the change** 
In addition to [#20409 ](https://github.com/bitnami/charts/pull/20409) not enable metrics-svc with podMonitor

**Benefits** 
Removing the need for ClusterIP service when podMonitor enabled

**Possible drawbacks**

**Applicable issues**

**Additional information**


peplc:
Hi @jotamartos. can you check it please? thanks

jotamartos:
Hi @peplc,

Sorry for the delay in reviewing this Pull Request. First of all, thank you for your contribution and for taking the time to propose these changes. 

The metrics service can be enabled independently of the creation of the ServiceMonitor or PodMonitor objects. With your changes, the metrics' service can't be created if you do not enable ServiceMonitor (which is disabled by default) and you won't be able to access the necessary Redis' (the server itself) metrics. 

With the current configuration, if you enable the metrics' service, you will see that there are 2 containers in each pod. The second container is the metrics one and it gets information from the Redis server. You can obtain that information by querying the port 9121.

```
$ curl localhost:9121/metrics
# HELP go_gc_duration_seconds A summary of the pause duration of garbage collection cycles.
# TYPE go_gc_duration_seconds summary
go_gc_duration_seconds{quantile="0"} 0
go_gc_duration_seconds{quantile="0.25"} 0
go_gc_duration_seconds{quantile="0.5"} 0
go_gc_duration_seconds{quantile="0.75"} 0
go_gc_duration_seconds{quantile="1"} 0
go_gc_duration_seconds_sum 0
go_gc_duration_seconds_count 0
# HELP go_goroutines Number of goroutines that currently exist.
# TYPE go_goroutines gauge
go_goroutines 6
...
```

As I mentioned above, this information is independent of the ServiceMonitor or PodMonitor objects creation and the service must be enabled to be able to access it.

peplc:
Thanks @jotamartos. I updated it with not enable metrics-svc with podMonitor

jotamartos (changes requested):
Hi again @peplc,

The metrics service is required for the podMonitor metrics to work. Please note that these metrics are exposed through the metrics service http-port (https://github.com/bitnami/charts/blob/main/bitnami/redis/templates/metrics-svc.yaml#L33) and if it's not enabled, you won't be able to access the metrics.

peplc:
ok @jotamartos. i was wrong. thanks for your time

PR #20555, opened 2023-10-31
[bitnami/postgresql] Fix readiness probe

### LINKED ISSUE #20434: postgresql readiness probe exec - second command is never run
### Name and Version

bitnami/postgresql 13.1.5

amd64

1. Minikube
2. default values
3. `helm install postgresql --version 13.1.5 bitnami/postgresql`

Readiness check which contains two commands check pg_isready and for file as well

```yaml
exec:
  command:
    - /bin/sh
    - -c
    - -e
    - |
      exec pg_isready -U "test" -d "dbname=boom" -h 127.0.0.1 -p 5432
      [ -f /opt/bitnami/postgresql/tmp/.initialized ] || [ -f /bitnami/postgresql/.initialized ]
```

### What do you see instead?

Second command `[ -f /opt/bitnami/postgresql/tmp/.initialized ] || [ -f /bitnami/postgresql/.initialized ]` is never be run.

### Additional information

To reproduce it a large sql dump can be put to /docker-entrypoint-initdb.d directory. During the restoring process check the pod readiness status. It will be ready. This shouldn't happen if the second command is used in readinessProbe.

By the way, this [if statement](https://github.com/bitnami/charts/blob/d6092dc781b59106e434fa591c94e26494fa2f07/bitnami/postgresql/templates/_helpers.tpl#L303) could be put under values logic, not just contains check. As an example, I built my own image based on the bitnami postgresql image and put it into my private docker registry and as a result, I don't have this initialization readiness check.

joancafom:
### Description of the change

The default readiness probe consists of two commands that should be executed sequentially: _pg_isready_ and a file existence check (_bitnami/postgresql/.initialized_). Only the first command was being executed due to `exec`, which replaces the current PID and hence, skips following commands.

### Benefits

Default readiness probe is executed as intended.

### Possible drawbacks

None

### Applicable issues

- fixes #20434 

### Additional information

It is safe to exclude `exec` from `pg_isready`, as the readiness command is based of a shell script that will exit if any of the commands fail (`-e` option).

```
$ helm template postgresql bitnami/postgresql
...
            exec:
              command:
                - /bin/sh
                - -c
                - -e
                - |
                  pg_isready -U "postgres" -h 127.0.0.1 -p 5432
                  [ -f /opt/bitnami/postgresql/tmp/.initialized ] || [ -f /bitnami/postgresql/.initialized ]
```

### Checklist


migruiz4 (approved):
LGTM!

Issue #20564, opened 2023-10-31
Readiness and liveness probes randomly fail without cause


bramk:
### Name and Version

bitnami/zookeeper 12.1.4

### What architecture are you using?

amd64

### What steps will reproduce the bug?

On EKS with a statefulset with 3 replicas the liveness and readiness probes fail ever now and then.

This blog details the issue/solution: https://wolverineav.blog/tech/zookeeper-4-letter-word-netcat-no-response/

### Are you using any custom parameters or values?

No

### What is the expected behavior?

Under normal operation, every `ruok` call should receive an `imok` response causing the probes to report healthy.

### What do you see instead?

Every now and then there is no response which results in the probes to reported as unhealthy.

### Additional information

Adding `-q 1` to the nc command resolves the issue.

```yaml
      customLivenessProbe: |
        failureThreshold: 6
        initialDelaySeconds: 30
        periodSeconds: 10
        successThreshold: 1
        timeoutSeconds: 5
        exec:
          command: ['/bin/bash', '-c', 'echo "ruok" | timeout 2 nc -w 2 -q 1 localhost 2181 | grep imok']
      customReadinessProbe: |
        failureThreshold: 6
        initialDelaySeconds: 5
        periodSeconds: 10
        successThreshold: 1
        timeoutSeconds: 5
        exec:
          command: ['/bin/bash', '-c', 'echo "ruok" | timeout 2 nc -w 2 -q 1 localhost 2181 | grep imok']
```

carrodher:
Thank you for bringing this issue to our attention. We appreciate your involvement! If you're interested in contributing a solution, we welcome you to create a pull request. The Bitnami team is excited to review your submission and offer feedback. You can find the contributing guidelines [here](https://github.com/bitnami/charts/blob/main/CONTRIBUTING.md).

Your contribution will greatly benefit the community. Feel free to reach out if you have any questions or need assistance.

bramk:
Hi @carrodher, thank you for your response. Opened my first bitnami PR.

carrodher:
Thank you for opening this issue and submitting the associated Pull Request. Our team will review and provide feedback. Once the PR is merged, the issue will automatically close.

Your contribution is greatly appreciated!

PR #20572, opened 2023-10-31
[bitnami/zookeeper] Fix: randomly failing readiness/liveness probes

bramk:
### Description of the change

Delay readiness/liveness probe netcat calls sending FIN by 1 second to prevent occasional empty response to ruok request.

### Benefits

No more occasionally failing readiness/liveness probes

### Possible drawbacks

OSx BSD netcat does not support the q option. Is that a problem?

### Applicable issues

- fixes bitnami/charts#20564

### Additional information

I do not think it makes sense to make the dealyt configurable.

### Checklist


fmulero (commented):
Hi @bramk, thanks a lot for your contribution!

QQ, Does it have any impact in `probeCommandTimeout` ? Here you are adding a delay of 1s and the default value for the `probeCommandTimeout` is 2s. Should it be increased in 1 second also?

bramk:
Hi @fmulero 

> QQ, Does it have any impact in probeCommandTimeout ? Here you are adding a delay of 1s and the default value for the probeCommandTimeout is 2s. Should it be increased in 1 second also?

Hard to say.. 

I do not think the `-w 2` is affected by `-q 1`  and under normal circumstances zookeeper should still respond within single digit millisecond range anyway. Also because `.Values.probeCommandTimeout` is already used for both the `timeout` and the `-w` I guess it  is just about a sane default. Therefore IMHO 2 secs is still fine. In any case I see a major improvement over the old situation on our systems.

fmulero (approved):
LGTM

mveitas:
What would the process be to get this backported to an older version of the zookeeper chart. I am using `11.4.11` and running into this issue

mveitas:
I just added this to our existing chart as an override using the `customLivenessProbe` and the pods are restarting immediately. I am running with an older version of the chart (11.4.11)

```
  customLivenessProbe:
    exec:
      command:
        - /bin/bash
        - -c
        - echo "ruok" | timeout 2 nc -w 2 -q 1 localhost 2181 | grep imok
  customReadinessProbe:
    exec:
      command:
        - /bin/bash
        - -c
        - echo "ruok" | timeout 2 nc -w 2 -q 1 localhost 2181 | grep imok
```

PR #20675, opened 2023-11-07
[bitnami/charts] GitHub actions hardening

fmulero:
### Description of the change

* Use SHA reference for GitHub actions (`vmware-labs/vmware-image-builder-action` is an exception).
* Avoid the use of github context substitution in scripts to prevent script injections.
* Remove third-party actions that can be easily implemented with actions or scripts provided by GitHub.
* `aws-cli` installation and configuration is removed, this step is no needed in `ubuntu-latest`
* Include dependabot rules to update periodically the actions.
* Keep updated the actions.

### Benefits

Apply GitHub best practices, reduce attack surface and risks related to the use of third party actions.

### Possible drawbacks

Some actions has been updated to its latest major version. These actions are:
* apache/skywalking-eyes/header
* pozil/auto-assign-issue
* actions/stale

Due to that we could face some issues in the execution of the workflows:
* https://github.com/bitnami/charts/actions/workflows/license-headers.yml
* https://github.com/bitnami/charts/actions/workflows/moving-cards.yml
* https://github.com/bitnami/charts/actions/workflows/stale.yml
* https://github.com/bitnami/charts/actions/workflows/clossing-issues.yml

### Additional information

* [Security hardening for GitHub Actions](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions)

### Checklist

### DIFF
--- a/.github/workflows/ci-pipeline.yml
+++ b/.github/workflows/ci-pipeline.yml
@@ -81,7 +81,7 @@ jobs:
       - id: show-error
         name: Show error
         if: ${{ steps.get-chart.outputs.result != 'ok' }}
-        uses: actions/github-script@v6
+        uses: actions/github-script@d7906e4ad0b1822421a7e6a35d5ca353c962f410
         with:
           script: |
             let message='${{ steps.get-chart.outputs.error }}';
@@ -106,7 +106,7 @@ jobs:
       )
     name: VIB Verify
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11
         name: Checkout Repository
         with:
           ref: ${{ github.event.pull_request.head.ref }}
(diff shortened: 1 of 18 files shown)


.github/workflows/ci-pipeline.yml
```diff
@@ -157,13 +157,21 @@ jobs:
...
       - name: PR approval
         if: ${{ needs.vib-verify.result == 'success' }}
-        uses: hmarr/auto-approve-action@v3.0.0
+        uses: actions/github-script@d7906e4ad0b1822421a7e6a35d5ca353c962f410
         with:
-          pull-request-number: ${{ github.event.number }}
+          result-encoding: string
+          retries: 3
+          script: |
+            github.rest.pulls.createReview({
```
joancafom, line 165: Probably you have tested it already (and in case you don't because of permissions, we should be extra-careful when landing this), but sometimes in the past we have faced limitation in order to approve a PR because of the actor (i.e. the creator of the PR can't approve its own PR, so we had to use two different tokens). 

I don't see any mention to tokens here (neither before the change, but maybe it was something intrinsic of the action we used), but just as a reminder to us :)

joancafom (approved):

.github/workflows/ci-pipeline.yml
fmulero: Thanks a lot for the heads up, that's a good one. Automated PRs are created by `bitnami-bot` so the PR should be  approved by other user (this is configured in our repo to avoid self approvals) In this piece of code there is no reference to the token because it uses by default GitHub Actions user and we are allowing write permissions to pull-requests in that job: https://github.com/bitnami/charts/blob/c4b617b6b0ce71e438994bd35864a107f3026b44/.github/workflows/ci-pipeline.yml#L147-L148

Issue #19583, opened 2023-09-27
Chart doesn't define the proper API Public URL


jsmueller7:
### Name and Version

bitnami/supabase 0.4.6

### What architecture are you using?

amd64

### What steps will reproduce the bug?

The API URL isn't set properly when you don't use `kong.ingress.enabled`. The setting `global.publicURL` is ignored due to an obvious mistake in the `_helpers.tpl` file.

Here is the part in the `_helpers.tpl` that's buggy:

```
{{/*
Return the proper Supabase API Public URL
*/}}
{{- define "supabase.api.publicURL" -}}
{{- if .Values.publicURL -}}
{{- print .Values.publicURL -}}
```

Since there never is a value for `.Values.publicURL` this condition is never matched.

To fix please change from `.Values.publicURL` to `.Values.global.publicURL` 

### Are you using any custom parameters or values?

no

### What is the expected behavior?

`supabase.api.publicURL` should be set to the value of `.Values.global.publicURL` 

### What do you see instead?

> SUPABASE_PUBLIC_URL: "http://localhost:80"

### Additional information

I already provided the fix, so please get it done!

carrodher:
Thank you for bringing this issue to our attention. We appreciate your involvement! If you're interested in contributing a solution, we welcome you to create a pull request. The Bitnami team is excited to review your submission and offer feedback. You can find the contributing guidelines [here](https://github.com/bitnami/charts/blob/main/CONTRIBUTING.md).

Your contribution will greatly benefit the community. Feel free to reach out if you have any questions or need assistance.

PR #20686, opened 2023-11-08
[bitnami/supabase] FIX: Public URL

jsmueller7:
### Description of the change

As described [here](https://github.com/bitnami/charts/issues/19583#issue-1915967951)

### Benefits

If one doesn't enable the kong ingress the public URL will be recognized through setting the `Values.global.publicURL` field.

### Possible drawbacks

none.

### Applicable issues

https://github.com/bitnami/charts/issues/19583#issue-1915967951

### Additional information

It's a very plain and simple thing.

### Checklist

### DIFF
--- a/bitnami/supabase/Chart.yaml
+++ b/bitnami/supabase/Chart.yaml
@@ -53,4 +53,4 @@ maintainers:
 name: supabase
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/supabase
-version: 2.1.2
+version: 2.1.3


jsmueller7 (commented):
This adds some additional values to realtime.

1. Realtime needs to know the database user in order to write data into another scheme (realtime). Otherwise the data will be written to the public scheme.
2. Without the `DNS_NODES` env realtime won't work in a cluster.

javsalgar:
Hi!

Thank you so much for the PR! Could you fix the DCO issue and the merge conflict?

bitnami/supabase/templates/_helpers.tpl
```diff
@@ -22,8 +22,8 @@ Return the proper Supabase Studio Public URL
 Return the proper Supabase API Public URL
 */}}
 {{- define "supabase.api.publicURL" -}}
-{{- if .Values.publicURL -}}
-{{- print .Values.publicURL -}}
+{{- if .Values.global.publicURL -}}
```
aoterolorenzo, line 25: Shouldn't be at least a binding `.Values.global.publicURL` value in the _values.yaml_ file?

bitnami/supabase/values.yaml
```diff
@@ -950,6 +950,7 @@ realtime:
     DB_HOST: {{ include "supabase.database.host" . | quote }}
     DB_PORT: {{ include "supabase.database.port" . | quote }}
     DB_NAME: {{ include "supabase.database.name" . | quote }}
+    DB_USER: {{ include "supabase.database.user" . | quote }}
```
aoterolorenzo, line 953: What's the motivation of this changes? Seems to go further than just fix the public URL. Maybe unintentional changes?

jsmueller7: These changes aren't connected to the publicURL issue. I should have opend a pull request on it's own for this.

**The background of these change suggestions is:**
* Without knowing the `DB_USER` the realtime extension isn't able to write into the schema of it's extension (wheather it's called "realtime" or "_realtime" - in your deployment there exists at least a schema "realtime"). Instead realtime then writes it's information into the public schema, wich isn't what you want.
* After adding `DB_USER` you can add `DB_AFTER_CONNECT_QUERY: 'SET search_path TO <schema the extension shall use>'. Then it won't be using the public schema any more.
* Without specifying the `DNS_NODES` variable realtime can't be run it in a cluster.
* `ENABLE_TAILSCALE: "false"` actually shouldn't change anything because it sould be the default behavior.

bitnami/supabase/templates/_helpers.tpl
jsmueller7: You're right. You can reject this request. I had a missconfig in my own local values.

PR #20950, opened 2023-11-14
[bitnami/minio] Allow provisioning unversioned buckets

spanglerco:
### Description of the change

It isn't currently possible to provision unversioned buckets from the Helm chart. In MinIO, bucket versioning can be in three states: versioned, suspended, and unversioned. The Helm chart's existing bucket.versioning field is a boolean, and so it only allows buckets to be provisioned with suspended or versioned states. It's desirable to be able to provision unversioned buckets (i.e., where `mc version` has not been run), as otherwise, folders/prefixes are not automatically deleted when their contents are deleted.

This PR changes the bucket.versioning field to be an enum while maintaining backwards compatibility with boolean values. Allowed values:

| value | old behavior | new behavior |
|-------|---------------|----------------|
| (unspecified) | `mc version suspend` | `mc version suspend` |
| (empty) | `mc version suspend` | `mc version suspend` |
| false | `mc version suspend` | `mc version suspend` |
| true | `mc version enable` | `mc version enable` |
| "Versioned" | error | `mc version enable` |
| "Suspended" | error | `mc version suspend` |
| "Unchanged" | error | no call to `mc version` |
| other value | error | fail |

Note that the value is "Unchanged" instead of "Unversioned" because there's no way to set a bucket to be unversioned. That's just the initial state of a bucket. "Unchanged" will leave a bucket in whatever versioning state it's already in.

### Benefits

Clients will be able to provision unversioned buckets by supplying a version of "Unchanged".

### Possible drawbacks

Default behavior may be confusing because of backwards compatibility. I didn't do it in this PR, but I would be in favor of bumping the chart's major version and changing the default (the first two rows in the table above) to be "Unchanged".

### Additional information

Tested each case before and after this change using `helm template` with values such as:

```yaml
mode: distributed
provisioning:
  enabled: true
  buckets:
    - name: missing
      region: us-east-1
      withLock: false
    - name: empty
      region: us-east-1
      versioning: ""
      withLock: false
    - name: "false"
      region: us-east-1
      versioning: false
      withLock: false
    - name: "true"
      region: us-east-1
      versioning: true
      withLock: false
    - name: unchanged
      region: us-east-1
      versioning: Unchanged
      withLock: false
    - name: versioned
      region: us-east-1
      versioning: Versioned
      withLock: false
    - name: suspended
      region: us-east-1
      versioning: Suspended
      withLock: false
```

### Checklist


spanglerco:
Hi @CeliaGMqrz, is there any availability for someone to take a look at this PR? Thank you.

CeliaGMqrz (approved):
Hi @spanglerco 

First, sorry for the late response after vacation. 

Thanks for taking care of this. Looks good to me.

Thanks for your contribution!

PR #18889, opened 2023-08-27
[bitnami/common] new macro to merge a list of values with rendering 

jouve:
### Description of the change

new macro to merge a list of values with rendering 

### Benefits

support for fixing #18888

after this is integrated, all charts will be fixed by applying diffs like this one for labels  :
```diff
-{{- $podLabels := merge .Values.master.podLabels .Values.commonLabels }} 
+{{- $podLabels := include "common.tplvalues.merge" (dict "values" .Values.master.podLabels .Values.commonLabels "context" .) | fromYaml }} 
   selector:
     matchLabels: {{- include "common.labels.matchLabels" ( dict "customLabels" $podLabels "context" $ ) | nindent 6 }}
```
 and annotations
```diff
-  {{- $annotations := merge .Values.master.service.annotations .Values.commonAnnotations }}
+  {{- $annotations := include "common.tplvalues.merge" ( dict "values" ( list .Values.master.service.annotations .Values.commonAnnotations ) "context" . ) | fromYaml }}
```

the issues in charts can be found by searching :
* `$podLabels := merge`
* `$claimLabels := merge`
* `$labels := merge`
* `$annotations := merge`

### Applicable issues

- fixes #18888

### Checklist

- [x] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/). This is *not necessary* when the changes only affect README.md files.
- [x] Variables are documented in the values.yaml and added to the `README.md` using [readme-generator-for-helm](https://github.com/bitnami-labs/readme-generator-for-helm)
- [x] Title of the pull request follows this pattern [bitnami/<name_of_the_chart>] Descriptive title
- [x] All commits signed off and in agreement of [Developer Certificate of Origin (DCO)](https://github.com/bitnami/charts/blob/main/CONTRIBUTING.md#sign-your-work)

PR #20964, opened 2023-11-15
[template] Fix usage of common.tplvalues.merge partial

maxnitze:
### Description of the change

In #18889 the usage of a new partial was introduced to merge objects together. Unfortunately in the change the parameters were not given correctly to the partial.

### Benefits

_bugfix_

### Possible drawbacks

%

### Applicable issues

%

### Additional information

%

### Checklist

### DIFF
--- a/template/CHART_NAME/templates/daemonset.yaml
+++ b/template/CHART_NAME/templates/daemonset.yaml
@@ -11,14 +11,14 @@ metadata:
   labels: {{- include "common.labels.standard" ( dict "customLabels" .Values.commonLabels "context" $ ) | nindent 4 }}
     app.kubernetes.io/component: %%COMPONENT_NAME%%
   {{- if or .Values.%%MAIN_OBJECT_BLOCK%%.daemonsetAnnotations .Values.commonAnnotations }}
-  {{- $annotations := include "common.tplvalues.merge" (dict "values" (list .Values.%%MAIN_OBJECT_BLOCK%%.daemonsetAnnotations .Values.commonAnnotations) "context" .) | fromYaml }}
+  {{- $annotations := include "common.tplvalues.merge" (dict "values" (list .Values.%%MAIN_OBJECT_BLOCK%%.daemonsetAnnotations .Values.commonAnnotations) "context" .) }}
   annotations: {{- include "common.tplvalues.render" ( dict "value" $annotations "context" $ ) | nindent 4 }}
   {{- end }}
 spec:
   {{- if .Values.%%MAIN_OBJECT_BLOCK%%.updateStrategy }}
   updateStrategy: {{- toYaml .Values.%%MAIN_OBJECT_BLOCK%%.updateStrategy | nindent 4 }}
   {{- end }}
-  {{- $podLabels := include "common.tplvalues.merge" (dict "values" (list .Values.%%MAIN_OBJECT_BLOCK%%.podLabels .Values.commonLabels) "context" .) | fromYaml }}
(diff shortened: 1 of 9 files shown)


template/CHART_NAME/templates/daemonset.yaml
```diff
@@ -17,7 +17,7 @@ spec:
   {{- if .Values.%%MAIN_OBJECT_BLOCK%%.updateStrategy }}
   updateStrategy: {{- toYaml .Values.%%MAIN_OBJECT_BLOCK%%.updateStrategy | nindent 4 }}
   {{- end }}
-  {{- $podLabels := include "common.tplvalues.merge" (dict "values" .Values.%%MAIN_OBJECT_BLOCK%%.podLabels .Values.commonLabels "context" .) | fromYaml }}
+  {{- $podLabels := include "common.tplvalues.merge" (dict "values" (list .Values.%%MAIN_OBJECT_BLOCK%%.podLabels .Values.commonLabels) "context" .) | fromYaml }}
```
FraPazGal, line 20: Good catch and thanks for the PR @maxnitze! Could you also remove the `fromYaml` part from `template/` files? This was actually included in the `common.tplvalues.merge` [template](https://github.com/bitnami/charts/blob/a5c43a1c7a342dc39ba4d08d9aaced81b3228b54/bitnami/common/templates/_tplvalues.tpl#L35) so it's not necessary here.

maxnitze: Hey @FraPazGal , I'm pretty sure this is still necessary. In [line 37](https://github.com/bitnami/charts/blob/a5c43a1c7a342dc39ba4d08d9aaced81b3228b54/bitnami/common/templates/_tplvalues.tpl#L37) the `$dst` dictionary is rendered to yaml again (since that is the only way the templates can pass on things).

FraPazGal: You are right about the `toYaml` conversion at the end, I missed it. That said, the `fromYaml` doesn't maje any difference when rendering the template.

I made a couple of tests using some of our chart which do not implement the `fromYaml` and I haven't seen issues. This is probably because it doesn't matter whether the value is a string or an object (as you are just printing it). wdyt?

FraPazGal:
Sorry about this @maxnitze, it seems our automation tool missed your response and didn't ping. I added a comment on the previous review, please check it when you have some time.

template/CHART_NAME/templates/daemonset.yaml
maxnitze: Hm. Yeah, you might be right. Did not check all the places, but I think it should wirk either way.

I personally like working with "structures" better. But in this case both solutions are ok for me, I think. It's a constant "fromYaml-toYaml" with the common helpers anyways :)

If it's ok for you I would keep it as it is. If not, I can change it. Your call :)

maxnitze:
Ping, just to make sure the automation tool did not mess things up again :)

template/CHART_NAME/templates/daemonset.yaml
FraPazGal: Yes I agree with your reasoning. I have checked our charts and it seems the vast majority (except a few instances in `bintami/minio` and `bitnami/mlflow`) is using the `common.tplvalues.merge` without the `fromYaml`. Apologies for being so nit-picky here but I think we should remove it just to keep consistency with the rest of charts.

maxnitze:
I removed the `fromYaml` from all usages in the template. I also found some new usages of the partial, that had the same issue. They were merged in between the creation of my PR and now (see #21010 ).

PR #20966, opened 2023-11-15
[template] add support for non-equal service and container ports

maxnitze:
### Description of the change

The template allows for container and service ports, that are not equal. But the created service is only taking the service port into consideration.

In case the container and service ports are different the target port of the service should be set to the container port.

### Benefits

Template will work, when both ports are set differently.

### Possible drawbacks

None, that I know of.

### Applicable issues

%

### Additional information

https://kubernetes.io/docs/concepts/services-networking/service/#field-spec-ports

### Checklist


CeliaGMqrz (approved):
Hi @maxnitze, 

Sorry for the delay. Thanks for your contribution! LGTM

PR #21081, opened 2023-11-21
[bitnami/zookeeper] Make Zookeeper DefaultMode YAML 1.2 Compliant

schrieveslaach:
### Description of the change

With the YAML 1.2 spec octals must be prefixed with `0o` and not `0`. Parsers implementing the latest spec run into issues.

### Benefits

This commit makes the Zookeeper script mount compliant with YAML 1.2.

### Applicable issues

None at the moment

### Additional information

See background discussion here: https://github.com/dtolnay/serde-yaml/pull/225

### Checklist

### DIFF
--- a/bitnami/zookeeper/Chart.yaml
+++ b/bitnami/zookeeper/Chart.yaml
@@ -28,4 +28,4 @@ maintainers:
 name: zookeeper
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/zookeeper
-version: 12.3.3
+version: 12.3.4
--- a/bitnami/zookeeper/templates/statefulset.yaml
+++ b/bitnami/zookeeper/templates/statefulset.yaml
@@ -443,7 +443,7 @@ spec:
         - name: scripts
           configMap:
             name: {{ printf "%s-scripts" (include "common.names.fullname" .) }}
-            defaultMode: 0755
+            defaultMode: 0o755
         {{- if or .Values.configuration .Values.existingConfigmap }}
         - name: config
           configMap:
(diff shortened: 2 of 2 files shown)


andresbono:
Hi, thanks for opening this PR! I have some concerns about merging this PR because the [Kubernetes API Reference for v1.28](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.28/#configmapvolumesource-v1-core) is still using the YAML v1.1 format for octals:

> Must be an octal value between 0000 and 0777 [...] YAML accepts both octal and decimal values [...]. Defaults to 0644.

Could this cause some backwards incompatibility in some old (or current) versions of the k8s tooling? helm, kubectl... Would this change generate confudion for users as the official k8s API spec is documented using the old format? If we apply this update, shouldn't we apply it globally to all the charts for homogeneity?

What is the specific issue you are finding by sticking to YAML v1.1 format?

schrieveslaach:
> What is the specific issue you are finding by sticking to YAML v1.1 format?

I'm trying to parse the output of this Helm chart with Rust and the crate [kube](https://crates.io/crates/kube). This is needed for https://github.com/aixigo/PREvant/issues/146 (code will be public in a few days). I came across https://github.com/dtolnay/serde-yaml/pull/225 which showed Rust's serde library uses YAML 1.2 and thus my code failed because `defaultMode` was then a string and not a numeric value anymore.

However, I noticed that the chart also makes use of decimals (see TLS secrets) and thus I applied the same trick and this addresses following questions:

> Could this cause some backwards incompatibility in some old (or current) versions of the k8s tooling? helm, kubectl... Would this change generate confudion for users as the official k8s API spec is documented using the old format?

> If we apply this update, shouldn't we apply it globally to all the charts for homogeneity?

I was thinking about doing so but then I read in the contribution guide to only raise PRs for single charts. I'm up to do it for all charts. Just tell me what is your preferred approach.

andresbono (approved):
LGTM

schrieveslaach:
Thanks a lot.

schrieveslaach:
@andresbono, I did the same trick for Kafka (see #21086). Do you mind to check this out as well?

Issue #19393, opened 2023-09-19
[bitnami/postgresql] Wrong type in values schema for primary.service.nodePorts.postgresql field


psttf:
### Name and Version

bitnami/postgresql 12.11.2

### What architecture are you using?

None

### What steps will reproduce the bug?

Deploy bitnami/postgresql 12.11.2 Helm chart with `values.yaml` containing a numeric value for `primary.service.nodePorts.postgresql`

### Are you using any custom parameters or values?

numeric value for `primary.service.nodePorts.postgresql`

### What is the expected behavior?

Chart deployed

### What do you see instead?

```
Error: UPGRADE FAILED: values don't meet the specifications of the schema(s) in the following chart(s):
postgresql:
- primary.service.nodePorts.postgresql: Invalid type. Expected: string, given: integer
```

### Additional information

According to the [comments in values.yaml](https://github.com/bitnami/charts/blob/main/bitnami/postgresql/values.yaml#L603) `primary.service.nodePorts.postgresql` is a numeric field and it worked this way using the previous 12.11.1 chart version. Upgrading to 12.11.2 results in the error above since it now includes additional schema for this field.

PR #19335, opened 2023-09-18
Revert "[bitnami/*] Autogenerate json-schema files for all the charts"

fmulero:
### Description of the change

This change reverts bitnami/charts#19194 

### Applicable issues

* https://github.com/bitnami/charts/pull/19278
* https://github.com/bitnami/charts/issues/19284
* https://github.com/bitnami/charts/issues/19321
* https://github.com/bitnami/charts/issues/19274

### Additional information

Also PR #19195  should be freezed.

### Checklist

- [ ] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/). This is *not necessary* when the changes only affect README.md files.
- [ ] Variables are documented in the values.yaml and added to the `README.md` using [readme-generator-for-helm](https://github.com/bitnami-labs/readme-generator-for-helm)
- [X] Title of the pull request follows this pattern [bitnami/<name_of_the_chart>] Descriptive title
- [X] All commits signed off and in agreement of [Developer Certificate of Origin (DCO)](https://github.com/bitnami/charts/blob/main/CONTRIBUTING.md#sign-your-work)

Reverts 

Due to the issues reported and the

PR #19194, opened 2023-09-08
 [bitnami/*] Autogenerate json-schema files for all the charts

antgamdia:
### Description of the change

This PR simply performs a bulk update on every chart, generating (or updating) the json-schema file for the `values.yaml` using the [readme-generator-for-helm tool](https://github.com/bitnami-labs/readme-generator-for-helm).

### Benefits

The charts will have an up-to-date values.schema.json file, which, apart from the intrinsic validation benefits, will ease the UX while deploying them using [Kubeapps](https://kubeapps.dev/).

### Possible drawbacks

In the past, we (Kubeapps) ran into issues like https://github.com/bitnami-labs/readme-generator-for-helm/issues/15, but in recent versions, this should no longer be an issue.

### Applicable issues

N/A

### Additional information

TBD

### Checklist

- [ ] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/). This is *not necessary* when the changes only affect README.md files.
- [ ] Variables are documented in the values.yaml and added to the `README.md` using [readme-generator-for-helm](https://github.com/bitnami-labs/readme-generator-for-helm)
- [X] Title of the pull request follows this pattern [bitnami/<name_of_the_chart>] Descriptive title
- [X] All commits signed off and in agreement of [Developer Certificate of Origin (DCO)](https://github.com/bitnami/charts/blob/main/CONTRIBUTING.md#sign-your-work)

PR #21237, opened 2023-11-24
[bitnami/postgresql] update schema

tumido:
### Description of the change

The schema for PostgreSQL chart is severely lacking so I decided to update it.

I've noticed there was already an attempt https://github.com/bitnami/charts/pull/19194 which failed https://github.com/bitnami/charts/pull/19335.

 

### Benefits

The schema for `bitnami/postgresql` will be finally useful, and not just a gimmick. 

### Possible drawbacks

The schema update in itself is huge, so there may be hidden issues that are simply too hard to spot. However, I think it's better to fix the schema than keep using the current, very incomplete one.

### Applicable issues

https://github.com/bitnami/charts/issues/19393

### Additional information

I've scoped the update to `bitnami/postgresql` chart only, so we may have better success this time.

I've autogenerated the schema via https://github.com/bitnami/readme-generator-for-helm (commit 1)

And then I've addressed https://github.com/bitnami/charts/issues/19393 in a separate commit. This way it should be a bit easier to review.

### Checklist

### DIFF
--- a/bitnami/postgresql/Chart.yaml
+++ b/bitnami/postgresql/Chart.yaml
@@ -35,4 +35,4 @@ maintainers:
 name: postgresql
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/postgresql
-version: 13.2.16
+version: 13.2.17
--- a/bitnami/postgresql/values.schema.json
+++ b/bitnami/postgresql/values.schema.json
@@ -1,156 +1,2552 @@
 {
-  "$schema": "http://json-schema.org/schema#",
-  "type": "object",
-  "properties": {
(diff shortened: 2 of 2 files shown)


juan131 (changes requested):
Hi @tumido, thanks for this great contribution!

Unfortunately, we cannot accept this kind of changes right now given we don't have a GH action in place that uses the readme-generator to sync the JSON schema whenever the values are modified. The lack of this action is related to the limitations you pointed out on readme-generator that enforce to manually patch the result.
Without automatic syncing, we'd rather keep the schemas minimal to avoid values validation issues.

Issue #20682, opened 2023-11-08
[bitnami/nginx] 15.4.0 breaks cloneStaticSiteFromGit (dubious ownership in repository at '/app')


Cogitri:
### Name and Version

bitnami/nginx 15.4.0

### What architecture are you using?

amd64

### What steps will reproduce the bug?

1. Setup nginx with the values.yml below (and version >=15.4.0). The same config works with 15.3.0

### Are you using any custom parameters or values?

```yaml
cloneStaticSiteFromGit:
  enabled: true
  repository: <repo>
  branch: master
  interval: 15
```

### What is the expected behavior?

Synching should work

### What do you see instead?

Starting with 15.4.0, the following error is logged by the git-repo-syncher:

```
fatal: detected dubious ownership in repository at '/app'
```

Cogitri:
I suppose the cause of this is b17d6aeceb48ce6686b60ced1861753d9ea4596e . Setting `containerSecurityContext.enabled=false` mitigates the issue. That isn't great of course, so a proper fix would probably be to honour that setting in the init container which clones the repo.

Cogitri:
CC @javsalgar

Lavryana:
same problem. Exactly the same in version 15.4.3. I'm waiting for a solution, everything is CrashLoopBackOff with this version.

PR #21451, opened 2023-12-07
[bitnami/nginx] Set /app as safe directory when enabling gitSync

TheiLLeniumStudios:
### Description of the change

Added `/app` as a safe directory to global git config as newer versions of git require you to do that

### Applicable issues

- Fixes https://github.com/bitnami/charts/issues/20682

### Checklist


bitnami/nginx/templates/deployment.yaml
```diff
@@ -131,6 +131,7 @@ spec:
             - -ec
             - |
               [[ -f "/opt/bitnami/scripts/git/entrypoint.sh" ]] && source "/opt/bitnami/scripts/git/entrypoint.sh"
+              git config --global --add safe.directory /app
               while true; do
                   cd /app && git pull origin {{ .Values.cloneStaticSiteFromGit.branch }}
```
andresbono, line 136: I would avoid having to use the `git config --global` command as it tries to persist the configuration in `/.gitconfig` and that file may not be writable by an unprivileged user. I see it was already mentioned in the associated issue https://github.com/bitnami/charts/issues/20682#issuecomment-1844993985, and to workaround that, you are setting `HOME=/tmp`, but that is not set by default, right?

What if we just do the following instead?

```suggestion
              while true; do
                  cd /app && git -c safe.directory=/app pull origin {{ .Values.cloneStaticSiteFromGit.branch }}
```

andresbono (commented):
Thank you for your contribution! Could you please check my suggestion about your fix and let me know what do you think?

TheiLLeniumStudios:
@andresbono accepted your change. Can we get this merged now please?

johanneskastl:
Any news on this?

andresbono (approved):
LGTM, Thank you for your contribution!

PR #21542, opened 2023-12-12
[bitnami/influxdb] Allow separate persistence cfg for backups

kevin-delisle-peakpower:
- Fix failure to render when .Values.cronjob.securityContext is undefined (which is the current default in values.yaml)
- Add an optional configuration section under backup that allows the specification of distinct values for persistence. If this configuration section is not provided, the template will default to rendering using the existing top-level persistence configuration section to maintain compatibility.

### Benefits
Users can now define volumes for the Deployment and their backups separately without needing Kustomize or manual post-deployment hacks.

Example usage could include:
- using different StorageClasses for their backup volumes vs. their active database volumes.
- using a larger volume for backups vs. the main database volume to accommodate multiple full backups

### Possible drawbacks

There should not be any drawbacks to this PR; I have attempted to maintain backwards compatibility with existing use cases in the event that someone provides a values.yaml file that does not include the new section.

### Additional information

I would've committed the bugfix as a separate PR, but the template was failing to render without any changes in-place, so this fix is required to ensure the other commit works as expected.

If Bitnami releases patches for previous minor versions, I would recommend cherry-picking the first commit into a `5.11.2` release.

I've dumped the template out using permutations of both persistence modes to confirm that the `backup.persistence` section will take precedence over `persistence` for the backup volume, as well as to confirm that the templates still correctly render in the complete absence of that configuration section.

### Checklist

### DIFF
--- a/bitnami/influxdb/Chart.yaml
+++ b/bitnami/influxdb/Chart.yaml
@@ -37,4 +37,4 @@ maintainers:
 name: influxdb
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/influxdb
-version: 5.11.1
+version: 5.12.0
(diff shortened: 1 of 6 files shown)


corcoran (approved):
🔥

corcoran:
Bump

aoterolorenzo (approved):
LGTM

PR #21545, opened 2023-12-13
Remove explicit null assignment to nodePort in Kubernetes Service

### LINKED ISSUE #21501: nodePort: null -> Incorrect type. Expected "integer".
### Name and Version

bitnami/rabbitmq

None

Hi!

We're building application.yaml using `helm template rabbitmq`

In the generated file I see that `nodePort: null` is not a valid value according to the schema kubernetes://schema/io.k8s.api.core.v1.serviceport 

Ideally, I prefer that the generated yaml file should be a valid one. Maybe this is an issue not related to the schema itself.  Luckily it doesn't impact the runtime behavior anyhow. But the point is - when I remove `nodePort: null` from the file - the service keeps working identically. Is it needed?

### What do you see instead?

image

opencmit2:
### Description of the change

Remove explicit null assignment to nodePort in Kubernetes Service
### Benefits

In Kubernetes Service, it is not necessary to explicitly set nodePort to null. If nodePort is not set, its value defaults to null. This commit removes the explicit null assignment, simplifying the code.
### Applicable issues

- fixes #21501 

### Checklist

### DIFF
--- a/bitnami/rabbitmq/Chart.yaml
+++ b/bitnami/rabbitmq/Chart.yaml
@@ -30,4 +30,4 @@ maintainers:
 name: rabbitmq
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/rabbitmq
-version: 12.5.6
+version: 12.5.7
--- a/bitnami/rabbitmq/templates/svc.yaml
+++ b/bitnami/rabbitmq/templates/svc.yaml
@@ -45,9 +45,7 @@ spec:
     - name: {{ .Values.service.portNames.amqp }}
       port: {{ .Values.service.ports.amqp }}
       targetPort: amqp
-      {{- if (eq .Values.service.type "ClusterIP") }}
-      nodePort: null
-      {{- else if and (or (eq .Values.service.type "NodePort") (eq .Values.service.type "LoadBalancer")) (not (empty .Values.service.nodePorts.amqp)) }}
+      {{- if and (or (eq .Values.service.type "NodePort") (eq .Values.service.type "LoadBalancer")) (not (empty .Values.service.nodePorts.amqp)) }}
       nodePort: {{ .Values.service.nodePorts.amqp }}
       {{- end }}
     {{- end }}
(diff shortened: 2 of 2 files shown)


juan131 (changes requested):
Hi @opencmit2 

This safeguard mechanism was introduce long time ago on almost every Bitnami chart to prevent issues while upgrading the chart switching from NodePort/LoadBalancer service type to ClusterIP, see: 

- https://github.com/helm/charts/pull/13009

Did you find any issue with it or is this simply a "cosmetic" change?

opencmit2:
Hi @juan131 
By removing the explicit configuration of nodePort, the generated service yaml template can be made usable. When using the template with statefulset resources, initContainers may appear empty, and the definition of .Values.volumePermissions.enabled will become ineffective. I believe this issue can be resolved through substituting parameters, not merely by changing the sequence as shown in the diagram.
image

opencmit2:
Hi @juan131 
When switching back to ClusterIP from LoadBalancer or NodePort during a Helm update operation, Kubernetes will report an error because it does not allow changes to the already set nodePort.

In this case, you need to delete and recreate the service instead of trying to update it. This is a limitation of Kubernetes, not Helm.

So setting nodePort: null explicitly will cause the output of the helm template to need modification. If nodePort: null is not set explicitly, it will cause the switch from NodePort back to ClusterIP to fail. This is a trade-off issue.

Do you have any other solutions?

juan131:
> So setting `nodePort: null` explicitly will cause the output of the helm template to need modification

Why? I mean, I understand why the generated service would require modification, see:

```console
$ helm template rabbitmq oci://registry-1.docker.io/bitnamicharts/rabbitmq -s templates/svc.yaml | tee /dev/tty | kubectl apply -f -
---
# Source: rabbitmq/templates/svc.yaml
apiVersion: v1
kind: Service
metadata:
  name: rabbitmq
  namespace: "default"
  labels:
    app.kubernetes.io/instance: rabbitmq
    app.kubernetes.io/name: rabbitmq
    app.kubernetes.io/version: 3.12.10
    helm.sh/chart: rabbitmq-12.5.6
spec:
  type: ClusterIP
  sessionAffinity: None
  ports:
    - name: amqp
      port: 5672
      targetPort: amqp
      nodePort: null
    - name: epmd
      port: 4369
      targetPort: epmd
      nodePort: null
    - name: dist
      port: 25672
      targetPort: dist
      nodePort: null
    - name: http-stats
      port: 15672
      targetPort: stats
      nodePort: null
  selector:
    app.kubernetes.io/instance: rabbitmq
    app.kubernetes.io/name: rabbitmq
---
service/rabbitmq created
```

opencmit2:
Alright, do you have any good solutions?

juan131:
I might be missing something but I don't think there's any issue, at least from a functional point of view.

I agree the generated manifest might be "weird" and the `nodePort: null` line doesn't add any value from a developer's readability point of view but it's not a problem as far as Kubernetes manages it properly.

PR #21569, opened 2023-12-14
[bitnami/kafka] embedded jmx exporter agent

### LINKED ISSUE #21309: run JMX to Prometheus exporter for Kafka as java agent rather than side-car
### Name and Version

bitnami/kafka 18.0.3

Expose JVM process metrics (memory and CPU usage, e.g. `jvm_memory_bytes_used`) to Prometheus.

Kafka JMX metrics are currently exported via a side-car in the Kafka Chart: see container `jmx-exporter` in `bitnami/kafka/templates/statefulset.yaml`, enabled when `.Values.metrics.jmx.enabled` is true. As indicated in the README of the Prometheus JMX exporter project (https://github.com/prometheus/jmx_exporter), it is actually recommended to run it as a JVM agent rather than a separate process. One very concrete advantage is that it then also exposes process metrics (memory and CPU usage, e.g. `jvm_memory_bytes_used`).

What is available as `jvm_memory_bytes_used` can also be obtained from `java_lang_memory_heapmemoryusage_used` (that's the Prometheus metric name that corresponds to an attribute of the java.lang:type=Memory MBean), but having the process metrics available directly is more in-line with how it typically works for other Java processes.

The current situation makes system monitoring for Kafka inconsistent with other Java processes, such as for example bitnami/zookeeper. 

I ran a quick test with a customized bitnami/kafka docker image (added the prometheus agent jar file) to confirm this will just work fine. I can work on a proper PR where I replace the side-car with an agent-based implementation, if you agree this would be an improvement.

janvanbesien-ngdata:
### Description of the change

Run the jmx exporter for prometheus as an embedded agent in the Kafka broker JVM, rather than as a standalone process.

### Benefits

As explained in https://github.com/prometheus/jmx_exporter, this is the preferred approach. The most obvious benefit is that it then also exposes JVM process metrics (e.g. about memory usage etc).

fixes #21309 

### Possible drawbacks

None that I am aware of.

### Applicable issues

None

### Additional information

N/A

### Checklist

### DIFF
--- a/bitnami/kafka/Chart.yaml
+++ b/bitnami/kafka/Chart.yaml
@@ -42,4 +42,4 @@ maintainers:
 name: kafka
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/kafka
-version: 26.5.0
+version: 27.0.0
--- a/bitnami/kafka/README.md
+++ b/bitnami/kafka/README.md
@@ -542,21 +542,7 @@ The command removes all the Kubernetes components associated with the chart and 
 | `metrics.kafka.serviceAccount.name`                               | The name of the service account to use. If not set and `create` is `true`, a name is generated                                   | `""`                                                                                    |
 | `metrics.kafka.serviceAccount.automountServiceAccountToken`       | Allows auto mount of ServiceAccountToken on the serviceAccount created                                                           | `true`                                                                                  |
 | `metrics.jmx.enabled`                                             | Whether or not to expose JMX metrics to Prometheus                                                                               | `false`                                                                                 |
-| `metrics.jmx.kafkaJmxPort`                                        | JMX port where the exporter will collect metrics, exposed in the Kafka container.                                                | `5555`                                                                                  |
-| `metrics.jmx.image.registry`                                      | JMX exporter image registry                                                                                                      | `REGISTRY_NAME`                                                                         |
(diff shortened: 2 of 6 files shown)


migruiz4:
Hi @janvanbesien-ngdata,

I'm sorry but as mentioned by my colleague at https://github.com/bitnami/containers/pull/53817#issuecomment-1859812522, we can not accept this PR as it is because we want to keep the image minimal, and adding the Java agent would increase its attack surface.

Nevertheless, as mentioned by @OneCricketeer, an interesting alternative would be using an init container to copy the javaagent file into the container. That way, the javaagent would only be present when JMX metrics are enabled.

migruiz4 (commented):
Hi @janvanbesien-ngdata,

I'm sorry but as mentioned by my colleague at https://github.com/bitnami/containers/pull/53817#issuecomment-1859812522, we can not accept this PR as it is because we want to keep the image minimal, and adding the Java agent would increase its attack surface.

Nevertheless, as mentioned by @OneCricketeer, an interesting alternative would be using an init container to copy the javaagent file into the container. That way, the javaagent would only be present when JMX metrics are enabled.

janvanbesien-ngdata:
Ok, thanks for the feedback.

migruiz4:
You are welcome, if you are not familiar with init-containers or have any questions about how to implement the suggestion please let me know.

Issue #21360, opened 2023-12-03
[bitnami/mariadb] Request for Modification in MariaDB Helm Chart to Support HashiCorp Vault Integration


sobrito:
### Name and Version

bitnami/mariadb 14.1.4

### What is the problem this feature will solve?

The current issue is the incompatibility between Bitnami's MariaDB Helm chart and HashiCorp's Vault Secrets service. The key for the MariaDB password needs to be mariadb-root-password, mariadb-replication-password and mariadb-password but Vault Secrets does not accept hyphens (-) in secret names, limiting the integration between these two services.

### What is the feature you are proposing to solve the problem?

I am proposing two potential solutions:

- Change the defaults keys from mariadb-root-password, mariadb-replication-password and mariadb-password to mariadb_root_password, mariadb_replication_password and mariadb_password, allowing direct integration with Vault, which accepts underscores (_) in names.
- Add an option in values.yaml to set a custom name for the secret key, providing flexibility to adapt to different secret management systems.

### What alternatives have you considered?

The alternatives considered were: not using the HCPVaultSecretsApp for dynamic password generation and manually entering the password, which would compromise the security and automation of the process.

Issue #7953, opened 2021-10-27
Helm upgrade fails if mysql root password is unknown !


pankaj-dahiya:
**Which chart**:
The latest one! - version doesn't matter here.

**Describe the bug**
I rotated the root password of my MySQL database via hasicorp-vault. So the root password is unknown to me.

**To Reproduce**
1. Run bitnami/mysql chart with root password as "Hello@123" in values.yaml.
2. Rotate root password with the help of vault - ( https://www.vaultproject.io/docs/secrets/databases/mysql-maria )
3. Helm upgrade and restart the MySQL pod.
4. Mysql pod went into CrashLoopBackOff.

**Expected behaviour**
Helm upgrade should not ask for a root password as the root password is the "ROOT-PASSWORD" which is a secret and should not be shared at every helm upgrade.

**Version of Helm and Kubernetes**:
Doesn't matter.

rafariossaa:
Hi,
  This kind of functionality using vaults is not supported. It acts as an external client and the chart deployment doesn't know anything about the new password, and this password is needed in the initialization scripts.

PR #21790, opened 2023-12-29
[bitnami/mysql] Allow undo secret generation when using secret injector like vault.

NeuronAddict:
When using a vault to inject secrets in the pod, there is no need to create secrets with the password values.

In my case, secrets are injected as files by vault inside pods with the option below: 

```yaml
mysql:
  auth:
    usePasswordFiles: true
    customPasswordFiles:
    root: /vault/secrets/root-db-password.txt
    user: /vault/secrets/db-password.txt
```

And I use annotations to inject the files inside vault/secrets/. There is no secret generation.

But when I try to upgrade my chart, I get the following error:

```

    'auth.rootPassword' must not be empty, please add '--set auth.rootPassword=$MYSQL_ROOT_PASSWORD' to the command. To get the current value:

        export MYSQL_ROOT_PASSWORD=$(kubectl get secret --namespace "contrast-ui-xm-dev" contrast-mysql -o jsonpath="{.data.mysql-root-password}" | base64 -d)
```

This is because of the `template/secrets.yaml` file, we try to retrieve the password event if we don't need to store it in a secret.

file mysql/template/secret.yaml
```yaml
{{- /*
Copyright VMware, Inc.
*/}}

{{- $host := include "mysql.primary.fullname" . }}
{{- $port := print .Values.primary.service.ports.mysql }}
{{/* This line is interpreted even if we don't need secret */ }}
{{- $rootPassword := include "common.secrets.passwords.manage" (dict "secret" (include "mysql.secretName" .) "key" "mysql-root-password" "length" 10 "providedValues" (list "auth.rootPassword") "context" $) | trimAll "\"" | b64dec }}
{{- $password := include "common.secrets.passwords.manage" (dict "secret" (include "mysql.secretName" .) "key" "mysql-password" "length" 10 "providedValues" (list "auth.password") "context" $) | trimAll "\"" | b64dec }}
{{/* even when false, the upgrade fail whit an error */ }}
{{- if eq (include "mysql.createSecret" .) "true" }}
```

 ### Description of the change

With this change, when the parameter createSecret is false, we do not try to get the secret and the rootPassword.

### Benefits

We can use the vault injector in a clean way and upgrade the chart without error.

### Possible drawbacks

When the parameter to false, the user is responsible for the secret management.

With this configuration, the behavior with the parameters : 
```
creteSecret: false
.Values.serviceBindings.enabled : true
```
Will change.

Before: secret mychart-svcbind-root is created
After: secret mychart-svcbind-root is not created

Same behavior when the username changes.

### Applicable issues

I think this will fix errors like this (but I can't test the same configuration): https://github.com/bitnami/charts/issues/7953

Maybe we can report this PR on issues like this: https://github.com/bitnami/charts/issues/21360

### Checklist

### DIFF
--- a/bitnami/mysql/README.md
+++ b/bitnami/mysql/README.md
@@ -82,7 +82,7 @@ The command removes all the Kubernetes components associated with the chart and 
 ### MySQL common parameters
 
 | Name                               | Description                                                                                                                                                                         | Value                   |
-| ---------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ----------------------- |
+|------------------------------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|-------------------------|
 | `image.registry`                   | MySQL image registry                                                                                                                                                                | `REGISTRY_NAME`         |
 | `image.repository`                 | MySQL image repository                                                                                                                                                              | `REPOSITORY_NAME/mysql` |
 | `image.digest`                     | MySQL image digest in the way sha256:aa.... Please note this parameter, if set, will override the tag                                                                               | `""`                    |
@@ -105,6 +105,7 @@ The command removes all the Kubernetes components associated with the chart and 
 | `initdbScriptsConfigMap`           | ConfigMap with the initdb scripts (Note: Overrides `initdbScripts`)                                                                                                                 | `""`                    |
 | `startdbScripts`                   | Dictionary of startdb scripts                                                                                                                                                       | `{}`                    |
 | `startdbScriptsConfigMap`          | ConfigMap with the startdb scripts (Note: Overrides `startdbScripts`)                                                                                                               | `""`                    |
+| `createSecret`                     | Create secrets for passwords according configuration. Set to false if you use secret injection like vault-injector.                                                                 | `true`                  |
 
 ### MySQL Primary parameters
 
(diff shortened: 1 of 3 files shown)


bitnami/mysql/README.md
```diff
@@ -105,6 +105,7 @@ The command removes all the Kubernetes components associated with the chart and
 | `initdbScriptsConfigMap`           | ConfigMap with the initdb scripts (Note: Overrides `initdbScripts`)                                                                                                                 | `""`                    |
 | `startdbScripts`                   | Dictionary of startdb scripts                                                                                                                                                       | `{}`                    |
 | `startdbScriptsConfigMap`          | ConfigMap with the startdb scripts (Note: Overrides `startdbScripts`)                                                                                                               | `""`                    |
+| `createSecret`                     | Create secrets for passwords according configuration. Set to false if you use secret injection like vault-injector.                                                                 | `true`                  |
```
jotamartos, line 108: This is not a parameter in the values.yaml file so it can't be added here. It's a function [defined in the helpers file](https://github.com/bitnami/charts/blob/main/bitnami/mysql/templates/_helpers.tpl#L139)

jotamartos (changes requested):
As I mentioned, the mysql.createSecret function is defined in the helpers file. Please do modify that function to return false if usePasswordFiles is enabled.

PR #21866, opened 2024-01-05
[bitnami/drupal] - feat: Added tlsWwwPrefix value to handle www subdomain to default SSL cert

g0blin79:
This PR basically adds what is just contained in the `bitnami/wordpress` helm chart.

## Description of the change
Added `tlsWwwPrefix` in the `values.yaml` file and the relative snippet in the `ingress.yaml` template to create an ingress manifest structure just like this:

```yaml
  tls:
  - hosts:
    - example.com
    - www.example.com
    secretName: example.com-tls
```

## Benefits

The ingress section now can handle the common situation in which I want to create a 301 redirect from www.domain.com to domain.com using the same SSL certificate.


dgomezleon (dismissed):
LGTM. Thanks for contributing

dgomezleon (approved):

PR #21891, opened 2024-01-08
[bitnami/rabbitmq] Add auth.existingSecret to support external sealed secrets

### LINKED ISSUE #21359: [bitnami/rabbitmq] Request for Modification in RabbitMQ Helm Chart to Support HashiCorp Vault Integration
### Name and Version

bitnami/rabbitmq 12.5.5

The current issue is the incompatibility between Bitnami's RabbitMQ Helm chart and HashiCorp's Vault Secrets service. The key for the RabbitMQ password needs to be rabbitmq-password, but Vault Secrets does not accept hyphens (-) in secret names, limiting the integration between these two services.

I am proposing two potential solutions:

- Change the default key from rabbitmq-password to rabbitmq_password, allowing direct integration with Vault, which accepts underscores (_) in names.
- Add an option in values.yaml to set a custom name for the secret key, providing flexibility to adapt to different secret management systems.

The alternatives I have considered include:

- Not using HCPVaultSecretsApp and looking for other password management solutions that are compatible with the current structure of the RabbitMQ Helm chart.
- Manually entering the password in the chart, which would reduce the security and automation of the process.

maxvisser:
Note: breaking change removal of auth.existingPasswordSecret

### Description of the change

I changed auth.existingPasswordSecret to be similarly implemented as the redis chart. Most of the helpers I implemented are directly copied.

### Benefits

As described in #21359 it would support all different kinds of password management solutions like Hashicorp Vault.

### Possible drawbacks

I dropped existingPasswordSecret as it would be confusing to keep it. This would mean it would be a breaking change for some users of the chart.

### Applicable issues

- fixes #21359

### Additional information
files used to implement this:
bitnami/redis/templates/_helpers.tpl
bitnami/redis/templates/master/application.yaml
bitnami/redis/templates/secret.yaml

### Checklist


maxvisser:
I believe it would be better to implement the API as other charts and make this a breaking change. (as it would work differently as before);

bitnami/rabbitmq/Chart.yaml
```diff
@@ -30,4 +30,4 @@ maintainers:
 name: rabbitmq
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/rabbitmq
-version: 12.6.2
+version: 13.6.2
```
carrodher, line 33:
```suggestion
version: 13.0.0
```

bitnami/rabbitmq/templates/secret.yaml
```diff
@@ -0,0 +1,23 @@
...
+  {{- end }}
+type: Opaque
+data:
+  {{- if (not .Values.auth.existingSecret ) }}
+  rabbitmq-password: {{ include "redis.password" . | b64enc | quote }}
+  {{- end }}
+  {{- if (not .Values.auth.existingErlangSecret ) }}
+  rabbitmq-erlang-cookie: {{ include "common.secrets.passwords.manage" (dict "secret" (include "common.names.fullname" .) "key" "rabbitmq-erlang-cookie" "length" 32 "providedValues" (list "auth.erlangCookie") "context" $) }}
+  {{- end }}
+{{- end }}
```
juan131, line 23: Please do not rename this secret template file to reduce the number of changes.

bitnami/rabbitmq/values.yaml
```diff
@@ -146,11 +146,14 @@ auth:
...
   securePassword: true
-  ## @param auth.existingPasswordSecret Existing secret with RabbitMQ credentials (must contain a value for `rabbitmq-password` key)
-  ## e.g:
-  ## existingPasswordSecret: name-of-existing-secret
+  ## @param auth.existingSecret The name of an existing secret with RabbitMQ credentials
+  ## NOTE: When it's set, the previous `auth.password` parameter is ignored
+  ##
+  existingSecret: ""
+  ## @param auth.existingSecretPasswordKey Password key to be retrieved from existing secret
+  ## NOTE: ignored unless `auth.existingSecret` parameter is set
```
juan131, line 154: Introducing backwards incompatibilities comes with a high cost (specially in terms of UX). Therefore, we should avoid it unless it's strictly necessary.

Here you have a difference alternative so we can keep this change as a minor:

- Mark current `auth.existingPasswordSecret` as deprecated but keeping it:

```yaml
  ## @param auth.existingPasswordSecret Existing secret with RabbitMQ credentials (must contain a value for `rabbitmq-password` key)
  ## DEPRECATED. Use auth.existingSecret instead
  ##
  existingPasswordSecret: ""
  ## @param auth.existingSecret The name of an existing secret with RabbitMQ credentials
  ## NOTE: When it's set, the previous `auth.password` parameter is ignored
  ##
  existingSecret: ""
  ## @param auth.existingSecretPasswordKey Password key to be retrieved from existing secret
  ## NOTE: ignored unless `auth.existingSecret` parameter is set 
  ##
  existingSecretPasswordKey: ""
```

- Then, on `_helpers.tpl` (and other potential templates where the secret name is required) use `coalesce` so you give priority to the new parameter over the deprecated one. For instance:

```yaml
{{/*
Get RabbitMQ password secret name.
*/}}
{{- define "rabbitmq.secretName" -}}
{{- $existingSecret := coalesce .Values.auth.existingSecret .Values.auth.existingPasswordSecret -}}
{{- print (default (include "common.names.fullname" .) (tpl $existingSecret .)) -}}
{{- end -}}
```

juan131 (changes requested):
Thanks for this contribution! Please take a look to my comments, we should avoid breaking backwards compatibility.

bitnami/rabbitmq/templates/secret.yaml
maxvisser: do you want me to change it back into one file again also? I changed it to secret to show that I seperated the yamls inside to seperate files. (its now one secrets instead of multiple)

bitnami/rabbitmq/values.yaml
maxvisser: good suggestion. I found it hard to grasp how to make it backward compatible

maxvisser: this would still mean that we need to implement a fallback for the key if the secret name is provided (rabbitmq-password)

maxvisser:
Thank you for your review. 

I was thinking about implementing it for erlang aswell This would mean auth.existingErlangSecret would become also deprecated.

I was thinking about the following api:
auth.erlang.existingSecret
auth.erlang.existingErlangSecretKey 

I personally don't need it but it makes sense to implement it for erlang aswell. I believe this would be better as a seperate PR.

bitnami/rabbitmq/templates/_helpers.tpl
```diff
@@ -38,15 +38,52 @@ Return the proper Docker Image Registry Secret Names
...
+otherwise it generates a random value.
 */}}
-{{- define "rabbitmq.secretPasswordName" -}}
-    {{- if .Values.auth.existingPasswordSecret -}}
-        {{- printf "%s" (tpl .Values.auth.existingPasswordSecret $) -}}
+{{- define "getValueFromSecret" }}
+    {{- $len := (default 16 .Length) | int -}}
+    {{- $obj := (lookup "v1" "Secret" .Namespace .Name).data -}}
+    {{- if $obj }}
+        {{- index $obj .Key | trimAll "\"" | b64dec -}}
```
maxvisser, line 48: this is a change from the previous function; added a pipe to trim " as this was implemented in secrets.yaml:9; I believe to strip " from passwords. Its nice to keep if someone accidently adds quotations around a password.

bitnami/rabbitmq/templates/_helpers.tpl
```diff
@@ -240,7 +259,7 @@ otherwise it generates a random value.
     {{- $len := (default 16 .Length) | int -}}
     {{- $obj := (lookup "v1" "Secret" .Namespace .Name).data -}}
     {{- if $obj }}
-        {{- index $obj .Key | b64dec -}}
+        {{- index $obj .Key | trimAll "\"" | b64dec -}}
```
maxvisser, line 262: this trimAll was on line 9 in secrets.yaml; I believe its to trim accidently added quotations in passwords.

bitnami/rabbitmq/templates/secret.yaml
maxvisser: this is what I did.

maxvisser:
@juan131 not sure why the pipeline fails; seems like an exec timeout. all helm lint errors should be fixed.

juan131:
> I personally don't need it but it makes sense to implement it for erlang aswell. I believe this would be better as a seperate PR.

Yes please, let's keep it in a separate PR

maxvisser:
> > I personally don't need it but it makes sense to implement it for erlang aswell. I believe this would be better as a seperate PR.
> 
> Yes please, let's keep it in a separate PR

https://github.com/bitnami/charts/pull/22016

maxvisser:
Decided against adding a new value in this PR. Would be better to merge this and #21359 and have RFC if we want to change the API.

juan131 (dismissed):
LGTM

juan131 (approved):
LGTM

PR #21918, opened 2024-01-09
[bitnami/zookeeper] Enable automatic purging of snapshots by default

FHeilmann:
### Description of the change

The current zookeeper chart disables automatic purging of snapshots by setting `autopurge.purgeInterval` to 0. This results in snapshots accumulating indefinitely which will eventually cause all storage available to the zookeeper node to fill up. Once that happens, the zookeeper node will fail with errors that do not immediately point to a storage issue and, in the worst-case, will cause the entire deployment to become degraded (in our case, a clickhouse database, tables were set to read-only mode or were (incorrectly) reported as uninitialized).

The official clickhouse documentation calls this default behavior a ["time bomb that must be defused by the operator"](https://clickhouse.com/docs/en/operations/tips#zookeeper) and also suggests defaults for "a large deployment" which have been used in this PR.

Other available charts and documentation suggest similar defaults ([[1]](https://github.com/search?q=repo%3AAltinity%2Fclickhouse-operator%20autopurge&type=code), [[2]](https://www.ibm.com/support/pages/how-reduce-disk-space-logs-and-data-zookeeper), [[3]](https://docs.apigee.com/private-cloud/v4.19.01/apache-zookeeper-maintenance-tasks)). At the very least, all of them suggest to enable the feature, instead of leaving it disabled by default.

### Benefits

With this change, storage requirements of zookeeper will not grow indefinitely until they eventually fill up all available storage. Instead, once every hour all snapshots except the 10 newest will be purged.

This means users unaware of the "time bomb" defaults, who were almost guaranteed to run into issues with the current defaults eventually, will now be ok.

### Possible drawbacks

If, for any reason, the user of this chart expects all snapshots to be retained indefinitely, this change will modify default behavior and cause snapshots that were previously retained to be deleted.

### Applicable issues

- None

### Additional information

- We experienced this issue with a 1-shard-2-replica-2-zookeeper clickhouse deployment, where both zookeeper instances accumulated over 50 GB of snapshots in a very short time. These are some of the errors that we saw:
  - Shard: 
    - `Code: 667. DB::Exception: Table is not initialized yet. (NOT_INITIALIZED)`
  - Zookeeper: 
    - `ERROR [LearnerHandler-/....LearnerHandler@719] - Unexpected exception in LearnerHandler`
- After manually purging the snapshots (using `zkCleanup.sh /bitnami/zookeeper/data/ -n 10`) on both zookeeper nodes the database immediately recovered.

### Checklist


aoterolorenzo (approved):
Thanks for the contribution @FHeilmann ! LGTM

PR #21994, opened 2024-01-11
[bitnami/contour] Add contour multi az service

### LINKED ISSUE #21684: [bitnami/contour] multiple envoy LoadBalancer services for multi-AZ deployments
### Name and Version

15.0.1

Some cloud-proivder does not support region-wide LoadBalancers only zone-wide LoadBalancers. 
Which lead to unavailable services if the zone with the LoadBalancer is not available. 

In the case you do only have a zone-wide LoadBalancer and want an HA deployment of the ingress-controller you have to deploy a LoadBalancer for every zone. Like [here (scaleway/nginx)](https://www.scaleway.com/en/docs/tutorials/k8s-kapsule-multi-az/#nginx-ingress-controller-as-a-stateless-multi-az-application) described.

In common you have to deploy a additional LoadBalancer-Services for every zone with an annotation like `load-balancer.hetzner.cloud/location: hel1` or `service.beta.kubernetes.io/scw-loadbalancer-zone: fr-par-1`. 

Then failover between the LoadBalancers (and so zones) is done over DNS-checking. 

Right now the solution would be the extraDeploy filed - i think.

I am not that much aware about the internal structure of the Chart. But i could imagine a new field under `enovy.service` which name something like `replicas` and a list `replicas.separateAnnotations`.
Other suggestions are welcome.

using the extraDeploy field

wsn-dev:
### Description of the change

This change implements the possibility to render multiple services with different annotations and loadBalancerIP.

### Benefits

A multi region support with different loadbalancer services.

### Applicable issues

- fixes #21684

### Checklist


dgomezleon (approved):
LGTM. Thanks for contributing

PR #22233, opened 2024-01-16
[bitnami/keycloak] Support 'tpl' on ingress.secrets

gkacskovics:
### Description of the change

The change enables the usage of `tpl` function for the ingress.secrets list of objects in the `values.yaml`.

The `tpl` function is heavily used by the chart and it would help me avoiding to fork the repository if I could use the `tpl` function also in the `tls-secret.yaml`.

### Benefits

> The tpl function allows developers to evaluate strings as templates inside a template.

### Possible drawbacks

N/A

### Applicable issues

- fixes #

### Checklist


bitnami/keycloak/templates/tls-secret.yaml
```diff
@@ -9,16 +9,16 @@ SPDX-License-Identifier: APACHE-2.0
...
   labels: {{- include "common.labels.standard" ( dict "customLabels" $.Values.commonLabels "context" $ ) | nindent 4 }}
   {{- if $.Values.commonAnnotations }}
   annotations: {{- include "common.tplvalues.render" ( dict "value" $.Values.commonAnnotations "context" $ ) | nindent 4 }}
   {{- end }}
 type: kubernetes.io/tls
 data:
-  tls.crt: {{ .certificate | b64enc }}
-  tls.key: {{ .key | b64enc }}
+  tls.crt: {{ tpl .certificate $ | b64enc }}
+  tls.key: {{ tpl .key $ | b64enc }}
```
joancafom, line 21: I see the value of using `tpl` in `name`, but not really that much here. At the end of the day, it would be strange some kind of templating applied to a certificate or key content, isn't it?

bitnami/keycloak/templates/tls-secret.yaml
```diff
@@ -9,16 +9,16 @@ SPDX-License-Identifier: APACHE-2.0
 apiVersion: v1
 kind: Secret
 metadata:
-  name: {{ .name }}
+  name: {{ tpl .name $ }}
```
joancafom, line 12: Could you please use our own function for this instead of directly using `tpl`? It is used further below in L16:

```suggestion
  name: {{- include "common.tplvalues.render" ( dict "value" .name "context" . ) }}
```

joancafom (commented):
Thanks for your contribution. I added some comments as a review, could you please check them?

bitnami/keycloak/templates/tls-secret.yaml
gka-regnology: Yes, sure, I could do that. You can see the usage of both throughout the chart. I could not fully understand why isn't always "common.tplValues.render" used. My only guess was simpler cases are used with tpl and more complex ones with  "common.tplValues.render". This looked simpler for me.

bitnami/keycloak/templates/tls-secret.yaml
gka-regnology: Imagine including the chart as a subchart into an umbrella chart. The umbrella chart has several ingresses, all using a wildcard certificate.

In this following example the certificate is for "*.something.com". 

```
global:
  keycloak:
    hostname: keycloak.something.com
  ingress:
    tlsCert: somecert
    tlsCert: somekey


keycloak:
  (...)
  ingress:
    (...)
    hostname: "{{ .Values.global.keycloak.hostname }}"
    secrets:
      - name: "{{ printf \"%s-tls\" .Values.global.keycloak.hostname }}"
        certificate: "{{ .Values.global.ingress.tlsCert }}"
        key: "{{ .Values.global.ingress.tlsKey }}"
```

My goal is that the umbrella chart can be configured with minimal values from the outside. I would like to hide the complex configuration of the individual charts as much as possible.

bitnami/keycloak/templates/tls-secret.yaml
gkacskovics: FYI, this other guy is me, from the wrong account.

joancafom (approved):

bitnami/keycloak/templates/tls-secret.yaml
joancafom: Gotcha! I see your point now, thanks 😄

bitnami/keycloak/templates/tls-secret.yaml
joancafom: > I could not fully understand why isn't always "common.tplValues.render" used

`tpl` is the legacy approach used in our charts and `common.tplValues.render` the new one. In newer charts (and features) you will exclusively see the latter being used 😁

PR #22250, opened 2024-01-17
[bitnami/redis] - add support for additional-endpoints in service/pod monitor

shlomitubul:
**Description of the change**
 Add support for passing endpoints for Service/Pod-Monitor, so users can scrape metrics from both Redis and sentinel.
 Tested as working in both default & new values

**Benefits**
scraping multiple containers using single redis exporter containers

**Possible drawbacks**
none

**Applicable issues**

**Additional information**


bitnami/redis/values.yaml
```diff
@@ -1816,6 +1826,16 @@ metrics:
     ## @param metrics.podMonitor.targetLimit Limit of how many targets should be scraped
     ##
     targetLimit: false
+    ## @param metrics.podMonitor.additionalEndpoints  Additional endpoints to scrape (e.g sentinel)
+    ##
+    additionalEndpoints: []
+      # uncomment in order to scrape sentinel metrics
```
shlomitubul, line 1832: there is no port named http-metrics, it only valid when using ServiceMonitoer & metrics-svc, i'll  fix this comment

dgomezleon (dismissed):
LGTM. Thanks for contributing

dgomezleon (approved):

PR #22401, opened 2024-01-18
[bitnami/flux] fix: :lock: Move service-account token auto-mount to pod declaration

javsalgar:

### Description of the change

This PR sets all ServiceAccount automountServiceAccountToken=false by default, as the token mounting should be set in the pod declaration instead (if a new pod uses this service account and the token gets automatically mounted, it could be problematic in terms of security). This PR also adds automountServiceAccountToken in the pod declaration as a new value, which can be configured by users in case they want to use an external token.

### Benefits

Charts become more security-compliant

### Possible drawbacks

n/a

### Checklist


bitnami/flux/values.yaml
```diff
@@ -218,6 +218,9 @@ kustomizeController:
   ## @param kustomizeController.args Override default container args (useful when using custom images)
   ##
   args: []
+  ## @param kustomizeController.automountServiceAccountToken Mount Service Account token in pod
+  ##
+  automountServiceAccountToken: true
```
joancafom, line 223: shouldn't it be set to `false` instead?

bitnami/flux/values.yaml
```diff
@@ -664,6 +667,9 @@ helmController:
   ## @param helmController.args Override default container args (useful when using custom images)
   ##
   args: []
+  ## @param helmController.automountServiceAccountToken Mount Service Account token in pod
+  ##
+  automountServiceAccountToken: true
```
joancafom, line 672: shouldn't it be set to `false` instead?

bitnami/flux/values.yaml
```diff
@@ -1112,6 +1118,9 @@ sourceController:
   ## @param sourceController.args Override default container args (useful when using custom images)
   ##
   args: []
+  ## @param sourceController.automountServiceAccountToken Mount Service Account token in pod
+  ##
+  automountServiceAccountToken: true
```
joancafom, line 1123: shouldn't it be set to `false` instead?

bitnami/flux/values.yaml
```diff
@@ -1716,6 +1725,9 @@ notificationController:
   ## @param notificationController.args Override default container args (useful when using custom images)
   ##
   args: []
+  ## @param notificationController.automountServiceAccountToken Mount Service Account token in pod
+  ##
+  automountServiceAccountToken: true
```
joancafom, line 1730: shouldn't it be set to `false` instead?

bitnami/flux/values.yaml
```diff
@@ -2162,6 +2174,9 @@ imageAutomationController:
   ## @param imageAutomationController.args Override default container args (useful when using custom images)
   ##
   args: []
+  ## @param imageAutomationController.automountServiceAccountToken Mount Service Account token in pod
+  ##
+  automountServiceAccountToken: true
```
joancafom, line 2179: shouldn't it be set to `false` instead?

bitnami/flux/values.yaml
```diff
@@ -2608,6 +2623,9 @@ imageReflectorController:
   ## @param imageReflectorController.args Override default container args (useful when using custom images)
   ##
   args: []
+  ## @param imageReflectorController.automountServiceAccountToken Mount Service Account token in pod
+  ##
+  automountServiceAccountToken: true
```
joancafom, line 2628: shouldn't it be set to `false` instead?

bitnami/flux/values.yaml
javsalgar: They require access to the K8s API so they need to mount the SA token

bitnami/flux/values.yaml
javsalgar: They require access to the K8s API so they need to mount the SA token

bitnami/flux/values.yaml
javsalgar: They require access to the K8s API so they need to mount the SA token

bitnami/flux/values.yaml
javsalgar: They require access to the K8s API so they need to mount the SA token

bitnami/flux/values.yaml
javsalgar: They require access to the K8s API so they need to mount the SA token

bitnami/flux/values.yaml
javsalgar: They require access to the K8s API so they need to mount the SA token

joancafom (approved):

PR #22676, opened 2024-01-23
[bitnami/redis] Fix the PodMonitor implementation

wilmardo:
### Description of the change

The current `PodMonitor` in the chart isn't working, it has the labels and port of the metrics service not the pods. I think someone got confused that the `PodMonitor` looks at the `Service`:

> The metrics service is required for the podMonitor metrics to work
https://github.com/bitnami/charts/pull/20517#pullrequestreview-1739776974

This is not the case, the `PodMonitor` scrapes the pods directly without needing a `Service` so the `PodMonitor` selector labels and port should match the pods. Not the `Service`

Relevant pod yaml for the labels 
```yaml
apiVersion: v1
kind: Pod
metadata:
  labels:
    app.kubernetes.io/component: node
    app.kubernetes.io/instance: redis1
    app.kubernetes.io/name: redis
    app.kubernetes.io/version: 7.2.4
    controller-revision-hash: redis1-node-65778dc7bc
    helm.sh/chart: redis-1.0.0
    statefulset.kubernetes.io/pod-name: redis1-node-0

```
And for the port:

```yaml
apiVersion: v1
kind: Pod
spec:
  containers:
  - image: docker.io/bitnami/redis-exporter:1.56.0-debian-11-r0
    ports:
    - containerPort: 9121
      name: metrics
      protocol: TCP
```

### Benefits

Current setup isn't working, made these changes and the pods came available as targets and the metrics were scraped.

### Checklist


wilmardo:
We can start a discussion if the `PodMonitor` even serves a function in this chart. Since scraping through the service also scrapes every instance like the `PodMonitor` does:

> It will scrape all pods behind the service, because the Service maintains an Endpoints object. That's where Prometheus discovers the targets from.
https://github.com/prometheus-operator/prometheus-operator/issues/3119#issuecomment-631409690


In the current state it is broken as well so I would be surprised if anyone is using it at the moment so if removal is on the table this feels like the moment.

IMO it is the choice of providing a sensible default (`ServiceMonitor`) vs flexibility and supporting all usecases and edgecases (`ServiceMonitor` and `PodMonitor`) both available.

bitnami/redis/templates/metrics-svc.yaml
```diff
@@ -3,7 +3,7 @@ Copyright VMware, Inc.
 SPDX-License-Identifier: APACHE-2.0
 */}}
 
-{{- if .Values.metrics.enabled }}
+{{- if and .Values.metrics.enabled .Values.metrics.service.enabled }}
```
fmulero, line 6: I think we could use `.Values.metrics.serviceMonitor.enabled` WDYT? The new value won't be required.

bitnami/redis/templates/podmonitor.yaml
```diff
@@ -48,5 +48,5 @@ spec:
       - {{ include "common.names.namespace" . | quote }}
   selector:
     matchLabels: {{- include "common.labels.matchLabels" ( dict "customLabels" .Values.commonLabels "context" $ ) | nindent 6 }}
-      app.kubernetes.io/component: metrics
+      app.kubernetes.io/component: node
```
fmulero, line 51: Why are you using this label? I think we are not using in latests versions.

fmulero (commented):
Hi @wilmardo 👋  thanks a lot for your contribution!

I left a few comments, Could you take them a look?

bitnami/redis/templates/metrics-svc.yaml
wilmardo: Hmm I think it would be nice to make it possible to retain the metrics `Service` without the `ServiceMonitor`. When not using the prometheus operator you can annotate the `Service` to have it scraped without using the `ServiceMonitor`:
```yaml
kind: Service
annotations:
  prometheus.io/port: "80"
  prometheus.io/scrape: "true"
```

Pretty sure those might still be around in the wild

bitnami/redis/templates/podmonitor.yaml
wilmardo: Good catch! We are using sentinel so I had this one in my pods but I see that it isn't used for the master/replica setup. Removed it to just use the `matchLabels`:

```yaml
        labels:
          app.kubernetes.io/instance: redis2
          app.kubernetes.io/name: redis
```

Should be specific enough

wilmardo:
@fmulero Thanks for the review! Updates made and ready for review again :)

bitnami/redis/templates/metrics-svc.yaml
fmulero: Ok, LGTM

fmulero (approved):
Thanks a lot @wilmardo  for your contribution. LGTM 👍

PR #22698, opened 2024-01-24
[bitnami/thanos] Add revisionHistoryLimit option for each component

jakuboskera:
### Description of the change

Adds revisionHistoryLimit option for each Thanos component

### Benefits

Able to select how many revisions I want to keep in Kubernetes cluster for each component, e.g. when using GitOps there is no need to have revisionHistoryLimit greater than 1 as rollback is done using Git.

### Possible drawbacks

none

### Applicable issues

none

### Additional information

none

### Checklist


bitnami/thanos/Chart.yaml
```diff
@@ -35,4 +35,4 @@ maintainers:
 name: thanos
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/thanos
-version: 12.21.2
+version: 12.21.3
```
FraPazGal, line 38: We should bump the minor version per semver policy as we are introducing a new feature.

```suggestion
version: 12.22.0
```

FraPazGal (commented):
Thanks for the contribution @jakuboskera! I added a minor comment as a review, please check it when you have some time.

bitnami/thanos/Chart.yaml
jakuboskera: Yeah that's true, I was about that it is part of almost all Helm Charts, and it is missing here, so I bumped the minor. But I leave the decision up to you.

FraPazGal (approved):
LGTM!

Issue #22694, opened 2024-01-24
[bitnami/mlflow]: automountServiceAccountToken has been moved


angegar:
### Name and Version

bitnami/mlflow:0.6.0

### What architecture are you using?

None

### What steps will reproduce the bug?

Deploy the chart version 0.3.0
Upgrade the chart to version 0.6.0
Deployment fails

### What do you see instead?

Service account permission issues

### Additional information

The root cause is that the automountServiceAccountToken has been moved out of the serviceAccountSection to be place in :** {{ .Values.tracking.automountServiceAccountToken }}**

The chart version has been incremented with a minor whereas I think it should be a major because the user interface changed.

javsalgar:
Hi!

Could you explain the use case? Are you using mlflow with some sort of Kubernetes integration?

angegar:
Yes sorry,

I am using mlfow helm chart, and the configuration field automountServiceAccountToken  has been moved from the tracking.serviceAccount section to the tracking one. Thus, I guess it is a breaking change.

javsalgar:
Hi, 

We did not update the minor version because in the original chart we were not mounting the service account token. However, could you let us know why you require the SA token to be in the pod? We can revert the change and then bump the major version again. If you want to speed up the process, you can submit a PR and we will review it.

angegar:
It is because I spin up infrastructure such as a PostGreSQL. The infrastructure is managed with GCP config connector so that I need to be able request the object status in an init container before starting mlflow.

PR #22699, opened 2024-01-24
[bitnami/mlflow] feat: 🔒 make S3 secret optional

schmoesta:
### Description of the change

This patch introduces a new value `externalS3.useCredentialsInSecret` that lets users opt out of setting AWS credentials for connecting to S3 using a kubernetes secret.

When this value is set to `true` (the default), the chart behaves in the exact same way as before, creating and/or referencing a secret for setting the `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` variables.

When this value is set to `false`, `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` are not set and no secret is created and/or referenced.

### Benefits

Previously, enabling external S3 meant that the chart was setting the `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` environment variables with values from a kubernetes secret. However, users may want to use other means of providing these credentials, e.g. on EKS, where it's possible to use a web identity token file, e.g. using [IRSA](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html). However, if `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` are set, they will take precedence, because they come first in the credentials provider chain.

### Possible drawbacks

None that I am immediately aware of.

### Applicable issues

None

### Additional information

See [here](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/credentials.html#configuring-credentials) for how boto3 looks for credentials.

### Checklist

### DIFF
--- a/bitnami/mlflow/Chart.yaml
+++ b/bitnami/mlflow/Chart.yaml
@@ -43,4 +43,4 @@ name: mlflow
 sources:
 - https://github.com/bitnami/containers/tree/main/bitnami/mlflow
 - https://github.com/mlflow/mlflow
-version: 0.6.1
+version: 0.7.0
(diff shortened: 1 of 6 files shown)


schmoesta:
Possibly #22694 needs to be addressed first, I believe that the previous version increment might indeed have been incorrect. Let me know what you decide and if I need to change the version in the PR.

dgomezleon (approved):
LGTM. Thanks for contributing

PR #22736, opened 2024-01-26
[bitnami/aspnet-core] fix: 🐛 Set workingDir empty by default, add param `useContainercommand`

### LINKED ISSUE #22558: [bitnami/aspnet-core] deployment.yaml should use default command from image instead of `dotnet`
### Name and Version

bitnami/aspnet-core 5.3.0

In deployment.yaml ,Currently when using prebuild image, I need to set args like `args: ["my-web-app.dll"] `  
But I already defined `ENTRYPOINT` in Dockerfile, it should use default command in image.

Removed default command `dotnet`, if command is null, it will use default command in image.
`workingDir` is also predefined in Dockerfile, it should be empty when `.Values.appFromExternalRepo.enabled=false` 
```yaml
      containers:
        - name: aspnet-core
          image: {{ include "aspnet-core.image" . }}
          imagePullPolicy: {{ .Values.image.pullPolicy | quote }}
          securityContext:
            runAsUser: {{ .Values.containerSecurityContext.runAsUser }}
            runAsNonRoot: {{ .Values.containerSecurityContext.runAsNonRoot }}
          {{- end }}
          {{- if .Values.lifecycleHooks }}
          {{- end }}
          {{- if .Values.appFromExternalRepo.enabled }}
          workingDir: /app
          {{- end }}
          command:
          {{- if .Values.command }}
          {{- include "common.tplvalues.render" (dict "value" .Values.command "context" $) | nindent 12 }}
          {{- else if .Values.appFromExternalRepo.enabled }}
          {{- include "common.tplvalues.render" (dict "value" .Values.appFromExternalRepo.startCommand "context" $) | nindent 12 }}
          {{- end }}
```

_No response_

yip102011:
### Description of the change

Set command and workingDir empty by default when using custom image `.Values.appFromExternalRepo.enabled=false`.

### Benefits

Visual Studio default Dockerfile has defined `ENTRYPOINT ["dotnet", "my-app.dll"]` and `WORKDIR /app`
For using custom image `.Values.appFromExternalRepo.enabled=false`.
- Set `command` to empty to respect Dockerfile `ENTRYPOINT`, and I don't have to set `args: ["my-app.dll"]`.
- Set `workingDir` to empty to respect Dockerfile `WORKDIR` so i can control working dir in Dockerfile

### Possible drawbacks

n/a

### Applicable issues

- fixes #22558

### Additional information

n/a

### Checklist


FraPazGal (commented):
Hello @yip102011, I agree having the `workingDir` hardcoded is not a good practice and could cause practice. As such, only using it when `appFromExternalRepo.enabled=true` or even parametrised seems like a good option.

That said, I have some concerns regarding the `command` change. We have that trailing `else` so our `bitnami/aspnet-core` image, which by defaults has `bin/bash` as [entrypoint](https://github.com/bitnami/containers/blob/b2294543f243c453d44aebc30ace8f9aaa3b4476/bitnami/aspnet-core/7/debian-11/Dockerfile#L58), can run dotnet in the chart by default. Having already the option of customising both `command` and `args`, I don't think the change is necessary given we'll need to modify our container's default behaviour.

yip102011:
How about I add a new parameter `useContainerCommand`, default `false`, command when be empty if `comment` is not override , `.Values.appFromExternalRepo.enabled=false` and `useContainerCommand=true`.

bitnami/aspnet-core/templates/deployment.yaml
```diff
@@ -114,12 +114,16 @@ spec:
...
+          {{- if .Values.appFromExternalRepo.enabled }}
           workingDir: /app
+          {{- end }}
           command:
           {{- if .Values.command }}
           {{- include "common.tplvalues.render" (dict "value" .Values.command "context" $) | nindent 12 }}
           {{- else if .Values.appFromExternalRepo.enabled }}
           {{- include "common.tplvalues.render" (dict "value" .Values.appFromExternalRepo.startCommand "context" $) | nindent 12 }}
+          {{- else if .Values.useContainerCommand }}
+            {{- /* left empty if use container command */ -}}
```
FraPazGal, line 126: @yip102011 Correct me if I'm missing something but as it is now we'll just need to use a custom `command` and `vars` to replicate your Dockerfile command right? I agree it can be a minor nuisance caused by the chart not using the container's default, but I don't think this justifies adding a new parameter.

A middle point would be to differentiate between `command=null` and `!=null`. We could add the following logic to avoid setting anything when using `null`:

```suggestion
          {{- if .Values.command }}
          {{- include "common.tplvalues.render" (dict "value" .Values.command "context" $) | nindent 12 }}
          {{- else if kindIs "invalid" .Values.command }}
          {{- /* will use container container */ -}}
          {{- else if .Values.appFromExternalRepo.enabled }}
          {{- include "common.tplvalues.render" (dict "value" .Values.appFromExternalRepo.startCommand "context" $) | nindent 12 }}
```

wdyt?

yip102011:
ok, I agree the `kindIs "invalid"` solution.
I think is important to sperate the responsive between devops and developer.  
devops should not be required to know the dll name.

Imagine below conversation.
**devops guy**: hi, what is the command to run your program? I need to make the container image.
**developer**: I don't know, I click the start button on visual studio.
**devops guy**: ok, can I see your code for the dll name?
**developer**: no, we are diff team, you don't have permission.
**devops guy**: so what should I do?
**developer**: Not my problem.

bitnami/aspnet-core/values.yaml
```diff
@@ -80,6 +80,9 @@ image:
   ##
   pullSecrets: []
   debug: false
+## @param useContainerCommand Use container command instead of override it (useful when using custom images)
+##
+useContainerCommand: false
```
FraPazGal, line 85: This shouldn't be necessary any longer.

bitnami/aspnet-core/templates/deployment.yaml
```diff
@@ -114,10 +114,14 @@ spec:
...
           lifecycle: {{- include "common.tplvalues.render" (dict "value" .Values.lifecycleHooks "context" $) | nindent 12 }}
           {{- end }}
+          {{- if .Values.appFromExternalRepo.enabled }}
           workingDir: /app
+          {{- end }}
           command:
           {{- if .Values.command }}
           {{- include "common.tplvalues.render" (dict "value" .Values.command "context" $) | nindent 12 }}
+          {{- else if kindIs "invalid" .Values.command }}
+          {{- /* will use container container */ -}}
```
FraPazGal, line 124:
```suggestion
          {{- /* will use the container's command */ -}}
```

bitnami/aspnet-core/values.yaml
```diff
@@ -80,6 +80,9 @@ image:
...
 ## @param command Override default container command (useful when using custom images)
```
FraPazGal, line 86: We should add some comments about the new logic:

```suggestion
## @param command [object,nullable] Override custom pod command (useful when using custom images)
## If empty, it will default to "dotnet". Set to null to use the default container command 
```

FraPazGal (commented):
The PR looks almost ready @yip102011. I have added a few comments as a review

PR #22716, opened 2024-01-25
[bitnami/redis] Do not create master and replica serviceaccounts when using sentinel

wilmardo:
### Description of the change

Currently when choosing to use sentinel you get all the serviceaccounts not just the one needed.

Currently is creates the master and replica service accounts:
```
default          0         8h
redis1           0         73m
redis1-master    0         64s
redis1-replica   0         64s
redis2           0         51m
redis2-master    0         62s
redis2-replica   0         62s
```

### Benefits

Less clutter, removes unneeded resources. More inline with the other resources in the `master/` and `replica/` template folder. They all have this if check, this is the one resource where it was missing.

### Checklist

- [x] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/). This is *not necessary* when the changes only affect README.md files.
- [x] Variables are documented in the values.yaml and added to the `README.md` using [readme-generator-for-helm](https://github.com/bitnami/readme-generator-for-helm)
- [x] Title of the pull request follows this pattern [bitnami/<name_of_the_chart>] Descriptive title
- [x] All commits signed off and in agreement of [Developer Certificate of Origin (DCO)](https://github.com/bitnami/charts/blob/main/CONTRIBUTING.md#sign-your-work)

PR #22747, opened 2024-01-26
[bitnami/redis]: Fix standalone redis missing service account

### LINKED ISSUE #22746: Redis Standalone serviceAccount
### Name and Version

bitnami/redis 18.9.0

When we set architecture to standalone the serviceAccount was created, with the new Tag we miss that serviceAccount with standalone architecture

Leave it as enabled serviceAccount when use standalone

_No response_

Wielewout:
### Description of the change

Since 18.8.3 (https://github.com/bitnami/charts/pull/22716) no service account is created with architecture set to standalone.

In this case the serviceaccount and application condition didn't match for master (instead the condition for replicas was used for the master service account).

### Applicable issues

- fixes #22746

### Checklist

### DIFF
--- a/bitnami/redis/templates/master/serviceaccount.yaml
+++ b/bitnami/redis/templates/master/serviceaccount.yaml
@@ -3,7 +3,7 @@ Copyright VMware, Inc.
 SPDX-License-Identifier: APACHE-2.0
 */}}
 
-{{- if and .Values.master.serviceAccount.create (eq .Values.architecture "replication") (not .Values.sentinel.enabled) }}
+{{- if and .Values.master.serviceAccount.create (or (not (eq .Values.architecture "replication")) (not .Values.sentinel.enabled)) }}
 apiVersion: v1
 kind: ServiceAccount
 automountServiceAccountToken: {{ .Values.master.serviceAccount.automountServiceAccountToken }}
--- a/bitnami/redis/Chart.yaml
+++ b/bitnami/redis/Chart.yaml
@@ -34,4 +34,4 @@ maintainers:
 name: redis
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/redis
-version: 18.9.0
+version: 18.9.1


pinkfloydx33:
Our Flux managed install in our dev environment failed to auto-upgrade. The helm release times out waiting for the Statefulset which cannot start since the ServiceAccount got removed. Looking forward to a quick fix here!

kallaics:
Version 18.9.0 affected too with this issue.

bitnami/redis/templates/master/serviceaccount.yaml
```diff
@@ -3,7 +3,7 @@ Copyright VMware, Inc.
 SPDX-License-Identifier: APACHE-2.0
 */}}
 
-{{- if and .Values.master.serviceAccount.create (eq .Values.architecture "replication") (not .Values.sentinel.enabled) }}
+{{- if and .Values.master.serviceAccount.create (or (not (eq .Values.architecture "replication")) (not .Values.sentinel.enabled)) }}
```
migruiz4, line 6: I think this logic can be simplified by removing `.Values.architecture` from the equation.

The current conditional will cause the following:
* arch: standalone, sentinel: false -> master-SA created
* arch: standalone, sentinel: true -> Mode not supported, but master-SA enabled
* arch: replication, sentinel: false -> master-SA NOT created
* arch: replication, sentinel: true -> master-SA created

```suggestion
{{- if and .Values.master.serviceAccount.create (not .Values.sentinel.enabled) }}
```

The proposed conditional will cause the following:
* arch: standalone, sentinel: false -> master-SA created
* arch: standalone, sentinel: true -> Mode not supported and master-SA NOT enabled
* arch: replication, sentinel: false -> master-SA created
* arch: replication, sentinel: true -> master-SA NOT created

migruiz4 (commented):
Hi @Wielewout,

Could you please take a look at my suggestion?

bitnami/redis/templates/master/serviceaccount.yaml
Wielewout: The last two results are reversed, so the original logic gives:
> The current conditional will cause the following:
> 
> arch: standalone, sentinel: false -> master-SA created
> arch: standalone, sentinel: true -> Mode not supported, but master-SA enabled
> arch: replication, sentinel: false -> master-SA created
> arch: replication, sentinel: true -> master-SA NOT created

If the standalone/sentinel enabled combintation is a "don't care", then the suggested logic is also fine. Though maybe there should be a clear failure message or something then that it's not supported.

```
{{- if and (eq .Values.architecture "standalone") (.Values.sentinel.enabled) -}}
fail "A standalone redis architecture with sentinel is not supported"
{{- end -}}
```

If the idea is to just install it as configured (even though it's unsupported), then the original logic is the one to go for.

Any changes here should also be portrayed over here btw: https://github.com/bitnami/charts/blob/bc4b1dcd3ba113f5a1e75b6a65ba522acc241dc5/bitnami/redis/templates/master/application.yaml#L7

Wielewout: Probably we should create a helper for both master and replication so that there is no duplicated logic to keep in sync. First let's agree on what the logic inside should be :smile:

PR #22752, opened 2024-01-26
[bitnami/opensearch] Allow templating of secret names

maxnitze:
### Description of the change

Put secret names through `common.tplvalues.render` helper, before rendering to containers.

### Benefits

Templating of extra env vars possible.

### Possible drawbacks

None

### Applicable issues

None

### Additional information

None

### Checklist

### DIFF
--- a/bitnami/opensearch/templates/_helpers.tpl
+++ b/bitnami/opensearch/templates/_helpers.tpl
@@ -422,7 +422,11 @@ Return true if an authentication credentials secret object should be created
 Return the Opensearch authentication credentials secret name
 */}}
 {{- define "opensearch.secretName" -}}
-{{- default (include "common.names.fullname" .) .Values.security.existingSecret  -}}
+{{- if .Values.security.existingSecret -}}
+    {{- printf "%s" (tpl .Values.security.existingSecret $) -}}
+{{- else -}}
+    {{- printf "%s" (include "common.names.fullname" .) -}}
+{{- end -}}
 {{- end -}}
 
 {{/*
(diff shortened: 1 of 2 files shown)


bitnami/opensearch/templates/_helpers.tpl
```diff
@@ -422,7 +422,7 @@ Return true if an authentication credentials secret object should be created
 Return the Opensearch authentication credentials secret name
 */}}
 {{- define "opensearch.secretName" -}}
-{{- default (include "common.names.fullname" .) .Values.security.existingSecret  -}}
+{{- include "common.tplvalues.render" (dict "value" .Values.security.existingSecret "context" $) | default (include "common.names.fullname" .) -}}
```
migruiz4, line 425: Could you please use this approach instead? I would prefer keeping the `default` as it is or replace it with `if/else` as in other charts if its content gets too long:
```suggestion
{{- default (include "common.names.fullname" .) (tpl .Values.security.existingSecret $) -}}
```

In this case, because only 'string' values are expected, we can use `tpl` directly, similar to other charts:
https://github.com/bitnami/charts/blob/1c7dfc16b41dd75f08aa261efb74fa302ec9f2f4/bitnami/wordpress/templates/_helpers.tpl#L70-L79

bitnami/opensearch/templates/_helpers.tpl
```diff
@@ -438,7 +438,7 @@ Return true if a TLS password secret object should be created
 Return the Opensearch TLS password secret name
 */}}
 {{- define "opensearch.tlsPasswordsSecret" -}}
-{{- default (printf "%s-tls-pass" (include "common.names.fullname" .)) .Values.security.tls.passwordsSecret -}}
+{{- include "common.tplvalues.render" (dict "value" .Values.security.tls.passwordsSecret "context" $) | default (printf "%s-tls-pass" (include "common.names.fullname" .)) -}}
```
migruiz4, line 441: Same applies here:
```suggestion
{{- default (printf "%s-tls-pass" (include "common.names.fullname" .)) (tpl .Values.security.tls.passwordsSecret $) -}}
```

migruiz4 (changes requested):
Thank you for your contribution @maxnitze. 
Could you please take a look at my comments?

bitnami/opensearch/templates/_helpers.tpl
maxnitze: Updated.

bitnami/opensearch/templates/_helpers.tpl
maxnitze: Updated.

migruiz4 (approved):
Thank you very much for applying the suggestions, looks good to me!

PR #22807, opened 2024-01-28
[bitnami/postgresql] replace initialDelaySeconds by startupProbe

jfcoz:
### Description of the change

Replace long initialDelaySeconds in livenessProbes/readinessProbes with startupProbe enabled.

### Benefits

Pods becomes Ready faster

### Applicable issues

- fixes #

### Checklist


joancafom:
Hi @jfcoz 

Tuning this type of parameters is tricky because we should ensure that this works well in every type of clusters and not just a specific configuration. IMHO, I would not update them if they do not solve a real issue. 

The cool thing about the params is that they allow people to set the right values for their set up.

jfcoz:
Hi @joancafom 

Currently the initialDelaySeconds is 3 times the periodSeconds, it is clearly used to wait for the service to start before the first liveness/readiness probes, and this is the goal of the startupProbe.

This is not really a problem. The only problem is that each time we use bitnami/postgresql as a dependency, we need to override the default values to correctly setup probes, and it would be easier to get a good setup by default.

Theses new defaults will also divide startup time by 3.

joancafom:
> Currently the initialDelaySeconds is 3 times the periodSeconds, it is clearly used to wait for the service to start before the first liveness/readiness probes, and this is the goal of the startupProbe.

@jfcoz I see your point here 😁 Yes, I agree it would be better to use the _startupProbe_s to wait for the services to start and reduce the times in the other already enabled ones.

>This is not really a problem. The only problem is that each time we use bitnami/postgresql as a dependency, we need to override the default values to correctly setup probes, and it would be easier to get a good setup by default.

> Theses new defaults will also divide startup time by 3.

Nevertheless, although I agree on enabling the _startupProbe_s, I think the "eventual time to start" (that is, the time it takes for `startUpProbe` + `livenessProbe`/`readinessProbe` to mark the pods as live/ready) should be kept the same. The goal of having these parameters exposed is that users can adapt the values to their needs.

We have to take into account that not every user has the same cluster set up, and while certain configurations can be good for some users (and speed up times) it may impact others running different requirements 😞 . So I think that a change like this that modifies the default values can indirectly affects others deploying the software on different (and maybe slower) machines.

PR #22935, opened 2024-01-31
Revert "[bitnami/redis] feat: :lock: Enable networkPolicy"

onedr0p:
https://github.com/bitnami/charts/issues/22934
https://github.com/bitnami/charts/issues/22922
https://github.com/bitnami/charts/pull/22738#issuecomment-1917502517

Reverts bitnami/charts#22738

PR #22738, opened 2024-01-26
[bitnami/redis] feat: :lock: Enable networkPolicy

javsalgar:
Signed-off-by: Javier Salmeron Garcia <jsalmeron@vmware.com>

### Description of the change

This PR normalizes the use of NetworkPolicy in the chart. Adds all Bitnami standards for NetworkPolicies as well as enabling it by default, in order to comply with security checklists.

### Benefits

More security in the chart

### Applicable issues

- fixes #

### Checklist

- [ ] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/). This is *not necessary* when the changes only affect README.md files.
- [x] Variables are documented in the values.yaml and added to the `README.md` using [readme-generator-for-helm](https://github.com/bitnami/readme-generator-for-helm)
- [x] Title of the pull request follows this pattern [bitnami/<name_of_the_chart>] Descriptive title
- [x] All commits signed off and in agreement of [Developer Certificate of Origin (DCO)](https://github.com/bitnami/charts/blob/main/CONTRIBUTING.md#sign-your-work)

PR #22978, opened 2024-01-31
[bitnami/redis] fix: env value for REDIS_EXPORTER_WEB_LISTEN_ADDRESS

### LINKED ISSUE #22922: [bitnami/redis] REDIS_EXPORTER_WEB_LISTEN_ADDRESS incorrectly set causing metrics pod to crashloop
### Name and Version

bitnami/redis 18.11.0

amd64

1. depoly redis-master via `HEAD`, enable `metrics`, but don't change any readinessProbe related values
2. observe `metrics` fails to perform readiness probe

I am, but I believe the default values (with `metrics.enabled = true`) will also trigger this condition. Here are the relevant values that I am setting: 
```yaml
auth:
  enabled: true
  password: "superUltraMegaSecret"
master:
  resources:
    limits:
      cpu: 4000m
      memory: 256Mi
    requests:
      cpu: 500m
      memory: 128Mi
metrics:
  enabled: true
  resources:
    limits:
      cpu: 500m
      memory: 256Mi
    requests:
      cpu: 100m
      memory: 64Mi
namespace: monitoring
networkPolicy:
  enabled: false
prometheusRule:
  enabled: true
  # https://github.com/bitnami/charts/blob/dfbf28240eca7b8732d83d07f10d7a3ca0788e3a/bitnami/redis/values.yaml
  rules: []
replica:
  autoscaling:
    enabled: true
    minReplicas: 2
    maxReplicas: 10
    targetCPU: "75"
  resources:
    limits:
      cpu: 1500m
      memory: 256Mi
    requests:
      cpu: 500m
      memory: 128Mi
serviceMonitor:
  enabled: true
```

The `metrics` pod should not crash.

### What do you see instead?

The `metrics` pod crashes, rendering the `master` unavailable.

### Additional information

I believe this change is the culprit: https://github.com/bitnami/charts/commit/f1c7b0d6c691bccf60c3294e1e607184ee462e1e#r138079079

a5r0n:
### Description of the change
fix regression intrudeced in #22738

closes #22922 #22935
see https://github.com/bitnami/charts/pull/22738#issuecomment-1917502517

### Applicable issues

closes #22922 #22935
see https://github.com/bitnami/charts/pull/22738#issuecomment-1917502517

- fixes #

### Checklist


a5r0n:
@javsalgar can you please take a look?

bitnami/redis-cluster/templates/redis-statefulset.yaml
```diff
@@ -342,7 +342,7 @@ spec:
               value: {{ template "redis-cluster.tlsCACert" . }}
             {{- end }}
             - name: REDIS_EXPORTER_WEB_LISTEN_ADDRESS
-              value: {{ printf ":%d" .Values.metrics.contanierPorts.http }}
+              value: value: ":{{ .Values.metrics.containerPorts.http }}"
```
mars64, line 345: Typo: `value: value:`

orangewolf:
though this moves in the correct direction, I don't think the result will be quite right. My reading of this would be that the default value will be `9121` but that the correct value for this field should be something more like `0.0.0.0:9121`

a5r0n:
The value will be `:9121` and is just fine, actually it is the same as 0.0.0.0:9122, anyway, right now the chart is broken so any fix should be merged ASAP..

bitnami/redis/Chart.yaml
```diff
@@ -34,4 +34,4 @@ maintainers:
 name: redis
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/redis
-version: 18.11.0
+version: 18.12.0
```
lytxashleywilliams, line 37: Would this not need to be `18.13.0`

Looks to be failing the tests due to non-matching version bumps `num_version_bumps`

```
script: let message='Detected changes in charts without version bump in Chart.yaml. Charts changed: 2. Version bumps detected: 1';
```

bitnami/redis-cluster/templates/redis-statefulset.yaml
```diff
@@ -342,7 +342,7 @@ spec:
               value: {{ template "redis-cluster.tlsCACert" . }}
             {{- end }}
             - name: REDIS_EXPORTER_WEB_LISTEN_ADDRESS
-              value: {{ printf ":%d" .Values.metrics.containerPorts.http }}
```
javsalgar, line 345: Could you change it to `%v`? that should fix the issue

bitnami/redis/Chart.yaml
javsalgar: As this is not adding any feature, I'd say it should be a patch bump

javsalgar (changes requested):
Thank you so much for the PR! Changes requested

javsalgar:
As this is critical, I included this fix on the PR I previously created fixing the network policy https://github.com/bitnami/charts/pull/22955

a5r0n:
@javsalgar 
Combining unrelated fixes in the same PR is a bad idea, especially when there is already an open PR by volunteer, it doesn't help to get new contributors when their work is rejected at the end for no good reason, you could add the suggestions you requested without waiting for my approval.

PR #23410, opened 2024-02-13
[bitnami/common] feat: :sparkles: Add support for resource presets

javsalgar:

### Description of the change

This PR adds support for a future improvement in our charts: resource presets. Currently, Bitnami charts do not set the `resources` object by default. While it is true that this is a task of the operator to adapt the resources to their use case, we do not explicitly mention in our documentation that a deployment without resources is not recommended for production. 

The PR adds a helper for defining the resources based on a set of presets (nano, micro, small...), which will be configured in our charts by default to the smallest size that works in our testing. This value will be called `resourcesPreset` , and will be superseded if the user sets the already existing `resources` value. Additionally, we will add warnings explaining users that the should set the `resources` object, regardless of whether they use the `resourcesPreset` value or not.

This is an example of this would look in the chart:

```
          {{- if .Values.metrics.resources }}
          resources: {{- toYaml .Values.metrics.resources | nindent 12 }}
          {{- else if ne .Values.metrics.resourcesPreset "none" }}
          resources: {{- include "common.resources.preset" (dict "type" .Values.metrics.resourcesPreset) | nindent 12 }}
          {{- end }}
```

Additionally, in the NOTES.txt section we would have something like this

```
{{- include "common.warnings.resources" (dict "sections" (list "" "metrics" "volumePermissions") "context" $) }}
```

Before bumping the major version, the value would look like this

```
## @param metrics.resourcesPreset Set the container resources based on a preset (ignored if metrics.resources is set). Possible values: none, nano, micro, small, medium, large, 2xlarge
resourcesPreset: "none"
```

After we perform a major bump, we will set the nalue to one of the presets so our charts set resources by default following the security best practices, but alerting the user that these should not be used in production.

### Benefits

Charts more secure.

### Possible drawbacks

Users will need to get used to resource management

### Checklist

### DIFF
--- a/bitnami/common/Chart.yaml
+++ b/bitnami/common/Chart.yaml
@@ -23,4 +23,4 @@ name: common
 sources:
   - https://github.com/bitnami/charts
 type: library
-version: 2.14.1
+version: 2.15.0
--- a/bitnami/common/templates/_resources.tpl
+++ b/bitnami/common/templates/_resources.tpl
@@ -1,0 +1,50 @@
+{{/*
+Copyright VMware, Inc.
+SPDX-License-Identifier: APACHE-2.0
+*/}}
+
+{{/* vim: set filetype=mustache: */}}
+
+{{/*
+Return a resource request/limit object based on a given preset.
+These presets are for basic testing and not meant to be used in production
+{{ include "common.resources.preset" (dict "type" "nano") -}}
+*/}}
+{{- define "common.resources.preset" -}}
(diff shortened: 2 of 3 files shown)


migruiz4 (approved):
LGTM!

fmulero (approved):
Brilliant! Thanks a lot @javsalgar

klamas1:
Are there any plans to redefine presets?
The presence of limits contradicts one of the principles of combating unreasonable CPU throttling

PR #23460, opened 2024-02-14
[bitnami/kafka] Add useHelmHooks value to allow install with --wait

Wielewout:
### Description of the change

This PR adds the option `provisioning.useHelmHooks` to allow disabling the helm hooks for the provisioning job. By default they will be enabled as before.
Disabling the hooks in combination with setting `provisioning.waitForKafka` to `true` (default value) there still won't be any failure with backoff happening.

### Benefits

* Allows to use `--wait` or even `--atomic` when trying to install the chart with provisioning.
* Allows to use helm's chart tester tool for `ct install` tests (the tool depends on the `--wait` flag).

### Possible drawbacks

* The job (and pod) will not be automatically removed (when disabling helm hooks)

### Additional information

Not having this option blocks us completely from using chart tester running on an umbrella chart with kafka as a dependency. In that umbrella chart there are init containers which wait until provisioned topics are present.

### Checklist

### DIFF
--- a/bitnami/kafka/Chart.yaml
+++ b/bitnami/kafka/Chart.yaml
@@ -42,4 +42,4 @@ maintainers:
 name: kafka
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/kafka
-version: 26.8.5
+version: 26.9.0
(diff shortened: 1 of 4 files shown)


FraPazGal (approved):
Thanks for the contribution @Wielewout, the changes LGTM!

PR #23544, opened 2024-02-15
[bitnami/common] feat: :sparkles: Add ephemeral-storage to resources preset

javsalgar:

### Description of the change

Security checklists also require the `ephemeral-storage` section to be set.  This is used for emptyDir volumes that do not have `medium: Memory`. In this sense, these should not be very high (except for charts where we download datasets like tensorflow-resent or clone git repos like deepspeed).  Then, it could be used by applications that write temporary data or logs that we have not detected.

We've set a very low request value (50Mi) and then increased a bit based on the instance size. However, we should not expect the ephemeral-storage to be highly consumed, otherwise we should check the application logic.

### Applicable issues

- fixes #

### Checklist

### DIFF
--- a/bitnami/common/Chart.yaml
+++ b/bitnami/common/Chart.yaml
@@ -6,7 +6,7 @@ annotations:
   licenses: Apache-2.0
 apiVersion: v2
 # Please make sure that version and appVersion are always the same.
-appVersion: 2.14.1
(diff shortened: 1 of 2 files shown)


fmulero (dismissed):
Great! 👍

PR #23568, opened 2024-02-16
[bitnami/mariadb] Replace deprecated mysqladmin healthchecks with mariadb-admin

biclighter81:
### Description of the change

Mariadb marked the admin cli `mysql-admin` as deprecated and recommend the usage of `mariadb-admin` which is feature compatible. I've just modified the healthchecks so that they use the new cli.

### Benefits

A future release of mariadb won't break our healthchecks.

### Checklist


fevisera:
Hi @biclighter81,

Thank you very much for creating the pull request (PR). I noticed that mariadb-admin is available as a symbolic link since mariadb:10.4.6, as indicated by [this link](https://mariadb.com/kb/en/mariadb-admin).

I was wondering if you could provide some details on the deprecation. It would be helpful if we could access an execution log containing the warning or deprecation messages from MariaDB, just for reference.

Additionally, could you please resolve any conflicts with the chart version?

Thank you!

PR #23577, opened 2024-02-16
[bitnami/odoo] remove rollingUpdate from values.yaml

zzorica:
### Description of the change

When setting values of 

```yaml
updateStrategy:
  type: Recreate
```
...deployment fails because in default values.yaml there is:
```yaml
updateStrategy:
  rollingUpdate: {}
 ```
...which gets into template when rendered:
```yaml
updateStrategy:
  type: Recreate
  rollingUpdate: {}
```
..so it fails with:
```
cannot patch "odoo" with kind Deployment: Deployment.apps "odoo" is invalid: spec.strategy.rollingUpdate: Forbidden: may not be specified when strategy `type` is 'Recreate'
```

### Benefits

This will alow to use  `Recreate` updateStrategy so new pods can mount existing PVC not held by old pod anymore.

### Possible drawbacks

None

### Applicable issues

None that I know off.

### Checklist


bitnami/odoo/values.yaml
```diff
@@ -357,7 +357,6 @@ schedulerName: ""
 ##
 terminationGracePeriodSeconds: ""
 ## @param updateStrategy.type Odoo deployment strategy type
-## @param updateStrategy.rollingUpdate Odoo deployment rolling update configuration parameters
```
juan131, line 360:
```suggestion
## @param updateStrategy.rollingUpdate [object,nullable] Odoo deployment rolling update configuration parameters
```

bitnami/odoo/values.yaml
```diff
@@ -369,7 +368,6 @@ terminationGracePeriodSeconds: ""
 ##
 updateStrategy:
   type: RollingUpdate
-  rollingUpdate: {}
```
juan131, line 372:
```suggestion
  rollingUpdate: null
```

juan131 (changes requested):
Thanks for this contribution @zzorica !!

I think we should use the "nullable" modifier instead but keep the value and document it. Please check my suggestions.

bitnami/odoo/values.yaml
zzorica: Hey @juan131! TIL! Didn't know it will be ignored if `null`. Nice!

juan131 (approved):
LGTM

Issue #21791, opened 2023-12-30
[bitnami/schema-registry] just won't start in SSL


ETisREAL:
### Name and Version

bitnami/schema-registry:16.2.5

### What architecture are you using?

None

### What steps will reproduce the bug?

1. If I specify https listener (ex. listeners: https://0.0.0.0:8081) the container will return this error:
```
ERROR ==> In order to configure HTTPS access, you must mount your ssl.keystore.jks (and optionally the ss │
│ l.truststore.jks) to the /opt/bitnami/schema-registry/certs directory.'. 
```
I circumvented this issue by "manually" mounting the `ssl.keystore.jks` and `ssl.truststore.jks` from a secret as extra volumes
``` yaml
extraVolumes:
  - name: credentials
    secret:
      secretName: schema-registry-certs
extraVolumeMounts:
  - name: credentials
    mountPath: /opt/bitnami/schema-registry/certs/
```

### What do you see instead?

Here we get to the core of the issue. I am using Kafka Strimzi as a Kafka cluster. 
1. I setup a tls user and retrieved the user.crt and user.key and converted them to a jks keystore and truststore for the auth.kafka.jksSecret, a 'dedicated' secret just with these keystore and trustore. This is the command I used to create the secret `kubectl create secret generic schema-registry-kafka-secret --from-file=./kafka.truststore.jks --from-file=./kafka-0.keystore.jks`
2. I created a keystore and trustore for the schema registry client by first retrieving the cluster CA and creating CSR, signed by the CA, and then converted them into jks formats as well. I then created the secret with the following command: `kubectl create secret generic schema-registry-certs-secret --from-file=./schema-registry.truststore.jks --from-file=./schema-registry-0.keystore.jks`.
Here is the chart auth section: 

``` yaml
auth:
  tls:
    enabled: true
    jksSecret: "schema-registry-certs-secret"
    keystorePassword: "changeit"
    truststorePassword: "changeit"
    clientAuthentication: NONE
  
  kafka:
    jksSecret: "schema-registry-kafka-secret"
    tlsEndpointIdentificationAlgorithm: https
    keystorePassword: "changeit"
    truststorePassword: "changeit"
    saslMechanism: PLAIN
```

and this configuration causes this error when I install the chart: `cp: cannot create regular file '/opt/bitnami/schema-registry/certs/schema-registry.truststore.jks': Read-only file system`.

[...]

PR #23583, opened 2024-02-17
[bitnami/schema-registry] Fixed SS init script to copy ssl.keystore and ssl.trustore to correct folder for HTTPS setup

ETisREAL:
As highlighted in the issue #21791 it was necessary to update the Stateful Set init script to copy the ssl.keystore.jks and ssl.trustore.jsk files (if present) over to `/opt/bitnami/schema-registry/certs/` to enable HTTPS setup


bitnami/schema-registry/Chart.yaml
```diff
@@ -34,4 +34,4 @@ maintainers:
 name: schema-registry
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/schema-registry
-version: 16.8.0
+version: 17.0.0
```
Mauraza, line 37:
```suggestion
version: 16.8.1
```

Mauraza (changes requested):
Hi @ETisREAL,

Thanks for the contribution!!!! I think it is not necessary to upgrade to major, it is not a breaking change.

bitnami/schema-registry/Chart.yaml
ETisREAL: Sure thing :)

Mauraza (approved):
LGTM!

ETisREAL:
Thank you @Mauraza :) Have a nice WE

PR #23880, opened 2024-02-23
[bitnami/mongodb] Fix invalid yaml by default

nisanohana3:
### Description of the change

- Only add datadir sub path if it is not empty in order to ensure a valid yaml template
- fixes a bug where node port was rendered with null instead of not being rendered. nodePort is an integer and it can't be null.

### Benefits

Valid YAML

### Possible drawbacks

None

### Checklist


bitnami/mongodb/templates/standalone/svc.yaml
```diff
@@ -50,8 +50,6 @@ spec:
       targetPort: mongodb
       {{- if and (or (eq .Values.service.type "LoadBalancer") (eq .Values.service.type "NodePort")) .Values.service.nodePorts.mongodb }}
       nodePort: {{ .Values.service.nodePorts.mongodb }}
-      {{- else if eq .Values.service.type "ClusterIP" }}
-      nodePort: null
```
Mauraza, line 54: I think it doesn't need to remove these lines. 

> fixes a bug where node port was rendered with null instead of not being rendered. nodePort is an integer and it can't be null.

NodePort is null in case the type is CluterIP.

Mauraza: It's a warning there is no issue. If you check, it is in a lot of the charts.

nisanohana3: Got Ya, what about the other change with the subpath ?

nisanohana3: Hey, any update

yardenshoham:
It's not stale, we are waiting for a review

Mauraza:
Hi @nisanohana3, 

The PR looks good, except for the change of the null. c ould you solve the conflict?

nisanohana3:
sure - the change for the null is reverted

Mauraza (approved):
Hi @nisanohana3,

bitnami/mongodb/Chart.yaml
```diff
@@ -39,4 +39,4 @@ maintainers:
 name: mongodb
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/mongodb
-version: 15.1.0
+version: 15.1.0
```
Mauraza, line 42:
```suggestion
version: 15.1.1
```

PR #23903, opened 2024-02-26
[bitnami/matomo] add independent configuration of cronjob resources

### LINKED ISSUE #23690: Ability to setup custom resources for cronjobs
### Name and Version

bitnami/matomo 

Currently the resources assigned to cronjobs are the same as the one of the main pod, without the possibility to do override it: https://github.com/bitnami/charts/blob/main/bitnami/matomo/templates/cronjob.yaml#L113

There is a problem running these cronjobs when you are using RWO volumes, like mentioned on [this issue](https://github.com/bitnami/charts/issues/20568). A workaround is to setup a `podAffinity` but it is not bullet proof as it will not work if the node on which the main matomo pod is running is already full. Removing the resources assigned to the cronjobs would then ensure they got executed. Not an ideal solution to have cronjobs without assigned resources, but at least it would work

Default behavior would still be to inherit default resources assigned to matomo, but with the possibility to override it

_No response_

reiSh6phoo9o:
### Description of the change

Adds the ability to independently configure resources for cronjobs.

### Benefits

This fixes a problem when using RWO storage, where you need to ensure cronjobs can get scheduled on the same node as the main matomo deployment.

### Possible drawbacks

None

### Applicable issues

- fixes #23690

### Checklist


bitnami/matomo/Chart.yaml
```diff
@@ -39,4 +39,4 @@ maintainers:
 name: matomo
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/matomo
-version: 5.3.2
+version: 5.4.2
```
carrodher, line 42:
```suggestion
version: 5.4.0
```

bitnami/matomo/Chart.yaml
```diff
@@ -39,4 +39,4 @@ maintainers:
 name: matomo
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/matomo
-version: 5.4.1
+version: 5.4.2
```
joancafom, line 42: There probably was a collision with versioning when integrating more recent changes from `main` branch to this one and the version is now outdated. I believe this should be a minor bump, as you are adding a new feature

```suggestion
version: 5.5.0
```

joancafom (commented):
Thanks for your changes! I left you a small comment as a review, please take a look at it whenever you have the chance 😁

joancafom (approved):

viceice:
This is a Breaking Change! Now the normal resources are no longer applied when no resource for the jobs are defined 😞 

This should be reverted or a fallback to the global resources should be added as fix.

PR #23931, opened 2024-02-27
[bitnami/discourse] Update plugins to compatible versions

notz:
### Description of the change

This change updates all installed plugins to the latest version (same as install does) and then sets all plugins to a compatible version.

### Benefits

This change should fix upgrades with outdated plugins and fix install if a plugin already has a newer version.

### Applicable issues

- fixes #

### Checklist


Mauraza:
Hi @notz,

Is this only necessary when doing an upgrade? because if so, I think that if it is only in the upgrade process, it should be added to the documentation.

notz:
> Hi @notz,
> 
> Is this only necessary when doing an upgrade? because if so, I think that if it is only in the upgrade process, it should be added to the documentation.

No, there can also be issue on new installs. Because plugins often support already beta versions of the next discord release and this changes can be incompatible. 

So the the command plugin:pull_compatible_all is for upgrades and new installs.

The plugin:update_all is only for updates/upgrades. But without you can break the installation, because of some outdated plugins with default settings (discourse.persistPlugins=true). And you have to manually pull the plugins and restart the pod to re-create the assets before start.

Mauraza:
Could you add a variable to run this command for no execution by default when exiting a plugin?

notz:
> Could you add a variable to run this command for no execution by default when exiting a plugin?

Sorry, i don't understand your wish. Perhaps you can explain it in more detail.

notz:
> Something simar to:
> 
> ```yaml
>               {{- if .Values.discourse.plugins.update }}
>                 RAILS_ENV=production bundle exec rake plugin:update_all
>                 RAILS_ENV=production LOAD_PLUGINS=0 bundle exec rake plugin:pull_compatible_all
>              {{- end }}
> ```

So what do you think about that option?

| Name | Description | Value |
|------|-------------|-------|
| discourse.updatePlugins | Updates plugins to a compatible version on container initialization | false | 

I'm unsure about the default setting. Because i think everyone that uses plugins want to use compatible versions and that discourse starts without failure on first install and/or upgrade.

Mauraza:
In my opinion, it's okay that the default is true.

notz:
@Mauraza I have updated the pr accordingly.

bitnami/discourse/templates/deployment.yaml
```diff
@@ -119,6 +119,12 @@ spec:
               {{- range $plugin := .Values.discourse.plugins }}
               RAILS_ENV=production bundle exec rake plugin:install repo={{ $plugin }}
               {{- end }}
+              {{- if .Values.discourse.updatePlugins }}
+              {{- if .Values.discourse.persistPlugins }}
+              RAILS_ENV=production bundle exec rake plugin:update_all
+              {{- end }}
+              RAILS_ENV=production LOAD_PLUGINS=0 bundle exec rake plugin:pull_compatible_all
+              {{- end }}
```
Mauraza, line 127: In this case, maybe it makes more sense to separate 'if'.What do you think?
```suggestion
              {{- if .Values.discourse.persistPlugins }}
              RAILS_ENV=production bundle exec rake plugin:update_all
              {{- end }}
              {{- if .Values.discourse.updatePlugins }}
              RAILS_ENV=production LOAD_PLUGINS=0 bundle exec rake plugin:pull_compatible_all
              {{- end }}
```

Mauraza: the and should be the same as use only un variable. 
The options are:

- Only use one variable: 
```yaml
             {{- if .Values.discourse.updatePlugins }}
                RAILS_ENV=production bundle exec rake plugin:update_all
                RAILS_ENV=production LOAD_PLUGINS=0 bundle exec rake plugin:pull_compatible_all
             {{- end }}
```
- The suggestion
```yaml
              {{- if .Values.discourse.persistPlugins }}
                RAILS_ENV=production bundle exec rake plugin:update_all
              {{- end }}
              {{- if .Values.discourse.updatePlugins }}
                RAILS_ENV=production LOAD_PLUGINS=0 bundle exec rake plugin:pull_compatible_all
              {{- end }}
```

notz: The output is the same. But run of  `update_all` is useless, if plugins are not persistet, because the are alwas installed before and are at the same version.

I'm fine with your suggestion, but i would rename the variable to `compatiblePlugins`. Because it's a different use case and the plugins are always updated.

Mauraza: Ok! When you have time please update this and resolve the conflicts.

notz:
@Mauraza I have updated the patch

Mauraza (approved):
Hi @notz , thanks again for the contribution!! LGTM!

PR #23941, opened 2024-02-27
[bitnami/kibana] feat: :sparkles: :lock: Add readOnlyRootFilesystem support

javsalgar:

### Description of the change

This PR performs the necessary changes to enable readOnlyRootFilesystem. This consists of:

- Adding emptyDir directories in all container writable paths
- Update image with one compatible with an emptydir conf folder
- Add runAsGroup

IMPORTANT: In order to avoid potentially breaking changes, this PR sets the readOnlyRootFilesystem value to `false`.

### Benefits

More security in the chart

### Possible drawbacks

Potential unforeseen issues with customizations

### Applicable issues

- fixes #

### Checklist


alemorcuq (approved):

bitnami/kibana/templates/deployment.yaml
```diff
@@ -97,6 +97,34 @@ spec:
...
+            - /bin/bash
+          args:
+            - -ec
+            - |
+                #!/bin/bash
+
+                . /opt/bitnami/scripts/libfs.sh
+                . /opt/bitnami/scripts/kibana-env.sh
+
+                if ! is_dir_empty "$SERVER_DEFAULT_PLUGINS_DIR"; then
```
nick-jones, line 121: Where is `SERVER_DEFAULT_PLUGINS_DIR` expected to be defined? `kibana-env.sh` makes no reference to it. I _can_ see `SERVER_PLUGINS_DIR`, however.

This is causing the script to fail for us, although I can't guarantee we've configured everything correctly:

```
/opt/bitnami/scripts/libfs.sh: line 61: 1: missing directory
```

javsalgar: It is defined here: https://github.com/bitnami/containers/blob/main/bitnami/kibana/8/debian-12/rootfs/opt/bitnami/scripts/kibana-env.sh#L96

Are you changing the default image?

nick-jones: Thanks for clarifying. Indeed, we're overriding the tag:

```
    image:
      tag: 8.8.2
```

Seems that isn't a good idea, so we'll look to stop doing that. Thanks again :+1:

Issue #24011, opened 2024-03-01
[bitnami/keycloak] set Statefulset specific annotations


UXabre:
### Name and Version

bitnami/keycloak:*

### What is the problem this feature will solve?

for some stuff, e.g. stakater reloader, annotations need to be set on a statefulset (or deployment or whatever you want to "reload" on a change).
Currently I can add this via commonAnnotations but this pollutes everything in this case as it is only needed (and wanted) for the statefulset

### What is the feature you are proposing to solve the problem?

allow something like statefulset.annotation or statefulset.extraAnnotations so that secret.reloader.stakater.com/reload can be set

### What alternatives have you considered?

Using commonAnnotations but that pollutes everything (and also this [secret.reloader.stakater.com/reload] would be strange to see on the other resources)

javsalgar:
Hi!


Did you try this?

```
## @param podAnnotations Annotations for Keycloak pods
## ref: https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/
##
podAnnotations: {}
```

Not sure if that would be enough or you would need a specific one for the statefulset object.

UXabre:
Hi @javsalgar , 

It's actually to use it with stakater/reloader, so that keycloak can reload on a change of, for instance, certificates or some secret containing the database password.

E.g.
```
kind: StatefulSet
metadata:
  annotations:
    reloader.stakater.com/auto: "true"
spec:
  template:
    metadata:
    ....
```
Will restart the SS on every change of any mounted secret/configmap (pretty cool stuff btw)

More info in: https://github.com/stakater/Reloader

So it must be on the top level statefulset unfortunately.
BTW, I am willing to implement this, haven't come around to it just yet

javsalgar:
Awesome, then please let us know when you submit the PR and we will take a look. Thank you so much!

UXabre:
@javsalgar PR was created, I've added a field "statefulsetAnnotations", to be somewhat in line with "commonAnnotations" but I'll hear it from you or the community if this is the way

carrodher:
Thank you for opening this issue and submitting the associated Pull Request. Our team will review and provide feedback. Once the PR is merged, the issue will automatically close.

Your contribution is greatly appreciated!

PR #24097, opened 2024-03-05
[bitnami/keycloak] Add support for annotations on statefulset

UXabre:
### Description of the change

This PR adds the ability to add a annotations on the statefulset of keycloak, this is to support special use-case w.r.t. for instance stakater reloader, as the latter adds the ability to reload the rollout of keycloak when, for instance, the database password changes or something else. Again, this is just an example...

### Benefits

The benefit is that we support more integrations, e.g. like with stakater reloader. In the past only commonAnnotations were available, but they "pollute" (in this case) all the resources to include these annotations which might lead to unwanted side-effects

### Possible drawbacks

I don't really think there are any

### Applicable issues

- fixes #https://github.com/bitnami/charts/issues/24011

### Checklist

- [X ] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/). This is *not necessary* when the changes only affect README.md files.
- [X ] Title of the pull request follows this pattern [bitnami/<name_of_the_chart>] Descriptive title


joancafom (approved):
LGTM! Thanks for you contribution 🚀

joancafom (approved):

PR #24339, opened 2024-03-11
[bitnami/matomo] Separate config for cronjobs

Xartos:
### Description of the change

I wanted to be able to have separate config for the cronjobs compared to the matomo deployment.
So I extracted persistent volume and resources config.
I also added `podSecurityContext` and `extraEnvVars` for both cronjobs

### Benefits

One major benefit is that you can now disable persistence for the cronjobs since they were unable to be scheduled if you had persistence on matomo and `ReadWriteOnce` volumes.
And otherwise this adds more ability to setup the cronjobs a bit differently than the matomo deployment to match the needs of the cluster.

### Possible drawbacks

There's a bit more config now which might be perceived as complex. But I think that the benefits overcomes the drawbacks.

### Applicable issues

- fixes #

### Additional information

This will cause some disruption of the config where some config were moved from one place to another so I opted to bump a minor version for this.

### Checklist

### DIFF
--- a/bitnami/matomo/Chart.yaml
+++ b/bitnami/matomo/Chart.yaml
@@ -39,4 +39,4 @@ maintainers:
 name: matomo
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/matomo
-version: 5.5.1
+version: 6.0.0
--- a/bitnami/matomo/README.md
+++ b/bitnami/matomo/README.md
@@ -398,6 +398,8 @@ helm install my-release --set persistence.existingClaim=PVC_NAME oci://REGISTRY_
 | `cronjobs.taskScheduler.podAnnotations`                                    | Additional pod annotations                                                                                        | `{}`             |
 | `cronjobs.taskScheduler.podLabels`                                         | Additional pod labels                                                                                             | `{}`             |
 | `cronjobs.taskScheduler.resources`                                         | Set container requests and limits for different resources like CPU or memory (essential for production workloads) | `{}`             |
+| `cronjobs.taskScheduler.persistence.enable`                                | Enable persistence using Persistent Volume Claims                                                                 | `true`           |
+| `cronjobs.taskScheduler.persistence.existingClaim`                         | A manually managed Persistent Volume Claim                                                                        | `""`             |
 | `cronjobs.archive.enabled`                                                 | Whether to enable scheduled mail-to-task CronJob                                                                  | `true`           |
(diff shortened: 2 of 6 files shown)


Mauraza:
Hi @Xartos,

Thanks for the contribution!! Your changes made the tests fail, so I just updated it, and let's see how is going.

Mauraza:
The helm list is failing, could you check it?
```bash
helm lint .
==> Linting .
[ERROR] templates/cronjob.yaml: unable to parse YAML: error converting YAML to JSON: yaml: line 80: did not find expected '-' indicator
[ERROR] templates/cronjob.yaml: unable to parse YAML: error converting YAML to JSON: yaml: line 80: did not find expected '-' indicator

Error: 1 chart(s) linted, 1 chart(s) failed
```

Xartos:
@Mauraza I totally missed to test with the default config :sweat_smile: But I should've solved that issue now :slightly_smiling_face:

Mauraza:
Hi @Xartos,

Could you add a small note in the README.md about this change? I think the users should be notified.

Xartos:
@Mauraza Do you not think this is enough

https://github.com/bitnami/charts/pull/24339/files#diff-190bc6e7ebe50a8f1a98ec55511936b5066371f377f79ceacf2f83b3c6a37f7fR475-R478

If not would you like me do expand even more in that section or would you like me do add something in some other section?

bitnami/matomo/Chart.yaml
```diff
@@ -39,4 +39,4 @@ maintainers:
 name: matomo
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/matomo
-version: 5.5.1
+version: 5.6.0
```
Mauraza, line 42:
```suggestion
version: 6.0.0
```

bitnami/matomo/README.md
```diff
@@ -456,6 +472,11 @@ Find more information about how to deal with common errors related to Bitnami's
 
 ## Upgrading
 
+### To 5.6.0
```
Mauraza, line 475:
```suggestion
### To 6.0.0
```

Mauraza (changes requested):
That looks good but I think if someone with the cronjobs disabled the upgrade it is going to fail, so it should be major.

Mauraza (approved):
Hi @Xartos,
Thanks again for the contribution! LGTM!

viceice (commented):
broken because of a typo `enabled` vs `enable`

#24682

PR #24346, opened 2024-03-11
[bitnami/vault] fix: don't create unused CSI Provider or injector resources

rloporp:
### Description of the change

Some auxiliary resources like service accounts and network policies are still created when using the default values (where CSI Provider and injector are not deployed). With this change, those resources will only be created when actually needed.

### Benefits

* Unused resources will not be deployed.

### Possible drawbacks

* There may be a corner case (not identified) when the CSI Provider and injector are deployed without those auxiliary resources.

### Applicable issues

None

### Checklist


juan131 (approved):
Thanks for the fix! LGTM

PR #24530, opened 2024-03-18
[bitnami/postgresql] existingExtendedConfigmap not honored since chart v14.2.1

### LINKED ISSUE #24528: [bitnami/postgresql] existingExtendedConfigmap and extendedConfiguration broken since chart v14.2.1
### Name and Version

bitnami/postgresql] v14.2.1+

When using a value file with the`"existingExtendedConfigmap"` tag pointing to a configmap, the postgresql configuration params are not applied when deploying the chart
It works up to v14.1.3, broken since c14.2.1

The problem is the same when`extendedConfiguration`, ie the content of the file is discard/not used

values.yaml:
```yaml
    primary:
      existingExtendedConfigmap: budget2-db-dev-postgresql-config 
```
budget2-db-dev-postgresql-config  ConfigMap:
```yaml
    apiVersion: v1
    kind: ConfigMap
    metadata:
      name: budget2-db-dev-postgresql-config
    data:
      override.conf: |+
        shared_buffers=256MB
        work_mem=128MB
        maintenance_work_mem=128MB
        temp_buffers=64MB
        ...
 ```

The chart deploys without problem with the postgresql parameters overridden from the config map

### What do you see instead?

The chart deploys without overriding the postgresql config parameters

### Additional information

This is caused by this code addition in `"template/primary/statefulset.yaml"` in v14.2.1:
```yaml
     volumeMounts:
        - name: empty-dir
          mountPath: /tmp
          subPath: tmp-dir
        - name: empty-dir
          mountPath: /opt/bitnami/postgresql/conf
          subPath: app-conf-dir
        - name: empty-dir
          mountPath: /opt/bitnami/postgresql/tmp
          subPath: app-tmp-dir
        - name: empty-dir
          mountPath: /opt/bitnami/postgresql/logs
          subPath: app-logs-dir
```
Removing the added lines related to `"/opt/bitnami/postgresql/conf"` from `template/primary/statefulset.yaml` in any version >v14.2.1+ makes the chart work again
```yaml
        - name: empty-dir
          mountPath: /opt/bitnami/postgresql/conf
          subPath: app-conf-dir
```
Q:
Why has this been added to the chart, and in what use case it is useful?

It may be related to issue #16623

titou10titou10:
removed mounty empty-dir on /opt/bitnami/postgresql/conf

### Description of the change

fix #24528 , ie remove 
```yaml
    - name: empty-dir
          mountPath: /opt/bitnami/postgresql/conf
          subPath: app-conf-dir
```
From `template/primary/statefulset.yaml` introduced in chart v14.2.1 and broke the usage of `"existingExtendedConfigmap"` in the chart


andresbono:
Hi, as you can see in the results of the CI/CD, the empty dir volume is actually required:

```
[pod_id=postgresql-ptest-0] postgresql 12:22:24.06 INFO  ==> Generating local authentication configuration
[pod_id=postgresql-ptest-0] /opt/bitnami/scripts/libpostgresql.sh: line 234: /opt/bitnami/postgresql/conf/pg_hba.conf: Read-only file system
[pod_id=postgresql-ptest-0] /opt/bitnami/scripts/libfile.sh: line 41: /opt/bitnami/postgresql/conf/postgresql.conf: Read-only file system
```

Because the pod is using a read-only root file system by default to enhance the security features.

If that enhancement conflicts with `existingExtendedConfigmap`, we should address the fix in some other way... Does it make sense to you?

titou10titou10:
I understand
> If that enhancement conflicts with existingExtendedConfigmap, we should address the fix in some other way... 

Yes it does, so now the usage of`existingExtendedConfigmap`is broken
I'm not sure to uderstand why the directory where people can add extra config to postgresql must be read-only (for security reasons?)...So, ho can we use`existingExtendedConfigmap`or even`extendedConfiguration` that create a file in this directory too. The 2 are broken now...
I have no idea on how to fix it except by doing what I propose ...

PR #24544, opened 2024-03-19
[bitnami/keycloak] Avoid initContainer execution with 3rd party Keycloak images

### LINKED ISSUE #24192: [bitnami/keycloak] `app-quarkus-dir` is empty
### Name and Version

bitnami/keycloak 14.2.4

amd64

There seems to be a seemingly random chance that `/opt/bitnami/keycloak/lib/quarkus/` is empty, due to which https://github.com/bitnami/charts/blob/b9201f64c174a9e0cee184e02e48e31624695568/bitnami/keycloak/templates/statefulset.yaml#L90-L99 fails.

I have no idea how this is possible, it worked the last few days, and now it doesn't work anymore.
Even a downgrade didn't help.

Output of `helm get --namespace $NAMESPACE values keycloak`
```yaml
USER-SUPPLIED VALUES:
args:
- start-dev
- --import-realm
auth:
  adminPassword: <redacted>
  adminUser: <redacted>
command:
- /opt/keycloak/bin/kc.sh
extraEnvVars:
- name: KC_HOSTNAME_STRICT
  value: "false"
extraVolumeMounts:
- mountPath: /opt/keycloak/data/import
  name: keycloak-realm
- mountPath: /opt/keycloak/providers
  name: keycloak-plugins
extraVolumes:
- configMap:
    name: keycloak-realm-config
  name: keycloak-realm
- name: keycloak-plugins
  persistentVolumeClaim:
    claimName: keycloak-plugin-volume
image:
  registry: quay.io
  repository: keycloak/keycloak
  tag: 24.0
ingress:
  hostname: <redacted>
initContainers:
- image: <redacted>
  imagePullPolicy: Always
  name: keycloak-plugin-initializer
  volumeMounts:
  - mountPath: /opt/keycloak/providers
    name: keycloak-plugins
metrics:
  enabled: true
podAnnotations:
  dapr.io/app-id: keycloak
  dapr.io/app-port: "80"
  dapr.io/config: tracing
  dapr.io/enabled: "true"
postgresql:
  auth:
    database: <redacted>
    postgresPassword: <redacted>
    username: <redacted>
production: true
proxy: edge
readinessProbe:
  initialDelaySeconds: 60
  timeoutSeconds: 10
```

The chart starts.

### What do you see instead?

```sh
kubectl logs --namespace $NAMESPACE keycloak-0 --container init-quarkus-directory
cp: cannot stat '/opt/bitnami/keycloak/lib/quarkus/*': No such file or directory
```

### Additional information

I think the easiest solution would be to continue even if the directory is empty.
However, I don't know how and why this directory is empty, and/or what it means

menzbua:
### Description of the change

The initContainer would be executed, although when you are using a 3rd party Keycloak image. The initContainer fails because in the 3rd party image, the folder /quarkus not exists. This fix evaluates whether you are using the bitnami/keycloak image or another. When you are using the Bitnami image the initContainer would be executed when not the initContainer would not be executed.

### Benefits

With this change, it is possible to use 3rd party images with the helm chart.

### Possible drawbacks

The bitnami/keycloak image name is hard-coded. If there is any change, the initContainer not starts on the Bitnami image anymore.

### Applicable issues

- fixes #24192

### Checklist


jotamartos (changes requested):
Thanks for the suggestion but this will break some automations we have. Could you please follow a similar approach to the one we have in Milvus? 

https://github.com/bitnami/charts/blob/main/bitnami/milvus/values.yaml#L288

The enableDefaultInitContainers parameter allows you run the init container or not depending on your needs.

menzbua:
Okay, I will do that in the coming days.

menzbua:
I've added the parameter as described above. Please check again. Thanks

bitnami/keycloak/Chart.yaml
```diff
@@ -33,4 +33,4 @@ maintainers:
 name: keycloak
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/keycloak
-version: 19.3.3
+version: 19.3.4
```
jotamartos, line 36: As you are including a new parameter, please bump the chart to a new minor version

```suggestion
version: 19.4.0
```

jotamartos (commented):
The changes look great. Just a last minor request

jotamartos (approved):

jotamartos (approved):
Changes are now ready to be merged! Thanks!

lknite:
I'm working to use quay.io/phasetwo/keycloak-crdb .

If I specify that as my image in the bitnami keycloak helm chart:
```
cp: cannot stat '/opt/bitnami/keycloak/lib/quarkus/*': No such file or directory
```

If I then add:
```
enableDefaultInitContainers: false
```

Any idea why I then end up with the following in the logs?  (appreciate any ideas, a bit out of my element here)
```
Appending additional Java properties to JAVA_OPTS: -Djgroups.dns.query=keycloak-headless.keycloak.svc.cluster.local
Keycloak - Open Source Identity and Access Management
Find more information at: https://www.keycloak.org/docs/latest
Usage:
kc.sh [OPTIONS] [COMMAND]
Use this command-line tool to manage your Keycloak cluster.
Options:
-cf, --config-file <file>
                     Set the path to a configuration file. By default, configuration properties are
                       read from the "keycloak.conf" file in the "conf" directory.
-h, --help           This help message.
-v, --verbose        Print out error details when running this command.
-V, --version        Show version information
Commands:
  build                   Creates a new and optimized server image.
  start                   Start the server.
  start-dev               Start the server in development mode.
  export                  Export data from realms to a file or directory.
  import                  Import data from a directory or a file.
  show-config             Print out the current configuration.
  tools                   Utilities for use and interaction with the server.
    completion            Generate bash/zsh completion script for kc.sh.
Examples:
  Start the server in development mode for local development or testing:
      $ kc.sh start-dev
  Building an optimized server runtime:
      $ kc.sh build <OPTIONS>
  Start the server in production mode:
      $ kc.sh start <OPTIONS>
  Enable auto-completion to bash/zsh:
      $ source <(kc.sh tools completion)
  Please, take a look at the documentation for more details before deploying in
production.
Use "kc.sh start --help" for the available options when starting the server.
Use "kc.sh <command> --help" for more information about other commands.
```

menzbua:
You have to specify startup args in the values file for custom images. Here is an example, it could differ depending on what you want:
```yaml
args:
  - start-dev
```

lknite:
It looks like the helm chart parameter extraStartupArgs exists, and sets the env var KEYCLOAK_EXTRA_ARGS, but that variable is not passed to the container.  Is that the parameter that you would expect to work for this purpose?

menzbua:
The Bitnami config values not working for third party Keycloak images. For my perspective, the values `production: true` and `production: true` are only working with the Bitnami Keycloak image. You have to map these variables to env Variables that your container understands. Your values file could look like this:
```yaml
  image:
    debug: true
    registry: quay.io
    repository: phasetwo/keycloak-crdb
    tag: 24.0.3

  enableDefaultInitContainers: false
  args:
    - start
  extraEnvVars:
    - name: KC_PROXY
      value: edge
```
Keep in mind that KC_PROXY is deprecated:
https://www.keycloak.org/server/reverseproxy

lknite:
Looks like my 3rd party image is using /opt/keycloak for files rather than the /opt/bitnami/keycloak used by the default image.

I could create a Dockerfile which adds a symbolic link from /opt/bitnami/keycloak to /opt/keycloak but that feels like the wrong solution.

In the StatefulSet template there is an initContainer but it uses a hardcoded path:
```
cp -r /opt/bitnami/keycloak/lib/quarkus/* /quarkus
```
If that instead used a templated value, then I could adjust the templated value over to /opt/keycloak/lib/quarkus .

I could also use an initContainer for this, but if I use enableDefaultInitContainers: false, it disables all initContainers:
```
.Values.initContainers
```

What sounds like the best solution here?

What about this:  'enableDefaultInitContainers: false' could disable the one default initContainer, but still allow initContainers if they have been defined by .Values.initContainers ?

menzbua:
You can configure your own initContainers:
https://github.com/bitnami/charts/blob/3f528aaf16f9d54c6564e225dda54d87bf46dddd/bitnami/keycloak/values.yaml#L503

lknite:
But if '.enableDefaultInitContainers' is set to false they will be ignored.

> You can configure your own initContainers:
> 
> https://github.com/bitnami/charts/blob/3f528aaf16f9d54c6564e225dda54d87bf46dddd/bitnami/keycloak/values.yaml#L503

lknite:
First attempt at a pull request with the else added: https://github.com/bitnami/charts/pull/25285

PR #24549, opened 2024-03-19
[bitnami/redis] Add support for dedicated values for sentinel master service

shlomitubul:
### Description of the change
The current implementation of the chart uses the sentinel.service values to configure the Sentinel master service. This approach limits the flexibility to specify dedicated configurations specifically for the master service, as any changes affect both the Sentinel service and the master service simultaneously. To address this limitation, this PR introduces a new set of values under sentinel.masterService. These values are specifically dedicated to configuring the Sentinel master service, allowing for more granular control over its behavior.

This PR is designed to be fully backward-compatible. Existing deployments that rely on sentinel.service for their configuration will continue to function as before without any required changes. The new sentinel.masterService values are optional and are only used when explicitly defined by the user. 

### Benefits
Users can now apply specific configurations to the Sentinel master service independently from the Sentinel service

### Possible drawbacks

None

### Checklist


shlomitubul:
@aoterolorenzo When I think about that, we might fully separate sentinel service from sentinel-master service vaules,
the use case is for example if we want to have sentinel-master service as LB with some annotation/label or any other config that is specific to this service, currently, we try to use sentinel-master service feature but there since both services share the same value if sentinel-master is LB the regular sentinel service is also LB, same for any other config.
I'm willing to open dedicated PR for this, WDYT?

bitnami/redis/templates/podmonitor.yaml
```diff
@@ -45,7 +45,7 @@ spec:
       {{- if .honorLabels }}
       honorLabels: {{ .honorLabels }}
       {{- end }}
-      {{- with concat .Values.metrics.podMonitor.relabelings .Values.metrics.podMonitor.relabellings }}
```
shlomitubul, line 48: This is not related to the PR context changes, just a bug fix since the helm context is not correct here

bitnami/redis/README.md
```diff
@@ -815,14 +815,27 @@ helm install my-release --set master.persistence.existingClaim=PVC_NAME oci://RE
...
 | `sentinel.service.clusterIP`                                 | Redis&reg; Sentinel service Cluster IP                                                                                                                                                                                       | `""`                             |
-| `sentinel.service.createMaster`                              | Enable master service pointing to the current master (experimental)                                                                                                                                                          | `false`                          |
+| `sentinel.service.createMaster`                              | DEPRECATED. Use sentinel.masterService.createMaster instead                                                                                                                                                                  | `false`                          |
```
aoterolorenzo, line 818: We should not mark a feature as deprecated while we mark the alternative as experimental. IMO one of them has to be 'stable', so I would mark this still as stable and usable. Also, I think that people may still want to apply configurations just to both services, so we should stay stable here.

aoterolorenzo (changes requested):

shlomitubul:
@aoterolorenzo is there anything else that needs to be fixed/changed?

sviatlo:
Hello @aoterolorenzo @carrodher! Is there any way to push the review of this PR? I'd love to start using this feature. Thank you.

bitnami/redis/templates/podmonitor.yaml
shlomitubul: moved to separate PR https://github.com/bitnami/charts/pull/25237

andresbono: Thank you for splitting this other fix in an independent PR. The team will review #25237 soon

bitnami/redis/Chart.yaml
```diff
@@ -36,4 +36,4 @@ maintainers:
 name: redis
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/redis
-version: 19.1.2
+version: 19.1.3
```
andresbono, line 39: I think we should bump the minor version here, as we are adding new features with new parameters.

andresbono (commented):
Thank you very much for your contribution. This PR is really similar to these other changes for the influxdb chart: https://github.com/bitnami/charts/pull/24586. I wonder if we shouldn't be applying a similar approach in both PRs:

* Have a parameter (let's say `sentinel.masterService.ownConfig=true`) that is the flag that activates the `sentinel.masterService` block only if set explicitly for backwards compatibility. That way, you can have full control over the settings in the master service. If not, it is not possible to set master service values to empty/null if they aren't empty/null in the other service (because of the `coalesce`).
* In `sentinel/service.yaml`, load the service block with something like `{{- $masterService := ternary .Values.sentinel.masterService .Values.sentinel.service .Values.sentinel.masterService.ownConfig }}`. That would avoid a bunch of `coalesce` functions.

Does it make sense to you?

shlomitubul:
@andresbono I see that the license header check has started to fail recently without change from my side, is something changing lately?

DrFaust92:
[andresbono](https://github.com/andresbono) any chance we can push this?

andresbono:
After giving a second round of review, I think it's better to drop `sentinel.masterService.createMaster` and `sentinel.masterService.ownConfig` in favor of `sentinel.masterService.enabled`. By setting `sentinel.masterService.enabled=true`, the block of values in `sentinel.masterService` will be applied, and can be different than the ones set in `sentinel.service`.

I have pushed a commit addressing those changes to not miss the momentum and merge the PR.

> @andresbono I see that the license header check has started to fail recently without change from my side, is something changing lately?

Yes, thank you. Fixed by merging from the `main` branch.

andresbono (approved):
LGTM. Thank you so much for your contribution!

PR #24586, opened 2024-03-20
[bitnami/influxdb] Allow separate persistence cfg for backups

kevin-delisle-peakpower:
I'm back! I missed the approval of the PR last time (was bogged down in other work and it slipped my mind), so I've rebased it, and tidied it up a bit.

- Add an optional configuration section under backup that allows the specification of distinct values for persistence. If this configuration section is not provided, the template will default to rendering using the existing top-level persistence configuration section to maintain compatibility.

### Benefits
Users can now define volumes for the Deployment and their backups separately without needing Kustomize or manual post-deployment hacks.

Example usage could include:
- using different StorageClasses for their backup volumes vs. their active database volumes.
- using a larger volume for backups vs. the main database volume to accommodate multiple full backups

### Possible drawbacks

There should not be any drawbacks to this PR; I have attempted to maintain backwards compatibility with existing use cases in the event that someone provides a values.yaml file that does not include the new section.

### Additional information

I've dumped the template out using permutations of both persistence modes to confirm that the `backup.persistence` section will take precedence over `persistence` for the backup volume, as well as to confirm that the templates still correctly render in the complete absence of that configuration section.

### Checklist


bitnami/influxdb/README.md
```diff
@@ -304,7 +306,7 @@ There are K8s distribution, such as OpenShift, where you can dynamically define
 | `ingress.pathType`         | Ingress path type                                                                                                                | `ImplementationSpecific` |
 | `ingress.apiVersion`       | Force Ingress API version (automatically detected if not set)                                                                    | `""`                     |
 | `ingress.hostname`         | Default host for the ingress resource (evaluated as template)                                                                    | `influxdb.local`         |
-| `ingress.path`             | Ingress path*' in order to use this                                                                                              | `/`                      |
+| `ingress.path`             | Ingress path\*' in order to use this                                                                                             | `/`                      |
```
kevin-delisle-peakpower, line 309: This change is being inserted when I run the `readme-generator` utility; not sure if that's a problem.

kevin-delisle-peakpower:
@carrodher I've sent the invite with Write access to bitnami-bot.

kevin-delisle-peakpower:
@andresbono Any word on whether or not we can merge this? I don't want it to get closed due to inactivity like last time 😄

bitnami/influxdb/templates/pvc-backup.yaml
```diff
@@ -2,26 +2,17 @@
...
-  {{- if or .Values.persistence.annotations .Values.commonAnnotations }}
-  {{- $annotations := include "common.tplvalues.merge" ( dict "values" ( list .Values.persistence.annotations .Values.commonAnnotations ) "context" . ) }}
-  annotations: {{- include "common.tplvalues.render" ( dict "value" $annotations "context" $) | nindent 4 }}
+{{-
+/*
+Prefer .Values.backup.persistence, but fall back to .Values.persistence if not present.
+*/
+-}}
+{{- if .Values.backup.enabled }}
+  {{- if and .Values.backup.persistence .Values.backup.persistence.enabled }}
```
andresbono, line 11: I see you are trying to keep backwards compatibility in case some user does not define the new `backup.persistence` block... But note that you are defining that block in the `values.yaml`. Those values are the default ones. So I'm not sure if we actually need to cover that case... Does it make sense?

bitnami/influxdb/templates/_helpers.tpl
```diff
@@ -109,3 +109,32 @@ Get the InfluxDB&trade; initialization scripts secret.
...
 {{- end -}}
+
+{{- /*
+Helper template for determining which persistence settings to use for backup PVC.
+Allows backwards-compatibility for users that have not defined a backups.persistence dict in their values files.
+*/}}
+{{- define "influxdb.backup.pvc" }}
+{{- $ := index . 0 }}
+{{- $persistence := index . 1 }}
+kind: PersistentVolumeClaim
```
andresbono, line 120: If possible, I would try to avoid including the resource definition in the helper. But please, let's clarify first my other comment.

andresbono (commented):
Thank you for your contribution! Could you please check my comments?

bitnami/influxdb/templates/pvc-backup.yaml
kevin-delisle-peakpower: Whoops!

Would it make more sense to set the default value of `backup.persistence.enabled` to `false` so that users don't end up with an additional volume provisioned that they weren't expecting?

kevin-delisle-peakpower: The main reason I covered the case was that I assumed that people might be running these backups on the existing volume (though, I suppose that depends on a lot of factors, like whether or not their environment supports ReadWriteMany mode for the bound volume, and is binding that volume for both the database Pod and the backup Job)

bitnami/influxdb/templates/_helpers.tpl
kevin-delisle-peakpower: I'm not quite sure where I should put the template for this. I moved it to a template function because I couldn't figure out a way to set a common variable for rendering the template in the `.yaml` file itself, and `_helpers.tpl` was the only template at the time. I realize that this function doesn't fit the established pattern in there; should I move it to its own `.tpl` file?

andresbono: I would say that the proper place is in `pvc-backup.yaml` and the values for the fields would be typically set with [`coalesce`](https://helm.sh/docs/chart_template_guide/function_list/#coalesce) for example to keep compatibility. But in this case, the scenario might be more complex though...

kevin-delisle-peakpower:
I noticed there's no tests, but I ran through a few scenarios to evaluate the output of the chart:
```sh
# Should create no PVCs, use the same volume claim for both (still not sure if this makes sense)
helm template test . -f values.yaml --set backup.enabled=true --set backup.persistence.enabled=false --set persistence.existingClaim=foo

# Should create one PVC for the backup
helm template test . -f values.yaml --set backup.enabled=true --set backup.persistence.enabled=true --set persistence.existingClaim=foo

# Should create one PVC for the database
helm template test . -f values.yaml --set backup.enabled=true --set backup.persistence.enabled=true --set backup.persistence.existingClaim=foo --set persistence.size=16Gi

# Should create one PVC for the backup with size=16Gi
helm template test . -f values.yaml --set backup.enabled=true --set backup.persistence.enabled=true --set persistence.enabled=true --set persistence.existingClaim=foo --set backup.persistence.size=16Gi

# Should create no PVCs, reference existing claims for both
helm template test . -f values.yaml --set backup.enabled=true --set backup.persistence.enabled=true --set persistence.existingClaim=foo  --set backup.persistence.existingClaim=bar

# Should only use foo and ignore bar (since backup.persistence.enabled=false)
helm template test . -f values.yaml --set backup.enabled=true --set backup.persistence.enabled=false --set persistence.existingClaim=foo  --set backup.persistence.existingClaim=bar
```

I am still worried that the default behaviour for many (most?) consumers using this chart would be to have the database's PVC set to ReadWriteOnce mode, which would prevent the CronJob from ever launching.

Ultimately, if the Bitnami team thinks its okay to swap the CronJob to using a separate persistent volume as a default for backups, I can change the default value of `backup.persistence.enabled` back to true. You could argue that it's a breaking change, though, since anyone who's been consuming this chart might find that any additional scripts they've written will now need to run on a different target.

bitnami/influxdb/Chart.yaml
```diff
@@ -18,23 +18,23 @@ annotations:
...
-- name: common
-  repository: oci://registry-1.docker.io/bitnamicharts
-  tags:
-  - bitnami-common
-  version: 2.x.x
+  - name: common
+    repository: oci://registry-1.docker.io/bitnamicharts
+    tags:
+      - bitnami-common
+    version: 2.x.x
```
andresbono, line 25:
```suggestion
- name: common
  repository: oci://registry-1.docker.io/bitnamicharts
  tags:
  - bitnami-common
  version: 2.x.x
```

bitnami/influxdb/Chart.yaml
```diff
@@ -18,23 +18,23 @@ annotations:
...
 icon: https://bitnami.com/assets/stacks/influxdb/img/influxdb-stack-220x234.png
 keywords:
-- influxdb
-- tick
-- database
-- timeseries
+  - influxdb
+  - tick
+  - database
+  - timeseries
```
andresbono, line 33:
```suggestion
- influxdb
- tick
- database
- timeseries
```

bitnami/influxdb/Chart.yaml
```diff
@@ -18,23 +18,23 @@ annotations:
...
 maintainers:
-- name: VMware, Inc.
-  url: https://github.com/bitnami/charts
+  - name: VMware, Inc.
+    url: https://github.com/bitnami/charts
```
andresbono, line 36:
```suggestion
- name: VMware, Inc.
  url: https://github.com/bitnami/charts
```

bitnami/influxdb/Chart.yaml
```diff
@@ -18,23 +18,23 @@ annotations:
...
 name: influxdb
 sources:
-- https://github.com/bitnami/charts/tree/main/bitnami/influxdb
-version: 6.0.6
+  - https://github.com/bitnami/charts/tree/main/bitnami/influxdb
```
andresbono, line 39:
```suggestion
- https://github.com/bitnami/charts/tree/main/bitnami/influxdb
```

The YAML is still valid with both levels of indentation but it will be reverted in the next automatic update causing unnecessary commit changes.

bitnami/influxdb/values.yaml
```diff
@@ -926,6 +926,36 @@ backup:
   ## @param backup.retentionDays Retention time in days for backups (older backups are deleted)
   ##
   retentionDays: 10
+
+  ## Persistence parameters
+  ##
+  persistence:
```
andresbono, line 932:
```suggestion
  persistence:
    ## @param backup.persistence.ownConfig Prefer independent own persistence parameters to configure the backup volume
    ## When set to `false` (for backwards compatibility), the rest of the persistence parameters below will be ignored.
    ## This parameter will be set to `true` and removed in a future release.
    ##
    ownConfig: false
```

andresbono (commented):
This looks great now! Thank you so much. Could you check my suggestions and run your `helm template` scenarios again?

The main point of my suggestions is to introduce a new different parameter `backup.persistence.ownConfig` to have control on what set of persistence parameters should be applied. I consider it is important to have a different one than `backup.persistence.enabled` to cover the case in which you want persistence for the database but not for the backup for example.

As we are dealing with application persisted data here, we should be extra safe :)

kevin-delisle-peakpower:
> This looks great now! Thank you so much. Could you check my suggestions and run your `helm template` scenarios again?
> 
> The main point of my suggestions is to introduce a new different parameter `backup.persistence.ownConfig` to have control on what set of persistence parameters should be applied. I consider it is important to have a different one than `backup.persistence.enabled` to cover the case in which you want persistence for the database but not for the backup for example.
> 
> As we are dealing with application persisted data here, we should be extra safe :)

I'm a bit confused; if someone wants persistence for the database and not the backup, they would set `persistence.enabled: true` (which is the default) and `backup.persistence.enabled: false`. If someone decides to use `backup.persistence.enabled: true`, the template use the values/default values under the `backup.persistence` section, not the `persistence` section (or at least, that's what I was aiming for).

If I'm missing the point, could you give me some contrived values.yaml examples to help me see the issue?

bitnami/influxdb/README.md
```diff
@@ -129,6 +129,8 @@ As an alternative, you can use of the preset configurations for pod affinity, po
 The data is persisted by default using PVC(s). You can disable the persistence setting the `persistence.enabled` parameter to `false`.
 A default `StorageClass` is needed in the Kubernetes cluster to dynamically provision the volumes. Specify another StorageClass in the `persistence.storageClass` or set `persistence.existingClaim` if you have already existing persistent volumes to use.
 
+If you would like to define persistence settings for a backup volume that differ from the persistence settings for the database volume, you may do so under the `backup.persistence` section of the configuration. If this section is undefined, but `backup.enabled` is set to true, the backup volume will be defined using the `persistence` parameter section.
```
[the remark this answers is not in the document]
kevin-delisle-peakpower: Just to clarify: if someone doesn't set `backup.persistence.ownConfig: true`, and they have `persistence.existingClaim` set to some volume, should the effective value of `backup.persistence.existingClaim` be the same? That would lead to binding of the same target volume, which is the RWO problem I mentioned.

bitnami/influxdb/Chart.yaml
kevin-delisle-peakpower: Yeah, my YAML formatting plugin has been aggressively asserting itself. I'll change them back

bitnami/influxdb/README.md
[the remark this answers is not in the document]
andresbono: If someone has `backup.persistence.ownConfig: false` (which is the proposed default, at least on this first version) the chart will have the very same behavior than the current version and any `backup.persistence.*` parameter will be ignored.

Any bugs present in the current version will still be there unless you explicitly set `backup.persistence.ownConfig: true`. Then, you have all the flexibility you want, right?

andresbono:
> but it would also create issues for anyone using persistence.existingClaim with ReadWriteOnce volumes by default as well.

I feel like I still don't fully understand the concern here. What if you give me an example of the `values.yaml` that could cause trouble?

> Regardless, I think this makes a strong argument for a major version bump of the chart, and a step in an upgrade guide to avoid whichever pitfall is created by the approach we settle on.

I would do the change incrementally. We can go ahead with the PR and `backup.persistence.ownConfig: false` by default to be safe and then in the next major version (maybe grouped with other changes), we get rid of the `backup.persistence.ownConfig` flag and we change the default behavior. ➕ to the upgrade guide. We use to do that as part of the major upgrading notes in the README.

Does it make sense to you?

kevin-delisle-peakpower:
@andresbono Sorry, I've been really busy all of this week with work, but I haven't forgotten about this PR.

My concerns about `ReadWriteOnce` are probably misplaced. Technically, as long as the node the volume is bound to has compute and memory available for the CronJob (and has the right tolerations), this should still work.

I was mostly worried because I've previously seen behaviour in which certain drivers treat ReadWriteOnce in the way the newer `ReadWriteOncePod` feature is intended to work: https://github.com/bottlerocket-os/bottlerocket/issues/2417#issuecomment-1247395609

When I was testing this out myself (EKS v1.23 with aws-ebs-csi for storage), the cluster was throwing VolumeAttachment errors. I suspect it wasn't scheduling the CronJob on the same node as the Deployment, but that's an issue with something outside of the scope of this chart.

andresbono:
> @andresbono Sorry, I've been really busy all of this week with work, but I haven't forgotten about this PR.

No problem at all! We totally understand.

andresbono:
Hi @kevin-delisle-peakpower, if you prefer, you can grant me access to your fork and I can help pushing the proposed suggestions. Let me know if you prefer that. Thank you!

kevin-delisle-peakpower:
Honestly, I would love that. I know this has been taking too long and I've been swamped with a bunch of OIDC and AWS IAM refactoring work.

andresbono:
Hi @kevin-delisle-peakpower, I missed the invitation to your fork. It was forwarded to my personal email, and the invitation now expired. But no worries, I have created a follow-up PR: https://github.com/bitnami/charts/pull/26453. I'm preserving your original commits as well 😄.

Closing in favor of https://github.com/bitnami/charts/pull/26453.

Thank you so much for your contribution!!

PR #24603, opened 2024-03-21
[bitnami/thanos] Fix thanos objstoreConfig and storegatewayConfigMap

### LINKED ISSUE #24536: thanos helm chart does not render with existingObjstoreSecret
### Name and Version

bitnami/thanos 14.0.1

None

1. Use the newest thanos helm chart version 14.0.1
2. Configure existingObjstoreSecret

```
existingObjstoreSecret: "custom-s3-config"
```

### What do you see instead?

```
Error: template: thanos/charts/thanos/templates/storegateway/statefulset-sharded.yaml:53:44: executing "thanos/charts/thanos/templates/storegateway/statefulset-sharded.yaml" at <include "thanos.objstoreConfig" .>: error calling include: template: thanos/charts/thanos-query-ssl/templates/_helpers.tpl:127:61: executing "thanos.objstoreConfig" at <.Values.objstoreConfig>: can't evaluate field Values in type int
```

### Additional information

it seems like the following PR https://github.com/bitnami/charts/pull/24014 has introduced this problem.

dragoangel:
### Description of the change

Do not render `checksum/objstore-configuration` annotation if `existingObjstoreSecret` is set, plus move `indexCacheConfig` and `bucketCacheConfig` checksum generation from `objstore-configuration` to `storegateway-configuration` as they only used by storage gateway, and other components of thanos should not be redeployed when they are changed.

### Benefits

Get working deployment when existingObjstoreSecret is set

### Possible drawbacks

None

### Applicable issues

- fixes #24536

### Checklist

### DIFF
--- a/bitnami/thanos/Chart.yaml
+++ b/bitnami/thanos/Chart.yaml
@@ -35,4 +35,4 @@ maintainers:
 name: thanos
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/thanos
-version: 14.0.1
+version: 14.0.2
--- a/bitnami/thanos/templates/_helpers.tpl
+++ b/bitnami/thanos/templates/_helpers.tpl
@@ -123,15 +123,9 @@ Return true if a secret object should be created
 Return the object store config
 */}}
 {{- define "thanos.objstoreConfig" -}}
+{{- if and .Values.objstoreConfig (not .Values.existingObjstoreSecret) }}
 objstore.yml: |-
   {{- include "common.tplvalues.render" (dict "value" .Values.objstoreConfig "context" $) | b64enc | nindent 2 }}
-{{- if .Values.indexCacheConfig }}
-index-cache.yml: |-
-  {{- include "common.tplvalues.render" (dict "value" .Values.indexCacheConfig "context" $) | b64enc | nindent 2 }}
-{{- end }}
-{{- if .Values.bucketCacheConfig }}
-bucket-cache.yml: |-
-  {{- include "common.tplvalues.render" (dict "value" .Values.bucketCacheConfig "context" $) | b64enc | nindent 2 }}
 {{- end }}
 {{- end -}}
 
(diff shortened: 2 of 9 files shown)


jotamartos (approved):
Changes LGTM, thank you for you contribution.

jkroepke:
Seems like introduced a breaking-change.

It moved the index-cache.yaml to an different configmap, however the path to that config has been not changed, too:

https://github.com/bitnami/charts/blob/9a0fece43caabdd9fe4de507cf2197945b04ac6d/bitnami/thanos/templates/storegateway/statefulset.yaml#L134

If configured, thanos may not come up, becaue it misses the index.yaml

dragoangel:
> Seems like introduced a breaking-change.
> 
> It moved the index-cache.yaml to an different configmap, however the path to that config has been not changed, too:
> 
> https://github.com/bitnami/charts/blob/9a0fece43caabdd9fe4de507cf2197945b04ac6d/bitnami/thanos/templates/storegateway/statefulset.yaml#L134
> 
> If configured, thanos may not come up, becaue it misses the index.yaml

Why? As it has same dir as was for other configmap:
https://github.com/bitnami/charts/blob/9a0fece43caabdd9fe4de507cf2197945b04ac6d/bitnami/thanos/templates/storegateway/statefulset.yaml#L129

jkroepke:
> Why? As it has same dir as was for other configmap:

Correct, but this PR moves the index-cache.yaml away from objstore.yml into a different ConfigMap

dragoangel:
> > Why? As it has same dir as was for other configmap:
> 
> Correct, but this PR moves the index-cache.yaml away from objstore.yml into a different ConfigMap

Which not mounted... Okay looks I missed this part :((

jkroepke:
An other user hits this error as well https://github.com/bitnami/charts/issues/24678

dragoangel:
My intention was that caches used here should be available only for storegateway and not added where they should not exist, but looks I missed configure mounts and change arguments.

jkroepke:
> My intention was that caches used here should be available only for storegateway

We have already `Values.storegateway.config` which extactly does this. It seems like `indexCacheConfig` is an history implementation

https://github.com/bitnami/charts/blob/9a0fece43caabdd9fe4de507cf2197945b04ac6d/bitnami/thanos/templates/storegateway/statefulset.yaml#L140

In case of `indexCacheConfig`, `storegateway.config` is an drop-in replacement. Only `bucketCacheConfig` is broken for now.

jkroepke:
@dragoangel 

Another issue is that the values inside the configmap are base64 encoded now.

https://github.com/bitnami/charts/blob/4d9c1ab019c483d78f695d5ea1a03792738f4526/bitnami/thanos/templates/_helpers.tpl#L135-L146

@jotamartos FYI: `bucketCacheConfig` and `indexCacheConfig` are broken now.

PR #24636, opened 2024-03-23
[bitnami/schema-registry]  bugfix #24545

### LINKED ISSUE #24545: [bitnami/schema-registry] Change default file mode to 0600 so that container crash can be recovered
### Name and Version

bitnami/schema-registry 17.1.1

amd64

When connecting to Kafka via SASL_SSL, `schema-registry.truststore.jks` and `schema-registry-X.keystore.jks` are copied to `/opt/bitnami/schema-registry/certs` from `/certs` (which is a projected volume, the jks are pulled in from secrets) when the container starts. Due to [this line](https://github.com/bitnami/charts/blob/9ee0e0505c76b6b82d389792abce0a6951210ba0/bitnami/schema-registry/templates/statefulset.yaml#L293), the files are created with mode `0400`. If the container crashes (not the Pod!), any subsequent restart will fail with
> cp: cannot create regular file '/opt/bitnami/schema-registry/certs/schema-registry.truststore.jks': Permission denied

since the file is not writable by the owner.

_No response_

Crash of the container should be recoverable. Why not set the `defaultMode` to `0600`?

### What do you see instead?

> cp: cannot create regular file '/opt/bitnami/schema-registry/certs/schema-registry.truststore.jks': Permission denied

### Additional information

_No response_

lkad:
### Description of the change

The permission of the key file being set to 400 caused issues during container restarts where the initialization process failed to copy the key file due to inability to overwrite existing files. To mitigate this, a check has been added during initialization to verify if the key file already exists. If it does, the copying process is skipped, thus avoiding errors resulting from attempts to overwrite existing files due to restarts.
### Applicable issues

- fixes #24545

### Checklist


andresbono (approved):
LGTM. Thank you so much for your contribution!

PR #21937, opened 2024-01-10
[bitnami/keycloak] Add revisionHistoryLimit variable for statefulset

johannes-engler-mw:
### Description of the change

Adds the variable revisionHistoryLimit to the Keycloak statefulset specification.

### Benefits

With this change its possible to change the amount of controller revisions for the statefulset.

### Checklist

- [X] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/). This is *not necessary* when the changes only affect README.md files.
- [X] Variables are documented in the values.yaml and added to the `README.md` using [readme-generator-for-helm](https://github.com/bitnami/readme-generator-for-helm)
- [X] Title of the pull request follows this pattern [bitnami/<name_of_the_chart>] Descriptive title
- [X] All commits signed off and in agreement of [Developer Certificate of Origin (DCO)](https://github.com/bitnami/charts/blob/main/CONTRIBUTING.md#sign-your-work)

PR #24690, opened 2024-03-26
[bitnami/zookeeper] Add revisionHistoryLimit variable for statefulset

Farfaday:
### Description of the change

Adds the variable revisionHistoryLimit to the zookeeper statefulset specification.

### Benefits

With this change its possible to change the amount of controller revisions for the statefulset.

### Checklist


bitnami/zookeeper/values.yaml
```diff
@@ -246,6 +246,9 @@ args: []
 ## @param replicaCount Number of ZooKeeper nodes
 ##
 replicaCount: 1
+## @param revisionHistoryLimitCount Number of controller revisions to keep
```
migruiz4, line 249: Please update the description:

```suggestion
## @param revisionHistoryLimit The number of old history to retain to allow rollback
```

migruiz4 (changes requested):
Hi @Farfaday,

I have noticed that this same change has been implemented in the `bitnami/keycloak` chart here (https://github.com/bitnami/charts/pull/21937)

But could you please rename the value as `revisionHistoryLimit` to be aligned with the rest of the charts in our catalog?

bitnami/zookeeper/values.yaml
Farfaday: I updated my commit :)

migruiz4 (approved):
Thank you very much for applying my suggestion, looks good to me!

PR #24706, opened 2024-03-27
 [bitnami/kubernetes-event-exporter] Separate adding common and pod annotations in deployment.yaml

### LINKED ISSUE #24676: [bitnami/kubernetes-event-exporter] Using podAnnotations to inject vault annotations does not work
### Name and Version

bitnami/kubernetes-event-exporter

None

Add podAnnotations in values.yaml. No matter how I try to escape the curlybraces ( using backticks {{ "{{" }} notation ) I always get the error function "secret" not defined

```
podAnnotations:
    vault.hashicorp.com/role: "kubernetes-apps"
    vault.hashicorp.com/log-level: "Info"
    vault.hashicorp.com/agent-init-first: "true"
    vault.hashicorp.com/agent-inject: "true"
    vault.hashicorp.com/agent-inject-secret-password: "kubernetes-secrets/data/tooling/es-node/k8s_events_exporter"
    vault.hashicorp.com/agent-inject-template-password: |
        {{`{{- with secret "kubernetes-secrets/data/tooling/es-node/k8s_events_exporter" -}}
          export ES_USERNAME="k8s_events_exporter"
          export ES_PASSWORD="{{ .Data.data.k8s_events_exporter }}"
          {{- end }}`}}
```

pod annotations added to the deployment and init container for vault created

### What do you see instead?

error function "secret" not defined. 

### Additional information

_No response_

epwickey:
When trying to merge common and pod annotations, vault injections cannot be added since they are parsed.

### Description of the change

The change separates separately includes common and pod annotations instead f merging them before adding the. 
### Benefits

This change will allow to add vault inject annotations to the pods. With the previous way of merging the common and pod annotations vault templates were parsed and throwing errors
### Applicable issues

- fixes #24676 

### Checklist

### DIFF
--- a/bitnami/kubernetes-event-exporter/templates/deployment.yaml
+++ b/bitnami/kubernetes-event-exporter/templates/deployment.yaml
@@ -26,9 +26,11 @@ spec:
     metadata:
       annotations:
         checksum/config: {{ include "kubernetes-event-exporter.config" . | sha256sum }}
-        {{- if or .Values.podAnnotations .Values.commonAnnotations }}
-        {{- $podAnnotations := include "common.tplvalues.merge" ( dict "values" ( list .Values.podAnnotations .Values.commonAnnotations ) "context" . ) }}
-        {{- include "common.tplvalues.render" ( dict "value" $podAnnotations "context" $ ) | nindent 8 }}
+        {{- if .Values.podAnnotations }}
+        {{- include "common.tplvalues.render" (dict "value" .Values.podAnnotations "context" $) | nindent 8 }}
+        {{- end }}
+        {{- if .Values.commonAnnotations }}
+        {{- include "common.tplvalues.render" (dict "value" .Values.commonAnnotations "context" $) | nindent 8 }}
         {{- end }}
       labels: {{- include "common.labels.standard" ( dict "customLabels" $podLabels "context" $ ) | nindent 8 }}
         app.kubernetes.io/component: kubernetes-event-exporter
--- a/bitnami/kubernetes-event-exporter/Chart.yaml
+++ b/bitnami/kubernetes-event-exporter/Chart.yaml
@@ -31,4 +31,4 @@ maintainers:
 name: kubernetes-event-exporter
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/kubernetes-event-exporter
-version: 3.0.0
+version: 3.0.1


bitnami/kubernetes-event-exporter/templates/deployment.yaml
```diff
@@ -26,9 +26,11 @@ spec:
...
       annotations:
         checksum/config: {{ include "kubernetes-event-exporter.config" . | sha256sum }}
-        {{- if or .Values.podAnnotations .Values.commonAnnotations }}
-        {{- $podAnnotations := include "common.tplvalues.merge" ( dict "values" ( list .Values.podAnnotations .Values.commonAnnotations ) "context" . ) }}
-        {{- include "common.tplvalues.render" ( dict "value" $podAnnotations "context" $ ) | nindent 8 }}
+        {{- if .Values.podAnnotations }}
+        {{- include "common.tplvalues.render" (dict "value" .Values.podAnnotations "context" $) | nindent 8 }}
+        {{- end }}
+        {{- if .Values.commonAnnotations }}
+        {{- include "common.tplvalues.render" (dict "value" .Values.commonAnnotations "context" $) | nindent 8 }}
```
juan131, line 33: The problem with this approach is that it doesn't protect the chart from including duplicated annotations (e.g. if the same annotation key is used both in `podAnnotations` and `commonAnnotations` parameters)

juan131 (changes requested):
Thanks so much for your PR!

Please correct me if I'm wrong but you're addressing the use case of using functions such as 'secret' in the templates, right? I mean, sth like this would work with the current approach:
```yaml
podAnnotations:
  foo: |
    {{- with .Values.config }}
    export LOG_LEVEL={{ .logLevel }}
    {{- end }}
```
While you're finding issues with:
```yaml
podAnnotations:
  foo: |
    {{- with secret "bar" }}
    export LOG_LEVEL={{ .Data.data.logLevel }}
    {{- end }}
```

bitnami/kubernetes-event-exporter/templates/deployment.yaml
juan131: Try using:
```yaml
podAnnotations:
  vault.hashicorp.com/agent-inject-template-password: |
    {{ printf "{{`{{- with secret \"kubernetes-secrets/data/tooling/es-node/k8s_events_exporter\" }}\n  export ES_USERNAME=\"k8s_events_exporter\"\n  export ES_PASSWORD=\"{{ .Data.data.k8s_events_exporter }}\"\n  {{- end }}" }}`}}" }}
```

Issue #24251, opened 2024-03-07
[bitnami/*] MAJOR CHANGE: Security defaults


javsalgar:
We are currently working on improving the overall security in our charts. During the following days, we will publish PRs with the following titles:

[bitnami/APP] feat!: :lock: :boom: Improve security defaults

## Changes

These PRs perform the following changes:

### Set `runAsGroup` to `1001`

Up until now, Bitnami charts were using user `1001` and group `root` (following [Openshift standards](https://access.redhat.com/documentation/en-us/openshift_container_platform/4.7/pdf/images/images.pdf)). However, there are Kubernetes platforms that do not allow using group `root`. With the new changes, containers will no longer need group `root` to work, becoming `1001` the new default `runAsGroup`.

This behavior can be reverted by setting `runAsGroup` back to `0`.

### Set `readOnlyRootFilesystem` to `true`

Using immutable filesystems is a mandatory requirement in most security checklists (such as NSA or MITRE). To allow this, we needed to change all writable paths of containers to `emptyDir` volumes. In some cases, we needed to add extra init containers that copy folders like `conf`, `plugins` or `logs`. This is because mounted volumes substitute the contents of the original folders.

*IMPACT*: This may cause issues with customization scripts like `initScripts` or using custom `command` and `args`. Using older versions of Bitnami containers may cause issues as they may not contain the required bash changes to allow `readOnlyRootFilesystem`.

This behavior can be reverted by setting `readOnlyRootFilesystem` back to `false`.

### Enable `resourcePresets` by default

Having pods without any resource requests or limits is strongly discouraged. These containers may cause major problems like draining node resources. Currently, our charts warn the user if the `resources` object is not set. In addition to this, we have the `resourcesPreset` value to help users test different resource configurations, but this is not recommended for production. In the new changes, we set the `resourcesPreset` value to the minimum size that works in our internal testing. This is *by no means* the recommended size by the Bitnami team, but it is the minimum size to perform basic testing of the solution. We insist that users should set their own `resource` values according to their needs and use case.

*IMPACT*: Users already setting the `resource` value should be unaffected by this change. Those that were not setting resources may see a performance drop in their workloads as they would be configured to a minimal value for basic testing. If that is the case, we strongly advise you to set the `resources` value before performing any upgrade.

This behavior can be reverted by setting `resourcesPreset` back to `none`

### Perform automatic adaptation of `securityContext` sections in Openshift

[...]

PR #24716, opened 2024-03-27
[bitnami/tomcat] feat!: :lock: :boom: Improve security defaults

javsalgar:
BREAKING CHANGE

### Description of the change

This major bump changes the following security defaults:

- `runAsGroup` is changed from `0` to `1001`
- `readOnlyRootFilesystem` is set to `true`
- `resourcesPreset` is changed from `none` to the minimum size working in our test suites (NOTE: `resourcesPreset` is not meant for production usage, but `resources` adapted to your use case).
- `global.compatibility.openshift.adaptSecurityContext` is changed from `disabled` to `auto`.
- The `networkPolicy` section has been normalized amongst all Bitnami charts. Compared to the previous approach, the values section has been simplified (check the Parameters section) and now it set to `enabled=true` by default. Egress traffic is allowed by default and ingress traffic is allowed by all pods but only to the ports set in `containerPorts` and `extraContainerPorts`.

### Benefits

- Better compliance with NSA and MITRE security checklists
- Improved compatibility with Openshift

### Possible drawbacks

This could potentially break any customization or init scripts used in your deployment. If this is the case, change the default values to the previous ones.

### Applicable issues

- https://github.com/bitnami/charts/issues/24251

### Checklist


bitnami/tomcat/templates/_pod.tpl
```diff
@@ -150,6 +150,24 @@ containers:
     volumeMounts:
       - name: data
         mountPath: /bitnami/tomcat
+      - name: empty-dir
+        mountPath: /opt/bitnami/tomcat/webapps
+        subPath: app-webapps-dir
```
fmulero, line 155: `/opt/bitnami/tomcat/webapps` is "virtually mounted" and persisted in the `/bitnami/tomcat` volume. If we mount this empty volume, we will remove installed apps every time the pod is removed.

fmulero (changes requested):
Thanks a lot @javsalgar 

I left a comment, could you review it?

bitnami/tomcat/templates/_pod.tpl
javsalgar: Actually, we're not persisting the webapps, looking at the container logic. What we're doing in the container logic is checking if the webapps folder is empty or not, but we're not applying any persistence logic to that folder.

```
    # Fix to make upgrades from old images work
    # Before, we were persisting 'data' dir instead of 'webapps', causing errors when restoring persisted data
    if ! is_dir_empty "$TOMCAT_WEBAPPS_DIR" || ! is_dir_empty "${TOMCAT_VOLUME_DIR}/data"; then
        info "Persisted webapps detected"
        if [[ ! -e "$TOMCAT_WEBAPPS_DIR" && -e "${TOMCAT_VOLUME_DIR}/data" ]]; then
            warn "Detected legacy configuration directory path ${TOMCAT_VOLUME_DIR}/conf in volume"
            warn "Creating ${TOMCAT_BASE_DIR}/webapps symlink pointing to ${TOMCAT_VOLUME_DIR}/data"
            ln -sf "${TOMCAT_VOLUME_DIR}/data" "${TOMCAT_BASE_DIR}/webapps"
        fi
    else
        info "Ensuring Tomcat directories exist"
        ensure_dir_exists "$TOMCAT_WEBAPPS_DIR"
        # Use tomcat:root ownership for compatibility when running as a non-root user
        am_i_root && configure_permissions_ownership "$TOMCAT_WEBAPPS_DIR" -d "775" -f "664" -u "$TOMCAT_DAEMON_USER" -g "root"

        if is_boolean_yes "$TOMCAT_INSTALL_DEFAULT_WEBAPPS"; then
            info "Deploying Tomcat from scratch"
            cp -rp "$TOMCAT_BASE_DIR"/webapps_default/* "$TOMCAT_WEBAPPS_DIR"

            # These applications have been enabled for historical reasons, and do not pose any security threat
            tomcat_enable_application examples
            tomcat_enable_application docs
            if is_boolean_yes "$TOMCAT_ALLOW_REMOTE_MANAGEMENT"; then
                # These applications should not be enabled by default, for security reasons
                info "Enabling remote connections for manager and host-manager applications"
                tomcat_enable_application manager
                tomcat_enable_application host-manager
            fi
        else
            info "Skipping deployment of default webapps"
        fi

    fi
```

However, we should check in an init container if the container has populated the webapps directory.

fmulero: I think I am missing something. In the postunpack.sh file: https://github.com/bitnami/containers/blob/7fa47487ccc2519fdfeffdff630d0ed38cbc3c71/bitnami/tomcat/10.1/debian-12/rootfs/opt/bitnami/scripts/tomcat/postunpack.sh#L37-L39

We can see that `${TOMCAT_BASE_DIR}/webapps` is a symlink to `$TOMCAT_WEBAPPS_DIR` (by default `/bitnami/tomcat/webapps`). If we mount an emptyDir over `${TOMCAT_BASE_DIR}/webapps` we will need to recreate the logic above in the chart.

fmulero (approved):
LGTM

PR #24730, opened 2024-03-28
bitnami/redis: allow to set containerSecurityContext on kubectl container to fix issue e.g. with OpenShift

anessi:
### Description of the change

In https://github.com/bitnami/charts/issues/4082 a new feature has been implemented to allow exposing a single service to access the active Redis master. The `kubectl` POD that is used to update the labels has a hard-coded `runAsUser` of `0`. This causes the following issues:
* Uses the root user which is not necessary to run kubectl and therefore a security issue
* Will not work on e.g. OpenShift where the user is dynamically assigned and must not be `0`

### Benefits

Adds security and makes it work on OpenShift.
Allows the user to fully customize the `securityContext`.

### Possible drawbacks

More complexity as more configuration options are given.

### Applicable issues

No issue, but related to https://github.com/bitnami/charts/issues/4082
FYI: @anthosz and @juan131

### Checklist

I was not able to fully run `readme-generator-for-helm` because the newest version complains about values with `@skip`. Seems this is broken.


juan131 (approved):
Thanks so much for this enhancement! LGTM

PR #24763, opened 2024-04-01
bitnami/kibana - Adapt the securityContext also for the init container

### LINKED ISSUE #24762: Kibana fails to open secret content on OpenShift
### Name and Version

bitnami/kibana 11.0.1

amd64

Relevant only for OpenShift

```yaml
elasticsearch:
  security:
    tls:
      enabled: true
      existingSecret: < SecretName >
      usePemCerts: true
```

Kibana should be able to access the certificates in the secret

### What do you see instead?

Kibana returns a permissions error

### Additional information

The root cause if the `securityContext` of the `initContainer` not being templated for OpenShift

ygalblum:
### Description of the change

The kibana chart includes a mechanism to adapt the `securityContext` when used on OpenShift. This mechanism was used only for the main container, but not for the initContainer. According to K8S's behavior, the `securityContext` of the initContainer is also applied to the main container. As a result, when passing secrets to the container, kibana fails to open the file due to permissions. The fix applies the same mechanism for the initContainer.

### Benefits

Without this change, kibana is unable to access the provided certificates. The chart offers a workaround of disabling the `securityContext` altogether. But, I believe it's less desiarable.

### Possible drawbacks

Don't know any

### Applicable issues

- fixes #24762

### Checklist

- [X ] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/). This is *not necessary* when the changes only affect README.md files.
- [ X] Title of the pull request follows this pattern [bitnami/<name_of_the_chart>] Descriptive title
- [ X] All commits signed off and in agreement of [Developer Certificate of Origin (DCO)](https://github.com/bitnami/charts/blob/main/CONTRIBUTING.md#sign-your-work)

### DIFF
--- a/bitnami/kibana/Chart.yaml
+++ b/bitnami/kibana/Chart.yaml
@@ -32,4 +32,4 @@ maintainers:
 name: kibana
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/kibana
-version: 11.0.1
+version: 11.0.2
--- a/bitnami/kibana/templates/deployment.yaml
+++ b/bitnami/kibana/templates/deployment.yaml
@@ -101,7 +101,7 @@ spec:
           image: {{ include "kibana.image" . }}
           imagePullPolicy: {{ .Values.image.pullPolicy | quote }}
           {{- if .Values.containerSecurityContext.enabled }}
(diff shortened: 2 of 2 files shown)


dgomezleon (approved):
LGTM

PR #24786, opened 2024-04-02
[bitnami/solr] Add revisionHistoryLimit variable for statefulset

Farfaday:
### Description of the change

Adds the variable revisionHistoryLimit to the solr statefulset specification.

### Benefits
With this change its possible to change the amount of controller revisions for the statefulset.
### Checklist


bitnami/solr/templates/statefulset.yaml
```diff
@@ -16,6 +16,7 @@ metadata:
 spec:
   podManagementPolicy: {{ .Values.podManagementPolicy }}
   replicas: {{ .Values.replicaCount }}
+  revisionHistoryLimit: {{ .Values.revisionHistoryLimit }}
```
Mauraza, line 19:
```suggestion
{{ if .Values.revisionHistoryLimit }}
  revisionHistoryLimit: {{ .Values.revisionHistoryLimit }}
 {{- end }}
```

bitnami/solr/values.yaml
```diff
@@ -171,6 +171,9 @@ extraEnvVarsSecret: ""
 ## @param replicaCount Number of solr replicas
 ##
 replicaCount: 3
+## @param revisionHistoryLimit The number of old history to retain to allow rollback
+##
+revisionHistoryLimit: 10
```
Mauraza, line 176:
```suggestion
revisionHistoryLimit: ""
```

bitnami/solr/README.md
```diff
@@ -209,6 +209,7 @@ You can enable this initContainer by setting `volumePermissions.enabled` to `tru
 | Name                                                | Description                                                                                                                                                                                                | Value            |
 | --------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ---------------- |
 | `replicaCount`                                      | Number of solr replicas                                                                                                                                                                                    | `3`              |
+| `revisionHistoryLimit`                              | The number of old history to retain to allow rollback                                                                                                                                                      | `10`             |
```
Mauraza, line 212:
```suggestion
| `revisionHistoryLimit`                              | The number of old history to retain to allow rollback                                                                                                                                                      | ``             |
```

Mauraza (commented):
Hi @Farfaday,

Thanks for the contribution! I want don't force the user from using revisionHistoryLimit, so I think we can add an 'if'. what do you think?

bitnami/solr/templates/statefulset.yaml
Farfaday: The default value for revisionHistoryLimit is 10, so my change proposal is not a breaking change, it keeps the defaults. Putting this inside an if block or not setting the value, does not "unset" the value at the end in kubernetes.

BUT: if the user sets the value to 0 as an integer (which is natural to me), then the if condition will be false.
The revisionHistoryLimit won't be set by helm, so kubernetes will use the default value for revisionHistoryLimit which is... 10. A user setting revisionHistoryLimit to 0 (integer) in values.yaml will at the end then get 10. :face_with_spiral_eyes: 
This would be a bug to me.
I would recommend against the if block and the empty value in the values.yaml

Mauraza: You are right about the default value, but if "you set this field to 0, will result in cleaning up all the history of your Deployment thus that Deployment will not be able to roll back"
https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#clean-up-policy.

It would be nice if you put something about the value 0, in values.yaml to clarify things.

Mauraza:
Hi @Farfaday, 

Could you check my last comment?  for the rest, LGTM!

Mauraza (approved):
Hi @Farfaday,

Thanks again for the contribution!!

PR #22557, opened 2024-01-23
[bitnami/rabbitmq] fix: make toBytes handle numbers with a decimal dot

tomasz-torcz-airspace-intelligence:
Replace `toBytes` function with the one lifted from
https://github.com/redpanda-data/helm-charts/blob/main/charts/redpanda/templates/_helpers.tpl#L199C1-L236C12

Redpanda's charts are Apache Version 2.0 licensed.

Signed-off-by: Tomasz Torcz <tomasz.torcz@airspace-intelligence.com>

### Description of the change

Fix `toBytes` function to handle values with decimal dot, like `1.5Gi`.

### Benefits

Previously such values were silently ignored, causing a generation of a faulty rabbitmq.conf.

### Possible drawbacks

None.

### Applicable issues

- fixes https://github.com/bitnami/charts/issues/4028#issuecomment-710029668

### Checklist

- [X] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/). This is *not necessary* when the changes only affect README.md files.
- [X] Variables are documented in the values.yaml and added to the `README.md` using [readme-generator-for-helm](https://github.com/bitnami/readme-generator-for-helm)
- [X] Title of the pull request follows this pattern [bitnami/<name_of_the_chart>] Descriptive title
- [X] All commits signed off and in agreement of [Developer Certificate of Origin (DCO)](https://github.com/bitnami/charts/blob/main/CONTRIBUTING.md#sign-your-work)

PR #24866, opened 2024-04-04
[bitnami/rabbitmq] Fix misleading memoryHighWatermark.value description in values

alexhfmnn:
### Description of the change

The description of memoryHighWatermark.value is wrong, when memoryHighWatermark.type is set to "absolute". According to the current description, "you can also use an absolute value, e.g.: 256MB". However, the [helpers function](https://github.com/bitnami/charts/blob/0c725da8074070f67e8f1efe65e194a262a75f06/bitnami/rabbitmq/templates/_helpers.tpl#L177) only accepts for example mebibytes instead of megabytes, so either "265M" or "265Mi" as syntax

### Applicable issues

- fixes #

### Checklist


jotamartos:
Hi @alexhfmnn,

Thank you for your suggestion. I took a look at it and I found that the chart doesn't work when you try to use it 

```
diff --git a/bitnami/rabbitmq/values.yaml b/bitnami/rabbitmq/values.yaml
index adbee7ee51..e55d42c914 100644
--- a/bitnami/rabbitmq/values.yaml
+++ b/bitnami/rabbitmq/values.yaml
@@ -234,17 +234,17 @@ onlineSchedulers: ""
 memoryHighWatermark:
   ## @param memoryHighWatermark.enabled Enable configuring Memory high watermark on RabbitMQ
   ##
-  enabled: false
+  enabled: true
   ## @param memoryHighWatermark.type Memory high watermark type. Either `absolute` or `relative`
   ##
-  type: "relative"
+  type: "absolute"
   ## Memory high watermark value.
   ## @param memoryHighWatermark.value Memory high watermark value
   ## The default value of 0.4 stands for 40% of available RAM
   ## Note: the memory relative limit is applied to the resource.limits.memory to calculate the memory threshold
-  ## You can also use an absolute value, e.g.: 256MB
+  ## You can also use an absolute value, e.g.: 256Mi
   ##
-  value: 0.4
+  value: 256Mi
 ## @param plugins List of default plugins to enable (should only be altered to remove defaults; for additional plugins use `extraPlugins`)
 ##
 plugins: "rabbitmq_management rabbitmq_peer_discovery_k8s"
```

I get this error when installing the chart 

```
Error: INSTALLATION FAILED:
... 
 is invalid SI quantity
Suffixes can be: b | B | k | K | m | M | g | G | Ki | Mi | Gi or without any Suffixes
```

It seems this previous change broke this functionality

https://github.com/bitnami/charts/pull/22557

I'll take a look at it and see how we can fix it.

alexhfmnn:
Hi @jotamartos,

I found out that both "absolute" and "relative" need to have resource limits defined under resource.limits.memory so that the value for total_memory_available_override_value can be set accordingly in the RabbitMQ Config-Secret. According to the [RabbitMQ docs](https://www.rabbitmq.com/docs/configure#config-items), this key should be defined if the application is operated in containers, for example "when the value is set to 4 GB, the node will believe it is running on a machine with 4 GB of RAM" - so that the defined resource limit of the Pod is complied with

I added another commit that makes the Chart render again

jotamartos (approved):
Thanks for your contribution.

saravanan-ai:
This is still open we get
{{ MEM_HIGH_WATERMARK }} is invalid SI quantity
Suffixes can be: b | B | k | K | m | M | g | G | Ki | Mi | Gi or without any Suffixes

we are using helm templating and we are using it like 
memoryHighWatermark:
  enabled: true
  type: absolute
  value: "{{ MEM_HIGH_WATERMARK }}"
  
  and in dev,prod,qa values.yaml files we use it like
  MEM_HIGH_WATERMARK: 512Mi

alexhfmnn:
Hi @saravanan-ai,
are you using the latest Helm chart version?

Still seems to work fine for me with this example on the latest version:
```console
$ cat <<EOF > rmq_values.yaml
resources:
  limits:
    memory: 2Gi
memoryHighWatermark:
  enabled: true
  type: "absolute"
  value: 512Mi
EOF
```

```console
$ helm upgrade --install --dry-run rabbitmq oci://registry-1.docker.io/bitnamicharts/rabbitmq -n rabbitmq -f rmq_values.yaml --version 14.6.2 | grep "rabbitmq\.conf" -A 1 | yq '."rabbitmq.conf" | @base64d' | grep "vm_memory_high_watermark" -B 3 -A 1
```

Issue #24938, opened 2024-04-05
[bitnami/keycloak] Fail to get theme in version 23+ since chart 20.0.0 is going readonly filesystem


AlexisDucastel:
### Name and Version

bitnami/keycloak 20.0.0

### What architecture are you using?

amd64

### What steps will reproduce the bug?

In several releases of keycloak deployed via Helm with chart "20.0.0", different clusters, different config, we have the template that is not loading, with this error in logs :

```
2024-04-05 08:43:58,179 WARN  [org.keycloak.services] (executor-thread-1) KC-SERVICES0075: Failed to get theme request: java.lang.RuntimeException: Temporary directory /opt/bitnami/keycloak/bin/../data/tmp does not exist and it was not possible to create it.
2024-04-05T10:43:58.180068787+02:00 	at org.keycloak.quarkus.runtime.integration.QuarkusPlatform.getTmpDirectory(QuarkusPlatform.java:152)
2024-04-05T10:43:58.180072936+02:00 	at org.keycloak.encoding.GzipResourceEncodingProviderFactory.initCacheDir(GzipResourceEncodingProviderFactory.java:70)
2024-04-05T10:43:58.180075966+02:00 	at org.keycloak.encoding.GzipResourceEncodingProviderFactory.create(GzipResourceEncodingProviderFactory.java:29)
```

### Are you using any custom parameters or values?

Values example : 
```yaml
        auth:
          <REDACTED>

        postgresql:
          auth:
            <REDACTED>

        ingress:
          annotations:
            cert-manager.io/cluster-issuer: letsencrypt
            nginx.ingress.kubernetes.io/proxy-buffer-size: 16k
          certManager: true
          enabled: true
          hostname: <REDACTED>
          ingressClassName: nginx
          pathType: Prefix
          tls: true

        production: true
        proxy: edge
        service:
          type: ClusterIP

        resources:
          limits:
            memory: 1Gi
          requests:
            cpu: 500m
            memory: 1Gi
```

### What is the expected behavior?

The theme should load properly

### What do you see instead?

The theme is not loaded, leading to 500 HTTP error and unfunctional administration console (hopefully, login interface still works without theme, even if graphically-broken)

Admin :
![image](https://github.com/bitnami/charts/assets/1911945/be6c4b9a-4cd2-470e-9ea7-693006b88efe)

Login:
![image](https://github.com/bitnami/charts/assets/1911945/4dd25911-05a9-4144-a32e-d235891c720a)

Account console:
[...]

PR #24939, opened 2024-04-05
[bitnami/keycloak] fix: :bug: Add empty-dir in data folder

javsalgar:

### Description of the change
This PR makes the /opt/bitnami/keycloak/data folder writable by using an empty-dir

### Applicable issues

- fixes https://github.com/bitnami/charts/issues/24938
- fixes https://github.com/bitnami/containers/issues/64802

### Checklist

### DIFF
--- a/bitnami/keycloak/templates/statefulset.yaml
+++ b/bitnami/keycloak/templates/statefulset.yaml
@@ -284,6 +284,9 @@ spec:
             - name: empty-dir
               mountPath: /opt/bitnami/keycloak/lib/quarkus
               subPath: app-quarkus-dir
+            - name: empty-dir
+              mountPath: /opt/bitnami/keycloak/data
+              subPath: app-data-dir
             {{- if or .Values.configuration .Values.existingConfigmap }}
             - name: keycloak-config
               mountPath: /bitnami/keycloak/conf/keycloak.conf
--- a/bitnami/keycloak/Chart.yaml
+++ b/bitnami/keycloak/Chart.yaml
@@ -33,4 +33,4 @@ maintainers:
 name: keycloak
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/keycloak
-version: 20.0.0
+version: 20.0.1


carrodher (approved):

bitnami/keycloak/templates/statefulset.yaml
```diff
@@ -284,6 +284,9 @@ spec:
             - name: empty-dir
               mountPath: /opt/bitnami/keycloak/lib/quarkus
               subPath: app-quarkus-dir
+            - name: empty-dir
```
JBKoerber, line 287: We have a use case where we want to pre-populate our keycloak dockerfile image with existing data.
This particular mount over an existing data folder, basically gives us an empty folder, without our preload data once this is mounted.

can we have a values override added .e.g.?

            {{- if .Values.enableDataVolumeMount }}

we coded an example and were using it but would like this in future bitnami versions
image
image

juan131: @JBKoerber could you please try to install the chart using the `values.yaml` below and let us know if that works for you? That should avoid the emptydir to ovewrite the data you pre-populated in the image.

```yaml
initContainers:
  - name: prepare-write-dirs
    image: "{{ template \"keycloak.image\" . }}"
    imagePullPolicy: "{{ .Values.image.pullPolicy }}"
    command:
      - /bin/bash
    args:
      - -ec
      - |
        . /opt/bitnami/scripts/liblog.sh

        info "Copying writable dirs to empty dir"
        # In order to not break the application functionality we need to make some
        # directories writable, so we need to copy it to an empty dir volume
        cp -r --preserve=mode /opt/bitnami/keycloak/data /emptydir/app-data-dir
        info "Copy operation completed"
    volumeMounts:
      - name: empty-dir
        mountPath: /emptydir
```

JBKoerber: @juan131 Thank you so much for the quick followup! It works, and  I was able to confirm using your recommended initContainers override code above.   It now has data/import files we were looking for using that approach.

juan131: Thanks for confirming the workaround works @JBKoerber! I created [this PR](https://github.com/bitnami/charts/pull/28550) to adopt it by default so you don't need to use the custom values.

PR #25040, opened 2024-04-08
[bitnami/nginx] fix: correct ternary when tls is enabled

Rod-Sol:
### Description of the change

In deployment.yaml file, it seems that the ternary used by probes is incorrect since the use of the value ```tls.enabled``` done by commit e3d035dfe226568e07b40b6f561d591199e2706f. When we deactivate the TLS, the deployment could not start because a connection attempt is done in HTTPS and not HTTP.

### Benefits

Fix issue when TLS is disabled.

### Possible drawbacks

HTTPS doesn't work as intended, I didn't test this kind of configuration

### Applicable issues

-

### Additional information

-

### Checklist

### DIFF
--- a/bitnami/nginx/Chart.yaml
+++ b/bitnami/nginx/Chart.yaml
@@ -34,4 +34,4 @@ maintainers:
 name: nginx
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/nginx
-version: 16.0.1
+version: 16.0.2
--- a/bitnami/nginx/templates/deployment.yaml
+++ b/bitnami/nginx/templates/deployment.yaml
@@ -252,21 +252,21 @@ spec:
           {{- else if .Values.livenessProbe.enabled }}
           livenessProbe: {{- include "common.tplvalues.render" (dict "value" (omit .Values.livenessProbe "enabled") "context" $) | nindent 12 }}
             tcpSocket:
-              port: {{ ternary "http" "https" .Values.tls.enabled }}
+              port: {{ ternary "https" "http" .Values.tls.enabled }}
           {{- end }}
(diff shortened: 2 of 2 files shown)


juan131 (approved):
Thanks for the fix! LGTM

Cogitri:
This seems to break the deployment for me: `Warning  Unhealthy  0s (x4 over 15s)  kubelet            Readiness probe errored: strconv.Atoi: parsing "https": invalid syntax`

Rod-Sol:
@Cogitri I think I found your issue. In deployment.yaml:

``` 
          ports:
            {{- if .Values.containerPorts.http }}
            - name: http
              containerPort: {{ .Values.containerPorts.http }}
            {{- end }}
            {{- if .Values.containerPorts.https }}
            - name: https
            {{- end }}
```

If you use TLS, I think you should also set the https inside the values.yaml file
``` 
## @param containerPorts.http Sets http port inside NGINX container
## @param containerPorts.https Sets https port inside NGINX container
##
containerPorts:
  http: 8080
  https: ""
```
@juan131 I think there is some modification missing with the TLS update of the chart, should I add a message inside the previous PR to the author ?

bobbeeke:
> This seems to break the deployment for me: `Warning Unhealthy 0s (x4 over 15s) kubelet Readiness probe errored: strconv.Atoi: parsing "https": invalid syntax`

Same issue here. Upgraded chart 16.0.1 -> 16.0.2
I use it simply as a proxy to forward traffic to a http endpoint in my cluster. No https.

juan131:
Thanks for reporting the issue! I just created https://github.com/bitnami/charts/pull/25091 to address it

PR #20934, opened 2023-11-14
[bitnami/nginx] feat: disable http port if containerPorts.http is empty

robinelfrink:
### Description of the change

This change will disable the `http` port if `containerPorts.http` is empty.

### Benefits

When using `NodePort` and `http` is not used, it will still occupy a port on the node, even if unused.

### Possible drawbacks

If `containerPorts.http` is not set, and the default `readinessProbe` is used, the pod will never reach readiness.

### Applicable issues

Useful if using only the `https` port.

My specific use-case: I have another application using a defined node port. On installation the nginx-chart is installed first, causing a random node port being assigned to `http`, with a chance of being the same as the defined port for the other application.

### Checklist

- [X] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/). This is *not necessary* when the changes only affect README.md files.
- [X] Title of the pull request follows this pattern [bitnami/<name_of_the_chart>] Descriptive title
- [X] All commits signed off and in agreement of [Developer Certificate of Origin (DCO)](https://github.com/bitnami/charts/blob/main/CONTRIBUTING.md#sign-your-work)

PR #25059, opened 2024-04-08
[bitnami/nginx] Fix ternary on probe port name selection

krmcbride:
### Description of the change

Fixes a backwards ternary in the nginx deployment that would result in the nginx pod probes failing when `tls.enabled=false` was set, due to the probes trying to use a `https` port that didn't exist.

At first I just reversed the `http` and `https` parts but that didn't cut it, because whether or not there is a port named `https` at all depends on `containerPorts.https`, not `tls.enabled`.

### Benefits

The end result is that `tls.enabled=false` continues to default probe ports to `http`, and the probes succeed.  Setting `containerPorts.https` to a number updates the probes to use the `https` port, which I assume is the original intended behavior.

### Possible drawbacks

I did not attempt to replicate all of the [ports logic](https://github.com/bitnami/charts/blob/66f3026/bitnami/nginx/templates/deployment.yaml#L237-L248) in the probe ternary. The ternary continues to assume that if both `https` and `http` ports are configured, the `https` port should be used for probes.

### Checklist


bitnami/nginx/templates/deployment.yaml
```diff
@@ -252,21 +252,21 @@ spec:
           {{- else if .Values.livenessProbe.enabled }}
           livenessProbe: {{- include "common.tplvalues.render" (dict "value" (omit .Values.livenessProbe "enabled") "context" $) | nindent 12 }}
             tcpSocket:
-              port: {{ ternary "http" "https" .Values.tls.enabled }}
+              port: {{ ternary "http" "https" (empty .Values.containerPorts.https) }}
```
fmulero, line 255: Your proposal could be enough but keeping in mind how we are setting the ports some lines above and the proposal done in #20934, I'd use something like (not tested):
```suggestion
              port: {{ ternary "https" "http" (and (not  .Values.containerPorts.http)  .Values.containerPorts.https) }}
```

WDYT?

fmulero (commented):
Thanks a lot @krmcbride  for the fix!!!

I left a little suggestion could you give it a glance?

bitnami/nginx/templates/deployment.yaml
krmcbride: Makes sense to me, however I get nothing but `wrong type for value; expected bool; got string` if I don't use `empty`, so I've pushed a commit that combines both.  Tested it and the `https` port is used _if and only if_ `containerPorts.http` is empty and `containerPorts.https` is set, else all combinations result in `http` being used.

fmulero (approved):
I've just fixed the conflicts.

LGTM

PR #25091, opened 2024-04-10
[bitnami/nginx] Define by default a https container port now TLS is enabled

juan131:
### Description of the change

Ensure there's a defined https container port by default since TLS is enabled.

### Benefits

Readiness & liveness probe no longer fail.

### Possible drawbacks

None

### Applicable issues

Issue reported at https://github.com/bitnami/charts/pull/25040

### Additional information

Bug introduced at https://github.com/bitnami/charts/commit/e3d035dfe226568e07b40b6f561d591199e2706f & https://github.com/bitnami/charts/pull/25040

### Checklist


Rod-Sol:
Hello @juan131,

By doing this, it seems that the Chart will have both `http` and `https` ports exposed inside the container, for example in the 'deployment.yaml' (you also have it in `service.yaml`):
```
{{- if .Values.containerPorts.http }}
    - name: http
      containerPort: {{ .Values.containerPorts.http }}
{{- end }}
{{- if .Values.containerPorts.https }}
    - name: https
{{- end }}
 ```

This doesn't seems right if we want to expose one but not the other ? It will also add a var for the Nginx image that I think will be used by the Nginx configuration: `NGINX_HTTPS_PORT_NUMBER`

In my opinion, the `https` port should not be set by default, and so the TLS also ? Or if it's the new default behavior, maybe adding info about how to disable `https` to have only `http` in the `values.yaml` ?

**Comment by juan131** [comment]
Hi @Rod-Sol 

Yes, this would enable **by default** both ports in NGINX containers and the associated service. However, please note it's just a default behavior and it's possible to disable the `http` one if that's desired using the values below:

```yaml
containerPorts:
  http: ""
```

andresbono (approved):

PR #25103, opened 2024-04-10
[bitnami/thanos] Enable Ingress only when component is enabled

jakuboskera:
### Description of the change

When some Thanos component is disabled and Ingress of this component is enabled, the Ingress component should not be rendered.

E.g. In general values file (same for multiple instances) I will create a general configuration of component bucketweb. Then in values file for each instance I will just enable or disable this bucketweb component.

```yaml
# values.general.yaml

bucketweb:
  enabled: false
  ingress:
    enabled: true
```

```yaml
# values.cluster-1.yaml

# In this case Ingress resource will be rendered as whole
# component is enabled and ingress is enabled in general by default

bucketweb:
  enabled: true
```

```yaml
# values.cluster-2.yaml

# In this case Ingress resource won't be rendered as whole
# component is disabled even ingress is enabled in general by default

bucketweb:
  enabled: false
```

Values applied in way `helm install thanos bitnami/thanos -f values.general.yaml -f values.cluster-N.yaml`

### Benefits

It will reduce the necessary configuration for multiple instances of the same Chart using different values ​​files from general to specific.

### Possible drawbacks

N/A

### Applicable issues

N/A

### Additional information

N/A

### Checklist

### DIFF
--- a/bitnami/thanos/Chart.yaml
+++ b/bitnami/thanos/Chart.yaml
@@ -35,4 +35,4 @@ maintainers:
 name: thanos
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/thanos
-version: 15.0.4
+version: 15.0.5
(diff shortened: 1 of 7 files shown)


juan131 (approved):

PR #25121, opened 2024-04-10
[bitnami/redis] nullify dbfilename config to prevent RDB creation

### LINKED ISSUE #6389: dump.rdb file is exist when RDB persistence disabled (default installation)
**Which chart**:
redis

**Describe the bug**
dump.rdb file exists when RDB persistence disabled on fresh installation

**To Reproduce**
1. Install redis `helm install redis bitnami/redis`
2. Execute shell on it `kubectl exec -it redis2-master-0 /bin/bash`
3.  Check data dir
```
I have no name!@redis-master-0:/$ ls /data
appendonly.aof	dump.rdb
I have no name!@redis-master-0:/$ ls -la  /data
total 16
drwxrwxrwx 2 root root 4096 May 17 13:06 .
drwxr-xr-x 1 root root 4096 May 17 13:06 ..
-rw-r--r-- 1 1001 root  818 May 17 13:17 appendonly.aof
-rw-r--r-- 1 1001 root  175 May 17 13:06 dump.rdb
```

**Expected behavior**
If RDB persistence is disabled then my **assumption** that there should be no dumb.rdb file as aof is in use or documentation/comments in values.yaml should be extended.

**Additional context**
Why I took attention to this. 
```
I have no name!@redis-master-0:/$ cat /opt/bitnami/redis/etc/redis.conf 
# User-supplied configuration:
# Enable AOF https://redis.io/topics/persistence#append-only-file
appendonly yes
# Disable RDB persistence, AOF persistence already enabled.
save ""I have no name!@redis-master-0:/$ ls -lha /data
total 6.0G
drwxrwsrwx 2 root 1001 4.0K May 17 06:11 .
drwxr-xr-x 1 root root 4.0K May 15 18:54 ..
-rw-r--r-- 1 1001 1001 3.3G May 17 08:35 appendonly.aof
-rw-r--r-- 1 1001 1001 2.8G May 15 18:58 dump.rdb
```
RDB file is quite big on dev instance and maybe it shouldn't exist at all and save some space on disk.
I am not pointing this as an example(focus on a fresh install) because probably somebody could save rdb file manually via redis-cli or api.

shlomitubul:
### Description of the change

Currently, even though the chart default config is disabling RDB and using AOF with diskless replication in each full sync or pod restarts RDB file created, by nullifying dbfilename config key we prevent this behavior

### Benefits
potentially reduce IO and storage usage

### Possible drawbacks
even though we defaulting to diskless with no rdb file, it might confuse some users

### Applicable issues

- fixes #6389 

### Checklist


juan131 (approved):
Thanks for this improvement! LGTM

shlomitubul:
@juan131 Thanks for the quick review and approval! I see that some checks failed but there are no logs or hints as to why that happened, can you retrigger CI?

juan131:
Some Goss tests are failing:

```
FAILED	bbf9e22a	Command redis-master-role stdout	0.000	goss	ERRORS: [Command: redis-master-role: stdout: Expected "object: *bytes.Reader" to have patterns ["master","redis-replicas-0","redis-replicas-1","redis-replicas-2"] the missing elements were ["redis-replicas-0","redis-replicas-1","redis-replicas-2"]] FAILURES: [Command: redis-master-role: stdout: Expected "object: *bytes.Reader" to have patterns ["master","redis-replicas-0","redis-replicas-1","redis-replicas-2"] the missing elements were ["redis-replicas-0","redis-replicas-1","redis-replicas-2"]]
FAILED	682ffbcf	Command redis-set-key-value-pairs exit-status
```

This comes from:

- https://github.com/bitnami/charts/blob/main/.vib/redis/goss/goss.yaml#L9-L13
- https://github.com/bitnami/charts/blob/main/.vib/redis/goss/goss.yaml#L21-L28

I'll try to take a look later to figure out what's going on.

shlomitubul:
@juan131 Although redis load and works if I set dbfilename to "" it did throw errors like:
`Failed trying to rename the temp DB into  in MASTER <-> REPLICA synchronization: No such file or directory`
Then Redis triggers full sync again and again since he can't create a temp AOF filename (as now he is an empty string), which seems to be constructed from dbfilename , till I find a solution for this I'm converting this to a draft

shlomitubul:
@juan131 After exploring my Redis deployment, I realized I didn't fully understand how diskless replication works. It turns out that setting repl-diskless-sync yes tells Redis that during a full-sync event, instead of forking a child process to write the RDB to disk and then loading it back into memory, the master will write the RDB directly to the replica socket. However, this might not affect how the replica loads the RDB. For this, we have another parameter called repl-diskless-load with three options:

"disabled": The default setting, where the RDB file is written to disk first and then loaded from memory.
"swapdb": Loads the RDB into memory without writing anything to disk. This can use double the memory and might cause OOM.
"on-empty-db": If no dataset exists (like with a new replica), then load from memory; otherwise, write to and load from disk.
If someone wants to fully disable RDB creation, they might use the "swapdb" option. Just to clarify, there is a good reason why this is not the default since it could cause OOM and increase the risk of data loss. 

Im closing this PR since it not relevant anymore
(https://raw.githubusercontent.com/redis/redis/7.2.0/redis.conf#:~:text=repl%2Ddiskless%2Dload%20disabled)

PR #25124, opened 2024-04-11
[bitnami/odoo] Add pre-install/pre-upgrade job for module installation.

samaroon:
### Description of the change

This PR introduces a new Kubernetes job defined in the Odoo Helm chart, which is triggered during the pre-install and pre-upgrade phases. The job, named `odoo-addons-install`, utilizes arguments (`args`) and environment variables (`env`) from the `values` file to configure and run the Odoo modules installation. This setup ensures that all necessary modules are installed and configured before any system updates or upgrades are applied, improving stability and consistency across deployments.

### Benefits

This job enhances the stability and reliability of the Odoo system by ensuring all necessary modules are pre-installed, which prevents conflicts during upgrades.

### Checklist
- [x] Variables are documented in the values.yaml and added to the `README.md` using [readme-generator-for-helm] (https://github.com/bitnami/readme-generator-for-helm)

### DIFF
--- a/bitnami/odoo/templates/addons-install-job.yaml
+++ b/bitnami/odoo/templates/addons-install-job.yaml
@@ -93,7 +93,7 @@ spec:
         volumeMounts:
           - name: odoo-data
             mountPath: /bitnami/odoo
-      restartPolicy: OnFailure
+      restartPolicy: Never
       volumes:
         - name: odoo-data
           {{- if .Values.persistence.enabled }}
--- a/bitnami/odoo/templates/addons-install-job.yaml
+++ b/bitnami/odoo/templates/addons-install-job.yaml
@@ -1,0 +1,105 @@
+{{- /*
+Copyright VMware, Inc.
+SPDX-License-Identifier: APACHE-2.0
+*/}}
+{{- if .Values.upgradeJob.enabled }}
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: {{ include "common.names.fullname" . }}-addons-install
+  namespace: {{ .Release.Namespace | quote }}
+  labels: {{- include "common.labels.standard" (dict "customLabels" .Values.commonLabels "context" $) | nindent 4 }}
+  annotations:
+    "helm.sh/hook": pre-install,pre-upgrade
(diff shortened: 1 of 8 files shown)


bitnami/odoo/values.yaml
```diff
@@ -114,6 +114,9 @@ odooPassword: ""
 ## @param odooSkipInstall Skip Odoo installation wizard
 ##
 odooSkipInstall: false
+## @param odooSkipModulesUpdate Skip Odoo update wizard
+##
+odooSkipModulesUpdate: true
```
jotamartos, line 119:
```suggestion
## @param odooSkipModulesUpdate Skip update of modules during the deployment
##
odooSkipModulesUpdate: no
```

This would change the default behavior. I would not skip this by default

https://github.com/bitnami/containers/blob/main/bitnami/odoo/17/debian-12/rootfs/opt/bitnami/scripts/odoo-env.sh#L96

bitnami/odoo/templates/addons-install-job.yaml
```diff
@@ -0,0 +1,112 @@
...
+    spec:
+      {{- include "odoo.imagePullSecrets" . | nindent 6 }}
+      containers:
+        - name: odoo-addons-install
+          image: {{ template "odoo.image" . }}
+          imagePullPolicy: {{ .Values.image.pullPolicy | quote }}
+          {{- if .Values.upgradeJob.command }}
+          command: 
+            {{- include "common.tplvalues.render" (dict "value" .Values.command "context" $) | nindent 12 }}
+          {{- end }}
```
jotamartos, line 29: If you do not provide a command, it'll basically run another pod with Odoo, will not simply update the plugins.

jotamartos (changes requested):
Thanks for your suggestion but I have some doubts. Could you please check my comments.

bitnami/odoo/values.yaml
jotamartos: Sorry, my fault. In that case, use "false" instead of "no" but the idea is to follow the same approach we have in the container.

jotamartos (commented):
Updated suggestion

PR #25133, opened 2024-04-11
[bitnami/kafka] :lady_beetle: Fix password replace for >10 users

### LINKED ISSUE #25132: Placeholders are incorrectly replaced in KAFKA_CLIENT_PASSWORDS for configuration with >10 users
### Name and Version

bitnami/kafka 28.0.3

None

1. Add more than 10 users to `sasl.client.users`
2. Inspect `/opt/bitnami/kafka/config/server.properties` inside controller kontainer
3. See that the passwords for user at index 10-19 are all set to the password of user at index 1 with the second character of the index as the postfix.
    - For instance, if user[1] has password _PASS_, the password for user[10] will be _PASS0_, the password for user[11] will be _PASS1_, and so on

A list of users with more than 10 elements

The passwords should correspond with the passwords in the secret `kafka-user-passwords` at key `client-passwords`

### What do you see instead?

Given that user[1] has password _PASS_, the password for user[10] is _PASS0_, the password for user[11] is _PASS1_, and so on, instead of their respective passwords

### Additional information

The error seems to be in [bitnami/kafka/templates/scripts-configmap.yaml](https://github.com/bitnami/charts/blob/3614ed59f79b15912f5ec2382588cb6c2453f2de/bitnami/kafka/templates/scripts-configmap.yaml#L308-L310) (line 308 - 310), where `i = 1` will replace:
- `password-placeholder-1` with the right password (eg. `PASS`)
- `password-placeholder-10` with the same password but `0` as a postfix (eg. `PASS0`)
- `password-placeholder-11` with the same password but `1` as a postfix (eg. `PASS1`)
- and so on

nfsvtrifork:
Fixes #25132 

Matching on `"` at the end for the `password-placeholder` ensures that only `password-placeholder-1` is matched and not `password-placeholder-10`, `password-placeholder-11`, etc.

### DIFF
--- a/bitnami/kafka/Chart.yaml
+++ b/bitnami/kafka/Chart.yaml
@@ -16,7 +16,7 @@ annotations:
     - name: os-shell
       image: docker.io/bitnami/os-shell:12-debian-12-r18
 apiVersion: v2
-appVersion: 3.7.1
+appVersion: 3.7.0
 dependencies:
 - condition: zookeeper.enabled
   name: zookeeper
@@ -42,4 +42,4 @@ maintainers:
 name: kafka
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/kafka
-version: 28.0.3
+version: 28.0.4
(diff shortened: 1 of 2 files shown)


dgomezleon (approved):
LGTM

PR #25281, opened 2024-04-19
[bitnami/spring-cloud-dataflow] Remove Skipper ConfigMap and Secrets when Skipper is disabled

miller79:
### Description of the change

This change was made to fix a discovered issue from the values documentation mentioning that `externalDatabase.existingPasswordSecret` and `externalDatabase.existingPasswordKey` is deprecated in favor of using `existingSecret` and `existingSecretPasswordKey` within `dataflow` and `skipper`. When changing to the expected way but having skipper disabled, the skipper `externaldb-secret.yaml` is still included and fails to base64 encode the password since it is empty. The best solution was not to include the file at all. I applied the same logic to skippers `configmap.yaml` since it is not used when enabled.

### Benefits

This change fixes an issue when disabling skipper and mariadb but not including a password for the skipper database. The following is a bare minimum values.yaml that would break before the changes:

```yaml
skipper:
  enabled: false

mariadb:
  enabled: false

externalDatabase:
  hibernateDialect: org.hibernate.dialect.SQLServer2016Dialect
  driver: com.microsoft.sqlserver.jdbc.SQLServerDriver
  dataflow:
    url: 'jdbc:sqlserver://host.docker.internal;databaseName=spring;encrypt=false'
    username: test
    password: test
```

### Possible drawbacks

There should be no drawbacks for this change as it only applies when skipper is disabled.

### Applicable issues

No issue created for this.

### Checklist

### DIFF
--- a/bitnami/spring-cloud-dataflow/Chart.yaml
+++ b/bitnami/spring-cloud-dataflow/Chart.yaml
@@ -53,4 +53,4 @@ maintainers:
 name: spring-cloud-dataflow
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/spring-cloud-dataflow
-version: 28.0.0
+version: 28.0.1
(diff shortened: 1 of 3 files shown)


dgomezleon (approved):
LGTM

PR #25285, opened 2024-04-19
bitnami/keycloak add else when '.Values.enableDefaultInitContainers: false', otherwise all initContainers are disabled

lknite:
### Description of the change

Adds an 'else' to the .Values.enableDefaultInitContainers allowing for initContainers to still be specified, even if the default initContainers are disabled.

### Benefits

It is understood that folks sometimes want to specify initContainers and that is good, so we should still allow initContainers to be specified, even if someone decides to disable the default initContainers (needed to run 3rd party images).

### Possible drawbacks

None come to mind.  It's a pretty small change with big benefit.

### Applicable issues

- fixes #24544

### Additional information

I'm not a pro at helm chart template files, so please check my work.  Suggestions welcome!

### Checklist

### Notes
Been awhile since I commited a change... if something else is needed just let me know.

### DIFF
--- a/bitnami/keycloak/Chart.yaml
+++ b/bitnami/keycloak/Chart.yaml
@@ -33,4 +33,4 @@ maintainers:
 name: keycloak
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/keycloak
-version: 21.0.2
+version: 21.0.3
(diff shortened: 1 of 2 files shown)


lknite:
I'm going to retry this with a single commit.

PR #25219, opened 2024-04-18
[bitnami/airflow] Fix clone of initial repositories

jotamartos:
### Description of the change

The git repository didn't have permissions to edit the /etc/ssh folder so the initialization failed

### Benefits

Initialization works and repo can be cloned 

### Applicable issues

- fixes https://github.com/bitnami/charts/issues/24935

### Checklist

- [X] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/). This is *not necessary* when the changes only affect README.md files.
- [NA] Variables are documented in the values.yaml and added to the `README.md` using [readme-generator-for-helm](https://github.com/bitnami/readme-generator-for-helm)
- [X] Title of the pull request follows this pattern [bitnami/<name_of_the_chart>] Descriptive title
- [X] All commits signed off and in agreement of [Developer Certificate of Origin (DCO)](https://github.com/bitnami/charts/blob/main/CONTRIBUTING.md#sign-your-work)

PR #25287, opened 2024-04-19
[bitnami/nginx] fix git clone initcontainers

dunn:
See https://github.com/bitnami/charts/pull/25219


bitnami/nginx/Chart.yaml
```diff
@@ -34,4 +34,4 @@ maintainers:
 name: nginx
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/nginx
-version: 16.0.4
+version: 16.0.5
```
migruiz4, line 37: Bumping chart version because 16.0.5 has already been released https://github.com/bitnami/charts/blob/35ef5e68076a9e7e43b36c3d8e5d9589946b88b6/bitnami/nginx/Chart.yaml#L37

```suggestion
version: 16.0.6
```

PR #25328, opened 2024-04-23
[bitnami/vault] Enabling Vault run in one single Namepace and with that, add more tha…

matheusmgon:
Small change to able run multiples Bitnami Vault helm chart in one single Cluster.


bitnami/vault/templates/injector/mutating-webhook.yaml
```diff
@@ -40,4 +40,6 @@ webhooks:
           operator: NotIn
           values:
             - injector
+    namespaceSelector:
+{{ tpl (toYaml (((.Values.injector.webhook)).namespaceSelector | default .Values.injector.namespaceSelector)) . | indent 6}}
```
migruiz4, line 44: I wouls like to suggest using the helper `common.tplvalues.render` in combination with 
```suggestion
    namespaceSelector: {{- include "common.tplvalues.render" (dict "value" (default .Values.injector.namespaceSelector ((.Values.injector.webhook)).namespaceSelector) "context" $) | nindent 6 }}

```

migruiz4: Additionally, the value `injector.namespaceSelector` does not exist in the values.yaml.

Is there a reason to have both `injector.webhook.namespaceSelector` and `injector.namespaceSelector`?

migruiz4 (changes requested):
Hi @matheusmgon,

Thank you very much for your contribution! Could you please take a look at my comments?

bitnami/vault/templates/injector/mutating-webhook.yaml
matheusmgon: There are no reason to add `injector.namespaceSelector`, it was a mistake.
Now I add the way like you sent and plus I add a condition to check if this info was sent.
Just to be sure, nothing else.

matheusmgon:
I really don't know why license file is failing 🤔

bitnami/vault/values.yaml
```diff
@@ -1387,6 +1387,18 @@ injector:
...
+  ## Example:
+  ## webhook:
+  ##   namespaceSelector:
+  ##     matchExpressions:
+  ##     - key: kubernetes.io/metadata.name
+  ##       operator: In
+  ##       values:
+  ##         - "{{ .Release.Namespace }}"
+  webhook: {}
+
```
migruiz4, line 1401: Please use the following structure:

```suggestion
  webhook: 
    ## @param injector.webhook Enabling specify which namespace you want to work with webhook
    ## Example:
    ## webhook:
    ##   namespaceSelector:
    ##     matchExpressions:
    ##     - key: kubernetes.io/metadata.name
    ##       operator: In
    ##       values:
    ##         - "{{ .Release.Namespace }}"
    namespaceSelector: {}
```

This way users should not be confused about what `webhook.*` values can be configured or not, and leaves room for new values to be added in the future.

migruiz4 (commented):
Please take a look at the new comments

bitnami/vault/values.yaml
matheusmgon: Someting failed during the creation of the Readme.md
Do you know why @migruiz4 ?

migruiz4 (commented):
Looks like the branch is all messed up, could you please fix it?

I would recommend recreating it or creating a new PR if necessary

bitnami/vault/values.yaml
migruiz4: My bad, I fixed it in the suggestion above.

The README is autogenerated based on the `@param` tag, which was missing `.namespaceSelector`.
It should be `@param injector.webhook.namespaceSelector`

migruiz4:
Thank you for creating #25629, we will continue there

PR #25336, opened 2024-04-23
[bitnami/thanos] Use endpoint group optionally for Store Gateways

### LINKED ISSUE #25317: [bitnami/thanos] Add support for `endpoint-group` in store gateway
### Name and Version

bitnami/thanos 15.1.0

HA Store replicas are references using the `endpoint` rather than the `endpoint-group` flag. This means fanout is used, rather than round-robin.

Highly available store API components should be referenced using the `endpoint-group` flag, instead of `endpoint`. This uses round-robing rather than fan-out to distribute requests.

_No response_

verejoel:
### Description of the change

Introduces an optional flag to use `endpoint-group` when adding store gateway StoreAPI endpoints to the querier component.

### Benefits

This is preferable to using `endpoint`, as a round-robin approach is taken to load-balancing requests to these endpoints, rather than fanout.

### Possible drawbacks

None that I am aware of. We ran Thanos in production for years with this configuration.

### Applicable issues

- fixes #25317

### Checklist


juan131 (approved):
Thanks for this enhancement! LGTM

PR #25397, opened 2024-04-25
[bitnami/mongodb] Simplify and fix externalAccess configuration

### LINKED ISSUE #16719: [bitnami/mongodb] Enabling TLS using a custom certificate runs into multiple issues. - user creation / current_primary detection / all probes
### Name and Version

bitnami/mongodb 13.9.4

amd64

1. Create a TLS certificate with a single wildcard domain. `*.mongodb.example.com`
2. Enable TLS in the chart with `requireTLS`.
3. 
 a) The root users are unable to be created during the startup
 b) New nodes are unable to register with the primary host 
 c) The probes fail to run commands
 d) The metrics server is unable to connect to `mongod`

> *The root cause of this issue is being restricted by a TLS Certificate which can only have a single domain.*

The certificate cannot have `127.0.0.1`, `localhost` or any of the k8s domains on there. 

```yaml
tls: 
  enabled: true 
  standalone/replicaset/hidden/arbiter:
  existingSecret(s): [ "mongodb-example-com-cert" ]
externalAccess: 
  enabled: true
  autoDiscovery: 
    enabled: true
  annotations:
      external-dns.alpha.kubernetes.io/hostname: "{{ .targetPod }}.mongodb.example.com"
  hidden:
    enabled: true
    service:
      annotations:
        external-dns.alpha.kubernetes.io/hostname: "{{ .targetPod }}.mongodb.example.com"
```

SIDE NOTE: We have enabled autodiscovery to mount the `/shared` volume and populate `info.txt`. 
Our in use DNS address does not match the value returned by the autoDiscovery script since we are using the external-dns won AWS.
Our FQDN is `releasename-0.mongodb.example.com` and autoDiscovery returns the AWS load balancer address. 
If we switch auto-discovery off then the `MONGODB_ADVERTISED_HOSTNAME` reverts to the loadbalancer IPs which we do not want.

---

The mongodb database initializes correctly:
- root user and additional users are created 
- replicaset config is configured and nodes join existing replica sets.

The probes are able to connect to mongod using tls
The metrics server connects to mongod using tls

### What do you see instead?

- All commands of `mongosh` time out since the TLS requirements are not met. 
- The script continues, configuring mongo.conf and restarting the mongod server.
- Mongod is then running but without any users or a replicaset configured. 

### Additional information

We have implemented significant workarounds to get the chart working. I think these are good candidates for being included in the Chart (and additionally the Container) 

Below I will describe the various points where the restrictions 

---

### Description

The `MONGODB_ADVERTISED_HOSTNAME` envar is populated through `autoDiscovery.sh` and takes a default through the container envvars. It is used as the hostname when configuring the replica set and also used to compare against the hostname returned from an existing replica set to verify if the current node is a primary. 

The `MONGODB_INITIAL_PRIMARY_HOST` env var is used as the host when configuring a new secondary node onto an existing replicaset.

The `current_primary` variable is used in the `setup.sh` script and populated through a mongosh function. The chart only uses the hostnames for the headless service. 

The environment variables can be overwritten through the chart `extraEnvVars:` but this is not sufficient. 
`MONGODB_ADVERTISED_HOSTNAME` is overwritten in the `setup.sh` script, either through the autoDiscover.sh initi container or using other details for the load balancer. 

### Issues 

All three of these variables cause issues with our TLS set-up. 

The current_primary node is unable to be found. The replicaset configuration is unable to be set, and even if it was able to connect a hostname would be added which would also fail TLS. 

### Workaround 

First we override the environment variables with our domain name: 
```
extraEnvVars:
      - name: MONGODB_INITIAL_PRIMARY_HOST
        value: {{ printf "%s-0.%s" (include "mongodb.fullname" .) "mongodb.example.com" }}

      - name: MONGODB_ADVERTISED_HOSTNAME
        value: $(MY_POD_NAME).mongodb.example.com
```

This sets the defaults to use our Valid Domain name and will ensure that the replicaset config can be done. 

We alter the `args:` of the container to run a preparatory script. This script runs first to perform some small changes, it also *change the `/scripts/setup.sh` file,  before running the altered script*

The preparatory script first inserts our Valid FQDN for that pod into the response from the initContainer.
```
    echo -n "$MONGODB_ADVERTISED_HOSTNAME" > /shared/info.txt
```

Next it takes `/scripts/setip.sh` and replaces the connection string for the `current_primary` variable with our preferred host list. 
NOTE: This requires us to mount an additional emptyDir volume since we cannot alter the script in place

```
    {{- $portNumber := int .Values.service.ports.mongodb }}
    {{- $fullname := include "mongodb.fullname" . }}
    {{- $releaseNamespace := include "mongodb.namespace" . }}
    {{- $clusterDomain := .Values.mongodb.clusterDomain }}
    {{- $loadBalancerIPListLength := len .Values.externalAccess.service.loadBalancerIPs }}
    {{- $mongoList := list }}
    {{- $mongoListTLS := list }} <----------  New line added to construct connection string to tls addresses.
    {{- range $e, $i := until $replicaCount }}
    {{- $mongoList = append $mongoList (printf "%s-%d.%s-headless.%s.svc.%s:%d" $fullname $i $fullname $releaseNamespace $clusterDomain $portNumber) }}
    {{- $mongoListTLS = append $mongoListTLS (printf "%s-%d.%s:%d" $fullname $i "mongodb.example.com" $portNumber) }} #### <---------- New line added to construct connection string to tls addresses
    {{- end }}

    CONNECTION_STRING_OLD="\"{{ join "," $mongoList }}\""
    CONNECTION_STRING_NEW="\"{{ join "," $mongoListTLS }}\""

    CONNECTION_STRING_NEW="$CONNECTION_STRING_NEW --tls --tlsCertificateKeyFile=/certs/mongodb.pem --tlsCAFile=/certs/mongodb-ca-cert"

    info "Old Connection String: $CONNECTION_STRING_OLD"
    info "New Connection String: $CONNECTION_STRING_NEW"

    for sourceFile in "/scripts/setup.sh" "/scripts/setup-hidden.sh"; do
      if [[ -f "$sourceFile" ]]; then
        destFile="/alteredScripts/$(basename -- "$sourceFile")"
        info "Amending $sourceFile and moving to $destFile..."

        ESCAPED_CONNECTION_STRING_OLD=$(printf '%s\n' "$CONNECTION_STRING_OLD" | sed -e 's/[]\/$*.^[]/\\&/g');
        ESCAPED_CONNECTION_STRING_NEW=$(printf '%s\n' "$CONNECTION_STRING_NEW" | sed -e 's/[\/&]/\\&/g')

        sed "s/$ESCAPED_CONNECTION_STRING_OLD/$ESCAPED_CONNECTION_STRING_NEW/g" $sourceFile > $destFile

        chmod 0755 "$destFile"
      fi
    done
```

We can now run `/alteredScripts/setup.sh` or `/alteredScripts/setup-hidden.sh` as required.

We have alsorequired a wait to ensure the hostnames we want to reach are ready.
```
    for mongodb_host in "${MONGODB_INITIAL_PRIMARY_HOST}" "${MONGODB_ADVERTISED_HOSTNAME}"; do
      info "Testing host: $mongodb_host..."
      for i in {1..180}; do
        if getent ahosts "$mongodb_host"; then
          break
        fi
        sleep 1
        debug "Waiting for $mongodb_host to be ready... $i"
        if [ "$i" == "180" ]; then
          warn "Unable to connect to host: $mongodb_host"
          exit 1
        fi
      done
    done
```
This waits for 3 minutes, if It still fails the pod stops and restarts.
We have increased the startupProdeDelay by 2 minutes to avoid a CrashLoopBackoff from slowing down the startup. 
(Perhaps a better balance can be found between the wait in the pod and the risk of triggering a long `crashLoopBackoff`?)  

### BONUS FEATURE - Attach to existing replicaSet 

In our prep script we have also added an option to completely overwrite the connection string for current_primary using a hardcoded string from the values.

```
    {{- if .Values.global.existingReplicaSetConnectionString }}
    CONNECTION_STRING_NEW="{{ .Values.global.existingReplicaSetConnectionString }}\""
    {{- end }}
```

Ensuring that the replicaset name, key and connection string match an existing replicaset it will deploy all nodes as secondaries for that replica set. 

We are using this for migration purposes. 
Once the data has been synced we can move the primary to the cluster, update our connection strings, remove the additional value and destroy the old nodes.

### Suggested Chart Changes

To implement similar in the chart some / all of these changes could be made:

- Provide a method to manually set MONGODB_ADVERTISED_HOSTNAME even when autoDiscovery is disabled 
  - OR
  - provide a way to customise autoDiscovery so that it can take/output our custom domain. This initContainer could also verify the DNS address is reachable. 
- Provide a method to change the structure of the hosts generated in the host list for populating current_primary. This way we can guarantee a TLS connection. 
- Bonus: Provide a way to override the host list when populating current_primary. This will enable deploying all nodes as part of an existing replica set. 

The environment variable override is working well, some way of tieing these changes together and avoiding bad combinations would be good to find.

### Description

The `libmongodb.sh` script runs commands against localhost (`127.0.0.1`) during initialization. These operations createall the users and importantly configure the connection to a replicaset. 

### Issues 

All these commands fail through TLS unless `127.0.0.1` is part of the certificate. (Uncommon for external certificates)

### Workaround 

During setup the mongod server is bound to the localhost IP. 

There are two options with this work around: 
1. Provide a way of connecting to `127.0.0.1` using a domain in the certificate
2. Use the external domain and alter the `mongodb.conf` to bind all ips _before_ the first time `mongod` is started allowing connections from outside the pod.

We preferred option 1. 

#### Workaround 1

First add a `localhost.mongodb.example.com` domain to our certificate and route that domain to `127.0.0.1` 
```
hostAliases:
- ip: 127.0.0.1
  hostnames:
  - "localhost.mongodb.example.com"
```

We add another step to the prepatory script described above to alters the `libmongodb.sh` file. 

```
sed -i "s/127.0.0.1/localhost.mongodb.example.com/g" /opt/bitnami/scripts/libmongodb.sh
```

This file can be altered in place and then you can continue as normal!

> NOTE: I have been unable to find a way to add a pod specific domain to the hosts file. If this is possible we could remove the need for a `localhost.*` domain. 

#### Workaround 2: 

Similar to the workaround 1 we replace the `127.0.0.1` in libmongo, this time with the advertised hostname. 

Since this will resolve to an external IP we also need to enable bindAll Ips before mongod start. 

Something like: 
```
sed -i "s/127.0.0.1/${MONGODB_ADVERTISED_HOSTNAME}/g" /opt/bitnami/scripts/libmongodb.sh


# Load libraries
. /opt/bitnami/scripts/libmongodb.sh
# Load environment
. /opt/bitnami/scripts/mongodb-env.sh
mongodb_set_listen_all_conf "$MONGODB_CONF_FILE"
```

There is a risk that a bad actor will come along and create a root user before you can since the server is now exposed, but it is minimal. 

### Suggested Chart Changes 

The fix for this will require a change to the scripts in the container. 

- Use an environment variable to be used in place of `127.0.0.1` in the libmongo.sh script. (`MONGODB_LOCALHOST_NAME`?)
- The chart changes to provide this environment variable. 

It could also be done to ensure that this host only resolves to 127.0.0.1, or use the chart to set it as a default hostalias.

### Description

The probes connect with the TLS options, but do not specify a hostname which defaults to `127.0.0.1`. 

### Workaround 

This issue has forced me to provide custom probes which are only altered to provide the hostname of `MONGODB_ADVERTISED_HOSTNAME`

We could also use `MONGODB_LOCALHOST_NAME` / `localhost.mongodb.example.com` if it is available through the TLS certificate and configured in the `hostAliases`. 

This has the benefit of routing traffic internally, so it is directly checking on the pod health and not implicitly checking network connectivity.

### Description

The metrics sidecar by default uses the `localhost` as the host in the connection string. 

The metrics sidecar is unable to connect to mongod as it fails the tls check

### Workaround

I have manually defined the args for the side car to use localhost.mongodb.example.com. 

### Suggested Chart Changes 

Having access to `MONGODB_ADVERTISED_HOSTNAME`, `/shared/info.txt` or `MONGODB_LOCALHOST_NAME` would make a more DRY approach to correcting this connection string. 

(There are other issues with metrics, but that is outside the scope of this issue)

fmulero:
### Description of the change

Allow the use of external certificates signed for external names. Current chart requires several internal names and IPs signed in the certificates provided by the user. That restriction makes almost impossible the use of this chart with external names and TLS enabled.

### Benefits

Allow the use of external certificates and names in MongoDB chart.

### Possible drawbacks

None

### Applicable issues

- fixes #16719

### Additional information

These changes were tested in the following scenario:
1. Empty cluster created with k3d:
```console
$ k3d cluster create
```
2. Certificates and secrets have been created with following script:
```bash
#!/bin/bash
if [[ ! -f  ca.crt ]]; then
  openssl genrsa -out ca.key 2048
  openssl req -x509 -new -nodes -key ca.key -subj "/CN=example.com" -sha256 -days 1024 -out ca.crt
fi

for i in 0 1 arbiter; do
  if [[ ! -f "mongodb-${i}.crt" ]]; then
    openssl genrsa -out "mongodb-${i}.key" 2048
    openssl req -new -key "mongodb-${i}.key" -subj "/CN=mongodb-${i}.example.com" -out "mongodb-${i}.csr"
    openssl x509 -req -in "mongodb-${i}.csr" -CA ca.crt -CAkey ca.key -CAcreateserial -out "mongodb-${i}.crt" -days 1024 -sha256
  fi
  kubectl create secret tls "mongodb-${i}-pem"  --cert="mongodb-${i}.crt" --key="mongodb-${i}.key"
  kubectl patch secret "mongodb-${i}-pem" -p="{\"data\":{\"ca.crt\": \"$(cat ca.crt | base64 -w0 )\"}}"
done
```
3. To avoid  issues with name resolution this configuration has been added to the `coredns` configmap:
```
rewrite name mongodb-0.example.com mongodb-0.mongodb-headless.default.svc.cluster.local
rewrite name mongodb-1.example.com mongodb-1.mongodb-headless.default.svc.cluster.local
```
4. Values used:
```yaml
architecture: replicaset
replicaCount: 2
tls:
  enabled: true
  autoGenerated: false
  replicaset:
    existingSecrets:
      - mongodb-0-cert
      - mongodb-1-cert
  arbiter:
    existingSecret: mongodb-arbiter-cert
externalAccess:
  enabled: true
  service:
    loadBalancerIPs:
      - 10.10.0.3
      - 10.10.0.4
    annotationsList:
      - external-dns.alpha.kubernetes.io/hostname: mongodb-0.example.com
      - external-dns.alpha.kubernetes.io/hostname: mongodb-1.example.com
  publicNames:
    - mongodb-0.example.com
    - mongodb-1.example.com
```

### Checklist

### DIFF
--- a/bitnami/mongodb/Chart.yaml
+++ b/bitnami/mongodb/Chart.yaml
@@ -39,4 +39,4 @@ maintainers:
 name: mongodb
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/mongodb
-version: 15.1.5
+version: 15.2.0
--- a/bitnami/mongodb/templates/backup/cronjob.yaml
+++ b/bitnami/mongodb/templates/backup/cronjob.yaml
@@ -103,8 +103,8 @@ spec:
                 {{- if .Values.externalAccess.service.loadBalancerIPs }}
                 - -i {{ join "," .Values.externalAccess.service.loadBalancerIPs }}
                 {{- end }}
-                {{- if .Values.tls.extraDnsNames }}
-                - -n {{ join "," .Values.tls.extraDnsNames }}
+                {{- if or .Values.tls.extraDnsNames .Values.externalAccess.publicNames }}
+                - -n {{ join "," ( concat .Values.tls.extraDnsNames .Values.externalAccess.publicNames ) }}
                 {{- end }}
               {{- if .Values.tls.resources }}
               resources: {{- toYaml .Values.tls.resources | nindent 16 }}
--- a/bitnami/mongodb/templates/hidden/statefulset.yaml
+++ b/bitnami/mongodb/templates/hidden/statefulset.yaml
@@ -158,8 +158,8 @@ spec:
             {{- if .Values.externalAccess.hidden.service.loadBalancerIPs }}
             - -i {{ join "," .Values.externalAccess.hidden.service.loadBalancerIPs }}
             {{- end }}
-            {{- if .Values.tls.extraDnsNames }}
-            - -n {{ join "," .Values.tls.extraDnsNames }}
+            {{- if or .Values.tls.extraDnsNames .Values.externalAccess.publicNames }}
+            - -n {{ join "," ( concat  .Values.tls.extraDnsNames .Values.externalAccess.publicNames ) }}
             {{- end }}
           {{- if .Values.tls.resources }}
           resources: {{- toYaml .Values.tls.resources | nindent 12 }}
--- a/bitnami/mongodb/templates/replicaset/external-access-svc.yaml
+++ b/bitnami/mongodb/templates/replicaset/external-access-svc.yaml
@@ -19,8 +19,9 @@ metadata:
   labels: {{- include "common.labels.standard" ( dict "customLabels" $root.Values.commonLabels "context" $ ) | nindent 4 }}
     app.kubernetes.io/component: mongodb
     pod: {{ $targetPod }}
-  {{- if or $root.Values.externalAccess.service.annotations $root.Values.commonAnnotations }}
-  {{- $annotations := include "common.tplvalues.merge" ( dict "values" ( list $root.Values.externalAccess.service.annotations $root.Values.commonAnnotations ) "context" $ ) }}
+  {{- if or $root.Values.externalAccess.service.annotations $root.Values.commonAnnotations $root.Values.externalAccess.service.annotationsList}}
+  {{- $exclusiveAnnotations := ternary ( dict ) (index $root.Values.externalAccess.service.annotationsList $i) ( lt (len $root.Values.externalAccess.service.annotationsList ) $i ) }}
+  {{- $annotations := include "common.tplvalues.merge" ( dict "values" ( list $root.Values.externalAccess.service.annotations $root.Values.commonAnnotations $exclusiveAnnotations ) "context" $ ) }}
   annotations: {{- include "common.tplvalues.render" ( dict "value" $annotations "context" $) | nindent 4 }}
   {{- end }}
 spec:
--- a/bitnami/mongodb/templates/replicaset/scripts-configmap.yaml
+++ b/bitnami/mongodb/templates/replicaset/scripts-configmap.yaml
@@ -73,7 +73,7 @@ data:
     export MONGODB_ADVERTISED_HOSTNAME="$(<${SHARED_FILE})"
     {{- else }}
     ID="${MY_POD_NAME#"{{ $fullname }}-"}"
-    export MONGODB_ADVERTISED_HOSTNAME=$(echo '{{ .Values.externalAccess.service.loadBalancerIPs }}' | tr -d '[]' | cut -d ' ' -f "$(($ID + 1))")
+    export MONGODB_ADVERTISED_HOSTNAME=$(echo '{{ coalesce .Values.externalAccess.publicNames .Values.externalAccess.service.loadBalancerIPs }}' | tr -d '[]' | cut -d ' ' -f "$(($ID + 1))")
     {{- end }}
     {{- else if eq .Values.externalAccess.service.type "NodePort" }}
     ID="${MY_POD_NAME#"{{ $fullname }}-"}"
@@ -113,7 +113,6 @@ data:
       {{- $fullname := include "mongodb.fullname" . }}
       {{- $releaseNamespace := include "mongodb.namespace" . }}
       {{- $clusterDomain := .Values.clusterDomain }}
-      {{- $loadBalancerIPListLength := len .Values.externalAccess.service.loadBalancerIPs }}
       {{- $mongoList := list }}
       {{- range $e, $i := until $replicaCount }}
       {{- $mongoList = append $mongoList (printf "%s-%d.%s-headless.%s.svc.%s:%d" $fullname $i $fullname $releaseNamespace $clusterDomain $portNumber) }}
--- a/bitnami/mongodb/templates/replicaset/statefulset.yaml
+++ b/bitnami/mongodb/templates/replicaset/statefulset.yaml
@@ -158,8 +158,8 @@ spec:
             {{- if .Values.externalAccess.service.loadBalancerIPs }}
             - -i {{ join "," .Values.externalAccess.service.loadBalancerIPs }}
             {{- end }}
-            {{- if .Values.tls.extraDnsNames }}
-            - -n {{ join "," .Values.tls.extraDnsNames }}
+            {{- if or .Values.tls.extraDnsNames .Values.externalAccess.publicNames }}
+            - -n {{ join "," ( concat .Values.tls.extraDnsNames .Values.externalAccess.publicNames ) }}
             {{- end }}
           {{- if .Values.tls.resources }}
           resources: {{- toYaml .Values.tls.resources | nindent 12 }}
@@ -243,7 +243,7 @@ spec:
             - name: K8S_SERVICE_NAME
               value: "{{ include "mongodb.service.nameOverride" . }}"
             - name: MONGODB_INITIAL_PRIMARY_HOST
-              value: {{ printf "%s-0.$(K8S_SERVICE_NAME).$(MY_POD_NAMESPACE).svc.%s" (include "mongodb.fullname" .) .Values.clusterDomain }}
+              value: {{ ternary ( printf "%s-0.$(K8S_SERVICE_NAME).$(MY_POD_NAMESPACE).svc.%s" (include "mongodb.fullname" .) .Values.clusterDomain ) ( first .Values.externalAccess.publicNames ) ( empty .Values.externalAccess.publicNames ) }}
             - name: MONGODB_REPLICA_SET_NAME
               value: {{ .Values.replicaSetName | quote }}
             {{- if and .Values.replicaSetHostnames (not .Values.externalAccess.enabled) }}
--- a/bitnami/mongodb/templates/standalone/dep-sts.yaml
+++ b/bitnami/mongodb/templates/standalone/dep-sts.yaml
@@ -152,8 +152,8 @@ spec:
             {{- if .Values.externalAccess.service.loadBalancerIPs }}
             - -i {{ join "," .Values.externalAccess.service.loadBalancerIPs }}
             {{- end }}
-            {{- if .Values.tls.extraDnsNames }}
-            - -n {{ join "," .Values.tls.extraDnsNames }}
+            {{- if or .Values.tls.extraDnsNames .Values.externalAccess.publicNames }}
+            - -n {{ join "," ( concat .Values.tls.extraDnsNames .Values.externalAccess.publicNames ) }}
             {{- end }}
           {{- if .Values.tls.resources }}
           resources: {{- toYaml .Values.tls.resources | nindent 12 }}
--- a/bitnami/mongodb/values.yaml
+++ b/bitnami/mongodb/values.yaml
@@ -129,7 +129,7 @@ diagnosticMode:
 image:
   registry: docker.io
   repository: bitnami/mongodb
-  tag: 7.0.8-debian-12-r2
+  tag: 7.0.8-debian-12-r4
   digest: ""
   ## Specify a imagePullPolicy
   ## ref: https://kubernetes.io/docs/concepts/containers/images/#pre-pulled-images
@@ -823,6 +823,9 @@ externalAccess:
   ## @param externalAccess.enabled Enable Kubernetes external cluster access to MongoDB(&reg;) nodes (only for replicaset architecture)
   ##
   enabled: false
+  ## @param externalAccess.publicNames Array of public names. The size should be equal to the number of replicas.
+  ## 
+  publicNames: []
   ## External IPs auto-discovery configuration
   ## An init container is used to auto-detect LB IPs or node ports by querying the K8s API
   ## Note: RBAC might be required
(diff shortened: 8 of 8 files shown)


rrileyca:
Hey @fmulero thanks for looking into this. 

Does this work? One of the issues is that Mongo will see the external DNS name in the Replica config (which can be seen in `mongosh` with `rs.config()`) and that won't match its hostname, so it will think it's not in the cluster. Have you tested this?

fmulero:
> Hey @fmulero thanks for looking into this.
> 
> Does this work? One of the issues is that Mongo will see the external DNS name in the Replica config (which can be seen in `mongosh` with `rs.config()`) and that won't match its hostname, so it will think it's not in the cluster. Have you tested this?

At the moment I didn't face that issue, maybe because I am changing my coredns config to rewrite `mongodb-0.example.com` to `mongodb-0.mongodb-headless.default.svc.cluster.local`

fmulero:
Hi @rrileyca, I think I reproduced the problem you mentioned. To fix it I configured the cluster using the external names and I added an init container to ensure that we can resolve the external name before init the cluster.

rrileyca:
Hey @fmulero, thanks this is defnitely a useful feature. 

Does it work without the coreDNS edits in the kubernetes cluster though? I think the external DNS name has to match the hostname of the Pod, and failing a hostname match it does a DNS lookup and tries to match the IP's of the Replicas to its own IP. I don't see any changes to the hostname of the Pod, just the services.

The reason changing the coreDNS config works (I believe) is because you can put the public DNS names like `mongo-1.mydns.com` in the Mongo ReplicaSet config, which coreDNS then translates to the service name like `mongodb.svc.cluster.local` and resolves the Pod IP. Mongo sees that IP address matches its local interface address and then assumes it is part of the cluster. 

I think the problem here is that when you resolve the public DNS name of a node, it will resolve to the Loadbalancer IP address which is not the Pod IP. Mongo will not think it is part of the cluster in this case. 

Does your PR assume that you have a DNS name that always translates to the Pod IP? If so, how do you keep this DNS entry up to date? I've only used external-dns for Loadbalancers and Ingresses.

fmulero:
HI @rrileyca thanks for taking care of this. I've tested it following these steps:
1. Create the certificates with the script described in the description.
2. Run a regular installation with the values below:
```yaml
architecture: replicaset
replicaCount: 2
tls:
  enabled: true
  autoGenerated: false
  replicaset:
    existingSecrets:
      - mongodb-0-cert
      - mongodb-1-cert
  arbiter:
    existingSecret: mongodb-arbiter-cert
externalAccess:
  enabled: true
  service:
    type: LoadBalancer
    publicNames:
      - 'mongodb-0.example.com'
      - 'mongodb-1.example.com'
```
That will create 2 external services and the pods will wait for the `mongodb-0.example.com` name resolution.

3. Once the external services have their external ip addresses we can upgrade the deployment adding `hostAliases` to resolve the external names:
```yaml
architecture: replicaset
replicaCount: 2
tls:
  enabled: true
  autoGenerated: false
  replicaset:
    existingSecrets:
      - mongodb-0-cert
      - mongodb-1-cert
  arbiter:
    existingSecret: mongodb-arbiter-cert
externalAccess:
  enabled: true
  service:
    type: LoadBalancer
    publicNames:
      - 'mongodb-0.example.com'
      - 'mongodb-1.example.com'
hostAliases: &aliases
  - ip: 35.243.236.68
    hostnames:
      - mongodb-0.example.com
  - ip: 34.75.99.90
    hostnames:
      - mongodb-1.example.com
arbiter:
  hostAliases: *aliases
```
4. At this point our pods are stuck waiting for name resolution. In this case, we have changed the StatefulSet adding the hostAliases, we have to remove old pods.
```console
$ kubectl delete pods mongodb-0 mongodb-arbiter-0
pod "mongodb-0" deleted
pod "mongodb-arbiter-0" deleted
```
5. New pods start running and get ready. Now if we run the following commands we can see how mongodb-0 is primary and mongodb-1 is secondary:
```console
$ kubectl exec -it mongodb-0 -c mongodb -- mongosh --tls --tlsCertificateKeyFile=/certs/mongodb.pem --tlsCAFile=/certs/mongodb-ca-cert --tlsAllowInvalidHostnames --port $MONGODB_PORT_NUMBER  --eval "db.hello().isWritablePrimary"
true
$ kubectl exec -it mongodb-0 -c mongodb -- mongosh --tls --tlsCertificateKeyFile=/certs/mongodb.pem --tlsCAFile=/certs/mongodb-ca-cert --tlsAllowInvalidHostnames --port $MONGODB_PORT_NUMBER  --eval "db.hello().secondary"
false
$ kubectl exec -it mongodb-1 -c mongodb -- mongosh --tls --tlsCertificateKeyFile=/certs/mongodb.pem --tlsCAFile=/certs/mongodb-ca-cert --tlsAllowInvalidHostnames --port $MONGODB_PORT_NUMBER  --eval "db.hello().isWritablePrimary"
false
$ kubectl exec -it mongodb-1 -c mongodb -- mongosh --tls --tlsCertificateKeyFile=/certs/mongodb.pem --tlsCAFile=/certs/mongodb-ca-cert --tlsAllowInvalidHostnames --port $MONGODB_PORT_NUMBER  --eval "db.hello().secondary"
true
```

From my understanding everything looks good.

rafariossaa (approved):
LGTM.

rrileyca:
thanks @fmulero. Should those instructions regarding the hostAliases be added to the README? As I understand it, there is no way to deploy using public names in a single `helm install`. You need to deploy it, let the IPs be allocated, and then add the `hostAliases` and run a `helm update` afterward.

fmulero:
I've just added a note about `hostAliases`. From my understanding that is useful in testing environments. In the real world, developers or SREs should be able to update their DNS records, using external-dns (the chart allows it) or manually configuring the DNS server records. In that case the pods will wait for name resolution and then they'll continue with the startup process.

PR #25430, opened 2024-04-26
[bitnami/node-exporter] Update PSP annotation to dynamic value

### LINKED ISSUE #21510: Implementing containerSecurityContext.seccompProfile breaks compatibility with PodSecurityPolicy
### Name and Version

bitnami/node-exporter 3.9.0

amd64

On a k8s v1.23 cluster with PodSecurityPolicy enabled, install node-exporter 3.9.0 or newer (included in kube-prometheus versions after 8.21.6)

`helm -n kube-prometheus --version=8.22.2 --repo=https://raw.githubusercontent.com/bitnami/charts/archive-full-index/bitnami  install kube-prometheus kube-prometheus`

No

The node-exporter daemonset should run abiding by the built-in kube-prometheus-node-exporter-kube-prometheus PSP that gets created and set up.

```
$ kubectl -n kube-prometheus get daemonsets.apps kube-prometheus-node-exporter
NAME                            DESIRED   CURRENT   READY   UP-TO-DATE   AVAILABLE   NODE SELECTOR   AGE
kube-prometheus-node-exporter   3         3         3       3            3           <none>          2m31s
```

### What do you see instead?

The pods fail to run due to PSP violations:
```
54s         Warning   FailedCreate             daemonset/kube-prometheus-node-exporter                    Error creating: pods "kube-prometheus-node-exporter-" is forbidden: PodSecurityPolicy: unable to admit pod: [spec.securityContext.hostNetwork: Invalid value: true: Host network is not allowed to be used spec.securityContext.hostPID: Invalid value: true: Host PID is not allowed to be used spec.volumes[0]: Invalid value: "hostPath": hostPath volumes are not allowed to be used spec.volumes[1]: Invalid value: "hostPath": hostPath volumes are not allowed to be used spec.containers[0].hostPort: Invalid value: 9100: Host port 9100 is not allowed to be used. Allowed ports: [] pod.metadata.annotations[container.seccomp.security.alpha.kubernetes.io/node-exporter]: Forbidden: seccomp may not be set]
```

### Additional information

Commits such as https://github.com/bitnami/charts/pull/20514 for https://github.com/bitnami/charts/issues/20000 break compatibility with PodSecurityPolicy.
This is because AFAICT  "RuntimeDefault" is not a valid seccomp profile according to the PodSecurityPolicy API. It may only be valid for PSA, not PSP.  If I try to set the PSP annotation  `seccomp.security.alpha.kubernetes.io/allowedProfileNames: 'RuntimeDefault'`  (which is how seccomp is controlled in PSP), it fails:
```
# podsecuritypolicies.policy "kube-prometheus-node-exporter-kube-prometheus" was not valid:
# * metadata.annotations[seccomp.security.alpha.kubernetes.io/allowedProfileNames]: Invalid value: "RuntimeDefault": must be a valid seccomp profile
```

Instead, the default seccomp profile as understood by PSP is called 'runtime/default'.

So then I tried setting the PSP annotation to allow the 'runtime/default' seccomp profile, which is accepted as valid.
However if I try to modify the seccomp profile used by the node-exporter daemonset to also use 'runtime/default', it is rejected:
```
DaemonSet.apps \"kube-prometheus-node-exporter\" is invalid: spec.template.spec.containers[0].securityContext.seccompProfile.type: Unsupported value: \"runtime/default\": supported values: \"Localhost\", \"RuntimeDefault\", \"Unconfined\
```

So there is no valid value of seccomp profile that is accepted by PSP and also matches the daemonset.
The only way to avoid this is to not have any seccomp profile set in the containerSecurityContext of the node-exporter daemonset at all. However, due to longstanding Helm bugs (https://github.com/helm/helm/issues/9027 , https://github.com/helm/helm/issues/12488 ) it is not possible to remove the value by setting `containerSecurityContext.seccompProfile.type` to null.

That being the case there is no way to run versions of node-exporter newer than 3.8.5 on a cluster using PodSecurityPolicy (at least with k8s 1.23).
I understand PSP is deprecated but it will not be removed from kubernetes until v1.25 and some people may still be using it.

sbathgate:
### Description of the change

This change introduces a specific dynamic annotation in the PodSecurityPolicy for the node-exporter Helm chart, specifically to resolve an issue with the seccomp profile settings. By allowing these values to be configurable:

```
seccomp.security.alpha.kubernetes.io/defaultProfileName: runtime/default
seccomp.security.alpha.kubernetes.io/allowedProfileNames: runtime/default
```
we enable users to specify the required seccomp profiles, thus enhancing the chart's adaptability to different security requirements and Kubernetes cluster configurations.

### Benefits

- Increased Compatibility: The update ensures that the node-exporter can be deployed in environments with strict PodSecurityPolicies that enforce seccomp profiles, thereby expanding the chart's usability.
- Security Compliance: Users can align the seccomp profiles with their organization's security policies, ensuring compliance without needing to manually edit or override the deployed policies.
- Flexibility: By making the seccomp profile annotations configurable, the Helm chart can be easily adjusted to meet future changes in Kubernetes security practices or updates in seccomp.

### Possible drawbacks

- Complexity: Introducing additional configuration options can increase the complexity of the chart, potentially making it more challenging for new users to configure.

### Applicable issues

- fixes #21510

### Additional information

To mitigate conflicts and maintain compatibility with existing deployments, the default configuration will retain an empty annotation. The necessity for this change is highlighted by the issue described in #21510. The proposed example annotation is a crucial adjustment to prevent the PodSecurityPolicy from blocking the Daemonset from spawning new pods. This adjustment was tested in my local cluster and successfully resolved the specific PSP configuration issue encountered. This ensures that the chart remains flexible and adaptable to diverse security environments without introducing disruptions to current users.

### Checklist

### DIFF
--- a/bitnami/node-exporter/Chart.yaml
+++ b/bitnami/node-exporter/Chart.yaml
@@ -28,4 +28,4 @@ maintainers:
 name: node-exporter
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/node-exporter
-version: 4.1.2
+version: 4.2.0
--- a/bitnami/node-exporter/Chart.yaml
+++ b/bitnami/node-exporter/Chart.yaml
@@ -28,4 +28,4 @@ maintainers:
 name: node-exporter
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/node-exporter
-version: 4.1.1
+version: 4.1.2
--- a/bitnami/node-exporter/README.md
+++ b/bitnami/node-exporter/README.md
@@ -212,6 +212,7 @@ Installing the Node Exporter chart in `Deployment` mode will overwrite `hostNetw
 | `serviceMonitor.honorLabels`                        | honorLabels chooses the metric's labels on collisions with target labels                                                                                                                                          | `false`                         |
 | `serviceMonitor.attachMetadata`                     | Attaches node metadata to discovered targets                                                                                                                                                                      | `{}`                            |
 | `serviceMonitor.sampleLimit`                        | Per-scrape limit on number of scraped samples that will be accepted.                                                                                                                                              | `""`                            |
+| `podSecurityPolicy.annotations`                     | Annotations for Pod Security Policy. Evaluated as a template.                                                                                                                                                     | `{}`                            |
 
 Specify each parameter using the `--set key=value[,key=value]` argument to `helm install`. For example the following command sets the `minReadySeconds` of the Node Exporter Pods to `120` seconds.
 
--- a/bitnami/node-exporter/templates/psp.yaml
+++ b/bitnami/node-exporter/templates/psp.yaml
@@ -10,8 +10,9 @@ metadata:
   name: {{ template "common.names.fullname.namespace" . }}
   namespace: {{ include "common.names.namespace" . | quote }}
   labels: {{- include "common.labels.standard" ( dict "customLabels" .Values.commonLabels "context" $ ) | nindent 4 }}
-  {{- if .Values.commonAnnotations }}
-  annotations: {{- include "common.tplvalues.render" ( dict "value" .Values.commonAnnotations "context" $ ) | nindent 4 }}
+  {{- if or .Values.podSecurityPolicy.annotations .Values.commonAnnotations }}
+  {{- $annotations := include "common.tplvalues.merge" ( dict "values" ( list .Values.podSecurityPolicy.annotations .Values.commonAnnotations ) "context" . ) }}
+  annotations: {{- include "common.tplvalues.render" ( dict "value" $annotations "context" $) | nindent 4 }}
   {{- end }}
 spec:
   privileged: false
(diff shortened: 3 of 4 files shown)


bitnami/node-exporter/Chart.yaml
```diff
@@ -28,4 +28,4 @@ maintainers:
 name: node-exporter
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/node-exporter
-version: 4.1.1
+version: 4.1.2
```
migruiz4, line 31: Because this PR adds a new feature, version should be bumped a minor instead of a patch:
```suggestion
version: 4.2.0
```

migruiz4 (changes requested):
Thank you for your contribution @sbathgate! 
Your changes look good to me, but could you please bump a minor version?

migruiz4 (approved):

PR #25450, opened 2024-04-29
[bitnami/nginx-ingress-controller] deployment fail because of OpenShift restricted-v2 issue

Darren-Bell:
### Description of the change
Fix deployment of the nginx-ingress-controller to an Open Shift cluster

### Benefits
Deployment to OpenShit with restricted-v2 policy will not fail because of initContainer with missing containerSecurityContext

### Possible drawbacks
None AFAIK.


carrodher:
Thanks for your contribution, could you please rebase from `main` and bump the chart version in the Chart.yaml file?

Issue #25399, opened 2024-04-26
[harbor] Allow providing all secrets out-of-band


meln5674:
### Name and Version

bitnami/harbor 21.1.1

### What is the problem this feature will solve?

Presently, some Secrets in the harbor chart are "baked into", that is, they require providing sensitive values in the values.yaml, which is not secure when using GitOps.

### What is the feature you are proposing to solve the problem?

The following changes should allow for zero secrets to be created by the chart, if desired:

1. Add a `trivy.existingEnvVarsSecret` field that will point to a secret currently created by `trivy/trivy-secret-envvars.yaml`, do not create the secret if it is provided
2. Move the redis URL from `jobservice/jobservice-config-secret.yaml` at config.yaml's worker_pool.redis_url to `jobservice/jobservice-secret-envvars.yaml` at a separate key `JOB_SERVICE_POOL_REDIS_URL` which the jobservice respects (see https://github.com/goharbor/harbor/blob/main/src/jobservice/config/config.go#L40C42-L40C68)
3. Change `jobservice/jobservice-config-secret.yaml` from a Secret to a ConfigMap, now that the only sensitive value has been removed
4. Add a `jobservice.existingEnvVarsSecret` field that will point to a secret currently created by `jobservice/jobservice-secret-envvars.yaml`, do not create if provided
5. Do not create `core/core-secret.yaml` if the data section would be empty (i.e. if both `core.existingSecret` and `core.secretName` are both provided).

With these changes made, all Secrets are within an if-block, meaning it is possible to configure the chart to emit no Secrets.

I have these changes working in a local copy. I am happy to update the docs and submit a PR if these would be accepted as-is or with minor modifications.

### What alternatives have you considered?

Using a different chart.

carrodher:
Thank you for bringing this issue to our attention. We appreciate your involvement! If you're interested in contributing a solution, we welcome you to create a pull request. The Bitnami team is excited to review your submission and offer feedback. You can find the contributing guidelines [here](https://github.com/bitnami/charts/blob/main/CONTRIBUTING.md).

Your contribution will greatly benefit the community. Feel free to reach out if you have any questions or need assistance.

cthtrifork:
I am also interested in this. Not closed.

PR #25453, opened 2024-04-29
[bitnami/harbor] Enable zero-secret output

meln5674:
### Description of the change

This patch makes the following changes:

* Added value trivy.existingEnvVarsSecret
* Added value jobservice.existingEnvVarsSecret
* Moved redis URL from jobservice config secret to env vars secret
* Switched jobservice config from Secret to ConfigMap now that no sensitive values are present
* Don't create core service if it would be empty

### Benefits

It is now possible to deploy the chart while creating all secrets out-of-band (i.e. kubectl). This means that no sensitive values are required to be in the `values.yaml`. This is more secure when running via GitOps.

Because the jobservice configuration no longer contains the (potentially) sensitive redis URL, it can be stored as a ConfigMap, which makes it easier to view.

All changes should be backwards compatible.

### Possible drawbacks

None known.

### Applicable issues

- fixes https://github.com/bitnami/charts/issues/25399

### Checklist


carrodher:
Thanks for your contribution, could you please rebase from `main`? There is an issue with one of the GH actions which is totally unrelated to your PR but will require to use the latest changes in the `main` branch

meln5674:
Let me know if there's anything needed on my end besides the rebase to get the check to pass.

javsalgar:
It seems that your branch is 31 commits behind main, could you sync your fork? The issue with the action should be fixed by then

meln5674:
Please excuse my impatience, but its been 3 weeks since I've made the requested change, and all checks are passing. I rebased again yesterday. Please let me know if there are any further obstacles to merging that I can address.

bitnami/harbor/templates/core/core-secret.yaml
```diff
@@ -3,6 +3,7 @@ Copyright Broadcom, Inc. All Rights Reserved.
 SPDX-License-Identifier: APACHE-2.0
 */}}
 
+{{- if or (not .Values.core.existingSecret) (not .Values.core.secretName) }}
```
dgomezleon, line 6: This two variables are not use fot the same purpose (a bit confusing I think). I suggest this change:
```suggestion
{{- if (not .Values.core.existingSecret) }}
```

dgomezleon (changes requested):
Thanks a lot for contributing and sorry for the delay. I apply a couple of suggestions. Could you please take a look?

bitnami/harbor/templates/core/core-secret.yaml
meln5674: You are correct that they are used for different things, however, if you check the data section, you'll notice that if both secrets are provided, the secret in this file will be empty. While this is not technically "wrong", it does mean that it A) still creates a secret (when the goal of the PR is to enable it to output zero secrets) and B) prevents the user from creating a secret out-of-band with the same name (as the chart would overwrite it). By checking if at least one of those secrets is not already provided, we can avoid outputting an unnecessary empty secret.

Let me know if I've misunderstood your comment.

dgomezleon (approved):
LGTM

PR #25496, opened 2024-05-02
[bitnami/vault]  Add support for vault oracle plugin persistence storage 

Archanashetty7:
Added support for vault Oracle plugin

### Description of the change

Additional mount path and volume claim template is added for vault oracle plugin support

### Benefits

- The Oracle database plugin is not bundled in the core Vault code tree and can be found at its own git repository here: [hashicorp/vault-plugin-database-oracle](https://github.com/hashicorp/vault-plugin-database-oracle)
- When vault oracle plugin is registered ,vault internally writes into a directory "/tmp" for which an external persistenance storage volume is needed.
- vault oracle plugin is not supported by vault opensource edition, so this will help to enable oracle database secret engine for dynamic secrets engine.

### Possible drawbacks

N/A

### Applicable issues

NA
### Additional information

- This change only works with  custom image with plugin binary installed.
- For linux/amd64, pre-built binaries can be found at [the releases page](https://releases.hashicorp.com/vault-plugin-database-oracle)
- Before running the plugin you will need to have the Oracle Instant Client library installed. These can be downloaded from Oracle. The libraries will need to be placed in the default library search path or defined in the ld.so.conf configuration files.

### Checklist

- [X] Variables are documented in the values.yaml and added to the README.md using [readme-generator-for-helm](https://github.com/bitnami/readme-generator-for-helm)

### DIFF
--- a/bitnami/vault/Chart.yaml
+++ b/bitnami/vault/Chart.yaml
@@ -1,4 +1,4 @@
-# Copyright VMware, Inc.
+# Copyright Broadcom, Inc. All Rights Reserved.
 # SPDX-License-Identifier: APACHE-2.0
 
 annotations:
@@ -35,4 +35,4 @@ maintainers:
 name: vault
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/vault
-version: 1.0.4
+version: 1.1.0
--- a/bitnami/vault/templates/server/statefulset.yaml
+++ b/bitnami/vault/templates/server/statefulset.yaml
@@ -274,5 +274,8 @@ spec:
         {{- if .Values.server.persistence.selector }}
         selector: {{- include "common.tplvalues.render" (dict "value" .Values.server.persistence.selector "context" $) | nindent 10 }}
         {{- end }}
+    {{- if .Values.server.persistence.extraVolumeClaimTemplates }}
+    {{- include "common.tplvalues.render" (dict "value" .Values.server.persistence.extraVolumeClaimTemplates "context" $) | nindent 4 }}
+    {{- end }}
   {{- end }}
 {{- end }}
--- a/bitnami/vault/values.yaml
+++ b/bitnami/vault/values.yaml
@@ -368,6 +368,10 @@ server:
   ## @param server.extraVolumeMounts Optionally specify extra list of additional volumeMounts for the Vault Server container(s)
   ##
   extraVolumeMounts: []
+  ## e.g.
+  ##- name: plugin
+  ##  mountPath: /tmp
+
   ## @param server.sidecars Add additional sidecar containers to the Vault Server pod(s)
   ## e.g:
   ## sidecars:
@@ -726,6 +730,21 @@ server:
     ## @param server.persistence.existingClaim The name of an existing PVC to use for persistence
     ##
     existingClaim: ""
+    ## @param server.persistence.extraVolumeClaimTemplates  Add additional VolumeClaimTemplates for enabling any plugins or any other purpose to the Vault Server pod(s)
+    ##
+    extraVolumeClaimTemplates: []
+    ## e.g:
+    ##- metadata:
+    ##    name: plugin
+    ##    annotations:
+    ##  spec:
+    ##    accessModes:
+    ##      - "ReadWriteOnce"
+    ##    resources:
+    ##      requests:
+    ##        storage: "10Gi"
+    ##    storageClassName: nfs-client
+
   ## @section Vault Server Metrics Parameters
   ##
 
--- a/bitnami/vault/README.md
+++ b/bitnami/vault/README.md
@@ -346,17 +346,18 @@ The [Bitnami vault](https://github.com/bitnami/containers/tree/main/bitnami/vaul
 
 ### Source Conttroller Persistence Parameters
 
-| Name                               | Description                                                                             | Value                 |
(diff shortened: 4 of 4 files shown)


bitnami/vault/Chart.yaml
```diff
@@ -35,4 +35,4 @@ maintainers:
 name: vault
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/vault
-version: 1.0.4
+version: 1.0.5
```
fmulero, line 38: New feature is added, according to [semver](https://semver.org), a new minor version should be bumped.
```suggestion
version: 1.1.0
```

bitnami/vault/templates/server/statefulset.yaml
```diff
@@ -230,6 +230,8 @@ spec:
               mountPath: /bitnami/vault/config
             - name: home
               mountPath: /bitnami/vault/home
+            - name: plugin
+              mountPath: /tmp
```
fmulero, line 234: If you accept my comment below, we could delegate this to the `extraVolumeMounts`.

bitnami/vault/templates/server/statefulset.yaml
```diff
@@ -274,5 +276,32 @@ spec:
...
+        {{- range .Values.server.persistence.accessModes }}
+          - {{ . | quote }}
+        {{- end }}
+        resources:
+          requests:
+            storage: {{ .Values.server.persistence.size | quote }}
+        {{- include "common.storage.class" (dict "persistence" .Values.server.persistence "global" .Values.global) | nindent 8 }}
+        {{- if .Values.server.persistence.selector }}
+        selector: {{- include "common.tplvalues.render" (dict "value" .Values.server.persistence.selector "context" $) | nindent 10 }}
+        {{- end }}
```
fmulero, line 305: What do you think about replacing this piece of code with something like this:
```suggestion
        {{- if .Values.server.persistence.extraVolumeClaimTemplates }}
        {{- include "common.tplvalues.render" (dict "value" .Values.server.persistence.extraVolumeClaimTemplates "context" $) | nindent 8 }}
        {{- end }}
```

fmulero (changes requested):
HI @Archanashetty7, thanks a lot for this contribution.

I see your changes very interesting but I have a little concern: this volume is only needed if you use certain plugins and this change will create it always if persistence is enabled (and it is enabled by default). What do you think if we delegate this decision to the user and we allow that configuration? I mean, users could set its own `VolumeClaimTemplate` according to their needs and mounting it with the `extraVolumeMounts`. I think that approach is more flexible.

Archanashetty7:
> HI @Archanashetty7, thanks a lot for this contribution.
> 
> I see your changes very interesting but I have a little concern: this volume is only needed if you use certain plugins and this change will create it always if persistence is enabled (and it is enabled by default). What do you think if we delegate this decision to the user and we allow that configuration? I mean, users could set its own `VolumeClaimTemplate` according to their needs and mounting it with the `extraVolumeMounts`. I think that approach is more flexible.

Hi @fmulero , thanks for the quick response
yes, we can allow user to configure volumeclaimtemplate according to their needs

bitnami/vault/Chart.yaml
Archanashetty7: Sure I will update chart version

bitnami/vault/templates/server/statefulset.yaml
Archanashetty7: okay, This will also introduce a new param extraVolumeClaimTemplates in values.yaml right?

bitnami/vault/templates/server/statefulset.yaml
```diff
@@ -274,5 +274,8 @@ spec:
         {{- if .Values.server.persistence.selector }}
         selector: {{- include "common.tplvalues.render" (dict "value" .Values.server.persistence.selector "context" $) | nindent 10 }}
         {{- end }}
+        {{- if .Values.server.persistence.extraVolumeClaimTemplates }}
+        {{- include "common.tplvalues.render" (dict "value" .Values.server.persistence.extraVolumeClaimTemplates "context" $) | nindent 4 }}
+        {{- end }}
```
fmulero, line 279: Sorry @Archanashetty7 The indentation in my first comment was wrong, my failure. It should be like this:

```suggestion
    {{- if .Values.server.persistence.extraVolumeClaimTemplates }}
    {{- include "common.tplvalues.render" (dict "value" .Values.server.persistence.extraVolumeClaimTemplates "context" $) | nindent 4 }}
    {{- end }}
```

bitnami/vault/values.yaml
```diff
@@ -368,6 +368,10 @@ server:
   ## @param server.extraVolumeMounts Optionally specify extra list of additional volumeMounts for the Vault Server container(s)
   ##
   extraVolumeMounts: []
+  ## e.g. for oracle plugin installation:
```
fmulero, line 371: Please avoid references to third party companies in the documentation.

fmulero (commented):
Hi @Archanashetty7 I left some comments. Sorry for being so picky.

bitnami/vault/values.yaml
Archanashetty7: I have updated

fmulero (approved):
Thanks a lot @Archanashetty7! LGTM

Archanashetty7:
thank you @fmulero

Issue #19173, opened 2023-09-07
[bitnami/mastodon] Use of go templating inside of values.yaml for mastodon chart breaks Argo CD ApplicationSets


jessebot:
### Name and Version

bitnami/mastodon 2.1.3

### What architecture are you using?

amd64

### What steps will reproduce the bug?

On k3s using Argo CD 2.8.3, if you try to deploy an ApplicationSet for mastodon which uses a generator (which also using [go templates](https://argo-cd.readthedocs.io/en/latest/operator-manual/applicationset/GoTemplate/)), using a values.yaml like the one listed below for instance, it will try to pull that from the external [generator](https://argo-cd.readthedocs.io/en/latest/operator-manual/applicationset/Generators/).

Example ApplicationSet that uses a custom defaultConfig via values.yaml:
https://github.com/small-hack/argocd-apps/blob/main/mastodon/mastodon_argocd_appset.yaml

### Are you using any custom parameters or values?

```yaml
            ## [string] Default config for Mastodon in the form of environment variables
            defaultConfig: |
              DB_HOST: {{ include "mastodon.database.host" . | quote }}
              DB_PORT: {{ include "mastodon.database.port" . | quote }}
              DB_NAME: {{ include "mastodon.database.name" . | quote }}
              DB_USER: {{ include "mastodon.database.user" . | quote }}
              ES_ENABLED: {{ .Values.enableSearches | quote }}
              ES_HOST: {{ include "mastodon.elasticsearch.host" . | quote }}
              ES_PORT: {{ include "mastodon.elasticsearch.port" . | quote }}
              WEB_DOMAIN: {{ include "mastodon.web.domain" . | quote }}
              LOCAL_DOMAIN: {{ .Values.localDomain | quote }}
              LOCAL_HTTPS: {{ .Values.local_https | quote }}
              DEFAULT_LOCALE: {{ .Values.defaultLocale | quote }}
              STREAMING_API_BASE_URL: {{ include "mastodon.streaming.url" . | quote }}
              REDIS_HOST: {{ include "mastodon.redis.host" . | quote }}
              REDIS_PORT: {{ include "mastodon.redis.port" . | quote }}
              SMTP_PORT: {{ .Values.smtp.port | quote }}
              SMTP_FROM_ADDRESS: {{ .Values.smtp.from_address | quote }}
              SMTP_DOMAIN: {{ .Values.smtp.domain | quote }}
[...]

PR #25519, opened 2024-05-03
[bitnami/mongodb] Network policy review

### LINKED ISSUE #25340: Allow templating NetworkPolicy ingressNSMatchLabels values?
### Name and Version

bitnami/mongodb 15.1.5

I'm looking to lock down NetworkPolicies to only allow ingress traffic from the same namespace as the deployment.

Conceptually, I was thinking I could configure the chart like so:
```yaml
networkPolicy:
  allowExternal: false
  ingressNSMatchLabels:
    kubernetes.io/metadata.name: "{{ .Release.Namespace }}"
```

However the chart [takes the raw](https://github.com/bitnami/charts/blob/288a544c640bd8850b6e2b40c254d246952524a9/bitnami/mongodb/templates/networkpolicy.yaml#L82) `ingressNSMatchLabels` values, so I cannot use `{{ .Release.Namespace }}` here.

Is it possible to render these values to accommodate this use case?  Is there another way already possible in the chart to do this that I'm not seeing?

The verbiage in the README is a little confusing I also think. `allowExternal` is a value, but if I understand correctly it's not really about external so much as requiring the pod to have a specific label, regardless of the namespace.  Similarly on the description for `ingressNSMatchLabels`:  `Labels to match to allow traffic from other namespaces` but it doesn't look like the namespace restriction is in place to begin with.

This issue is specific to the MongoDB chart but it looks like most NetworkPolicies in this repository follow the same template, so for consistency sake, it may be scoped to more charts than just this.

fmulero:
### Description of the change

Review network policy features to allow users cover their needs without forcing specific labels.

### Benefits

Users can restrict access to their deployments in a cleaner way, without leaving open doors to specific labeled pods.

### Possible drawbacks

None

### Applicable issues

- fixes #25340

### Checklist


javsalgar:
Users in the past have reported that using templates inside values.yaml may potentially break argo-cd deployments: https://github.com/bitnami/charts/issues/19173

If we went on with this change we may add issues to all charts in argo-cd. I would add an if-else or something similar to avoid adding that templated values. We may need to see what happens in other charts that we currently use templated values.

rafariossaa (approved):
LGTM

fmulero:
@javsalgar I've just applied a possible solution. It works but the templates are becoming more complicated. Any thoughts?

rafariossaa (dismissed):
LGTM

javsalgar (approved):

PR #25566, opened 2024-05-06
[bitnami/argo-cd] Chart modification to activate external Redis

### LINKED ISSUE #15064: [bitnami/argocd] External redis with username and password does not connect
### Name and Version

bitnami/argocd@latest

1. Create a secret with username and password and set up a user in redis with username and password
2. Try to use `externalRedis` with just a password/ using the secret (username isn't possible to offer)
3. Install chart

_No response_

Expect to connect to redis

### What do you see instead?

Redis wait pods fail to connect saying password incorrect or user is disabled. 

### Additional information

I installed using the redis dependency with no issues. I'd like to, however, use my own redis instance. 

Scott

myeong-han:
### Description of the change
This PR contains changes to improve the issue of calling `externalRedis.host` without referencing `externalRedis.existingSecret` when `redis.enabled` specified in `values.yaml` is `false`.

### Benefits
When deploying argo-cd, you can use an existing Redis instead of deploying a new Redis.
Screenshot 2024-05-06 at 11 22 03 PM
 

### Applicable issues

- fixes #15064

### Additional information
To enable externalRedis, two values in `values.yaml` must be disabled.
- `redis.enabled=false`
- `redisWait.enabled=false`

Specifies the label of externalRedis Service.
- `externalRedis.selector`

### Checklist


dgomezleon (changes requested):
Hi @myeong-han ,

I think you are assuming that Redis is always used (internal or external). Note that most of the `else` conditions should also validate if an external Redis is configured. Otherwise, the logic below should not be added. Could you please double-check it taking into account if Redis is not used?

myeong-han:
I hadn't thought about it. This is because the `argocd.redisHost` value used in `templates/server/deployment.yaml` was determined by `redis.enabled` in `./templates/_helpers.tpl`.

However, we confirmed that the value is not a problem even if redis is not used, so we did not modify the definition.

Please let me know if there is any mistake.

### Additional information
Required value
- `redis.enabled=false`
- `redisWait.enabled=false`
- `externalRedis.enabled=true`

dgomezleon (approved):
LGTM

dgomezleon (approved):

PR #25839, opened 2024-05-14
[bitnami/postgresql] Only load existing K8s Secret when needed

### LINKED ISSUE #25644: Postgresql Secret Issues with Cyberark Conjur
### Name and Version

bitnami/postgresql 15.2.9

None

1. Create a Cyberark Conjur Secret with replication and admin user credentials
2. Use that Secret as source for the `global.postgresql.auth.existingSecret` value
3. Add the Cyberark Conjur Sidecar to the helm values
4. Try to deploy

The helm chart gets deployed with the Cyberark Conjur Sidecar which then reads and updates the Secret in `global.postgresql.auth.existingSecret` to contain the correct values. Postgres then uses those values to start up

### What do you see instead?

The helm chart deployment fails due to this error:

```
error: execution error at (postgresql/templates/secrets.yaml:15:27): 
PASSWORDS ERROR: The secret "postgresql-app-secret" does not contain the key "postgres-password"
```

### Additional information

The Secret `postgresql-app-secret` initially only contains information for conjur. Those information are then used by the Cyberark Conjur Sidecar to update the k8s secret with the real values. So initially the `postgres-password` key doesn't exist in the k8s Secret but conjur will insert it as soon as it starts.

ChristianRaoulis:
### Description of the change

This change removes the early secret read from the generell execution and only reads the secrets when those information are mandatory for the chart.

### Benefits

This allows so use secret manager like Cyberark Conjur.

### Applicable issues

- fixes #25644

### Checklist


ChristianRaoulis:
It is the first time ever I'm working on Helm Charts.
Please look through it carefully 😄

rafariossaa:
Hi,
  I am sorry but I think I don't fully understand how this solves the issue.
  In your issue you mentioned providing and `existingSecret` and the template already includes a condition to not render the secret when `existingSecret` is provided. In this case, adding a condition to the helm variables would not help here.
  I don't know how Conjur works but based on the error message, I would say that the issue is that the provided `existingSecret` does not include the `postgres-password` key.
  If counjur updates the secrets after the pod is started, probably it already has picked an empty value (or not the right value) for the password and that wrong value would be used for the configuration. In order to pick the right value the pod would need to be restarted and if that is already persisted the new value would be ignored.

ChristianRaoulis:
Hay,

My issue is basically just that helm doesn't continue to deploy the postgresql helm chart because the secret doesn't contain the `postgres-password` key at that time. 

Conjur will take care about the secret and insert the key at the corret time. I'm not sure how this works "under the hood" in conjur but it works for my Redis and RabbitMQ which are both bitnami helm charts too.

By preventing helm from loading the existing Secret when it is not required to load it, it should not run into the `The secret "postgresql-app-secret" does not contain the key "postgres-password"` error anymore and just continue to deploy the chart. Conjur will "initialize" the k8s secret and the postgresql pods will be able to get the secret values at runtime.

ChristianRaoulis:
I've tried to visualize what conjur does in k8s

ConjurFlow drawio

I hope that helps understanding my problem

rafariossaa:
Could be in that case to have more sense to move the `if` in line 29 to the very beginning ? So if there is not need to create the secret no variables need to be calculated.
WYT ?

PR #25856, opened 2024-05-14
bitnami/kafka Fix for sed in kafka-init.sh

### LINKED ISSUE #25500: [bitnami/kafka] sed expression in init container of broker has default delimiter, that lead to errors using passwords with '/'
### Name and Version

bitnami/kafka 28.0.4

amd64

If you set password for sasl with '/' symbol in it, so sed will exit with error, because it is used default delimiter in init script
For example you have such sasl config in values.yaml
```
sasl:
  client:
    users:
      - user
    passwords:
      - passw/ord
```
The problem is on line 140 of scripts-configmap.yaml
https://github.com/bitnami/charts/blob/bb1ece6f1cb4110af1a35ea23bb16e03a76c7595/bitnami/kafka/templates/scripts-configmap.yaml#L140

_No response_

There is no error in sed, so we can use password without escaping special characters in it
 

### What do you see instead?

sed gets error

### Additional information

Use non-printable delimiter like this in scripts-configmap.yaml on line 113
https://github.com/bitnami/charts/blob/bb1ece6f1cb4110af1a35ea23bb16e03a76c7595/bitnami/kafka/templates/scripts-configmap.yaml#L113C9-L113C29

vyaradaikin:
Use non-printable delimiter to avoid sed errors with delimiter symbols in passwords

### Description of the change

Fixes sed string to be correct with passwords that use / symbols (standard sed delimiter symbol)

### Benefits

No sed errors in case of using / symbols in passwords

### Applicable issues

- fixes #25500

### Checklist

### DIFF
--- a/bitnami/kafka/Chart.yaml
+++ b/bitnami/kafka/Chart.yaml
@@ -42,4 +42,4 @@ maintainers:
 name: kafka
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/kafka
-version: 28.2.3
+version: 28.2.4
--- a/bitnami/kafka/templates/scripts-configmap.yaml
+++ b/bitnami/kafka/templates/scripts-configmap.yaml
@@ -137,7 +137,8 @@ data:
     replace_placeholder() {
       local placeholder="${1:?missing placeholder value}"
       local password="${2:?missing password value}"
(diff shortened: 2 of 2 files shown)


juan131 (approved):
Thanks for this enhancement! LGTM

PR #25861, opened 2024-05-14
[bitnami/template] Correct image tag in template

dan-hill2802:
Correcting the tag in `/templates/CHART_NAME/values.yaml` to match the image tags on Docker Hub. Looks like it was missed in #23658


carrodher (approved):
Thanks!

Issue #25698, opened 2024-05-13
[bitnami/rabbitmq] Cluster does not recover after reboot, even with forceBoot: true


razvanphp:
### Name and Version

bitnami/rabbitmq 12.15.0

### What architecture are you using?

amd64

### What steps will reproduce the bug?

We run this chart in TrueNAS server, deployed with fluxCD. With 3 pods, restart the k3s node, and cluster will not recover.

### Are you using any custom parameters or values?

```yaml
apiVersion: helm.toolkit.fluxcd.io/v2beta1
kind: HelmRelease
metadata:
  name: rabbitmq
  namespace: vampirebyte
spec:
  releaseName: rabbitmq
  chart:
    spec:
      chart: rabbitmq
      version: 12.15.0
      sourceRef:
        kind: HelmRepository
        name: bitnami
        namespace: flux-system
  interval: 1h
  values:
    extraPlugins: rabbitmq_mqtt rabbitmq_web_mqtt
    image:
      debug: true
    # diagnosticMode:
    #   enabled: true
    clustering:
      forceBoot: true
    auth:
      tls:
        enabled: true
        existingSecret: rabbitmq-tls-secret
        existingSecretFullChain: true
    replicaCount: 3
    ingress:
      enabled: true
      hostname: mq.[blablabla]
      tls: true
      existingSecret: rabbitmq-tls-secret
      extraPaths:
      - path: /ws
        pathType: Prefix
        backend:
          service:
            name: rabbitmq
            port:
              number: 15675
      annotations:
        traefik.ingress.kubernetes.io/router.entrypoints: websecure
        traefik.ingress.kubernetes.io/router.tls: 'true'
    extraConfiguration: |
      log.console = true

      web_mqtt.ssl.port       = 15676
      web_mqtt.ssl.backlog    = 1024
      web_mqtt.ssl.cacertfile = /opt/bitnami/rabbitmq/certs/ca_certificate.pem
      web_mqtt.ssl.certfile   = /opt/bitnami/rabbitmq/certs/server_certificate.pem
      web_mqtt.ssl.keyfile    = /opt/bitnami/rabbitmq/certs/server_key.pem

      management.tcp.port       = 15672
      management.ssl.port       = 15671
      management.ssl.cacertfile = /opt/bitnami/rabbitmq/certs/ca_certificate.pem
      management.ssl.certfile   = /opt/bitnami/rabbitmq/certs/server_certificate.pem
[...]

Issue #16081, opened 2023-04-17
rabbitmq clustering force boot should be true by default


hronix:
### Name and Version

bitnami/rabiitmq

### What is the problem this feature will solve?

Problem:
By default clustering is on. When you update nodes with new ami on AWS it can with high probability leave last working pod which is NOT 0. And when rabbitmq-0 will start it can't, problem described in details here: https://stackoverflow.com/questions/60407082/rabbit-mq-error-while-waiting-for-mnesia-tables

It makes default configuration doesn't work properly

### What is the feature you are proposing to solve the problem?

set "clustering.forceBoot"is true by default

corico44:
Hello @hronix,

Thank you so much for reporting this issue. Would you be open to contributing this change via a PR? We'd be glad to review and help with the release.

michaelklishin:
I'm a RabbitMQ core team member. Force booting nodes was never meant to be used by default. 
@hronix @corico44 this is an awful, awful workaround and a catastrophically bad default if you ask me.

On Kubernetes, the need to force boot nodes usually comes from an unfortunately picked readiness probes. Two relevant doc sections:

 * How [nodes rejoin their peers](https://www.rabbitmq.com/docs/clustering#restarting) when they restart
 * How [readiness probes affect nodes rejoining](https://www.rabbitmq.com/docs/clustering#restarting-readiness-probes) their cluster
 * [Kubernetes-specific deployment requirements](https://www.rabbitmq.com/docs/cluster-formation#peer-discovery-k8s)

michaelklishin:
@javsalgar please see my comment above. It would be great if the RabbitMQ core team was consulted before changes like this are made default.

mkuratczyk:
In the https://github.com/rabbitmq/cluster-operator we use `podManagementPolicy: Parallel` so in the case of a complete outage for example, all nodes will be started without waiting for node 0 to be ready first.

javsalgar:
Hi @michaelklishin,

Thanks for letting us know. However, checking the code I do not see that the user suggestion was actually implemented in the chart, which still has the default boot value set to false:

```
  ## @param clustering.forceBoot Force boot of an unexpectedly shut down cluster (in an unexpected order).
  ## forceBoot executes 'rabbitmqctl force_boot' to force boot cluster shut down unexpectedly in an unknown order
  ## ref: https://www.rabbitmq.com/rabbitmqctl.8.html#force_boot
  ##
  forceBoot: false
```

[...]

PR #25873, opened 2024-05-14
RabbitMQ: explain how to properly avoid a deployment deadlock

michaelklishin:
See https://github.com/bitnami/charts/issues/25698 and https://github.com/bitnami/charts/issues/16081#issuecomment-2106462797.

For now, we can at least recommend the right thing to do.

### DIFF
--- a/bitnami/rabbitmq/README.md
+++ b/bitnami/rabbitmq/README.md
@@ -268,17 +268,17 @@ extraConfiguration: |
 RabbitMQ nodes assume their peers come back online within five minutes (by default). With the `OrderedReady` pod management policy is used
 with a readiness probe that implicitly requires a fully booted node, the deployment can deadlock:
 
- - Kubernetes will expect the first node to pass a readiness probe
- - The readiness probe may require a fully booted node
- - The node will fully boot after it detects that its peers have come online
- - Kubernetes will not start any more pods until the first one boots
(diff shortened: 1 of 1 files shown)


michaelklishin:
I will take care of the Markdown linter failures.

bitnami/rabbitmq/README.md
```diff
@@ -263,43 +263,35 @@ extraConfiguration: |
   log.console.formatter = json
 ```
 
-### Recover the cluster from complete shutdown
+## How to Avoid Deadlocked Deployemnts After a Cluster-Wide Restart
 
-> IMPORTANT: Some of these procedures can lead to data loss. Always make a backup beforehand.
+RabbitMQ nodes assume their peers come back online within five minutes (by default). With the `OrderedReady` pod management policy is used
```
cvijdea-bd, line 268: should this be "*When* the `OrderedReady` pod ..."?

michaelklishin: Thanks, corrected.

michaelklishin:
https://github.com/michaelklishin/charts-1/commit/6fd17b789a382ac04eb2847f710bff681caf7040 does included a `Signed-off-By`, I don't know what else the DCO workflow wants me to do.

michaelklishin:
Moved to https://github.com/bitnami/charts/pull/25874.

PR #25874, opened 2024-05-14
RabbitMQ: explain how to properly avoid an `OrderedReady`-induced deployment deadlock (take 2)

michaelklishin:
Explain the problem and two widely used solutions
instead of recommending force booting nodes.

This PR includes `Signed-off-by` commits.

See https://github.com/bitnami/charts/issues/25698 and https://github.com/bitnami/charts/issues/16081#issuecomment-2106462797.

For now, we can at least recommend the right thing to do.

### DIFF
--- a/bitnami/rabbitmq/README.md
+++ b/bitnami/rabbitmq/README.md
@@ -272,14 +272,17 @@ RabbitMQ nodes assume their peers come back online within five minutes (by defau
 - The node will fully boot after it detects that its peers have come online
 - Kubernetes will not start any more pods until the first one boots
 
-Using [Bitnami RabbitMQ Cluster Operator](https://hub.docker.com/r/bitnamicharts/rabbitmq-cluster-operator) is the easiest solution.
-
-Alternatively, the following combination of deployment settings avoids the problem:
+The following combination of deployment settings avoids the problem:
 
 - Use `podManagementPolicy: "Parallel"` to boot multiple cluster nodes in parallel
 - Use `rabbitmq-diagnostics ping` for readiness probe
 
-Note that forcing nodes to boot is **not a solution** and doing so **can be dangerous**. Forced booting is a last resort mechanism in RabbitMQ that helps make remaining cluster nodes recover and rejoin each other after a permanent loss of some of their former peers. In other words, forced booting a node is an emergency event recovery procedure.
+This combination is what [the official RabbitMQ Cluster Operator](https://www.rabbitmq.com/kubernetes/operator/operator-overview) uses,
+for example.
+
+Note that forcing nodes to boot is **not a solution** and doing so **can be dangerous**.
+Forced booting is a last resort mechanism in RabbitMQ that helps make remaining cluster
+nodes recover and rejoin each other after a permanent loss of some of their former peers. In other words, forced booting a node is an emergency event recovery procedure.
 
 To learn more, see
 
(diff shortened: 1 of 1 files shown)


razvanphp:
LGTM. Thank you for putting your time into this, eases the frustration and pain.

bitnami/rabbitmq/README.md
```diff
@@ -263,35 +263,31 @@ extraConfiguration: |
...
-with a readiness probe that implicitly requires a fully booted node, the deployment can deadlock:
+RabbitMQ nodes assume their peers come back online within five minutes (by default). When the `OrderedReady` pod management policy is used with a readiness probe that implicitly requires a fully booted node, the deployment can deadlock:
 
 - Kubernetes will expect the first node to pass a readiness probe
 - The readiness probe may require a fully booted node
 - The node will fully boot after it detects that its peers have come online
 - Kubernetes will not start any more pods until the first one boots
 
-Using [RabbitMQ Cluster Operator](https://www.rabbitmq.com/kubernetes/operator/operator-overview) is the easiest solution.
+Using [Bitnami RabbitMQ Cluster Operator](https://hub.docker.com/r/bitnamicharts/rabbitmq-cluster-operator) is the easiest solution.
```
michaelklishin, line 275: If this is [the RabbitMQ core team's Operator](https://www.rabbitmq.com/kubernetes/operator/operator-overview) then it's OK to recommend but I don't understand what doe Bitnami have to do with it. It was not developed by Bitnami and our team never refers to it as the "Bitnami Operator" for that reason.

If that is a different Operator, then I must say that I have never signed up for recommending a different one and the sentence must be taken out completely.

bitnami/rabbitmq/README.md
```diff
@@ -263,35 +263,31 @@ extraConfiguration: |
   log.console.formatter = json
 ```
 
-## How to Avoid Deadlocked Deployments After a Cluster-Wide Restart
+### How to Avoid Deadlocked Deployments After a Cluster-Wide Restart
```
michaelklishin, line 266: This deserves a level 2 header because it is a crucially important problem and not just "a yet another configuration topic", as regularly appearing discussions and issues suggest.

bitnami/rabbitmq/README.md
michaelklishin: Someone from the core team has confirmed it to me that Bitnami engineers did not contribute to https://www.rabbitmq.com/kubernetes/operator/operator-overview.

@carrodher it is NOT OK to call a project not developed by Bitnami "a Bitnami Operator".

michaelklishin:
@carrodher thank you for the review. As noted above, I have a major problem with https://github.com/bitnami/charts/pull/25874/commits/3afe6072838cfeb77ddce720f10d03e8d13e826c removing a link to the official (the project is maintained by the RabbitMQ core team, the docs are on rabbitmq.com, you get the idea) K8S Operator, so I have backed out a link to
a "Bitnami Cluster Operator" because I have no idea what that project is, and I never meant to recommend it.

If the "Bitnami Cluster Operator" uses the same combination of recommended settings, then
both of them can be mentioned, or none of the Operators (the README can recommend the pod management policy and readiness probe combination only).

If in a subsequent commit (outside of this PR) the Bitnami Operator is added, I'd like to know where its source lives so that Team RabbitMQ can assess what it does.

But again, I DO NOT CONSENT to my links being changed to an Operator I have never meant to recommend in this PR. Add it in a separate PR if you want but the RabbitMQ core team recommends what we develop, know and document.

michaelklishin:
Since I do not consent to some of the changes applied in https://github.com/bitnami/charts/commit/3afe6072838cfeb77ddce720f10d03e8d13e826c, I will submit another PR.

PR #25878, opened 2024-05-14
[bitnami/prometheus] Updated prometheus server local cluster address

krufab:
Updated prometheus server local cluster address, as it pointed to the wrong service name.

### Description of the change

The post install message generates this cluster url: prometheus.prometheus.svc.cluster.local
However, the url is wrong, as it uses the `common.names.fullname` template instead of the `prometheus.server.fullname`.
Using the correct template, the generated cluster url is: **prometheus-server.prometheus.svc.cluster.local**.

For reference, also the alert manager and thanos internal url are generated in a similar way:
- template "prometheus.alertmanager.fullname"
- template "prometheus.thanos-sidecar.fullname"

### Benefits

The cluster url for the prometheus-server service is now correct.

### Checklist


fmulero (approved):
LGTM. Thanks a lot @krufab for your contribution

PR #25916, opened 2024-05-15
RabbitMQ: explain how to safely avoid an `OrderedReady`-induced deployment deadlock

michaelklishin:
Explain the problem and a widely used solution
instead of recommending force booting nodes.

These practices are generally used by existing
RabbitMQ K8S Cluster Operators but this PR
intentionally does not recommend any of them.

References https://github.com/bitnami/charts/issues/25698 and https://github.com/bitnami/charts/issues/16081.

### DIFF
--- a/bitnami/rabbitmq/README.md
+++ b/bitnami/rabbitmq/README.md
@@ -285,9 +285,7 @@ To learn more, please consult RabbitMQ documentation guides:
 
 #### Do Not Force Boot Nodes on a Regular Basis
 
-Note that forcing nodes to boot is **not a solution** and doing so **can be dangerous**.
-Forced booting is a last resort mechanism in RabbitMQ that helps make remaining cluster
-nodes recover and rejoin each other after a permanent loss of some of their former peers. In other words, forced booting a node is an emergency event recovery procedure.
+Note that forcing nodes to boot is **not a solution** and doing so **can be dangerous**. Forced booting is a last resort mechanism in RabbitMQ that helps make remaining cluster nodes recover and rejoin each other after a permanent loss of some of their former peers. In other words, forced booting a node is an emergency event recovery procedure.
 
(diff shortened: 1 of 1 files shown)


bitnami/rabbitmq/README.md
```diff
@@ -263,41 +263,31 @@ extraConfiguration: |
...
-    --set auth.erlangCookie=ERLANG_COOKIE
-```
-
-> Note: You need to substitute the placeholders `REGISTRY_NAME` and `REPOSITORY_NAME` with a reference to your Helm chart registry and repository. For example, in the case of Bitnami, you need to use `REGISTRY_NAME=registry-1.docker.io` and `REPOSITORY_NAME=bitnamicharts`.
+#### Do Not Force Boot Nodes on a Regular Basis
 
-More information: [Clustering Guide: Restarting](https://www.rabbitmq.com/clustering.html#restarting).
+Note that forcing nodes to boot is **not a solution** and doing so **can be dangerous**.
+Forced booting is a last resort mechanism in RabbitMQ that helps make remaining cluster
+nodes recover and rejoin each other after a permanent loss of some of their former peers. In other words, forced booting a node is an emergency event recovery procedure.
```
carrodher, line 290:
```suggestion
Note that forcing nodes to boot is **not a solution** and doing so **can be dangerous**. Forced booting is a last resort mechanism in RabbitMQ that helps make remaining cluster nodes recover and rejoin each other after a permanent loss of some of their former peers. In other words, forced booting a node is an emergency event recovery procedure.
```

michaelklishin:
I will have to submit another PR because of https://github.com/bitnami/charts/pull/25916/checks?check_run_id=25008225640. My default GitHub account email is not what DCO expects and even if I squash and push, that workflow somehow does not detect the changes.

PR #25941, opened 2024-05-16
[bitnami/etcd] Fix inconsistent between readme and code for etcd

doudoubobo:
It should be `auth.rbac.create` not `auth.rbac.enabled` since https://github.com/bitnami/charts/commit/c3f6f9a4a6f5f81038a33245e198557ba4b19024


migruiz4 (approved):
Nice catch! Thank you for your contribution!

PR #26342, opened 2024-05-22
[bitnami/kafka] #25646 Kafka config with yaml parameter map

### LINKED ISSUE #25646: [bitnami/kafka] add support for helm parameter maps for extraConfig
### Name and Version

bitnami/kafka 28.1.1

With the migration to Kafka 3.5, an `extraConfig` parameter was added to the chart.
We regret that this is just a text value instead of a helm parameter map.
This choice blocks the overloading of a certain setting via additional values.yaml layers. 
With the current implementation, you need to duplicate all `extraConfig` if you want to change a single parameter in it.

As an example, instead of this
```yaml
extraConfig: |
  max.message.bytes = 1048576
  offsets.topic.replication.factor = 3
  default.replication.factor = 3
  delete.topic.enable = true
  auto.create.topics.enable = false
```
we would prefer to have
```yaml
extraConfig:
  max.message.bytes: 1048576
  offsets.topic.replication.factor: 3
  default.replication.factor: 3
  delete.topic.enable: true
  auto.create.topics.enable: false
```
to allow overloading.

helm parameter maps to allow overloading the extraConfig.

rubenvw-ngdata:
### Description of the change

Use a yaml parameter map instead of a flat file for the extraConfig parameter. This allows for easier overloading both between the default and broker / controller values; and between multiple values.yaml.

### Benefits

You can now overload config values easily.

### Applicable issues

- fixes #25646

### Checklist


bitnami/common/templates/_propertiesvalues.tpl
```diff
@@ -0,0 +1,5 @@
+{{- define "common.properties.render" -}}
```
migruiz4, line 1: Please add this function inside `kafka/_helpers.tpl` instead.

This can not be added inside common due to several reasons:
- This function is specific to Kafka, and I don't see any other chart benefitting from this in a short-term, so there is no point adding it to the common chart at the moment. If another chart could use this functionality in the future we could move it to the common chart.
- Changes to several charts in the same PR are not allowed in our contributing guidelines. To modify the bitnami/common chart, this change should be performed in a separate PR, merged and then update the common chart dependency in Kafka.

bitnami/kafka/Chart.yaml
```diff
@@ -42,4 +42,4 @@ maintainers:
 name: kafka
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/kafka
-version: 28.3.0
+version: 28.3.1
```
migruiz4, line 45: This change breaks backward compatibility, so this must be a major release and therefore be documented in the README.md.

As an alternative, I would suggest keeping the value `extraConfig` as it is, and instead add this functionality in a new value `extraConfigYaml`. So we would have two methods to extend the configuration: as plaintext and as a yaml object.

Because kafka settings usually include dots, I would understand not all users may like the yaml approach which would require quoting the object keys.

bitnami/common/templates/_propertiesvalues.tpl
```diff
@@ -0,0 +1,5 @@
...
+{{- range $key, $value := . }}
+{{ $key }}={{ $value }}
```
migruiz4, line 3: Rendering the `$value` as a template would be useful I think

```suggestion
{{ $key }}={{ tpl $value $ }}
```

bitnami/kafka/Chart.yaml
migruiz4: Additionally, the current `extraConfig` is rendered as a template, which may allow other use-cases that YAML format may not.

bitnami/kafka/values.yaml
```diff
@@ -115,9 +115,9 @@ extraInit: ""
 ##
 config: ""
 existingConfigmap: ""
-## @param extraConfig Additional configuration to be appended at the end of the generated Kafka configuration file.
+## @param extraConfig Additional configuration in yaml format to be appended at the end of the generated Kafka configuration file.
```
migruiz4, line 118: Could you please add a working example?
Something similar to:

```
  ## E.g.
  ## extraConfig:
  ##   'index.interval.bytes': 4096
  ##
```

migruiz4 (commented):
Hi @rubenvw-ngdata,

As mentioned in my comments, I would appreciate it if you could consider adding this feature as a new value instead of introducing a non-backward compatible change.

bitnami/common/templates/_propertiesvalues.tpl
rubenvw-ngdata: moved as proposed

bitnami/common/templates/_propertiesvalues.tpl
rubenvw-ngdata: fixed as proposed

bitnami/kafka/Chart.yaml
rubenvw-ngdata: I've changed it and followed your extraConfigYaml proposal.

migruiz4 (commented):
Thank you very much for applying my suggestions!

I see the `tpl` I suggested was causing some issues. I tested your branch locally and using `common.tplvalues.render` should fix the issue.

bitnami/kafka/Chart.yaml
```diff
@@ -40,4 +40,4 @@ maintainers:
 name: kafka
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/kafka
-version: 29.1.2
+version: 29.1.3
```
migruiz4, line 43: Please bump a minor version

```suggestion
version: 29.2.0
```

We use semver.org for versioning, which translates into:
- Major version for non-backward compatible changes.
- Minor version for new values.
- Patches for bug fixes and enhancements for existing features.

migruiz4:
Regarding your comment, the reason to keep both `extraConfig` and `extraConfigYaml` is mostly to not introduce a non-backward compatible change, and therefore force a new major.

Additionally, with the plaintext `extraConfig` you could use templates at a higher level, for example:

```
extraConfig: |
   {{- $iterator := list "a" "b" "c" }}
   {{- range $i := $iterator }}
   kafka.setting.$i=false
   {{-end}}
   {{- if .Values.metrics.enabled }}
   ...
   {{- else }}
   ...
   {{- end }}
```

Can not ensure anyone is using the value `extraConfig` at that level, but by replacing it we would be removing this feature and forcing users to perform manual action to migrate to the new YAML object format.

migruiz4 (approved):

PR #26461, opened 2024-05-27
[bitnami/mlflow] added support for google storage

### LINKED ISSUE #22720: [bitnami/mlflow] MLFlow: Support additional artifact storage providers beyond S3
### Name and Version

bitnami/mlflow 0.6.0

The MLFlow chart (and docker image?) do not support using google cloud storage (or Microsoft cloud storage) for storing artifacts; while MLFlow itself does support it. 

Supporting this will make it easer to integrate MLFlow artifacts into the cloud provider storage.

We are coming from the https://community-charts.github.io/helm-charts mlflow helm chart, which does have support for it (but is otherwise outdated). 

What I propose is that we extend the [Helm chart](https://github.com/bitnami/charts/tree/main/bitnami/mlflow) configuration to allow the user to choose a cloud storage type, and corresponding configuration. 

I think that the [docker image](https://github.com/bitnami/containers/tree/main/bitnami/mlflow) can just be adjusted to also install the `google-cloud-storage` `azure-storage-blob` and `azure-identity` packages.  I cannot see the build details, and so cannot see if these are not already installed; but I guess they are not..

See: https://mlflow.org/docs/latest/tracking/artifacts-stores.html#google-cloud-storage

The alternative that I have implemented for ourselves right now is to use the XML HMAC api of google cloud, to make it compatible with MLFlow; and it works; but it would be easier for many people if support is added to MLFlow.

https://github.com/bitnami/charts/pull/22699

dhrp:
### Description of the change

This adds a section to the MLFlow template to configure Google Cloud Storage, and changes to the deployment template to make it support GCS. -- A user has two options: To use a secret for storing their application credentials, or to not configure application credentials; and use workload identity.

### Benefits

This adds support for natively storing artifacts with Google Cloud Storage. It is a feature that MLflow supports but didn't have a way to configure. 

### Possible drawbacks

None really; more configuration? -- I do see that; had GCS support been envisioned from day one configuration may have been structured slightly different. 

### Applicable issues

- fixes #22720
- Depends on: https://github.com/bitnami/containers/pull/67246

### Additional information

Things to double check:

- [x] do any init container scripts need to be modified? -- mostly they are related to minio
- [x] is there additional documentation needed anywhere?

### Checklist


bitnami/mlflow/templates/tracking/deployment.yaml
```diff
@@ -190,6 +195,16 @@ spec:
             - name: MLFLOW_S3_ENDPOINT_URL
               value: {{ printf "%s://%s:%v" (include "mlflow.v0.s3.protocol" .) (include "mlflow.v0.s3.host" .) (include "mlflow.v0.s3.port" .) | quote }}
             {{- end }}
+            {{- if .Values.externalGCS.enabled }}
+            {{- if .Values.externalGCS.useCredentialsInSecret }}
```
fmulero, line 199: I see a gap here, What happens if `externalGCS.enabled=true` and `externalGCS.useCredentialsInSecret=false`? Should users set additional env vars? Should users set additional permissions/roles to the serviceAccount created for mlflow if you are running the chart on GKE cluster?

fmulero (changes requested):
Thanks a lot @dhrp for your contribution I see it very useful.

About the dependency with `google-cloud-storage` I see you've opened a PR to include it in the container. Not sure if that PR will be accepted but I think we should advise in the chart about this, to let users know that they will need to extend this image. Anyway I would wait for the resolution of that PR.

I also left some comments related to the implementation, Could you review them?

bitnami/mlflow/templates/tracking/deployment.yaml
dhrp: on GCS one common approach is to use Workload Identity. -- What you then do is annotate the serviceaccount with the Google ServiceAccount you'd like to use (this is already supported by the chart), and grant that serviceaccount access to the bucket. It all then "just works" no secrets required.

dhrp:
Ah; I just noticed that the changes to the Docker Image supposedly have been released! 

See: https://github.com/bitnami/containers/pull/67246

I will test this shortly and update this PR to make it merge-able when I can confirm everything works.

SachinMaharana:
> Ah; I just noticed that the changes to the Docker Image supposedly have been released!
> 
> See: [bitnami/containers#67246](https://github.com/bitnami/containers/pull/67246)
> 
> I will test this shortly and update this PR to make it merge-able when I can confirm everything works.

We are using some custom stuff to support the gcs(passing args to containers for example), but would like to use this instead . If any more work is to be done or pending, i would be willing to contribute.

jotamartos (changes requested):
Sorry for the late review. Please take a look at my comments and fix the conflicts

nmauti-malt:
Hello,

I'm currently looking for support of GCS in this Helm Chart and I think this PR is exactly what I'm looking for (especially now that the Docker image include `google-storage`). However this PR seems inactive... What is the best way to contribute / reopen it @jotamartos? I can reopen another MR with @dhrp changes and reviewing your comments if it's ok for you?

Thanks,

fmulero:
Yes, feel free to open a new PR. You can start from this point, using the @dhrp changes, we will be happy to review it and provide feedback. [Here](https://github.com/bitnami/charts/blob/master/CONTRIBUTING.md) you can find the contributing guidelines.

Issue #23959, opened 2024-02-28
[bitnami/mlflow] Artifacts Issue  in UI - botocore.errorfactory.NoSuchKey: An error occurred (NoSuchKey) when calling the ListObjectsV2 operation: The specified key does not exist.


iamhritik290799:
### Name and Version

bitnami/mlflow and 0.7.4

### What architecture are you using?

amd64

### What steps will reproduce the bug?

We can simply use the bitnami/mlflow helm chart version 0.7.4 and provide the below s3 details like this 
```
externalS3.bucket = mlflow-bucket-name
externalS3.useCredentialsInSecret = false
externalS3.host = mlflow-bucket-name.s3.eu-central-1.amazonaws.com

externalS3.existingSecretAccessKeyIDKey = root-user
externalS3.existingSecretKeySecretKey = root-password
externalS3.port = 443
externalS3.protocol = https
externalS3.serveArtifacts = true
```
### Are you using any custom parameters or values?

Nope

### What is the expected behavior?

Able to see artifacts in the mlflow UI Console

### What do you see instead?

Getting this error in the logs and it's just showing Loading in the UI screen.

```
2024/02/28 09:46:35 ERROR mlflow.server: Exception on /ajax-api/2.0/mlflow/artifacts/list [GET]
Traceback (most recent call last):
  File "/opt/bitnami/python/lib/python3.10/site-packages/flask/app.py", line 1463, in wsgi_app
    response = self.full_dispatch_request()
  File "/opt/bitnami/python/lib/python3.10/site-packages/flask/app.py", line 872, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/opt/bitnami/python/lib/python3.10/site-packages/flask/app.py", line 870, in full_dispatch_request
    rv = self.dispatch_request()
  File "/opt/bitnami/python/lib/python3.10/site-packages/flask/app.py", line 855, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)  # type: ignore[no-any-return]
  File "/opt/bitnami/python/lib/python3.10/site-packages/mlflow/server/handlers.py", line 494, in wrapper
    return func(*args, **kwargs)
  File "/opt/bitnami/python/lib/python3.10/site-packages/mlflow/server/handlers.py", line 535, in wrapper
    return func(*args, **kwargs)
  File "/opt/bitnami/python/lib/python3.10/site-packages/mlflow/server/handlers.py", line 953, in _list_artifacts
[...]

PR #25294, opened 2024-04-21
[bitnami/mlflow] - make MLFLOW_S3_ENDPOINT_URL env option in tracking mlflow container

iamhritik:
### Description of the change
This PR modifies the mlflow tracking deployment template to allow setting `MLFLOW_S3_ENDPOINT_URL` as an optional environment variable. Previously, this variable was mandatory, causing issues in the code and resulting in errors when loading artifacts in the mlflow UI.

### Benefits

- Users can now optionally set `MLFLOW_S3_ENDPOINT_URL` in tracking.extraEnvVars, providing more flexibility in configuring the mlflow deployment.
- Resolves the issue where the presence of `MLFLOW_S3_ENDPOINT_URL` caused errors when loading artifacts in the mlflow UI.

### Possible drawbacks
- No known drawbacks at this time.

### Applicable issues
- fixes #23959

### Additional information
No additional information

PR #26462, opened 2024-05-27
[bitnami/mlflow] fix issue 26397 - allow MLFLOW_S3_ENDPOINT without minio

### LINKED ISSUE #26397: [bitnami/mlflow] MLFLOW_S3_ENDPOINT_URL not inject if i dont want to use minio
### Name and Version

mlflow/1.2.0

None

it was working fine before and today I am upgrading my mlflow to the latest mlfow 2.13.0. 

In my values.yaml
I need to set minio to false due to company policy. 
```yaml
minio:
  ## @param minio.enabled Enable/disable MinIO&reg; chart installation
  ## to be used as an objstore for Mlflow
  ##
  enabled: false
```

But I have my custom s3 host,  e.g my company provided s3-compatible host:
```yaml
externalS3:
  host: "objectstore.xyz.com"
  port: 443
  ...
```

In this [PR](https://github.com/bitnami/charts/pull/25294/files) , it introduced that i must enable minio otherwise my custom value won't able to inject. 

I have to temporary remove the PR introduced condition.

_No response_

_No response_

### What do you see instead?

NA. 

### Additional information

_No response_

dhrp:
### Description of the change

Changes the condition to check whether either Minio OR externalS3 is enabled.

- [x] Double check if we shouldn't use the condition in _helpers.tpl

[edit] moved setting the variable to inside the already existing check if s3 is enabled. This simplifies it.

### Benefits

This fixes a problem / bug in the chart where an external S3 address cannot be set unless Minio is also enabled; which is unwanted / undesirable.

### Possible drawbacks

none really

### Applicable issues

- fixes #26397

### Checklist


andresbono:
Hi @dhrp, thank you very much for contributing to this Helm chart.

This PR is reverting the changes introduced in #25294, sent on an attempt to fix #23959. Wouldn't merging your proposed changes cause regressions for the users affected by #23959?

In that issue, we came to the conclusion that `MLFLOW_S3_ENDPOINT_URL` was only required when MinIO was enabled (maybe that conclusion is wrong?) and users requiring to set the envar in some other scenarios would always be able to set it via `extraEnvVars` like this:

```console
$ helm template oci://registry-1.docker.io/bitnamicharts/mlflow --version 1.2.2 --set minio.enabled=false --set tracking.extraEnvVars[0].name=MLFLOW_S3_ENDPOINT_URL,tracking
.extraEnvVars[0].value='https://my-s3-endpoint:443' | grep -A1 MLFLOW_S3_ENDPOINT_URL
            - name: MLFLOW_S3_ENDPOINT_URL
              value: https://my-s3-endpoint:443
```

Can you share your thoughts on this? Thank you!

heinzelmannm (approved):
LGTM

bitnami/mlflow/templates/tracking/deployment.yaml
```diff
@@ -186,10 +188,6 @@ spec:
                   key: {{ include "mlflow.v0.s3.secretAccessKeyKey" . | quote }}
             {{- end }}
             {{- end }}
-            {{- if .Values.minio.enabled }}
```
frittentheke, line 189: Maybe not removing the condition (see side effects https://github.com/bitnami/charts/pull/26462#issuecomment-2146834438), but additionally checking if e.g. `externalS3.host` is defined. That would clearly indicate an external S3 is configured and the URL needs to be set. Independently from minIO being enabled.

dhrp: I see; and you are completely right. In fact, I had first fixed it in that way; and then "simplified" it; reintroducing the error.

dhrp:
@frittentheke was right; I had actually first fixed it right; and then simplified it wrong. When S3 is enabled and Minio is not, the host does not need to be set when you are using Amazon AWS S3

bitnami/mlflow/templates/tracking/deployment.yaml
andresbono: The new condition is equivalent to `{{- if (include "mlflow.v0.s3.enabled" .) }}`, a helper function: https://github.com/bitnami/charts/blob/1da77913298e55e7e06228deb16712dac54564fa/bitnami/mlflow/templates/_helpers.tpl#L554-L561

Maybe we can use that one instead for clarity.

---

However, I'm still missing something, the actual condition that makes `MLFLOW_S3_ENDPOINT` necessary to be set. Wouldn't this PR be a regression for https://github.com/bitnami/charts/issues/23959#issuecomment-2063750385?

andresbono:
I left a suggestion and a concern as an inline comment. Please, check it when you have a chance! Thank you.

bitnami/mlflow/templates/tracking/deployment.yaml
frittentheke: @andresbono `MLFLOW_S3_ENDPOINT` is relevant if a different S3 service from AWS is used.
This could either be minIO as provided by the chart or some other S3-comptible service. And the latter is what is currently not configurable.

As for the helper you are correct .. that could simply be used as conditional to keep the code DRY.

frittentheke:
> I left a suggestion and a concern as an inline comment. Please, check it when you have a chance! Thank you.

I replied  in the thread @andresbono .

Would be nice to get the ability to use external non-AWS S3.

bitnami/mlflow/templates/tracking/deployment.yaml
dhrp: ok; I changed the logic to have use the helper

dhrp:
I've resolved the comment regarding the helper and rebased on main.

andresbono:
Thank you, @dhrp, @frittentheke. I reopened https://github.com/bitnami/charts/pull/26462#discussion_r1650708286. Sharing my concerns there.

bitnami/mlflow/templates/tracking/deployment.yaml
frittentheke: @andresbono Let's first look at what MLflow actually documents and uses to configure S3:

* https://mlflow.org/docs/latest/tracking/artifacts-stores.html#supported-storage-types-for-the-artifact-store
* https://mlflow.org/docs/latest/python_api/mlflow.environment_variables.html#mlflow.environment_variables.MLFLOW_S3_ENDPOINT_URL

Since they re using the boto3 library, it's not required to set an endpoint in case "original" AWS S3 is used, yes.
It's possible, see https://docs.aws.amazon.com/general/latest/gr/s3.html, to define regional endpoints, but it's again not required for AWS S3.

But with the current state of the PR there is a conditional asing if either `.Values.minio.enabled` **OR**  `.Values.externalS3.host` has anything but an empty string. If so an ENV var `MLFLOW_S3_ENDPOINT` is defined.

As for your question, you have me a little confused. If **neither** `minio.enabled`, not a distinct `externalS3.host` is set no ENV var `MLFLOW_S3_ENDPOINT_URL` is filled, having boto3 apply it's default, which is AWS and their S3 endpoints.

dhrp: That's right. AWS S3 does not need a region set (AWS is terrible in properly explaining this though but I double checked) -- AWS bucket names are globally unique and boto actually uses the hostname e.g. bucket hostname e.g. `bar.s3.amazonaws.com` to lookup what to connect to. There may be exceptions where an endpoint needs te set (GovCloud), but then `MLFLOW_S3_ENDPOINT` should suffice. 

See amongst others this issue: https://github.com/aws/aws-cli/issues/646

andresbono:
Thank you for the additional information. However, I still have the concern that this PR will introduce a regression and we will face #23959 again. I'm not saying we do not have to address #26397, but I'm not sure if this PR is the proper fix.

I gave this PR a try. Let me share my findings:

<details>
<summary><b>Scenario 1: Store artifacts in external S3-compatible storage (MinIO) ✅</b></summary>

To take the minio subchart out of the equation, I'm deactivating it and explicitly installing it with hardcoded credentials.

1. Install MinIO

`values-minio.yaml` customizations:
```yaml
auth:
  rootPassword: somepassword
defaultBuckets: test-mlflow
provisioning:
  enabled: true
  extraCommands:
    - mc anonymous set download provisioning/test-mlflow
```

```console
$ helm install minio oci://registry-1.docker.io/bitnamicharts/minio --version 14.6.16 --values /tmp/values-minio.yaml
```

2. Install mlflow (with the changes from this PR)
`values-mlflow-minio.yaml` customizations:
```yaml
run:
  source:
    configMap:
      # Example taken from the MLFlow UI
      test.py: |
        import mlflow
        from sklearn.model_selection import train_test_split
        from sklearn.datasets import load_diabetes
        import sklearn.ensemble

        # set the experiment id
        mlflow.set_experiment(experiment_id="0")

        mlflow.autolog()
        db = load_diabetes()

        X_train, X_test, y_train, y_test = train_test_split(db.data, db.target)

        # Create and train models.
        rf = sklearn.ensemble.RandomForestRegressor(n_estimators=100, max_depth=6, max_features=3)
        rf.fit(X_train, y_train)

        # Use the model to make predictions on the test dataset.
        predictions = rf.predict(X_test)

minio:
  enabled: false

externalS3:
  host: minio.default.svc.cluster.local
  port: 9000
  protocol: http
  bucket: test-mlflow
  accessKeyID: admin
  accessKeySecret: somepassword
```

```console
$ helm install mlflow-minio ./bitnami/mlflow --values /tmp/values-mlflow-minio.yaml
```

3. Create experiment run

```console
$ kubectl exec -it deployment/mlflow-minio-run -- python test.py
2024/07/01 14:19:27 INFO mlflow.utils.autologging_utils: Created MLflow autologging run with ID '7fe8ec68831b4eeea1680fe52753d45a', which will track hyperparameters, performance metrics, model artifacts, and lineage information for the current sklearn workflow
```

4. Check artifacts in MinIO ✅

```
$ kubectl exec -it deployment/minio -- mc ls local/test-mlflow/0
[2024-07-01 14:25:36 UTC]     0B 7fe8ec68831b4eeea1680fe52753d45a/
```

5. Check artifacts via UI ✅

```
$ kubectl get secret --namespace default mlflow-minio-tracking -o jsonpath="{.data.admin-password }" | base64 -d
$ kubectl port-forward svc/mlflow-minio-tracking 8080:80
```

http://localhost:8080/#/experiments/0/runs/7fe8ec68831b4eeea1680fe52753d45a/artifacts

image

6. Check mlflow logs ✅

```
$ kubectl logs deploy/mlflow-minio-tracking
```
</details>

<details>
  <summary><b>Scenario 2: Store artifacts in AWS S3 ❌</b></summary>

1. Create an AWS S3 bucket

```
test-mlflow-ab891a3
```

2. Install mlflow (with the changes from this PR)

`values-mlflow-s3.yaml` customizations:
```yaml
run:
  source:
    configMap:
      # Example taken from the MLFlow UI
      test.py: |
        import mlflow
        from sklearn.model_selection import train_test_split
        from sklearn.datasets import load_diabetes
        import sklearn.ensemble

        # set the experiment id
        mlflow.set_experiment(experiment_id="0")

        mlflow.autolog()
        db = load_diabetes()

        X_train, X_test, y_train, y_test = train_test_split(db.data, db.target)

        # Create and train models.
        rf = sklearn.ensemble.RandomForestRegressor(n_estimators=100, max_depth=6, max_features=3)
        rf.fit(X_train, y_train)

        # Use the model to make predictions on the test dataset.
        predictions = rf.predict(X_test)

minio:
  enabled: false

externalS3:
  host: test-mlflow-ab891a3.s3.amazonaws.com
  bucket: test-mlflow-ab891a3
  useCredentialsInSecret: false

tracking:
  extraEnvVars:
    - name: AWS_ACCESS_KEY_ID
      value: ASIA***
    - name: AWS_SECRET_ACCESS_KEY
      value: '***'
    - name: AWS_SESSION_TOKEN
      value: '***'
```

```console
$ helm install mlflow-s3 ./bitnami/mlflow --values /tmp/values-mlflow-s3.yaml
```

3. Create experiment run

```console
$ kubectl exec -it deployment/mlflow-s3-run -- python test.py
2024/07/01 15:12:56 INFO mlflow.utils.autologging_utils: Created MLflow autologging run with ID 'f9e2c06af11b4ff787ee400e5912d75b', which will track hyperparameters, performance metrics, model artifacts, and lineage information for the current sklearn workflow
```

4. Check artifacts in S3 ⚠️ 

```
$ aws s3 ls s3://test-mlflow-ab891a3/test-mlflow-ab891a3/0/
                           PRE f9e2c06af11b4ff787ee400e5912d75b/
```

They exist but the objects are uploaded in a root directory named like the bucket.

4. Check artifacts via UI ❌

```
$ kubectl get secret --namespace default mlflow-s3-tracking -o jsonpath="{.data.admin-password }" | base64 -d
$ kubectl port-forward svc/mlflow-s3-tracking 8080:80
```

http://localhost:8080/#/experiments/0/runs/f9e2c06af11b4ff787ee400e5912d75b/artifacts

image

Artifacts cannot be retrieved.

6. Check mlflow logs ❌

```
$ kubectl logs deploy/mlflow-s3-tracking
2024/07/01 15:17:11 ERROR mlflow.server: Exception on /ajax-api/2.0/mlflow/artifacts/list [GET]
Traceback (most recent call last):
  File "/opt/bitnami/python/lib/python3.10/site-packages/flask/app.py", line 1473, in wsgi_app
    response = self.full_dispatch_request()
...
  File "/opt/bitnami/python/lib/python3.10/site-packages/botocore/client.py", line 1021, in _make_api_call
    raise error_class(parsed_response, operation_name)
botocore.errorfactory.NoSuchKey: An error occurred (NoSuchKey) when calling the ListObjectsV2 operation: The specified key does not exist.
```

</details>

I guess the error in "Scenario 2" is raised because the objects are not written in the correct path (the root directory named like the bucket I believe). The documentation warns about this: https://mlflow.org/docs/2.14.1/tracking/artifacts-stores.html#setting-bucket-region.

With `bitnami/mlflow:1.4.12`, "Scenario 2" is working as expected and "Scenario 1" would just need to explicitly set the `MLFLOW_S3_ENDPOINT_URL` env-var:

```yaml
tracking:
  extraEnvVars:
    - name: MLFLOW_S3_ENDPOINT_URL
      value: 'http://minio.default.svc.cluster.local:9000'
```

Let me know your thoughts on this @dhrp, @frittentheke.

frittentheke:
@andresbono @javsalgar Could you kindly check if my explanation (https://github.com/bitnami/charts/pull/26462#issuecomment-2209054770) holds true and this PR could move forward?

andresbono:
That made the trick, @frittentheke. Thank you so much for looking into it.

For external AWS S3, `externalS3.host=s3.amazonaws.com` works flawlessly, there is no need to include the bucket name in the host (virtual-hosted–style) as it leads to a malformed endpoint uri.

> Edit: Maybe it makes sense to also expose the `MLFLOW_BOTO_CLIENT_ADDRESSING_STYLE` (https://github.com/mlflow/mlflow/blob/0ebbefc8c075165b558dfb481cf1e39d5cb53003/mlflow/environment_variables.py#L605) variable via the Helm chart / values.yaml, to make the externalS3 configuration as flexible as mlFlow itself? @dhrp ?

A user could pass the `MLFLOW_BOTO_CLIENT_ADDRESSING_STYLE` via the `tracking.extraEnvVars` value if needed.

andresbono (dismissed):
The PR is ready to go. Thank you so much for the contribution! 🎉

frittentheke:
@andresbono should this auto-merge or CI are the failures expected?

andresbono:
> @andresbono should this auto-merge or CI are the failures expected?

No, the CI failures are not expected. Let me look into it.

andresbono (approved):

andresbono:
Finally merged!! Thank you so much for being collaborative and digging into our suggestions and questions!!

PR #26600, opened 2024-06-03
[bitnami/grafana-loki] Remove unnecessary memcached image definitions

carrodher:
We are setting the memcached image in the _values.yaml_ overriding the one present in the subchart. Since the version set in the _values.yaml_ and the one used in the subchart are the same, there is no need to do this


fmulero (approved):
LGTM

PR #26603, opened 2024-06-03
[bitnami/grafana-loki] Don't enforce version label when it's not needed

carrodher:
We are [enforcing version label](https://github.com/bitnami/charts/blob/f9566f39a9fc443c2f228434efb2c94e24e2e980/bitnami/grafana-loki/templates/compactor/deployment.yaml#L12-L16) when it's not needed because the image is the same one everywhere and it is the one used for the appVersion in the Chart definition (`.Values.loki.image`).

To be merged after https://github.com/bitnami/charts/pull/26600


juan131 (approved):
LGTM

PR #27080, opened 2024-06-10
[bitnami/redis] Fix sentinel PDB condition

fmulero:
### Description of the change

Fix mistake in the condition what enables the PodDisruptionBudgets for sentinel.

### Benefits

Remove unnecessary PDB created when sentinel is not enabled.

### Possible drawbacks

None

### Applicable issues

- Follow up #26355

### Checklist

- [NA] Variables are documented in the values.yaml and added to the `README.md` using [readme-generator-for-helm](https://github.com/bitnami/readme-generator-for-helm)


bitnami/redis/templates/sentinel/pdb.yaml
```diff
@@ -3,7 +3,7 @@ Copyright Broadcom, Inc. All Rights Reserved.
 SPDX-License-Identifier: APACHE-2.0
 */}}
 
-{{- if and (eq .Values.architecture "replication") (not .Values.sentinel.enabled) .Values.replica.pdb.create }}
+{{- if and (eq .Values.architecture "replication") .Values.sentinel.enabled .Values.replica.pdb.create }}
```
wouterh, line 6: For backwards compatibility, this needs to be:

```
{{- $pdb := coalesce .Values.replica.pdb .Values.pdb }}
{{- if and (eq .Values.architecture "replication") .Values.sentinel.enabled $pdb.create }}
```

Chart version 19.3.0 used to create a PDB for sentinel nodes based on `.Values.pdb`.

wouterh (commented):
I believe there's something wrong with the condition for creating the master PDB too. In my configuration, it created a PDB for the masters while the master statefulset is not created:

```bash
$ kubectl -n redis get pdb -o jsonpath="{.items[].metadata.name} {.items[].spec.selector} {.items[].status.expectedPods}"
redis-master {"matchLabels":{"app.kubernetes.io/component":"master","app.kubernetes.io/instance":"redis","app.kubernetes.io/name":"redis"}} 0

$ kubectl -n redis get sts
NAME         READY   AGE
redis-node   3/3     472d
```

bitnami/redis/templates/sentinel/pdb.yaml
fmulero: Version 19.3.0 was using the same PDB for all pods created by this helm chart, without any distinction between replicas, master or sentinel pods, what was a failure. Anyway I've just changed the templates to add backwards compatibility

bitnami/redis/templates/master/pdb.yaml
```diff
@@ -2,8 +2,8 @@
 Copyright Broadcom, Inc. All Rights Reserved.
 SPDX-License-Identifier: APACHE-2.0
 */}}
-
-{{- if and .Values.master.pdb.create (gt (int64 .Values.master.count) 0) (or (eq .Values.architecture "replication") (not .Values.sentinel.enabled)) }}
+{{- $pdb := coalesce .Values.pdb .Values.master.pdb }}
```
juan131, line 5: Shouldn't it be the other way around? I mean, in this way `.Values.pdb` takes precedence over `.Values.master.pdb`?

juan131 (changes requested):

bitnami/redis/templates/master/pdb.yaml
fmulero: Good question. With current change If we have a values file coming from an older version with `pdb` defined, this will be used. If we call the coalesce function in the opposite way, the default values will be always applied and the `pdb` value will be ignored

juan131: Newer values should take precedence over older ones. That's what we always do with deprecated values.

It's true that the values we've deprecated before were usually unset by default. That said, it's the user responsibility to update their values according the parameters available in newer versions and it would be very unusual to give priority to parameter that no longer exists.

fmulero: I agree with your idea in general but in this particular case your approach won't work, because `master.pdb` is set to an object by default and this will take precedence over `pdb` unless the user explicitly remove that object in the values.

juan131: ahh! Good point, it makes sense

juan131 (approved):

PR #27434, opened 2024-06-18
[bitnami/external-dns]  Fix crd.create=true

gnom4ik:
CRD doesn't work, this solving the problem )

### Description of the change

Added annotation that 
### Benefits

Helm chart will start work if crd.create=true

### Possible drawbacks
---

### Applicable issues

- fixes #

### Additional information

This correction adds an annotation to the external-dns CRD. There are no other changes. The consequence of the change is that the chart will start working correctly (currently, enabling the CRD does not result in installation due to the absence of the annotation)."

### Checklist


fmulero (commented):
Hi @gnom4ik, Thanks a lot for your contribution.

Why is that annotation required? The upstream [CRDs](https://github.com/kubernetes-sigs/external-dns/blob/v0.14.2/docs/contributing/crd-source/crd-manifest.yaml) don't have that annotation.  If the annotation is required beacuase external-dns is using any private/reserved API, it should be communicated to the upstream project to solve the issue.

fmulero:
I've just seen a related [PR](https://github.com/kubernetes-sigs/external-dns/pull/4488) fixing this problem on external-dns side. Unfortunately there is no new releases with those changes. We should wait for the next release

jetersen:
Can we not have something that patches it and once the next release comes out it will be patched from their side.
Currently we been waiting over a month for a release from external-dns and several comments on the PR with no success: https://github.com/kubernetes-sigs/external-dns/pull/4488

fmulero:
We could land this change but CRDs are automatically updated and it could be reverted without any advice.

jetersen:
Can't we fix this in the bitnami bot? [`375ee3b` (#25962)](https://github.com/bitnami/charts/pull/25962/commits/375ee3ba1adfabacb8c8ccb29fb97b78c15e92c1)

juan131:
Hi @jetersen 

I'm afraid we cannot include patches or point to specific commits in our automations to sync changes. Until there's an upstream release, you can use [kustomize](https://kustomize.io) (or similar tools) to apply the changes in the Bitnami chart, sth like:

```bash
helm template external-dns oci://registry-1.docker.io/bitnamicharts/external-dns --set crd.create=true (...) | kubectl apply -k (...)
```

jetersen:
Not stale. Still relevant

juan131 (dismissed):

juan131 (approved):

pschichtel:
@juan131: Out of interest: Assuming that the next release will definitely include the upstream fix, wouldn't it work just fine including the the annotation now? The bot would update the CRD upon the next release, which would have the change anyway. Or will the bot occasionally verify consistency with the upstream and overwrite the change?

fmulero:
> @juan131: Out of interest: Assuming that the next release will definitely include the upstream fix, wouldn't it work just fine including the the annotation now? The bot would update the CRD upon the next release, which would have the change anyway. Or will the bot occasionally verify consistency with the upstream and overwrite the change?

With this [change](https://github.com/bitnami/charts/pull/27434/commits/d1c556deb6cea44a2981473f793fc4cb9046b60b) the bot won't know where is the source for that CRD and it won't try to update it. We should add the header again once external-dns is released

PR #27467, opened 2024-06-20
[bitnami/external-dns] Disable pod disruption budget per default value

bernermic:
### Description of the change

Disable pod disruption budget as default to avoid errors  with non evictable external-dns pod on node updates when there can be only one replica of external-dns.

### Benefits

Pod disruption budget does not hinder a node restart.

### Applicable issues

There is none yet. I observed this behavior in our cluster. Node could not be drained because of pdb on external-dns.

### Checklist

- [x] Variables are documented in the values.yaml and added to the `README.md`´


rafariossaa:
Hi,
  I am sorry but in all our catalog, by default, we enable this parameter. So, I would like to keep this coherent with the rest of the catalog.
  This parameter can be easily set to the desired value when deploying the chart.

PR #27490, opened 2024-06-21
[bitnami/rabbitmq] Add instructions for using different rmq versions

andrelegault:
### Description of the change

Add instructions for using different rmq versions.

### Benefits

More people are aware that the version of rabbitmq is customizable.

### Possible drawbacks

Different image tags are used

### Checklist


carrodher:
Thank you for taking the time to create this PR.

Each Helm chart functions as a package that is tested with a specific combination of source code (Helm chart templates) and container images. Deviating from this tested combination may lead to unpredictable results, which we cannot guarantee. While the workaround you proposed might work, it is not recommended as it has not been tested. 

Including this workaround in the README would imply it has official support, so we cannot accept this contribution at this time.

PR #27534, opened 2024-06-26
[bitnami/kafka] Add custom port for advertised listeners

asazanoff:
### Description of the change

Add custom port config to be added in advertised.listeners with custom external access domain param

### Benefits

If you have Kafka behind reverse proxy it should tell clients not internal nodeport but public port that had been assigned with reverse proxy configuration.

### Possible drawbacks

I dont see limitations now

### Applicable issues

No issues found

### Checklist


bitnami/kafka/Chart.yaml
```diff
@@ -40,4 +40,4 @@ maintainers:
 name: kafka
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/kafka
-version: 29.3.4
+version: 29.3.5
```
fmulero, line 43: In this PR a new feature is being added, according to [semver](https://semver.org) a minor version should be bumped.

```suggestion
version: 29.4.0
```

fmulero (commented):
Thanks a lot @asazanoff for you contribution.

I am trying to understand the scenario where this change could be useful and I don't see it clearly. Why don't the values `...service.ports.external`  fit your needs? Don't you need a service exposing those ports or a networkpolicy to manage the traffic?

bitnami/kafka/Chart.yaml
asazanoff: @fmulero in `externalAccess.controller/broker.service.ports.external` we set the port Kafka is listening inside pod. This parameter is applicable when service type is LoadBalancer, but I need NodePort.

BTW when I set domain `cluster.example.com`  in `externalAccess.controller.service.domain` and  `33333` in `externalAccess.controller.service.nodePorts`, file `server.properties` has `cluster.example.com:33333` as external listener address. It is OK when you exposing Kafka external port directly from cluster.

In my case I have reverse proxy outside cluster that is forwarding traffic from `example.com:9095` to `cluster.example.com:33333`. For domain it is simple, just change `externalAccess.controller.service.domain` to `example.com`. But port stais as 33333. With this PR I can set port in  `server.properties` for domain so it can accept port `9095`, and in `server.properties` it shows as `example.com:9095` as expected.

Also bumped version and fixed typo

fmulero: I am not sure if I am missing something but in your example after the changes you exposed, `cluster.example.com:33333` won't work. I mean if you set this configuration (please amend me if I am wrong):
```yaml
externalAccess:
  enabled: true
  controller:
    service:
      type: NodePort
      nodePorts: 
        - 32333
        - 32334
        - 32335
      domain: "mykafka.example.com"
      advertisedExternalPort: "9095"
```
Your NodePort service won't work on port 32333. Is your reverse proxy redirecting the traffic internally? using the internal service instead of the NodePort? Sorry for being so picky but I need to understand the scenario and your changes because once these changes are landed we have to maintain it.

asazanoff: Hi!

Just tried with values you provided, and it creates three nodeport services as expected with ports 32333-32335. It is correct. 

```yaml
apiVersion: v1
kind: Service
metadata:
  name: kafka-controller-1-external
  namespace: "my-namespace"
  labels:
    app.kubernetes.io/instance: kafka
    app.kubernetes.io/name: kafka
    app.kubernetes.io/version: 3.7.0
    helm.sh/chart: kafka-29.3.4
    app.kubernetes.io/component: kafka
    pod: kafka-controller-1
spec:
  type: NodePort
  publishNotReadyAddresses: false
  ports:
    - name: tcp-kafka
      port: 9094
      nodePort: 32333
      targetPort: external
  selector:
    app.kubernetes.io/instance: kafka
    app.kubernetes.io/name: kafka
    app.kubernetes.io/part-of: kafka
    app.kubernetes.io/component: controller-eligible
    statefulset.kubernetes.io/pod-name: kafka-controller-1
```

Connecting to <node-address>:32333 is successfull, so nothing changed here.

Then, I have a reverse-proxy, for example nginx. This reserse proxy is external, lives as standalone installation, for example in docker-compose or as linux binary, it doesnt matter.  The thing we should know is that we want to have an access to Kafka with reverse-proxy on address mykafka.example.com:9095. Reverse-proxy accepts traffic from mykafka.example.com:9095 and translates it to <node-address>:32333.

With this Kafka also should know that it is getting traffic not only from <node-address>:32333, but it should know about mykafka.example.com:9095, because it is an address that kafka clients use. Domain could be set with its own `externalAccess.controller.service.domain` parameter, but port now stais as 32333, which is not suitable for us, because kafka should know the external 9095 port which clients use.

Hope I could tell you reasons of changes clearly

asazanoff: @fmulero Hi! Any news about this PR? just updated branch with new commits in original repo

asazanoff: @fmulero ping

fmulero: Sorry about my very late response, I was out of office.

If I understood well, communications are working because all clients are routed through the reverse proxy. If your client tries to connect to port 32333, it will be redirected to the advertised host and port. 

To be perfectly blunt I have mixed feelings. I see this feature useful if we open a little the scope, and the advertised port can be set for ClusterIP and LoadBalancer type services, but right now I see it like a very particular solution. (I have to say that I don't see any other solution if you are forced to use NodePort services).

PR #27536, opened 2024-06-26
[github-actions] Add Kubescape score action

migruiz4:
### Description of the change

This additional CI action checks that contributions do not reduce the Kubescape score.

Tested at https://github.com/migruiz4/charts/pull/19

### Additional information

* The action obtains the score of the affected chart from the `main` branch using the bitnami/kubescape image.
* The score obtained is truncated to two decimal digits precision.
* Then, the score is set as `threshold` for the VIB kubescape action, which will return not passed if the obtained score is lower than the threshold score.

### Checklist

### DIFF
--- a/.github/workflows/ci-pipeline.yml
+++ b/.github/workflows/ci-pipeline.yml
@@ -305,16 +305,13 @@ jobs:
           fi
   chart-score:
     runs-on: ubuntu-latest
-    needs: [get-chart, update-pr]
+    needs: [get-chart]
     name: Get chart latest Kubescape score
+    permissions:
+      contents: read
     outputs:
       threshold: ${{ steps.set-output.outputs.threshold }}
-    if: |
-      needs.get-chart.outputs.result == 'ok' &&
-      needs.update-pr.outputs.result == 'skip' &&
-      (
-        contains(github.event.pull_request.labels.*.name, 'verify') || (github.event.action == 'labeled' && github.event.label.name == 'verify')
-      )
+    if: needs.get-chart.outputs.result == 'ok'
     steps:
       - uses: actions/checkout@a5ac7e51b41094c92402da3b24376905380afc29
         name: Checkout 'main' branch
--- a/.github/workflows/ci-pipeline.yml
+++ b/.github/workflows/ci-pipeline.yml
@@ -303,9 +303,63 @@ jobs:
           else
             echo "result=skip" >> $GITHUB_OUTPUT
           fi
-  vib-verify:
+  chart-score:
     runs-on: ubuntu-latest
     needs: [get-chart, update-pr]
+    name: Get chart latest Kubescape score
+    outputs:
+      threshold: ${{ steps.set-output.outputs.threshold }}
+    if: |
+      needs.get-chart.outputs.result == 'ok' &&
+      needs.update-pr.outputs.result == 'skip' &&
+      (
+        contains(github.event.pull_request.labels.*.name, 'verify') || (github.event.action == 'labeled' && github.event.label.name == 'verify')
+      )
+    steps:
+      - uses: actions/checkout@a5ac7e51b41094c92402da3b24376905380afc29
+        name: Checkout 'main' branch
+        with:
+          path: charts-main
+          repository: bitnami/charts
+          fetch-depth: 1
+      - name: Install helm
+        run: |
+          HELM_TARBALL="helm-v3.8.1-linux-amd64.tar.gz"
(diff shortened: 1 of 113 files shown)


.vib/airflow/vib-verify.json
```diff
@@ -57,6 +57,12 @@
               "password": "ComplicatedPassword123!4"
             }
           }
+        },
+        {
+          "action_id": "kubescape",
+          "params": {
+            "threshold": {VIB_ENV_KUBESCAPE_SCORE_THRESHOLD}
```
carrodher, line 64: Just for consistency, other env. vars are quoted
```suggestion
            "threshold": "{VIB_ENV_KUBESCAPE_SCORE_THRESHOLD}"
```

migruiz4: The missing quotes are in purpose because of variable types.

The `threshold` variable is of type `number`, therefore it requires to be set to:

```
            "threshold": 97.0
```

By adding the quotes, it switches to `string` type, causing an error:
```
            "threshold": "97.0"
```

migruiz4: Problem is most linters detect this as an error:
```
"threshold": {VIB_ENV_KUBESCAPE_SCORE_THRESHOLD}
```

But it is not a problem because string substitution happens during the `vmware-image-builder-action` step.

migruiz4: For instance, the result would look like this:

```
VIB_ENV_KUBESCAPE_SCORE_THRESHOLD=99.99
"threshold": {VIB_ENV_KUBESCAPE_SCORE_THRESHOLD} -> "threshold": 99.99 (type: number)
"threshold": "{VIB_ENV_KUBESCAPE_SCORE_THRESHOLD}" -> "threshold": "99.99" (type: string)
```

Adding the quotes results in the error `expected 'number' got 'string'`

.github/workflows/ci-pipeline.yml
```diff
@@ -303,9 +303,63 @@ jobs:
           else
             echo "result=skip" >> $GITHUB_OUTPUT
           fi
-  vib-verify:
+  chart-score:
     runs-on: ubuntu-latest
     needs: [get-chart, update-pr]
```
fmulero, line 308: Could we remove the `update-pr` dependency? We are calculating the score form main branch so I think we can run `update-pr` and `chart-score` in parallel

fmulero (commented):
Hi @migruiz4 

I left some comments, could you give them a glance?

.github/workflows/ci-pipeline.yml
```diff
@@ -303,9 +303,63 @@ jobs:
...
+    name: Get chart latest Kubescape score
+    outputs:
```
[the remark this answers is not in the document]
migruiz4: It wasn't needed in the test pull request

.github/workflows/ci-pipeline.yml
migruiz4: The reason why I chose to wait for `update-pr` is because the `chart-score` step does not make sense unless the `vib-verify` is executed right after, but I guess that by executing both in parallel we would be saving some extra time.

.github/workflows/ci-pipeline.yml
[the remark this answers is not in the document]
migruiz4: Maybe because repository is public, will add them in case they are required in other use cases

.github/workflows/ci-pipeline.yml
migruiz4: Updated

.github/workflows/ci-pipeline.yml
```diff
@@ -303,9 +303,63 @@ jobs:
...
+        uses: addnab/docker-run-action@v3
+        name: Get main branch kubescape score
+        env:
+          CHART: ${{ needs.get-chart.outputs.chart }}
+        with:
+          image: bitnami/kubescape:3.0.3
+          options: -v  /home/runner/work/charts/charts/charts-main:/charts -v /tmp:/out -e CHART
+          run: |
+            if [ -d "/charts/bitnami/${CHART}" ]; then
+              kubescape scan framework MITRE,NSA,SOC2,cis-v1.23-t1.0.1 /charts/bitnami/${CHART} --format json -o /out/report.json
```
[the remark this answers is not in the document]
fmulero: We can leave as it is, I don't think it changes very often.

fmulero (approved):
LGTM

PR #27585, opened 2024-06-28
[bitnami/argo-cd] Make it possible to run ArgoCD in HA mode

maxnitze:
### Description of the change

When the `replicaCount` of the application controller or the server deployments in the ArgoCD chart are set to a number higher than one, an additional env variable is set on the respective container `ARGOCD_CONTROLLER_REPLICAS` and `ARGOCD_API_SERVER_REPLICAS`.

In addition a change was added, that allows the application controller to be installed as a StatefulSet. I cannot find direct confirmation in the docs for it, but it seems the sharding of clusters, that is enabled when running with multiple application controller instances, is only working, when the pods are running as part of a stateful set. Otherwise they can't figure out their shard number and some clusters are not handled by any replica anymore (see here for example: https://github.com/argoproj/argo-cd/issues/17016).

The "official" Helm chart also installs the application controller as a StatefulSet (see https://github.com/argoproj/argo-helm/tree/main/charts/argo-cd/templates/argocd-application-controller). There are cases, where it is installed as a Deployment, but that's for a (currently in alpha) feature called [Dynamic Cluster Distribution](https://argo-cd.readthedocs.io/en/stable/operator-manual/dynamic-cluster-distribution/).

Also a validation was added, that multiple replicas of the application controller can only be set if the kind is changed to a StatefulSet (see "Benefits" for an explanation why).

### Benefits

As per [ArgoCD documentation on high availability](https://argo-cd.readthedocs.io/en/stable/operator-manual/high_availability/), both the application controller and the server should have an environment variable set to the number of replicas in case they are run with multiple.

The application controller uses this value to "shard clusters across multiple controller replicas". The server uses it to "divide [the limit of concurrent login requests (`ARGOCD_MAX_CONCURRENT_LOGIN_REQUESTS_COUNT`)](https://argo-cd.readthedocs.io/en/stable/operator-manual/user-management/#failed-logins-rate-limiting) between each replica." The repo-server has no such environment variable.

Especially on the application controller, if  the variable is not set, then the different replica  will concurrently try to do the same operations on the clusters. Which results in constantly out-of-sync applications and other issues with the apps.

### Possible drawbacks

None for the application itself.

### Is this change breaking?

There are some interesting cases though.

#### Installations with Multiple Replicas of the Application Controller

In case you already have an installation with multiple replicas of the application controller, the chart will fail to validate, if you have not set the `.Values.controller.kind` to `"StatefulSet"`. But in that case your application would not be running properly anyways. I had a similar setup for a week, or so, and the controller pods failed every couple of minutes due to memory issues created by the fact, that they concurrently worked on my clusters.

#### Env variable is set already in `extraEnvVars`

In case a user set the respective env variable already in the `extraEnvVars` block, this leads to a duplicate entry:

```yaml
spec:
  replicas: 3
  templates:
    spec:
      containers:
        - name: controller
          env:
            - name: ARGOCD_CONTROLLER_REPLICAS
              value: "3" # <-- from the .Values.controller.replicaCount
            # ...
            - name: ARGOCD_CONTROLLER_REPLICAS
              value: "3" # <-- from .Values.controller.extraEnvVars
```

This is not a problem, since according to the [documentation](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#environment-variables) the last occurrence of an environment variable takes precedence.

#### Env variable is set already in `extraEnvVarsCM` or `extraEnvVarsSecret`

This one is a bit more tricky. According to the docs, the environment variables from `env` take precedence over those from `envFrom`. So in theory this could overwrite a value set in the `extraEnvVarsCM` or `extraEnvVarsSecret`.

BUT: This is only an issue, if  those values differ. And since this value should be the same value as the number of replicas, this is a misconfiguration, from my perspective. I don't actually know what the implications of a wrong value are though. I could not find detailed info on the sharding of the application controller, but reading the linked docs I assume this could lead to clusters not being served by any of the replicas!

### Applicable issues

None

### Additional information

There are some issues and PRs in the ArgoCD repo talking about this:
- https://github.com/argoproj/argo-cd/issues/17124
- https://github.com/argoproj/argo-cd/pull/17167
- https://github.com/argoproj/argo-cd/issues/17016#issuecomment-1937079998
- 

### Checklist


maxnitze:
Converted this back to a draft. I ran into some major issues with the sharding of the application controller today. Apparently it only really works, when the controller is installed as a StatefulSet :(

bitnami/argo-cd/templates/server/deployment.yaml
```diff
@@ -169,6 +169,10 @@ spec:
               containerPort: {{ .Values.server.containerPorts.metrics }}
               protocol: TCP
           env:
+            {{- if gt .Values.server.replicaCount 1 }}
```
fmulero, line 172: Numbers coming from values files are considered floats.
```suggestion
            {{- if gt ( int .Values.server.replicaCount ) 1 }}
```

bitnami/argo-cd/templates/_helpers.tpl
```diff
@@ -256,6 +256,19 @@ Return the Redis&reg; port
 {{- end -}}
 {{- end -}}
 
+{{/*
+Validate Application Controller config
+*/}}
+{{- define "argocd.validateValues.controller.config" -}}
+{{- if gt .Values.controller.replicaCount 1 }}
```
fmulero, line 263: Numbers coming from values files are considered floats.
```suggestion
{{- if gt ( int .Values.controller.replicaCount ) 1 }}
```

bitnami/argo-cd/templates/application-controller/_pod.tpl
```diff
@@ -0,0 +1,262 @@
...
+  {{- end }}
+  {{- end }}
+  ports:
+    - name: metrics
+      containerPort: {{ .Values.controller.containerPorts.metrics }}
+      protocol: TCP
+  env:
+    - name: ARGOCD_APPLICATION_CONTROLLER_NAME
+      value: {{ include "argocd.application-controller" . }}
+    {{- if gt .Values.server.replicaCount 1 }}
```
fmulero, line 146:
```suggestion
    {{- if gt ( int .Values.server.replicaCount ) 1 }}
```

fmulero (changes requested):
Hi @maxnitze Thanks a lot for your contribution, this will be very useful for all users!

I left some comments, Could you give them a glance?

maxnitze:
Hey @fmulero ,

thanks for the review! Looking at the Pod partial I saw, that the whole indentation was wrong. In the file itself, but also the calls to `nindent` were not adapted accordingly. Fixed that as well as your other remarks :)

bitnami/argo-cd/templates/_helpers.tpl
```diff
@@ -256,6 +256,19 @@ Return the Redis&reg; port
...
+*/}}
+{{- define "argocd.validateValues.controller.config" -}}
+{{- if gt (int .Values.controller.replicaCount) 1 }}
+    {{- if and .Values.controller.dynamicClusterDistribution.enabled (not (eq .Values.controller.kind "Deployment")) }}
+Argo CD: When running in HA mode with dynamic cluster distribution enabled, the application controller must be installed as a Deployment.
+    {{- else if and (not .Values.controller.dynamicClusterDistribution.enabled) (not (eq .Values.controller.kind "StatefulSet")) }}
+Argo CD: When running in HA mode, the application controller must be installed as a StatefulSet.
+    {{- end -}}
+{{- end -}}
+{{- end -}}
```
fmulero, line 270: I think this validation is not correct or I am understanding the opposite. What should happen if I run this?
```
$ helm install argo-cd bitnami/argo-cd --set server.replicaCount=3 --set controller.replicaCount=3 --set controller.kind=Deployment
```
From your comments I understand that I should face a validation error

maxnitze: There is the feature [Dynamic Cluster Distribution](https://argo-cd.readthedocs.io/en/stable/operator-manual/dynamic-cluster-distribution/). If that feature is enabled, the clusters will dynamically be distributed between all of the existing nodes. In this case the `REPLICAS` env variable is ignored.

maxnitze: And yes, the command from you above should fail. And it does, right?

maxnitze: Hm. Kind-of. The condition is the opposite, since it is printing the message, if the validation fails.

The whole block only renders, when the replica count is > 1. Line 226 is true when dynamic cluster distribution is not enabled and the controller is not deployed as a StatefulSet.

fmulero: Please try it, this is what I see from my side
```console
helm install argo-cd bitnami/argo-cd --set server.replicaCount=3 --set controller.replicaCount=3 --set controller.kind=Deployment
NAME: argo-cd
LAST DEPLOYED: Thu Jul 11 14:13:57 2024
NAMESPACE: default
STATUS: deployed
REVISION: 1
TEST SUITE: None
NOTES:
CHART NAME: argo-cd
CHART VERSION: 6.6.0
APP VERSION: 2.11.4

** Please be patient while the chart is being deployed **

1. Access your Argo CD installation:
  Execute the following commands:

  kubectl port-forward --namespace default svc/argo-cd-server 8080:80 &
  export URL=http://127.0.0.1:8080/
  echo "Argo CD URL: http://127.0.0.1:8080/"

2. Execute the following commands to obtain the Argo CD credentials:

  echo "Username: \"admin\""
  echo "Password: $(kubectl -n default get secret argocd-secret -o jsonpath="{.data.clearPassword}" | base64 -d)"

WARNING: There are "resources" sections in the chart not set. Using "resourcesPreset" is not recommended for production. For production installations, please set the following values according to your workload needs:
  - controller.resources
  - repoServer.resources
  - server.resources
+info https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/
```

maxnitze: Yeah. You are right. I think there is something broken with the validations in general... I think the `validateValues` helper misses this block in the end

```
{{- if $message -}}
{{-   printf "\nVALUES VALIDATION:\n%s" $message -}}
{{- end -}}
```

maxnitze: At least that is part of some of the other charts, that I checked.

maxnitze: And it should probably be `{{-   printf "\nVALUES VALIDATION:\n%s" $message | fail -}}`, so that the deployment fails. This is very in-consequent throughout all the charts though. `fluentd` fails. `flux` does not. `airflow` also does not fail. `logstash` does.

fmulero: I completely agree. Most of our charts fail and print an error message  when the validation is not successful. In this case there was a failure in that piece of code  and I thought the problem was in you code. Sorry about that.

fmulero (approved):
Thanks a lot @maxnitze for your contribution.

LGTM

bitnami/argo-cd/templates/_helpers.tpl
maxnitze: Nothing to be sorry about. Did not test this error case tbh. Just thought I can "integrate" in the existing implemtation. But glad we fixed another issue in the chart 🙂

PR #27586, opened 2024-06-28
postgres exporter monitor all databases

KorenP1:
Currently there is no option to monitor all databases with postgres exporter without mentioning the databases names and creating them in the init scripts. (.Values.auth.database). Postgres Exporter does let this feature work if you dont mention any databases in the URI and that is why i think the required statement is not correct so i removed it.

### DIFF
--- a/bitnami/postgresql/templates/primary/statefulset.yaml
+++ b/bitnami/postgresql/templates/primary/statefulset.yaml
@@ -532,7 +532,7 @@ spec:
             {{- end }}
           {{- end }}
           env:
-            {{- $database := required "In order to enable metrics you need to specify a database (.Values.auth.database or .Values.global.postgresql.auth.database)" (include "postgresql.v1.database" .) }}
+            {{- $database := include "postgresql.v1.database" . }}
             - name: DATA_SOURCE_URI
               value: {{ printf "127.0.0.1:%d/%s?sslmode=disable" (int (include "postgresql.v1.service.port" .)) $database }}
             {{- $pwdKey := ternary (include "postgresql.v1.adminPasswordKey" .) (include "postgresql.v1.userPasswordKey" .) (or (eq $customUser "postgres") (empty $customUser)) }}
--- a/bitnami/postgresql/templates/primary/statefulset.yaml
+++ b/bitnami/postgresql/templates/primary/statefulset.yaml
@@ -532,9 +532,8 @@ spec:
             {{- end }}
           {{- end }}
           env:
-            {{- $database := include "postgresql.v1.database" . }}
             - name: DATA_SOURCE_URI
-              value: {{ printf "127.0.0.1:%d/%s?sslmode=disable" (int (include "postgresql.v1.service.port" .)) $database }}
+              value: {{ printf "127.0.0.1:%d/?sslmode=disable" (int (include "postgresql.v1.service.port" .)) }}
             {{- $pwdKey := ternary (include "postgresql.v1.adminPasswordKey" .) (include "postgresql.v1.userPasswordKey" .) (or (eq $customUser "postgres") (empty $customUser)) }}
             {{- if .Values.auth.usePasswordFiles }}
             - name: DATA_SOURCE_PASS_FILE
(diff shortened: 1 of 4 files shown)


bitnami/postgresql/templates/primary/statefulset.yaml
```diff
@@ -532,7 +532,7 @@ spec:
             {{- end }}
           {{- end }}
           env:
-            {{- $database := required "In order to enable metrics you need to specify a database (.Values.auth.database or .Values.global.postgresql.auth.database)" (include "postgresql.v1.database" .) }}
+            {{- $database := include "postgresql.v1.database" . }}
```
fmulero, line 535: The read statefulset has exactly [the same line](https://github.com/bitnami/charts/blob/main/bitnami/postgresql/templates/read/statefulset.yaml#L435) and probably same changes are required. Could you give it a glance?

fmulero (commented):
Hi @KorenP1, thanks a lot for your contribution.

I left a little comment  and I have a general doubt. In the PR description you mentioned that you need this change to monitor all databases but I am not sure about that. I've just deployed the existing chart (version 15.5.10) and the exporter is monitoring any new database created after the deployment, so I am not sure about your initial assumption. Could you share more details?

bitnami/postgresql/templates/primary/statefulset.yaml
KorenP1: How have you done it?
The main issue is that I do not want to create databases on initialization and when you do not mention any databases the helm chart "required" statement won't let you install the chart, moreover, its kinda weird you monitor any database because the URI is dependent in the .Values.auth.database value

KorenP1: Hi, i have looked it further and i have committed another fix, would be awesome if you look at it.



COMMIT_MESSAGE:

Looks like there is no usage of the DB NAME part in the DATA_SOURCE_URI environment variable which means there is no need for the $database part, the postgres exporter always monitoring all the dbs, also if created after the initialization and not mentioned in the DATA_SOURCE_URI env var.

In conclusion the PR is for two reasons.
1: making DATA_SOURCE_URI env var more "right" and related to what it actually does
2: More importantly, removing the required statement in the $database helm variable assignment in order for users to be able to deploy the postgres exporter although not assigning .Values.auth.database variable for the db creation in the initialization of the postgres

fmulero: I agree, is a little bit weird. You can deploy version 15.5.10 setting `metrics.enabled=true` and `auth.database=test` and `postgresql-exporter` will monitor all the databases in the postgresql server, including the new ones after the deploy. As you said the database name in the URI seems not doing anything and I think we shouldn't require `auth.database` to set it.

bitnami/postgresql/templates/primary/statefulset.yaml
```diff
@@ -532,9 +532,8 @@ spec:
             {{- end }}
           {{- end }}
           env:
-            {{- $database := required "In order to enable metrics you need to specify a database (.Values.auth.database or .Values.global.postgresql.auth.database)" (include "postgresql.v1.database" .) }}
             - name: DATA_SOURCE_URI
-              value: {{ printf "127.0.0.1:%d/%s?sslmode=disable" (int (include "postgresql.v1.service.port" .)) $database }}
+              value: {{ printf "127.0.0.1:%d/?sslmode=disable" (int (include "postgresql.v1.service.port" .)) }}
```
fmulero, line 536: Could you check also the read statefulset ? That file has the same piece of code:
https://github.com/bitnami/charts/blob/08e8c5859d21cd76feeba0474a545c35680b163e/bitnami/postgresql/templates/read/statefulset.yaml#L435-L437

fmulero (changes requested):

bitnami/postgresql/templates/primary/statefulset.yaml
KorenP1: Looks like the same is on the read replicas one, committing now. Thanks!

fmulero (approved):
LGTM

PR #27603, opened 2024-06-30
[bitnami/jenkins] Add Jenkins Password Secret and Secret Key Values

speedythesnail:
### Description of the change

This change enables Jenkins to use an existing password secret, rather than having to provide a password in the chart.

### Benefits

This change enables better GitOps by allowing people to provide an existing secret name, rather than having to put a password in their custom values.
The values file would then be able to be safely stored in a git repository, without the risk of secrets being exposed.

### Possible drawbacks

None known, this change should be backwards compatible with the existing chart as it changes no defaults.

### Checklist


bitnami/jenkins/Chart.yaml
```diff
@@ -35,4 +35,4 @@ maintainers:
 name: jenkins
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/jenkins
-version: 13.4.6
+version: 13.4.7
```
jotamartos, line 38:
```suggestion
version: 13.5.0
```

Moving to a new minor version as you are adding new parameters.

jotamartos (changes requested):
You also need to update the NOTES.txt file with the new way to get the password. 

```
templates/NOTES.txt
45:  echo Password: $(kubectl get secret --namespace {{ include "common.names.namespace" . }} {{ template "common.names.fullname" . }} -o jsonpath="{.data.jenkins-password}" | base64 -d)
```

PR #27628, opened 2024-07-02
[bitnami/argo-cd] Add native support for ConfigManagementPlugins

maxnitze:
### Description of the change

Adding support for [ArgoCD Config Management Plugins (CMP)](https://argo-cd.readthedocs.io/en/stable/operator-manual/config-management-plugins/) to the chart.

> Argo CD's "native" config management tools are Helm, Jsonnet, and Kustomize. If you want to use a different config management tools, or if Argo CD's native tool support does not include a feature you need, you might need to turn to a Config Management Plugin (CMP).

Adding those plugins creates a bit of overhead, as there are some volumes that need to be shared between the repo server and the sidecar container of the plugin and the sidecar configuration in general. This is automatically added when the plugins are enabled.

### Benefits

ConfigManagementPlugins can be enabled directly from the chart.

#### Example with the argocd-vault-plugin

Here's an example of how I would be using it for a CMP that adds support for the [argocd-vault-plugin](https://github.com/argoproj-labs/argocd-vault-plugin/).

```yaml
repoServer:
  extraVolumes:
    - name: vault-configuration
      secret:
        secretName: argo-cd-vault-configuration

  configManagementPlugins:
    enabled: true
    additionalBinaries:
      - name: argocd-vault-plugin
        url: https://github.com/argoproj-labs/argocd-vault-plugin/releases/download/v1.18.1/argocd-vault-plugin_1.18.1_linux_amd64
    plugins:
      - name: avp-helm
        spec:
          init:
            command:
              - /bin/sh
              - -c
              - |
                #!/usr/bin/env bash
                set -Eeuo pipefail
                
                # add all repositories from dependencies of this chart
                for REPO_URL in $(helm dependency list | tail -n+2 | tr -s '[:space:]' | cut -f3)
                do
                  # skip entries that reference local charts
                  if [[ "${REPO_URL}" == file://* ]]; then continue; fi
                  helm repo add $(echo -n "${REPO_URL}" | base64 -w0) "${REPO_URL}"
                done
                
                # finally download the charts dependencies
                helm dependency build
          generate:
            command:
              - /bin/sh
              - -c
              - |
                #!/usr/bin/env bash
                set -Eeuo pipefail
                
                # do some magic with the parameters
                # see https://github.com/argoproj-labs/argocd-vault-plugin/issues/566 for an example
                
                helm template ${HELM_RELEASE_NAME} \
                  --namespace ${ARGOCD_APP_NAMESPACE} \
                  ${HELM_VALUE_FILES} \
                  ${HELM_PARAMETERS} \
                  ${HELM_EXTRA_PARAMETERS} \
                  ./ |
                argocd-vault-plugin generate - -c /run/secrets/vault-configuration/config.yaml
          allowConcurrency: true
          lockRepo: false
        sidecar:
          image:
            repository: alpine/helm
            tag: 3.12.3
          extraEnvVars:
            # root filesystem is read-only, so we need to move the Helm home folders to /tmp
            - name: HELM_CACHE_HOME
              value: /tmp/helm/cache
            - name: HELM_CONFIG_HOME
              value: /tmp/helm/config
            - name: HELM_DATA_HOME
              value: /tmp/helm/data
          extraVolumeMounts:
            - mountPath: /run/secrets/vault-configuration/config.yaml
              name: vault-configuration
              subPath: config.yaml
```

### Possible drawbacks

The change tries to make everything as much configurable as possible. So no drawbacks there (hopefully).

The change in general is opt-in. So no undesired changes to existing installations.

### Applicable issues

None

#### `argocd.tplvalues.merge-with-precedence` Helper

I introduced a new helper for merging values from a list with precedence to later entries.

This was necessary, because the helper `common.tplvalues.merge` has one major "flaw", that I needed fixed here:

```yaml
context1:
  enabled: true
  name: 1001
  isRunning: true
context2:
  enabled: false
  name: 1234
  isRunning: false
```

When I merge these two with `include "common.tplvalues.merge" (dict "values" (list .context1 .context2) "context" $)`, the result is

```
mergedContext:
  enabled: true
  name: 1234
  isRunning: true
```

I don't know if this is a bug or a feature (the comment of the helper says, that "precedence is consistent with http://masterminds.github.io/sprig/dicts.html#merge-mustmerge").

Since I was merging container security contexts here, I wanted to be give the option to deactivate the `readOnlyRootFilesystem`, for example. Or disable the context completely.

The result when using the new helper `include "argocd.tplvalues.merge-with-precedence" (dict "values" (list .context1 .context2) "context" $)` is

```
mergedContext:
  enabled: false
  name: 1234
  isRunning: false
```

### Checklist


maxnitze:
Not sure why the VIB Verify check failed here. But it also failed for other PRs. Maybe somethings wrong with the action?

andresbono (commented):
Thank you very much for your contribution, this could be a great addition! I left some initial questions. Let's discuss them before we go more in depth!

bitnami/argo-cd/values.yaml
```diff
@@ -3151,6 +3151,40 @@ repoServer:
   ## @param repoServer.extraArgs Add extra args to the default repo server args
   ##
   extraArgs: []
+
+  ## ref: https://argo-cd.readthedocs.io/en/stable/operator-manual/config-management-plugins/
+  ##
+  configManagementPlugins:
+    ## @param repoServer.configManagementPlugins.enabled Whether the config management plugins are enabled or not.
+    enabled: false
```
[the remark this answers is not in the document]
maxnitze: We could, yeah. This is the pattern we use for most features though, right? I also like the fact, that for debugging purposes you can just disable it here without having to delete the whole plugins config.

bitnami/argo-cd/values.yaml
```diff
@@ -3151,6 +3151,40 @@ repoServer:
...
+    ## @param repoServer.configManagementPlugins.additionalBinaries Additional binaries necessary for the config management plugins.
+    ## Will be downloaded from the given URL and made available as `name` in the configured `additionalBinariesDir` of the plugin.
+    ## If a custom script is used to download the binary, make sure to eventually provide your binary in an executable format in the
+    ## shared volume mounted to `/additional-binaries`.`
+    ## E.g.
+    ## additionalBinaries:
+    ##   - name: my-custom-binary
+    ##     url: https://www.example.com/my-custom-binary-1.2.3
+    ##     customScript: ""
+    additionalBinaries: []
```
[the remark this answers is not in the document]
maxnitze: While I was thinking about it: Maybe we can skip the "customScript" part and recommend to use an own init container or custom image in that case. But for the easy cases, where the binary can just be downloaded from a given URL, I would prefre to keep the functionality.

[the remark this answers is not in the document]
andresbono: Yes, at the very least, we should skip the `customScript` part. Will you update the PR? Once we have that removed, let's see if it makes sense to keep simplifying the implementation. Thanks!

maxnitze:
Updated the chart and removed the custom script option from the binaries.

andresbono:
@maxnitze, thank you for addressing this. Let me take a look at the changes to get back to you. No worries about the conflicts, I'll take care of resolving them.

bitnami/argo-cd/templates/repo-server/deployment.yaml
```diff
@@ -137,6 +137,40 @@ spec:
...
+            - /additional-binaries/argocd-cmp-server
+          {{- if .Values.repoServer.containerSecurityContext.enabled }}
+          securityContext: {{- omit .Values.repoServer.containerSecurityContext "enabled" | toYaml | nindent 12 }}
+          {{- end }}
+          volumeMounts:
+            - mountPath: /additional-binaries
+              name: cmp-additional-binaries
+        {{- if .Values.repoServer.configManagementPlugins.additionalBinaries }}
+        - name: download-additional-binaries
+          image: {{ include "common.images.image" (dict "imageRoot" (dict "repository" "curlimages/curl" "tag" "latest") "global" .Values.global) }}
```
andresbono, line 157: Any container image should be maintained by Bitnami and configured in a standard way from `values.yaml`. You could use `bitnami/os-shell`. Look at the `volumePermissions.image` value to follow a similar pattern.

Do you have any concerns with that?

maxnitze: No issues at all! I remember thinking about this when I added the init-container. I think I just forgot about it. Anyway, added it now!

bitnami/argo-cd/templates/repo-server/deployment.yaml
```diff
@@ -273,9 +309,82 @@ spec:
...
+              mountPath: /home/argocd/cmp-server/plugins
+            {{- end }}
           {{- if .Values.repoServer.extraVolumeMounts }}
           {{- include "common.tplvalues.render" (dict "value" .Values.repoServer.extraVolumeMounts "context" $) | nindent 12 }}
           {{- end }}
+        {{- if .Values.repoServer.configManagementPlugins.enabled }}
+        {{- range $plugin := .Values.repoServer.configManagementPlugins.plugins }}
+        {{- $additionalBinariesDir := $plugin.additionalBinariesDir | default "/usr/local/bin" | trimSuffix "/" }}
+        - name: {{ $plugin.name }}
+          image: {{ include "common.images.image" (dict "imageRoot" $plugin.sidecar.image "global" $.Values.global) }}
```
andresbono, line 324: I think it would be better to simplify all this and let users decide what will be the sidecar preferred structure. It simplify our maintenance at the same time it allows total flexibility for users.

It would be the same approach that we follow in `.Values.repoServer.sidecars` for example.

bitnami/argo-cd/templates/repo-server/deployment.yaml
```diff
@@ -273,9 +309,82 @@ spec:
...
+            - configMapRef:
+                name: {{ include "common.tplvalues.render" (dict "value" $plugin.sidecar.extraEnvVarsCM "context" $) }}
+            {{- end }}
+            {{- if $plugin.sidecar.extraEnvVarsSecret }}
+            - secretRef:
+                name: {{ include "common.tplvalues.render" (dict "value" $plugin.sidecar.extraEnvVarsSecret "context" $) }}
+            {{- end }}
+          {{- if $plugin.sidecar.resources }}
+          resources: {{- toYaml $plugin.sidecar.resources | nindent 12 }}
+          {{- else if and $plugin.sidecar.resourcePreset (ne $plugin.sidecar.resourcesPreset "none") }}
```
andresbono, line 357: There is a typo here, but in any case, refer to my previous comment about simplifying all this code and allow users to set their sidecars config explicitly.

```suggestion
          {{- else if and $plugin.sidecar.resourcesPreset (ne $plugin.sidecar.resourcesPreset "none") }}
```

bitnami/argo-cd/values.yaml
```diff
@@ -3153,6 +3153,61 @@ repoServer:
...
+        digest: ""
+        pullPolicy: IfNotPresent
+        ## Optionally specify an array of imagePullSecrets.
+        ## Secrets must be manually created in the namespace.
+        ## ref: https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/
+        ## e.g:
+        ## pullSecrets:
+        ##   - myRegistryKeySecretName
+        ##
+        pullSecrets: [ ]
```
andresbono, line 3185: Should this image be added to `argocd.imagePullSecrets`?

andresbono (commented):
Hi, I just added some comments to the PR. Could you please check? Thank you!

bitnami/argo-cd/templates/_helpers.tpl
```diff
@@ -191,6 +191,24 @@ Create the name of the service account to use for Dex
 {{- end -}}
 {{- end -}}
 
+{{/*
```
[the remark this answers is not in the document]
maxnitze: Yep, I would really like to move this to the common chart. I posted my reasoning to the description of the PR. But here's the condensed version of it:

Let's assume we have these two dicts:

```yaml
context1:
  enabled: true
  name: 1001
  isRunning: true
  someOtherKey: "someOtherValue"
context2:
  enabled: false
  name: 1234
  isRunning: false
  anotherKey: "anotherValue"
```

When we merge both with `common.tplvalues.merge` (from the common chart), the resulting dict is

```yaml
mergedContext:
  enabled: true
  name: 1234
  isRunning: true
  someOtherKey: "someOtherValue"
  anotherKey: "anotherValue"
```

So even though we would assume the values from `context2` to have precedence (which they have for `name`, for example), the boolean values stay the same.

In general it is not possible to overwrite any keys from `context1` with values that are _falsy_.

I'm not sure if this is a bug or by design (the docs say that "precedence is consistent with http://masterminds.github.io/sprig/dicts.html#merge-mustmerge"). This helper here is designed to circumvent that.

bitnami/argo-cd/templates/repo-server/deployment.yaml
maxnitze: This block is kind of the whole point of this PR. There are some mounts that have to be mounted to specific directories, like the CMP binary, the `plugins` folder or the `plugin.yml`.

The user still has the option to completely set it by him or herself (through `.Values.repoServer.sidecars`). But removing this _opinionated_ block defeats the whole purpose of this PR.

maxnitze:
I'm back from a longer vacation and looked through your remarks. Had some comments myself, but fixed most of them.

On the `argocd.tplvalues.merge-with-precedence` helper topic, should I create a PR in the common chart and update this afterwards? Or would it be okay to keep it as is and create the PR in the common chart afterwards (I guess there will be some discussion, if that is a valid case for the common chart or not).

andresbono:
> On the `argocd.tplvalues.merge-with-precedence` helper topic, should I create a PR in the common chart and update this afterwards?

Yes, please. Let's do it like that. Once the new PR for the bitnami/common library is merged, we can continue the final changes here.

Thank you!

maxnitze:
Created #29566 to discuss what the correct fix for the issue is. Will come back here once that issue is closed.

bitnami/argo-cd/templates/repo-server/deployment.yaml
```diff
@@ -137,6 +137,42 @@ spec:
...
+          args:
+            - |-
+              {{- range $additionalBinary := .binaries }}
+              # downloading {{ $additionalBinary.name }}
+              curl -L {{ $additionalBinary.url }} -o /additional-binaries/{{ $additionalBinary.name }}
+              chmod +x /additional-binaries/{{ $additionalBinary.name }}
+              {{- end }}
+          volumeMounts:
+            - mountPath: /additional-binaries
+              name: cmp-additional-binaries
```
juan131, line 172: These two steps look overcomplicated to me. I guess we're using `bitnami/os-shell` to download extra binaries because `bitnami/argo-cd` lacks `curl`, right?

bitnami/argo-cd/templates/repo-server/deployment.yaml
```diff
@@ -273,9 +309,82 @@ spec:
...
+          {{- end }}
+          {{- if $plugin.sidecar.customCommand }}
+          command: {{- include "common.tplvalues.render" (dict "value" $plugin.sidecar.customCommand "context" $) | nindent 12 }}
+          {{- else }}
+          command:
+            - {{ $additionalBinariesDir }}/argocd-cmp-server
+          {{- end }}
+          {{- if $plugin.sidecar.customArgs }}
+          args: {{- include "common.tplvalues.render" (dict "value" $plugin.sidecar.customArgs "context" $) | nindent 12 }}
+          {{- end }}
```
juan131, line 341: We always use `xxx.command` and `xxx.args` as paramaters name for users to customize their custom command/args. For consistency with the catalog, please use this naming.

bitnami/argo-cd/templates/repo-server/deployment.yaml
juan131: The thing is... This assumes a very specific object structure for each element in the `plugins` array. 
The main value the PR adds is automating the `volumeMounts` that are required for the sidecars to work. Could we sth like this instead?
```yaml
        {{- range $sidecar := .Values.repoServer.configManagementPlugins.sidecars }}
        {{- include "common.tplvalues.render" (dict "value" $sidecar "context" $) | nindent 8 }}
          volumeMounts:
            - name: cmp-{{ $sidecar.name }}
              mountPath: /home/argocd/cmp-server/config/plugin.yaml
              subPath: plugin.yaml
            - name: cmp-server-plugins
              mountPath: /home/argocd/cmp-server/plugins
            - name: cmp-additional-binaries
              mountPath: {{ $additionalBinariesDir }}
            - name: cmp-server-tmp-dir
              mountPath: /tmp
        {{- end }}
```
This way, we help users setting up mounted directories while we keep flexible enough for users to customize the sidecars.

juan131 (changes requested):

juan131:
Closing due to lack of activity, feel free to reopen it if the work is resumed.

maxnitze:
Still have it on my TODO list.  Sorry for that. Will re-open once I find the time to do it!

PR #27734, opened 2024-07-03
Postgres exporter URI fix

KorenP1:
The postgres exporter is monitoring all the dbs in the postgresql cluster no matter what you assign in the URI.

The postgres exporter will by default do the initial connection through the db of the username you are using for instance when you use "postgres" user the exporter will query all the dbs through connecting to database "postgres", when you'll use user "test" the exporter will query all the dbs through connecting database "test".

You are able to assign the db you want to connect to through the URI of the postgres exporter after the slash sign.

**### BUG**: Lets say you are using user "test" and there is no db "test", the exporter will not work and the connection will fail.
When using a custom username and using the URI in the postgres exporter that currently set, the connection will probably fall because it will search for a db with that name that might exist or not. the connection will not fall if the a database the same name as the username name exists.

Postgres always have a database named "postgres" for its internal uses and this helm chart always deploys an admin user.

**### Conclustion**: We can use in the URI the postgres db because we are admin and the postgres database does always exist.


KorenP1:
@fmulero

KorenP1:
@alemorcuq

KorenP1:
@fmulero @alemorcuq 
Hi, I would like to get a review. Kinda important for my project
Thanks.

fmulero (commented):
Thanks a lot @KorenP1  for taking care of this.

I ran some tests and maybe we should force the `postgres` user for metrics (not sure). For instance, if we deploy setting `auth.username`, the metrics seems not complete, there is no information for new databases.

Steps to reproduce it:
1. Install setting `auth.username`:
```
helm install postgresql bitnami/postgresql --set auth.username=test  --set auth.password=salchichon --set metrics.enabled=true
```
2. Connect to the server using the test user:
```
$ export POSTGRES_PASSWORD=salchichon
$ kubectl run postgresql-client --rm --tty -i --restart='Never' --namespace default --image docker.io/bitnami/postgresql:16.3.0-debian-12-r18 --env="PGPASSWORD=$POSTGRES_PASSWORD"       --command -- psql --host postgresql -U test -d postgres -p 5432
If you don't see a command prompt, try pressing enter.

postgres=> create database salami;
CREATE DATABASE
postgres=> 
```
3. Check metrics. At this point we can't see any metric about the new database.

KorenP1:
I suggest you look at the postgres exporter container logs

I think it is because what i have said in the first comment, it will try connect to a database named test, as the username. And the fix i committed supposed to fix it.

fmulero (approved):
I thought I was using your changes but I was wrong. My fault.

LGTM

PR #27736, opened 2024-07-03
[bitnami/apache] Uses common.compatibility.renderSecurityContext template to render containerSecurityContext values in container specs

### LINKED ISSUE #27633: global.compatibility.openshift.adaptSecurityContext only applies to pod securityContext, not container specs
### Name and Version

bitnami/apache 11.2.6

amd64

See https://github.com/bitnami/charts/blob/apache/11.2.6/bitnami/apache/templates/deployment.yaml#L58 and compare to container specs.

No

Setting global.compatibility.openshift.adaptSecurityContext to `auto` or `force` applies to `securityContext` of container specs. 

### What do you see instead?

Show uid and gid for `preserve-logs-symlinks` initContainer:

```yaml
spec:
  template:
    spec:
      initContainers:
        - name: preserve-logs-symlinks
          securityContext:
            runAsUser: 1001
            runAsGroup: 0
```

### Additional information

_No response_

dchandekstark:
### Description of the change

Use common.compatibility.renderSecurityContext template to render containerSecurityContext values in container specs.

### Benefits

Consistency with other Bitnami charts, and presumably the intended behavior.

### Possible drawbacks

It may be possible that a user relies on the current (incorrect) default behavior in a OpenShift environment - which would be to set `securityContext.runAsUser: 1001` and `securityContext.runAsGroup: 0` in the container specs (excepting the main `apache` container).  In this case, the default setting `global.compatibility.openshift.adaptSecurityContext: auto` would likely remove the securityContext config for container specs.  If in fact the user wants the pre-commit behavior, they can set global.compatibility.openshift.adaptSecurityContext to `disabled`. 

### Applicable issues

- fixes #27633 

### Checklist

### DIFF
--- a/bitnami/airflow/CHANGELOG.md
+++ b/bitnami/airflow/CHANGELOG.md
@@ -1,8 +1,13 @@
 # Changelog
 
-## 18.3.6 (2024-07-03)
+## 18.3.7 (2024-07-03)
 
-* [bitnami/airflow] Release 18.3.6 ([#27682](https://github.com/bitnami/charts/pull/27682))
+* [bitnami/airflow] Release 18.3.7 ([#27742](https://github.com/bitnami/charts/pull/27742))
+
+## <small>18.3.6 (2024-07-03)</small>
+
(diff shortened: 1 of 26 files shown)


dchandekstark:
Arg, I've borked this PR.  Will open new one.

PR #27856, opened 2024-07-09
[bitnami/keycloak] Added nodeSelector and toleration for postgres

Exobitt:
### Description of the change

Made it possible to parse nodeSelector and tolerations to postgres, if not using an external database.

### Benefits

if using different nodes for storage or stateful apps, this makes it possible to deploy postgres and its data to those nodes

### Possible drawbacks

n/a

### Applicable issues

n/a

### Additional information

The reason this is made, is because in our usecase, we have longhorn deployed on certain nodes, which handles data and stateful applications and operators.
Without parsing nodeSelector and toleration towards postgres, we're forced to setup and use an external database.

Of course, we could just deploy a postgresql chart. But that creates more dependencies, that's not packed together

### Checklist

### DIFF
--- a/bitnami/keycloak/CHANGELOG.md
+++ b/bitnami/keycloak/CHANGELOG.md
@@ -1,8 +1,12 @@
 # Changelog
 
-## 21.6.0 (2024-07-08)
+## 21.6.1 (2024-07-09)
 
-* [bitnami/keycloak] Add support for minReadySeconds ([#27550](https://github.com/bitnami/charts/pull/27550))
+* [bitnami/keycloak] Added nodeSelector and toleration for postgres ([#27856](https://github.com/bitnami/charts/pull/27856))
+
+## 21.6.0 (2024-07-09)
+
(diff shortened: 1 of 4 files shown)


fmulero (commented):
Thanks a lot @Exobitt for your contribution and such detailed explanation.

I've been taking it a glance and I think these changes are not required. You can use all existing values in the subchart just using the `postgresql` prefix.

Please give it a try:
```console
$ helm install kc oci://registry-1.docker.io/bitnamicharts/keycloak --version 21.4.5 --set
 postgresql.primary.nodeSelector.test=keycloak
...
$ kubectl get sts/kc-postgresql -o yaml | yq '.spec.template.spec.nodeSelector' --
test: keycloak
```

We only add a subset of values for subcharts, just only those we need to change or we need to document in some way.

Exobitt:
Hi again

That works just fine. Thank you!

I'll close this pr

PR #27436, opened 2024-06-18
[bitnami/keycloak] fix Keycloak HTTP schema for edge proxy mode

singhbaljit:
### Description of the change

Uses HTTPS for admin host url when using proxy=edge.

### Benefits

Fixes the issue to make admin console accessible, otherwise we get `redirect_uri` incorrect error.

### Possible drawbacks

N/A

### Applicable issues

- fixes #27315

### Checklist

- [x] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/). This is *not necessary* when the changes only affect README.md files.
- [x] Variables are documented in the values.yaml and added to the `README.md` using [readme-generator-for-helm](https://github.com/bitnami/readme-generator-for-helm)
- [x] Title of the pull request follows this pattern [bitnami/<name_of_the_chart>] Descriptive title
- [x] All commits signed off and in agreement of [Developer Certificate of Origin (DCO)](https://github.com/bitnami/charts/blob/main/CONTRIBUTING.md#sign-your-work)

Issue #26175, opened 2024-05-21
KC_PROXY deprecated, use proxy-headers


Loucool111:
### Name and Version

bitnami/keycloack 21.2.1

### What architecture are you using?

amd64

### What steps will reproduce the bug?

1. Deploy helm chart with the following values:

```yaml
replicaCount: 1
proxy: "edge"
production: true

cache:
  enabled: true
  stackName: kubernetes

service:
  type: NodePort

ingress:
  enabled: true

  path: "/*"
  hostname: "[REDACTED]"

  ingressClassName: alb

postgresql:
  enabled: false

externalDatabase:
  [REDACTED]
``` 

### What is the expected behavior?

normal startup

### What do you see instead?

Logs about deprecated KC_PROXY config
```
2024-05-20 22:56:14,193 WARN  [org.key.qua.run.cli.Picocli] (main) The following used options or option values are DEPRECATED and will be removed in a future release:
	- proxy: Use proxy-headers.
Consult the Release Notes for details.
```
### Additional information

When using proxy: edge, what should be the config?

dalbani:
I guess fixing the chart will first require fixing the Docker image, as reported at https://github.com/bitnami/containers/issues/65190.

Note that Keycloak 25 introduced even more changes when it comes to hostnames: https://www.keycloak.org/2024/06/keycloak-2500-released.html#_new_hostname_options.

andresbono:
Thank you for linking both issues. You are correct, the container image needs to be adapted as well.

tanadeau:
Still true

andresbono:
I have just set the https://github.com/bitnami/charts/labels/on-hold label as https://github.com/bitnami/containers/issues/65190 needs to be fixed first.

Kajot-dev:
This should be fixed with  #27890 and #28530

lucasfcnunes:
One question I have about `proxy: edge` to `proxyHeaders: xforwarded (or forwarded)` is the arg `--http-enabled true` in the documentation.
Seems that the PR #27890 and #28530 don't address it by adding a flag for `--http-enabled true`.

https://www.keycloak.org/docs/latest/upgrading/index.html#deprecated-proxy-option

<img src="https://github.com/user-attachments/assets/c3ed292d-0416-4f79-8689-2047c940c42b" width="50%"/>

Kajot-dev:
[...]

Issue #25963, opened 2024-05-17
[bitnami/keycloak] Admin URL no longer works for chart versions higher than 21.0.2


CaptainKrby:
### Name and Version

bitnami/keycloak:21.2.1

### What architecture are you using?

None

### What steps will reproduce the bug?

Here is my values.yaml :

````yaml
keycloak:
  replicaCount: 1
  production: true
  proxy: edge
  httpRelativePath: "/auth/"
  podAntiAffinityPreset: hard
  clusterDomain: k8s-valid.local

  auth:
    adminUser: root
    adminPassword: "xxxx"

  resources:
    requests:
      cpu: 2
      memory: 512Mi
    limits:
      cpu: 3
      memory: 1024Mi

  ingress:
    enabled: true
    ingressClassName: "nginx"
    hostname: keycloak-dr.company.net
    path: "{{ .Values.httpRelativePath }}"
    pathType: ImplementationSpecific
    servicePort: http
    annotations:
      nginx.ingress.kubernetes.io/app-root: /auth/realms/realm-internal/account/#
      nginx.ingress.kubernetes.io/affinity: cookie
    extraTls:
      - hosts:
          - keycloak-dr.company.net

  adminIngress:
    enabled: true
    ingressClassName: "nginx"
    hostname: keycloak-dr-admin.company.net
    path: "{{ .Values.httpRelativePath }}"
    pathType: ImplementationSpecific
    servicePort: http
    annotations:
      nginx.ingress.kubernetes.io/app-root: "{{ .Values.httpRelativePath }}"
      nginx.ingress.kubernetes.io/affinity: cookie
    extraTls:
      - hosts:
          - keycloak-dr-admin.company.net

  networkPolicy:
    enabled: false

  cache:
    enabled: true
    stackName: kubernetes
    stackFile: "/opt/bitnami/keycloak/conf/cache-ispn.xml"

  postgresql:
    enabled: true
    architecture: replication
    auth:
      username: keycloak
      password: "xxxxxx"
      database: keycloakdb
      postgresPassword: "yyyy"
      replicationPassword: "zzzz"
    primary:
      persistence:
        existingClaim: keycloak-dr-primary-pvc
    readReplicas:
      replicaCount: 1
      podAntiAffinityPreset: hard
      persistence:
        existingClaim: keycloak-dr-read-pvc
```` 

### What do you see instead?

https://keycloak-dr-admin.company.net/admin/ returns :

404: Not Found

### Additional information

[...]

PR #25386, opened 2024-04-25
[bitnami/keycloak]: add url env vars to fix admin ingress access

MrFreezeex:
### Description of the change

Replace KC_HOSTNAME_ADMIN by KC_HOSTNAME_ADMIN_URL to include the scheme (http or https) and add KC_HOSTNAME_URL for the frontend to function correctly. These variable allow the front to function correctly if we have the tls option enabled. Before this commit the front was always attempting to load some resources over http while it may be exposed in https.

### Benefits
Keycloak admin ingress works if tls/https is configured

### Possible drawbacks

If some users defined `KC_HOSTNAME` or `KC_HOSTNAME_URL` in extraEnv it could prevent keycloak to start, I reduced the risk by adding this variable only if the ingress is not enabled though

### Checklist

- [x] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/). This is *not necessary* when the changes only affect README.md files.
- [ ] Variables are documented in the values.yaml and added to the `README.md` using [readme-generator-for-helm](https://github.com/bitnami/readme-generator-for-helm)
- [x] Title of the pull request follows this pattern [bitnami/<name_of_the_chart>] Descriptive title
- [x] All commits signed off and in agreement of [Developer Certificate of Origin (DCO)](https://github.com/bitnami/charts/blob/main/CONTRIBUTING.md#sign-your-work)

PR #27867, opened 2024-07-10
[bitnami/keycloak] fix conditions for KC_HOSTNAME_URL

singhbaljit:
### Description of the change

#25386 added usage of `KC_HOSTNAME_ADMIN_URL` and `KC_HOSTNAME_URL`; however, the implementation has many issues. One such issue (dealing with `edge` proxy mode) was fixed earlier in #27436; another issue is documented in #25963. This PR is resolving yet another issue. Per [v1 docs](https://www.keycloak.org/server/hostname-deprecated), `KC_HOSTNAME_URL` and `KC_HOSTNAME_ADMIN_URL` are independent environment variables. Since the chart makes two separate ingresses (one public and one admin), these variables should be defined independently based on the ingress values. Without this, in the proxy edge mode, Keycloak is returning Kubernetes internal DNS when calling `.well-known/openid-configuration` from within the same cluster using internal services since `KC_HOSTNAME_URL` is not defined.

### Benefits

* Fixes the issue of returning the correct frontend URL
* This will help migrate to [v2 hostname](https://www.keycloak.org/server/hostname), see #26175

### Possible drawbacks

None, this is a bug fix

### Checklist

### DIFF
--- a/bitnami/keycloak/CHANGELOG.md
+++ b/bitnami/keycloak/CHANGELOG.md
@@ -1,8 +1,12 @@
 # Changelog
 
-## 21.6.0 (2024-07-08)
+## 21.6.1 (2024-07-10)
 
-* [bitnami/keycloak] Add support for minReadySeconds ([#27550](https://github.com/bitnami/charts/pull/27550))
+* [bitnami/keycloak] fix conditions for KC_HOSTNAME_URL ([#27867](https://github.com/bitnami/charts/pull/27867))
+
+## 21.6.0 (2024-07-09)
+
+* [bitnami/keycloak] Add support for minReadySeconds (#27550) ([bf357f9](https://github.com/bitnami/charts/commit/bf357f93bb2ad28d3a27826f4bae8a65a0bc3af5)), closes [#27550](https://github.com/bitnami/charts/issues/27550)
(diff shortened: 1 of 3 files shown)


fmulero (approved):
Thanks a lot @singhbaljit for taking care of this issue and share all the documentation and follow up steps.

LGTM

wkbrd:
I updated to the newest chart today and now the chart does not work.  The KC_HOSTNAME_URL in my environment is resolving to https://.

This causes Keycloak to not start:
Invalid base URL for FrontEnd URLs: https://
2024-07-11T12:41:49.937012733-04:00 ERROR: Expected scheme-specific part at index 6: https:
2024-07-11T12:41:49.937043946-04:00 For more details run the same command passing the '--verbose' option. Also you can use '--help' to see the details about the usage of the particular command.

The KC_HOSTNAME_URL was previously undefined in my environment.

I am using production=true, ingress.enabled=true, ingress.hostname=, and ingress.extraHosts[0].name=,ingress.extraHosts[0]

This PR seems to assume that ingress.hostname is non-blank.  Can we change this new section perhaps to check if ingress.hostname is defined/non-blank?

Issue #25233, opened 2024-04-18
[bitnami/keycloak] Unable to use Ingress rule without a host defined


wkbrd:
### Name and Version

bitnami/keycloak 21.0.1

### What architecture are you using?

None

### What steps will reproduce the bug?

Attempting to deploy Keycloak helm chart with ingress.enabled=true should not require a hostname to be defined.  An ingress does not require the host to be filled in.

### Are you using any custom parameters or values?

ingress.enabled=true

### What is the expected behavior?

Ingress run should be created without the "host: ..." line, though the rest of the rule is defined and traffic flows properly.

### What do you see instead?

Helm chart does not deploy and gives an error message

javsalgar:
Hi!

Could you try using the `extraHosts` section? There you can use an empty `host` value, which may work for your use case

wkbrd:
That ended up working.  Thank you!

PR #27890, opened 2024-07-10
[bitnami/keycloak] Add support for proxy-headers

Kajot-dev:
### Description of the change

This PR implements new proxy setting introduced in Keycloak. It's a helm chart equivalent to https://github.com/bitnami/containers/pull/67957 and is meant to be used together

### Benefits

Not using deprecated options

### Applicable issues

- fixes #65190

### Checklist

- [x] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/). This is *not necessary* when the changes only affect README.md files.
- [x] Variables are documented in the values.yaml and added to the `README.md` using [readme-generator-for-helm](https://github.com/bitnami/readme-generator-for-helm)
- [x] Title of the pull request follows this pattern [bitnami/<name_of_the_chart>] Descriptive title
- [x] All commits signed off and in agreement of [Developer Certificate of Origin (DCO)](https://github.com/bitnami/charts/blob/main/CONTRIBUTING.md#sign-your-work)

PR #27912, opened 2024-07-11
bitnami/keycloak Safer variable set of KC_HOSTNAME_URL

### LINKED ISSUE #27911: KC_HOSTNAME_URL is set when it should not be (regression)
### Name and Version

bitnami/keycloak 21.6.1

amd64

Used the chart with:
I am using `proxy=edge,production=true`, `ingress.enabled=true`, `ingress.hostname=`, and `ingress.extraHosts[0].name=,ingress.extraHosts[0].path`

The chart does not work properly (pod goes into crashloop)

The `KC_HOSTNAME_URL` in my environment is resolving to `https://`.

This is due to the proxy hostname being used by Keycloak when proxy=edge is enabled.

This causes Keycloak to not start:
```
Invalid base URL for FrontEnd URLs: https://
2024-07-11T12:41:49.937012733-04:00 ERROR: Expected scheme-specific part at index 6: https:
2024-07-11T12:41:49.937043946-04:00 For more details run the same command passing the '--verbose' option. Also you can use '--help' to see the details about the usage of the particular command.
```

Yes.
```
proxy=edge, production=true, ingress.enabled=true, ingress.hostname=, and ingress.extraHosts[0].name=,ingress.extraHosts[0]
```

pod comes up and works properly.

### What do you see instead?

pod goes into crash loop

wkbrd:
### Description of the change

Bug fix for KC_HOSTNAME_URL being set when ingress.hostname is undefined.  It should not output the variable in this case.

### Benefits

Restores regression from recent change in this PR: https://github.com/bitnami/charts/pull/27867

### Possible drawbacks

N/A

### Applicable issues

- fixes #27911 
- related to #25233

### Additional information

See issue for more details.

### Checklist


singhbaljit:
For what use-case would the hostname be empty? I think the hostname is required when the ingress is enabled.

wkbrd:
With the proxy=edge value set, Keycloak will determine use the provided proxy headers to determine the fully URL that is used in web links/etc.  This can be paired with ingress.enabled=true,ingress.extraHosts[0].name=,ingress.extraHosts[0].path=/someprefix to allow an Ingress rule that is created without a hostname being defined in the Ingress object.

Therefore, I don't think the hostname is required.

wkbrd:
The use of extraHosts for this scenario was recommended in #25233 previously

singhbaljit:
Got it, I think the same should also apply for the admin ingress too then.

wkbrd:
Likely so, though I was not using it.  Would you like me to update the PR to add that check to the admin ingress?

singhbaljit:
I'm not using the `extraHosts`, it doesn't impact me. But for consistency between the ingresses, I think we should add it, unless @javsalgar disagrees.

singhbaljit:
I guess this was another miss from #25386 😄

wkbrd:
PR updated to add a check for admin ingress, too.  Thanks for your review!

wkbrd:
> I'm not using the `extraHosts`, it doesn't impact me. But for consistency between the ingresses, I think we should add it, unless @javsalgar disagrees.

Done now :-)

singhbaljit:
Just a side note, this might all have to be revisited with [v2 hostname](https://www.keycloak.org/server/hostname) and proxy headers (#27890). Based on the defaults, I think the recommended approach is to use full URLs in in `hostname`/`hostname-admin` since `hostname-strict` is enabled and `hostname-backchannel-dynamic` is disabled by default.

wkbrd:
Hi @rafariossaa, please let me know if you have questions regarding this PR. It restores functionality that regressed with a previous merge. Thanks!

rafariossaa (approved):
LGTM

PR #28112, opened 2024-07-16
[github-action] Fix kubescape score step for common chart

migruiz4:
### Description of the change

Fixes an issue with the 'chart-score' step in the CI pipeline that affects the common chart.

It omits the untar step when there are no sub-charts.


juan131 (approved):
LGTM

PR #28178, opened 2024-07-19
[github-actions] Replace CD pipeline in favor of OCI releases

migruiz4:
### Description of the change

This PR replaces the current CD pipeline in favor of the new index update workflow.

From now on, charts will be published exclusively in OCI format instead of .tgz and the index will be automatically updated pointing to the OCI releases.

Changes included in this PR:
- New 'index-update' workflow is added.
- Because the only job that remains is the 'push-tag' job, the 'cd-pipeline' workflow title has been renamed.


.github/workflows/push-tag.yml
```diff
@@ -0,0 +1,88 @@
...
+        with:
+          path: charts
+          fetch-depth: 2 # to be able to obtain files changed in the latest commit
+      - id: get-chart
+        name: 'Get modified charts'
+        run: |
+          cd charts
+          files_changed="$(git show --pretty="" --name-only)"
+          # Adding || true to avoid "Process exited with code 1" errors
+          charts_dirs_changed="$(echo "$files_changed" | xargs dirname | grep -o "bitnami/[^/]*" | sort | uniq || true)"
```
rafariossaa, line 34:
```suggestion
          charts_dirs_changed="$(echo "$files_changed" | xargs dirname | grep -o "bitnami/[^/]*" | sort -u || true)"
````

rafariossaa (changes requested):

.github/workflows/push-tag.yml
migruiz4: I do not think we should address that kind of topic here.
That section remains unchanged from the original workflow.

In this PR these are the changes I performed:

- Renamed the file
- Removed the 'vib-publish' step
- Removed the 'update-index' step
- Removed the 'push-promotion' step
- Removed the 'slack-notif' step

Because of the file rename changes may not be as visible as they should, I will try to address that.

migruiz4: It should be easier to see now that I separated it in two commits: https://github.com/bitnami/charts/pull/28178/commits/0433d24ae556a9ac2d9cff4f45a1c69096e1a542

Although Github Diff still shows the changes as removing an entire file and adding it again.

rafariossaa (approved):
LGTM

migruiz4:
Hi there!

I have updated this PR to include all the changes in a single PR, by having all workflows in the 'main' branch.

The reason why I'm moving all '[Index]' workflows to the main branch is that, since we started squashing the index branch to a single commit, we are losing version control of those files.

migruiz4:
The new workflow has been tested here: https://github.com/migruiz4/charts/actions/runs/10043677410

migruiz4:
Reverted the 'Move index workflows to main branch' change

PR #28189, opened 2024-07-22
[bitnami/solr] Change permissions of the Solr data directory

soroush:
* When using a volume permission change container, change permissions of the Solr data directory, as well as its content.
* Bump patch version. Now 9.3.8

### Description of the change

Make the initialization container change the owner of the data directory itself. Otherwise, Solr will not be able to create files in it and will get stuck in a crash loop. The same chown is being done on the zookeeper data directory, but it seems it is forgotten on Solr.

### Benefits

With this change, the permission system works. Otherwise, the chart fails to deploy on restricted settings. For example, if you use Talos and host path volumes, the existing init container fails.

### Possible drawbacks

None.

### Applicable issues

None

### Checklist

### DIFF
--- a/bitnami/solr/CHANGELOG.md
+++ b/bitnami/solr/CHANGELOG.md
@@ -1,8 +1,12 @@
 # Changelog
 
-## 9.3.7 (2024-07-16)
+## 9.3.8 (2024-07-22)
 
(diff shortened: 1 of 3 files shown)


dgomezleon (approved):

PR #28252, opened 2024-07-24
[bnitnami/mysql] Update documentation to use bash instead of sh

### LINKED ISSUE #28195: [MySQL] Default documentation contains bash script, but executes it via `sh`
### Name and Version

oci://registry-1.docker.io/bitnamicharts/mysql:11.1.4

ARM64

1. Follow https://github.com/bitnami/charts/blob/main/bitnami/mysql/README.md
2. Specifically use the following scriptlet in your helm chart:
```yaml
initdbScripts:
  my_init_script.sh: |
    #!/bin/sh
    if [[ $(hostname) == *master* ]]; then
      echo "Master node"
      mysql -P 3306 -uroot -prandompassword -e "create database new_database";
    else
      echo "No master node"
    fi
```
(This is an exact copy from the docs)

The script should work without any error, and create `new_database` on the master node.

### What do you see instead?

Since the script is executed by `/bin/sh` it prints 
```
/bin/sh: 2: [[: not found
```

This is correct as per my understanding the POSIX shell does not know `[[`, it only knows `[`

### Additional information

I believe today `bash` is widely adopted, as such instead of rewriting the script to be POSIX compliant it might be easier to just execute it using `/usr/bin/bash`

Additionally, I would make the following two and a half changes to the script:
1. The script uses `-uroot -prandompassword`, would suggest changing to `-uroot -p"$MYSQL_ROOT_PASSWORD"` as that variable is available.
2. An even better alternative than `MYSQL_ROOT_PASSWORD` would be to use:
```bash
                  password_aux="${MYSQL_ROOT_PASSWORD:-}"
                  if [[ -f "${MYSQL_ROOT_PASSWORD_FILE:-}" ]]; then
                      password_aux=$(cat "$MYSQL_ROOT_PASSWORD_FILE")
                  fi
```
3. Replace `master` in the script with `primary` as the default name for the `primary` host is now "primary", and no longer "master"

florian-besser:
### Description of the change
Update docs as per https://github.com/bitnami/charts/issues/28195
- Use bash
- Default primary node name includes "primary", not "master"
- Read root password from env / file instead of hardcoding

### Benefits

The docs can be copy/pasted without breaking

### Possible drawbacks

None that I could see.

### Applicable issues

- fixes #28195

### Additional information

I'm unclear whether you want me to increase the chart's version.
https://github.com/bitnami/charts/blob/main/CONTRIBUTING.md says it must be done, whereas the PR default template says it is not necessary for pure README updates. I've included a bump to be safe.

### Checklist


dgomezleon (approved):

dgomezleon (approved):

PR #27253, opened 2024-06-17
[bitnami/argo-cd] fix: Use the right port name for livenessProbe.

mathieulaude:
### Description of the change

Fix regression introduced by #26367. See https://github.com/bitnami/charts/pull/26367#issuecomment-2173270646
The applicationset-controller Pod is currently backoff restarting because of this erroneous liveness port.

### Benefits

No more restarts for the applicationset controller.

### Possible drawbacks

No drawback

### Checklist

- [X] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/). This is *not necessary* when the changes only affect README.md files.
- [X] Variables are documented in the values.yaml and added to the `README.md` using [readme-generator-for-helm](https://github.com/bitnami/readme-generator-for-helm)
- [X] Title of the pull request follows this pattern [bitnami/<name_of_the_chart>] Descriptive title
- [X] All commits signed off and in agreement of [Developer Certificate of Origin (DCO)](https://github.com/bitnami/charts/blob/main/CONTRIBUTING.md#sign-your-work)

PR #28348, opened 2024-07-24
[bitnami/argo-cd] Remove usage of health check that does not exist (#…

ricoapon:
…26547)

ArgoCD contains a dedicated health-check URL for all applications except the applicationset controller.

### Description of the change

Fix regression introduced by https://github.com/bitnami/charts/pull/26367 and https://github.com/bitnami/charts/pull/27253.
The applicationset-controller pod is currently backoff restarting because of the incorrect liveness probe.

Some additional explanation: ArgoCD does indeed have specific URLs that are meant for health checks. This is done on all the pods that are running, except for the applicationset-controller. When implementing a change in this Bitnami Helm chart, it was already found by @mathieulaude that this didn't work. He modified the port, but this does not resolve the issue. The problem is that the whole path /healthz does not exist.

I did some research in the ArgoCD repository to backup my claim. You can look at the code, but the following issues/PRs also backup my claim:
* https://github.com/argoproj/argo-cd/issues/15304 (closed)
* https://github.com/argoproj/argo-cd/issues/10878 (still open, but from 2022)
* https://github.com/argoproj/argo-cd/pull/17691 (not merged, because the maintainer added a review comment that the "/healthz" url should just be implemented

### Benefits

No more restarts for the applicationset controller.

### Possible drawbacks

No drawback

### Checklist


bitnami/argo-cd/Chart.yaml
```diff
@@ -38,5 +38,5 @@ maintainers:
     url: https://github.com/bitnami/charts
 name: argo-cd
 sources:
-  - https://github.com/bitnami/charts/tree/main/bitnami/argo-cd
-version: 6.6.9
+- https://github.com/bitnami/charts/tree/main/bitnami/argo-cd
+version: 6.7.0
```
juan131, line 42: Please bump the patch version instead:
```suggestion
version: 6.6.10
```

bitnami/argo-cd/templates/applicationset/deployment.yaml
```diff
@@ -119,8 +119,7 @@ spec:
           livenessProbe: {{- include "common.tplvalues.render" (dict "value" .Values.applicationSet.customLivenessProbe "context" $) | nindent 12 }}
           {{- else if .Values.applicationSet.livenessProbe.enabled }}
           livenessProbe:
-            httpGet:
-              path: /healthz
+            tcpSocket:
```
juan131, line 122: This would affect Kubescape score give this tool recommends using different probes for liveness and readiness checks. An alternative would be using sth such as "pgrep" to check the process is alive:

```suggestion
            exec:
              command:
                - pgrep
                - -f
                - argocd-applicationset-controller
```

Please give it a try.

> Note: I know this is stupid, but it's important to avoid the chart to obtain bad scores in these security scans.

juan131 (changes requested):
Thanks for your contribution! Please check my comments

bitnami/argo-cd/Chart.yaml
ricoapon: Seems logical. Done!

bitnami/argo-cd/templates/applicationset/deployment.yaml
ricoapon: I read some blogs about using different liveness and readiness probe. Everybody does seem to agree that they should be different, so I am fine with updating this. But just having it different for the sake of being different is indeed a little stupid. It is what it is. Hopefully ArgoCD will implement the health-check properly, and that would solve the issue!

Thanks for suggesting the `pgrep` command. I didn't know this existed! Tested it, and it worked as expected. I implemented this.

juan131 (approved):
LGTM

juan131:
> Is there some preference I should be aware of for future changes?

We do commit squashing when merging the PRs, so feel free to use any strategy that works for you

PR #28519, opened 2024-07-25
[bitnami/keycloak] Allow support for gce based ingress controllers

### LINKED ISSUE #28357: [bitnami/keycloak] Cannot load AdminUI with gce based ingress
### Name and Version

bitnami/keycloak:21.7.4

amd64

I am trying to deploy Keycloak with APISIX as the ingress-controller. APISIX requires the path to be specified as /* instead of / in the ingress definition.

Due to this, post-deployment, the KC_HOSTNAME_URL env var gets specified as https://keycloak.docker.internal/* instead of https://keycloak.docker.internal/
Source: https://github.com/bitnami/charts/blob/2b15502a03c16d5c38877244650ff0d1f8f764b9/bitnami/keycloak/templates/statefulset.yaml#L230

Appears, it will also affect the KC_HOSTNAME_ADMIN_URL env var
Source: https://github.com/bitnami/charts/blob/2b15502a03c16d5c38877244650ff0d1f8f764b9/bitnami/keycloak/templates/statefulset.yaml#L223

```
auth:
  adminUser: admin
  adminPassword: 2ewF79kM0R

ingress:
  enabled: true
  ingressClassName: apisix
  hostname: keycloak.docker.internal
  path: /*
  annotations:
    cert-manager.io/cluster-issuer: my-issuer
  tls: true

proxyAddressForwarding: true
proxy: edge
proxyHeaders: xforwarded
production: true

service:
  type: ClusterIP
```

Keycloak UI should load successfully

### What do you see instead?

The Keycloak UI is forever stuck in **Loading the Admin UI** state.
Various static requests on the browser are appended with /* in the URI which causes a 404 error.

image

### Additional information

_No response_

rajan123456:
### Description of the change

Adds new options ingress.controller and adminIngress.controller to support ingress controllers such as GCE and APISIX, which require * to be appended in the ingress.path, such that the * suffix is not added into KC_HOSTNAME_ADMIN_URL and KC_HOSTNAME_URL in case the controller type is set as gce (default is default)

The specification of ingress.controller is inspired from the bitnami/harbor helm chart

### Benefits

Allow users to deploy with Ingress controllers such as GCE and APISIX and not break Admin UI functionality

### Possible drawbacks

NA

### Applicable issues

- fixes #28357

### Additional information

NA

### Checklist

### DIFF
--- a/bitnami/keycloak/templates/statefulset.yaml
+++ b/bitnami/keycloak/templates/statefulset.yaml
@@ -221,7 +221,7 @@ spec:
                 {{ ternary "https://" "http://" ( or .Values.adminIngress.tls (eq .Values.proxy "edge") (not (empty .Values.proxyHeaders)) ) -}}
                 {{- include "common.tplvalues.render" (dict "value" .Values.adminIngress.hostname "context" $) -}}
                 {{- if eq .Values.adminIngress.controller "default" }}
-                    {{- include "common.tplvalues.render" (dict "value" .Values.adminIngress.path "context" $) }}
+                  {{- include "common.tplvalues.render" (dict "value" .Values.adminIngress.path "context" $) }}
                 {{- else if eq .Values.adminIngress.controller "gce" }}
                   {{- $path := .Values.adminIngress.path -}}
                   {{- if hasSuffix "*" $path -}}
@@ -236,7 +236,7 @@ spec:
                 {{ ternary "https://" "http://" ( or .Values.ingress.tls (eq .Values.proxy "edge") (not (empty .Values.proxyHeaders)) ) -}}
                 {{- include "common.tplvalues.render" (dict "value" .Values.ingress.hostname "context" $) -}}
                 {{- if eq .Values.ingress.controller "default" }}
-                    {{- include "common.tplvalues.render" (dict "value" .Values.ingress.path "context" $) }}
+                  {{- include "common.tplvalues.render" (dict "value" .Values.ingress.path "context" $) }}
                 {{- else if eq .Values.ingress.controller "gce" }}
                   {{- $path := .Values.ingress.path -}}
                   {{- if hasSuffix "*" $path -}}
--- a/bitnami/keycloak/templates/statefulset.yaml
+++ b/bitnami/keycloak/templates/statefulset.yaml
@@ -215,39 +215,35 @@ spec:
             - name: KEYCLOAK_EXTRA_ARGS
               value: {{ .Values.extraStartupArgs | quote }}
             {{- end }}
-            {{- if and .Values.adminIngress.enabled .Values.adminIngress.hostname (eq .Values.adminIngress.controller "default") }}
+            {{- if and .Values.adminIngress.enabled .Values.adminIngress.hostname }}
             - name: KC_HOSTNAME_ADMIN_URL
               value: |-
                 {{ ternary "https://" "http://" ( or .Values.adminIngress.tls (eq .Values.proxy "edge") (not (empty .Values.proxyHeaders)) ) -}}
                 {{- include "common.tplvalues.render" (dict "value" .Values.adminIngress.hostname "context" $) -}}
-                {{- include "common.tplvalues.render" (dict "value" .Values.adminIngress.path "context" $) }}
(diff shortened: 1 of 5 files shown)


djas19:
Related to PR #28520

bitnami/keycloak/templates/statefulset.yaml
```diff
@@ -215,19 +215,39 @@ spec:
...
+            - name: KC_HOSTNAME_ADMIN_URL
+              value: |-
+                {{ ternary "https://" "http://" ( or .Values.adminIngress.tls (eq .Values.proxy "edge") (not (empty .Values.proxyHeaders)) ) -}}
+                {{- include "common.tplvalues.render" (dict "value" .Values.adminIngress.hostname "context" $) -}}
+                {{- $path := .Values.adminIngress.path -}}
+                {{- if hasSuffix "*" $path -}}
+                  {{- $path = trimSuffix "*" $path -}}
+                {{- end -}}
+                {{- include "common.tplvalues.render" (dict "value" $path "context" $) }}
             {{- end }}
```
dgomezleon, line 234: I would avoid duplicated lines. What about something like this to reduce the code?

```suggestion
            {{- if and .Values.adminIngress.enabled .Values.adminIngress.hostname }}
            - name: KC_HOSTNAME_ADMIN_URL
              value: |-
                {{ ternary "https://" "http://" ( or .Values.adminIngress.tls (eq .Values.proxy "edge") (not (empty .Values.proxyHeaders)) ) -}}
                {{- include "common.tplvalues.render" (dict "value" .Values.adminIngress.hostname "context" $) -}}
                {{- if eq .Values.adminIngress.controller "default" }}
                  {{- include "common.tplvalues.render" (dict "value" .Values.adminIngress.path "context" $) }}
               {{- else if eq .Values.adminIngress.controller "gce" }}
                 {{- $path := .Values.adminIngress.path -}}
                 {{- if hasSuffix "*" $path -}}
                   {{- $path = trimSuffix "*" $path -}}
                 {{- end -}}
                 {{- include "common.tplvalues.render" (dict "value" $path "context" $) }}
             {{- end }}
           {{- end }}
```

Please double-check correct indentation.

bitnami/keycloak/templates/statefulset.yaml
```diff
@@ -215,19 +215,39 @@ spec:
...
+            - name: KC_HOSTNAME_URL
+              value: |-
+                {{ ternary "https://" "http://" ( or .Values.ingress.tls (eq .Values.proxy "edge") (not (empty .Values.proxyHeaders)) ) -}}
+                {{- include "common.tplvalues.render" (dict "value" .Values.ingress.hostname "context" $) -}}
+                {{- $path := .Values.ingress.path -}}
+                {{- if hasSuffix "*" $path -}}
+                  {{- $path = trimSuffix "*" $path -}}
+                {{- end -}}
+                {{- include "common.tplvalues.render" (dict "value" $path "context" $) }}
             {{- end }}
```
dgomezleon, line 251: The same as above

dgomezleon (changes requested):
The changes LGMT. I just added some minor changes to reduce it a little bit.

bitnami/keycloak/templates/statefulset.yaml
rajan123456: Thanks, incorporated

bitnami/keycloak/templates/statefulset.yaml
rajan123456: Thanks, incorporated

dgomezleon (approved):

JKrehling:
This breaks any ingress controller that requires an asterisk but is not GCP.  
You should never create a conditional based around the name of a specific ingress class

dgomezleon:
> This breaks any ingress controller requiring an asterisk but not GCP. You should never create a conditional based around the name of a specific ingress class

Hi @JKrehling ,
This does not seem to be a breaking change since it keeps the default behavior and only changes environment variables. It is true that it could have used the existing `ingressClassName` param instead.

Feel free to create a separate issue by filling the Issue template if you find more problems. If you're interested in contributing a solution, we welcome you to create a pull request.

Issue #28161, opened 2024-07-18
[bitnami/keycloak] new hostname env var does not add httpRelativePath


gtudan:
### Name and Version

bitnami/keycloak:21.7.1

### What architecture are you using?

amd64

### What steps will reproduce the bug?

Install the chart with a httpRelativePath configured

### Are you using any custom parameters or values?

```yaml
keycloak:
  enabled: true
  httpRelativePath: /auth/
  ingress:
    enabled: true
    hostname: myHost.local
 ```

### What is the expected behavior?

The env var KC_HOSTNAME_URL should match the full URL: https://myHost.local/auth

### What do you see instead?

The variable only contains the hostname, but not the path:
```
KC_HOSTNAME_URL:              https://myHost.local
```

javsalgar:
Hi!

Thank you so much for reporting. I will forward it to the team but as it is not a critical feature we cannot guarantee an ETA. If you want to speed up the process, would you like to submit a PR?

gtudan:
Thanks Javier for looking into this. I've created a PR that addresses this.

javsalgar:
Thank you so much! The team will take a look

Issue #28121, opened 2024-07-16
[bitnami/keycloak] proxy:edge and httpRelativePath in combination are broken


MaxRink:
### Name and Version

bitnami/keycloak 21.6.3

### What architecture are you using?

None

### What steps will reproduce the bug?

we currently use 
```
proxy: edge
httpRelativePath: "/auth/"
ingress:
      enabled: true
      ingressClassName: keycloak
      hostname: keycloak.mytld.de
```

based upon that config i wouldd expect keycloak to listen on keycloak.mytld.de/auth/realms/master for the master real, but that isnt happening, its keycloak.mytld.de/realms/master
Thus all login into the UI fails, as all URLs returned by keyclaok in its rednered resources are wrong (/resources instead of /auth/resources for example).

Manually adding /auth to the  `KC_HOSTNAME_URL` makes everything work again, but IMO im not a friend of that approach
Keycloak should probably use x-forwarded headers behind an ingress instead of listening on an hostname

### What do you see instead?

A broken keycloak

juan131:
Hi @MaxRink 

We just merged a PR (see https://github.com/bitnami/charts/pull/27890), could you please give a try to the version `21.7.0` as soon as it gets released (usually within a couple of hours)? Sorry for the incoveniences

davidfrickert:
> Hi @MaxRink
> 
> We just merged a PR (see #27890), could you please give a try to the version `21.7.0` as soon as it gets released (usually within a couple of hours)? Sorry for the incoveniences

hey, not sure if this is the proper channel but this seems to be a breaking change since previously "proxy" held "edge" but after this change it can only have "forwarded" or "xforwarded", thus breaking existent installations.

davidfrickert:
also it seems that setting "proxy" to "forwarded" does not even work because the chart does not allow setting production mode if "proxy" is not "edge"

MaxRink:
> also it seems that setting "proxy" to "forwarded" does not even work because the chart does not allow setting production mode if "proxy" is not "edge"

Yup, that PR seems a bit hasty.
We use flux so i used postRenderers for implementing that with the 21.6.x releases
```yaml
  postRenderers:
    - kustomize:
        patches:
          # nuke the ENV var "KC_HOSTNAME_URL" from the keycloak statefulset
          - target:
              version: v1
              kind: StatefulSet
              name: keycloak
            patch: |
              apiVersion: apps/v1
              kind: StatefulSet
              metadata:
                name: keycloak
[...]

PR #28523, opened 2024-07-25
bitnami/Keycloak] Allow manual override fo KC_HOSTNAME_URL and KC_ADMIN_HOSTNAME_URL

### LINKED ISSUE #28357: [bitnami/keycloak] Cannot load AdminUI with gce based ingress

djas19:
### Description of the change

Allow users to provide their own value of the KC_HOSTNAME_URL and KC_ADMIN_HOSTNAME_URL parameters.

### Benefits

Currently a lot of assumptions are made when constructing these environment variables. This behaviour can be wanted in simple setups but it is impossible to foresee all possible scenarios. To accommodate this we can allow users to override this themselves but fallback on autogeneration if not provided.

### Possible drawbacks

The change is backwards compatible so no drawbacks are expected.

### Applicable issues

- fixes #28357 

### Additional information

Inspired out of discussions in issues: #28357 #28121 and #28161 

Closes: #28520 

### Checklist

### DIFF
--- a/bitnami/keycloak/CHANGELOG.md
+++ b/bitnami/keycloak/CHANGELOG.md
@@ -1,16 +1,12 @@
 # Changelog
 
+## 22.0.1 (2024-08-01)
+
+* bitnami/Keycloak] Allow manual override fo KC_HOSTNAME_URL and KC_ADMIN_HOSTNAME_URL ([#28523](https://github.com/bitnami/charts/pull/28523))
+
 ## 22.0.0 (2024-07-29)
 
-* [bitnami/keycloak] Release 22.0.0 ([#28563](https://github.com/bitnami/charts/pull/28563))
-
-## 21.8.0 (2024-07-26)
-
-* [bitnami/keycloak] Allow support for gce based ingress controllers (#28519) ([87b60d7](https://github.com/bitnami/charts/commit/87b60d7526474cc22fd8295732f6b1ed7b3771a6)), closes [#28519](https://github.com/bitnami/charts/issues/28519)
-
-## <small>21.7.6 (2024-07-26)</small>
-
-* [bitnami/keycloak] Fix invalid value of proxy headers when legacy proxy is used (#28530) ([372d263](https://github.com/bitnami/charts/commit/372d2638677330da509c8ff2783a2efd48484d45)), closes [#28530](https://github.com/bitnami/charts/issues/28530)
+* [bitnami/keycloak] Release 22.0.0 (#28563) ([81162c4](https://github.com/bitnami/charts/commit/81162c45a2a9759ac00ae26ad0bb5310af4597e4)), closes [#28563](https://github.com/bitnami/charts/issues/28563)
 
 ## 21.8.0 (2024-07-26)
 
--- a/bitnami/keycloak/CHANGELOG.md
+++ b/bitnami/keycloak/CHANGELOG.md
@@ -12,6 +12,14 @@
 
 * [bitnami/keycloak] Fix invalid value of proxy headers when legacy proxy is used (#28530) ([372d263](https://github.com/bitnami/charts/commit/372d2638677330da509c8ff2783a2efd48484d45)), closes [#28530](https://github.com/bitnami/charts/issues/28530)
 
+## 21.8.0 (2024-07-26)
+
+* [bitnami/keycloak] Allow support for gce based ingress controllers (#28519) ([87b60d7](https://github.com/bitnami/charts/commit/87b60d7526474cc22fd8295732f6b1ed7b3771a6)), closes [#28519](https://github.com/bitnami/charts/issues/28519)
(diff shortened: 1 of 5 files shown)


djas19:
I don't see how to fix the CI issue. Could somebody give me some pointers on how to setup this fork to successfully run the CI?

EC-Sol:
> I don't see how to fix the CI issue. Could somebody give me some pointers on how to setup this fork to successfully run the CI?

According to the PR checklist:
- [ ] Variables are documented in the values.yaml and added to the README.md using [readme-generator-for-helm](https://github.com/bitnami/readme-generator-for-helm)

The README.md file is automatically generated and updated if the variables are properly documented in the values.yaml. If you have directly modified the README.md file, you should remove those changes.

carrodher:
Thanks for your contribution! The issue is that the automation trying to update the README is not able to do it due to the lack of permissions in your branch/fork. Could you please add @bitnami-bot as developer/maintainer in your fork so it can update the README?

djas19:
Hi @carrodher, the bitnami-bot is added as a maintainer on the repo (invite pending) but the pipeline still fails.

carrodher:
Thanks! The changelog was properly updated!
The problem now seems to be conflicts caused by recent PRs such as https://github.com/bitnami/charts/pull/28519.

Could you please rebase from `main` and solve the conflicts?

singhbaljit:
Chart v22 and Keycloak v25 was released. I don't these variables are still [supported anymore](https://www.keycloak.org/docs/25.0.2/release_notes/#new-hostname-options):

> Hostname v2 options are supported by default, as the old hostname options are deprecated and will be removed in the following releases. You should migrate to them as soon as possible. New options are activated by default, so Keycloak will not recognize the old ones.

singhbaljit:
Can't we move these env vars to the config map instead of being defined directly on the statefulset? If you want to override them, you still can by defining them directly on the statefulset.

djas19:
@singhbaljit As this is blocking our updates/releases I wanted to get a fix out quickly hence why the implementation is presented this way. also providing these via `extraEnvVars` doesn't work in the current setup and caused errors for duplicate keys. Maybe that would be possible if it would be on the in the config map but I didn't look into that. Looking at #28611 it doesn't look like this would solve the issue for us and a similar approach to this or moving those variables to the config map too would be needed as it still doesn't take into account wildcards in the ingress path and/or `httpRelativePath`

singhbaljit:
@djas19 the `extraEnvVars` won't work currently because those variables get rendered on the statefulset, and there are already these variables defined, hence the duplicate key error. However, if you revise this PR to instead just move the definition of these two variables into the config map instead of adding the conditional logic to override, it will allow you to still meet your need (which to override these variables), without the conditional logic or extra parameters to the chart. I understand the need for a quick solution; to me, the moving of the variables is less code/quicker. Introducing new parameters has long term impact, because they can't be easily removed without breaking change. You can see how the override works already with the database variables. There are values being defined in the config map, and overrides defined in the statefulset. This is an established pattern in the chart/Kubernetes in general.

singhbaljit:
see alternative, https://github.com/bitnami/charts/pull/28838

jotamartos:
Sorry for the late reply. As #28838 is already merged, I understand this PR is not necessary anymore. Could you please confirm so we close it? 

Thanks

djas19:
Confirmed

djas19:
closing as #28838 is merged

PR #28543, opened 2024-07-26
[bitnami/kibana] add publicBaseUrl param

mike12806:
### Description of the change

This change maps a configuration item in server.publicBaseURL to the Kibana config file (server.publicBaseUrl). See https://www.elastic.co/guide/en/kibana/8.14/settings.html#server-publicBaseUrl

### Benefits

Kibana throws an annoying pop-up on each initial load asking you to set this variable. I'm sure it probably also provides a security benefit, but i'm primarily annoyed with the pop up :) 

### Possible drawbacks

There's no real drawbacks and arguably this could be added as a parameter in "extraConfiguration", however given Kibana is pushing this as an issue in the UI, it feels more appropriate to be an explicit configuration option IMO 

### Applicable issues

None

### Checklist

### DIFF
--- a/bitnami/kibana/templates/configmap.yaml
+++ b/bitnami/kibana/templates/configmap.yaml
@@ -23,6 +23,7 @@ data:
     server.basePath: {{ .Values.configuration.server.basePath | quote }}
     {{- end }}
     server.rewriteBasePath: {{ .Values.configuration.server.rewriteBasePath }}
+    server.publicBaseUrl: {{ .Values.configuration.publicBaseUrl}}
     {{- if .Values.extraConfiguration }}
     {{- include "common.tplvalues.render" ( dict "value" .Values.extraConfiguration "context" $ ) | nindent 4 }}
     {{- end }}
--- a/bitnami/kibana/templates/configmap.yaml
+++ b/bitnami/kibana/templates/configmap.yaml
@@ -22,8 +22,8 @@ data:
     {{- if .Values.configuration.server.basePath }}
     server.basePath: {{ .Values.configuration.server.basePath | quote }}
     {{- end }}
-    {{ if .Values.configuration.server.publicBaseUrl }}
-    server.publicBaseUrl: {{ .Values.configuration.server.publicBaseUrl | quote }}
+    {{- if or .Values.configuration.server.publicBaseUrl .Values.ingress.enabled }}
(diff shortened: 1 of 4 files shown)


bitnami/kibana/templates/configmap.yaml
```diff
@@ -22,6 +22,9 @@ data:
     {{- if .Values.configuration.server.basePath }}
     server.basePath: {{ .Values.configuration.server.basePath | quote }}
     {{- end }}
+    {{ if .Values.configuration.server.publicBaseUrl }}
+    server.publicBaseUrl: {{ .Values.configuration.server.publicBaseUrl | quote }}
+    {{- end }}
```
juan131, line 27: I wonder whether it makes sense to set this configuration property based on the ingress hostname (when provided). Sth like this:
```suggestion
    {{- if or .Values.configuration.server.publicBaseUrl .Values.ingress.enabled }}
    server.publicBaseUrl: {{ default .Values.ingress.hostname .Values.configuration.server.publicBaseUrl }}
    {{- end }}
```

juan131 (commented):
Thanks so much for this PR @mike12806 !! Please take a look to my comments when you have a chance.

mike12806:
> Thanks so much for this PR @mike12806 !! Please take a look to my comments when you have a chance.

Makes sense to me! Just accepted your change - thanks!

juan131 (approved):
LGTM

ftqualifio:
> The publicly available URL that end-users access Kibana at. Must include the protocol, hostname, port (if different than the defaults for http and https, 80 and 443 respectively), and the [server.basePath](https://www.elastic.co/guide/en/kibana/current/settings.html#server-basePath) (if configured).

`.Values.ingress.hostname` doesn't include the protocol or port.

RytisLT:
> > The publicly available URL that end-users access Kibana at. Must include the protocol, hostname, port (if different than the defaults for http and https, 80 and 443 respectively), and the [server.basePath](https://www.elastic.co/guide/en/kibana/current/settings.html#server-basePath) (if configured).
> 
> `.Values.ingress.hostname` doesn't include the protocol or port.

yep, this breaks my kibana deployment. I had `server.publicBaseUrl` as `extraConfiguration`, now it is set incorrectly because protocol is missing and kibana refuses to boot. Also I cannot change it with `extraConfiguration` because it creates duplicate key

PR #28581, opened 2024-07-30
[bitnami/common] Allow handling of new secrets after initial installation

cwrau:
### Description of the change

This changes the handling of non-existing secrets that should be generated so they will be generated if `failOnNew` is `true`.

### Benefits

This allows the user to add new secrets to a managed secret without having to specify it and letting this chart generate it.

### Possible drawbacks

Not that I can think of.

### Applicable issues

Non that I created, but we sometimes do https://github.com/teutonet/teutonet-helm-charts/blob/main/charts/base-cluster/templates/monitoring/kube-prometheus-stack/oauth-proxy-secret.yaml?plain=1#L14C204-L14C356 to work around this problem.

### Checklist

- [x] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/). This is _not necessary_ when the changes only affect README.md files.

### DIFF
--- a/bitnami/common/CHANGELOG.md
+++ b/bitnami/common/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## 2.21.0 (2024-07-30)
+## 2.21.0 (2024-07-31)
 
 * [bitnami/common] Allow handling of new secrets after initial installation ([#28581](https://github.com/bitnami/charts/pull/28581))
 
--- a/bitnami/common/CHANGELOG.md
+++ b/bitnami/common/CHANGELOG.md
@@ -1,8 +1,12 @@
 # Changelog
 
-## 2.20.5 (2024-07-16)
+## 2.21.0 (2024-07-30)
 
-* [bitnami/common] [bitnami/wordpress] Use global.storageClass for fallback, not override ([#24863](https://github.com/bitnami/charts/pull/24863))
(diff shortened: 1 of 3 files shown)


juan131:
Hi @cwrau 

I'm sorry but the changes view that GH offers it's not the best in this case and I'm unable to fully understand the changes... Could you do a quick summary of the behavior before & after your changes? Thanks in advance.

cwrau:
> Hi @cwrau
> 
> I'm sorry but the changes view that GH offers it's not the best in this case and I'm unable to fully understand the changes... Could you do a quick summary of the behavior before & after your changes? Thanks in advance.

Currently, when you add a secret to be generated after the chart was initially installed, the upgrade fails saying that the value has to be supplied, even if `failOnNew` is `false`. To be honest, I don't know what that flag is for, even when I happened to have a scenario where the corresponding logic hit, it would result in an empty secret, which is not desirable.

This changes this logic in the case of `failOnNew`, to not fail when a secret get's added.

juan131 (commented):
Thanks for providing the details! I've been reviewing the changes and it seems your approach is more robust and streamlined.

In any case, I'll ask for a second review before approving it.

javsalgar (approved):

juan131 (approved):

javsalgar:
It's this issue related with this change? https://github.com/bitnami/charts/issues/28770

PR #28598, opened 2024-07-31
[bitnami/mariadb-galera] fix: ginkgo tests

fmulero:
### Description of the change

Perform rolling update instead of scaling down and up the STS. To achieve it we added a new function in `common-utils` to annotate the statefulset template. 

### Benefits

The test is a little bit faster and we get some seconds extra to avoid the volume clean up performed in some environments.
The new function in `common-utils` allows us to have more control about the restart and we can restart the same STS several times.

### Possible drawbacks

None

### Checklist

### DIFF
--- a/.vib/mariadb-galera/ginkgo/mariadb_test.go
+++ b/.vib/mariadb-galera/ginkgo/mariadb_test.go
@@ -10,6 +10,7 @@ import (
 	. "github.com/onsi/gomega"
 	appsv1 "k8s.io/api/apps/v1"
 	batchv1 "k8s.io/api/batch/v1"
+	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 )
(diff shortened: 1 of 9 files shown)


dgomezleon (dismissed):
LGTM

.vib/mariadb-galera/ginkgo/mariadb_test.go
dgomezleon, line 1: Could you add the code to remove the jobs? Something like

```
			By("deleting the job once it has succeeded")
			err = c.BatchV1().Jobs(namespace).Delete(ctx, createDBJobName, metav1.DeleteOptions{})
			Expect(err).NotTo(HaveOccurred())
```
...
```

```
			By("deleting the job once it has succeeded")
			err = c.BatchV1().Jobs(namespace).Delete(ctx, deleteDBJobName, metav1.DeleteOptions{})
			Expect(err).NotTo(HaveOccurred())
```

Comment (fmulero) [reply .vib/mariadb-galera/ginkgo/mariadb_test.go:1]: Just a quick note, If we remove explicitily the jobs we won't remove the pods created by them. Instead of oding that  I've just set the TTL for the job. With this solution the jobs and the pods created should be removed after covering its time to live.

File: .vib/mariadb-galera/ginkgo/mariadb_test.go [hunk 1]
```diff
@@ -70,17 +74,16 @@ var _ = Describe("MariaDB Galera", Ordered, func() {
...
 
-			Eventually(func() (*appsv1.StatefulSet, error) {
-				return c.AppsV1().StatefulSets(namespace).Get(ctx, stsName, getOpts)
-			}, timeout, PollingInterval).Should(WithTransform(getAvailableReplicas, BeZero()))
-
-			By("scaling up to the original replicas")
-			ss, err = utils.StsScale(ctx, c, ss, origReplicas)
-			Expect(err).NotTo(HaveOccurred())
+			// Wait for the new annotation in the existing pods
+			for i := int(origReplicas) - 1; i >= 0; i-- {
```
fmulero, line 82: Mariadb-galera pods takes around 1 min to spin up and the rollout restart rebuilds the pods from the last one (mariadb-galera-2 > mariadb-galera-1 > mariadb-galera-0). If we start the loop in the first pod we can face a timeout.

dgomezleon (approved):

Issue #28590, opened 2024-07-30
[bitnami/keycloak] use of regex path broken due to recent change


singhbaljit:
### Name and Version

bitnami/keycloak 21.8.0

### What architecture are you using?

None

### What steps will reproduce the bug?

When using RegEx path matching for ingress (i.e. with `nginx.ingress.kubernetes.io/use-regex` annotation), the hostname passed to the controller via environment variables includes the regex, which leads to errors are Keycloak boot.

```
ERROR: Unexpected error when starting the server in (production) mode
ERROR: Failed to start quarkus
ERROR: Invalid base URL for FrontEnd URLs: [redacted]/(realms|js|resources|admin)/.*
ERROR: Illegal character in path at index 42: [redacted]/(realms|js|resources|admin)/.*
``` 

### Are you using any custom parameters or values?

`path = "/(realms|js|resources|admin)/.*"`

### What do you see instead?

The hostname includes regex path while it shouldnt be included.

singhbaljit:
For anyone curious, path matching is done so that only certain paths are exposed, per [Keycloak recommendation](https://www.keycloak.org/server/reverseproxy#_exposed_path_recommendations).

singhbaljit:
There has been a lot of back and forth on these hostname variables recently. Since the v22 of the chart was released with Keycloak v25, we should simplify configurations to match [v2 hostname options](https://www.keycloak.org/docs/25.0.2/release_notes/#new-hostname-options), see also #28586.

carrodher:
Thank you for bringing this issue to our attention. We appreciate your involvement! If you're interested in contributing a solution, we welcome you to create a pull request. The Bitnami team is excited to review your submission and offer feedback. You can find the contributing guidelines [here](https://github.com/bitnami/charts/blob/main/CONTRIBUTING.md).

Your contribution will greatly benefit the community. Feel free to reach out if you have any questions or need assistance.

singhbaljit:
I created https://github.com/bitnami/charts/pull/28838 as an alternative to https://github.com/bitnami/charts/pull/28523.

singhbaljit:
Once the #28838 is merged, you can override the hostname parameter accordingly using `extraEnvVars`.

cmdub:
[...]

PR #28611, opened 2024-07-31
[bitnami/keycloak] use hostname v2 options

### LINKED ISSUE #28586: [binamit/keycloak] Helm chart using deprecated hostname v1 env variable
### Name and Version

bitnami/keycloak 22.0

amd64

1. Deploy helm chart

Helm chart updated to use the new hostname v2 parameters.

### What do you see instead?

Keycloak 25 has changed the way hostnames work. They've released documentation on hostnames v2 and have deprecated hostname v1 parameters. This helm chart uses hostname v1 parameters as signified by the warning on startup:

```
2024-07-30 02:31:40,081 WARN  [org.keycloak.quarkus.runtime.cli.Picocli] (main) The following used run time options are UNAVAILABLE and will be ignored during build time:
 	- hostname-url: Available only when hostname:v1 feature is enabled.
```

singhbaljit:
### Description of the change

Use [Hostname v2](https://www.keycloak.org/docs/latest/upgrading/index.html#new-hostname-options) options as v1 options are not used by default.

### Benefits

Use updated configurations in alignment with upstream Keycloak project.

### Possible drawbacks

Enabling hostname-strict when hostname is more secure, but may cause issues for some users.

### Applicable issues

- fixes #28586

### Additional information

This has be merged after the image is updated, see https://github.com/bitnami/containers/pull/70466.

### Checklist

### DIFF
--- a/bitnami/keycloak/templates/statefulset.yaml
+++ b/bitnami/keycloak/templates/statefulset.yaml
@@ -218,7 +218,7 @@ spec:
               value: {{ .Values.extraStartupArgs | quote }}
             {{- end }}
             {{- if and .Values.adminIngress.enabled .Values.adminIngress.hostname }}
-            - name: KC_HOSTNAME_ADMIN_URL
+            - name: KEYCLOAK_HOSTNAME_ADMIN
               value: |-
                 {{ ternary "https://" "http://" ( or .Values.adminIngress.tls (eq .Values.proxy "edge") (not (empty .Values.proxyHeaders)) ) -}}
                 {{- include "common.tplvalues.render" (dict "value" .Values.adminIngress.hostname "context" $) -}}
@@ -233,7 +233,7 @@ spec:
                 {{- end }}
             {{- end }}
             {{- if and .Values.ingress.enabled .Values.ingress.hostname }}
-            - name: KC_HOSTNAME_URL
+            - name: KEYCLOAK_HOSTNAME
               value: |-
                 {{ ternary "https://" "http://" ( or .Values.ingress.tls (eq .Values.proxy "edge") (not (empty .Values.proxyHeaders)) ) -}}
                 {{- include "common.tplvalues.render" (dict "value" .Values.ingress.hostname "context" $) -}}
--- a/bitnami/keycloak/CHANGELOG.md
+++ b/bitnami/keycloak/CHANGELOG.md
@@ -1,8 +1,12 @@
 # Changelog
 
-## 22.0.2 (2024-08-06)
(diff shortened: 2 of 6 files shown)


bitnami/keycloak/templates/statefulset.yaml
```diff
@@ -231,7 +231,9 @@ spec:
                 {{- end }}
             {{- end }}
             {{- if and .Values.ingress.enabled .Values.ingress.hostname }}
-            - name: KC_HOSTNAME_URL
+            - name: KEYCLOAK_HOSTNAME_STRICT
+              value: "true"
```
singhbaljit, line 235: I'm on the fence whether this should be automatically enabled. It is more secure, but may be not valid in some cases.

singhbaljit: According to the [docs](https://www.keycloak.org/server/hostname):
> Should always be set to true in production, unless your reverse proxy overwrites the Host header. If enabled, the hostname option needs to be specified.

Perhaps its value is based on the `production` chart parameter.

dgomezleon: What about parameterizing it?

djas19:
Keep in mind #28121 is still open/broken. I presume this change will have impact for some users as we don't fully control how hostnames/urls are constructed causing broken behaviour in setups with wildcards or pathsuffixes like `httpRelativePath`

bitnami/keycloak/templates/statefulset.yaml
singhbaljit: added the parameter with default as `false` so its not a breaking change

singhbaljit:
@djas19 this is only focusing on option name change, to avoid making this PR too complicated. `proxy` is deprecated anyways; so I think we should focus on removing it instead. I've opened a related issue as well, #28590.

dgomezleon (commented):
The changes LGTM. Just a couple of things:
* Bump a minor version (22.1.0) since it adds a new value
* Check DCO for all the commits

singhbaljit:
Good catch, done and done. Waiting for https://github.com/bitnami/containers/pull/70466...

singhbaljit:
The conflict is due to the fact that v24 is now removed. 😄 . So, I'll just undo the v24 changes.

singhbaljit:
Container PR is merged. Do I need to update the image tag in this PR or will that be automatic/elsewhere?

singhbaljit:
Rebased, and ready to merge

dgomezleon (approved):

PR #28681, opened 2024-08-06
[bitnami/external-dns] Allow rfc2136-zone to specify multiple zones

### LINKED ISSUE #28645: [bitnami/external-dns] External dns rfc2136 zone cannot supply more than 2 zones.
### Name and Version

bitnami/external-dns 8.3.3

Currently the helm template only allows for a single zone, however external-dns can support more than one zone by adding extra argument variables similar to how the domain filter works. 

Currently the only way to add another zone is to add it with 
```yaml
extraArgs:
  rfc2136-zone: extra-zone.com
```

However since this is a map and not a array it would only be able to support one additional zone.

This will support more than one zone, however, due to the limitations there might be additional providers that allow multiple arguments that cannot be set due to how extraArgs works. Need feedback if this should be changed to how most other bitnami charts do it via. My recommendation is that this should be changed from a map to a array to be inline with other charts to reduce a dependency when arguments are added that support multiple flags. 

```yaml
extraArgs: []
```

This fix will change the value 

```yaml
rfc2136:
  zone: ""
```

to

```yaml
rfc2136:
  zones: []
```

This will change the default values file to support more than two zones.

And discuss if changing to `extraArgs: []` makes sense

The following can work for another zone, another option is manually editing the template and maintaining yourself.
```yaml
extraArgs:
  rfc2136-zone: extra-zone.com
```

opencmit2:
### Description of the change

- Modified Helm template to handle single string or list values for extraArgs.

This change allows specifying more than two rfc2136 zones in the values.yaml file, addressing the limitation where only two zones could be provided.

### Applicable issues

- fixes #28645

### Checklist


bitnami/external-dns/templates/deployment.yaml
```diff
@@ -346,12 +346,18 @@ spec:
...
+                {{- if kindIs "slice" $value }}
+                  {{- range $val := $value }}
+            - --{{ $key }}={{ $val }}
+                  {{- end }}
+                {{- else if $value }}
             - --{{ $key }}={{ $value }}
-              {{- else }}
+                {{- else }}
             - --{{ $key }}
+                {{- end }}
```
fmulero, line 360: Not sure why we are enforcing users to use a map or yaml structure here. I think we should keep it as simple and flexible as possible to avoid these problems. Something like this would cover user needs and it doesn't need to rewrite external-dns args in yaml format:
```suggestion
            {{- if .Values.extraArgs }}
            {{ ( include "common.tplvalues.render" .Values.extraArgs . )| indent 12  | trim }}
            {{- end }}
```
I'd create a new major version changing the type of the `extraArgs` value and explaining the change in the README file.

fmulero (commented):
Hi @opencmit2 thanks a lot for your contribution.

It looks good but I have a little concern, because there is unnecessary complexity around this `extraArgs` feature. Could you give it a glance and share your thoughts?

bitnami/external-dns/templates/deployment.yaml
opencmit2: Although I haven't experimented with the method you mentioned, it does simplify the extraArgs

opencmit2:
@fmulero I did what you said, but I didn't test it successfully and it's not clear if it's compatible with previous versions

opencmit2:
image

fmulero (approved):

fmulero (approved):

TomyLobo:
this PR would be useful to me

fmulero (approved):
LGTM

PR #28801, opened 2024-08-09
[bitnami/rabbitmq-cluster-operator] Add AdmissionWebhookConfig toggle

silveiralexf:
### Description of the change

This PR proposes a new toggle to allow users to skip creation of `AdmissionWebhookConfig` resource.

### Benefits

This toggle can be useful for when users have restricted cluster-wide permissions and are only allowed to deploy namespaced scope resources.

### Possible drawbacks

The only drawback is the fact that users would have to manually create the `AdmissionWebhookConfig` by themselves.

### Applicable issues

- N/A

### Additional information

Since the default value for the toggle is set to `false` we would keep compatibility with the current behaviour, without risking a breaking change for users that are unaware or don't require the use of this toggle -- only users who explicitly opt-in would make use of it.

### Checklist


fmulero (approved):
Thanks a lot @silveiralexf  for your contribution.

LGTM

PR #28841, opened 2024-08-12
[bitnami/juypterhub] postgres image missing when postgres is disabled (fluxCD/argoCD)

leasley199:
### Description of the change
When using fluxCD or argoCD `and` using an external postgres instances, the postgres image is lost in translation and causes the following error.
```
Helm install failed for release my_namespace/jupyterhub with chart jupyterhub@7.2.14: template: jupyterhub/templates/hub/deployment.yaml:82:20: executing "jupyterhub/templates/hub/deployment.yaml" at <include "common.images.image" (dict "imageRoot" .Values.postgresql.image "global" .Values.global)>: error calling include: template: jupyterhub/charts/common/templates/_images.tpl:12:39: executing "common.images.image" at <.imageRoot.registry>: can't evaluate field registry in type interface {}
``` 
This creates a catchall so that the deploy user/manifest can still lint the postgres image to the init container and deploy successfully. 

### Benefits

fluxCD and ArgoCD deployments are happy. For some odd reason this is not happening in non CD deployments. Believe it to be due to helm rendering validation baked into flux and argo.  

### Possible drawbacks
none to my knowledge, if `postgres.image` is not set, uses defaults from helm chart.

### Applicable issues
None 

### Additional information

Simply just allows the deployer to specify the postgres image for the init-container, currently fails and prevents deployments in CD tools. 

### Checklist

### DIFF
--- a/bitnami/jupyterhub/CHANGELOG.md
+++ b/bitnami/jupyterhub/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## 7.2.15 (2024-08-13)
+## 7.2.15 (2024-08-21)
 
 * [bitnami/juypterhub] postgres image missing when postgres is disabled (fluxCD/argoCD) ([#28841](https://github.com/bitnami/charts/pull/28841))
 
--- a/bitnami/jupyterhub/CHANGELOG.md
+++ b/bitnami/jupyterhub/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## 7.2.15 (2024-08-12)
+## 7.2.15 (2024-08-13)
 
 * [bitnami/juypterhub] postgres image missing when postgres is disabled (fluxCD/argoCD) ([#28841](https://github.com/bitnami/charts/pull/28841))
 
--- a/bitnami/jupyterhub/CHANGELOG.md
+++ b/bitnami/jupyterhub/CHANGELOG.md
@@ -1,8 +1,12 @@
 # Changelog
 
-## 7.2.14 (2024-07-31)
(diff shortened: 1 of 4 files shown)


fmulero (changes requested):
Nice catch!!! thanks a lot @leasley199 to work on this fix.

This is happening because the postgresql subchart is using  the `postgresql.enabled` value as condition: https://github.com/bitnami/charts/blob/9da9f3b33f7abfb1a04f42de47a9802c27166d18/bitnami/jupyterhub/Chart.yaml#L19-L22

Please take a glance to the solution adopted in concourse to follow the same pattern: https://github.com/bitnami/charts/blob/9da9f3b33f7abfb1a04f42de47a9802c27166d18/bitnami/concourse/templates/web/deployment.yaml#L70-L76

I think is worth to follow the same solution.

leasley199:
> Nice catch!!! thanks a lot @leasley199 to work on this fix.
> 
> This is happening because the postgresql subchart is using the `postgresql.enabled` value as condition:
> 
> https://github.com/bitnami/charts/blob/9da9f3b33f7abfb1a04f42de47a9802c27166d18/bitnami/jupyterhub/Chart.yaml#L19-L22
> 
> Please take a glance to the solution adopted in concourse to follow the same pattern:
> 
> https://github.com/bitnami/charts/blob/9da9f3b33f7abfb1a04f42de47a9802c27166d18/bitnami/concourse/templates/web/deployment.yaml#L70-L76
> 
> I think is worth to follow the same solution.

much better approach updated my branch with the suggested changes :) many thank!

fmulero (approved):
Thanks a lot @leasley199 

LGTM

PR #28843, opened 2024-08-12
[bitnami/flux] Added Lookup Flag when Installing CRDs

### LINKED ISSUE #28812: [bitnami/flux] Use Helm Lookup for CRDs
### Name and Version

bitnami/flux 2.3.14

If there is already an existing Custom Resource Definition on the Cluster when installing Flux, I am unable to install it with a chart. 

I desire to port a helm chart with Flux as a dependency so that I can safely template the HelmRepository/OCIRepository/HelmChart/HelmRelease CRDs. The reason I want to do this is that I have to provide a helm chart as a final product, but the growing complexity of the chart demands extra-layers of preprocessing and flexible deployment, whilst I am limited on what tools I can use, though operators are fair game.

This, however, becomes an issue when I have multiple deployments of the same chart onto the same cluster. Whilst the controllers should conduct a leader-election and prevent any operator issues, the chart itself is fighting me.

I propose adding a flag or some template that can be added to the values.yaml that when installing the CRDs will do a lookup for those CRDs already existing. This should allow for deployments.

I have considered just controlling this via InstallCRDs when installing or copying the chart and making a custom version.

The first makes installation harder and makes automation more complex, whereas having a custom version makes future updates more annoying.

jl-beast:
### Description of the change

Added a flag to the values that allows a check to see if the CRDs already exist when installing the chart.

### Benefits

Gives users more control over the install of the CRDs when installing Flux, preventing scenarios wherein you could not install flux because the CRDs already existed.

### Possible drawbacks

Added Chart Complexity and Variance.

### Applicable issues

- fixes #28812 

### Additional information

I locally tested both `helm install` and `helm template`, both with and without `--dry-run=server`. 

I am not particularly certain what to do regarding the:
`# Conditional: .Values.____.installCRDs`
on the top of each CRD file, given that I added an additional if statement. Let me know if that is relevant.

### Checklist


jotamartos (changes requested):
Hi @jl-beast,

Thanks for taking the time to create this PR. However, I can't accept this as the next version of the solution will overwrite your changes. There is a GitHub action that updates those crds based on the files in the upstream project. You can learn more about that [here](https://github.com/bitnami/charts/blob/main/.github/workflows/ci-pipeline.yml#L209).

There are 2 options here: 
* Update the "Conditional" line to only include one single if in those crd files
* Check how the upstream project enables that feature

jl-beast:
@jotamartos 

I have reviewed and updated the conditionals.

However, given that files may contain multiple different CRDs when downloaded from their remote release sources, this means that lookup must be attuned to selecting a particular one. For this purpose, I selected the first in file.

This does however introduce additional risk, in that, if a kubernetes clusters include "Provider", but not "Alert" (for example), then it will attempt to install nonetheless with the setting. My thought, however, is that the cluster's operations with a deleted CRD for flux should already break it for each controller. That is likely a bigger concern for the cluster operator than lookup failing on a chart.

jotamartos (changes requested):
Could you investigate how this is done in the upstream chart and update the templates based on that?

PR #28938, opened 2024-08-20
[bitnami/mlflow] Add support for Google Cloud Storage as artifact store

mautini:
### Description of the change

This PR follows this one https://github.com/bitnami/charts/pull/26461 and integration of `google-cloud-storage` library in bitnami mlflow Docker image: https://github.com/bitnami/containers/pull/67246

It adds support to Google Cloud Storage as backend store for mlflow artifacts. The use could specify credentials via workload identity or application credential.

### Benefits

User will be able to use Google Storage as backend for artifact storage in mlflow tracking. This is useful for other that migrate from another mlflow instance (running with GCS as backend) or that cannot use other cloud provider.

### Possible drawbacks

More complexity to the helm chart / configuration? But it's a feature supported by mlflow so I think we have to expose it.

### Checklist


andresbono:
Thank you @mautini for this contribution! The CI failed for some reason. Let me retry it and I will review the PR. Thanks!

bitnami/mlflow/templates/_helpers.tpl
```diff
@@ -648,6 +654,33 @@ Return the S3 secret access key inside the secret
...
+{{- end -}}
+
+{{/*
+Return the GCS bucket
+*/}}
+{{- define "mlflow.v0.gcs.bucket" -}}
+    {{- if and (not .Values.minio.enabled) (not .Values.externalS3.host) .Values.externalGCS.bucket -}}
+        {{- print .Values.externalGCS.bucket -}}
+    {{- end -}}
+{{- end -}}
```
andresbono, line 682: I think this helper is not adding any value. We could just get the value from `.Values.externalGCS.bucket` as we are already wrapping the calls to the helper with the same condition, right?

```suggestion
```

I guess it's to follow the same approach than for `mlflow.v0.s3.bucket`, but in that case, the helper has a relevant if/else.

bitnami/mlflow/templates/tracking/deployment.yaml
```diff
@@ -141,12 +141,17 @@ spec:
             {{- end }}
             {{- if (include "mlflow.v0.s3.enabled" .) }}
             - --artifacts-destination=s3://{{ include "mlflow.v0.s3.bucket" . }}
+            {{- else if (include "mlflow.v0.gcs.enabled" .) }}
+            - --artifacts-destination=gs://{{ include "mlflow.v0.gcs.bucket" . }}
```
andresbono, line 145:
```suggestion
            - --artifacts-destination=gs://{{ .Values.externalGCS.bucket }}
```

bitnami/mlflow/templates/tracking/deployment.yaml
```diff
@@ -141,12 +141,17 @@ spec:
...
             {{- else }}
             - --artifacts-destination={{ .Values.tracking.persistence.mountPath }}/mlartifacts
             {{- end }}
             {{- if and (not (include "mlflow.v0.s3.serveArtifacts" .)) (include "mlflow.v0.s3.enabled" .) }}
             - --default-artifact-root=s3://{{ include "mlflow.v0.s3.bucket" . }}
             - --no-serve-artifacts
+            {{- else if and (not .Values.externalGCS.serveArtifacts) (include "mlflow.v0.gcs.enabled" .) }}
+            - --default-artifact-root=gs://{{ include "mlflow.v0.gcs.bucket" . }}
```
andresbono, line 153:
```suggestion
            - --default-artifact-root=gs://{{ .Values.externalGCS.bucket }}
```

bitnami/mlflow/templates/tracking/deployment.yaml
```diff
@@ -273,6 +288,11 @@ spec:
               {{- if .Values.tracking.persistence.subPath }}
               subPath: {{ .Values.tracking.persistence.subPath }}
               {{- end }}
+            {{- if and (include "mlflow.v0.gcs.enabled" .) .Values.externalGCS.useCredentialsInSecret .Values.externalGCS.existingSecret }}
+            - name: google-application-credentials
```
andresbono, line 292: There is no volume called like that, should it be `gcs-key`? Or the other way around.

bitnami/mlflow/templates/tracking/deployment.yaml
```diff
@@ -273,6 +288,11 @@ spec:
...
+              mountPath: /bitnami/gcs/key.json
+              subPath: key.json
```
andresbono, line 294: Is this combination of `mountPath` and `subPath` configuring the mount volume as expected?

bitnami/mlflow/values.yaml
```diff
@@ -1421,3 +1422,20 @@ externalS3:
...
   bucket: "mlflow"
   serveArtifacts: true
+
+## @section External Google Cloud Storage parameters
+## All of these values are only used when minio.enabled is set to false and externalS3 is not configured (host is empty)
+## @param externalGCS.bucket GCS bucket name. Activate gcs artifact storage if set
+## @param externalGCS.googleCloudProject Google Cloud Project to use (optional, needed when using "default application credentials")
+## @param externalGCS.useCredentialsInSecret Whether to read the GCS application credentials from a secret
+## @param externalGCS.existingSecret Name of an existing secret key containing the application credentials file (used only if useCredentialsInSecret is true)
+## @param externalGCS.existingSecretKey Key in the existing secret containing the application credentials (must be set if existingSecret is set)
```
andresbono, line 1432: How does auth work if a user does not provide the credentials in a secret? Are these parameters actually optional?

andresbono (commented):
Thank you for your contribution!! 🎉  I left some suggestions and questions. Could you please look at them?

Also, could you share if you performed some specific validation of the changes on your own env? To better understand the use case as well.

bitnami/mlflow/templates/tracking/deployment.yaml
mautini: You're right. I took this part from the original MR there: https://github.com/bitnami/charts/pull/26461 as I don't use this authentication method. Let's rename it to `gcs-key`

bitnami/mlflow/templates/tracking/deployment.yaml
mautini: I also took that from the original MR. I just checked, and there was a problem with the volume configuration. I push a commit. (`path: key.json` vs `path: /bitnami/gcs/key.json`)

I tested locally using minikube and the following commands (if you want to test on your side):

```shell
kubectl create namespace test-bitnami
kubectl -n test-bitnami create secret generic gcs-secret --from-file=gcs_keys=./keys.json
kubectl -n test-bitnami apply -f deployment.yaml
```

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: wait-deployment
spec:
  replicas: 1
  selector:
    matchLabels:
      app: wait-app
  template:
    metadata:
      labels:
        app: wait-app
    spec:
      containers:
        - name: wait-container
          image: busybox
          command: ['sh', '-c', 'while true; do sleep 3600; done']
          ports:
            - containerPort: 80
          volumeMounts:
            - name: gcs-key
              mountPath: /bitnami/gcs/key.json
              subPath: key.json
      volumes:
        - name: gcs-key
          secret:
            secretName: gcs-secret
            items:
              - key: gcs_keys
                path: key.json
```

The `keys.json` is correctly mounted at `/bitnami/gcs/key.json` and this is the path specified for `GOOGLE_APPLICATION_CREDENTIALS` env variable.

mautini:
Hi @andresbono, thanks a lot for your careful review. I applied your suggested changes, responded to your questions and pushed a fix for the mounting of `key.json`. Let me know if there is something left.

bitnami/mlflow/templates/_helpers.tpl
```diff
@@ -648,6 +654,24 @@ Return the S3 secret access key inside the secret
...
+
+{{/*
+Return whether artifacts should be served from GCS
+*/}}
+{{- define "mlflow.v0.gcs.serveArtifacts" -}}
+    {{- if .Values.externalGCS.serveArtifacts  -}}
+        {{- true }}
+    {{- end -}}
+{{- end -}}
```
andresbono, line 673: Let's remove this helper too

```suggestion
```

bitnami/mlflow/templates/tracking/deployment.yaml
```diff
@@ -310,4 +330,12 @@ spec:
         {{- if .Values.tracking.extraVolumes }}
         {{- include "common.tplvalues.render" (dict "value" .Values.tracking.extraVolumes "context" $) | nindent 8 }}
         {{- end }}
+        {{- if and (include "mlflow.v0.gcs.enabled" .) .Values.externalGCS.useCredentialsInSecret .Values.externalGCS.existingSecret }}
```
andresbono, line 333:
```suggestion
        {{- if and (include "mlflow.v0.gcs.enabled" .) .Values.externalGCS.useCredentialsInSecret }}
```

bitnami/mlflow/templates/tracking/deployment.yaml
```diff
@@ -273,6 +288,11 @@ spec:
               {{- if .Values.tracking.persistence.subPath }}
               subPath: {{ .Values.tracking.persistence.subPath }}
               {{- end }}
+            {{- if and (include "mlflow.v0.gcs.enabled" .) .Values.externalGCS.useCredentialsInSecret .Values.externalGCS.existingSecret }}
```
andresbono, line 291:
```suggestion
            {{- if and (include "mlflow.v0.gcs.enabled" .) .Values.externalGCS.useCredentialsInSecret }}
```

bitnami/mlflow/values.yaml
andresbono, line 1432:
```suggestion
## @param externalGCS.existingSecret Name of an existing secret key containing the application credentials file (required when useCredentialsInSecret is true)
## @param externalGCS.existingSecretKey Key in the existing secret containing the application credentials (required when useCredentialsInSecret is true)
```

bitnami/mlflow/templates/tracking/deployment.yaml
andresbono: Thank you for double-checking it!

bitnami/mlflow/values.yaml
andresbono: Oh that's nice. Thank you for providing the context!

andresbono (commented):
Thank you for the additional context!

I just added a few minor suggestions. If you agree with them, you can apply them and the PR is ready to go 🚀

mautini:
All good for me, thanks!

andresbono (approved):
LGTM, thank you so much!

PR #29095, opened 2024-08-28
[bitnami/rabbitmq] update values.schema.json

sturman:
### Description of the change
Update `values.schema.json` for the RabbitMQ helm chart

### Benefits
Allows generating `RabbitmqValues` object so it can be used in the `cdk8s` project. [cdk8s doc](https://cdk8s.io/docs/latest/cli/import/#values-schema)

### Checklist

### DIFF
--- a/bitnami/rabbitmq/CHANGELOG.md
+++ b/bitnami/rabbitmq/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## 14.6.10 (2024-08-28)
+## 14.6.10 (2024-08-29)
 
 * [bitnami/rabbitmq] update values.schema.json ([#29095](https://github.com/bitnami/charts/pull/29095))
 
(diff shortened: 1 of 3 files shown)


migruiz4:
Hi @sturman,

I'm sorry but I will have to reject this PR. 

In the past we attempted to include a complete `values.schema.json` in our charts, but it ended up causing conflicts with users deployments and we had to revert it (https://github.com/bitnami/charts/pull/19335). 

The reason why we had to revert the change was that having a full schema was too strict on installations and it had conflict with some values that are designed to accept different types of values.

To generate a new 'full' `values.schema.json`, we first need to investigate what changes have to be performed in the schema to avoid any issues and implement a solution in the values metadata tags (e.g. `@param`, etc.) and the [readme-generator](https://github.com/bitnami/readme-generator-for-helm) tool (which would be used to automatically generate and maintain the schemas).

Once the tool is adapted, we could evaluate re-enabling the automatic schema generation as in the previous PR, but we currently lack the bandwidth to implement the required changes ourselves.

PR #29201, opened 2024-09-04
[bitnami/postgresql] collect metrics as postgres user

### LINKED ISSUE #20247: [bitnami/postgresql] permission denied for function pg_ls_waldir
### Name and Version

bitnami/postgres

None

Error in Postgres container:
```
permission denied for function pg_ls_waldir
```
Error in Metrics container:
```
ts=2023-10-16T09:48:46.558Z caller=collector.go:199 level=error msg="collector failed" name=wal duration_seconds=0.01408122 err="pq: permission denied for function pg_ls_waldir"
```

_No response_

This command should be executed in init container by default.

```
GRANT EXECUTE ON FUNCTION pg_ls_waldir() TO <User>
```

### What do you see instead?

User doesn't have the permission to use the function.

### Additional information

_No response_

### LINKED ISSUE #29156: [postgresql] when specifying a custom user metrics can't be collected anymore
### Name and Version

bitnami/postgresql 15.5.12

None

1. Set `auth.username` to `myuser`
2. Set `metrics.enabled` to `true`
3. Install chart and observe logs of metrics container

`--set auth.username=myuser --set metrics.enabled=true`

Metrics should be collected from all databases regardless of whether a custom user is specified or not.

### What do you see instead?

Once a custom user is specified this user is used to collect metrics and has insufficient permissions.

### Additional information

This breaks for us in 15.5.12 as since that version all databases are monitored instead of only the database the custom user has access to.

In my opinion the `postgres` administrative user should be used to collect metrics. This also solves other related issues like #20247 I believe.

If you agree I'm happy to provide a PR for this.

tbalzer:
### Description of the change

This change intends to change the credentials used by the `metrics` container to those of the `postgres` user, regardless of whether a custom user is also created or not.

### Benefits

The `postgres` user has administrative privileges which allows this user to actually monitor all databases correctly, avoiding issues with permissions like in #29156 and #20247

### Possible drawbacks

`metrics` container has full administrative privileges, but this is also the case if no custom user is created. Using the custom user for this however does not seem to be the right approach. It could be improved further by using a dedicated monitoring/metrics user.

### Applicable issues

- fixes #29156
- fixes #20247 

### Checklist


tbalzer:
Hey @jotamartos , what do you think regarding these changes?

bitnami/postgresql/templates/primary/statefulset.yaml
```diff
@@ -534,19 +534,18 @@ spec:
           env:
             - name: DATA_SOURCE_URI
               value: {{ printf "127.0.0.1:%d/postgres?sslmode=disable" (int (include "postgresql.v1.service.port" .)) }}
-            {{- $pwdKey := ternary (include "postgresql.v1.adminPasswordKey" .) (include "postgresql.v1.userPasswordKey" .) (or (eq $customUser "postgres") (empty $customUser)) }}
             {{- if .Values.auth.usePasswordFiles }}
             - name: DATA_SOURCE_PASS_FILE
-              value: {{ printf "/opt/bitnami/postgresql/secrets/%s" $pwdKey }}
+              value: {{ printf "/opt/bitnami/postgresql/secrets/%s" (include "postgresql.v1.adminPasswordKey" .) }}
```
ltsuda, line 539: Hello, even though I have zero knowledge on all of this, are these changes working?

If we look at the following configuration, `adminPasswordKey` and `adminPasswordKey` is only used if  `existingSecret` is set, then `auth.*` field would be ignored.
https://github.com/bitnami/charts/blob/main/bitnami/postgresql/values.yaml#L162

Should these `metrics` configuration considere both, `auth.*` and/or `existingSecret` fields? I'm asking because I copied these changes and I still have this error `ERROR:  permission denied for function pg_ls_waldi`

jotamartos (approved):
Accepting this as this is the default approach when no custom user is defined. Ideally, we should create a separate user for the metrics, but that's something that can be done in a different contribution.

diranged:
@jotamartos Is there a reason this wasn't also done to the HA chart?

Stevenpc3:
This doesn't solve for "custom" user or in postgresql-ha

javolek:
The Problem is actually back again since 16.4.7, because this change was effectivelly reverted - see: ca8f9307b80e6a192b51d333fbe8b927b2949be1. See also 8f4bc7b (16.2.4)

jotamartos:
Thanks for pinging us about this. I'll review the changes and update the thread as soon as I have more information.

> @jotamartos Is there a reason this wasn't also done to the HA chart?

@diranged PRs should only affect one single asset so we approved and merged this one affecting PostgreSQL. If you want to apply the same changes in PostgreSQL HA, you can always [contribute](https://github.com/bitnami/charts/blob/main/CONTRIBUTING.md). Our team will be more than happy to review the changes.

jotamartos:
As this PR is already merged, I created a separate issue here

https://github.com/bitnami/charts/issues/36320

PR #29270, opened 2024-09-06
[bitnami/minio]: fix missing `metrics.enabled` docs.

### LINKED ISSUE #29218: [bitnami/minio]: `metrics.prometheusRule` also requires `metrics.enabled` which doesn't exist.
### Name and Version

bitnami/minio 14.6.7

the docs for `metrics.prometheusRule` mention `metrics.enabled` but `metrics.enabled` isn't documented in `values.yaml`. Weirdly, `metrics.serviceMonitor` doesn't require `metrics.enabled`.

Karitham:
### Description of the change

Add documentation for `metrics.enabled`, currently required by `metrics.prometheusRule.enabled`

### Applicable issues

- fixes #29218

### Additional information

I didn't want to have to bump major by adding `metrics.enabled` to `metrics.serviceMonitor` checks, so I just added a todo comment. Let me know if instead we should add it and bump major.

### Checklist


bitnami/minio/templates/servicemonitor.yaml
```diff
@@ -2,7 +2,7 @@
 Copyright Broadcom, Inc. All Rights Reserved.
 SPDX-License-Identifier: APACHE-2.0
 */}}
-
+{{- /* TODO: On major version upgrade, add .Values.metrics.enabled to that check */}}
```
juan131, line 5: Given the default values is `false`, there shouldn't be any problem on adding the parameter on a minor version? I mean, there shouldn't be any upgradability issue, right?

bitnami/minio/values.yaml
```diff
@@ -1188,6 +1188,8 @@ metrics:
   ## To allow public access without authentication for prometheus metrics set environment as follows.
   ##
   prometheusAuthType: public
+  ## @param metrics.enabled Enable metrics
+  enabled: false
```
juan131, line 1192: For consistency with other charts:
```suggestion
  ## @param metrics.enabled Enable the export of Prometheus metrics
  ##
  enabled: false
```

bitnami/minio/templates/servicemonitor.yaml
Karitham: The people that are running with `metrics.serviceMonitor.enabled=true` but don't use the `metrics.prometheusRule` should have had no reason (at least no documented reason) to set `metrics.enabled=true`, so we would be removing the service monitor. This sounds like a breaking change to me but maybe I see it the wrong way

juan131: Fair enough! Let's keep the TODO then.

juan131 (approved):
LGTM

PR #29316, opened 2024-09-09
[bitnami/kafka] Allow for domain name override

Sudmota:
### Description of the change

This PR allows the server.properties to integrate external domains into the advertised listeners. When `listeners.overrideDomain` is set to `true`, the listeners will appear in the format `<listener_name>://<pod_name>.<cluster_domain>:<port>` on all listeners.

### Benefits

This configuration is to be used when an external JKS, matching a specific domain other than `cluster.local` is used, and ensures all SSL connections (interbroker, controller, client and external) are validated through that certificate.

### Possible drawbacks

None identified

### Applicable issues

None identified

### Additional information

N/A

### Checklist

### DIFF
--- a/bitnami/kafka/CHANGELOG.md
+++ b/bitnami/kafka/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## 30.1.0 (2024-09-09)
+## 30.1.0 (2024-09-10)
 
 * [bitnami/kafka] Allow for domain name override ([#29316](https://github.com/bitnami/charts/pull/29316))
 
--- a/bitnami/kafka/CHANGELOG.md
+++ b/bitnami/kafka/CHANGELOG.md
@@ -1,8 +1,12 @@
 # Changelog
 
-## 30.0.5 (2024-08-20)
+## 30.1.0 (2024-09-09)
 
-* bitnami/kafka Fix pem auth with custom encrypted private key ([#28618](https://github.com/bitnami/charts/pull/28618))
+* [bitnami/kafka] Allow for domain name override ([#29316](https://github.com/bitnami/charts/pull/29316))
+
+## <small>30.0.5 (2024-08-23)</small>
+
+* bitnami/kafka Fix pem auth with custom encrypted private key (#28618) ([96b751e](https://github.com/bitnami/charts/commit/96b751e3eb0a6acba28e0fcbca907bb2de88fdf5)), closes [#28618](https://github.com/bitnami/charts/issues/28618)
 
 ## <small>30.0.4 (2024-08-14)</small>
 
--- a/bitnami/kafka/Chart.yaml
+++ b/bitnami/kafka/Chart.yaml
@@ -40,4 +40,4 @@ maintainers:
 name: kafka
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/kafka
-version: 30.0.5
+version: 30.1.0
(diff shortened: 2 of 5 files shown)


juan131:
Hi @Sudmota 

I find this new parameter confusing.

In both cases the cluster domain will be substituted in the advertised listeners based on the `clusterDomain` chart parameter (which is set by default to `cluster.local`). The difference seems to be related on removing the K8s svc name and namespace from the FQDN.

Can't you set any custom advertised listeners using the `listeners.advertisedListeners` parameter in case the default generated ones are no valid for you?

Sudmota:
Hello @juan131 !

Thank you for your prompt reply.
Indeed it would work, but we are seeking to enforce that each broker hosts is advertising its own pod name, instead of a fixed variable that all pods would equally advertised.

In short:
**Desired:**
Pod 1: CLIENT://pod1.example.com:9092
Pod 2: CLIENT://pod2.example.com:9092

**Undesired: What would happen by changing listeners.advertisedListeners: "CLIENT://test.example.com:9092"**
Pod 1: CLIENT://test.example.com:9092
Pod 2: CLIENT://test.example.com:9092

Which is what we are trying to address.

juan131:
Hi @Sudmota 

The current logic already uses `MY_POD_NAME` env. var which is unique per pod:

```bash
replace_placeholder "advertised-address-placeholder" "${MY_POD_NAME}.{{ $fullname }}-${POD_ROLE}-headless.{{ $releaseNamespace }}.svc.{{ $clusterDomain }}"
```

Could you please provide the chart parameters (provided via `values.yaml` or using the `--set` flag) you used to install your Helm release? Please exclusively provide the parameters you customized avoiding the ones with default values.

> Note: you can easily obtain the above parameters using `helm get values RELEASE_NAME`

I'd like to reproduce the behaviour you're mentioning.

Sudmota:
Hello @juan131 !

Here it goes. This is my whole values.yaml file as of now. Everything else inherits from the defaults. I have replaced the clusterDomain and trust/keystore passwords with valid ones.

```
clusterDomain: example.com
listeners:
  overrideDomain: true
tls:
  existingSecret: kafka-jks
  keystorePassword: <some-password>
  truststorePassword: <some-other-password>
kraft:
  controllerQuorumVoters: 0@mykafka-controller-0.example.com:9093,1@mykafka-controller-1.example.com:9093,2@mykafka-controller-2.example.com:9093
```

With this, this returns (for node 2):
```
advertised.listeners=CLIENT://mykafka-controller-2.example.com:9092,INTERNAL://mykafka-controller-2.example.com:9094
```

Whereas this line exists:
```
replace_placeholder "advertised-address-placeholder" "${MY_POD_NAME}.{{ $fullname }}-${POD_ROLE}-headless.{{ $releaseNamespace }}.svc.{{ $clusterDomain }}"
```

What we're trying to accomplish is having this:
```
replace_placeholder "advertised-address-placeholder" "${MY_POD_NAME}.{{ $clusterDomain }}"
```

juan131:
Hi @Sudmota 

With the current chart, replacing the placeholder results on the configuration below (assuming pod names is `kafka-controller-2`):

```
advertised.listeners=CLIENT://kafka-controller-2.kafka-controller-headless.test.svc.cluster.local:9092,INTERNAL://kafka-controller-2.kafka-controller-headless.test.svc.cluster.local:9094
```

This implies a unique advertised listener per Kafka broker. The only difference with your proposal is that K8s svc name and namespace is included in the FQDN.

PR #29374, opened 2024-09-12
[bitnami/external-dns] Fixed oci workload identity secret creation

dcrn:
### Description of the change

This resolves an issue when using workload identity with the external-dns oci provider. When `oci.useWorkloadIdentity` is true, the secret is needed to configure the oci provider as such, but secret generation depends on the field `oci.privateKeyFingerprint` being set even though it's not used.

I've updated the conditions to use either `oci.useWorkloadIdentity` or `oci.privateKeyFingerprint` to determine if the secret should be generated or not.

### Benefits

- Fixes an issue where `oci.privateKeyFingerprint` must be specified even when it's not needed.

### Applicable issues

None

### Checklist

### DIFF
--- a/bitnami/external-dns/CHANGELOG.md
+++ b/bitnami/external-dns/CHANGELOG.md
@@ -1,8 +1,12 @@
 # Changelog
 
-## 8.3.7 (2024-09-05)
+## 8.3.8 (2024-09-12)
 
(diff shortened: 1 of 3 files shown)


andresbono (approved):
Thank you for contributing with this fix :tada:
LGTM

PR #29376, opened 2024-09-12
[bitnami/redis] Use common password manager to handle password

LeoColomb:
### Description of the change

Switch to the common helper `common.secrets.passwords.manage` for managing the password instead of the local undocumented helper.

### Benefits

* Streamline with the other charts
* Use a documented helper
* Benefit from the common helper features

### Applicable issues

The custom helper was introduced by https://github.com/bitnami/charts/pull/8486

### Checklist

### DIFF
--- a/bitnami/redis/templates/_helpers.tpl
+++ b/bitnami/redis/templates/_helpers.tpl
@@ -222,34 +222,13 @@ Get the password key to be retrieved from Redis&reg; secret.
 {{- end -}}
 {{- end -}}
 
-
-{{/*
-Returns the available value for certain key in an existing secret (if it exists),
-otherwise it generates a random value.
-*/}}
-{{- define "getValueFromSecret" }}
-    {{- $len := (default 16 .Length) | int -}}
-    {{- $obj := (lookup "v1" "Secret" .Namespace .Name).data -}}
-    {{- if $obj }}
-        {{- index $obj .Key | b64dec -}}
-    {{- else -}}
-        {{- randAlphaNum $len -}}
-    {{- end -}}
-{{- end }}
-
 {{/*
 Return Redis&reg; password
 */}}
 {{- define "redis.password" -}}
-{{- if or .Values.auth.enabled .Values.global.redis.password }}
-    {{- if not (empty .Values.global.redis.password) }}
-        {{- .Values.global.redis.password -}}
-    {{- else if not (empty .Values.auth.password) -}}
-        {{- .Values.auth.password -}}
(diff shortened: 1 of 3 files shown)


bitnami/redis/templates/_helpers.tpl
```diff
@@ -222,34 +222,11 @@ Get the password key to be retrieved from Redis&reg; secret.
...
-{{- if or .Values.auth.enabled .Values.global.redis.password }}
-    {{- if not (empty .Values.global.redis.password) }}
-        {{- .Values.global.redis.password -}}
-    {{- else if not (empty .Values.auth.password) -}}
-        {{- .Values.auth.password -}}
-    {{- else -}}
-        {{- include "getValueFromSecret" (dict "Namespace" (include "common.names.namespace" .) "Name" (include "redis.secretName" .) "Length" 10 "Key" (include "redis.secretPasswordKey" .))  -}}
-    {{- end -}}
-{{- end -}}
+{{ include "common.secrets.passwords.manage" (dict "secret" (include "redis.secretName" .) "key" (include "redis.secretPasswordKey" .) "providedValues" (list "global.redis.password" "auth.password") "length" 10 "skipB64enc" true "skipQuote" true "context" $) }}
```
jotamartos, line 229:
```suggestion
{{- if or .Values.auth.enabled .Values.global.redis.password }}
    {{ include "common.secrets.passwords.manage" (dict "secret" (include "redis.secretName" .) "key" (include "redis.secretPasswordKey" .) "providedValues" (list "global.redis.password" "auth.password") "length" 16 "skipB64enc" true "skipQuote" true "context" $) }}
{{- end }}
```

The password was previously generated under that condition. Apart from that, the default length was 16

jotamartos (changes requested):
Thanks for your contribution, please take a look at my comment.

bitnami/redis/templates/_helpers.tpl
LeoColomb: Thanks for your review. Suggestion applied.
I still believed the default length was 10, though, with the argument when calling the previous function. https://github.com/bitnami/charts/pull/29376/files#diff-ddd93ad4c7925e478912f6a48d16e6ac7d562f34083126c97736d210470f7e04L250

jotamartos: You are right @LeoColomb, I didn't notice that and saw that the default length in the function was 16. Accepting the change now

jotamartos (approved):

PR #29562, opened 2024-09-22
[bitnami/wordpress] templates/ingress.yaml: fix wrong www.www tls name

### LINKED ISSUE #29190: [bitnami/wordpress] nginx from-to-www-redirect from example.com to www.example.com requires DNS record for www.www.example.com
### Name and Version

bitnami/wordpress 23.1.10

For Ingress-NGINX users, setting the `nginx.ingress.kubernetes.io/from-to-www-redirect` Ingress annotation to `true` enables an automatic redirect from `example.com` to `www.example.com` and vice-versa.

The direction of the redirect [depends on which host](https://github.com/kubernetes/ingress-nginx/blob/main/docs/user-guide/nginx-configuration/annotations.md#redirect-fromto-www) is first in the Ingress `hosts` list:

1. if it's `www.example.com`, then the redirect is `example.com` to `www.example.com`
2. if it's `example.com`, then the redirect is `www.example.com` to `example.com`

Of course, you still need to ensure TLS is available for both hosts.

Now the issue with the bitnami/wordpress chart: it is not possible to use the first redirect option, `example.com` to `www.example.com`, without creating a DNS record for `www.www.example.com`.

This is due to [line 53 in `bitnami/wordpress/templates/ingress.yaml`](https://github.com/bitnami/charts/blob/main/bitnami/wordpress/templates/ingress.yaml#L53) which checks whether the `nginx.ingress.kubernetes.io/from-to-www-redirect` annotation is being set on the Ingress; if so, then it always adds the additional TLS host `{{ printf "www.%s" (tpl .Values.ingress.hostname $) | quote }}`.

Thus, if you've set `.Values.ingress.hostname` to `www.example.com` so it appears first in the `hosts` list per the requirements of the annotation, then you now have a CertificateRequest for `www.example.com` and `www.www.example.com` which will not resolve until there's a DNS record for both.

This is awkward and feels unintentional. [Line 53 of `bitnami/wordpress/templates/ingress.yaml`](https://github.com/bitnami/charts/blob/main/bitnami/wordpress/templates/ingress.yaml#L53) should check whether `.Values.ingress.hostname` is already prefixed with `www.` and not create the extra host if it is.

It should be possible to use an `ingress.hostname` value of `www.example.com` _and_ set the `nginx.ingress.kubernetes.io/from-to-www-redirect` annotation, without a TLS host for `www.www.example.com` being added.

Generally, if `.Values.ingress.hostname` is prefixed with `www.`, no TLS host for `www.{{ .Values.ingress.hostname}}` should be added.

### What do you see instead?

A TLS host for `www.www.example.com` is added when using an `ingress.hostname` of `www.example.com` with `nginx.ingress.kubernetes.io/from-to-www-redirect` annotation.

EnDjeee:
### Description of the change

This should fix [#29190 ](https://github.com/bitnami/charts/issues/29190).

### Benefits

It should now be possible to use an ingress.hostname value of www.example.com and set the nginx.ingress.kubernetes.io/from-to-www-redirect annotation, without a TLS host for www.www.example.com being added.

### Applicable issues

- fixes #29190

### Checklist


jotamartos (changes requested):
Please take a look at my suggestions in https://github.com/bitnami/charts/pull/29564

EnDjeee:
> Please take a look at my suggestions in #29564

suggestion committed

jotamartos (approved):

PR #29563, opened 2024-09-22
[bitnami/nginx] templates/ingress.yaml: fix wrong www.www tls name

EnDjeee:
### Description of the change

This should fix the issue mentioned in #29190, which impacts the **nginx**, **wordpress** and **drupal** charts.

### Benefits

It should now be possible to use an ingress.hostname value of [www.example.com](http://www.example.com/) and set the nginx.ingress.kubernetes.io/from-to-www-redirect annotation, without a TLS host for [www.www.example.com](http://www.www.example.com/) being added.

### Checklist

### DIFF
--- a/bitnami/nginx/CHANGELOG.md
+++ b/bitnami/nginx/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## 18.2.1 (2024-09-22)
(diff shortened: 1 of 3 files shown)


migruiz4 (approved):
Thank you for your contribution @EnDjeee, looks good to me!

PR #29565, opened 2024-09-22
[bitnami/mongodb] Fixed arbiter anti affinity rules to prevent it from running on the same node of a primary/secondary replica

santimar:
### Description of the change

Modified arbiter `podAntiAffinity` to prevent it from running on the same node of a primary/secondary replica

### Benefits

When deploying the chart using `architecture=replicaset`, a replicaset with 1 primary, 1 secondary and 1 arbiter is created by default.
Primary and secondary pod are created on the `mongodb` statefulset and the arbiter is created on `mongo-arbiter` statefulset.
Primary/secondary and arbiter have different `antiAffinity` rules, so it may happen that the arbiter is scheduled on the same node of the primary/secondary replica.

This is a problem since, if a node with a primary/secondary replica goes down, the arbiter is supposed to participate in  election for primary.
If both the regular replica and the arbiter goes down, the replicaset can't elect a new primary and this defect the whole purpose of having an arbiter.

This change will allow for a better resiliency against node crashes. 

### Possible drawbacks

None

### Additional information

I've also hardcoded 1 replica in arbiter, since it's not safe nor advised to use multiple arbiters.
see: https://www.mongodb.com/docs/manual/core/replica-set-arbiter/#concerns-with-multiple-arbiters

Arbiter replica was already defaulting to 1, I just made it explicit. 

### Checklist

### DIFF
--- a/bitnami/mongodb/Chart.yaml
+++ b/bitnami/mongodb/Chart.yaml
@@ -39,4 +39,4 @@ maintainers:
 name: mongodb
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/mongodb
-version: 15.6.25
+version: 15.6.26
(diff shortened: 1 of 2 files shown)


juan131 (approved):
Thanks so much! Your changes make sense in order to make the chart more robust. LGTM

PR #29573, opened 2024-09-23
[bitnami/keycloak] Fix underlying env var for `.Values.proxyHeaders`

### LINKED ISSUE #29533: Setting keycloak proxy headers is broken
### Name and Version

bitnami/keycloak 21.4.1

amd64

1. Using a reverse proxy that sets the `X-Forwarded-*` headers (Traefik in our case)
2. Set `proxyHeaders` to `xforwarded`
3. notice that the header is not respected

`ingress.enable: false` because we use a custom ingress route for Traefik

The expected behaviour is that keycloak logs the client IP address by reading the `X-Forwarded-For` header set by our reverse proxy. 

### What do you see instead?

When setting `proxyHeaders: xforwarded` in values.yaml keycloak continues to log the internal IP of our reverse-proxy. 

### Additional information

I have dived into the chart I have found the issue. The env var set here: https://github.com/bitnami/charts/blob/ea6a6a963f9c8a01fc9425301263560b69b79c6d/bitnami/keycloak/templates/configmap-env-vars.yaml#L20 should be called `KC_PROXY_HEADERS` not `KEYCLOAK_PROXY_HEADERS`. If I manually set this env var then the header is respected in the keycloak logs. Unless something has changes between keycloak v24.0.5 (chart v21.4.1) and keycloak v25 (chart v on main) then you can ignore and close this issue.

matthijswolters-rl:
### Description of the change

This PR aims to fix the naming of a keycloak env var that allows the client IP to be read from forwarded headers when using a reverse-proxy. The naming of the env var was incorrect meaning that the value `proxyHeaders` set in `values.yaml` did not achieve its intended goal. After this fix the client IP is correctly set in the keycloak logs.

The relevant keycloak documentation page can be found here: https://www.keycloak.org/server/reverseproxy#_configure_the_reverse_proxy_headers

### Benefits

The client IP forwarded by a reverse proxy in either the `Forwarded` header or the `X-Forwarded-*` headers can now successfully be read by keycloak.

### Possible drawbacks

No known drawbacks.

### Applicable issues

- fixes #29533 

### Additional information

It might be valuable to do a check of all env vars set by this chart. The keycloak docs say that all keycloak env vars are prefixed with `KC_*` rather than with `KEYCLOAK_*`. The relevant keycloak docs are here: https://www.keycloak.org/server/configuration#_formats_for_configuration

### Checklist

### DIFF
--- a/bitnami/keycloak/CHANGELOG.md
+++ b/bitnami/keycloak/CHANGELOG.md
@@ -1,8 +1,12 @@
 # Changelog
 
-## 22.2.6 (2024-09-19)
+## 22.2.7 (2024-09-23)
 
-* [bitnami/keycloak] Release 22.2.6 ([#29542](https://github.com/bitnami/charts/pull/29542))
+* [bitnami/keycloak] Fix underlying env var for `.Values.proxyHeaders` ([#29573](https://github.com/bitnami/charts/pull/29573))
+
+## <small>22.2.6 (2024-09-19)</small>
+
+* [bitnami/keycloak] Release 22.2.6 (#29542) ([754f9e2](https://github.com/bitnami/charts/commit/754f9e2365ab6d0786a997e4d582e2b6ce92c237)), closes [#29542](https://github.com/bitnami/charts/issues/29542)
 
 ## <small>22.2.5 (2024-09-16)</small>
 
--- a/bitnami/keycloak/CHANGELOG.md
+++ b/bitnami/keycloak/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## 22.2.7 (2024-09-23)
+## 22.2.7 (2024-09-24)
 
 * [bitnami/keycloak] Fix underlying env var for `.Values.proxyHeaders` ([#29573](https://github.com/bitnami/charts/pull/29573))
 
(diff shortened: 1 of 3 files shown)


Kajot-dev:
I do not think it rigth. You are correct that `KC_PROXY_HEADERS` is the correct env to set the value directly.
But that skips validation. `KEYCLOAK_PROXY_HEADERS` is meant for https://github.com/bitnami/containers/blob/f75effba29a8029a866c9b41bcf444c10fe6e98f/bitnami/keycloak/25/debian-12/rootfs/opt/bitnami/scripts/libkeycloak.sh#L248 which earlier validates the value

FYI `KEYCLOAK_*` variables are used and validated by Bitnami scripts A LOT. Using `KEYCLOAK_*` instead of `KC_*` is **intended**. See: https://github.com/bitnami/containers/blob/f75effba29a8029a866c9b41bcf444c10fe6e98f/bitnami/keycloak/25/debian-12/rootfs/opt/bitnami/scripts/keycloak-env.sh#L26

Kajot-dev:
Related PRs:
https://github.com/bitnami/charts/pull/27890
https://github.com/bitnami/containers/pull/67957
aaand a fix https://github.com/bitnami/charts/pull/28530

PR #29578, opened 2024-09-23
[bitnami/rabbitmq-cluster-operator] Add PodDistruptionBidget (PBDs)

pinkavaj:
### Description of the change

Add PodDisruptionBudget (PBD) to allow run the RabbitMQ Operator in HA mode.  This is useful for production, because when the `messaging-topology-operator` pod is eg re-scheduled the `validating-webhook` cannot be called and this hinders cluster usability, if running only in single pod with no PDB.

### Benefits

Reduces time when the usability of the K8s cluster is reduced.

### Possible drawbacks

Adds PDB even when running in single pod mode, I see no issue with this, seems this is also default for most of the present charts.

### Checklist


juan131 (approved):
Thanks for this great contribution! LGTM

PR #29645, opened 2024-09-27
[bitnami/opensearch] Fix typo in dashboards.fullname helper

lindhe:
### Description of the change

This change fixes what looks like a typo in the definition of the fullname helper template for the dashboard component.

### Benefits

I assume this was a typo. Looks like a copy-paste error based on the corresponding helper template for the data component.

### Possible drawbacks

This is technically a breaking change, so I guess we cannot publish it without a major version bump. Before I change the chart version, please confirm to me that we should do that.

### Applicable issues

n/a

### Additional information

If I were you, I'd double check if the same issue exists in the bitnami/elasticsearch chart.

### Checklist


lindhe:
BTW, I've not tested this _at all_. I just looks like an obvious typo so I wanted to point it out to you. Do what you will with this PR. :)

carrodher:

lindhe:
Done!

lindhe:
Ah, sorry. I did have it on the first commit, missed it on the chart bump. Will fix it. Do Bitnami have a recommended Git hook that helps me not forget?

lindhe:
Come to think of it, I think I made that via GitHub web GUI. Have you not checked this box in the repo settings?

image

bitnami/opensearch/Chart.yaml
```diff
@@ -30,4 +30,4 @@ maintainers:
 name: opensearch
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/opensearch
-version: 1.3.7
+version: 2.0.0
```
dgomezleon, line 33: I would go with a patch instead:
```suggestion
version: 1.3.8
```

dgomezleon (commented):
The change LGTM. Just a minor change regarding the version.

bitnami/opensearch/Chart.yaml
lindhe: Sure, we can do that if you want. I just want to be clear that this might be a breaking change for some people, since the old config and new config will have different outcomes. While the old outcome was probably not the intended outcome, it's possible that people learned to expect these exact resource names and thus depend on that old incorrect behavior.

It's not so clear-cut whether bug fixes are breaking changes or not. If we had an app with an API and the API spec stayed the same but a bug was fixed, I think it's fair to call it a non-breaking change. But in our case, we have no API (or perhaps the whole chart can be considered the API spec), so it's hard to tell what's breaking or not.

lindhe: But you are the maintainer, so if you want to call this a non-breaking change that's fine by me.

dgomezleon (dismissed):

dgomezleon (approved):

PR #29667, opened 2024-09-30
[actions] CI pipeline: Check hardcoded images

migruiz4:
### Description of the change

Adds a new step to the CI pipeline that checks the chart does not include hardcoded images in its templates.

The goal of this feature is to avoid accidentally merging charts that have containers defined directly in the manifests instead of using the values.yaml.

The action searches for any `image:` key in the `${CHART}/templates` folder and checks it is formatted as `image: {{ ... }}`.

Images formatted as plain text will cause the workflow to fail, e.g. `image: foo/bar`

### Checklist

### DIFF
--- a/.github/workflows/ci-pipeline.yml
+++ b/.github/workflows/ci-pipeline.yml
@@ -114,7 +114,7 @@ jobs:
             if [[ $image != {{*}} ]]; then
               hardcoded_images+=("${image}")
             fi
-          done <<< "$(grep -REoh "image:\s+[\"']*.+[\"']*\s*$" "charts/bitnami/${CHART}/templates" | sed "s/image: [\"']*//" | sed "s/[\"']*$//")"
+          done <<< "$(grep -REoh "^\s*image:\s+[\"']*.+[\"']*\s*$" "charts/bitnami/${CHART}/templates" | sed "s/image: [\"']*//" | sed "s/[\"']*$//")"
 
           echo "${hardcoded_images[@]}"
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-pipeline.yml
```diff
@@ -89,6 +89,52 @@ jobs:
...
+        name: Look for hardcoded images
+        env:
+          CHART: ${{ needs.get-chart.outputs.chart }}
+        run: |
+          hardcoded_images=()
+          while read -r image; do
+            if [[ $image != {{*}} ]]; then
+              hardcoded_images+=("${image}")
+            fi
+          done <<< "$(grep -REoh "image:\s+[\"']*.+[\"']*\s*$" "charts/bitnami/${CHART}/templates" | sed "s/image: [\"']*//" | sed "s/[\"']*$//")"
```
juan131, line 117: Should the 1st regexp be like this instead?
```diff
-image:\s+[\"']*.+[\"']*\s*$
+^\s*image:\s+[\"']*.+[\"']*\s*$
```
Regarding the last "sed", maybe it's simpler to use `tr -d '"'`

migruiz4: I will add the `^\s*` to the regexp for extra protection in case of mismatch / false positive.

Regarding the `tr -d` I'm not so sure, since I want the command to remove both single and double quotes exclusively from the end of the string. For example, from `"{{ include "helper.example" . }}"` I want to keep inner quotes, not remove every quote.

juan131 (approved):

PR #29668, opened 2024-09-30
[bitnami/common] Add common.tplvalues.merge-overwrite helper

### LINKED ISSUE #29566: [bitnami/common] `common.tplvalues.merge` ignores precedence on falsy values
### Name and Version

bitnami/common 2.23.0

None

1. Have two dicts with the following values in your chart:

```yaml
dict1:
  enabled: true
  name: 1001
  myList:
    - "value1"
  someOtherKey: "someOtherValue"

dict2:
  enabled: false
  name: 1234
  myList: []
  anotherKey: "anotherValue"
```

2. Call the `common.tplvalues.merge` on the two dicts

```gotpl
{{- include "common.tplvalues.merge" (dict "values" (list .Values.dict2 .Values.dict1) "context" $) }}
```

EDIT: swapped the two dicts in this example, since the precedence in the helper is "from left to right". So "earlier" dicts have higher precedence than "later" ones (which is quite un-intuitive to me; see https://github.com/bitnami/charts/issues/29566#issuecomment-2382986146).

_No response_

I would expect to have a merged context like this:

```yaml
mergedContext:
  enabled: false
  name: 1234
  myList: []
  someOtherKey: "someOtherValue"
  anotherKey: "anotherValue"
```

### What do you see instead?

The following merged context:

```yaml
mergedContext:
  enabled: true
  name: 1234
  myList:
    - "value1"
  someOtherKey: "someOtherValue"
  anotherKey: "anotherValue"
```

### Additional information

Stumbled upon this issue while working on #27628 .

I'm not entirely sure if this is a bug or "dy-design". The docs state, that precedence in the helper is "consistent with http://masterminds.github.io/sprig/dicts.html#merge-mustmerge".

If this is by-design (or should not be fixed due to backwards compatibility reasons), I would propose to add a new helper, that has the behavior I would expect from this one already:

```gotpl
{{/*
Merge a list of values that contains template after rendering them.
Later values taking higher precedence. So merging two dicts with values `true` and `false` for the same key will result
in the value of the latter on that key. When using the default merge functionality this will always result in `true`.
Usage:
{{ include "common.tplvalues.merge-with-precedence" ( dict "values" (list .Values.path.to.the.Value1 .Values.path.to.the.Value2) "context" $ ) }}
*/}}
{{- define "common.tplvalues.merge-with-precedence" -}}
{{- $dst := dict -}}
{{- range .values -}}
{{- $val := include "common.tplvalues.render" (dict "value" . "context" $.context "scope" $.scope) | fromYaml }}
{{- range $key := keys $val }}
{{- $_ := set $dst $key (get $val $key) }}
{{- end }}
{{- end -}}
{{ $dst | toYaml }}
{{- end -}}
```

maxnitze:
### Description of the change

Adding a new helper to the Bitnami common chart. This helper uses the sprig function [mergeOverwrite](https://masterminds.github.io/sprig/dicts.html#mergeoverwrite-mustmergeoverwrite) to merge multiple dictionaries into one.

### Benefits

* merging two or more dicts with precedence consistent with [mergeOverwrite](https://masterminds.github.io/sprig/dicts.html#mergeoverwrite-mustmergeoverwrite)
   * precedence "from **right to left**"
* fixes issues, where it is not possible to overwrite "truthy" values with "falsy" values, even when the "falsy" values should have precedence according to the docs
   * for more see #29566 and https://github.com/Masterminds/sprig/issues/134

### Possible drawbacks

None

### Applicable issues

- needed for #27628 
- fixes #29566 
- Masterminds/sprig issue for reference https://github.com/Masterminds/sprig/issues/134

### Additional information

%

### Checklist

### DIFF
--- a/bitnami/common/Chart.yaml
+++ b/bitnami/common/Chart.yaml
@@ -6,7 +6,7 @@ annotations:
   licenses: Apache-2.0
 apiVersion: v2
 # Please make sure that version and appVersion are always the same.
-appVersion: 2.23.0
+appVersion: 2.24.0
 description: A Library Helm Chart for grouping common logic between bitnami charts. This chart is not deployable by itself.
 home: https://bitnami.com
 icon: https://bitnami.com/downloads/logos/bitnami-mark.png
--- a/bitnami/common/Chart.yaml
+++ b/bitnami/common/Chart.yaml
@@ -23,4 +23,4 @@ name: common
 sources:
   - https://github.com/bitnami/charts/tree/main/bitnami/common
 type: library
-version: 2.23.0
+version: 2.24.0
--- a/bitnami/common/CHANGELOG.md
+++ b/bitnami/common/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## 2.24.0 (2024-09-30)
+## 2.24.0 (2024-10-01)
 
 * [bitnami/common] Add common.tplvalues.merge-overwrite helper ([#29668](https://github.com/bitnami/charts/pull/29668))
 
--- a/bitnami/common/CHANGELOG.md
+++ b/bitnami/common/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## 2.24.0 (2024-10-01)
+## 2.24.0 (2024-10-02)
 
 * [bitnami/common] Add common.tplvalues.merge-overwrite helper ([#29668](https://github.com/bitnami/charts/pull/29668))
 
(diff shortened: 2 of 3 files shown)


bitnami/common/Chart.yaml
```diff
@@ -23,4 +23,4 @@ name: common
 sources:
   - https://github.com/bitnami/charts/tree/main/bitnami/common
 type: library
-version: 2.23.0
+version: 2.24.0
```
juan131, line 26: Please also bump the `appVersion`. We keep them sync'ed for this special chart

juan131 (changes requested):
Thanks so much for this great contribution! Take a look at my comments

juan131 (approved):

PR #29683, opened 2024-10-01
[bitnami/zookeeper] Rename metrics port and expose scheme, and tlsConfig for Istio compatibility

cpepper96:
### Description of the change

Hi folks, this PR is to rename the `port` name for the metrics Service and to expose the `scheme` and `tlsConfig` fields for the ServiceMonitor. 

### Benefits

This will enable Prometheus metrics to be configured properly when deploying ZooKeeper to a cluster configured with Istio. The metrics service port name is `tcp-metrics` causing Istio to [automatically select](https://istio.io/latest/docs/ops/configuration/traffic-management/protocol-selection/#explicit-protocol-selection) the TCP protocol for this service which causes an error. By changing this to `http-metrics` Istio will correctly select the `http` protocol.

Similarly for the ServiceMonitor object, exposing the `scheme` and `tlsConfig` will enable Prometheus metrics to be integrated with Istio. 

### Possible drawbacks

I made the `scheme`, and `tlsConfig` values optional and empty by default to ensure backwards compatibility.

### Checklist


bitnami/zookeeper/templates/metrics-svc.yaml
```diff
@@ -21,6 +21,9 @@ spec:
     - name: tcp-metrics
       port: {{ .Values.metrics.service.port }}
       targetPort: metrics
+      {{- if .Values.metrics.service.appProtocol }}
+      appProtocol: {{ .Values.metrics.service.appProtocol }}
+      {{- end }}
```
juan131, line 26: Can't we simply rename the port name to `http-metrics`?

bitnami/zookeeper/Chart.yaml
```diff
@@ -28,4 +28,4 @@ maintainers:
 name: zookeeper
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/zookeeper
-version: 13.4.14
+version: 13.4.15
```
juan131, line 31: Please bump minor version instead given you're adding new chart parameters

juan131 (changes requested):
Thanks so much for this contribution! Please take a look at my comments when you have a chance.

bitnami/zookeeper/templates/metrics-svc.yaml
cpepper96: Yep! That works as well, we would just need to update the port name for the ServiceMonitor to match. 

I considered that approach but I thought that might be too big of a lift and potentially break some deployments. What are your thoughts?

juan131: What deployments do you refer to? I don't think other than metrics collector such as Prometheus would be interested on connecting to Zookeeper metrics service port. And, given we'll be adapting the ServiceMonitor in the same PR, I don't expect breaking changes.
In any case, you can always test your changes by running a `helm upgrade` after applying your changes and see if you find any errors.

cpepper96: I was thinking if someone was deploying/managing the ServiceMonitor outside of the helm chart they might not pick up on that port name change. That would probably be a rare case though.

I'd prefer to change the port name to solve this problem anyways so if it's good with you then I can implement that!

juan131: Yes please, change the port name so the `metrics.service.appProtocol` parameter isn't required.

cpepper96: Made those updates and updated the PR title/description to reflect that. Thanks for the help!

juan131 (approved):
LGTM

PR #29724, opened 2024-10-02
[bitnami/argo-cd] fix: provide a functional validation function for validation

mFranz82:
### Description of the change

Since a [change](https://github.com/mFranz82/charts/commit/8d231f86e78a6ad07b5f588a0eedebf7c21d0f9b) made 3 month before the templating fails if using ` redis.enabled=false `

```
Error: 
execution error at (argo-cd/templates/server/deployment.yaml:100:50): 
If the redis dependency is disabled you need to add an external redis host
```

The error is triggered at this point in Notes.txt
```
{{- if $message -}}
{{- end -}}
```

And occurs because the validation function could never be successful:
```
{{/*
Validate external Redis config
*/}}
{{- define "argocd.validateValues.externalRedis" -}}
{{- if not .Values.redis.enabled -}}
Argo CD: If the redis dependency is disabled you need to add an external redis port
{{- end -}}
{{- end -}}
```

As you can see in the snippet above the validating function always fails if ```redis.enabled = false```

### Benefits

You should be able again to install the chart without using internal redis.

### Possible drawbacks

No

### Applicable issues

No

### Additional information

Generally I think that the validation should just produce warnings and the templating should not fail. I am not sure why @maxnitze introduced the pattern. 

### Checklist

### DIFF
--- a/bitnami/argo-cd/templates/_helpers.tpl
+++ b/bitnami/argo-cd/templates/_helpers.tpl
@@ -289,7 +289,12 @@ Validate external Redis config
 */}}
 {{- define "argocd.validateValues.externalRedis" -}}
 {{- if not .Values.redis.enabled -}}
+    {{- if not .Values.externalRedis.port -}}
 Argo CD: If the redis dependency is disabled you need to add an external redis port
+    {{- end -}}
+    {{- if not .Values.externalRedis.host -}}
+Argo CD: If the redis dependency is disabled you need to add an external redis host
+    {{- end -}}
 {{- end -}}
 {{- end -}}
 
--- a/bitnami/argo-cd/CHANGELOG.md
+++ b/bitnami/argo-cd/CHANGELOG.md
@@ -1,8 +1,12 @@
 # Changelog
 
-## 7.0.15 (2024-10-01)
+## 7.0.16 (2024-10-08)
 
-* [bitnami/argo-cd] Release 7.0.15 ([#29682](https://github.com/bitnami/charts/pull/29682))
+* [bitnami/argo-cd] fix: provide a functional validation function for validation ([#29724](https://github.com/bitnami/charts/pull/29724))
+
+## <small>7.0.15 (2024-10-01)</small>
+
+* [bitnami/argo-cd] Release 7.0.15 (#29682) ([6b89f04](https://github.com/bitnami/charts/commit/6b89f04f14ccd7aa78811dea9642754899c102da)), closes [#29682](https://github.com/bitnami/charts/issues/29682)
 
 ## <small>7.0.14 (2024-09-26)</small>
 
--- a/bitnami/argo-cd/Chart.yaml
+++ b/bitnami/argo-cd/Chart.yaml
@@ -39,4 +39,4 @@ maintainers:
 name: argo-cd
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/argo-cd
-version: 7.0.15
+version: 7.0.16


maxnitze:
Hey @mFranz82 ,

Ypou can find more info on that specific change in this discussion: https://github.com/bitnami/charts/pull/27585#discussion_r1673910634

The main reason for the change is, that before it did nothing. The messages were collected but never printed. The default behavior should be the failing of the chart, as stated by @fmulero :

> I completely agree. Most of our charts fail and print an error message when the validation is not successful. In this case there was a failure in that piece of code and I thought the problem was in you code. Sorry about that.

Imho the issue is, that the validation should only yield the message when redis is not enabled **AND** externalRedis is also not enabled:

```gotpl
{{/*
Validate external Redis config
*/}}
{{- define "argocd.validateValues.externalRedis" -}}
{{- if and (not .Values.redis.enabled) (not .Values.externalRedis.enabled) -}}
Argo CD: If the redis dependency is disabled you need to add an external redis port
{{- end -}}
{{- end -}}
```

Side-Note: The validation for redis seems a bit weird:
* Why doesn't the first one `argocd.validateValues.redis` contain this validation already?
* What should the second one `argocd.validateValues.externalRedis` actually validate? The error text says something about a port?

mFranz82:
Hi @maxnitze 
Thanks for providing me further infos. 
I thought that the initial inventor of the ```argocd.validateValues.externalRedis``` validation rule just want's to check if a host and port for external redis is defined and to make it run I simply wrote this rule without further considering whether the rules even make sense. 

```
{{- define "argocd.validateValues.externalRedis" -}}
{{- if not .Values.redis.enabled -}}
    {{- if not .Values.externalRedis.port -}}
Argo CD: If the redis dependency is disabled you need to add an external redis port
    {{- end -}}
    {{- if not .Values.externalRedis.host -}}
Argo CD: If the redis dependency is disabled you need to add an external redis host
    {{- end -}}
{{- end -}}
{{- end -}}
```

Do we want to merge it like this or do you think we should fundamentally revise the rules?

maxnitze:
I have a couple of smaller issues with the current setup.  And your fix does not address them all.

# It does not matter whether or not `externalRedis` is enabled

This is a valid setup, from your validation:

```yaml
redis:
  enabled: false
externalRedis:
  enabled: false
  host: some-host
  port: 1337
```

# Both `redis` and `externalRedis` can be enabled (or disabled).

Is it a valid case, that both `redis.enabled` and `externalRedis.enabled` are `true`? In many cases when using Postgres this is a valid case. Because the first indicates, that you are using Postgres and the second, that you want an external one. If both are disabled you are using an in-memory db, for example. But is this also a valid case for Redis in ArgoCD? Can you run it without any dedicated Redis installed? Is there some in-memory solution?

IMHO the validation could be something like this:

```yaml
{{/*
Validate Redis config
*/}}
{{- define "argocd.validateValues.redis" -}}
{{- if and .Values.redis.enabled .Values.externlaRedis.enabled }}
Argo CD: You must not enable both, redis dependency and external redis
{{- end -}}
{{- if and .Values.redis.enabled .Values.redis.auth.existingSecret }}
    {{- if not .Values.redis.auth.existingSecretPasswordKey -}}
Argo CD: You need to provide existingSecretPasswordKey when an existingSecret is specified in redis dependency
    {{- end -}}
{{- else if and (not .Values.redis.enabled) .Values.externalRedis.existingSecret }}
    {{- if not .Values.externalRedis.existingSecretPasswordKey -}}
Argo CD: You need to provide existingSecretPasswordKey when an existingSecret is specified in redis
    {{- end }}
{{- end -}}
{{- end -}}

{{- define "argocd.validateValues.externalRedis" -}}
{{- if .Values.externalRedis.enabled -}}
    {{- if not .Values.externalRedis.port -}}
Argo CD: If the redis dependency is disabled you need to add an external redis port
    {{- end -}}
    {{- if not .Values.externalRedis.host -}}
Argo CD: If the redis dependency is disabled you need to add an external redis host
    {{- end -}}
{{- end -}}
{{- end -}}
```

TL;DR: added a block to validate only one of the two is enabled in the first helper and changed the second helper to check whether `externalRedis` is enabled (instead of `redis` is disabled).

What do you think about that?

migruiz4:
Hi there,

Thank you for the contribution and all the detailed discussion!

This PR looks good to me as it is, although, as @maxnitze mentions this setup would skip the validation and fail at startup:
```
redis:
  enabled: false
externalRedis:
  enabled: false
  host: some-host
  port: 1337
```
 
Now, IMHO it is the value `externalRedis.enabled` that does not make sense unless Argo CD could run without Redis, which is not supported according to [this discussion](https://github.com/argoproj/argo-cd/discussions/6618) and the ArgoCD [documentation](https://argo-cd.readthedocs.io/en/stable/developer-guide/architecture/components/#redis).

All of our charts use `externalRedis` automatically when `redis.enabled=false` (similar to `externalDatabase)`, and I don't see a reason why the argo-cd chart should be different.

My idea of a complete fix for this would be to remove the `externalRedis.enabled` value, adapt any associated conditional, and release it in a major version of the chart, but that would be out of the original scope of this PR.

Let me know what you think before I merge this PR, thanks!

migruiz4 (dismissed):

mFranz82:
Hi @migruiz4 
Thanks for your feedback. 

As @maxnitze suggested I finally changed the validation rule for external redis to this:

```
{{- define "argocd.validateValues.externalRedis" -}}
{{- if .Values.externalRedis.enabled -}}
    {{- if not .Values.externalRedis.port -}}
Argo CD: If the redis dependency is disabled you need to add an external redis port
    {{- end -}}
    {{- if not .Values.externalRedis.host -}}
Argo CD: If the redis dependency is disabled you need to add an external redis host
    {{- end -}}
{{- end -}}
{{- end -}}
```

I think we should leave it as it is for now and revise it at a later date...
Thank you for your participation!

bitnami/argo-cd/templates/_helpers.tpl
```diff
@@ -288,8 +288,13 @@ Argo CD: You need to provide existingSecretPasswordKey when an existingSecret is
 Validate external Redis config
 */}}
 {{- define "argocd.validateValues.externalRedis" -}}
-{{- if not .Values.redis.enabled -}}
+{{- if not .Values.externalRedis.enabled -}}
```
migruiz4, line 291: This one line would be very problematic because it won't allow deploying with default values (redis subchart):
```
$ helm install --dry-run argo-cd bitnami/argo-cd
Error: INSTALLATION FAILED: execution error at (argo-cd/templates/NOTES.txt:60:4): 
VALUES VALIDATION:
Argo CD: If the redis dependency is disabled you need to add an external redis host
```

My suggestion would be to revert to:
```suggestion
{{- if not .Values.redis.enabled -}}
```

Or alternatively, if we want to consider the value `externalRedis.enabled`, then:
```suggestion
{{- if and .Values.externalRedis.enabled (not .Values.redis.enabled) -}}
```

migruiz4 (changes requested):
Please take a look at my comment

bitnami/argo-cd/templates/_helpers.tpl
mFranz82: Hey @migruiz4 
Fortunately you discovered the error 👍 

I recently revert the change I made! 

Sorry and thanks!

migruiz4 (approved):
Looks good to me! Thank you very much!

PR #29746, opened 2024-10-02
[bitnami/influxdb] Set default updateStrategy to Recreate

p4block:
If RollingUpdate, a crashloopbackoff pod will be created, as only 1 influxdb instance can grab the same database

### Benefits
Users will be able to upgrade influxdb without manual intervention

### Possible drawbacks
This change required replacing the deployment in my Argo instance, patching from RollingUpdate to Recreate is iffy. 

### Applicable issues
Crashloopbackoff because a second influx instance cannot start on a database that is being used by another. Its log output will print something along the lines of "boltdb timeout"

### Additional information
This is a faster fix than changing the chart to use a StatefulSet, but I don't know if that would be preferable for this app.

### Checklist


carrodher:
Thank you for initiating this pull request. We appreciate your effort. Yep, commits are not signed, [here](https://github.com/bitnami/charts/pull/29746/checks?check_run_id=30978806752) you can see the specific commands you can run to sign them.

Feel free to reach out if you have any questions or need assistance with the signing process.

p4block:
@carrodher signed them off now, please let me know if there's anything else

jotamartos:
Hi @p4block,

Thank you for taking the time to contribute to the project. We've been evaluating the changes and this error is caused by a mix between the default configuration and the environment you are running the application into. In this case, the accessMode for the volume is set to ReadWriteOnce, so if you have multiple nodes, the second pod may use a different node during the upgrade causing the issue you mentioned. You can easily solve this issue by configuring the pod affinity parameters or by changing the strategy type during the installation/upgrade. 

If you want to contribute, you can add a warning message that lets the users know about this 

``` 
We detect that you enable data persistence using RWO volumes. Please note that you may find issues running rolling updates. A few alternatives you have are:

- Set `influxdb.podAffinityPreset` to `hard` to ensure new pods are scheduled in the same K8s nodes.
- Set `influxdb.updateStrategy.type` to `Recreate`. Please avoid this option if you plan to scale your solution with more replicas.
```

I'm going to close the ticket now but please reopen it if necessary.

PR #29796, opened 2024-10-07
[bitnami/opensearch] Enable automated setup of snapshot policies

### LINKED ISSUE #29601: [bitnami/opensearch] Enable automated snapshots for OpenSearch
### Name and Version

bitnami/opensearch 1.3.4

To enable consistent backups, OpenSearch uses [Snapshots](https://opensearch.org/docs/latest/tuning-your-cluster/availability-and-recovery/snapshots/index/). It is pretty easy to configure snapshots using the OpenSearch dashboard, but this requires manual intervention and is hard to automate. As far as I can tell, there is no mechanism to configure this using declarative configuration in OpenSearch, so the user must interact with the OpenSearch API.

This is a pretty involved process where the user must first send a request to "register a repository" to store the snapshots in and then configure snapshot policy, all using the OpenSearch REST API. Even advanced users, who are already familiar with this process and how to use the API, have no easy way to automate this process today.

I suggest we implement a [`post-install` chart hook](https://helm.sh/docs/topics/charts_hooks/) that performs registration of a repository and then (optionally) creates a snapshot policy.

If we implement this as opt-in, this should be a non-breaking change.

In the future, this could be expanded upon to also include policy changes during upgrade/rollback operations or creating multiple policies.

I have written a Job that does this already, but since it's not integrated into the Helm chart it's a semi-automatic process that I need to manually trigger after installing the chart. It would be nicer if it was available out-of-the-box.

lindhe:
### Description of the change

This change lets users opt-in to automatically initialize snapshot repositories and snapshot policies in OpenSearch using a post-install chart hook.

### Benefits

Before this change, configuring snapshot policies was a fairly complex task that requires interacting with the OpenSearch API manually. This change automates that process and allows for easy configuration via the Helm chart's values. I believe this will enable more people to have an OpenSearch instance that is properly configured so consistent backups can be taken.

### Possible drawbacks

Considering this is an opt-in feature, I don't see any significant draw-backs.

### Applicable issues

- fixes #29601

### Additional information

I've tried to split the PR into logically separate commits, so if we want the PVC creation to be a separate PR that's easy to do. Please inspect my commits for more info!

- good with objects for repo and policy?
- strange interdependence between persistance.enabled and snapshots.enabled

### Checklist


bitnami/opensearch/README.md
```diff
@@ -94,20 +94,19 @@ initScriptsSecret=special-scripts-sensitive
...
-    mountPath: /snapshots
+snapshots:
+  enabled: true
+  persistence:
+    enabled: true
 snapshotRepoPath: "/snapshots"
 ```
 
+For details, please refer to the `snapshots.*` documentation below.
+
```
lindhe, line 109: Is this section of the README even relevant anymore? Should I remove it?

bitnami/opensearch/templates/_helpers.tpl
```diff
@@ -822,3 +822,44 @@ Return true if a TLS credentials secret object should be created
...
+*/}}
+{{- define "opensearch.snapshots.mountPath" -}}
+{{- required "Value snapshotRepoPath must be set!" $.Values.snapshotRepoPath -}}
+{{- end -}}
+
+{{/*
+Create name for snapshot API repo data ConfigMap
+*/}}
+{{- define "opensearch.snapshots.repoDataConfigMap" -}}
+{{- printf "%s-repo-data" (include "opensearch.snapshots.fullname" $) -}}
```
lindhe, line 850: Here (and in the matching helper for the policy cm name) I'm skipping the truncation to 63 characters. If the truncation leads to the suffix being dropped, we will have a resource collision, so I thought this was safer. I believe that the config map name is not limited to 63 characters, so it should be safe? What do you think?

bitnami/opensearch/values.yaml
```diff
@@ -3408,3 +3408,181 @@ dashboards:
...
+      snapshot_config:
+        repository: default # This must match a repo name in `snapshots.repositories`
+  ## @param snapshots.repositories [default: SNAPSHOT_REPO_API_BODY] Each object represents a snapshot repository in YAML form, which will be converted to JSON and then passed as the HTTP body data to the OpenSearch REST API.
+  ## NOTE: The field `settings.location` in each repo must match the snapshot repo path configured in `snapshotRepoPath`, if its `type` is `fs`.
+  ##
+  repositories:
+    default:
+      settings:
+        location: /snapshots # This must match the snapshot repo path configured in `snapshotRepoPath`
+      type: fs
```
lindhe, line 3571: I'd like some input on this design decision. Is it good to have the API body object as YAML like this? On the one hand, I like that everything is YAML so it's easy to work with. On the other hand, this kind of hides the fact that these are not "regular" values but actually YAML representations of OpenSearch API payloads. Might be clearer if we include them as JSON strings, but I think that would make it easier for users to introduce typos in their configs.

WDYT?

bitnami/opensearch/values.yaml
```diff
@@ -3408,3 +3408,181 @@ dashboards:
...
+      - ReadWriteMany
+    ## @param snapshots.persistence.annotations Persistent Volume Claim annotations
+    ##
+    annotations: {}
+    ## @param snapshots.persistence.dataSource Custom PVC data source
+    ##
+    dataSource: {}
+    ## @param snapshots.persistence.enabled Enable persistence using Persistent Volume Claims
+    ##
+    enabled: false
```
lindhe, line 3499: I'm not convinced that it makes sense to let users disable PVC for snapshots. What kind of use-case is that? I think it is a strange interdependence between `snapshots.enabled` and `snapshots.persistence.enabled` right now; only `snapshots.enabled` should be required, in my opinion. But keeping this makes it more similar to other config in the chart.

WDYT?

jotamartos:
Sorry for the delay in getting back to you. This is just a quick message to let you know that we are going to review it and will provide feedback as soon as possible.

lindhe:
Thanks, I really appreciate the ping to let me know you will get to it soon! 👍

bitnami/opensearch/values.yaml
```diff
@@ -3408,3 +3408,181 @@ dashboards:
...
+      enabled: true
+      snapshot_config:
+        repository: default # This must match a repo name in `snapshots.repositories`
+  ## @param snapshots.repositories [default: SNAPSHOT_REPO_API_BODY] Each object represents a snapshot repository in YAML form, which will be converted to JSON and then passed as the HTTP body data to the OpenSearch REST API.
+  ## NOTE: The field `settings.location` in each repo must match the snapshot repo path configured in `snapshotRepoPath`, if its `type` is `fs`.
+  ##
+  repositories:
+    default:
+      settings:
+        location: /snapshots # This must match the snapshot repo path configured in `snapshotRepoPath`
```
jotamartos, line 3570:
```suggestion
        location: /snapshots
```

As you added a note above, I'd remove the comment here

bitnami/opensearch/values.yaml
jotamartos: As this is a yaml file, I think this configuration format is correct but you can use any other format here and change the template as required.

jotamartos (commented):
I just added some comments. Please take a look at them and let me know so we finish reviewing this PR. Thanks for your contribution

bitnami/opensearch/values.yaml
lindhe: Yeah, I agree, it feels a bit redundant. Let's remove it and also let's make sure the README generator works with my new parameters.

bitnami/opensearch/README.md
lindhe: @jotamartos Do you think I can remove it?

bitnami/opensearch/values.yaml
lindhe: Certainly. I was just wondering what would provide the best user experience for the end user/consumer of the chart.

It also makes a difference for how we document the parameters. I think it's bad if the `@param` for this object becomes a copy-paste of the OpenSearch documentation. That will be a nightmare for you to maintain. If we just document this as a opaque blob, it's easier to say "this object must be an API object according to the OpenSearch documentation" and then they can look up the examples there.

I'll try and make it a JSON implementation to see if that looks better. I'll get back once I've updated my subjective opinion on the matter.

bitnami/opensearch/values.yaml
jotamartos: I prefer keeping the "snapshots.persistence.enabled" parameter and set it to true by default to follow the same approach we have for other solutions. For example, we mount an emptyVolume if persistence is not configured in MySQL

```
  {{- if and .Values.primary.persistence.enabled .Values.primary.persistence.existingClaim }}
        - name: data
          persistentVolumeClaim:
            claimName: {{ tpl .Values.primary.persistence.existingClaim . }}
  {{- else if not .Values.primary.persistence.enabled }}
        - name: data
          emptyDir: {}
...
```

You can follow that approach here. This case covers the scenarios where you can't create/configure a volume and need to disable persistence.

bitnami/opensearch/README.md
jotamartos: If we have a configuration for the snapshots, you can remove the usage of extraVolumes and extraVolumeMounts

jotamartos:
Please take a look at the error while generating the README ```error=Please ensure all *.registry params include the [default: REGISTRY_NAME] modifier in the chart bitnami/opensearch/values.yaml file
```

**Comment by lindhe** [comment]
> Please take a look at the error while generating the README
> 
> ```
> error=Please ensure all *.registry params include the [default: REGISTRY_NAME] modifier in the chart bitnami/opensearch/values.yaml file
> ```

I'm aware of the error. It relates to this discussion: https://github.com/bitnami/charts/pull/29796#discussion_r1819152621

If we want to go for keeping the API body as YAML format, do you have any idea how we should go about configuring the `@param` argument for the documentation?

File: bitnami/opensearch/values.yaml (hunk 4 above)
Comment (lindhe) [reply bitnami/opensearch/values.yaml:3499]: > This case covers the scenarios where you can't create/configure a volume and need to disable persistence.

I don't see any value for them to have `snapshots.enabled=true` unless they also have persistent volumes. But whatever – I'm happy to do what you suggest here and keep the config option as already configured.

File: bitnami/opensearch/values.yaml (hunk 3 above)
Comment (lindhe) [reply bitnami/opensearch/values.yaml]: I just realized that my sanity checks (in [policies-cm.yaml](https://github.com/bitnami/charts/blob/c29ade4c05217605be43471a1338d1352c68af2c/bitnami/opensearch/templates/snapshots/policies-cm.yaml#L20) and [repos-cm.yaml](https://github.com/bitnami/charts/blob/c29ade4c05217605be43471a1338d1352c68af2c/bitnami/opensearch/templates/snapshots/repos-cm.yaml#L19-L24)) are entirely dependent on parsing the YAML. I think they are valuable to have, and it would be very hard to implement them unless we keep these objects as YAML code.

Thus, I'm marking this as resolved. If you disagree, please unresolve!

File: bitnami/opensearch/README.md (hunk 1 above)
Comment (lindhe) [reply bitnami/opensearch/README.md:109]: I think I know what to do with the README as soon as we have landed in a decision regarding `snapshots.persistence.enabled` in the other thread. 👍

File: bitnami/opensearch/values.yaml (hunk 4 above)
Comment (jotamartos) [reply bitnami/opensearch/values.yaml:3499]: >Some components have persistence.enabled=true by default, but not all. At least dashboards.persistence.enabled=false by default. So there is no common approach to follow here.

As you are configuring an "extra" feature, I'd disable the persistence by default. Users can always enable it when deploying the solution

>We must not set snapshots.persistence.enabled=true by default unless we also set snapshots.enabled=true by default also. Otherwise, we'll just mount volumes that are never used, which makes no sense.

That's correct. 

>I suggest we remove the snapshots.persistence.enabled

I don't like this idea. I prefer providing the users with the possibility to manage this information. 

> or snapshots.enabled=false,snapshots.persistence.enabled=false by default.

This is the approach I'd follow. Sounds good?

**Comment by jotamartos** [comment]
Just updated the thread above. Please ping me with a new comment when you update the PR. Thanks

File: bitnami/opensearch/values.yaml (hunk 4 above)
Comment (lindhe) [reply bitnami/opensearch/values.yaml:3499]: Sounds all good. Thanks for the input. 👍

File: bitnami/opensearch/README.md (hunk 1 above)
Comment (lindhe) [reply bitnami/opensearch/README.md:109]: Considering we have decided to keep `false,false` as the defaults, I think think it's helpful to the users if we document that both needs to be set to true when enabling snapshots. Thus, I think we stay with my current changes like they are now.

**Comment by lindhe** [comment]
Great. I'll just run one last test by installing it in my local cluster to try and verify nothing is obviously broken. I'll ping you once I'm done with that.

File: bitnami/opensearch/templates/snapshots/init-job.yaml [hunk 6]
```diff
@@ -0,0 +1,89 @@
...
+          imagePullPolicy: {{ .Values.snapshots.image.pullPolicy }}
+          name: {{ include "opensearch.snapshots.fullname" . }}
+          {{- if .Values.snapshots.containerSecurityContext.enabled }}
+          securityContext: {{- include "common.compatibility.renderSecurityContext" (dict "secContext" .Values.snapshots.containerSecurityContext "context" $) | nindent 12 }}
+          {{- end }}
+          {{- if .Values.snapshots.resources }}
+          resources: {{- toYaml .Values.snapshots.resources | nindent 12 }}
+          {{- else if ne .Values.snapshots.resourcesPreset "none" }}
+          resources: {{- include "common.resources.preset" (dict "type" .Values.snapshots.resourcesPreset) | nindent 10 }}
+          {{- end }}
```
lindhe, line 66: Seems like something's wrong with the resources. I'll check what's up.

lindhe: Was a typo in nindent. Should be 12, not 10. Pushing fix in a minute.

lindhe:
I found a typo in the Network Policy that I fixed. When debugging this, I also added a flag to curl to make debugging things a lot easier when curl crashes.

I've tested this now and it seems to work! Assuming you approve my recent changes, I think we are ready to merge! 🚀

lindhe:
Hey @jotamartos , if you have time to check on my latest changes that would be great. 🙂

lindhe:
After merging main I'm getting a CI error I don't believe I got before:

image

Is this a new check? I have no idea where to find the images warning list or why it's wrong.

lindhe:
@jotamartos I've resolved the conflicts now. If you could have a look at the latest changes and review before the next conflict emerge, that would be awesome! 👍

bitnami/opensearch/values.yaml
```diff
@@ -3405,3 +3405,181 @@ dashboards:
     ## @param dashboards.persistence.dataSource Custom PVC data source
     ##
     dataSource: {}
+
+## @section OpenSearch Snapshots Parameters
+snapshots:
+  ## @param snapshots.command Override default container command (useful when using custom images)
+  ##
+  command:
+    - /usr/bin/curl
```
jotamartos, line 3414: Is it necessary to add this here? I was thinking about writing the command directly in the template file

bitnami/opensearch/values.yaml
```diff
@@ -3405,3 +3405,181 @@ dashboards:
...
+    privileged: false
+    readOnlyRootFilesystem: true
+    allowPrivilegeEscalation: false
+    capabilities:
+      drop: ["ALL"]
+    seccompProfile:
+      type: "RuntimeDefault"
+  ## @param snapshots.enabled Enable automatic setup of repositories and snapshot policies
+  ##
+  enabled: false
```
jotamartos, line 3443: I'd move this to the beginning of the section to give more visibility

jotamartos (commented):
Two minor comments. Please have a look

bitnami/opensearch/values.yaml
lindhe: Not necessary at all, having it hard-coded in the template should work just as well for our purposes. Just that this lets the user customize the command, should they need to. We already let them change image, and if the do that it would very likely be required to change the command also (e.g. because `curl` exists in another path in their custom image).

There are `command` overrides in other parts of the `values.yaml` file too, the only difference here is that I included the default value as part of the `values.yaml` file instead of using logic in the template to check for custom values and then conditionally changing the command.

No rights or wrongs here, really. Just preference. But I do think that we need to accommodate for custom `command` if we want to let users provide custom images.

What do you think?

bitnami/opensearch/values.yaml
lindhe: Sure. I sorted it alphabetically, but visibility makes sense. I'll move it.

lindhe: Should `snapshots.persistence.enabled` be moved to the top of `snpashots.persistence`, so they are similar?

lindhe: I've moved both to the top. If you don't want the same for persistence, I'll just revert the commit for you. Please mark this conversation as resolved if you are happy with the solution!

PR #29818, opened 2024-10-08
[bitnami/contour] Expose NodePort for the metrics port in the Envoy service

ethervoid:
### Description of the change

When deploying Contour behind an existing load balancer, such as an AWS ALB, health checks are often required for the target group.

The Envoy health check endpoint is not exposed, making it unavailable for such checks. This change adds the option to expose the metrics port as a NodePort, enabling health checks to be performed.

### Benefits

You're able to configure your Envoy service to expose the metrics as NodePort so you're able to perform external, from the Cluster, health checks to Envoy.

For example, if you're using a [TargetGroupBinding](https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.2/guide/targetgroupbinding/targetgroupbinding/) you are not able to define those health checks pointing to the `/ready` endpoint

### Possible drawbacks

Nodes shouldn't be publicly accessible but if you define this NodePort those endpoints would be accessible from the outside.

### Checklist


bitnami/contour/Chart.yaml
```diff
@@ -32,4 +32,4 @@ maintainers:
 name: contour
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/contour
-version: 19.0.4
+version: 19.0.5
```
migruiz4, line 35: Could you please bump a minor version? (19.1.0)

bitnami/contour/templates/service_helpers.yaml
```diff
@@ -131,6 +131,15 @@ The default envoy service spec
...
       {{- end }}
+    - name: metrics
+      port: {{ .Values.envoy.service.ports.metrics }}
+      protocol: TCP
+      targetPort: {{ .Values.envoy.service.targetPorts.metrics }}
+      {{- if and (or (eq .Values.envoy.service.type "NodePort") (eq .Values.envoy.service.type "LoadBalancer")) (not (empty .Values.envoy.service.nodePorts.metrics)) }}
+      nodePort: {{ .Values.envoy.service.nodePorts.https }}
+      {{- else if eq .Values.envoy.service.type "ClusterIP" }}
+      nodePort: null
+      {{- end }}
```
migruiz4, line 142: In order to not modify the default behavior and give users control over whether to expose or not the metrics port in their service, I think we should make this conditional to a value, for example `envoy.service.exposeMetrics` which defaults to `false`.

```suggestion
    {{- if .Values.envoy.service.exposeMetrics }}
    - name: metrics
      port: {{ .Values.envoy.service.ports.metrics }}
      protocol: TCP
      targetPort: {{ .Values.envoy.service.targetPorts.metrics }}
      {{- if and (or (eq .Values.envoy.service.type "NodePort") (eq .Values.envoy.service.type "LoadBalancer")) (not (empty .Values.envoy.service.nodePorts.metrics)) }}
      nodePort: {{ .Values.envoy.service.nodePorts.https }}
      {{- else if eq .Values.envoy.service.type "ClusterIP" }}
      nodePort: null
      {{- end }}
    {{- end }}
```

migruiz4 (commented):
Thank you for your contribution @ethervoid! Could you please take a look at my comments?

migruiz4 (approved):
Thank you very much for applying my suggestions @ethervoid! Looks good to me!

PR #29865, opened 2024-10-10
[bitnami/postgresql-ha] Fix authentication and add configuration options to backup cronjob

lyeq:
### Description of the change
This PR fixes an authentication problem with the backup cronjob and introduces new configuration options for it.

If the username is not equal to `postgres`, this account doesn't have full access to the database. It is therefore not possible to perform a full backup (`pg_dumpall`) with this user. Also because of a misspelling, the password of the default user is used instead of that of the admin user. This PR changes the username for the backup cronjob to `postgres` and fixes the issue with the password.

Additionally the common configuration options `extraEnvVars`, `extraEnvVarsCM`, `extraEnvVarsSecret`, `extraVolumes` and `extraVolumeMounts` are added to the container of the backup cronjob.

### Benefits

- Default backup command works, even if the parameter `username` is not equal to `postgres`
- Better configurability of backup job

### Possible drawbacks

- Backup job always runs under user with the highest privileges

### Applicable issues

None

### Additional information

None

### Checklist


bitnami/postgresql-ha/templates/backup/cronjob.yaml
```diff
@@ -66,7 +66,7 @@ spec:
                     {{- if (include "postgresql-ha.postgresqlCreateSecret" .) }}
                     key: password
                     {{- else }}
-                    {{- if ( index ( lookup "v1" "Secret" (include "common.names.namespace" .) (include "postgresql-ha.postgresqlSecretName" .) ) ".data.postgres-password" ) }}
+                    {{- if ( index ( lookup "v1" "Secret" (include "common.names.namespace" .) (include "postgresql-ha.postgresqlSecretName" .) ) "data.postgres-password" ) }}
```
dgomezleon, line 69: Why do you need this change instead of keeping it as in `witness-statefulset.yaml` or `statefulset.yaml`?

dgomezleon (changes requested):
Hi @lyeq 

Thanks for contributing

I added a pair of suggestions. I think it would be great to share the difference of output with this change (e.g. using `helm template`) for clarification.

lyeq:
I'm sorry, I thought I would know how the lines
https://github.com/bitnami/charts/blob/6db8a2e736e2a60be2d3848d85f49d236a43136c/bitnami/postgresql-ha/templates/postgresql/statefulset.yaml#L217-L219
and
https://github.com/bitnami/charts/blob/6db8a2e736e2a60be2d3848d85f49d236a43136c/bitnami/postgresql-ha/templates/postgresql/witness-statefulset.yaml#L209-L211
are supposed to be used, but I can't figure it out.

First I thought it just checks, if a key `postgres-password` exists in the secret but this is not what this code does and neither does it make sense at this position (if the username is not `postgres` the password should also not be the one of the user `postgres`).

After that, I thought one can add a key `.data.postgres-password` with a value evaluating to `true` to the password secret, then the `postgres-password` is used instead of the `password`.
But I'm looking up a secret and can only control what's under the key `data`, the key `.data.postgres-password` never exists.

How is that supposed to work? According to my understanding, this condition can never be true and the key of the secret is always `password` and not `postgres-password`. Am I missing something?

lyeq:
I added a commit 5ac82c6f to the PR which fixes a problem I introduced with the change of the user in the backup cronjob.

If the secret is not given by the user, it doesn't exist at template evaluation time. I added that logic and outsourced the check into the template helper `postgresql-ha.postgresqlSeparatePostgresPassword`.

Now if a key `postgres-password` exists in the password secret (regardless if it is an existing secret or not), that value is used, otherwise the value of the key `password`.

dgomezleon (approved):
LGTM. Thanks!

PR #29872, opened 2024-10-11
[bitnami/sealed-secrets] Fix combination of additionalNamespaces, rbac.clusterRole=false and rbac.namespacedRoles=true

BWagenerGenerali:
### Description of the change

Currently the sealed-secret chart fails to render with the following minimal values:
```yaml
additionalNamespaces: ["test"]
rbac:
  clusterRole: false
  namespacedRoles: true
```

This configuration causes namespaced roles to be created for every entry in `additionalNamespaces`. Within the definition of the namespaced roles some `include`s are called using `.` instead of `$`. Since this is happening within a `range` `.` does not reference the root context anymore.

The error message produced:

`Error: template: sealed-secrets/templates/rolebinding.yaml:55:24: executing "sealed-secrets/templates/rolebinding.yaml" at <include "sealed-secrets.namespacedRoleName" .>: error calling include: template: sealed-secrets/templates/_helpers.tpl:46:14: executing "sealed-secrets.namespacedRoleName" at <.Values.rbac.namespacedRolesName>: can't evaluate field Values in type string`

This PR fixes the creation of namespaced roles and role bindings.

### Benefits

The creation of namespaced roles and role bindings for `additionalNamespaces` works.

### Checklist

### DIFF
--- a/bitnami/sealed-secrets/templates/rolebinding.yaml
+++ b/bitnami/sealed-secrets/templates/rolebinding.yaml
@@ -52,7 +52,7 @@ subjects:
 apiVersion: rbac.authorization.k8s.io/v1
 kind: RoleBinding
 metadata:
-  name: {{ printf "%s" (include "sealed-secrets.namespacedRoleName" .) }}
+  name: {{ printf "%s" (include "sealed-secrets.namespacedRoleName" $) }}
   namespace: {{ $additionalNamespace }}
   labels: {{- include "common.labels.standard" ( dict "customLabels" $.Values.commonLabels "context" $ ) | nindent 4 }}
   {{- if $.Values.rbac.labels }}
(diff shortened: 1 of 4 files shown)


bitnami/sealed-secrets/templates/rolebinding.yaml
```diff
@@ -64,11 +64,11 @@ metadata:
...
   apiGroup: rbac.authorization.k8s.io
   kind: Role
-  name: {{ printf "%s" (include "sealed-secrets.namespacedRoleName" .) }}
+  name: {{ printf "%s" (include "sealed-secrets.namespacedRoleName" $) }}
 subjects:
   - apiGroup: ""
     kind: ServiceAccount
     name: {{ include "sealed-secrets.serviceAccountName" $ }}
-    namespace: {{ include "sealed-secrets.namespace" $ }}
+    namespace: {{ include "common.names.namespace" $ | quote }}
```
juan131, line 72: Then we can remove `sealed-secrets.namespace` function from the `_helpers.tpl`, right?

juan131 (commented):
Thanks so much for your contribution! Please check my comments.

bitnami/sealed-secrets/templates/rolebinding.yaml
BWagenerGenerali: I can still find 23 matches for the include "sealed-secrets.namespace" in other files.
Maybe it's better to replace those usages including this one in a separate PR.

What do you think?

juan131 (approved):
LGTM

BWagenerGenerali:
Great! There's one failing check: `[CI/CD] CI Pipeline / VIB Verify (pull_request_target) Failing after 55m`

Is there anything I can do about that?

juan131:
> Great! There's one failing check: `[CI/CD] CI Pipeline / VIB Verify (pull_request_target) Failing after 55m`
> 
> Is there anything I can do about that?

We were solving internal issues in the pipeline, sorry for the inconveniences.

PR #29879, opened 2024-10-13
[bitnami/rabbitmq] Use common password/secrets manager to handle credentials

LeoColomb:
### Description of the change

Switch to the common helpers `common.secrets.passwords.manage` & `common.secrets.lookup` for managing the password instead of the local undocumented helper.

### Benefits

* Streamline with the other charts
* Use documented helpers
* Benefit from the common helpers features and maintenance

### Checklist

### DIFF
--- a/bitnami/rabbitmq/templates/secrets.yaml
+++ b/bitnami/rabbitmq/templates/secrets.yaml
@@ -6,8 +6,8 @@ SPDX-License-Identifier: APACHE-2.0
 {{- $host := printf "%s.%s.svc.%s" (include "common.names.fullname" .) (include "common.names.namespace" .) .Values.clusterDomain }}
 {{- $port := print .Values.service.ports.amqp }}
 {{- $user := print .Values.auth.username }}
-{{- $password := include "rabbitmq.password" . }}
-{{- $erlangCookie := include "rabbitmq.erlangCookie" . }}
+{{- $password := include "common.secrets.passwords.manage" (dict "secret" (include "rabbitmq.secretPasswordName" .) "key" (include "rabbitmq.secretPasswordKey" .) "length" 16 "providedValues" (list "auth.password") "skipB64enc" true "skipQuote" true "context" $) }}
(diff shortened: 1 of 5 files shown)


bitnami/rabbitmq/templates/secrets.yaml
```diff
@@ -6,8 +6,8 @@ SPDX-License-Identifier: APACHE-2.0
 {{- $host := printf "%s.%s.svc.%s" (include "common.names.fullname" .) (include "common.names.namespace" .) .Values.clusterDomain }}
 {{- $port := print .Values.service.ports.amqp }}
 {{- $user := print .Values.auth.username }}
-{{- $password := include "rabbitmq.password" . }}
-{{- $erlangCookie := include "rabbitmq.erlangCookie" . }}
+{{- $password := include "common.secrets.passwords.manage" (dict "secret" (include "rabbitmq.secretPasswordName" .) "key" (include "rabbitmq.secretPasswordKey" .) "length" 16 "providedValues" (list "auth.password") "skipB64enc" true "skipQuote" true "context" $) }}
```
juan131, line 9: Shouldn't you remove the `"skipQuote" true` parameter and directly use the resulting value at line 24 as you did for erlang cookie?

juan131 (commented):
Thanks so much for this great contribution! Please check my comments.

bitnami/rabbitmq/templates/secrets.yaml
LeoColomb: Thanks for your review, @juan131.
I thought about it, but I'm not sure whether it would match its use further:
https://github.com/bitnami/charts/blob/5fb0e97d9336d40d86c3295637d4233218b8afea/bitnami/rabbitmq/templates/secrets.yaml#L63

The quotes might be problematic, aren't they?

juan131: Ahh good catch! I haven't seen that block. Keep it as it's then. Thanks

juan131 (approved):
LGTM

PR #29881, opened 2024-10-14
[bitnami/redis] add extraPortsEnabled

shecodesmagic:
### Description of the change

This change introduces an optional flag, extraPortsEnabled, allowing users to control whether the extraPorts defined for the sentinel service should also be exposed in the headless service. If set to true, the extraPorts will be exposed in the headless service.

### Benefits

Allows users to expose additional ports (e.g., metrics, custom services) via the headless service for better integration and monitoring.
Provides flexibility by making this feature optional through the sentinel.service.headless.extraPortsEnabled flag.

### Possible drawbacks

None known. Existing deployments without sentinel.service.headless.extraPortsEnabled set to true remain unaffected.

### Applicable issues

- fixes #

### Checklist


migruiz4 (changes requested):
I'm sorry for the late response @shecodesmagic., could you please apply the following changes?

- Since this PR adds a new value, please bump a minor version (`20.3.0`).
- I think that copying the ports from the sentinel service should not be the way to proceed, since indeed they are different services that serve different purposes. Instead, I would suggest adding the value `service.headless.extraPorts`.
- Please make sure that all the commits in the PR are signed, otherwise the DCO check will fail.

shecodesmagic:
@migruiz4 thanks for the feedback. Ill update the PR accordingly.

Issue #28630, opened 2024-08-01
[bitnami/keycloak] new ports for metrics and health endpoints


oberlon:
### Name and Version

bitnami/keycloak 22.0.0

### What architecture are you using?

amd64

### What steps will reproduce the bug?

Keycloak changed the ports for metrics and health ([https://www.keycloak.org/docs/25.0.0/release_notes/#management-port-for-metrics-and-health-endpoints](https://www.keycloak.org/docs/25.0.0/release_notes/#management-port-for-metrics-and-health-endpoints))

### What is the expected behavior?

Get keycloak metrics

### What do you see instead?

404 page

ikatlinsky:
It would be nice to have a native solution in a helm chart to access metrics and health endpoints, currently, there are 2 options (as far as I can see):
- use `KC_LEGACY_OBSERVABILITY_INTERFACE` variable to rollback to keycloak <= 24 behavior 
- manually configure additional ports and mappings for pod/service/ingress

javsalgar:
Hi!

Thank you so much for the feature request. I will forward it to the team but as it is not a critical feature we cannot guarantee an ETA. If you want to speed up the process, you can contribute with a PR and the team will check it.

krankkkk:
For the next guy on how to enable metrics the mean time with the networkPolicy enabled:


```
  extraContainerPorts:
    - name: metrics
      containerPort: 9000
      protocol: TCP

  extraEnvVars:
    - name: KC_METRICS_ENABLED
      value: "true"
    #- name: KC_HTTP_METRICS_HISTOGRAMS_ENABLED
    #  value: "true"
    #- name: KC_CACHE_METRICS_HISTOGRAMS_ENABLED
    #  value: "true"

  metrics:
    enabled: true
    service:
      extraPorts:
        - name: metrics
          port: 9000
          protocol: TCP
          targetPort: metrics
    serviceMonitor:
      enabled: true
      port: metrics
      endpoints:
        - path: '{{ include "keycloak.httpPath" . }}metrics'
    prometheusRule:
      enabled: true

  networkPolicy:
    extraIngress:
    - from:
      - namespaceSelector:
          matchLabels:
            kubernetes.io/metadata.name: <your prometheus-namespace>
        podSelector:
          matchLabels:
            app.kubernetes.io/name: prometheus
      ports:
        - protocol: TCP
          port: 9000

```

nix-power:
[...]

PR #29889, opened 2024-10-14
[bitnami/keycloak] adds port 9000 to the default ingress policy

maschmi:
### Description of the change

Edits the network policy and allows port 9000 for ingress to keycloak. 

### Benefits

With Keycloak 25 the default ports for metrics and health endpoints changed from 8080/8443 to 9000. This minimal change allows for accessing the metrics and health endpoints, without automatically exposing them as service or to the world via an ingress.

### Possible drawbacks

This change basically opens port 9000 for all the clients that can access the keycloak server on the other ports. It may be favorable to create a special policy allowing only special resources, e.g. prometheus,  to access this port. However, this would need some discussion on how to do this best.

Further, this change does not create a service to access this port. However, metrics and health should be accessed per pod and not round-robin. I'm not that familiar what the best practice for this use case is. I'm happy to learn how to do it better.

### Applicable issues

[bitnami/keycloak] new ports for metrics and health endpoints #28630 

### Additional information

This minimal change basically should get the discussion going. I'm happy to learn and make it better.
I am a bit unsure if this fix should be counted as a patch or a minor version upgrade. I've decided to treat it like a patch.

### Checklist

### DIFF
--- a/bitnami/keycloak/CHANGELOG.md
+++ b/bitnami/keycloak/CHANGELOG.md
@@ -1,8 +1,12 @@
 # Changelog
 
+## 24.0.1 (2024-10-14)
+
+* [bitnami/keycloak] adds port 9000 to the default ingress policy ([#29889](https://github.com/bitnami/charts/pull/29889))
+
 ## 24.0.0 (2024-10-08)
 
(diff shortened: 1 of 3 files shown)


maschmi:
I will work more on this PR as pointed out in the discussion on #28630. I'm sorry to have it opened to soon. I will therefore close it for now.

PR #29817, opened 2024-10-08
[bitnami/contour]: Fix invalid extraVolumeMounts

uubk:
### Description of the change

The default backend's deployment.yaml currently contains the following code:
```yaml
          {{- if .Values.defaultBackend.extraVolumeMounts }}
          volumeMounts: {{- include "common.tplvalues.render" ( dict "value" .Values.contour.extraVolumeMounts "context" $ ) | nindent 12 }}
          {{- end }}
```
Note that the if condition uses the correct variable (`defaultBackend.extraVolumeMounts`) whereas the render call uses the incorrect one (`contour.extraVolumeMounts`). This looks like a simple copy and paste mistake.

### Benefits

The default backends' volumes get rendered correctly.

### Checklist

- [X] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/). This is *not necessary* when the changes only affect README.md files.
- [X] Title of the pull request follows this pattern [bitnami/<name_of_the_chart>] Descriptive title
- [X] All commits signed off and in agreement of [Developer Certificate of Origin (DCO)](https://github.com/bitnami/charts/blob/main/CONTRIBUTING.md#sign-your-work)

PR #29890, opened 2024-10-14
[bitnami/contour]: Fix invalid volume indentation

uubk:
The defaultBackend currently indents volumes to the same level as an individual container (8), however, it actually needs to be one level further up (6).

### Description of the change
Currently, we have 
```
      containers:
        - name: default-backend
          <snip>
        {{- if .Values.defaultBackend.sidecars }}
        {{- include "common.tplvalues.render" ( dict "value" .Values.defaultBackend.sidecars "context" $) | nindent 8 }}
        {{- end }}
        {{- if .Values.defaultBackend.extraVolumes }}
        volumes: {{- include "common.tplvalues.render" ( dict "value" .Values.defaultBackend.extraVolumes "context" $ ) | nindent 8 }}
        {{- end }}
```
... note that volumes is on the same indent level as an individual container. It should actually be one level up. This change does just that. I unfortunately missed this when upstreaming the original fix in bitnami/charts#29817.

### Benefits

Extra volumes for the default backends no longer lead to errors.

### Possible drawbacks

None.

### Applicable issues

None.

### Additional information

None.

### Checklist

### DIFF
--- a/bitnami/contour/CHANGELOG.md
+++ b/bitnami/contour/CHANGELOG.md
@@ -1,8 +1,12 @@
 # Changelog
 
-## 19.2.0 (2024-10-11)
+## 19.2.1 (2024-10-15)
 
-*  [bitnami/contour] fixed wrong envoy-service-name arg value in contour when envoy service name is changed ([#29556](https://github.com/bitnami/charts/pull/29556))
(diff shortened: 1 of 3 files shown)


uubk:
> There are no changes in values or manifests, did you miss to upload them?

My bad ... yes. I amended the first commit to include them now.

PR #29900, opened 2024-10-15
[bitnami/mastodon] feat!: :sparkles: :boom: :recycle: Bump appVersion to 4.3 and refactor init containers

javsalgar:
BREAKING CHANGE

### Description of the change
This PR adds new keys that are present in Mastodon 4.3 and performs a refactor of the init containers and job resources. This is necessary to perform a better management of the resources. The assets precompile container requires large resources but the create admin requires a small one. Before it was not possible to differentiate the resources between them.

### Applicable issues

- fixes #

### Checklist


juan131 (approved):
LGTM

juan131:
Please perform upgrade tests to double-check changes are as less disruptive as possible

applejag:
Damn, I did not read the patch notes carefully enough.

When using `helmfile apply -i` this chart generated new credentials for the encryption key every time. I applied some changes two-three times, and so the initial encryption key was lost and now everything got encrypted with a lost key. (would be nice if the bitnami chart didn't generate a password at all to prevent this from happening)

I tried looking for the first password that the bitnami chart used in the `helm history mastodon` but in my stress to fix this I created too many history items and so the first password is really lost to the ether.

And because I'm a noob and didn't do backups on Postgres then this means all is lost forever. I now have to recreate my entire account with all of my followings on someone else's instance instead of migrating my account...

Welp, it has been interesting I guess. But this is where my multi-year experiment of hosting my own Mastodon ends. It has not really been a fun experience. Fuck damn. Frustrating...

PR #29998, opened 2024-10-19
[bitnami/keycloak] Added providers to the list of writable dirs

casell:
### Description of the change

Added /opt/bitnami/keycloak/providers to the list of dirs copied from base image to empty-dir

### Benefits

It enables extensions/plugins/themes provisioning when readonly filesystem is enabled

### Additional information

The providers directory contains plugins from the base image, so replacing it with a new empty dir is not enough, i.e. metrics plugin will be removed.

### Checklist


migruiz4 (approved):
Thank you for your contribution @casell! Looks good to me!

p4block:
This change despite being shipped in a minor required me to (very lightly) intervene in the values.yaml, as I was creating this empty volume by hand before

image

Happy to see it by default in any case, however, should the same be done for the `themes` folder? It's empty by default so there's no loss of functionality if I do so, but I do need to mount an empty volume to it to copy my themes at startup.

PR #29986, opened 2024-10-18
[bitnami/discourse] include discourse.extraEnvVars in the "install-plugins" init-container

VasquezPonte:
### Description of the change

When installing discourse behind a firewall, the "install-plugins" init-container cannot download the plugins without the proxy-settings defined in “discourse.extraEnvVars”:

```yaml
discourse:
  extraEnvVars:
  - name: http_proxy
    value: http://my.proxy
  - name: https_proxy
    value: http://my.proxy
  - name: no_proxy
    value: '.svc,.svc.cluster.local'
```

### Benefits

The "install-plugins" init-container can download the plugins.

PR #30026, opened 2024-10-21
[bitnami/discourse] include discourse.extraEnvVars in the "install-plugins" init-container

VasquezPonte:
### Description of the change

When installing discourse behind a firewall, the "install-plugins" init-container cannot download the plugins without the proxy-settings defined in “discourse.extraEnvVars”:

```
discourse:
  extraEnvVars:
  - name: http_proxy
    value: http://my.proxy
  - name: https_proxy
    value: http://my.proxy
  - name: no_proxy
    value: '.svc,.svc.cluster.local'
```

### Benefits

The "install-plugins" init-container can download the plugins.

See: https://github.com/bitnami/charts/pull/29986

### DIFF
--- a/bitnami/discourse/CHANGELOG.md
+++ b/bitnami/discourse/CHANGELOG.md
@@ -1,8 +1,13 @@
 # Changelog
 
-## 15.0.3 (2024-10-21)
+## 15.0.4 (2024-10-21)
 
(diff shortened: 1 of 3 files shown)


juan131 (approved):
LGTM! Thanks for the contribution

PR #30062, opened 2024-10-23
set imagePullPolicy of mlflow container to Always when latest tag is …

### LINKED ISSUE #30061: [bitnami/mlflow] imagePullPolicy of mlflow container is not set according to docstring
### Name and Version

bitnami/mlflow:2.0.2

None

```
helm template --values=values.yaml --set image.tag=latest bitnami/mlflow > foo.bar
```

image.tag=latest

`imagePullPolicy` of the mlflow contaienr should be `Always`

### What do you see instead?

`imagePullPolicy` of the mlflow contaienr should be `IfNotPresent`

### Additional information

_No response_

hahahannes:
…used; fix #30061

### Description of the change

According to the docstring the `imagePullPolicy` is supposed to be `Always` when the latest tag is used.
This is now the case with the change.

### Benefits

The latest image will always be used.

### Possible drawbacks

It might be unclear to the user why their `imagePullPolicy` value is not used. Maybe the logic should be changed so that it will be only `Always` if `imagePullPolicy` is empty as well?

### Applicable issues

- fixes #30061 

### Checklist

### DIFF
--- a/bitnami/mlflow/CHANGELOG.md
+++ b/bitnami/mlflow/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## 2.0.3 (2024-10-23)
+## 2.0.3 (2024-10-24)
 
 * set imagePullPolicy of mlflow container to Always when latest tag is … ([#30062](https://github.com/bitnami/charts/pull/30062))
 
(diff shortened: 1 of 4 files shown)


hahahannes:
Hi @dgomezleon, alright! I just noticed it because I did not set it to Always because of the comment.

PR #30079, opened 2024-10-24
[bitnami/rabbitmq] Allow missing memory limit when memoryHighWatermark is absolute

jonasbadstuebner:
### Description of the change

I get an `You enabled configuring memory high watermark using a relative limit.` error when I remove the memory limit and set
```yaml
memoryHighWatermark:
  enabled: true
  type: absolute
  value: 40Gi
```
That's because the chart validation does not check if the type is actually `relative` before looking for a memory limit.

### Benefits

I can set an absolute high watermark without memory limit specified.

### Possible drawbacks

None afaik.

### Checklist

### DIFF
--- a/bitnami/rabbitmq/CHANGELOG.md
+++ b/bitnami/rabbitmq/CHANGELOG.md
@@ -1,8 +1,12 @@
 # Changelog
 
-## 15.0.3 (2024-10-14)
+## 15.0.4 (2024-10-28)
 
-* [bitnami/rabbitmq] Use common password/secrets manager to handle credentials ([#29879](https://github.com/bitnami/charts/pull/29879))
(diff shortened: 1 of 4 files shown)


fmulero (commented):
Thanks @jonasbadstuebner for your contribution.

I've been taking a glance to the solution and I am not pretty sure if it could hide other issues. Have you seen this piece of code? https://github.com/bitnami/charts/blob/a5afdbe173b219f39f03487275228266bb4a4b48/bitnami/rabbitmq/values.yaml#L498

`resources.limits.memory` seems mandatory even if `memoryHighWatermark.type` is `absolute`

jonasbadstuebner:
Thank you for having a look.

That's a good point. I would propose to also remove this config option, if the memory limit is not set. The override is not necessary in that case, because RabbitMQ can use all available memory on the machine (in theory), where it will auto-detect how much that is.

fmulero (approved):
Nice!

Thanks a lot @jonasbadstuebner. LGTM

PR #30106, opened 2024-10-28
[bitnami/postgresql-ha] fix POSTGRES_POSTGRES_PASSWORD env var when generating secrets

### LINKED ISSUE #30105: [bitnami/postgresql-ha] Chart does not work properly without setting pg credentials via values or using existingSecret
### Name and Version

bitnami/postgresql-ha 14.2.34

amd64

1. Install postgresql-ha with the provided values
2. Experience the postgres pod on a crash loop because `POSTGRES_POSTGRES_PASSWORD` is not passed to the pod

```yaml
postgresql:
  replicaCount: 1
  database: keycloak
  username: keycloak
  initdbScripts:
    init.sql: |
      \connect keycloak;
      CREATE SCHEMA IF NOT EXISTS keycloak AUTHORIZATION keycloak;
```

When not setting an existingSecret or passing the password for postgres user in values, I expect that the chart generates a password and sets `POSTGRES_POSTGRES_PASSWORD` on the postgres pod so that it initializes properly.

### What do you see instead?

When not setting an existingSecret or passing the password for postgres user in values, the chart generates a password for the postgres user but does not set `POSTGRES_POSTGRES_PASSWORD` on the postgres pod - so it crash loops.
This forces me to either set a credential for postgres user via values or use existingSecret, even though these secrets are generated properly if no credentials are passed via values.

davidfrickert:
### Description of the change

When using a username that is not "postgres" and no credentials or existingSecret is passed via values, POSTGRES_POSTGRES_PASSWORD env var needs to be passed to postgres pod. 
This PR fixes a condition that should not have been negated, so that when username is not "postgres" and credentials are generated, we pass the env var to the pod.

### Applicable issues

- fixes #30105

### Checklist


javsalgar:
Hi!

Thank you so much for the PR! Could you bump the chart patch version?

davidfrickert:
> Hi!
> 
> Thank you so much for the PR! Could you bump the chart patch version?

Hi, no worries!
Done

migruiz4:
Hi @davidfrickert,

I'm sorry but the logic is correct as it is.

```
{{- if and (not (eq (include "postgresql-ha.postgresqlUsername" .) "postgres")) (or (not (include "postgresql-ha.postgresqlCreateSecret" .)) (include "postgresql-ha.postgresqlPasswordProvided" .)) }}
```

This could be translated as:
```
IF
username != postgresql
AND
(
  existingSecret was provided
  OR
  postgresPassword was provided
)
THEN set POSTGRES_POSTGRES_PASSWORD (which should be available in either the existing secret or the provided password)
```

Here you can see the logic for the helpers [postgresql-ha.postgresqlCreateSecret](https://github.com/bitnami/charts/blob/postgresql-ha/14.3.5/bitnami/postgresql-ha/templates/_helpers.tpl#L322-L338) and [postgresql-ha.postgresqlPasswordProvided](https://github.com/bitnami/charts/blob/postgresql-ha/14.3.5/bitnami/postgresql-ha/templates/_helpers.tpl#L144-L163)

I tried to reproduce your issue using the values you shared at `https://github.com/bitnami/charts/issues/30105` but I wasn't able to reproduce your issue using the latest version of the `bitnami/keycloak` chart:
```
$ cat user.yaml
replicaCount: 1
database: keycloak
username: keycloak
initdbScripts:
  init.sql: |
    \connect keycloak;
    CREATE SCHEMA IF NOT EXISTS keycloak AUTHORIZATION keycloak;
$ helm install postgresql-ha oci://registry-1.docker.io/bitnamicharts/postgresql-ha -f user.yaml
...
$ kubectl get pods
postgresql-ha-pgpool-64d6d495b6-g9x7m   1/1     Running   0          2m43s
postgresql-ha-postgresql-0              1/1     Running   0          2m43s
postgresql-ha-postgresql-1              1/1     Running   0          2m43s
postgresql-ha-postgresql-2              1/1     Running   0          2m43s
```

bitnami/postgresql-ha/templates/postgresql/statefulset.yaml
```diff
@@ -189,7 +189,7 @@ spec:
               value: {{ .Values.persistence.mountPath | quote }}
             - name: PGDATA
               value: {{ printf "%s/%s" .Values.persistence.mountPath "data" | quote }}
-            {{- if and (not (eq (include "postgresql-ha.postgresqlUsername" .) "postgres")) (or (not (include "postgresql-ha.postgresqlCreateSecret" .)) (include "postgresql-ha.postgresqlPasswordProvided" .)) }}
+            {{- if and (not (eq (include "postgresql-ha.postgresqlUsername" .) "postgres")) (or ((include "postgresql-ha.postgresqlCreateSecret" .)) (include "postgresql-ha.postgresqlPasswordProvided" .)) }}
```
migruiz4, line 192: I need to further investigate this issue, but there is currently bug either on this line, where the logic should be simplified as:
```
{{- if and (not (eq (include "postgresql-ha.postgresqlUsername" .) "postgres")) }}
```

In this line:
https://github.com/bitnami/charts/blob/68b3f324b6b96947cf232b9f127677dbbb264c07/bitnami/postgresql-ha/templates/postgresql/secrets.yaml#L19
Where the condition to create the postgres-password is always true

Or in the container logic, where container initialization will fail with the following error if password for `postgres` user is not provided when a different user is used:
```
postgresql-repmgr 16:01:42.52 INFO  ==> Creating repmgr user: repmgr
2024-10-31 16:01:42.544 GMT [233] FATAL:  password authentication failed for user "postgres"
2024-10-31 16:01:42.544 GMT [233] DETAIL:  User "postgres" has no password assigned.
        Connection matched file "/opt/bitnami/postgresql/conf/pg_hba.conf" line 8: "host     all              all       0.0.0.0/0    md5"
psql: error: connection to server at "127.0.0.1", port 5432 failed: FATAL:  password authentication failed for user "postgres"
```

davidfrickert: Yeah maybe this is not the correct fix but it seems something is strange indeed.

PR #30135, opened 2024-10-30
[bitnami/minio] fix secret creation check

c4tz:
### Description of the change

The "Option to use a secret" introduced in `14.7.0` also brought a small bug with it: The [check if the secret should be created](https://github.com/bitnami/charts/blob/minio/14.8.0/bitnami/minio/templates/secrets.yaml#L6) expects a boolean, but ["include" always returns a string ](https://github.com/helm/helm/issues/11231). Because it is not used directly (only in the [minio.createSecret](https://github.com/bitnami/charts/blob/minio/14.8.0/bitnami/minio/templates/_helpers.tpl#L95) template), even setting `auth.useSecret` to `false` will result in the string "false". This string is then interpreted as / casted to a boolean in the corresponding `if`-clause, where it always is translated to `true` because it is not empty. (If you set `auth.useSecret` to an empty string, the secret will not be created)

### Benefits

`auth.useSecret` actually works as expected:
* If set to `true` the secret will be created (if `existingSecret` is empty)
* If set to `false` the secret won't be created

### Possible drawbacks
I don't see any (yet)

### Applicable issues
-

### Additional information
There would also be the option to change all `if (include "minio.createSecret" .)`s so that they check the string's content instead of only for emptiness. But this would create a larger diff, so I chose not to do it that way.

### Checklist

### DIFF
--- a/bitnami/minio/templates/_helpers.tpl
+++ b/bitnami/minio/templates/_helpers.tpl
@@ -93,9 +93,8 @@ Get the root password key.
 Return true if a secret object should be created
 */}}
 {{- define "minio.createSecret" -}}
-{{- if .Values.auth.existingSecret -}}
-{{- else -}}
-    {{- .Values.auth.useSecret -}}
+{{- if and (not .Values.auth.existingSecret) .Values.auth.useSecret -}}
+    {{- true -}}
 {{- end -}}
 {{- end -}}
 
--- a/bitnami/minio/CHANGELOG.md
+++ b/bitnami/minio/CHANGELOG.md
@@ -1,8 +1,13 @@
 # Changelog
 
-## 14.8.1 (2024-10-31)
+## 14.8.2 (2024-10-31)
 
-* [bitnami/minio] Release 14.8.1 ([#30140](https://github.com/bitnami/charts/pull/30140))
+* [bitnami/minio] fix secret creation check ([#30135](https://github.com/bitnami/charts/pull/30135))
+
+## <small>14.8.1 (2024-10-31)</small>
+
+* [bitnami/*] Remove wrong comment about imagePullPolicy (#30107) ([a51f9e4](https://github.com/bitnami/charts/commit/a51f9e4bb0fbf77199512d35de7ac8abe055d026)), closes [#30107](https://github.com/bitnami/charts/issues/30107)
+* [bitnami/minio] Release 14.8.1 (#30140) ([baef6e7](https://github.com/bitnami/charts/commit/baef6e77282c85b2689cc72415669b7db6666e63)), closes [#30140](https://github.com/bitnami/charts/issues/30140)
 
 ## 14.8.0 (2024-10-18)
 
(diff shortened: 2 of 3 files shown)


bitnami/minio/templates/_helpers.tpl
```diff
@@ -93,9 +93,9 @@ Get the root password key.
 Return true if a secret object should be created
 */}}
 {{- define "minio.createSecret" -}}
-{{- if .Values.auth.existingSecret -}}
+{{- if or (.Values.auth.existingSecret) (eq .Values.auth.useSecret false) -}}
 {{- else -}}
-    {{- .Values.auth.useSecret -}}
+    {{- true -}}
 {{- end -}}
 {{- end -}}
```
migruiz4, line 100: I know this was like this prior to your PR, but could you please simplify this as:
```
{{- define "minio.createSecret" -}}
{{- if and (not .Values.auth.existingSecret) .Values.auth.useSecret -}}
    {{- true -}}
{{- end -}}
```

This was a really nice catch! Although the value `.Values.auth.useSecret` is of type boolean, what happened here is what I consider of the weirdest behaviors with helm.

What happens with this helpers is the following:
```
{{- define "my_helper" -}}
{{- if .var }}
{{- true -}} <-- when '{{if include "my_helper" .}}' is called, this is interpreted as 'true'
{{- else -}}
{{- false -}} <-- when '{{if include "my_helper" .}}' is called, this is also interpreted as 'true'
{{- end -}}
```

The actual solution to this is to:
```
{{- define "my_helper" -}}
{{- if .var }}
{{- true -}} <-- when '{{if include "my_helper" .}}' is called, this is interpreted as 'true'
{{- end -}} 
{{- end -}}  <-- empty return would be interpreted as 'false'
```

In this case, using empty `if` and return value based on input value only makes logic more complex, leading to this bug you catched.

migruiz4 (commented):
Nice catch @c4tz!  Could you please take a look at my comment?

bitnami/minio/templates/_helpers.tpl
c4tz: Ah, thank you for providing a more sleek solution, looks even cleaner!

Yes, that's what I tried to say in my PR description. Sorry if that was written in a confusing way. :)

c4tz:
I accidentally "closed" this PR by force-pushing to my repo (wanted to have a clean history without unnecessary clutter). 

Reopening with this comment.

c4tz:
Why is this being closed again? 🤔 It isn't stale at all.

c4tz:
@migruiz4 the publishing of the new version seems to have failed: https://github.com/bitnami/charts/actions/runs/11661918124 :(

Also, the notification for this failure failed (because the Slack channel might be archived).

migruiz4:
Thanks for the heads up! We are currently facing an issue with our CI/CD pipeline, we will rerun the affected workflows once the issues are solved.

c4tz:
Oh, okay! I thought there might be a problem with my commit.

Hope you can figure out a solution for that issue soon and don't have to spend much time on it :)

PR #30236, opened 2024-11-06
[bitnami/keycloak] Add servicemonitor apiVersion value

catadumitru:
### Description of the change

Adds a configurable apiVersion for ServiceMonitor, enabling compatibility with custom monitoring services like Azure’s azmonitoring.coreos.com/v1

### Checklist

### DIFF
--- a/bitnami/keycloak/CHANGELOG.md
+++ b/bitnami/keycloak/CHANGELOG.md
@@ -1,8 +1,12 @@
 # Changelog
 
-## 24.1.0 (2024-11-05)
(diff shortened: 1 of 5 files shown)


fmulero (commented):
Hi @catadumitru 

Unfortunately we can't accept your change because, even it currently works, we can't guarantee the spec would be compatible in the future with other CRDs or different versions and it could lead into several support cases.

As workarounds you could use the `extraDeploy` feature or `helm template` + kustomize to deploy the Azure ServiceMonitor

PR #30300, opened 2024-11-07
[bitnami/mariadb] fix!: :bug: :sparkles: :boom: Set correct serviceName and add Password Update Job

javsalgar:
BREAKING CHANGE

### Description of the change

This PR fixes how the StatefulSets were using as `serviceName` a non-headless service. This was causing that DNS entries were not being registered. In addition to this, we add a new feature, a password update job. This job uses helm hooks to update the mariadb passwords. This is disabled by default.

### Benefits
Correct usage of headless services and better password management

### Possible drawbacks
Manual operation to update the statefulset

### Applicable issues

- fixes #

### Checklist


bitnami/mariadb/templates/primary/statefulset.yaml
```diff
@@ -30,10 +32,14 @@ spec:
         {{- if (include "mariadb.primary.createConfigmap" .) }}
         checksum/configuration: {{ include (print $.Template.BasePath "/primary/configmap.yaml") . | sha256sum }}
         {{- end }}
+        {{- if .Values.passwordUpdateJob.enabled }}
+        charts.bitnami.com/password-last-update: {{ now | date "20060102150405" | quote }}
```
javsalgar, line 36: I'm using charts.bitnami.com as prefix because, when we add this kind of annotations, we get several users asking if this is a standard annotation. This way we explicitly say that this comes from our own definitions, no Kubernetes standard.

bitnami/mariadb/templates/update-password/job.yaml
```diff
@@ -0,0 +1,241 @@
+{{- /*
+Copyright Broadcom, Inc. All Rights Reserved.
+SPDX-License-Identifier: APACHE-2.0
+*/}}
+
+{{- if .Values.passwordUpdateJob.enabled }}
+apiVersion: batch/v1
```
juan131, line 7: Don't we have a `common.capabilities.XXX.apiVersion` for this?

bitnami/mariadb/templates/update-password/job.yaml
```diff
@@ -0,0 +1,241 @@
...
+kind: Job
+metadata:
+  name: {{ printf "%s-password-update" (include "common.names.fullname" .) | trunc 63 | trimSuffix "-" }}
+  namespace: {{ include "common.names.namespace" . | quote }}
+  {{- $versionLabel := dict "app.kubernetes.io/version" ( include "common.images.version" ( dict "imageRoot" .Values.image "chart" .Chart ) ) }}
+  {{- $labels := include "common.tplvalues.merge" ( dict "values" ( list .Values.commonLabels $versionLabel ) "context" . ) }}
+  labels: {{- include "common.labels.standard" ( dict "customLabels" $labels "context" $ ) | nindent 4 }}
+    app.kubernetes.io/part-of: mariadb
+    app.kubernetes.io/component: update-job
```
juan131, line 16: `$versionLabel` is only required for secondary components which version doesn't match appVersion. In this case, give you're using default image, you can simplify:
```suggestion
  labels: {{- include "common.labels.standard" ( dict "customLabels" .Values.commonLabels "context" $ ) | nindent 4 }}
    app.kubernetes.io/part-of: mariadb
    app.kubernetes.io/component: update-job
```

bitnami/mariadb/templates/update-password/job.yaml
```diff
@@ -0,0 +1,241 @@
...
+  {{- $defaultAnnotations := dict "helm.sh/hook" "pre-upgrade" "helm.sh/hook-delete-policy" "hook-succeeded" }}
+  {{- $annotations := include "common.tplvalues.merge" ( dict "values" ( list .Values.commonAnnotations $defaultAnnotations ) "context" . ) }}
+  annotations: {{- include "common.tplvalues.render" ( dict "value" $annotations "context" $ ) | nindent 4 }}
+spec:
+  backoffLimit: {{ .Values.passwordUpdateJob.backoffLimit }}
+  template:
+    metadata:
+      {{- $podLabels := include "common.tplvalues.merge" ( dict "values" ( list .Values.passwordUpdateJob.podLabels .Values.commonLabels $versionLabel ) "context" . ) }}
```
juan131, line 24: Following up previous comment:
```suggestion
      {{- $podLabels := include "common.tplvalues.merge" ( dict "values" ( list .Values.passwordUpdateJob.podLabels .Values.commonLabels ) "context" . ) }}
```

bitnami/mariadb/templates/update-password/job.yaml
```diff
@@ -0,0 +1,241 @@
...
+            - |
+              {{- if .Values.usePasswordFiles }}
+              # We need to load all the secret env vars to the system
+              for file in $(find /bitnami/mariadb/secrets -type f); do
+                  env_var_name="$(basename $file)"
+                  echo "Exporting $env_var_name"
+                  export $env_var_name="$(< $file)"
+              done
+              {{- end }}
+              primary_host={{ include "mariadb.primary.fullname" . }}-0.{{  printf "%s-headless" (include "common.names.fullname" .) | trunc 63 | trimSuffix "-" }}
```
juan131, line 69: Do we support N primaries? Why are we targeting a specific pod?

bitnami/mariadb/templates/update-password/job.yaml
```diff
@@ -0,0 +1,241 @@
...
+              echo "Starting password update job"
+              if [[ -f /job-status/root-password-changed ]]; then
+                echo "Root password already updated. Skipping"
+              else
+                echo "Updating root password"
+                mysql -uroot -p$MARIADB_PREVIOUS_ROOT_PASSWORD -h$primary_host --port={{ .Values.primary.containerPorts.mysql }} -e "SET PASSWORD for 'root'@'%' = PASSWORD('$MARIADB_NEW_ROOT_PASSWORD');"
+                touch /job-status/root-password-changed
+                echo "Root password successfully updated"
+              fi
```
juan131, line 78: Indentation is inconsistent with previous blocks.

bitnami/mariadb/templates/update-password/job.yaml
```diff
@@ -0,0 +1,241 @@
...
+              {{- if not (empty .Values.auth.username) }}
+              if [[ -f /job-status/password-changed ]]; then
+                echo "User password already updated. Skipping"
+              else
+                echo "Updating user password"
+                mysql -u$MARIADB_USER -p$MARIADB_PREVIOUS_PASSWORD -h$primary_host --port={{ .Values.primary.containerPorts.mysql }} -e "SET PASSWORD for '$MARIADB_USER'@'%' = PASSWORD('$MARIADB_NEW_PASSWORD');"
```
juan131, line 84: You can potentially load MariaDB libraries and use wrappers such as `mysql_remote_execute`:
- https://github.com/bitnami/containers/blob/main/bitnami/mariadb/11.5/debian-12/rootfs/opt/bitnami/scripts/libmariadb.sh#L697
And replace `echo` references with `info/debug` ones

bitnami/mariadb/templates/update-password/job.yaml
```diff
@@ -0,0 +1,241 @@
...
+          {{- end }}
+          {{- if .Values.passwordUpdateJob.customLivenessProbe }}
+          livenessProbe: {{- include "common.tplvalues.render" (dict "value" .Values.passwordUpdateJob.customLivenessProbe "context" $) | nindent 12 }}
+          {{- end }}
+          {{- if .Values.passwordUpdateJob.customReadinessProbe }}
+          readinessProbe: {{- include "common.tplvalues.render" (dict "value" .Values.passwordUpdateJob.customReadinessProbe "context" $) | nindent 12 }}
+          {{- end }}
+          {{- if .Values.passwordUpdateJob.customStartupProbe }}
+          startupProbe: {{- include "common.tplvalues.render" (dict "value" .Values.passwordUpdateJob.customStartupProbe "context" $) | nindent 12 }}
+          {{- end }}
```
juan131, line 187: Are these required in a job?

bitnami/mariadb/templates/update-password/job.yaml
```diff
@@ -0,0 +1,241 @@
...
+          volumeMounts:
+            - name: empty-dir
+              mountPath: /tmp
+              subPath: tmp-dir
```
juan131, line 191: Is it required?

bitnami/mariadb/templates/update-password/new-secret.yaml
```diff
@@ -0,0 +1,28 @@
+{{- /*
+Copyright Broadcom, Inc. All Rights Reserved.
+SPDX-License-Identifier: APACHE-2.0
+*/}}
+
+{{- if and .Values.passwordUpdateJob.enabled (eq ( include "mariadb.createSecret" . ) "true") (not ( include "mariadb.createPreviousSecret" . )) (not .Values.passwordUpdateJob.previousPasswords.existingSecret) }}
```
juan131, line 6: Can we replace `eq ( include "mariadb.createSecret" . ) "true"` with `(include "mariadb.createSecret" . )` or does it fail?

bitnami/mariadb/values.yaml
```diff
@@ -575,6 +575,10 @@ primary:
     ##     timeoutSeconds: 300
     ##
     sessionAffinityConfig: {}
+    ## @param primary.service.headless.annotations Annotations of the headless service
```
juan131, line 578: Remove deprecated values:
```console
$ grep -E "Deprecated:|DEPRECATED:" bitnami/mariadb/values.yaml
## @param global.storageClass DEPRECATED: use global.defaultStorageClass instead
```

juan131 (changes requested):
Thanks so much for this awesome PR! Please check my comments

bitnami/mariadb/templates/update-password/job.yaml
javsalgar: We only support 1 primary but still we are using a StatefulSet to have a stable pod domain name. We want to use the DNS entry because it could happen something like this:

- The first operation changes the root password
- The readiness probe fails because the root password has changed
- When trying to update the user password, the service gets removed from the normal, non-headless service, and therefore starts failing with `connection timeout`

The job (and therefore the upgrade) gets stuck

bitnami/mariadb/templates/_helpers.tpl
```diff
@@ -117,10 +117,40 @@ Return the secret with MariaDB credentials
...
+{{/*
+Return the secret with previous MariaDB credentials
+*/}}
+{{- define "mariadb.update-job.previousSecretName" -}}
+    {{- if .Values.passwordUpdateJob.previousPasswords.existingSecret -}}
+        {{- /* The new secret is managed externally */ -}}
+        {{- tpl .Values.passwordUpdateJob.previousPasswords.existingSecret $ -}}
+    {{- else if .Values.passwordUpdateJob.previousPasswords.rootPassword -}}
+        {{- /* The new secret is managed externally */ -}}
+        {{- printf "%s-previous-secret" (include "common.names.fullname" .) | trunc 63 | trimSuffix "-" -}}
```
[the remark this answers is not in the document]
javsalgar: I meant the secret containing the new password, let me rephrase it

bitnami/mariadb/templates/_helpers.tpl
```diff
@@ -117,10 +117,40 @@ Return the secret with MariaDB credentials
...
+    {{- else -}}
+        {{- /* The new secret managed by the helm chart. We use the current secret name as it has the old password */ -}}
+        {{- include "common.names.fullname" . -}}
+    {{- end -}}
+{{- end -}}
+
+{{/*
+Return the secret with previous MariaDB credentials
+*/}}
+{{- define "mariadb.update-job.newSecretName" -}}
```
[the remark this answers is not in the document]
javsalgar: We are doing it in other charts:

domain of the helper (or group of helpers) -> kebab case
name of the helper -> camel case

bitnami/mariadb/templates/update-password/job.yaml
javsalgar: We don't. We have it from cronjobs but not for jobs

bitnami/mariadb/templates/update-password/job.yaml
javsalgar: If they tried to follow DISA/CIS guidelines, they might want to set them. In the end we decided to not add them in jobs because in most cases it did not make sense.

bitnami/mariadb/templates/update-password/new-secret.yaml
javsalgar: As it is a major bump, let me change the createSecret helper because right now it is returning a string with "true" or "false"

javsalgar: Actually not, I do not know why we were using `eq`. Changing it

bitnami/mariadb/templates/update-password/job.yaml
javsalgar: Done

bitnami/mariadb/templates/update-password/job.yaml
javsalgar: Fixed

bitnami/mariadb/templates/_helpers.tpl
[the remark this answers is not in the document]
javsalgar: Example: https://github.com/bitnami/charts/blob/main/bitnami/dremio/templates/_helpers.tpl#L96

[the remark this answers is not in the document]
juan131: I don't like it but feel free to leave it as it is

bitnami/mariadb/templates/update-password/job.yaml
juan131: > - The readiness probe fails because the root password has changed
> - When trying to update the user password, the service gets removed from the normal, non-headless service

Are we sure about that only applies to the "normal" service? I guess you'll have to add the block below on the headless service (which lacks it) in order to guarantee that you can access the primary replica using the stable pod domain name:

```yaml
spec:
  publishNotReadyAddresses: true
```

bitnami/mariadb/templates/update-password/job.yaml
juan131: > In the end we decided to not add them in jobs because in most cases it did not make sense.

I'm confused.. Default value for `passwordUpdateJob.livenessProbe.enabled` and `passwordUpdateJob.readinessProbe.enabled` is `true`.

javsalgar: Leftover from copy-paste, removed

bitnami/mariadb/templates/update-password/job.yaml
javsalgar: Just in case I added it

juan131 (approved):
LGTM

juan131 (approved):

PR #30303, opened 2024-11-07
[bitnami/keycloak] Fix metrics ports and servicemonitor

### LINKED ISSUE #28975: [bitnami/keycloak] Keycloak Version 25.x.x and Metrics Exposure
### Name and Version

bitnami/keycloak-22.x.x

Starting with Keycloak version 25.x.x, a management interface has been introduced that exposes the /metrics endpoint. When the main Keycloak HTTP server is configured with `tls.enabled`, the `HTTP` server on the management interface also operates in TLS mode.

However, there appears to be a potential bug in Keycloak: while the management interface exposes /metrics over HTTPS, the realm-specific metrics endpoints (e.g., realms/{realm}/metrics) are still exposed via the main Keycloak HTTP server, which might run without `TLS`, i.e., on plain `HTTP`.

In the meantime, the ServiceMonitor configuration template only supports the default HTTP scheme, with no option to configure scrapers to collect metrics via an https-based endpoint. This limitation means that metrics from the management interface, which are served over HTTPS, cannot be gathered directly using the existing ServiceMonitor configuration.

servicemonitor.yaml

```yaml
spec:
  {{- if .Values.metrics.serviceMonitor.jobLabel }}
  jobLabel: {{ .Values.metrics.serviceMonitor.jobLabel }}
  {{- end }}
  endpoints:
    {{- $defaultEndpoint := pick .Values.metrics.serviceMonitor "port" "interval" "scrapeTimeout" "relabelings" "metricRelabelings" "honorLabels" "scheme" }}
    {{- $defaultScheme := "http" }}
    {{- $endpoints := ternary (.Values.metrics.serviceMonitor.endpoints) (list (dict "path" .Values.metrics.serviceMonitor.path)) (empty .Values.metrics.serviceMonitor.path) }}
    {{- range $endpoints }}
    {{- $endpoint := merge . $defaultEndpoint }}
    - port: {{ $endpoint.port | quote }}
      scheme: {{ default $defaultScheme $endpoint.scheme | quote }}
      {{- if eq (default $defaultScheme $endpoint.scheme) "https" }}
      tlsConfig:
        insecureSkipVerify: {{ default true $endpoint.insecureSkipVerify }}
      {{- end }}
      path: {{ include "common.tplvalues.render" ( dict "value" $endpoint.path "context" $) }}
      {{- if $endpoint.interval }}
      interval: {{ $endpoint.interval }}
      {{- end }}
      {{- if $endpoint.scrapeTimeout }}
      scrapeTimeout: {{ $endpoint.scrapeTimeout }}
      {{- end }}
      {{- if $endpoint.relabelings }}
      relabelings: {{- include "common.tplvalues.render" ( dict "value" $endpoint.relabelings "context" $) | nindent 6 }}
      {{- end }}
      {{- if $endpoint.metricRelabelings }}
      metricRelabelings: {{- include "common.tplvalues.render" ( dict "value" $endpoint.metricRelabelings "context" $) | nindent 6 }}
      {{- end }}
      {{- if $endpoint.honorLabels }}
      honorLabels: {{ $endpoint.honorLabels }}
      {{- end }}
    {{- end }}
  namespaceSelector:
    matchNames:
      - {{ include "common.names.namespace" . | quote }}
  selector:
    matchLabels: {{- include "common.labels.matchLabels" ( dict "customLabels" .Values.commonLabels "context" $ ) | nindent 6 }}
      {{- if .Values.metrics.serviceMonitor.selector }}
      {{- include "common.tplvalues.render" (dict "value" .Values.metrics.serviceMonitor.selector "context" $) | nindent 6 }}
      {{- end }}
      app.kubernetes.io/component: metrics
{{- end }}
```

values.yaml

```yaml
serviceMonitor:
     enabled: true
    endpoints:
      - path: '{{ include "keycloak.httpPath" . }}metrics'
        scheme: https
        insecureSkipVerify: true
        port: metrics
      - path: '{{ include "keycloak.httpPath" . }}realms/{{ .Values.adminRealm }}/metrics'
        scheme: http
        port: http
    interval: 30s
    scrapeTimeout: ""
    labels: {}
    selector: {}
    relabelings: []
    metricRelabelings: []
    honorLabels: false
    jobLabel: ""
 ```

### PR DESCRIPTION
**Pull request by erwanval**
### Description of the change

Fixes metrics collection by:
* adding the possibility to overwrite the ```port``` for each serviceMonitor's endpoint
* adding ```scheme``` and ```tlsConfig``` support to serviceMonitor
* adding all required ports to ```keycloak-metrics``` service

As explained in #28975, a recent update changed ```/metrics``` to be exposed through the management port (9000). However, the realms' metrics (```/realms/<realm>/metrics```) are still exposed through the http (8080), or https (8443) ports.
This causes 2 issues:
* Metrics service only expose the 9000 port, and the service monitor has both ```/metrics``` and ```/realms/master/metrics``` endpoints configured in default values. The latter fails as it's not exposed on the management port, and http port isn't present in the service (nor serviceMonitor)
* When ```tls.enabled=true``` is defined, ```/metrics``` also fails because the management port is now exposed over HTTPS. Since there is no ```scheme``` defined in the service monitor, it tries over HTTP, which ends up in an error.

### Benefits

ServiceMonitor now works out of the box with default values.yaml, for both endpoints.
Also, serviceMonitor now works when `tls.enabled=true`.

### Possible drawbacks

The metrics service port's name changed. With default values, previously the port named ```http``` was 9000. Now it's the actual http port from the container (8080). Port 9000 has been named ```metrics```, the same as in the container. The first port of the service is still the 9000, and `metrics` has been defined as default in the serviceMonitor to ensure backward compatibility.   
It should be transparent in most cases, but it might be a breaking change for some users who use that service for something else (outside the chart) and expect port 9000 to be named `http`.

### Applicable issues

- fixes #28975

### Additional information

Note that, as of now, realms metrics are exposed on both HTTP and HTTPS when TLS is enabled (HTTP isn't disabled), so the serviceMonitor can collect these through any, as long as the right port and scheme are defined. Example of values below:
```yaml
tls:
  enabled: true
  usePem: true
  existingSecret: keycloak-internal-tls
metrics:
  enabled: true
  serviceMonitor:
    enabled: true
    endpoints:
      - path: '/metrics'
        scheme: https
        tlsConfig:
          insecureSkipVerify: true
      - path: '/realms/master/metrics'
        port: http
      - path: '/realms/<my-realm>/metrics'
        port: https
        scheme: https
        tlsConfig:
          insecureSkipVerify: true
```

### Checklist

### DIFF
--- a/bitnami/keycloak/CHANGELOG.md
+++ b/bitnami/keycloak/CHANGELOG.md
@@ -1,8 +1,12 @@
 # Changelog
 
-## 24.1.0 (2024-11-05)
+## 24.2.0 (2024-11-07)
 
-* [bitnami/keycloak] switches keycloak-metrics service to point to new port ([#30095](https://github.com/bitnami/charts/pull/30095))
+* [bitnami/keycloak] Fix metrics ports and servicemonitor ([#30303](https://github.com/bitnami/charts/pull/30303))
+
+## 24.1.0 (2024-11-06)
+
+* [bitnami/keycloak] switches keycloak-metrics service to point to new port (#30095) ([8ca86ae](https://github.com/bitnami/charts/commit/8ca86ae9ecb2b375735787001188e5c7757d181b)), closes [#30095](https://github.com/bitnami/charts/issues/30095)
 
 ## <small>24.0.5 (2024-11-04)</small>
 
--- a/bitnami/keycloak/CHANGELOG.md
+++ b/bitnami/keycloak/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## 24.2.0 (2024-11-07)
+## 24.2.0 (2024-11-08)
 
 * [bitnami/keycloak] Fix metrics ports and servicemonitor ([#30303](https://github.com/bitnami/charts/pull/30303))
 
--- a/bitnami/keycloak/Chart.yaml
+++ b/bitnami/keycloak/Chart.yaml
@@ -33,4 +33,4 @@ maintainers:
 name: keycloak
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/keycloak
-version: 24.1.0
+version: 24.2.0
--- a/bitnami/keycloak/README.md
+++ b/bitnami/keycloak/README.md
@@ -540,29 +540,33 @@ As an alternative, you can use of the preset configurations for pod affinity, po
 
 ### Metrics parameters
 
-| Name                                       | Description                                                                                                               | Value   |
-| ------------------------------------------ | ------------------------------------------------------------------------------------------------------------------------- | ------- |
-| `metrics.enabled`                          | Enable exposing Keycloak statistics                                                                                       | `false` |
-| `metrics.service.ports.http`               | Metrics service HTTP port                                                                                                 | `9000`  |
-| `metrics.service.annotations`              | Annotations for enabling prometheus to access the metrics endpoints                                                       | `{}`    |
-| `metrics.service.extraPorts`               | Add additional ports to the keycloak metrics service (i.e. admin port 9000)                                               | `[]`    |
-| `metrics.serviceMonitor.enabled`           | Create ServiceMonitor Resource for scraping metrics using PrometheusOperator                                              | `false` |
-| `metrics.serviceMonitor.port`              | Metrics service HTTP port                                                                                                 | `http`  |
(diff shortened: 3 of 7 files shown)


dgomezleon (approved):
LGTM

dgomezleon:
Thanks for contributing and for the detailed explanation.

PR #30368, opened 2024-11-08
[bitnami/keycloak] KEYCLOAK_HOSTNAME present even if KEYCLOAK_PROXY_HEADERS is set

### LINKED ISSUE #30182: [bitnami/keycloak] KEYCLOAK_HOSTNAME present even if KEYCLOAK_PROXY_HEADERS is set
### Name and Version

bitnami/keycloak:24.0.4

None

Into the Helm chart of Keycloak, the configuration variable `KEYCLOAK_HOSTNAME` is always set once the ingress is enabled.

However, as stated into the Keycloak startup log, when this variable is set, the content of the variable `KEYCLOAK_PROXY_HEADERS` is never taken into account.

_No response_

When the helm `proxyHeaders` variable is set, the `KEYCLOAK_HOSTNAME` into the config map should not be set.

### What do you see instead?

The `KEYCLOAK_HOSTNAME` is always set when ingress is enabled.

### Additional information

_No response_

RDemarneffe:
### Description of the change

When the helm proxyHeaders variable is set, the KEYCLOAK_HOSTNAME into the config map should not be set.

### Benefits

Improve chart to respect the keycloak documentation.

### Possible drawbacks

None

### Applicable issues

- fixes #30182

### Checklist


bitnami/keycloak/templates/configmap-env-vars.yaml
```diff
@@ -35,7 +35,7 @@ data:
       {{- include "common.tplvalues.render" (dict "value" $path "context" $) }}
     {{- end }}
   {{- end }}
-  {{- if and .Values.ingress.enabled .Values.ingress.hostname }}
+  {{- if and .Values.ingress.enabled (and .Values.ingress.hostname (not (empty .Values.proxyHeaders))) }}
```
fmulero, line 38: From the description I understand this line should be:

```suggestion
  {{- if and .Values.ingress.enabled .Values.ingress.hostname (empty .Values.proxyHeaders) }}
```

If `proxyHeaders` is set, `KEYCLOAK_HOSTNAME` shouldn't be filled.

fmulero (changes requested):
Thanks a lot @RDemarneffe for your contribution.

I've just left a comment. Maybe I didn't understand the issue but I think the condition you set is almost the opposite. BTW Could you provide an scenario to reproduce the issue and test your changes?

bitnami/keycloak/templates/configmap-env-vars.yaml
RDemarneffe: I have updated the code.

bitnami/keycloak/templates/configmap-env-vars.yaml
```diff
@@ -35,7 +35,7 @@ data:
       {{- include "common.tplvalues.render" (dict "value" $path "context" $) }}
     {{- end }}
   {{- end }}
-  {{- if and .Values.ingress.enabled .Values.ingress.hostname }}
+  {{- if and .Values.ingress.enabled .Values.ingress.hostname (empty .Values.proxyHeaders) }}
   KEYCLOAK_HOSTNAME: |-
     {{ ternary "https://" "http://" ( or .Values.ingress.tls (eq .Values.proxy "edge") (not (empty .Values.proxyHeaders)) ) -}}
```
fmulero, line 40: Here `proxyHeaders` will be always empty, so we can remove it:
```suggestion
    {{ ternary "https://" "http://" ( or .Values.ingress.tls (eq .Values.proxy "edge") ) -}}
```

bitnami/keycloak/templates/configmap-env-vars.yaml
```diff
@@ -35,7 +35,7 @@ data:
       {{- include "common.tplvalues.render" (dict "value" $path "context" $) }}
     {{- end }}
   {{- end }}
-  {{- if and .Values.ingress.enabled .Values.ingress.hostname }}
+  {{- if and .Values.ingress.enabled .Values.ingress.hostname (empty .Values.proxyHeaders) }}
```
fmulero, line 38: Does it apply to KEYCLOAK_HOSTNAME_ADMIN also?  https://github.com/bitnami/charts/blob/6ec5f9131660ec1735896e3a9f8885b6495955ee/bitnami/keycloak/templates/configmap-env-vars.yaml#L24-L25

fmulero (commented):
Thanks a lot @RDemarneffe for your contribution!

I reproduced the issue you mentioned in https://github.com/bitnami/charts/pull/30368 and how these changes fix it. But I have a couple of doubts below, Could you give them a glance?

bitnami/keycloak/templates/configmap-env-vars.yaml
RDemarneffe: Yes, you are right. I removed it.

bitnami/keycloak/templates/configmap-env-vars.yaml
RDemarneffe: I think it's the same but it's not well documented into Keycloak documentation (https://www.keycloak.org/server/reverseproxy).
I updated the code.

fmulero (approved):
LGTM

fmulero:
Thanks a lot @spantaleev for such detailed explanation.

To be perfectly blunt, I was trying to reproduce the issue reported in #30182 with several chart versions (24.0.4, 24.0.3 24.0.2, 24.0.1 and 24.0.0) and today I am not able to see the message mentioned by @RDemarneffe  (It seems that [I could do it](https://github.com/bitnami/charts/pull/30368#pullrequestreview-2457818144) in the past). I am happy to revert this change but I'd like to get more information about the original issue. Please @RDemarneffe could you share more details about #30182?

RDemarneffe:
In my configuration, I use one Keycloak instance that is accessed through different public URL. I use nginx as a reverse proxy.

Into the Keycloak configuration, I must use the `proxyHeaders: xforwarded` to reply with the right public url. But, with the chart before this change, I always have the `hostname` set (I enable the ingress into the `values.yaml`). But, with the `hostname` set for Keycloak, the public url of Keycloak is not dynamically resolved from the X-Forwarded-* headers.

fmulero:
I've just opened https://github.com/bitnami/charts/pull/31227 to revert this change. @RDemarneffe Please share your configuration and how the change will impact it in that PR

RDemarneffe:
> I've just opened #31227 to revert this change. @RDemarneffe Please share your configuration and how the change will impact it in that PR

Hi @fmulero,

I would like to deploy the "Fully dynamic URLs." case of the documentation https://www.keycloak.org/server/hostname#_fully_dynamic_urls but, this configuration requires no `hostname` configuration parameter set.

I use the same Keycloak instance as identity provider for differents products with different dns. Each product use the keycloak instance through a different URL (keycloak url: `https://identity.<product dns>`). To configure Keycloak to support this deployment configuration, I must be able to:
* Configure Ingress with multiple hostnames. The part of my helm chart configuration is the following:
```
## Keycloak ingress parameters
## ref: https://kubernetes.io/docs/user-guide/ingress/
##
ingress:
  ## @param ingress.enabled Enable ingress record generation for Keycloak
  ##
  enabled: true
  ## @param ingress.ingressClassName IngressClass that will be be used to implement the Ingress (Kubernetes 1.18+)
  ## This is supported in Kubernetes 1.18+ and required if you have more than one IngressClass marked as the default for your cluster .
  ## ref: https://kubernetes.io/blog/2020/04/02/improvements-to-the-ingress-api-in-kubernetes-1.18/
  ##
  ingressClassName: "nginx"
  ## @param ingress.hostname Default host for the ingress record (evaluated as template)
  ##
  hostname: identity.<first-product-dns>
  ## @param ingress.hostnameStrict Disables dynamically resolving the hostname from request headers.
  ## Should always be set to true in production, unless your reverse proxy overwrites the Host header.
  ## If enabled, the hostname option needs to be specified.
  ##
  #
  # Let it as false because we use multiple hostname for Keycloak (with different DNS suffixes)
  hostnameStrict: false
  ## @param ingress.annotations [object] Additional annotations for the Ingress resource. To enable certificate autogeneration, place here your cert-manager annotations.
  ## Use this parameter to set the required annotations for cert-manager, see
  ## ref: https://cert-manager.io/docs/usage/ingress/#supported-annotations
  ## e.g:
  ## annotations:
  ##
  annotations:
    # A common issue with Keycloak and nginx is that the proxy buffer may be too
    # small for what Keycloak is trying to send. This will result in a Bad Gateway
    # (502) error. There are many issues around the internet about this. The
    # solution is to increase the buffer size of nginx. This can be done by
    # creating an annotation in the ingress specification
    # See: https://stackoverflow.com/questions/57503590/upstream-sent-too-big-header-while-reading-response-header-from-upstream-in-keyc
    nginx.ingress.kubernetes.io/proxy-buffer-size: "256k"
  ## @param ingress.tls Enable TLS configuration for the host defined at `ingress.hostname` parameter
  ## TLS certificates will be retrieved from a TLS secret with name: `{{- printf "%s-tls" (tpl .Values.ingress.hostname .) }}`
  ## You can:
  ##   - Use the `ingress.secrets` parameter to create this TLS secret
  ##   - Rely on cert-manager to create it by setting the corresponding annotations
  ##   - Rely on Helm to create self-signed certificates by setting `ingress.selfSigned=true`
  ##
  tls: true
  ## @param ingress.extraHosts An array with additional hostname(s) to be covered with the ingress record
  ## e.g:
  ## extraHosts:
  ##   - name: keycloak.local
  ##     path: /
  ##
  extraHosts:
    - name: identity.<second-product-dns>
      path: /
  ## @param ingress.extraTls The tls configuration for additional hostnames to be covered with this ingress record.
  ## see: https://kubernetes.io/docs/concepts/services-networking/ingress/#tls
  ## extraTls:
  ## - hosts:
  ##     - keycloak.local
  ##   secretName: keycloak.local-tls
  ##
  extraTls:
    - hosts:
        - identity<first-product-dns>
      secretName: cluster-wildcard-tls
    - hosts:
        - identity.<second-product-dns>
      secretName: ${replace(dnsName, ".", "-")}-cluster-wildcard-tls
```
* Avoid setting the `KEYCLOAK_HOSTNAME` environment variable.

If the `KEYCLOAK_HOSTNAME` environment variable is set, Keycloak always reply with the `identity.<first-product-dns>` url (this url is also referenced into the tokens emitted) and the second product is not able to validate / trust the tokens.


By reading the documentation, I think we have 2 options:
* Adding a separate configuration parameter for setting or not the `KEYCLOAK_HOSTNAME` environment variable;
* Using the value of ´ingress.hostnameStrict` to prevent setting the `KEYCLOAK_HOSTNAME` environment variable.

I think the second one is the best.

Regards,
Renaud

fmulero:
Please let's follow the discussion in https://github.com/bitnami/charts/pull/31227

PR #30393, opened 2024-11-11
[bitnami/nats] Expose `publishNotReadyAddresses` on headless service

### LINKED ISSUE #30392: [bitnami/nats] Expose `publishNotReadyAddresses` in Headless Service
### Name and Version

bitami/nats 8.4.8

We deploy NATS in AWS EKS and we use a Network Load Balancer for the `nats` service.

We use the AWS Load Balancer Controller with Readiness Gates enabled so we can do controlled rolling updates of the NATS cluster.

The Readiness Gates make it so that K8s will use the Load Balancer's target group health status before shutting down an old pod.

The problem is that, by default, NATS uses the `nats` (load balancer) service for clustering. With the readiness gates and scaling from zero, we see the health checks never pass which prevents the cluster from forming.

Our solution was to change it so NATS uses the `nats-headless` service instead and manually adding `publishNotReadyAddresses` on the headless service so that DNS works and the NATS cluster can form.

Expose the `publishNotReadyAddresses` config in the Headless Service.
Maintain the existing default of `publishNotReadyAddresses: false` so operators have to explicitly set it to `true`

_No response_

rblaine95:
### Description of the change

* Add the `publishNotReadyAddresses` config to NATS Headless Service
* Default to `false` (existing behavior)

### Benefits

Allow operators to specify `service.headless.publishNotReadyAddresses=true` if desired.

### Applicable issues

- fixes #30392 

### Checklist

### DIFF
--- a/bitnami/nats/CHANGELOG.md
+++ b/bitnami/nats/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 8.4.9 (2024-11-11)
 
-* [bitnami/nats] Expose `publishNotReadyAddresses` headless service ([#30393](https://github.com/bitnami/charts/pull/30393))
+* [bitnami/nats] Expose `publishNotReadyAddresses` on headless service ([#30393](https://github.com/bitnami/charts/pull/30393))
 
 ## <small>8.4.8 (2024-11-08)</small>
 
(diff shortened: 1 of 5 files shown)


juan131 (approved):

PR #30397, opened 2024-11-11
[bitnami/kube-prometheus] Add dummy values.yaml to CRDs subchart

carrodher:
Add _values.yaml_ since it is required by CI/CD tools and processes.


bitnami/kube-prometheus/charts/kube-prometheus-crds/values.yaml
```diff
@@ -0,0 +1,8 @@
+# Copyright Broadcom, Inc. All Rights Reserved.
+# SPDX-License-Identifier: APACHE-2.0
+
+## bitnami/common
```
jotadrilo, line 4:
```suggestion
## bitnami/kube-prometheus-crds
```

PR #30403, opened 2024-11-11
[github-actions] Add support for legacy .tgz index branch

migruiz4:
### Description of the change

This PR adds an additional job to the CD pipeline workflow.

This additional step will update the `deprecated-index` branch in preparation for the index.yaml migration to OCI https://github.com/bitnami/charts/pull/30402

**IMPORTANT NOTES:** 
- We will keep `.tgz` releases for 3 months after this change, where the legacy index using `.tgz` will be published into the 'deprecated-index' branch.  
- **After the deprecation period ends** the branch will be permanently removed and new releases of the charts will only be available through the OCI repository.

### DIFF
--- a/.github/workflows/cd-pipeline.yml
+++ b/.github/workflows/cd-pipeline.yml
@@ -165,6 +165,86 @@ jobs:
             echo "Could not update the index after $max_attempts attempts"
             exit 1
           fi
+  update-deprecated-index:
+    runs-on: ubuntu-latest
+    needs:
+      - vib-publish
+    name: Update branch deprecated-index
+    steps:
+      - uses: actions/download-artifact@fa0a91b85d4f404e444e00e005971372dc801d16
+        with:
+          path: ~/artifacts
+      # If we perform a checkout of the main branch, we will find conflicts with the submodules
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683
+        with:
+          ref: 'deprecated-index'
+          path: 'deprecated-index'
+          # The token is persisted in the local git config and enables scripts to run authenticated git commands.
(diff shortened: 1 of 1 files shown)


.github/workflows/cd-pipeline.yml
```diff
@@ -161,6 +161,86 @@ jobs:
...
+        run: |
+          # Configure AWS account
+          export $(printf "AWS_ACCESS_KEY_ID=%s AWS_SECRET_ACCESS_KEY=%s AWS_SESSION_TOKEN=%s" $(aws sts assume-role --role-arn ${AWS_ASSUME_ROLE_ARN} --role-session-name GitHubCharts --query "Credentials.[AccessKeyId,SecretAccessKey,SessionToken]" --output text))
+          # Extract chart release metadata from the publish report file
+          vib_publish_report_file=$(find ~/artifacts -name "report.json" -print -quit)
+          chart_name=$(jq -re '.actions|map(select(.action_id == "helm-publish"))[0] | .application.name' $vib_publish_report_file)
+          chart_version=$(jq -re '.actions|map(select(.action_id == "helm-publish"))[0] | .application.version' $vib_publish_report_file)
+          # Download published asset
+          mkdir download
+          aws s3 cp s3://${{ secrets.AWS_S3_BUCKET }}/bitnami/${chart_name}-${chart_version}.tgz download/
```
carrodher, line 207: We can pull the chart directly from DockerHub, the fetched artifact is the same tgz stored as the one stored in s3. In that way, we reduce the traffic to s3 and it can continue working once s3 is stopped.

With that change, we can also get rid of the AWS account configuration.

```suggestion
          helm pull oci://registry-1.docker.io/bitnamicharts/${chart_name} --version ${chart_version} -d download
```

carrodher (approved):

PR #30512, opened 2024-11-19
[bitnami/mongodb] fix: Modify access modes to match YAML 1.2 schema

eugene-marchanka:
### Description of the change

This change updates the access modes to match the YAML 1.2 schema

### Benefits

This change is preventing manifests from being malformed due to using YAML autoformatters that comply with the YAML 1.2 schema.

### Possible drawbacks

None

### Applicable issues

- fixes #

### Checklist

### DIFF
--- a/bitnami/mongodb/templates/arbiter/statefulset.yaml
+++ b/bitnami/mongodb/templates/arbiter/statefulset.yaml
@@ -287,7 +287,7 @@ spec:
         - name: common-scripts
           configMap:
             name: {{ printf "%s-common-scripts" (include "mongodb.fullname" .) }}
-            defaultMode: 0555
+            defaultMode: 0o555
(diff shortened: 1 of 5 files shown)


dus7eh:
There's a certain level of ambiguity there. You've updated the values to be compliant with YAML 1.2 spec.  
However, I don't see it explicitly stated that Kubernetes supports `0oNNN` notation. Helm (v3.17) acts transparently here and generates configs with the same notation, when applied to kuberetes cluster the installation succeeds although kubernetes docs [here](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.32/#configmapprojection-v1-core) state that:

> defaultMode: ... Must be an octal value between 0000 and 0777 or a decimal value between 0 and 511. ...

Hence, I'm not sure how to treat this: an undocumented kubernetes feature or non-compliance in the charts.

PR #30520, opened 2024-11-19
[bitnami/redis] CA shouldn't be mandatory when TLS is enabled

### LINKED ISSUE #30305: [bitnami/redis] More flexibility needed for handling CA certificate
### Name and Version

bitnami/redis 20.2.1

In order to use TLS, Redis needs 

> a CA certificate bundle file or path to be used as a trusted root when validating certificates [[source](https://redis.io/docs/latest/operate/oss_and_stack/management/security/encryption/#:~:text=In%20order%20to%20support%20TLS%2C%20Redis%20must%20be%20configured%20with%20a%20X.509%20certificate%20and%20a%20private%20key.%20In%20addition%2C%20it%20is%20necessary%20to%20specify%20a%20CA%20certificate%20bundle%20file%20or%20path%20to%20be%20used%20as%20a%20trusted%20root%20when%20validating%20certificates)]

`redis-cli` is a little more flexible in that it will:

> If neither cacert nor cacertdir are specified, the default system-wide trusted root certs configuration will apply. [[source](https://valkey.io/topics/cli/#:~:text=If%20neither%20cacert%20nor%20cacertdir%20are%20specified%2C%20the%20default%20system%2Dwide%20trusted%20root%20certs%20configuration%20will%20apply.) (Disclaimer: these are `valkey` docs, but `redis-cli --help` says the same thing]

As such, `bitnami/redis` also insists upon being given a CA certificate.

This presents a problem when working with `cert-manager` which doesn't always include a `ca.crt` when issuing certificates:

> Additionally, if the Certificate Authority is known, the corresponding CA certificate will be stored in the secret with key ca.crt. For example, with the ACME issuer, the CA is not known and ca.crt will not exist in acme-crt-secret. [[source](https://cert-manager.io/v1.1-docs/concepts/certificate/#:~:text=Additionally%2C%20if%20the%20Certificate%20Authority%20is%20known%2C%20the%20corresponding%20CA%20certificate%20will%20be%20stored%20in%20the%20secret%20with%20key%20ca.crt.%20For%20example%2C%20with%20the%20ACME%20issuer%2C%20the%20CA%20is%20not%20known%20and%20ca.crt%20will%20not%20exist%20in%20acme%2Dcrt%2Dsecret.)]

This leads to trouble with `bitnami/redis` because I'm left with two options:
1. Create a custom secret from the `cert-manager` secret that also includes a manually retrieved copy of [the appropriate root certificate](https://crt.sh/?id=9314791).
2. Hack a way around.

Option 1 doesn't work great for a CA like Let's Encrypt which issues certificates with a TTL of 90 days.
Option 2 is the route I've been forced to go, but the path forward is brittle and hacky.

I've created [`tdg5/bitnami-redis-ca-cert-shenanigans`](https://github.com/tdg5/bitnami-redis-ca-cert-shenanigans) to document the working hack I've found and to demonstrate that it should be easier than this.

I'm happy to help contribute a solution, but I'd need some guidance on the right way to go about simplifying this problem.

I think the best option is to make the source of the CA certificate more configurable (e.g. don't require it to be a part of the TLS certificate). This would allow injecting custom root certificates and/or using trusted root certs already available in the docker image.

I don't love that this still requires mounting extra volumes to the pods (if the system doesn't have an acceptable root cert), but at least its a more explicitly supported path. At a minimum, it seems like forcing the CA certificate path to start with `/opt/bitnami/redis/certs` severely limits options and should be made more flexible.

Another less good option (IMHO) would be to make `cacert` an optional argument for uses of `redis-cli` since `redis-cli` will fallback to using the `default system-wide trusted root certs`. I've found that redis server will work if I tell it the CA is `tls.crt`, but `redis-cli` will not cooperate. This could allow for making the CA cert more generally optional. I worry though that this is not very explicit and maybe only solves a subset of awkwardness.

Maybe I'm missing something, but it seems that the only alternative would be to craft a custom secret from the `cert-manager` certificate secret that also includes the CA, but that makes automation very difficult so it's not an option I'm willing to accept.

juan131:
### Description of the change

This PR makes it optional to have include a CA (`ca.crt`) on TLS secrets when TLS is enabled and existing secrets are used.

### Benefits

As explained in the linked issue, the CA isn't mandatory and the chart should be flexible enough to not require it.

### Possible drawbacks

None

### Applicable issues

- fixes #30305

### Additional information

N/A

### Checklist


dgomezleon (dismissed):
LGTM

dgomezleon (approved):

PR #30523, opened 2024-11-19
[bitnami/flux] Fix incorrect argument mapping and add missing events-addr argument

erfan-272758:
## Description
This pull request addresses two key issues in the current Bitnami Helm chart for FluxCD:

### 1. Fix Argument Mapping for Notification Controller
- The Notification Controller uses two primary arguments to manage communication:
  - `events-addr`: The address where the controller listens for events from other controllers (e.g., `helm-controller`, `kustomize-controller`). Events received on this address trigger Alert objects.
  - `receiverAddr`: The address where the controller listens for webhooks from external sources (e.g., Git repositories). Requests sent to this address trigger Receiver objects.
- **Problem: Incorrect Mapping in Current Chart**
  - The `receiver` key maps to `receiverAddr`, which is correct.
  - However, the `webhook` key is incorrectly mapped to `events-addr`.
- **Impact of the Incorrect Mapping:**
  - Users configuring the chart with the `webhook` key would expect it to configure the webhook listener (`receiverAddr`), but instead, it configures the event listener (`events-addr`).
  - Similarly, the `receiver` key, which users would expect to configure the event listener (`events-addr`), instead configures the webhook listener (`receiverAddr`).
- **Solution (Applied in This PR):**
  - `webhook` key → `receiverAddr` (Webhook listener).
  - `receiver` key → `events-addr` (Event listener).
- This fix aligns the key mappings with their intended functionality, making the chart configuration more intuitive and accurate.

### 2. Add Missing `events-addr` Argument for Other Controllers
- The `events-addr` argument was not initialized for the following controllers:
  - `helm-controller`
  - `kustomize-controller`
  - `source-controller`
  - `image-automation-controller`
  - `image-reflector-controller`
- **Solution:**
  - This PR adds the `events-addr` argument for these controllers, with its value set to the address of the Notification Controller's service.

## Impact
- Fixes the misconfigured argument mapping for the Notification Controller, ensuring it works as intended.
- Ensures all controllers properly communicate with the Notification Controller by initializing the `events-addr` argument.

## Testing
- Verified that the Notification Controller correctly handles events via `events-addr` and webhooks via `receiverAddr`.
- Confirmed other controllers now correctly use the `events-addr` argument to communicate with the Notification Controller.

## Checklist
- [x] Updated argument mappings for `webhook` and `receiver` keys in the Notification Controller.
- [x] Added the `events-addr` argument to the relevant controllers.
- [x] Verified changes function correctly across all controllers.

## Notes
These changes ensure that all FluxCD controllers are correctly configured for communication and align with the expected behavior of the Notification Controller.

### DIFF
--- a/bitnami/flux/CHANGELOG.md
+++ b/bitnami/flux/CHANGELOG.md
@@ -1,8 +1,8 @@
 # Changelog
 
-## 2.3.22 (2024-11-19)
+## 2.3.22 (2024-11-29)
 
-* Fix Incorrect Argument Mapping and Add Missing events-addr Argument in Bitnami FluxCD Chart ([#30523](https://github.com/bitnami/charts/pull/30523))
+* [bitnami/flux] Fix incorrect argument mapping and add missing events-addr argument ([#30523](https://github.com/bitnami/charts/pull/30523))
 
 ## <small>2.3.21 (2024-11-07)</small>
 
--- a/bitnami/flux/CHANGELOG.md
+++ b/bitnami/flux/CHANGELOG.md
@@ -1,8 +1,13 @@
 # Changelog
 
-## 2.3.21 (2024-11-07)
+## 2.3.22 (2024-11-19)
 
-* [bitnami/flux] Release 2.3.21 ([#30263](https://github.com/bitnami/charts/pull/30263))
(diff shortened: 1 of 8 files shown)


erfan-272758:
Is there anyone?

PR #30531, opened 2024-11-19
[bitnami/mongodb] feat: allow setting `targetLabel` and `podTargetLabels` array on ServiceMonitor

umarhussain15:
### Description of the change

This update to MongoDB chart will allow users to set targetLabels and podTargetLabels arrays on ServiceMonitor resource in the chart. By default, the arrays are set empty and there is a check in the ServiceMonitor file to only put these array if they are set and array is not empty.

Spec ref: https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/api.md#servicemonitorspec

### Benefits

Allows users to further enrich Prometheus metrics with new labels when they are scraped by the server.
### Possible drawbacks
None

### Applicable issues

None
### Checklist


bitnami/mongodb/values.schema.json
```diff
@@ -1,232 +1,3525 @@
 {
```
migruiz4, line 1: Could you please keep the `values.schema.json` as it is?

migruiz4 (commented):
Thank you for your contribution @umarhussain15!

Changes look good to me, but I can not accept this PR until the changes to the `values.schema.json` have been reverted.

We have had issues in the past where having an schema that fully covers the values.yaml is 'too strict' and ends up causing severe issues with many values, especially those that are rendered as template or allow different value types (string/array/object).

bitnami/mongodb/values.schema.json
umarhussain15: I have reverted the changes in the schema

Issue #30597, opened 2024-11-22
Provisioning job failing due to sleep timeout not being sufficient enough for minio cluster to be operational upon "mc admin service restart" command in provisioning job


choudham:
### Name and Version

14.8.5

### What architecture are you using?

amd64

### What steps will reproduce the bug?

Deploy Minio Helm chart version 14.8.5

Provisioning job fails since the data drives are not up

![image](https://github.com/user-attachments/assets/db639def-94ad-472f-a482-89d4ccd61368)

Screenshot 2024-11-22 at 10.41.00 AM.png
Based on the above upstream change to address the race condition in helm chart we have 5 sec sleep but based on the observation our environment it took more time forit to stabilize (approximately 3 mins … see prometheus graph)

### Are you using any custom parameters or values?
```yaml
USER-SUPPLIED VALUES:
auth:
  rootPassword: XXXX
  rootUser: XXXX
defaultBuckets: ""
global:
  defaultStorageClass: default
image:
  registry: docker.io
  repository: bitnami/minio
  tag: 2024.8.3-debian-12-r1
metrics:
  enabled: true
  prometheusRule:
    enabled: true
  serviceMonitor:
    enabled: true
    labels:
      prometheus: default
      release: platform-monitoring
mode: distributed
persistence:
  size: 300Gi
podAnnotations:
  prometheus.io/path: /minio/v2/metrics
  prometheus.io/port: "9000"
  prometheus.io/scrape: "true"
provisioning:
  buckets:
  - anonymous: true
    lifecycle:
    - disabled: false
      expiry:
        days: 30
        noncurrentVersionDays: 0
      id: 30dRetention
    name: test-A
    versioning: false
    withLock: false
  - lifecycle:
    - disabled: false
      expiry:
        days: 30
        noncurrentVersionDays: 0
      id: 30dRetention
    name: test-B
    versioning: false
    withLock: false
  - lifecycle:
    - disabled: false
      expiry:
        days: 30
        noncurrentVersionDays: 0
      id: 30dRetention
    name: test-C
    versioning: false
    withLock: false
  - lifecycle:
    - disabled: false
      expiry:
        days: 30
        noncurrentVersionDays: 0
      id: 30dRetention
    name: tunnel
    quota:
      size: 30GiB
      type: set
    versioning: false
    withLock: false
  - lifecycle:
    - disabled: false
      expiry:
        days: 30
        noncurrentVersionDays: 0
      id: 30dRetention
    name: test123
    versioning: false
[...]

PR #30718, opened 2024-12-02
[bitnami/minio]: Templating sleep sleep time so that it can be passed using values.yaml

choudham:
### Description of the change

This templates the sleep time for minio-provisioning job keeping the default at 5 sec.

### Benefits

Easy to change using values.yaml

### Applicable issues

(https://github.com/bitnami/charts/issues/30597)

### Checklist
Issue opened for this
(https://github.com/bitnami/charts/blob/main/CONTRIBUTING.md#sign-your-work)


patila-hpe (approved):

dgomezleon (commented):
Thanks for contributing.

I would suggest adding the new parameter to `values.yaml`. Otherwise, the README.md will be overwrite in a future PR and the note will be lost. Also, since it adds a new parameter, use version 14.9.0 instead.

bitnami/minio/values.yaml
```diff
@@ -235,6 +235,7 @@ provisioning:
   ## @param provisioning.enabled Enable MinIO&reg; provisioning Job
   ##
   enabled: false
+  sleepTime: 5
```
dgomezleon, line 238: You need to add a comment before the parameter. Otherwise, the README automation fails. For example.
```suggestion
  ## @param provisioning.sleepTime Sleep time before checking Minio availability
  ##
  sleepTime: 5
```

bitnami/minio/templates/provisioning-job.yaml
```diff
@@ -143,10 +143,10 @@ spec:
 
               mc admin service restart {{ $minioAlias }} --wait --json;
 
-              # Adding a sleep to ensure that the check below does not cause
+              # Adding a sleep and templating it to ensure that the check below does not cause
```
dgomezleon, line 146:
```suggestion
              # Adding a sleep to ensure that the check below does not cause
```

dgomezleon (changes requested):
Minor suggestions required to fix README.md automation.

bitnami/minio/values.yaml
choudham: @dgomezleon ACK and updated

bitnami/minio/Chart.yaml
```diff
@@ -34,4 +34,6 @@ maintainers:
 name: minio
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/minio
-version: 14.8.6
+version: 14.9.0
+
+
```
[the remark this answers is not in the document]
dgomezleon: I applied the other suggestion. Please bump the version again, this time to 14.10.0, and I will merge the change.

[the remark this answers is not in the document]
choudham: @dgomezleon : Updated the version. Please verify and help us with the merge

choudham:
@dgomezleon : All looks good in terms of the check. Please review and help us merge this :)

dgomezleon (approved):

Issue #30680, opened 2024-11-29
Postgresql metrics container CreateContainerConfigError when replication mode.


song2park:
### Name and Version

bitnami/postgresql 16.2.3

### What architecture are you using?

None

### What steps will reproduce the bug?

1. enable `metrics` and make it `replication` mode.
```yaml
image:
  registry: docker.io
  repository: bitnami/postgresql
  tag: 16.3.0-debian-12-r23

architecture: replication

auth:
  postgresPassword: "test"
  rootPassword: "test"
  database: "testdb"
  replicationPassword: "testrepl"

metrics:
  enabled: true
```

2. install the chart on kubernetes
```
helm upgrade --install post-test bitnami/postgresql -f .\values.yaml
```

### What is the expected behavior?

Metrics sidecar of read replica is successfully launched like that of primary pod.

### What do you see instead?

![image](https://github.com/user-attachments/assets/817aeb32-d937-4e87-9e2e-dcc2199cfcbc)

I got a `CreateContainerConfigError` error.

The error description is `Error: couldn't find key password in Secret default/post-test-postgresql`.

It turns out it uses `password` not `postgres-password`.

### statefulset.yaml in primary
```yaml
# ...
          env:
            - name: DATA_SOURCE_URI
              value: {{ printf "127.0.0.1:%d/postgres?sslmode=disable" (int (include "postgresql.v1.service.port" .)) }}
            {{- if .Values.auth.usePasswordFiles }}
            - name: DATA_SOURCE_PASS_FILE
              value: {{ printf "/opt/bitnami/postgresql/secrets/%s" (include "postgresql.v1.adminPasswordKey" .) }}
            {{- else }}
            - name: DATA_SOURCE_PASS
              valueFrom:
                secretKeyRef:
                  name: {{ include "postgresql.v1.secretName" . }}
                  key: {{ include "postgresql.v1.adminPasswordKey" . }}
            {{- end }}
            - name: DATA_SOURCE_USER
              value: "postgres"
            {{- if .Values.metrics.extraEnvVars }}
            {{- include "common.tplvalues.render" (dict "value" .Values.metrics.extraEnvVars "context" $) | nindent 12 }}
            {{- end }}
# ...
```

### statefulset.yaml in read
```yaml
# ...
          env:
            - name: DATA_SOURCE_URI
              value: {{ printf "127.0.0.1:%d/postgres?sslmode=disable" (int (include "postgresql.v1.service.port" .)) }}
[...]

PR #30720, opened 2024-12-03
[bitnami/postgresql] use adminPassword for metrics user when custom user is not set

song2park:
### Description of the change

When `auth.username` is not set and `metrics.enabled: true`, it still looks up `auth.password` for metrics user, leading to `CreateContainerConfigError` on replication mode.

As the default value of `DATA_SOURCE_USER` is `postgres`, its value must be `auth.postgresPassword` if `auth.username` is not set.

### Benefits

Resolve error when creating replica postgresql pod with metrics enabled.

### Applicable issues

#30680

### DIFF
--- a/bitnami/postgresql/CHANGELOG.md
+++ b/bitnami/postgresql/CHANGELOG.md
@@ -1,8 +1,12 @@
 # Changelog
 
-## 16.2.3 (2024-11-27)
+## 16.2.4 (2024-12-03)
 
-* [bitnami/postgresql] Release 16.2.3 ([#30645](https://github.com/bitnami/charts/pull/30645))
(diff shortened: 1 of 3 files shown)


javsalgar (approved):

gaelreyrol:
Hi @song2park, I am sorry to ask that here but, why has it not been applied the primary statefulset too ?

song2park:
> Hi @song2park, I am sorry to ask that here but, why has it not been applied the primary statefulset too ?

Hi @gaelreyrol 

To be honest, I don't understand why primary statefulset uses admin only.
I was hurry at the time, so I didn't wanna change much because I don't know all the history.
IMAO, `metrics` block shud be same as the one in `secondary` as `customUser` on each pod has same permissions.

Are you guna change it?

Issue #30681, opened 2024-11-29
[bitnami/keycloak] "Error: secret "keycloak-externaldb" not found" when trying to install helm chart using external database


federicochiesa:
### Name and Version

bitnami/keycloak 24.2.2

### What architecture are you using?

amd64

### What steps will reproduce the bug?

1. Install with the provided values file
2. `kubectl describe pod keycloak-0 -n sso`
3. I get the error in the title

### Are you using any custom parameters or values?

```yaml
nameOverride: "keycloak"
namespaceOverride: "sso"
image:
  registry: docker.io
  repository: bitnami/keycloak
  tag: 26.0.6-debian-12-r0

adminRealm: "master"

production: true

proxy: "edge"

replicaCount: 2

resourcesPreset: "medium"

livenessProbe:
  enabled: true
  initialDelaySeconds: 300
  periodSeconds: 1
  timeoutSeconds: 5
  failureThreshold: 1
  successThreshold: 1

readinessProbe:
  enabled: true
  initialDelaySeconds: 30
  periodSeconds: 10
  timeoutSeconds: 1
  failureThreshold: 1
  successThreshold: 1

autoscaling:
  enabled: true
  minReplicas: 1
  maxReplicas: 3
  
  behavior:
    scaleUp:
      stabilizationWindowSeconds: 120
      selectPolicy: Max
      policies: []
    scaleDown:
      stabilizationWindowSeconds: 300
      selectPolicy: Max
      policies:
        - type: Pods
          value: 1
          periodSeconds: 300

postgresql:
  enabled: false

externalDatabase:
  host: "xxxxxxxxxxxxx"
  port: 5432
  user: keycloak_kubernetes
  database: keycloak_kubernetes
  password: "xxxxxxxxxxx"

logging:
  output: default
  level: WARN

```

### What is the expected behavior?

It should connect to the DB using the provided values, which are the same as indicated in the readme.

### What do you see instead?
```
Events:
  Type     Reason     Age                   From               Message
  ----     ------     ----                  ----               -------
  Normal   Scheduled  29m                   default-scheduler  Successfully assigned sso/keycloak-0 to xxxxxxxxxx
  Normal   Pulling    29m                   kubelet            Pulling image "docker.io/bitnami/keycloak:26.0.6-debian-12-r0"
[...]

PR #30725, opened 2024-12-03
[bitnami/keycloak] fix external-db-secret not using the specified namespace

federicochiesa:
As per issue #30681 I'm submitting this fix.


javsalgar:
Hi!

Thank you for the PR! Could you bump the patch `version` in Chart.yaml?

PR #30802, opened 2024-12-05
fix: azure secrets

ion-elgreco:
Currently when you set useCredentialsInSecret it will require all secrets to be set, but this doesn't make sense since you only want 1 out of the 3 (connection_String, access_key or client_secret) to be set. This PR adds a conditional check if they key is provided in the values.yaml otherwise the env secret is not defined.


jotamartos (approved):
Changes LGTM! Thank you for your contribution.

PR #30807, opened 2024-12-05
[bitnami/kibana] Template hostname everywhere

jplitza:
### Description of the change

Previously, if `.Values.extraConfiguration."server.publicBaseUrl"` was specified (with a dot in the key), that key was inserted into the configmap twice, causing Kibana to error out.

Furthermore, since the introduction of the fallback to the ingress hostname, this setting couldn't be templated anymore (which was possible before).

This commit allows the templating as well as specifying it with the dot notation.

### Benefits

Upgrades from old chart versions that specified the dotted extraConfiguration won't break the deployment.

Also, templating the publicBaseUrl is possible again.

### Possible drawbacks

None.

### Applicable issues

- fixes #

### Checklist


jplitza:
Wow, auto-closing due to staleness over the holidays... thanks.

@carrodher @gongomgra Do you have the permissions to "reopen" this PR, as suggested? Because I don't.

bitnami/kibana/templates/configmap.yaml
```diff
@@ -22,8 +22,8 @@ data:
     {{- if .Values.configuration.server.basePath }}
     server.basePath: {{ .Values.configuration.server.basePath | quote }}
     {{- end }}
-    {{- if or .Values.configuration.server.publicBaseUrl .Values.ingress.enabled }}
-    server.publicBaseUrl: {{ default .Values.ingress.hostname .Values.configuration.server.publicBaseUrl }}
+    {{- if or .Values.configuration.server.publicBaseUrl ( get .Values.configuration "server.publicBaseUrl" ) .Values.ingress.enabled }}
```
gongomgra, line 25: Please use the `coalesce` method here

bitnami/kibana/templates/configmap.yaml
```diff
@@ -22,8 +22,8 @@ data:
...
+    server.publicBaseUrl: {{ include "common.tplvalues.render" ( dict "value" ( or .Values.configuration.server.publicBaseUrl ( get .Values.configuration "server.publicBaseUrl" ) .Values.ingress.hostname ) "context" $ ) }}
```
gongomgra, line 26: Not sure if we should accept `"server.publicBaseUrl"` as a valid value. We would have to accept it (or even add it) for all other parameters in our charts. Can you provide us with more details on why do you need that format to be available? My understanding is that the format of the current `values.yaml` file is simple enough.

```
## @param configuration [object] Kibana configuration
##
configuration:
  server:
    basePath: ""
    host: "0.0.0.0"
    publicBaseUrl: ""
    rewriteBasePath: false
```

gongomgra (changes requested):
Thanks for your contribution! I have added a couple of comments to your changes.

carrodher:
Thank you for initiating this pull request. We appreciate your effort. This is just a friendly reminder that signing your commits is important. Your signature certifies that you either authored the patch or have the necessary rights to contribute to the changes. You can find detailed information on how to do this in the [“Sign your work” section of our contributing guidelines](https://github.com/bitnami/charts/blob/master/CONTRIBUTING.md#sign-your-work).

Feel free to reach out if you have any questions or need assistance with the signing process.

bitnami/kibana/templates/configmap.yaml
gongomgra: I've been checking this with a teammate and he has some concerns on these changes. He wonders if there would be a value collision in case `server.publicUrl` is also set in extraConfig. @javsalgar can you provide further details?

gongomgra: @jplitza can you check my comment when you have a chance?

jplitza: @gongomgra I didn't understand the comment last week. But while writing this comment, I think I understood the problem: While we use the correct Helm value here if it's specified in `extraConfiguration`,  it will still be pasted twice into the `kibana.yml`. I will remove the coalesce here.

IMO, when you add new keys to the `kibana.yml` that someone could have specified in `extraConfiguration` before (like `server.publicBaseUrl` here), you should increase the major version of this chart. I tried to make that change backwards compatible with this PR, but that's simply not possible.

gongomgra:
@jplitza it looks like your changes are failing for the `helm lint` action, while it doesn't fail in `main` branch. Can you check it?

```
$ gh pr checkout https://github.com/bitnami/charts/pull/30807
$ helm lint .
==> Linting .
[ERROR] templates/: parse error at (kibana/templates/configmap.yaml:23): unexpected right paren

Error: 1 chart(s) linted, 1 chart(s) failed
```

jplitza:
Sorry, there indeed was a spurious `)`. Removed it.

gongomgra (approved):
Thanks for your contribution!

Issue #30850, opened 2024-12-10
New container images security assessments


carrodher:
On December 10th, 2024, we released a minor version for every Helm chart in the Bitnami catalog. This major version introduced a new security mechanism to detect the usage of non-standard containers while installing or upgrading Helm chart releases.

Replacing the original containers that are shipped with the charts will result in installation errors like the one below:
```
⚠ ERROR: Original containers have been substituted for unrecognized ones. Deploying this chart with non-standard containers is likely to cause degraded security and performance, broken chart features, and missing environment variables.

Unrecognized images:

- [REGISTRY]/[REPOSITORY]:[TAG]
(...)
```

### Why did Bitnami introduce this mechanism?

There are several reasons to introduce this change:

1. Bitnami charts are designed, tested, and validated on multiple platforms using a specific set of Bitnami and Tanzu Application Catalog containers. Therefore, Bitnami cannot guarantee the same quality standards levels when the original container images are replaced.
2. Non-standard container images are likely to cause degraded security and performance, broken chart features, and missing environment variables.
3. Incapacity to provide efficient support for non-standard container images given the Bitnami support team's limited bandwidth.

### Can I use different image tags?

Yes, replacing default image tags is considered a valid use case. For instance, a user might be interested in using the Bitnami PostgreSQL chart using Bitnami PostgreSQL _15.x_ container images (note the chart uses _16.x_ images by default).

That said, it is expected that you see some warnings on the chart notes given that the specific chart & image combination wasn't tested and validated by Bitnami on any platform.

### Can I skip the security verification?

Yes, we introduced a new global parameter on every chart that can be used for that purpose, see:

|Name|Description|Value|
| -------- | ------- | ------- |
|_global.security.allowInsecureImages_|By default, this chart verifies that the original container images that were designed, tested, and validated are used. This option makes the chart skip the verification step and proceed.|_false_|

At your own risk, you bypass the verification specifying the parameter adding the `--set global.security.allowInsecureImages=true` argument to `helm install` or `helm upgrade` commands or, alternatively, adding the block below to your values YAML file:
```yaml
global: 
  security: 
    allowInsecureImages: true
```

### Should I expect any issues upgrading my existing chart releases?

No issue is expected if you upgrade your existing releases using original container images. If you replace them, you'll have to skip the security verification.

### Should I expect errors consuming Bitnami charts from mirror registries?

[...]

PR #30863, opened 2024-12-10
[bitnami/appsmith] Detect non-standard images

carrodher:
You can find more information about this change at https://github.com/bitnami/charts/issues/30850.

Implemented thanks to [those changes](https://github.com/bitnami/charts/pull/30851) in btinami/common.


bitnami/appsmith/values.yaml
```diff
@@ -1133,6 +1133,11 @@ backend:
     ##   GKE, AWS & OpenStack)
     ##
     storageClass: ""
+  ## Security parameters
+  ##
+  security:
+    ## @param global.security.allowInsecureImages Allows skipping image verification
+    allowInsecureImages: false
```
jotamartos, line 1140: This is not the global section

PR #31079, opened 2024-12-17
[bitnami/harbor] made generating existing/default secret secrets an option

### LINKED ISSUE #25399: [harbor] Allow providing all secrets out-of-band

cthtrifork:
### Description of the change

This adds an option to disable generation of templated secrets. Default is set to true to keep current behaviour intact.

### Benefits

We have several scanners, such as `polaris` which is picking up that secrets are created by the chart. We are using external-secrets and want to create all secrets out of band.

### Applicable issues

- fixes #25399

### Checklist


cthtrifork:
This fix is still relevant. Reopening

javsalgar:
Hi!

Thank you for the PR! Could you fix the merge issues?

migruiz4:
I'm sorry for the late response @cthtrifork.

Although the changes in this PR are reasonable, after checking more in-depth how the env variable secret is handled in our harbor chart I think it would be better to refactor them rather than an additional feature that workarounds the issue that we may need to remove in the future.

Initially, I was thinking why harbor didn't handle this as other charts: if `existingSecret` is provided, use it and do not generate the default one.

Then, I saw that the default secret is supposed to merge with the `existingSecret` when provided, in which case it would make sense to implement a value that allowed skipping that generation.

But then I saw that, when `existingSecret` / `existingEnvVarsSecret` is provided, the default secret is omitted... See the following:
- Core: https://github.com/bitnami/charts/blob/54531ce059d1156a615749485e11e2f30143ea16/bitnami/harbor/templates/core/core-dpl.yaml#L172-L177
- Registry: https://github.com/bitnami/charts/blob/54531ce059d1156a615749485e11e2f30143ea16/bitnami/harbor/templates/registry/registry-dpl.yaml#L180-L185
- Trivy: https://github.com/bitnami/charts/blob/54531ce059d1156a615749485e11e2f30143ea16/bitnami/harbor/templates/trivy/trivy-sts.yaml#L149-L154
- Job Service: https://github.com/bitnami/charts/blob/54531ce059d1156a615749485e11e2f30143ea16/bitnami/harbor/templates/jobservice/jobservice-dpl.yaml#L186-L191

I think that detail was missed when the secret merge feature was added at https://github.com/bitnami/charts/pull/28989, which additionally removed the condition to not create the secret when existing secret is provided was removed.

In this situation, this leaves us with 2 options:

- Revert this commit #28989, so the secrets are not created if `existingSecret` is provided, and because the merge feature is not being used anyway.
- Honor the merge feature contribution, but add a feature called `mergeSecretEnvVars` (default `false`) that: 
  - When `existingSecret=something && mergeSecretEnvVars=false` skips secret generation and uses `existingSecret`.
  - When `existingSecret=something && mergeSecretEnvVars=true` uses the merged secret.

I will double check this with the team just in case, but wanted to share this with you.

migruiz4:
After discussing this with the team, we decided to revert the merge feature (https://github.com/bitnami/charts/pull/28989).

The feature hasn't really been working for almost a year and not a single user have complained, so we think the way here is to simplify and revert it.

By reverting the change, this PR should not be necessary since secrets won't be created if `existingSecret` / `existingEnvVarsSecret` is provided.

cthtrifork:
Great work looking into this!

PR #31159, opened 2024-12-25
[bitnami/harbor] Upgrade postgres to version 17.2.0

lusu007:
### Description of the change

The commit 017c8c368e4740be7e4d465cbb9b0c54b2ca829a should update the postgres version to version 17.x. But it did not. The patch release after this commit upgrades Postgres to version 14 and not 17. This is also referenced in the README.md of the chart. To solve this issue, this commit really updates Postgres to version 17.2.

### Benefits

Use newest Postgres version.

### Possible drawbacks

Manual upgrade of Postgres is needed (again).

### Applicable issues

None

### Additional information

I have Harbor with PostgreSQL version 17.2 in production without any issues.

### Checklist


carrodher:
According to the "Harbor Components" section from https://goharbor.io/docs/2.12.0/install-config/, the default/recommended PostgreSQL major version is 14, that's the reason why we're overriding the container image of the PostgreSQL subchart.
By default, the latest PostgreSQL subchart bundles PostgreSQL 17, but we need to override its version in the `values.yaml` to match the version recommended by the upstream maintainers.

lusu007:
Thank you for your response! I appreciate you taking the time to provide clarification. I want to point out a few additional considerations regarding the PostgreSQL version and its relationship with Harbor.

The documentation you referenced mentions PostgreSQL 14 as the default version. However, there is no explicit recommendation that PostgreSQL 14 is the only supported version or the version users should always use. It appears to be more of a default choice that hasn’t been updated yet rather than an official recommendation.

Furthermore, the Harbor maintainers have supported users upgrading to newer PostgreSQL versions, as seen in discussions like [Harbor Issue #20800](https://github.com/goharbor/harbor/issues/20800). This suggests that newer versions (e.g., PostgreSQL 15 or even 17) are compatible or becoming so, rather than strictly adhering to PostgreSQL 14.

As noted, there has already been progress toward upgrading PostgreSQL versions in certain Harbor-related contexts, like [Harbor Issue #19512](https://github.com/goharbor/harbor/issues/19512), where efforts were made to move to PostgreSQL 15.

carrodher:
Since Harbor is a product maintained by VMware, let me try to ping the Harbor maintainers internally to clarify what the default version is and what the plans are to support more recent versions in the short term.

Issue #8916, opened 2022-02-05
[bitnami/ghost] `smtpExistingSecret` value is ignored


shosti:
### Name and Version

bitnami/ghost 16.0.7

### What steps will reproduce the bug?

```bash
$ helm template ghost bitnami/ghost --version 16.0.7 \
    --set ghostHost=ghost.local \
    --set smtpUser=bob \
    --set smtpExistingSecret=mysecret  | grep -a5 SMTP
```

### Are you using any custom parameters or values?

See the reproduction command.

### What is the expected behavior?

From the documentation, I expected `GHOST_SMTP_PASSWORD` to be set and pull its value from `smtpExistingSecret`.

### What do you see instead?

`GHOST_SMTP_PASSWORD` isn't set unless `smtpPassword` is set. 

### Additional information

I can workaround by setting `smtpPassword` to a junk value, in which case `GHOST_SMTP_PASSWORD` is actually pulled from `existingSecret` (*not* `smtpExistingSecret`):

```bash
$ helm template ghost bitnami/ghost --version 16.0.7 --set ghostHost=ghost.local --set smtpUser=bob --set smtpPassword="-" --set existingSecret=mysecret  | grep -a5 SMTP
              value: "80"
            - name: GHOST_EXTERNAL_HTTPS_PORT_NUMBER
              value: "443"
            - name: GHOST_SKIP_BOOTSTRAP
              value: "no"
            - name: GHOST_SMTP_USER
              value: "bob"
            - name: GHOST_SMTP_PASSWORD
              valueFrom:
                secretKeyRef:
                  name: mysecret
                  key: smtp-password
```
But this isn't what the documentation would lead me to believe is correct behavior.

carrodher:
Please, [note the Bitnami team is not actively monitoring GitHub Issues for this Helm chart](https://github.com/bitnami/charts/tree/master/bitnami/ghost#community-supported-solution). This issue will stay open for 20 days to allow the community to contribute, after 15 days without activity the issue will be marked as stale being closed after 5 days.

If you would like to contribute to this Helm chart by fixing a bug or adding a new feature, feel free to create a PR, the Bitnami team will be happy to review it and provide feedback.

PR #31179, opened 2024-12-28
[bitnami/Ghost] Correctly handle smtpExistingSecret value

AverageMarcus:
### Description of the change

Correctly handles setting either `.Values.smtpPassword` or `.Values.smtpExistingSecret`. Default smtp-password is now created in its own Secret if an existing is not provided.

### Benefits

Fixes the need to provide a dummy `smtpPassword` value when using an external secret.

### Possible drawbacks

The SMTP secret, if an existing Secret resource is not provided, is now in its own Secret resource rather than the one used for the Ghost auth credentials (to be more in-line with the database secret)

### Applicable issues

- fixes https://github.com/bitnami/charts/issues/8916

### Checklist

### DIFF
--- a/bitnami/ghost/templates/deployment.yaml
+++ b/bitnami/ghost/templates/deployment.yaml
@@ -204,13 +204,11 @@ spec:
             - name: GHOST_SMTP_USER
               value: {{ .Values.smtpUser | quote }}
             {{- end }}
-            {{- if .Values.smtpPassword }}
             - name: GHOST_SMTP_PASSWORD
               valueFrom:
                 secretKeyRef:
-                  name: {{ include "common.secrets.name" (dict "existingSecret" .Values.existingSecret "context" $) }}
-                  key: {{ include "common.secrets.key" (dict "existingSecret" .Values.existingSecret "key" "smtp-password") }}
-            {{- end }}
+                  name: {{ include "ghost.smtpSecretName" . }}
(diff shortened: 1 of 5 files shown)


bitnami/ghost/templates/deployment.yaml
```diff
@@ -204,13 +204,11 @@ spec:
             - name: GHOST_SMTP_USER
               value: {{ .Values.smtpUser | quote }}
             {{- end }}
-            {{- if .Values.smtpPassword }}
```
alvneiayu, line 207: Hi, the pipelines are failing because you need to keep this. If not you are deploying a secret without the key smtp-password

bitnami/ghost/templates/secrets.yaml
```diff
@@ -22,9 +22,23 @@ data:
   ghost-password: {{ randAlphaNum 10 | b64enc | quote }}
   {{- end }}
   {{- end }}
-  {{- if and .Values.smtpPassword (not .Values.smtpExistingSecret) }}
+---
+{{- end }}
+
+{{- if not .Values.smtpExistingSecret }}
```
alvneiayu, line 28: you need to check here the smtpPassword, you are creating an empty secret without data in this case.

alvneiayu (changes requested):
Thanks a lot for sending a PR. I included some comments.

PR #31208, opened 2025-01-03
[bitnami/moodle] #31199 Fix invalid imagePullSecrets on container level

### LINKED ISSUE #31199: [bitnami/moodle] imagePullSecrets on container level is against spec
### Name and Version

bitnami/moodle 4.5.1

amd64

1. Use bitnami/moodle 4.5.1 chart and custom values.yaml
2. Add some `certificates.customCAs` entries
3. Deployment will fail with `".spec.template.spec.initContainers[name="certificates"].imagePullSecrets: field not declared in schema"` due to `imagePullSecrets` for init container "certificates" are defined on container level, not Pod level. This is against spec: https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#containers
4. Line of issue is here: https://github.com/bitnami/charts/blob/1fe7c59f415c46e5b1e002229d5d2027f167a320/bitnami/moodle/templates/deployment.yaml#L100C19-L100C20

```
certificates:
  ## @param certificates.customCAs Defines a list of secrets to import into the container trust store
  ##
  customCAs:
    - secret: my-custom-ca
```

Deploy without error

### What do you see instead?

Deployment fails during spec validation.

### Additional information

Remove the `imagePullSecrets` stanza from init container "certificates"

RalfBarth:
### Description of the change

Defining imagePullSecrets on container level is against spec (https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#containers). They have to be defined on pod level, which is already covered.

### Benefits

Being able to deploy without error with `certificates.customCAs` defined.

### Possible drawbacks

None as `imagePullSecrets` on pod level are already covered.

### Applicable issues

- fixes #31199 

### Checklist

### DIFF
--- a/bitnami/moodle/values.yaml
+++ b/bitnami/moodle/values.yaml
@@ -927,6 +927,7 @@ certificates:
   ## @skip certificates.image.tag Container sidecar image tag (immutable tags are recommended)
   ## @param certificates.image.digest Container sidecar image digest in the way sha256:aa.... Please note this parameter, if set, will override the tag
   ## @param certificates.image.pullPolicy Container sidecar image pull policy
+  ## @param certificates.image.pullSecrets Container sidecar image pull secrets
   ##
   image:
     registry: docker.io
@@ -937,6 +938,14 @@ certificates:
     ## ref: https://kubernetes.io/docs/concepts/containers/images/#pre-pulled-images
     ##
     pullPolicy: IfNotPresent
+    ## Optionally specify an array of imagePullSecrets.
+    ## Secrets must be manually created in the namespace.
+    ## ref: https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/
+    ## e.g:
+    ## pullSecrets:
+    ##   - myRegistryKeySecretName
+    ##
+    pullSecrets: []
 ## @section NetworkPolicy parameters
 
 ## Network Policy configuration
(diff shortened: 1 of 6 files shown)


bitnami/moodle/values.yaml
```diff
@@ -938,14 +937,6 @@ certificates:
...
     ##
     pullPolicy: IfNotPresent
-    ## Optionally specify an array of imagePullSecrets.
-    ## Secrets must be manually created in the namespace.
-    ## ref: https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/
-    ## e.g:
-    ## pullSecrets:
-    ##   - myRegistryKeySecretName
-    ##
-    pullSecrets: []
```
juan131, line 948: I'm fine with removing the usage of this parameter at container spec, but it is still required to be used on the `moodle.imagePullSecrets` helper, right?

RalfBarth: You're right, good catch! I suggest leaving the `pullSecrets` list in to be able to provide a pull secret for the certificates container if needed. The linked helper will take care of that. 

Note, according to https://github.com/bitnami/charts/blob/5fd0be75c1a4714f324acd55a0a304ab20a46647/bitnami/common/templates/_images.tpl#L35 the use of `common.images.pullSecrets` seems to be deprecated and you might consider refactoring to the suggested replacement `common.images.renderPullSecrets`

juan131: Yes, you can replace the deprecated helper with the new one.
Please also revert the changes that remove the parameter from the `values.yaml` given the parameter would actually be useful.

RalfBarth: Done.

juan131 (approved):
LGTM

Issue #23032, opened 2024-02-02
Helm upgrade failed: template: rabbitmq/templates/networkpolicy.yaml:47:64: executing "rabbitmq/templates/networkpolicy.yaml" at <.Values.rts.networkPolicy.extraEgress>: nil pointer evaluating interface {}.networkPolicy


tufberg:
### Name and Version

bitnami/rabbitmq 12.9.1

### What architecture are you using?

None

### What steps will reproduce the bug?

I've enabled networkpolicy and extraEgress in my HelmRelease using Flux

```yaml
    networkPolicy:
      enabled: true
      allowExternal: true
      extraIngress:
        - ports:
            - port: 1883
            - port: 8883
      extraEgress:
        - ports:
            - port: 6379
              protocol: TCP
            - port: 1883
              protocol: TCP
            - port: 8883
              protocol: TCP
          to:
            - podSelector:
                - matchLabels:
                    - app.kubernetes.io/instance: redis
```

Doing this I'll receive an error deploying this HelmRelease.
```
Helm upgrade failed: template: rabbitmq/templates/networkpolicy.yaml:47:64: executing "rabbitmq/templates/networkpolicy.yaml" at <.Values.rts.networkPolicy.extraEgress>: nil pointer evaluating interface {}.networkPolicy
```

### What is the expected behavior?

It should enable the egress Network policy.

### What do you see instead?

I checked the template code for **[networkPolicy](bitnami/rabbitmq/templates/networkpolicy.yaml)** and I see this line which seems wrong

`{{- include "common.tplvalues.render" ( dict "value" .Values.rts.networkPolicy.extraEgress "context" $ ) | nindent 4 }}`

Shouldn´t it be?

`{{- include "common.tplvalues.render" ( dict "value" .Values.networkPolicy.extraEgress "context" $ ) | nindent 4 }}`

carrodher:
Thank you for bringing this issue to our attention. We appreciate your involvement! If you're interested in contributing a solution, we welcome you to create a pull request. The Bitnami team is excited to review your submission and offer feedback. You can find the contributing guidelines [here](https://github.com/bitnami/charts/blob/main/CONTRIBUTING.md).

Your contribution will greatly benefit the community. Feel free to reach out if you have any questions or need assistance.

PR #31210, opened 2025-01-03
[bitnami/rabbitmq] remove undefined rts key from network policy template

JustinHallquist:
### Description of the change

Removes a undefined `rts` key from `networkpolicy.yaml`

Currently, modifying the `.Values.extraEgress` fails chart rendering as the template is looking for an `rts` key, which is neither specified in the documentation of the `.Values` nor present _anywhere_ else in the chart for reference points as to how to use it.

The error produced when `.Values.extraEgress` is defined:
```
Helm upgrade failed: template: rabbitmq/templates/networkpolicy.yaml:47:64: executing "rabbitmq/templates/networkpolicy.yaml" at <.Values.rts.networkPolicy.extraEgress>: nil pointer evaluating interface {}.networkPolicy
```

The work around is to define both `.Values.extraEgress`  and `.Values.rts.extraEgress` to pass the initial feature flag and enable the block render

By removing the undefined `rts` key, the workaround is no longer necessary and the template matches the `.Values.extraEgress` documentation

### Benefits

*  `.Values.extraEgress` expected behavior reestablished
* Restores `.Values.extraEgress` documentation validity
* Matches the `extraIngress` interface

### Possible drawbacks

Nonbackwards compatibility for anyone who added a `.Values.rts` block to work around the issue

### Applicable issues

- fixes [#23032](https://github.com/bitnami/charts/issues/23032)

### Checklist

### DIFF
--- a/bitnami/rabbitmq/Chart.yaml
+++ b/bitnami/rabbitmq/Chart.yaml
@@ -30,4 +30,4 @@ maintainers:
 name: rabbitmq
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/rabbitmq
-version: 15.2.1
+version: 16.0.0
--- a/bitnami/rabbitmq/Chart.yaml
+++ b/bitnami/rabbitmq/Chart.yaml
@@ -30,4 +30,4 @@ maintainers:
 name: rabbitmq
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/rabbitmq
-version: 16.0.0
+version: 15.2.2
--- a/bitnami/rabbitmq/CHANGELOG.md
+++ b/bitnami/rabbitmq/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## 16.0.0 (2025-01-03)
+## 15.2.2 (2025-01-07)
 
 * [bitnami/rabbitmq] remove undefined rts key from network policy template ([#31210](https://github.com/bitnami/charts/pull/31210))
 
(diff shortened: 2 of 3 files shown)


JustinHallquist:
Is there a way to see the failed spec for the VIB Verify test?
All I can see is `Error: Execution graph succeeded, however some tasks didn't pass the verification.` and an execution graph url link I do not have access to view.

Aside from that, it seems unlikely that this change would break the test since the runtime-parameters.yaml does not specify a `networkPolicy.extraEgress` so it would have been feature flagged off with or without the update.

For some additional digging, I did run the test on two different clusters:
* EKS cluster which passed (I did use a ClusterIP rather than a Load Balancer)
* local k3s cluster which failed 2775 -> 2777 mnensia perms (the rest of the tests passed) so I believe that's more likely due to my storage class provisioner defaults and nothing to do with network policies

Would you happen to have any suggestions for next steps here on my end?

bitnami/rabbitmq/Chart.yaml
```diff
@@ -30,4 +30,4 @@ maintainers:
 name: rabbitmq
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/rabbitmq
-version: 15.2.1
+version: 16.0.0
```
dgomezleon, line 33:
```suggestion
version: 15.2.2
```

dgomezleon (changes requested):
Just a minor change related to the versioning

dgomezleon (approved):

PR #31271, opened 2025-01-09
[bitnami/keycloak] add option to use dedicated version bound headless service for jgroups discovery

### LINKED ISSUE #31072: [bitnami/keycloak] Keycloak rolling upgrades with Infinispan
### Name and Version

bitnami/keycloak latest

During rolling upgrade of Keycloak to the next Major/Minor version it could happen that the first pod that is updated can not start up due to breaking changes in Infinispan. Basically the new Pod can not connect to the infinispan cluster. 

To mitigate this, it would be nice if the Keycloak pods only connect to infinispan nodes that are running the same version. This way the cache is busted during an upgrade, but it will always be possible to do a rolling update. With the new persistent-session feature, also the sessions will not be lost anymore.

This solution is inspired by: https://github.com/keycloak/keycloak/discussions/21067#discussioncomment-8855429 

1. Include the version on the name of the headless-service (e.g. ...-keycloak-headless-25-0) and let the headless service only connect to Keycloak pods running the corresponding version. E.g. use a newly introduced label on the Keycloak pods as additional selector.
2. Update the StatefulSet to include the new pod label. 
3. Adapt `JAVA_OPTS_APPEND` to use the new name of the headless service, e.g.: `jgroups.dns.query=%s-headless-%s.%s.svc.%s`

The Major/Minor version could be read from the image tag that is being used: `regexReplaceAll "^([0-9]+)\\.([0-9]+).*" .Values.image.tag "$1-$2"`

_No response_

antikalk:
Add possibility to version bind the ispn discovery. The new config property is `.cache.useHeadlessServiceWithAppVersion`. This solution is inspired by: https://github.com/keycloak/keycloak/discussions/21067#discussioncomment-8855429 

### Description of the change

Added a config property `.cache.useHeadlessServiceWithAppVersion`. When this is set to `false` (default) the behavior stays the same as before and the Keycloak ISPN Cache will continue to use the default headless service for discovery. 

If the value is set to true:
- New headless service "`(include "common.names.fullname" .)-headless-ispn-(replace "." "-" .Chart.AppVersion)`"  is created that contains an additional selector for `app.kubernetes.io/app-version: {{ .Chart.AppVersion }}`. This headless service will only discover for keycloak pods that are running the corresponding appVersion
- Env variable `JAVA_OPTS_APPEND` is adapted to use the new headless service for jgroups discovery.

An additional label `app.kubernetes.io/app-version: {{ .Chart.AppVersion }}` has been added to the statefulset to allow for selecting pods by their app version.

### Benefits

Enable the possibility to do rolling upgrades, even with breaking changes in the infinispan update. Previously, when there were breaking changes with infinispan, the new version Keycloak pods can not startup, as they would try to connect to the incompatible cluster. This rendered it necessary to kill all Keycloak pods to proceed with the upgrade, causing a downtime.

### Possible drawbacks

With the new property enabled, when deploying a new app version the infinispan cache will not be synched over to the new Keycloak instances, as these will create a new cluster for themselves. This might result in data loss and slower response times shortly after the update.

### Applicable issues

- fixes #31072

### Checklist

### DIFF
--- a/bitnami/keycloak/CHANGELOG.md
+++ b/bitnami/keycloak/CHANGELOG.md
@@ -1,12 +1,13 @@
 # Changelog
 
-## 24.4.0 (2025-01-XX)
+## 24.4.0 (2025-01-09)
 
-* [bitnami/keycloak] Add support to bind the headless-service to the app version to allow for rolling upgrades, even with breaking changes in the infinispan cache (closes [#31072](https://github.com/bitnami/charts/issues/31072)).
+* [bitnami/keycloak] add option to use dedicated version bound headless service for jgroups discovery ([#31271](https://github.com/bitnami/charts/pull/31271))
 
-## 24.3.2 (2024-12-20)
+## <small>24.3.2 (2024-12-30)</small>
 
-* [bitnami/keycloak] Update KEYCLOAK_ADMIN env variables deprecation ([#30636](https://github.com/bitnami/charts/pull/30636))
+* [bitnami/*] Fix typo in README (#31052) ([b41a51d](https://github.com/bitnami/charts/commit/b41a51d1bd04841fc108b78d3b8357a5292771c8)), closes [#31052](https://github.com/bitnami/charts/issues/31052)
+* [bitnami/keycloak] Update KEYCLOAK_ADMIN env variables deprecation (#30636) ([668bd27](https://github.com/bitnami/charts/commit/668bd2772c5ea45af7b1b57141c1776ccf4169f4)), closes [#30636](https://github.com/bitnami/charts/issues/30636)
 
 ## <small>24.3.1 (2024-12-16)</small>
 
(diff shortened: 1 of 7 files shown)


dgomezleon (approved):
LGTM. Thanks for this great contribution

nix-power:
What possible data-loss is expected in the  version upgrade ?

antikalk:
@nix-power Since the persistent session feature is now fully supported by Keycloak since version 26, User Sessions should not get lost anymore - as those are now persisted in the database by default.

What will be lost though would be usually more short-lived items that are persisted to the cache only like authentication sessions, login failures or action tokens. If you would be forced to recreate the infinispan cluster on the Keycloak upgrade (when the new version does not sync with the old one) - you would anyway experience the same data loss when force upgrading. 

What could make sense though is to only activate this new property, when on a keycloak upgrade the infinispan versions are incompatible to each other. Enabling you to do a rolling upgrade without force killing every Keycloak Pod. Then after the update, revert to setting the new property to false. This way you would mitigate unneccessary cache recreations on upgrades where the infinispan does not break.

PR #31374, opened 2025-01-15
bitnami/zipkin Remove quote for the host

ptrekelsseg:
This is needed because it will later be combined with the port in the configmap.

### Description of the change

Removed the quote from the host value so it can be combined in configmap with the port.

### Benefits

Now it is possible to use external database.

### Possible drawbacks

None

### Applicable issues

None

### Additional information

Nothing.

### Checklist

### DIFF
--- a/bitnami/zipkin/CHANGELOG.md
+++ b/bitnami/zipkin/CHANGELOG.md
@@ -1,8 +1,12 @@
 # Changelog
 
-## 1.1.1 (2025-01-07)
+## 1.1.2 (2025-01-15)
 
-* [bitnami/zipkin] Release 1.1.1 ([#31228](https://github.com/bitnami/charts/pull/31228))
+* bitnami/zipkin Remove quote for the host ([#31374](https://github.com/bitnami/charts/pull/31374))
+
+## <small>1.1.1 (2025-01-07)</small>
+
+* [bitnami/zipkin] Release 1.1.1 (#31228) ([a71714c](https://github.com/bitnami/charts/commit/a71714cd224837c124b989f7ffd5b795ca0fb699)), closes [#31228](https://github.com/bitnami/charts/issues/31228)
 
 ## 1.1.0 (2024-12-10)
 
--- a/bitnami/zipkin/CHANGELOG.md
+++ b/bitnami/zipkin/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## 1.1.2 (2025-01-20)
+## 1.1.2 (2025-01-21)
 
 * bitnami/zipkin Remove quote for the host ([#31374](https://github.com/bitnami/charts/pull/31374))
 
--- a/bitnami/zipkin/CHANGELOG.md
+++ b/bitnami/zipkin/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## 1.1.2 (2025-01-15)
+## 1.1.2 (2025-01-20)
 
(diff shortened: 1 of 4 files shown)


fmulero (changes requested):
Thanks a lot @ptrekels for your contribution.

I don't have clear the error your are trying to fix. If I understood well the issue is here:  https://github.com/bitnami/charts/blob/422b3a72023f2c47fded9e3d696c855540d1b158/bitnami/zipkin/templates/configmap.yaml#L16

So it should be fixed there with something like this (not tested):
```yaml
CASSANDRA_CONTACT_POINTS: {{ printf "%s:%d" ( include "zipkin.cassandra.host" . ) ( include "zipkin.cassandra.port" . ) }}
```

Could you share more details about your issue and how to reproduce it?

ptrekels:
@fmulero indeed the error is in that line 16 of the configmap template. As the host string would be quoted there this would introduce something like "aa.bb.com":1234  This of course is no legal yaml syntax and so my helm complained about it.

Steps to reproduce it was simply trying to use an external cassandra db..

And printf fix is indeed also possible and maybe a better fix.

fmulero (approved):
I've just added a little change. As I said I didn't the test my suggested change and there was a little failure.

ptrekelsseg:
I did not used this fix correctly until now it seems. And it seems there is still an issue. The zipkin.cassandra.host is quoted and so introduces the issue I think.

CASSANDRA_CONTACT_POINTS: {{ printf "%s:%d" ( include "zipkin.cassandra.host" . | replace "\"" "" ) ( include "zipkin.cassandra.port" . | int ) }}

'unquoting' it seems to fix the issue for me, but don't now if this is the best solution?

fmulero:
Sorry @ptrekelsseg  I don't fully understand, What issue are you facing now? Could you share how to reproduce it?

ptrekelsseg:
I'm simply using a values file like this:
`
storageType: cassandra3
externalDatabase:
  host: cascluster-dev-service
  port: 9042
  user: cassandra-zipkin
  existingSecret: cascluster-user-zipkin
  existingSecretPasswordKey: password
  cluster:
    datacenter: dev
  keyspace: bitnami_zipkin
cassandra:
  enabled: false
`
I checked out the bitname charts locally and I'm able to reproduce the issue with:
helm template test bitnami/charts/bitnami/zipkin/ --values testoverlays/dev/zipkin-values.yaml --show-only templates/configmap.yaml

Here the external db host value is quoted: https://github.com/bitnami/charts/blob/main/bitnami/zipkin/templates/_helpers.tpl#L32
And here it is used: https://github.com/bitnami/charts/blob/main/bitnami/zipkin/templates/configmap.yaml#L16

And like I said, I thought I tested your change. But was wrong and I was still using my old fix.. So I did not notice your fix still gives an issue.

ptrekelsseg:
2 possible solutions I tested:
``` yaml
CASSANDRA_CONTACT_POINTS: {{ printf "%s:%d" ( trimAll "\"" (include "zipkin.cassandra.host" .) )  ( include "zipkin.cassandra.port" . | int ) }}
CASSANDRA_CONTACT_POINTS2: {{ printf "%s:%d" ( include "zipkin.cassandra.host" . | replace "\"" "" ) ( include "zipkin.cassandra.port" . | int ) }}
```

ptrekelsseg:
@fmulero : I pushed a proposal in https://github.com/bitnami/charts/pull/31866

PR #31377, opened 2025-01-15
[bitnami/thanos] feature: allow configuring the Thanos ruler rule-file via ruler.ruleFile parameter

### LINKED ISSUE #31375: Set the ruler config file flag via the helm chart
### Name and Version

bitnami/thanos 15.9.2

Right now, in the thanos chart, the ruler's `--rule-file=/conf/rules/*.yml` arg is hardcoded in the ruler Statefulset.

I'd like to see the `--rule-file` as a configurable parameter, so that we can change the directory or file extension (`yaml` vs `yml`, for instance). It would look like this:

```yaml
ruler:
  ## @param ruler.ruleFile Change the location Thanos checks for rule files
  ##
  ruleFile: "/conf/rules/*.yml"
```

It would still default to the current location so it does not introduce a breaking change. I will work on this PR.

none, as I don't want to have to override the entire args section.

jessebot:
### Description of the change

Create a `ruler.ruleFile` config parameter for Thanos ruler.

### Benefits

This allows users to specify a different directory or file extension to check for Thanos ruler rule files. Example would be to check for `.yaml` vs `.yml` files, so users can more easily import rules from prometheus, for instance.

### Possible drawbacks

none that I can think of

### Applicable issues

- fixes #31375

### Checklist


juan131 (approved):
Thanks for your contribution! LGTM

PR #19841, opened 2023-10-09
[bitnami/*] Skip image's tag in the README files of the Bitnami Charts

jotamartos:
### Description of the change

In order to standardize the documentation, we would like to remove the images' tag information from the README files. This change is related to this other previous PR https://github.com/bitnami/charts/pull/19835 

### Benefits

Make the documentation neutral for all distros

### Checklist

- [NA] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/). This is *not necessary* when the changes only affect README.md files.
- [NA] Variables are documented in the values.yaml and added to the `README.md` using [readme-generator-for-helm](https://github.com/bitnami-labs/readme-generator-for-helm)
- [X] Title of the pull request follows this pattern [bitnami/<name_of_the_chart>] Descriptive title
- [X] All commits signed off and in agreement of [Developer Certificate of Origin (DCO)](https://github.com/bitnami/charts/blob/main/CONTRIBUTING.md#sign-your-work)

PR #31474, opened 2025-01-20
[bitnami/keycloak] Include image.tag value in Keycloak README.md

EddyMM:
### Description of the change

Explicitly mention in the README.md that the image tag for a custom Keycloak image can be specified using the `image.tag` value.

Including the tag in the `image.repository` value results in the helm chart's `image.tag` value overwriting it.

### Without Tag:

```yaml
image:
  repository: docker.io/../custom-keycloak:26.1.0
```

Helm Diff:

```yaml
image: docker.io/.../custom-keycloak:26.1.0-debian-12-r0
```

### With Tag:

```yaml
image:
  repository: docker.io/../custom-keycloak
  tag: 26.1.0
```

Helm Diff:

```yaml
image: docker.io/.../custom-keycloak:26.1.0
```

### Benefits

- Consumers of `bitnami/keycloak` will know which image value to use to specify (`tag` in this case) for the custom keycloak image they're using

### Checklist


carrodher:
The `image.tag` field is omitted from the table with all the parameters on purpose to standardize the README for other products like [Tanzu Application Catalog](https://www.vmware.com/products/app-platform/tanzu-application-catalog) where those tags can point to different distros and not only Debian 12, see https://github.com/bitnami/charts/pull/19841.

EddyMM:
Thanks for clarifying @carrodher . Is there any other documentation for someone to discover the `image.tag` if not from the README?

carrodher:
All parameters are defined in the _values.yaml_, which provides a short explanation; see https://github.com/bitnami/charts/blob/8a2a477bc67c7badb2fb32a1949d19b8801f7699/bitnami/keycloak/values.yaml#L102.

In fact, that's the metadata used to autogenerate the table that appears in the README, so the information is the same.

EddyMM:
Got it, thanks Carlos.

PR #31526, opened 2025-01-23
ci: Skip hardcoded images check on NOTES.txt files

juan131:
### Description of the change

This PR adapts the "Look for hardcoded images" job on "CI Pipeline" workflow so we skip looking for hardcoded images on 
NOTES.txt files.

We usually include instructions to create "client pods" on the notes and, when using images for the clients not included in the chart (e.g. using a `bitnami/golang:latest` to compile some Go example app and use it as a client) it's useful to use hardcoded images with rolling tags.


fmulero (approved):
LGTM

Issue #31527, opened 2025-01-23
[bitnami/etcd] issue with preupgrade job during scale down


sachinphogat:
### Name and Version

bitnami/etcd 11.0.2

### What architecture are you using?

None

### What steps will reproduce the bug?

1. Deploy bitnami/etcd 11.0.2 with replicaCount as 3.
2. Modify the replicaCount as 1.
3. PreUpgrade job will fail with below error.

**Logs:**
```
Error: etcdclient: no available endpoints
etcd 10:45:46.41 ERROR ==> Unable to list members, are all members healthy?
```
### Are you using any custom parameters or values?
```yaml
    auth:
      client:
        caFilename: ca.crt
        certFilename: tls.crt
        certKeyFilename: tls.key
        enableAuthentication: true
        existingSecret: etcd-etcd-client-tls
        secureTransport: true
        useAutoTLS: false
      peer:
        caFilename: ca.crt
        certFilename: tls.crt
        certKeyFilename: tls.key
        enableAuthentication: true
        existingSecret: etcd-etcd-peer-tls
        secureTransport: true
        useAutoTLS: false
      rbac:
        create: false
    autoCompactionRetention: 3h
    customLivenessProbe:
      exec:
        command:
        - /opt/bitnami/scripts/etcd/healthcheck.sh
      failureThreshold: 5
      initialDelaySeconds: 60
      periodSeconds: 30
      successThreshold: 1
      timeoutSeconds: 5
    defrag:
      enabled: true
    replicaCount: 3
```
### What is the expected behavior?

Should be able to scale down helm chart without any issues.

### What do you see instead?
```
etcd 10:45:46.37 INFO  ==>
etcd 10:45:46.37 INFO  ==> Welcome to the Bitnami etcd container
etcd 10:45:46.37 INFO  ==> Subscribe to project updates by watching https://github.com/bitnami/containers
etcd 10:45:46.38 INFO  ==> Did you know there are enterprise versions of the Bitnami catalog? For enhanced secure software supply chain features, unlimited pulls from Docker, LTS support, or application customization, see Bitnami Premium or Tanzu Application Catalog. See https://www.arrow.com/globalecs/na/vendors/bitnami/ for more information.
etcd 10:45:46.38 INFO  ==>

Error: etcdclient: no available endpoints
etcd 10:45:46.41 ERROR ==> Unable to list members, are all members healthy?
```

carrodher:
[...]

PR #31539, opened 2025-01-24
[bitnami/etcd] Fix: issue with preupgrade job during scale down

sachinphogat:
### Description of the change

Removed condition for adding INITIAL_CLUSTER env variable to preupgrade job based on replica count. It should be added even when replicaCount is 1.

### Benefits

Due to above change etcd preupgrade will get succeeded when we are scaling down from 3 replicas to 1.

### Possible drawbacks

No possible drawbacks known.

### Applicable issues

- fixes #https://github.com/bitnami/charts/issues/31527

### Checklist


taliesins:
Looks like work around for now is just to use a replicaCount greater then 1, until this merged in.

This is why:
https://github.com/bitnami/containers/blob/main/bitnami/etcd/3.5/debian-12/rootfs/opt/bitnami/scripts/etcd/preupgrade.sh#L30

What I don't like about this pre upgrade job is that it uses a pre-upgrade Helm hook which means that the job gets executed every single time. When executing Helm from Ansible or Terraform, I don't want to see this resource change every time. I really wish we could use another approach (i.e. init container): https://trstringer.com/helm-hooks-anti-pattern/

alvneiayu (approved):

Issue #31646, opened 2025-01-28
[bitnami/jaeger]: cqlsh fails with multiple contact points, option to disable


beyimjan:
### Name and Version

bitnami/jaeger 5.1.5

### What is the problem this feature will solve?

1. With comma separated list of peers
2. I get this error
```
21:01:11.71 INFO  ==> Connecting to the Cassandra instance ip1,ip2,ip3:9042

Traceback (most recent call last):
  File "/opt/bitnami/cassandra/bin/cqlsh.py", line 88, in <module>
    main(sys.argv[1:], CASSANDRA_PATH)
  File "/opt/bitnami/cassandra/bin/../pylib/cqlshlib/cqlshmain.py", line 2321, in main
    shell = Shell(hostname,
            ^^^^^^^^^^^^^^^
  File "/opt/bitnami/cassandra/bin/../pylib/cqlshlib/cqlshmain.py", line 393, in __init__
    load_balancing_policy=WhiteListRoundRobinPolicy([self.hostname]),
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/bitnami/cassandra/bin/../lib/cassandra-driver-internal-only-3.29.0.zip/cassandra-driver-3.29.0/cassandra/policies.py", line 425, in __init__
  File "/opt/bitnami/cassandra/bin/../lib/cassandra-driver-internal-only-3.29.0.zip/cassandra-driver-3.29.0/cassandra/policies.py", line 426, in <listcomp>
  File "/opt/bitnami/python/lib/python3.11/socket.py", line 974, in getaddrinfo
    for res in _socket.getaddrinfo(host, port, family, type, proto, flags):
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
socket.gaierror: [Errno -2] Name or service not known
```
### What is the feature you are proposing to solve the problem?

I wish there were a way to disable the database check step to address this issue in my multi-node Cassandra setup, especially since cqlsh doesn't support multiple contact points.

carrodher:
Thank you for bringing this issue to our attention. We appreciate your involvement! If you're interested in contributing a solution, we welcome you to create a pull request. The Bitnami team is excited to review your submission and offer feedback. You can find the contributing guidelines [here](https://github.com/bitnami/charts/blob/main/CONTRIBUTING.md).

[...]

PR #31654, opened 2025-01-29
[bitnami/jaeger] Add option to skip Cassandra readiness check

beyimjan:
This PR helps mitigate issue #31646 by adding a configuration option to skip the Cassandra readiness check when not needed.

Add `collector.skipCassandraCheck` parameter to control whether the 
Cassandra readiness check init container is created. When set to true,
the init container will be skipped.

The default value is set to false to maintain backward compatibility
with existing deployments and ensure the readiness check continues to
run by default.

### Checklist


fmulero (commented):
Thanks a lot @beyimjan for your contribution!

I know this change could help you in your use case but don't you think is worth to adapt the test to that scenario? I mean, if `jaeger.cassandra.host` returns a list of hosts, check if any of them is ready. I think this could be  a better solution in this case.

beyimjan:
I want to make sure the implementation correctly handles both single and multiple hosts. Since `host` can be either `"host1"` or `"host1,host2,..."`, I suggest using `IFS=',' read -ra ADDR <<< "$host"` to properly split multiple hosts.

```bash
IFS=',' read -ra ADDR <<< "$host"
for h in "${ADDR[@]}"; do
  echo "Processing $h"
done
```

This way, whether `host` contains a single value (`host="host1"`) or multiple values (`host="host1,host2,host3"`), the loop processes them correctly.

Should I open another PR for this change, or can I continue in this one?

fmulero:
Thanks @beyimjan, please continue in this one

PR #31840, opened 2025-02-08
[bitnami/postgresql] use adminPassword for metrics user when custom user is not set on primary

gaelreyrol:
### Description of the change

This PR applies the same changes from #30720 to the primary statefulset.

### Benefits

Resolve error when creating primary postgresql pod with metrics enabled.

### Applicable issues

#30680


juan131 (approved):

PR #31849, opened 2025-02-10
[index] Update gchat-notification workflow

migruiz4:
### Description of the change

Updates the gchat-notification workflow location


fmulero (approved):
LGTM

PR #31987, opened 2025-02-19
Update README.md - readinessProbe path

eljeffeg:
The readinessProbe.path stated it was for the liveness path


carrodher:
Thanks for the fix! Please note the README is automatically generated using the information from the _values.yaml_, could you please apply the fix at https://github.com/bitnami/charts/blob/main/bitnami/nginx/values.yaml#L402 and bump the chart version in the _Chart.yaml_? Thanks

PR #32110, opened 2025-02-21
[bitnami/mariadb] Set `usePasswordFiles=true` by default

migruiz4:
### Description of the change

### Benefits

### Possible drawbacks

None

### Applicable issues

None

### Additional information

N/A

### Checklist


gongomgra (approved):

samip5:
Breaking change as a minor again? Why?

migruiz4:
Hi @samip5,

Although this changes the default behavior, we do not bump major versions unless we consider backward-compatibility would be affected.

Did you encounter any issue related to this change?

monoxane:
This just took out our primary db nodes with a patch automerge from renovate, secondary was fine but the primary required additional values in the secret that were not previously necessary.

samip5:
It seems Bitnami cannot be trusted with versioning according to semver.

migruiz4:
Hi @monoxane,
 
Could you please provide more details about your case in a new issue? 
That way we can work on a fix as soon as possible.

samip5:
> Hi @monoxane,
> 
>  
> 
> Could you please provide more details about your case in a new issue? 
> 
> That way we can work on a fix as soon as possible.

The fix is to not introduce breaking changes in a minor please.

migruiz4:
I'm sorry for the inconveniences, we will revert this change as soon as possible (#32158).

Regarding the major/minor version discussion, as I previously mentioned if we strictly follow semver guidelines, major versions are meant for non-backward compatible changes.

In retrospective, it is easy to say that this could have been a major at first, but if this change would have been seamless as we expected we wouldn't even be having this discussion.

We will investigate this issue and try to improve our CI/CD pipeline to help us detect this kind of issues before the releases take place.

migruiz4:
After some further investigation, I found it is a bug with the existing `usePasswordfiles` feature when `auth.username` is empty.

After fixing the issue, we expect users should be able to upgrade without issues.

This bug was omitted by our CI/CD pipeline because we do set `auth.username` in our testing values: (see [runtime-parameters.yaml](https://github.com/bitnami/charts/blob/7f0db451da4b457ce14ef6c0617c5d1e98916a07/.vib/mariadb/runtime-parameters.yaml#L4))

samip5:
If this was a single instance where Bitnami's charts have introduced similarly a major breakage in a minor, this wouldn't be such an issue but as this is at least 2nd time in a short period of time. We the end-users of these charts expect better from Bitnami.

So I don't think the revert is not needed, it should be reverted and done via a major bump instead.

migruiz4:
Version 20.4.1 was released 15 hours ago. I would appreciate if you could confirm that using the new version fixes the issue.

As I previously mentioned, upgrading from any 20.x.x version to 20.4.1 should be seamless, so no major upgrade is required.

samip5:
> That said, I’d like to remind everyone that Bitnami charts are an open-source project provided completely free for users.

I would like to remind you that they are are not fully open source when the actual code in the container images are behind a cascade of different somewhat close-source build pipelines but that's a different problem.

PR #32117, opened 2025-02-21
[bitnami/redis] Set `usePasswordFiles=true` by default

migruiz4:
### Description of the change

### Benefits

### Possible drawbacks

None

### Applicable issues

None

### Additional information

N/A

### Checklist


fmulero (dismissed):
LGTM

fmulero (approved):

bitnami/redis/templates/replicas/application.yaml
```diff
@@ -308,9 +308,9 @@ spec:
             - /bin/bash
             - -c
             - |
-              if [[ -f '/secrets/redis-password' ]]; then
-              export REDIS_PASSWORD=$(cat /secrets/redis-password)
-              fi
+              {{- if .Values.usePasswordFiles }}
```
JRemitz, line 311: Should this be `.Values.auth.usePasswordFiles`?

bitnami/redis/templates/sentinel/statefulset.yaml
```diff
@@ -520,9 +520,9 @@ spec:
             - /bin/bash
             - -c
             - |
-              if [[ -f '/secrets/redis-password' ]]; then
-                  export REDIS_PASSWORD=$(cat /secrets/redis-password)
-              fi
+              {{- if .Values.usePasswordFiles }}
```
JRemitz, line 523: Same feedback - should this be `.Values.auth.usePasswordFiles`?

adam-lebon:
I am deploying redis without authentication, since today the deployment is broken : redis pods are looking for a secret which doesn't exists (I don't need any secret since I disabled authentication)

Here are my values, I haven't had time to debug yet, so I'll stick to version 20.9.0
```yaml
---
architecture: replication
auth:
  enabled: false
master:
  count: 1
  resourcePreset: 2xlarge
  persistence:
    enabled: false
replica:
  persistence:
    enabled: false
sentinel:
  enabled: true
```

mariob1626:
@migruiz4

michielswaanen:
Our was broken, fixed it by adding `auth.usePasswordFiles=false`

xamma:
> I am deploying redis without authentication, since today the deployment is broken : redis pods are looking for a secret which doesn't exists (I don't need any secret since I disabled authentication)
> 
> Here are my values, I haven't had time to debug yet, so I'll stick to version 20.9.0
> 
> ```yaml
> ---
> architecture: replication
> auth:
>   enabled: false
> master:
>   count: 1
>   resourcePreset: 2xlarge
>   persistence:
>     enabled: false
> replica:
>   persistence:
>     enabled: false
> sentinel:
>   enabled: true
> ```

Yes this breaks the deployment of Redis deployed with auth disabled since there is no password file.
We need a conditional check to handle the password file only if auth.enabled is true.
Also for the startup script there needs to be a check if the password file exists first, I think. 
Rolled back to 20.9.0 as well

bitnami/redis/templates/replicas/application.yaml
Stevenpc3: It is also broken here https://github.com/bitnami/charts/pull/32117/files#diff-a1b6b9755fd3835c9419effa6f1c64c60a41f5dd2b9eecf02824e1d04c68a2f9R291

bitnami/redis/templates/master/application.yaml
```diff
@@ -288,9 +288,9 @@ spec:
             - /bin/bash
             - -c
             - |
-              if [[ -f '/secrets/redis-password' ]]; then
-              export REDIS_PASSWORD=$(cat /secrets/redis-password)
-              fi
+              {{- if .Values.usePasswordFiles }}
```
Stevenpc3, line 291: ```.Values.auth.usePasswordFiles```

bitnami/redis/values.yaml
```diff
@@ -162,7 +162,7 @@ auth:
   existingSecretPasswordKey: ""
   ## @param auth.usePasswordFiles Mount credentials as files instead of using an environment variable
   ##
-  usePasswordFiles: false
+  usePasswordFiles: true
```
Stevenpc3, line 165: changing a default value in a chart is 99% of the time a "BREAKING" change.  This affects all downstream users as they will not have to modify any charts that use this.  Was this confirmed to not be a breaking change?  It certainly looks like one... aside from the incorrect references in the templates.

Stevenpc3: The version number should likely have not been a minor bump.

PR #32125, opened 2025-02-21
[index] Update archive-full-index workflow dependencies

migruiz4:
Update dependencies of the archive-full-index workflow.


fmulero (approved):
Is worth to add dependabot config for this branch?

migruiz4:
@fmulero  I think dependabot can not be enabled for this branch because it is squashed

PR #32128, opened 2025-02-21
[bitnami/prometheus] Add toggle for Prometheus Kubernetes Service

ricardoapl:
### Description of the change

I would like to not expose Prometheus, but rather use a reverse proxy sidecar that handles mTLS

However, the Kubernetes Service manifest assumes we want to expose Prometheus - it will always target its container port

### Benefits

Users can decide whether or not they want to expose Prometheus

### Possible drawbacks

None

### Applicable issues

None

### Additional information

None

### Checklist


migruiz4 (commented):
Hi @ricardoapl, thank you for your contribution and I'm sorry for the late response.

I would like to make some suggestions for this PR:

- Please bump a minor version, since this adds a new feature for the chart.
- In the NOTES.txt, there is a message that tells users how to access prometheus service. It doesn't make sense when the service is not enabled.
- The ingress.yaml also relies on this service to work. Please adapt it accordingly.
- Since users may not be aware about this behavior, we should add a blocker condition during deployment when `ingress.enabled=yes and service.enabled=false`. This is done at the values validation helper [here](https://github.com/bitnami/charts/blob/cbfec4f962119b505b5c81b6d61448506b7a6216/bitnami/prometheus/templates/_helpers.tpl#L66-L78). See [prometheus.server.validateValues.thanosObjectStorageConfig](https://github.com/bitnami/charts/blob/cbfec4f962119b505b5c81b6d61448506b7a6216/bitnami/prometheus/templates/_helpers.tpl#L80-L87) as reference.

If you have any issues implementing my suggestions please let me know and I will assist with anything needed.

PR #32360, opened 2025-03-07
[bitnami/external-dns] Fix TXT AES key being read from incorrect secret

### LINKED ISSUE #27828: [bitnami/external-dns] Cannot use both `txtEncrypt.aesKey` and `cloudflare.secretName`
### Name and Version

bitnami/external-dns 8.0.2

None

Try to install the helm chart with the values I provided.

```yaml
provider: cloudflare
cloudflare:
  secretName: cloudflare-token
txtEncrypt:
  enabled: true
  aesKey: some-random-key
```

### What do you see instead?

Installation fails with the following error:

```
cloudflare.apiKey is required if cloudflare.apiToken is not provided
```

### Additional information

The `external-dns.createSecret` template in [`_helpers.tpl`](https://github.com/bitnami/charts/blob/main/bitnami/external-dns/templates/_helpers.tpl) forces the secret defined in [`secret.yaml`](https://github.com/bitnami/charts/blob/main/bitnami/external-dns/templates/secret.yaml) to be created if the `txtEncrypt.aesKey` is provided. The cloudflare block in the latter seems to assume that it will only be reached if `cloudflare.secretName` isn't set and thus either `cloudflare.apiKey` or `cloudflare.apiToken` need to be set.

luryus:
### Description of the change

Fix an issue where the external-dns chart references a wrong secret when trying to read the TXT record encryption AES key in deployment.yaml. When a provider-specific existing secret is configured (e.g. `google.serviceAccountSecret`), the deployment tried to read the AES key from _that_ secret. The correct secret to use here would be the default one created by this chart.

Fixed by changing the `external-dns.txtEncryptKeySecretName` template to directly reference the default secret, not the provider-specific existing secrets.

Additionally fixed some typos in the relevant comments in values.yaml.

### Benefits
Allows configuring both TXT encryption AES keys and existing secret for provider authentication.

### Possible drawbacks
While this is a pretty clear bug fix, it might be a breaking change for users that have explicitly worked around this (manually added the AES key as `txt_aes_encryption_key` in the existing provider secret).

### Applicable issues

- fixes #27828

### Checklist


juan131 (approved):
Thanks for the fix! LGTM

PR #32381, opened 2025-03-10
[bitnami/mongodb] #27482 - fix replicaset config handling

### LINKED ISSUE #27482: [bitnami/mongodb]  Infinite loop in replicaSetConfigurationSettings.sh
### Name and Version

bitnami/mongodb 15.1.15

amd64

Hi, when deploying mongodb using `replicaSetConfigurationSettings.enabled = true` the script [replicaSetConfigurationSettings.sh](https://github.com/bitnami/charts/blob/main/bitnami/mongodb/templates/replicaset/scripts-configmap.yaml#L212) never exits its while loop, and outputs logs forever. 

The custom configuration is correctly applied (confirmed using `rs.config()` in `mongosh`) so I am not sure why it never finishes. 

Looking at the script it seems `desiredEqualsactual` is always false so we never enter the if condition [here](https://github.com/bitnami/charts/blob/main/bitnami/mongodb/templates/replicaset/scripts-configmap.yaml#L276) so the loop never exits.

```yaml
mongodb:
  architecture: replicaset
  replicaCount: 3
  replicaSetConfigurationSettings:
    enabled: true
    configuration:
      members[0].priority: 1
      members[1].priority: 1
      members[2].priority: 1
```

[replicaSetConfigurationSettings.sh](https://github.com/bitnami/charts/blob/main/bitnami/mongodb/templates/replicaset/scripts-configmap.yaml#L212) should finish and stop printing logs.

### What do you see instead?

[replicaSetConfigurationSettings.sh](https://github.com/bitnami/charts/blob/main/bitnami/mongodb/templates/replicaset/scripts-configmap.yaml#L212) never exits and prints logs forever.

### Additional information

The logs
image

jkraml-staffbase:
### Description of the change

- Fix evaluation of `rs.conf().members[…]` properties to avoid endless loop in helper script

### Benefits

- It is now possible to configure per-member settings from the Helm chart's values (it already claims this is possible, which it was not)

### Possible drawbacks

- (not aware of any)

### Applicable issues

- fixes #27482

### Additional information

This uses `mongosh` as a JS interpreter to evaluate properties of the JSON object.

### Checklist


bitnami/mongodb/templates/replicaset/scripts-configmap.yaml
```diff
@@ -254,28 +254,34 @@ data:
       # read rs.conf again and store it. settings format is '"<key>" : <value>,'
       currentRsConf=$(mongosh ${usernameAndPassword} --eval 'rs.conf()')
 
-      desiredEqualsactual=unknown
+      desiredEqualsActual=unknown
```
jkraml-staffbase, line 257: typo fix

bitnami/mongodb/templates/replicaset/scripts-configmap.yaml
```diff
@@ -254,28 +254,34 @@ data:
...
       settingsToConfigure=""
       for key in ${!desiredRsConf[@]}; do
         value=${desiredRsConf[$key]}
-        if ! $(echo "\"${currentRsConf}"\" | grep -q -e "${key}: ${value},"); then
```
jkraml-staffbase, line 261: This was the problematic line: per-member settings would not appear in `currentRsConf` in a way that grep would find. The script thus looped endlessly, always trying to set the same setting.

bitnami/mongodb/templates/replicaset/scripts-configmap.yaml
```diff
@@ -254,28 +254,34 @@ data:
...
-           if [[ $key =~ ^members\[[0-9]+\]\..+ ]]; then
-            memberIndex=$(echo $key | grep -o -E '[0-9]+')
-            nodeConfigKey=${key#*.}
-            settingsToConfigure="${settingsToConfigure}cfg.members[${memberIndex}].${nodeConfigKey} = ${value}; "
+        if [[ $key =~ ^members\[[0-9]+\]\..+ ]]; then
+          # Replica set member specific setting
+          if "$(mongosh "cfg=${currentRsConf}; cfg.${key}" 2>/dev/null)" != "${value}"; then
```
fmulero, line 263: Sorry @jkraml-staffbase  I don't fully understand this line. Could you explain it? I see these errors when I try to execute it:
```console
MongoshInvalidInputError: [COMMON-10001] Invalid URI: cfg={....}
```

jkraml-staffbase: @fmulero sorry, I must've made a cop&paste mistake between what I tested in our cluster and what I committed. This line is missing `--eval`.

The idea is that we want to be able to evaluate the config which is given as e.g. `members[1].priority = 1`. The script recognizes `members[1].priority` as the key of a settings, but the old  grep variant would not be able to find this in the output (which was `{…,"members":[{"priority": 1, …}]}`.

We use `mongosh` as a javascript interpreter here.

The expression sets `cfg` to the previously retrieved value if `rs.conf()` and then tries to extract the desired value by replacing `cfg.${key}`, giving for example `cfg.members[0].priority`, which would be a valid javascript expression.

I'll push a fixed commit right away.

jkraml-staffbase: (sorry I didn't reply earlier, Firefox apparently does not refresh the tab on restart, so I didn't see your comment :laughing:)

fmulero: 👍

fmulero (approved):
Thanks a lot @jkraml-staffbase for your contribution.

LGTM

jkraml-staffbase:
@fmulero I'm very sorry, but I created a bug here. The `if`s are missing the proper brackets `[` `]` for evaluating the conditions. :facepalm: 

Unfortunately, this affects also the previously working general settings and thus makes the chart worse than it was.

Would you like me to provide a fix or to just revert for now?

jkraml-staffbase:
here's the fix: https://github.com/bitnami/charts/pull/32591

again, I'm very sorry for this stupid copy&paste error.

PR #32382, opened 2025-03-10
[bitnami/harbor] Revert merge chart generated env vart secret with existing secret

### LINKED ISSUE #25399: [harbor] Allow providing all secrets out-of-band

migruiz4:
This reverts commit 84fae4bdd3985037f1ccfe60156b0c3dbf760a3e.

### Description of the change

In August 2024 we merged this PR (#28989) which added a feature that merges the data from the provided secret values `<component>.existingEnvVarsSecret` with the secret generated by the chart.

This secret was not really used anywhere, since the PR did not update the secret selection on the deployments, meaning that the 'merged' secret was generated but never used. See

- Core: https://github.com/bitnami/charts/blob/54531ce059d1156a615749485e11e2f30143ea16/bitnami/harbor/templates/core/core-dpl.yaml#L172-L177
- Registry: https://github.com/bitnami/charts/blob/54531ce059d1156a615749485e11e2f30143ea16/bitnami/harbor/templates/registry/registry-dpl.yaml#L180-L185
- Trivy: https://github.com/bitnami/charts/blob/54531ce059d1156a615749485e11e2f30143ea16/bitnami/harbor/templates/trivy/trivy-sts.yaml#L149-L154
- Job Service: https://github.com/bitnami/charts/blob/54531ce059d1156a615749485e11e2f30143ea16/bitnami/harbor/templates/jobservice/jobservice-dpl.yaml#L186-L191

Recently, some users complained about the chart secret being generated although they were providing their own existing secrets which led us to noticing this issue.

This PR reverts the merge logic, simplifying the chart templates, since we think users haven't really being taking any advantages of this:

- Those providing existingSecrets were using the secret they provided, but the chart was generating the additional secret which was never used.
- Those who did not provide an existingSecret were using the default one, since there were no values to merge.

### Benefits

Simplifies chart logic and removes unused feature.

### Possible drawbacks

None known. 

### Applicable issues

- fixes #25399
- closes #31079
- reverts #28989

### Checklist


bitnami/harbor/templates/core/core-secret-envvars.yaml
```diff
@@ -3,29 +3,6 @@ Copyright Broadcom, Inc. All Rights Reserved.
 SPDX-License-Identifier: APACHE-2.0
 */}}
 
-{{- $secretName := .Values.core.existingEnvVarsSecret -}}
```
fmulero, line 6: I think you should replace the code here with this condition:
```
{{- if not .Values.core.existingEnvVarsSecret }}
```

fmulero (changes requested):
Nice catch. Thanks a lot Migue!

I've just left a comment, I think you missed a condition in core secret

fmulero (dismissed):
LGTM

fmulero (approved):

PR #32394, opened 2025-03-11
[bitnami/mongodb] Update dns-check init container condition

### LINKED ISSUE #30798: [bitnami/mongodb] dnsCheck needs $MY_POD_NAMESPACE to be set
### Name and Version

bitnami/mongodb: 16.3.2

arm64

If you look into the definition of the dnsCheck, you find this:

```yaml
- name: dns-check
  image: {{ include "mongodb.externalAccess.dnsCheck.image" . }}
  imagePullPolicy: {{ .Values.externalAccess.dnsCheck.image.pullPolicy | quote }}
  command:
    - /bin/bash
  args:
    - -ec
    - |
      # MONGODB_INITIAL_PRIMARY_HOST should be resolvable
      while ! (getent ahosts "{{ include "mongodb.initialPrimaryHost" . }}" | grep STREAM); do
        sleep 10
      done
  securityContext: {{- include "common.compatibility.renderSecurityContext" (dict "secContext" .Values.containerSecurityContext "context" $) | nindent 12 }}
  {{- end }}
  {{- if .Values.externalAccess.dnsCheck.resources }}
  resources: {{- toYaml .Values.externalAccess.dnsCheck.resources | nindent 12 }}
  {{- else if ne .Values.externalAccess.dnsCheck.resourcesPreset "none" }}
  resources: {{- include "common.resources.preset" (dict "type" .Values.externalAccess.dnsCheck.resourcesPreset) | nindent 12 }}
  {{- end }}
{{- end -}}
```

It includes `mongodb.initialPrimaryHost`, which itself includes the namespace as `$(MY_POD_NAMESPACE)` instead of helm variable. But the environment variable MY_POD_NAMESPACE is not set in the context of the dnsCheck. So externalAccess is unable to work.

### What do you see instead?

dnsCheck failing with MY_POD_NAMESPACE: command not found

migruiz4:
### Description of the change

Updates the `dns-check` init container conditional to not include `.Values.externalAccess.service.domain`.

The value `externalAccess.service.domain` didn't interfere in any way with the `dns-check` init container, which led us to two possible scenarios:

- When `externalAccess.service.publicNames` is set, the dns-check will ensure that indeed the pod is able to resolve the DNS.
- When `externalAccess.service.publicNames` is **NOT** set, but `externalAccess.service.domain` is, the `dns-check` will perform a check on the pod's FQDN, e.g. `mongodb-0.svc.mongodb-hl.my-ns.cluster.local`.

This second scenario doesn't really make sense, because the headless service is configured with `publishNotReadyAddresses=true`, and performing a DNS check on a headless service is unnecessary.

### Benefits

Simplifies the logic and fixes an existing issue with the `dns-check` init container, where check would never pass because variable `MY_POD_NAMESPACE` was never defined.

### Possible drawbacks

None known.

### Applicable issues

- fixes #30798

### Checklist

### DIFF
--- a/bitnami/mongodb/CHANGELOG.md
+++ b/bitnami/mongodb/CHANGELOG.md
@@ -1,8 +1,12 @@
 # Changelog
 
-## 16.4.6 (2025-02-27)
+## 16.4.7 (2025-03-11)
 
-* [bitnami/mongodb] Use actual hostname instead of localhost for mongodb_exporter URI string ([#32192](https://github.com/bitnami/charts/pull/32192))
+* [bitnami/mongodb] Update dns-check init container condition ([#32394](https://github.com/bitnami/charts/pull/32394))
+
+## <small>16.4.6 (2025-03-10)</small>
+
+* [bitnami/mongodb] Use actual hostname instead of localhost for mongodb_exporter URI string (#32192) ([cbfec4f](https://github.com/bitnami/charts/commit/cbfec4f962119b505b5c81b6d61448506b7a6216)), closes [#32192](https://github.com/bitnami/charts/issues/32192)
 
 ## <small>16.4.5 (2025-02-21)</small>
 
--- a/bitnami/mongodb/Chart.yaml
+++ b/bitnami/mongodb/Chart.yaml
@@ -39,4 +39,4 @@ maintainers:
 name: mongodb
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/mongodb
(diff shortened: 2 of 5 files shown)


jotamartos (approved):
The PR LGTM but if we remove usage from .Values.externalAccess.service.domain here, it's almost unused in the chart. I think we can change the logic to use .Values.externalAccess.service.publicNames and remove the other parameter.

PR #32401, opened 2025-03-11
[bitnami/grafana] Add `optional` option to the extraEnvVars

yardenshoham:
For a use-case where the secret might not exist but still we would like grafana to run.

### Description of the change

We add the option to set the `optional` field in `extraEnvVarsCM` and `extraEnvVarsSecret`.

### Benefits

When `extraEnvVarsSecret` references optional environment variables, it's OK for Grafana to run without them. We don't want to block it from running.

### Possible drawbacks

None. The old defaults are kept.

### Additional information

It's not a breaking change.

### Checklist


jotamartos (approved):
Thank you for your contribution.

PR #32409, opened 2025-03-11
[bitnami/*] Add tanzuCategory annotation

carrodher:
This PR adds a new annotation named `tanzuCategory` whose possible values are `service`, `application`, or `clusterUtility`.

_PS: don't worry about the failing action; we can manually merge PRs like this one, modifying more than one chart at a time PR_

### DIFF
--- a/bitnami/common/Chart.yaml
+++ b/bitnami/common/Chart.yaml
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: APACHE-2.0
 
 annotations:
+  tanzuCategory: service
   category: Infrastructure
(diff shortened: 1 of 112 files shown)


bitnami/common/Chart.yaml
```diff
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: APACHE-2.0
 
 annotations:
+  tanzuCategory: service
```
mpermar, line 5: This one is an exception. I don't think we want or need a category here.

mpermar (approved):

PR #32520, opened 2025-03-19
[bitnami/valkey-cluster] Add support for templating nodes, replicas, resources, and persistent storage values

mbrancato:
### Description of the change

This PR adds support to template the values input for the nodes, replicas, resources, and persistent storage of a cluster.

### Benefits

Currently, this helm chart does not allow configuration by shard. This PR allows for abstraction when used as a subchart, or when the user just wants to configure things differently.

For example, this abstraction becomes possible:

```yaml
global:
  cache:
    shards: 18
    replicas: 3

cluster:
  nodes: "{{ .Values.global.cache.shards | int | mul .Values.global.cache.replicas | int }}"
  replicas: "{{ .Values.global.cache.replicas }}"
```

### Possible drawbacks

This should be compatible with all existing usage.

### Checklist

### DIFF
--- a/bitnami/valkey-cluster/templates/valkey-statefulset.yaml
+++ b/bitnami/valkey-cluster/templates/valkey-statefulset.yaml
@@ -20,7 +20,7 @@ spec:
   {{- $podLabels := include "common.tplvalues.merge" ( dict "values" ( list .Values.valkey.podLabels .Values.commonLabels ) "context" . ) }}
   selector:
     matchLabels: {{- include "common.labels.matchLabels" ( dict "customLabels" $podLabels "context" $ ) | nindent 6 }}
-  replicas: {{ .Values.cluster.nodes }}
+  replicas: {{ if eq (typeOf .Values.cluster.nodes) "float64" -}}{{ .Values.cluster.nodes }}{{- else -}}{{ tpl .Values.cluster.nodes . | int }}{{- end }}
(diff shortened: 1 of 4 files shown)


bitnami/valkey-cluster/templates/valkey-statefulset.yaml
```diff
@@ -20,7 +20,7 @@ spec:
   {{- $podLabels := include "common.tplvalues.merge" ( dict "values" ( list .Values.valkey.podLabels .Values.commonLabels ) "context" . ) }}
   selector:
     matchLabels: {{- include "common.labels.matchLabels" ( dict "customLabels" $podLabels "context" $ ) | nindent 6 }}
-  replicas: {{ .Values.cluster.nodes }}
+  replicas: {{ if eq (typeOf .Values.cluster.nodes) "float64" -}}{{ .Values.cluster.nodes }}{{- else -}}{{ tpl .Values.cluster.nodes . | int }}{{- end }}
```
migruiz4, line 23: I'm sorry but I'm not understanding, why the `typeOf` is required? Can't this be simplified like this?

```suggestion
replicas: {{ tpl .Values.cluster.nodes . | int }}
```

For those values that need to be quoted, like environment variables values: {{ tpl .Values.cluster.nodes $ | int | quote }}

migruiz4 (commented):
Hi @mbrancato,

Thank you very much for your contribution! Could you please take a look at my comment?

Issue #32541, opened 2025-03-20
[bitnami/nginx] Feature Request: Allow disabling of automatic initContainers in NGINX chart


vineetguptadev:
### Name and Version

bitnami/nginx 19.0.2

### What is the problem this feature will solve?

Description:
## Current Behavior
The NGINX chart automatically creates initContainers under certain conditions:

1. When `containerSecurityContext.readOnlyRootFilesystem` is true, it creates a `preserve-logs-symlinks` initContainer that requires bash and performs filesystem operations
2. The chart has no way to disable these automatically created initContainers

## Issues
This causes problems when:
1. Running in environments with strict OPA policies that don't allow readOnlyRootFilesystem
2. Using custom container images that don't include `bash` and only `sh`
3. Needing to completely disable the log preservation mechanism

## Desired Behavior
Add a new configuration option like `disableDefaultInitContainers: false` (default) that when set to `true` would skip the automatically generated initContainers while still allowing custom ones via `.Values.initContainers`.

## Suggested Implementation
Modify the conditional statements in deployment.yaml:

```yaml
{{- if and (not .Values.disableDefaultInitContainers) (or .Values.containerSecurityContext.readOnlyRootFilesystem .Values.cloneStaticSiteFromGit.enabled .Values.initContainers) }}
initContainers:
{{- if and (not .Values.disableDefaultInitContainers) .Values.containerSecurityContext.readOnlyRootFilesystem }}
  - name: preserve-logs-symlinks
    ...
{{- end }}

### What is the feature you are proposing to solve the problem?

A configuration flag disableDefaultInitContainers: true|false (defaulting to false) that allows users to completely skip the automatically created initContainers in the NGINX chart, particularly the preserve-logs-symlinks initContainer that's generated when containerSecurityContext.readOnlyRootFilesystem is true. This would allow users with strict OPA policies or custom images without bash to still use the chart without having to fork it.

sleeyax:
Stale bots are the worst. This is still an issue!

PR #32577, opened 2025-03-23
[bitnami/nginx] feat(nginx-optional-initcontainers): Enable optional initcontainers

vineetguptadev:
### Description of the change

This change introduces a simple but effective way to completely disable all initContainers in the Bitnami NGINX Helm chart.

- Added a new configuration parameter:

Introduced initContainersDisabled flag in the values file
When set to true, it prevents all initContainers from being created
When not specified or set to false, existing behavior is preserved

- Modified conditional logic:

Changed the template condition that controls initContainers creation
Added an explicit check for not .Values.initContainersDisabled that takes precedence
Preserved all existing conditional logic for backward compatibility

- Non-breaking change:

Default behavior remains exactly the same
Users must explicitly opt-in to disable initContainers
No changes to any other parts of the template

### Addresses the specific issue:

Provides a solution to GitHub issue #32541
Gives users a clean way to completely disable initContainers without having to modify multiple unrelated settings
Simplifies deployments where initContainers are not needed or desired

### Benefits

- Simplified Configuration: Provides a single, clear parameter to disable all initContainers rather than requiring users to configure multiple unrelated settings.
- Reduced Resource Usage: Eliminates unnecessary init container executions, reducing resource consumption and speeding up pod startup time.
- Enhanced Control: Gives users explicit control over initContainer behavior without modifying core template logic.
- Better Compatibility with Security Environments: Some security-restricted environments prohibit or limit init containers - this change allows deployment in such environments.
- Improved Debugging: Makes it easier to troubleshoot deployment issues by allowing users to completely bypass initContainer logic.
- Cleaner Deployment Manifests: Results in cleaner, simpler Kubernetes manifests when initContainers are not needed.
- Preserves Existing Functionality: Maintains backward compatibility while adding new functionality, ensuring existing deployments continue to work.
- Reduces Fork Maintenance: Users can use the official chart with just a values override instead of maintaining a forked version with custom template modifications.
- Better Integration with CI/CD: Simplifies automated deployments by providing a clear flag to control initContainer behavior.
- Aligned with Kubernetes Best Practices: Follows the principle of providing explicit configuration options for deployment behaviors.

### Checklist


vineetguptadev:
@carrodher @javsalgar Updated the PR with the good PR now. Hope to et this reviewed ! Sincerely appreciate the guidance and help !

javsalgar:
Hi!

Thank you so much for the PR! In other charts, we call the value `enableDefaultInitContainers`. For consistency, could you rename it?

vineetguptadev:
> Hi!
> 
> Thank you so much for the PR! In other charts, we call the value `enableDefaultInitContainers`. For consistency, could you rename it?

Good Call @Absolutely, Just pushed the same.

vineetguptadev:
@migruiz4 May I request you to review this , appreciate the help.

bitnami/nginx/values.yaml
```diff
@@ -103,6 +103,10 @@ image:
   ## Set to true if you would like to see extra information on logs
   ##
   debug: false
+## @param enableDefaultInitContainers - boolean - optional
+## Disable all init containers regardless of other settings
+##
+enableDefaultInitContainers: false
```
migruiz4, line 109: Please update the default value to `true`.

bitnami/nginx/values.yaml
```diff
@@ -103,6 +103,10 @@ image:
   ## Set to true if you would like to see extra information on logs
   ##
   debug: false
+## @param enableDefaultInitContainers - boolean - optional
+## Disable all init containers regardless of other settings
```
migruiz4, line 107: The metadata structure is wrong
```suggestion
## @param enableDefaultInitContainers If set to false, disable all init containers except user-defined at `initContainer`.
```

migruiz4 (changes requested):
Hi @vineetguptadev,

Thank you for your contribution and I'm sorry for the late response.

There is an issue with this contribution, I think the value `enableDefaultInitContainers` should not skip the user-defined custom init containers (`.Values.initcontainers`).

My suggestion:
```
{{- if or .Values.initContainers (and .Values.enableDefaultInitContainers (or .Values.containerSecurityContext.readOnlyRootFilesystem .Values.cloneStaticSiteFromGit.enabled)) }}
initContainers:
{{- if .Values.enableDefaultInitContainers }}
// Default initContainers 
...
{{- end }}
{{- end }}
{{- end }}
```

Additionally, please set the value `enableDefaultInitContainers` to true, since the current naming is counter-intuitive:
- enable=false - Init containers enabled
- enable=true - Init containers disabled

bitnami/nginx/values.yaml
vineetguptadev: @migruiz4  thanks, done.

vineetguptadev:
@migruiz4  Can you please review it now if good to land ?

migruiz4 (approved):
Looks good to me! Thank you for your contribution!

PR #32588, opened 2025-03-25
Update README.md

Chagui05:
### Description of the change

In the updating password section it gave the architecture setting a value of "replication", instead of a "replicaset"->

### Benefits

It makes the instructions clearer, some may run into an error trying to put the  architecture as "replication"
### Additional information

It really is not a functional change, it was probably just a typo from someone


carrodher (approved):
LGTM, thanks

PR #32594, opened 2025-03-25
[bitnami/keycloak] Set `usePasswordFiles=true` by default

migruiz4:
### Description of the change

### Benefits

### Possible drawbacks

None

### Applicable issues

None

### Additional information

N/A

### Checklist


juan131 (changes requested):
Thanks so much for the PR @migruiz4 !! Please remember to bump the chart version and update the README

juan131 (approved):
LGTM

juan131 (approved):

olav-st:
@migruiz4 Does this actually work? 

I had to revert to `usePasswordFiles=false` because the admin password was not being set. I can't find  the environment variable that you are using (`KC_BOOTSTRAP_ADMIN_PASSWORD_FILE`) anywhere in the Keycloak documentation :thinking:

juan131:
Hi @olav-st 

What image version are you using? As you can see in the links below any env. variable (including `KC_BOOTSTRAP_ADMIN_PASSWORD`) listed there which is suffixed with `_FILE` would be supported:

- https://github.com/bitnami/containers/blob/main/bitnami/keycloak/26/debian-12/rootfs/opt/bitnami/scripts/keycloak-env.sh#L75
- https://github.com/bitnami/containers/blob/main/bitnami/keycloak/26/debian-12/rootfs/opt/bitnami/scripts/keycloak-env.sh#L97-L107

olav-st:
@juan131 I am using `docker.io/bitnami/keycloak:26.1.5-debian-12-r0` (I haven't overridden the image so it should be the default one)

But I think I found out what the problem is. I am using the following values to set the admin password from an existing secret:
```yaml
auth:
  adminUser: admin
  existingSecret: keycloak-admin-password
  passwordSecretKey: password
```

This does not seem to work when `usePasswordFiles=true` . I get a random admin password instead of the password from the secret

juan131:
I think I found the issue @olav-st 

The default database password is `password` when enabling PostgreSQL, see:

- https://github.com/bitnami/charts/blob/main/bitnami/keycloak/templates/_helpers.tpl#L180

Given the key you're using in your existing secret is also `password` it gets overwritten when it's mounted at the same path:

- https://github.com/bitnami/charts/blob/main/bitnami/keycloak/templates/statefulset.yaml#L159
- https://github.com/bitnami/charts/blob/main/bitnami/keycloak/templates/statefulset.yaml#L161

Please confirm me if that's the case and I'll send a PR to avoid this collision.

olav-st:
@juan131 Thanks for looking into it! 

I can confirm that if I change the key in the secret to something other than `password` then everything works. So it is indeed a name collision.

juan131:
Great! I just created a PR to address this

PR #32661, opened 2025-03-28
bitnami/external-dns: do not set EXTERNAL_DNS_RFC2136_TSIG_SECRET when rfc2136 provider is not used

paulfantom:
### Description of the change

This change will prevent setting EXTERNAL_DNS_RFC2136_TSIG_SECRET variable from a secret when `.Values.provider` is not set to `rfc2136`.

### Benefits

Prevent accidental misconfiguration. Especially when using chart as a downstream library.

### Possible drawbacks

none

### Applicable issues

- fixes #

### Additional information

I have a use case where I am using this helm chart as a dependency and allow for configuring two providers. To allow for easy debugging, I am preconfiguring this helm chart and setting `rfc2136.secretName` (secret is generated via other means and name is static). However when I want to use other provider (via `.Values.provider`), the `EXTERNAL_DNS_RFC2136_TSIG_SECRET` is still included in the Deployment object, which I would consider a bug.

### Checklist

### DIFF
--- a/bitnami/external-dns/Chart.yaml
+++ b/bitnami/external-dns/Chart.yaml
@@ -29,4 +29,4 @@ maintainers:
 name: external-dns
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/external-dns
-version: 8.7.10
+version: 8.7.11
--- a/bitnami/external-dns/templates/deployment.yaml
+++ b/bitnami/external-dns/templates/deployment.yaml
@@ -635,6 +635,7 @@ spec:
                   key: infoblox_wapi_password
             {{- end }}
             {{- end }}
+            {{- if eq .Values.provider "rfc2136" }}
             {{- if .Values.rfc2136.tsigSecret | or (and .Values.rfc2136.kerberosUsername .Values.rfc2136.kerberosPassword) | or .Values.rfc2136.secretName }}
             # RFC 2136 environment variables
             {{- if .Values.rfc2136.rfc3645Enabled }}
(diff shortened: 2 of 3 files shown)


bitnami/external-dns/templates/deployment.yaml
```diff
@@ -635,6 +635,7 @@ spec:
                   key: infoblox_wapi_password
             {{- end }}
             {{- end }}
+            {{- if eq .Values.provider "rfc2136" }}
             {{- if .Values.rfc2136.tsigSecret | or (and .Values.rfc2136.kerberosUsername .Values.rfc2136.kerberosPassword) | or .Values.rfc2136.secretName }}
```
jotamartos, line 639: I'd include the new condition inside of the existing one 

```suggestion
            {{- if and ( eq .Values.provider "rfc2136" ) ( .Values.rfc2136.tsigSecret | or (and .Values.rfc2136.kerberosUsername .Values.rfc2136.kerberosPassword) | or .Values.rfc2136.secretName ) }}
```

bitnami/external-dns/Chart.yaml
```diff
@@ -29,4 +29,4 @@ maintainers:
 name: external-dns
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/external-dns
-version: 8.7.8
+version: 8.7.9
```
jotamartos, line 32: Please bump the chart's version again as this version is already published

jotamartos (changes requested):

bitnami/external-dns/templates/deployment.yaml
jotamartos: You are completely right. In that case, to follow the same syntax, could you please revert the change so I accept it and merge the changes? Sorry for the noise

jotamartos (approved):

PR #32675, opened 2025-03-28
[bitnami/common] Add global.security.allowEmptyPassword

maxime1907:
### Description of the change

Add an option `global.security.allowEmptyPassword` to bypass empty password verification checks when using external vaults/secrets that manages this part

### Benefits

Creating via `.Values.extraDeploy` an ExternalSecret resource inside the bitnami release when upgrading from values based passwords to external vault secret based passwords

### Possible drawbacks

None as it keeps backward compatibility if the field is not present

### Applicable issues

None

### Additional information

None

### Checklist

### DIFF
--- a/bitnami/common/CHANGELOG.md
+++ b/bitnami/common/CHANGELOG.md
@@ -2,11 +2,13 @@
 
 ## 2.31.0 (2025-03-28)
 
-* [bitnami/common] Add global.security.allowEmptyPassword ([#32669](https://github.com/bitnami/charts/pull/32669))
+* [bitnami/common] Add global.security.allowEmptyPassword ([#32675](https://github.com/bitnami/charts/pull/32675))
 
 ## 2.30.0 (2025-02-19)
 
-* [bitnami/common] Add helper to check API versions ([#31969](https://github.com/bitnami/charts/pull/31969))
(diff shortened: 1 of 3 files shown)


maxime1907:
Any updates on this? @javsalgar

migruiz4:
Hi @maxime1907, 
Could you please provide an usage example where this feature is used? Maybe a branch in your fork where we can see a real use case for this feature.

The helper `common.errors.upgrade.passwords.empty` is usually used alongside other helpers such as `common.validations.values.mariadb.passwords`. 
It had a key role in the past before we implemented `common.secrets.passwords.manage`, since we previously tried to avoid the usage of the Helm `lookup` function.

For context, during upgrades, the helper `common.secrets.passwords.manage` will look for the existing (built-in) chart secret to recreate it without having to provide the passwords in the values, while during first installations it will use a randomly generated password unless the user provided one in the values.

I see two main inconveniences with this contribution:
- Calling the value `allowEmptyPassword` would be inaccurate, because the charts won't be deployed with empty passwords when set.
- What this PR does is skip the upgrade password check, but it does not prevent the chart from generating its built-in secret which could end up in different scenarios depending on the chart template:
  - The built-in Secret is recreated with a new random password, so the bootstrap password could be lost.
  - The deployment fails because the secret expects a password value that was not provided.

Additionally, in the benefits you mention: 
> Creating via .Values.extraDeploy an ExternalSecret resource inside the bitnami release when upgrading from values based passwords to external vault secret based passwords`
Isn't that already possible by using the `existingSecret` value?

PR #32686, opened 2025-03-31
[bitnami/opensearch] Update nodes heapSize to half the memory preset

migruiz4:
### Description of the change

By default, we are deploying the Opensearch nodes with `resourcesPreset: small` (except for 'data' nodes).

The `small` preset sets the following resources:

```
    Limits:
      cpu:                750m
      ephemeral-storage:  2Gi
      memory:             768Mi
    Requests:
      cpu:                500m
      ephemeral-storage:  50Mi
      memory:             512Mi
```

But we were setting the HEAP_SIZE to 128m.

Having heap size so low may sometimes cause issues during deployment, so this PR increases it to 256m, which actually fits the recommendation of using half the memory assigned.

### Benefits

Prevents issue while deploying, where some plugins will fail to install because of memory heap being too low.

### Possible drawbacks

None known. 

This PR does not increase the pod memory assignment, it only redistributes the Opensearch / Lucene memory assignment to 1/2 each.

### Checklist


fmulero (approved):
LGTM

PR #32814, opened 2025-04-04
[bitnami/etcd] Simplify condition

jachym-tousek-keboola:
I noticed a condition which is redundant so I'm simplifying it. It's also possible the condition is incorrect.


javsalgar:
Hi!

Thank you so much for the PR! Could you bump the chart patch version?

jachym-tousek-keboola:
Like this? :)

Also I'm unsure... should the condition be

a. just disasterRecovery.enabled (this is equivalent to the existing code)
```
{{- if .Values.disasterRecovery.enabled }}
```

or was the intention for it to be

b. if either disasterRecovery.enabled or startFromSnapshot.enabled?
```
{{- if or .Values.disasterRecovery.enabled .Values.startFromSnapshot.enabled }}
```

Checking the code b. would make more sense to me but the existing code is equivalent to a.

javsalgar:
About your question, the reviewers will check it (I'm doing the triage). It seems there's an issue with the DCO, could you check it?

jachym-tousek-keboola:
fixed

PR #32827, opened 2025-04-06
[bitnami/rabbitmq] Update README.md

thecatontheflat:
Fixes outdated link


migruiz4:
Hi @thecatontheflat, thank you for reporting this issue!

In this case, the link in the bitnami/rabbitmq README is correct, but I think we accidentally removed the parameters section in the bitnami/common README.

I think the correct fix here would be to restore the common README instead of patching the affected links.

migruiz4:
I have created this PR restoring the missing section: #32861

thecatontheflat:
Thanks! I am closing this PR then

Issue #32684, opened 2025-03-31
[bitnami/flink] Flink global.defaultStorageClass is not working


wklee610:
### Name and Version

bitnami/flink 1.4.5

### What architecture are you using?

None

### What steps will reproduce the bug?

If only specify defaultStorageClass in an existing Helm chart and run "helm upgrade --install", PV and PVC won't be created.

### Are you using any custom parameters or values?

global.defaultStorageClass

### What is the expected behavior?

Create PV, PVC

### What do you see instead?

nothing

javsalgar:
Hi,

Does the issue happen in a fresh instance? Or does it only happen when upgrading?

wklee610:
It happens in a fresh instance.

fmulero:
Hi @wklee610 thanks for using bitnami/charts.

Sorry but I am a little bit confused, Flink chart doesn't have any PV, PVC or PersistentVolumeClaimTemplate in its code. Do you think they are missing? Would you like to contribute by creating a PR adding them? The Bitnami team will be happy to review it and provide feedback. [Here](https://github.com/bitnami/charts/blob/main/CONTRIBUTING.md) you can find the contributing guidelines.

wklee610:
First, I'm sorry for confusing you. I thought it was supported because there was defaultStorageClass in this code.
After looking into it, it seems that PV and PVC are not provided by default in Flink.
Normally, people use it in a way like S3 or HDFS or upload a JAR file in the Flink Web UI but in my case, I kept running "kubectl cp" while uploading a JAR file on k8s for testing purposes, so, it seemed more convenient to have PV/PVC in Flink (If I was using it correctly).

Of course, I would like to contribute by creating a PR (Adding volumeClaimTemplates or delete defaultStorageClass in global).

I don't know whether I should delete it or add it, so if you make any decision, I'll try to make a PR.

Thank you so much :)

wklee610:
Hello, @fmulero 
I've submitted a PR for adding PVC support based on your suggestion. I’d appreciate it if you could take a look when you have time. Thanks a lot!

https://github.com/bitnami/charts/pull/32828

PR #32828, opened 2025-04-06
[bitnami/flink] Add support for PersistentVolume and PersistentVolumeClaim

wklee610:
### Description of the change

This PR adds support for defining `PersistentVolumeClaim`s in the Flink chart.  
It introduces configurable persistence settings for both the `JobManager` and `TaskManager` components.

The chart did not previously support persistence, and this change is based on feedback suggesting it would be a useful addition.

### Benefits

- Enables users to configure persistent volumes for Flink workloads and use.

### Possible drawbacks

- None.

### Applicable issues

- None.

### Additional information

- New fields were added to bottom of `values.yaml`. (Persistence)
- Add `pvc.yaml` in templates/jobmanager
- Change deployment.yaml and statefulset.yaml in Charts
- Change Persistence section in README.md 

### Checklist


bitnami/flink/templates/jobmanager/pvc.yaml
```diff
@@ -0,0 +1,25 @@
...
+  {{- if .Values.commonAnnotations }}
+  annotations: {{- include "common.tplvalues.render" ( dict "value" .Values.commonAnnotations "context" $ ) | nindent 4 }}
+  {{- end }}
+spec:
+  accessModes:
+    - {{ .Values.persistence.accessMode | quote }}
+  resources:
+    requests:
+      storage: {{ .Values.persistence.size | quote }}
+  storageClassName: {{ .Values.global.defaultStorageClass | quote }}
```
juan131, line 24: You should use our `common.storage.class` helper instead to set the storage class name:
```suggestion
  {{- include "common.storage.class" (dict "persistence" .Values.persistence "global" .Values.global) | nindent 2 }}
```

bitnami/flink/templates/jobmanager/pvc.yaml
```diff
@@ -0,0 +1,25 @@
...
+  namespace: {{ include "common.names.namespace" . | quote }}
+  labels: {{- include "common.labels.standard" ( dict "customLabels" .Values.commonLabels "context" $ ) | nindent 4 }}
+    app.kubernetes.io/part-of: flink
+    app.kubernetes.io/component: jobmanager
+  {{- if .Values.commonAnnotations }}
+  annotations: {{- include "common.tplvalues.render" ( dict "value" .Values.commonAnnotations "context" $ ) | nindent 4 }}
+  {{- end }}
+spec:
+  accessModes:
+    - {{ .Values.persistence.accessMode | quote }}
```
juan131, line 20: `persistence.accessMode` parameter should be renamed to `persistence.accessModes` and act as an array:
```suggestion
  {{- if not (empty .Values.persistence.accessModes) }}
  accessModes:
  {{- range .Values.persistence.accessModes }}
    - {{ . | quote }}
  {{- end }}
  {{- end }}
```

juan131 (changes requested):
Thanks so much for the contribution! As you mentioned, you should split the parameter so they can be independently set for each component. That said, could you please provide info about the use case you want to cover? Why is it necessary to store data on these components?

wklee610:
It is common practice to manage JAR files in Flink by integrating with external storage systems such as S3 or HDFS. https://github.com/bitnami/charts/issues/32684 However, there are cases in production environments where using external storage is not feasible due to security or network constraints. In such scenarios, uploading the JAR files directly to a Persistent Volume (PV) during the CI/CD pipeline can be a more suitable approach in terms of operations and management. This is why I am suggesting this method.

juan131:
Please correct me if I'm wrong but, in your use case, you don't use these new "data" volumes to store data generated by Flink components. Instead, you use them to "upload" JARs to be consumed by them, am I right?

wklee610:
Yes, you are right.

juan131:
Hi @wklee610 

Thanks for confirming so!

In this case, we should take a different approach given Bitnami charts use `persistence.*` parameters to store data generated by the different apps & servers we have in the catalog.

When we want to support ways to make it easier for developers to upload custom JARs (or other kind of custom objects) to be consumed, we explore other alternatives. For instance, check this examples:

- Airflow support loading custom DAGs by mounting them from a ConfigMap or cloning it from a GIT repository: https://github.com/bitnami/charts/tree/main/bitnami/airflow#load-dag-files
- ASP.NET Core supports mounting the custom app via GIT repository and also from an existing PVC: https://github.com/bitnami/charts/tree/main/bitnami/aspnet-core#deploying-your-custom-aspnet-core-application

juan131:
Hi @wklee610 

Please note Bitnami charts are developed and tested using Bitnami images. Using them in combination with custom images is possible given these charts are very flexible and can be adapted to custom setups. This is the reason why we have parameters such as `extraEnvVars`, `sidecars`, `initContainers`, `extraVolumes` or `extraVolumeMounts`, so users can add any custom env. variable, container, volume, etc. to the default pod structure.

However, it's not suitable, in terms of maintenance, to support specific parameters that are designed for use cases not supported by Bitnami images.

PR #32846, opened 2025-04-07
[bitnami/keycloak] fix to enable empty hostname

JoKrefting:
### Description of the change

- this fixes `hostname` no allowed to be empty within ingress.yaml

### Benefits

It will fully enable an empty hostname without exception within ingress.yaml to be able to disable `KEYCLOAK_HOSTNAME` property.

### Possible drawbacks

None.

### Applicable issues

- None known

### Checklist

### DIFF
--- a/bitnami/keycloak/CHANGELOG.md
+++ b/bitnami/keycloak/CHANGELOG.md
@@ -1,8 +1,12 @@
 # Changelog
 
-## 24.5.1 (2025-04-07)
+## 24.5.2 (2025-04-07)
 
-* [bitnami/keycloak] The secret name of the externaldb is inherited by the .Release.Name and not by the (include "common.names.fullname" .) function. ([#32823](https://github.com/bitnami/charts/pull/32823))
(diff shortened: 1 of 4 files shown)


JoKrefting:
@migruiz4 you are right. I didn't notice that tls was true in my setup and hence just noticed that the ingress controller crashed due to my empty hostname definition. 
So the PR is more optional i believe as the condition itself makes still sense to avoid that others might run into same issue that i faced. But i'm fine as well if you prefer just closing it.

migruiz4:
I'm sorry @JoKrefting I appreciate your time and effort, but I will dismiss this contribution and close this PR.

As I previously mentioned, the default ingress `tls` entry is kind of a especial case, where users set `ingress.tls=true` + `cert-manager/ingress.secrets/ingress.selfSigned` and a route for the ingress.hostname will be set.

The reason why I can not accept this contribution is because, by accepting it there could be a case where `ingress.hostname` and `ingress.extraTls` are empty and `ingress.tls` is enabled. That would result in the rendering of a wrong YAML:

```
...
tls:
  // Nothing here
```

PR #32852, opened 2025-04-07
[bitnami/kubernetes-event-exporter] Add clusterName config setting to values

ChrisMcD1:
### Description of the change

There is a configuration setting for the kubernetes cluster name, which all users in a multi-cluster environment should be setting. The current behavior of the chart is to throw the entire `config` value into the ConfigMap, so no changes are needed inside of the actual template for this to get the desired effect.

Here is where the configuration is consumed. https://github.com/resmoio/kubernetes-event-exporter/blob/907ce911f7ea8b347aa1a469aaac9c5b56e78dec/pkg/exporter/config.go#L28

### Benefits

Makes it easier for users to set up a useful kubernetes-event-exporter instance.

### Possible drawbacks

Under this same standard, we could end up adding _every_ configuration key to the chart, which I don't know would be entirely helpful. This key is so standard though, that I don't feel that drawback is critical.

### Applicable issues

Didn't create an issue for it.

### Checklist


jotamartos (approved):
LGTM! Thank you for your contribution.

PR #32857, opened 2025-04-07
[bitnami/postgresql-ha] PGPOOL_POSTGRES_PASSWORD mapping from postgres-password

### LINKED ISSUE #32807: [bitnami/postgresql-ha] pgpool is mapping PGPOOL_POSTGRES_PASSWORD env. var. from wrong secretKeyRef
### Name and Version

bitnami/postgresql-ha-15.3.8

None

Deploy HelmChart using different passwords (values) defined in secret "my-db-secret", and use this secret file in "postgresql.existingSecret: my-db-secret".

postgresql-ha-15.3.8.tgz/postgresql-ha/templates/pgpool/deployment.yaml

```yaml
            {{- if .Values.postgresql.usePasswordFiles }}
            - name: PGPOOL_POSTGRES_PASSWORD_FILE
              value: "/opt/bitnami/pgpool/secrets/pgpool-password"
            {{- else }}
            - name: PGPOOL_POSTGRES_PASSWORD
              valueFrom:
                secretKeyRef:
                  name: {{ include "postgresql-ha.postgresqlSecretName" . }}
                  key: password
            {{- end }}

```
key should be "postgres-password", not "password".

Secret file content example below.

How to test:
```
kubectl exec --tty --stdin -n <my_namespace> pod/my-postgresql-ha-pgpool-b9c77db9d-q5wrt -- bash -c "env | sort"
```

=> check content of PGPOOL_POSTGRES_PASSWORD
=> using wrong password
=> pgpool keeps restarting (due to invalid probe)...

secret-postgresql.yaml
```yaml
apiVersion: v1
kind: Secret
metadata:
  name: my-db-secret
  labels:
    {{ toYaml .Values.global.labels | nindent 4 }}
type: Opaque
stringData:
  postgres-password: pwd1
  password: pwd2
  repmgr-password: pwd3
```

PGPOOL_POSTGRES_PASSWORD should use password from correct secretKeyRef.

### What do you see instead?

Password from the other key.

rstribrn:
### Description of the change

Fixes PGPOOL_POSTGRES_PASSWORD mapping from postgres-password
(see https://github.com/bitnami/charts/issues/32807)

### Benefits

Correct usage of different type of passwords.

### Possible drawbacks

Unknown.

### Applicable issues

- fixes #32807

### Additional information

Not tested myself.

### Checklist

Test different variants of (un)defined passwords in secretKeyRef.


javsalgar:
Hi!

Thank you so much for the PR! Could you bump the chart patch version and fix the DCO issue?

bitnami/postgresql-ha/templates/pgpool/deployment.yaml
```diff
@@ -227,7 +227,7 @@ spec:
               valueFrom:
                 secretKeyRef:
                   name: {{ include "postgresql-ha.postgresqlSecretName" . }}
-                  key: password
+                  key: postgres-password
```
fmulero, line 230: Password secret could not contain `postgres-password` key if `postgresql.username == 'postgres'`:  https://github.com/bitnami/charts/blob/375ee44a7787e8dde6573924fb87994a83e64d2f/bitnami/postgresql-ha/templates/postgresql/secrets.yaml#L18-L24

fmulero (commented):
Thanks a lot @rstribrn for your contribution.

This solves the problem you 've reported but I think the solution is not complete. Could you take a glance to the comment below?

bitnami/postgresql-ha/templates/pgpool/deployment.yaml
rstribrn: Thanks for reviewing the code. Unfortunately, I found no documentation about postgresqlUsername at 
https://github.com/bitnami/charts/tree/main/bitnami/postgresql
or
https://github.com/bitnami/charts/tree/main/bitnami/postgresql-ha

There is just a piece of information at
https://github.com/bitnami/charts/tree/main/bitnami/postgresql#automated-update-using-a-password-update-job
, but about "auth.postgresPassword".

See https://github.com/bitnami/charts/tree/main/bitnami/postgresql#global-parameters as well.
No information about what you're mentioning here.

So this is other (documentation)  bug?

fmulero: The `include` key means that `postgresql-ha.postgresqlUsername` is [named template](https://helm.sh/docs/chart_template_guide/named_templates/), and you can find the definition for that template in the `_helpers.tpl` file: https://github.com/bitnami/charts/blob/8e7d44a8fd17b6ab8c32c904fcc34c4aa5b23375/bitnami/postgresql-ha/templates/_helpers.tpl#L75-L80


About the concrete suggestion, if you try to install the chart with the default values, setting only `pgpool.usePasswordFiles=false` to use this piece of code, you could see the pgpool pod hang with a `CreateContainerConfigError` because the `postgres-password` doesn't exists.

```suggestion
                  {{- if not (eq (include "postgresql-ha.postgresqlUsername" .) "postgres") }} 
                  key: postgres-password
                  {{- else }}
                  key: password
                  {{- end }}
```

rstribrn: So,
if I understand you correctly, you're just saying the whole pull request is wrong and we should close it.
Right?

I've no problem with this since I just wanted to help a little.
Let me know.

fmulero: No and sorry if you thought that. We appreciate any help and any contribution.

Just apply this suggestion, give it a try and check if it fits to you:

```suggestion
                  {{- if not (eq (include "postgresql-ha.postgresqlUsername" .) "postgres") }} 
                  key: postgres-password
                  {{- else }}
                  key: password
                  {{- end }}
```

PR #32865, opened 2025-04-08
[bitnami/redis] If labeling isMaster failed, add a retry.

mcpovel:
Added a continue to bypass file removal if kubectl failed.


javsalgar:
Hi!

Thank you so much for the PR!

- Could you bump the patch version of the chart?
- Could you also change the PR title to something more meaningful? It automatically becomes part of the changelog

bitnami/redis/templates/scripts-configmap.yaml
```diff
@@ -853,7 +853,7 @@ data:
 
         if [ -f "/etc/shared/current" ]; then
             echo "new master elected, updating label(s)..."
-            kubectl label pod --field-selector metadata.name="$(< "/etc/shared/current")" isMaster="true" --overwrite
+            kubectl label pod --field-selector metadata.name="$(< "/etc/shared/current")" isMaster="true" --overwrite || contimue
```
kppullin, line 856:
```suggestion
            kubectl label pod --field-selector metadata.name="$(< "/etc/shared/current")" isMaster="true" --overwrite || continue
```

mcpovel: Sorry for my typo, agreed

FewKinG:
I noticed some errors with Redis deployments in `replication` architecture using a masterService. Sometimes when a new master is elected, the Kubernetes resources do not get updated. The part on the Redis side seems okay, there actually is 1 master and 1 replica. However, I end up in states where both replicas bear the label `app.kubernetes.io/role: slave` and thus no pod is matched by the selector for the `-master`-service.

Workloads trying to connect to Redis using the master service's DNS name fail to connect because the service does not match any pod.
Manually deleting both replica Pods solved the issue temporarily, but I'd prefer a more stable long-term solution that does not require manual intervention ;)

From what I can tell the error stems from the same problem described in this issue. In the logs of the `kubectl-shared` container I see errors when an attempt was made to update the labels:

```
new master elected, updating label(s)...                                                                                                                                              
Error from server (InternalError): Internal error occurred: failed calling webhook "validate.kyverno.svc-fail": [...]
```

Even after several hours, no retry is attempted. In my view, this needs to be fixed for the masterService to be considered usable.
Reviewing this issue's chat history I can see no reason why this fix has not been and should not be merged.

Please consider reopening this issue and merging the proposed fix. Thank you very much.

PS: I can see the reasoning behind automatically closing issues after a certain time of inactivity. But automatically labelling them as `solved` seems wrong to me, this issue being an obvious example.

PR #32955, opened 2025-04-10
[bitnami/grafana-loki] Add support to override client_max_body_size in nginx

iamjunaidali:
### Description of the change

 
In this PR I have added functionality to override `client_max_body_size`, by default the value is `1m` but in a few use cases we need to increase it and currently we have no support to override it.
 
### Benefits

Benefits of adding this will avoid monkey patching the original helm chart and ease to configure `client_max_body_size`.

### Possible drawbacks

I don't see any drawbacks for this.

### Applicable issues

- fixes #

### Checklist


bitnami/grafana-loki/Chart.yaml
```diff
@@ -56,4 +56,4 @@ maintainers:
 name: grafana-loki
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/grafana-loki
-version: 4.7.10
+version: 4.7.11
```
fmulero, line 59: According to semver.org, we should bump a minor version here:

```suggestion
version: 4.8.0
```

fmulero (changes requested):

PR #33040, opened 2025-04-16
Dynamic tpl rendering of the external database port

KorenP1:
Title says everything :)

### DIFF
--- a/bitnami/keycloak/Chart.yaml
+++ b/bitnami/keycloak/Chart.yaml
@@ -34,4 +34,4 @@ maintainers:
 name: keycloak
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/keycloak
-version: 24.5.4
+version: 24.5.5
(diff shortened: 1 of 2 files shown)


KorenP1:
@fmulero

KorenP1:
@juan131 
Sorry for bothering

fmulero (commented):
Thanks @KorenP1 for your contribution.

Looks good to me, but could you share at least one scenario where we can take advantage of this change?

fmulero (approved):

KorenP1:
> Thanks @KorenP1 for your contribution.
> 
> Looks good to me, but could you share at least one scenario where we can take advantage of this change?

@fmulero 
Sure! In my use case, my application can connect either to an internal keycloak that is deployed by the bitnami keycloak helm chart or either to an already deployed keycloak internally or externally. In my helm there is a function that returns the relevant port by the specific use case and i want to tpl run it inside the port value. Thanks.

PR #33173, opened 2025-04-24
[bitnami/common] add namespaces to extraPodAffinityTerms for affinities

klamas1:
# Fix indentation and namespace handling in affinity helpers

## Changes

1. Fixed indentation in `common.affinities.pods.hard`:
   - `namespaces` is now at the same level as `labelSelector`
   - Fixed indentation for `common.tplvalues.render` from `nindent 10` to `nindent 8`

2. Unified namespace handling in both helpers (`soft` and `hard`):
   - Current namespace is automatically added in the base rule via `{{ .context.Release.Namespace }}`
   - Current namespace is also automatically added in additional rules via `{{ $.context.Release.Namespace }}` because in range
   - Additional namespaces are added after the current namespace via `common.tplvalues.render`

3. Preserved dashes before `include` to prevent line breaks when using `nindent`

## Example of correct YAML structure

```yaml
  - labelSelector:
      matchLabels:
        ...
    namespaces:
      - current-namespace
      - additional-namespace1
      - additional-namespace2
    topologyKey: ...
```

## Testing

Changes were tested on Helm chart rendering with various namespace configurations.
### Possible drawbacks

Not found

### Checklist


bitnami/common/templates/_affinities.tpl
```diff
@@ -97,6 +97,13 @@ preferredDuringSchedulingIgnoredDuringExecution:
           {{- range $key, $value := .extraMatchLabels }}
           {{ $key }}: {{ $value | quote }}
           {{- end }}
+      {{- if .namespaces }}
+      namespaces:
+        - {{ $.context.Release.Namespace }}
+        {{- with .namespaces }}
+        {{- include "common.tplvalues.render" (dict "value" . "context" $) | nindent 8 }}
+        {{- end }}
+      {{- end }}
```
juan131, line 106: Are you introducing a new `namepaces` parameter to the helper? According to the helper description extra namespaces should be added using `extraNamespaces` parameter.

juan131 (changes requested):

bitnami/common/templates/_affinities.tpl
klamas1: I am not adding a new namespaces parameter to the helper. I am using the already existing extraPodAffinityTerms parameter, which previously did not support specifying namespaces, which did not allow them to be set for additional affinity terms. Thus, additional namespaces can be specified there, and this does not violate compatibility with the helper description — extraNamespaces is still supported for the main term, and namespaces inside extraPodAffinityTerms is used for additional conditions.

juan131: Got it! Thanks so much for explaining the details

bitnami/common/Chart.yaml
```diff
@@ -23,4 +23,4 @@ name: common
 sources:
   - https://github.com/bitnami/charts/tree/main/bitnami/common
 type: library
-version: 2.30.0
+version: 2.31.0
```
juan131, line 26: In this case, I'd bump patch version instead. I don't think we're introducing a new feature but fixing a bug given it addresses indentation errors and it make the helper to take into account sth it should already consider.

juan131 (changes requested):
The changes LGTM, but please ensure you merge the latest changes in the origin's `main` branch in your fork's branch and then, update the PR. Thanks in advance.

bitnami/common/Chart.yaml
klamas1: It's fair, I'll do it

juan131 (approved):
LGTM

PR #33194, opened 2025-04-26
[bitnami/keycloak] Allow tpl of external database + external user

KorenP1:
Allow tpl of external database + external user


dgomezleon (approved):

PR #33198, opened 2025-04-27
[bitnami/postgresql] Global postgres fullname override, needed when the fullname needs to be shared across subcharts.

KorenP1:
### Description of the change

Added value `global.postgresql.fullnameOverride`.

### Benefits

Global PostgreSQL fullname override, needed when the fullname needs to be shared across subcharts like the other global PostgreSQL values.

### Possible drawbacks

None

### Applicable issues

None

### Additional information

The function changed to a single ternary operation for compact and read reasons.

### Checklist


KorenP1:
@gongomgra @juan131

juan131:
Hi @KorenP1 

This isn't required, we solve it in other charts by creating a specific helper to get the PostgreSQL fullname. For instance, check the helper below:

- https://github.com/bitnami/charts/blob/main/bitnami/airflow/templates/_helpers.tpl#L71-L77

KorenP1:
@juan131 @gongomgra 
If there are any questions I can email you guys privately

juan131:
Thanks for explaining your use case @KorenP1

The issue here is that one of the sub-charts your chart deploys (Airflow in the example) uses another sub-chart as db backend (PostgreSQL in the example) hence you need to know the svc name where to connect (which is affected by the fullname override).

Please correct me if I'm wrong but this can be achieved without using global parameters:

```yaml
(...)
postgresql:
  fullnameOverride: pg
airflow:
  postgresql:
    enabled: false
  externalDatabase:
    host: pg
```

KorenP1:
@juan131 
Thanks.
yes this will absolutely work. The thing is i want to do it dynamically with the following values example
```
postgresql:
  fullnameOverride: pg
airflow:
  postgresql:
    enabled: false
  externalDatabase:
    host: '{{ .Values.global.postgresql.fullnameOverride }}'
```

juan131:
All right! IMO your is quiet a specific use case, but, given the change doesn't introduce any technical concern I'm okay with accepting it. Thanks for the contribution.

juan131 (approved):

PR #33492, opened 2025-05-07
[bitnami/keycloak] fix: preserve timestamps in init script

msvechla:
### Description of the change

With the latest Keycloak version, the timestamps of providers is compared to the timestamp of the latest provider build.

When starting keycloak with the `--optimized` flag, keycloak will throw an error when the timestamp of the providers is more recent than it was during the build.

> A provider JAR was updated since the last build, please rebuild for this to be fully utilized.

As we do not preserve the timestamp in the init script, this error is always thrown when starting keycloak as optimized.

This is fixed by preserving the timestamp in the initContainer `cp` commands.

see: https://github.com/keycloak/keycloak/issues/38893

### Benefits

keycloak can start again in optimized mode

### Applicable issues

- fixes https://github.com/keycloak/keycloak/issues/38893

### Checklist


msvechla:
@carrodher could you take a look at this? This is breaking keycloak startup for everyone that uses the `--optimized` flag starting with version `26.2.0`.

Let me know if anything needs to be adapted or clarified!

Thanks a lot!

msvechla:
@migruiz4 FYI I updated the version again and rebased, as something else was merged before this MR.

msvechla:
@migruiz4 @carrodher I rebased and updated the version again, as something else was merged that bumped the version. Can we get this merged?

Thanks a lot for your help!

PR #33525, opened 2025-05-07
[bitnami/prometheus] Allow tpl for ingress hostnames and allow ingress annotations to be both map object and string

KorenP1:
Allowing the tpl of ingress hostnames to match bitnami standards and allow the dynamic allocation of hostnames based on global configuration

Allowing the ingress annotations to be also type string. (Was only map object type)
This allows using string tpl in the annotations key.
Example values that is now allowed:
```yaml
myGlobalMap:
  a: b
  c: d
  e: f
  
server:
  ingress:
    enabled: true
    annotations: '{{ .Values.myGlobalMap | toYaml }}'
```


juan131 (approved):
Thanks for the improvement! LGTM

PR #33593, opened 2025-05-09
[bitnami/postgresql] Postgresql global fullnameOverride fix

KorenP1:
As for the last PR about global postgresql fullnameOverride. #33198 It is currently not functioning as expected.
The postgresql chart uses the `"common.names.fullname"` function and a local function called `"postgresql.v1.primary.fullname"` with some other naming functions.

When I changed the function `"postgresql.v1.primary.fullname"` in the last PR I wasnt paying attention to the other naming functions. and also the still using `"common.names.fullname"` function.

In order to solve this i have overriden the function `"common.names.fullname"` to use the `"common.names.dependency.fullname"` function (Couldn't override with the same function because it would end up in a recursion) which does the same but gets arguments for it. I have added an if else to override with the `global.postgresql.fullnameOverride` value. Also added .Chart.IsRoot validation because after overriding a function it is getting overriden in the whole context including subcharts. In this postgresql chart case its just in case this chart will somewhen use subchart which I dont think it will.

This whole PR is supposed to be a dialog with the mods because there are alot of ways to implement this `global.postgresql.fullnameOverride` feature, I chose the one i am offering here but i will describe also other ways that have come to my mind to implement it.

1. Abandon this feature and revert to a commit before the PR Mentioned above.

2. Overriding the function as i described earlier

3. Creating a new function that retrieves the general fullname and change every occurance of `"common.names.fullname"` to that new function

4. Implement some feature in the common chart


KorenP1:
@javsalgar @juan131 
Sorry for the misunderstanding in the last PR
Hopefully this one explains it better. Im availalbe for any questions.
Thanks. Koren.

juan131 (changes requested):
We should try not to override any function from the common library. The option 3 you mentioned might be a better solution:

> Creating a new function that retrieves the general fullname and change every occurrence of "common.names.fullname" to that new function

PR #33672, opened 2025-05-13
[bitnami/grafana] Deterministic admin password secret + saving password for future uses if .Values.admin.password is not set

KorenP1:
Currently the grafana deployment will rollout restart every upgrade if no admin password is set because it will randomize it while using the function randAlphaNumeric which returns a random password but is generated twice. Once while creating the secret and a second time while including it in the deployments pods annotation which will give different results and end up in a different sha256sum result. which means that the annotation is not representing the hashing of the secret.
To solve this i used a seed and changed the randomizing of the password to `(now | unixEpoch | sha256sum | b64enc | trunc 10)`

Another problem is that if we choose to randomize the password, the password will change every upgrade too and that will result in an unwanted restart to the grafana pod. ALSO IN CASE YOU DONT USE PERSISTENCE YOUR PASSWORD WILL CHANGE INSIDE THE GRAFANA.

To solve this i have added lookup function to see if the secret already exists and it will use the last generated password instead of generating a new one.

### EDIT
Changed to use the common chart generate password function instead. It is deterministic after first upgrade.
It is a standard in other bitnami charts.
To fix the issue that its not deterministic in the first run, maybe should use a seed with the now | unixEpoch method mentioned above, but that is for the common chart and not here.


jotamartos (approved):
Changes LGTM! Thank you for your contribution.

PR #33689, opened 2025-05-14
[bitnami/grafana] Grafana uniq dashboard volumes

KorenP1:
When using grafana dashboard configMap feature (.Values.dashboardsConfigMaps) and mapping multiple dashboards to the same configMapName, the deployment will fail becaues the templating will try to mount same volume twice or more, We have to make the .spec.template.spec.volumes Uniq.


juan131 (approved):
LGTM! Thanks for the fix

PR #33811, opened 2025-05-21
Added image.tag parameter

AsphaltExpert:
Added image.tag parameters to "Redis(R) Image parameters" and "Metrics Parameters" tables

### Description of the change

This is a very small change: you can set an image tag for the metrics, but it was not clearly mentioned and was completely missing from the parameters table.

### Applicable issues

- fixes #

### Checklist


carrodher:
Thanks for your contribution! Unfortunately, we can't accept this change, as those parameters are intentionally omitted from the README table due to how frequently they change. To handle that, we use the `@skip` metadata in the parameter definition — see this example for reference:
https://github.com/bitnami/charts/blob/3525eb1e80566d54c52d0e94b9f5af368b358c22/bitnami/redis/values.yaml#L108

PR #33834, opened 2025-05-22
[bitnami/spark] Spark optional worker

pckroon:
### Description of the change
I added 2 values to the bitnami/spark chart: `worker.enabled` and `master.enabled` (both defaulting to `true`), which enable more flexible use of the chart. In my specific case, I'd like to only deploy the master in my cluster, while I run the workers elsewhere in my network.

### Possible drawbacks

Users setting both `worker.enabled=false` and `master.enabled=false` don't get a lot of use out of this chart. Also, I don't know how hard it would be to set `worker.enabled=true` and `master.enabled.false`, and then configure the worker to use an external master.

### Additional information

This is my first helm related PR, and I don't have a very sane dev environment set up (cypress, goss), so I hope the existing tests still pass and are sufficient.

### Checklist


pckroon:
Hi! 
Very welcome :)
I can imagine someone having either an already-existing out-of-cluster master, or even a master running in a different cluster, and only wanting to run add only workers. Definitely a rare scenario, but IMO plausible, and it's a pretty simple flag.

PR #33843, opened 2025-05-22
[bitnami/keycloak] Keycloak generate certificate global cluster domain support

KorenP1:
Title

### DIFF
--- a/bitnami/keycloak/templates/tls-secret.yaml
+++ b/bitnami/keycloak/templates/tls-secret.yaml
@@ -46,7 +46,7 @@ data:
 {{- $secretName := printf "%s-crt" (include "common.names.fullname" .) }}
 {{- $ca := genCA "keycloak-ca" 365 }}
 {{- $releaseNamespace := include "common.names.namespace" . }}
-{{- $clusterDomain := .Values.clusterDomain }}
+{{- $clusterDomain := default .Values.clusterDomain .Values.global.clusterDomain }}
 ---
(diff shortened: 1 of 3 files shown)


bitnami/keycloak/templates/tls-secret.yaml
```diff
@@ -46,7 +46,7 @@ data:
 {{- $secretName := printf "%s-crt" (include "common.names.fullname" .) }}
 {{- $ca := genCA "keycloak-ca" 365 }}
 {{- $releaseNamespace := include "common.names.namespace" . }}
-{{- $clusterDomain := .Values.clusterDomain }}
+{{- $clusterDomain := default .Values.clusterDomain .Values.global.clusterDomain }}
```
juan131, line 49: `global.clusterDomain` isn't a recognized chart parameter in the `values.yaml`

juan131 (changes requested):
Thanks so much for your contribution! Please check my comments.

bitnami/keycloak/templates/tls-secret.yaml
KorenP1: It is recognized in other bitnami charts so i added support here, do you think we should add it in the values.yaml here too?

KorenP1: @juan131 
My bad, i dont know why i though it exists in other bitnami charts but looks like it dont. Thanks.

PR #33894, opened 2025-05-27
[bitnami/kube-prometheus] add thanos ruler CR

### LINKED ISSUE #28525: [bitnami/thanos] Add integration for Thanos Ruler with PrometheusRule via ruleSelector
### Name and Version

bitnami/thanos latest

The current configuration only allows Ruler to interact with recordings and alertings via configmap, a big no if a project is already running hundreds of records of PrometheusRules to manage Prometheus records and alerts.

Just like in the kube-prometheus-stack, using [rulerSelector](https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/values.yaml#L4467) for discovering rules will make the overall experience much more easier to automate.

Using the kube-prometheus-stack just to deploy the Thanos Ruler, which wouldn’t be the best option…

### LINKED ISSUE #32826: [bitnami/thanos] Support creation of ThanosRuler
### Name and Version

bitnami/thanos

Currently, Thanos Ruler can be only be deployed in `thanos` chart by means of a [Statefulset](https://github.com/bitnami/charts/blob/main/bitnami/thanos/templates/ruler/statefulset.yaml).

Prometheus Operator supports creation of Thanos Ruler [using a CR](https://prometheus-operator.dev/docs/api-reference/api/#monitoring.coreos.com/v1.ThanosRuler).

Using a CR takes away complexity from the Statefulset, but also allows the usage of [PrometheusRule](https://prometheus-operator.dev/docs/api-reference/api/#monitoring.coreos.com/v1.PrometheusRule) objects to define rules in a more flexible manner (using rule selectors to state which Thanos Ruler is the target for a given rule).

This is already the way [kube-prometheus-stack deploys it](https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/templates/thanos-ruler/ruler.yaml).

This issue relates to https://github.com/bitnami/charts/issues/28525

Given we probably don't want to get rid of the Satefulset deployment for backwards compatibility, we should probably add a new template to deploy it using the `ThanosRuler` CR, and then put it behind a feature flag that selects how ruler will be deployed.

Having the CR created as additional resources, but it is not pretty since the auxiliary resources (secret, service, etc...) are controlled by the same feature flag that enables ruler.

We could maybe change that logic to allow creating all those resources but skipping Statefulset creation (and then one can add the CR as extra manifests), but I think having the CR inside as an option would be a best choice.

kir4h:
### Description of the change

This PR adds support for creating a [ThanosRuler CR](https://prometheus-operator.dev/docs/getting-started/design/#thanosruler).

`ThanosRuler` is already available as a `Statefulset` on [Bitnami Thanos chart](https://github.com/kir4h/bitnami-charts/tree/main/bitnami/thanos/templates/ruler), but there is no option to create it using the CRD deployed by Prometheus Operator. After some brainstorming on https://github.com/bitnami/charts/issues/32826 , the decision was to [add it as part of kube-prometheus chart](https://github.com/bitnami/charts/issues/32826#issuecomment-2817796931).

cc @juan131 

### Benefits

Using a CR for Thanos ruler takes away complexity from the Statefulset, but also allows the usage of [PrometheusRule](https://prometheus-operator.dev/docs/api-reference/api/#monitoring.coreos.com/v1.PrometheusRule) objects to define rules in a more flexible manner (using rule selectors to state which Thanos Ruler is the target for a given rule, or allowing to decouple the deployment of rules from ruler itself).

This is already the way [kube-prometheus-stack deploys it](https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/templates/thanos-ruler/ruler.yaml).

### Possible drawbacks

None that I can think of.

### Applicable issues

- fixes #32826 
- fixes #28525

### Additional information

I have done some rendering tests with different configurations and a basic deployment test with:

* [Thanos chart](https://github.com/bitnami/charts/tree/main/bitnami/thanos) with default values
* This chart with default values, but enabling ruler

```console
❯ helm ls
NAME           	NAMESPACE	REVISION	UPDATED                              	STATUS  	CHART                 	APP VERSION
kube-prometheus	default  	3       	2025-05-20 09:46:41.319475 +0200 CEST	deployed	kube-prometheus-11.2.0	0.82.2
thanos         	default  	3       	2025-05-13 08:26:18.067957 +0200 CEST	deployed	thanos-16.0.6         	0.38.0
```

Values used for kube-prometheus

```yaml
thanosRuler:
  enabled: true
  queryConfig:
    config:
    - static_configs:
      - "dnssrv+_http._tcp.thanos-query.default.svc.cluster.local"
```

Ruler running

```console
❯ k get po | grep ruler
thanos-ruler-kube-prometheus-thanos-ruler-0           2/2     Running   4 (23h ago)   7d
```

Ruler UI looks as expected

image

image

### Checklist


bitnami/kube-prometheus/values.yaml
```diff
@@ -3307,3 +3307,565 @@ kubeProxy:
...
+    ##         basic_auth:
+    ##           username: some_user
+    ##           password: some_pass
+    ##       static_configs:
+    ##         - alertmanager.thanos.io
+    ##       scheme: http
+    ##       timeout: 10s
+    ##       api_version: v2
+    ##
+    config: {}
```
juan131, line 3686: Shouldn't this configuration be auto-generated base on `alertmanager.*` parameters?

bitnami/kube-prometheus/values.yaml
```diff
@@ -3307,3 +3307,565 @@ kubeProxy:
...
+  ## @param thanosRuler.initContainers InitContainers allows adding initContainers to the pod definition
+  ## Those can be used to e.g. fetch secrets for injection into the ThanosRuler configuration from external sources.
+  ##
+  initContainers: []
+  ## @param thanosRuler.priorityClassName Priority class assigned to the Pods
+  ##
+  priorityClassName: ""
+  ## @param thanosRuler.portName Port name used for the pods and governing service. Defaults to web
+  ##
+  portName: "web"
```
juan131, line 3862: Default value not required in parameter description
```suggestion
  ## @param thanosRuler.portName Port name used for the pods and governing service
  ##
  portName: web
```

juan131 (changes requested):
Thanks so much for this amazing contribution, we really appreciate it! Please take a look to my comments when you have a chance.

bitnami/kube-prometheus/Chart.yaml
```diff
@@ -51,4 +51,4 @@ maintainers:
 name: kube-prometheus
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/kube-prometheus
-version: 11.1.13
+version: 11.2.0
```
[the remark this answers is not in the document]
kir4h: > Ensure you add the new container image under `annotations.images`

Thanos uses a single binary, so also a single docker image. The image was already present at https://github.com/bitnami/charts/blob/main/bitnami/kube-prometheus/Chart.yaml#L15-L16 since it's used for Thanos sidecar. 

I am not sure of the purpose of the key under `annotation.images` ; I think as it's metadata it's used for chart info so that entry should cover every usage of the image (please let me wrong if otherwise and I need to add an extra entry with different key)

bitnami/kube-prometheus/templates/thanos-ruler/networkpolicy.yaml
```diff
@@ -0,0 +1,113 @@
...
+        - port: 53
+          protocol: TCP
+    # Communicate with other ruler instances via headless service
+    - ports:
+        - port: {{ .Values.thanosRuler.containerPorts.http }}
+        - port: {{ .Values.thanosRuler.containerPorts.grpc }}
+      to:
+        - podSelector:
+            matchLabels: {{ include "common.labels.matchLabels" ( dict "customLabels" $podLabels "context" $ ) | nindent 14 }}
+              app.kubernetes.io/component: thanos-ruler
```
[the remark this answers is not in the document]
kir4h: Good point! I used Thanos chart as base, so I copied over that rule from https://github.com/bitnami/charts/blob/main/bitnami/thanos/templates/ruler/networkpolicy.yaml#L36-L43

But I see no reason why rulers should talk to each other, and haven't found any signs of it under Thanos docs. 
Since ruler is kind of a "piece of Prometheus", I checked Prometheus docs and on HA mode they don't talk to each other (https://promlabs.com/blog/2023/08/31/high-availability-for-prometheus-and-alertmanager-an-overview/). 

Additionally, ruler configuration doesn't include any argument related to service discovery of other rulers, so I'm removing this one. (and following lines since they reference minio, which is not part of this chart)

bitnami/kube-prometheus/values.yaml
```diff
@@ -3307,3 +3307,565 @@ kubeProxy:
...
+  ## Configure Container Security Context
+  ## ref: https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-container
+  ## @param thanosRuler.containerSecurityContext.enabled Enable container security context
+  ## @param thanosRuler.containerSecurityContext.runAsUser User ID for the container
+  ## @param thanosRuler.containerSecurityContext.runAsNonRoot Enable running as non-root
+  ##
+  containerSecurityContext:
+    enabled: true
+    runAsUser: 1001
+    runAsNonRoot: true
```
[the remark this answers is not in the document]
kir4h: Thanks! missed adapting this one (kept the format from the other chart). I will adapt it.

Regarding `containerSecurityContext`, I've been reading and my preferred approach would be not to include it.
Rationale:
* Not supported directly by the operator. We currently do some strategic merge patching to include it
* From https://prometheus-operator.dev/docs/api-reference/api/#monitoring.coreos.com/v1.ThanosRulerSpec  
> The current container names are: thanos-ruler and config-reloader. Overriding containers is entirely outside the scope of what the maintainers will support and by doing so, you accept that this behaviour may break at any time without notice.
* Operator containers have some security restrictions already in place, which I think should cover most cases https://github.com/prometheus-operator/prometheus-operator/commit/d46ae5145ed96b69647741f79100326d2f2a4786

wdyt?

bitnami/kube-prometheus/values.yaml
kir4h: Definitely. Again, ported from Thanos chart, that didn't include Alertmanager.

I'm thinking of having:
* ability to inject config through existing secret (already in place)
* a default config that uses `alertmanager.*` parameters to point to alertmanager deployed by this chart

What I'm not sure about:
* should we add ability not to have any alertmanager config even if the chart deploys one?
* should we add ability to deploy a static config ignoring the one deployed by the chart? 

Both are edge cases in my mind, typically if you deploy an alertmanager and a ruler with this chart, you will want ruler to talk to that Alertmanager. But maybe there are scenarios and adding support for that will give more flexibility?

bitnami/kube-prometheus/values.yaml
```diff
@@ -3307,3 +3307,565 @@ kubeProxy:
...
+  ##     memory: 1024Mi
+  ## ref: https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/
+  ##
+  resources: {}
+  ## @param thanosRuler.containerPorts.http HTTP container port
+  ## @param thanosRuler.containerPorts.grpc GRPC container port
+  ##
+  containerPorts:
+    http: 10902
+    grpc: 10901
```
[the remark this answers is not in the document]
kir4h: good catch! I need to inject them as `additionalArgs`

kir4h:
> Thanks so much for this amazing contribution, we really appreciate it! Please take a look to my comments when you have a chance.

Thanks for the quick review! Tough one to review since it's a long PR.
 I have started implementing the requested changes (applied all suggestions, need to finish some of the comments) will continue tomorrow morning

kir4h:
@juan131 I've submitted the changes.

Some comments:

* disabled alertmanager configuration on prometheus when thanos ruler is enabled ( I think it makes sense, since we don't want both sending the same alerts)
* ports are not exposed through the CR, so had to use the `additionalArgs` field. I think I need to refine the readme generator annotation, I added a skip and now defaults to "undefined", not sure what I should force default to in description (I could also hardcode these port values in the templates, and then concatenate `additionalArgs` leaving it empty by default if you think it's better
* for alertmanager config, I have added a third scenario 1) use existing secret if provided 2) use config if provided 3) fallback to chart alertmanager if alertmanager enabled

Regarding conversation resolution on this PR: should I resolve my self or let you do it?

bitnami/kube-prometheus/templates/thanos-ruler/networkpolicy.yaml
```diff
@@ -0,0 +1,113 @@
...
+    # Communicate with query-frontend
+    - ports:
+        - port: {{ .Values.queryFrontend.service.ports.http }}
+        - port: {{ .Values.queryFrontend.containerPorts.http }}
+      to:
+        - podSelector:
+            matchLabels: {{- include "common.labels.matchLabels" ( dict "customLabels" $podLabels "context" $ ) | nindent 14 }}
+              app.kubernetes.io/component: query-frontend
+
+    {{- end }}
```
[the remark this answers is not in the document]
juan131: Nope, feel free to skip the validation. Thanks for the details.

bitnami/kube-prometheus/templates/prometheus/prometheus.yaml
```diff
@@ -41,7 +41,7 @@ spec:
     alertmanagers:
     {{- if .Values.prometheus.alertingEndpoints }}
     {{- include "common.tplvalues.render" (dict "value" .Values.prometheus.alertingEndpoints "context" $) | nindent 6 }}
-    {{- else if .Values.alertmanager.enabled }}
+    {{- else if and .Values.alertmanager.enabled (not .Values.thanosRuler.enabled) }}
```
juan131: I think we should relax this and throw a warning instead (via `kube-prometheus.validateValues` helper). It's up to the user to configure alerts twice.

bitnami/kube-prometheus/templates/thanos-ruler/secret-alertmanager-config.yaml
```diff
@@ -0,0 +1,30 @@
...
+      {{- include "common.tplvalues.render" (dict "value" .Values.thanosRuler.alertmanagersConfig.config "context" $) | b64enc | nindent 4 }}
+    {{ else }}
+    alertmanagers:
+    - static_configs:
+      - {{ template "kube-prometheus.alertmanager.fullname" . }}.{{ include "common.names.namespace" . }}:{{ .Values.alertmanager.service.ports.http }}
+      scheme: http
+      timeout: 10s
+      api_version: v2
+      path_prefix: {{ .Values.alertmanager.routePrefix }}
+    {{- end }}
```
juan131, line 31: There are no credentials on this default configuration. I guess that, in this case, the configuration should be kept in a ConfigMap instead of a secret, does it make sense?

bitnami/kube-prometheus/values.yaml
juan131: > should we add ability not to have any alertmanager config even if the chart deploys one? should we add ability to deploy a static config ignoring the one deployed by the chart?

It's enough with what you did given a custom config can be achieved via existing secret.

bitnami/kube-prometheus/templates/thanos-ruler/service-headless.yaml
```diff
@@ -0,0 +1,33 @@
+{{- /*
+Copyright Broadcom, Inc. All Rights Reserved.
+SPDX-License-Identifier: APACHE-2.0
+*/}}
+
+{{- if and (.Values.thanosRuler.enabled) (.Values.thanosRuler.service.additionalHeadless) }}
```
juan131: No parenthesis required here:
```suggestion
{{- if and .Values.thanosRuler.enabled .Values.thanosRuler.service.additionalHeadless }}
```

bitnami/kube-prometheus/templates/thanos-ruler/thanosruler.yaml
```diff
@@ -0,0 +1,164 @@
...
+*/}}
+
+{{- if .Values.thanosRuler.enabled }}
+apiVersion: monitoring.coreos.com/v1
+kind: ThanosRuler
+metadata:
+  name: {{ template "kube-prometheus.thanosRuler.name" . }}
+  namespace: {{ include "common.names.namespace" . | quote }}
+  {{- $labels := include "common.tplvalues.merge" ( dict "values" ( list .Values.thanosRuler.service.labels .Values.commonLabels ) "context" . ) }}
+  labels: {{ include "common.labels.standard" ( dict "customLabels" $labels "context" $ ) | nindent 4 }}
```
juan131: We should include a `app.kubernetes.io/version` label based on the Thanos image version (instead of relying on Prometheus Operator one) and also removed the usage of `.Values.thanosRuler.service.labels` which is only intended for the service:
```suggestion
  {{- $versionLabel := dict "app.kubernetes.io/version" (include "common.images.version" (dict "imageRoot" .Values.thanosRuler.image "chart" .Chart)) }}
  {{- $labels := include "common.tplvalues.merge" (dict "values" (list .Values.commonLabels $versionLabel) "context" .) }}
  labels: {{ include "common.labels.standard" (dict "customLabels" $labels "context" .) | nindent 4 }}
```
Please note you can apply this to every manifest under `bitnami/kube-prometheus/templates/thanos-ruler` folder.

bitnami/kube-prometheus/values.yaml
```diff
@@ -3307,3 +3307,564 @@ kubeProxy:
...
+  ##
+  ruleSelector: {}
+  ## @param thanosRuler.evalInterval How frequently to evaluate rules
+  ##
+  evalInterval: "1m"
+  ## @param thanosRuler.clusterName Used to set the 'ruler_cluster' label
+  ##
+  clusterName: ""
+  ## @param thanosRuler.additionalArgs Additional arguments for the ThanosRuler container.
+  ## @skip thanosRuler.additionalArgs
```
juan131:
```suggestion
  ## @param thanosRuler.additionalArgs [array] Additional arguments for the ThanosRuler container.
```

bitnami/kube-prometheus/templates/prometheus/prometheus.yaml
kir4h: > I think we should relax this and throw a warning instead (via `kube-prometheus.validateValues` helper). It's up to the user to configure alerts twice.

Thinking about it again, that condition didn't make sense: It's ok for both to send alerts to Alertmanager, what we probably don't want is for them to handle the same alerts so I have instead added a validation for rule selectors.

`kube-prometheus.validateValues` throws an error (so more than a warning), is that ok?

bitnami/kube-prometheus/templates/thanos-ruler/secret-alertmanager-config.yaml
kir4h: Yes, default configuration does not hold any credentials (it could potentially include basic auth, but I haven't included an option for it aligning with what we do for Prometheus).

~I kept it as a secret to reuse the same resource we use when the config is provided, but I will move this to a configmap instead~

I just remembered why it's a secret. `alertmanagersConfig` in the CR is of type [secretKeySelector](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.31/#secretkeyselector-v1-core)  (see https://prometheus-operator.dev/docs/api-reference/api/#monitoring.coreos.com/v1.ThanosRulerSpec)

juan131: Good point! Keep it as it is then

bitnami/kube-prometheus/templates/prometheus/prometheus.yaml
juan131: Yes, it's okay to use `kube-prometheus.validateValues`

bitnami/kube-prometheus/templates/thanos-ruler/ingress.yaml
```diff
@@ -0,0 +1,59 @@
...
+*/}}
+
+{{- if and .Values.thanosRuler.enabled .Values.thanosRuler.ingress.enabled }}
+apiVersion: {{ include "common.capabilities.ingress.apiVersion" . }}
+kind: Ingress
+metadata:
+  name: {{ include "kube-prometheus.thanosRuler.fullname" . }}
+  namespace: {{ include "common.names.namespace" . }}
+  {{- $versionLabel := dict "app.kubernetes.io/version" (include "common.images.version" (dict "imageRoot" .Values.thanosRuler.image "chart" .Chart)) }}
+  {{- $labels := include "common.tplvalues.merge" (dict "values" (list .Values.commonLabels $versionLabel) "context" .) }}
```
juan131:
```suggestion
  {{- $labels := include "common.tplvalues.merge" (dict "values" (list .Values.thanosRuler.ingress.labels .Values.commonLabels $versionLabel) "context" .) }}
```

bitnami/kube-prometheus/templates/thanos-ruler/secrets-tls.yaml
```diff
@@ -0,0 +1,48 @@
+{{- /*
```
juan131, line 1: Please rename this file to `ingress-tls-secret.yaml`

bitnami/kube-prometheus/templates/thanos-ruler/service.yaml
```diff
@@ -0,0 +1,64 @@
...
+*/}}
+
+{{- if .Values.thanosRuler.enabled }}
+apiVersion: v1
+kind: Service
+metadata:
+  name: {{ include "kube-prometheus.thanosRuler.fullname" . }}
+  namespace: {{ include "common.names.namespace" . }}
+  {{- $versionLabel := dict "app.kubernetes.io/version" (include "common.images.version" (dict "imageRoot" .Values.thanosRuler.image "chart" .Chart)) }}
+  {{- $labels := include "common.tplvalues.merge" (dict "values" (list .Values.commonLabels $versionLabel) "context" .) }}
```
juan131:
```suggestion
  {{- $labels := include "common.tplvalues.merge" (dict "values" (list .Values.thanosRuler.service.labels .Values.commonLabels $versionLabel) "context" .) }}
```

bitnami/kube-prometheus/templates/thanos-ruler/servicemonitor.yaml
```diff
@@ -0,0 +1,48 @@
...
+*/}}
+
+{{- if and .Values.thanosRuler.enabled .Values.thanosRuler.serviceMonitor.enabled }}
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  name: {{ include "kube-prometheus.thanosRuler.fullname" . }}
+  namespace: {{ include "common.names.namespace" . }}
+  {{- $versionLabel := dict "app.kubernetes.io/version" (include "common.images.version" (dict "imageRoot" .Values.thanosRuler.image "chart" .Chart)) }}
+  {{- $labels := include "common.tplvalues.merge" (dict "values" (list .Values.commonLabels $versionLabel) "context" .) }}
```
juan131:
```suggestion
  {{- $labels := include "common.tplvalues.merge" (dict "values" (list .Values.thanosRuler.serviceMonitor.labels .Values.commonLabels $versionLabel) "context" .) }}
```

juan131 (commented):
This is almost ready! Just a few very minor comments

bitnami/kube-prometheus/values.yaml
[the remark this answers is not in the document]
juan131: Can we at least rename the parameters to be alike other components? We can simplify the object to only have the fields supported by the operator

kir4h:
> This is almost ready! Just a few very minor comments

apologies, did a copy paste on latest review (labels thing) without realizing I had to include the specific values on each of them.

I think I checked all the comments, but let me know if I missed any

juan131 (approved):
Amazing contribution! LGTM

PR #33951, opened 2025-05-28
[bitnami/external-dns] Fix bug with trailing comma in azure.json

### LINKED ISSUE #33882: azure.json cannot be parsed when working with workload identity
### Name and Version

bitnami/external-dns 8.8.3

Azure Kubernetes Services

We are using external-dns togehter with Azure DNS. external-dns authenticates using Azure workload identity.
After updating from 8.8.2 to 8.8.3 the external-dns pod crashes during startup with this error message:

`failed to read Azure config file '/etc/kubernetes/azure.json': failed to parse Azure config file '/etc/kubernetes/azure.json': invalid character '}' looking for beginning of object key string`

We use the following parameters with the helm chart, where `useWorkloadIdentityExtension` is causing the issue.

```yaml
azure:
  useWorkloadIdentityExtension: true
  resourceGroup: rg-name
  subscriptionId: our-subscription-id
  tenantId: our-tenant-id
```

This configuration results in the following JSON file stored in the chart's Kubernetes secret:

```json
{
  "tenantId": "our-tenant-id",
  "subscriptionId": "our-subscription-id",
  "resourceGroup": "rg-name",
  "useWorkloadIdentityExtension":  true,
}
```

Note the comma at the end of the last line. This is causing the issue. If I remove it and recreate the secret, the pod starts successfully.

With 8.8.3, using workload identity with Azure DNS should not cause the external-dns pod to crash.

### Additional information

The [_helpers.tpl](https://github.com/bitnami/charts/blob/main/bitnami/external-dns/templates/_helpers.tpl) defines the variable `"external-dns.azure-credentials"` which always adds `"useWorkloadIdentityExtension":  true,` when using workload identity, no matter if some more values are following.

Could a solution be to either add the line with or without a trailing comma, depending on whether the managed identity extension values are set afterwards? Is there even a scenario in which workload identity and managed identity can be used together?

MaxAnderson95:
This PR resolves #33882.

A breaking change was introduced in `external-dns` `0.17.0` which now (correctly) fails to parse a previously innocuous bug in how the `azure.json` was generated in this chart. Since the introduction of workload identity support in this chart (and likely before), the Azure credentials were generated by constructing a raw JSON string. When using only workload identity, this resulted in an invalid `azure.json` containing a trailing comma on the last key-value pair:

```json
{
  "tenantId": "11111111-1111-1111-1111-111111111111",
  "subscriptionId": "0000000-0000-0000-0000-000000000000",
  "resourceGroup": "foo-group",
  "useWorkloadIdentityExtension":  true,
}
```

In `external-dns` `0.16.1` this was apparently accepted as valid JSON and did not cause a parsing error. Since the upgrade to chart version `8.8.3` which included `external-dns` `0.17.0`, this is now causing a parsing error:

```
failed to read Azure config file '/etc/kubernetes/azure.json': failed to parse Azure config file '/etc/kubernetes/azure.json': invalid character '}' looking for beginning of object key string
```

### Description of the change

This PR uses the `toJson` helm function to dynamically generate the JSON object, eliminating the need to worry about key order or trailing commas.

### Benefits

The method is much cleaner rather than constructing a raw string.

### Possible drawbacks

One slight drawback is that *falsey* values now appear in the JSON object passed to `azure.json`. 

Before:
```json
{
  "tenantId": "11111111-1111-1111-1111-111111111111",
  "subscriptionId": "0000000-0000-0000-0000-000000000000",
  "resourceGroup": "foo-group",
  "useWorkloadIdentityExtension":  true,
}
```

After:
```json
{
  "aadClientId": "",
  "aadClientSecret": "",
  "cloud": "",
  "resourceGroup": "foo-group",
  "subscriptionId": "0000000-0000-0000-0000-000000000000",
  "tenantId": "11111111-1111-1111-1111-111111111111",
  "useManagedIdentityExtension": false,
  "useWorkloadIdentityExtension": true,
  "userAssignedIdentityID": ""
}
```

I tested in one of my test clusters and had no issues, and I also reviewed the [relevant parsing code](https://github.com/kubernetes-sigs/external-dns/blob/v0.17.0/provider/azure/config.go) and found no indication it would handle *falsey* values differently than non-existent keys. 

### Applicable issues

- Fixes #33882

### Checklist


bitnami/external-dns/templates/_helpers.tpl
```diff
@@ -168,33 +168,16 @@ region = {{ .Values.aws.region }}
...
-  {{- else if and .Values.azure.useManagedIdentityExtension (not .Values.azure.userAssignedIdentityID) }}
-  "useManagedIdentityExtension": true
-  {{- end }}
-}
+{{/*
+The object shape in the `values.yaml` and `azure.json` are identical, except for the key "secretName".
+This takes the entire .Values.azure object, removes the helm-specific key "secretName", 
+and passes the rest of the object directly to the `toJson` function.
+*/}}
+{{ omit .Values.azure "secretName" | toJson }}
```
pinkfloydx33, line 178: If falsey values are a concern, we can remove those values from the copy of the dict before serializing it

```
{{- $copy := omit .Values.azure "secretName" }} 
{{- range $k, $v := $copy }} 
{{- if empty $v }} 
{{- $_ := unset $copy $k }} 
{{- end }} 
{{- end }} 
{{ $copy | toJson }} 
```

Typed from my phone so may not be perfect, but you get the idea

MaxAnderson95: I considered that at first, but I thought about what happens if an additional `bool` property was added to the `.Values.azure` object in the future that has a default value of `true`, but the user explicitly sets it to `false`. This would strip the entire key out of the object. Maybe i'm overthinking it?

migruiz4 (commented):
Hi there!

Thank you for your contribution @MaxAnderson95! I agree that using `toJson` here is the way to go, but I'm worried about transforming the entire `.Values.azure` into `azure.json`.

In the future, extra values may be added under `.Values.azure` that we may not want to add into the azure.json and cause issues for existing deployments by accident.

I would rather keep logic similar to how it currently is:

- Build a dict from scratch.
- Add key-values individually.
- Then convert the dict using `toJson` (instead of printing it line by line as it currently is).

It is not as pretty, but it would give us better control over what is included in the final Json. 
If external-dns adds a new supported field for the `azure.json`, then we just need to add a new value to the values.yaml and update the helper.

bitnami/external-dns/templates/_helpers.tpl
pinkfloydx33: Oops... Good point about the explicitly false values. That's what I get for posting at 1am

MaxAnderson95:
@migruiz4 can you re-review this?

MaxAnderson95:
Sorry for the notification spam. I screwed up trying to fix a merge conflict. Anyway I think we're good now.

MaxAnderson95:
@migruiz4 sorry to bother, but do you mind re-reviewing the changes you requested? I keep having to re-base this PR because changes are being made in other branches.

PR #34022, opened 2025-05-31
[bitnami/minio] fix: apiIngress tls values were ignored

Idotno-1:
### Description of the change

I found out that the following values were not taken into consideration in the templates. This change generates manifests based on those values by modifying the file `templates§tls-secrets.yaml`

```
# previously ignored values
minio:
  apiIngress:
    selfSigned: true # ignored
    secrets: [...] # secrets not created
```

### Benefits

API Ingress TLS secrets are now created. 

### Possible drawbacks

- If the console and API ingress wish to use the same tls secret, they will need to each provide the values, as the secrets name are not the same (`<hostname>-tls` vs `<hostname>-api-tls`).

- For existing users using both ingress with the same tls secrets, this would be breaking
```
# This config currently works because the console and API ingress use the same secret name `<hostname>-tls`
minio:
  ingress:
    # [...]
    hostname: minio.local

    tls: true

    secrets:
    - name: minio.local-tls
      key: "xxx"
      certificate: "xxx"

  apiIngress:
    # [...]
    hostname: minio.local
    path: api/

    tls: true
```

To resolve both of these drawbacks, we could replace the `-api-tls` suffix by the old `-tls` suffix only. However that would break in the following case as we would have two Secret manifest with the same name:

```
minio:
  ingress:
    hostname: minio.local
    tls: true
    selfSigned: true
    # [...]
  apiIngress:
    hostname: minio.local
    tls: true
    selfSigned: true
    # [...]
```

### Testing

With a custom local file `values.more.yaml`:
```
ingress:
  enabled: true

  tls: true

  selfSigned: true

  secrets:
    - name: custom.local-tls
      key: "example"
      certificate: "example"

apiIngress:
  enabled: true

  tls: true

  selfSigned: true

  secrets:
    - name: custom.local-api-tls
      key: "example"
      certificate: "example"
```

I ran a this before and after, and compared the results: 

```bash
cd bitnami/minio/
helm dependency build
helm template . -f values.more.yaml --output-dir /tmp/out
cat /tmp/out/minio/templates/tls-secrets.yaml
```

### Checklist


Idotno-1:
Due to #34037, this PR might be out-of-date now

Idotno-1:
Hello @fmulero, this looks solved ! Have a great day and thanks maintaining theses charts !

Issue #34255, opened 2025-06-08
Allow conditionally removing spec.replicas without creating HPA


nfriday:
### Name and Version

bitnami/thanos

### What is the problem this feature will solve?

Thanos deployments/statefulsets are templated to include the `spec.replicas` field unless autoscaling is enabled via, for example `.Values.receive.autoscaling.enabled`.

e.g. https://github.com/bitnami/charts/blob/fd5e9ffa796b16bb3b55f570cea32915304fc3d1/bitnami/thanos/templates/receive/statefulset.yaml#L21-L23

However, specifying auto-scaling also creates a corresponding HPA resource, e.g. https://github.com/bitnami/charts/blob/fd5e9ffa796b16bb3b55f570cea32915304fc3d1/bitnami/thanos/templates/receive/hpa.yaml#L6

It would be useful to disable the `spec.replicas` field **without** creating a HPA in scenarios where users of the chart wish to supply their own HPA (via KEDA, for example).

In particular in the case of Thanos ingesting receivers, users (like me!) may want to use prometheus metrics (via a KEDA ScaledObject HPA) to scale their ingesting receiver StatefulSet rather than the chart-provided HPA.

### What is the feature you are proposing to solve the problem?

Provide Helm values option, e.g. `.Values.receive.autoscaling.createHorizontalPodAutoscaler` defaulted to `true` (to ensure it is a non-breaking change).

### What alternatives have you considered?

I have considered using Kustomize to patch out the field post-rendering for my specific use case, but this would be more convenient!

carrodher:
Thank you for opening this issue and submitting the associated Pull Request. Our team will review and provide feedback. Once the PR is merged, the issue will automatically close.

Your contribution is greatly appreciated!

nfriday:
Commenting as this is still desired. Is there anything I can adjust on the PR to help get this in?

carrodher:
Sorry for the delay, due to other initiatives in the team, we are taking a bit more than usual to review PRs. I just removed the `stale` label to prevent the bot from closing the issue while it is queued for review.

nfriday:
Resolved by: https://github.com/bitnami/charts/pull/34256

PR #34256, opened 2025-06-08
[bitnami/thanos] Conditional creation of HorizontalPodAutoscaler resources for Thanos components

nfriday:
### Description of the change

Conditionally specify `replicas` field on scalable Deployment/StatefulSet resources for Thanos components based on new per-component `externalAutoscaler` values.

Values are defaulted to `false` to ensure this is a non-breaking change.

### Benefits

Allow users to omit spec.replicas from Deployment/StatefulSet manifests and bring their own scaling mechanisms (e.g. a KEDA controlled HPA for scaling using Prometheus metrics).

### Possible drawbacks

n/a

### Applicable issues

https://github.com/bitnami/charts/issues/34255

### Additional information

n/a

### Checklist


migruiz4:
Hi @nfriday,

Thank you for your contribution and sorry for the late response.

I'm sorry but I think that adding an extra value  `*.createHorizontalPodAutoscaler` shouldn't be the way to go in this case, since the final goal would be to omit the `spec.replicas` from the statefulsets.

The `autoscaling.enabled` already does the job of whether the HPA should be created or not.

I would instead suggest something like:
```
  {{- if not .Values.ruler.autoscaling.enabled .Values.ruler.externalAutoscaler }}
  replicas: {{ .Values.ruler.replicaCount }}
  {{- end }}
```

I feel this way the feature would be more simple, intuitive and easier to maintain.

The value name would be trivial, I thought of `externalAutoscaler` since we usually reference components not managed by the chart as `external`, e.g. externalDatabase.

In the values yaml, description should say something like.
```
## @param ruler.externalAutoscaler If set to true, enable the management of pods autoscaling using external sources, e.g. KEDA autoscaling. The value ruler.replicaCount will be ignored when set to true.
##
externalAutoscaler: false
```

nfriday:
Thanks @migruiz4, and please do not apologize. That you are generous enough to respond at all is incredibly welcome.

I agree with your feedback and have updated this PR accordingly.

bitnami/thanos/values.yaml
```diff
@@ -4873,13 +4885,15 @@ receiveDistributor:
   ## @param receiveDistributor.autoscaling.maxReplicas Maximum number of Thanos Receive Distributor replicas
   ## @param receiveDistributor.autoscaling.targetCPU Target CPU utilization percentage
   ## @param receiveDistributor.autoscaling.targetMemory Target Memory utilization percentage
+  ## @param receive.externalAutoscaler Specifies whether the Thanos Receive Distributor deployment is scaled by an external autoscaler
```
migruiz4: Small typo that is preventing the automatic README generation

```suggestion
  ## @param receiveDistributor.externalAutoscaler Specifies whether the Thanos Receive Distributor deployment is scaled by an external autoscaler
```

PR #34297, opened 2025-06-10
[template] move apiVersion on first element (or kind to second)

alagoutte:
Minor change but align all template to have apiVersion on the first element (or Kind to second...)

help when diff template


fmulero (approved):
Thanks @alagoutte for this contribution.

LGTM

PR #34562, opened 2025-06-19
[bitnami/kibana] Support TLS and Authentication in Kibana dashboard import script

### LINKED ISSUE #34366: Kibana Init Script Fails to Import Saved Objects Due to HTTP/HTTPS Mismatch When TLS is Enabled
### Name and Version

bitnami/elasticsearch 22.0.4

None

1. Deploy Kiabana with TLS enabled.
2. Create a dashboard and export it.
3. Create a configmap with the dashboard information.
4. Add the configmap name to the savedObjects.configmap value.
5. Kibana will init, but it won't load the savedObjects

Below is my values.yaml section for kibana
```yaml
kibana:
  enabled: true
  fullnameOverride: "monitoring-kibana" 
  ## @param updateStrategy.type Set up update strategy for Kibana installation.
  ## Set to Recreate if you use persistent volume that cannot be mounted by more than one pods to makesure the pods are destroyed first.
  updateStrategy:
    type: Recreate    
  ingress:
    enabled: true
    pathType: Prefix
    hostname: xx.yy.zz
    extraTls:
    - hosts:
        - xx.yy.zz
      secretName: wildcardcert-tls-secret   
    ingressClassName: "nginx"  
    annotations:
      nginx.ingress.kubernetes.io/rewrite-target: /
      nginx.ingress.kubernetes.io/backend-protocol: HTTPS
 
  extraEnvVars:
    - name: "NODE_OPTIONS"
      value: "--max-old-space-size=756"
  elasticsearch:
    hosts:
      - 'ofs-platform-service-elasticsearch'
    port: '9200'
    security:
      auth:
        enabled: true
        createSystemUser: true   
        # default in the elasticsearch chart is elastic

        kibanaUsername: "elastic"
        elasticsearchPasswordSecret: 'ofs-platform-service-elasticsearch'
      tls:
        # Instruct kibana to connect to elastic over https
        enabled: true
        # Bit of a catch 22, as you will need to know the name upfront of your release

        existingSecret: monitoring-elasticsearch-coordinating-crt # or just 'elasticsearch-coordinating-crt' if the release name happens to be 'elasticsearch'
        # # As the certs are auto-generated, they are pemCerts so set to true

        usePemCerts: true    
  tls:
    enabled: true
    # AutoGenerate TLS certs for kibana

    autoGenerated: true    

  extraConfiguration:
    xpack:
      encryptedSavedObjects:
        encryptionKey: 7884134987903fcc6692e072980304ab386b62fb1b22f7c5f689856cf7481f4e
    monitoring:
      ui:
        ccs:
          enabled: false 
  ## @param forceInitScripts Force execution of init scripts
  ##
  forceInitScripts: true        
  service:
    ## @param service.ports.http Kubernetes Service port
    ##
    ports:
      http: 5601
  savedObjects:        
    ## @param savedObjects.configmap Configmap containing NDJSON files to be imported during Kibana initialization (evaluated as a template)
    ##
    configmap: saved-dashboard
    ## @param savedObjects.overwrite Overwrite saved objects
    overwrite: true   
```

savedObjects should appear in kibana

### What do you see instead?

savedObjects are not loaded

### Additional information

The problem seems to be that when TLS is enabled for Kibana, the background Kibana instance (started by kibana_custom_init_scripts) listens on **https://localhost:5601**, but the curl command in the **import-saved-objects.sh** script attempts to connect via http://localhost:5601 (by default, if https:// is not specified). This leads to the "Empty reply from server" error because curl is speaking plain HTTP to an HTTPS endpoint.

As a temporary solution, I have modified the curl command in my import-saved-objects.sh script to explicitly use https and include `--insecure` to bypass certificate validation for this localhost connection during initialization. And it worked for me.

Here's the modified curl command:

```
curl -s --connect-timeout 60 --max-time 60 -XPOST https://localhost:5601/api/saved_objects/_import -H 'kbn-xsrf:true' --form file=@${f} --insecure
```
This allowed the saved objects to be imported successfully during initialization.

For context, I'm generating this custom script dynamically via a Helm ConfigMap. This allows me to configure the curl command to correctly use https and the `--insecure` flag based on the `elasticsearch.kibana.tls.enabled` and `elasticsearch.kibana.tls.autoGenerated` settings in my Helm chart values.

For full context, my deployment utilizes a nested Helm chart structure: Elasticsearch is deployed as a subchart, and Kibana is a subchart of that Elasticsearch subchart. The necessary ConfigMap resources are dynamically generated using the following Helm templates, which are defined in the main chart's templates, not within the Kibana or Elasticsearch subcharts themselves.

For reference:

First, here's the ConfigMap that holds the actual .ndjson dashboard files:
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: {{ .Values.importSavedDashboard.configmapName }}
data:
{{- $files := .Files.Glob (printf "%s*.ndjson" .Values.importSavedDashboard.path) }}
{{- range $path, $file := $files }}
  {{ base $path }}: |
{{ toString $file | indent 4 }}
{{- end }}
```
And here's the ConfigMap that defines the custom-import-saved-objects.sh script, which is responsible for importing those dashboards. As mentioned previously, this script dynamically configures curl to use https and includes the `--insecure` flag based on the TLS settings in my chart values, addressing the issue of curl attempting a plain HTTP connection to an HTTPS endpoint.

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: custom-import-saved-objects
  labels: {{- include "common.labels.standard" ( dict "customLabels" .Values.commonLabels "context" $ ) | nindent 4 }}
  {{- end }}
data:
  {{- $useHttps := false }}
  {{- $insecureSkipVerify := false }}

{{- /* Check if Kibana's frontend TLS is enabled directly via .Values.elasticsearch.kibana.tls.enabled */}}
  {{- if and 
      (hasKey .Values.elasticsearch "kibana") 
      (hasKey .Values.elasticsearch.kibana "tls") 
      (hasKey .Values.elasticsearch.kibana.tls "enabled") 
      (eq .Values.elasticsearch.kibana.tls.enabled true) 
  }}
    {{- $useHttps = true }}
    {{- /* If autoGenerated is true, it implies self-signed, so we'll need -k */}}
    {{- if and 
        (hasKey .Values.elasticsearch.kibana.tls "autoGenerated") 
        (eq .Values.elasticsearch.kibana.tls.autoGenerated true) 
    }}
      {{- $insecureSkipVerify = true }}
    {{- end }}
  {{- end }}


  {{- $protocol := "http" }}
  {{- if $useHttps }}
  {{- $protocol = "https" }}
  {{- end }}

  {{- $kibanaPort := int .Values.elasticsearch.kibana.containerPorts.http }}
  {{- /* Assuming 5601 is used for HTTPS as well */}}

  {{- $savedObjectsUrl := printf "%s://localhost:%d/api/saved_objects/_import%s" $protocol $kibanaPort (ternary "?overwrite=true" "" .Values.elasticsearch.kibana.savedObjects.overwrite) }}

  custom-import-saved-objects.sh: |
    #!/bin/bash
    echo "==> Saved objects import"

    # Add curl flags for TLS handling
    CURL_TLS_FLAGS=""
    {{- if $useHttps }}
      {{- if $insecureSkipVerify }}
    CURL_TLS_FLAGS="-k" # Use -k for self-signed or untrusted certs
      {{- else }}
    # If not insecure, you'd ideally use --cacert. This script does not include logic
    # to find or mount the CA cert here. If your certs are from a known CA, this might work
    # without -k, but for internal communication, -k is often pragmatic for auto-generated.
      {{- end }}
    {{- end }}


    # Add authentication flags
    CURL_AUTH_FLAGS=""
    # This checks for the elastic password from the standard Bitnami file mount
    if [[ -f "/opt/bitnami/kibana/secrets/elasticsearch-password" ]]; then
      # Read the password, stripping any trailing newline characters
      KIBANA_PASSWORD=$(< "/opt/bitnami/kibana/secrets/elasticsearch-password" tr -d '\n')
      echo "DEBUG: Found elasticsearch-password file."
      # Correctly form CURL_AUTH_FLAGS without problematic extra quotes
      CURL_AUTH_FLAGS="-u elastic:${KIBANA_PASSWORD}"

    else
      echo "ERROR: elasticsearch-password file NOT found at /opt/bitnami/kibana/secrets/elasticsearch-password"
    fi

    {{- if .Values.elasticsearch.kibana.savedObjects.urls }}
    {{- $totalURLs := len .Values.elasticsearch.kibana.savedObjects.urls }}
    echo "Total saved objects NDJSON URLs to import: {{ $totalURLs }}"
    {{- range $i, $url := .Values.elasticsearch.kibana.savedObjects.urls }}
    echo "Importing saved objects from NDJSON in url {{ add $i 1 }} out of {{ $totalURLs }}: {{ $url }}"
    download_tmp_file="$(mktemp)"
    curl "{{ $url }}" > "${download_tmp_file}.ndjson"
    curl -s --connect-timeout 60 --max-time 60 -XPOST {{ $savedObjectsUrl }} -H 'kbn-xsrf:true' --form file=@"${download_tmp_file}.ndjson" ${CURL_TLS_FLAGS} ${CURL_AUTH_FLAGS}
    {{- end }}
    {{- end }}
    {{- if .Values.elasticsearch.kibana.savedObjects.configmap }}
    echo "Searching for dashboard NDJSON files from ConfigMap mounted in /bitnami/kibana/saved-objects"
    ndjson_file_list_tmp="$(mktemp)"
    find /bitnami/kibana/saved-objects -type f -regex ".*\.ndjson" > $ndjson_file_list_tmp
    while read -r f; do
      case "$f" in
        *.ndjson)
          echo "Importing $f"
          curl -s --connect-timeout 60 --max-time 60 -XPOST {{ $savedObjectsUrl }} -H 'kbn-xsrf:true' --form file=@"${f}" ${CURL_TLS_FLAGS} ${CURL_AUTH_FLAGS}
          ;;
        *)
          echo "Ignoring $f"
          ;;
      esac
    done < $ndjson_file_list_tmp
    {{- end }}
    echo "==> End of Saved objects import"
```
Below is the relevant snippet from my values.yaml
```yaml
##monitoring
importSavedDashboard:
  configmapName: &importSavedDashboardConfigmap "kibana-dashboards"
  path: dashboards/
elasticsearch:
  kibana:
    savedObjects:        
      configmap: *importSavedDashboardConfigmap
      overwrite: true   
    extraVolumes: 
      - name: custom-saved-objects-init-scripts
        configMap:
          name: custom-import-saved-objects
          defaultMode: 0755
    extraVolumeMounts:
      - name: custom-saved-objects-init-scripts
        mountPath: /docker-entrypoint-initdb.d/custom-saved-objects-import/
```

I hope this information was helpful, and that the issue will be resolved by Bitnami as soon as possible.

aparnams1999:
### Description of the change

This PR updates the Kibana import-saved-objects.sh init script to properly support environments where TLS is enabled and authentication is required. It dynamically chooses between HTTP and HTTPS based on .Values.tls.enabled and adds optional authentication via user credentials (either from mounted password files or environment variables).

### Benefits

- Ensures the saved objects import succeeds in secured Kibana environments.
- Supports both TLS and non-TLS deployments.
- Adds compatibility with authentication-protected Kibana instances.
- Improves robustness and flexibility of the init script.

### Possible drawbacks

- If incorrect credentials are provided, the import will fail.
- The -k (insecure) option is used for TLS, which skips cert verification — suitable for internal dev environments but not ideal for production.

### Applicable issues

- fixes #34366

### Checklist

### DIFF
--- a/bitnami/kibana/CHANGELOG.md
+++ b/bitnami/kibana/CHANGELOG.md
@@ -1,8 +1,12 @@
 # Changelog
 
-## 12.1.3 (2025-06-24)
+## 12.1.4 (2025-06-26)
 
-* [bitnami/kibana] :zap: :arrow_up: Update dependency references ([#34604](https://github.com/bitnami/charts/pull/34604))
+* [bitnami/kibana] Support TLS and Authentication in Kibana dashboard import script ([#34562](https://github.com/bitnami/charts/pull/34562))
+
+## <small>12.1.3 (2025-06-24)</small>
+
+* [bitnami/kibana] :zap: :arrow_up: Update dependency references (#34604) ([ae475c8](https://github.com/bitnami/charts/commit/ae475c80a52127cbdab36276ed2d071c4495c785)), closes [#34604](https://github.com/bitnami/charts/issues/34604)
 
 ## <small>12.1.2 (2025-06-13)</small>
 


bitnami/kibana/templates/saved-objects-configmap.yaml
```diff
@@ -14,34 +14,55 @@ metadata:
...
-  {{- $savedObjectsUrl := printf "localhost:%d%s/api/saved_objects/_import%s" (int .Values.containerPorts.http) (include "kibana.basePath" .) (ternary "?overwrite=true" "" .Values.savedObjects.overwrite) }}
+  {{- $savedObjectsUrl := printf "%s://localhost:%d%s/api/saved_objects/_import%s" (ternary "https" "http" .Values.tls.enabled) (int .Values.containerPorts.http) (include "kibana.basePath" .) (ternary "?overwrite=true" "" .Values.savedObjects.overwrite) }}
   import-saved-objects.sh: |
     #!/bin/bash
     echo "==> Saved objects import"
+    CURL_TLS_FLAGS=""
+    CURL_AUTH_FLAGS=""
+
+    {{- if .Values.tls.enabled }}
+    CURL_TLS_FLAGS="-k"
```
fmulero: Instead of not verifying the peer connection, Have you considered this option?

```suggestion
    CURL_TLS_FLAGS="--cacert /opt/bitnami/kibana/config/certs/server/ca.crt"
```

fmulero (changes requested):
Thanks a lot @aparnams1999 for taking care of this issue and contributing with this PR. We appreciate it a lot.

I've just left a comment with a suggestion. Could you take it a look?

bitnami/kibana/templates/saved-objects-configmap.yaml
aparnams1999: I did initially try using the certificate with --cacert, but encountered the following error:
```
* Server certificate:
*  subject: CN=monitoring-kibana
*  subjectAltName does not match localhost
* SSL: no alternative certificate subject name matches target host name 'localhost'
```
This happens because the TLS certificate is issued for `monitoring-kibana`, not `localhost`, so the hostname verification fails.

As a temporary workaround, I used `-k` to bypass certificate validation. However, if `.Values.tls.enabled` is `true`, we can and should avoid -k by calling Kibana using its service name (`{{ include "common.names.fullname" . }}`) instead of `localhost`.

fmulero (approved):
LGTM. Thanks a lot for your contribution!

PR #34583, opened 2025-06-23
template(values).yaml: remove none value from preset

alagoutte:
None is not a valid value for ressource preset (see _resources.tpl file)


alagoutte:
Update for add missing sign off

May be need also to update [templates/deployments.yaml](https://github.com/bitnami/charts/blob/main/template/CHART_NAME/templates/deployment.yaml#L114) for don't check none value...

migruiz4:
Hi @alagoutte,

The resourcesPreset "none" option is not defined in the _resources.tpl, but each manifest resources definition includes this:
```yaml
  {{- if .Values.defaultInitContainers.prepareConfig.resources }}
  resources: {{- toYaml .Values.defaultInitContainers.prepareConfig.resources | nindent 4 }}
  {{- else if ne .Values.defaultInitContainers.prepareConfig.resourcesPreset "none" }}
  resources: {{- include "common.resources.preset" (dict "type" .Values.defaultInitContainers.prepareConfig.resourcesPreset) | nindent 4 }}
  {{- end }}
```

PR #34641, opened 2025-06-25
[bitnami/valkey] Fix key on condition to create PodDisruptionBudget

rastut:
### Description of the change

Fixes a bug in the Helm template condition that prevents the PodDisruptionBudget (PDB) from being rendered when using the `standalone` architecture.

The condition incorrectly references `.Values.primary.count`, which is not defined in the `values.yaml`. It has been updated to use `.Values.primary.replicaCount`, which is the correct and documented field for specifying the number of primary replicas.

### Benefits

Ensures that the PDB is rendered as expected when `primary.pdb.create` is `true` and `primary.replicaCount > 0`.

### Possible drawbacks

If any users were unknowingly relying on `primary.count` (despite it not being documented), their configuration might stop working. However, this is unlikely and would indicate misalignment with the intended values schema.

### Checklist

### DIFF
--- a/bitnami/valkey/CHANGELOG.md
+++ b/bitnami/valkey/CHANGELOG.md
@@ -1,8 +1,12 @@
 # Changelog
 
-## 3.0.15 (2025-06-18)
+## 3.0.16 (2025-06-25)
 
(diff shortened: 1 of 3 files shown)


migruiz4 (approved):
Thank you for your contribution @rastut! Looks good to me!

PR #34701, opened 2025-06-28
[bitnami/milvus] Add GCS support to _helpers.tpl

asalfaifi:
### Description of the change

This pull request modifies the `_helpers.tpl` file within the `milvus` Helm chart templates. The primary purpose of this change is to enable support for Google Cloud Storage (GCS) as a backend for Milvus, specifically by adapting the `wait-for-s3` init container's logic.

The `wait-for-s3` init container, when checking object storage availability, currently expects the `RequestID` field within the body of the response from `curl` requests for AWS S3. However, GCS responses do not include this field in the body; instead, they provide a similar identifier (`x-guploader-uploadid`) within the response header. This change updates the templating logic to correctly parse the GCS-specific header when GCS is configured, ensuring the init container successfully validates the storage connection.

### Benefits

* **Adds GCS Compatibility:** Extends Milvus's object storage options to include Google Cloud Storage, providing more flexibility for users.
* **Improved Storage Provisioning:** Ensures the `wait-for-s3` init container correctly identifies successful GCS connections, preventing startup failures when using GCS.
* **Wider Cloud Adoption:** Allows Milvus users to seamlessly integrate with their existing GCS infrastructure.

### Possible drawbacks

* No known drawbacks, as the change is conditional and only applies when GCS is explicitly configured. It maintains existing AWS S3 compatibility.

### Applicable issues

- couldn't find an open issue bitnami/charts repository

This change specifically targets the logic used by the `wait-for-s3` init container. No new values are introduced in `values.yaml` as the logic leverages existing configuration parameters to determine whether AWS S3 or GCS is being used.

### Checklist

    *Note: The `version` field in `Chart.yaml` has been incremented from `15.0.8` to `15.1.0` to reflect the addition of new functionality.*
    *Note: This change does not introduce new configurable variables, thus no updates to `values.yaml` or `README.md` for variable documentation are required.*
    *Note: The PR title will be `[bitnami/milvus] Add GCS support to _helpers.tpl`.*
    *Note: This is automatically handled by GitHub's UI when "Sign off and commit changes" is selected.*

### DIFF
--- a/bitnami/milvus/Chart.yaml
+++ b/bitnami/milvus/Chart.yaml
@@ -51,4 +51,4 @@ maintainers:
 name: milvus
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/milvus
-version: 15.0.8
+version: 15.1.0
--- a/bitnami/milvus/templates/_helpers.tpl
+++ b/bitnami/milvus/templates/_helpers.tpl
@@ -824,6 +824,8 @@ Init container definition for waiting for the database to be ready
 
           if curl --max-time 5 "${s3_host}" $params_cert | grep "RequestId"; then
              return 0
+          elif curl --head --max-time 5 "${s3_host}" $params_cert | grep "x-guploader-uploadid"; then
+             return 0
           else
              return 1
           fi


bitnami/milvus/templates/_helpers.tpl
```diff
@@ -824,6 +824,8 @@ Init container definition for waiting for the database to be ready
 
           if curl --max-time 5 "${s3_host}" $params_cert | grep "RequestId"; then
              return 0
+          elif curl --head --max-time 5 "${s3_host}" $params_cert | grep "x-guploader-uploadid"; then
+             return 0
```
juan131: I'd use an OR condition in the grep instead:

```suggestion
          if curl --max-time 5 "${s3_host}" $params_cert | grep -E "RequestId|x-guploader-uploadid"; then
             return 0
```

bitnami/milvus/Chart.yaml
```diff
@@ -51,4 +51,4 @@ maintainers:
 name: milvus
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/milvus
-version: 15.0.8
+version: 15.1.0
```
juan131: Please bump patch version instead:
```suggestion
version: 15.0.9
```

juan131 (changes requested):
Thanks so much for your contribution! Please check my comments

asalfaifi:
Thank you for your guidance @juan131 
I have committed to what you have pointed out to me to modify 👍

asalfaifi:
I have checked the change to the if statement, and it will not work
The `elif` is meant to check the header of the curl request instead of the body

PR #32442, opened 2025-03-13
[bitnami/opensearch] allow to configure trafficDistribution

rossifumax:
Hello !

### Description of the change

Add trafficDistribution feature on svc-headless of Opensearch. 
This feature is available since k8s v1.31.

### Benefits

Take advantage of the new feature trafficDistribution.
cf. https://kubernetes.io/docs/concepts/services-networking/service/#traffic-distribution

### Possible drawbacks

Normally, no drawbacks.

### Applicable issues

- fixes #31730

### Checklist

- [X] Chart version bumped in `Chart.yaml` according to [semver](http://semver.org/). This is *not necessary* when the changes only affect README.md files.
- [X] Variables are documented in the values.yaml and added to the `README.md` using [readme-generator-for-helm](https://github.com/bitnami/readme-generator-for-helm)
- [X] Title of the pull request follows this pattern [bitnami/<name_of_the_chart>] Descriptive title
- [X] All commits signed off and in agreement of [Developer Certificate of Origin (DCO)](https://github.com/bitnami/charts/blob/main/CONTRIBUTING.md#sign-your-work)

PR #34773, opened 2025-07-02
[bitnami/memcached] Allow trafficDistribution field to be specified

Asgavar:
### Description of the change

Similarly to https://github.com/bitnami/charts/pull/32442 and https://github.com/bitnami/charts/pull/32443, this MR introduces possibility to set `Service.spec.trafficDistribution` field in the `memcached` chart: https://kubernetes.io/docs/concepts/services-networking/service/#traffic-distribution

### Possible drawbacks

Virtually none - we're not setting it at all by default, so potential caveats and Kubernetes version requirements are completely up to the chart's end users.

### Checklist

### DIFF
--- a/bitnami/memcached/templates/service.yaml
+++ b/bitnami/memcached/templates/service.yaml
@@ -21,6 +21,9 @@ spec:
   {{- if .Values.service.sessionAffinityConfig }}
   sessionAffinityConfig: {{- include "common.tplvalues.render" (dict "value" .Values.service.sessionAffinityConfig "context" $) | nindent 4 }}
   {{- end }}
+  {{- if .Values.service.trafficDistribution }}
+  trafficDistribution: {{ .Values.service.trafficDistribution }}
+  {{- end }}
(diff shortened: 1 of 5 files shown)


bitnami/memcached/templates/service.yaml
```diff
@@ -21,6 +21,9 @@ spec:
   {{- if .Values.service.sessionAffinityConfig }}
   sessionAffinityConfig: {{- include "common.tplvalues.render" (dict "value" .Values.service.sessionAffinityConfig "context" $) | nindent 4 }}
   {{- end }}
+  {{- if not (empty .Values.service.trafficDistribution) }}
```
jotamartos: What do you think about this change? As it expects a string, it simplifies the code

```suggestion
  {{- if .Values.service.trafficDistribution }}
```

Asgavar: Ah right, sure thing - applied in the current revision

jotamartos (approved):

Issue #27549, opened 2024-06-26
[bitnami/redis] Long release name breaks statefulset pods


kkopachev:
### Name and Version

bitnami/redis 19.5.5

### What architecture are you using?

None

### What steps will reproduce the bug?

```
helm install aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa oci://registry-1.docker.io/bitnamicharts/redis --set architecture=standalone --set master.persistence.enabled=false
```

where release name is 41 character long

### Are you using any custom parameters or values?

--set architecture=standalone --set master.persistence.enabled=false

### What is the expected behavior?

Expect either error on `helm install` time or release name truncated further to account for longer pod names in statefulset 

### What do you see instead?

Helm deploys successfully, but pods in replica set fail to start with 
```
create Pod aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa-redis-master-0 in StatefulSet aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa-redis-master failed error: 
Pod "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa-redis-master-0" is invalid: metadata.labels: Invalid value: "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa-redis-master-df5fb664f": must be no more than 63 characters
```

PR #34803, opened 2025-07-04
[bitnami/redis] Ensure that Redis resources are less than 63 chars

mboutet:
### Description of the change

Make sure to truncate the Statefulsets and Services' name to 63 characters.

### Benefits

This guarantees that all the resources are created successfully.

### Possible drawback

None, as I made sure to truncate the original `fullname` while keeping the differentiating suffix as-is (to prevent resource name clash in case the `fullname` is really long).

### Applicable issues

- fixes https://github.com/bitnami/charts/issues/27549

### Additional information

I bumped the major version as it could technically cause resources to be recreated due a change in the `metadata.name` of some Statefulsets and Services. In practice, however, it should not happen as the truncation only happens if really needed, so users already using the chart without issues won't get any truncation. On the other hand, the major bump is also warranted due to the complexity of the chart and the regression risk of this PR even though I tried to not miss anything.

### Checklist

### DIFF
--- a/bitnami/redis/templates/_helpers.tpl
+++ b/bitnami/redis/templates/_helpers.tpl
@@ -6,6 +6,45 @@ SPDX-License-Identifier: APACHE-2.0
 {{/* vim: set filetype=mustache: */}}
 
 {{/*
+Return the proper Redis master fullname.
+
+The 0000000000 is to take into account the 10-charaters hash that the StatefulSet
+appends to the name of the StatefulSet in the pod's controller-revision-hash label.
+A label value can only have 63 characters.
+*/}}
+{{- define "redis.master.fullname" -}}
+{{- printf "%s-%s" ((include "common.names.fullname" .) | trunc ((sub 63 ("-master-0000000000" | len)) | int) | trimSuffix "-") "master" -}}
+{{- end -}}
+
+{{/*
+Return the proper Redis replicas fullname
+*/}}
+{{- define "redis.replicas.fullname" -}}
+{{- printf "%s-%s" ((include "common.names.fullname" .) | trunc ((sub 63 ("-replicas-0000000000" | len)) | int) | trimSuffix "-") "replicas" -}}
+{{- end -}}
+
+{{/*
+Return the proper Redis sentinel fullname
+*/}}
+{{- define "redis.sentinel.fullname" -}}
+{{- printf "%s-%s" ((include "common.names.fullname" .) | trunc ((sub 63 ("-node-0000000000" | len)) | int) | trimSuffix "-") "node" -}}
+{{- end -}}
+
+{{/*
+Return the proper Redis headless fullname
+*/}}
+{{- define "redis.headless.fullname" -}}
+{{- printf "%s-%s" ((include "common.names.fullname" .) | trunc ((sub 63 ("-headless" | len)) | int) | trimSuffix "-") "headless" -}}
+{{- end -}}
+
+{{/*
+Return the proper Redis metrics fullname
+*/}}
+{{- define "redis.metrics.fullname" -}}
+{{- printf "%s-%s" ((include "common.names.fullname" .) | trunc ((sub 63 ("-metrics" | len)) | int) | trimSuffix "-") "metrics" -}}
+{{- end -}}
+
+{{/*
 Return the proper Redis image name
 */}}
 {{- define "redis.image" -}}
--- a/bitnami/redis/Chart.yaml
+++ b/bitnami/redis/Chart.yaml
@@ -39,4 +39,4 @@ maintainers:
 name: redis
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/redis
-version: 21.2.10
+version: 22.0.0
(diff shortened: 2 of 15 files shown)


mboutet:
Testing on my end, but encountering few issues with my changes, so I'm setting the PR as draft for now.

EDIT: I think it's good now.

mboutet:
~Actually, I think there's still an issue with one label having a too long value, but this label comes from the common bitnami chart, so I would have to fix it there :/ ~

It was the value of the `controller-revision-hash` label in the end.

bitnami/redis/templates/_helpers.tpl
```diff
@@ -5,6 +5,45 @@ SPDX-License-Identifier: APACHE-2.0
...
+{{- define "redis.replicas.fullname" -}}
+{{- printf "%s-%s" ((include "common.names.fullname" .) | trunc ((sub 63 ("-replicas-0000000000" | len)) | int) | trimSuffix "-") "replicas" -}}
+{{- end -}}
+
+{{/*
+Return the proper Redis sentinel fullname
+*/}}
+{{- define "redis.sentinel.fullname" -}}
+{{- printf "%s-%s" ((include "common.names.fullname" .) | trunc ((sub 63 ("-node-0000000000" | len)) | int) | trimSuffix "-") "node" -}}
+{{- end -}}
```
mboutet, line 31: Since the `Service` also uses this helper, it could result in it being recreated for some users. Truncating the equivalent of `0000000000` is only strictly needed for the `Statefulset` in reality.

bitnami/redis/templates/_helpers.tpl
```diff
@@ -5,6 +5,45 @@ SPDX-License-Identifier: APACHE-2.0
...
+
+{{/*
+Return the proper Redis headless fullname
+*/}}
+{{- define "redis.headless.fullname" -}}
+{{- printf "%s-%s" ((include "common.names.fullname" .) | trunc ((sub 63 ("-headless" | len)) | int) | trimSuffix "-") "headless" -}}
+{{- end -}}
```
jotamartos, line 38: I like the idea and see it as a good improvement. However, all our solutions use a simpler method

```suggestion
{{/*
Return the proper Redis headless fullname
*/}}
{{- define "redis.headless.fullname" -}}
{{ printf "%s-headless" (include "common.names.fullname" .) | trunc 63 | trimSuffix "-" }}
{{- end -}}
```

I'd follow that approach here (and the other methods) too. 

However, the common.names.fullname function can be improve to truncate the strings the way you are suggesting. 

https://github.com/bitnami/charts/blob/main/bitnami/common/templates/_names.tpl#L26

You can update that function to accept another string to perform the "sub" method. That way, that function can be added to other charts as well and the whole community will benefit from that.

mboutet: Although I think it's a good idea, it would also change the signature of `common.names.fullname` in a backward incompatible manner, no? Unless conditional logic is added to `common.names.fullname` to accept both the current and new signature with the custom string to use to calculate the additional truncation. I don't mind doing that, but before I do, I just want to make sure if we're agreeing on the approach of having `common.names.fullname` handle both signatures.

jotamartos:
Sorry @mboutet, I didn't receive the notification about your comments. Will review everything again and update the PR

bitnami/redis/templates/_helpers.tpl
jotamartos: I simply suggested to improve the function to accept a new parameter to perform the "sub" method. The function won't do anything unless you set that new parameter. You can later update the function in the Redis' chart and increase the version to a new major so we avoid issues when updating between minor versions. Sounds good?

jotamartos:
Hi @mboutet,

I just updated the thread with more information. Let me know if it's clear now.

jotamartos:
Sorry @mboutet, I lost track of this ticket. Let me review it again and provide feedback

bitnami/redis/templates/_helpers.tpl
```diff
@@ -5,6 +5,45 @@ SPDX-License-Identifier: APACHE-2.0
...
 
+{{/*
+Return the proper Redis master fullname.
+
+The 0000000000 is to take into account the 10-charaters hash that the StatefulSet
+appends to the name of the StatefulSet in the pod's controller-revision-hash label.
+A label value can only have 63 characters.
+*/}}
+{{- define "redis.master.fullname" -}}
+{{- printf "%s-%s" ((include "common.names.fullname" .) | trunc ((sub 63 ("-master-0000000000" | len)) | int) | trimSuffix "-") "master" -}}
```
jotamartos, line 16: Current fullnames do not include the "0000000000" string, so it's not necessary to remove that

```
➜  redis git:(main) k get all                      
NAME                   READY   STATUS    RESTARTS   AGE
pod/redis-master-0     1/1     Running   0          2m29s
pod/redis-replicas-0   1/1     Running   0          2m28s
pod/redis-replicas-1   1/1     Running   0          91s
pod/redis-replicas-2   1/1     Running   0          51s

NAME                     TYPE        CLUSTER-IP     EXTERNAL-IP   PORT(S)    AGE
service/redis-headless   ClusterIP   None           <none>        6379/TCP   2m29s
service/redis-master     ClusterIP   10.0.92.51     <none>        6379/TCP   2m29s
service/redis-replicas   ClusterIP   10.0.203.103   <none>        6379/TCP   2m29s

NAME                              READY   AGE
statefulset.apps/redis-master     1/1     2m30s
statefulset.apps/redis-replicas   3/3     2m30s
```

Issue #34823, opened 2025-07-07
[bitnami/moodle] Helm value hostAliases in moodle deployment not used (Solution in additional information)


marhode:
### Name and Version

bitnami/moodle all versions

### What steps will reproduce the bug?

1. Setting the hostAliases value in the values.yaml file
2. Rolling out the Helm Chart
3. hostAliases value is present in the Helm values File, but not in the deployment

### Are you using any custom parameters or values?

Yes i used custom testing parameters:

hostAliases:
  - ip: "127.0.0.2"
    hostnames:
      - "dummy.local"
  - ip: "127.0.0.1"
    hostnames:
      - "dummy.test"

### What is the expected behavior?

We expect that the hostAliases value in the values.yaml will be used in the moodle deployment.

### What do you see instead?

The hostAliases value in the moodle deployment stays the same because it is hard coded in the moodle deployment template.

### Additional information

https://artifacthub.io/packages/helm/bitnami/moodle/27.0.0?modal=template&template=deployment.yaml
**Current implementation of the deployment template:**

```
42 {{- if .Values.hostAliases}}
43     hostAliases: {{- include "common.tplvalues.render" (dict "value" Values.hostAliases "context" $) | nindent 8 }}
44 {{- end }}
...
66     hostAliases:
67        - ip: "127.0.0.1"
68          hostnames:
69            - "status.localhost"
```

**The tested and working solution:**

```
{{- if .Values.hostAliases }}
      hostAliases: {{- include "common.tplvalues.render" (dict "value" .Values.hostAliases "context" $) | nindent 8 }}
{{- else }}
      hostAliases:
        - ip: "127.0.0.1"
          hostnames:
            - "status.localhost"
{{- end }}
```

javsalgar:
Hi! Thank you so much for reporting! Indeed, that should fix the issue, would you like to submit a PR with the solution?

PR #34839, opened 2025-07-08
[bitnami/moodle] Fix hostAliases Bug in Moodle deployment.yaml

marhode:
This change will fix the existing issue and Bug regarding the hostAliases value in the moodle Helm Chart.

### Applicable issues

- fixes https://github.com/bitnami/charts/issues/34823

### Checklist

### DIFF
--- a/bitnami/kubernetes-event-exporter/CHANGELOG.md
+++ b/bitnami/kubernetes-event-exporter/CHANGELOG.md
@@ -1,8 +1,12 @@
 # Changelog
 
-## 3.5.7 (2025-07-09)
+## 3.6.0 (2025-07-10)
 
-* [bitnami/kubernetes-event-exporter] :zap: :arrow_up: Update dependency references ([#34918](https://github.com/bitnami/charts/pull/34918))
+* [bitnami/kubernetes-event-exporter] Add tpl ability to .Values.Config ([#34855](https://github.com/bitnami/charts/pull/34855))
+
+## <small>3.5.7 (2025-07-09)</small>
+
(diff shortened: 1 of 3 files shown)


carrodher:

bitnami/moodle/templates/deployment.yaml
```diff
@@ -41,6 +41,11 @@ spec:
       automountServiceAccountToken: {{ .Values.automountServiceAccountToken }}
       {{- if .Values.hostAliases }}
       hostAliases: {{- include "common.tplvalues.render" (dict "value" .Values.hostAliases "context" $) | nindent 8 }}
+      {{- else }}
+      hostAliases:
+        - ip: "127.0.0.1"
+          hostnames:
+            - "status.localhost"
```
fmulero: I've been checking other solutions based on Apache, like Wordpress and the approach is slightly different. The deployment code is like this:
```suggestion
      {{- if .Values.hostAliases }}
      hostAliases: {{- include "common.tplvalues.render" (dict "value" .Values.hostAliases "context" $) | nindent 8 }}
      {{- end }}
```

And the default value for `hostAliases`  in the `values.yaml` file is:
```yaml
hostAliases:
  - ip: "127.0.0.1"
    hostnames:
      - "status.localhost"
```
https://github.com/bitnami/charts/blob/60c443a3bb96ee6ca41f920168a95d3a0744af89/bitnami/wordpress/values.yaml#L311-L319

Could we follow the same approach here?

fmulero (changes requested):
Good catch!!! Thanks a lot @marhode for your contribution!

I've just left a comment. Could you give it a glance?

bitnami/moodle/templates/deployment.yaml
marhode: Done

Issue #34853, opened 2025-07-08
[bitnami/ghost] subPath is ignored when usePasswordFiles: true


insekticid:
https://github.com/bitnami/charts/blob/b7eeced384731a9f23f6db2c494e39ed9d7f49bf/bitnami/ghost/templates/deployment.yaml#L330

there is bug when usePasswordFiles: true, you cannot set subPath for 'ghost-data' dir

carrodher:
Thank you for opening this issue and submitting the associated Pull Request. Our team will review and provide feedback. Once the PR is merged, the issue will automatically close.

Your contribution is greatly appreciated!

PR #34854, opened 2025-07-08
[bitnami/ghost] subPath fix

insekticid:
subPath fix

image

- Fixes https://github.com/bitnami/charts/issues/34853


carrodher:

insekticid:
@carrodher done in separate commit

bitnami/ghost/templates/deployment.yaml
```diff
@@ -327,6 +327,9 @@ spec:
               subPath: tmp-dir
             - name: ghost-data
               mountPath: /bitnami/ghost
+            {{- if (and (not .Values.usePasswordFiles) .Values.persistence.subPath) }}
```
fmulero: Why are you enforcing `.Values.usePasswordFiles == false` ? Is not enough with this:
```suggestion
            {{- if .Values.persistence.subPath }}
```

insekticid: Because if you set persistence.subPath: xyz and usePasswordFiles:false

You will get 

- name: ghost-data
  mountPath: /bitnami/ghost
  subPath: xyz
  subPath: xyz

fmulero: I've just seen what you said 🤯. There is typo in around the `ghost-secrets` mountPath, it doesn't need the subpath as far as I know, that value is only for the persistent volume. I think this piece of code should be removed:  https://github.com/bitnami/charts/blob/a23cddee5e484c4c10d7e5ff00f24958800dfae3/bitnami/ghost/templates/deployment.yaml#L337-L339

Once it's done, we don't need to check the `usePasswordFiles` value before adding the subPath to the persistent volume.

fmulero (changes requested):

fmulero (approved):
LGTM

PR #34968, opened 2025-07-09
[bitnami/kube-prometheus] add extraRbac

Poil:
### Description of the change

Allow extra rbac

### Benefits

Allow to poll specific endpoint like
https://docs.aws.amazon.com/eks/latest/userguide/view-raw-metrics.html

### Applicable issues

Fetch EKS control plane metrics https://docs.aws.amazon.com/eks/latest/userguide/view-raw-metrics.html

### Checklist


bitnami/kube-prometheus/Chart.yaml
```diff
@@ -51,4 +51,4 @@ maintainers:
 name: kube-prometheus
 sources:
 - https://github.com/bitnami/charts/tree/main/bitnami/kube-prometheus
-version: 11.2.11
+version: 11.2.12
```
juan131: Please bump minor version instead given you're adding a new parameter:
```suggestion
version: 11.3.0
```

bitnami/kube-prometheus/values.yaml
```diff
@@ -3308,6 +3308,10 @@ rbac:
   create: true
   pspEnabled: true
 
+## @param extraRbac Extra Rbac for the Prometheus clusterrole
+##
+extraRbac: []
```
juan131: We do support this in other charts via the `rbac.rules` parameter, let's stick to the format we used on them for consistency:

```yaml
  ## @param rbac.rules Custom RBAC rules to set on Prometheus ClusterRole
  ## e.g:
  ## rules:
  ##   - apiGroups:
  ##       - ""
  ##     resources:
  ##       - pods
  ##     verbs:
  ##       - get
  ##       - list
  ##
  rules: []
```

juan131: That said and, given that we have 2 ClusterRole(s) (one for prometheus and another for prometheus-operator):

- https://github.com/bitnami/charts/blob/main/bitnami/kube-prometheus/templates/prometheus/clusterrole.yaml
- https://github.com/bitnami/charts/blob/main/bitnami/kube-prometheus/templates/prometheus-operator/clusterrole.yaml

We might need to split this parameter in two so both can be customize separately:

```suggestion
  ## @param rbac.rules.operator Custom RBAC rules to set on Prometheus Operator ClusterRole
  ## @param rbac.rules.prometheus Custom RBAC rules to set on Prometheus ClusterRole
  ## e.g:
  ## rules:
  ##   operator: []
  ##   prometheus:
  ##     - apiGroups:
  ##         - ""
  ##       resources:
  ##         - pods
  ##       verbs:
  ##         - get
  ##         - list
  ##
  rules:
    operator: []
    prometheus: []
```

juan131 (changes requested):
Thanks so much for your contribution! Please check my comments when you have a chance.

bitnami/kube-prometheus/values.yaml
Poil: Thanks, I'm on holidays for two weeks, please keep it opens I will fix that when I come back
Regards

juan131 (approved):
LGTM! Thanks for the contribution

Issue #29789, opened 2024-10-06
[bitnami/apisix] failed on restart of container


maipal-c:
### Name and Version

bitnami/apisix:3.5.0

### What architecture are you using?

arm64

### What steps will reproduce the bug?

1 install the chart using the values below
2. simulate restart (keep memory requirement low so you will see OOM killed and the container will get recreated within the pod.)
3.  when the container gets recreated (not in the new pod) it fails to start with the following error
```
"2024/10/06 13:26:22 [emerg] 1#1: bind() to unix:/usr/local/apisix/logs/worker_events.sock failed (98: Address already in use)"
```

### Are you using any custom parameters or values?

default values with coraza wasm plugin enabled on the control plane & data plane 
```yaml
dataPlane:
  extraConfig:
        wasm:
           plugins:
              - name: coraza-filter
                priority: 7999
                file: /tmp/wasm-plugins/coraza-proxy-wasm.wasm
     extraVolumes:
        - name: wasm-plugins
          emptyDir: {}
     extraVolumeMounts:
        - name: wasm-plugins
          mountPath: "/tmp/wasm-plugins"
     initContainers:
        - name: attach-wasm-plugins
          image: busybox
          securityContext:
             capabilities:
                drop:
                   - ALL
             privileged: false
             runAsUser: 1001
             runAsGroup: 1001
             runAsNonRoot: true
             readOnlyRootFilesystem: true
             allowPrivilegeEscalation: false
             seccompProfile:
                type: RuntimeDefault
          volumeMounts:
             - name: wasm-plugins
               mountPath: "/tmp/wasm-plugins"
          command:
             - "sh"
             - "-c"
             - |
                cd /tmp/wasm-plugins ;
                wget https://github.com/corazawaf/coraza-proxy-wasm/releases/download/0.5.0/coraza-proxy-wasm-0.5.0.zip ;
                unzip coraza-proxy-wasm-0.5.0.zip ;
                rm coraza-proxy-wasm-0.5.0.zip

controlPlane: <same_as_above>
```

### What is the expected behavior?

on recreation of container, it should start normally.

### What do you see instead?

"2024/10/06 13:26:22 [emerg] 1#1: bind() to unix:/usr/local/apisix/logs/worker_events.sock failed (98: Address already in use)"

### Additional information

same issue on apisix github repo, [here](https://github.com/apache/apisix/issues/11376)

possible fix - 

In Apisix's official helm chart they have a lifecycle hook -
```yaml
lifecycle:
[...]

PR #34987, opened 2025-07-10
[bitnami/apisix] fix: Apisix clean uds

### LINKED ISSUE #34385: [bitnami/apisix] Container can't be restarted correctly after crash
### Name and Version

bitnami/apisix:5.0.2

amd64

This is a follow-up to issue #29789.
Currently, after a crash, the only working solution is manual container recreation.

Perhaps the chart implementation should remove the socket file on startup?
I attempted the following configuration:

```yaml
      lifecycle:
        postStart:
          exec:
            command:
            - /bin/sh
            - -c
            - |
              sleep 5;
              rm /usr/local/apisix/logs/worker_events.sock
```

However, this approach didn't work for me. If this cleanup is necessary, maybe it should be incorporated directly into the chart?

Current error after container restart:

```
2025/06/13 03:57:59 [emerg] 1#1: bind() to unix:/usr/local/apisix/logs/worker_events.sock failed (98: Address already in use)
nginx: [emerg] bind() to unix:/usr/local/apisix/logs/worker_events.sock failed (98: Address already in use)
2025/06/13 03:57:59 [emerg] 1#1: bind() to unix:/usr/local/apisix/logs/worker_events.sock failed (98: Address already in use)
nginx: [emerg] bind() to unix:/usr/local/apisix/logs/worker_events.sock failed (98: Address already in use)
2025/06/13 03:57:59 [emerg] 1#1: bind() to unix:/usr/local/apisix/logs/worker_events.sock failed (98: Address already in use)
nginx: [emerg] bind() to unix:/usr/local/apisix/logs/worker_events.sock failed (98: Address already in use)
2025/06/13 03:57:59 [emerg] 1#1: bind() to unix:/usr/local/apisix/logs/worker_events.sock failed (98: Address already in use)
nginx: [emerg] bind() to unix:/usr/local/apisix/logs/worker_events.sock failed (98: Address already in use)
2025/06/13 03:57:59 [emerg] 1#1: bind() to unix:/usr/local/apisix/logs/worker_events.sock failed (98: Address already in use)
nginx: [emerg] bind() to unix:/usr/local/apisix/logs/worker_events.sock failed (98: Address already in use)
2025/06/13 03:57:59 [emerg] 1#1: still could not bind()
nginx: [emerg] still could not bind()
```

Crash is probably an upstream problem, but container cleanup may be a problem with the chart.

```yaml
project: default
source:
  repoURL: registry-1.docker.io/bitnamicharts
  targetRevision: 5.0.2
  helm:
    parameters:
      - name: dataPlane.ingress.enabled
        value: 'true'
      - name: etcd.replicaCount
        value: '1'
    valuesObject:
      controlPlane:
        enabled: true
        lifecycleHooks:
          postStart:
            exec:
              command:
              - /bin/sh
              - -c
              - |
                sleep 5;
                rm /usr/local/apisix/logs/worker_events.sock`
      dataPlane:
        config:
          apisix:
            enable_ipv6: true
        service:
          externalIPs:
            - 192.168.9.252
            - 3.3.3.3
            - 1.1.1.1
  chart: apisix
destination:
  server: https://kubernetes.default.svc
  namespace: ingress-apisix
```

### What do you see instead?

The container after the crash should start

SpectatorNan:
### Description of the change

add auto clean uds file  before start

### Benefits

fix apsixi start failed err: handle nginx: [emerg] bind() to unix:/usr/local/apisix/logs/worker_events.sock failed (98: Address already in use) error

### Possible drawbacks

- N/A

### Applicable issues

- fixes #34385 , #29789
- support bitnami/containers/issues/82469

### Additional information

- N/A

### Checklist


SpectatorNan:
image
handle this err

@migruiz4  when to review?

bitnami/apisix/templates/_helpers.tpl
```diff
@@ -336,6 +336,16 @@ Init container definition for waiting for the database to be ready
       ln -sf /opt/bitnami/apisix/deps /usr/local/apisix
       ln -sf /opt/bitnami/apisix/openresty/luajit/share/lua/*/apisix /usr/local/apisix
       mkdir -p /usr/local/apisix/logs
+      # Clean up UDS file before starting APISIX
```
migruiz4: I'm sorry I didn't notice that you had already included this in the init container.

Then replacing the container command with this shouldn't be needed.

SpectatorNan: I'm very sorry, but I have tried many times to reproduce this issue perfectly, and I haven't been able to.

But from the log output, I can see that the prepare init container only executes once (when it is created for the first time or deleted and recreated). The behavior I observed in the logs, ```Back-off restarting failed container apisix in pod```, suggests that it skips the initContainers and directly executes the main container.

Therefore, I believe that checking before the command execution starts is a more reliable approach. The removal operation in the initContainer may be unnecessary because, based on the logs from my local tests, this UDS file never exists.

If it's necessary to remove redundant code snippets, I think it would be better to remove the check code snippet from the initContainer. If you think it should be removed, I will proceed with removing the check code snippet from the initContainer.

migruiz4: That sounds okay. I agree to keep the removal in the pod command and remove it from the init container.

From documentation, I would say that the init container should be triggered every time the container is restarted, but can't ensure if that would cover every corner case so keeping it in the command sounds good to me.

Could you please remove the section from the init container?

Issue #35313, opened 2025-07-28
Introduce context level includes in nginx


cono:
### Name and Version

bitnami/nginx 21.0.8

### What is the problem this feature will solve?

We don't have ability to load modules without re-definition of the whole nginx.conf.

This will definitely resolve issues that already happened a couple of times: https://github.com/bitnami/charts/issues/4857 https://github.com/bitnami/charts/issues/11851 https://github.com/bitnami/charts/issues/4121
And prevent future creation of such issues.

### What is the feature you are proposing to solve the problem?

In nginx documentation for each directive they specify what context it should be placed in, e.g.:
```
Syntax:	load_module file;
Default:	—
Context:	main
This directive appeared in version 1.9.11.
```
https://nginx.org/en/docs/ngx_core_module.html#load_module

I propose to introduce includes for each context, instead of server_block, call them by the name of the context. So I can provide particular snippet that is going to be included in existing config of main/events/http/... or any other contexts.

This change definitely requires changes to docker image default nginf.conf config.

carrodher:
Thank you for bringing this issue to our attention. We appreciate your involvement! If you're interested in contributing a solution, we welcome you to create a pull request. The Bitnami team is excited to review your submission and offer feedback. You can find the contributing guidelines [here](https://github.com/bitnami/charts/blob/main/CONTRIBUTING.md).

Your contribution will greatly benefit the community. Feel free to reach out if you have any questions or need assistance.

cono:
@carrodher not trying to be rude. But what is usual contribution approval time? I'm asking as I need this feature in order to accomplish task at work, so I wonder how much this gonna take. Or should I just go via my own helm chart implementation.

first part of it: https://github.com/bitnami/containers/pull/83786

dgomezleon:
Hi @cono ,

I will check it today.

PR #35347, opened 2025-07-30
[bitnami/nginx] Add support for context-based configuration includes

cono:
This is continuation of https://github.com/bitnami/containers/pull/83786 and attempts to fix: https://github.com/bitnami/charts/issues/35313

> I propose to introduce includes for each context, instead of server_block, call them by the name of the context. So I can provide particular snippet that is going to be included in existing config of main/events/http/... or any other contexts.

What has been done as part of this PR:

Add contextIncludes parameters for main, events, and http contexts
Add support for multiple external ConfigMaps per context
Enable users to load modules, customize events, and add http directives

How I tested it locally.
1. Created following file:
```
# Test values for context includes functionality
contextIncludes:
  main: |
    # Load WebDAV module
    load_module /opt/bitnami/nginx/modules/ngx_http_dav_module.so;

    # Set worker processes
    worker_processes auto;

  events: |
    # Increase worker connections
    worker_connections 2048;

    # Use epoll for better performance
    use epoll;

  http: |
    # Enable gzip compression
    gzip on;
    gzip_vary on;
    gzip_types text/plain application/json application/javascript text/css;

    # Security headers
    add_header X-Frame-Options DENY;
    add_header X-Content-Type-Options nosniff;

# Test external ConfigMaps as well
existingContextMainConfigmaps:
  - "nginx-custom-modules"
  - "nginx-main-directives"

existingContextHttpConfigmaps:
  - "nginx-security-config"
  - "nginx-performance-config"
```

Ran following command: `helm template test-nginx . -f test-values.yaml | less`
Checked by my eyes, so issue still can be present, but resulting config looks sane for me.


bitnami/nginx/values.yaml
```diff
@@ -625,6 +625,38 @@ existingServerBlockConfigmap: ""
 ## NOTE: This will override streamServerBlock
 ##
 existingStreamServerBlockConfigmap: ""
+## @param contextIncludes.main Custom configuration to be added to the NGINX main context
```
dgomezleon: This param comment is repeated

cono:
Rebased and fixed description of the param

PR #35536, opened 2025-08-07
[bitnami/redis] Add support to redis master service with useHostnames false

### LINKED ISSUE #29217: [bitnami/redis] Automatic labeling of master pod fails if useHostnames is set to false
### Name and Version

bitnami/redis 20.0.5

amd64

- run chart with useHostnames=false (due to DNS issues causing tilt mode)
- enable sentinel
- enable masterService

```yaml
useHostnames: false
sentinel:
  enabled: true
  masterService:
    enabled: true
```

master pod is labeled with `isMaster="true"`

### What do you see instead?

no such label on master

### Additional information

kubectl uses the content of `/etc/shared/current` for the pod selector in `update-master-label.sh`.
The content of `/etc/shared/current` is written by sentinel from `${REDIS_MASTER_HOST/.*}` in `start-sentinel.sh`.
when using useHostnames=false REDIS_MASTER_HOST contains not `redis-node-0.redis.svc.cluster.local`, which would return `redis-node-0` as `current`, but instead the IP address e.g. `172.22.140.140` which returns `172` as `current`.
thus kubectl matches no pod and updates no labels. 
Unfortunately, the output of kubectl is also not logged - only the echo statement with `new master elected, updating label(s)...`.
Patching the script configmap with `echo "$(< "/etc/shared/current")"` helped to debug this as it returned `172` instead of `redis-node-0`.

This also applies to the delete patch for the label `app.kubernetes.io/role` which will be fixed in 20.0.6 (annotation -> label).

To fix this either sentinel writes a non-resolved value for `REDIS_MASTER_HOST` into `/etc/shared/current`
or the kubectl script looks up the pod by IP if `current` is numerical. such as:
`kubectl get -n <namespace>  --output json  pods | jq '.items[] | select(.status.podIP=="<IP>")' | jq .metadata.name`
though this would likely require further RBAC permissions.

fmendez89:
### Description of the change

This PR improves the way the current Redis master pod is labeled by Sentinel by introducing synchronization and avoiding race conditions when multiple pods detect a failover. It adds the following changes:

- A shared coordination mechanism using flock to avoid race conditions when labeling pods.
- Skips relabeling if the master hasn't changed, preventing unnecessary updates.
- Adds support for resolving the master pod name by IP when useHostnames is disabled.
- Improves the label management logic by adding isMaster=true to the current master and removing app.kubernetes.io/role.

### Benefits

- Avoids concurrent relabeling issues across Sentinel instances.
- Reduces unnecessary pod label updates.
- Ensures correct label assignment in clusters with or without useHostnames.
- Supports both IP and hostname-based master detection.
- Aligns the master labeling logic with the actual Sentinel-detected master.

### Possible drawbacks

- Requires that the container has permission to use file locking (should be fine on most Linux distros).
- Relies on a shared writable volume mounted at /etc/shared across pods.

### Applicable issues

Fixes #29217

### Additional information

This feature is particularly useful when using external services that rely on pod labels (e.g., a Service selector for the master role). It ensures the label is always up-to-date based on Sentinel decisions.

### Checklist

- [X] Chart version bumped in Chart.yaml according to [semver](http://semver.org/). This is not necessary when the changes only affect README.md files.
- [X] Variables are documented in the values.yaml and added to the README.md using [readme-generator-for-helm](https://github.com/bitnami/readme-generator-for-helm)
- [X] Title of the pull request follows this pattern [bitnami/redis] Add support to redis master service with useHostnames false


bitnami/redis/templates/scripts-configmap.yaml
```diff
@@ -1000,25 +1009,79 @@ metadata:
...
+
+            if [ -f "/etc/shared/previous" ]; then
+                previous_master="$(< "/etc/shared/previous")"
+
+                if [ "$current_master" == "$previous_master" ]; then
+                    echo "Master has not changed ($current_master), skipping label update and cleaning state"
+                    rm -f /etc/shared/current /etc/shared/previous
+                    continue
+                fi
+            fi
```
juan131: Nice improvement!

bitnami/redis/templates/scripts-configmap.yaml
```diff
@@ -1000,25 +1009,79 @@ metadata:
...
   update-master-label.sh: |
     #!/bin/bash
+    set -euo pipefail
+
+    LOCK_FILE="/etc/shared/update-master-label.lock"
+
+    {{- if eq .Values.useHostnames false }}
+    resolve_pod() {
+        local pod_ip="$1"
+        kubectl get pods -o custom-columns=":metadata.name" --field-selector=status.podIP="$pod_ip" --no-headers 2>/dev/null | head -n1
```
juan131: I think you need to update RBAC for this:

- https://github.com/bitnami/charts/blob/main/bitnami/redis/templates/role.yaml#L29

bitnami/redis/templates/scripts-configmap.yaml
```diff
@@ -1000,25 +1009,79 @@ metadata:
...
-                kubectl label pod --field-selector metadata.name="$(< "/etc/shared/previous")" isMaster="false" --overwrite
+                previous_master="$(< "/etc/shared/previous")"
+                echo "Previous master: $previous_master"
+
+                {{- if eq .Values.useHostnames false }}
+                previous_pod="$(resolve_pod "$previous_master")"
+                echo "Previous master IP resolved to pod $previous_pod"
+                {{- else }}
+                previous_pod="$previous_master"
+                {{- end }}
```
juan131: The resulting logic if `.Values.useHostnames` is `true` is:
```bash
                previous_master="$(< "/etc/shared/previous")"
                echo "Previous master: $previous_master"

                previous_pod="$previous_master"
```
Isn't `previous_pod` redundant in this case?

bitnami/redis/templates/scripts-configmap.yaml
```diff
@@ -1000,25 +1009,79 @@ metadata:
...
+    LOCK_FILE="/etc/shared/update-master-label.lock"
+
+    {{- if eq .Values.useHostnames false }}
+    resolve_pod() {
+        local pod_ip="$1"
+        kubectl get pods -o custom-columns=":metadata.name" --field-selector=status.podIP="$pod_ip" --no-headers 2>/dev/null | head -n1
+    }
+    {{- end }}
+
+    flock -n "$LOCK_FILE" bash <<'EOF'
```
juan131: We should add a safeguard mechanism and skip this is `flock` isn't available in the system (e.g. running `which flock` or sth like that)

juan131 (changes requested):
Thanks so much for this great PR! Absolutely brilliant, please check my comments.

bitnami/redis/templates/scripts-configmap.yaml
fmendez89: No it won't be necessary as it is just listing the pods, and the rbac allows this.

It's the same as the labeling command used before the PR.

`kubectl label pod --field-selector=metadata.name="$pod_name" isMaster="true" --overwrite`

This command does not need "get" as is just listing pods with the field selector. 

*IF* it had used 
`kubectl label pod <pod_name> isMaster="true" --overwrite`
Then the get permission would have been needed.

I believe this is something kubernetes/kubectl needs to refine, but anyways it should work fine.

bitnami/redis/templates/scripts-configmap.yaml
fmendez89: Ok, that would be nice improvement. I'll add it.

bitnami/redis/templates/scripts-configmap.yaml
fmendez89: Yes, i thought it would be nice to state that from that point the variable is the pod name and not the possibility to be the IP. It's almost the same structure as the current_master few lines above.

Maybe this way is better to understand? (Exact copy from the current_master)

```shell
previous_master="$(< "/etc/shared/previous")"
echo "Previous master: $previous_master"

previous_pod="$previous_master"
{{- if eq .Values.useHostnames false }}
previous_pod="$(resolve_pod "$previous_master")"
echo "Previous master IP $previous_master resolved to pod $previous_pod"
{{- end }}
```

Happy to hear your advices.

bitnami/redis/templates/scripts-configmap.yaml
juan131: Thanks for the detailed explanation!

bitnami/redis/templates/scripts-configmap.yaml
juan131: What about this?
```bash
previous_pod="$(< "/etc/shared/previous")"
echo "Previous master: $previous_pod"

{{- if eq .Values.useHostnames false }}
previous_pod="$(resolve_pod "$previous_pod")"
echo "Previous master IP resolved to pod $previous_pod"
{{- end }}
```

fmendez89: And what if we choose to name the variable more explicitly ? could it make sense then?

previous_master could be an IP or pod name, but previous_master_pod_name that only could be the pod name. 

```shell
previous_master="$(< "/etc/shared/previous")"
echo "Previous master: $previous_master"

{{- if eq .Values.useHostnames false }}
previous_master_pod_name="$(resolve_pod "$previous_master")"
echo "Previous master IP $previous_master resolved to pod $previous_pod"
{{- else }}
previous_master_pod_name="$previous_master"
{{- end }}
```

fmendez89: And what if we choose to name the variable more explicitly ? could it make sense then?

previous_master could be an IP or pod name, but previous_master_pod_name that only could be the pod name.

```shell
{{- if eq .Values.useHostnames false }}
previous_master="$(< "/etc/shared/previous")"
previous_master_pod_name="$(resolve_pod "$previous_master")"
echo "IP $previous_master resolved to previous master: $previous_master_pod_name"
{{- else }}
previous_master_pod_name="$(< "/etc/shared/previous")"
echo "Previous master: $previous_master_pod_name"
{{- end }}
```

juan131: Great

fmendez89: I wanted to share a new approach which has come to my mind recently. To me it seems much simpler. I haven't tried it yet, but should work. The key here is to change field-selector key if using hostnames false or true.

```shell
current_master=$(< /etc/shared/current)
[ -f /etc/shared/previous ] && previous_master=$(< /etc/shared/previous) || previous_master=""

{{- if .Values.useHostnames }}
selector_key="metadata.name"
{{- else }}
selector_key="status.podIP"
{{- end }}

if [ -n "$current_master" ]; then
    echo "Labeling new master $current_master"
    kubectl label pod --field-selector="$selector_key=$current_master" isMaster="true" --overwrite || echo "Failed to label master"
    kubectl label pod --field-selector="$selector_key=$current_master" app.kubernetes.io/role- || echo "Failed to remove role label"
fi

if [ -n "$previous_master" ]; then
    echo "Previous master: $previous_master"
    echo "Removing master label from previous master $previous_master"
    kubectl label pod --field-selector="$selector_key=$previous_master" isMaster="false" --overwrite || echo "Failed to remove label"
fi
```

This way it doesn't need to resolve the ip, it uses the field-selector to label the pod.

juan131: Nice! Please implement that solution 👏🏼

bitnami/redis/templates/scripts-configmap.yaml
fmendez89: I've added a check so you can see the approach, but let me test it before any approve.

bitnami/redis/templates/scripts-configmap.yaml
```diff
@@ -1000,25 +1009,66 @@ metadata:
...
+
+    main_logic() {
+        while true; do
+            while [ ! -f "/etc/shared/current" ] && [ ! -f "/etc/shared/terminate" ]; do
+                sleep 1
+            done
+
+            if [ -f "/etc/shared/current" ]; then
+                current_master=$(< /etc/shared/current)
+                [ -f /etc/shared/previous ] && previous_master=$(< /etc/shared/previous) || previous_master=""
```
juan131: Please split this in 2 lines:
```suggestion
                previous_master=""
                [ -f /etc/shared/previous ] && previous_master=$(< /etc/shared/previous)
```

bitnami/redis/templates/scripts-configmap.yaml
```diff
@@ -1000,25 +1009,66 @@ metadata:
...
+
+                echo "Current master: $current_master"
+
+                if [ -n "$previous_master" ] && [ "$current_master" = "$previous_master" ]; then
```
juan131: Is the 1st condition redundant?
```suggestion
                if [ "$current_master" = "$previous_master" ]; then
```

bitnami/redis/templates/scripts-configmap.yaml
```diff
@@ -1000,25 +1009,66 @@ metadata:
...
+                    echo "Master ($current_master = $previous_master) has not changed, skipping label update and cleaning state"
```
juan131: In the previous `echo` you already displayed the variable value:
```suggestion
                    echo "Master has not changed, skipping label update and cleaning state"

```

fmendez89: Ok, removing it

bitnami/redis/templates/scripts-configmap.yaml
fmendez89: sure, it's better that way

bitnami/redis/templates/scripts-configmap.yaml
fmendez89: yes, good catch

fmendez89:
Hi @juan131, how can I reopen de PR to push the changes?

carrodher:
Thanks for your contribution! Could you please bump the chart version in the _Chart.yaml_? This is necessary to test the changes and cut a new release. Probably it was missing after the latest rebase.

carrodher:
The chart version present in the Chart.yaml should be bumped

fmendez89:
> Any updates?

Sorry @ZebcoWeb, I haven't had time the last weeks. I wanted to test the scripts on our staging environment before merging. I don't want to mess with the deployments of the entire world, mine's it's enough. Although, if you want to try them or have already, please let me know. 
I'm looking forward to try them next week. So you can expect this PR to be merged soon if everything is working fine.

fmendez89:
> Hi @fmendez89, I had to use your own branch directly because I needed this update, and I can say that after 2 weeks of testing, it works well. And I'm still waiting for it to be merged into the main repository.

Wow, nice! thanks for the feedback. I'll try to push the changes this weekend then.

fmendez89:
Bumped chart version, I believe that's it, let me know if anything else is needed.

bitnami/redis/templates/scripts-configmap.yaml
```diff
@@ -815,12 +819,17 @@ data:
...
+      #!/bin/bash
+      # https://download.redis.io/redis-stable/sentinel.conf
+
+      {{- if .Values.useHostnames }}
+      echo "${6/.*}" > /etc/shared/current
+      echo "${4/.*}" > /etc/shared/previous
+      {{- else }}
+      echo "$6" > /etc/shared/current
+      echo "$4" > /etc/shared/previous
+      {{- end }}
```
juan131: Please correct me if I'm wrong but this change in the indentation is unexpected:
```suggestion
  push-master-label.sh: |
    #!/bin/bash
    # https://download.redis.io/redis-stable/sentinel.conf
    {{- if .Values.useHostnames }}
    echo "${6/.*}" > /etc/shared/current
    echo "${4/.*}" > /etc/shared/previous
    {{- else }}
    echo "$6" > /etc/shared/current
    echo "$4" > /etc/shared/previous
    {{- end }}
```

fmendez89: Hi, @juan131 sorry for the delay. Could you open the MR again? I'm pushing these requested changes now.

fmendez89: Just pushed the changes, please you check them now

PR #35667, opened 2025-08-07
[bitnami/redis] Fix Redis sentinel initialization with single replica

Nepoxx:
## Description

The `get_service_ip` function was called unconditionally during master initialization but only defined when external access is enabled. This caused single-replica Redis deployments to fail during startup when using the default internal-only configuration.

The fix makes the initialization logic consistent with the comparison logic by using `get_full_hostname` when external access is disabled.

### Description of the change

This change fixes a conditional logic mismatch in the Redis chart's startup script template. The `get_service_ip()` function was being called unconditionally during master initialization (line 158) but was only defined when `sentinel.externalAccess.enabled` is true. This caused Redis containers to fail during startup in single-replica deployments using the default configuration where external access is disabled.

The fix adds the same conditional logic used in the comparison section to the initialization section, ensuring consistency between initialization and comparison phases.

### Benefits

- Fixes Redis sentinel mode in single-replica deployments with default configuration
- Maintains compatibility with existing multi-replica deployments  
- Ensures consistent conditional logic throughout the startup script
- No breaking changes to existing functionality

### Possible drawbacks

None identified. The change only affects the problematic code path and maintains all existing behavior for configurations where external access is enabled.

### Applicable issues

This addresses startup failures in single-replica Redis deployments where sentinel is unable to promote the replica to master due to undefined function calls.

### Additional information

The issue occurs specifically when:
- `replica.replicaCount: 1` (single replica)  
- `sentinel.externalAccess.enabled: false` (default configuration)

The fix ensures the startup script works correctly in both external access and internal-only configurations.

### Checklist

### Root Cause
The bug existed in `/bitnami/redis/templates/scripts-configmap.yaml` at line 158:
```bash
master_in_persisted_conf="$(get_service_ip "$SVC_NAME")"
```

This line called `get_service_ip()` unconditionally, but the function was only defined when:
```yaml
{{- if and .Values.sentinel.externalAccess.enabled .Values.sentinel.externalAccess.service.loadBalancerIP }}
```

### Fix Applied
Changed the unconditional call to use the same conditional logic as the comparison section:
```bash
{{- if and .Values.sentinel.externalAccess.enabled .Values.sentinel.externalAccess.service.loadBalancerIP }}
master_in_persisted_conf="$(get_service_ip "$SVC_NAME")"
{{- else }}
master_in_persisted_conf="$(get_full_hostname "$HOSTNAME")"
{{- end }}
```

### Tested Scenarios
- ✅ Single replica with internal access (default)
- ✅ Multiple replicas with internal access
- ✅ Single/multiple replicas with external access

### DIFF
--- a/bitnami/redis/CHANGELOG.md
+++ b/bitnami/redis/CHANGELOG.md
@@ -1,12 +1,12 @@
 # Changelog
 
-## 22.0.7 (2025-08-25)
+## 22.0.7 (2025-08-27)
 
 * [bitnami/redis] Fix Redis sentinel initialization with single replica ([#35667](https://github.com/bitnami/charts/pull/35667))
 
 ## <small>22.0.6 (2025-08-26)</small>
 
-* [bitnami/charts/issues/35435] Fix Redis configuration when using sentinel and 1 replica ([#36170](https://github.com/bitnami/charts/pull/36170))
+* [bitnami/charts/issues/35435] Fix Redis configuration when using sentinel and 1 replica (#36170) ([7bc9738](https://github.com/bitnami/charts/commit/7bc9738651c99dc1a9d960da83f4502d9c2d352e)), closes [#36170](https://github.com/bitnami/charts/issues/36170)
 
 ## <small>22.0.5 (2025-08-23)</small>
 


gongomgra:
Hi @Nepoxx 

Thanks for reporting this issue and for contributing with a solution! Can you please rebase your branch with updated `main` branch and fix the conflicts so we can merge your changes?

PR #35748, opened 2025-08-11
[bitnami/rabbitmq] Migrate deprecated queue_leader_locator configuration

bpfoster:
### Description of the change

When running the Bitnami RabbitMQ helm chart, I observe the following warning logged:
```
[warning] <0.985.0> Deprecated features: `queue_master_locator`: queue-master-locator is deprecated. queue-leader-locator should be used instead (allowed values are 'client-local' and 'balanced')
```

This feature looks to be deprecated beginning in RabbitMQ 4.0.

According to [the docs](https://www.rabbitmq.com/docs/clustering#replica-placement), the `queue_leader_locator` value of `min-masters` is mapped to `queue_leader_locator = balanced`.

### Benefits

This will update the configuration generated by the chart to use the new, corresponding, configuration for this.

### Possible drawbacks

None noted

### Applicable issues

- fixes #

### Checklist

### DIFF
--- a/bitnami/rabbitmq/CHANGELOG.md
+++ b/bitnami/rabbitmq/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## 16.0.14 (2025-08-12)
+## 16.0.14 (2025-08-13)
 
 * [bitnami/rabbitmq] Migrate deprecated queue_leader_locator configuration ([#35748](https://github.com/bitnami/charts/pull/35748))
 


bitnami/rabbitmq/values.yaml
```diff
@@ -448,7 +448,7 @@ configuration: |-
   load_definitions = {{ .Values.loadDefinition.file }}
   {{- end }}
   # queue master locator
-  queue_master_locator = {{ .Values.queue_master_locator }}
+  queue_leader_locator = {{ .Values.queue_leader_locator }}
```
gongomgra: Thanks for your contribution! Can you add backward compatibility to the old parameter using the `coalesce` function?

bpfoster: Thanks @gongomgra this is a good point.
I wasn't sure we could use a simple `coalesce` here because the values can be different.  Refactored out to a helper function - please let me know if this works for you, or if you'd like another approach.

gongomgra: Thank you for adding the helper! I'm marking this conversation as resolved.

Issue #35327, opened 2025-07-29
[bitnami/kafka] NetworkPolicy should be created when match object is enabled


michal-marszalek-h2oai:
### Name and Version

bitnami/kafka 32.3.8

### What steps will reproduce the bug?

**Description**
Network policies for the broker and controller should be created only when those components are deployed. Currently, the Helm chart does not support disabling the broker's NetworkPolicy independently when the broker itself is disabled.

**Current Behavior**
All NetworkPolicies are toggled globally using the value:
```yaml
.Values.networkPolicy.enabled
```

This does not account for scenarios where the broker is disabled, but the controller is still in use.

**Expected Behavior**
NetworkPolicies should be conditionally created based on the enabled state of their respective components.

Proposed Solutions
Add component-specific flags, similar to how PodDisruptionBudgets (PDBs) are handled:
```yaml
.Values.broker.networkPolicy.create

.Values.controller.networkPolicy.create
```
Conditionally create NetworkPolicies based on the replica count, as is done in the broker's StatefulSet:

```yaml
{{- if gt $replicaCount 0 }}
  # create network policy
{{- end }}
````

**Benefits**
Improved flexibility for deployments that do not include all components.
Consistent behavior with other configurable resources like PDBs and StatefulSets.
Avoids creating unused or unnecessary Kubernetes resources.

### What is the expected behavior?

NetworkPolicies should be conditionally created based on the enabled state of their respective components.

Proposed Solutions
Add component-specific flags, similar to how PodDisruptionBudgets (PDBs) are handled:
```yaml
.Values.broker.networkPolicy.create

.Values.controller.networkPolicy.create
```
Conditionally create NetworkPolicies based on the replica count, as is done in the broker's StatefulSet:

```yaml
{{- if gt $replicaCount 0 }}
  # create network policy
{{- end }}
````

### What do you see instead?

When the broker is disabled unused NetworkPolicy is created that cannot be disabled without disabling NetworkPolicy for the controller.

javsalgar:
Hi!

Thank you so much for reporting! I will forward it to the team but as it is not a critical feature we cannot guarantee an ETA. If you want to speed up the process, you can submit a PR and the team will take a look.

dashashutosh80:
@javsalgar / @michal-marszalek-h2oai I have raised a PR with possible fix for this issue.

carrodher:
[...]

PR #35771, opened 2025-08-13
[bitnami/kafka] feat: conditional implementation of network policy

dashashutosh80:
### Description of the change

The fix addresses the core issue where NetworkPolicies were created regardless of whether the corresponding components were actually running, as mentioned in https://github.com/bitnami/charts/issues/35327:
**Before:** NetworkPolicies created when `networkPolicy.enabled: true` (even with `broker.replicaCount: 0`)
**After:** NetworkPolicies only created when:
`networkPolicy.enabled: true` AND
Either `replicaCount > 0` OR `autoscaling.hpa.enabled: true`

### Benefits

- Consistency: Now matches the exact same conditional logic used by StatefulSets
- Resource Efficiency: Eliminates unused NetworkPolicy objects
- Logical Correctness: NetworkPolicies only exist when pods exist to be governed by them
- Backward Compatibility: Existing configurations continue to work as expected
- Future-Proof: Accounts for HPA scenarios where pods might be scaled from 0

### Additional information
Even though validation is in place for controller-eligible components to be enabled/running with message `kafka: Missing controller-eligible nodes`,  conditional has been added to the NetworkPolicy for controller as well to maintain consistency across templates and not depend on any external validation.

### Checklist

### DIFF
--- a/bitnami/kafka/CHANGELOG.md
+++ b/bitnami/kafka/CHANGELOG.md
@@ -1,8 +1,12 @@
 # Changelog
 
-## 32.3.14 (2025-08-07)
+## 32.4.0 (2025-08-13)
 
-* [bitnami/kafka] :zap: :arrow_up: Update dependency references ([#35615](https://github.com/bitnami/charts/pull/35615))
+* [bitnami/kafka] feat: conditional implementation of network policy ([#35771](https://github.com/bitnami/charts/pull/35771))
+
+## <small>32.3.14 (2025-08-07)</small>
+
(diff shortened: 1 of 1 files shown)


dgomezleon (commented):
The changes LGTM. However, considering it a bugfix, I think the version should be bumped to `32.3.15`. Does it make sense?

dashashutosh80:
@dgomezleon My opinion is that this is an improvement rather than a bug fix as even now the charts installation, upgrades and components' functionalities are not affected. An additional NP is created for broker even when broker pods are not present, effectively leaving the NP unused. For this reason, I updated the minor version of chart and not patch version.

dgomezleon (approved):

